PR #555, opened 2016-11-21
[C] detach Behaviors and Triggers on VE finalization

StephaneDelcroix:
### Description of Change ###

Detach Triggers and Behaviors on VE finalizer, because they're not IDisposable...

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=44074

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/BindableObject.cs
+++ b/Xamarin.Forms.Core/BindableObject.cs
@@ -163,6 +163,14 @@ namespace Xamarin.Forms
 			return bpcontext != null && bpcontext.Binding != null;
 		}
 
+		internal bool GetIsDefault(BindableProperty targetProperty)
+		{
+			if (targetProperty == null)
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Core/BindableObject.cs
```diff
@@ -163,6 +163,14 @@ internal bool GetIsBound(BindableProperty targetProperty)
 			return bpcontext != null && bpcontext.Binding != null;
 		}
 
+		internal bool GetIsDefault(BindableProperty targetProperty)
```
StephaneDelcroix: this is to avoid calling the defaultValueCreator is the BP is not used

adrianknight89:
+1 if this is a better fix.

samhouts:
See #701 for related crash.

PR #558, opened 2016-11-24
Fix weak references when getting targets

adrianknight89:
### Description of Change ###

When using `WeakReference`, one should not rely on `IsAlive` when it returns true as it could mean one of two things and we don't know which one. For more info, see bug description as well as the links in it.

It appears that XF is using a mix of `WeakReference` (.NET 1.1) and `WeakReference<T>` (4.5). For future references, `WeakReference` should probably be banned from being used.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=47203

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/MessagingCenter.cs
+++ b/Xamarin.Forms.Core/MessagingCenter.cs
@@ -91,7 +91,7 @@ namespace Xamarin.Forms
 			List<Tuple<WeakReference, Action<object, object>>> actionsCopy = actions.ToList();
 			foreach (Tuple<WeakReference, Action<object, object>> action in actionsCopy)
 			{
-				if (action.Item1.IsAlive && actions.Contains(action))
+				if (action.Item1.Target != null && actions.Contains(action))
 					action.Item2(sender, args);
 			}
 		}
--- a/Xamarin.Forms.Platform.WP8/NavigationMenuRenderer.cs
+++ b/Xamarin.Forms.Platform.WP8/NavigationMenuRenderer.cs
@@ -72,8 +72,9 @@ namespace Xamarin.Forms.Platform.WinPhone
 				var weakRef = new WeakReference(hubTile);
 				SizeChanged += (sender, args) =>
 				{
-					if (weakRef.IsAlive)
-						((HubTile)weakRef.Target).Size = GetSize();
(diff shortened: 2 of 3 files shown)


Xamarin.Forms.Platform.WP8/NavigationMenuRenderer.cs
```diff
@@ -72,8 +72,9 @@ void UpdateItems(System.Windows.Controls.Grid grid)
 				var weakRef = new WeakReference(hubTile);
 				SizeChanged += (sender, args) =>
 				{
-					if (weakRef.IsAlive)
-						((HubTile)weakRef.Target).Size = GetSize();
+					var hTile = (HubTile)weakRef.Target;
+					if (hTile != null)
+						hTile.Size = GetSize();
```
adrianknight89: Admittedly, this is WP8, but the usage was incorrect.

StephaneDelcroix:
Hey @adrianknight89. This is a good one. there was no need to open a bug report for it.

Xamarin.Forms.Core/MessagingCenter.cs
```diff
@@ -91,7 +91,7 @@ static void InnerSend(string message, Type senderType, Type argType, object send
 			List<Tuple<WeakReference, Action<object, object>>> actionsCopy = actions.ToList();
 			foreach (Tuple<WeakReference, Action<object, object>> action in actionsCopy)
 			{
-				if (action.Item1.IsAlive && actions.Contains(action))
+				if (action.Item1.Target != null && actions.Contains(action))
 					action.Item2(sender, args);
```
adrianknight89: Btw, it appears that the tuple decouples the subscriber from the action; however, we check to see if the subscriber is alive. I'm not sure if we want to or expect to see the subscriber stick around until the action completes. So an adjustment to the loop would look like this:

```
object subscriber = action.Item1.Target;

if(subscriber == null || !actions.Contains(action))
   continue;

action.Item2(sender, args);
GC.KeepAlive(subscriber);
```

Just a thought. Obviously, if the action had a strong reference to the subscriber, the GC would keep it alive, but this is just enforcing pseudo parent-child relationship.

PR #561, opened 2016-11-24
[Android] Fix NRE in packager dispose

adrianknight89:
### Description of Change ###

When `VisualElementPackager` is disposing itself, it doesn't check if its element is null. You could run the attached repro in the bug description and see that the user is nullifying `Element` and then disposing the packager.

```
{
	if (disposing)
	{
		this.SetElement((VisualElement) null);
		if (this.Packager != null) {

			// exception
			this.Packager.Dispose ();
			this.Packager = null;
		}

		this.Tracker.Dispose();
		this.Tracker = (VisualElementTracker) null;
	}

	base.Dispose(disposing);
}
```

This change is in line with the UAP counterpart. iOS side doesn't use `Element` after setting it to null.

P.S. I'm not sure why iOS seems to be the only platform that sets `Element` to null. Was this on purpose?

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=39228

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/VisualElementPackager.cs
+++ b/Xamarin.Forms.Platform.Android/VisualElementPackager.cs
@@ -52,10 +52,13 @@ namespace Xamarin.Forms.Platform.Android
 					_childViews = null;
 				}
 
-				_renderer.Element.ChildAdded -= _childAddedHandler;
-				_renderer.Element.ChildRemoved -= _childRemovedHandler;
+				if (_renderer.Element != null)
+				{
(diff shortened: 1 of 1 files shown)


hartez:
This doesn't really fix the underlying problem in [39228](https://bugzilla.xamarin.com/show_bug.cgi?id=39228); the problem there was the order in which they cleaned up resources. But the null check does help prevent crashes in future custom renderers, so I'm happy to add it.

PR #563, opened 2016-11-25
[iOS] Platform specifics to control ScrollView content touch delay

adrianknight89:
### Description of Change ###

iOS `ScrollView` is by default delaying content touches, which means when a touch gesture begins, an implicit timer will trigger and `ScrollView` will decide, based on user action within the timer span, whether it should handle the touch or pass it to its content. The default option is useful when we swipe fast and expect that it's a scroll.

However, this seems to be problematic when `ScrollView` is wrapped in a detail of an MDP. When the content of the scroll view is a `Slider` and the user slides the track fast, the slider will not win the gesture from `ScrollView` which in turn opens the master panel of the MDP.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=41778

### API Changes ###

Added:
 - `ShouldDelayContentTouches`
 - Getters/Setters for `ShouldDelayContentTouches`

### PR Checklist ###


Xamarin.Forms.Platform.iOS/Renderers/PhoneMasterDetailRenderer.cs
```diff
@@ -171,7 +171,7 @@ protected override void Dispose(bool disposing)
 
 				if (_tapGesture != null)
 				{
-					if (_clickOffView != null && _clickOffView.GestureRecognizers.Contains(_panGesture))
+					if (_clickOffView != null && _clickOffView.GestureRecognizers.Contains(_tapGesture))
```
rmarinho: please in a new pr..

jassmith:
Everything but the double get API looks fine.

PR #564, opened 2016-11-26
[Android] Vector files can be used as Image source

adrianknight89:
### Description of Change ###

Android `Image` now supports vector files. Note that the bug description also mentions `Button.Image` not supporting vectors. I have a solution for that, but it's not a complete solution.

The test code has three images in landscape, square, and portrait sizes and tests them in different aspects.

To expand on the `Button` issue, although vectors can be loaded and shown, when `Text` is empty, the images do not resize properly. I have done some manual work such as converting vectors to bitmap drawables and then scaled bitmaps on layout change, but this resulted in pixelated images, so this PR only addresses the `Image` part of the bug.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=47923 (partially)

### PR Checklist ###


Xamarin.Forms.Platform.Android/Renderers/ImageRenderer.cs
```diff
@@ -71,40 +71,52 @@ void UpdateAspect()
...
 				}
-				catch (TaskCanceledException)
-				{
-				}
-				catch (IOException ex)
+
+				if (bitmap == null && source is FileImageSource)
 				{
-					Log.Warning("Xamarin.Forms.Platform.Android.ImageRenderer", "Error updating bitmap: {0}", ex);
+					Control.SetImageResource(ResourceManager.GetDrawableByName(((FileImageSource)Element.Source).File));
```
StephaneDelcroix: can't you get/create a bitmap from the vector source ?

adrianknight89: The problem with that approach is that converting vectors to bitmaps is an expensive procedure especially on the `argb8888` setting which allocates 4 bytes per pixel. If you had a list of vectors and you navigated to and from the page frequently, it's easy to end up with an `OutOfMemoryError` exception. Of course, one solution would be to cache the results. `SetImageResource()` seems to be doing everything including resizing out of the box. However, the downside is that decoding is done on the UI thread. If this creates a latency concern, then we'd need to consider caching one again.

This change is by no means an advanced solution. It's just providing basic support for vectors.

adrianknight89:
Added support for view cells and updated sample code to demonstrate the changes. Also improved upon the original changes so that they are not applied if the element is no longer available (due to awaiting). Moved variables around to so that they are defined when we need them.

rmarinho:
Can you rebase please @adrianknight89  ? thanks

adrianknight89:
Done.

rmarinho:
Sorry @adrianknight89 we made some changes on Docs and need to rebase all pr's with latest master so it builds ok and we can run tests, can you rebase again the latest commits from master. 
Thanks.

adrianknight89:
Done.

PR #567, opened 2016-11-27
[Android] Fix issue with the use of null propagation operator

adrianknight89:
### Description of Change ###

When propagating null checks using `?` and accessing a property, one should give the expression a default value or else an NRE can be thrown.

`renderer?.ViewGroup?.ChildCount != 0` throws an exception when `renderer` is null. To fix this, we could default the expression to `0` like so:

`(renderer?.ViewGroup?.ChildCount ?? 0) != 0`

or use old-school null checks. I chose the second option because ReSharper isn't smart enough to remove the squiggly line under `renderer`.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=38989

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/ConditionalFocusLayout.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ConditionalFocusLayout.cs
@@ -26,11 +26,11 @@ namespace Xamarin.Forms.Platform.Android
 			(aView as EntryCellView)?.EditText.SetOnTouchListener(this);
 
 			var viewCell = item as ViewCell;
-			if (viewCell == null || viewCell?.View == null)
+			if (viewCell?.View == null)
 				return;
 
 			IVisualElementRenderer renderer = Platform.GetRenderer(viewCell.View);
-			if (renderer?.ViewGroup?.ChildCount != 0)
+			if (renderer?.ViewGroup != null && renderer.ViewGroup.ChildCount != 0)
 				(renderer.ViewGroup.GetChildAt(0) as EditText)?.SetOnTouchListener(this);
 
 			foreach (Element descendant in viewCell.View.Descendants())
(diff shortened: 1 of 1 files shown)


bentmar:
I think this is the type or crash i get sometimes! can you show me what the stacktrace looks like? im having a hard time creating a repro for my issue, and if this is a solution for it i can stop trying :) @adrianknight89

adrianknight89:
@StephaneDelcroix If you look at the bug description and repro, the user is using a custom view cell renderer and overriding `GetCellCore` method. In there, he's inflating an `axml` to use as native view and returning the new native view. What happens is that as program flow continues, `ListViewAdapter` is trying to apply touch listeners to this new cell and the cell does not have a renderer hence the NRE. The original implementation of `GetCellCore` creates a renderer. 

After thinking about this, I believe my fix is not the right one (although it could perhaps be nice to throw an argument null exception if renderer is null for visual guidance). It appears to me the proper fix might be invoking the base method with the new native cell instead of returning the new native cell.

```
protected override Android.Views.View GetCellCore(Cell item, Android.Views.View convertView, ViewGroup parent, Context context)
{
	var nativeView = convertView;

	if (nativeView == null)
		nativeView = (context as Activity).LayoutInflater.Inflate(Resource.Layout.CustomViewCell, null);

	return base.GetCellCore(item, nativeView, parent, context);
}
```

That said, I will let you guys decide how to proceed. The changes in this PR are not applying touch listeners if the renderer is null. Should we throw an exception instead?

hartez:
I agree that the original submitter's custom renderer might have issues and might not work they way they expect. Regardless, `ApplyTouchListenersToSpecialCells()` should be able to run without crashing if a renderer isn't set. (It was clearly written with that intent; otherwise it would just assume the renderer is not null). 

I whipped up an alternate PR which tests this specific scenario and does the null checking in a more concise manner: https://github.com/xamarin/Xamarin.Forms/pull/587

adrianknight89:
@hartez Great.

PR #570, opened 2016-11-28
[iOS] Avoid using DrawRect on base VisualElementRenderer

rmarinho:
### Description of Change ###
We shouldn't override DrawRect, this method should only used when we  want to provide full implementation of what we are drawing on the screen. Since we provide a some information to the UIView Layer directly in out tracker update method, overriding this method prevents IOS of passing the info to the Layer properly . 

So this fix avoids that by configuring without overriding any Draw methods.
### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=48158

### API Changes ###
None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla48158.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla48158.cs
@@ -6,12 +6,6 @@ using Xamarin.UITest;
 using NUnit.Framework;
 #endif
 
-// Apply the default category of "Issues" to all of the tests in this assembly
-// We use this as a catch-all for tests which haven't been individually categorized
-#if UITEST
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla48158.cs
```diff
@@ -0,0 +1,35 @@
...
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+// Apply the default category of "Issues" to all of the tests in this assembly
+// We use this as a catch-all for tests which haven't been individually categorized
+#if UITEST
+[assembly: NUnit.Framework.Category("Issues")]
+#endif
```
hartez: You don't need lines 9-13; the category only needs to be applied to the namespace once. It's already in `_Template.cs`.

PR #571, opened 2016-11-28
[Android] Control WebView state on pause/resume

adrianknight89:
### Description of Change ###

Unlike iOS, Android `WebView` won't stop playing media if the screen is locked. There doesn't seem to be a direct way to continue playback on resume (could be achieved possibly via html/js parsing), but other necessary code is executed on resume.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=48195

### PR Checklist ###


rmarinho:
This seems code that should be in a Custom renderer for the particular issue. OOB we should just behave like Android does it, since this could be a Platform characteristic someone is relying on, and we don't want to introduce a breaking change

adrianknight89:
Honestly, this should be default Android behavior not even default XF behavior. I cannot think of a good reason why anybody would not want to do this.

ceblanda:
THANKS SO MUCH!
I also added ControlPlayback("Pause"); to the Dispose method of the WebViewRenderer because the video does not stop until the object is garbage collected.

PR #572, opened 2016-11-29
Don't run Command CanExecute on incorrect inherited binding context type

hartez:
### Description of Change ###

On WinRT/UWP during the process of binding, items in a `ListView` initially inherit the binding context of the `ListView` before getting a more specific binding context. If the `DataTemplate` of the `ListView` includes an item which has both 

- a bound `Command` which includes a `CanExecute` method and
- a bound `CommandParameter`

the initial binding of the item will trigger a check on the `CanExecute` method (to determine if `IsEnabled` should be true). The `CanExecute` method attempts to cast the binding context to the correct type and throws an `InvalidCastException`. The `InvalidCastException` is not fatal, but it causes the rest of the `DataTemplate` binding to fail, and the `ListView` items will appear blank (though in the most recent version of XF they are still tappable). 

This change checks the type before executing the command and does not execute the command if the type is incorrect; it also checks the type before checking `CanExecute` and returns `false` if the type is incorrect.

### Bugs Fixed ###

- [47971 – XF UWP ListView Items no longer display ImageCell is the problem](https://bugzilla.xamarin.com/show_bug.cgi?id=47971)

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Command.cs
+++ b/Xamarin.Forms.Core/Command.cs
@@ -3,7 +3,7 @@ using System.Windows.Input;
 
 namespace Xamarin.Forms
 {
-	public sealed class Command<T> : Command
+	public sealed class Command<T> : Command 
 	{
 		public Command(Action<T> execute) : base(o => execute((T)o))
 		{
@@ -63,9 +63,9 @@ namespace Xamarin.Forms
 				{
 					return _canExecute(parameter);
 				}
-				catch (InvalidCastException ex)
+				catch (Exception ex)
 				{
-					Log.Warning("Command", $"CanExecute can't run with the current BindingContext; this is probably a temporary inherited context issue: {ex}");
+					Log.Warning("Command", $"Error determining whether the Command can execute: {ex}");
 				}
 
 				return false;
--- a/Xamarin.Forms.Core/Command.cs
+++ b/Xamarin.Forms.Core/Command.cs
@@ -11,7 +11,14 @@ namespace Xamarin.Forms
 				throw new ArgumentNullException("execute");
 		}
 
-		public Command(Action<T> execute, Func<T, bool> canExecute) : base(o => execute((T)o), o => canExecute((T)o))
+		public Command(Action<T> execute, Func<T, bool> canExecute) 
+			: base(o =>
+			{
+				if (o is T)
+				{
+					execute((T)o);
+				}
+			}, o => o is T && canExecute((T)o))
 		{
 			if (execute == null)
 				throw new ArgumentNullException("execute");
@@ -58,18 +65,7 @@ namespace Xamarin.Forms
 		public bool CanExecute(object parameter)
 		{
 			if (_canExecute != null)
-			{
-				try
-				{
-					return _canExecute(parameter);
-				}
-				catch (Exception ex)
-				{
-					Log.Warning("Command", $"Error determining whether the Command can execute: {ex}");
-				}
-
-				return false;
-			}
+				return _canExecute(parameter);
 
 			return true;
 		}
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Core/Command.cs
```diff
@@ -58,7 +58,18 @@ public Command(Action execute, Func<bool> canExecute) : this(o => execute(), o =
 		public bool CanExecute(object parameter)
 		{
 			if (_canExecute != null)
-				return _canExecute(parameter);
+			{
+				try
+				{
+					return _canExecute(parameter);
```
StephaneDelcroix: `_canExecute` can not throw, unless it's in user code

StephaneDelcroix:
> [...] and we have to think about what's going on when the parameter is null.

by that, I mean testing if `T` is a value or reference type, and only continuing with a `null` value in case of a reference type.

Also, to make things super easy, `Nullable<>` are reference types that can be null...

hartez:
Taking the suggestion from @StephaneDelcroix and just avoiding the `InvalidCastException` in the first place.

Xamarin.Forms.Core/Command.cs
```diff
@@ -3,15 +3,22 @@
 
 namespace Xamarin.Forms
 {
-	public sealed class Command<T> : Command
+	public sealed class Command<T> : Command 
 	{
 		public Command(Action<T> execute) : base(o => execute((T)o))
```
campersau: What about this?

hartez: Yep. Missed that one. Fixing it now.

Xamarin.Forms.Core/Command.cs
```diff
@@ -3,15 +3,31 @@
...
+			{
+				if (o is T)
+				{
+					execute((T)o);
+				}
+			})
 		{
 			if (execute == null)
+			{
 				throw new ArgumentNullException("execute");
```
StephaneDelcroix: while you're at it, change that line as well, and use `nameof()`

Xamarin.Forms.Core/Command.cs
```diff
@@ -3,15 +3,31 @@
...
+	public sealed class Command<T> : Command 
 	{
-		public Command(Action<T> execute) : base(o => execute((T)o))
+		public Command(Action<T> execute) 
+			: base(o =>
+			{
+				if (o is T)
+				{
+					execute((T)o);
+				}
```
StephaneDelcroix: what if `o` is `null` ?

Xamarin.Forms.Core/Command.cs
```diff
@@ -3,15 +3,31 @@
...
 
-		public Command(Action<T> execute, Func<T, bool> canExecute) : base(o => execute((T)o), o => canExecute((T)o))
+		public Command(Action<T> execute, Func<T, bool> canExecute) 
+			: base(o =>
+			{
+				if (o is T)
+				{
+					execute((T)o);
+				}
+			}, o => o is T && canExecute((T)o))
```
StephaneDelcroix: what if `o` is `null` ?

hartez:
@StephaneDelcroix Added some unit tests and handling for value types and Nullable<T>. Let me know if I missed any combinations or cases.

Xamarin.Forms.Core/Command.cs
```diff
@@ -8,31 +9,51 @@ public sealed class Command<T> : Command
...
+			var t = typeof(T);
+
+			// The parameter is null. Is T Nullable?
+			if (Nullable.GetUnderlyingType(t) != null)
+			{
+				return true;
+			}
+
+			// Not a Nullable, if it's a value type then null is not valid
+			return !t.GetTypeInfo().IsValueType;
```
StephaneDelcroix: I'd swap the 2 checks. first check if it's a reference type, and if it's not, start to wonder about Nullable...

PR #440, opened 2016-10-10
[iOS] More button on TabbedPage should work

adrianknight89:
### Description of Change ###

iOS `TabbedPage` shows `More` button when tab items can't fit on screen, but tapping on this button sends you to the first tab. This PR fixes the issue. Note that the `Edit` button is missing, and this is XF change. See #398. In my opinion, this button should be enabled and we should find out what kind of error it causes with a `NavigationPage`. I don't believe a feature should be hidden to hide a bug. When I tested `Edit`, it worked fine though I don't think I tested it with a `NavigationPage`.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=22986

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [ ] Consolidate commits as makes sense

PR #573, opened 2016-11-29
[iOS] Enabling TabbedPage More button for content pages

adrianknight89:
### Description of Change ###

According to #440, enabling `More` button on a `TabbedPage` creates issues when the tab items under this button are navigation page instances because the controller for `More` is also a navigation controller. This PR will enable the button in a limited way so that only `ContentPage` instances can be packed under it. Using content pages can be handy if trivial information such as about, ToS, basic settings, etc. can be shown on a single page.  

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=22986

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/TabbedRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/TabbedRenderer.cs
@@ -28,7 +28,24 @@ namespace Xamarin.Forms.Platform.iOS
 			get { return base.SelectedViewController; }
 			set
 			{
+				if (ReferenceEquals(value, MoreNavigationController))
+				{
+					if(CustomizableViewControllers != null)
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.iOS/Renderers/TabbedRenderer.cs
```diff
@@ -28,7 +28,24 @@ public override UIViewController SelectedViewController
 			get { return base.SelectedViewController; }
 			set
 			{
+				if (ReferenceEquals(value, MoreNavigationController))
+				{
+					if(CustomizableViewControllers != null)
+						throw new NotSupportedException("Editing the order of tab items under the More button is not supported.");
```
adrianknight89: Check against this property because we don't want non-content pages moved to the list.

PR #560, opened 2016-11-24
[iOS] Platform specifics to allow visual elements to become first responder

adrianknight89:
### Description of Change ###

#20 has added `CancelsTouchesInView = false` in order to allow custom renderers to process gestures; however, this has the adverse affect of subviews not canceling tap gestures at page level. In turn, `View.EndEditing(true)` is called and results in regression as explained in the below bug description.

Added a platform specific option for any visual element to become first responder so that `ShouldReceiveTouch` of the page gesture recognizer can be set to false. This is useful in cases such as when a chat application does not need to dismiss keyboard when a `Send` button is hit.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=45723

### API Changes ###

Added:
 - `CanBecomeFirstResponder`
 - Getters/Setters for `CanBecomeFirstResponder`

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #576, opened 2016-11-29
[Android] Platform specifics to control WebView state on app lifecycle change

adrianknight89:
### Description of Change ###

Convert #571 into platform specifics.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=48195

### API Changes ###

Added:
 - `ShouldHandleWebViewStateOnLifecycleChange`
 - Getters and setters for `ShouldHandleWebViewStateOnLifecycleChange`

### PR Checklist ###


adrianknight89:
@ceblanda That's a good question, and I thought about it before. To me, it looked like everything was going away so at that point, it shouldn't matter if the platform specific option is true or false. Perhaps XF team can comment.

Tbh, after all this work, I'm still not 100% sure how renderers work. I have run into cases where they were being disposed but the Element seemed to stay or both went away at the same time. I'd like to see an article on how renderers work from an architectural standpoint. Sometimes it seems like they act like a middle man between Element and NativeView/Control yet other times they are the native view. My mind is looking for a consistent pattern but can't find it. :)

I'll test this part of the code further tonight.

programmation:
@adrianknight89 This has always been the case, and it's not even consistent between platforms. For example, `WebView` is the renderer on iOS but is inside a container on Android. `BoxView` is the renderer on both platforms. But `Label` and `Entry` and most of the rest are all inside containers. The only consistent pattern I've been able to discern is that `Element` is the bindable object, and `Control` is the actual UI object. I'd like to think it's being done for performance reasons, but it makes renderers very awkward to work with because you never know what you're getting. The abstraction also leaks badly because the inconsistency means you have to know the internals of the specific renderer in order to work with it. I'd imagine that the inconsistency also makes the layout engine more complex too, but I haven't looked into that in detail.

adrianknight89:
@programmation I think this is also one of the reasons why `RecycleElement` strategy of `ListView` doesn't work well. It's either renderers aren't being disposed properly or when new renderers are created for cell re-use, they don't initialize native elements properly. I'm all for consistency because it makes development easy. Performance is always an issue, but of course we don't know how big of an issue it is until we see it. I'd still stick to a consistent design and find workarounds for renderers that become performance bottlenecks.

Add to that layouts not setting [`Control`](https://bugzilla.xamarin.com/show_bug.cgi?id=45860) property.

StephaneDelcroix:
@rmarinho to rebase and trigger the tests. Thanks

rmarinho:
build

adrianknight89:
@rmarinho What's happening with this and #560?

rmarinho:
this needs docs

rmarinho:
Hey @adrianknight89 we are trying to clean some of these old pr's you mind rebase this and update the docs and reopen when that's done? 

Thanks

ceblanda:
Did this ever get released?

PR #577, opened 2016-11-30
[Android] Fix NavigationPage dispose crash when it parents a MasterDetailPage

adrianknight89:
### Description of Change ###

When a `MasterDetailPage` is nested under `NavigationPage`, hitting the back button triggers disposal of `NavigationPage` and results in a hard-crash.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=45702

### PR Checklist ###


Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
```diff
@@ -173,13 +152,52 @@ protected override void Dispose(bool disposing)
 				if (_drawerLayout != null && _drawerListener != null)
 				{
 					_drawerLayout.RemoveDrawerListener(_drawerListener);
```
adrianknight89: The issue is `_drawerLayout` is disposed because of the above `foreach` loop. This block of code should be executed at the end so that we can remove drawer listeners safely before disposing the MDP.

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
```diff
@@ -173,13 +152,52 @@ protected override void Dispose(bool disposing)
...
+				{
+					_drawerToggle.Dispose();
+					_drawerToggle = null;
+				}
+
+				if (_backgroundDrawable != null)
+				{
+					_backgroundDrawable.Dispose();
+					_backgroundDrawable = null;
+				}
```
adrianknight89: Dispose `_backgroundDrawable` in case it's not null.

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
```diff
@@ -173,13 +152,52 @@ protected override void Dispose(bool disposing)
 				if (_drawerLayout != null && _drawerListener != null)
 				{
 					_drawerLayout.RemoveDrawerListener(_drawerListener);
+				}
+
+				if (_drawerListener != null)
```
adrianknight89: This can be moved to its own if block since the previous `_drawerLayout != null && _drawerListener != null` may evaluate to false.

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
```diff
@@ -499,7 +514,7 @@ void RegisterToolbar()
 			if (renderer == null)
 				return;
 
-			_drawerLayout = (DrawerLayout)renderer;
+			_drawerLayout = renderer;
```
adrianknight89: `renderer` is always a `DrawerLayout`.

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
```diff
@@ -173,13 +152,52 @@ protected override void Dispose(bool disposing)
...
+					}
+
+					var navController = (INavigationPageController)Element;
+
+					navController.PushRequested -= OnPushed;
+					navController.PopRequested -= OnPopped;
+					navController.PopToRootRequested -= OnPoppedToRoot;
+					navController.InsertPageBeforeRequested -= OnInsertPageBeforeRequested;
+					navController.RemovePageRequested -= OnRemovePageRequested;
+				}
```
adrianknight89: Execute this logic at the end.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla45702.cs
```diff
@@ -0,0 +1,40 @@
...
+						HorizontalTextAlignment = TextAlignment.Center,
+						VerticalTextAlignment = TextAlignment.Center,
+						Text =
+							"Pressing Back button to go to home screen should not crash the app when MasterDetailPage is nested under NavigationPage"
+					}
+				}
+			};
+		}
+	}
+}
```
hartez: This whole test class can probably just be dropped. It doesn't test the issue in 45702, and I'm not sure there even _is_ a way to test that from inside ControlGallery.

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
```diff
@@ -173,13 +152,52 @@ protected override void Dispose(bool disposing)
...
+
+				if (_backgroundDrawable != null)
+				{
+					_backgroundDrawable.Dispose();
+					_backgroundDrawable = null;
+				}
 
 				Current = null;
 
+				// dispose child renderers at the end to avoid compliations with MasterDetailPage being nested under NavigationPage
```
hartez: Let's update this comment to something more descriptive, like
"We dispose the child renderers after cleaning up everything related to Drawer Layout in case one of the children is a Master Detail Page (which may dispose of the Drawer Layout)"

adrianknight89:
Done.

hartez:
@adrianknight89 Sorry for the trouble, but could you rebase this? I think the `.pfx` files are out of date and it's failing to build.

adrianknight89:
Done.

hartez:
@adrianknight89 Thanks!

hartez:
Only failing test is the broken CellsGalleryImageUrlCellList test on iOS 10; not an issue with these changes.

PR #581, opened 2016-11-30
[WinRT/UWP] Apply BackgroundColor to Stepper buttons

pauldipietro:
### Description of Change ###

When presently used in WinRT/UWP, the `Stepper` can potentially have the background color run the width of the screen due to the color being applied on the panel behind its buttons opposed to the buttons themselves (which are transparent to let the color show through). By adding a `DependencyProperty` on the `StepperControl` which works on the buttons themselves, this can be made to look more as expected.

A present side effect is that changing the color while one of the buttons is not enabled will make the button appear as that color instead of gray until it returns to that state  (e.g. the stepper is at its minimum value of 0 and the color changes; the minus stepper will appear as that color until it is incremented, then decremented) but it's probably an edge case at best.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=48236

### API Changes ###

(In the StepperControl implementations in WinRT/UWP)
- public static readonly DependencyProperty ButtonBackgroundColorProperty
- public Color ButtonBackgroundColor

### Behavioral Changes ###

This will potentially change how the Stepper looks in a user's app, but should be more as expected without the color running over. There's the edge case of changing the color, which I didn't particularly look that far into. If a user really wanted to use the `BackgroundColor` as before, they could presumably override `UpdateBackgroundColor` on a custom renderer and use the `BackgroundColor` as before.

### PR Checklist ###


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla48236.cs
```diff
@@ -0,0 +1,48 @@
...
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+// Apply the default category of "Issues" to all of the tests in this assembly
+// We use this as a catch-all for tests which haven't been individually categorized
+#if UITEST
+[assembly: NUnit.Framework.Category("Issues")]
```
hartez: This attribute only needs to be declared once for the assembly; you can remove if from this file.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla48236.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Bugzilla, 48236, "[WinRT/UWP] BackgroundColor for Stepper behaves differently compared to iOS to Android", PlatformAffected.WinRT)]
+	public class Bugzilla48236 : TestContentPage
+	{
+		protected override void Init()
+		{
+			var stepper = new Stepper
```
hartez: Could we add some explanatory text for whoever is verifying this test? There's nothing to tell the person running it whether it failed.

Xamarin.Forms.Platform.UAP/StepperControl.cs
```diff
@@ -18,6 +18,8 @@ public sealed class StepperControl : Control
 		public static readonly DependencyProperty IncrementProperty = DependencyProperty.Register("Increment", typeof(double), typeof(StepperControl),
 			new PropertyMetadata(default(double), OnIncrementChanged));
 
+		public static readonly DependencyProperty ButtonBackgroundColorProperty = DependencyProperty.Register("ButtonBackgroundColor", typeof(Color), typeof(StepperControl), new PropertyMetadata(default(Color), OnButtonBackgroundColorChanged));
```
hartez: We should probably use `nameof(ButtonBackgroundColor)`.

Xamarin.Forms.Platform.WinRT/StepperControl.xaml.cs
```diff
@@ -15,6 +15,8 @@ public sealed partial class StepperControl : UserControl
 		public static readonly DependencyProperty IncrementProperty = DependencyProperty.Register("Increment", typeof(double), typeof(StepperControl),
 			new PropertyMetadata(default(double), OnIncrementChanged));
 
+		public static readonly DependencyProperty ButtonBackgroundColorProperty = DependencyProperty.Register("ButtonBackgroundColor", typeof(Color), typeof(StepperControl), new PropertyMetadata(default(Color), OnButtonBackgroundColorChanged));
```
hartez: Another `nameof` opportunity.

hartez:
The test failing on iOS 10 (CellsGalleryImageUrlCell) has nothing to do with these changes; that failure can be ignored.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla48236.cs
```diff
@@ -20,6 +14,10 @@ public class Bugzilla48236 : TestContentPage
 	{
 		protected override void Init()
 		{
+			// Refer to https://bugzilla.xamarin.com/show_bug.cgi?id=48236 for this issue. WinRT/UWP had
+			// an issue where the background color would be used on the area containing the buttons, 
+			// potentially causing the color to run the width of the screen. Only the buttons should have
+			// a background color.
```
hartez: The comments are good, but I meant on-screen text for anyone manually running the test. So they have some idea of what constitutes success and failure (e.g., the on-screen text for [36014](https://github.com/xamarin/Xamarin.Forms/blob/74cb5c4a97dcb123eb471f6b1dffa1267d0305aa/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla36014.cs) ).

PR #582, opened 2016-11-30
[UWP] Fix CommandBar being clipped when expanded

jimmgarrido:
### Description of Change ###

It appears that setting the CommandBar Height was causing it to be clipped when expanded by pressing the "..." button. It also seemed to cause the "..." button to not even appear on Mobile.

This fix also defines the default CommandBar background because by removing the Height in the page styles another issue appeared on some pages where the expanded area of CommandBar was transparent. This was a result of the entire CommandBar being transparent since the default color was not being applied, e.g. in TabbedPageRenderer it was returning [at this line](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.UAP/TabbedPageRenderer.cs#L297).

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=47295

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla47295.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla47295.cs
@@ -1,0 +1,39 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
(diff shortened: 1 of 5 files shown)


StephaneDelcroix:
see #588 for a rebased version

hartez:
This fixes the "command labels are cut off" problem, but it introduces a different problem when the bar's color is set:

Uneven Title Background

I've created https://github.com/xamarin/Xamarin.Forms/pull/594 as an alternative that addresses both problems.

jimmgarrido:
Perfect! I had noticed the uneven expansion but figured this fix would be alright in the meantime 😄

PR #578, opened 2016-11-30
[iOS] Prevent multiple ListView cells from being swiped simultaneously

adrianknight89:
### Description of Change ###

iOS `ListView` is allowing multiple cells to be swiped to show their context action menus at the same time. This PR will fix that. It seems that each cell has its own dedicated `ScrollView` and they are not aware of each other.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=43735

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #584, opened 2016-12-01
[iOS] Preserve context menu state on orientation change

adrianknight89:
### Description of Change ###

If a view cell has its context menu open and you change device orientation, this closes the menu. Can you merge this after #578? Both touch the same file and it will be easier to rebase in that order.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=48237

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/ContextScrollViewDelegate.cs
+++ b/Xamarin.Forms.Platform.iOS/ContextScrollViewDelegate.cs
@@ -45,14 +45,15 @@ namespace Xamarin.Forms.Platform.iOS
 		GlobalCloseContextGestureRecognizer _globalCloser;
 
 		bool _isDisposed;
-
+		UIScrollView _scrollView;
 		UITableView _table;
 
-		public ContextScrollViewDelegate(UIView container, List<UIButton> buttons, bool isOpen)
+		public ContextScrollViewDelegate(UIScrollView scrollView, UIView container, List<UIButton> buttons, bool isOpen)
 		{
 			IsOpen = isOpen;
 			_container = container;
 			_buttons = buttons;
+			_scrollView = scrollView;
 
 			for (var i = 0; i < buttons.Count; i++)
 			{
@@ -62,11 +63,17 @@ namespace Xamarin.Forms.Platform.iOS
 				ButtonsWidth += b.Frame.Width;
 				_finalButtonSize = b.Frame.Width;
 			}
+
+			if (IsOpen)
+			{
+				RemoveHighlight();
+				AddClosers();
+			}
 		}
 
 		public nfloat ButtonsWidth { get; }
 
-		public Action ClosedCallback { get; set; }
+		public NSAction ClosedCallback { get; set; }
 
 		public bool IsOpen { get; private set; }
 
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Platform.iOS/ContextScrollViewDelegate.cs
```diff
@@ -140,50 +147,57 @@ public override void WillEndDragging(UIScrollView scrollView, PointF velocity, r
...
+			NSAction close = () =>
+			{
+				if (UIDevice.CurrentDevice.CheckSystemVersion(8, 0))
+					RestoreHighlight();
+
+				IsOpen = false;
+				_scrollView.SetContentOffset(new PointF(0, 0), true);
+
+				ClearCloserRecognizer();
+			};
```
adrianknight89: Move `close` declaration outside the `while` loop to avoid unnecessary initialization in each iteration.

Xamarin.Forms.Platform.iOS/ContextScrollViewDelegate.cs
```diff
@@ -45,14 +45,15 @@ internal class ContextScrollViewDelegate : UIScrollViewDelegate
 		GlobalCloseContextGestureRecognizer _globalCloser;
 
 		bool _isDisposed;
-
+		UIScrollView _scrollView;
 		UITableView _table;
 
-		public ContextScrollViewDelegate(UIView container, List<UIButton> buttons, bool isOpen)
+		public ContextScrollViewDelegate(UIScrollView scrollView, UIView container, List<UIButton> buttons, bool isOpen)
```
adrianknight89: Passing scroll view so we can add closers to it and its cell on orientation change which causes dispose. Doing this also allows us to change the signature of several methods so that we no longer need to pass the scroll view over and over.

Xamarin.Forms.Platform.iOS/ContextScrollViewDelegate.cs
```diff
@@ -196,24 +210,25 @@ protected override void Dispose(bool disposing)
 
 			if (disposing)
 			{
+				ClearCloserRecognizer();
```
adrianknight89: Since we have _`scrollView`, we can dispose its closers properly. This method also nullifies `_table`.

Xamarin.Forms.Platform.iOS/ContextScrollViewDelegate.cs
```diff
@@ -196,24 +210,25 @@ protected override void Dispose(bool disposing)
...
 				ClosedCallback = null;
 
-				_table = null;
 				_backgroundView = null;
 				_container = null;
 
 				_buttons = null;
+				_scrollView = null;
```
adrianknight89: Release object reference.

rmarinho:
I was going to say this is intended  behaviour .. but @jassmith  can confirm it.

adrianknight89:
@rmarinho If you go from landscape to portrait, then the context menu stays open. The issue is only with portrait to landscape, but the fix addresses that.

rmarinho:
I think we can't accept this but @jassmith can have the final world.

jassmith:
Sorry after review we cant take this

PR #597, opened 2016-12-03
Fix Bug 39359

fengrui358:
### Description of Change ###

UriImageSource does cache failed HTTP responses

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=39359


StephaneDelcroix:
this is a dupe of #599

PR #461, opened 2016-10-15
Return group instead of internal class on ItemTapped of grouped ListView

adrianknight89:
### Description of Change

`ItemTapped` of `ListView` returns `ItemTappedEventArgs` that is supposed to be showing item tapped as well as the group it lives in. As of now, XF is returning an internal object `TemplatedItemsList`. This PR will ensure that parameters are directly convertible to user-defined models.

```
void OnItemTapped(Object sender, ItemTappedEventArgs e)
{
    var item = (MyItem)(e.Item);
    var group = (MyGroup)(e.Group);
}
```

There is no point in returning an internal member. Similarly, users should not be using reflection. This will be a breaking change for those who were using reflection. So, please make sure to communicate the change.
### Bugs Fixed
- https://bugzilla.xamarin.com/show_bug.cgi?id=45092
### PR Checklist
- [ ] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #598, opened 2016-12-03
[iOS] Fix ItemTapped and ItemSelected event args in Recycle caching strategy

adrianknight89:
### Description of Change ###

In `Recycle` caching strategy, `ItemTapped` and `ItemSelected` are returning null for the current item. It appears that the binding context for cells is null. I'm not sure if the fix needs to be fixing the `BindingContext` or if these changes are fine. The sample code tests both caching modes.

There is also another PR for another issue: #461 

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=48271

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla48271.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla48271.cs
@@ -80,14 +80,16 @@ namespace Xamarin.Forms.Controls.Issues
 			Content = listView;
 		}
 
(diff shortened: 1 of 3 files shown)


adrianknight89:
@StephaneDelcroix I'm going to retract this. The original repro was using a custom renderer which I wasn't paying attention to. 

@davidbritch Would you mind updating your [sample](https://developer.xamarin.com/guides/xamarin-forms/custom-renderer/viewcell/) to also work with `RecycleElement`? When I tried to use this caching strategy, I was prompted to implement `INativeElementView` on the custom renderer. It's either things needs to be implemented differently for `RecycleElement` to work or there is a bug somewhere that needs a fix.

davidbritch:
In this sample the cell custom renderers need updating to work for `RecycleElement`. I've made the required changes and will be pushing out later today.

PR #600, opened 2016-12-04
[iOS] Suggested fix for 39908

adrianknight89:
### Description of Change ###

When you tap the `Back` button on iOS `NavigationPage` rapidly, this will mess up XF's navigation stack because iOS doesn't seem to be calling `PopViewController` for every page being removed.

@rmarinho `ViewDidDisappear` will be called sequentially for every page in `PageRenderer`, so perhaps we can sync the navigation stack there?

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=39908

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
@@ -86,20 +86,20 @@ namespace Xamarin.Forms.Platform.iOS
 
 			if (!_appeared || _disposed)
 				return;
-
-			var navigationPage = Element.Parent as NavigationPage;
-			if (navigationPage != null)
-			{
-				int managedStackCount = navigationPage.Navigation.NavigationStack.Count;
-				IVisualElementRenderer renderer = Platform.GetRenderer(navigationPage);
-				var navigationController = renderer as UINavigationController;
-				int? nativeStackCount = navigationController?.ViewControllers.Length;
-				if (managedStackCount > nativeStackCount)
-					await ((INavigationPageController)renderer.Element).PopAsyncInner(animated, true);
-			}
-
 			_appeared = false;
 			PageController.SendDisappearing();
+
+			var navigationPage = Element.Parent as NavigationPage;
+			if (navigationPage == null)
+				return;
+
+			int managedStackCount = navigationPage.Navigation.NavigationStack.Count;
+			IVisualElementRenderer renderer = Platform.GetRenderer(navigationPage);
+			var navigationController = renderer as UINavigationController;
+
+			int? nativeStackCount = navigationController?.ViewControllers.Length;
+			if (nativeStackCount.HasValue && managedStackCount > nativeStackCount.Value)
+				await ((INavigationPageController)renderer.Element).PopAsyncInner(animated, true);
 		}
 
 		public override void ViewDidLoad()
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
```diff
@@ -80,7 +80,7 @@ public override void ViewDidAppear(bool animated)
 			UpdateStatusBarPrefersHidden();
 		}
 
-		public override void ViewDidDisappear(bool animated)
+		public override async void ViewDidDisappear(bool animated)
```
StephaneDelcroix: making this `async void` will be `fire and forget`, and the thread calling this will continue wether or not this method is completed. That's probably _not_ what we want.

Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
```diff
@@ -89,6 +89,18 @@ public override void ViewDidDisappear(bool animated)
...
+
+			var navigationPage = Element.Parent as NavigationPage;
+			if (navigationPage == null)
+				return;
+
+			int managedStackCount = navigationPage.Navigation.NavigationStack.Count;
+			IVisualElementRenderer renderer = Platform.GetRenderer(navigationPage);
+			var navigationController = renderer as UINavigationController;
+
+			int? nativeStackCount = navigationController?.ViewControllers.Length;
```
StephaneDelcroix: it's probably cheaper, and more clear to wrap this in a `if (navigationController != null)` than playing with `int?`

like this:
```
if (navigationController != null && navigationController.ViewControllers.Length <= managedStackCount)
    await your stuff;
```

Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
adrianknight89: So what's a workaround for this? I need to await `PopAsyncInner` or else we end up with the same issue. Changing the return signature of `ViewDidDisappear` is also not an option.

StephaneDelcroix:
Don't use await, but wait synchronously for the task to be complete

--
Stéphane Delcroix
+32 475 96 22 99

adrianknight89:
@StephaneDelcroix Let me know if another change is needed.

Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
```diff
@@ -89,6 +90,20 @@ public override void ViewDidDisappear(bool animated)
...
+			if (navigationPage == null)
+				return;
+
+			int managedStackCount = navigationPage.Navigation.NavigationStack.Count;
+			IVisualElementRenderer renderer = Platform.GetRenderer(navigationPage);
+
+			var navigationController = renderer as UINavigationController;
+			if (navigationController != null && managedStackCount > navigationController.ViewControllers.Length)
+			{
+				Task<Page> task = ((INavigationPageController)renderer.Element).PopAsyncInner(animated, true);
```
rmarinho: The fix should be on NavigationPage and not here.. I think the code in NavigationPAge is not good and we should fix it.. this looks more like a patch  ..

adrianknight89: Well, this is still better than a broken system. I'll try to find a solution on `NavigationRenderer` tonight, but I'd consider including this for now instead of potentially waiting for months to fix the navigation stack.

adrianknight89:
@rmarinho I think we need to leverage `PopToViewController` in addition to `PopViewController` to fix this issue. Do we expect Appearing/Disappearing events to be fired sequentially on each page in addition to fixing the XF navigation stack?

PR #602, opened 2016-12-04
Localized test code to Japanese.

qwerty2501:
### Description of Change ###
In ja-JP string format convert 0.95d  with "{0:P2}" to "95.00%".
but excepted string format convert to "95.00 %" in test code.
So I modified test code and refactored.
### Bugs Fixed ###
- Nothing.

### API Changes ###
Added:
 - Nothing.

Changed:
 - Nothing.

### Behavioral Changes ###
-Nothing.

### PR Checklist ###

- [x ] Has tests (if omitted, state reason in description)
- [x ] Rebased on top of master at time of PR
- [x ] Changes adhere to coding standard
- [x ] Consolidate commits as makes sense


qwerty2501:
I understand why testng in tr-TR.
And, Nothing reason why testing for ja-JP.
Thank you.

PR #603, opened 2016-12-04
[XamlC] compile ThicknessTypeConverter

StephaneDelcroix:
### Description of Change ###

[XamlC] compile ThicknessTypeConverter

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/CompiledConverters/ThicknessTypeConverter.cs
+++ b/Xamarin.Forms.Build.Tasks/CompiledConverters/ThicknessTypeConverter.cs
@@ -1,0 +1,51 @@
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+
+using Xamarin.Forms.Xaml;
+
+namespace Xamarin.Forms.Core.XamlC
+{
+	class ThicknessTypeConverter : ICompiledTypeConverter
+	{
+		public IEnumerable<Instruction> ConvertFromString(string value, ModuleDefinition module, BaseNode node)
+		{
+			if (!string.IsNullOrEmpty(value)) {
+				double l, t, r, b;
+				var thickness = value.Split(',');
+				switch (thickness.Length) {
+				case 1:
+					if (double.TryParse(thickness[0], NumberStyles.Number, CultureInfo.InvariantCulture, out l))
+						return GenerateIL(module, l);
+					break;
+				case 2:
+					if (double.TryParse(thickness[0], NumberStyles.Number, CultureInfo.InvariantCulture, out l) &&
+					    double.TryParse(thickness[1], NumberStyles.Number, CultureInfo.InvariantCulture, out t))
+						return GenerateIL(module, l, t);
+					break;
+				case 4:
+					if (double.TryParse(thickness[0], NumberStyles.Number, CultureInfo.InvariantCulture, out l) &&
+					    double.TryParse(thickness[1], NumberStyles.Number, CultureInfo.InvariantCulture, out t) &&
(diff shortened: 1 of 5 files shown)


activa:
May I ask why you are consistently ignoring coding style requirements for Xamarin.Forms? I've commented about this on 2 of your other commits but you ignored that entirely and merged the code anyway.

Don't you think it's important to have a consistent coding style across all code?

Xamarin.Forms.Core/ThicknessTypeConverter.cs
```diff
@@ -3,6 +3,7 @@
 
 namespace Xamarin.Forms
 {
+	[Xaml.ProvideCompiled("Xamarin.Forms.Core.XamlC.ThicknessTypeConverter")]
```
StephaneDelcroix: it references the class by name, so the Core doesn't hold a reference on the BuildTask assembly

Xamarin.Forms.Build.Tasks/CompiledConverters/ThicknessTypeConverter.cs
```diff
@@ -0,0 +1,51 @@
...
+					break;
+				case 4:
+					if (double.TryParse(thickness[0], NumberStyles.Number, CultureInfo.InvariantCulture, out l) &&
+					    double.TryParse(thickness[1], NumberStyles.Number, CultureInfo.InvariantCulture, out t) &&
+						double.TryParse(thickness[2], NumberStyles.Number, CultureInfo.InvariantCulture, out r) &&
+					    double.TryParse(thickness[3], NumberStyles.Number, CultureInfo.InvariantCulture, out b))
+						return GenerateIL(module, l, t, r, b);
+					break;
+				}
+			}
```
StephaneDelcroix: this is very similar to the runtime `ThicknessTypeConverter` so far

Xamarin.Forms.Build.Tasks/CompiledConverters/ThicknessTypeConverter.cs
```diff
@@ -0,0 +1,51 @@
...
+			throw new XamlParseException($"Cannot convert \"{value}\" into {typeof(Thickness)}", node);
+		}
+
+		IEnumerable<Instruction> GenerateIL(ModuleDefinition module, params double[] args)
+		{
+			foreach (var d in args)
+				yield return Instruction.Create(OpCodes.Ldc_R8, d);
+			var thicknessCtor = module.Import(typeof(Thickness)).Resolve().Methods.FirstOrDefault(md => md.IsConstructor && md.Parameters.Count == args.Length);
+			var thicknessCtorRef = module.Import(thicknessCtor);
+			yield return Instruction.Create(OpCodes.Newobj, thicknessCtorRef);
```
StephaneDelcroix: this produces the following IL
```
IL_00b9:  ldc.r8 0.
IL_00c2:  ldc.r8 20.
IL_00cb:  ldc.r8 0.
IL_00d4:  ldc.r8 20.
IL_00dd:  newobj instance void valuetype [Xamarin.Forms.Core]Xamarin.Forms.Thickness::'.ctor'(float64, float64, float64, float64)
```

equivalent to this c# code
```
new Thickness(0d,20d,0d,20);
```

a sharp mind would notice that using ldc.r8 for loading 0, or 20, takes 9B, and that's way too much. like if those bytes were for free ! when this IL is equally as good
```
ldc.i4.0
conv.r8
```

and takes only 2Bytes.

This is a nice observation, but that would make this code very hard to understand. That's why we do this kind of optimization as a post-generation step, here for i8 https://github.com/xamarin/Xamarin.Forms/pull/611/files#diff-6b2666065ac83ca89999db8a704b717eR247, and I should write a conversion for r8 as well, and submit it upstream

PR #498, opened 2016-11-02
[iOS] Add ability to toggle selector actions on and off when an input field is long clicked

adrianknight89:
### Description of Change ###

There is no way on iOS to disable selector actions when an input field is long clicked. This functionality may be needed in cases such as when preventing users from editing a password entry via the clipboard. I wanted to implement the same behavior on Android, but AppCompat seems to be ignoring all efforts. For now, it's left for another time.

This feature is currently implemented for `Entry`, `Picker`, `TimePicker`, and `DatePicker`.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=37311
- https://bugzilla.xamarin.com/show_bug.cgi?id=38480

### API Changes ###

Added:
 - 6 methods and 1 property for each class
 - `SelectorAction` enum

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #608, opened 2016-12-05
[iOS] MasterDetailPage master icon should be changeable from any page on the stack

adrianknight89:
### Description of Change ###

At the moment, `Master` icon can only be changed if you are on the first page of the detail (`NavigationPage`). This only addresses the phone side of it since I do not have an iPad. Also, in the bug description, the OP wanted the ability to change icon tinting independently of `BarTextColor` which controls tinting for both icons and title. This could probably be a future update to go in an [effect library](https://github.com/xamarin/Xamarin.Forms/pull/498#issuecomment-260752019).

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=32865

### PR Checklist ###


Xamarin.Forms.Platform.iOS/Renderers/NavigationRenderer.cs
```diff
@@ -644,34 +642,7 @@ void UpdateLeftBarButtonItem(ParentingViewController containerController, Page p
 			if ((firstPage != pageBeingRemoved && currentChild != firstPage && NavigationPage.GetHasBackButton(currentChild)) || _parentMasterDetailPage == null)
```
adrianknight89: This expression seems complicated, and it probably makes sense to not complicate it even further. So I refactored below code into its own method and let phone master detail renderer call it freely.

rmarinho:
Needs rebase

adrianknight89:
Done.

PR #609, opened 2016-12-05
[Android] Double tapping on context action should not crash

adrianknight89:
### Description of Change ###

If a cell context menu is open and you double tap an action, it shouldn't crash.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=45027

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla45027.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla45027.cs
@@ -1,0 +1,58 @@
+﻿using System.Collections.Generic;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+// Apply the default category of "Issues" to all of the tests in this assembly
+// We use this as a catch-all for tests which haven't been individually categorized
+#if UITEST
+[assembly: NUnit.Framework.Category("Issues")]
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Bugzilla, 45027, "App crashes when double tapping on ToolbarItem or MenuItem very quickly", PlatformAffected.Android)]
+	public class Bugzilla45027 : TestContentPage // or TestMasterDetailPage, etc ...
+	{
+		protected override void Init()
+		{
+			var list = new List<int>();
+			for (var i = 0; i < 10; i++)
+				list.Add(i);
+
+			var listView = new ListView
+			{
+				ItemsSource = list,
+				ItemTemplate = new DataTemplate(() =>
+				{
+					var label = new Label();
+					label.SetBinding(Label.TextProperty, new Binding("."));
+
+					return new ViewCell
+					{
+						View = new ContentView
+						{
+							Content = label,
+						},
+						ContextActions = { new MenuItem
+						{
+							Text = "Action"
+						},
+						new MenuItem
+						{
+							Text = "Delete",
+							IsDestructive = true
+						} }
+					};
+				})
+			};
+
+			Content = listView;
+		}
+	}
+}
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla45027.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla45027.cs
@@ -7,12 +7,6 @@ using Xamarin.UITest;
 using NUnit.Framework;
 #endif
 
-// Apply the default category of "Issues" to all of the tests in this assembly
-// We use this as a catch-all for tests which haven't been individually categorized
-#if UITEST
-[assembly: NUnit.Framework.Category("Issues")]
-#endif
-
 namespace Xamarin.Forms.Controls.Issues
 {
 	[Preserve(AllMembers = true)]
@@ -24,6 +18,19 @@ namespace Xamarin.Forms.Controls.Issues
 			var list = new List<int>();
 			for (var i = 0; i < 10; i++)
 				list.Add(i);
+
+			var stackLayout = new StackLayout
+			{
+				Orientation = StackOrientation.Vertical,
+				Children =
+				{
+					new Label
+					{
+						Text = "Long tap list items to display context menu. Double tapping each action rapidly should not crash.",
+						HorizontalTextAlignment = TextAlignment.Center
+					}
+				}
+			};
 
 			var listView = new ListView
 			{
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.Android/CellAdapter.cs
```diff
@@ -257,6 +256,9 @@ bool HandleContextMode(AView view, int position)
 
 		void OnActionItemClickedImpl(IMenuItem item)
 		{
+			if (ActionModeContext == null)
```
rmarinho: Why is it null if it's open ?

adrianknight89: This seems to be some kind of race condition. When you tap an action, `OnActionItemClicked` is called. This invokes `_supportActionMode?.Finish()` which triggers `OnDestroyActionMode`. Now, inside this method, we are setting `ActionModeContext` to nul. By the time, the menu disappears from screen, we can tap on the action again in which case we end up with a crash because `ActionModeContext` is null.

Another solution is to do `item.SetEnabled(false);` inside `OnActionItemClicked`. I changed the PR to use this version instead. This will stop `OnActionItemClicked` from being called twice.

Xamarin.Forms.Platform.Android/CellAdapter.cs
```diff
@@ -82,14 +82,15 @@ internal AView ContextView
...
+			item.SetEnabled(false);
 			_actionMode?.Finish();
 			return true;
 		}
 
 		bool global::Android.Support.V7.View.ActionMode.ICallback.OnActionItemClicked(global::Android.Support.V7.View.ActionMode mode, IMenuItem item)
 		{
 			OnActionItemClickedImpl(item);
-
+			item.SetEnabled(false);
```
rmarinho: Let's add a comment here so we know why we had this..

adrianknight89: Done.

adrianknight89:
@rmarinho Actually, we need to disable all items not just the current one. You can tap one item and immediately tap another which will crash the app. The old change doesn't have this problem.

adrianknight89:
@rmarinho Nvm. Found it.

rmarinho:
Can you rebase please. thanks

adrianknight89:
Done.

rmarinho:
It's failing 100 tests.. please check android UITests

adrianknight89:
@rmarinho Changed the place where the menu was being cleared. Should be fixed now.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla45027.cs
```diff
@@ -0,0 +1,58 @@
...
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+// Apply the default category of "Issues" to all of the tests in this assembly
+// We use this as a catch-all for tests which haven't been individually categorized
+#if UITEST
```
hartez: You can remove the duplicate declaration of this attribute here.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla45027.cs
```diff
@@ -0,0 +1,58 @@
...
+[assembly: NUnit.Framework.Category("Issues")]
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Bugzilla, 45027, "App crashes when double tapping on ToolbarItem or MenuItem very quickly", PlatformAffected.Android)]
+	public class Bugzilla45027 : TestContentPage // or TestMasterDetailPage, etc ...
```
hartez: Since this is a race condition (and since UI test taps are so slow), I'm guessing we can't reliably make this an automated UI test.
Can you add some instructions for the tester (e.g. "long press an item, tap Action twice quickly")?

hartez:
That Android test failure (45926) is because of a bad test which has been fixed in master - it'll go away once you rebase.

adrianknight89:
Done.

CSkoubo:
Hi Adrian, i've been working on a fix for this issue too. I think you hide the real problem with your solution. The issue is that the `Item` in `OnActionItemClicked` is `null`. I would much rather just have it 
`If (Item == null) return true;`

rmarinho:
Keeps failing tests on iOS 10 :(

adrianknight89:
Why does it fail iOS tests if the change was made for Android? 

@CSkoubo I will test your suggestion tonight. I think I had checked `Item` before, and it wasn't null for me, but it's been a while since the time of the PR, so I'll go back and verify. Thanks. :)

adrianknight89:
@CSkoubo So, I looked at this again. I wasn't able to reproduce a null `Item`. When `onActionItemClicked` is called, Android seems to be providing the item that was clicked. Here's the [documentation](https://goo.gl/m7UOQb).

Note that `ActionModeContext` is null. If you look at the commit history, I was handling this first, but as @rmarinho suggested `ActionModeContext` should not be null if the menu is open.

The if blocks below seem to be never hit for me:

```
public bool OnActionItemClicked(ActionMode mode, IMenuItem item)
{
	if (item == null)
	{
		System.Diagnostics.Debug.WriteLine("Item is null");
		return true;
	}
	//mode.Menu.Clear();
	OnActionItemClickedImpl(item);
	_actionMode?.Finish();
	return true;
}

bool global::Android.Support.V7.View.ActionMode.ICallback.OnActionItemClicked(global::Android.Support.V7.View.ActionMode mode, IMenuItem item)
{
	if (item == null)
	{
		System.Diagnostics.Debug.WriteLine("Item is null");
		return true;
	}
	//mode.Menu.Clear();
	OnActionItemClickedImpl(item);
	_supportActionMode?.Finish();
	return true;
}
```

When I was searching for a solution, `mode.Menu.Clear();` seemed to do the trick. To be honest, I cannot think of a better solution.

CSkoubo:
@adrianknight89 you are right! sorry for that. Spend some time looking into this.

So either use your current solution or use the "enabled" as you said -perhaps a "canExecute" would be of use. I haven't been able to get my local XF to work, so i am unable to investigate some more. (only having a mac, so need to install a VM)

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla45027.cs
rmarinho: @adrianknight89 can you add the comment?

adrianknight89: @rmarinho Already added.

pratius:
@adrianknight89 Yes, Can you please provide the small POC in Xamarin form? Because it's giving exception  for me on "**ListView with ContextActions throws NullReferenceException when Multi-touch Tapped on empty and non-empty cell**"

PR #611, opened 2016-12-05
[XamlC] fix loading ulongs, optimize bytecode for longs

StephaneDelcroix:
### Description of Change ###

[XamlC] fix loading ulongs, optimize bytecode for longs

The code had a casting issue while generating IL for ulongs.

This patches, in addition to fix the issue, also optimize `ldc.i8` calls by 3 to 7B (from an original 9B) when the [u]long being loaded fits an int.

This optimization has been send to upstream as well https://github.com/jbevain/cecil/pull/307

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #614, opened 2016-12-06
Optimize stlod ldloc

StephaneDelcroix:
### Description of Change ###

replace (2 to 6B)
```
ldloc x
stloc x
```
by (1 to 4B)
```
dup
stloc x
```

this gain is quite small (~1%)

require #611 and #612 

The current tests are covering this already

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/MethodBodyExtensions.cs
+++ b/Xamarin.Forms.Build.Tasks/MethodBodyExtensions.cs
@@ -12,6 +12,7 @@ namespace Xamarin.Forms.Build.Tasks
 				throw new ArgumentNullException(nameof(self));
 
 			self.OptimizeLongs();
+			self.OptimizeStLdLoc();
 			self.OptimizeMacros();
 		}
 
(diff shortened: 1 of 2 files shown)


Xamarin.Forms.Build.Tasks/MethodBodyExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+			var method = self.Method;
+			for (var i = 0; i < self.Instructions.Count; i++) {
+				var instruction = self.Instructions[i];
+				if (instruction.OpCode.Code != Code.Ldc_I8)
+					continue;
+				var l = (long)instruction.Operand;
+				if (l < int.MinValue || l > int.MaxValue)
+					continue;
+				ExpandMacro(instruction, OpCodes.Ldc_I4, unchecked((int)l));
+				self.Instructions.Insert(++i, Instruction.Create(OpCodes.Conv_I8));
```
samhouts: You're only going to iterate through every other instruction if you do `++i` here. Is that intentional?

StephaneDelcroix:
Yes. We insert an instruction so we 'i'

Xamarin.Forms.Build.Tasks/MethodBodyExtensions.cs
StephaneDelcroix: btw, this was part of a preview PR, and was only moved in this

PR #617, opened 2016-12-06
Allow subscriber to be collected if MessagingCenter is the only reference to it

hartez:
### Description of Change ###

If the target of a MessagingCenter callback is the subscriber, keep a weak reference to the subscriber to allow it to be garbage collected.

Also refactored MessagingCenter slightly because I got tired of miscounting the angle brackets.

### Bugs Fixed ###

- [45926 – MessagingCenter prevents subscriber from being collected](https://bugzilla.xamarin.com/show_bug.cgi?id=45926)

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


hartez:
@StephaneDelcroix This one is ready for another look.

PR #618, opened 2016-12-07
[UWP] Fix TextBox style for foreground focus color

pauldipietro:
### Description of Change ###

An issue exists with the Entry in UWP where the text color won't update while focused (the bug report referred to one of the Forms samples projects using a behavior to check against a numerical value); after some poking around it looks like the cause of this was stemming from the template where the value wasn't being applied on the Focused VisualState, but using it on the ContentElement itself allows things to function as normal.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=41054

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/FormsTextBoxStyle.xaml
+++ b/Xamarin.Forms.Platform.UAP/FormsTextBoxStyle.xaml
@@ -173,6 +173,7 @@
 						                  Foreground="{ThemeResource SystemControlForegroundBaseHighBrush}" FontWeight="Normal"
 						                  Margin="0,0,0,8" Grid.Row="0" Visibility="Collapsed" x:DeferLoadStrategy="Lazy" />
 						<ScrollViewer x:Name="ContentElement" AutomationProperties.AccessibilityView="Raw"
+                                      Foreground="{TemplateBinding ForegroundFocusBrush}"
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.UAP/FormsTextBoxStyle.xaml
```diff
@@ -173,6 +173,7 @@
 						                  Foreground="{ThemeResource SystemControlForegroundBaseHighBrush}" FontWeight="Normal"
 						                  Margin="0,0,0,8" Grid.Row="0" Visibility="Collapsed" x:DeferLoadStrategy="Lazy" />
 						<ScrollViewer x:Name="ContentElement" AutomationProperties.AccessibilityView="Raw"
+                                      Foreground="{TemplateBinding ForegroundFocusBrush}"
```
pauldipietro: The code on [L141-143](https://github.com/xamarin/Xamarin.Forms/blob/bad77d0ee4b9d3a2b642cbc3c35d87e848d5ece3/Xamarin.Forms.Platform.UAP/FormsTextBoxStyle.xaml#L141-L143) of this template seems like it could be removed and not make a difference, but I left it for the time being.

PR #619, opened 2016-12-07
[Core] Register renderers conditionally based on OS version

adrianknight89:
### Description of Change ###

On iOS, `ExportRendererAttribute` allows us to supply two different renderers for MDP on phones and tablets with `UIUserInterfaceIdiom` attribute parameter. We can build upon this to extend the behavior of the exporter such that different renderers can be utilized based on OS version.

We can now do the following:

```
[assembly:ExportRenderer(typeof(Entry), typeof(MyRenderer1), MaximumSdkVersion = 9)]
namespace fgfg789.iOS
{
	public class MyRenderer1 : EntryRenderer
	{
		public MyRenderer1()
		{
			; // Put a breakpoint here. It won't be hit because the renderer 
                          // has not been registered on iOS 10.
		}
	}
}

[assembly: ExportRenderer(typeof(Entry), typeof(MyRenderer2), MinimumSdkVersion = 10)]
namespace fgfg789.iOS
{
	public class MyRenderer2 : EntryRenderer
	{
		public MyRenderer2()
		{
			; // on iOS 10+, This renderer will be registered.
		}
	}
}
```

We can also use more than two renderers if we like so long as we make sure only one renderer is registered per version:

```
[assembly: ExportRenderer(typeof(Entry), typeof(MyRenderer1), MinimumSdkVersion = 6, MaximumSdkVersion = 7)]
[assembly: ExportRenderer(typeof(Entry), typeof(MyRenderer2), MinimumSdkVersion = 8, MaximumSdkVersion = 9]
[assembly: ExportRenderer(typeof(Entry), typeof(MyRenderer3), MinimumSdkVersion = 10)]
```

For further discussion, see [here](https://forums.xamarin.com/discussion/80341/methods-for-registering-renderers-during-runtime-instead-of-using-exportrenderer-attribute#latest).

Note that I do not know how to get OS version in RT (not sure if it's possible either) so I left it out.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/HandlerAttribute.cs
+++ b/Xamarin.Forms.Core/HandlerAttribute.cs
@@ -11,9 +11,9 @@ namespace Xamarin.Forms
 			HandlerType = handler;
 		}
 
-		internal Type HandlerType { get; private set; }
+		internal Type HandlerType { get; }
 
-		internal Type TargetType { get; private set; }
+		internal Type TargetType { get; }
 
 		public virtual bool ShouldRegister()
 		{
--- a/Xamarin.Forms.Platform.WinRT/ExportRendererAttribute.cs
+++ b/Xamarin.Forms.Platform.WinRT/ExportRendererAttribute.cs
@@ -9,10 +9,13 @@ namespace Xamarin.Forms.Platform.WinRT
 #endif
 {
 	[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
-	public sealed class ExportRendererAttribute : HandlerAttribute
+	public sealed class ExportRendererAttribute : BaseExportRendererAttribute
 	{
 		public ExportRendererAttribute(Type handler, Type target) : base(handler, target)
 		{
+#if WINDOWS_UWP
+			MajorVersion = Convert.ToInt32(WindowsDeviceInfo.GetOSVersion().Split('.')[0]);
+#endif
 		}
 	}
 
--- a/Xamarin.Forms.Platform.WinRT/WindowsDeviceInfo.cs
+++ b/Xamarin.Forms.Platform.WinRT/WindowsDeviceInfo.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using Windows.Foundation;
 using Windows.Graphics.Display;
+using Windows.System.Profile;
 using Windows.UI.Xaml;
 
 #if WINDOWS_UWP
@@ -109,6 +110,21 @@ namespace Xamarin.Forms.Platform.WinRT
 		void OnOrientationChanged(DisplayInformation sender, object args)
 		{
 			CurrentOrientation = GetDeviceOrientation(sender.CurrentOrientation);
+		}
+
+		internal static string GetOSVersion()
+		{
+#if WINDOWS_UWP
+			string deviceFamilyVersion = AnalyticsInfo.VersionInfo.DeviceFamilyVersion;
+			ulong version = ulong.Parse(deviceFamilyVersion);
+			ulong major = (version & 0xFFFF000000000000L) >> 48;
+			ulong minor = (version & 0x0000FFFF00000000L) >> 32;
+			ulong build = (version & 0x00000000FFFF0000L) >> 16;
+			ulong revision = version & 0x000000000000FFFFL; 
+			return $"{major}.{minor}.{build}.{revision}"; 
+#else
+			throw new NotSupportedException();
+#endif
 		}
 	}
 }
--- a/Xamarin.Forms.Platform.iOS/ExportRendererAttribute.cs
+++ b/Xamarin.Forms.Platform.iOS/ExportRendererAttribute.cs
@@ -4,9 +4,9 @@ using UIKit;
 namespace Xamarin.Forms
 {
 	[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
-	public sealed class ExportRendererAttribute : HandlerAttribute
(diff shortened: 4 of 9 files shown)


Xamarin.Forms.Core/HandlerAttribute.cs
```diff
@@ -5,19 +5,60 @@ namespace Xamarin.Forms
...
+
+			var isSupported = (bool)FilteredSdkVersionInformation[0];
+			var minimumSdkVersion = (int)FilteredSdkVersionInformation[1];
+			var maximumSdkVersion = (int)FilteredSdkVersionInformation[2];
+
+			if(minimumSdkVersion <= MajorVersion.Value && MajorVersion.Value <= maximumSdkVersion)
+				return isSupported;
+
+			// This means either the OS version has not originally fallen into the supplied range or 
+			// the OS version has changed and the supplied range is no longer valid.
```
adrianknight89: Not sure if we should throw an exception here or return true so that the renderer is registered.

Xamarin.Forms.Core/HandlerAttribute.cs
```diff
@@ -5,19 +5,60 @@ namespace Xamarin.Forms
 	[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
 	public abstract class HandlerAttribute : Attribute
 	{
+		protected HandlerAttribute(Type handler, Type target, object[] filteredSdkVersionInformation) : this(handler, target)
```
StephaneDelcroix: we don't do 'magic' `object[]` parameter in public API. it is likely to throw ~~on line 33~~ on `Validate()`, if the array doesn't have the right size, or the worn element types

Also, there's no need to add ctor parameters to the attribute. you can add properties, and those will be available as optional named param when you declare the attribute. That's the beauty, and weirdness, of attributes: attributes had optional parameters long before C# got it, and `DebuggerDisplayAttribute` had string interpolation long before c#6...

adrianknight89: I added a ctor parameter because `UIUserInterfaceIdiom` was added as such. Instead of `object[]`, what would be a proper data type? The reason why I added it in the first place was breaking up `{isSupported, minimumSdkVersion, maximumSdkVersion}` didn't make sense because `isSupported` didn't make sense on its own. Would it be okay if there were three properties (perhaps `isSupported can be named `isFilteredSdkVersionSupported`)?

adrianknight89:
@StephaneDelcroix Can you review again? I'll also work on tests.

adrianknight89:
@StephaneDelcroix Added tests as well.

Xamarin.Forms.Platform.WinRT/WindowsDeviceInfo.cs
```diff
@@ -110,5 +111,20 @@ void OnOrientationChanged(DisplayInformation sender, object args)
 		{
 			CurrentOrientation = GetDeviceOrientation(sender.CurrentOrientation);
 		}
+
+		internal static string GetOSVersion()
+		{
+#if WINDOWS_UWP
```
samhouts: We can get rid of this now.

Xamarin.Forms.Platform.WinRT/ExportRendererAttribute.cs
```diff
@@ -9,10 +9,13 @@ namespace Xamarin.Forms.Platform.WinRT
 #endif
 {
 	[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
-	public sealed class ExportRendererAttribute : HandlerAttribute
+	public sealed class ExportRendererAttribute : BaseExportRendererAttribute
 	{
 		public ExportRendererAttribute(Type handler, Type target) : base(handler, target)
 		{
+#if WINDOWS_UWP
```
samhouts: We can get rid of this now.

Xamarin.Forms.Platform.iOS/ExportRendererAttribute.cs
```diff
@@ -23,7 +23,10 @@ public ExportRendererAttribute(Type handler, Type target) : base(handler, target
 
 		public override bool ShouldRegister()
 		{
-			return !Idiomatic || Idiom == UIDevice.CurrentDevice.UserInterfaceIdiom;
+			if (!(!Idiomatic || Idiom == UIDevice.CurrentDevice.UserInterfaceIdiom))
```
samhouts: A little hard to parse this. Can we make it easier to understand without the double negative? :)

adrianknight89:
The same functionality needs to be implemented in the new projects (Tizen, WPF, GTK, etc.). Let me know if this feature can be merged so I can re-work it.

rmarinho:
@adrianknight89 yes we will be merging this, so if you can get this updated it would be great.

rmarinho:
any luck running the new stuff @adrianknight89  ?

adrianknight89:
Unfortunately no. At this point, I wouldn't mind a team member owning this. Most of the work is done already.

rookiejava:
In tizen, you can get th e platform version as follow : 
`Tizen.System.Information.TryGetValue<string>("tizen.org/feature/platform.version", out platformVersion);
`

For more details, please refer to [API Reference](https://developer.tizen.org/dev-guide/csapi/api/Tizen.System.Information.html#Tizen_System_Information_TryGetValue__1_System_String___0__) or [Developer Guide](https://developer.tizen.org/development/guides/.net-application/device-settings-and-systems/system-and-runtime-information).

PR #621, opened 2016-12-07
[iOS] Fix IsNavigationBarTranslucent platform-specific not working

jimmgarrido:
### Description of Change ###

Fixes setting `IsNavigationBarTranslucent = true` not working due to `EdgesForExtendedLayout` always being set to `UIRectEdge.None`.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=49104

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


jimmgarrido:
There is an empty area below the page content that appears when the navbar is translucent due to the container not being adjusted. #659 is a better fix as it does adjust the container as well.

PR #622, opened 2016-12-08
[XamlC] Allow compilation of IValueProviders

StephaneDelcroix:
### Description of Change ###

`IValueProvider`s tagged with the appropriate attribute will be bypassed
and the compiled version, if found, will be used.

This first version contains a compiled version of Setter's
IValueProvider and it already reduces the amount of generated IL by 39%
in, e.g. StyleTests.

It's a huge gain because XamlC no longer have to generate
ServiceProviders for those, so the methodbody is smaller, takes less
time to jit, less time to execute and nothing is invoked at runtime,
which probably saves a tons of time as well, as most IValueProvider
implementation heavily uses reflection.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/CompiledValueProviders/SetterValueProvider.cs
+++ b/Xamarin.Forms.Build.Tasks/CompiledValueProviders/SetterValueProvider.cs
@@ -1,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+
+using Xamarin.Forms.Xaml;
+using Xamarin.Forms.Build.Tasks;
+
+namespace Xamarin.Forms.Core.XamlC
+{
+	class SetterValueProvider : ICompiledValueProvider
+	{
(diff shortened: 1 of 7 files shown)


Xamarin.Forms.Build.Tasks/CompiledValueProviders/SetterValueProvider.cs
```diff
@@ -0,0 +1,40 @@
...
+			var bpRef = (new BindablePropertyConverter()).GetBindablePropertyFieldReference((string)bpNode.Value, module, bpNode);
+
+			TypeReference _;
+			var setValueRef = module.Import(module.Import(typeof(Setter)).GetProperty(p => p.Name == "Value", out _).SetMethod);
+
+			//push the setter
+			yield return Instruction.Create(OpCodes.Ldloc, vardefref.VariableDefinition);
+
+			//push the value
+			foreach (var instruction in ((ValueNode)valueNode).PushConvertedValue(context, bpRef, valueNode.PushServiceProvider(context, bpRef: bpRef), true, false))
```
samhouts: Small request: can we use named arguments for the two `bool`s at the end? :)

PR #623, opened 2016-12-08
Exception clarity in Xamarin.Forms.Platform.WinRT/VisualElementRenderer.cs.

buzzytom:
### Description of Change ###

Xamarin.Forms.Platform.WinRT/VisualElementRenderer.cs throws a vague NullReferenceException when SetNativeControl is called when the Element property is null. I'm new to GitHub (from Bitbucket) and the first time changing the Xamarin source. Does this require tests?

### Bugs Fixed ###

- None

### API Changes ###

Changed:
 - SetNativeControl in Xamarin.Forms.Platform.WinRT/VisualElementRenderer.cs.

### Behavioral Changes ###

Throws an Exception with custom message instead of a runtime induced NullReferenceException.

### PR Checklist ###

- [ ] Has tests (see description)

### DIFF
--- a/Xamarin.Forms.Platform.WinRT/VisualElementRenderer.cs
+++ b/Xamarin.Forms.Platform.WinRT/VisualElementRenderer.cs
@@ -325,7 +325,7 @@ namespace Xamarin.Forms.Platform.WinRT
 			Children.Add(control);
 
 			if (Element == null)
-				throw new InvalidOperationException("When calling SetNativeControl the Element property must have a non-null value.");
+				throw new InvalidOperationException("Changes to the native control require the Element property must have a non-null value.");
 
 			Element.IsNativeStateConsistent = false;
 			control.Loaded += OnControlLoaded;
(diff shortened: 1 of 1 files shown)


adrianknight89:
@buzzytom You can remove the curly braces since there is only one line under `if`. Also, I'm not sure about `Exception`. Maybe `InvalidOperationException` makes more sense. Remove `(of the ViewRenderer)`. You're throwing the exception from `VisualElementRenderer`.

buzzytom:
I have removed the braces and changed the exception.

adrianknight89:
Wouldn't the stack trace reveal the method name anyways?

buzzytom:
Yeah, which is why I think a more generic exception message would make sense.

StephaneDelcroix:
@buzzytom what are you doing to get that exception ? It shouldn't happen under normal circumstances, and if it does, we'd better fix the root cause...

StephaneDelcroix:
but thanks for your first PR anyway. the first one is always the harder one

jassmith:
Thank you for your PR. While this check does make this failure case more obvious, we prefer at this time to simply document the requirement. We have alerted the docs team to the need.

PR #518, opened 2016-11-09
[C] Track and clear CurrentNavigationTask value

pauldipietro:
### Description of Change ###

Per a couple reports it appears as if the `CurrentNavigationTask` value was not being cleared when using `PopAsync`.

Update: Added a baseline attempt at a counter and only clear the `CurrentNavigationTask` if the active navigation task count is at 0 and moved the other change to its own PR.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=31171

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #625, opened 2016-12-08
[C] Track and clear CurrentNavigationTask value

pauldipietro:
### Description of Change ###

Reopening in a new PR from #518 because force pushing to a branch after a PR is closed doesn't let you reopen it ☹️ 

This has some repetitive code put in a method (save for one spot where `t.Result` is returned) and the task count. Had to rebase it anyway.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=31171

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/NavigationPage.cs
+++ b/Xamarin.Forms.Core/NavigationPage.cs
@@ -61,6 +61,8 @@ namespace Xamarin.Forms
 
 		internal Task CurrentNavigationTask { get; set; }
 
+		internal int ActiveNavigationTaskCount { get; set; }
+
 		Stack<Page> INavigationPageController.StackCopy
 		{
 			get
@@ -114,12 +116,13 @@ namespace Xamarin.Forms
 
 		public async Task<Page> PopAsync(bool animated)
 		{
+			ActiveNavigationTaskCount++;
 			if (CurrentNavigationTask != null && !CurrentNavigationTask.IsCompleted)
 			{
 				var tcs = new TaskCompletionSource<bool>();
 				Task oldTask = CurrentNavigationTask;
 				CurrentNavigationTask = tcs.Task;
-				await oldTask;
+				await oldTask.ContinueWith((t) => ResolveActiveNavigation(t));
 
 				Page page = await ((INavigationPageController)this).PopAsyncInner(animated);
 				tcs.SetResult(true);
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Core/NavigationPage.cs
```diff
@@ -61,6 +61,8 @@ public Color Tint
 
 		internal Task CurrentNavigationTask { get; set; }
 
+		internal int ActiveNavigationTaskCount { get; set; } = 0;
```
adrianknight89: Isn't ` = 0;` redundant?

Xamarin.Forms.Core/NavigationPage.cs
```diff
@@ -114,12 +116,13 @@ public Task<Page> PopAsync()
...
 		public async Task<Page> PopAsync(bool animated)
 		{
+			ActiveNavigationTaskCount++;
 			if (CurrentNavigationTask != null && !CurrentNavigationTask.IsCompleted)
 			{
 				var tcs = new TaskCompletionSource<bool>();
 				Task oldTask = CurrentNavigationTask;
 				CurrentNavigationTask = tcs.Task;
-				await oldTask;
+				await oldTask.ContinueWith((t) => ResolveActiveNavigation(t));
```
adrianknight89: I think the signature can change to passing `oldTask` to a method and awaiting the method instead. Similar pattern can be observed in other cases.

Xamarin.Forms.Core/NavigationPage.cs
```diff
@@ -128,7 +131,12 @@ public Task<Page> PopAsync()
 
 			Task<Page> result = ((INavigationPageController)this).PopAsyncInner(animated);
 			CurrentNavigationTask = result;
-			return await result;
+			return await result.ContinueWith((t) =>
+			{
+				if (t.Status == TaskStatus.RanToCompletion)
+					UpdateActiveNavigationTaskCount();
+				return t.Result;
```
adrianknight89: In this case, we could also probably call `ResolveActiveNavigation` and then get the result.

Xamarin.Forms.Core/NavigationPage.cs
```diff
@@ -364,6 +381,13 @@ void InsertPageBefore(Page page, Page before)
 				Pushed(this, args);
 		}
 
+		Task ResolveActiveNavigation(Task task)
+		{
+			if (task.Status == TaskStatus.RanToCompletion)
```
adrianknight89: What happens if the [task status](https://msdn.microsoft.com/en-us/library/system.threading.tasks.taskstatus(v=vs.110).aspx) is not `RanToCompletion`?

pauldipietro: It can probably be removed outright because the value should be decremented regardless, yes.

pauldipietro:
Will come back to this later because the tests are failing for some reason and I don't think it's perfect right now.

PR #626, opened 2016-12-09
Deprecate iOS 8 and earlier

pauldipietro:
### Description of Change ###

Deprecating iOS 8 and earlier by removing usage of `IsiOS7OrNewer` `IsiOS8OrNewer` (and since we're now always on iOS 9+, `IsiOS9OrNewer` shouldn't be needed either).

You were good operating system versions, but alas, your time has come. RIP.

### Bugs Fixed ###

N/A

### API Changes ###

N/A

### Behavioral Changes ###

Stuff not on iOS 9 or newer can and will likely blow up.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/ContextActionCell.cs
+++ b/Xamarin.Forms.Platform.iOS/ContextActionCell.cs
@@ -169,9 +169,7 @@ namespace Xamarin.Forms.Platform.iOS
 				_scroller = new UIScrollView(new RectangleF(0, 0, width, height));
 				_scroller.ScrollsToTop = false;
 				_scroller.ShowsHorizontalScrollIndicator = false;
-
-				if (Forms.IsiOS8OrNewer)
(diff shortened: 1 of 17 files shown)


adrianknight89:
This is great. Now we just need to move to Android 21+.

rmarinho:
IOS8 tests are failing at the time of merge but that's normal since we are removing iOS8 support.

AntRemo:
Is it possible this refactor was incorrect?
https://github.com/xamarin/Xamarin.Forms/pull/626/files/764c1736a62b72e7d2497c363fd37318bcba943d#diff-71148fd2814c3d17185db42cc58913ffR14

Should this change be 
`return base.GetDesiredSize(widthConstraint, heightConstraint); 
`

or just remove the override entirely.

Thanks
@rmarinho @StephaneDelcroix  @pauldipietro

PR #524, opened 2016-11-12
[iOS/Critical] Fix ListView memory leaks

adrianknight89:
### Description of Change ###

iOS `ListView` is leaking for various reasons. Although the bug description mentions issues with `RecycleElement`, this applies to either caching strategy. Since this seems to be a serious issue, can this PR be given priority?

See comments for details.

P.S. It was mentioned that the same issue occurs on Android, but I wasn't able to reproduce it on AppCompat (at least not with the sample code in this PR).

EDIT: Also included the fix for another bug concerning context actions not being properly disposed.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=43941
- https://bugzilla.xamarin.com/show_bug.cgi?id=32206
- https://bugzilla.xamarin.com/show_bug.cgi?id=27194 (possibly fixes this too. Download links seem to be down.)

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #632, opened 2016-12-10
[iOS] Dispose pickers properly and automate GC checks

adrianknight89:
### Description of Change ###

Went through renderers for `Picker`, `TimePicker`, and `DatePicker` to add/enhance dispose methods. Also added a UI test to automate GC checks. This should be run on top of #524.

### Bugs Fixed ###

- N/A

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/DatePickerRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/DatePickerRenderer.cs
@@ -22,6 +22,7 @@ namespace Xamarin.Forms.Platform.iOS
 	{
 		UIDatePicker _picker;
 		UIColor _defaultTextColor;
+		bool _disposed;
 
 		IElementController ElementController => Element as IElementController;
 
@@ -29,7 +30,10 @@ namespace Xamarin.Forms.Platform.iOS
 		{
 			base.OnElementChanged(e);
 
-			if (e.OldElement == null)
+			if (e.NewElement == null)
+				return;
+
+			if (Control == null)
 			{
 				var entry = new NoCaretField { BorderStyle = UITextBorderStyle.RoundedRect };
 
@@ -55,13 +59,10 @@ namespace Xamarin.Forms.Platform.iOS
 				SetNativeControl(entry);
 			}
 
-			if (e.NewElement != null)
-			{
-				UpdateDateFromModel(false);
-				UpdateMaximumDate();
-				UpdateMinimumDate();
-				UpdateTextColor();
-			}
+			UpdateDateFromModel(false);
+			UpdateMaximumDate();
+			UpdateMinimumDate();
+			UpdateTextColor();
 		}
 
 		protected override void OnElementPropertyChanged(object sender, PropertyChangedEventArgs e)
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Platform.iOS/Renderers/DatePickerRenderer.cs
```diff
@@ -22,14 +22,18 @@ public class DatePickerRenderer : ViewRenderer<DatePicker, UITextField>
...
 		{
 			base.OnElementChanged(e);
 
-			if (e.OldElement == null)
+			if (e.NewElement == null)
+				return;
+
+			if (Control == null)
 			{
 				var entry = new NoCaretField { BorderStyle = UITextBorderStyle.RoundedRect };
```
StephaneDelcroix: this will re-create an entry if the renderer is reused, instead of re-using it

StephaneDelcroix: scratch that, sorry, the `Control == null` prevents this

Xamarin.Forms.Platform.iOS/Renderers/PickerRenderer.cs
```diff
@@ -138,13 +139,51 @@ void UpdateTextColor()
...
+				}
+
+				if (Control != null)
+				{
+					Control.EditingDidBegin -= OnStarted;
+					Control.EditingDidEnd -= OnEnded;
+				}
+
+				if(Element != null)
+					((INotifyCollectionChanged)Element.Items).CollectionChanged -= RowsCollectionChanged;
```
adrianknight89: @StephaneDelcroix I'm not 100% sure about this one. In a working situation, `oldElement` and `newElement` should be changed in a balanced way. When the object is being disposed, `newElement` should be set to null and `OnElementChanged` should be fired. This should unregister the handler because `oldElement` is not null.

This is just extra measure in case things don't work. Perhaps it makes sense to not do this and potentially end up with a leak so it could be investigated further and the root cause can be fixed.

adrianknight89:
@jassmith  Some renderers seem to be calling `base.OnElementChanged(e)` at the end whereas most call it at the beginning. I'd like to see a consistent approach. So maybe you guys could discuss this internally and make necessary adjustments across solution if needed.

Xamarin.Forms.Platform.iOS/Renderers/TimePickerRenderer.cs
```diff
@@ -10,17 +10,34 @@ public class TimePickerRenderer : ViewRenderer<TimePicker, UITextField>
...
 			if (disposing)
 			{
-				Control.EditingDidBegin -= OnStarted;
-				Control.EditingDidEnd -= OnEnded;
+				_defaultTextColor = null;
 
-				_picker.ValueChanged -= OnValueChanged;
+				if (_picker != null)
+				{
+					_picker.RemoveFromSuperview();
```
rmarinho: This is done on the tracker or packager i think..

adrianknight89: Where is it done exactly?

StephaneDelcroix:
Rebased. waiting for ui tests

PR #634, opened 2016-12-10
[UWP] Fix Navigate/Switch to exist page performance and lost state problem in real phone

huangjinshe:
### Description of Change ###

When we navigate to some page and navigate to other page and back or switch exist Details page, or change App.MainPage to exist page, the page last state fill lost, like ScrollViewer position...etc, and also the navigate/switch speed still very slow. This will fix that problem.
(Note: **Navigate animation unable to use currently after apply this**, but I think temporarily it's fine, the navigate animation too ugly, I think no one will use animation for now)

The basic idea is that use Canvas, put pages to the canvas, and when navigate to new page, hide the old page, navigate back just show it again. It will  more faster navigate or switch to the exist page now, especially when the page has a a lot of controls or ListView, Because the Phone is using ARM, it would be a big performance problem if every time navigate/switch will unload and load exist page such like a new page. So I just change the idea

Currently improved type of pages:
1.App.MainPage
2.NavigationPage
3.MasterDetailPage

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=46856
https://bugzilla.xamarin.com/show_bug.cgi?id=48682
https://bugzilla.xamarin.com/show_bug.cgi?id=38260

### API Changes ###

None

Added:
 - Xamarin.Forms.Platform.UWP.MasterDetailControl Add Detail , its' a  Canvas. also add internal method CheckContentIfExist
 - Xamarin.Forms.Platform.UWP.MasterDetailControl Add Detail , its' a  Canvas. also add internal method CheckContentIfExist
 - Xamarin.Forms.Platform.WinRT/PageControl.xaml Add Canvas in the content, replace Content and add internal method CheckContentIfExist

Changed:
 - Xamarin.Forms.Platform.UWP.MasterDetailControl Detail rename to DetailContent
 - Xamarin.Forms.Platform.UAP/MasterDetailPageRenderer change logic: do not clear the page every time switch pages(for that manually fire SendDisappearing and SendAppearing)
 - Xamarin.Forms.Platform.WinRT/NavigationPageRenderer.cs change logic: do not clear the page every time switch pages(for that manually fire SendDisappearing and SendAppearing)
 - Xamarin.Forms.Platform.WinRT/Platform change logic: do not clear the page every time switch pages

### Behavioral Changes ###

In the original mode: every time when use these:
App.MainPage = existPage;
NavigationPage.push(existPage) or pop();
MasterDetailPage.Detail = existPage;

it will always clean and dispose the old page, when user want to switch back the last page, it need to reload all things, that's a really bad design, **we should recommend developer always should use exist page more now**, because every open or display a page unload and load it will cause to much performance, also last page state will lose(Like ListView or ScrollViever scroll position will reset)

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/VisualElement.cs
+++ b/Xamarin.Forms.Core/VisualElement.cs
@@ -791,7 +791,8 @@ namespace Xamarin.Forms
 
 			if (!GetIsDefault(TriggersProperty)) {
 				var triggers = GetValue(TriggersProperty) as AttachedCollection<Trigger>;
-				triggers.DetachFrom(this);
+				if (null != triggers)
+                    triggers.DetachFrom(this);
 			}
 		}
 	}
--- a/Xamarin.Forms.Platform.UAP/MasterDetailControl.cs
+++ b/Xamarin.Forms.Platform.UAP/MasterDetailControl.cs
@@ -4,6 +4,7 @@ using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Media;
 using Xamarin.Forms.PlatformConfiguration.WindowsSpecific;
+using System.Linq;
 
 namespace Xamarin.Forms.Platform.UWP
 {
@@ -14,8 +15,8 @@ namespace Xamarin.Forms.Platform.UWP
 
 		public static readonly DependencyProperty MasterTitleProperty = DependencyProperty.Register("MasterTitle", typeof(string), typeof(MasterDetailControl), new PropertyMetadata(default(string)));
 
-		public static readonly DependencyProperty DetailProperty = DependencyProperty.Register("Detail", typeof(FrameworkElement), typeof(MasterDetailControl),
-			new PropertyMetadata(default(FrameworkElement)));
+		public static readonly DependencyProperty DetailContentProperty = DependencyProperty.Register("DetailContent", typeof(Canvas), typeof(MasterDetailControl),
+			new PropertyMetadata(default(Canvas)));
 
 		public static readonly DependencyProperty IsPaneOpenProperty = DependencyProperty.Register("IsPaneOpen", typeof(bool), typeof(MasterDetailControl), new PropertyMetadata(default(bool)));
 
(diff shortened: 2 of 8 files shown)


huangjinshe:
In this PR, I also fixed a BUG Xamarin.Forms\Xamarin.Forms.Core\VisualElement

from:
```

if (!GetIsDefault(TriggersProperty)) {
	var triggers = GetValue(TriggersProperty) as AttachedCollection<Trigger>;
	triggers.DetachFrom(this);
}
```

to:
```

if (!GetIsDefault(TriggersProperty)) {
	var triggers = GetValue(TriggersProperty) as AttachedCollection<Trigger>;
	if (null != triggers)
            triggers.DetachFrom(this);
}

```

That will cause : `System.Runtime.InteropServices.SEHException` before.

huangjinshe:
Also I think maybe this PR will resolve some memory leak problems. before every time navigate to exist page (because original design is unload and reload it again), there is a memory issue, specially there are a ListView, but after we changed now, we don't need unload and load again, it will no more cause that problem(at least the memory leak not so seriously).

huangjinshe:
**Update:** Fix when phone orientation changed, the page not display correctly,
Because the new design, sometimes when the phone changed the orientation, the last page and currently page will display together, also the width and height not recalculate again.

Team, please give me any response about this change. It would be very great you guys could release a beta, get some feedback from other developers.

huangjinshe:
Last commit should be fix Bug: https://bugzilla.xamarin.com/show_bug.cgi?id=41843

huangjinshe:
Last commit should be fix:
When App.MainPage = old exist navigation page, the back button will not show.(my new design problem)
Also I fixed when on UWP desktop version the Master split button not display problem.(old BUG, details:
https://forums.xamarin.com/discussion/comment/235782/#Comment_235782)

jassmith:
Hi!

This proposal touches on another idea we have been considering for a long time. Unfortuantely we cant take this as implemented because it would cause memory leaks for some usage patterns. However I think the proposal we have put forth solves your issue in a much more elegant and extensible way.

https://forums.xamarin.com/discussion/84504/proposal-configurable-retention-of-contentpage-renderers

huangjinshe:
@jassmith ,No, I don't agree with you, currently the exist page load too slow on UWP, not only because the previous page disposed when navigate to other, Also because this:
`
ContentPage.Content = renderer.ContainerElement;

`
I don't really get it why it too slow, (in my case my page have a lot of control(Maybe ListView), it's load once already(slow enough), and navigate to it again, this will take at least 2 seconds, even I'm not disposed the page). That's the exactly reason why I change the `Content` to a `Canvas`.

I'm do this change not for cause memory leak, it's for fix memory leak, Don't you think currently design will case memory leak already? If you not, please more test, maybe you should download some sample from Bugzilla. Yes, I know what you considering, Of course if you want add:
`
contentPage.SetRetainsRenderer(true)

`
I could check the previous page, and decide clean it or not by that. But, I think `Canvas` must be leave in here as `Content`, because as I said before every time you set the `Content`, it will too slow for show the page. I could also check if you want to clean the previous page, also remove it from `Canvas`. you only mention a property, but you shouldn't close the all design idea here, or you could please fix and make a better design here, because we can't take this shit really. My customers already blame enough how the our project worse every day!

Final thing: did you just ignore other Bugs which I fixed??  you guys take a half year not fixed, and when I fixed by myself  you just ignore the hole thing??

huangjinshe:
I explain use another way: I have an exist page will take 5 seconds now(based currently Xamarin.Forms design), If only remove clean page or your idea(actually same), it would still need 2.5 seconds around. That's not a complete solution for fixing performance problem.

PR #636, opened 2016-12-12
[UWP] Allow for proper dismissal of soft keyboard

pauldipietro:
### Description of Change ###

As mentioned in the below bug report, a page with entries on it suffer from a situation where the soft keyboard can't be dismissed. This can be resolved by utilizing the `InputPane` class to show or hide the soft keyboard when the focus changes. The soft keyboard is also utilized in situations such as tablet mode on devices like the Surface.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=45067

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


pauldipietro:
The intention here seems to have been a bit mislead so this is just being closed for now.

PR #637, opened 2016-12-12
[Core] Account for TranslationX/Y when laying out children into bounding region

adrianknight89:
### Description of Change ###

In the below bug description, the user is having an issue when a `ViewCell` is scrolled off the screen, its `View` loses `TranslationX`. The change I made fixes this issue; however, I'm not 100% sure about the change as the layouting system is still a black box to me.

The sample code should demonstrate the solution. You can also give `PosX` a starting value so the cells are initially laid out with a non-zero `TranslationX`.

Note that this is a change in Core, and I do not know how to unit test it.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=44690

### PR Checklist ###


StephaneDelcroix:
translationX and Y properties are never handled in Core, but in the different platforms (VisualElementTracker)

that's probably were you should look at. trying to figure out why the view isn't displayed with the correct translationX.

jassmith:
The fix for this issue needs to be done in the renderers. Moving things here does a pre layout transform instead of a post layout transform.

PR #641, opened 2016-12-13
[iOS] Register effects for ScrollView and WebView

adrianknight89:
### Description of Change ###

On iOS, `ScrollView` and `WebView` can't attach effects. This seems to be due to the fact that `SetElement` from `IVisualElementRenderer` is implemented and it does not register effects. 

Note that any renderer that does not derive from `ViewRenderer` has another problem which is that `ViewRenderer`'s `OnRegisterEffect` is not called and `Control` is not set. Currently, **page renderers** implement `IVisualElementRenderer` and they only set `Container` to `View`. Was this by design (i.e. they do not have `Control`)?

`ScrollViewRenderer` and `WebViewRenderer` are the only **control renderers** that implement the same interface. For these, I'm not exactly sure what to set `Container` and `Control`. Right now, they are set to `this` and `NativeView` respectively.

I looked at Android. It also has the same issues with several renderers, but Android confuses me even more when it comes to figuring out the values for `Container` and `Control`, so this is something you should look into.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=49648
- https://bugzilla.xamarin.com/show_bug.cgi?id=45874

### API Changes ###

Changed:
 - `IEffectControlProvider` is part of `IVisualElementRenderer`
 - Moved `EffectUtilities` into Core.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/EffectUtilities.cs
+++ b/Xamarin.Forms.Core/EffectUtilities.cs
@@ -1,0 +1,16 @@
+namespace Xamarin.Forms
+{
+	internal static class EffectUtilities
+	{
+		public static void RegisterEffectControlProvider(IEffectControlProvider self, IElementController oldElement, IElementController newElement)
+		{
+			IElementController controller = oldElement;
+			if (controller != null && controller.EffectControlProvider == self)
+				controller.EffectControlProvider = null;
+
+			controller = newElement;
+			if (controller != null)
+				controller.EffectControlProvider = self;
+		}
+	}
+}
--- a/Xamarin.Forms.Platform.iOS/IVisualElementRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/IVisualElementRenderer.cs
@@ -3,7 +3,7 @@ using UIKit;
 
 namespace Xamarin.Forms.Platform.iOS
 {
-	public interface IVisualElementRenderer : IDisposable, IRegisterable, IEffectControlProvider
+	public interface IVisualElementRenderer : IDisposable, IRegisterable
 	{
 		VisualElement Element { get; }
 
--- a/Xamarin.Forms.Platform.iOS/IVisualElementRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/IVisualElementRenderer.cs
@@ -3,7 +3,7 @@ using UIKit;
 
 namespace Xamarin.Forms.Platform.iOS
 {
-	public interface IVisualElementRenderer : IDisposable, IRegisterable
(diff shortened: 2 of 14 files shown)


Xamarin.Forms.Core/EffectUtilities.cs
```diff
@@ -0,0 +1,16 @@
+namespace Xamarin.Forms
+{
+	internal static class EffectUtilities
```
adrianknight89: This can be moved to Core. Android applies the same functionality.

Xamarin.Forms.Platform.iOS/VisualElementRenderer.cs
```diff
@@ -73,6 +73,16 @@ protected bool AutoTrack
 			}
 		}
 
+		public static void RegisterEffect(Effect effect, UIView container, UIView control = null)
```
adrianknight89: Since `RegisterEffect` is the same across all renderers, this can be refactored into its own method.

Xamarin.Forms.Platform.iOS/IVisualElementRenderer.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Xamarin.Forms.Platform.iOS
 {
-	public interface IVisualElementRenderer : IDisposable, IRegisterable
+	public interface IVisualElementRenderer : IDisposable, IRegisterable, IEffectControlProvider
```
adrianknight89: Should any `VisualElement` have the ability to attach effects? I thought this would be okay. If not, this could be moved back.

Xamarin.Forms.Core/EffectUtilities.cs
rmarinho: We have a internals namespace, maybe move it there.

Xamarin.Forms.Platform.iOS/IVisualElementRenderer.cs
rmarinho: I think that will break  existing renderers from other people ..

jassmith: because if you made a renderer using the IVisualElementRenderer directly, your code would no longer compile with this change.

adrianknight89: Removed.

LeRondPoint:
Thanks for the initiative, just hit this bug.  Hopefully it will be pulled in soon.

PR #643, opened 2016-12-14
Update NameScope.cs

JeanCollas:
### Description of Change ###

Debug message corrected (parameter named passed instead of parameter value).

### Bugs Fixed ###

```
Exception:

System.ArgumentException: An element with the same key already exists in NameScope
Parameter name: name

```

### PR Checklist ###

Debug message corrected

### DIFF
--- a/Xamarin.Forms.Core/Internals/NameScope.cs
+++ b/Xamarin.Forms.Core/Internals/NameScope.cs
@@ -21,7 +21,7 @@ namespace Xamarin.Forms.Internals
 		void INameScope.RegisterName(string name, object scopedElement)
 		{
 			if (_names.ContainsKey(name))
-				throw new ArgumentException("An element with the same key already exists in NameScope", "name");
+				throw new ArgumentException($"An element with the same key '{name}' already exists in NameScope", name);
 
 			_names[name] = scopedElement;
 		}


StephaneDelcroix:
Thanks for this PR Jean. This change unfortunately doesn't bring much value. closing

JeanCollas:
Are you serious? What is a PR with value? Are you going to let the bug for the next releases?

StephaneDelcroix:
what's the bug ? The 2 argument of the ArgumentException ctor must be the name of the faulty parameter

https://msdn.microsoft.com/en-us/library/sxykka64(v=vs.110).aspx

and that's what we do.

`Dictionary<TKey,TValue>.Add()` also throws `System.ArgumentException: An item with the same key has already been added.` when the key is a duplicate.

JeanCollas:
Right, then it is the behavior and the exception thrown that should be adjusted: it not consistent with the method just below:

			try
			{
				((INameScope)this).RegisterName(name, scopedElement);
			}
			catch (ArgumentException)
			{
				throw new XamlParseException(string.Format("An element with the name \"{0}\" already exists in this NameScope", name), xmlLineInfo);
			}

The ArgumentException should be caught and rethrown in a XamlParseException to explain what happens.

djcparker:
Knowing the name of the key that caused the exception can be crucial for debugging.   I personally hate exception  messages that are vague.

PR #499, opened 2016-11-02
[UWP] Add platform specifics for SearchBar's suggestion area

pauldipietro:
### Description of Change ###

Our current implementation of the `SearchBar` on UWP uses the [AutoSuggestBox](https://msdn.microsoft.com/library/windows/apps/Windows.UI.Xaml.Controls.AutoSuggestBox), which is the current standard. However, it does not take advantage of the dropdown that is available, effectively treating it as an empty search entry. This has a side effect of causing the bar to scroll up to the top of the screen if the suggestion area would not fit. The reason it always shows as empty is due to a lack of suggestions being set as the `ItemsSource`.

To rectify this potentially awkward behavior, a platform specific feature can be added which defaults the `AutoSuggestBox` control's `AutoMaximizeSuggestionArea` to false, so that it simply does not scroll on the screen when it would otherwise occur. We can also extend upon the behavior and allow for developers to take advantage of this suggestion area functionality if desired by providing a `BindableProperty` (appropriately) called `SuggestionsProperty`, which is of the type `ObservableCollection<string>`. Users can use `On<Windows>.Suggestions()` to operate on this collection and modify it directly, as needed.

In order to do this, we have a `TextChangedAction` of type `Action` that allows the user to handle this to their liking, which gets used in the `OnTextChanged` method in the `SearchBarRenderer` if the action has been explicitly set by the user (it is otherwise null). The `SuggestionChosen` event is handled inside the renderer and set the text is set to whatever has been chosen by the user.

searchbar

searchbar2

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=45924

### API Changes ###

Added:
- public static readonly BindableProperty SuggestionsProperty
- public static ObservableCollection<string> GetSuggestions(BindableObject element)
- public static void SetSuggestions(BindableObject element, ObservableCollection<string> collection)
- public static ObservableCollection<string> Suggestions(this IPlatformElementConfiguration<Windows, FormsElement> config)

- public static readonly BindableProperty TextChangedActionProperty
- public static Action GetTextChangedAction(BindableObject element)
- public static Action GetTextChangedAction(this IPlatformElementConfiguration<Windows, FormsElement> config)
- public static IPlatformElementConfiguration<Windows, FormsElement> SetTextChangedAction(this IPlatformElementConfiguration<Windows, FormsElement> config, Action value)

### Behavioral Changes ###

None presumed, as by default the behavior is brought more in line with the other platforms.

### PR Checklist ###

[...]

PR #646, opened 2016-12-14
[UWP] Set SearchBar's AutoMaximizeSuggestionArea to false

pauldipietro:
### Description of Change ###

The feature in #499 was deemed as something that be moved into an effect, but there is one element of it that is actually a fix via setting `AutoMaximizeSuggestionArea` to false to prevent the SearchBar from shifting up awkwardly to accommodate a suggestion area which doesn't exist (explained and displayed in that PR).

(I had been meaning to get around to extracting this)

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=45924

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


pauldipietro:
Can we merge this?

PR #649, opened 2016-12-15
Update NameScope.cs throwing more precise except

JeanCollas:
### Description of Change ###

Update to give a more comprehensive exception and a consistent behavior among methods (same kind of exception thrown, with relevant information for developers)

### Bugs Fixed ###

The following exception message is not very helpful for developers:

    Exception:
    
    System.ArgumentException: An element with the same key already exists in NameScope
    Parameter name: name

### API Changes ###

Changes the type of Exception thrown by the method INameScope.RegisterName(string name, object scopedElement)

### PR Checklist ###

Update to give a more comprehensive exception and a consistent behavior among methods (same kind of exception thrown, with relevant information for developers)

Adjustment to the PR https://github.com/xamarin/Xamarin.Forms/pull/643

### DIFF
--- a/Xamarin.Forms.Core/Internals/NameScope.cs
+++ b/Xamarin.Forms.Core/Internals/NameScope.cs
@@ -21,7 +21,7 @@ namespace Xamarin.Forms.Internals
 		void INameScope.RegisterName(string name, object scopedElement)
 		{
 			if (_names.ContainsKey(name))
-				throw new ArgumentException("An element with the same key already exists in NameScope", "name");
+                		throw new XamlParseException(string.Format("An element with the name \"{0}\" already exists in this NameScope", name), new XmlLineInfo());
 
 			_names[name] = scopedElement;
 		}
@@ -32,7 +32,7 @@ namespace Xamarin.Forms.Internals
 			{
 				((INameScope)this).RegisterName(name, scopedElement);
 			}
-			catch (ArgumentException)
+			catch (XamlParseException)
 			{
 				throw new XamlParseException(string.Format("An element with the name \"{0}\" already exists in this NameScope", name), xmlLineInfo);
 			}
--- a/Xamarin.Forms.Xaml/RegisterXNamesVisitor.cs
+++ b/Xamarin.Forms.Xaml/RegisterXNamesVisitor.cs
@@ -31,17 +31,7 @@ namespace Xamarin.Forms.Xaml
 		{
 			if (!IsXNameProperty(node, parentNode))
 				return;
-			try
-			{
-				((IElementNode)parentNode).Namescope.RegisterName((string)node.Value, Values[parentNode]);
-			}
-			catch (ArgumentException ae)
-			{
-				if (ae.ParamName != "name")
-					throw ae;
-				throw new XamlParseException(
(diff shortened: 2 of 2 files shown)


StephaneDelcroix:
I'll need a use case of what you're doing to get that exception, and it that context I'll have to understand why the exception you get is not sufficient

this is a breaking change

JeanCollas:
Hello @StephaneDelcroix 

I got this exception just because I had copy/duplicated a part of XAML that had a name defined, in a "normal" page. 
The exception that we get in result is not developer-friendly at all, as we have to browse the whole xaml to find the duplicated key. 
It happened to me many times, I guess I am not the only one, and it is a quite frustrating message with no extra information. That kind of message is a reason why I "could" say I dislike Xamarin. Yes debugging xamarin is sometimes painful, no need to add extra difficult bits when not necessary. 

One of the first objectives of Xamarin is to reduce the development time, I think this PR is exactly in that direction.

Anyway the current behavior is not optimal as the check is performed outside the method that known what is wrong, and it adds an additional try/catch which could be avoided.

StephaneDelcroix:
Please open a bug report for this, and point me to it, and I'll come up with a proper solution.

JeanCollas:
It is actually not a single issue, it is a big issue of the framework.
Today I got these, what do you do with this information?

```
Thread finished: <Thread Pool> #4
Exception:

System.ArgumentNullException: Value cannot be null.
Parameter name: name

The thread 'Unknown' (0x4) has exited with code 0 (0x0).
Exception:

System.ArgumentNullException: Value cannot be null.
Parameter name: name

Exception:

System.ArgumentNullException: Value cannot be null.
Parameter name: name

Exception:

System.ArgumentNullException: Value cannot be null.
Parameter name: name

Exception:

System.ArgumentNullException: Value cannot be null.
Parameter name: name

Thread finished: <Thread Pool> #3
Exception:

System.ArgumentNullException: Value cannot be null.
Parameter name: name

The thread 'Unknown' (0x3) has exited with code 0 (0x0).
Thread finished: <Thread Pool> #5
Exception:

System.ArgumentNullException: Value cannot be null.
Parameter name: name

The thread 'Unknown' (0x5) has exited with code 0 (0x0).
Thread finished: <Thread Pool> #11
Exception:

System.ArgumentNullException: Value cannot be null.
Parameter name: name

The thread 'Unknown' (0xb) has exited with code 0 (0x0).
Exception:

System.ArgumentNullException: Value cannot be null.
Parameter name: name

Thread finished: <Thread Pool> #9
The thread 'Unknown' (0x9) has exited with code 0 (0x0).
Thread started: <Thread Pool> #12
```

Ok, you can break and sometimes get more details by searching deep into the innerexceptions or stacktraces, but why are they not displayed when available?

StephaneDelcroix:
@JeanCollas this is not the right forum to share the issues you have with Xamarin.Forms. Please fill bug reports with simple projects on how to reproduce the issue you're facing. Point me to them, and I'll look at them. We know that our error reporting can be enhanced, and this looks like a nice opportunity.

StephaneDelcroix:
we already have test for this https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Xaml.UnitTests/Issues/Issue2450.xaml.cs#L31 
that's why a new bug report is required

PR #653, opened 2016-12-16
Remove license comments

adrianknight89:
I'm not strong with licenses, but now that XF is open source, is this needed?


StephaneDelcroix:
Yes, it is, for the following reason (as stated in the licence itself):

> // The above copyright notice and this permission notice shall be included in
> // all copies or substantial portions of the Software.

You'll find some other copyright notices here and there. Everything has been reviewed at the time the code was open-sourced, and the one remaining should remains.

PR #654, opened 2016-12-16
Enable 24-hr formatting for Android TimePicker

hardcodet:
### Description of Change ###

The renderer for the Android time picker opens the time picker dialog with a hardcoded 12-hr setting despite the fact the the platform allows to determine whether 24 hr time formatting was configured or not.

### Bugs Fixed ###

Time picker should respect the 12hr/24hr setting of the user on Android.

### PR Checklist ###

Enable 24-hr time formatting depending on Android system settings instead of the hard coded 12-hour setting.

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/TimePickerRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/TimePickerRenderer.cs
@@ -3,6 +3,7 @@ using System.ComponentModel;
 using Android.App;
 using Android.Content.Res;
 using Android.Widget;
+using Android.Text.Format;
 using ADatePicker = Android.Widget.DatePicker;
 using ATimePicker = Android.Widget.TimePicker;
 using Object = Java.Lang.Object;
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.Android/Renderers/TimePickerRenderer.cs
```diff
@@ -92,7 +93,8 @@ void OnClick()
 			TimePicker view = Element;
 			ElementController.SetValueFromRenderer(VisualElement.IsFocusedPropertyKey, true);
 
-			_dialog = new TimePickerDialog(Context, this, view.Time.Hours, view.Time.Minutes, false);
+			bool is24HourFormat = DateFormat.Is24HourFormat(Context);
```
adrianknight89: Instead of creating a local variable, you could pass the method call in the constructor.

hardcodet: I could, but how would that improve readability of the code?

adrianknight89: I suppose it's a matter of preference. You're calling a method named `Is24HourFormat` and declaring a variable with the same name. From the method name , we could infer the same thing.

hardcodet: I was never a fan of nesting method calls within method calls, for reasons of readability and ease of debugging. I don't have strong feeling here though, it's not my code after all - I can change this tomorrow.

adrianknight89: I don't have strong feeling either. Perhaps the actual reviewers can comment.

StephaneDelcroix:
I *think* we have one or multiple issues open for this, and we should dig that out

PR #656, opened 2016-12-19
Fixed property name

LWChris:
Not sure if wrong or intended. This edit was proposed just by looking at the surrounding code. It seems like a typo.

The unit tests suggest this was a working fixture in spite of the name being seemingly wrong. I just want to discuss the code, not actually make a PR, but GitHub won't let me.

Please verify if this breaks or fixes something, and if the latter, why the unit tests for the FactoryMethods work although the name was wrong.

### Description of Change ###

The name of the static XmlName object for the FactoryMethod property seems to contain the x prefix of the namespace, too. I removed the x.

Maybe the Xaml parser breaks for x:FactoryMethod attribute (although I'd rather expect it to not have worked before)?

No additional unit test because the existing ones should cover.

### Bugs Fixed ###

I don't see an issue tracker here, so none?

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Xaml/XmlName.cs
+++ b/Xamarin.Forms.Xaml/XmlName.cs
@@ -10,7 +10,7 @@ namespace Xamarin.Forms.Xaml
 		public static readonly XmlName xName = new XmlName("x", "Name");
 		public static readonly XmlName xTypeArguments = new XmlName("x", "TypeArguments");
 		public static readonly XmlName xArguments = new XmlName("x", "Arguments");
-		public static readonly XmlName xFactoryMethod = new XmlName("x", "xFactoryMethod");
+		public static readonly XmlName xFactoryMethod = new XmlName("x", "FactoryMethod");
 		public static readonly XmlName xDataType = new XmlName("x", "DataType");
 		public static readonly XmlName Empty = new XmlName();
 


StephaneDelcroix:
Thanks for this first PR @LWChris. This is not a bug, that string is a magic string only used internally by the parser to maintain state. It could as well be
```
public static readonly XmlName xFactoryMethod = new XmlName("x", "🌈🏢🦅");
```

(I'm wondering if actually it shouldn't be...)

So it's not a bug, and that's something we won't change.

For your next PRs (they are welcome), do not start by looking at the code in search of a bug. Most of the code is there (or was there) for a reason. Instead, start with an actual bug you can reproduce and fix it.

LWChris:
Hi @StephaneDelcroix, I was searching for the x:ClassModifier, so I stumbled upon the code.  Being a programmer, when I see something that looks like wrong code, I ask about it. Git doesn't seem to provide a "comment" function for merged code like it does for commits.

> So it's not a bug, and that's something we won't change.

Well, I'd probably still fix it... We all know, if it stays like that, then in [insert any timespan here] someone will use this very static members' Name properties to retrieve the respective attribute names, and it will work for all tags except for `FactoryMethod`, which will not be unit tested etc. etc.
I've seen programs crash because someone fixed a cosmetic error by adding the missing colon of a label's text.

It's your decision though.

StephaneDelcroix:
Hi @LWChris,

Implementing x:ClassModifier and x:FieldModifier is definitely an addition we could consider.

If you plan to do this, here's how you should ideally proceed:
- Start discussing this as a proposal on https://forums.xamarin.com/categories/xamarin-forms-evolution. It might look straightforward at first, but it's not that simple as the templates for xaml file creation will need to be changed so the `partial class` in code behind doesn't list any accessibility modifier (that could differ from the one in the `xaml.g.cs`).
- If you're looking into actually providing an implementation, that should be done in `XamlGTask.cs`. And the additional attributes will have to be marked as ignored in the visitors of both xaml inflators.

PR #659, opened 2016-12-21
Put the child view underneath the Navigation Bar when translucent (iOS)

skaman:
### Description of Change ###

Under iOS, only if the Navigation Bar is translucent, the content is placed underneath the Navigation
Bar and not aligned to the bottom.

### Bugs Fixed ###

- https://forums.xamarin.com/discussion/27745/navigationrender-issue

### API Changes ###

None

### Behavioral Changes ###

Before this patch, if the Navagion Bar was setted to translucent, underneath the navigation bar the user see the background of the navigation controller. After the patch he'll see the child view (on iOS this should be the right behaviour for translucent navigation bar)

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/NavigationRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/NavigationRenderer.cs
@@ -154,7 +154,7 @@ namespace Xamarin.Forms.Platform.iOS
 
 			var toolbar = _secondaryToolbar;
 			// Use 0 if the NavBar is hidden or will be hidden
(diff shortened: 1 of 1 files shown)


jimmgarrido:
This is a similar fix as #621

skaman:
True, the #621 maybe need also a fix for the container height. If the content go underneath the navigation bar, the height of the navigation bar should not be considered for the container height (it's like the navigation is invisibile and the container fit the screen).

rmarinho:
@jimmgarrido can you check if this is covered by #621 if so we can close this one.

jimmgarrido:
@rmarinho @skaman #621 does not adjust the container's size for this change and results in an empty area below the page content. #621 should be closed in favor of this pull request.

PR #660, opened 2016-12-22
[XamlC] compile ListStringTypeConverter

StephaneDelcroix:
### Description of Change ###

compile the ListStringTypeConverter, used in Style and the new OnPlatform

### Behavioral Changes ###

slightly longer IL, but no reflection calls, and less method calls

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/CompiledConverters/ListStringTypeConverter.cs
+++ b/Xamarin.Forms.Build.Tasks/CompiledConverters/ListStringTypeConverter.cs
@@ -20,9 +20,6 @@ namespace Xamarin.Forms.Core.XamlC
 			}
 			var parts = value.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries).Select(s => s.Trim()).ToList();
 
-			var list = module.Import(typeof(List<>)).Resolve().Methods;
-			if (list == null)
-				yield break;
 			var listCtor = module.Import(typeof(List<>)).Resolve().Methods.FirstOrDefault(md => md.IsConstructor && md.Parameters.Count == 1 && md.Parameters[0].ParameterType.FullName == "System.Int32");
 			var listCtorRef = module.Import(listCtor);
 			listCtorRef = module.Import(listCtorRef.ResolveGenericParameters(module.Import(typeof(List<string>)), module));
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Build.Tasks/CompiledConverters/ListStringTypeConverter.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+	class ListStringTypeConverter : ICompiledTypeConverter
+	{
+		public IEnumerable<Instruction> ConvertFromString(string value, ModuleDefinition module, BaseNode node)
+		{
+			if (value == null) {
+				yield return Instruction.Create(OpCodes.Ldnull);
+				yield break;
+			}
+			var parts = value.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries).Select(s => s.Trim()).ToList();
```
samhouts: I wonder if there is a scenario where users would want a whitespace entry to occupy space in a list of strings. Perhaps providing labels for a chart where they want to skip every other item, for example?

samhouts: I wonder if there is a scenario where users would want a whitespace entry to occupy space in a list of strings. Perhaps providing labels for a chart where they want to skip every other item, for example?

Xamarin.Forms.Xaml.UnitTests/CompiledTypeConverter.xaml
```diff
@@ -4,7 +4,8 @@
 			 x:Class="Xamarin.Forms.Xaml.UnitTests.CompiledTypeConverter"
 			 RectangleP="0,1,2,4"
 			 RectangleBP="4,8,16,32"
-			 BackgroundColor="Pink">
+			 BackgroundColor="Pink"
+			 List="Foo, Bar">
```
samhouts: If we're explicitly removing whitespace entries, should we test for that?

Xamarin.Forms.Build.Tasks/CompiledConverters/ListStringTypeConverter.cs
samhouts: I wonder if there is a scenario where users would want a whitespace entry to occupy space in a list of strings. Perhaps providing labels for a chart where they want to skip every other item, for example?

StephaneDelcroix: All scenarios are possible... the ListStringTypeConverter (the new compiled version, and the old one) are only used for `VisualElement.StyleClass` where it is ok to RemoveEmptyEntries, but it's not required.

PR #661, opened 2016-12-26
[Android] Hide Fragments before removing them to allow custom animations

adrianknight89:
### Description of Change ###

As of now, when the navigation stack is being navigated back, Fragments are being removed which does not allow for custom transition animations. This change attempts to hide them first and remove them when the transition is done.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=50787

### API Changes ###

Added:
 - `protected virtual int TransitionDuration { get; set; } = 220;`
 This seems to be the time interval before the transition timer triggers. By default, it's set to 220ms but the custom animations I provided use 500ms.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
@@ -651,7 +651,7 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 			// We don't currently support fragment restoration, so we don't need to worry about
 			// whether the commit loses state
 			transaction.CommitAllowingStateLoss();
-		
+
 			// The fragment transitions don't really SUPPORT telling you when they end
 			// There are some hacks you can do, but they actually are worse than just doing this:
 
--- a/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
@@ -624,10 +624,10 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 					while (_fragmentStack.Count > 1 && popPage)
 					{
 						Fragment currentToRemove = _fragmentStack.Last();
-                        _fragmentStack.RemoveAt(_fragmentStack.Count - 1);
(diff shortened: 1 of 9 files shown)


adrianknight89:
I'm not 100% sure if using two transactions is a good idea or whether `TransitionDuration` and animation durations in the xml files will run in sync. For example, if I set `android:duration="500"`, does Android guarantee that the slide will complete in exactly 500ms? I believe it should be okay if the timer kicks off on or after the slide animation is complete but not before as this would remove the fragment while it's sliding.

There is also the possibility of using `Replace` instead of hide & remove, which I do not know how to use at the moment.

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
```diff
@@ -612,7 +613,8 @@ Task<bool> SwitchContentAsync(Page view, bool animated, bool removed = false, bo
 					{
 						Fragment currentToRemove = fragments.Last();
 						fragments.RemoveAt(fragments.Count - 1);
-						transaction.Remove(currentToRemove);
+						transaction.Hide(currentToRemove);
```
ahmedalejo: Hi @adrianknight89 , doensn´t this by passes the `animated` parameter?

adrianknight89: I don't think it does. `SwitchContentAsync` has an `animated` parameter. We set up transitions only if this flag is true:

```
if (animated)
     SetupPageTransition(transaction, !removed);
```

I tested the sample code I provided with non-animated push and pop. It works fine.

rmarinho:
Needs rebase @adrianknight89

adrianknight89:
@rmarinho Sorry, it took a while. Should be done.

hartez:
We're going to be addressing this problem in a different way in an upcoming version.

hartez:
The original plan to implement this got put on the back burner; reopening this so we can determine if we want to use this solution for now.

adrianknight89:
I'll re-do it on Wednesday since re-basing this will be difficult.

kaushalyacs:
@adrianknight89 When this will be released?

PR #663, opened 2016-12-30
[Android] Keyboard should not change layout while dismissing on Entry and Editor

adrianknight89:
### Description of Change ###

When the soft keyboard is dismissing on `Entry` and `Editor`, it changes characters. This seems to happen randomly instead of following a consistent pattern unless you use a Google keyboard (tested on S6) in which case it seems to happen every time.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=43867

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla43867.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla43867.cs
@@ -26,19 +26,21 @@ namespace Xamarin.Forms.Controls.Issues
 				{
 					new Label
 					{
-						Text = "Focus and unfocus each element 10 times. Observe that the soft keyboard does not show different characters while hiding."
+						Text = "Focus and unfocus each element 10 times using the Back button. Observe that the soft keyboard does not show different characters while hiding. Now repeat the test by tapping off of the element."
 					},
 					new Entry
 					{
 						WidthRequest = 250,
 						HeightRequest = 50,
-						BackgroundColor = Color.AntiqueWhite
+						BackgroundColor = Color.AntiqueWhite,
+						Keyboard = Keyboard.Numeric
 					},
 					new Editor
 					{
 						WidthRequest = 250,
 						HeightRequest = 50,
-						BackgroundColor = Color.BurlyWood
+						BackgroundColor = Color.BurlyWood,
+						Keyboard = Keyboard.Numeric
 					}
 				}
 			};
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla43867.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla43867.cs
@@ -1,0 +1,47 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using Xamarin.Forms.PlatformConfiguration;
+using Xamarin.Forms.PlatformConfiguration.AndroidSpecific;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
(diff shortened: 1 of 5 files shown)


adrianknight89:
@hartez Since you're looking at the other keyboard issue, any way this can be given priority as well?

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla43867.cs
```diff
@@ -0,0 +1,47 @@
...
+			{
+				Spacing = 10,
+				VerticalOptions = LayoutOptions.Center,
+				Children =
+				{
+					new Label
+					{
+						Text = "Focus and unfocus each element 10 times. Observe that the soft keyboard does not show different characters while hiding."
+					},
+					new Entry
```
hartez: Your test doesn't set the keyboards to numeric, so the effect your testing against never happens.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla43867.cs
```diff
@@ -0,0 +1,47 @@
...
+
+			Content = new StackLayout
+			{
+				Spacing = 10,
+				VerticalOptions = LayoutOptions.Center,
+				Children =
+				{
+					new Label
+					{
+						Text = "Focus and unfocus each element 10 times. Observe that the soft keyboard does not show different characters while hiding."
```
hartez: You should specify that the tester also needs to verify that the problem doesn't occur when hitting the 'back' button; I've never seen this problem reproduced just from tapping off of the Entry/Editor to unfocus them. This might be different on other hardware/emulators, but the test should cover both.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla43867.cs
adrianknight89: Sorry, the issue exists for other kinds of keyboard types as well although in the bug description I had mentioned numeric keyboard only. Made the change to make it aligned with the bug.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla43867.cs
adrianknight89: Done.

rmarinho:
It's failing a test 45926 on Android, can you check @adrianknight89  ?

Xamarin.Forms.Platform.Android/Renderers/EditorEditText.cs
```diff
@@ -23,13 +23,12 @@ bool IDescendantFocusToggler.RequestFocus(global::Android.Views.View control, Fu
...
-			{
-				EventHandler handler = OnBackKeyboardPressed;
-				if (handler != null)
-					handler(this, EventArgs.Empty);
-			}
-			return base.OnKeyPreIme(keyCode, e);
+			if (keyCode != Keycode.Back || e.Action != KeyEventActions.Down)
+				return base.OnKeyPreIme(keyCode, e);
+
+			this.HideKeyboard();
```
rmarinho: can't we drop the this ?

adrianknight89: Since 'HideKeyboard' is an extension method, `this` is needed.

adrianknight89:
@rmarinho I just ran 45926, and it seems to be passing for me. Can you run it again to see if it wasn't an intermittent issue?

rmarinho:
@adrianknight89 yap .. we have alot of runs queued but it's up there to be run again.

hartez:
@rmarinho @adrianknight89 Just checked it, the failure on 45926 is that the version in this PR is slightly out-of-date. There was a change to that test checked in to master a few days back. So it's not the changes in this PR which are causing that failure.

PR #664, opened 2017-01-02
TargetNullValue & FallbackValue implemented

akgulebubekir:
### TargetNullValue & FallbackValue features added to Binding class ###

**TargetNullValue**: The value that is used in the target when the value of the source is null.
(This feature exists on Windows platforms. [See more ](https://msdn.microsoft.com/en-us/library/system.windows.data.bindingbase.targetnullvalue(v=vs.110).aspx))

**FallbackValue:** Gets or sets the value to use when the binding is unable to return a value.
(This feature exists on Windows platforms. [See more](https://msdn.microsoft.com/en-us/library/system.windows.data.bindingbase.fallbackvalue%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396))

Usage: 
`<Label Text="{Binding User.Name, TargetNullValue='Unknown', FallbackValue='-'}"/>`

- **Unknown**: when User.Name is `null`
- **-** : If `BindingContext `does not have property such `User` or `Name`

### Bugs Fixed ###

- 

### API Changes ###

Added:
 - object : FallbackValue Property in BindingBase
 - object : TargetNullValue  property in BindingBase

### Behavioral Changes ###
-None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Binding.cs
+++ b/Xamarin.Forms.Core/Binding.cs
@@ -82,9 +82,11 @@ namespace Xamarin.Forms
 			}
 		}
 
-		internal string UpdateSourceEventName {
+		internal string UpdateSourceEventName
+		{
 			get { return _updateSourceEventName; }
-			set {
+			set
+			{
 				ThrowIfApplied();
 				_updateSourceEventName = value;
 			}
@@ -116,7 +118,7 @@ namespace Xamarin.Forms
 			object src = _source;
 			base.Apply(src ?? newContext, bindObj, targetProperty);
 
-			object bindingContext = src ?? Context ?? newContext;
+			object bindingContext = src ?? Context ?? newContext ?? TargetNullValue;
 			if (_expression == null && bindingContext != null)
 				_expression = new BindingExpression(this, SelfPath);
 
@@ -125,7 +127,7 @@ namespace Xamarin.Forms
 
 		internal override BindingBase Clone()
 		{
-			return new Binding(Path, Mode) { Converter = Converter, ConverterParameter = ConverterParameter, StringFormat = StringFormat, Source = Source, UpdateSourceEventName = UpdateSourceEventName };
+			return new Binding(Path, Mode) { Converter = Converter, ConverterParameter = ConverterParameter, StringFormat = StringFormat, Source = Source, UpdateSourceEventName = UpdateSourceEventName, TargetNullValue = TargetNullValue, FallbackValue = FallbackValue };
 		}
 
 		internal override object GetSourceValue(object value, Type targetPropertyType)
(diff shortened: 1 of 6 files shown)


StephaneDelcroix:
That https://forums.xamarin.com/categories/xamarin-forms-evolution is the right place to discuss new API proposition. This is usually only used to discuss the implementation.

- `TargetNullValue` can be done with a Converter
- What happens if `TargetNullValue` is not set ? does it fall back to `FallBackValue` ?
- see previous question: there's no tests at all
- the Clone() in TypedBinding will ignore those 2 properties
- `TargetNullValue` is related to the `Converter` which is defined in `Binding`, not `BindingBase`. The same applies to `FallBackValue` which is related to the `Path`, and shouldn't be in `BindingBase` either.
- there are some unnecessary formatting change, but I don't really care about that.

For the reasons above, I'm going to close this PR. Don't get me wrong, I think it could be a nice addition, but this should be discussed first. So please open a proposal in the evolution forum. Thanks a lot.

StephaneDelcroix:
I changed my mind. this PR is salvageable and doesn't require much of discussion, only agreement. I'm reopening for discussion

StephaneDelcroix:
Note that having `null` by default as `FallbackValue` is a breaking change as right now we do not update the binding if the path can't be evaluated.

akgulebubekir:
> **TargetNullValue can be done with a Converter**
> Consider this as a ** syntatic sugar ** that suppose to be in a language. Also `Converter` does not works if model does not implement `INotifyPropertyChanged` interface. 

> **What happens if `TargetNullValue `is not set ? does it fall back to `FallBackValue `?**
> Nope, `FallBackValue` does not get involve to binding process if binding path is correct and exists.

> **The `Clone() `in `TypedBinding `will ignore those 2 properties**
> Will be first commit on tomorrow.

> **~~TargetNullValue is related to the Converter which is defined in Binding, not BindingBase. The same applies to FallBackValue which is related to the Path, and shouldn't be in BindingBase either.~~ it is defined in BindingBase on other platforms, so it's fine**
> I couldn't get your point

> **There are some unnecessary formatting change, but I don't really care about that.**
> Routine Visual Studio shortcut pressing :)  `CTRL - K + D`

jamesl77:
TargetNullValue & FallbackValue are very useful, I hope they get implemented well and sooner(yesterday).
I need to write unnecessary code sometimes in both UI and view-model to overcome the lack of these.

StephaneDelcroix:
Closing. It's breaking. We definitely consider adding this to the binding system, but not right now.

ctrexler:
I love everything your team has done. Just want to be another data point letting you know that this feature would be very appreciated, as I've now encountered some trouble with this that I'm now forced to find some kind of workaround for. Thanks!

opcodewriter:
Ditto. I have a long XAML background, and this feature is very useful, and like @ctrexler said, I also needed to find (not so nice) workarounds.

PR #667, opened 2017-01-04
[WinRT/UWP] Handle Button click routing via Tapped

pauldipietro:
### Description of Change ###

A Button being used inside of a ListView cell on WinRT/UWP will presently trigger a click/tap event for the button and the cell it belongs to due to the routing not being handled. Because the `Click` event can't handle routing itself, the `Tapped` event can be used in place of it to perform the same function while also setting the `TappedRoutedEventArgs`'s `Handled` property to true, preventing the unintended second tap from occurring. If the Button is not enabled, the taps on the button will instead go through to the cell.

Per the [MSDN docs](https://msdn.microsoft.com/en-us/library/windows/apps/windows.ui.xaml.uielement.tapped.aspx):

> The input system processes an action where the user clicks the left mouse button while over the element as a *Tapped* action. The event doesn't fire until the left mouse button is released. Mouse input doesn't produce Holding events by default, no matter how long a mouse button is held down, or which button is held.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=46632

### API Changes ###

None

### Behavioral Changes ###

Per the MSDN docs it seems like Tapped is the newer and more preferable way to handle click/tap events. I assume that as the button exists on WinRT/UWP now, if someone had a part of the screen that could be tapped and the button was on top of it, the tap would go through as it does with the ListView cells, but it seems like an edge case since I believe it's reasonable to want the button to stop bubbling routing events when it's the focal point of a click/tap.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla46632.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla46632.cs
@@ -1,0 +1,81 @@
+﻿using System;
+using System.Collections.Generic;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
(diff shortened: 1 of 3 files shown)


rmarinho:
@pauldipietro any feedback on this?

pauldipietro:
This was brought up back during the meeting and I think @samhouts wanted to look into it more because of the accessibility changes

samhouts:
Confirmed that this change does, indeed, break the narrator's ability to trigger buttons on mobile touch devices. Also, I'm fairly confident that #1133 will resolve this, if not affect it in some way. Closing this PR.

PR #669, opened 2017-01-09
Fix for nullreferenceexception in TabbedRenderer on iOS when it has are no children (yet) 

jorisvergeer:
When a TabbedPage is rendered on iOS with no set CurrentPage (i.e. CurrentPage = null), this results in a NullReferenceException. 

This occurs for example when the children are loaded/added async.

### Description of Change ###

Add null check in ChildViewControllerForStatusBarHidden() in TabbedRenderer

### Bugs Fixed ###

None

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description) I don't know how to write automated integration tests.

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/TabbedRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/TabbedRenderer.cs
@@ -233,7 +233,11 @@ namespace Xamarin.Forms.Platform.iOS
 
 		public override UIViewController ChildViewControllerForStatusBarHidden()
 		{
-			return GetViewController(Tabbed.CurrentPage);
+			var current = Tabbed.CurrentPage;
+			if (current == null)
+				return null;
+
+			return GetViewController(current);
 		}
 
 		void UpdateCurrentPagePreferredStatusBarUpdateAnimation()


rmarinho:
Can you add a UITest for it?

jorisvergeer:
I will add tests when I have time. (Somewhere this week)
Isn't it a but dubious that there are currently no tests for this control/renderer?

rmarinho:
@jorisvergeer for the platform/renderers side we have UITests. If you write a UITest with a custom renderer that reproduces this should work. and cover this issue. 

Thanks.

rmarinho:
@jorisvergeer can you please add a UITest so we don't regress on this. 

Thanks

jorisvergeer:
@rmarinho Is there any documentation about how to get the UItest project to work?

All I get for every test when I do "run tests" is:
`TestFixtureSetUp failed in IssuesSetup`

I tried with the Xamarin.Forms.Core.iOS.UITests project on both a real device and the simulator. They all give the same results. (I followed instructions from https://developer.xamarin.com/guides/testcloud/uitest/working-with/testing-on-devices/ios/)

PR #671, opened 2017-01-10
Defer initialization of Application.Properties until first use

hartez:
### Description of Change ###

Currently the built-in Application.Properties dictionary is initialized as part of the Application constructor, which adds overhead to application startup even if the application doesn't use the feature. 

This change defers the initialization of the Properties feature until first use.

No new tests.

### Bugs Fixed ###

- None

### API Changes ###

- None

### Behavioral Changes ###

- Slightly faster application startup
- Slight delay on first access of Properties

### PR Checklist ###


hartez:
@StephaneDelcroix The task starts when `GetPropertiesAsync()` is called. Accessing `.Result` doesn't cause the task to start running, it just blocks until the task is complete.

PR #672, opened 2017-01-10
Reduce overhead of pushing existing navigation stack 

hartez:
### Description of Change ###

Removes the StackCopy member of NavigationPage and replaces it with an enumerable to avoid the Stack construction and Reverse operation overhead. 

### Bugs Fixed ###

- None

### API Changes ###

- Removes `StackCopy` from `INavigationPageController`
- Adds `Page Peek(int depth);`
- Adds `IEnumerable<Page> Pages { get; }`

### Behavioral Changes ###

- None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/INavigationPageController.cs
+++ b/Xamarin.Forms.Core/INavigationPageController.cs
@@ -7,7 +7,8 @@ namespace Xamarin.Forms
 {
 	public interface INavigationPageController
 	{
-		Page SecondToLast { get;  }
+		Page Peek(int depth);
+
 		IEnumerable<Page> Pages { get; }
 
 		int StackDepth { get; }
--- a/Xamarin.Forms.Core/INavigationPageController.cs
+++ b/Xamarin.Forms.Core/INavigationPageController.cs
@@ -7,7 +7,8 @@ namespace Xamarin.Forms
 {
 	public interface INavigationPageController
 	{
-		Stack<Page> StackCopy { get; }
+		Page SecondToLast { get;  }
+		IEnumerable<Page> Pages { get; }
 
 		int StackDepth { get; }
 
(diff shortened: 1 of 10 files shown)


jassmith:
👍

Xamarin.Forms.Core/INavigationPageController.cs
```diff
@@ -7,7 +7,8 @@ namespace Xamarin.Forms
 {
 	public interface INavigationPageController
 	{
-		Stack<Page> StackCopy { get; }
+		Page SecondToLast { get;  }
```
StephaneDelcroix: can't this be calculated from `IEnumerable<Page> Pages {get;}` ?

shouldn't it be internal ?

hartez:
I never liked SecondToLast; I have another idea I hate a lot less.

hartez:
@StephaneDelcroix I've nixed SecondToLast and replaced it with `Page Peek(int depth)`, which is just as fast and also more generally useful.

Xamarin.Forms.Core/NavigationPage.cs
```diff
@@ -61,17 +62,18 @@ public Color Tint
...
 			{
-				var result = new Stack<Page>(PageController.InternalChildren.Count);
-				foreach (Page page in PageController.InternalChildren)
-					result.Push(page);
-				return result;
+				return null;
 			}
+
+			return (Page)PageController.InternalChildren[PageController.InternalChildren.Count - depth - 1];
 		}
```
djcparker: Peek(-1) would be a bad thing, right?  Should probably return null if depth < 0?

hartez: Good point.

hartez: Okay, `Peek(-1)` now returns `null`.

PR #673, opened 2017-01-10
Make ExpressionSearch for RelativeLayouts lazy

hartez:
### Description of Change ###

Currently Forms intializes the ExpressionSearch at application startup, even if the application doesn't use RelativeLayout anywhere.

This change makes the initialization of ExpressionSearch occur on first use of a RelativeLayout.

### Bugs Fixed ###

- None

### API Changes ###

- None

### Behavioral Changes ###

- None

### PR Checklist ###


hartez:
> is the difference measurable at all ? it's just a single object creation after all

I thought the answer was "yes", and then I checked and I was measuring the wrong line of code. After measuring the correct line, the startup savings isn't worth the `Lazy<T>` overhead.

PR #674, opened 2017-01-10
Improve registry startup and lookup performance

hartez:
Optimize search for Effect resolution name
Avoid multiple collection of assemblies during startup
Cache handler lookups

### Description of Change ###

Don't resolve Effect resolution short name unless the assembly actually has effects.

If the Dependency Service hasn't already been initialized, pass the assembly list to it for initialization (so the Dependency Service doesn't have to compile that list again for itself).

When  looking up handler types, cache null results to avoid traversing the class hierarchy on subsequent lookups.

### Bugs Fixed ###

- None

### API Changes ###

- None

### Behavioral Changes ###

- None 

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/DependencyService.cs
+++ b/Xamarin.Forms.Core/DependencyService.cs
@@ -70,6 +70,16 @@ namespace Xamarin.Forms
 				assemblies = assemblies.Union(Registrar.ExtraAssemblies).ToArray();
 			}
 
+			Initialize(assemblies);
+		}
+
+		internal static void Initialize(Assembly[] assemblies)
+		{
+			if (s_initialized)
+			{
+				return;
+			}
+
 			Type targetAttrType = typeof(DependencyAttribute);
 
 			// Don't use LINQ for performance reasons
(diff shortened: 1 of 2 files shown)


Xamarin.Forms.Core/DependencyService.cs
```diff
@@ -70,6 +70,16 @@ static void Initialize()
 				assemblies = assemblies.Union(Registrar.ExtraAssemblies).ToArray();
 			}
 
+			Initialize(assemblies);
+		}
+
+		internal static void Initialize(Assembly[] assemblies)
+		{
+			if (s_initialized)
```
StephaneDelcroix: why don't you do this check in the parameterless Initialize() ?

hartez: The only path to the parameterless `Initialize()` already checks `s_initialized` before calling it.

That said, checking it inside of `Initialize()` would be clearer; I'll update it.

hartez:
@StephaneDelcroix It doesn't do much - on an Android device for an app with enough assemblies, it can shave a few milliseconds off of startup.

PR #677, opened 2017-01-11
[WinRT/UWP] Enable selection in password entry

pauldipietro:
### Description of Change ###

There is presently an issue in WinRT/UWP where entries with `Password = true` do not allow selection of text. This is a larger issue on the desktop than mobile, but is an issue for both. What this PR changes in particular is that selection is allowed, but also handles a variety of other inputs. The code is appropriately commented due to being a process that involves

- Caching the size of the selection of text for deletion purposes
- Caching whether the Ctrl key is in a pressed state and handling the key routing
- Handling any use of the Shift key, Tab key, arrow keys, or Home/End/PgUp/PgDown keys (which can be used to select text or leave the entry)
- Handle any use of the Copy (C), Cut (X), Undo (Z), or Redo (Y) keys in the instance that the Ctrl key is in being actively pressed to prevent those shortcuts from working (which emulates the behavior of a `PasswordBox`)
- Remove any text if the aforementioned cached selection size is greater than 0 (in other instances a backspace will simply remove the single character)

Note that this should fix the core problem with being unable to select the text, but does not fix three issues:
- If the back button is pressed with the cursor mid-password, the trailing character gets removed.
- Inserting a character/pasting into in the middle of the password causes unintended results.
- The context menu still permits use those keyboard shortcuts which are otherwise disabled. An issue is that WinRT does not support the use of `Clipboard` so this makes things more difficult. If the text is copied, however, it is only copied as obfuscation characters. This can possibly be revisited in the future.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=45277

### API Changes ###

None

### Behavioral Changes ###

The only one that was very clearly visible was the fact that the password characters can be copied, but again, those are only the obfuscation characters (●).

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.WinRT/FormsTextBox.cs
+++ b/Xamarin.Forms.Platform.WinRT/FormsTextBox.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Threading;
 using System.Threading.Tasks;
+using Windows.System;
 using Windows.UI.Core;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
@@ -44,6 +45,7 @@ namespace Xamarin.Forms.Platform.WinRT
 		bool _cachedSpellCheckSetting;
 		CancellationTokenSource _cts;
 		bool _internalChangeFlag;
+		int _cachedSelectionLength;
 
 		public FormsTextBox()
 		{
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.WinRT/FormsTextBox.cs
```diff
@@ -237,19 +240,75 @@ static void OnIsPasswordChanged(DependencyObject dependencyObject, DependencyPro
...
+		// handled accordingly.
+		protected override void OnKeyDown(KeyRoutedEventArgs e)
+		{
+			if (IsPassword)
 			{
-				SelectionLength = 0;
-				SelectionStart = Text.Length;
+				// The _ctrlKeyActive flag is used to track if that key is pressed; we want to pretend Ctrl is not actively being
+				// used by treating the key routing as handled.
+				if (e.Key == VirtualKey.Control)
```
hartez: Is this double check intentional?

Xamarin.Forms.Platform.WinRT/FormsTextBox.cs
```diff
@@ -237,19 +240,75 @@ static void OnIsPasswordChanged(DependencyObject dependencyObject, DependencyPro
...
+				{
+					if (e.Key == VirtualKey.Control)
+					{
+						_ctrlKeyActive = true;
```
hartez: Rather than tracking this with a member variable, you could probably just do something like
`var ctrl = Window.Current.CoreWindow.GetKeyState(VirtualKey.Control).HasFlag(CoreVirtualKeyStates.Down);` here; then you don't have to worry about `OnKeyUp`.

Xamarin.Forms.Platform.WinRT/FormsTextBox.cs
pauldipietro: Nope 😃

Xamarin.Forms.Platform.WinRT/FormsTextBox.cs
pauldipietro: Good idea, made that change and it looks to behave the same.

Xamarin.Forms.Platform.WinRT/FormsTextBox.cs
```diff
@@ -237,19 +239,65 @@ static void OnIsPasswordChanged(DependencyObject dependencyObject, DependencyPro
...
+		{
+			if (IsPassword)
 			{
-				SelectionLength = 0;
-				SelectionStart = Text.Length;
+				// The ctrlDown flag is used to track if the Ctrl key is pressed; if it's actively being used and the most recent
+				// key to trigger OnKeyDown, then treat it as handled.
+				var ctrlDown = Window.Current.CoreWindow.GetKeyState(VirtualKey.Control).HasFlag(CoreVirtualKeyStates.Down);
+
+				if (ctrlDown && e.Key == VirtualKey.Control)
```
hartez: This `if` block seems unnecessary. You're checking whether the Ctrl key is down and the Ctrl key is the one being pressed.

pauldipietro: Yeah, you're right; I was concerned about the handling but it should be taken care of later via `base.OnKeyDown(e)` so I'll remove it.

PR #679, opened 2017-01-12
Add MonoAndroid70 to Forms/Map .nuspec 

kingces95:
Users want to upgrade to Xamarin.Support.* v25 but cannot because XF locks its version at v23 (in .nuspec, see version="[23.3.0]"). If we simply unlocked the version then they can reference v25 however those assemblies reference Android SDK API 25 which if not installed will break their build. So we have to make sure that by default, when they upgrade, API 25 is present. The VS SDK installer is not owned by Xamarin's division so are forced to wait till May for the installer to add API 25 during a VS upgrade. That upgrade will also include Xamarin.Android (XA) which will support MonoAndroid70. XA will automatically update Android csproj files at load time to MonoAndroid71. Hence, for users that simply upgrade VS in May, XF knows that if referring project is MonoAndroid70 or greater than API 25 is present and XS.* v25 can be referenced hence the change in the .nuspec.

XS v25 also removed an event signature so we changed to using the java methods to hook and unhook hence the source code changes.

### DIFF
--- a/.nuspec/Xamarin.Forms.Maps.nuspec
+++ b/.nuspec/Xamarin.Forms.Maps.nuspec
@@ -22,6 +22,12 @@
         <dependency id="Xamarin.Android.Support.v7.AppCompat" version="[23.3.0]"/>
         <dependency id="Xamarin.Forms$IdAppend$" version="$version$"/>
       </group>
+      <group targetFramework="MonoAndroid70">
+        <dependency id="Xamarin.GooglePlayServices.Maps" version="29.0.0.1"/>
+        <dependency id="Xamarin.Android.Support.v7.MediaRouter" version="23.3.0"/>
+        <dependency id="Xamarin.Android.Support.v7.AppCompat" version="23.3.0"/>
+        <dependency id="Xamarin.Forms$IdAppend$" version="$version$"/>
+      </group>
     </dependencies>
     <references>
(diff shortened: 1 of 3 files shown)


brandonrisell:
This change allows a project to get into some strange states.

For example:
* We updated all nuget packages after the most recent Xamarin release, which updated our Android Support libraries.
* We then decided to add the share plugin, https://github.com/jguertl/SharePlugin, which worked great in development, until we tried a release build and the linker choked.
* I put in an issue, https://github.com/jguertl/SharePlugin/issues/44, and downgraded support libraries to 24.2.1 so that share plugin worked.
* James Montemagno put out a beta build that is built against 25.1.1 yesterday, https://github.com/jguertl/SharePlugin/issues/44#issuecomment-283564209.
* I updated Support packages again, and installed beta build of share plugin. Working great.
* Today, decided to add deep linking to android, and find out that Xamarin.Forms.AppLinks won't install because it requires v23.3.0 of the Android support libraries.
* So I'm back to downgrading.

Now it's not that much work to downgrade and upgrade the packages, it's just a bit odd that Xamarin.Forms lets you update to android support versions that no other forms packages support. I feel like maybe the better approach is for the user who needs newer versions of the Android Support packages to utilize `-IgnoreDependencies`, and then install the versions that they specifically need.

PR #695, opened 2017-01-17
[WinRT/UWP] Fix Opacity behavior with ProgressBar

pauldipietro:
### Description of Change ###

An issue is present in WinRT/UWP where the `ActivityIndicator` shows up on the screen even if its `IsRunning` value is initially set to false. The present implementation of this modifies the `Opacity` value on the control to either 1 or 0 if the indicator is visible or not. This causes some unintended behavior because the fact that `Element.Opacity` being set to 1 causes the `VisualElementTracker` to set the control's Opacity back to that value. Users presently have to use workarounds to get `IsRunning = false` working correctly, and in the small chance they might not want to use something smaller than a 1 value for the opacity, it has to be explicitly set after setting `IsRunning` to true.

Adding in an `ElementOpacity` value via a `FormsProgressBar` class allows that value to be set by the user to a different value if desired. The added in templates then bind to that `ElementOpacity` value opposed to Control's `Opacity` so as to not cause any negative interactions between the two. This keeps the current behavior the same (by setting the Opacity on a non-running indicator opposed to setting its `Visibility` to `Collapsed`) while fixing two issues.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=37290

### API Changes ###

Add the `FormsProgressBar` class (inherits from `Windows.UI.Xaml.Controls.ProgressBar`) to the WinRT project (linked to by UWP as well)

### Behavioral Changes ###

None expected

### PR Checklist ###


rmarinho:
Can you rebase @pauldipietro ?

hartez:
@pauldipietro You're missing an entry in the `.nuspec` for the FormsProgressBarStyle `.xbf` file.

pauldipietro:
Updated; let me know if it's incorrect in any way.

hartez:
@pauldipietro Looks like there's still an entry missing - I think in the WinRT section? For tablet?

pauldipietro:
@hartez Added, thanks.

rmarinho:
@hartez can you do a final review?

PR #526, opened 2016-11-14
Platform specifics naming adjustments

pauldipietro:
### Description of Change ###

Renaming of some platform specifics stuff per @davidbritch's request -- namely `AdjustsFontSizeToFitWidth` to `IsAdjustsFontSizeToFitWidthEnabled` and `SetOffscreenPageLimitProperty`  to `SetOffscreenPageLimit`.

Edit: Updated to only use the latter.

### Bugs Fixed ###

N/A

### API Changes ###

See description changes.

### Behavioral Changes ###

N/A

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #696, opened 2017-01-17
[iOS] Remove "Enabled" usage on AdjustsFontSizeToFitWidth

pauldipietro:
### Description of Change ###

There was lingering usage of "Enabled" in naming from #526. This corrects those names while we should still be able to since this is only in 2.3.4-pre1.

### Bugs Fixed ###

N/A

### API Changes ###

See changes

### Behavioral Changes ###

N/A

### PR Checklist ###


pauldipietro:
Docs updated. @rmarinho this is only in 2.3.4-pre1

PR #701, opened 2017-01-18
[Core] Fix NullReferenceException in VisualElement finalizer.

kvpt:
### Description of Change ###

When `EventTrigger`, `DataTrigger` or `MultiTrigger` are attached to a `VisualElement`, the finalizer throw a `NullReferenceException` which crash the application after the page disappear. 
It's because these inherit from `TriggerBase` and not `Trigger`.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=51503

### API Changes ###
Nothing

### Behavioral Changes ###

### PR Checklist ###


StephaneDelcroix:
Excellent catch, thank you

PR #704, opened 2017-01-23
[macOS] Modern/iOS styled scrollbars for the ListView and ScrollView

MarkoBL:
### Description of Change ###

This PR adds modern/iOS styled vertical scrollbars to the ListView and ScrollView, as instroduced in macOS 10.7 https://developer.apple.com/library/content/releasenotes/MacOSX/WhatsNewInOSX/Articles/MacOSX10_7.html.

### Bugs Fixed ###

None

### API Changes ###

None

### Behavioral Changes ###

Before this PR, the macOS platform didn't show scrollbars at all.

### PR Checklist ###


rmarinho:
How does this work in previous versions? Won't it crash if the API isn't available  ?

MarkoBL:
I think it's going to crash, but I haven't tried it. Are you still supporting 10.6 with Xamarin.Mac? I can't even select it as target in Xamarin Studio in the info.plist, the oldest is 10.7.

rmarinho:
Please open again master now. THanks

PR #707, opened 2017-01-23
[WinRT/UWP] Do not allow IsPresented use in Split mode on desktop

pauldipietro:
### Description of Change ###

The `IsPresented` property should not function on WinRT/UWP in Split mode (whereas one can use Popover mode instead); currently setting `IsPresented = false` doesn't hide the pane when Split is being used, but it can subsequently be changed causing some unintended behavior with the master page. A check is done to prevent false from being set while also allowing the initial layout behavior to still display the master pane when using Split mode.

A side note is that the reproduction can trigger an `InvalidOperationException` when clicking around between the modes and changing the `IsPresented` value which I believe appears to be unrelated to this and involves how core handles `IsPresented` changing.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=51236

### API Changes ###

None

### Behavioral Changes ###

Technically speaking, people could be working around the behavior, so some unintended changes might occur, but I don't know how likely this is.

### PR Checklist ###


Xamarin.Forms.Platform.UAP/MasterDetailPageRenderer.cs
```diff
@@ -281,6 +281,11 @@ void UpdateDetailTitle()
 
 		void UpdateIsPresented()
 		{
+			// Ignore the IsPresented value being set to false for Split mode on desktop and allow the master
+			// view to be made initially visible
+			if (Device.Idiom == TargetIdiom.Desktop && Control.IsPaneOpen && Element.MasterBehavior != MasterBehavior.Popover)
```
StephaneDelcroix: we could Log() something here in debug mode.

PR #182, opened 2016-05-18
[UWP/8.1] ListView respects keyboard selection

pauldipietro:
### Description of Change ###

In both UWP/Win 8.1, if a keyboard was being used to make a selection out of a list of items, the `SelectedItem` value on the `Element` was not being updated due to its not being changed via tap/click, although `OnControlSelectionChanged` would still fire and update the `SelectedIndex` value on the `ListView`.

In this change, the `ShouldCustomHighlight` bool property is removed, but its general behavior in checking that the platform isn't UWP and is the Phone idiom are shifted into the `OnControlSelectionChanged` method itself. `OnControlSelectionChanged` is then always hooked and unhooked so that it may be called when the selection changes, and if the `Element.SelectedItem` value hasn't changed, we can reasonably assume that this selection change was made via keyboard and not tap or click, then update that value accordingly.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=41072

### API Changes ###

Removed:
- (private) bool ShouldCustomHighlight

### Behavioral Changes ###

None expected.

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #713, opened 2017-01-24
[All] Basic Accessibility Support

samhouts:
### Description of Change ###

Adds several new attached properties that will enable native screen readers to discover and interact with Xamarin.Forms controls.

As described in XFE proposal: https://forums.xamarin.com/discussion/85909/basic-accessibility-support

### Bugs Fixed ###

- Provides an alternate fix for [Bugzilla 41072](https://bugzilla.xamarin.com/show_bug.cgi?id=41072) that will actually fire the Forms `ItemSelected` event when selected by Keyboard and Narrator. (#182)

### API Changes ###

Added:
 - `string Accessibility.Hint { get; set; } //Attached Bindable Property`
 - `bool Accessibility.IsInAccessibleTree { get; set; } //Attached Bindable Property`
 - `VisualElement Accessibility.LableledBy {  get; set; } //Attached Bindable Property`
 - `string Accessibility.Name { get; set; } //Attached Bindable Property`

### Behavioral Changes ###

[Win] The change made in 1836e18 might cause the `ItemSelected` event to fire more than once. Was unable to cause this behavior, but I see the potential for it.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Accessibility.cs
+++ b/Xamarin.Forms.Core/Accessibility.cs
@@ -1,0 +1,14 @@
+﻿
+namespace Xamarin.Forms
+{
+	public class Accessibility
+	{
+		public static readonly BindableProperty HintProperty = BindableProperty.Create("Hint", typeof(string), typeof(Element), default(string));
+
+		public static readonly BindableProperty IsInAccessibleTreeProperty = BindableProperty.Create("IsInAccessibleTree", typeof(bool?), typeof(Element), null);
+
+		public static readonly BindableProperty LabeledByProperty = BindableProperty.Create("LabeledBy", typeof(VisualElement), typeof(Element), default(VisualElement));
+
+		public static readonly BindableProperty NameProperty = BindableProperty.Create("Name", typeof(string), typeof(Element), default(string));
+	}
+}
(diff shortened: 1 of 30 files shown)


Xamarin.Forms.Core/Accessibility.cs
```diff
@@ -0,0 +1,14 @@
+﻿
+namespace Xamarin.Forms
+{
+	public class Accessibility
+	{
+		public static readonly BindableProperty HintProperty = BindableProperty.Create("Hint", typeof(string), typeof(Element), default(string));
```
StephaneDelcroix: does it makes sense to only allow that on `VisualElement`? why do we need `Element` for ? `Cells` ?

huangjinshe:
I'm not sure this is the problem source, but after I update my master source code, I always get `System.Runtime.InteropServices.MissingInteropDataException:“ComTypeMarshalling_MissingInteropData” error, and the application will crash when I release the UWP app with .net Native.

It will fixed until I add my project Default.rd.xml:
` <Type Name="Windows.UI.Xaml.Automation.Peers.AccessibilityView" MarshalObject="Required Public" />`

But after that, I realize there are some other problem such like: the page trigger not working(the page aways keep the first state).

Do you have any suggestion for this? Thank you.

PR #396, opened 2016-09-30
[Android] Numeric keyboard should be able to process all available keys

adrianknight89:
### Description of Change

Android numeric keyboard displayed 14 keys (https://developer.xamarin.com/recipes/cross-platform/xamarin-forms/controls/choose-keyboard-for-entry/), but only the digits along with the dot character were processed. This ensures that all characters can be input.

Steps to verify changes:
- Install Google Keyboard from the play store
- Change your input language to Swedish
- Run sample code
- Initially, `KeyListener` should listen for the 14 keys.
- Change keyboard type by tapping on the button
- `KeyListener` should be set to the proper listener

`InputType = InputTypes.ClassNumber | InputTypes.NumberFlagSigned | InputTypes.NumberFlagDecimal` does not work (tested on S6 / M) probably because of Android fragmentation. 

IMO, entry validation should be left to the developer (see user's request). The simplest way to validate numbers is with `double.Parse(text, CultureInfo.CurrentCulture)`, but for whatever reason custom validation might be required or the UI might be tasked to react as the entry changes from a valid number to an invalid number or vice versa.

You might see [this bug](https://bugzilla.xamarin.com/show_bug.cgi?id=43867) while testing the sample. Please note that this is an existing bug.
### Bugs Fixed
- https://bugzilla.xamarin.com/show_bug.cgi?id=42000

PR #719, opened 2017-01-25
Add localized listener for Android numeric input

hartez:
### Description of Change ###

Proposing this as an alternative fix to [PR 396](https://github.com/xamarin/Xamarin.Forms/pull/396)

The DigitsKeyListener in Android has `.` [hard-coded as the decimal separator](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/text/method/DigitsKeyListener.java#63). This change introduces a LocalizedDigitsKeyListener which checks the current decimal separator for the device language. If the separator is `.`, it defers to the built-in DigitsKeyListener; if not, it uses an instance of LocalizedDigitsKeyListener with the correct decimal separator.

This preserves the default Android behavior (not allowing the user to enter an invalid number format) while allowing for "alternate" decimal separators (e.g. `,`). 

This change also provides an easily overridable method for providing an alternate NumericKeyListener for users who need to disallow negative/decimal numbers or provide alternate acceptable characters.

A test page is included, but it's manual - there's currently no UI test facility for changing the language settings of the device mid-test.

### Bugs Fixed ###

- [42000 – Unable to use comma (",") as decimal point](https://bugzilla.xamarin.com/show_bug.cgi?id=42000)

### API Changes ###

- None

### Behavioral Changes ###

- This change makes the numeric keyboard on Android allow negative numbers by default; the previous default did _not_ allow negative numbers. I believe this new default follows the law of least surprise, but feel free to argue for the previous default behavior in the comments.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Cells/EntryCellRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Cells/EntryCellRenderer.cs
@@ -1,5 +1,7 @@
 using System.ComponentModel;
 using Android.Content;
+using Android.Text;
+using Android.Text.Method;
 using Android.Views;
 
 namespace Xamarin.Forms.Platform.Android
@@ -56,6 +58,14 @@ namespace Xamarin.Forms.Platform.Android
 				UpdateHeight();
 		}
 
+		protected NumberKeyListener GetDigitsKeyListener(InputTypes inputTypes)
+		{
+			// Override this in a custom renderer to use a different NumberKeyListener 
+			// or to filter out input types you don't want to allow 
+			// (e.g., inputTypes &= ~InputTypes.NumberFlagSigned to disallow the sign)
+			return LocalizedDigitsKeyListener.Create(inputTypes);
+		}
+
 		void OnEditingCompleted()
 		{
 			var entryCell = (IEntryCellController)Cell;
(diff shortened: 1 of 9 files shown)


Xamarin.Forms.Platform.Android/Cells/EntryCellRenderer.cs
```diff
@@ -56,6 +58,14 @@ protected override void OnCellPropertyChanged(object sender, PropertyChangedEven
 				UpdateHeight();
 		}
 
+		protected NumberKeyListener GetDigitsKeyListener(InputTypes inputTypes)
```
samhouts: `virtual` ?

Xamarin.Forms.Platform.Android/Renderers/EditorRenderer.cs
```diff
@@ -86,6 +87,14 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
 			base.OnElementPropertyChanged(sender, e);
 		}
 
+		protected NumberKeyListener GetDigitsKeyListener(InputTypes inputTypes)
```
samhouts: `virtual` ?

Xamarin.Forms.Platform.Android/Cells/EntryCellRenderer.cs
hartez: Fixed. Thanks.

Xamarin.Forms.Platform.Android/Renderers/EditorRenderer.cs
hartez: Fixed. Thanks.

PR #541, opened 2016-11-17
Deprecate MessagingCenter in favor of Messaging

adrianknight89:
### Description of Change ###

Marking `MessagingCenter` obsolete and introducing a new service as suggested in #514. I updated internal references so that there is no reference to `MessagingCenter`. At the moment, I'm duplicating code between the old and new services. I'm not sure if this is the right approach or if the common code should be abstracted into another file.

EDIT: I moved argument validations to public methods and validated `message`. Since this is a new service, in my opinion, message names should not be allowed to be empty.

The easiest way for people to change their source code is to replace references to `MessagingCenter.` with `Messaging.Instance.` and ensure that their message names are not empty.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=46601

### API Changes ###

Added:
 - `IMessaging`
 - `Messaging`

Changed:
 - `MessagingCenter` is marked obsolete.

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #723, opened 2017-01-27
Make MessagingCenter testable

hartez:
### Description of Change ###

Proposing this as an alternative to [PR 541](https://github.com/xamarin/Xamarin.Forms/pull/541). This change is designed to allow for easier testing of components which utilize `MessagingCenter` by introducing an interface (`IMessagingCenter`) which can be faked/mocked/substituted in automated tests (see the `TestMessagingCenterSubstitute()` method in `MessagingCenterTests` for an example). 

This is essentially the same interface proposed in 541, but built on top of the changes from [PR 617](https://github.com/xamarin/Xamarin.Forms/pull/617). Also, this version allows the continued use of the original `MessagingCenter` API without deprecation; the static methods on `MessagingCenter` simply call through to the new singleton instance, allowing the code to be shared rather than duplicated.

### Bugs Fixed ###

- [46601 – MessagingCenter is static](https://bugzilla.xamarin.com/show_bug.cgi?id=46601)

### API Changes ###

Added:
```
public interface IMessagingCenter
{
	void Send<TSender, TArgs>(TSender sender, string message, TArgs args) where TSender : class;
	void Send<TSender>(TSender sender, string message) where TSender : class;
	void Subscribe<TSender, TArgs>(object subscriber, string message, Action<TSender, TArgs> callback, TSender source = null) where TSender : class;
	void Subscribe<TSender>(object subscriber, string message, Action<TSender> callback, TSender source = null) where TSender : class;
	void Unsubscribe<TSender, TArgs>(object subscriber, string message) where TSender : class;
	void Unsubscribe<TSender>(object subscriber, string message) where TSender : class;
}
```

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/MessagingCenter.cs
+++ b/Xamarin.Forms.Core/MessagingCenter.cs
@@ -22,9 +22,7 @@ namespace Xamarin.Forms
 
 	public class MessagingCenter : IMessagingCenter
 	{
-		static readonly MessagingCenter s_instance = new MessagingCenter();
-
-		public static IMessagingCenter Instance => s_instance;
+		public static IMessagingCenter Instance { get; } = new MessagingCenter();
 
 		class Sender : Tuple<string, Type, Type>
 		{
--- a/Xamarin.Forms.Core/MessagingCenter.cs
+++ b/Xamarin.Forms.Core/MessagingCenter.cs
@@ -2,12 +2,30 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
-using System.Runtime.CompilerServices;
 
 namespace Xamarin.Forms
 {
-	public static class MessagingCenter
+	public interface IMessagingCenter
 	{
+		void Send<TSender, TArgs>(TSender sender, string message, TArgs args) where TSender : class;
+
+		void Send<TSender>(TSender sender, string message) where TSender : class;
+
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Core/MessagingCenter.cs
```diff
@@ -2,12 +2,30 @@
...
+		void Unsubscribe<TSender, TArgs>(object subscriber, string message) where TSender : class;
+
+		void Unsubscribe<TSender>(object subscriber, string message) where TSender : class;
+	}
+
+	public class MessagingCenter : IMessagingCenter
+	{
+		static readonly Lazy<MessagingCenter> s_instance = new Lazy<MessagingCenter>(() => new MessagingCenter());
+
+		public static IMessagingCenter Instance => s_instance.Value;
```
StephaneDelcroix: what's the point of the `Lazy` ? Allocations ? I don't think we gain anything measurable by instantiating a `Lazy<MC>` (and the Func) instead of a `MessagingCenter` in `.cctor` (static constructor), and at the end of the day, we have allocated one object more...

hartez: Now that I think about it more, there really is no point. Initializing it in the static constructor is just as thread-safe and there's no startup performance gain because Core uses it for several features, so it always gets created when you start the application no matter what. 

I'll update it to use the static constructor.

PR #725, opened 2017-01-27
[UWP] connect the actual ObservableCollection to the ItemsSource

StephaneDelcroix:
### Description of Change ###

[UWP] connect the actual ObservableCollection to the ComboBox.ItemsSource as it looks like ComboBox doesn't correctly connect to INPC or INPP on a non-ObservableCollection<> ItemsSource. We'd have to look at the ComboBox source to be 100% sure...

The workaround is then to expose our collection as `internal`

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=51642

### PR Checklist ###


pauldipietro:
@StephaneDelcroix: On my end, it looks like that the changes break this on Android (both FormsActivity/AppCompat) and the items aren't showing up. iOS seems okay.

There's another issue I'm seeing too where if you add a label to the repro and have it set the text to whatever the `SelectedItem` is when `SelectedIndexChanged` fires, the first time something is selected, it's null. That appears to be occurring [here](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/Picker.cs#L181) because the event's getting invoked before the `SelectedItem` is updated; swapping their order appears to fix the issue (it also looks vaguely similar to [39407](https://bugzilla.xamarin.com/show_bug.cgi?id=39407)). If you want, I could make that change as another PR for history clarity, or you could include it in this one, since I'm assuming it's legitimate?

StephaneDelcroix:
@pauldipietro push to this branch, it's fine

pauldipietro:
I pushed that change which you can take a peek at, but Android still has issues with the items not showing up.

TomQv:
When will this be merged?
Its desperatly awaited...

TomQv:
When will this be merged?
Its desperatly awaited...

PR #699, opened 2017-01-18
[XamlC] Produce correct mdb files

StephaneDelcroix:
### Description of Change ###

- Update to cecil 0.10.0-beta
- no longer duplicate methods (used to mess with sequence_points matching)
- force using symbol reader and writers
- allow debugging of Xamarin.Forms.Xaml.UnitTests

I can't automate easily the check making sure the mdb file is modified, and correct. But as our relationship is based on trust, that's ok.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=48378

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #726, opened 2017-01-27
[XamlC] support portable pdb

StephaneDelcroix:
### Description of Change ###

[XamlC] support portable pdb. Reuse Cecil logic (https://github.com/mono/cecil/commit/898fb6b8cb03319937448d93bdf6eaa3dd2c6008#diff-d60c9973465fb52c71c5ba9af324992aR687) based on file availability to pick between Pdb (portable or native) or Mdb.

Portable Pdb will be the default symbol format for mono in Cycle10. We HAVE to support this.

~~This PR actually contains a single commit but is built on top of #699. Merge that one before.~~

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=51778

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/DebugXamlCTask.cs
+++ b/Xamarin.Forms.Build.Tasks/DebugXamlCTask.cs
@@ -46,7 +46,7 @@ namespace Xamarin.Forms.Build.Tasks
 			using (var assemblyDefinition = AssemblyDefinition.ReadAssembly(Assembly, new ReaderParameters {
 				ReadWrite = true,
 				ReadSymbols = DebugSymbols,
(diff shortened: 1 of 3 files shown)


marek-safar:
Cecil now supports this officially and even simpler code can be used. See https://github.com/jbevain/cecil/pull/322 for more details

StephaneDelcroix:
@marek-safar I know, but for some reason cecil can't find the symbol assembly at runtime using reflection while being used in our task. This is the reason why we manually reference the symbol reader providers...

marek-safar:
Do you have the exception with the type it cannot find?

PR #731, opened 2017-01-31
Fix disappearing ListView text when changing color settings on Windows

hartez:
### Description of Change ###

Modified the `WrapperControl` class which hosts `ViewCell` to inherit from `Canvas` rather than `ContentControl`; it now responds correctly to color preferences changes.

This modification also fixes the problems reported in [32847](https://bugzilla.xamarin.com/show_bug.cgi?id=32847), so the ugly hacks previously used to fix those issues (the `IWrapperAware` interface and several event listeners attached directly to the individual components of `DatePicker` and `TimePicker`) are no longer necessary.

### Bugs Fixed ###

- [40139 – Changing the Windows 10 System Theme Color causes ListView text to disappear.](https://bugzilla.xamarin.com/show_bug.cgi?id=40139)
- [47701 – Label inside ViewCell from ListView will clear if updated after await Task.Delay(anyNumber)](https://bugzilla.xamarin.com/show_bug.cgi?id=47701)

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


huangjinshe:
@hartez I'm not very sure but current master source code can't build, the visual studio:2015 output the error:

> 5>C:\Program Files (x86)\MSBuild\Microsoft\.NetNative\ARM\ilc\IlcInternals.targets(936,5): error : MCG0024: MCG0024:UnresolvableTypeReference 在“Assembly(Name=Xamarin.Forms.Platform.UAP, Version=2.0.0.0, Culture=neutral, PublicKeyToken=null)”找到了无法解析的类型引用“Xamarin.Forms.Platform.UWP.FormsDatePicker”。请检查生成系统中的引用。引用缺失或者程序集缺少预期类型。


I tried to search anything about FormsDatePicker in the solution, but not found.
Could you please check that there is something not remove clearly after you remove FormsDatePicker? Thank you.

PR #736, opened 2017-02-01
Xaml empty service provider

StephaneDelcroix:
### Description of Change ###

Do not generate ServiceProvider if they're not used in Custom MarkupExtensions.

### Bugs Fixed ###

- ask @jonathanpeppers 

### API Changes ###

Added:
```
[AttributeUsage(AttributeTargets.Class, Inherited = false)]
public sealed class AcceptEmptyServiceProviderAttribute : Attribute
```

### Behavioral Changes ###

/

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/SetPropertiesVisitor.cs
+++ b/Xamarin.Forms.Build.Tasks/SetPropertiesVisitor.cs
@@ -279,6 +279,7 @@ namespace Xamarin.Forms.Build.Tasks
 			else if (vardefref.VariableDefinition.VariableType.ImplementsGenericInterface("Xamarin.Forms.Xaml.IMarkupExtension`1",
 				out markupExtension, out genericArguments))
 			{
+				var acceptEmptyServiceProvider = vardefref.VariableDefinition.VariableType.GetCustomAttribute(module.ImportReference(typeof(AcceptEmptyServiceProviderAttribute))) != null;
 				if (vardefref.VariableDefinition.VariableType.FullName == "Xamarin.Forms.Xaml.BindingExtension")
 					foreach (var instruction in CompileBindingPath(node, context, vardefref.VariableDefinition))
 						yield return instruction;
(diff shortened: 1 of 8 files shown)


jonathanpeppers:
@StephaneDelcroix I had to merge [several](https://github.com/Hitcents/Xamarin.Forms/commits/hitcents) of your PRs to try this as we were on top of 2.3.3.

At first glance, it seems our compiled binary went down about 6MB after putting `[AcceptEmptyServiceProvider]` on several of our markup extensions. Then I added back ARMv7 and our app is once again working on 32 bit.

There is one thing in our codebase that is getting an error in XamlC now, but we are messing with it to figure out what is wrong. We also need to QA the app in general to make sure nothing broke.

Are there more XamlC changes in 2.3.4 that would mean we should upgrade?

jonathanpeppers:
@StephaneDelcroix here is an example of XAML that isn't compiling anymore:

```xml
<ContentView
  x:Class="ToE.Forms.SomeAmazingView">
  <ContentView.Resources>
    <ResourceDictionary>
      <Style x:Key="Default" TargetType="ContentView">
        <Setter Property="Content">
          <Setter.Value>
            <Label Text="Bananas!" />
          </Setter.Value>
        </Setter>
    </ResourceDictionary>
  </ContentView.Resources />
</ContentView>
```

The style is just applied in code-behind on certain situations. The Style existing seems be what breaks XamlC, it just says:
```
ToE.Forms.SomeAmazingView.xaml: error : The given key was not present in the dictionary.
```

StephaneDelcroix:
@jonathanpeppers see #684

jonathanpeppers:
@StephaneDelcroix #684 fixes the regression except for one part.

This xaml above has `<Setter.Value></Setter.Value>` which appears to break unless you remove it. (I just removed it from our codebase since it is ContentProperty).

The error I get is:
```
ToE.Forms.SomeAmazingView.xaml(50,10): error : Position 50:10. Missing Value for Setter
```
The line number is pointing to the opening `<Style>` tag I believe. 

I don't think this regression is related to this PR, though.

jonathanpeppers:
Since Monday, we have this fix live on the app store. Seems to be working great.

PR #746, opened 2017-02-06
[*] ScaleX and ScaleY

StephaneDelcroix:
### Description of Change ###

ScaleX and ScaleY properties. See https://forums.xamarin.com/discussion/84709/add-scalex-and-scaley-properties for the proposal

### Bugs Fixed ###

/

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###


AndroidNative/FormsViewGroup/src/com/xamarin/forms/platform/android/FormsViewGroup.java
```diff
@@ -72,7 +72,8 @@ public void sendBatchUpdate (
 			float rotation, 
 			float rotationX, 
 			float rotationY,
-			float scale,
+			float scaleX,
+			float scaleY
```
samhouts: missing `,`

AndroidNative/FormsViewGroup/src/com/xamarin/forms/platform/android/FormsViewGroup.java
```diff
@@ -88,8 +89,8 @@ public void sendBatchUpdate (
 		setRotation (rotation);
 		setRotationX (rotationX);
 		setRotationY (rotationY);
-		setScaleX (scale);
-		setScaleY (scale);
+		setScaleX (scaleX);
+		setScaleY (sclaeY);
```
samhouts: typo

Xamarin.Forms.Platform.Android/VisualElementTracker.cs
```diff
@@ -230,7 +230,7 @@ void SetElement(VisualElement oldElement, VisualElement newElement)
 						UpdateRotationX();
 					if (oldElement.RotationY != newElement.RotationY)
 						UpdateRotationY();
-					if (oldElement.Scale != newElement.Scale)
+					if (oldElement.Scale != newElement.Scale || oldElement.ScaleX != newElement.ScaleY || oldElement.ScaleY != newElement.ScaleY)
```
samhouts: `oldElement.ScaleX != newElement.ScaleX`

Xamarin.Forms.Platform.WinRT/VisualElementTracker.cs
```diff
@@ -188,7 +188,7 @@ protected virtual void OnPropertyChanged(object sender, PropertyChangedEventArgs
 			{
 				UpdateScaleAndRotation(Element, Container);
 			}
-			else if (e.PropertyName == VisualElement.ScaleProperty.PropertyName)
+			else if (e.PropertyName == VisualElement.ScaleProperty.PropertyName || e.PropertyName == VisualElement.ScaleXProperty.PropertyName ||)e.PropertyName == VisualElement.ScaleYProperty.PropertyName
```
samhouts: `||)` ?

programmation:
Shouldn't the `Scale` property now work in terms of `ScaleX` and `ScaleY`? 
It appears to me that it's now possible to set `Scale` independently of `ScaleX` and `ScaleY`, whereas I would have thought that setting `Scale` should automatically set both of the others (to the same value).
Did I miss where that happens?

StephaneDelcroix:
@programmation we can not remove the Scale property, as it would break backward compatibility, so you can see Scale as a global scaling factor. The actual scaling factors are computed as follows:

```
var actualScaleX = Scale * ScaleX;
var actualScaleY = Scale * ScaleY;
```

programmation:
Aha. Thanks for explaining. 
It would probably be worth highlighting that subtlety in the release notes. 
It makes sense when I think about it some more as it's hard to see how else we could have a meaningful getter for `Scale`.

StephaneDelcroix:
@jassmith could you rebuild and push the FormsView project ? or make sure the project files are in source control ? Thanks

Xamarin.Forms.Platform.Android/VisualElementTracker.cs
```diff
@@ -305,9 +305,18 @@ void UpdateNativeView(object sender, EventArgs e)
...
+											   (float)view.Rotation,
+											   (float)view.RotationX,
+											   (float)view.RotationY,
+											   (float)view.ScaleX,
+											   (float)view.ScaleY,
+											   _context.ToPixels(view.TranslationX),
+											   _context.ToPixels(view.TranslationY));
 			}
 			else
 			{
```
hartez: The fast renderers on Android will ignore the independent X and Y scale properties initially; we'd need to update `FormsViewGroup.SendViewBatchUpdate` to handle the ScaleX and ScaleY.
But now that I look at it, we could probably just combine those two methods into a single method which takes a `View`.

hartez: It doesn't matter to me if that happens in this PR or a separate one.

StephaneDelcroix:
see #957. closing the discussion here

PR #747, opened 2017-02-06
Make Entry completed behavior on UWP/WinRT match Android/iOS

hartez:
### Description of Change ###

Hitting Enter on the soft keyboard on WinRT and UWP now dismisses the soft keyboard (to make the behavior match the Forms Android/iOS behaviors). 

Also fixes the internal UnfocusControl method to behave like Android/iOS.

### Bugs Fixed ###

- [45067 – No way of cleanly dismissing soft keyboard in UWP](https://bugzilla.xamarin.com/show_bug.cgi?id=45067)

### API Changes ###

- None

### Behavioral Changes ###

- None 

### PR Checklist ###


JKennedy24:
Looking at this it looks like UWP will `TryHide` the Keyboard on Enter pressed. Should it not Unfocus the Control?

How do I "Complete" an Entry using the keyboard?

JKennedy24:
While looking at UWP Entries, does this also fix the issue where the first entry in a ScrollView takes focus when the ScrollView is clicked, explained here: [https://bugzilla.xamarin.com/show_bug.cgi?id=52416](https://bugzilla.xamarin.com/show_bug.cgi?id=52416)

EmilAlipiev:
this is  not working on 4.8. when i press enter on entry, it wont unfocus

PR #748, opened 2017-02-06
Split Windows RuntimePlatform into UWP and WinRT

hartez:
### Description of Change ###

Splits the `Windows` option for `RuntimePlatform` into separate `WinRT` and `UWP` values, allowing for separate `OnPlatform` values for each platform (and distinguishing between the two anywhere else it might be called for).

Also added long-overdue `UWP` option to `PlatformAffected`.

### Bugs Fixed ###

None

### API Changes ###

Anywhere which previously checked for a `Device.RuntimePlatform` of `Windows` will now need to check `WinRT` and `UWP` individually.

### Behavioral Changes ###

`Device.RuntimePlatform` now returns `WinRT` on WinRT and `UWP` on UWP.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Xaml/XamlParser.cs
+++ b/Xamarin.Forms.Xaml/XamlParser.cs
@@ -269,7 +269,14 @@ namespace Xamarin.Forms.Xaml
 					continue;
 				try {
 					if (targetPlatform != Device.RuntimePlatform)
+					{
+						// Special case for Windows backward compatibility
+						if (targetPlatform == "Windows" &&
+						    (Device.RuntimePlatform == Device.UWP || Device.RuntimePlatform == Device.WinRT))
+							continue;
+						
 						prefixes.Add(prefix);
+					}
 				} catch (InvalidOperationException) {
 					prefixes.Add(prefix);
 				}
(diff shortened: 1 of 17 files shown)


StephaneDelcroix:
When was the new RuntimePlatform released ? Are we breaking any stable release with that ?

Also note that this is breaking xmlns declarations if you're using native views in xaml...

hartez:
@StephaneDelcroix The new RuntimePlatform is from [2.3.4.184-pre1](https://releases.xamarin.com/pre-release-xamarin-forms-2-3-4-184-pre1/), so we're not breaking anything stable.

hartez:
@StephaneDelcroix I added a special case to make the xmlns declarations backward compatible.

Xamarin.Forms.Xaml/XamlParser.cs
```diff
@@ -269,7 +269,14 @@ static IList<string> PrefixesToIgnore(IList<KeyValuePair<string, string>> xmlns)
 					continue;
 				try {
 					if (targetPlatform != Device.RuntimePlatform)
+					{
+						// Special case for Windows backward compatibility
+						if (targetPlatform == "Windows" &&
```
StephaneDelcroix: I'd prefer this being written like this

```
if (targetPlatform != Device.RuntimePlatform && (targetPlatform != "Windows" || (Device.RuntimePlatform != Device.UWP && Device.RuntimePlatform != Device.WinRT)))
    prefixes.Add(prefix);
```

PR #755, opened 2017-02-09
Unfocus Entry on Enter pressed

JKennedy24:
### Description of Change ###

Unfocus UWP Entry on Enter Pressed

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.WinRT/EntryRenderer.cs
+++ b/Xamarin.Forms.Platform.WinRT/EntryRenderer.cs
@@ -111,6 +111,7 @@ namespace Xamarin.Forms.Platform.WinRT
 				return;
 
 			((IEntryController)Element).SendCompleted();
+			Control?.Unfocus();
 		}
 
 		void UpdateAlignment()


rmarinho:
@ez can you see how this works with your fix #747

Xamarin.Forms.Platform.WinRT/EntryRenderer.cs
```diff
@@ -111,6 +111,7 @@ void TextBoxOnKeyUp(object sender, KeyRoutedEventArgs args)
 				return;
 
 			((IEntryController)Element).SendCompleted();
+			Control?.Unfocus();
```
hartez: Why do you want the TextBox to lose focus when you hit Enter?

JKennedy24: These Entries are used mostly for Data Entry Forms / User Name / Password boxes.
The Editor is used for Multi Line - Where Enter is handled differently

To me Enter means "Complete - I'm finished with this text box" So the logical thing to mean was to unfocus the box.

StephaneDelcroix:
Part of this issue is fixed in #747. the other part is not valid (according to @hartez )

PR #760, opened 2017-02-13
Fixes unlimited Image sizing in UWP ImageCell and aligns lack of styling …

limefrogyank:
…to be similar to Android ImageCell.

### Description of Change ###

Added a fixed height to the `Image` element of `ImageCell`.  This stops large images from expanding the `Grid` container.  Also, added some styling changes to the overall `ImageCell` using `Margin` so that there is a 1 pixel gap between items and centering of the two text fields.

No tests as these are just UI property value changes.

### Bugs Fixed ###

[https://bugzilla.xamarin.com/show_bug.cgi?id=47962](https://bugzilla.xamarin.com/show_bug.cgi?id=47962)

### API Changes ###

None

### Behavioral Changes ###

NA

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/Resources.xaml
+++ b/Xamarin.Forms.Platform.UAP/Resources.xaml
@@ -375,7 +375,7 @@
 	</DataTemplate>
 
 	<DataTemplate x:Key="ImageCell">
-		<Grid>
+		<Grid Margin="10,0,10,1">
 			<Grid.ColumnDefinitions>
 				<ColumnDefinition Width="Auto" />
 				<ColumnDefinition />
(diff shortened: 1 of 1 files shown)


selaromdotnet:
was this not pulled? I am hitting the same issue reported in relation to this PR here: https://bugzilla.xamarin.com/show_bug.cgi?id=47962

on android, ios, image cell appears fine, but uwp is a large image and extends beyond screen view. there was concern about changing default uwp behavior, but I think all three platforms should match if this viewcell is used...

is there any way to fix this without having this PR having been pulled?

StephaneDelcroix:
this is a breaking change, and anyway hardcoding values is not the way to go.

PR #762, opened 2017-02-14
[WinRT/UWP] Open Picker dropdown when calling Focus

pauldipietro:
### Description of Change ###

The Picker presently focuses the `ComboBox` but doesn't open the dropdown when `Focus()` is called on it. The dropdown itself (essentially a `Popup`) is separate from the `ComboBox` (`FormsComboBox`) that is used. Hooking into `GotFocus` and tracking whether the dropdown was just closed lets this function in line with what the other platforms do.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=52266

### API Changes ###

None

### Behavioral Changes ###

The unfortunate side effect here is that this is a behavior change from what presently occurs which appears in line with the default behavior of the platform (focusing the `ComboBox` only focuses the `ComboBox` and does not open the dropdown). We could alternatively implement a platform specific behavior which allows the user to define if this focus behavior opens the popup (or vice versa).

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla52266.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla52266.cs
@@ -1,0 +1,38 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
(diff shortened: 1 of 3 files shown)


rmarinho:
Agree this is a breaking change but for me is ok because it will be more similar to what a user will expect.

jassmith:
Yeah this is fine

PR #764, opened 2017-02-14
IEditorController

kingces95:
### Description of Change ###

### API Changes ###

Added Editor.SendCompleted as explicit implementation of IEditorController.

### DIFF
--- a/Xamarin.Forms.Core/Editor.cs
+++ b/Xamarin.Forms.Core/Editor.cs
@@ -68,8 +68,7 @@ namespace Xamarin.Forms
 			return _platformConfigurationRegistry.Value.On<T>();
 		}
 
-		void IEditorController.SendCompleted() => SendCompleted();
-		internal void SendCompleted()
+		void IEditorController.SendCompleted()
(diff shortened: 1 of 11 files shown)


Xamarin.Forms.Core/Editor.cs
```diff
@@ -68,6 +68,7 @@ public Editor()
 			return _platformConfigurationRegistry.Value.On<T>();
 		}
 
+		void IEditorController.SendCompleted() => SendCompleted();
```
jassmith: Missing new line

kingces95: Happy to add it. It's just that those two members are so closely related. Would you like me to add a new line for all such members going forward or would you like to reconsider?

kingces95:
Any objection to updating docs at the end of the exercise?

Xamarin.Forms.Core/Editor.cs
jassmith: Always have the space

PR #767, opened 2017-02-15
IWebViewController

kingces95:

Describe your changes here.

None

### API Changes ###

Added IWebViewController which contains internal members of WebView. Explicitly implemented IWebViewController on WebView with implementations delegating to their corresponding internal members.

### Behavioral Changes ###

None

### PR Checklist ###

Want to rebase onto https://github.com/xamarin/Xamarin.Forms/pull/766 before generating docs.

### DIFF
--- a/Xamarin.Forms.Core/WebView.cs
+++ b/Xamarin.Forms.Core/WebView.cs
@@ -41,24 +41,22 @@ namespace Xamarin.Forms
 
 		bool IWebViewController.CanGoBack {
 			get { return CanGoBack; }
-			set { CanGoBack = value; }
+			set { SetValue(CanGoBackPropertyKey, value); }
 		}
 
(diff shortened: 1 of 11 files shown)


Xamarin.Forms.Core/WebView.cs
```diff
@@ -39,12 +39,22 @@ public WebView()
...
 
+		bool IWebViewController.CanGoBack {
+			get { return CanGoBack; }
+			set { CanGoBack = value; }
+		}
+
 		public bool CanGoBack
 		{
 			get { return (bool)GetValue(CanGoBackProperty); }
 			internal set { SetValue(CanGoBackPropertyKey, value); }
```
StephaneDelcroix: well, at this point, remove the `internal` on setters, and you can drop the interface as WebView is already public

PR #768, opened 2017-02-15
INavigationMenuController

kingces95:

### Describe your changes here.

None

### API Changes

Added INavigationMenuController which contains internal members of NavigationMenu. Explicitly implemented INavigationMenuController on NavigationMenu with implementations delegating to their corresponding internal members.

### Behavioral Changes

None

### PR Checklist

Want to rebase onto #766 before generating docs.

### DIFF
--- a/Xamarin.Forms.Core/NavigationMenu.cs
+++ b/Xamarin.Forms.Core/NavigationMenu.cs
@@ -3,11 +3,11 @@ using System.Collections.Generic;
 using System.Linq;
 using Xamarin.Forms.Platform;
 
-namespace Xamarin.Forms
+namespace Xamarin.Forms.Internals
 {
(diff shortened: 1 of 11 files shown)


Xamarin.Forms.Core/NavigationMenu.cs
```diff
@@ -64,6 +64,8 @@ public void Remove(Page target)
 			return _platformConfigurationRegistry.Value.On<T>();
 		}
 
+		void INavigationMenuController.SendTargetSelected(Page target) => SendTargetSelected(target);
+
 		internal void SendTargetSelected(Page target)
```
StephaneDelcroix: why are we keeping the internal one when we now have a public `SendTargetSelected()` ?

PR #529, opened 2016-11-15
[iOS] Labels with WordWrap or CharacterWrap will Expand

samhouts:
### Description of Change ###

iOS `LabelRenderer` will now expand the width of the `Label` to the width constraint if the `Label`'s `LineBreakMode` is set to `WordWrap` (default) or `CharacterWrap`.

### Bugs Fixed ###
- [Bug 28650 - In a Listview on iOS, "andExpand" does not expand when text is two lines long](https://bugzilla.xamarin.com/show_bug.cgi?id=28650)

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #776, opened 2017-02-18
[iOS] Prevent "NaN is not a valid value for width" exception

samhouts:
### Description of Change ###

Fix regression introduced in #529 allowing an infinite width to be returned for `Label's on iOS, which led to an exception in nested and potentially horizontally infinite layouts. 

### Bugs Fixed ###

- [Bug 52533 - Xamarin.Forms v2.3.4.192-pre2 Regression: System.ArgumentException: NaN is not a valid value for width](https://bugzilla.xamarin.com/show_bug.cgi?id=52533)

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


rooby:
This is a fairly intrusive bug, does anyone know when the next release is scheduled?

PR #782, opened 2017-02-23
Remove InternalsVisibleTo from Core to XF.Platforms.*

kingces95:
## Description of Change ###

We hired contractors to build out new backends. They will not live in the XF repo so internal functionality in Core used by our renderers needs to be exposed publicly for use by their renderers. 

Core allows our renderers to access its internal members by applying InternalsVisibleToAttribute (IVT) to itself for each platform assembly. This PR comments out those IVT attributes and fixes the resulting build errors. This is done by either (1) exposing an internal class publicly and changing it's namespace to `Xamarin.Forms.Internals` or (2) exposing a member publicly and adding a `[EditorBrowsable(EditorBrowsableState.Never)]` to the member. This essentially creates a new scope between `internal` and `public` that is expressed as membership in XF.Internals namespace or the presence of `EditorBrowsable` on a public member and is only used by renderer implementors to feed state and events back into core. Customers should expect no support of this scope and should not bind to any of it's members.

The `EditorBrowsable` approach I present as preferable to controller interfaces for hiding members from isense. The attribute is more explicit, doesn't require exposing an additional public interface, nor require changing any of the callsites, is easier to apply, and yet still preserves isesne for the XF team (EditorBrowsable only affects isense of referenced assemblies, not referenced projects). What's not to like! Going forward we can convert the existing controller interfaces to use the attribute approach (recommended) or we could add new controller interfaces on top of the attribute approach as the two are not exclusive. 

IMHO, ultimately, the contractors code should live in the XF repo (maybe in another branch) and each core/platform assembly pair should be merged into a single assembly. This would eliminate this extra scope as well as the need for `ExportRenderer` and forwarder types for the core renderers. This conversion from dynamic to static binding could only speed up our load times. Assembly merging has also been shown to result in smaller assemblies after linking. 

### Bugs Fixed ###

None

### API Changes ###

None. The public surface area we support is unchanged. 

### Behavioral Changes ###

None. The change should result in no behavioral changes. 

### PR Checklist ###


Xamarin.Forms.Core/IPinchGestureController.cs
```diff
@@ -1,6 +1,6 @@
-﻿namespace Xamarin.Forms
+﻿namespace Xamarin.Forms.Internals
 {
-	internal interface IPinchGestureController
+	public interface IPinchGestureController
```
jassmith: Does not need to be in Internals namespace

Xamarin.Forms.Core/IPanGestureController.cs
```diff
@@ -1,6 +1,6 @@
-﻿namespace Xamarin.Forms
+﻿namespace Xamarin.Forms.Internals
 {
-	internal interface IPanGestureController
+	public interface IPanGestureController
```
jassmith: Does not need to be in internals namespace

Xamarin.Forms.Core/MenuItem.cs
```diff
@@ -17,7 +18,8 @@ public class MenuItem : BaseMenuItem, IMenuItemController
 
 		public static readonly BindableProperty IconProperty = BindableProperty.Create("Icon", typeof(FileImageSource), typeof(MenuItem), default(FileImageSource));
 
-		internal static readonly BindableProperty IsEnabledProperty = BindableProperty.Create("IsEnabled", typeof(bool), typeof(ToolbarItem), true);
+		[EditorBrowsable(EditorBrowsableState.Never)]
```
jassmith: Use controller interface for this class

Xamarin.Forms.Core/MultiPage.cs
```diff
@@ -159,15 +161,17 @@ protected virtual void UnhookContent(T content)
 		{
 		}
 
-		internal static int GetIndex(T page)
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public static int GetIndex(T page)
```
jassmith: Use controller interface for this class

Xamarin.Forms.Core/TapGestureRecognizer.cs
```diff
@@ -35,7 +36,8 @@ public int NumberOfTapsRequired
 
 		public event EventHandler Tapped;
 
-		internal void SendTapped(View sender)
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public void SendTapped(View sender)
```
jassmith: Use controller interface for this class

Xamarin.Forms.Core/VisualElement.cs
```diff
@@ -86,7 +87,8 @@ public partial class VisualElement : Element, IAnimatable, IVisualElementControl
 
 		public static readonly BindableProperty MinimumHeightRequestProperty = BindableProperty.Create("MinimumHeightRequest", typeof(double), typeof(VisualElement), -1d, propertyChanged: OnRequestChanged);
 
-		internal static readonly BindablePropertyKey IsFocusedPropertyKey = BindableProperty.CreateReadOnly("IsFocused", typeof(bool), typeof(VisualElement), default(bool),
+		[EditorBrowsable(EditorBrowsableState.Never)]
```
jassmith: Use controller interface for this class

Xamarin.Forms.Core/ListView.cs
```diff
@@ -210,7 +212,8 @@ public SeparatorVisibility SeparatorVisibility
 			set { SetValue(SeparatorVisibilityProperty, value); }
 		}
 
-		internal ListViewCachingStrategy CachingStrategy { get; private set; }
+		[EditorBrowsable(EditorBrowsableState.Never)]
```
jassmith: Use controller interface for this class

Xamarin.Forms.Core/Grid.cs
```diff
@@ -212,7 +212,8 @@ internal override void ComputeConstraintForView(View view)
 			view.ComputedConstraint = result;
 		}
 
-		internal override void InvalidateMeasureInternal(InvalidationTrigger trigger)
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public override void InvalidateMeasureInternal(InvalidationTrigger trigger)
```
jassmith: Lets make a method on the Controller interface which calls this instead.

Xamarin.Forms.Core/Element.cs
```diff
@@ -158,7 +159,8 @@ internal IPlatform Platform
 		}
 
 		// you're not my real dad
-		internal Element RealParent { get; private set; }
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public Element RealParent { get; private set; }
```
jassmith: Use controller interface for this class

Xamarin.Forms.Core/IPanGestureController.cs
kingces95: It was moved because it was previously internal. Will move it back to XF namespace and expose publicly. Suggest you review to make sure it's the API you want exposed.

Xamarin.Forms.Core/IPinchGestureController.cs
kingces95: It was moved because it was previously internal. Will move it back to XF namespace and expose publicly. Suggest you review to make sure it's the API you want exposed.

Xamarin.Forms.Core/MenuItem.cs
kingces95: This is a static member. What convention would you like to use to expose it?

Xamarin.Forms.Core/BindableObject.cs
```diff
@@ -91,7 +91,8 @@ public void SetValue(BindablePropertyKey propertyKey, object value)
 			SetValue(propertyKey.BindableProperty, value, false, false);
 		}
 
-		protected internal static void SetInheritedBindingContext(BindableObject bindable, object value)
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public static void SetInheritedBindingContext(BindableObject bindable, object value)
 		{
```
kingces95: https://github.com/xamarin/Xamarin.Forms/blob/7cbcd2b2608bb9d2d25d9677417f3e9f9fa44800/Xamarin.Forms.Platform.iOS/Platform.cs#L209 + others

Xamarin.Forms.Core/BindableObject.cs
```diff
@@ -171,7 +172,8 @@ internal bool GetIsDefault(BindableProperty targetProperty)
 			return GetContext(targetProperty) == null;
 		}
 
-		internal object[] GetValues(BindableProperty property0, BindableProperty property1)
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public object[] GetValues(BindableProperty property0, BindableProperty property1)
```
kingces95: No references. Revert to internal.

Xamarin.Forms.Core/BindableObject.cs
```diff
@@ -202,7 +204,8 @@ internal object[] GetValues(BindableProperty property0, BindableProperty propert
 			return values;
 		}
 
-		internal object[] GetValues(BindableProperty property0, BindableProperty property1, BindableProperty property2)
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public object[] GetValues(BindableProperty property0, BindableProperty property1, BindableProperty property2)
 		{
```
kingces95: https://github.com/xamarin/Xamarin.Forms/blob/7cbcd2b2608bb9d2d25d9677417f3e9f9fa44800/Xamarin.Forms.Platform.iOS/Renderers/FontExtensions.cs#L141 + others

Xamarin.Forms.Core/BindableObject.cs
```diff
@@ -321,12 +324,14 @@ internal void SetValue(BindableProperty property, object value, bool fromStyle)
 			SetValue(property, value, fromStyle, true);
 		}
 
-		internal void SetValueCore(BindablePropertyKey propertyKey, object value, SetValueFlags attributes = SetValueFlags.None)
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public void SetValueCore(BindablePropertyKey propertyKey, object value, SetValueFlags attributes = SetValueFlags.None)
```
kingces95: No references. Revert to internal.

Xamarin.Forms.Core/BindableObject.cs
```diff
@@ -321,12 +324,14 @@ internal void SetValue(BindableProperty property, object value, bool fromStyle)
...
 		{
 			SetValueCore(propertyKey.BindableProperty, value, attributes, SetValuePrivateFlags.None);
 		}
 
-		internal void SetValueCore(BindableProperty property, object value, SetValueFlags attributes = SetValueFlags.None)
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public void SetValueCore(BindableProperty property, object value, SetValueFlags attributes = SetValueFlags.None)
```
kingces95: https://github.com/xamarin/Xamarin.Forms/blob/7cbcd2b2608bb9d2d25d9677417f3e9f9fa44800/Xamarin.Forms.Core/ListView.cs#L277 + many others

Xamarin.Forms.Core/MenuItem.cs
kingces95: Actually, there already is a controller. Let me know if the public surface area of that controller is not to your liking.

Xamarin.Forms.Core/MultiPage.cs
kingces95: Do you have a preference for how we expose static members? We could just leave them as is for now and if a customer needs those methods just tell them to use the isense hidden ones.

Xamarin.Forms.Core/ListView.cs
kingces95: Already exposed via interface. Really, we should go back and remove the explicit implementations of the controller interfaces from our code, replace them all with the EditorBrowser idiom, and revert all the callsites back to directly accessing the member.

jassmith: IFooController interfaces are there for when customers need to actually use the methods for implementing renderers. Simply hidden from intellisense is there for when they are not. That is the distinction we need to draw with Controller vs EditorBrowsable

Xamarin.Forms.Core/BindableObject.cs
jassmith: This particular call site could just be made to be ((IElementController)Element).SetValueFromRenderer (IsRefreshingProperty, true);

Xamarin.Forms.Core/ListView.cs
kingces95: Yep, got that. I'm suggesting we change how we implement the interfaces (not whether we need them). I'd suggest moving from implicit implementation to explicit implementation with the addition of the EditorBrowser attribute. That way our callsites don't go through the interface and so are cleaner. Customers could do the same but will prob go through the interfaces because that is how they'll get isense as well as know that we're supporting the method.

Xamarin.Forms.Core/BindableObject.cs
kingces95: I propose we do a BindableObject scrub in a subsequent PR and keep this a mechanical change. There's a lot of changes here and it'd be good to get them committed as a block with the invariant that no non-trivial control flow changes were made. Then a later investigator could look at the note and dismiss the PR as the source of whatever bug she's looking at.

Xamarin.Forms.Core/NumericExtensions.cs
```diff
@@ -1,14 +1,18 @@
...
+	public static class NumericExtensions
 	{
+
+		[EditorBrowsable(EditorBrowsableState.Never)]
 		public static double Clamp(this double self, double min, double max)
 		{
 			return Math.Min(max, Math.Max(self, min));
 		}
 
+		[EditorBrowsable(EditorBrowsableState.Never)]
```
StephaneDelcroix: idem

Xamarin.Forms.Core/Performance.cs
```diff
@@ -1,22 +1,25 @@
...
 using System.Diagnostics;
 using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Text;
 
-namespace Xamarin.Forms
+namespace Xamarin.Forms.Internals
 {
-	internal static class Performance
+	public static class Performance
```
StephaneDelcroix: this doesn't need to be public, it's a tool we use but other implementors might want to use other ones

Xamarin.Forms.Core/ReflectionExtensions.cs
```diff
@@ -1,32 +1,38 @@
...
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Linq;
 using System.Reflection;
 
-namespace Xamarin.Forms
+namespace Xamarin.Forms.Internals
 {
-	internal static class ReflectionExtensions
+	public static class ReflectionExtensions
```
StephaneDelcroix: I think this is only used by the Binding system and the Xaml parser. no need to open it. everyone can rewrite the same

Xamarin.Forms.Core/VisualElement.cs
kingces95: We agreed statics will just get the Editor attribute.

Xamarin.Forms.Core/Element.cs
kingces95: The controller interface declares RealParent so it is on the controller interface.

Xamarin.Forms.Core/BindableObject.cs
kingces95: I'd guess the `params` keyword was not used here in order to prevent GC allocation of the array; that signature could be an optimization. For this PR, I'd prefer not to make any control flow or signature modifications to ensure there are no regressions. More to the point, this API was not requested to be put on a controller iface. So this change is not expanding our supported surface area.

Xamarin.Forms.Core/Application.cs
```diff
@@ -48,11 +49,13 @@ public IAppLinks AppLinks
 			}
 		}
 
+		[EditorBrowsable(EditorBrowsableState.Never)]
```
kingces95: Yes

Xamarin.Forms.Core/ActionSheetArguments.cs
```diff
@@ -1,9 +1,9 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
 
-namespace Xamarin.Forms
+namespace Xamarin.Forms.Internals
 {
-	internal class ActionSheetArguments
+	public class ActionSheetArguments
```
kingces95: Online doc?

Xamarin.Forms.Core/Application.cs
```diff
@@ -48,11 +49,13 @@ public IAppLinks AppLinks
...
+		public static void SetCurrentApplication(Application value) => Current = value;
+
 		public static Application Current
 		{
 			get { return s_current; }
-			internal set
-			{
+			set {
```
kingces95: Hehe, no such luck...

Xamarin.Forms.Core/Color.cs
```diff
@@ -22,11 +24,14 @@ public static Color Default
 			get { return new Color(-1d, -1d, -1d, -1d, Mode.Default); }
 		}
 
-		internal bool IsDefault
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public bool IsDefault
```
kingces95: I'd rather this PR not change the public surface area exposed to non-platform implementors. Let's leave those for another PR.

Xamarin.Forms.Core/Color.cs
```diff
@@ -22,11 +24,14 @@ public static Color Default
...
 		{
 			get { return _mode == Mode.Default; }
 		}
 
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public static void SetAccent(Color value) => Accent = value;
```
kingces95: Before the EditorBrowsable we would have done that because moving the method to the internals namespace as an extension was the only way to indicate to customers they shouldn't bind to the method. With the browser attribute we can leave the logic where it was originally placed. Not saying it shouldn't be an extension for other reasons but as far as IVT is concerned this change is smaller and gets the job done.

Xamarin.Forms.Core/DeviceInfo.cs
```diff
@@ -2,17 +2,18 @@
 using System.ComponentModel;
 using System.Runtime.CompilerServices;
 
-namespace Xamarin.Forms
+namespace Xamarin.Forms.Internals
 {
-	internal abstract class DeviceInfo : INotifyPropertyChanged, IDisposable
+	public abstract class DeviceInfo : INotifyPropertyChanged, IDisposable
```
kingces95: Yeah, good idea. Make 'em even harder to bind to. Let's do that in a separate change PR right after we merge this one.

Xamarin.Forms.Core/Grid.cs
```diff
@@ -212,6 +212,11 @@ internal override void ComputeConstraintForView(View view)
 			view.ComputedConstraint = result;
 		}
 
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public void InvalidateMeasureInernalNonVirtual(InvalidationTrigger trigger)
```
kingces95: As far as I can tell, the only reason we explicitly implemented the methods on the controller interface was to hide them from isense. Unfortunately, this also had the side effect of breaking all our call sites and requiring us to sprinkle the idiom Foo.Bar -> ((IController)Foo).Bar everywhere. By using the browser attribute we get the desired design effect (hide from isense) but we don't break our call sites which can simply remain Foo.Bar. So I'd actually suggest we revert all explicit implementations and add the browser attribute and at the same time revert the callsites back from ((IController)Foo).Bar to simply Foo.Bar.

Xamarin.Forms.Core/NavigationProxy.cs
```diff
@@ -1,17 +1,19 @@
...
 {
-	internal class NavigationProxy : INavigation
+	public class NavigationProxy : INavigation
 	{
 		INavigation _inner;
 		Lazy<List<Page>> _modalStack = new Lazy<List<Page>>(() => new List<Page>());
 
 		Lazy<List<Page>> _pushStack = new Lazy<List<Page>>(() => new List<Page>());
 
+		[EditorBrowsable(EditorBrowsableState.Never)]
```
kingces95: Good point. On the subsequent Internals scrub I'll check to see if adding the attribute to the class simply hides all the members.

Xamarin.Forms.Core/NumericExtensions.cs
kingces95: Agreed. Could be. But let's save changing the public surface area for another PR. The goal of this PR is just to remove IVT and as we can see that's a pretty big change as is.

Xamarin.Forms.Core/Performance.cs
kingces95: It's in the internals namespace so its not part of our supported public surface.

Xamarin.Forms.Core/ReflectionExtensions.cs
kingces95: Again, in the internals namespace so it's not part of the supported public surface area. It's prob here so it can be shared across all the platforms. If we want to refactor to pull this into each platform or into a utils assembly we can do that in a separate PR.

Xamarin.Forms.Core/BindableObject.cs
kingces95: As the API is browser attributed it doesn't constitute part of our supported public surface area and as such could be changed back to internal at a later state. I have no objection to doing SetvalueCore -> SetValueFromRenderer scrub but just not as part of this PR; conversion to SetValueFromRenderer is not necessary for IVT removal and as a non-trivial change could potentially break something.

mdizzy:
Out of curiosity, will this be published in an official release? I'm building some renderers that need things like SendTapped on the TapGestureRecognizer, basically I'm creating irregular shaped views and in order to do so need to implement custom gesture handling.

rmarinho:
@mdizzy is already in public beta 2.3.5-pre

PR #785, opened 2017-02-25
Fix for Android Visibility/Opacity crash 

hartez:
### Description of Change ###

Restoring the visibility of Views which have an Alpha value between 0 and 1 causes a crash due to a negative width/height in `VisualElementTracker`'s `UpdateLayout` method.

### Bugs Fixed ###

- [51238 – Transparent Grid causes Java.Lang.IllegalStateException: Unable to create layer for Platform_DefaultRenderer](https://bugzilla.xamarin.com/show_bug.cgi?id=51238)

### API Changes ###

None

### Behavioral Changes ###

None 

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla51238.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla51238.cs
@@ -1,6 +1,10 @@
-﻿using NUnit.Framework;
-using Xamarin.Forms.CustomAttributes;
+﻿using Xamarin.Forms.CustomAttributes;
 using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
 
 namespace Xamarin.Forms.Controls.Issues
 {
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla51238.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla51238.cs
@@ -2,44 +2,17 @@
 using Xamarin.Forms.Internals;
 
 #if UITEST
-using Xamarin.UITest;
 using NUnit.Framework;
 #endif
 
 namespace Xamarin.Forms.Controls.Issues
 {
 	[Preserve(AllMembers = true)]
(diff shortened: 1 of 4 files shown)


adrianknight89:
OP indicated that the issue started happening in 2.3.3-pre3. Looking at commit history and stacktrace, it might have something to do with https://github.com/xamarin/Xamarin.Forms/commit/87a79d14aa28bf77b70ad889a37964202ed4a302.

```
01-20 11:23:35.923 E/AndroidRuntime(24322): java.lang.IllegalStateException: Unable to create layer for Platform_DefaultRenderer
	at android.os.MessageQueue.nativePollOnce(Native Method)
	at android.os.MessageQueue.next(MessageQueue.java:323)
	at android.os.Looper.loop(Looper.java:143)
	at android.app.ActivityThread.main(ActivityThread.java:7229)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1230)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1120)
```

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla51238.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Bugzilla, 51238,
+		"Transparent Grid causes Java.Lang.IllegalStateException: Unable to create layer for Platform_DefaultRenderer",
+		PlatformAffected.Android)]
+	public class Bugzilla51238 : TestContentPage
+	{
+#if UITEST
```
samhouts: You're freaking me out, having this first. :P

hartez:
@adrianknight89 Thanks for tracking down that change - you're right, the bug did show up then. Turns out that change revealed a problem with negative size values in `VisualElementTracker`.

hartez:
@samhouts Can I get another review for the new fix? This one has a _lot_ less code :)

adrianknight89:
Great :)

Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
```diff
@@ -3,6 +3,8 @@
 using System.Collections.ObjectModel;
 using System.Collections.Specialized;
 using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
```
samhouts: Don't think we need these.

PR #787, opened 2017-02-26
Bug 51536, initial iOS cell layout is incorrect when cell has context actions …

VitalyKnyazev:
### Description of Change ###
Bug 51536, initial cell content is not aligned properly when cell has context actions or ListView is in recycle mode, but content gets aligned properly after scrolling out and back. See attached screenshots and sample project, add it to Xamarin.Forms solution.

### Bugs Fixed ###
https://bugzilla.xamarin.com/show_bug.cgi?id=51536

### API Changes ###
None

### Behavioral Changes ###
Not sure why this check was there in the first place, could not find any side effects.

### PR Checklist ###

before
after
[App1.zip](https://github.com/xamarin/Xamarin.Forms/files/802417/App1.zip)


Xamarin.Forms.Platform.iOS/ContextActionCell.cs
```diff
@@ -87,7 +87,7 @@ public override void LayoutSubviews()
 		{
 			base.LayoutSubviews();
 
-			if (_scroller == null || (_scroller != null && _scroller.Frame.Width == ContentView.Bounds.Width))
```
VitalyKnyazev: When "_scroller.Frame.Width == ContentView.Bounds.Width" is true then no further layout is happening and cell view is broken

MichaelRumpler:
There is another bug regarding Cell layout and ContextActions on iOS: https://bugzilla.xamarin.com/show_bug.cgi?id=33450
Will this be fixed with this PR too?

VitalyKnyazev:
> @rmarinho Maybe the fix is check if the height is also the same?

Thanks, check for both width and height works too, updated PR.

VitalyKnyazev:
@rmarinho Indentation fixed, added test case but not sure it is as expected.

rmarinho:
@VitalyKnyazev yap that's it, but seems that test file doesn't have the propper formating. We use tabs. You can read on the Readme about coding style .

VitalyKnyazev:
@rmarinho Fixed, some lines were copy/pasted.

rmarinho:
There's 1 test failing 

Bugzilla31330 can you take a look? 

Thanks

VitalyKnyazev:
@samhouts done, please review

gentledepp:
I just found out that this actually is our issue!
It would be so great to have it fixed!!! Thanks @VitalyKnyazev

gentledepp:
@samhouts 
VitalyKnyazev renamed the bug. Could you now please confirm? We are eagerly waiting for that fix. :-|

gentledepp:
His lead to the following regression bug: https://bugzilla.xamarin.com/show_bug.cgi?id=53834

Don't you guys have Xamarin.UITests for xamarin.forms?
I mean you are writing and talking about "quality improvements" all the time nowerdays - I'm just wondering how you are supposed to achieve that without tests? :-|

PR #802, opened 2017-03-07
Setup a ConditionalWeakTable in ListProxy to hold weak references so that a ListView will work when connected to a Weak Source

PureWeen:
### Description of Change ###

If a collection implements INotifyCollectionChanged with a Weak Reference then the ListProxy's subscription will get finalized prematurely by the GC since there are no roots holding onto the WeakProxy or the CollectionChanged subscription. I added a ConditionalWeakTable keyed on the ListProxy to hold the WeakProxy and the NotifyCollectionChangedEventHandler this way when there are no more roots holding onto ListProxy they will still get disposed. All current unit tests pass and I added an additional one that demonstrates the bug.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=31415
I first came across this issue when trying to use a ReactiveList with Xamarin Forms
https://github.com/reactiveui/ReactiveUI/issues/806

### API Changes ###

None

### Behavioral Changes ###

There shouldn't be any. When there are no more roots referencing the ListProxy the subscriptions will get cleaned up as is demonstrated by the unit tests.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/ListProxyTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/ListProxyTests.cs
@@ -422,5 +422,97 @@ namespace Xamarin.Forms.Core.UnitTests
 				return Items.GetEnumerator ();
 			}
 		}
+
+		[Test]
+		public void WeakToWeak()
+		{
+			WeakCollectionChangedList list = new WeakCollectionChangedList();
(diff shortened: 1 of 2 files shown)


ghuntley:
<3

PR #806, opened 2017-03-09
Add EditorBrowsableAttribute to public classes/interfaces in XF.Internals

kingces95:
### Description of Change ###

Add `[EditorBrowsable(EditorBrowsableState.Never)]` to classes and interfaces in XF.Internals hides these classes and all their members from isense when XF.Core is referenced as an assembly. Our policy is that customers should not bind to these members. Hiding them in this fashion makes it more difficult for them to bind to them by mistake. 

When referenced as a csproj the members will still appear in isense. 

### Bugs Fixed ###

None

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/NativeBindingHelpers.cs
+++ b/Xamarin.Forms.Core/NativeBindingHelpers.cs
@@ -10,9 +10,9 @@ using static System.String;
 
 namespace Xamarin.Forms.Internals
 {
+	[EditorBrowsable(EditorBrowsableState.Never)]
 	public static class NativeBindingHelpers
 	{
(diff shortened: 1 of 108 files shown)


Xamarin.Forms.Core/NativeBindingHelpers.cs
```diff
@@ -10,9 +10,9 @@
 
 namespace Xamarin.Forms.Internals
 {
+	[EditorBrowsable(EditorBrowsableState.Never)]
 	public static class NativeBindingHelpers
```
StephaneDelcroix: why is this public again ?

PR #807, opened 2017-03-09
Implicit impl of controller ifaces

kingces95:
### Description of Change ###

Convert from explicit to implicit implementation + `[EditorBrowsable(EditorBrowsableState.Never)]` of controller interfaces. Either way the members of the controller iface are hidden from customers however the latter has the advantage that it does not hide the members from the XF team (we use a ProjectReference to reference core and in that case the attribute is ignored by isense) which allows us to simplify all our call sites from `((ICtrlIface)Foo).Bar` to `Foo.Bar`. I've update the callsites for only one controller interface in this PR. If/when consensus has been reached on this change of convention then I'll merge this PR and create a second PR that updates the remaining callsites. 

### Bugs Fixed ###

None

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/CellTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/CellTests.cs
@@ -44,7 +44,7 @@ namespace Xamarin.Forms.Core.UnitTests
 			bool emitted = false;
 			cell.Appearing += (sender, args) => emitted = true;
 
-			((ICellController)cell).SendAppearing ();
+			cell.SendAppearing ();
 			Assert.True (emitted);
 			Assert.True (cell.OnAppearingSent);
 			Assert.False (cell.OnDisappearingSent);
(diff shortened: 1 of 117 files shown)


StephaneDelcroix:
I'm leaning on the other way. Implements all internals interfaces explicitly.

Our platform code is, or will be, used as reference for other implementation and we should consume the exact same API

kingces95:
I'd argue that we still consume the same the API as our customers. The difference is in _how_ the API is consumed. We consume the core API directly (which, IMHO, results in more readable code) while our customers consume the API through an interface -- but in either case we're consuming the same API. 

On the other hand, your point is well taken that the implementation of the renderer and custom renderer do not look the same even if they are consuming the same API. It just that IMHO the benefit of allowing customers to copy/paste from our renderers doesn't outweigh the cost of the less readable callsites.

@jassmith, it's your show, let me know if you'd like me to proceed simplifying the callsites or to hold off.

rmarinho:
Can you rebase @kingces95

PR #483, opened 2016-10-22
[Android] UI tests for InputTransparent and IsEnabled fixes

adrianknight89:
### Description of Change

I'm able to run Android UITests now and wanted to submit tests for #344. In the process, I realized this needed a better fix.

### Bugs Fixed

- https://bugzilla.xamarin.com/show_bug.cgi?id=46458

### PR Checklist
- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #808, opened 2017-03-10
UI tests for InputTransparent and fixes for Android/Windows

hartez:
### Description of Change ###

This change adds UI tests for the InputTransparent property on all of the controls. It also makes the behavior of InputTransparent on Android consistent with the other platforms (prior to this, most of the Android controls were _always_ input transparent).

This change also incorporates the UI tests and part of the Android fixes from [PR 483](https://github.com/xamarin/Xamarin.Forms/pull/483) (with some slight modifications). Getting those tests passing on Windows ended up incidentally fixing some other open issues.

Also, replaces ugly MSPaint test image with a picture of my brother's dog.

### Bugs Fixed ###

- [44096 – Controls still participate in hit testing when inside a Grid, StackLayout, and ContentView that has IsEnabled = false](https://bugzilla.xamarin.com/show_bug.cgi?id=44096) (for Windows)
- [43007 – Controls nested in StackLayout need to inherit IsEnabled property](https://bugzilla.xamarin.com/show_bug.cgi?id=43007) (For Windows)
- [46458 – Grid.IsEnabled property is not working](https://bugzilla.xamarin.com/show_bug.cgi?id=46458) (For Windows)
- [44176 – InputTransparent fails if BackgroundColor not explicitly set on Android](https://bugzilla.xamarin.com/show_bug.cgi?id=44176) 

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
```diff
@@ -93,7 +93,7 @@ void IOnClickListener.OnClick(AView v)
 		public override bool OnInterceptTouchEvent(MotionEvent ev)
 		{
 			if (Element.InputTransparent && Element.IsEnabled)
-				return false;
+				return true;
```
samhouts: Ok. So, this is what I think you're saying. The `Button` is a child of the `ViewGroup`. It receives a touch event. That event hits this method first to see if it should flow through to the `Button`. Previously, if the `Element` was `InputTransparent`, it _would_ flow through to the `Button`. But the `Button` did not handle `onTouch`, so it wouldn't actually respect the `InputTransparent` value. Now, this `ViewGroup` will intercept the event and use _its_ `onTouch` handler, which _does_ take the `InputTransparent` value into account. Cool.

adrianknight89:
@hartez  I submitted #483 a while back. Please take that into account as well. Also `IsEnabled` property doesn't seem to be working well on Android.

rmarinho:
The InputTransparentTests are failing @hartez

hartez:
@adrianknight89 Apologies, I hadn't realized there was already a PR dealing with this. I'll spend some time tomorrow pulling in your UI tests and making sure my changes didn't break any of them (or, if they did, that they were supposed to).

hartez:
This now includes the tests from [PR 483](https://github.com/xamarin/Xamarin.Forms/pull/483).

Xamarin.Forms.Platform.Android/Renderers/MotionEventHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+			// Let the container know that we're "fake" handling this event
+			renderer.NotifyFakeHandling();
+
+			return true;
+		}
+
+		public void UpdateElement(VisualElement element)
+		{
+			_isInViewCell = false;
+			_element = element;
```
rmarinho: I worry we will be keeping this reference around since we don't clear or set to null our instance of MotionEventHelper.

Xamarin.Forms.Platform.Android/Renderers/LabelRenderer.cs
```diff
@@ -23,6 +23,8 @@ public class LabelRenderer : ViewRenderer<Label, TextView>
 		FormsTextView _view;
 		bool _wasFormatted;
 
+		readonly MotionEventHelper _motionEventHelper = new MotionEventHelper();
```
rmarinho: Can we remove readonly and set this to null on Dispose ? Just worry we could leak by setting a element reference on MotionEventHelper

rmarinho:
@hartez can you resolve this conflit, don't want to make the wrong choice :)

rmarinho:
Rebased with changes i did to used shared project for tests

Xamarin.Forms.Platform.Android/Renderers/LabelRenderer.cs
hartez: Which object are you worried about leaking?

rmarinho: The renderer not being colected because MotionEventHelper will keep a reference to it.

rmarinho: Sorry.. it keeps a reference to the element.. not the renderer.

hartez: The only thing which holds a reference to MotionEventHelper is the renderer; when the renderer is collected, MotionEventHelper goes away, too. Setting it to null during dispose doesn't make any difference.

PR #811, opened 2017-03-13
UWP MapRenderer NRE and Location fixes

erdennis13:
### Description of Change ###

1. If UWP MapRenderer is navigated away from, before the user location
is found, an uncaught NRE will be thrown.
2. UWP MapRenderer updates are not enforced to be on UI thread
3. Order of operations in MapRenderer would cause user location to never
be shown.

Fixes:
1. Added null checks in UpdateUserIsShowing and LoadUserPosition
methods.
2. Called map position updates on UI thread.
3. Changed order of operations in OnElementChanged to allow for user
geocoordinates to be shown.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=53113

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Maps.UWP/MapRenderer.cs
+++ b/Xamarin.Forms.Maps.UWP/MapRenderer.cs
@@ -62,10 +62,10 @@ namespace Xamarin.Forms.Maps.WinRT
 				if (mapModel.Pins.Any())
 					LoadPins();
 
-                if (Control != null)
-                    await Control.Dispatcher.RunIdleAsync(async (i) => await MoveToRegion(mapModel.LastMoveToRegion, MapAnimationKind.None));
+                if (Control == null) return;
 
-                await UpdateIsShowingUser();
+				await Control.Dispatcher.RunIdleAsync(async (i) => await MoveToRegion(mapModel.LastMoveToRegion, MapAnimationKind.None));
+				await UpdateIsShowingUser();
             }
 		}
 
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Maps.UWP/MapRenderer.cs
```diff
@@ -62,8 +62,10 @@ public class MapRenderer : ViewRenderer<Map, MapControl>
 				if (mapModel.Pins.Any())
 					LoadPins();
 
-				await UpdateIsShowingUser();
-				await Control.Dispatcher.RunIdleAsync(async (i) => await MoveToRegion(mapModel.LastMoveToRegion, MapAnimationKind.None));
+                if (Control != null)
```
rmarinho: maybe if(Control == null) return;

rmarinho:
Some issues with formatting, please only commit the changes needed.

erdennis13:
@rmarinho code formatting issues are resolved

Xamarin.Forms.Maps.UWP/MapRenderer.cs
```diff
@@ -256,15 +261,15 @@ void LoadUserPosition(Geocoordinate userCoordinate, bool center)
 				};
 			}
 
-			if (Control.Children.Contains(_userPositionCircle))
+			if (Control?.Children?.Contains(_userPositionCircle) == true)
```
rmarinho: maybe it's better in the beginning of the method to do
``` 
if (Control == null || Element == null) return; 
```

erdennis13: Agreed. This is reflected with commit

Xamarin.Forms.Maps.UWP/MapRenderer.cs
```diff
@@ -235,6 +240,8 @@ void UpdateVisibleRegion()
 
 		void LoadUserPosition(Geocoordinate userCoordinate, bool center)
 		{
+			if (Control == null || Element == null) return;
```
rmarinho: Sorry reviewing this again, can we do the same we did here on 
so we can keep the code as it is there.. 

`UpdateIsShowingUser`

Xamarin.Forms.Maps.UWP/MapRenderer.cs
```diff
@@ -176,20 +177,22 @@ void LoadPin(Pin pin)
 						LoadUserPosition(userPosition.Coordinate, moveToLocation);
 				}
 
+				if (Control == null || Element == null) return;
```
erdennis13: @rmarinho I'm happy to keep the null checks consistent. The reason I included a check on both lines 180 and 167, is because if the user exits the Map page while the geolocater is still operating, the check on line 167 would not be enough.

Xamarin.Forms.Maps.UWP/MapRenderer.cs
```diff
@@ -62,9 +62,11 @@ public class MapRenderer : ViewRenderer<Map, MapControl>
 				if (mapModel.Pins.Any())
 					LoadPins();
 
-				await UpdateIsShowingUser();
+                		if (Control == null) return;
+
 				await Control.Dispatcher.RunIdleAsync(async (i) => await MoveToRegion(mapModel.LastMoveToRegion, MapAnimationKind.None));
```
BladeMF: Guys, LastMoveToRegion can be null and it isn't checked anywhere here.

Xamarin.Forms.Maps.UWP/MapRenderer.cs
```diff
@@ -174,6 +177,8 @@ void LoadPin(Pin pin)
 						LoadUserPosition(userPosition.Coordinate, moveToLocation);
```
BladeMF: The above code (using myGeolocator) crashes if the location is not enabled. LocationStatus returns PositionStatus.NotInitialized.

PR #814, opened 2017-03-14
Bug 53275 - IProvideValueTarget TargetProperty always null

workgroupengineering:
### Description of Change ###

Implements discovery of IProvideValueTarget.TargetProperty 

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=53275: When in custom MarkupExtension you using IProvideValueTarget to get TargetProperty is always null, Because in this [line](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Xaml/XamlServiceProvider.cs#L17 )  you always setted null the targetProperty parameter of XamlServiceProvider constructor.

### API Changes ###

Nothing.

### Behavioral Changes ###

The discovery of TargetProperty happens in lazy mode. When it called for the first time TargetProperty Check That the TargetObject exists in static field That is of type BindableProperty and That its name is equal to the name of the property plus the postfix "Property", if it is not found performs the search for a property with same name we are searching.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Xaml.UnitTests/Issues/Bz53275.cs
+++ b/Xamarin.Forms.Xaml.UnitTests/Issues/Bz53275.cs
@@ -1,0 +1,72 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using System.Xml;
+
+using Xamarin.Forms.Core.UnitTests;
+using System.Reflection;
+
+namespace Xamarin.Forms.Xaml.UnitTests.Issues
+{
+	class GetTargetPropertyName : IMarkupExtension
+	{
+		public object ProvideValue(IServiceProvider provider)
+		{
+
+			if (provider == null) return false;
+
+			//create a binding and assign it to the target
+			var service = provider.GetService(typeof(IProvideValueTarget)) as IProvideValueTarget;
+			if (service == null) return false;
+
+			//we need dependency objects / properties
+			var propertyName = string.Empty;
+			var property = service.TargetProperty;
+			if (property is BindableProperty)
+			{
+				propertyName = ((BindableProperty)property).PropertyName;
+			}
+			else if(property is PropertyInfo)
+			{
+				propertyName = ((PropertyInfo)property).Name;
+			}
+			return propertyName;
+		}
+	}
+
+	[TestFixture]
+	class Bz53275:BaseTestFixture
+	{
+		IXamlTypeResolver typeResolver;
+
+		[SetUp]
+		public override void Setup()
+		{
+			base.Setup();
+			var nsManager = new XmlNamespaceManager(new NameTable());
+			nsManager.AddNamespace("local", "clr-namespace:Xamarin.Forms.Xaml.UnitTests.Issues;assembly=Xamarin.Forms.Xaml.UnitTests");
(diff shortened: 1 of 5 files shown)


StephaneDelcroix:
Thanks for this PR. The `TargetProperty` is not `always null`. It's null only with XamlC off. I haven't yet looked at your code as you still have to sign the CLA, but using a lazy doesn't appear to be the best approach

workgroupengineering:
Hi, Thanks for clarification.
Why using a lazy doesn't appear to be the best approach?

Xamarin.Forms.Xaml/XamlServiceProvider.cs
```diff
@@ -91,19 +92,63 @@ public void Add(Type type, object service)
 
 	class XamlValueTargetProvider : IProvideParentValues, IProvideValueTarget
 	{
+		
+		Lazy<object> targetProperty;
 		public XamlValueTargetProvider(object targetObject, INode node, HydratationContext context, object targetProperty)
```
StephaneDelcroix: the targetProperty should come from the caller, not induced by this method. Also, this code ignore the parameter value.

Xamarin.Forms.Xaml.UnitTests/Issues/Bz53275.cs
```diff
@@ -0,0 +1,72 @@
...
+			var nsManager = new XmlNamespaceManager(new NameTable());
+			nsManager.AddNamespace("local", "clr-namespace:Xamarin.Forms.Xaml.UnitTests.Issues;assembly=Xamarin.Forms.Xaml.UnitTests");
+			nsManager.AddNamespace("x", "http://schemas.microsoft.com/winfx/2006/xaml");
+
+			typeResolver = new Internals.XamlTypeResolver(nsManager, XamlParser.GetElementType, Assembly.GetCallingAssembly());
+		}
+
+		[Test]
+		public void TestGetTargetProperty()
+		{
```
StephaneDelcroix: this is not testing an equal behavior with or without XamlC on.

StephaneDelcroix:
if you don't get what I mean in the comment, I can have a look a propose a different fix

Xamarin.Forms.Xaml/XamlServiceProvider.cs
workgroupengineering: >the targetProperty should come from the caller, not induced by this method. 

I don't put out constructor, because i don't know all Xamarin framework and I don't know the whole part where the constructor XamlValueTargetProvider is called.
>Also, this code ignore the parameter value.

It 's true in part, targetProperty is ignored only when it is null. Then see the line 106.

Xamarin.Forms.Xaml.UnitTests/Issues/Bz53275.cs
workgroupengineering: Because, i am a beginner on Xamarin Framework.
Can you tell me how can i do testing for XamlC?
Thank you.

StephaneDelcroix:
hey @workgroupengineering, I'm working on this in parallel, trying to figure out the best approach to solve this

workgroupengineering:
@StephaneDelcroix  Thanks for the [suggestion](http://bit.ly/2m3JDWX).

workgroupengineering:
Hi @StephaneDelcroix , any news?

StephaneDelcroix:
See #847. @workgroupengineering close this PR if you're fine with the alternative.

PR #815, opened 2017-03-14
Resource loading

StephaneDelcroix:
### Description of Change ###

More generic resourceLoading mechanism for the Previewer, so the same API can be reused to load different kind of resources.

### Bugs Fixed ###

/

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/XamlCTask.cs
+++ b/Xamarin.Forms.Build.Tasks/XamlCTask.cs
@@ -176,7 +176,7 @@ namespace Xamarin.Forms.Build.Tasks
 
 						Logger.LogString(2, "   Replacing {0}.InitializeComponent ()... ", typeDef.Name);
 						Exception e;
-						if (!TryCoreCompile(initComp, initCompRuntime, rootnode, out e)) {
+						if (!TryCoreCompile(initComp, initCompRuntime, rootnode, resource.Name, out e)) {
 							success = false;
 							Logger.LogLine(2, "failed.");
 							(thrownExceptions = thrownExceptions ?? new List<Exception>()).Add(e);
@@ -194,6 +194,8 @@ namespace Xamarin.Forms.Build.Tasks
 							initComp.Body.Optimize();
 							Logger.LogLine(2, "done");
 						}
+
+						Logger.LogLine(2, "");
 
 						if (outputGeneratedILAsCode)
 							Logger.LogLine(2, "   Decompiling option has been removed. Use a 3rd party decompiler to admire the beauty of the IL generated");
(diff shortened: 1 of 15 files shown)


Xamarin.Forms.Build.Tasks/XamlCTask.cs
```diff
@@ -195,6 +195,8 @@ public override bool Execute(out IList<Exception> thrownExceptions)
 							Logger.LogLine(2, "done");
 						}
 
+						Logger.LogLine(2, "");
```
alanmcgovern: Is this a debugging helper?

Xamarin.Forms.Xaml/XamlLoader.cs
```diff
@@ -127,10 +131,14 @@ static void Visit (RootNode rootnode, HydratationContext visitorContext)
...
 			//the Previewer might want to provide it's own xaml for this... let them do that
-			if (Xamarin.Forms.Xaml.Internals.XamlLoader.XamlFileProvider != null && (xaml = Xamarin.Forms.Xaml.Internals.XamlLoader.XamlFileProvider(type)) != null)
+			//the check at the end is preferred (using ResourceLoader). keep this until all the previewers are updated
+
+#pragma warning disable 0618
+			var xaml = Internals.XamlLoader.XamlFileProvider?.Invoke(type);
+#pragma warning restore 0618
+
+			if (xaml != null && ResourceLoader.ResourceProvider != null)
 				return xaml;
```
alanmcgovern: Why are you checking that `ResourceProvider != null` here? I presume the idea is that we will *either* be using `XamlFileProvider` *or* we'll be using `ResourceLoader.ResourceProvider` which means if xaml is non-null then ResourceProvider will always be null.

Xamarin.Forms.Build.Tasks/XamlCTask.cs
StephaneDelcroix: no, it's not. adding that white line makes the build output clearer

Xamarin.Forms.Xaml/XamlLoader.cs
StephaneDelcroix: this `ResourceProvider != null` check is not really required, it's only make sure that if you were foolish enough to set both, the ResourceProvider will prevail.

alanmcgovern: I think you've just hit the nail on the head - if you set both, and XamlFileProvider does give valid data, then ResourceProvider will *not* prevail. Was that expected?

StephaneDelcroix: 🤦‍♂️

StephaneDelcroix: this is a case when I should rebase to hide my embarrassment

PR #824, opened 2017-03-20
Convert Android maps to GetMapAsync call

jassmith:
### Description of Change ###

Move to GetMapAsync API for better forwards compatibility with map renderer on Android

### Bugs Fixed ###

None

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Maps.Android/MapRenderer.cs
+++ b/Xamarin.Forms.Maps.Android/MapRenderer.cs
@@ -165,13 +165,19 @@ namespace Xamarin.Forms.Maps.Android
 
 			if (_init)
 			{
-				MoveToRegion(Element.LastMoveToRegion, false);
-				OnCollectionChanged(Element.Pins, new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
-				_init = false;
+				if (NativeMap != null)
+				{
+					MoveToRegion(Element.LastMoveToRegion, false);
+					OnCollectionChanged(Element.Pins, new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
+					_init = false;
+				}
 			}
 			else if (changed)
 			{
-				UpdateVisibleRegion(NativeMap.CameraPosition.Target);
+				if (NativeMap != null)
+				{
+					UpdateVisibleRegion(NativeMap.CameraPosition.Target);
+				}
 				MoveToRegion(Element.LastMoveToRegion, false);
 			}
 		}
--- a/Xamarin.Forms.Maps.Android/MapRenderer.cs
+++ b/Xamarin.Forms.Maps.Android/MapRenderer.cs
@@ -16,7 +16,7 @@ using Math = System.Math;
 namespace Xamarin.Forms.Maps.Android
 {
 	public class MapRenderer : ViewRenderer<Map, MapView>,
-		GoogleMap.IOnCameraChangeListener
+		GoogleMap.IOnCameraChangeListener, IOnMapReadyCallback
 	{
 		const string MoveMessageName = "MapMoveToRegion";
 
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Maps.Android/MapRenderer.cs
```diff
@@ -81,7 +80,8 @@ protected override void Dispose(bool disposing)
  					NativeMap.SetOnCameraChangeListener(null);
  					NativeMap.InfoWindowClick -= MapOnMarkerClick;
  					NativeMap.Dispose();
-  				}		  				
+					NativeMap = null;
+				 }
```
StephaneDelcroix: your keyboard is auto-inserting spaces randomly.

kentcb:
Just tried this. Whilst it does indeed fix the crash, now none of the standard stuff works, like `HasZoomEnabled`, `HasScrollEnabled`, and `MoveToRegion`. @jassmith  Is this intentional? Is it being fixed in two stages?

kentcb:
Actually, the problem is more subtle. We have a custom map renderer that extends the in-built one. However, we also need to know when the map is ready, so we too implement `IOnMapReadyCallback` and call `GetMapAsync` in our `OnElementChanged`. This seems to prevent the base class `OnMapReady` from executing. I suspect our renderer is doing the wrong thing and am gonna overhaul it - will report back.

kentcb:
Yeah, on closer inspection I think it would be best if we just didn't inherit the built-in renderer at all. It doesn't seem to be designed for extensibility. For example, we need greater control over icons used for the pins, but there is no hook for the generation of `MarkOptions`. I'm assuming this is why we created a custom renderer in the first place (not sure because I didn't write it).

Will try using the built-in renderer code as a starting point for our own renderer instead.

rmarinho:
@kentcb feel free to make a pr to add for example a virtual method so one could override and extend on the custom renderer a marker creation.

Xamarin.Forms.Maps.Android/MapRenderer.cs
```diff
@@ -372,5 +357,23 @@ void UpdateVisibleRegion(LatLng pos)
...
+
+			map.SetOnCameraChangeListener(this);
+			map.InfoWindowClick += MapOnMarkerClick;
+
+			map.UiSettings.ZoomControlsEnabled = Map.HasZoomEnabled;
+			map.UiSettings.ZoomGesturesEnabled = Map.HasZoomEnabled;
+			map.UiSettings.ScrollGesturesEnabled = Map.HasScrollEnabled;
+			map.MyLocationEnabled = map.UiSettings.MyLocationButtonEnabled = Map.IsShowingUser;
+			SetMapType();
+		}
```
jcmanke: Could OnMapReady be made a protected virtual method, so that subclasses can do setup on the NativeMap as well?

kentcb:
@rmarinho no time right now (release due next week). Another problem I'd like to see fixed is `Pin` being `sealed` makes it impossible to extend in order to add your own app-specific data to it (which the renderer could then pick up and apply).

kentcb:
Grrrr, can't do that because it the renderer relies on `internal` methods and properties. :sob:

kentcb:
I've managed to get this working for our scenario. Had to resort to reflection to invoke several `internal` members. As a note to myself, these are the important hooks as I see it:

* make `Pin` extensible
* expose a hook that converts a `Pin` to a `Marker`
* expose a hook for when the map is ready
* allow consumers to forgo `InfoWindowClick` and instead use `MarkerClick`

Also, some bugs I noticed (some may be due to the specifics of our scenario - just want to note for later):

* an initially set location is not moved to? I had to fix this in `OnMapReady`
* initially set pins are not displayed? Again, had to fix in `OnMapReady`

jcmanke:
@kentcb I've created a fork to work on this, feel free to contribute when you can.

https://github.com/jcmanke/Xamarin.Forms/tree/map-renderer-extensibility

PR #829, opened 2017-03-21
Add gesture disposed check to FrameRenderer

hartez:
### Description of Change ###

This applies the same fix from [PR 706](https://github.com/xamarin/Xamarin.Forms/pull/706) to FrameRenderer.

### Bugs Fixed ###

- [45330 – System.ObjectDisposedException: Cannot access a disposed object. Object name: 'Android.Views.GestureDetector'.](https://bugzilla.xamarin.com/show_bug.cgi?id=45330)

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/FrameRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FrameRenderer.cs
@@ -83,6 +83,13 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 					ScaleGestureDetectorCompat.SetQuickScaleEnabled(_scaleDetector.Value, true);
 				handled = _scaleDetector.Value.OnTouchEvent(e);
 			}
+
+			if (_gestureDetector.IsValueCreated && _gestureDetector.Value.Handle == IntPtr.Zero)
+			{
+				// This gesture detector has already been disposed, probably because it's on a cell which is going away
+				return handled;
+			}
+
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.Android/AppCompat/FrameRenderer.cs
```diff
@@ -83,6 +83,13 @@ bool IOnTouchListener.OnTouch(AView v, MotionEvent e)
 					ScaleGestureDetectorCompat.SetQuickScaleEnabled(_scaleDetector.Value, true);
 				handled = _scaleDetector.Value.OnTouchEvent(e);
 			}
+
+			if (_gestureDetector.IsValueCreated && _gestureDetector.Value.Handle == IntPtr.Zero)
+			{
+				// This gesture detector has already been disposed, probably because it's on a cell which is going away
```
mattrichnz: This comment needs to be updated - "it's on a cell which is going away", perhaps change to "it's on a view which is going away".

hartez: The situation being addressed is a gesture detector which has been disposed on the Java side because the list view cell which contains it has been collected - see the description from the [original PR](https://github.com/xamarin/Xamarin.Forms/pull/706).

mattrichnz: OK I see, however I have also seen the gesture detector being disposed when a List View is not being used hence my comment. Particularly noticeable when using Tarjan.

PR #830, opened 2017-03-22
[Core] Bz53561: Grid's element spanning was not rendering correctly.

BradChase2011:
### Description of Change ###

There was an issue with grids where if an element was longer(or taller) than another element then the first element would incorrectly set the opposing height or width of the column/row and skip measuring the second element till a later time causing an incorrect size of the grid compared to the elements inside.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=53561

### API Changes ###

GridCalc had some reclaculation changes.

Changed:
double MeasuredStarredColumns()
double MeasureStarredRows()

### Behavioral Changes ###

Will need to check the other passes for Autos and see if the same thing needs done(most likely will).

### PR Checklist ###


Xamarin.Forms.Xaml.UnitTests/Issues/Bz53561.xaml
```diff
@@ -0,0 +1,18 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
```
StephaneDelcroix: this doesn't need a xaml unites, but a "normal" one in code

BradChase2011: Yea I'll fix up the unit test a bit.  I couldn't get that project to compile.  There seemed to be alot of issues with it, maybe because I used vs2017?  I'll try my best to get it right today, is the project supposed to compile?

StephaneDelcroix: you wouldn't get any release at all if our unit tests projects were failing to build, or tests were failing to pass

BradChase2011:
@StephaneDelcroix Ok I put the changes back up.  Sorry I didnt realize there was a unit test project for the Core project.  I removed the Xaml test as its not needed.  Other than that we should be good to go.  When I get time i am going to test the Auto sizing as well which I believe will be wrong(lets hope not)

BradChase2011:
@StephaneDelcroix  Wait to test this again in the morning.  I found that the Auto calculations were definitely causing issues.  I described it in the bugzilla post. https://bugzilla.xamarin.com/show_bug.cgi?id=53561 .  I will check code in in the morning.

BradChase2011:
Well I just hosed this branch up.  I wanted to rebase and ended up merging, then trying to revert the merge and now it says 533 files changed.

BradChase2011:
@StephaneDelcroix I put the changes up a bit ago, have we made any decisions on what we want to do about this PR?

StephaneDelcroix:
I need to review this again

BradChase2011:
@StephaneDelcroix I would recommend rebuilding the Grid.  We have actually written our own GridCalc at this point and are using that instead.  The Grid still has numerous other issues that are causing bad layouts.  This was mainly an example of something that is not correct.  For us, when we leave an android app and come back every label is set to one character.  Same thing when we expand or contract panels.  After we rewrote it all those problems went away.

I would suggest taking a peak at the measure/arrange code and try to remove the measure call out of the arrange pass.  That will get alot close to the main issues.

opcodewriter:
I also have issues with Grid and other common controls (StackPanel, Image). It's very frustrating after years there are still problems with these controls.

rmarinho:
@BradChase2011 can you rebase? thanks

rmarinho:
Please reopen after addressing feedback and with rebase.

PR #842, opened 2017-03-24
Align iOS, Android, Windows handling of tap gesture event bubbling

hartez:
### Description of Change ###

This change brings iOS, Android, and WinRT/UWP all into alignment on how they treat taps/clicks on a control within a container. 

Previously, if a container (Grid, ContentView, Frame, etc.) had a tap gesture assigned to it and contained controls (Label, Box, Entry, etc.) which do not have gestures assigned, the platforms all behaved differently when those controls were tapped. On each platform, some of the controls would bubble the tap gesture up to the container and some would not, and those lists were different on each platform. 

With this change, controls which become interactive on tap/click (Entry, Editor, Button, etc.) will *not* bubble the tap gesture to the container. Controls which do not (Label, Image, Box, Frame) will allow the gesture to bubble up to their container. The two lists are now consistent across the platforms.

### Bugs Fixed ###

- This change partially addresses the problems described in [45453](https://bugzilla.xamarin.com/show_bug.cgi?id=45453) (in that the behavior of the controls on all of the platforms should now be consistent when the controls are enabled; the behavior inconsistencies Adrian points out in the Bugzilla discussion are still present for controls with `IsEnabled` set to false).

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/GestureBubblingTests.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/GestureBubblingTests.cs
@@ -1,0 +1,190 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+
+#if UITEST
+using NUnit.Framework;
+using Xamarin.UITest.Queries;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	// This is similar to the test for 35477, but tests all of the basic controls to make sure that they all exhibit
+	// the same behavior across all the platforms. The question is whether tapping a control inside of a frame
+	// will trigger the frame's tap gesture; for most controls it will not (the control itself absorbs the tap),
+	// but for non-interactive controls (box, frame, image, label) the gesture bubbles up to the container.
+
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.None, 00100100, "Verify that the tap gesture bubbling behavior is consistent across the platforms", PlatformAffected.All)]
+	public class GestureBubblingTests : TestNavigationPage
+	{
+		const string TargetAutomationId = "controlinsideofframe";
+		ContentPage _menu;
+
+#if UITEST
+		[Test, TestCaseSource(nameof(TestCases))]
+		public void VerifyTapBubbling(string menuItem, bool frameShouldRegisterTap)
+		{
+			var results = RunningApp.WaitForElement(q => q.Marked(menuItem));
+
+			if (results.Length > 1)
+			{
+				var rect = results.First(r => r.Class.Contains("Button")).Rect;
+
+				RunningApp.TapCoordinates(rect.CenterX, rect.CenterY);
+			}
+			else
+			{
+				RunningApp.Tap(q => q.Marked(menuItem));
+			}
+
+			// Find the start label
+			RunningApp.WaitForElement(q => q.Marked("Start"));
+
+			// Find the control we're testing
+			var result = RunningApp.WaitForElement(q => q.Marked(TargetAutomationId));
+			var target = result.First().Rect;
+
+			// Tap the control
+			var y = target.CenterY;
+
+			// In theory we want to tap the center of the control. But Stepper lays out differently than the other controls,
+			// (it doesn't center vertically within its layout), so we need to adjust for it until someone fixes it
+			if (menuItem == "Stepper")
+			{
+				y = target.Y;
+			}
+
+			RunningApp.TapCoordinates(target.CenterX, y);
+
+			if (menuItem == nameof(DatePicker) || menuItem == nameof(TimePicker))
+			{
+				// These controls show a pop-up which we have to cancel/done out of before we can continue
+#if __ANDROID__
+				var cancelButtonText = "Cancel";
+#elif __IOS__
+				var cancelButtonText = "Done";
+#else
+				var cancelButtonText = "X";
+#endif
+				RunningApp.WaitForElement(q => q.Marked(cancelButtonText));
+				RunningApp.Tap(q => q.Marked(cancelButtonText));
+			}
+
+			if (frameShouldRegisterTap)
+			{
+				RunningApp.WaitForElement(q => q.Marked("Frame was tapped"));
+			}
+			else
+			{
+				RunningApp.WaitForElement(q => q.Marked("Start"));
+			}
+		}
+#endif
+
+		ContentPage CreateTestPage(View view)
+		{
+			var instructions = new Label
+			{
+				Text = "Tap the frame below. The label with the text 'No taps yet' should change its text to 'Frame was tapped'."
+			};
+
+			var label = new Label { Text = "Start" };
+
+			var frame = new Frame();
+			frame.Content = new StackLayout { Children = { view } };
+
+			var rec = new TapGestureRecognizer { NumberOfTapsRequired = 1 };
+			rec.Tapped += (s, e) => { label.Text = "Frame was tapped"; };
+			frame.GestureRecognizers.Add(rec);
+			
+			var layout = new StackLayout();
+		
+			layout.Children.Add(instructions);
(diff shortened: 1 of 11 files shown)


kingces95:
Just to be explicit about which controls bubble and which do not, the UITest tests the following matrix:

Those that bubble are 

- Image,
- Label, 
- BoxView, 
- Frame, 

And those that do not are 

- Entry, 
- Editor, 
- Button, 
- SearchBar, 
- DatePicker, 
- TimePicker, 
- Slider, 
- Switch, 
- Stepper.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/GestureBubblingTests.cs
```diff
@@ -0,0 +1,199 @@
...
+			{
+				return (BuildMenu().Content as Layout).InternalChildren.SelectMany(
+					element => (element as Layout).InternalChildren.Select(view => new object[]
+					{
+						(view as Button).Text,
+						_controlsWhichShouldAllowTheTapToBubbleUp.Contains((view as Button).Text)
+					}));
+			}
+		}
+
```
kingces95: LINQ?

```C#
var layout = (Layout)BuildMenu().Content;
var result =
	from Layout element in layout.InternalChildren
	from Button button in element.InternalChildren
	let text = button.Text
	select new object[]
	{
		text,
		_controlsWhichShouldAllowTheTapToBubbleUp.Contains(text)
	};
return result;

**Comment by hartez** [comment]
> Why not fix the cases where the control is disabled instead of just the cases where the control is enabled? Do we not yet agree on the behavior?

Because it's a very difficult problem and I don't want to let perfect be the enemy of good. 

It's a difficult problem because Windows considers controls with `IsEnabled == false` to basically be black holes for the purpose of input, and so far I haven't found a good way around that. iOS seems to have the opposite opinion; anything not enabled just ignores input and lets it bubble up. Android is flexible about this. The correct answer is probably to choose the Windows or iOS behavior and bend the other platforms to that, but I haven't found an easy/elegant way to do that on iOS or Windows. 

Until someone does figure out an answer to that question, I figured getting things in order for enabled controls was better than the total chaos we currently have.

File: Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/GestureBubblingTests.cs [hunk 2]
```diff
@@ -0,0 +1,199 @@
...
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.None, 00100100, "Verify that the tap gesture bubbling behavior is consistent across the platforms", PlatformAffected.All)]
+	public class GestureBubblingTests : TestNavigationPage
+	{
+		const string TargetAutomationId = "controlinsideofframe";
+		ContentPage _menu;
+
+#if UITEST
+		[Test, TestCaseSource(nameof(TestCases))]
+		public void VerifyTapBubbling(string menuItem, bool frameShouldRegisterTap)
```
kingces95: Could we expand the test case to include `bool isDisabled` to test the behavior of disabled controls? Also, might we want to position a sibling `BoxView` just under the control being tested and test InputTransparent? Seems like a good idea to have a single test that covers all these cases, no?

hartez: We have a test like this for `InputTransparent`: https://github.com/xamarin/Xamarin.Forms/pull/808

hartez: And any test for disabled controls wouldn't pass on all the platforms at this point, because they are not consistent for disabled controls. I'd rather leave that test for whoever fixes that bug.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/GestureBubblingTests.cs
```diff
@@ -0,0 +1,199 @@
...
+			get
+			{
+				return (BuildMenu().Content as Layout).InternalChildren.SelectMany(
+					element => (element as Layout).InternalChildren.Select(view => new object[]
+					{
+						(view as Button).Text,
+						_controlsWhichShouldAllowTheTapToBubbleUp.Contains((view as Button).Text)
+					}));
+			}
+		}
```
hartez: Updated to query syntax.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/GestureBubblingTests.cs
```diff
@@ -0,0 +1,199 @@
...
+			var testView = view();
+			testView.AutomationId = TargetAutomationId;
+
+			button.Clicked += (sender, args) => PushAsync(CreateTestPage(testView));
+
+			return button;
+		}
+
+		// These controls should allow the tap gesture to bubble up to their container; everything else should absorb the gesture
+		readonly List<string> _controlsWhichShouldAllowTheTapToBubbleUp = new List<string> { nameof(Image), nameof(Label), nameof(BoxView), nameof(Frame) };
```
kingces95: Eh, git hub is not using the full width. Is the a switch somewhere?

hartez: I reformatted it to be a more friendly width. AFAIK there's no way to get GH to use the full screen width unless you use a custom stylesheet or something.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/GestureBubblingTests.cs
kingces95: And it could break people... So, just to sure, when InputTransparent is set the events, assuming they are not swallowed/handled, are routed to what's behind the control as opposed to its parent?

hartez: When `InputTransparent` is true, the events are not swallowed or handled by the control at all; they're routed to whatever is behind/below the control and if they're still not handled, they bubble up to the container/parent.

kingces95: Ah. Even when the control is disabled? You mentioned windows controls swallow events when disabled. That's not the case if the control is InputTransparent?

adrianknight89:
@hartez 

> Controls which do not (Label, Image, Box, Frame) will allow the gesture to bubble up to their container.

I just tested 2.3.5-pre2. While the above statement is true on iOS, it isn't on Android. I have the following setup:

```
<StackLayout Orientation="Vertical" Spacing="10" VerticalOptions="Center">
        <Grid WidthRequest="250" HeightRequest="50" HorizontalOptions="Center" VerticalOptions="Center" BackgroundColor="Red">
            <Grid.GestureRecognizers>
                <TapGestureRecognizer Tapped="TapGestureRecognizer_Tapped"></TapGestureRecognizer>
            </Grid.GestureRecognizers>
            <Label x:Name="Label1" Text="InputTransparent=True" InputTransparent="True" />
        </Grid>

        <Grid WidthRequest="250" HeightRequest="50" HorizontalOptions="Center" VerticalOptions="Center" BackgroundColor="Yellow">
            <Grid.GestureRecognizers>
                <TapGestureRecognizer Tapped="TapGestureRecognizer_Tapped2"></TapGestureRecognizer>
            </Grid.GestureRecognizers>
            <Label x:Name="Label2" Text="InputTransparent=False"/>
        </Grid>
</StackLayout>
```

The only way for `Label` to bubble up a tap event is to set `InputTransparent` to true. I believe the intention was to let labels bubble up regardless of `InputTransparent` values.

adrianknight89:
I believe #902 might have fixed the issue, so I'll wait for -pre3.

PR #844, opened 2017-03-29
Map renderer extensibility

jcmanke:
### Description of Change ###

Makes the Xamarin.Forms.Maps projects more extensible, by unsealing the Pin class and adding hooks in the MapRenderer classes for customization of map and pins.

### API Changes ###

Added:
- protected virtual void MapRenderer.OnMapReady(GoogleMap map) (Android)
- protected virtual MarkerOptions MapRenderer.CreateMarker(Pin pin) (Android)
- protected virtual IMKAnnotation MapRenderer.CreateAnnotation(Pin pin) (iOS)
- Pin.LabelProperty

Changed:

- IOnMapReadyCallback.OnMapReady(GoogleMap map) calls protected OnMapReady for map initialization 
- public sealed class Pin => public class Pin
- Pin.Label now uses backing BindableProperty LabelProperty

### Behavioral Changes ###

No behavioral changes to existing functionality, reduces need for workarounds when subclassing MapRenderer.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Maps.Android/MapRenderer.cs
+++ b/Xamarin.Forms.Maps.Android/MapRenderer.cs
@@ -181,6 +181,33 @@ namespace Xamarin.Forms.Maps.Android
 				MoveToRegion(Element.LastMoveToRegion, false);
 			}
 		}
+		
+		protected virtual void OnMapReady(GoogleMap map)
+		{
+			if (map == null)
+			{
+				return;
+			}
+			
+			map.SetOnCameraChangeListener(this);
+			map.InfoWindowClick += MapOnMarkerClick;
+			
+			map.UiSettings.ZoomControlsEnabled = Map.HasZoomEnabled;
+			map.UiSettings.ZoomGesturesEnabled = Map.HasZoomEnabled;
+			map.UiSettings.ScrollGesturesEnabled = Map.HasScrollEnabled;
+			map.MyLocationEnabled = map.UiSettings.MyLocationButtonEnabled = Map.IsShowingUser;
+			SetMapType();
+		}
+		
+		protected virtual MarkerOptions CreateMarker(Pin pin)
+		{
+			var opts = new MarkerOptions();
+			opts.SetPosition(new LatLng(pin.Position.Latitude, pin.Position.Longitude));
+			opts.SetTitle(pin.Label);
+			opts.SetSnippet(pin.Address);
+			
+			return opts;
+		}
 
 		void AddPins(IList pins)
 		{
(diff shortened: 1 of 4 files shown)


pauldipietro:
I think this should go through the [Xamarin.Forms Evolution forum](https://forums.xamarin.com/categories/xamarin-forms-evolution) first.

jcmanke:
Apologies if I'm doing things out-of-order. [Idea posted in Evolution forum.](https://forums.xamarin.com/discussion/91907/map-renderer-extensibility/)

Xamarin.Forms.Maps.Android/MapRenderer.cs
```diff
@@ -51,6 +51,24 @@ public override SizeRequest GetDesiredSize(int widthConstraint, int heightConstr
 		{
 			return new SizeRequest(new Size(Context.ToPixels(40), Context.ToPixels(40)));
 		}
+		
+		public virtual void OnMapReady(GoogleMap map)
```
jassmith: Can you please convert this to a public non-virtual method which them calls a protected virtual method?

There are a series of issues with public virtual methods, mostly centering around the fact that you cant sanitize inputs/outputs if issues crop up in the future. By moving to a 2 method solution with a protected virtual we retain the required flexibility.

jassmith: In fact now that I think of it, we should leave the explicity implementation of this method, and then have that call the protected virtual method. This reduces the external API surface area while retaining the desired effect.

Xamarin.Forms.Maps/Pin.cs
```diff
@@ -2,17 +2,15 @@
 
 namespace Xamarin.Forms.Maps
 {
-	public sealed class Pin : BindableObject
+	public class Pin : BindableObject
```
jassmith: For the moment I think I'm okay with unsealing this class. We will need to evaluate the other renderers to make sure there isn't a reason this could go wrong.

Xamarin.Forms.Maps.Android/MapRenderer.cs
jcmanke: Would you prefer to have the protected virtual method take the GoogleMap parameter, or expect overrides to use the NativeMap property?

jassmith: I much prefer needed data passed to methods rather than fetched from properties. The more idempotent the function, the better.

vrassouli:
I'm trying to extend the Map and MapRenderer to support custom tile providers.
But when I inherit from MapRenderer, some features don't work. Like MoveToRegion method, or IsShowingUser property.

PR #846, opened 2017-03-29
Don't auto reenable the cancel button

Depechie:
### Description of Change ###

iOS SearchBar renderer will now no longer auto reenable the cancel button when text is being cleared.

### Bugs Fixed ###

When an user has a custom renderer with following code : Control.ShowsCancelButton = false;
Inside the OnElementPropertyChanged() method, the current implementation of the control on iOS would reverse this setting each time the search bar would be cleared from input text.

For reference blogpost about the problem and current work around [http://depblog.weblogs.us/2016/10/27/xamarin-forms-ios-disable-cancel-button-on-searchbar/](http://depblog.weblogs.us/2016/10/27/xamarin-forms-ios-disable-cancel-button-on-searchbar/), this commit will fix that problem.

### API Changes ###

Removed:
 - Control.ShowsCancelButton = !string.IsNullOrEmpty(Control.Text);

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/SearchBarRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/SearchBarRenderer.cs
@@ -158,8 +158,6 @@ namespace Xamarin.Forms.Platform.iOS
 
 		void UpdateCancelButton()
 		{
-			Control.ShowsCancelButton = !string.IsNullOrEmpty(Control.Text);
-
 			// We can't cache the cancel button reference because iOS drops it when it's not displayed
 			// and creates a brand new one when necessary, so we have to look for it each time
 			var cancelButton = Control.FindDescendantView<UIButton>();


rmarinho:
Hi, this is not the best way to fix it... Can you create a virtual method that your renderer can override and change it.. this is way we don't break other people.

Depechie:
@rmarinho ow... to be clear 'bugs' in the frameworks are always left as is then? Even if you guys fix them? Because current behaviour ( even if people now depend on it ) is wrong!
And to be honest I don't think we are breaking anything because only way people will notice this is they created a custom renderer to actually dismiss the button and that would not work... so my guess is that nobody would be bothered or implemented their on work around.

If all of this makes no sense I will try the virtual method ;)

rmarinho:
Well yeah we can reliable say if someone is already making a fix for this, and we don't won't them to break/change there code even if we are fixing the bug/issue, thats just how the way things go. 

Yeah the virtual method is the way to go, keeping the old behaviour, and allowing to be override. 

Thanks for understanding

rmarinho:
Hi @Depechie please reopen when you address this using the virtual method

PR #851, opened 2017-03-31
[Xaml[C]] supports 'using:' xmlns declarations

StephaneDelcroix:
### Description of Change ###

[Xaml[C]] supports 'using:' xmlns declarations

```
	xmlns:local="using:Xamarin.Forms.Xaml.UnitTests"
```

This behaves exactly like `clr-namespace` without any other parameters

### Bugs Fixed ###

/

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###


Xamarin.Forms.Xaml/XmlnsHelper.cs
```diff
@@ -20,6 +20,19 @@ public static void ParseXmlns(string xmlns, out string typeName, out string ns,
 		{
 			typeName = ns = asm = targetPlatform = null;
 
+			xmlns = xmlns.Trim();
+
+			if (xmlns.StartsWith("using:", StringComparison.Ordinal)) {
+				ParseUsing(xmlns, out typeName, out ns, out asm, out targetPlatform);
+				return;
```
samhouts: Just curious...why not do `else`?

Xamarin.Forms.Xaml/XmlnsHelper.cs
```diff
@@ -46,5 +59,17 @@ public static void ParseXmlns(string xmlns, out string typeName, out string ns,
...
 		}
+
+		static void ParseUsing(string xmlns, out string typeName, out string ns, out string asm, out string targetPlatform)
+		{
+			typeName = ns = asm = targetPlatform = null;
+
+			foreach (var decl in xmlns.Split(';')) {
+				if (decl.StartsWith("using:", StringComparison.Ordinal)) {
+					ns = decl.Substring(6, decl.Length - 6);
+					continue;
```
samhouts: This doesn't seem useful.

Xamarin.Forms.Xaml/XmlnsHelper.cs
```diff
@@ -46,5 +59,17 @@ public static void ParseXmlns(string xmlns, out string typeName, out string ns,
 					typeName = decl;
 			}
 		}
+
+		static void ParseUsing(string xmlns, out string typeName, out string ns, out string asm, out string targetPlatform)
+		{
+			typeName = ns = asm = targetPlatform = null;
+
+			foreach (var decl in xmlns.Split(';')) {
```
samhouts: So if the user does something silly, like include more than one `using`, then the last one wins. Is this something we need to have a test for to establish the behavior as expected?

Xamarin.Forms.Xaml/XmlnsHelper.cs
StephaneDelcroix: avoid unnecessary nesting ? no real reason for one option or the other. If you think I did that to optimize the compiled code, well, you're thinking too much :)

Xamarin.Forms.Xaml/XmlnsHelper.cs
StephaneDelcroix: silly user gets silly results.

I left the foreach, like in cld-namespace, because at some point in the future someone will surely asks for supporting targetPlatform

Xamarin.Forms.Xaml/XmlnsHelper.cs
StephaneDelcroix: for the same reason stated before. to keep in sync, and be future-ish proof

samhouts: fair enough. the `continue` doesn't hurt anything, just seemed superfluous at this time. :)

StephaneDelcroix:
@jassmith do you think this should support (as of NOW with the goal we have) `targetPlatform` ?

PR #852, opened 2017-04-01
Make IAppLinks.RegisterLink etc awaitable to allow catching exceptions

dominik-weber:
### Description of Change ###

With the current implementation the app crashes when the thumbnail url can't be loaded on iOS. This change makes the RegisterLink etc. methods awaitable (return Task) so the dev can catch exceptions and avoid unhandled exception crashes.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=42061

### API Changes ###

Changed:
 - void IAppLinks.RegisterLink => Task IAppLinks.RegisterLink
 - void IAppLinks.DeregisterLink => Task IAppLinks.DeregisterLink

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android.AppLinks/AndroidAppLinks.cs
+++ b/Xamarin.Forms.Platform.Android.AppLinks/AndroidAppLinks.cs
@@ -35,19 +35,21 @@ namespace Xamarin.Forms.Platform.Android.AppLinks
 			_client.Connect();
 		}
 
-		public void DeregisterLink(IAppLinkEntry appLink)
+		public Task DeregisterLink(IAppLinkEntry appLink)
 		{
 			RemoveFromIndexItemAsync(appLink.AppLinkUri.ToString());
+			return Task.FromResult<object>(null);
 		}
 
-		public void DeregisterLink(Uri appLinkUri)
+		public Task DeregisterLink(Uri appLinkUri)
 		{
 			RemoveFromIndexItemAsync(appLinkUri.ToString());
+			return Task.FromResult<object>(null);
 		}
 
-		public async void RegisterLink(IAppLinkEntry appLink)
+		public Task RegisterLink(IAppLinkEntry appLink)
 		{
-			await IndexItemAsync(appLink);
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.Android.AppLinks/AndroidAppLinks.cs
```diff
@@ -35,19 +35,21 @@ public AndroidAppLinks(Context context)
 			_client.Connect();
 		}
 
-		public void DeregisterLink(IAppLinkEntry appLink)
+		public Task DeregisterLink(IAppLinkEntry appLink)
 		{
 			RemoveFromIndexItemAsync(appLink.AppLinkUri.ToString());
+			return Task.FromResult<object>(null);
```
jerone: `null` ?

dominik-weber: It doesn't matter what result we use, we just need a completed Task to convert this synchronous method to an async one. The result is actually converted from `Task<whatever>` to `Task` so the result is ignored anyway.
The same task/result is used here: https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/NavigationProxy.cs#L201

dominik-weber: If there is an existing guideline or practice how this should be handled in the Forms repo please let me know and I'll change it.

rmarinho:
Hum, this is a breaking change.. .. not sure we can accept this.

dominik-weber:
@rmarinho I can't see where this should break any apps - especially since the currrent implementation just crashes when the image can't be loaded. Could you please explain?

rmarinho:
@dominik-weber if anyone is implementing the IAppLinks interface they will have to change the code.

Can we try fix the image loading only ?

rmarinho:
Closing for now. I will revise this and try to fix the loading image issue

PR #859, opened 2017-04-06
OnResume assumed that _application had been set

appbureauet:
When OnResume is called it should not expect the _application to be set already.
I have an app where I use one activity for showing a splash screen and then it starts an intent to show the main activity containing the application.
This used to work before 2.3.4.224.

### Description of Change ###

I have added a null check for the `_application` object before it's used.

### Bugs Fixed ###

OnResume crashes if the `_application` has not been set.

### API Changes ###

None

### Behavioral Changes ###

### PR Checklist ###


rmarinho:
Failing test is not from this pr

PR #862, opened 2017-04-07
Do not emit implict_op if from/to are same type

kingces95:
### Description of Change ###

After bisection I found that change `eacbc1acf4cc` in `GetImplicitOperatorTo` causes the reproduction of https://bugzilla.xamarin.com/show_bug.cgi?id=54012 to throw an AccessViolationException in https://github.com/xamarin/Xamarin.Forms/blob/c65a9a8c57d93b0eff38fd8b06223f6a04688c59/Xamarin.Forms.Platform.WinRT/LabelRenderer.cs#L20. Note the bug reports a different issue; this is a regression  that blocks investigation of the reported issue.

Expected `MasterDetailNavigation.dll` for 54012 to successfully pass peverify but actually observed errors. Expected xaml compiler to emit initlocals for all method bodies but actually did not. Expected xaml compiler to not emit op_implicit if `fromType` equals `toType` but actually emits op_implicit when that is the case. Either actual behavior produces unverifiable code. 

Changed xaml compiler to emit initlocals for all MethodDefinitions. Changed xaml compiler to not emit op_implicit when from type equals to type. 

### Bugs Fixed ###

None. Discovered regression while investigating another bug. 

### API Changes ###

None.

### Behavioral Changes ###

None.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/DebugXamlCTask.cs
+++ b/Xamarin.Forms.Build.Tasks/DebugXamlCTask.cs
@@ -73,6 +73,7 @@ namespace Xamarin.Forms.Build.Tasks
 						if (initCompRuntime == null) {
 							Logger.LogString(2, "   Creating empty {0}.__InitComponentRuntime ...", typeDef.Name);
 							initCompRuntime = new MethodDefinition("__InitComponentRuntime", initComp.Attributes, initComp.ReturnType);
+							initCompRuntime.Body.InitLocals = true;
 							Logger.LogLine(2, "done.");
 							Logger.LogString(2, "   Copying body of InitializeComponent to __InitComponentRuntime ...", typeDef.Name);
 							initCompRuntime.Body = new MethodBody(initCompRuntime);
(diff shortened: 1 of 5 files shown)


kingces95:
Here is the bad IL: The op_Implicit call takes a string but is passed V_10 which is a `FormattedString`. 
    
    IL_03bc:  ldsfld     class [Xamarin.Forms.Core]Xamarin.Forms.BindableProperty [Xamarin.Forms.Core]Xamarin.Forms.Label::FormattedTextProperty
    IL_03c1:  ldloc.s    V_10
    IL_03c3:  call       class [Xamarin.Forms.Core]Xamarin.Forms.FormattedString [Xamarin.Forms.Core]Xamarin.Forms.FormattedString::op_Implicit(string)
    IL_03c8:  callvirt   instance void [Xamarin.Forms.Core]Xamarin.Forms.BindableObject::SetValue(class [Xamarin.Forms.Core]Xamarin.Forms.BindableProperty,

PR #876, opened 2017-04-19
[XamlC] Check param type in op_Implicit

StephaneDelcroix:
### Description of Change ###

Who could have guessed that picking the first implicit operator with the right return type without checking the input parameter would not work all the time ?

What were the chances that it would fail ?

1 out of a trillion. No more. No less.

But still that particular unlikely case happened. That's really bad luck.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=55347

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/TypeReferenceExtensions.cs
+++ b/Xamarin.Forms.Build.Tasks/TypeReferenceExtensions.cs
@@ -249,9 +249,16 @@ namespace Xamarin.Forms.Build.Tasks
 			if (TypeRefComparer.Default.Equals(fromType, toType))
 				return null;
 
-			var implicitOperatorsOnFromType = fromType.GetMethods(md => md.IsPublic && md.IsStatic && md.IsSpecialName && md.Name == "op_Implicit", module);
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Build.Tasks/TypeReferenceExtensions.cs
```diff
@@ -249,9 +249,16 @@ public static MethodReference GetImplicitOperatorTo(this TypeReference fromType,
 			if (TypeRefComparer.Default.Equals(fromType, toType))
 				return null;
 
-			var implicitOperatorsOnFromType = fromType.GetMethods(md => md.IsPublic && md.IsStatic && md.IsSpecialName && md.Name == "op_Implicit", module);
-			var implicitOperatorsOnToType = toType.GetMethods(md => md.IsPublic && md.IsStatic && md.IsSpecialName && md.Name == "op_Implicit", module);
+			var implicitOperatorsOnFromType = fromType.GetMethods(md =>    md.IsPublic
```
samhouts: This looks insane on GitHub, but whatever. :)

PR #877, opened 2017-04-19
Calling API not available in all versions

kingces95:
### Description of Change ###

Calling an API not available in all versions of Android.

### Bugs Fixed ###

https://forums.xamarin.com/discussion/comment/267383/#Comment_267383
https://bugzilla.xamarin.com/show_bug.cgi?id=55230

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/FastRenderers/ButtonRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/FastRenderers/ButtonRenderer.cs
@@ -346,7 +346,8 @@ namespace Xamarin.Forms.Platform.Android.FastRenderers
 		void Initialize()
 		{
 			// Fixes issues with AppCompatButton elevation (z-index) 
-			StateListAnimator = null;
+			if (Forms.IsLollipopOrNewer)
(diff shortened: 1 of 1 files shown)


kingces95:
Looks like @StephaneDelcroix removed the call entirely. We should remove the comment too if the call's not needed. @rmarinho checked in the file. Do we not need that call? The comment makes it seem like it's a fix for a regression..

kingces95:
@rmarinho, you noted in the channel that because this is a fast renderer that it cannot be a regression. From that I assume it does not need to be backported.

rmarinho:
Sorry may i expressed myself wrong, it has to be backported to 2.3.5 , but **is not a regression** because it's a new feature, it's just a bugfix on a new feature that needs to be cherrypicked to 2.3.5 branch for 2.3.5-pre2

kingces95:
Cherry-picked to 2.3.5; https://github.com/xamarin/Xamarin.Forms/commit/2dba30562cb4db7114207b7095d5660a4ca616a7

PR #885, opened 2017-04-26
Bugzilla55607 - [iOS] Opening a ListView Cell's context action will disable context actions for all other ListViews

bruzkovsky:
### Description of Change ###

Added check for instance of UITableView, so that only one ContextAction can be open **per ListView**.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=55588
https://bugzilla.xamarin.com/show_bug.cgi?id=55607 (duplicate)

### API Changes ###

None.

### Behavioral Changes ###

None.

### PR Checklist ###


Ruddy2007:
@bruzkovsky I have ran into a problem with this commit. You can get a null exception on this line. Im just going to guess that in `(ContextScrollViewDelegate)scrollViewBeingScrolled.Delegate` - `scrollViewBeingScrolled` can be null?

PR #552, opened 2016-11-20
[Android] Ability to skip adding status bar underlay

adrianknight89:
### Description of Change ###

AppCompat adds status bar underlay when the top-level layout is being constructed, and there is no way to skip this process. When `AdjustResize` is used, we end up with a status bar that has no background color AND the underlay below it.

Also changed variable and method names for retrieving status bar height to status bar underlay height to avoid confusion when previously-named `GetStatusBarHeight()` starts returning zero. Moved API >= 21 checks into the method to simplify callers.

Not sure how to add this test code since it's changing `MainPage`: https://gist.github.com/adrianknight89/e35d769961fc367edad9d19ca73b0679

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=47548

### API Changes ###

Added:
 - protected bool ShouldAddStatusBarUnderlay = true;

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #892, opened 2017-05-02
Remove status bar underlay view in Android AppCompat

hartez:
### Description of Change ###

Previously, Android status bar background color was set by creating a view to be laid out behind the status bar and setting its background color. This required an extra view, a query for the height of the status bar, several adjustments to layouts to account for the size of the view, and a search through application resources to find the "colorPrimaryDark" value, all of which affect application startup (on older devices, this took as much as 100 ms). Additionally, this caused an issue when setting the soft input adjustment mode to "resize"; a gap would appear between the status bar and the application.

This change removes the status bar underlay. Status bar background color is still determined by the "colorPrimaryDark" value (the default for AppCompat). Changing the status bar color at runtime is handled by the `Window.SetStatusBarColor()` method. 

By removing the status bar underlay and removing the status bar height adjustments, the gap from [issue 47548](https://bugzilla.xamarin.com/show_bug.cgi?id=47548) (see the screenshots in [PR 552](https://github.com/xamarin/Xamarin.Forms/pull/552) ) goes away.

With this change, making an application full screen no longer [requires reflection to remove the status bar](https://forums.xamarin.com/discussion/comment/237310/#Comment_237310).

### Bugs Fixed ###

- [47548 – There is no way to skip adding status bar underlay on AppCompat](https://bugzilla.xamarin.com/show_bug.cgi?id=47548)

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
@@ -45,7 +45,6 @@ namespace Xamarin.Forms.Platform.Android
 		AndroidApplicationLifecycleState _previousState;
 
 		bool _renderersAdded, _isFullScreen;
-		int _statusBarHeight = -1;
 
 		// Override this if you want to handle the default Android behavior of restoring fragments on an application restart
 		protected virtual bool AllowFragmentRestore => false;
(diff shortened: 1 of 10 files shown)


Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -184,7 +185,11 @@ protected override void OnCreate(Bundle savedInstanceState)
 
 			OnStateChanged();
 
-			AddStatusBarUnderlay();
+			if (Forms.IsLollipopOrNewer)
+			{
+				// Allow for the status bar color to be changed
+				Window.AddFlags(WindowManagerFlags.DrawsSystemBarBackgrounds);
```
hartez: What's wrong with the formatting?

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -105,7 +103,10 @@ public override bool OnOptionsItemSelected(IMenuItem item)
 
 		public void SetStatusBarColor(AColor color)
 		{
-			_statusBarUnderlay.SetBackgroundColor(color);
+			if (Forms.IsLollipopOrNewer)
```
kingces95: If older than no StatusBarColor for you, huh?

PR #893, opened 2017-05-02
[iOS Maps] Pin customization in iOS MapRenderer

jcmanke:
### Description of Change ###

- Continuation of #844 to create customization feature parity between Android and iOS.

[Evolution Forum Post](https://forums.xamarin.com/discussion/91907/map-renderer-extensibility/)

### API Changes ###

Changed:
- public class MapDelegate : ViewRenderer => MapDelegate : ViewRenderer<Map, MKMapView>
- private void AttachGestureToPin(MKPinAnnotationView mapPin, IMKAnnotation annotation) => protected void AttachGestureToPin(MKAnnotationView mapPin, IMKAnnotation annotation) // So the Pin.Clicked event can be used when GetViewForAnnotation is overridden and/or native pin is rendered as a different subclass MKAnnotationView

Removed:
 - MapDelegate class // Methods moved into MapRenderer class

### Behavioral Changes ###

Should not change any existing functionality in the base library, makes it easier to customize pins when extending the iOS MapRenderer.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Maps.iOS/MapRenderer.cs
+++ b/Xamarin.Forms.Maps.iOS/MapRenderer.cs
@@ -20,22 +20,29 @@ using Xamarin.Forms.Platform.MacOS;
 namespace Xamarin.Forms.Maps.MacOS
 #endif
 {
-	internal class MapDelegate : MKMapViewDelegate
+	public class MapDelegate : MKMapViewDelegate
 	{
+		protected Map _map;
+
 		// keep references alive, removing this will cause a segfault
 		readonly List<object> List = new List<object>();
-		Map _map;
+
 		object _lastTouchedView;
 		bool _disposed;
 
-		internal MapDelegate(Map map)
+		public MapDelegate(Map map)
 		{
 			_map = map;
 		}
 
+		public override void RegionChanged(MKMapView mapView, bool animated)
+		{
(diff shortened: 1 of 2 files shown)


jcmanke:
I'm still personally debating if the delegate methods should all be contained in the MapDelegate class or be methods on the MapRenderer class, but the half-and-half methodology currently in use just seems wrong.

Xamarin.Forms.Maps.iOS/MapRenderer.cs
```diff
@@ -20,145 +20,11 @@ namespace Xamarin.Forms.Maps.iOS
...
-				_map = null;
-				_lastTouchedView = null;
-			}
-
-			base.Dispose(disposing);
-		}
-	}
-
-	public class MapRenderer : ViewRenderer
+	public class MapRenderer : ViewRenderer<Map, MKMapView>
```
StephaneDelcroix: We can't change the base class definition, it will break every renderer inheriting from it

StephaneDelcroix:
This is a breaking change. Closing this PR. Please propose an alternate fix that doesn't change the base class and reopen the PR (or create a new one)

jcmanke:
@StephaneDelcroix please see new PR #1065

PR #898, opened 2017-05-03
Fix bugzilla55542

rmarinho:
### Description of Change ###

This issue comes from January related with #674 , what happens is View was being cached with a null renderer.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=55542

### Behavioral Changes ###

We don't cache null renderers 

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Registrar.cs
+++ b/Xamarin.Forms.Core/Registrar.cs
@@ -21,6 +21,9 @@ namespace Xamarin.Forms.Internals
 
 		public void Register(Type tview, Type trender)
 		{
+			//avoid caching null renderers
+			if (trender == null)
+				return;
 			_handlers[tview] = trender;
 		}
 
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Core/Registrar.cs
```diff
@@ -21,6 +21,9 @@ public class Registrar<TRegistrable> where TRegistrable : class
 
 		public void Register(Type tview, Type trender)
 		{
+			//avoid caching null renderers
```
hartez: We _do_ want to cache null renderer results if they're valid, so we don't have to do multiple `GetCustomAttribute` calls (see the comments in `GetHandlerType`.

It looks like what you really want is to avoid overwriting an existing handler with `null`; something like this:
```
if (trender == null && _handlers.ContainsKey(tview)) 
    return;
```

So if a handler has already been registered, don't overwrite it with 'null'; if there's no handler, we can cache the `null` result.

rmarinho: Yap crossed my mind that.. but decided null wasn't important .. will take your suggestion.

rmarinho: After further revision and discussion we found that the problem is with the lookup system that needs more work, and we will go with just reverting caching null for now

rmarinho: @hartez can you aprove?

PR #899, opened 2017-05-03
[Android] Add null check before creating and drawing canvas bugzilla 55559 & 45602

pineapple216:
[Android] Add null check before creating and drawing canvas bugzilla 55559 & 45602

### Description of Change ###

Bugs 

### Bugs Fixed ###
[Android] Exception when setting the backgroundcolor of a frame
https://bugzilla.xamarin.com/show_bug.cgi?id=55559

NullPointerException when applying style
https://bugzilla.xamarin.com/show_bug.cgi?id=45602

### API Changes ###
None.

Added:
a null check on the _bitmap object before creating a canvas with it.

### Behavioral Changes ###

None

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description): No, not necessary.

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/FrameRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/FrameRenderer.cs
@@ -188,14 +188,17 @@ namespace Xamarin.Forms.Platform.Android
 					e.PropertyName == Frame.OutlineColorProperty.PropertyName ||
 					e.PropertyName == Frame.CornerRadiusProperty.PropertyName)
 				{
-					using (var canvas = new ACanvas(_normalBitmap))
+					if (_normalBitmap != null)
 					{
-						int width = Bounds.Width();
-						int height = Bounds.Height();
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.Android/Renderers/FrameRenderer.cs
```diff
@@ -188,14 +188,17 @@ void FrameOnPropertyChanged(object sender, PropertyChangedEventArgs e)
 					e.PropertyName == Frame.OutlineColorProperty.PropertyName ||
 					e.PropertyName == Frame.CornerRadiusProperty.PropertyName)
 				{
-					using (var canvas = new ACanvas(_normalBitmap))
+					if (_normalBitmap != null)
```
StephaneDelcroix: this could be instead
```
if (normalBitmap == null)
    return;
```

pineapple216: I've to admit, it's more readable.

PR #903, opened 2017-05-04
Enable building FormsViewGroup.jar from the command line

hartez:
### Description of Change ###

This adds a Gradle project to build the FormsViewGroup.jar from the command line; no need to install Android Studio or Eclipse. It works on Windows and macOS. To build a new `.jar` just navigate to the `AndroidNative` folder and run 

- Windows PowerShell: `.\gradlew createJar --rerun-tasks`
- macOS Terminal: `./gradlew createJar --rerun-tasks`

The resulting `formsviewgroup.jar` will be put in `Xamarin.Forms.Platform.Android.FormsViewGroup\Jars`.

Before the first time you run it, you'll need to create a `local.properties` file in the `AndroidNative` folder so Gradle can find your Android SDK. It needs one line to set the `sdk.dir` property. For example:

- Windows: `sdk.dir=C\:\\Users\\cfinley\\AppData\\Local\\Android\\Sdk`
- macOS: `sdk.dir=/Users/cfinley/Library/Developer/Xamarin/android-sdk-macosx`

It may download a bunch of stuff the first time it runs; after that it'll be much faster.
	
If you don't already have Gradle installed:

- Windows: `choco install gradle`
- macOS: `brew install gradle`


hartez:
Marking this "do-not-merge" until I've gotten feedback from more people on whether it works on their configurations.

.gitignore
```diff
@@ -46,4 +46,10 @@ Xamarin.Forms.Controls/secrets.txt
...
 *.tss
-*.bak
\ No newline at end of file
+*.bak
+AndroidNative/local.properties
+AndroidNative/FormsViewGroup/build/
+AndroidNative/build/
+AndroidNative/.gradle/
+AndroidNative/gradlew
+AndroidNative/gradlew.bat
```
StephaneDelcroix: those 2 last ones are part of this PR. shouldn't be ignored

StephaneDelcroix:
After installing and uninstalling almost all possible item in the SDK manager to get rid of the exception stacktraces in the (failing) build output, it turned out those weren't failure.

to get the script working, I had to `chmod +x path/to/aapt`

StephaneDelcroix:
Added the instruction from this description in the sources. and merging without further review as it doesn't touch any sensible part of the project

PR #889, opened 2017-04-28
[Evolution] Toolbar a11y

samhouts:
### Description of Change ###

Building on #713, this would add VoiceOver/TalkBack/Narrator capability to ToolBarItems. Included in this change would be limited support for setting accessibility properties on the MasterDetailPage toggle button and the NavigationBar back button.

### `ToolbarItem`s
- [All] `ToolbarItem`s with `Text` will already read the `Text` if no `Accessibility.Name` or `.Hint` are supplied. No change.
- [iOS/UWP] `Accessibility.Name` will _replace_ the `Text` that is read aloud on iOS and UWP. This is consistent with the platforms.
- [Android] `Accessibility.Name` and/or `Accessibility.Hint` will _completely replace_ the `Text` that is both **visible** and read aloud on Android. This is a limitation of API<26 and may be resolved later to be consistent with iOS and UWP.

### `MasterDetailPage` Toggle Button
- [iOS/UWP] Set `Accessibility.Name` and `Accessibility.Hint` on a `MasterDetailPage` to set the text that is read aloud on the toggle button. Setting these properties will have no effect on Android.
- [Android] Add string resources to the Android project to set the text that is read aloud on the `MasterDetailPage` toggle button. 
```
 <string name="mdp_open_a11y_name">Show Master Page</string>
 <string name="mdp_close_a11y_name">Hide Master Page</string>
```

### `NavigationPage` Back Button
- [Android] Set `Accessibility.Name` and `Accessibility.Hint` on a `Page` to set the text that is read aloud on the Back arrow in the Action Bar. This will have no effect on OS back buttons.
- [UWP] Not applicable. 
- [iOS] Not supported; will read the title of the page as usual.

See https://forums.xamarin.com/discussion/94289/basic-accessibility-support-toolbaritems.

### Bugs Fixed ###

- [Bug 52656 - ToolbarItem needs accessibility support](https://bugzilla.xamarin.com/show_bug.cgi?id=52656)

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #904, opened 2017-05-05
[Android] Remove the ". " on empty labels (Accessibility) on Fastrenderers

bentmar:
### Description of Change ###

Removes the ". " if a label has no text. Labels are today given a hint from the SetHint() method. The hint will allways include ". "  wich leads to empty labels being rendered with a dot instead of being empty. Also hints set by a custom renderer are ignored because of this.

This image shows what empty labels are rendered as  http://imgur.com/Min569d

### Bugs Fixed ###

The dot comes from a string.join with the accessibility values, but if those values are null we now return from that method before setting a hint.
result   http://imgur.com/PxoKvwi

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/FastRenderers/AccessibilityProvider.cs
+++ b/Xamarin.Forms.Platform.Android/FastRenderers/AccessibilityProvider.cs
@@ -8,8 +8,7 @@ namespace Xamarin.Forms.Platform.Android.FastRenderers
 	{
 		const string GetFromElement = "GetValueFromElement";
 		string _defaultContentDescription;
-		bool? _defaultFocusable;
-		string _defaultHint;
+		bool? _defaultFocusable;		
 		bool _disposed;
 
 		IVisualElementRenderer _renderer;
@@ -117,9 +116,9 @@ namespace Xamarin.Forms.Platform.Android.FastRenderers
 			Control.Focusable =
 				(bool)(value ?? (bool?)Element.GetValue(Accessibility.IsInAccessibleTreeProperty) ?? _defaultFocusable);
 		}
-
-		bool SetHint(string hint = GetFromElement)
-		{
+				
+		 bool SetHint()
+		 {
 			if (Element == null || Control == null)
 			{
 				return false;
@@ -137,18 +136,15 @@ namespace Xamarin.Forms.Platform.Android.FastRenderers
 				return true;
 			}
 
-			if (_defaultHint == null)
+			string accessibilityName = (string)Element.GetValue(Accessibility.NameProperty);
+			string accessibilityHint = (string)Element.GetValue(Accessibility.HintProperty);
+
+			if (string.IsNullOrWhiteSpace(accessibilityName) && string.IsNullOrWhiteSpace(accessibilityHint))
 			{
-				_defaultHint = textView.Hint;
+				return true;
 			}
 
-			string value = hint;
-			if (value == GetFromElement)
-			{
-				value = string.Join((String.IsNullOrWhiteSpace((string)(Element.GetValue(Accessibility.NameProperty))) || String.IsNullOrWhiteSpace((string)(Element.GetValue(Accessibility.HintProperty)))) ? "" : ". ", (string)Element.GetValue(Accessibility.NameProperty), (string)Element.GetValue(Accessibility.HintProperty));
-			}
-
-			textView.Hint = !string.IsNullOrWhiteSpace(value) ? value : _defaultHint;
+			textView.Hint = string.Join(". ", accessibilityName, accessibilityHint);
 
 			return true;
 		}
--- a/Xamarin.Forms.Platform.Android/FastRenderers/AccessibilityProvider.cs
+++ b/Xamarin.Forms.Platform.Android/FastRenderers/AccessibilityProvider.cs
@@ -144,12 +144,7 @@ namespace Xamarin.Forms.Platform.Android.FastRenderers
 				return true;
 			}
 
-		        string divider = ". ";
-
-			if (string.IsNullOrWhiteSpace(accessibilityName) || string.IsNullOrWhiteSpace(accessibilityHint))
-			{
-				divider = "";
-			}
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.Android/FastRenderers/AccessibilityProvider.cs
```diff
@@ -145,8 +145,7 @@ bool SetHint(string hint = GetFromElement)
 			string value = hint;
 			if (value == GetFromElement)
 			{
-				value = string.Join(". ", (string)Element.GetValue(Accessibility.NameProperty),
-					(string)Element.GetValue(Accessibility.HintProperty));
+				value = string.Join((String.IsNullOrWhiteSpace((string)(Element.GetValue(Accessibility.NameProperty))) || String.IsNullOrWhiteSpace((string)(Element.GetValue(Accessibility.HintProperty)))) ? "" : ". ", (string)Element.GetValue(Accessibility.NameProperty), (string)Element.GetValue(Accessibility.HintProperty));
```
StephaneDelcroix: This is correct, but it's a very complicated, and quite unreadable, way of writing:
```
var name = (string)Element.GetValue(Accessibility.NameProperty);
var hint = (string)Element.GetValue(Accessibility.HintProperty);
if (string.IsNullOrEmpty(name))
    return hint;
if (string.IsNullOrEmpty(hint))
    return name;
return $"{name}. {hint}"
```

bentmar: Agree, some other thoughts on the setHint method.

line 146 if (value == GetFromElement)
My eyes tells me this will allways be true since value is just a reference to hint wich  is allways pointing to GetFromElement. (I think there is a reason for this, but i just dont see it)

line 151 textView.Hint = !string.IsNullOrWhiteSpace(value) ? value : _defaultHint;
since value never will be null due to the join ". " the hint will allways be set to ". " if no Accessibility content is set.

wouldnt it be more efficient just to check if the accessibility properties has values, if not early return?

bentmar: have a look [http://imgur.com/Min569d](url)

bentmar: by not understanding the meaning of GetFromElement, i would do something like this
`bool SetHint()
		{
			if (Element == null || Control == null)
			{
				return false;
			}

			var textView = Control as TextView;
			if (textView == null)
			{
				return false;
			}

			// Let the specified Title/Placeholder take precedence, but don't set the ContentDescription (won't work anyway)
			if (((Element as Picker)?.Title ?? (Element as Entry)?.Placeholder) != null)
			{
				return true;
			}

			string accessibilityName = (string)Element.GetValue(Accessibility.NameProperty);
			string accessibilityHint = (string)Element.GetValue(Accessibility.HintProperty);

			if (string.IsNullOrWhiteSpace(accessibilityName) && string.IsNullOrWhiteSpace(accessibilityHint))
			{
				return true;
			}

			textView.Hint = string.Join(". ", accessibilityName, accessibilityHint);

			return true;
		}`

samhouts:
In #889, I extracted the methods that set `ContentDescription` to extension methods so we can have one place where we do this work. I also altered your original code to make it slightly more readable but have the same effect. See https://github.com/xamarin/Xamarin.Forms/pull/889/files#diff-bfdb4d69d20e6066fef7d616da826521R122. My thought is that we can take this patch as is for inclusion in 2.3.5, since I'm not sure that #889 will be included until 2.3.6, with the knowledge that this patch is temporary anyway.

Xamarin.Forms.Platform.Android/FastRenderers/AccessibilityProvider.cs
StephaneDelcroix: Please update the PR Description with the 'why' you're doing this. For historical purposes

bentmar:
Now i updated my code and the PR description

Xamarin.Forms.Platform.Android/FastRenderers/AccessibilityProvider.cs
```diff
@@ -137,19 +136,15 @@ bool SetHint(string hint = GetFromElement)
...
-			if (value == GetFromElement)
+			if (string.IsNullOrWhiteSpace(accessibilityName) && string.IsNullOrWhiteSpace(accessibilityHint))
 			{
-				value = string.Join(". ", (string)Element.GetValue(Accessibility.NameProperty),
-					(string)Element.GetValue(Accessibility.HintProperty));
+				return true;
 			}
 
-			textView.Hint = !string.IsNullOrWhiteSpace(value) ? value : _defaultHint;
+			textView.Hint = string.Join(". ", accessibilityName, accessibilityHint);
```
StephaneDelcroix: I no longer understand this PR.
If the name or the hint is null, do you still expect the ". " ?
Also, using string.Join is probably serious overkill

bentmar: I dont know i i did somethig wrong with commits, but that is the old commit you are looking at. I dont think I can be any clearer what I mean with this PR in the description, i even pasted images.

This is a newer commit where i changed the code ( its included in this PR)
https://github.com/xamarin/Xamarin.Forms/pull/904/commits/09b4ce162e8d18937198a5e1ec204e58be998e68

bentmar: Idont know what to expect if only one value is provided. Maybe someone at microsoft/xamarin (you?) has that answer, from the original code its not possible to understand how its supposed to work :P.  If the join is a no go we should probably clean it up in the standard renderer aswell.  Il update my code again. The meaning of this PR is solely to not give TextViews a hint of ". "

samhouts:
@bentmar Thank you very much for catching this and taking the time to submit this PR. Since we made some major conflicting changes in the a11y area, we decided to go ahead and create a new PR that includes a fix for this issue, as well as a UI test to prevent recurrence. We'll go ahead and close this one, but please feel free to offer comments on #915 if you see any issues. Thanks again!

PR #906, opened 2017-05-05
make OnShouldReturn protected

kadiryazgan:
There are situations where we don't want to dismiss the keyboard when return key is pressed.
Making OnShouldReturn protected will let custom renderers to disable ResignFirstResponder behaviour.

### Description of Change ###

OnShouldReturn is now a protected method.


hartez:
How would this change let custom renderers disable that behavior?

kadiryazgan:
By overriding it and not calling base method.

hartez:
I think you're missing the `virtual` keyword, then.

PR #801, opened 2017-03-04
Fixed issue where errant "." were appearing on droid controls due to …

BradChase2011:
…hints.

### Description of Change ###

Fixed issue where Android Hints were causing a "." to be placed on all entry controls.

### Bugs Fixed ###

Accessibility support on Android caused a ". " to be place on every control.

No tests believe to be needed. If there is another issue where certain hints are not being defaulted on certain setups then tests will probably be needed.

### API Changes ###

None

### Behavioral Changes ###

Android will now have errant ". " replaced with "" in the UI where Hint and Name are non existent.
### PR Checklist ###

- [X] Has tests (if omitted, state reason in description)
- [X] Rebased on top of master at time of PR
- [X] Changes adhere to coding standard
- [X] Consolidate commits as makes sense

PR #915, opened 2017-05-10
[Android] Remove the ". " on empty labels (Accessibility) on Fastrenderers

samhouts:
### Description of Change ###

This is an alternative to #904 (and revision to #801). This will clarify and simplify the code a bit more. 

Note that this code will be moved to extension methods if #889 is accepted.

Also adds a UI test to prevent recurrence.

Also removed the nullable parameters in the `AutomationPropertiesProvider` that are no longer necessary.

### Bugs Fixed ###

- Reported in #904 and #801
- [Bug 56444 - A small "dot" shows up as value in Android labels when the value is null/empty.](https://bugzilla.xamarin.com/show_bug.cgi?id=56444)

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/FastRenderers/AutomationPropertiesProvider.cs
+++ b/Xamarin.Forms.Platform.Android/FastRenderers/AutomationPropertiesProvider.cs
@@ -139,18 +139,17 @@ namespace Xamarin.Forms.Platform.Android.FastRenderers
 			return true;
 		}
 
-		static string ConcatenateNameAndHelpText(Element Element)
+		internal static string ConcatenateNameAndHelpText(Element Element)
 		{
-			string separator;
 			var name = (string)Element.GetValue(AutomationProperties.NameProperty);
 			var helpText = (string)Element.GetValue(AutomationProperties.HelpTextProperty);
 
-			if (string.IsNullOrWhiteSpace(name) || string.IsNullOrWhiteSpace(helpText))
-				separator = "";
-			else
-				separator = ". ";
+			if (string.IsNullOrWhiteSpace(name))
+				return helpText;
+			if (string.IsNullOrWhiteSpace(helpText))
+				return name;
 
-			return $"{name}{separator}{helpText}";
+			return $"{name}. {helpText}";
 		}
 
 		void SetLabeledBy()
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Platform.Android/FastRenderers/AutomationPropertiesProvider.cs
```diff
@@ -142,18 +132,27 @@ bool SetHint(string hint = GetFromElement)
...
+			string separator;
+			var name = (string)Element.GetValue(AutomationProperties.NameProperty);
+			var helpText = (string)Element.GetValue(AutomationProperties.HelpTextProperty);
+
+			if (string.IsNullOrWhiteSpace(name) || string.IsNullOrWhiteSpace(helpText))
+				separator = "";
+			else
+				separator = ". ";
+
+			return $"{name}{separator}{helpText}";
```
StephaneDelcroix: 1/ Are you sure you mean `IsNullOrWhiteSpace` and not `IsNullOrEmpty` ?

2/ Should be:
```
if (string.IsNullOrWhiteSpace(name)
    return helpText;
if (string.IsNullOrWhiteSpace(helpText)
    return name;

return $"{name}. {helpText}"
```
so you don't concatenate empty strings (this 2/ is only valid if you don't care about whitespaces

Xamarin.Forms.Platform.Android/FastRenderers/AutomationPropertiesProvider.cs
```diff
@@ -142,18 +132,27 @@ bool SetHint(string hint = GetFromElement)
...
-					(string)Element.GetValue(AutomationProperties.HelpTextProperty));
-			}
+			string value = ConcatenateNameAndHelpText(Element);
 
 			textView.Hint = !string.IsNullOrWhiteSpace(value) ? value : _defaultHint;
 
 			return true;
 		}
 
+		static string ConcatenateNameAndHelpText(Element Element)
```
StephaneDelcroix: I'd AggressiveInline this

Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
```diff
@@ -411,6 +411,20 @@ protected virtual bool SetHint()
 			return true;
 		}
 
+		static string ConcatenateNameAndHelpText(Element Element)
```
StephaneDelcroix: do we need this declared twice ?

BradChase2011:
@samhouts Saam, do you know how far away this is from being accepted?  Merge wise, do we need to merge this with other PRs or are we good just merging this in for the time being until its accepted?

PR #350, opened 2016-09-13
AppCompat should now be able to set title bar visibility programmatic…

adrianknight89:
### Description of Change ###

Created a bug 8 months ago that got no love. Decided to fix it myself. If you use AppCompat, you can't hide title bar (title bar controls are hidden but the background is still visible.) This change ensures that developers can toggle status bar visibility programmatically. Right now, there are two ways to achieve this:

- Call `Xamarin.Forms.Forms.SetTitleBarVisibility(param)`
- Add / clear fullscreen window flag

Updating title bar visibility should set the equivalent window flag or vice versa.

There is still more work to be done. For example, ~~there is no way at the moment to hide status bar on splash screen~~. Also not sure if `SystemUiFlags` works (haven't tested immersive stuff). I will either add to this PR to fix those or submit another in the future.

EDIT: 

Setting `<item name="android:windowFullscreen">true</item>` should work because this actually raises `OnWindowAttributesChanged()` with the fullscreen flag added.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=38019

### API Changes ###

Changed:
 - void LayoutRootPage(Page page, int width, int height) => internal static void LayoutRootPage(FormsAppCompatActivity activity, Page page, int width, int height)

- internal static AndroidTitleBarVisibility TitleBarVisibility { get; private set; } => internal static AndroidTitleBarVisibility TitleBarVisibility { get; set; }

PR #916, opened 2017-05-10
Do not layout page in OnWindowAttributesChanged on fullscreen flag change

adrianknight89:
### Description of Change ###

I wrote #350 a while ago so we could programmatically toggle the status bar on and off both at application start and during runtime. However, this was needed only because of the dummy underlay added to the overall view tree (which was recently removed by #892). We don't need to layout the root page anymore. This cuts down startup time by as much as 40 ms (at least in my case) if you were hiding the status bar behind the splash screen.

The following still works:

`<item name="android:windowFullscreen">true</item>` in `styles.xml` will not show the status bar at app start.

`global::Xamarin.Forms.Forms.SetTitleBarVisibility(Xamarin.Forms.AndroidTitleBarVisibility.Never);` in `MainActivity` will add `WindowManagerFlags.Fullscreen` flag to window attributes and hide the status bar.

Alternatively, one could add this flag manually in a dependency service call which should sync Form's `TitleBarVisibility`.

Please note that this PR assumes #892 will make it to production.

### Bugs Fixed ###

- N/A

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/Platform.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/Platform.cs
@@ -196,7 +196,7 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 		{
 			if (changed)
 			{
-				LayoutRootPage((FormsAppCompatActivity)_context, Page, r - l, b - t);
+				LayoutRootPage(Page, r - l, b - t);
 			}
 
 			Android.Platform.GetRenderer(Page).UpdateLayout();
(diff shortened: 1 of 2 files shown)


Xamarin.Forms.Platform.Android/AppCompat/Platform.cs
```diff
@@ -285,8 +285,9 @@ bool HandleBackPressed(object sender, EventArgs e)
 			return handled;
 		}
 
-		internal static void LayoutRootPage(FormsAppCompatActivity activity, Page page, int width, int height)
+		void LayoutRootPage(Page page, int width, int height)
```
adrianknight89: This can be private again.

PR #918, opened 2017-05-12
Remove VisualElement finalizer

hartez:
### Description of Change ###

The finalizer added to fix [44074](https://bugzilla.xamarin.com/show_bug.cgi?id=44074) causes an InvalidOperationException during the .NET finalization process because it attempts to operate on other objects which are not guaranteed to be alive and which contain ConditionalWeakTables which may or may not have already had their finalizers run. This ends up causing the ConditionalWeakTable to fail its integrity check. Detaching behaviors and triggers during finalization cannot safely be done, because there are no guarantees of the order that the objects are collected/finalized.

This change removes the finalizer so the exception does not occur.

This change also removes the test `TestBehaviorsAreDetachedBeforeGarbageCollection`, which actually tested whether behaviors were detached during finalization, which cannot safely be done.

The finalizer was added to fix [44074 – OnDetachingFrom for Behavior is never called, therefore memory can leak due to event handlers not being unbound](https://bugzilla.xamarin.com/show_bug.cgi?id=44074). 44074 is not a bug; there is no way to magically remove behaviors from controls as they are garbage collected, and there is no deterministic disposal mechanism for VisualElement. This is [documented behavior](https://developer.xamarin.com/guides/xamarin-forms/application-fundamentals/behaviors/creating/#Removing_a_Behavior_from_a_Control) - Behaviors have to be explicitly removed from controls. 

The potential for leaks only exists in the situation where an attached Behavior holds a reference to the VisualElement and something holds a reference to the Behavior; for many situations (e.g., the [XAML example in the documentation](https://developer.xamarin.com/guides/xamarin-forms/application-fundamentals/behaviors/creating/#Consuming_a_Xamarin.Forms_Behavior) ), this is not an issue. It could potentially be an issue if a Behavior reference is held Application-wide (e.g., in a Style), in which case the Behavior would need to be explicitly removed when its target is no longer relevant (e.g., in Page.Disappearing).

### Bugs Fixed ###

- [55365 – ~VisualElement crashes with System.Runtime.InteropServices.COMException](https://bugzilla.xamarin.com/show_bug.cgi?id=55365)

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla55365.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla55365.cs
@@ -1,75 +1,75 @@
-﻿using Xamarin.Forms.CustomAttributes;
-using Xamarin.Forms.Internals;
-using System.Collections;
-using System.Collections.Generic;
-using System.Collections.Specialized;
+﻿using System;
 using System.Collections.ObjectModel;
-using System;
-
+using System.Collections.Specialized;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
 #if UITEST
-using Xamarin.UITest;
 using NUnit.Framework;
-#endif
 
-// Apply the default category of "Issues" to all of the tests in this assembly
-// We use this as a catch-all for tests which haven't been individually categorized
-#if UITEST
-[assembly: NUnit.Framework.Category("Issues")]
 #endif
 
 namespace Xamarin.Forms.Controls.Issues
 {
 	[Preserve(AllMembers = true)]
 	[Issue(IssueTracker.Bugzilla, 55365, "~VisualElement crashes with System.Runtime.InteropServices.COMException", PlatformAffected.UWP)]
-	public class Bugzilla55365 : TestContentPage // or TestMasterDetailPage, etc ...
+	public class Bugzilla55365 : TestContentPage
 	{
+		readonly StackLayout _itemsPanel = new StackLayout();
+		readonly DataTemplate _itemTemplate = new DataTemplate(CreateBoxView);
+		readonly StackLayout _layout = new StackLayout();
+
+#if UITEST
+		[Test]
+		public void ForcingGCDoesNotCrash()
+		{
+			RunningApp.WaitForElement("Clear");
+			RunningApp.Tap("Clear");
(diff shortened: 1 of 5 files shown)


mattrichnz:
Thanks for the quick turnaround on this issue. Hopefully this fix can be provided in a XF 2.3.4 service release once merged.

kvpt:
Hi,

With this code removed, how to detach events suscribed during the behavior/trigger attachment.

In the [official example](https://developer.xamarin.com/guides/xamarin-forms/application-fundamentals/behaviors/creating/#Creating_a_Xamarin.Forms_Behavior), the **OnDetachingFrom** method is used.

Before this modification behavior/trigger was never removed and **OnDetachingFrom** method was never called (issue [44074](https://bugzilla.xamarin.com/show_bug.cgi?id=44074)) in contradiction to the documentation exemple in the link above.

In my application, a behavior is attached to listview elements's by the xaml and I can't access individual items to remove the attached behavior by code.

``` xaml
    <ListView>
        <ListView.ItemTemplate>
            <DataTemplate>
                <ViewCell>
                       ...
                        <Label>
                            <Label.Behaviors>
                                <behaviors:CustomVisualElementBehavior/>   <-- How to detach this ? -->
                            </Label.Behaviors>
                        </Label>
                       ...
                </ViewCell>
            </DataTemplate>
        </ListView.ItemTemplate>
    </ListView>
```

I want to be able to call the  **OnDetachingFrom** method to cancel an async method launched inside the **OnAttachedTo** method.

hartez:
> I want to be able to call the OnDetachingFrom method to cancel an async method launched inside the OnAttachedTo method.

I'm not sure a Behavior is the best way to accomplish your goal (I don't think this is how `OnDetachingFrom` is really meant to be used), but I don't know much about your specific situation so I might be wrong.

That said, one way for accomplish this would be to modify your Behavior to [detach itself when the control's `BindingContext` is set to `null`](https://github.com/hartez/XFBehaviorCleanupExample/blob/master/BehaviorCleanup2/BehaviorCleanup2/NumericValidationBehavior.cs). Then, when you want to clean up your Behaviors, you can simply [set the `ItemsSource` of your ListView to `null`](https://github.com/hartez/XFBehaviorCleanupExample/blob/master/BehaviorCleanup2/BehaviorCleanup2/MainPage.xaml.cs), and `OnDetachingFrom` will be called.

[Small example project here](https://github.com/hartez/XFBehaviorCleanupExample)

PR #451, opened 2016-10-13
[Android] Leverage new commit methods in Nougat

adrianknight89:
### Description of Change

API 24 has introduced new commit methods ([commitNow()](https://developer.android.com/reference/android/support/v4/app/FragmentTransaction.html#commitNow%28%29) and [commitNowAllowingStateLoss()](https://developer.android.com/reference/android/support/v4/app/FragmentTransaction.html#commitNowAllowingStateLoss%28%29)). I know that Xamarin Android Support Library v24 is in RC now, so this will have to wait until it's GA'ed, but I wanted to submit the work before I forget about it. In many ways, these methods should be preferable to using `executePendingTransactions()`. Nowhere in the code is `addToBackStack()` called, so these methods should not throw an `IllegalStateException`.

See [The many flavors of commit()](https://medium.com/@bherbst/the-many-flavors-of-commit-186608a015b1#.axs7wlxnd).

I haven't had a chance to test run the changes because I couldn't compile the Android Platform project with v24. I'll test it all when I can, but feel free to submit feedback now if you like.

I'll have to update this if #444 is merged.
### PR Checklist
- [ ] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [ ] Consolidate commits as makes sense

PR #919, opened 2017-05-12
[Android] Do not call DisallowAddToBackStack()

adrianknight89:
### Description of Change ###

Calling `DisallowAddToBackStack()` is costing me more than 100 ms. I'm not sure why we need this. As per the [documentation](https://developer.android.com/reference/android/app/FragmentTransaction.html#disallowAddToBackStack()), this method throws an exception if we have already called `AddToBackStack()`. Since it doesn't throw an exception in our case, it means we haven't. I also searched the codebase for any references and did not find any. Another purpose of this method is to make sure any subsequent calls to `AddToBackStack()` will throw an exception so it acts as a safety check.

I think we could let go of this safety check because, if we call `AddToBackStack()`, I'd assume that it was on purpose.

I have tested `NavigationPage`. Please test MDP to make sure all is okay.

Also referencing #451 so if it gets added, we should remove `DisallowAddToBackStack()` from there as well.

Maybe I'm not fully understanding the use case of this method, so feel free to leave feedback.

### Bugs Fixed ###

- N/A

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/MasterDetailContainer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/MasterDetailContainer.cs
@@ -69,7 +69,6 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 				{
 					// But first, if the previous occupant of this container was a fragment, we need to remove it properly
 					FragmentTransaction transaction = FragmentManager.BeginTransaction();
-					transaction.DisallowAddToBackStack();
 					transaction.Remove(_currentFragment);
 					transaction.SetTransition((int)FragmentTransit.None);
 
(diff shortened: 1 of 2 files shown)


StephaneDelcroix:
need test results

adrianknight89:
@hartez What's stopping this and #920 from being merged?

hartez:
@adrianknight89 Needs a second review, and it looks like the second review was pending passing UI tests. UI tests were failing for an unrelated reason - I've kicked off a new round. Assuming they pass, shouldn't be a problem to merge it.

rmarinho:
can you rebase @adrianknight89 so we can pass builds with fast renderes

hartez:
@adrianknight89 @rmarinho The CI build was just failing because of a pending Android bug which generates a warning. I updated the PR to turn off those warnings so it would build in CI and run the tests - they should be running now.

adrianknight89:
@hartez Does this require a rebase still then?

rmarinho:
think is better.

adrianknight89:
One of these days, VS will be nice and stop turning rebase into merge commits... Should be done. I think the rebase includes those `<NoWarn>CS0109</NoWarn>` tags?

PR #920, opened 2017-05-14
[Android] Avoid using GetGesturesFor when checking pinch gesture presence

adrianknight89:
### Description of Change ###

When processing touch events, `VisualElementRenderer` is checking if the current view has a pinch gesture. Instead of using `GetGesturesFor`, we can use LINQ's built-in `OfType` method. This seems to reduce processing time by as much as 8 ms (in my case).

Note that `GetGesturesFor` seems to be used in many different places. If there is room for improvement in those, one should take a look at. 

Also removed a redundant cast.

### Bugs Fixed ###

- N/A

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/PinchGestureHandler.cs
+++ b/Xamarin.Forms.Platform.Android/PinchGestureHandler.cs
@@ -18,7 +18,7 @@ namespace Xamarin.Forms.Platform.Android
 			get
 			{
 				View view = GetView();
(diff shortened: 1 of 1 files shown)


StephaneDelcroix:
need test results, then merge @hartez

hartez:
Tests failed for an unrelated reason, I've queued them up again.

hartez:
Incorporated this suggestion into #1135.

PR #921, opened 2017-05-15
[Android] Defer setting accent color during initialization

adrianknight89:
### Description of Change ###

We don't need to set accent color during initialization and can defer it until first use. This seems to reduce startup time by about 20 ms (theme attribute resolution + call to Color.FromUint()). Subscribing a new Message seems to add 1 ms which is negligible compared to the gain. Surprisingly, setting accent color after main page is loaded seems to be instantaneous.

### Bugs Fixed ###

- N/A

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Color.cs
+++ b/Xamarin.Forms.Core/Color.cs
@@ -31,8 +31,19 @@ namespace Xamarin.Forms
 		}
 
 		[EditorBrowsable(EditorBrowsableState.Never)]
-		public static void SetAccent(Color value) => Accent = value;
-		public static Color Accent { get; internal set; }
+		public static void SetAccent(Color value) => s_accent = value;
+
+		static Color s_accent;
+		public static Color Accent
+		{
+			get
+			{
+				if (Device.RuntimePlatform == Device.Android && s_accent.IsDefault)
+					MessagingCenter.Send(Application.Current.MainPage, Page.AccentColorSignalName);
+
+				return s_accent;
+			}
+		}
 
 		readonly float _a;
 
--- a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
@@ -9,9 +9,7 @@ using Android.Content;
 using Android.Content.Res;
 using Android.OS;
 using Android.Runtime;
-using Android.Support.V4.Content;
 using Android.Support.V7.App;
-using Android.Util;
 using Android.Views;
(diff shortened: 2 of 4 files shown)


Xamarin.Forms.Platform.Android/Forms.cs
```diff
@@ -172,16 +170,16 @@ static Color GetAccentColor()
 					if (sdkVersion <= 10)
```
adrianknight89: not sure if we need this version check?

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -9,9 +9,7 @@
 using Android.Content.Res;
 using Android.OS;
 using Android.Runtime;
-using Android.Support.V4.Content;
```
adrianknight89: redundant namespaces

Xamarin.Forms.Core/Color.cs
```diff
@@ -31,8 +31,19 @@ public bool IsDefault
...
-		public static void SetAccent(Color value) => Accent = value;
-		public static Color Accent { get; internal set; }
+		public static void SetAccent(Color value) => s_accent = value;
+
+		static Color s_accent;
+		public static Color Accent
+		{
+			get
+			{
+				if (Device.RuntimePlatform == Device.Android && s_accent.IsDefault)
```
adrianknight89: This is getting material accent color if Core accent color is default. Since `SetAccent` is a public method, it is possible that one sets accent color to default in which case we'd re-set it to theme-specific accent color. However, I'm not sure why anybody would do this on purpose.

adrianknight89:
@StephaneDelcroix You may be right, and I thought about it last night, but... in my tests, I observed the following:

`Context.Theme.ResolveAttribute(global::Android.Resource.Attribute.ColorAccent, value, true);` takes 3 ms
`(uint)value.Data` takes another 3 ms
`Color.FromUint();` takes between 10 and 12 ms

We're not deferring theme initialization but setting `Color.Accent` which is a Forms property. Also a bulk of this operation seems to come from creating a `Color` object (bitwise ops + HSL conversion).

Then again, I have no explanation as to why setting `Color.Accent` through a message call seems to be instantaneous. Perhaps, it's because another line of code interacts with the `Color` struct before we do, which moves it to another part of memory or register (sorry, not an optimization expert), and further interactions, at least in the near term, are fast.

StephaneDelcroix:
the time it takes is about the same, the measurement show different values as the device is probably under lighter load. but all in all it doesn't change the amount of processing required.

PR #922, opened 2017-05-15
[UWP] Use transparent as default TabbedPage bar background color

pauldipietro:
### Description of Change ###

When navigating between tabs that change the BarBackgroundColor, returning to a tab where the value would be the default would use the value of `SystemControlBackgroundChromeMediumLowBrush` rather than transparent which is what UWP uses (see the style [here](https://msdn.microsoft.com/en-us/library/windows/apps/mt299144.aspx) for reference, inside of the `PivotPanel`).

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=55674

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla55674.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla55674.cs
@@ -1,0 +1,229 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System.Linq;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Bugzilla, 55674, "[UWP] Using a navigation page changes the tabbed bar's color", PlatformAffected.UWP)]
+	public class Bugzilla55674 : TestTabbedPage
+	{
+		protected override void Init()
+		{
+			Children.Add(new FirstPage());
+			Children[0].Title = "Page One";
+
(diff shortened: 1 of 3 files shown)


jassmith:
Looks like this might break behavior when not inside a nav page?

hartez:
@pauldipietro  This might break if the Tabbed Page is inside of a Navigation Page - can you write a separate test with that scenario?

samhouts:
@pauldipietro Please see comments. Thanks! :)

samhouts:
@pauldipietro We still need a test for a TabbedPage inside of a NavigationPage, when you've got time to come back to this. Thanks!

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla55674.cs
```diff
@@ -0,0 +1,229 @@
...
+				Content = new StackLayout
+				{
+					Children =
+					{
+						new Label { Text = "Page Four / TabbedPage's Page Two"}
+					}
+				}
+			});
+
+			Children.Add(new NavigationPage(tabbedPage));
```
pauldipietro: I believe I had force pushed this addition here if you get a chance to check it out.

rmarinho:
Can you rebase please @pauldipietro  ?

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla55674.cs
samhouts: Excellent! While you're rebasing, can you put together some before/after screenshots for us? Thanks!!

rmarinho:
@pauldipietro can you open a new pr for this, no need for WINDOWS_UWP now.

PR #924, opened 2017-05-15
Fix bz56030; DynamicResource overriding

kingces95:
### Description of Change ###

Prior to this change Element tracked which DynamicResources would trigger a change in one of its properties by storing a List of BindableProperty/Key pairs where Key was the name of a DynamicResource. When a change to the DynamicResources was reported to the Element via Element.OnResourcesChanged the Element would compare the names of the changed DynamicResources to it's list of DynamicResource key names and when there was a match re-initialize the associated BindableProperty. This worked so long as the the Element had a unique set of Keys. However this bug demonstrates that's not always the case.

This bug constructed an example whereby a Label had two different DynamicResources bound to its  Label.TextColor BindableProperty. The first was bound via a Style declared at the application level and a second bound directly to the Label. Which should have precedence and control the TextColor? Clearly the one applied directly to the label should override the style. However this didn't always happen. Why? Because the Element stored both associations in it's list and would apply them in a "random" order (depending on the order the resources were declared) when the DynamicResources were refreshed.

The fix is to have Element store an Element's DynamicResources in a Dictionary keyed by its BindableProperties. This way an Element is guaranteed to associate a BindableProperty with a single DynamicResource and so long as the Element's DynamicResources are added in increasing order of precedence (as appears to be the case) then the last assignment will overwrite previous assignments leaving the highest precedent binding controlling the BindableProperty value. 

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=56030

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/DynamicResourceTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/DynamicResourceTests.cs
@@ -1,4 +1,5 @@
 ﻿using NUnit.Framework;
+using Xamarin.Forms.Internals;
 
 namespace Xamarin.Forms.Core.UnitTests
 {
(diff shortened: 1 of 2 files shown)


kingces95:
Agreed. I bet the original choice of variable names contributed to confusion that led to the bug. 

I could change `_dynamicResources` to `_dynamicallyBoundProperties` which would then be keyed by a BindableProperty to produce a DynamicResource name. Then when iterating that dictionary I could change the iteration variable from `dynR` to `dynamicallyBoundProperty` and `values` to `changingDynamicResource` so that the if statement in the join loop could then read:

    if (dynamicallyBoundProperty.Value != changingDynamicResource.Key) continue;

Pretty verbose. But either way the BP maps to a DR name and the changing DRs are delivered in a list of DR name/DR value pairs. So we're gonna have a `.Value` compared with `.Key`.

kingces95:
Ok!

StephaneDelcroix:
Also fixes https://bugzilla.xamarin.com/show_bug.cgi?id=56852. updating the description

PR #927, opened 2017-05-19
[Android] Correctly size content after toggling soft keyboard in Resize mode 

hartez:
### Description of Change ###

This change addresses two bugs when `WindowSoftInputModeAdjust` is set to `Resize` on Android.

When the soft keyboard is dismissed and the current page contains a ScrollView, this change forces the ScrollView's container to re-layout to accommodate the updated layout size.

When the soft keyboard is dismissed as part of a navigation pop operation, this change forces the page at the top of the stack to re-layout to accommodate the updated layout size.

This change also consolidates the identical base classes for the Entry and Editor renderers on Android and adds `Dispose` overrides to unhook the `OnKeyboardBackPressed` handlers.

### Bugs Fixed ###

- [45215 – AdjustResize Causes Content to Be Sized Incorrectly After Toggling Keyboard](https://bugzilla.xamarin.com/show_bug.cgi?id=45215)
- [55558 – Navigation bar back button with keyboard white space in adjust mode Resize](https://bugzilla.xamarin.com/show_bug.cgi?id=55558)

### API Changes ###

None

### Behavioral Changes ###

If a custom renderer derived from `EntryRenderer` or `EditorRenderer` is explicitly casting its `Control` to `EntryEditText` or `EditorEditText`, this change will cause a compile-time error. The target class of the cast will have to be updated to `FormsEditText`.

If a custom renderer derived from `EntryRenderer` or `EditorRenderer` is adding a handler for the `OnBackKeyboardPressed` event (formerly on `EditorEditText`), this change will cause a compile-time error. The handler will need to be modified to use the `OnKeyboardBackPressed` event instead.

### PR Checklist ###


Xamarin.Forms.Platform.Android/Renderers/FormsEditText.cs
```diff
@@ -6,11 +6,11 @@
 
 namespace Xamarin.Forms.Platform.Android
 {
-	public class EntryEditText : EditText, IDescendantFocusToggler
+	public class FormsEditText : EditText, IDescendantFocusToggler
```
jassmith: Why the rename?

hartez: Because it's used for both Entry and Editor.
Previously there were two classes (EntryEditText and EditorEditText) which were identical except for a transposed event name. Consolidating them makes things simpler and JITs fewer classes.

StephaneDelcroix: that might break custom renderers inheriting from ours.

Xamarin.Forms.Platform.Android/Renderers/EditorRenderer.cs
```diff
@@ -89,6 +89,26 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
...
+				return;
+			}
+
+			_disposed = true;
+
+			if (disposing)
+			{
+				if (Control != null)
+				{
+					Control.OnKeyboardBackPressed -= OnKeyboardBackPressed;
```
StephaneDelcroix: 👍

Xamarin.Forms.Platform.Android/Renderers/FormsEditText.cs
hartez: It might break a custom renderer which subclasses `EntryRenderer` or `EditorRenderer` and then explicitly casts `Control` to `EditText` or `EntryEditText`, but that's a pretty low risk and a very simple change if it does happen. I'm happy to take the risk. I'll call out the possible issues in the PR description so they make it into the release notes (and tag @jassmith on this comment to give him the opportunity to veto).

Xamarin.Forms.Platform.Android/Renderers/EditorRenderer.cs
hartez: It might. I've called out the potential break in the PR description; I don't think it's likely a problem (and it's an easy change to make), but again I'll give @jassmith a veto opportunity.

PR #933, opened 2017-05-24
[Android] Fix NullPointerException following RemovePage call

samhouts:
### Description of Change ###

Starting with API 25, calling `Navigation.RemovePage()` will cause any subsequent navigation calls to error with a `NullPointerException`:

```
Java.Lang.NullPointerException: Attempt to invoke virtual method 'void android.view.View.unFocus(android.view.View)' on a null object reference
  at java.lang.NullPointerException: Attempt to invoke virtual method 'void android.view.View.unFocus(android.view.View)' on a null object reference
  at at android.view.ViewGroup.removeViewInternal(ViewGroup.java:4702)
  at at android.view.ViewGroup.removeViewAt(ViewGroup.java:4665)
  at at android.support.v4.app.FragmentManagerImpl.moveFragmentToExpectedState(FragmentManager.java:1540)
  at at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1595)
```

This PR will stop removing the `PageContainer` for the `Page` being removed (i.e., the `View` that is represented by the `fragment`) on `RemovePage` when the API level is >= 25. Based on the stack trace and the existence of two new issues on the Android issue tracker with a similar stack trace ([[1]](https://issuetracker.google.com/issues/37469028) [[2]](https://issuetracker.google.com/37138029)), it appears that removing a fragment will now automatically remove its related `View` without checking to see if it has already been removed and disposed.

### Bugs Fixed ###

- [Bug 53179 - PopAsync crashing after RemovePage when support packages are updated to 25.1.1](https://bugzilla.xamarin.com/show_bug.cgi?id=53179)

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla53179.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla53179.cs
@@ -6,16 +6,10 @@ using Xamarin.UITest;
 using NUnit.Framework;
 #endif
 
-// Apply the default category of "Issues" to all of the tests in this assembly
-// We use this as a catch-all for tests which haven't been individually categorized
-#if UITEST
-[assembly: NUnit.Framework.Category("Issues")]
-#endif
-
 namespace Xamarin.Forms.Controls.Issues
 {
 	[Preserve(AllMembers = true)]
-	[Issue(IssueTracker.Bugzilla, 9953179, "PopAsync crashing after RemovePage when support packages are updated to 25.1.1", PlatformAffected.Android)]
+	[Issue(IssueTracker.Bugzilla, 53179, "PopAsync crashing after RemovePage when support packages are updated to 25.1.1", PlatformAffected.Android)]
 	public class Bugzilla53179 : TestNavigationPage
 	{
 		class TestPage : ContentPage
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla53179.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla53179.cs
@@ -1,0 +1,81 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+// Apply the default category of "Issues" to all of the tests in this assembly
+// We use this as a catch-all for tests which haven't been individually categorized
+#if UITEST
+[assembly: NUnit.Framework.Category("Issues")]
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla53179.cs
```diff
@@ -0,0 +1,89 @@
...
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+// Apply the default category of "Issues" to all of the tests in this assembly
+// We use this as a catch-all for tests which haven't been individually categorized
+#if UITEST
+[assembly: NUnit.Framework.Category("Issues")]
```
hartez: This only needs to be in the project once, you can remove it from here.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla53179.cs
```diff
@@ -0,0 +1,89 @@
...
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Bugzilla, 9953179, "PopAsync crashing after RemovePage when support packages are updated to 25.1.1", PlatformAffected.Android)]
```
hartez: "9" trick.

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
```diff
@@ -534,18 +534,24 @@ void RegisterToolbar()
...
 			{
-				rendererToRemove.View.RemoveFromParent();
-				rendererToRemove.Dispose();
+				var containerToRemove = (PageContainer)rendererToRemove.View.Parent;
+
+				rendererToRemove.View?.RemoveFromParent();
+
+				rendererToRemove?.Dispose();
+
+				if ((int)Build.VERSION.SdkInt < 25)
```
hartez: This needs some kind of comment so we remember in the future why this API check is happening (especially since it might just be an Android bug and we might be able to remove it once the bug is fixed).

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla53179.cs
samhouts: We should probably remove it from the template then...?

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla53179.cs
samhouts: bah!

7934345:
https://bugzilla.xamarin.com/show_bug.cgi?id=53179

Come on, we need it to be fixed in next version!!!

LeonBoon:
When will this branch be merged?

PR #935, opened 2017-05-24
Align Layout transparency behavior between Android, iOS, Windows

hartez:
### Description of Change ###

This change brings iOS, Windows, and Android into alignment with regard to Layout transparency. 

Previously, Layouts with sufficiently low `Opacity` were input transparent on iOS but not on Windows or Android; Layouts with a `BackgroundColor` of `Transparent` or `Default` were input transparent on Android but not on Windows or iOS; and Layouts with `InputTransparent` were input transparent on iOS, mostly input transparent on Android, and not input transparent at all on Windows. 

With this change, input transparency is entirely a function of the `InputTransparent` flag; if a layout is marked `InputTransparent`, then clicks/taps/gestures will pass through it to controls below. Visual transparency (i.e., `Opacity` and the transparency of the background color) no longer has any bearing on whether a layout is is input transparent. 

### Bugs Fixed ###

- Layouts marked InputTransparent on Windows were not actually input transparent
- [25943 – [Android] TapGestureRecognizer does not work with a nested StackLayout](https://bugzilla.xamarin.com/show_bug.cgi?id=25943)

### API Changes ###

None

### Behavioral Changes ###

- On Android, layouts with a BackgroundColor of Clear or Default are no longer effectively InputTransparent
- On iOS, layouts with an Opacity <= 0.01 are no longer effectively InputTransparent

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Platform.cs
+++ b/Xamarin.Forms.Platform.iOS/Platform.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using CoreGraphics;
 using Foundation;
 using UIKit;
 using RectangleF = CoreGraphics.CGRect;
@@ -478,6 +479,30 @@ namespace Xamarin.Forms.Platform.iOS
 
 		internal class DefaultRenderer : VisualElementRenderer<VisualElement>
 		{
+			public override UIView HitTest(CGPoint point, UIEvent uievent)
+			{
+				// UIview hit testing ignores objects which have an alpha of less than 0.01 
+				// (see https://developer.apple.com/reference/uikit/uiview/1622469-hittest)
+				// To prevent layouts with low opacity from being implicitly input transparent, 
+				// we need to temporarily bump their alpha value during the actual hit testing,
+				// then restore it. If the opacity is high enough or user interaction is disabled, 
+				// we don't have to worry about it.
+
+				nfloat old = Alpha;
+				if (UserInteractionEnabled && old <= 0.01)
+				{
+					Alpha = (nfloat)0.011;
+				}
+
+				var result = base.HitTest(point, uievent);
+
+				if (UserInteractionEnabled && old <= 0.01)
+				{
+					Alpha = old;
+				}
+
+				return result;
+			}
 		}
 	}
 }
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla25943.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla25943.cs
@@ -1,0 +1,62 @@
+using System.Diagnostics;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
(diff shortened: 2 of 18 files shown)


Vinnie1991:
Could this PR please merged to the master?
It fixes a high priority bug in my app.

Xamarin.Forms.Platform.iOS/Platform.cs
```diff
@@ -478,6 +479,30 @@ static void PresentPopUp(UIWindow window, UIAlertController alert, ActionSheetAr
...
+				// (see https://developer.apple.com/reference/uikit/uiview/1622469-hittest)
+				// To prevent layouts with low opacity from being implicitly input transparent, 
+				// we need to temporarily bump their alpha value during the actual hit testing,
+				// then restore it. If the opacity is high enough or user interaction is disabled, 
+				// we don't have to worry about it.
+
+				nfloat old = Alpha;
+				if (UserInteractionEnabled && old <= 0.01)
+				{
+					Alpha = (nfloat)0.011;
```
rmarinho: Doesn't this cause the all UIView do redraw ?

hartez: No. The next CALayer update won't happen until after this method is done. Since we're setting the Alpha back to its original value before exiting the method, nothing visibly happens. And since we're only touching a CALayer property, the UIView's Draw method is never triggered.

Xamarin.Forms.Platform.iOS/Platform.cs
```diff
@@ -478,6 +479,30 @@ static void PresentPopUp(UIWindow window, UIAlertController alert, ActionSheetAr
...
 		internal class DefaultRenderer : VisualElementRenderer<VisualElement>
 		{
+			public override UIView HitTest(CGPoint point, UIEvent uievent)
+			{
+				// UIview hit testing ignores objects which have an alpha of less than 0.01 
+				// (see https://developer.apple.com/reference/uikit/uiview/1622469-hittest)
+				// To prevent layouts with low opacity from being implicitly input transparent, 
+				// we need to temporarily bump their alpha value during the actual hit testing,
+				// then restore it. If the opacity is high enough or user interaction is disabled, 
+				// we don't have to worry about it.
```
durandt: Hi,
Great that issue 25943 has a fix on its way. I just want to make sure I understand the impact of it.
If one would have used `View.Opacity=0.0` instead of `View.Enabled = false` to disable a view, would this change the behavior from being input-transparent to handling the input?
I agree this is a misuse of Opacity, I just wanted to be sure I understood the fix.
Would it make sense to modify Alpha only if `0 < old <= 0.01` ?

MartinZikmund:
This change was quite a breaking change in two of my apps. The default XAML behavior in UWP and WPF is that when Background is clear, the visual is click-through, but If it has some children, you can still interact with them. This is very important especially for mapping apps where you need map on top and they layout some control above it. If I set the layer to `InputTransparent` user cannot interact with any controls in the layer in any way.

MartinZikmund:
Please see my StackOverflow question that details my problem further : https://stackoverflow.com/questions/47414685/overlay-layer-without-interactivity-in-xamarin-forms

bukal260:
I have exactly the same problem as @MartinZikmund described. If someone have a workaround let me know.

PR #937, opened 2017-05-25
Moving to version 2.0.9 of Xamarin.UITest

hartez:
### Description of Change ###

Bumps the UI test library version to 2.0.9 so we can run UI tests on machines with XCode 8.3.


hartez:
Test failures are because of the failing 38731 test, they can be ignored.

PR #940, opened 2017-05-29
[XamlG] use random (short) file names for XamlG outputs

StephaneDelcroix:
### Description of Change ###

Use random short filenames for XamlG outputs.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=40077

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###


dotMorten:
I often review the generated files when trouble shooting compilation issues. I'm not so sure it's a good idea to use a random name. Also if you have this file open in VS and build again, a new file name is used, and you won't get a "reload file" window. It would be good if the file name stays intact. The crux of the problem is the file names are ridiculously long, causing issues on systems that can't handle exceeding MAX_PATH. I don't really understand why they need to be that long. Just using a random file name feels like a dirty quick fix to me :)

PR #941, opened 2017-05-29
Fix border on android buttons 

rmarinho:
### Description of Change ###

Don't allow negative border width .
Refactoring and add support for border properties to appcompact and fast renderer button  

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=36031

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/ButtonRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/ButtonRenderer.cs
@@ -1,14 +1,9 @@
 using System;
 using System.ComponentModel;
 using Android.Content;
-using Android.Content.Res;
 using Android.Graphics;
-using Android.Graphics.Drawables;
-using Android.Support.V4.Content;
 using Android.Support.V7.Widget;
 using Android.Util;
-using Xamarin.Forms.Internals;
-using GlobalResource = Android.Resource;
 using Object = Java.Lang.Object;
 using AView = Android.Views.View;
 using AMotionEvent = Android.Views.MotionEvent;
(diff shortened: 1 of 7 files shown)


Xamarin.Forms.Platform.Android/AppCompat/ButtonRenderer.cs
```diff
@@ -27,6 +23,7 @@ public class ButtonRenderer : ViewRenderer<Button, AppCompatButton>, AView.IOnAt
 
 		public ButtonRenderer()
 		{
+			System.Diagnostics.Debug.WriteLine("Slow Button!");
```
StephaneDelcroix: ?

StephaneDelcroix:
you comment says 

> Don't allow negative border width

but the code only checks for negative `cornerRadius`

Xamarin.Forms.Platform.Android/AppCompat/ButtonRenderer.cs
rmarinho: This is something we also added on non app compact button so users select fast renderers, should we remove it @ez ?

rmarinho:
@StephaneDelcroix my comment referes to border width (inset) calculation inside the renderer where if border radius was 0 it will have a negative value for the border width (inset) when drawing

Xamarin.Forms.Platform.Android/AppCompat/ButtonRenderer.cs
hartez: Yeah, that should have been removed a while ago.

PR #943, opened 2017-05-30
[C] use defaultValueCreator to set Frame default Padding

StephaneDelcroix:
### Description of Change ###

Use defaultValueCreator for setting Frame default Padding, so the value is treated as default and can be overridden by a Style.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=30765

### API Changes ###

Added:
 - protected virtual Thickness Layout.CreateDefaultPadding()
 - protected override Thickness Frame.CreateDefaultPadding()

### Behavioral Changes ###

Frame Padding set through Style should now be followed

### PR Checklist ###


jassmith:
Can you mark CreateDefaultPadding to internal virtual

PR #946, opened 2017-05-30
[previewer] make sure we do not crash even if the previewer doesn't s…

StephaneDelcroix:
### Description of Change ###

Make sure we do not crash even if the previewer doesn't sets the flag. This used to work, but somehow regressed when the previewer started to use the XamlLoader API...

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=56895

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Xaml/XamlLoader.cs
+++ b/Xamarin.Forms.Xaml/XamlLoader.cs
@@ -46,7 +46,7 @@ namespace Xamarin.Forms.Xaml.Internals
 			internal set {
 				xamlFileProvider = value;
 				//¯\_(ツ)_/¯ the previewer forgot to set that bool
-				DoNotThrowOnExceptions = true;
+				DoNotThrowOnExceptions = value != null;
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


Xamarin.Forms.Xaml/XamlLoader.cs
```diff
@@ -39,7 +39,17 @@ namespace Xamarin.Forms.Xaml.Internals
...
 	public static class XamlLoader
 	{
-		public static Func<Type, string> XamlFileProvider { get; internal set; }
+		static Func<Type, string> xamlFileProvider;
+
+		public static Func<Type, string> XamlFileProvider {
+			get { return xamlFileProvider; }
+			internal set {
+				xamlFileProvider = value;
+				//¯\_(ツ)_/¯ the previewer forgot to set that bool
```
alanmcgovern: 😢

StephaneDelcroix:
@jassmith tests are fixed

PR #949, opened 2017-05-30
[iOS,Android] Fix gap when setting Separator None and recycle element and on Android when using grouping

rmarinho:
### Description of Change ###

When using Recycle element we use the ContextActionCell, the problem here is we have 2 cells, 1 with the content and other that acts as a wrapper, and there's a 0.5 diff between them even when we say to the tableview to supress the separator. , seems IOS is somehow thinking ContextActionsCells has a separator.
We fix the problem by stretching the ContentCell to fit.

Also fixes logic for drawing separator on Android when using group headers. 

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=39802

### PR Checklist ###


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla39802.cs
```diff
@@ -0,0 +1,100 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System.Collections.ObjectModel;
+using System.Collections.Generic;
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Bugzilla, 939802, "Gap between ListView cells even if SeparatorVisablity is set to none ", PlatformAffected.iOS)]
```
hartez: Extra '9' :)

rmarinho:
Fixed @hartez

PR #835, opened 2017-03-22
Bz53803 Android app crashes during initialization due to invalid cast…

BradChase2011:
… of toolbar TextView.

### Description of Change ###

Changed the lookup for the TextView in the Toolbar instead of assuming the first element.

### Bugs Fixed ###

- [Bug 53803 - Android crashes upon initialization when using FormsApplicationActivity.](https://bugzilla.xamarin.com/show_bug.cgi?id=53803)

### API Changes ###

None.

### Behavioral Changes ###

FormsApplicationActivity should load without an InvalidCastException.

### PR Checklist ###

- [X] Has tests (if omitted, state reason in description)
- [X] Rebased on top of master at time of PR
- [X] Changes adhere to coding standard
- [X] Consolidate commits as makes sense

PR #631, opened 2016-12-09
[Android] Fix NavigationPage.BarTextColorProperty on API 21+ with FormsApplicationActivity

jimmgarrido:
### Description of Change ###

On Lollipop and above `NavigationPage.BarTextColorProperty` was not working when using `FormsApplicationActivity` because getting a reference to the title `TextView` with the `action_bar_title` resource Id returns null. Instead you have to get a reference to the view indirectly via the Toolbar.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=49384

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #961, opened 2017-06-02
Fix possible crash on API 21+ at launch when using Holo theme and FormsApplicationActivity

jimmgarrido:
### Description of Change ###

The changes made in #631 and #835 were causing an InvalidCastException when using the Holo theme on Lollipop or higher since that theme does not use a Toolbar.

This fixes it by first checking if there is the older ActionBarTextView on all Android versions and if not, _then_ checking for a Toolbar. This also includes a null check for the Toolbar for situations were it is not present, e.g. using a NoActionBar theme.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=57108

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Platform.cs
+++ b/Xamarin.Forms.Platform.Android/Platform.cs
@@ -852,7 +852,7 @@ namespace Xamarin.Forms.Platform.Android
 				int actionbarId = _context.Resources.GetIdentifier("action_bar", "id", "android");
 				if (actionbarId > 0)
 				{
-					var toolbar = (ViewGroup)((Activity)_context).FindViewById(actionbarId);
+					var toolbar = ((Activity)_context).FindViewById(actionbarId) as ViewGroup;
 					if (toolbar != null)
 					{
 						for (int i = 0; i < toolbar.ChildCount; i++)
--- a/Xamarin.Forms.Platform.Android/Platform.cs
+++ b/Xamarin.Forms.Platform.Android/Platform.cs
@@ -857,7 +857,8 @@ namespace Xamarin.Forms.Platform.Android
 					{
 						for (int i = 0; i < toolbar.ChildCount; i++)
 						{
-							if (toolbar.GetChildAt(i) is TextView textView)
+							var textView = toolbar.GetChildAt(i) as TextView;
+							if (textView != null)
 							{
 								actionBarTitleTextView = textView;
 								break;
--- a/Xamarin.Forms.Platform.Android/Platform.cs
+++ b/Xamarin.Forms.Platform.Android/Platform.cs
@@ -847,24 +847,27 @@ namespace Xamarin.Forms.Platform.Android
 			Color navigationBarTextColor = CurrentNavigationPage == null ? Color.Default : CurrentNavigationPage.BarTextColor;
 			TextView actionBarTitleTextView = null;
 
-			if(Forms.IsLollipopOrNewer)
+			if (Forms.IsLollipopOrNewer)
 			{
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.Android/Platform.cs
```diff
@@ -847,29 +847,30 @@ void SetActionBarTextColor()
 			Color navigationBarTextColor = CurrentNavigationPage == null ? Color.Default : CurrentNavigationPage.BarTextColor;
 			TextView actionBarTitleTextView = null;
 
-			if(Forms.IsLollipopOrNewer)
+			int actionBarTitleId = _context.Resources.GetIdentifier("action_bar_title", "id", "android");
```
jassmith: the only thing I don't like about this change is it changes the preference order of action_bar_title/toolbar usage to be inverse what it was before. It would be good to retain the preference ordering just in case there are any oddball scenarios out there.

jimmgarrido:
@jassmith Re-did the fix so it maintains the original preference order 👍

Xamarin.Forms.Platform.Android/Platform.cs
```diff
@@ -850,21 +850,24 @@ void SetActionBarTextColor()
...
+								actionBarTitleTextView = (TextView)toolbar.GetChildAt(i);
+								break;
+							}
 						}
 					}
 				}
 			}
-			else
+			
+			if(actionBarTitleTextView == null)
```
Therzok: Nit: `if (`

Xamarin.Forms.Platform.Android/Platform.cs
```diff
@@ -850,21 +850,24 @@ void SetActionBarTextColor()
...
-					for( int i = 0; i < toolbar.ChildCount; i++ )
+					var toolbar = (ViewGroup)((Activity)_context).FindViewById(actionbarId);
+					if (toolbar != null)
 					{
-						if( toolbar.GetChildAt(i) is TextView )
+						for (int i = 0; i < toolbar.ChildCount; i++)
 						{
-							actionBarTitleTextView = (TextView)toolbar.GetChildAt(i);
-							break;
+							if (toolbar.GetChildAt(i) is TextView)
```
Therzok: `GetChildAt(i)` is being called twice. Maybe use c#7:
`GetChildAt(i) is TextView textView` and use that in the inner block.

rmarinho: yeah let's fix this and do only 1 getChild

Xamarin.Forms.Platform.Android/Platform.cs
```diff
@@ -847,24 +847,27 @@ void SetActionBarTextColor()
...
+			if (Forms.IsLollipopOrNewer)
 			{
 				int actionbarId = _context.Resources.GetIdentifier("action_bar", "id", "android");
-				if(actionbarId > 0)
+				if (actionbarId > 0)
 				{
-					Toolbar toolbar = (Toolbar)((Activity)_context).FindViewById(actionbarId);
-					
-					for( int i = 0; i < toolbar.ChildCount; i++ )
+					var toolbar = (ViewGroup)((Activity)_context).FindViewById(actionbarId);
```
rmarinho: as ViewGroup instead of explicit cast here please..

rmarinho:
this is failing @jimmgarrido

jimmgarrido:
@rmarinho done 👍

PR #967, opened 2017-06-05
bug 40161; force layout of centered image in abs layout

kingces95:
### Description of Change ###

Expected swapping an image source of an image in an absolute layout with h/v options of center to render the swapped in image to it's native size. Actually observed swapped in image within an absolute layout rendering to the size of the original image. This happened because of an optimization in the layout engine which assumed re-computation of the size was unnecessary. The optimization was trying to take advantage of the fact that an absolute layout can sometimes dictate the size of it's children. The optimization was treating a child with h/v options _centered_ as one of those cases but it's not. With h/v options centered the child determines its own size in the case where more space was allocated to the image than the native image size. When more space is allocated the image does not fill that space and is the native size of the image. This change disables the optimization for h/v centered children causing their size to be recomputed. 

At the same time, the change enables the optimization for h/v _filled_ children. Filled children occupy the entire area allocated by the absolute layout; the absolute layout controls the the child size. In that case, the child can inherit the optimization status of the parent layout; if the parent size has not changed then the child size need not be recomputed during the next layout pass because the child size is the parent size.  

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=40161

### API Changes ###

### Behavioral Changes ###

None that are breaking.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla40161.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla40161.cs
@@ -1,0 +1,140 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Bugzilla, 40161, "Issue Description", PlatformAffected.Default)]
+	public class Bugzilla40161 : TestContentPage
+	{
+		// If an image is swapped out for another image must the size of an image be recomputed? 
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla40161.cs
```diff
@@ -0,0 +1,140 @@
...
+		{
+			RunningApp.Screenshot ("I am at Issue 40161");
+			RunningApp.WaitForElement (q => q.Marked ("REFRESH"));
+			RunningApp.Screenshot ("I see the first image");
+
+			RunningApp.Tap ("SWAP");
+			RunningApp.Tap ("REFRESH");
+			RunningApp.WaitForElement(q => q.Marked("step=0"));
+			RunningApp.Screenshot ("I swap the image");
+			RunningApp.WaitForElement(q => q.Marked("w=50"));
```
kingces95: The seth image width is not 50. So if it changed to 50 then the size has been recomputed.

PR #975, opened 2017-06-07
[Android]Dispose check on FastButtonRenderer to prevent crash

bentmar:
Check if buttonrenderer is disposed when making changes

### Description of Change ###

Check if disposed on elementChanged

### Bugs Fixed ###

Prevent crashes when making changes to a disposed renderer
"System.ObjectDisposedException: Cannot access a disposed object."

### API Changes ###

None

### Behavioral Changes ###

None


bentmar:
Ive made common dispose checks to other fastrenderers. It feels like this is the "easy" way to fix this, when a renderer gets disposed how come it still receives elementChanged calls?

PR #980, opened 2017-06-08
[UWP] Bug 56843; LayoutUpdated > SizeChanged; Layout Cycle

kingces95:
### Description of Change ###

Customer expected to be able to embed native TextBlock into AbsolutLayout but that actually generated a layout cycle. Looks like OnElementChanged hooked up Contol.LayoutUpdated so the native control could lay itself out when something changed. Turns out that event gets call just after the layout cycle is complete and doesn't appear to be documented as a place where actual layout should be preformed. Instead, we can use Control.SizeChanged which is only fired if the size of the control changes which should trigger a new layout of the view. See: 

https://blogs.msdn.microsoft.com/devdave/2008/05/27/layout-events-sizechanged-and-layoutupdated/

FYI, @hartez I just kinda made some of this stuff as my best guess as to what is going on. I added a test next to your other tests but that doesn't trigger the issue. So there is something about the wrapping of the test harness that is preventing it from reproducing and which is also prob why the bug slipped through. I attached a non XAML reproduction to the bug. 

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=56843

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.WinRT/NativeViewWrapperRenderer.cs
+++ b/Xamarin.Forms.Platform.WinRT/NativeViewWrapperRenderer.cs
@@ -68,7 +68,7 @@ namespace Xamarin.Forms.Platform.WinRT
 			if (e.OldElement == null)
 			{
 				SetNativeControl(Element.NativeElement);
-				Control.LayoutUpdated += (sender, args) => { ((IVisualElementController)Element)?.InvalidateMeasure(InvalidationTrigger.MeasureChanged); };
+				Control.SizeChanged += (sender, args) => { ((IVisualElementController)Element)?.InvalidateMeasure(InvalidationTrigger.MeasureChanged); };
 			}
 		}
 	}


kingces95:
Here is the meat of the reproduction:

``` C#
public MainPage() 
{
    var stackLayout = new StackLayout()
    {
        VerticalOptions = LayoutOptions.Start,
        HorizontalOptions = LayoutOptions.Center
    };

    var label = new Label
    {
        Text = "Hello World!",
        VerticalOptions = LayoutOptions.Center,
        HorizontalOptions = LayoutOptions.Center
    };

    var textBlock = new TextBlock
    {
        Text = "Text Block"
    };

    stackLayout.Children.Add(label);
    stackLayout.Children.Add(textBlock);

    Content = stackLayout;
}

File: Xamarin.Forms.ControlGallery.WindowsUniversal/MainPage.xaml.cs [hunk 1]
```diff
@@ -127,6 +129,30 @@ void AddNativeControls(NestedNativeControlGalleryPage page)
...
+				};
+
+				var label = new Label
+				{
+					Text = "Hello World!",
+					VerticalOptions = LayoutOptions.Center,
+					HorizontalOptions = LayoutOptions.Center
+				};
+
+				var bz56843TextBlock = new TextBlock
```
hartez: Since it doesn't successfully repro the problem, can we just remove the changes to this test? No point in making it more confusing than it already is.

kingces95: How about if I add a comment that we couldn't figure out why this didn't repro but for expediency sake we're moving on and just we're leaving it in for documentation?

hartez: But it doesn't document anything. It's just some random layout that makes the Windows version of this gallery page different from the other platforms. 

If you want to move it to another Page and document that "this is the (theoretical) repro for Bugzilla case 56843", I've got no objection to that; I just don't see the point of cluttering up the example page for this feature. 

I also wouldn't object to just removing it entirely, since we can't figure out how to make this bug reproduce in Control Gallery.

kingces95: Yeah, would have moved to a separate page but there was no precedence for that so... couldn't just copy paste. So I'll pull the "repro". It's not like we can't just look up the bug..

PR #983, opened 2017-06-09
[C] do not reapply a Binding with a Source when the Context changes

StephaneDelcroix:
### Description of Change ###

[C] do not reapply a Binding with a Source:
- when the Context changes,
- the view is parented
- the inherited context changes.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=57081

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Binding.cs
+++ b/Xamarin.Forms.Core/Binding.cs
@@ -20,6 +20,7 @@ namespace Xamarin.Forms
 		string _path;
 		object _source;
 		string _updateSourceEventName;
+		bool _hasSourceApplied;
 
 		public Binding()
 		{
@@ -81,6 +82,7 @@ namespace Xamarin.Forms
 			{
 				ThrowIfApplied();
 				_source = value;
+				_hasSourceApplied = false;
 			}
 		}
 
@@ -119,11 +121,17 @@ namespace Xamarin.Forms
 			object src = _source;
 			base.Apply(src ?? newContext, bindObj, targetProperty);
 
+			if (src != null && _hasSourceApplied)
+				return;
+			
 			object bindingContext = src ?? Context ?? newContext;
 			if (_expression == null && bindingContext != null)
 				_expression = new BindingExpression(this, SelfPath);
 
 			_expression.Apply(bindingContext, bindObj, targetProperty);
+
+			if (src != null)
+				_hasSourceApplied = true;
 		}
 
 		internal override BindingBase Clone()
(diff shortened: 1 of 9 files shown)


StephaneDelcroix:
@jassmith please review the alternate fix. the original one was blocking reapplying from all changes (propertychanged, ...)

StephaneDelcroix:
> @jimmgarrido: @stephane @davidortinau the updated #983 does not resolve https://bugzilla.xamarin.com/show_bug.cgi?id=57081. The original fix that had failing tests did resolve it however.

Szymaniuk:
In which Xamarin Forms release this COULD be released?
Thanks.

StephaneDelcroix:
@jimmgarrido was right, this doesn't fix the problem entirely

Xamarin.Forms.Core/Binding.cs
```diff
@@ -114,10 +114,15 @@ internal override void Apply(bool fromTarget)
...
-		internal override void Apply(object newContext, BindableObject bindObj, BindableProperty targetProperty)
+		internal override void Apply(object newContext, BindableObject bindObj, BindableProperty targetProperty, bool fromBindingContextChanged = false)
 		{
 			object src = _source;
-			base.Apply(src ?? newContext, bindObj, targetProperty);
+			var isApplied = IsApplied;
+
+			base.Apply(src ?? newContext, bindObj, targetProperty, fromBindingContextChanged: fromBindingContextChanged);
+
+			if (src != null && isApplied && fromBindingContextChanged)
```
kingces95: If I remove `fromBindingContextChanged` and change this line to simply `if (src != null && isApplied)` then all 80 unit tests in `BindableObjectUnitTests.cs`, including the two new ones, all pass. Let's include a test that fails.

StephaneDelcroix:
@kingces95 I can't write a test that fails without the `&& fromBindingContextChanged`, so the check might not be required. I still choose to leave it in as it shows the intent.

mariusbloemhof:
Is anyone aware of the fact that this issue has resurfaced from version 3.1?

It was working in version 2.5.1.52 and 3.0. I upgraded to 3.1 and we are seeing duplicate view model creation using navigation service

StephaneDelcroix:
@mariusbloemhof the unit test we added with this PR are still passing. Could you please open a new issue with a simple working reproduction so we can investigate ?

Thanks a lot

PR #984, opened 2017-06-09
Make Grid.AddHz/Vt span ortho dim

kingces95:
### Description of Change ###

User expected Grid.AddHorizontal to add a column which spans all rows. What actually happens is that the column spans all the rows except the last one _unless_ there are no rows in which case it creates a new row and spans that single row. 

The fix requires tweaking Grid.AddHorizontal. Grid.AddHorizontal calls Add to create the column. Add takes a view, left/right, and top/bottom. For simplicity consider just the arguments to create the columns: left/right. Starting with a grid with no columns calling add with 0,1 will create one column. First, note that the first index doesn't exist. When the first index doesn't exist that means create a new dimension. Second, note that the second index is one past the index to be created. Now consider the top/bottom values specifying the rows to span. To span all rows we'd need to pass 0, RowCount. However if no rows exist then 0 will create the row and we need to bump RowCount to 1. The same logic applies for Grid.AddVertical just transformed for that dimension. 

Now, consider our existing Grid.AddHorizontal code:

`Add(view, lastCol + 1, lastCol + 2, 0, Math.Max(1, lastRow));`

Notice that the last parameter, bottom, will be `lastRow` when there is more than a single row. This would be correct if `lastRow` were the number of rows however it's not -- it's the last valid row index. As such it actually excludes the last valid index when there is more than a single row; The existing logic works when there are no rows or when there is one row. 

I tested that the above logic works for the simple case in UWP. I haven't considered the merged row/column cases or tested the other platforms because the first case is breaking; we have to decide if we're take the break in the simple case before considering other cases. 

Grid.AddHorizontal calls no private or internal members so is essentially a helper method. So, as a work around, the user could copy/paste the fix here into their own extension method to build a method with the expected behavior.

Deprecation is also an option.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=42620

### API Changes ###

None added.

### Behavioral Changes ###

*Breaking Change*; Changes Grid.AddHorizontal and Grid.AddVertical.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/GridTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/GridTests.cs
@@ -39,6 +39,164 @@ namespace Xamarin.Forms.Core.UnitTests
 			Assert.Throws<ArgumentNullException> (() => layout.Children.Remove (null));
 		}
 
+		[TestFixture]
+		public class AddDimension : GridTests
+		{
+			[Datapoints]
+			public static IEnumerable<string> Operations = new[]
+			{
+				"HHH",
+				"HHV",
+				"HVH",
+				"HVV",
+				"VHH",
+				"VHV",
+				"VVH",
+				"VVV",
+
+				"RCRHVHVHVHVHV",
+
+				"HHHV",
+				"VVVH",
+
+				"RV",
+				"RH",
+				"CV",
+				"CH",
+
+				"RVRRV",
+				"CHCCH",
+			};
+
+			Grid _grid;
+
+			int _id = 0;
+			int _rowDef = 0;
+			int _colDef = 0;
+			int _totalWidth = 0;
+			int _totalHeight = 0;
+
+			public void AddHoizontal()
+			{
+				// new block gets new id
+				var id = _id++;
+
+				// adding column only increases height if no rows exist
+				if (_totalHeight == 0)
+					_totalHeight = 1;
+
+				// adding column always increased width by 1
+				_totalWidth++;
+
+				// column spans rows 0 to the last row
+				var row = 0;
+				var height = _totalHeight;
+
+				// column is always added at the end with a width of 1
+				var column = _totalWidth - 1;
+				var width = 1;
+
+				_grid.Children.AddHorizontal(
+					new Label()
+					{
+						Text = $"{id}: {column}x{row} {width}x{height}"
+					}
+				);
+			}
+			public void AddVertical()
+			{
+				// new block gets new id
+				var id = _id++;
+
+				// adding row only increases width if no columns exist
+				if (_totalWidth == 0)
+					_totalWidth = 1;
+
+				// adding row always increased height by 1
+				_totalHeight++;
+
+				// row spans columns 0 to the last column
+				var column = 0;
+				var width = _totalWidth;
+
+				// row is always added at the end with a height of 1
+				var row = _totalHeight - 1;
+				var height = 1;
+
+				_grid.Children.AddVertical(
+					new Label()
+					{
+						Text = $"{id}: {column}x{row} {width}x{height}"
+					}
+				);
+			}
+			public void AddRowDef()
+			{
+				_rowDef++;
+				_totalHeight = Math.Max(_rowDef, _totalHeight);
+
+				_grid.RowDefinitions.Add(new RowDefinition());
+			}
+			public void AddColumnDef()
+			{
+				_colDef++;
+				_totalWidth = Math.Max(_colDef, _totalWidth);
+
+				_grid.ColumnDefinitions.Add(new ColumnDefinition());
+			}
+
+			[TearDown]
+			public override void TearDown()
+			{
+				_grid = null;
+
+				_id = 0;
+				_rowDef = 0;
+				_colDef = 0;
+				_totalWidth = 0;
+				_totalHeight = 0;
+			}
+
+			[Theory]
+			public void AddDimensionTheory(string operations)
+			{
+				_grid = new Grid();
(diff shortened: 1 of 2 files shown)


Xamarin.Forms.Core/Grid.cs
```diff
@@ -353,13 +355,25 @@ public void AddVertical(View view)
...
+
+				// if no columns, create a column
+				if (columns == 0)
+					columns++;
 
-				Add(view, 0, Math.Max(1, lastCol), lastRow + 1, lastRow + 2);
+				Add(view, 0, columns, rows, rows + 1);
 			}
+
+			private int RowCount() => Math.Max(
```
StephaneDelcroix: AggressiveInline

kingces95: This isn't getting called during the layout pass; This is really just a helper routine.

kingces95:
ezgif com-optimize 1

Xamarin.Forms.Core/Grid.cs
StephaneDelcroix: right

StephaneDelcroix:
I don't get how the UITest is testing that the layout is correct, I'm probably missing something obvious...

kingces95:
The bug author expected a call to AddHorizontal on a grid with three rows to add a column that spans three rows. This test covers that case plus the permutations by asserting that each block is placed in the correct column/row and has the correct columnSpan/rowSpan. E.g. (`4: 1x0 1x3` asserts/reports block four should be the second column and have a height of three.)

kingces95:
Ah, thanks for the unit test tip. Though, I needed the UITest so I could see what the logic was generating. Now that it clear what it's supposed to be doing I'll expand the existing unit tests with the missing ones.

kingces95:
In general, the testing and probably user comprehension and certainly the implementation is complicated by the fact that there is not a column/row definition for each column and row. For example, if AddHorizontal is called on an empty grid then a 1x1 grid is created but _no definition is added_. Had we added a definition on implicit creation then all the logic would be a lot simpler. So simple that we probably wouldn't have needed these helper methods. 

What do you think about changing the logic to adding column/row definitions when a row or column is implicitly created? Too breaking?

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla42620.cs
```diff
@@ -0,0 +1,339 @@
...
+				AutomationId = $"{s_id++}",
+				TextColor = Color.Black,
+				BackgroundColor = color,
+				VerticalOptions = LayoutOptions.Center
+			};
+			layout.Children.Add(label, Rectangle.FromLTRB(0, 0, 1, 1), AbsoluteLayoutFlags.All);
+			var tgr = new TapGestureRecognizer();
+			tgr.Tapped += (x, o) =>
+			{
+				Grid.SetColumnSpan(layout, Grid.GetColumnSpan(layout) + 1);
```
kingces95: This is just for my own edification; The UITest doesn't tap on it. Just forgot to remove it.

Xamarin.Forms.Core.UnitTests/GridTests.cs
```diff
@@ -39,6 +39,164 @@ public void ThrowsOnNullRemove ()
...
+
+			Grid _grid;
+
+			int _id = 0;
+			int _rowDef = 0;
+			int _colDef = 0;
+			int _totalWidth = 0;
+			int _totalHeight = 0;
+
+			public void AddHoizontal()
```
StephaneDelcroix: Horizontal. not important

Xamarin.Forms.Core.UnitTests/GridTests.cs
```diff
@@ -39,6 +39,164 @@ public void ThrowsOnNullRemove ()
...
+
+				_id = 0;
+				_rowDef = 0;
+				_colDef = 0;
+				_totalWidth = 0;
+				_totalHeight = 0;
+			}
+
+			[Theory]
+			public void AddDimensionTheory(string operations)
```
StephaneDelcroix: multiple [TestCase] would make more sense than a [Theory]. I don't see this as a Theory, but a Test. Probably just me. Anyway, not important.

Xamarin.Forms.Core.UnitTests/GridTests.cs
kingces95: If we're testing AddVertical then we we should test AddHorizontal. The tests are the same just transposed. More testing is a good thing. Helps prevent bugs like this one. :)

Xamarin.Forms.Core.UnitTests/GridTests.cs
kingces95: Using a theory makes sense. The test case is parameterized; There is a single test that uses different data which is what a theory is designed to support.

Xamarin.Forms.Core.UnitTests/GridTests.cs
StephaneDelcroix: I meant the typo...

kingces95: Ah! Was wondering why more testing was a bad thing. :)

rmarinho:
@kingces95 1 failing UITest on IOS

kingces95:
Removed UITest but retain unit tests to unblock check-in due to failing UITest. Will resubmit/debug UITest after the fix is in.

PR #987, opened 2017-06-12
[iOS] workaround a weird iOS behavior.

StephaneDelcroix:
### Description of Change ###

Work around the fact the UITextField always returns 30 as fitting size for RoundedRect BorderStyle.
Also make sure the layout is invalidated when the font is changed.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=44774

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###


pirras-torres:
the overrided GetDesiredSize causes the height of the entries to be 15

PR #988, opened 2017-06-12
[WinRT/UWP] Add delay for layout in OnScrollToRequested

pauldipietro:
### Description of Change ###

When items are being added into a ScrollView on WinRT/UWP, it can cause ScrollToAsync to fail, particularly when scrolling to the end, as the items are added but have not yet been laid out, leading to incorrect scroll values in OnScrollToRequested. It appears as if the Android implementation of the ScrollViewRenderer does something [similar](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/Renderers/ScrollViewRenderer.cs#L284-L292) by running a small delay of up to 10ms; a similar usage here appears to alleviate the problem while not affecting the previous behavior (based on the use of the control gallery's ScrollView page).

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=36651

### API Changes ###

Adjust OnScrollToRequested to use async

### Behavioral Changes ###

A very small 10ms addition which should otherwise not be noticeable by the user.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla36651.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla36651.cs
@@ -1,0 +1,78 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System;
+
+#if UITEST
+using Xamarin.UITest;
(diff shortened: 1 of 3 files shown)


samhouts:
@pauldipietro Wanna check the build on this one?

johnshardman:
Could the IsInNativeLayout property be made for general use? At the moment, it's marked as for internal use only.

PR #993, opened 2017-06-14
[macOS] ScrollView's content are showing at End.

BradChase2011:
### Description of Change ###

Changed ScrollViewRenderer on macOS to align for Start,Center,End on contents that are smaller than the ScrollView panel.  It also fixes an issue where the coordinates being sent into the Forms control from the macOS control on ScrollY were inverted.

Tests omitted.
### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=56947

### API Changes ###
None

### Behavioral Changes ###

ScrollView's should now render at top left rather than bottom left, they should behave vertically for Start, Center, and End.  The correct Y coord should be correct for the Xamarin control.  The only remaining issue is when a window is resized vertically, the control doesnt resize correctly, when the window is resize horizontally it resets to the correct position.  I think this is outside of that control and maybe related to some other events not getting correctly passed down.  For instance some Page controls dont resize to the window at all when sized.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.MacOS/Renderers/ScrollViewRenderer.cs
+++ b/Xamarin.Forms.Platform.MacOS/Renderers/ScrollViewRenderer.cs
@@ -1,4 +1,4 @@
-﻿using System;
+﻿﻿using System;
 using System.ComponentModel;
 using AppKit;
 using RectangleF = CoreGraphics.CGRect;
@@ -19,10 +19,10 @@ namespace Xamarin.Forms.Platform.MacOS
 
 		public ScrollViewRenderer() : base(RectangleF.Empty)
 		{
+            ContentView = new FlippedClipView();
 			DrawsBackground = false;
-			ContentView.PostsBoundsChangedNotifications = true;
-			NSNotificationCenter.DefaultCenter.AddObserver(this, new Selector(nameof(UpdateScrollPosition)),
-				BoundsChangedNotification, ContentView);
+            ContentView.PostsBoundsChangedNotifications = false;
+			NSNotificationCenter.DefaultCenter.AddObserver(this, new Selector(nameof(UpdateScrollPosition)), BoundsChangedNotification, ContentView);
 			HasVerticalScroller = true;
 		}
 
@@ -128,8 +128,10 @@ namespace Xamarin.Forms.Platform.MacOS
 				Platform.SetRenderer(content, Platform.CreateRenderer(content));
 
 			_contentRenderer = Platform.GetRenderer(content);
-
+            (ContentView as FlippedClipView).ContentRenderer = _contentRenderer;
 			DocumentView = _contentRenderer.NativeView;
+
+            ContentView.PostsBoundsChangedNotifications = true;
 		}
 
 		void LayoutSubviews()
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.MacOS/Renderers/ScrollViewRenderer.cs
```diff
@@ -200,27 +191,77 @@ void UpdateBackgroundColor()
...
 		{
+            if ((ContentView as FlippedClipView) != null)
+                (ContentView as FlippedClipView).ContentRenderer = null;
+            
 			_contentRenderer?.NativeView?.RemoveFromSuperview();
 			_contentRenderer?.Dispose();
 			_contentRenderer = null;
 		}
+
+		private sealed class FlippedClipView : NSClipView
```
rmarinho: Lets have this in it's own file please mark internal

Xamarin.Forms.Platform.MacOS/Renderers/ScrollViewRenderer.cs
```diff
@@ -200,27 +191,77 @@ void UpdateBackgroundColor()
...
+            if (ScrollView == null)
+                return;
+            
+            if (ScrollView.ContentSize.Height >= ScrollView.Height)
+            {
+                RectangleF visibleRect = ContentView.DocumentVisibleRect();
+                if (visibleRect.Location.Y > -1)
+                    ScrollView.SetScrolledPosition(Math.Max(0, visibleRect.Location.X), Math.Max(0, ContentView.Frame.Height - visibleRect.Location.Y));
+            }
+            else
```
rmarinho: need style fixes here

Xamarin.Forms.Platform.MacOS/Renderers/ScrollViewRenderer.cs
BradChase2011: Will do.  I was hesitant to move it into its own file because it kinda has a singular purpose and was minimal code.  The only time I could see reusing it would be if something changes at the ground level of the layout system and we flipped the entire macos nsviews to more closely resemble forms.  I should have changes up in the AM.

BradChase2011:
@rmarinho Changes are up.

rmarinho:
Can we fix code style?

BradChase2011:
Ya gotta help me out on that.  What code style is off?  I keep looking at it and can't see it.

Xamarin.Forms.Platform.MacOS/Renderers/ScrollViewRenderer.cs
```diff
@@ -19,10 +19,10 @@ public class ScrollViewRenderer : NSScrollView, IVisualElementRenderer
 
 		public ScrollViewRenderer() : base(RectangleF.Empty)
 		{
+            ContentView = new FlippedClipView();
 			DrawsBackground = false;
-			ContentView.PostsBoundsChangedNotifications = true;
-			NSNotificationCenter.DefaultCenter.AddObserver(this, new Selector(nameof(UpdateScrollPosition)),
-				BoundsChangedNotification, ContentView);
+            ContentView.PostsBoundsChangedNotifications = false;
+			NSNotificationCenter.DefaultCenter.AddObserver(this, new Selector(nameof(UpdateScrollPosition)), BoundsChangedNotification, ContentView);
```
BradChase2011: That is so weird.  When viewed in my VS2015 instance it shows the correct number if tabs.  I'll remove them and tab them in.  Should be up soon.  Thanks.

BradChase2011: ok checked in.  I am beginning to think my merge tool is somehow adding spaces in.  I am going to reset all my settings and see if there is an option for tabs.  Can I ask what merge tool you guys use and are there specific settings?  I unfortunately am using 4 different computers with different VS instances from 2015, 2017, 2017 Preview, and VS macOS.  I need a different version at the current moment for each store deploy I make, for our in house code to xamarin's code.  Its getting difficult to track down which instance causes the merge issues as I now have merge tools for windows and a different one for macOS.  Its a mess!  I would love to have one singular approach or at least eliminate the problem child.

PR #454, opened 2016-10-13
Disable using EstimatedRowHeight for uneven rows with known heights

hartez:
### Description of Change

Modifies the ListViewRenderer and UnevenListViewDataSource to return an estimated row height of zero when the cell heights are known. This prevents iOS from attempting to use row height estimates unnecessarily, which causes scrolling issues (especially if the specified cell heights are very different from the minimum measured cell heights). 

There are tests for both estimated row height and known row height, but they are not automated (nothing in the UI test platform for detecting this sort of thing).
### Bugs Fixed
- [43313 – Adding an item to ListView ItemSource has unexpected animation with different height rows and HasUnevenRows is true](https://bugzilla.xamarin.com/show_bug.cgi?id=43313)
- [44830 – ListView not consistently focusing at the end of the list.](https://bugzilla.xamarin.com/show_bug.cgi?id=44830)

### API Changes
None

### Behavioral Changes
None

### PR Checklist
- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #994, opened 2017-06-15
[iOS] ListView with UnevenRows and Cell Heights will no longer be slow to load

samhouts:
### Description of Change ###

#454 turned off row height estimation for `ListView`s enabling `HasUnevenRows` when the first `ViewCell` had a non-zero `Height`. This requires every row to be measured on initial load to prepare the scroll area, so large item sources developed a significant load time.

Re-enabling estimation is ideal for performance; however, it will regress the issue resolved by #454. As a compromise, this change will cache known `ViewCell` heights to be used by `EstimateHeight`, which will be called once when the `ListView` loads and allow the scroll area to be accurately estimated (preserving the resolved status of Bugzilla 43313) while still providing a value for estimation quick enough to restore performance. 

Users may still see that the `ViewCell` constructors will be called more times than they were before #454, since we still need to query each `ViewCell` for its `Height`. However, this should be a value closer to 1x instead of 3-4x. 

If users are **not** concerned about adding rows and scrolling to them with smooth animation, users may now also specify a `RowHeight` on the `ListView`. This value will be used for initial estimation, and the actual measured heights of the rows will be used when they are displayed. This, in effect, returns the behavior back to pre-#454, reducing the constructor calls to just those necessary for initial display. 

### Bugs Fixed ###

- [Bug 56896 - ListViews for lists with many elements regressed in performance on iOS](https://bugzilla.xamarin.com/show_bug.cgi?id=56896)

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
@@ -373,7 +373,7 @@ namespace Xamarin.Forms.Platform.iOS
 
 				// We want to make sure we reset the cached defined row heights whenever this is called.
 				// Failing to do this will regress Bugzilla 43313 (strange animation when adding rows with uneven heights)
-				source.CacheDefinedRowHeights();
+				source?.CacheDefinedRowHeights();
 
 				if (_shouldEstimateRowHeight && !_estimatedRowHeight)
 				{
--- a/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
@@ -1,9 +1,11 @@
 ﻿using System;
 using System.Collections;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.ComponentModel;
 using System.Linq;
+using System.Threading.Tasks;
 using Foundation;
 using UIKit;
 using Xamarin.Forms.Internals;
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -364,14 +366,16 @@ void OnScrollToRequested(object sender, ScrollToRequestedEventArgs e)
...
-
 			var rowHeight = Element.RowHeight;
 			if (Element.HasUnevenRows && rowHeight == -1)
 			{
 				var source = _dataSource as UnevenListViewDataSource;
-				if (_shouldEstimateRowHeight)
+
+				// We want to make sure we reset the cached defined row heights whenever this is called.
+				// Failing to do this will regress Bugzilla 43313 (strange animation when adding rows with uneven heights)
+				source.CacheDefinedRowHeights();
```
rmarinho: we are using source without checking if it's null.. not sure `?. `will be more correct

samhouts: Good call! Fixed.

fschwiet:
Will there be an update to the documentation?  https://developer.xamarin.com/api/property/Xamarin.Forms.ListView.HasUnevenRows/#Remarks states:

```
When the app developer sets the ListView.HasUnevenRows property to true, the behavior of the list view still depends on the ListView.RowHeight property. First, if the developer either does not set the ListView.RowHeight property or sets it to -1, list view items are autosized to fit their contents. This is the desired behavior and the intended use case for a ListView.HasUnevenRows value of true, as noted above. Second, if the developer sets the ListView.RowHeight property to 0 or to a negative value other than -1, then all rows in the ListView will, irrespective of the height of their content, have the default height for the system. Third, and finally, if the developer sets ListView.RowHeight to a positive value, then all rows in the ListView will, irrespective of the height of their content, be as tall as ListView.RowHeight, as if ListView.HasUnevenRows had been set to false. 
```

PR #998, opened 2017-06-16
[C] reset the BindingContext on template change

StephaneDelcroix:
### Description of Change ###

When we remove an old template before setting a new one, all Bindings are unapplied (as part of unparenting).
This fix force a setting back the binding context when the an existing ControlTemplate is swapped for another one.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=43685

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/TemplatedViewUnitTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/TemplatedViewUnitTests.cs
@@ -27,5 +27,31 @@ namespace Xamarin.Forms.Core.UnitTests
             {
             }
         }
+
+		public class MyTemplate : StackLayout
+		{
+			public MyTemplate()
+			{
+				Children.Add(new ContentPresenter());
+			}
+		}
+
+		[Test]
+		public void BindingsShouldBeAppliedOnTemplateChange()
+		{
+			var template0 = new ControlTemplate(typeof(MyTemplate));
+			var template1 = new ControlTemplate(typeof(MyTemplate));
+			var label = new Label();
+			label.SetBinding(Label.TextProperty, ".");
+			var cv = new ContentView {
+				ControlTemplate = template0,
+				Content = label
+			};
+			cv.BindingContext = "Foo";
+
+			Assume.That(label.Text, Is.EqualTo("Foo"));
+			cv.ControlTemplate = template1;
+			Assert.That(label.Text, Is.EqualTo("Foo"));
+		}
 	}
 }
--- a/Xamarin.Forms.Core/ContentPage.cs
+++ b/Xamarin.Forms.Core/ContentPage.cs
@@ -22,5 +22,19 @@ namespace Xamarin.Forms
 				SetInheritedBindingContext(content, BindingContext);
 			}
 		}
+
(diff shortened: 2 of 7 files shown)


samhouts:
Unit tests aren't happy.

Xamarin.Forms.Core.UnitTests/TemplatedViewUnitTests.cs
```diff
@@ -27,5 +27,31 @@ public ExpectedView()
             {
             }
         }
+
+		public class MyTemplate : StackLayout
```
kingces95: Huh. This formatting is correct but the rest of the file uses spaces..

Xamarin.Forms.Core.UnitTests/TemplatedViewUnitTests.cs
```diff
@@ -27,5 +27,31 @@ public ExpectedView()
...
+			var template1 = new ControlTemplate(typeof(MyTemplate));
+			var label = new Label();
+			label.SetBinding(Label.TextProperty, ".");
+			var cv = new ContentView {
+				ControlTemplate = template0,
+				Content = label
+			};
+			cv.BindingContext = "Foo";
+
+			Assume.That(label.Text, Is.EqualTo("Foo"));
```
kingces95: Cool. So you use an assume to indicate this is not the test but an invariant of the the test itself.

Xamarin.Forms.Core/ContentPage.cs
```diff
@@ -22,5 +22,19 @@ protected override void OnBindingContextChanged()
 				SetInheritedBindingContext(content, BindingContext);
 			}
 		}
+
+		internal override void OnControlTemplateChanged(ControlTemplate oldValue, ControlTemplate newValue)
+		{
+			if (oldValue == null)
+				return;
+
+			base.OnControlTemplateChanged(oldValue, newValue);
```
kingces95: Why not have this before line 28? Is it always the case that the base class will similarly do nothing if the `oldValue` is null?

Xamarin.Forms.Core/ContentPage.cs
```diff
@@ -22,5 +22,19 @@ protected override void OnBindingContextChanged()
...
+			View content = Content;
+			ControlTemplate controlTemplate = ControlTemplate;
+			if (content != null && controlTemplate != null)
```
kingces95: Why the trap for `content == null`? Seems `TemplateUtilities.cs` L122 ensures that's the case. Should we throw instead?

Xamarin.Forms.Core.UnitTests/TemplatedViewUnitTests.cs
StephaneDelcroix: ¯\_(ツ)_/¯

Xamarin.Forms.Core/ContentPage.cs
StephaneDelcroix: safer, in case the base is modified one day

PR #1003, opened 2017-06-21
[XamlC] Allow overriding the assembly resolver and whether to compile

praeclarum:
### Description of Change ###

This adds two public properties to XamlC task to allow some of its hard coded defaults to be overridden.

1. You can now override its assembly resolver to use/reuse your own
2. You can change the default of whether to compile

### API Changes ###

Added:
 - `bool XamlCTask.ForceCompile { get; set; }`
 - `bool XamlCTask.CompileByDefault { get; set; }`
 - `IAssemblyResolver XamlCTask.DefaultAssemblyResolver { get; set; }`

### Behavioral Changes ###

This will cause the task to use the given assembly resolver instead of always creating its own.

This will also allow tools to force compilation.

### PR Checklist ###

    - XamlCTask doesn't have any tests

### DIFF
--- a/Xamarin.Forms.Build.Tasks/XamlCTask.cs
+++ b/Xamarin.Forms.Build.Tasks/XamlCTask.cs
@@ -19,6 +19,10 @@ namespace Xamarin.Forms.Build.Tasks
 		[Obsolete("OutputGeneratedILAsCode is obsolete as of version 2.3.4. This option is no longer available.")]
 		public bool OutputGeneratedILAsCode { get; set; }
 
+		public bool CompileByDefault { get; set; }
+
+		public IAssemblyResolver DefaultAssemblyResolver { get; set; }
+
 		internal string Type { get; set; }
 		internal MethodDefinition InitCompForType { get; private set; }
 		internal bool ReadOnly { get; set; }
@@ -35,7 +39,7 @@ namespace Xamarin.Forms.Build.Tasks
 				Logger.LogLine(1, "ReferencePath: \t{0}", ReferencePath.Replace("//", "/"));
 			Logger.LogLine(3, "DebugSymbols:\"{0}\"", DebugSymbols);
 			Logger.LogLine(3, "DebugType:\"{0}\"", DebugType);
-			var skipassembly = true; //change this to false to enable XamlC by default
+			var skipassembly = !CompileByDefault;
 			bool success = true;
 
 			if (!File.Exists(Assembly))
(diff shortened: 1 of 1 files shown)


praeclarum:
I'm actually torn between what I have here:

```csharp
bool XamlCTask.CompileByDefault { get; set; }
```

which still obeys the attributes. Or,

```csharp
bool XamlCTask.ForceCompilation { get; set; }
```

which would just force compilation.

Which do you prefer?

StephaneDelcroix:
I prefer 'ForceByDefault'. it at least avoids negating the bool later

Xamarin.Forms.Build.Tasks/XamlCTask.cs
```diff
@@ -44,24 +48,29 @@ public override bool Execute(out IList<Exception> thrownExceptions)
 				return true;
 			}
 
-			var resolver = new XamlCAssemblyResolver();
-			if (!string.IsNullOrEmpty(DependencyPaths))
+			var resolver = DefaultAssemblyResolver ?? new XamlCAssemblyResolver();
+			var xamlCResolver = resolver as XamlCAssemblyResolver;
+
+			if (xamlCResolver != null)
```
samhouts: So if `DefaultAssemblyResolver` is not a valid `XamlCAssemblyResolver`, then we don't do any of this work. Should we at least log a warning?

praeclarum: For me it's not a warning as most resolver won't support this. However I do see your point that we are ignoring properties at that point. I'm happy either way.

praeclarum:
This last commit was added per a discussion with @StephaneDelcroix on Slack.

It adds the ability to force compilation no matter what the user's attributes are.

Xamarin.Forms.Build.Tasks/XamlCTask.cs
praeclarum: I added the warning!

Xamarin.Forms.Build.Tasks/XamlCTask.cs
```diff
@@ -44,26 +49,34 @@ public override bool Execute(out IList<Exception> thrownExceptions)
...
+					var paths = ReferencePath.Replace("//", "/").Split(';');
+					foreach (var p in paths)
+					{
+						var searchpath = Path.GetDirectoryName(p);
+						Logger.LogLine(3, "Adding searchpath {0}", searchpath);
+						xamlCResolver.AddSearchDirectory(searchpath);
+					}
 				}
 			}
+			else {
```
StephaneDelcroix: I like how quickly @praeclarum understood the very special rules we have for randomly placing opening braces, and adopted it.

Xamarin.Forms.Build.Tasks/XamlCTask.cs
```diff
@@ -128,7 +141,7 @@ public override bool Execute(out IList<Exception> thrownExceptions)
 						if (Type != null)
 							skiptype = !(Type == classname);
 
-						if (skiptype) {
+						if (skiptype && !ForceCompile) {
 							Logger.LogLine(2, "Has XamlCompilationAttribute set to Skip and not Compile... skipped");
```
StephaneDelcroix: I'd add

```
if (skip type && ForceCompile)
    Logger.LogLine ("ForceCompile flag is set. Overriding preferences set by XamlCompilationAttribute");

```

or something like that.

PR #1004, opened 2017-06-21
Set the Id field for Android Views created by Forms

hartez:
### Description of Change ###

Because Forms doesn't set the `Id` value for Android Views it creates, some methods (most notably `BackStackRecord.configureTransitions`) which search for Views by `Id` and cast them were failing because all of the Views had the default of `NO_ID` (-1). 

This change sets the `Id` field for the various renderers and for the `ModalContainer` to prevent this issue.

### Bugs Fixed ###

- https://forums.xamarin.com/discussion/comment/280624/#Comment_280624
- https://forums.xamarin.com/discussion/81768/launching-a-modal-over-a-page-containing-a-map-maps-ad-m-cannot-be-cast-to-android-view-viewgroup

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls/TestCases.cs
+++ b/Xamarin.Forms.Controls/TestCases.cs
@@ -141,9 +141,10 @@ namespace Xamarin.Forms.Controls
 				var untrackedIssueCells = 
 					from issueModel in issueModels
 					where issueModel.IssueTracker == IssueTracker.None
-					orderby issueModel.Description 
+					orderby issueModel.IssueNumber descending, issueModel.Description 
 					select MakeIssueCell (issueModel.Name, issueModel.Description, issueModel.Action);
 
+				// TODO hartez 2017/06/21 09:12:51 Restore the original order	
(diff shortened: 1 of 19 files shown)


Xamarin.Forms.Controls/TestCases.cs
```diff
@@ -141,9 +141,10 @@ public TestCaseScreen ()
 				var untrackedIssueCells = 
 					from issueModel in issueModels
 					where issueModel.IssueTracker == IssueTracker.None
-					orderby issueModel.Description 
+					orderby issueModel.IssueNumber descending, issueModel.Description 
 					select MakeIssueCell (issueModel.Name, issueModel.Description, issueModel.Action);
 
+				// TODO hartez 2017/06/21 09:12:51 Restore the original order
```
samhouts: Don't forget about this guy!

hartez: Blergh. I did restore the order, then I forgot to remove the comment.

PR #1005, opened 2017-06-21
[UWP] ListView ItemSelected event will fire only once on selection changed

samhouts:
### Description of Change ###

Resolves regressions caused by #713.

Now calling `OnListItemClicked` when the native `ListView`s selection is changed and _not_ attempting to set the `SelectedItem` property of the `ListView` renderer at the same time. This was causing the second `ItemSelected` event to fire.

Calling `OnListItemClicked` when the selection changes ensures a consistent experience when using touch, mouse, keyboard, and Narrator navigation. UWP by default will select an item when it is focused by the keyboard, and using the spacebar/enter key will deselect the item. This PR preserves that default behavior. Deselecting an item will now fire an event, too, since that is a valid `SelectedItem` change. Note that this now allows the item returned by `Tapped` events to be null in these cases.

### Bugs Fixed ###

- [Bug 44886 - UWP Listview ItemSelected event triggered twice for each selection](https://bugzilla.xamarin.com/show_bug.cgi?id=44886)
- [Bug 42188 - ViewCell with focus does not recognise being tapped using keyboard on UWP](https://bugzilla.xamarin.com/show_bug.cgi?id=42188)

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla44886.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla44886.cs
@@ -1,0 +1,89 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.ComponentModel;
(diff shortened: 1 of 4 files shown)


rmarinho:
@samhouts Uitest failures

hartez:
@samhouts Needs rebase.

PR #1006, opened 2017-06-22
[All] Use IReflectableType when accessing the Registrar

praeclarum:
### Description of Change ###

Everywhere `Registrar.Registered.GetHandler` is called, try to use the `IReflectableType` to get type information instead of `GetType`.

This allows dynamic objects that implement their own types to be used by the Registrar.

There is already some precedent for support of dynamic objects: `BindingExpression.ApplyCore` does this same kind of check. This patch extends that support to the Registrar.

### API Changes ###

None

### Behavioral Changes ###

Users will be able to register dynamic objects in the Registrar where before they could not.

### PR Checklist ###

- [x] Has tests (2 new tests)

### DIFF
--- a/Xamarin.Forms.Platform.Android/Cells/BaseCellView.cs
+++ b/Xamarin.Forms.Platform.Android/Cells/BaseCellView.cs
@@ -190,7 +190,7 @@ namespace Xamarin.Forms.Platform.Android
 			Bitmap bitmap = null;
 			IImageSourceHandler handler;
 
-			if (source != null && (handler = Registrar.Registered.GetHandler<IImageSourceHandler>(source.GetType())) != null)
+			if (source != null && (handler = Registrar.Registered.GetHandlerForObject<IImageSourceHandler>(source)) != null)
 			{
 				try
 				{
--- a/Xamarin.Forms.Platform.Android/Cells/ViewCellRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Cells/ViewCellRenderer.cs
@@ -143,7 +143,7 @@ namespace Xamarin.Forms.Platform.Android
 				var renderer = GetChildAt(0) as IVisualElementRenderer;
 				var viewHandlerType = Registrar.Registered.GetHandlerTypeForObject(cell.View) ?? typeof(Platform.DefaultRenderer);
 				var reflectableType = renderer as System.Reflection.IReflectableType;
-				var rendererType = reflectableType != null ? reflectableType.GetTypeInfo().AsType() : (renderer != null ? renderer.GetType() : typeof(obj));
+				var rendererType = reflectableType != null ? reflectableType.GetTypeInfo().AsType() : (renderer != null ? renderer.GetType() : typeof(System.Object));
 				if (renderer != null && rendererType == viewHandlerType)
 				{
 					Performance.Start("Reuse");
--- a/Xamarin.Forms.Platform.Android/Cells/ViewCellRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Cells/ViewCellRenderer.cs
@@ -142,7 +142,9 @@ namespace Xamarin.Forms.Platform.Android
 
 				var renderer = GetChildAt(0) as IVisualElementRenderer;
 				var viewHandlerType = Registrar.Registered.GetHandlerTypeForObject(cell.View) ?? typeof(Platform.DefaultRenderer);
-				if (renderer != null && renderer.GetType() == viewHandlerType)
+				var reflectableType = renderer as System.Reflection.IReflectableType;
+				var rendererType = reflectableType != null ? reflectableType.GetTypeInfo().AsType() : (renderer != null ? renderer.GetType() : typeof(obj));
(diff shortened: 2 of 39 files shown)


Xamarin.Forms.Platform.Android/Cells/BaseCellView.cs
```diff
@@ -190,7 +190,10 @@ public void SetRenderHeight(double height)
 			Bitmap bitmap = null;
 			IImageSourceHandler handler;
 
-			if (source != null && (handler = Registrar.Registered.GetHandler<IImageSourceHandler>(source.GetType())) != null)
+			var rt = source as System.Reflection.IReflectableType;
+			var t = rt != null ? rt.GetTypeInfo().AsType() : source.GetType();
+
+			if (source != null && (handler = Registrar.Registered.GetHandler<IImageSourceHandler>(t)) != null)
```
StephaneDelcroix: the check for `source != null` happens to late. L194 is potentially accessing a `null` instance

what about 
```
var t = (source as IReflectableType) ?.GetTypeInfo().AsType() ?? source?.GetType();
if (t != null && ....)
```

Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
```diff
@@ -31,7 +31,10 @@ internal static class ImageViewExtensions
 			Bitmap bitmap = null;
 			IImageSourceHandler handler;
 
-			if (source != null && (handler = Internals.Registrar.Registered.GetHandler<IImageSourceHandler>(source.GetType())) != null)
+			var rt = source as System.Reflection.IReflectableType;
+			var t = rt != null ? rt.GetTypeInfo().AsType() : source.GetType();
+
+			if (source != null && (handler = Internals.Registrar.Registered.GetHandler<IImageSourceHandler>(t)) != null)
```
StephaneDelcroix: _idem_

praeclarum:
@StephaneDelcroix I agree with you and realized after I did all of this.

I would be happy to kill this PR and start over with a new function on the Registrar that hid all this ugliness.

praeclarum:
@StephaneDelcroix Do you have any suggestions for the name of `GetHandlerXXX`? Right now there's `GetHandler` and `GetHandlerType`. I propose to add these:

* `GetHandlerForObject`
* `GetHandlerTypeForObject`

If you don't like those, how about:

* `GetObjectHandler`
* `GetObjectHandlerType`

Xamarin.Forms.Platform.Android/Cells/BaseCellView.cs
praeclarum: Yeah I definitely screwed up there. I will be more careful on v2.

StephaneDelcroix:
I prefer the first pair. You @jassmith ??

praeclarum:
@StephaneDelcroix I have rewritten the PR to use the new methods. I have also added docs and unit tests for them.

Still happy to rename if @jassmith doesn't like them.

jassmith:
GetHandlerForObject is fine, rest of the PR looks fine, need to get the test ran then we can merge.

StephaneDelcroix:
@praeclarum here's the build failure

```
Compilation error: Xamarin.Forms.Platform.Android\Xamarin.Forms.Platform.Android.csproj 
Cells\ViewCellRenderer.cs(97, 139): error CS0246: The type or namespace name 'obj' could not be found (are you missing a using directive or an assembly reference?) ```

File: Xamarin.Forms.Platform.Android/Cells/ViewCellRenderer.cs [hunk 3]
```diff
@@ -92,8 +92,10 @@ public void Update(ViewCell cell)
 				Performance.Start();
 
 				var renderer = GetChildAt(0) as IVisualElementRenderer;
-				var viewHandlerType = Registrar.Registered.GetHandlerType(cell.View.GetType()) ?? typeof(Platform.DefaultRenderer);
-				if (renderer != null && renderer.GetType() == viewHandlerType)
+				var viewHandlerType = Registrar.Registered.GetHandlerTypeForObject(cell.View) ?? typeof(Platform.DefaultRenderer);
+				var reflectableType = renderer as System.Reflection.IReflectableType;
+				var rendererType = reflectableType != null ? reflectableType.GetTypeInfo().AsType() : (renderer != null ? renderer.GetType() : typeof(obj));
```
StephaneDelcroix: `typeof(obj)` ???

hartez:
@praeclarum This doesn't seem to be compiling.

Xamarin.Forms.Platform.Android/Cells/ViewCellRenderer.cs
praeclarum: Yep not sure how I missed this. Fixing...

praeclarum:
@hartez Sorry for the delay, fixing...

rmarinho:
build

rmarinho:
@praeclarum did you pushed the fix ?

PR #1021, opened 2017-06-28
[Android] Ignore PropertyChanged when Label is disposed

bentmar:
### Description of Change ###

Added a dispose check on elementpropertyChanged on FastLabelRenderer

### Bugs Fixed ###

Cannot acces disposed object, like this one https://forums.xamarin.com/discussion/comment/283251/#Comment_283251

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/FastRenderers/LabelRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/FastRenderers/LabelRenderer.cs
@@ -198,6 +198,9 @@ namespace Xamarin.Forms.Platform.Android.FastRenderers
 
 		protected virtual void OnElementPropertyChanged(object sender, PropertyChangedEventArgs e)
 		{
+			if (_disposed)
+			   return;
+			   
 			ElementPropertyChanged?.Invoke(this, e);
 
 			if (e.PropertyName == Label.HorizontalTextAlignmentProperty.PropertyName || e.PropertyName == Label.VerticalTextAlignmentProperty.PropertyName)


Xamarin.Forms.Platform.Android/FastRenderers/LabelRenderer.cs
```diff
@@ -198,6 +198,9 @@ protected virtual void OnElementChanged(ElementChangedEventArgs<Label> e)
 
 		protected virtual void OnElementPropertyChanged(object sender, PropertyChangedEventArgs e)
 		{
+			if (_disposed)
```
jassmith: How can this method ever git invoked when its been unhooked in the dispose handler? It seems that if this is actually happening the root cause of the bug needs to be addressed.

bentmar: I brought up that question a while back on a similar fix. Saw your post on the forum, nice to see that your looking for a better fix :)

bentmar: Ive seen this happening on image button label and frame (all fastrenderers). I would try creating a repo on the first prerelase containing fastrenderers (it had zero? dispose checks on any of the fast renderers). And it was crashing a lot because of accessing disposed objects

jassmith: I would kill for a repro showing an "imposible dispose" scenario for sure...

bentmar: a never found a solid way to repro it though :/ it was very random, Il see what i can wrap together next week

mattrichnz: On the right track with this today. Manually firing off GC collection is starting to break things. Putting together a repro project this afternoon,

mattrichnz: OK, we've found the following to be effective in creating a repro for these **ObjectDisposedException** problems:

1. Push/pop the navigation stack i.e. Push the demo page onto the stack and then pop it off again.
2. Simultaneously when pushing/popping, manually fire a GC collect to simulate memory pressure and run the GC sweep.
3. Use DataTriggers or bindings on a XAML view with a view-model that uses INotifyPropertyChanged. You need to get these bindings/triggers firing to see the problem.

Doing all three of these things, we have been able to repro ObjectDisposedException crashes  today on `Label` and `ProgressBar`. See our repro project here for an example:
https://bugzilla.xamarin.com/show_bug.cgi?id=57910

hartez:
This is addressed by PR #1033

PR #1024, opened 2017-06-28
[UWP] Fixes for usage of XF with .net native toolchain

rmarinho:
### Description of Change ###

Adding the xaml assembly so reflection works.
Add reference on rd.xml to use Windows.UI.Xaml.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=56795

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/Properties/Xamarin.Forms.Platform.UAP.rd.xml
+++ b/Xamarin.Forms.Platform.UAP/Properties/Xamarin.Forms.Platform.UAP.rd.xml
@@ -28,6 +28,7 @@
   <Library Name="Xamarin.Forms.Platform.UAP">
 
   	<!-- add directives for your library here -->
-
+    <Namespace Name="Windows.UI.Xaml" MarshalObject="Required All"/>
+  
   </Library>
 </Directives>
--- a/Xamarin.Forms.Platform.UAP/Properties/Xamarin.Forms.Platform.UAP.rd.xml
+++ b/Xamarin.Forms.Platform.UAP/Properties/Xamarin.Forms.Platform.UAP.rd.xml
@@ -28,7 +28,7 @@
   <Library Name="Xamarin.Forms.Platform.UAP">
 
   	<!-- add directives for your library here -->
-    <Namespace Name="Windows.UI.Xaml" MarshalObject="Required All"/>
+    <Namespace Name="Windows.UI.Xaml.Automation.Peers" MarshalObject="Required All"/>
   
   </Library>
 </Directives>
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.WinRT/WindowsBasePlatformServices.cs
```diff
@@ -74,10 +74,14 @@ public virtual Assembly[] GetAssemblies()
 			}
 
 			Assembly thisAssembly = GetType().GetTypeInfo().Assembly;
+			Assembly xamlAssembly = typeof(Xamarin.Forms.Xaml.IMarkupExtension).GetTypeInfo().Assembly;
```
samhouts: Not a big deal, but I feel like this should be down next to line 82.

Xamarin.Forms.Platform.UAP/Properties/Xamarin.Forms.Platform.UAP.rd.xml
```diff
@@ -28,6 +28,7 @@
   <Library Name="Xamarin.Forms.Platform.UAP">
 
   	<!-- add directives for your library here -->
-
+    <Namespace Name="Windows.UI.Xaml" MarshalObject="Required All"/>
```
samhouts: We can actually just do `<Namespace Name="Windows.UI.Xaml.Automation.Peers" MarshalObject="Required All"/> `. I think bringing in the whole namespace is bloaty.

samhouts: But if it works, it works. ;)

rmarinho: Oh nice. let me try that ..

Xamarin.Forms.Platform.WinRT/WindowsBasePlatformServices.cs
```diff
@@ -78,6 +78,10 @@ public virtual Assembly[] GetAssemblies()
 			if (!assemblies.Contains(thisAssembly))
 				assemblies.Add(thisAssembly);
 
+			Assembly xamlAssembly = typeof(Xamarin.Forms.Xaml.IMarkupExtension).GetTypeInfo().Assembly;
+			if (!assemblies.Contains(xamlAssembly))
+				assemblies.Add(xamlAssembly);
```
StephaneDelcroix: did we already had a dependency on that assembly ? or are we adding a new one ?

also, using `typeof(XamlLoader)` means more than using `IMarkupExtension` and helps understand which assembly we are talking about

PR #1027, opened 2017-06-28
[iOS] Fix for disposing a map causing a null reference

dkarzon:
### Description of Change ###

Maps renderer on iOS has been throwing a NullReferenceException when disposing due to the delegate property being null.
I have noticed this when using Xamarin.Forms v2.3.5.239-pre3 in conjunction with the TKCustomMap library.
Added a null check on the Delegate property in the Dispose method of the MapRenderer to make sure it isn't null before calling it's Dispose method.

### PR Checklist ###


rmarinho:
failing tests not related , just needs rebase

PR #1035, opened 2017-07-06
[core] Prevent canonicalization of '*.[.]' and '-0'

kingces95:
### Description of Change ###

User expects to be able to bind Entry to double, float, or decimal backed property and be able to enter "-0.1" and/or "0.1". What actually happens is that "-0" and "0." are both canonicalized to "0" preventing the user from entering their values. 

Under the hood, after each character is entered into the Entry the binding system attempts to convert the text to the bound property type. If successful then the text is replaced with the ToString value of the converted texted which is canonicalized. 

The fix is special case "-0" and "\d*." so that they no longer convert to 0.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=32871
Bug pointed out "\d*."; Testing found "-0" case

### API Changes ###

None

### Behavioral Changes ###

Entering "0." or "-0" into an Entry bound to a decimal, float, or double bound property will not convert to "0".

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/BindingExpression.cs
+++ b/Xamarin.Forms.Core/BindingExpression.cs
@@ -361,6 +361,7 @@ namespace Xamarin.Forms
 			}
 		}
 
+		static Type[] DecimalTypes = new[] { typeof(float), typeof(decimal), typeof(double) };
 		bool TryConvert(BindingExpressionPart part, ref object value, Type convertTo, bool toTarget)
 		{
 			if (value == null)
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Core/BindingExpression.cs
```diff
@@ -371,6 +372,16 @@ bool TryConvert(BindingExpressionPart part, ref object value, Type convertTo, bo
 			object original = value;
 			try
 			{
+				var stringValue = value as string ?? string.Empty;
+				// see: https://bugzilla.xamarin.com/show_bug.cgi?id=32871
+				// do not canonicalize "*.[.]"; "1." should not update bound BindableProperty
+				if (stringValue.EndsWith(".") && DecimalTypes.Contains(convertTo))
```
Therzok: Value is not modified until the end, so why not simply return false instead of throwing?

kingces95: Could do that. I don't have a great reason for one over the other.. Precedent? The current flow throws an exceptions so I stuck with that flow. In general, if I were starting from scratch, I'd avoid flows that throw (e.g. search out for a Convert.TryChangeType or equivalent) but when fixing bugs, I try to adopt whatever convention exists.

PR #1036, opened 2017-07-06
Add try-catch when reading properties dictionary on Windows

jimmgarrido:
### Description of Change ###
After storing non-primitive types in the properties dictionary, future attempts to deserialize the dictionary on Windows would result in a crash. Android and iOS already had a try-catch when reading the properties dictionary that would prevent this crash and output the exception to the log. This just adds the same try-catch to Windows.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=32116

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


paulanthonymorris:
Glad to see this fix. This is our top crash on UWP. Will this make it into 2.3.5?

PR #1033, opened 2017-07-06
[Android] Prevent ObjectDisposed exceptions on Fast Renderers

samhouts:
### Description of Change ###

Calling `Element.ClearValue(Platform.RendererProperty);` when the fast renderers are disposed will prevent the disposed renderer from being inadvertently returned later.

### Bugs Fixed ###

Numerous; TBD.

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #1040, opened 2017-07-07
[Android] Prevent ObjectDisposedException in ButtonBackgroundTracker

mattrichnz:
### Description of Change ###

Fixes an ObjectDisposedException thrown in the `ButtonDrawable.Draw` method. The drawable itself is held by the new class `ButtonBackgroundTracker` introduced in the refactoring PR #941 in XF 2.3.5.255-pre5.

Inside the `UpdateDrawableMethod()`, the field `_backgroundDrawable` is used to hold a reference to the drawable. Previously, we only recreate the drawable if it was null however this does not account for edge cases where the object has been disposed but is not null.

To ensure we have a valid instance of `ButtonDrawable`, we initialise a new instance.
Note: I experimented with simply checking the handle for IntPtr.Zero at this point, but this is not 100% effective. It appears there are certain cases whereby the object can be disposed between `_nativeButton.SetBackground(_backgroundDrawable);` and the native `Draw()' method being invoked.

This is particular noticeable when using advanced layout mechanisms such as Android's `RecyclerView`.

### Bugs Fixed ###

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/ButtonBackgroundTracker.cs
+++ b/Xamarin.Forms.Platform.Android/ButtonBackgroundTracker.cs
@@ -48,9 +48,7 @@ namespace Xamarin.Forms.Platform.Android
 			}
 			else
 			{
-				if (_backgroundDrawable == null)
-					_backgroundDrawable = new ButtonDrawable();
-
+				_backgroundDrawable = new ButtonDrawable();
 				_backgroundDrawable.Button = _button;
 
 				if (_drawableEnabled)


hartez:
@mattrichnz I believe that PR #1033 fixes this problem - if not, please re-open this PR and we'll take another look at it.

mattrichnz:
Thanks @hartez, we have tested #1033 but this does not resolve this particular issue.

As mentioned, there appear to be edge cases where "_backgroundDrawable" is getting disposed but is not null so throws an `ObjectDisposedException` when it is accessed. Prior to #941, this bug did not occur.

We tested a number of different solutions and without rewriting the entire implementation found that ensuring the drawable is recreated each time resolved the issue. Please feel free to improve on this however.

PR #1045, opened 2017-07-09
[XamlC] change module MVID

StephaneDelcroix:
### Description of Change ###

When XamlC modifies a module, modify the module MVID too.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=57931

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###


migueldeicaza:
Jon mentions that it would be nice to only update the MVID if the resulting assembly changes, not every time, otherwise we end up rebuilding everything

PR #1050, opened 2017-07-10
[UWP] Remove competing state recording Title visibility

kingces95:
### Description of Change ###

Cache of whether or not a UWP page's title was not invalidated. Remove the cache so the value is retrieved directly from the control (just like the surrounding properties). 

Cache was being invalidated via https://github.com/xamarin/Xamarin.Forms/blob/248d0176b5a437536f90e54ea5ea34e6b0d12099/Xamarin.Forms.Platform.UAP/MasterDetailControl.cs#L307.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=51509

### API Changes ###

None

### Behavioral Changes ###

Honor value set via `NavigationPage.SetHasNavigationBar(..., false);`

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/MasterDetailPageRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/MasterDetailPageRenderer.cs
@@ -14,7 +14,6 @@ namespace Xamarin.Forms.Platform.UWP
 	{
 		Page _master;
 		Page _detail;
-		bool _showTitle;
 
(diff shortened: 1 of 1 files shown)


samhouts:
Hmm. I tested this on the repro on the bug, and it does not appear to fix the issue. On UWP Desktop, that white bar should be gone.

badnavbar

PR #1051, opened 2017-07-11
[Android] Set TimePicker Time Before Unfocusing

jimmgarrido:
### Description of Change ###
On Android, users who checked TimePicker.Time in the Unfocused event would get the previously set Time instead of the newly chosen one. This fixes the issue by setting the Time property before setting the IsFocused property and brings it in line with the behavior of [DatePicker](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/Renderers/DatePickerRenderer.cs#L108). 

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=58071

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


hartez:
@jimmgarrido Can we write a UI test for this? And DatePicker, while you're at it.

PR #1056, opened 2017-07-14
[UWP] Prevent crash when using DateTime.MinValue for DatePicker

jimmgarrido:
### Description of Change ###

On UWP, if you try to set the DatePicker.MinimumDate to DateTime.MinValue when the UTC offset is positive an `ArgumentOutOfRangeException` is thrown because the time gets converted to UTC which makes it go out of range.

This fix catches the exception and specifies the date as UTC in order to prevent it from being converted.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=57855

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.WinRT/DatePickerRenderer.cs
+++ b/Xamarin.Forms.Platform.WinRT/DatePickerRenderer.cs
@@ -90,13 +90,25 @@ namespace Xamarin.Forms.Platform.WinRT
 		void UpdateMaximumDate()
 		{
 			DateTime maxdate = Element.MaximumDate;
-			Control.MaxYear = new DateTimeOffset(maxdate.Date);
+			Control.MaxYear = new DateTimeOffset(maxdate);
 		}
 
 		void UpdateMinimumDate()
 		{
 			DateTime mindate = Element.MinimumDate;
-			Control.MinYear = new DateTimeOffset(mindate);
+
+			try
+			{
+				Control.MinYear = new DateTimeOffset(mindate);
+			}
+			catch (ArgumentOutOfRangeException)
+			{
+				// This will be thrown when mindate equals DateTime.MinValue and the UTC offset is positive
+				// because the resulting DateTimeOffset.UtcDateTime will be out of range. In that case let's
+				// specify the Kind as UTC so there is no offset.
+				mindate = DateTime.SpecifyKind(mindate, DateTimeKind.Utc);
+				Control.MinYear = new DateTimeOffset(mindate);
+			}
 		}
 
 		void UpdateTextColor()


Xamarin.Forms.Platform.WinRT/DatePickerRenderer.cs
```diff
@@ -96,7 +96,19 @@ void UpdateMaximumDate()
...
-			Control.MinYear = new DateTimeOffset(mindate);
+
+			try
+			{
+				Control.MinYear = new DateTimeOffset(mindate);
+			}
+			catch(ArgumentOutOfRangeException)
+			{
+				// This will be thrown if you try to set MinYear to DateTime.MinValue when the UTC offset is positive
+				// because the time is out range once it's converted. In that case, specify the Kind as UTC
```
samhouts: > time is out range

typo, might confuse future dev

Xamarin.Forms.Platform.WinRT/DatePickerRenderer.cs
```diff
@@ -96,7 +96,19 @@ void UpdateMaximumDate()
...
+			try
+			{
+				Control.MinYear = new DateTimeOffset(mindate);
+			}
+			catch(ArgumentOutOfRangeException)
+			{
+				// This will be thrown if you try to set MinYear to DateTime.MinValue when the UTC offset is positive
+				// because the time is out range once it's converted. In that case, specify the Kind as UTC 
+				// so it doesn't get converted.
+				mindate = DateTime.SpecifyKind(mindate, DateTimeKind.Utc);
```
samhouts: I'm sure there's a reason, but why don't we just always get the Date value and use that instead of adding a `try catch` and additional logic?

jimmgarrido: We actually do do that in `UpdateMaximumDate()` and that was the first thing I tried. But Date just [returns another DateTime](https://msdn.microsoft.com/en-us/library/system.datetime.date(v=vs.110).aspx) with the Time set to 0:00 so in this case the crash still happens.

I think what we should do is always specify the incoming value as Utc and avoid the extra object from getting the Date. This would eliminate the `try/catch` and also prevent similar cases where the original date might get converted to the day before (or day after for MaxDate).

Xamarin.Forms.Platform.WinRT/DatePickerRenderer.cs
```diff
@@ -90,13 +90,25 @@ void UpdateDate(DateTime date)
 		void UpdateMaximumDate()
 		{
 			DateTime maxdate = Element.MaximumDate;
-			Control.MaxYear = new DateTimeOffset(maxdate.Date);
```
jimmgarrido: Date just returns another DateTime object with the same date so it's redundant.

Xamarin.Forms.Platform.WinRT/DatePickerRenderer.cs
```diff
@@ -90,13 +90,25 @@ void UpdateDate(DateTime date)
...
 		void UpdateMinimumDate()
 		{
 			DateTime mindate = Element.MinimumDate;
-			Control.MinYear = new DateTimeOffset(mindate);
+
+			try
+			{
+				Control.MinYear = new DateTimeOffset(mindate);
+			}
+			catch (ArgumentOutOfRangeException)
```
jimmgarrido: Turns out a try/catch is really just the simplest fix because the exception should only be thrown in this one scenario.

jimmgarrido:
This should be okay to merge now. I didn't make any changes other than making the comment more clearer because this seems like an edge case so most people probably won't hit the logic inside the catch.

PR #1059, opened 2017-07-19
Animation callbacks directly on invoke params

sichy:
### Description of Change ###

I've added callbacks straight to the Animation functions. They are often called from void sync methods, without the Task API cleverness and for simplicity it is useful to be able to add when completed callback without the TaskCompletion.

### API Changes ###

Changed:
- public static Task FadeTo(this VisualElement view, double opacity, uint length = 250, Easing easing = null, Action finished = null) // Action Finished
- public static Task LayoutTo(this VisualElement view, Rectangle bounds, uint length = 250, Easing easing = null, Action finished = null) // Action Finished
- public static Task RelRotateTo(this VisualElement view, double drotation, uint length = 250, Easing easing = null, Action finished = null) // Action Finished
- public static Task RelScaleTo(this VisualElement view, double dscale, uint length = 250, Easing easing = null, Action finished = null) // Action Finished
- public static Task RotateTo(this VisualElement view, double rotation, uint length = 250, Easing easing = null, Action finished = null) // Action Finished
- public static Task RotateXTo(this VisualElement view, double rotation, uint length = 250, Easing easing = null, Action finished = null) // Action Finished
- public static Task RotateYTo(this VisualElement view, double rotation, uint length = 250, Easing easing = null, Action finished = null) // Action Finished
- public static Task ScaleTo(this VisualElement view, double scale, uint length = 250, Easing easing = null, Action finished = null) // Action Finished
- public static Task TranslateTo(this VisualElement view, double x, double y, uint length = 250, Easing easing = null, Action finished = null) // Action Finished

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/ViewUnitTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/ViewUnitTests.cs
@@ -282,7 +282,7 @@ namespace Xamarin.Forms.Core.UnitTests
 			var view = new View {IsPlatformEnabled = true, Platform = new UnitPlatform ()};
 			Ticker.Default = new BlockingTicker ();
 
-			view.FadeTo (0.1);
+            view.FadeTo (0.1, finished:() => Console.WriteLine("TestFadeTo.Finished"));
 
 			Assert.True (Math.Abs (0.1 - view.Opacity) < 0.001);
 		}
--- a/Xamarin.Forms.Core/ViewExtensions.cs
+++ b/Xamarin.Forms.Core/ViewExtensions.cs
@@ -19,7 +19,7 @@ namespace Xamarin.Forms
 			view.AbortAnimation("SizeTo");
 		}
 
-		public static Task<bool> FadeTo(this VisualElement view, double opacity, uint length = 250, Easing easing = null)
+		public static Task<bool> FadeTo(this VisualElement view, double opacity, uint length = 250, Easing easing = null, Action finished = null)
 		{
 			if (view == null)
 				throw new ArgumentNullException("view");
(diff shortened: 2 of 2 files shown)


sichy:
Cla signed.

Xamarin.Forms.Core.UnitTests/ViewUnitTests.cs
```diff
@@ -282,7 +282,7 @@ public void TestFadeTo ()
 			var view = new View {IsPlatformEnabled = true, Platform = new UnitPlatform ()};
 			Ticker.Default = new BlockingTicker ();
 
-			view.FadeTo (0.1);
+            view.FadeTo (0.1, finished:() => Console.WriteLine("TestFadeTo.Finished"));
```
hartez: This _demonstrates_ the callback, but it doesn't _test_ that the callback is invoked (i.e., this test wouldn't fail even if `finished?.Invoke();` were missing from the `FadeTo` method). This needs a test which would fail if the callback weren't invoked. And all of the other updated methods would need similar tests.

Xamarin.Forms.Core/ViewExtensions.cs
```diff
@@ -35,12 +35,19 @@ public static Task<bool> FadeTo(this VisualElement view, double opacity, uint le
 					v.Opacity = f;
 			};
 
-			new Animation(fade, view.Opacity, opacity, easing).Commit(view, "FadeTo", 16, length, finished: (f, a) => tcs.SetResult(a));
+			new Animation(fade, view.Opacity, opacity, easing)
+                .Commit(view, "FadeTo", 16, length,
```
hartez: All of this PR needs to be converted from spaces to tabs.

hartez:
@sichy I'm not sure I understand what you're trying to accomplish with this change. Is there something this change allows which cannot be accomplished by something like: 
`view.FadeTo(0.1).ContinueWith(task => Console.WriteLine("finished"));` ?

sichy:
Yes the continue with can be used, but the usual APIs in IOS and Droid users are used to have a callback on the actual method.

I designed this for my colleague who used the ViewExtensions a lot (swipeable generic frame for swipe cells and similar) and him being "old" ObjC he has issues with the usual Task API concepts, hence this callback system.

hartez:
Then I don't think you actually need changes to Forms; you just need to include your own convenience extension methods in your own projects. Add something like

```
public static Task<bool> FadeTo(this VisualElement view, double opacity, uint length = 250, Easing easing = null, Action finished = null)
{
	return view.FadeTo(opacity, length, easing).ContinueWith(f => {
		finished?.Invoke();
		return f.Result;
	});
}
```

for each of the methods and your colleague won't have to learn anything new.

sichy:
Good point with the extension methods. Lets close this one then.

PR #1061, opened 2017-07-20
[Maps] Change Map.LastMoveToRegion to public accessor

dkarzon:
### Description of Change ###

When using a custom map and renderer that inherit Map and MapRenderer the map will default it's centre to Rome when it is created from xaml.
This happens because the map has a property called `LastMoveToRegion` which can only be set internally and when using xaml to create the map gets set to Rome (See Map default constructor) and from the renderers `OnElementChanged` method there is a call made to `MoveToRegion`. Which is the only external place to set this property, however using a custom renderer we don't need to call `MoveToRegion` since we have access to the native map functions.

### API Changes ###

Changed:
 - internal MapSpan LastMoveToRegion => public MapSpan LastMoveToRegion

### Behavioral Changes ###

This change doesn't affect the behaviour of Xamarin.Forms.Maps at all but will help custom Map Renderers have more control over the map when navigating back and forward through pages with maps on them.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Maps/Map.cs
+++ b/Xamarin.Forms.Maps/Map.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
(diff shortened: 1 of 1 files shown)


rmarinho:
The best way to do this is to add a IMapController, a interface with methods we want to expose to the Renderer , similiar to what we do with IButtonController and others.

We can have a method called SetLastMoveRegion or something.

PR #1066, opened 2017-07-24
[Android] Prevent ObjectDisposedException in ButtonBackgroundTracker

mattrichnz:
### Description of Change ###

Fixes an ObjectDisposedException thrown in the `ButtonDrawable.Draw` method. The drawable is held by the class `ButtonBackgroundTracker` introduced in PR #941.

Inside the `UpdateDrawableMethod()`, the field `_backgroundDrawable` is used to hold a reference to the drawable. Previously, the drawable was only recreated if it was null however this **does not** account for edge cases where the object has been disposed but is not null.

To ensure we have a valid `ButtonDrawable`, we initialize a new instance.
Note: I experimented with simply checking the handle for IntPtr.Zero at this point, but this is not 100% effective. It appears there are certain cases whereby the object can be disposed between `_nativeButton.SetBackground(_backgroundDrawable);` and the native `Draw()' method being invoked.

This is particular noticeable when using advanced layout mechanisms such as Android's `RecyclerView`.

Note: This is the same as #1040 which was closed.

### Bugs Fixed ###

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/ButtonBackgroundTracker.cs
+++ b/Xamarin.Forms.Platform.Android/ButtonBackgroundTracker.cs
@@ -48,9 +48,7 @@ namespace Xamarin.Forms.Platform.Android
 			}
 			else
 			{
-				if (_backgroundDrawable == null)
-					_backgroundDrawable = new ButtonDrawable();
-
+				_backgroundDrawable = new ButtonDrawable();
 				_backgroundDrawable.Button = _button;
 
 				if (_drawableEnabled)


mattrichnz:
@hartez Same as #1040. Reopened.

samhouts:
@mattrichnz Can you please verify whether #1063 resolves this issue? If not, can you please provide a reproduction so I can test your fix? Thank you!

mattrichnz:
Thanks @samhouts. I will investigate this further on Monday.
#1063 won't solve it straight away as we have a custom collection view build on a RecyclerView, but I am interesting if applying the cleanup principles you mention will help. If not, will put together a repro project for you to look at.

mattrichnz:
@samhouts Have been doing further investigation into this today. This is the repro:
1. We have a page with our `RecyclerView` collection view. Everything lays out and works correctly.
2. We push a new page onto the stack. Load some images etc, enough to trigger a GC.Collect.
3. We pop the new page off the stack. ObjectDisposedException is thrown in `ButtonDrawable.Draw`.

Notes:
- When we create a view holder, we create a renderer for a XAML view which contains a button.
- Hence inside each view holder, a native fast button renderer will be tied to the portable element and internally each contains a `ButtonDrawable`.

Unfortunately, we can't handle `Dispose` in a similar way to #1063 as the page still exists in the navigation stack. Although we are not disposing the `RecyclerView`, it seems that whatever GC.Collect is doing is triggering cleanup of the native drawable inside the Button fast renderer although the .NET object still exists.

This is the edge case I mention in the PR description. Prior to the refactoring PR #941, this issue did not occur. Hence, the rationale behind the changes in this PR.

If you still require a reproduction project, can you please let me know. I will create one and DM you a link.

mattrichnz:
@samhouts @davidortinau Given the change in direction with XF 2.4, this issue can likely be addressed during the Fast Renderers refactor by the XF team. Will leave it at your digression to keep the issue open or close as part of your work.

Thanks!

rmarinho:
Since we don't have a test to verify this really fixes the problem and not a symptom we are not accepting this patch at the time. Please reopen if you have a reproducion and test.

PR #1070, opened 2017-07-26
[Android] Dispose check on GetDesiredSize() For FastButtonRenderer

bentmar:
return new SiseRequest if object is disposed

### Description of Change ###

Returns a new SiseRequest if button is disposed

### Bugs Fixed ###

https://forums.xamarin.com/discussion/comment/287671/#Comment_287671

### API Changes ###

none

### PR Checklist ###

- [ x] Rebased on top of master at time of PR
- [ x] Changes adhere to coding standard
- [ x] Consolidate commits as makes sense


samhouts:
@bentmar Can you verify for me whether #1033 and/or #1063 resolve this issue? If not, can you please provide a repro so I can test your fix? Thank you!

bentmar:
I cant verify sorry. It would be great if you included all the dispose fixes in a new pre release to test on. My understanding of the disposed object crashes is that they are very random.

rmarinho:
Since we don't have a test to verify this really fixes the problem and not a symptom we are not accepting this patch at the time. Pelase reopen if you have a reproducion and test.

PR #1074, opened 2017-07-31
Add support for backcompat hint flags

hartez:
### Description of Change ###

Add support for backwards-compatibility hint flags.

### API Changes ###

Added:
 - `public static void SetFlags(params string[] flags)` 

### Behavioral Changes ###

Android fast renderers now require setting flag `FastRenderers_Experimental`.

### DIFF
--- a/Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
@@ -42,7 +42,12 @@ namespace Xamarin.Forms.ControlGallery.Android
 			if (!Debugger.IsAttached)
 				Insights.Initialize(App.InsightsApiKey, ApplicationContext);
 
+#if TEST_EXPERIMENTAL_RENDERERS
+			Forms.SetFlags("FastRenderers_Experimental");
+#endif
+
 			Forms.Init(this, bundle);
+
 			FormsMaps.Init(this, bundle);
 			AndroidAppLinks.Init(this);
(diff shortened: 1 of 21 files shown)


rmarinho:
build

Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
```diff
@@ -42,7 +42,12 @@ protected override void OnCreate(Bundle bundle)
 			if (!Debugger.IsAttached)
 				Insights.Initialize(App.InsightsApiKey, ApplicationContext);
 
+#if TEST_EXPERIMENTAL_RENDERERS
+			Forms.SetFlags("FastRenderers_Experimental");
```
rmarinho: Shouldn' we use our namespace ?

hartez: I thought that when we discussed it we decided that our stuff doesn't need namespaces, but we would recommend that third parties using this should prefix with namespaces.

bugadani:
Is there a particular reason flags are basically magic strings? Wouldn't an enum be a better choice? Since the Flags class is internal, we can't use it to have some compiler support for autocompletion and preventing typos, making it annoying to dig through documentation to find the correct flag name and what it does. Also, is there a reason this is being done in runtime, instead of compile time? Coming from an embedded world, wasting code memory when it could be done in an objectively better way bugs me. I'm concerned that this is a rushed change.

PR #1075, opened 2017-07-31
[Core, Android] Fix to address the elusive 'Sharing violation on path PropertyStore.forms.tmp'

dgeller-OUHSC:
### Description of Change ###

Note: we still haven't gotten a satisfactory way to reproduce the actual bug, but this seems to be a fix for it, hence no tests.

-Added a semaphore to Application.SetPropertiesAsync() to wrap some logic that is an unconvincing guard against concurrent FileAccess calls.
-made several methods asynchronous to honor a later asynchronous call to Application.SetPropertiesAsync() within the OnStateChanged() method of FormsAppCompatActivity
-change the the deserialization piece of saving application Properties to file readonly to help with collisions

Also, I did not rebase, I simply merged.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=57469

We had been seeing the 'Sharing violation on path PropertyStore.forms.tmp' bug in our application almost daily before these changes on certain phones. I have been running these changes for two weeks in production with no problems and no further sharing violations.

### API Changes ###

None

### Behavioral Changes ###

Does not make any behavioral changes (beyond attempt to fix the specified bug)

### PR Checklist ###


jassmith:
build

rmarinho:
@jassmith see SucceededWithIssues  .. needs docs it sems

dgeller-OUHSC:
@rmarinho Is the doc error something I should change?

RogerHardiman:
Please please please merge in the Semaphore to stop the crash during the Save.

The big goes right back to Xamarin Forms 1.x and https://bugzilla.xamarin.com/show_bug.cgi?id=28709 logged the bug and logged a fix using semaphores.

Xamarin went off and implemented a different fix that did not use an lock. Their fix has bugs and needed to use an atomic TestAndSet and did not so there js still a path where two simultaneous calls to SavePropAsync can arrive in the file writing code.
 All my own applications implement some locking before calling the SavePropAsync function.

Xamarin.Forms.Core/Application.cs
```diff
@@ -314,6 +317,7 @@ void OnPopCanceled()
 			if (_saveAgain)
 				await DependencyService.Get<IDeserializer>().SerializePropertiesAsync(Properties);
 			_isSaving = _saveAgain = false;
+			SaveSemaphore.Release();
```
hartez: This should be in a `try...finally` block so we can guarantee that the semaphore is released if there's an exception while saving the properties.

hartez:
Also, note that we don't need to add a UI test for this bug in this PR; we accidentally created one in #1151 (that test is now failing consistently on Android because of https://bugzilla.xamarin.com/show_bug.cgi?id=57469).

Xamarin.Forms.Core/Application.cs
dgeller-OUHSC: should I catch and log the exception similar to what's done in Xamarin.Forms.Platform.Android.Deserializer.DeserializePropertiesAsync()?

RogerHardiman:
While we are in this part of the codebase I've spotted another possible bug.

I cannot see anything that prevents the Deserialization Task running at the same time as the Serialization Task when the Serialization Task is in the store.Delete() and store.MoveFile() code.

It would be a rare condition but it would be possible for the following sequence to ocure
```
  Serialization    calls store.Delete
  Deserialization  reads the store file. This will fail as the store file has been deleted
  Serialization    calls store.MoveFile to rename the .tmp store file
```

Some extra locking would be needed.

RogerHardiman:
Removed last comment. It discussed a race condition where Reading the store file could occur in the middle of the store.Delete() and store.MoveFile (when saving). The issue is still there bit I realised XF only reads the store file once on App Init.

opcodewriter:
Wow, this issue is related to a bug submitted on Bugzilla in 2015...

rmarinho:
seems it needs a docs update

ashrafsarhan:
Any updates for this issue, I'm running with this issue mainly with Android on production.

PR #1082, opened 2017-08-03
[Android] Use 24-hour time format on TimePicker text if enabled

pauldipietro:
### Description of Change ###

A prior PR allowed for the use of 24-hour time on the `TimePicker`, but didn't set the text part of the control. While it's technically possible to add a test for this, it would be reliant upon knowing the setting of the device the test is being run on at the time, so it's been omitted for the moment.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=54645

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


rmarinho:
👍

PR #1085, opened 2017-08-04
[Core] Add ListViewCachingStrategy.RecycleElementAndDataTemplate

kingces95:
### Description of Change ###

Reduce ListView DataTemplate dynamism to enable caching DataTemplate by the type of the item in the ItemsSource. This in turn allows caching prototypical cells by the item type which are used under the hood to pre-compute height of a cell before scrolling on screen. 

One thing to keep in mind about iOS when doing a code reivew: When iOS returns a cell via `DequeueReusableCell` it will only reclaim the cell after it appears and disappears. So just getting a cell to measure but never attaching/rendering it _will leak the cell_. So if we're gonna create a cell just for measuring then we only want to create it once _per type_ instead of once _per index_ (although even once per index is better than what we do now which is create a new cell every time an index appears). 

Testing yet to do:
- [x] Improve existing strategy. 
- [x] Remove index -> type cache and instead re-ask for item each time and use it's type to hit type -> proto cell cache. 
- [x] Test multiple data templates.
- [ ] Test with image? 
- [ ] Test changing DataTemplateSelector invalidates cache of item types to prototypical cells.
- [x] Test grouping.
- [x] Re-test bug fixed by async height estimation to see if this fixes that bug so code added for that fix can be removed. See: https://github.com/xamarin/Xamarin.Forms/pull/994
- [x] Automate into a UITest
- [ ] _TBD_ "Flinging" a list view causes a ton of cells to be created and not reclaimed. Understanding why that happens may be a task for another day. 

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=52487

### API Changes ###

Added ListViewCachingStrategy.RecycleElementAndDataTemplate. When set OnSelectTemplate will be called once per item _type_ instead of once per _instance_. Consequently, a DataTemplate should be selected solely based on an items type and not any of it's data. Furthermore, the DataTemplate must be activated using the ctor that takes a Type.

One thing to note about iOS that drove the cache design: When iOS returns a

### Behavioral Changes ###

ListView activates only as many cells as are visible on screen. Finally. 

### PR Checklist ###

TODO: Commented out some code to fix other issues which may need to be added back if they're not fixed by this PR. 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla52487.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla52487.cs
@@ -1,0 +1,1028 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System.Linq;
+using System.ComponentModel;
+using System.Runtime.CompilerServices;
+using System;
+using System.Diagnostics;
+using System.Collections.Generic;
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Threading;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(
+		IssueTracker.Bugzilla,
+		52487,
+		"ListView with Recycle + HasUnevenRows generates lots (and lots!) of content view",
+		// https://bugzilla.xamarin.com/show_bug.cgi?id=52487
+		PlatformAffected.iOS
+	)]
+	public class Bugzilla52487 : TestContentPage
+	{
+#if __IOS__
+		const int MaxAskDelta = 0;
+		const int MaxViewDelta = 5;
+		const int MaxAttachDelta = 5;
+#elif __ANDROID__
+		const int MaxAskDelta = 0;
+		const int MaxViewDelta = 1;
+		const int MaxAttachDelta = 1;
+#else
+		const int MaxAskDelta = 0;
+		const int MaxViewDelta = int.MaxValue;
+		const int MaxAttachDelta = int.MaxValue;
+#endif
+
+		const int CountFontSize = 12;
+		const int CellFontSize = 12;
+		const int MinScrollDelta = 2;
+		const int ItemsCount = 1000;
+		const int GroupCount = 100;
+		const int DefaultItemHeight = 300 / 4;
+		const int MinimumItemHeight = 40;
+
+		// generate alternate item type when % item id is zero
+		const int ItemTypeModulous = 7;
+
+		// render half height when % item id is zero (RecycleElement or RecycleElementAndDataTemplate)
+		const int HalfHeightModulous = 5;
+
+		// select alternate cell type when % item id is zero (RecycleElement)
+		const int DataTemplateModulous = 3;
+
+		static Tuple<int, int, int> Mix = new Tuple<int, int, int>(255, 255, 100);
+		static Tuple<int, int, int> AltMix = new Tuple<int, int, int>(100, 100, 255);
+		static IEnumerable<Color> ColorGenerator(Tuple<int, int, int> mix)
+		{
+			double colorDelta = 0;
+			while (true)
+			{
+				colorDelta += 2 * Math.PI / 100;
+				var r = (Math.Sin(colorDelta) + 1) / 2 * 255;
+				var g = (Math.Sin(colorDelta * 2) + 1) / 2 * 255;
+				var b = (Math.Sin(colorDelta * 3) + 1) / 2 * 255;
+
+				if (mix != null)
+				{
+					r = (r + mix.Item1) / 2;
+					g = (g + mix.Item2) / 2;
+					b = (b + mix.Item3) / 2;
+				}
+
+				yield return Color.FromRgb((int)r, (int)g, (int)b);
+			}
+		}
+
+		[Preserve(AllMembers = true)]
+		class LazyReadOnlyList<V> : IReadOnlyList<V>
+			where V : class
+		{
+			int _count;
+			object _context;
+			List<WeakReference<V>> _items;
+			Action<int> _onAsk;
+			Func<LazyReadOnlyList<V>, int, object, V> _activate;
+
+			internal LazyReadOnlyList(
+				int count,
+				object context,
+				Action<int> onAsk,
+				Func<LazyReadOnlyList<V>, int, object, V> activate)
+			{
+				_count = count;
+				_context = context;
+				_onAsk = onAsk;
+				_activate = activate;
+				_items = new List<WeakReference<V>>(
+					Enumerable.Range(0, count)
+					.Select(o => new WeakReference<V>(null))
+				);
+			}
+
+			protected object Context
+			{
+				get { return _context; }
+				set { _context = value; }
+			}
+			protected IEnumerable<WeakReference<V>> WeakItems =>
+				_items;
+
+			public V this[int index]
+			{
+				get
+				{
+					_onAsk(index);
+
+					var weakItem = _items[index];
+
+					V item;
+					if (!weakItem.TryGetTarget(out item))
+					{
+						_items[index] =
+							new WeakReference<V>(
+								item = _activate(this, index, _context));
+					}
+
+					return item;
+				}
+			}
+
+			public int Count
+				=> _count;
+
+			public IEnumerator<V> GetEnumerator()
+			{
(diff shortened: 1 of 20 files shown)


mattrichnz:
This is really nice. Great work!

Xamarin.Forms.Core.UnitTests/DataTemplateSelectorTests.cs
```diff
@@ -90,4 +90,69 @@ public void NestingThrowsException ()
...
+
+			var selector = new TestDataTemplateSelector();
+			listView.ItemTemplate = selector;
+			Assert.That(selector.Counter == 0);
+
+			Assert.IsInstanceOf<ViewCell>(listView.TemplatedItems[0]);
+			Assert.That(selector.Counter == 1);
+
+			Assert.IsInstanceOf<ViewCell>(listView.TemplatedItems[1]);
+			Assert.That(selector.Counter == 1);
```
kingces95: Resolving the item for the second TemplatedItem hits the DataTemplate cache and so the counter remains at 1.

Xamarin.Forms.Core.UnitTests/DataTemplateSelectorTests.cs
```diff
@@ -90,4 +90,69 @@ public void NestingThrowsException ()
...
+				declarativeTemplate = new DataTemplate(typeof(ViewCell));
+				proceduralTemplate = new DataTemplate(() => new EntryCell());
+			}
+
+			protected override DataTemplate OnSelectTemplate(object item, BindableObject container)
+			{
+				Counter++;
+
+				if (item is string)
+					return declarativeTemplate;
```
kingces95: Given that (1) the DataTemplate is purely a function of the type of the item and (2) the DataTemplate declares it always returns a ViewCell (e.g. uses the Type ctor) let us know that all strings will be rendered with a ViewCell and allows us to cache that fact.

Xamarin.Forms.Core/DataTemplateSelector.cs
```diff
@@ -1,15 +1,39 @@
...
+			dataTemplate = OnSelectTemplate(item, container);
+			if (dataTemplate is DataTemplateSelector)
+				throw new NotSupportedException(
+					"DataTemplateSelector.OnSelectTemplate must not return another DataTemplateSelector");
+
+			if (recycle)
+			{
+				if (!dataTemplate.CanRecycle)
+					throw new NotSupportedException(
+						"RecycleElementAndDataTemplate requires DataTemplate activated with ctor taking a type.");
```
kingces95: If we don't add the new caching strategy `RecycleElementAndDataTemplate` then we cannot raise this exception.

Xamarin.Forms.Core/ItemsView.cs
```diff
@@ -86,13 +86,18 @@ static void OnItemsSourceChanged(BindableObject bindable, object oldValue, objec
...
 		{
-			var lv = b as ListView;
-			if (lv == null)
+			var listView = bindable as ListView;
+			if (listView == null)
 				return true;
 
-			return !(lv.CachingStrategy == ListViewCachingStrategy.RetainElement && lv.ItemTemplate is DataTemplateSelector);
+			var isRetainStrategy = listView.CachingStrategy == ListViewCachingStrategy.RetainElement;
+			var isDataTemplateSelector = listView.ItemTemplate is DataTemplateSelector;
```
kingces95: This is just refactor to make it clear that `RetainElement` cannot be used with a `DataTemplateSelector`.

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -366,35 +364,36 @@ void OnScrollToRequested(object sender, ScrollToRequestedEventArgs e)
 
 		void UpdateEstimatedRowHeight()
 		{
```
kingces95: This is just a refactor:

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -639,23 +642,23 @@ public UnevenListViewDataSource(ListViewDataSource source) : base(source)
 			{
 			}
 
-			internal void CacheDefinedRowHeights()
-			{
-				Task.Run(() =>
```
kingces95: Hopefully this work will also fix the bug that necessitated the introduction of this logic. This logic _asynchronously_ pre-computes the heights of cells. The idea is that the computation is done faster than the user scrolls so the heights are available by the time the user scrolls to the item.

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -692,22 +695,56 @@ internal nfloat GetEstimatedRowHeight(UITableView table)
 				return CalculateHeightForCell(table, firstCell);
 			}
 
-			public override nfloat EstimatedHeight(UITableView tableView, NSIndexPath indexPath)
```
kingces95: Hopefully, by caching prototypical cells, we can compute heights fast enough so we don't need to async generate actual/estimated heights. Seems to work on the test case.

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -1041,12 +1083,25 @@ public void UpdateGrouping()
 				_uiTableView.ReloadData();
 			}
 
-			protected Cell GetCellForPath(NSIndexPath indexPath)
+			protected ITemplatedItemsList<Cell> GetTemplatedItemsListForPath(NSIndexPath indexPath)
```
kingces95: Hoisting existing logic into a method.

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -692,22 +695,56 @@ internal nfloat GetEstimatedRowHeight(UITableView table)
...
+				}
+
+				var templatedItems = GetTemplatedItemsListForPath(indexPath);
+				return templatedItems.UpdateContent(protoCell, indexPath.Row);
 			}
 
 			public override nfloat GetHeightForRow(UITableView tableView, NSIndexPath indexPath)
 			{
-				var cell = GetCellForPath(indexPath);
+				var cell = GetPrototypicalCell(indexPath);
```
kingces95: Where it all comes together! If a prototypical cell exists for the the type of item at the index path then use that instead of recreating a view for each height measurement.

johot: Finally! This is what I changed in my own fork to fix this :D Great!

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -692,22 +671,42 @@ internal nfloat GetEstimatedRowHeight(UITableView table)
...
+
+			internal Cell GetPrototypicalCell(NSIndexPath indexPath)
 			{
-				if (_useEstimatedRowHeight)
-					return tableView.EstimatedRowHeight;
+				var itemTypeOrDataTemplate = default(object);
+
+				var cachingStrategy = List.CachingStrategy;
+				if (cachingStrategy == ListViewCachingStrategy.RecycleElement)
+					itemTypeOrDataTemplate = GetDataTemplateForPath(indexPath);
```
kingces95: This is where the existing Recycle strategy is improved at the cost of a slight break whereby a DataTemplate can no longer return a different type of cell. Requiring a DataTemplate always returns the same type of cell allows us to use the DataTemplate to hit a cache here to get a prototypical cell for height measuring. It's more expensive than the new recycle as we always have to ask for the DataTemplate for an item instead just using the item's type as the cache key.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla52487.cs
```diff
@@ -0,0 +1,1028 @@
...
+{
+	[Preserve(AllMembers = true)]
+	[Issue(
+		IssueTracker.Bugzilla,
+		52487,
+		"ListView with Recycle + HasUnevenRows generates lots (and lots!) of content view",
+		// https://bugzilla.xamarin.com/show_bug.cgi?id=52487
+		PlatformAffected.iOS
+	)]
+	public class Bugzilla52487 : TestContentPage
```
kingces95: XTC Nunit doesn't like our non-existent .ctor... Maybe the linker removed the default .ctor in the XTC case but not the VS case. Or maybe VS launches nunit in a different way than XTC.

PR #1087, opened 2017-08-05
[Android] Fix ListView memory leak

huy8687:
### Description of Change ###

Fix ListView memory leak.

When push and pop a page which contain a ListView many times. The binding between ViewCell and ViewModel is not removed => access disposed object exception.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=45722

### PR Checklist ###


rmarinho:
build

PR #1089, opened 2017-08-07
[UWP] Controls are ignoring their rendering routines.

BradChase2011:
### Description of Change ###
Moved the native control to add AFTER the events were attached to.  The issue was their render routines were being ignored because they had loaded during or shortly after the Children.Add causing the Loaded event to never be caught.  This fixes that issue.  All our views load perfectly now.

No need for tests, this is a simple and obvious change.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=58630

### API Changes ###

None.

### Behavioral Changes ###

UWP Views will now update correctly when sizing the window, validation, to everyday loading of views.  You no longer need to force controls to update.

### PR Checklist ###


rmarinho:
build

laarmen:
I have to ask: what's up with the "build" comments ?

rmarinho:
@laarmen testing integration with vsts and to build forks

PR #1094, opened 2017-08-10
Avoid exception when removing a page from navigation during Appearing handler

hartez:
### Description of Change ###

Remove unnecessary call to `ExecutePendingTransactions` which causes a "recursive entry to executePendingTransactions" error for Android Support libraries version 23 when removing a page from the navigation stack in response to another page's `Appearing` event. With Android Support libraries version 25 and up, this was throwing an exception with the error "FragmentManager is already executing transactions"; this change also fixes that issue.

### Bugs Fixed ###

- Problem caused by previous fix to [53179 – PopAsync crashing after RemovePage when support packages are updated to 25.1.1](https://bugzilla.xamarin.com/show_bug.cgi?id=53179#c42)

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


rmarinho:
build

rmarinho:
build

PR #1098, opened 2017-08-14
macOS: Make ListViewDataSource public

MarkoBL:
### Description of Change ###

Is it possible to make ListViewDataSource public? This makes it really easy to use it in custom ListViews, otherwise you have to copy lots of code into your own project and even that doesn't work at all, as there are function calls to other internal methods.

### PR Checklist ###


rmarinho:
Nah, i don't see this as public.. there's to much internal stuff as you said.

PR #1103, opened 2017-08-16
[iOS] Adjust SearchBar to allow functional Korean text input

pauldipietro:
### Description of Change ###

An issue currently exists where Korean text input causes characters to enter individually and not combine as they should; for example, ㅎ and ㅜ should convert to 후, but in the current behavior display separately. The TextChanged hook was causing a conflict in particular, but because the event only fires when text is entered, it's possible to track text entry and handle it separately from if `Element.Text` is otherwise set, allowing for the input to function as expected.

###  Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=51825

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


rmarinho:
build

allbdeveloper:
When this PR will be included in the stable release?

woojung3:
@allbdeveloper @pauldipietro 
I'm also curios when this PR will be included in the stable release.

PR #1034, opened 2017-07-06
[Android] Fix missing shadow on Button fast renderer

rmarinho:
### Description of Change ###

Fix missing states on fast button renderer

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [ ] Rebased on top of master at time of PR
- [ ] Changes adhere to coding standard
- [ ] Consolidate commits as makes sense

PR #1106, opened 2017-08-18
[Android] Fix bugzilla58868

rmarinho:
### Description of Change ###

Fix regression fromn #1034 

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=58868

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Platform.cs
+++ b/Xamarin.Forms.Platform.Android/Platform.cs
@@ -1134,17 +1134,23 @@ namespace Xamarin.Forms.Platform.Android
 			{
 				//On Material design the button states use Elevation property, we need to make sure
 				//we update the elevation of other controls to be over the previous one
-				if (!_minimumElevation.ContainsKey(i))
-					_minimumElevation[i] = GetChildAt(i).Elevation;
-				for (int j = 0; j < _minimumElevation.Count() - 1; j++)
+				if (Forms.IsLollipopOrNewer)
 				{
-					while (_minimumElevation[j] > _minimumElevation[j + 1])
+					if (!_minimumElevation.ContainsKey(i))
 					{
-						_minimumElevation[j + 1] = _minimumElevation[j] + 1;
-						GetChildAt(j + 1).Elevation = _minimumElevation[j + 1];
+						_minimumElevation[i] = GetChildAt(i).Elevation;
 					}
-					if (j == i)
-						break;
+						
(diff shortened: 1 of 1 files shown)


rmarinho:
build

Xamarin.Forms.Platform.Android/Platform.cs
```diff
@@ -1134,17 +1134,23 @@ protected override int GetChildDrawingOrder(int childCount, int i)
...
+					if (!_minimumElevation.ContainsKey(i))
 					{
-						_minimumElevation[j + 1] = _minimumElevation[j] + 1;
-						GetChildAt(j + 1).Elevation = _minimumElevation[j + 1];
+						_minimumElevation[i] = GetChildAt(i).Elevation;
+					}
+						
+					for (int j = 0; j < _minimumElevation.Count() - 1; j++)
+					{
+						while (_minimumElevation[j] > _minimumElevation[j + 1])
```
kingces95: We're O(n^3) here right? And won't this condition will always be false in the case where the new elevation has the same value as a previous elevation?

Xamarin.Forms.Platform.Android/Platform.cs
```diff
@@ -1134,17 +1134,23 @@ protected override int GetChildDrawingOrder(int childCount, int i)
 			{
 				//On Material design the button states use Elevation property, we need to make sure
 				//we update the elevation of other controls to be over the previous one
-				if (!_minimumElevation.ContainsKey(i))
-					_minimumElevation[i] = GetChildAt(i).Elevation;
-				for (int j = 0; j < _minimumElevation.Count() - 1; j++)
+				if (Forms.IsLollipopOrNewer)
 				{
-					while (_minimumElevation[j] > _minimumElevation[j + 1])
+					if (!_minimumElevation.ContainsKey(i))
```
kingces95: For this to work, shouldn't we remove this `if` statement? Then when we get a new index, we whack it's elevation into `_minimumElevation`. Then we re-establish strict ordering of `_minimumElement` in the loops below? Then if we don't have too many indices (so we're not worried about the perf of the loops below) then this will work.

Or if `_minimumElevation` already contains the key will the key's value will be the same as the new value? Put another way, will the new elevation always be the same as the old elevation for a given `i`? If so we should assert that fact and forgo the loops below (as we do not modify `_minimumElevation`).

kingces95:
Is it possible to UITest this?

PR #1108, opened 2017-08-18
Fix NRE when removing multiple pages in NavigationPageRenderer

hartez:
…ges with Android Support 25+

### Description of Change ###

The RemovePage method was doing some cleanup of the Page's container and renderer after removing the Page's Fragment; however, that same cleanup is already being done by the FragmentContainer itself. Aside from a very small performance hit, this isn't normally an issue, but when removing multiple pages the additional cleanup code can cause a NullReferenceException. 

This change removes the unnecessary cleanup code from RemovePage.

### Bugs Fixed ###

- [53179 – PopAsync crashing after RemovePage when support packages are updated to 25.1.1](https://bugzilla.xamarin.com/show_bug.cgi?id=53179#c52)

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


rmarinho:
👍

PR #1111, opened 2017-08-22
[UWP] Adjust PR #762 to instead use Element.Focused

pauldipietro:
### Description of Change ###

This is a potential different approach to the bug in #762 which was using `GotFocus` as the hook for focusing on the picker, which could cause issues when doing something such as tabbing between them; the normal UWP behavior doesn't open the dropdown. Basing the behavior on `Element.Focused` instead and ignoring `FocusState.Keyboard` lets the user tab between controls on the screen without opening the Picker. We can also grab the `Popup` (dropdown) part of the `ComboBox` template and use its open/closed events to track its state. This is similar to the original PR since without checking it the dropdown will just repeatedly open due to how focus works.

Related to the `FocusState`, if the user does something such as pressing enter on a button to call `Focus()` on a `Picker`, it will behave similarly to if it was tabbed to as we check for the `Keyboard` type. It's a small edge quirk, but in some ways is consistent with keyboard behavior in general.

### Bugs Fixed ###

#762

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.WinRT/FormsComboBox.cs
+++ b/Xamarin.Forms.Platform.WinRT/FormsComboBox.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Controls.Primitives;
 using Windows.UI.Xaml.Media.Animation;
 
 #if WINDOWS_UWP
@@ -18,6 +19,10 @@ namespace Xamarin.Forms.Platform.WinRT
 		internal bool IsFullScreen => Device.Idiom == TargetIdiom.Phone && Items != null && Items.Count > 5;
 
 		internal bool IsOpeningAnimated { get; private set; }
+
+		internal Popup Popup { get; private set; }
+
+		internal bool IsPopupOpen { get; private set; }
 
 		protected override void OnApplyTemplate()
 		{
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.WinRT/FormsComboBox.cs
```diff
@@ -19,6 +20,10 @@ public class FormsComboBox : ComboBox
 
 		internal bool IsOpeningAnimated { get; private set; }
 
+		internal Popup Popup { get; private set; }
```
samhouts: Looks like this could be `private` instead of `internal`.

Xamarin.Forms.Platform.WinRT/FormsComboBox.cs
```diff
@@ -49,6 +54,12 @@ protected override void OnApplyTemplate()
 					IsClosingAnimated = true;
 				}
 			}
+			Popup = (Popup)GetTemplateChild("Popup");
+			if (Popup != null)
+			{
+				Popup.Opened += (s, e) => IsPopupOpen = true;
```
samhouts: I wonder if we need to unsubscribe these somewhere, too.

pauldipietro: I made that first change mentioned above, but I'm not certain where where we would need to unsubscribe these, if we do.

pauldipietro:
It's likely just a better idea to revert #762 and leave what would be standard platform behavior, so I'm closing this for now.

PR #1113, opened 2017-08-22
[UWP] Adjust setting of Slider's BackgroundColor

pauldipietro:
### Description of Change ###

Setting the `BackgroundColor` on a `Slider` in UWP would also set it on the entire control's `Panel` via `UpdateBackgroundColor` in the `VisualElementRenderer`. Overriding this method on the renderer and removing the latter half of it prevents this from occurring, and only sets the color on the "rail" (`HorizontalTrackRect`) of the `Slider`, which is the correct behavior. No specific reproduction has been added as the behavior can be observed on the Slider Gallery page in the control gallery.

slider

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=57787

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.WinRT/SliderRenderer.cs
+++ b/Xamarin.Forms.Platform.WinRT/SliderRenderer.cs
@@ -71,6 +71,22 @@ namespace Xamarin.Forms.Platform.WinRT
 			}
 		}
 
+		protected override void UpdateBackgroundColor()
+		{
+			if (Control != null)
+			{
+				Color backgroundColor = Element.BackgroundColor;
+				if (!backgroundColor.IsDefault)
+				{
+					Control.Background = backgroundColor.ToBrush();
+				}
+				else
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.WinRT/SliderRenderer.cs
```diff
@@ -71,6 +71,22 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
 			}
 		}
 
+		protected override void UpdateBackgroundColor()
+		{
+			Color backgroundColor = Element.BackgroundColor;
```
samhouts: May as well move this into the `if`

pauldipietro: 👍

PR #1116, opened 2017-08-29
[Android] make NavigationBar support Transluscent status

bentmar:
If the app has flag Transluscentstatus then we should support that when creating the navbar. Now that we dont draw a statusbarunderlay anymore this opens up to many interesting UI possibilities

### Description of Change ###

Adds a padding and adjusts the height of the Navigationbar if Transluscentstatus flag is set to true.

### Bugs Fixed ###

- none

### API Changes ###

none

Changed:
 The way we calculate the height for the navigationbar.
 Adds padding to the navigationbar
Changes are only made if TransluscentStatus = true

### Behavioral Changes ###

Now this is possible
https://forums.xamarin.com/discussion/comment/293844/#Comment_293844
by adding this is MainActivity
```
 Window.AddFlags(WindowManagerFlags.TranslucentStatus);
```

### PR Checklist ###

- [ X] Changes adhere to coding standard
- [ X] Consolidate commits as makes sense

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
@@ -38,6 +38,7 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 		ActionBarDrawerToggle _drawerToggle;
 		FragmentManager _fragmentManager;
 		int _lastActionBarHeight = -1;
+		int _statusbarHeight;
 		AToolbar _toolbar;
 		ToolbarTracker _toolbarTracker;
 		DrawerMultiplexedListener _drawerListener;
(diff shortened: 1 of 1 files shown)


adrianknight89:
I think that if you are adding a padding, you need to account for how some pages (like MDP) are positioned during layout. Also, what happens if the user removes this Window flag through a DependencyService call? Will it get rid of the padding?

bentmar:
> you need to account for how some pages (like MDP) are positioned during layout.
https://forums.xamarin.com/discussion/comment/293844/#Comment_293844
this uses MDP, MDP has nothing to do with how a navigationpages navbar is layed out "?"

If you remove the Transluscent, it wouldn be transluscent anymore (thats all) and the next time a navbar is layed out, it doesnt add padding. 

This is what ive seen from testing on 5 different devices. 

Note that if you use transluscent today your navbar and its content gets hidden under the statusbar = you cant even use it (well you can but...)

adrianknight89:
> If you remove the Transluscent, it wouldn be transluscent anymore (thats all I can see) and the next time a navbar is layed out, it doesnt add padding.

I guess this is what I was getting at. To remove the padding, do you need to wait until the navbar is laid out next time or can you force it to layout once the translucent status is removed? I haven't tested these changes, so I'm only speculating. Maybe it's better to detect translucent changes and force navbar layout?

bentmar:
I see the transluscent status as a "either you have it, or you dont" I have never seen apps that toggles in and out of transluscent during runtime. And if a user were to toggle it on and off, then wouldnt it be possible to handle that forceLayout in a customrenderer?

jimmgarrido:
This will need to be fixed in ContentPage and TabbedPage as well. MasterDetailPage should be fine once the other ones are fixed, but it'll have to be tested with the different combination of Detail/Master pages just to be sure.

I opened up [a bug report](https://bugzilla.xamarin.com/show_bug.cgi?id=59185) with a repro project to track this.

bentmar:
@jimmgarrido 
Your bug report should be built on the latest prerelease of Xamarin forms. 
This does not fix anything in earlier releases since there is a "hardcoded" statusbar.

bentmar:
@jimmgarrido 

About your bug. TransluscentStatus behaviour is expected to layout content behind status bar if I'm not wrong here.

this is exactly what I'm trying to achieve :)
But we have to adjust the navbars layout to still make it look good.

Contentpages with no navbars are supposed to go under the statusbar, check the link I provided too see what I mean

kentcb:
@davidortinau is this going to be merged anytime soon? It's blocking us from upgrading XF. See [here](https://bugzilla.xamarin.com/show_bug.cgi?id=59185).

rmarinho:
@bentmar can you rebase to run some tests please ?

bentmar:
yes, if not today il do it this weekend!

bentmar:
https://github.com/xamarin/Xamarin.Forms/pull/1287 made a new PR that also covers TransluscentNavigation flag, and has some images. (im not a github pro hence a new pr)

rmarinho:
Closing since #1287 replaces this

PR #1120, opened 2017-08-31
[Android] Hide actionbar when MDP with NavigationPage with hidden NavigationBar

rmarinho:
### Description of Change ###

Make sure to respect what NavigationPage says regarding the NavigationBar if we have a MDP with a NavigationPage as a detail. This Works fine in AppCompat but was broken in non AppCompat

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=27731

### PR Checklist ###


samhouts:
@rmarinho We should probably rebase this one again to see if we can get those API19 and PreAppCompat green

PR #1126, opened 2017-09-05
[Core,MacOS]Desktop support

rmarinho:
### Description of Change ###

As Xamarin Forms expands to more backends, MacOS, GTK and WPF, users will need ways to use common features that are related to Desktop enviroment. 
Since XF started as focus on the mobile world, most of these features are not on our product and we will need to find new solutions for those.
Features to implement:

- **App Exit**
- **Right click support**
- **Menus - top menu bar/right click**
- **Menus - Accelerator**

### Bugs Fixed ###

- Provide links to bugs here

### API Changes ###

Changes to Application class, so we can Quit the app and for main menu 
```
public class Application
{
    public Menu MainMenu  { get; }
   
    public void Quit();

}

public interface IPlatformServices
{
    void QuitApplication();  
}
```

Menus
```
public abstract class Element : BindableObject, IElement, INameScope, IElementController 
{ 
     public static readonly BindableProperty MenuProperty = BindableProperty.CreateAttached("Menu", typeof(Menu), typeof(Element), null); 
}

public class Menu : BaseMenuItem, IList<Menu> 
{
    public string Text  {get; set;}
    public void Invalidate() => OnPropertyChanged(); 
    public ObservableCollection<MenuItem> Items  { get; }
}
```

Right click 
```

[Flags]
public enum ButtonsMask
{
    Primary = 1,
    Secondary = 2
}

public sealed class ClickGestureRecognizer : GestureRecognizer
{
    public static readonly BindableProperty CommandProperty = BindableProperty.Create(nameof(Command), typeof(ICommand), typeof(ClickGestureRecognizer), null);
    public static readonly BindableProperty CommandParameterProperty = BindableProperty.Create(nameof(CommandParameter), typeof(object), typeof(ClickGestureRecognizer), null);
    public static readonly BindableProperty NumberOfClicksRequiredProperty = BindableProperty.Create(nameof(NumberOfClicksRequired), typeof(int), typeof(ClickGestureRecognizer), 1);
    public static readonly BindableProperty ButtonsProperty = BindableProperty.Create(nameof(Buttons), typeof(ButtonsMask), typeof(ClickGestureRecognizer), ButtonsMask.Primary);
    public event EventHandler Clicked;
}

public class ClickedEventArgs : EventArgs
{
	public ButtonsMask Buttons { get; private set; }
	public object Parameter { get; private set; }
}


```

Accelerators

```

[TypeConverter(typeof(AcceleratorTypeConverter))]
public class Accelerator
{
	internal Accelerator(string text) { }

	public IEnumerable<string> Modififiers { get; set; }

	public IEnumerable<string> Keys { get; set; }

	public static Accelerator FromString(string text) {}
	public override string ToString()
	{
		return _text;
	}
}

public class AcceleratorTypeConverter : TypeConverter
{
	public override object ConvertFromInvariantString(string value)
	{
		if (value == null)
			return null;
		return Accelerator.FromString(value);
	}
}

public class MenuItem : BaseMenuItem, IMenuItemController
{
     public static readonly BindableProperty AcceleratorProperty = BindableProperty.CreateAttached(nameof(Accelerator), typeof(Accelerator), typeof(MenuItem), null);    
}

```
### Behavioral Changes ###

Platrforms will need to update` IPlatformServices QuitApplication ` method 

### PR Checklist ###


Xamarin.Forms.Core/ClickGestureRecognizer.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+using System.ComponentModel;
+using System.Windows.Input;
+
+namespace Xamarin.Forms
+{
+	[Flags]
+	public enum ButtonsMask
+	{
+		Primary = 1,
```
jassmith: 1 << 0;
1 << 1;

I personally find this easier to read for flags, no biggie.

Xamarin.Forms.Core/ClickGestureRecognizer.cs
```diff
@@ -0,0 +1,65 @@
...
+			set { SetValue(ButtonsProperty, value); }
+		}
+
+		public event EventHandler Clicked;
+
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public void SendClicked(View sender, ButtonsMask buttons)
+		{
+			ICommand cmd = Command;
+			if (cmd != null && cmd.CanExecute(CommandParameter))
```
jassmith: CommandParameter fetched with multiple GetValue

Xamarin.Forms.Core/Application.cs
```diff
@@ -49,6 +57,8 @@ public IAppLinks AppLinks
 			}
 		}
 
+		public Menu MainMenu => _menus;
```
jassmith: Maybe this should be an attached property as it wont really be serviced on all platforms

Xamarin.Forms.Core/Menu.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace Xamarin.Forms
+{
+	public class Menu : BaseMenuItem, IList<Menu>
+	{
+		readonly ObservableCollection<Menu> _menus = new ObservableCollection<Menu>();
+
+		ObservableCollection<MenuItem> _items = new ObservableCollection<MenuItem>();
```
jassmith: readonly, this way we know we dont need to unhook the event

Xamarin.Forms.Platform.Android/Forms.cs
```diff
@@ -486,6 +486,11 @@ static bool TryGetTextAppearance(int appearance, out double val)
 				return false;
 			}
 
+			public void QuitApp()
+			{
+				Internals.Log.Warning(nameof(AndroidPlatformServices), "Platform doesn't implement QuitApp");
```
jassmith: I mean... we could crash it... on purpose (kidding!)

Xamarin.Forms.Core/MenuItem.cs
```diff
@@ -6,6 +6,12 @@ namespace Xamarin.Forms
 {
 	public class MenuItem : BaseMenuItem, IMenuItemController
 	{
+		public static readonly BindableProperty AcceleratorProperty = BindableProperty.CreateAttached("Accelerator", typeof(string), typeof(MenuItem), null);
```
jassmith: I was thinking about the future of this API a bit and I think we need to be slightly more clever.

public class Accelerator { // name not good
  public static Accelerator FromString (string accel);
}

This way if we ever need to add an enum (god forbid) based API in the future we can.

Xamarin.Forms.Core/IPlatformServices.cs
```diff
@@ -32,5 +32,7 @@ public interface IPlatformServices
 		void StartTimer(TimeSpan interval, Func<bool> callback);
 
 		string RuntimePlatform { get; }
+
+		void QuitApp();
```
StephaneDelcroix: that'd be the first time (to the best of my knowledge) that we abbreviate Application into App in an API. Go for `QuitApplication` or `ExitApplication` instead.

Xamarin.Forms.Core/Menu.cs
```diff
@@ -0,0 +1,95 @@
...
+{
+	public class Menu : BaseMenuItem, IList<Menu>
+	{
+		readonly ObservableCollection<Menu> _menus = new ObservableCollection<Menu>();
+
+		readonly ObservableCollection<MenuItem> _items = new ObservableCollection<MenuItem>();
+
+		public Menu()
+		{
+			_items.CollectionChanged += (s, e) => OnPropertyChanged(nameof(Items));
```
StephaneDelcroix: I guess this is a temporary hack, but it redraws the entire menu

rmarinho: yeah it's true

Xamarin.Forms.Core/Accelerator.cs
```diff
@@ -0,0 +1,86 @@
...
+
+		internal Accelerator(string text)
+		{
+			if (string.IsNullOrEmpty(text))
+				throw new ArgumentNullException(nameof(text));
+			_text = text;
+		}
+
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public IEnumerable<string> Modififiers { get; set; }
```
alanmcgovern: typo :)

rmarinho: 🤦‍♂️

MarkoBL:
Please add something like tab order to the desktop platform (like I did for macOS https://github.com/xamarin/Xamarin.Forms/pull/1060), it's just so common to tab through entries and buttons.

Window manipulation should be added too (size, position, minimize, maximize, enterfullscreen) and the size of the whole screen. On macOS/UWP, I save the last position/size and when the user starts the app again, the window appears at the same position.

rogihee:
@MarkoBL Tab and focus is on their list, see the vNext roadmap:

**Accessibility Improvements**
Add direct support for focus order and tab index.

Xamarin.Forms.Platform.MacOS/Extensions/NSMenuExtensions.cs
```diff
@@ -0,0 +1,106 @@
...
+
+			bool hasModifierMask = accelerator.Modififiers.Count() > 1;
+
+			if (hasModifierMask)
+			{
+				nsMenuItem.KeyEquivalentModifierMask = 0;
+
+				for (int i = 0; i < accelerator.Modififiers.Count(); i++)
+				{
+					var modifierMast = accelerator.Modififiers.ElementAt(i).ToLower();
```
StephaneDelcroix: you mean `Mask`?

rmarinho: yes thanks

TheSamHughes:
Its macOS not MacOS, not sure why this seems to have been a requested change on this pull request, the file names are also incorrectly cased...

PR #1128, opened 2017-09-06
[Core] Removed rounding in AbsoluteLayout

ytn3rd:
### Description of Change ###
Removed rounding in AbsoluteLayout to ensure Android devices with certain sizes can still use proportional sizes to fill the device screen.

### Bugs Fixed ###
[https://bugzilla.xamarin.com/show_bug.cgi?id=40092]( https://bugzilla.xamarin.com/show_bug.cgi?id=40092)

### API Changes ###
None

### Behavioral Changes ###

People who previously had a 1px gap when using AbosluteLayouts with proportional x,y,width,height on Android will now potentially have their layouts become 1px different in size or position on certain devices. This shouldn't be a breaking change as previously they would have had a white line instead.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/AbsoluteLayout.cs
+++ b/Xamarin.Forms.Core/AbsoluteLayout.cs
@@ -224,7 +224,7 @@ namespace Xamarin.Forms
 
 			if (widthIsProportional)
 			{
-				result.Width = Math.Round(region.Width * bounds.Width);
+				result.Width = region.Width * bounds.Width;
 			}
 			else if (bounds.Width != AutoSize)
 			{
@@ -233,7 +233,7 @@ namespace Xamarin.Forms
 
 			if (heightIsProportional)
 			{
-				result.Height = Math.Round(region.Height * bounds.Height);
+				result.Height = region.Height * bounds.Height;
 			}
 			else if (bounds.Height != AutoSize)
 			{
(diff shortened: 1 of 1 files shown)


jassmith:
This is actually really close to right. Thought not exactly. What needs to be done is a little more subtle and thank you for this PR I failed to see it forever.

The right fix is to have a "Forms.Round" (naming not withstanding) method that rounds to the nearest device pixel. On iOS/UWP this is the same as Math.Round, on android however it is not! Android has pixel boundaries all over the place, so we need to DI in a round function that can round to the nearest pixel boundary, not the nearest integer.

ytn3rd:
What do you mean by DI?

Would the solution then be on Android to multiple the values by device scale, then round, then divide by device scale?
So device scale being 3, input value being 527.3 -> 1581.9 -> 1581 -> 527

Or from the example given in the bug report above with scale being 3.5

> 411.42857143 -> 1440.00 -> 1440.00 -> 411.42857143
> 411.52857143 -> 1440.35 -> 1440.00 -> 411.42857143
> 411.62857143 -> 1440.70 -> 1441.00 -> 411.7142857

Forms.DeviceRound or Forms.DisplayRound would be better IMO as .Round does not sound like its for screen display. Is there any specific display/screen classes it could go into?

jassmith:
Sorry for delay. Yes thats exactly what I mean. It needs to round to the nearest device pixel. And yes I agree the name I came up with on the fly was terrible, your names are better.

PaulVrugt:
any activity on this? Still waiting for the fix to be included an a xamarin.forms release

ytn3rd:
Sorry, nothing form my end @PaulVrugt .

PureWeen:
Closing in favor of
https://github.com/xamarin/Xamarin.Forms/pull/1761

PR #1130, opened 2017-09-06
[Android] Enables custom images (especially SVG) through the new IImageSourceHandler

softlion:
### Description of Change ###

Enables custom images (especially SVG) through the new IImageSourceHandler to work with tab bar icons on Android (like it does on iOS). And also on cell adapters and navigation page menu.

No test: there is no existing test for IImageSourceHandler (or can't find where).

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/ResourceManager.cs
+++ b/Xamarin.Forms.Platform.Android/ResourceManager.cs
@@ -24,6 +24,11 @@ namespace Xamarin.Forms.Platform.Android
 			if(drawable == null)
 			{
 				var bitmap = GetBitmap(resource, file) ?? BitmapFactory.DecodeFile(file);
+				if (bitmap == null)
+				{
+					var source = Registrar.Registered.GetHandler<IImageSourceHandler>(fileImageSource.GetType());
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -24,6 +24,11 @@ internal static Drawable GetFormsDrawable(this Resources resource, FileImageSour
 			if(drawable == null)
 			{
 				var bitmap = GetBitmap(resource, file) ?? BitmapFactory.DecodeFile(file);
+				if (bitmap == null)
+				{
+					var source = Registrar.Registered.GetHandler<IImageSourceHandler>(fileImageSource.GetType());
+					bitmap = source.LoadImageAsync(fileImageSource, Forms.Context).GetAwaiter().GetResult();
+				}
```
rmarinho: won't this lock the UI thread ?

softlion: Less than GetBitmap/DecodeFile.
It is coded like this on iOS already. 
This would make Android as extensible as iOS.

PR #1133, opened 2017-09-10
[UWP] Use ItemClick to re-enable use of enter key for selection on ListView

pauldipietro:
### Description of Change ###

A previous change involving the accessibility implementation had an effect which prevented the enter key from firing `ItemTapped`. We are able to instead use `IsItemClickEnabled` and use earlier behavior to make sure that the selected item is updated correctly.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=59248

### API Changes ###

None

### Behavioral Changes ###

None expected.

### PR Checklist ###


samhouts:
Note: this was in 2.4.0-pre3, but it was reverted from 2.4.0 before the next version was released. It will be addressed in the following version.

PR #1136, opened 2017-09-12
[C, iOS, Android] LayoutCompression

StephaneDelcroix:
### Description of Change ###

Enable Layout Compression for iOS and Android.

**NOTE: this feature is still in beta, and opt-in only**

Layout Compression allows the user to avoid creating renderers for Layouts (i.e. StackLayout, ContentView, ...).

Pros:
 - less renderers to create, manage and dispose
 - less UI elements on screen

Cons:
 - compressed layouts can't have BG colors, gesture recognizers
 - no transformation
 - ...

Sample:
```
	x:Class="Xamarin.Forms.Controls.CompressedLayout"
	Padding="0,20,0,0">
	<StackLayout Padding="6" CompressedLayout.IsHeadless="true">
		<Label Text="First"/>
		<StackLayout Padding="6" CompressedLayout.IsHeadless="true">
			<Label Text="Second"/>
			<ContentView Padding="6" CompressedLayout.IsHeadless="true" BackgroundColor="Pink" >
				<Label x:Name="this" Text="Third"/>
			</ContentView>
			<Label Text="Fourth"/>
		</StackLayout>
		<Label Text="Fifth"/>
	</StackLayout>
</ContentPage>
```
Enabling LayoutCompression for this layout only creates a Renderer for the page, and one for each label. The BackgroundColor property on the inner ContentView is ignored.

### Bugs Fixed ###

/

### API Changes ###

Added:
```
public static readonly BindableProperty CompressedLayout.IsHeadlessProperty;
```

### Behavioral Changes ###

/

### PR Checklist ###


Xamarin.Forms.Platform.Android/VisualElementPackager.cs
```diff
@@ -18,11 +18,16 @@ public class VisualElementPackager : IDisposable
 
 		IVisualElementRenderer _renderer;
 
-		public VisualElementPackager(IVisualElementRenderer renderer)
+		VisualElement _element;
+
+		IElementController ElementController => _element;// _renderer.Element as IElementController;
```
samhouts: Remove commented code

Xamarin.Forms.Core/Layout.cs
```diff
@@ -237,6 +237,13 @@ protected void UpdateChildrenLayout()
 			double w = Math.Max(0, width - Padding.HorizontalThickness);
 			double h = Math.Max(0, height - Padding.VerticalThickness);
 
+			var isHeadless = CompressedLayout.GetIsHeadless(this);
+			var headlessOffset = CompressedLayout.GetHeadlessOffset(this);
+			for (var i = 0; i < LogicalChildrenInternal.Count; i++) {
```
samhouts: Braces on their own lines. You're in Core territory, now! muwahaha

Xamarin.Forms.Platform.Android/Renderers/LocalizedDigitsKeyListener.cs
```diff
@@ -34,7 +34,9 @@ public static NumberKeyListener Create(InputTypes inputTypes)
 			if ((inputTypes & InputTypes.NumberFlagDecimal) == 0)
 			{
 				// If decimal isn't allowed, we can just use the Android version
+#pragma warning disable 0618 // Find issues with Android API usage
```
hartez: Which part of the next line is giving an obsolete warning?

Xamarin.Forms.Platform.Android/VisualElementPackager.cs
hartez: Can probably remove this comment.

Xamarin.Forms.Platform.iOS/RendererPool.cs
```diff
@@ -108,10 +108,14 @@ void FillChildrenWithRenderers(VisualElement element)
 				var child = logicalChild as VisualElement;
 				if (child != null)
 				{
+					if (CompressedLayout.GetIsHeadless(child)) {
+						child.IsPlatformEnabled = true;
+						FillChildrenWithRenderers(child);
+					} else {
```
hartez: Missing a tab on the next line.

Xamarin.Forms.Platform.Android/VisualElementPackager.cs
```diff
@@ -67,44 +70,49 @@ protected virtual void Dispose(bool disposing)
...
 		}
 
 		void AddChild(VisualElement view, IVisualElementRenderer oldRenderer = null, RendererPool pool = null, bool sameChildren = false)
 		{
 			Performance.Start();
 
-			if (_childViews == null)
-				_childViews = new List<IVisualElementRenderer>();
+			if (CompressedLayout.GetIsHeadless(view)) {
+				var _packager = new VisualElementPackager(_renderer, view);
```
samhouts: the underscore in the variable name is a little confusing here, I think.

Xamarin.Forms.Platform.iOS/VisualElementPackager.cs
```diff
@@ -13,16 +13,20 @@ public class VisualElementPackager : IDisposable
 
 		bool _isDisposed;
 
-		IElementController ElementController => Renderer.Element as IElementController;
+		IElementController ElementController => _element;// Renderer.Element as IElementController;
```
hartez: Can probably remove this comment.

adrianknight89:
> compressed layouts can't have BG colors, gesture recognizers
> no transformation

Are these limitations permanent or to be fixed in the future?

`<ContentView Padding="6" CompressedLayout.IsHeadless="true" BackgroundColor="Pink" >`

Should there be a compile-time / run-time error for this instead of silent fail of the bgcolor set? The reason I ask this is I don't like it when people pollute their visual tree with redundant stuff. If bgcolor has no use there, then it shouldn't be there?

StephaneDelcroix:
@adrianknight89 at this point we have no plan for supporting BgColor on headless layouts, nor do we have plans for validating that case

Xamarin.Forms.Platform.Android/Renderers/LocalizedDigitsKeyListener.cs
StephaneDelcroix: On X.A 7.5.0.15 (Alpha channel at this time): 'DigitsKeyListener.GetInstance(bool, bool)' is obsolete

```
[Register ("getInstance", "(ZZ)Landroid/text/method/DigitsKeyListener;", ""), Obsolete ("deprecated")]
public unsafe static DigitsKeyListener GetInstance (bool sign, bool @decimal);
```

Xamarin.Forms.Platform.Android/VisualElementPackager.cs
StephaneDelcroix: fixed

Xamarin.Forms.Platform.iOS/RendererPool.cs
StephaneDelcroix: fixed

Xamarin.Forms.Platform.iOS/VisualElementPackager.cs
StephaneDelcroix: fixed

Xamarin.Forms.Platform.Android/VisualElementPackager.cs
StephaneDelcroix: fixed

Xamarin.Forms.Platform.Android/VisualElementPackager.cs
StephaneDelcroix: fixed. at some point that variable was a field...

Xamarin.Forms.Core/Layout.cs
StephaneDelcroix: don't scratch that itch! :)

fixed by removing the braces.

Xamarin.Forms.Platform.Android/Renderers/LocalizedDigitsKeyListener.cs
StephaneDelcroix: fixed in master

alexsorokoletov:
@StephaneDelcroix thank you for this PR and this feature. Wondering, where can one read about limitations and why they are there. Maybe it's just me can't find the information

bentmar:
Can we please have some more information on how to use this?

davidortinau:
@bentmar I'll have a blog out shortly with more samples, and then docs to come. Really any layout that you don't need a background on, doesn't transform, and doesn't accept gestures is a candidate. So just add `CompressedLayout.IsHeadless="true"` to it. I suspect we may discover some other cons as we test further. So far the reduction of renderers created is looking good, especially if you have lots of layout nesting. In my page I saw nearly a 50% reduction, though granted it was a poorly constructed page in many respects to begin with.

alexsorokoletov:
https://blog.xamarin.com/3-big-things-explore-xamarin-forms-2-5-0-pre-release/

BlueRaja:
Seems like this should be on by default, and only disabled when there's a background/transform/gesture/explicitly disabled?

PR #869, opened 2017-04-14
[Evolution] Merged Dictionaries

adamped:
### Description of Change ###

The addition of Multiple MergedDictionaries into ResourceDictionary, as discussed in the Evolution forum: https://forums.xamarin.com/discussion/86308/multiple-merged-dictionaries. Allows multiple ResourceDictionaries to be added to the VisualElement root Resources property.

### Bugs Fixed ###

- N/A

### API Changes ###

Added:
 - ICollection<ResourceDictionary> MergedDictionaries // In ResourceDictionary.cs

### Behavioral Changes ###

The addition, of adding multiple ResourceDictionaries into a ResourceDictionary. No behavioral changes to existing functionality were made.

### PR Checklist ###

- [X] Has tests (if omitted, state reason in description)
- [X] Rebased on top of master at time of PR
- [X] Changes adhere to coding standard
- [X] Consolidate commits as makes sense

PR #1138, opened 2017-09-13
[Evolution] Merged Dictionaries

StephaneDelcroix:
**NOTE0: this PR replaces #869**
**NOTE1: Do not squash merge to preserve author information. Thanks**

### Description of Change ###

The addition of Multiple MergedDictionaries into ResourceDictionary, as discussed in the Evolution forum: https://forums.xamarin.com/discussion/86308/multiple-merged-dictionaries. Allows multiple ResourceDictionaries to be added to the VisualElement root Resources property.

### Bugs Fixed ###

- N/A

### API Changes ###

Added:
 - ICollection<ResourceDictionary> MergedDictionaries // In ResourceDictionary.cs

### Behavioral Changes ###

The addition, of adding multiple ResourceDictionaries into a ResourceDictionary. No behavioral changes to existing functionality were made.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/ResourceDictionary.cs
+++ b/Xamarin.Forms.Core/ResourceDictionary.cs
@@ -14,8 +14,9 @@ namespace Xamarin.Forms
 	{
 		static ConditionalWeakTable<Type, ResourceDictionary> s_instances = new ConditionalWeakTable<Type, ResourceDictionary>();
 		readonly Dictionary<string, object> _innerDictionary = new Dictionary<string, object>();
-
+		ResourceDictionary _mergedInstance;
 		Type _mergedWith;
+
 		[TypeConverter (typeof(TypeTypeConverter))]
 		public Type MergedWith {
 			get { return _mergedWith; }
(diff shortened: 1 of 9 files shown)


StephaneDelcroix:
@adamped I did some minor fixes here and there, added tests and support for XamlC. give me a thumbs up if you think it's good

adamped:
@StephaneDelcroix - why the removal of the Reset event in the CollectionChanged? If the developer, in code behind, for some reason, does this.Resources.MergedDictionaries.Clear();

It then triggers a Reset, then none of the OldItems will have their OnValueChanged event hooks removed, because the OldItems isn't populated on a Reset.

StephaneDelcroix:
@adamped trust your unit tests more. the Reset code is there https://github.com/xamarin/Xamarin.Forms/pull/1138/files#diff-210a3a9d3639265f22acd446c2e1d050R61

adamped:
@StephaneDelcroix so it is :) All good then.

opcodewriter:
Finally, it took ages for Xamarin to realize this is a much needed feature. I wonder how many more years will it need Xamarin to bring other basic\core features from Windows XAML to Xamarin Forms?

Xamarin.Forms.Core/ResourceDictionary.cs
```diff
@@ -5,19 +5,22 @@
...
 	public class ResourceDictionary : IResourceDictionary, IDictionary<string, object>
 	{
 		static ConditionalWeakTable<Type, ResourceDictionary> s_instances = new ConditionalWeakTable<Type, ResourceDictionary>();
 		readonly Dictionary<string, object> _innerDictionary = new Dictionary<string, object>();
-
+		ResourceDictionary _mergedInstance;
 		Type _mergedWith;
+
 		[TypeConverter (typeof(TypeTypeConverter))]
 		public Type MergedWith {
```
jassmith: Maybe we should deprecate this?

StephaneDelcroix: we will, but we need to implement `Source` for that

Makeloft:
Hi, guys!
Seems, that your implementation of MergedDictionaries functionality is not optimal.
Possible, it should looks like below.

```
    public class ResourceDictionary : IResourceDictionary, IDictionary<string, object>
    {
        public List<ResourceDictionary> MergedDictionaries { get; } =
            new List<ResourceDictionary>();

        public virtual object this[string key] => TryGetValue(key, out var value)
            ? value
            : throw new KeyNotFoundException(key);

        public virtual bool TryGetValue(string key, out object result)
        {
            var containsKey =
                base.TryGetValue(key, out var value) ||
                MergedDictionaries.LastOrDefault(d => d.TryGetValue(key, out value)) != null;

            result = value;
            return containsKey;
        }

        public virtual bool ContainsKey(string key) =>
            base.ContainsKey(key) ||
            MergedDictionaries.LastOrDefault(d => d.ContainsKey(key)) != null;
    }
```
At the current moment it is possible to make workable Custom Resource Dictionary based on INotifyCollectionChanged pattern (see: https://github.com/Makeloft/Ace/blob/master/Ace.Zest/Markup/ResourceDictionary.cs).

But it not possible use optimal approach because current implementation of  IResourceDictionary and IDictionary<string, object> into ResourceDictionary class is not virtual and it is not opportunity to override.

What do you think about this?

Makeloft:
For example you may rename current Xamarin.Forms.ResourceDictionary to Xamarin.Forms.ResourceDictionaryBase, made it internal and override by using "new" keyword implementation of IResourceDictionary and IDictionary<string, object>.

```
// first step: rename Xamarin.Forms.ResourceDictionary into Xamarin.Forms.ResourceDictionaryBase
// second step: override IResourceDictionary and IDictionary<string, object>

    public class ResourceDictionary : Xamarin.Forms.ResourceDictionaryBase
    {
        public List<ResourceDictionary> MergedDictionaries { get; } =
            new List<ResourceDictionary>();

        public new object this[string key] => TryGetValue(key, out var value)
            ? value
            : throw new KeyNotFoundException(key);

        public new bool TryGetValue(string key, out object result)
        {
            var containsKey =
                base.TryGetValue(key, out var value) ||
                MergedDictionaries.LastOrDefault(d => d.TryGetValue(key, out value)) != null;

            result = value;
            return containsKey;
        }

        public new bool ContainsKey(string key) =>
            base.ContainsKey(key) ||
            MergedDictionaries.LastOrDefault(d => d.ContainsKey(key)) != null;
    }
```

Makeloft:
Possible, it makes sense to review this PR about Merged Dictionaries https://github.com/xamarin/Xamarin.Forms/pull/1218

Thanks!

Sokoo92:
Hey guys, loving this feature, I'm using it in my application for brand theming. The only problem I have is that it doesn't work on iOS :( I use a build configuration for each brand in my application.
`#if BRAND1
            Resources.MergedDictionaries.Add(new BrandTheme1());
            brandName = "Brand1";
#endif

#if BRAND2
            Resources.MergedDictionaries.Add(new BrandTheme2());
            brandName = "Brand2";
#endif`
And I originally keep my App.xaml resource dictionary empty and Merge the themes I want at build this in my App constructor in App.xaml.cs. But when I run it on iOS it says there's no such key as "Primary" (a primary color in my app). So I thought I'd fill the Dictionary and let the MergedDictionaries override the styles at buildtime. But it doesn't. it keeps the styles I put in App.xaml ResourceDictionary and doesn't merge with the ones I tell it to. 

Everything works fine for Android, keeping my App.xaml empty, it merges the correct dictionary at buildtime, just iOS doesn't want to do that for some reason. Anyone knows why?

StephaneDelcroix:
@Sokoo92 please open an issue, with a very small reproduction project.

Sokoo92:
@StephaneDelcroix Okay just opened one :) https://github.com/xamarin/Xamarin.Forms/issues/2037

devmikew:
what is the syntax of the soure uri for dictionaries located in another assembly?

PR #1139, opened 2017-09-13
[Android] Make sure to clean the element reference

rmarinho:
### Description of Change ###

Fix a possible leak on the VisualElementPackager with CompressedLayout changes 

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/VisualElementPackager.cs
+++ b/Xamarin.Forms.Platform.Android/VisualElementPackager.cs
@@ -67,8 +67,7 @@ namespace Xamarin.Forms.Platform.Android
 				_renderer = null;
 			}
 
-			if (_element != null)
-				_element = null;
+			_element = null;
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.Android/VisualElementPackager.cs
```diff
@@ -66,6 +66,9 @@ protected virtual void Dispose(bool disposing)
 				}
 				_renderer = null;
 			}
+
+			if (_element != null)
```
bugadani: Why check? Just null it

rmarinho: you are right , is was doing some unsubscribe and then i removed that code..

rmarinho: will fix before merge..

PR #1142, opened 2017-09-15
[UWP] Use RawPixelsPerViewPixel for ScalingFactor

pauldipietro:
### Description of Change ###

UWP contains APIs relating to scale that aren't available on 8.1, namely `RawPixelsPerViewPixel` which simplifies getting the return value that we can use for the platform instead of checking `ResolutionScale` values.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=59404

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/WindowsDeviceInfo.cs
+++ b/Xamarin.Forms.Platform.UAP/WindowsDeviceInfo.cs
@@ -45,6 +45,9 @@ namespace Xamarin.Forms.Platform.UWP
 		{
 			get
 			{
+#if WINDOWS_UWP
+				return DisplayInformation.GetForCurrentView().RawPixelsPerViewPixel;
+#else
 				ResolutionScale scale = _information.ResolutionScale;
 				switch (scale)
 				{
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.WinRT/WindowsDeviceInfo.cs
```diff
@@ -51,6 +51,9 @@ public override double ScalingFactor
 		{
 			get
 			{
+#if WINDOWS_UWP
+				return DisplayInformation.GetForCurrentView().RawPixelsPerViewPixel;
```
dotMorten: I think it would be better to use this for both platforms:
   `((int)_information.ResolutionScale) / 100d`

That ensures if the enum gets more options, it'll still work.

jassmith: I agree

rmarinho:
Hey @pauldipietro needs rebase too , sorry. 
Thanks

PR #1143, opened 2017-09-15
Add iOS prototypical cell cache for LV RowHeight calcs

kingces95:
### Description of Change ###

Expect when an iOS `ListView` with `HasUnevenRows` computes the row height for an appearing cell, that does not require creating a new iOS view. Actually, we've been creating a new view every time we compute the height. 

The fix is to create an iOS cell (prototypical cell) for each type of `ViewCell`, then when the height of a `ViewCell` of that type needs to be computed, get (or create) the prototypical cell, update it's BindingContext, measure its height, and return that result.

How the type of the `ViewCell` is determined depends on the caching strategy. 

* RecycleElement -> the `ViewCell` type key is the `DataTemplate`. 

* RecycleElementAndDataTemplate -> the `ViewCell` type is the CLR `Type` of the binding context retrieved out of the `ItemsSource`. This differs from `RecycleElement` in that we need not keep asking the `DataTemplateSelector` for a `DataTemplate` as we'll assume it'll return the same `DataTemplate` for a given item type.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=52487

### API Changes ###

Add `CachingStrategy.RecycleElementAndDataTemplate`

### Behavioral Changes ###

For `RecycleElement`, we now assume every `DataTemplate` will return the same type of `ViewCell`. This is a breaking change which, we hope, won't bite anyone.

### PR Checklist ###

   - We still have Android failures which block the test. But we want to get this fix out sooner.

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
@@ -25,7 +25,6 @@ namespace Xamarin.Forms.Platform.iOS
 		KeyboardInsetTracker _insetTracker;
 		RectangleF _previousFrame;
 		ScrollToRequestedEventArgs _requestedScroll;
-		bool _shouldEstimateRowHeight = true;
 
 		FormsUITableViewController _tableViewController;
 		ListView ListView => Element;
@@ -212,7 +211,6 @@ namespace Xamarin.Forms.Platform.iOS
 						Control.SetContentOffset(offset, true);
 					});
 				}
-				_shouldEstimateRowHeight = true;
 
 				var listView = e.NewElement;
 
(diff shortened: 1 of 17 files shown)


Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -745,7 +744,7 @@ internal nfloat CalculateHeightForCell(UITableView tableView, Cell cell)
 
 					// Let the EstimatedHeight method know to use this value.
 					// Much more efficient than checking the value each time.
-					_useEstimatedRowHeight = true;
+					//_useEstimatedRowHeight = true;
```
rmarinho: if we don't need this maybe delete

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -366,35 +364,36 @@ void OnScrollToRequested(object sender, ScrollToRequestedEventArgs e)
...
+			if (!Element.HasUnevenRows || Element.RowHeight != -1)
 			{
 				Control.EstimatedRowHeight = 0;
 				_estimatedRowHeight = true;
+				return;
 			}
+
+			var source = _dataSource as UnevenListViewDataSource;
+
+			// We want to make sure we reset the cached defined row heights whenever this is called.
```
rmarinho: can we remove old code? Did we make sure we don't regress 43313 ??

rmarinho:
add  https://bugzilla.xamarin.com/show_bug.cgi?id=59489 as fixed too

PR #1149, opened 2017-09-19
[Android] Use bitwise ops on caching strategy in ListViewAdapter.cs

kingces95:
Expect all ListViewCachingStrategy tests to use bitwise operations after changing `ListViewCachingStrategy` enum to a flag. Actually, while trying to carve Android work out of the iOS work I cut too deeply and missed a few callsites in `ListViewAdapter.cs`. 

Continuation of #1143.


dgeller-OUHSC:
just FYI, on enum bitwise comparisons there's a nice hasflag option, IE:
if (cachingStrategy.HasFlag(ListViewCachingStrategy.RetainElement))

PR #1151, opened 2017-09-20
Prevent Flags/IsInitialized check from crashing Android app on restart

hartez:
### Description of Change ###

The first call to the `Forms.SetFlags` must come before `Forms.Init`; otherwise it throws an `InvalidOperationException`. However, Android apps can restart (and call `OnCreate` again) due to configuration changes (e.g., plugging in a hardware keyboard). The second call to `OnCreate` (and subsequent second call to `SetFlags`) was triggering the `InvalidOperationException` after a configuration change and crashing the application.

This change ensures that only the first call to `SetFlags` must come before `Init`; subsequent calls are ignored.

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.ControlGallery.Android/Activity1.cs
+++ b/Xamarin.Forms.ControlGallery.Android/Activity1.cs
@@ -165,7 +165,7 @@ namespace Xamarin.Forms.ControlGallery.Android
 					uiModeManager.DisableCarMode(DisableCarModeFlags.None);
 				}
 				
-				uiModeManager.EnableCarMode(EnableCarModeFlags.GoCarHome);
+				uiModeManager.EnableCarMode(EnableCarModeFlags.None);
 
 				// And put things back to normal so we can keep running tests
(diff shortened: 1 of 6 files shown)


Xamarin.Forms.ControlGallery.Android/Activity1.cs
```diff
@@ -146,6 +147,30 @@ public void Reset()
...
+			{
+				// We can force a restart by making a configuration change; in this case, we'll enter
+				// Car Mode. (The easy way to do this is to change the orientation, but ControlGallery
+				// handles orientation changes so they don't cause a restart.)
+
+				var uiModeManager = UiModeManager.FromContext(this);
+
+				if (uiModeManager.CurrentModeType == UiMode.TypeCar)
+				{
+					// If for some reason we're already in card mode, disable it
```
samhouts: car mode, not card mode

rmarinho:
@hartez Test is failing on LegacyRenderers

PR #1152, opened 2017-09-21
[Core] Allows setting SelectedItem before the ItemsSource, and compare with Equals() **Breaking change**

StephaneDelcroix:
### Description of Change ###

 - Allows setting the LV.SelectedItem before the ItemsSource (async load)
 - Compare items using `Equals()` instead of ReferenceEquals (to stay consistent with `List.IndexOf()`)

This is:
 - potentially breaking
 - not tested enough
  
### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=58451

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###


MelbourneDeveloper:
Thanks for looking in to this. These changes look good. However, I must point out that there still might be an issue here. It's now possible to set the ItemsSource after the SelectedItem. However, on the face of it, it seems that if the list inside the ItemsSource changes later, a refresh will not occur. I think that the code needs to add an event handler for CollectionChanged on INotifyCollectionChanged collections. 

If you have a look at line 72 here: https://github.com/MelbourneDeveloper/Adapt.Presentation/blob/master/Adapt.Presentation.Standard/Adapt/Presentation/Controls/AdaptListView.cs , you can see what we've had to do to make this work.

StephaneDelcroix:
@MelbourneDeveloper I considered that case, and left it for later. feel free to send a PR

StephaneDelcroix:
@jassmith docs fixed. please review

MelbourneDeveloper:
Was this pull request ever merged? Any plans to merge it?

jassmith:
I think with the breaking behavioral change we will have to let this one go :/ Even our own regression suite recognizes this is a break too far.

PR #1153, opened 2017-09-21
[Android] Prototypical Cell Cache for IsEnabled testing; UITest

kingces95:
### Description of Change ###

Expect checking if a cell is enabled does not always create a new view. Actually, before the stop gap fix, we had been creating a new view, binding, then checking if the view is enabled. As views are checked all the time to see if they are enabled this was a big leak. 

The stop gap fix, which this fix replaces, was to create and cache a new view, `_enabledCheckCell`, and rebind that view to determine if a cell is enabled. That worked fine so long as all cells used the same type of view to render themselves. However, when a `DataTemplateSelector` is used, then different cells can use different views so its not sufficient to cache a single view. 

This fix caches the correct view for a given type of cell which can be repeatedly rebound and checked to see if it's enabled. How the type of a cell is determined depends on the caching strategy.  

### Bugs Fixed ###

None; Discovered while fixing iOS caching.

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/ListViewAdapter.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ListViewAdapter.cs
@@ -9,6 +9,7 @@ using Android.Widget;
 using AView = Android.Views.View;
 using AListView = Android.Widget.ListView;
 using Xamarin.Forms.Internals;
+using System.Collections;
 
 namespace Xamarin.Forms.Platform.Android
 {
@@ -32,7 +33,7 @@ namespace Xamarin.Forms.Platform.Android
 		// To prevent a conflict in the event that a ListView supports both templates and non-templates, we will start the DataTemplate key at 2.
 
 		int _listCount = -1; // -1 we need to get count from the list
-		Cell _enabledCheckCell;
+		Dictionary<object, Cell> _prototypicalCellByTypeOrDataTemplate;
 
 		bool _fromNative;
 		AView _lastSelected;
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.Android/Renderers/ListViewAdapter.cs
```diff
@@ -309,6 +311,68 @@ public override AView GetView(int position, AView convertView, ViewGroup parent)
 			return layout;
 		}
 
+		internal void InvalidatePrototypicalCellCache()
+		{
+			_prototypicalCellByTypeOrDataTemplate.Clear();
```
kingces95: TODO: Invalidate the cache if the `DataTemplateSelector` changes. For now, just wanna get this in while I have a CI baseline.

Xamarin.Forms.Platform.Android/Renderers/ListViewAdapter.cs
```diff
@@ -309,6 +311,68 @@ public override AView GetView(int position, AView convertView, ViewGroup parent)
...
+				// cache prototypical cell by item type; Items of the same Type share
+				// the same DataTemplate (this is enforced by RecycleElementAndDataTemplate)
+				protoCell = GetCellForPosition(indexPath);
+				_prototypicalCellByTypeOrDataTemplate[itemTypeOrDataTemplate] = protoCell;
+			}
+
+			var templatedItems = GetTemplatedItemsListForPath(indexPath);
+			return templatedItems.UpdateContent(protoCell, indexPath);
+		}
+
```
kingces95: Will rename `indexPath` to `index` and `ForPath` to `ForIndex` before merging. For now, leave as is so as not to trigger another test run.

PR #1155, opened 2017-09-23
[Android] Prevent ObjectDisposedExceptions in ListViews with Header/FooterTemplates

samhouts:
### Description of Change ###

A continuation of #1063, this will do the same work for `HeaderTemplate` and `FooterTemplate`. Additionally, calling `RemoveAllViews` before calling `Dispose` is a terrible thing and leaves orphaned views behind. Also using a `_disposed` bool instead of checking for a non-null `HeaderView` just for added clarity and to prevent possible situations where everything may not be disposed.

### Bugs Fixed ###

- [Bug 48998 - Back on NavigationPage causes System.ObjectDisposedException](https://bugzilla.xamarin.com/show_bug.cgi?id=48998)
- [Bug 57910 - ObjectDisposedException in Xamarin.Forms.Platform.Android.Renderers.ProgressBarRenderer](https://bugzilla.xamarin.com/show_bug.cgi?id=57910)
- https://bugzilla.xamarin.com/show_bug.cgi?id=45330

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/EmbeddingTestBeds/Embedding.Droid/Embedding.Droid.csproj
+++ b/EmbeddingTestBeds/Embedding.Droid/Embedding.Droid.csproj
@@ -16,7 +16,7 @@
     <AndroidResgenFile>Resources\Resource.Designer.cs</AndroidResgenFile>
     <GenerateSerializationAssemblies>Off</GenerateSerializationAssemblies>
     <AndroidUseLatestPlatformSdk>True</AndroidUseLatestPlatformSdk>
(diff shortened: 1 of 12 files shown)


kingces95:
Eek! Got `System.ObjectDisposedException: Cannot access a disposed object. Object name: 'Android.Widget.ProgressBar'.` Looks like a different issue than the one you were going after tho. 

image

samhouts:
Hmm, that one should have been fixed by #1063. I haven't been able to repro it since that fix was put in place.

PR #1165, opened 2017-09-28
[Android] Add Android ListView IsNestedScrollingEnabled specific

cschwarz:
### Description of Change ###

Add IsNestedScrollingEnabled Android platform specific

### Bugs Fixed ###

None

### API Changes ###

Added:
 - `Xamarin.Forms.PlatformConfiguration.AndroidSpecific.ListView.IsNestedScrollingEnabled`

### Behavioral Changes ###

None

### PR Checklist ###


bentmar:
Just curious to what this does?

cschwarz:
If you use a ListView inside a ScrollView on Android, you can't scroll inside the ListView unless you set IsNestedScrollingEnabled.

Use case:
If you have a form with a few entries and use a ListView as table, you need to wrap the form in a ScrollView so you can see all elements on a small device.

rmarinho:
Hi, can you rebase please ? Thanks!

cschwarz:
I am not sure if you get a notification on a push, so to make sure it is rebased now.

samhouts:
Seems like this could be done easily with an effect or custom renderer.

cschwarz:
That argument applies to all platform specifics.

samhouts:
Not necessarily. We do try to limit the platform specifics to things that are 1. very common, or 2. can't be done with effects/custom renderers because of lifecycle and/or visibility. In this case, putting a `ListView` inside of a `ScrollView` is actually something that we recommend _not_ doing, so it may not make sense for us to include a property like this that seems to encourage it. (See https://developer.xamarin.com/guides/xamarin-forms/user-interface/listview/performance/#Improving_ListView_Performance for a few reasons why we don't recommend it.)

rmarinho:
Hi after some more internal review we are not taking this Platform specific.

Thanks

PR #1167, opened 2017-09-28
[XamlC] XamlResourceIdAttribute

StephaneDelcroix:
### Description of Change ###

Attribute partial classes generated from XamlG with a XamlResourceIdAttribute identifying the resource ID.

With that attribute, we can simplify the process of identifying the right resource for the right type in the XamlLoader and no longer have to guess or run expensive regex on the xaml.

This is the side effect of another branch I'm working on, but is nice on it's own, so it deserves it's PR.

### Bugs Fixed ###

/

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###


.nuspec/Xamarin.Forms.targets
```diff
@@ -11,36 +11,11 @@
 
 	<Import Project="$(MSBuildThisFileDirectory)Xamarin.Forms.DefaultItems.targets" Condition="'$(MSBuildSDKsPath)'!=''" />
 
-	<PropertyGroup>
```
StephaneDelcroix: This is only moving chairs and bumping the ABI version

Xamarin.Forms.Build.Tasks/XamlGTask.cs
```diff
@@ -16,10 +16,8 @@ public class XamlGTask : Task
 		internal static CodeDomProvider Provider = new CSharpCodeProvider();
 
 		[Required]
-		public string Source { get; set; }
-
+		public ITaskItem Source { get; set; }
```
StephaneDelcroix: use a TaskItem so we can extract metadata, like the ResourceName

Xamarin.Forms.Core/Xaml/XamlResourceIdAttribute.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+
+namespace Xamarin.Forms.Xaml
+{
+	[AttributeUsage(AttributeTargets.Class, Inherited = false, AllowMultiple = false)]
+	public sealed class XamlResourceIdAttribute : Attribute
+	{
+		public string RootNamespace { get; set; }
```
StephaneDelcroix: this property is needed for future use

StephaneDelcroix:
I'll rebase and revisit this when #1202 is merged

StephaneDelcroix:
rebased on top of #1202 merged to master

PR #1172, opened 2017-09-29
[Xaml[C]] ResourceDictionary.Source

StephaneDelcroix:
### Description of Change ###

Allow setting a Source to a ResourceDictionary. Source is an Uri, relative or absolute, that points to a resource in the same assembly.

Restriction: this can only be used from within Xaml

MergedWith is deprecated (obsolete).

Oh, and you no longer have to create code behind for resource dictionaries !

**Needs to be rebased when #1167 is merged** 

TODO:
- compile the RD with XamlC

### Bugs Fixed ###

/

### API Changes ###

Added:
 - Uri ResourceDictionary.Source {get; set;}

### Behavioral Changes ###

/

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/ResourceDictionary.cs
+++ b/Xamarin.Forms.Core/ResourceDictionary.cs
@@ -7,6 +7,8 @@ using System.Reflection;
 using Xamarin.Forms.Internals;
 using System.Collections.ObjectModel;
 using System.Collections.Specialized;
+using System.Globalization;
+using System.ComponentModel;
 
 namespace Xamarin.Forms
 {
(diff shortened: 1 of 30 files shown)


Xamarin.Forms.Core/ResourceDictionary.cs
```diff
@@ -36,6 +43,28 @@ public class ResourceDictionary : IResourceDictionary, IDictionary<string, objec
 			}
 		}
 
+		[TypeConverter(typeof(RDSourceTypeConverter))]
```
StephaneDelcroix: ok, here's the magic...

if the type is attributed with a TypeConverter, the converter will be executed, and it's return value used as value for the setter.

during the execution of that converter we call SetAndLoadSource, which sets `_source` and loads the actual RD. When the converter returns the Uri, and the xaml engine sets it to this property, we no_op and exits.

this is the reason why it only works from xaml...

Xamarin.Forms.Core/ResourceDictionary.cs
```diff
@@ -256,5 +285,66 @@ void OnValuesChanged(params KeyValuePair<string, object>[] values)
...
+				return uri;
+			}
+
+			internal static string ComputeResourceId(Uri uri, string rootResourceId, string rootNs)
+			{
+				if (uri.IsAbsoluteUri)
+					return $"{rootNs}{uri.AbsolutePath.Replace('/','.')}";
+
+				var rootResourceIdLength = rootResourceId.Length;
+				rootResourceIdLength = rootResourceId.LastIndexOf('.', rootResourceIdLength - 1); //drop the .xaml
```
StephaneDelcroix: those length and index vars are used to avoid Substringing before the actual value is needed

Xamarin.Forms.Core/ResourceDictionary.cs
```diff
@@ -256,5 +285,66 @@ void OnValuesChanged(params KeyValuePair<string, object>[] values)
...
+				rootResourceIdLength = rootResourceId.LastIndexOf('.', rootResourceIdLength - 1); //drop the resource name
+
+				var path = uri.OriginalString;
+				var pathStartIndex = 0;
+
+				while (true) {
+					if (path.IndexOf("./", pathStartIndex, 2, StringComparison.OrdinalIgnoreCase) == pathStartIndex)
```
StephaneDelcroix: this would be equivalent of a StartsWith() with a startIndex parameter

PR #1173, opened 2017-09-29
[Android] Update FrameRenderer.cs

bill2004158:
1. fix bug: after update background color, the Outline is miss. 
2. created duplicated FrameDrawable (add duplicated event handler to _frame)
3. remove event handler when dispose


rmarinho:
#1385 fixes this

PR #463, opened 2016-10-16
[iOS] Add Platform Specific features for PrefersStatusBarHidden/UIStatusBarAnimation

pauldipietro:
### Description of Change

This is a Platform Specific feature which enables setting the visibility of the iOS status bar on a per-view basis. See some examples below:

output_1iqobs
output_qnofag
output_yi11ig

This is accomplished through the `StatusBarHiddenMode` and `UIStatusBarAnimation` enumerators, which contain the following values:

`StatusBarHiddenMode.Default`
`StatusBarHiddenMode.True`
`StatusBarHiddenMode.False`

`UIStatusBarAnimation.None`
`UIStatusBarAnimation.Fade`
`UIStatusBarAnimation.Slide`

Per [Apple's documentation](https://developer.apple.com/reference/uikit/uiviewcontroller/1621440-prefersstatusbarhidden), since iOS 8 the `prefersStatusBarHidden` method returns true on vertically compact environments. This represents the `StatusBarHiddenMode.Default` value, while the latter two apply the respective values to both the vertical and landscape orientations. This can be set as follows, for example:

`page.On<iOS>().PrefersStatusBarHidden(StatusBarHiddenMode.True)`

As one might guess, the `UIStatusBarAnimation` values represent the three possible manners by which the status bar can enter or leave the view. If a `Fade` or `Slide` value is set, a 0.25 second animation of [setNeedsStatusBarAppearanceUpdate](https://developer.apple.com/reference/uikit/uiviewcontroller/1621354-setneedsstatusbarappearanceupdat) will execute. That can be used like so:

`page.On<iOS>().SetPreferredStatusBarUpdateAnimation(UIStatusBarAnimation.Fade)`

Starting with the `Page` value on the `Platform` (which has been subsequently marked as internal in order for the `PlatformRenderer` to get at it), [childViewControllerForStatusBarHidden](https://developer.apple.com/reference/uikit/uiviewcontroller/1621451-childviewcontrollerforstatusbarh) is utilized to redirect the views down to `PageRenderer` as a baseline for setting the values on a per-view basis. An exception is that in a `TabbedPage`, setting the `PrefersStatusBarHidden` value directly on that page will update all of its contained pages.

As of this submission, tests can still be added, though they may be somewhat complex due to the need to check the status bar -- screenshots might be preferable.
### Bugs Fixed

This was technically not a bug, per se, but was reported under https://bugzilla.xamarin.com/show_bug.cgi?id=44803.
### API Changes

Added:
[...]

PR #1175, opened 2017-09-29
[iOS] Redirect PreferredStatusBarStyle to the Child View Controller

MarkoBL:
### Description of Change ###

UIApplication.StatusBarStyle is deprecated since iOS 9 (https://developer.apple.com/documentation/uikit/uiapplication/1622988-statusbarstyle) and doesn't work on Xamarin.Forms at all. If you use the black navigation bar style or a dark background color for the navigation page, the status bar always uses the default black color.

This PR simply redirects the calls to UIViewController.PreferredStatusBarStyle from the PlatformRenderer to the top view controller (NavigationRenderer or PageRenderer), therefore you can set the status bar color in your own renderer/ViewController:

```
	public class ExtNavigationRenderer : NavigationRenderer
	{
		public override UIStatusBarStyle PreferredStatusBarStyle()
		{
			return UIStatusBarStyle.LightContent;
		}
	}
```

### Behavioral Changes ###

There should be none.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/PlatformRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/PlatformRenderer.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Diagnostics;
+using System.Linq;
 using UIKit;
 using Xamarin.Forms.PlatformConfiguration.iOSSpecific;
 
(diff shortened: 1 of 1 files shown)


adrianknight89:
@pauldipietro Wasn't this fixed a couple months back or there was a platform specific option dealing with this? I can't remember, but I remember some kind of working being done with regards to PreferredStatusBarStyle vs child view controllers.

pauldipietro:
@adrianknight89  #463 was the PR having to do with `PrefersStatusBarHidden` which I believe you're referring to.

PR #1177, opened 2017-10-04
[Core] Add extra method that allows to specify default value for CanExecute on a Command

rmarinho:
### Description of Change ###

This pr resolves a issue where our `CanExecute(obj)` method on the `Command` will always return true if a `CanExecute` delegate wasn't specified. 
The problem happens that in a `Button` we don't have no way to figure if a user specified or not that delegate on a `Command`, and we need to decide to either respect the IsEnabled property from the button or the result of the command can execute. 

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=32899

### API Changes ###

Adding to `Command`
Added:
 - public bool CanExecute(object parameter, bool defaultCanExecute)

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Button.cs
+++ b/Xamarin.Forms.Core/Button.cs
@@ -188,8 +188,8 @@ namespace Xamarin.Forms
 		void CommandCanExecuteChanged(object sender, EventArgs eventArgs)
 		{
 			ICommand cmd = Command;
-			if (cmd != null)
-				IsEnabledCore = cmd.CanExecute(CommandParameter);
+			if(cmd != null)
+				IsEnabledCore = (cmd as Command)?.CanExecute(CommandParameter, IsEnabled) ?? cmd.CanExecute(CommandParameter);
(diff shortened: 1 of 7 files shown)


Xamarin.Forms.Core/Button.cs
```diff
@@ -188,8 +188,8 @@ protected override void OnPropertyChanging(string propertyName = null)
 		void CommandCanExecuteChanged(object sender, EventArgs eventArgs)
 		{
 			ICommand cmd = Command;
-			if (cmd != null)
-				IsEnabledCore = cmd.CanExecute(CommandParameter);
+			if(cmd != null)
+				IsEnabledCore = (cmd as Command)?.CanExecute(CommandParameter, IsEnabled) ?? cmd.CanExecute(CommandParameter);
```
StephaneDelcroix: this makes a lot of assumptions

PR #1181, opened 2017-10-05
[*] Expose the Flags to the Core

StephaneDelcroix:
Expose the Flags to the Core, through Device

as we aren't sure yet if this is the best API ever, decorate it with a nice [STAND BACK] attribute

New API:

```
public static IReadOnlyList<string> Device.Flags { get; private set; }

public static void Device.SetFlags(IReadOnlyList<string> flags)
```

### DIFF
--- a/Xamarin.Forms.Core/Device.cs
+++ b/Xamarin.Forms.Core/Device.cs
@@ -83,6 +83,7 @@ namespace Xamarin.Forms
 			set { s_platformServices = value; }
 		}
 
+		[EditorBrowsable(EditorBrowsableState.Never)]
 		public static IReadOnlyList<string> Flags { get; private set; }
 
 		[EditorBrowsable(EditorBrowsableState.Never)]
--- a/Xamarin.Forms.Core/Device.cs
+++ b/Xamarin.Forms.Core/Device.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.IO;
 using System.Reflection;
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Core/Device.cs
```diff
@@ -82,6 +83,15 @@ public static IPlatformServices PlatformServices
 			set { s_platformServices = value; }
 		}
 
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public static IReadOnlyList<string> Flags { get; private set; }
```
StephaneDelcroix: @jassmith the new API is now hidden

jalbertSyncroTech:
I'm getting a null exception in StaticResourceExtension ProvideValue() which calls Xamarin.Forms.Device.Flags.Contains("xamlDoubleImplicitOpHack") because Flags is null.

I think "Device.SetFlags(s_flags);" needs to be changed to "Device.SetFlags(Flags);", as the latter will create an empty list if s_flags is null.

StephaneDelcroix:
@jalbertSyncroTech thanks. fixed

jaysacco:
As far as I can tell, the bug that's fixed by Device.SetFlags(Flags) was introduced in 2.4 SR2, yet the fix has not made it into any 2.4 or 2.5 service releases.  When will it be released and how is it possible that no one else is hitting this bug?  Is there some workaround?

jaysacco:
I've worked around be doing Forms.SetFlags("WorkaroundForNullDeviceFlags"); but would still like to know when this will be released to the public

rmarinho:
@jaysacco this should be in the latest 2.4.0 and 2.5.0

jaysacco:
Thanks for the response.  I assumed as much but when I browse the source at those tags e.g. https://github.com/xamarin/Xamarin.Forms/blob/release-2.5.0-sr3/Xamarin.Forms.Xaml/MarkupExtensions/StaticResourceExtension.cs
the null check is missing on line 60.  After installing the nuget package, I hit the null reference until I put in the workaround noted above.  I can believe something's amiss with my nuget installation, but what about the github link?

PR #1188, opened 2017-10-09
[Android] Handle rapid taps when only single-tap recognizer is present

hartez:
### Description of Change ###

When an element has a TapGestureRecognizer with NumberOfTaps set to 1 (and no "double-tap" recognizer is present), tapping fast enough causes every other tap to be ignored. This is caused by confusion from the double-tap recognition component. This change addresses that scenario. 

### Bugs Fixed ###

- [59961 – TapGestureRecognizer with NumberOfTapsRequired="1" works only with double tap](https://bugzilla.xamarin.com/show_bug.cgi?id=59961)
- [59873 – Tap Gesture Recognizers not firing on frames](https://bugzilla.xamarin.com/show_bug.cgi?id=59873)
- *Possibly* [59863 – TapGestureRecognizer extremely finicky](https://bugzilla.xamarin.com/show_bug.cgi?id=59863). We'll need to get verification from Phillipp Sumi that this fixes his problem; other people jumped into the report and it's not clear if he was experiencing a side effect from this issue or another issue entirely.

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla59863_0.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla59863_0.cs
@@ -1,0 +1,92 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+using Xamarin.Forms.Core.UITests;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if UITEST
+	[Category(UITestCategories.Gestures)]
+#endif
+
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Bugzilla, 59863, "TapGestureRecognizer extremely finicky", PlatformAffected.Android)]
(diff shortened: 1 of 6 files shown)


hartez:
Only failing test is the restart test which is broken for unrelated reasons (and is fixed by #1194).

EmmanVazz:
When will this be merged? Unfortunately this is holding our new release back .

mackayn:
Would be great if this makes it into a service pack for 2.4.0 as we can't upgrade from 2.3.5-pre6 until then.

justinimel:
Also waiting for this to make it into a release for 2.4.0.  We rely on tap gesture recognizers heavily and they are completely unreliable in the current release.

justinimel:
I have pulled the latest version (.18342).   I can confirm it seems to solve most of the gesture based issues I was seeing.

However the one that you have listed as "possibly" I believe may still be an issue. I'll put some more information on the bug report.

adrianknight89:
@hartez @jassmith I'm seeing this issue as well even after upgrading to 2.5.0-pre2. I filed a new bug with a repro to compare gesture recognizer to button clicks. The difference is noticeable if you tap on them many times.

https://bugzilla.xamarin.com/show_bug.cgi?id=60341

flamecrow:
This issue is ridiculous and frustrating, I can't believe it still isn't fixed.

PR #1193, opened 2017-10-10
UITest for ListView caching

kingces95:
UITest only


rmarinho:
Do we still need this @kingces95 ? if so can you rebase? thanks

PR #1202, opened 2017-10-16
Use "correct" file paths for .xaml.g.cs files

StephaneDelcroix:
### Description of Change ###

The file path for partial classes were generated using the ResourceName for the file. For SAP(Shared Assets Projects, not the other SAP) projects, on some msbuild versions on some OSes, those resource names might contain the full path. That full path, concatenated to the current build directory, could sometimes ends up being too long for outdated file system and build engines...

This uses the same algorithm as the one used for UWP Pages on SAP projects, meaning it duplicates the relative directory from the project root.

This is now all correct and well, as God intended to on day 1, but delayed it until now.

~~**NOTE 1:** It might be important to notice that our current project templates are creating an App.xaml and MainPage.xaml files. Those being at the root of the SAP project will conflict with the default App.xaml and MainPage.xaml files on the UWP project. This is expected, and it's the way UWP works, so ` ¯\_(ツ)_/¯`~~

**NOTE 2:** `xbuild` doesn't see SAP projects the same way `msbuild` does. Projects using `xbuild` are therefore no longer supported.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=40077

### API Changes ###

I guess that Notes 1 and 2 are annoying enough, no need to break any API on top of that.

### PR Checklist ###


Xamarin.Forms.Build.Tasks/FixedCreateCSharpManifestResourceName.cs
```diff
@@ -1,29 +0,0 @@
-﻿using Microsoft.Build.Framework;
```
StephaneDelcroix: no longer required

.nuspec/Xamarin.Forms.targets
```diff
@@ -66,35 +42,42 @@
...
-		<ItemGroup>
-			<XamlGFiles Include="@(XamlFiles->'$(IntermediateOutputPath)%(ManifestResourceName).g$(DefaultLanguageSourceExtension)')"/>
-			<Compile Include="@(XamlGFiles)"/>
-			<FileWrites Include="@(XamlGFiles)"/>
-		</ItemGroup>
+	<Target Name="_ValidateMSBuild" BeforeTargets="_CheckForInvalidConfigurationAndPlatform">
+		<Error
+			Text="Xamarin.Forms require msbuild. xbuild is no longer supported."
+			Code="XF003"
+			Condition="'$(IsXbuild)' == 'true'" />
```
StephaneDelcroix: `IsXBuild`

that's not because it's case insensitive that you can type whatever you want

StephaneDelcroix:
after additional investigation, the `NOTE 1` isn't a problem as, for the same `App.xaml` file, the UWP head will generate `App.i.cs`, and XF will generate `App.xaml.g.cs`. so we're good

PR #1207, opened 2017-10-18
Stylesheets

StephaneDelcroix:
### Description of Change ###

Support cascading StyleSheets.
Support all selector (no media selectors)

#### Preliminary note
This code and feature is brand new. The API shouldn't change much, but some of the behaviors might be slightly altered in the future, depending on the feedback received.

#### Supported Selectors
|Selector|Example|Description|
|---|---|---|
|`.class`|`.header`|Selects all elements with the `StyleClass` property containing 'header'|
|`#id`|`#email`|Selects all elements with `StyleId` set to `email`. If `StyleId` is not set, fallback to `x:Name`. When using Xaml, always prefer `x:Name` over `StyleId`.|
|`*`|`*`|Selects all elements|
|`element`|`label`|Selects all elements of type Label (but not subclasses). Case irrelevant.|
|`^base`|`^contentpage`|Selects all elements with `ContentPage` as base class, including ContentPage itself. Case irrelevant. This selector isn't present in the CSS specification, and only applies to XF.|
|`element,element`|`label,button`|Selects all Buttons and all Labels|
|`element element`|`stacklayout label`|Selects all Labels inside of a StackLayout|
|`element>element`|`stacklayout>label`|Selects all Labels with StackLayout as a **direct** parent|
|`element+element`|`label+entry`|Selects all Entries **directly** after a Label|
|`element~element`|`label~entry`|Selects all Entries preceded by a Label|

#### Unsupported Selectors (for this version)
 - `[attribute]` selectors
 - `@media` or `@supports`selectors
 - `:` or `::` selectors

#### Selector combinations 

Selectors can be combined without limitation, like in `StackLayout > ContentView > label.email`. But keep it sane !

#### Precedence

Styles with matching selectors are all applied, on by one, in definition order. Styles defined on the item itself is always applied last.

This is the expected behavior in most cases, even if doesn't 100% match common CSS implementations.

Specificity, and specificity overrides (`!important`) are not supported. This is a known issue.

#### Supported properties
|Property|Applies to|Values (string literals are `grayed`, while types are _italic_)|Example|
|---|---|---|---|
|`background-color`|`VisualElement`|_color_ (see Colors) \| `initial`|`background-color: springgreen;`|
|`background-image`|`Page`|_string_ \| `initial `|`background-image: bg.png;`|
|`border-color`|`Button`, `Frame`|_color_ (see Colors) \| `initial `|`border-color: #9acd32;`|
|`border-width`|`Button`|_double_ \| `initial `|`border-width: .5;`|
|`color`|`Button`, `DatePicker`, `Editor`, `Entry`, `Label`, `Picker`, `SearchBar`, `TimePicker`|_color_ (see Colors) \| `initial `|`color: rgba(255, 0, 0, 0.3);`|
|`direction`|`VisualElement`|`ltr` \| `rtl` \| `inherit`, `initial `|`direction: rtl;`|
|`font-family`|`Button`, `DatePicker`, `Editor`, `Entry`, `Label`, `Picker`, `SearchBar`, `TimePicker`, `Span`|_string_ \| `initial `|`font-family: Consolas;`|
|`font-size`|`Button`, `DatePicker`, `Editor`, `Entry`, `Label`, `Picker`, `SearchBar`, `TimePicker`, `Span`|_double_  \| _namedsize_ (see NamedSize) \| `initial `|`font-size: 12;`|
|`font-style`|`Button`, `DatePicker`, `Editor`, `Entry`, `Label`, `Picker`, `SearchBar`, `TimePicker`, `Span`|`bold` \| `italic` \| `initial `|`font-style: bold;`|
|`height`|`VisualElement`|_double_ \| `initial `|`min-height: 250;|
|`margin`|`View`|_thickness_ (see Thickness) \| `initial `|`margin: 6 12;`|
|`margin-left`|`View`|_thickness_ (see Thickness) \| `initial `|`margin-left: 3;`|
|`margin-top`|`View`|_thickness_ (see Thickness) \| `initial `|`margin-top: 2;`|
|`margin-right`|`View`|_thickness_ (see Thickness) \| `initial `|`margin-right: 1;`|
|`margin-bottom`|`View`|_thickness_ (see Thickness) \| `initial `|`margin-bottom: 6;`|
|`min-height`|`VisualElement`|_double_ \| `initial `|`min-height: 50;|
|`min-width`|`VisualElement`|_double_ \| `initial `|`min-width: 112;|
|`opacity`|`VisualElement`|_double_ \| `initial `|`opacity: .3;`|
|`padding`|`Layout`, `Page`|_thickness_ (see Thickness) \| `initial `|`padding: 6 12 12;`|
|`padding-left`|`Layout`, `Page`|_double_ \| `initial `|`padding-left: 3;`|
|`padding-top`|`Layout`, `Page`| _double_ \| `initial ` |`padding-top: 4;`|
|`padding-right`|`Layout`, `Page`| _double_ \| `initial ` |`padding-right: 2;`|
|`padding-bottom`|`Layout`, `Page`| _double_ \| `initial ` |`padding-bottom: 6;`|
|`text-align`| `Entry`, `EntryCell`, `Label`, `SearchBar`|`left` \| `right` \| `center` \| `start` \| `end` \| `initial `. It is recommended to avoid `left` and `right` in non-ltr environments| `text-align: right;`|
|`visibility`|`VisualElement`|`true` \| `visible` \| `false` \| `hidden` \| `collapse` \| `initial `|`visibility: hidden;`|
|`width`|`VisualElement`|_double_ \| `initial `|`min-width: 320;|

#### Unsupported Common Properties
 - `all: initial`,
 - layout  properties (box, or grid). FlexLayout is coming, and it'll be CSS stylable,
 - shorthand properties, like `font`, `border`

#### Colors
 - one of the 140 `X11` color (https://en.wikipedia.org/wiki/X11_color_names), which happens to match CSS Colors, UWP predefined colors and XF Colors. Case insensitive
 - HEX: `#rgb`, `#argb`, `#rrggbb`, `#aarrggbb`
 - RGB: `rgb(255,0,0)`, `rgb(100%,0%,0%)` => values in range 0-255 or 0%-100%
 - RGBA: `rgba(255, 0, 0, 0.8)`, `rgba(100%, 0%, 0%, 0.8)` => opacity is 0.0-1.0
 - HSL: `hsl(120, 100%, 50%)` => h is 0-360, s and l are 0%-100%
 - HSLA: `hsla(120, 100%, 50%, .8)` => opacity is 0.0-1.0

#### Thickness
One, two, three or four values, separated by white spaces.
 - a single value indicates uniform thickness
 - two values indicates (resp.) vertical and horizontal thickness
 - three values indicates (resp.) top, horizontal (left and right) and bottom thickness
 - when using four values, they are top, right, bottom, left

> **IMPORTANT**: This differs from Xaml thickness definitions, which are
>  1. separated by commas (`,`),
>  2. are in the form of `uniform`, `horizontal, vertical` or `left, top, right, bottom`.

#### NamedSize
One of the following value, case insensitive. Exact meaning depends of the platform and the control
 - `default`,
 - `micro`,
 - `small`,
 - `medium`,
 - `large`

#### Initial
`initial ` is a valid value for all properties. It clears the value (resets to default) that was set from another Style.

#### Additional remarks
 - no inheritance supported, meaning no `inherit` value and that you can't set the font-size to a layout, and expect all the labels in that layout to inherit the value. The only exception is the `direction` property, which supports `inherit`, and that's the default value.
 - element are matched by name, no support for xmlns

##### XAML (preferred)
```
<ContentPage x:Class="...">
  <ContentPage.Resources>
    <StyleSheet Source="appresources/style.css" />
  </ContentPage.Resources>
</ContentPage>
```
the `Source` argument takes an Uri relative to the current xaml control, or relative to the application root if it starts with a `/`. The `style.css` has to be an EmbeddedResource.

alternatively, you can inline your style in a `CDATA` Section
```
<ContentPage x:Class="...">
  <ContentPage.Resources>
    <StyleSheet>
<![CDATA[
^contentpage {
    background-color: orange;
    padding: 20;
}

stacklayout > * {
    margin: 3;
}
]]>
    </StyleSheet>
  </ContentPage.Resources>
</ContentPage>
```

do not abuse of that second syntax.

##### in C#
From an embedded resource:
```
myPage.Resources.Add(StyleSheet.FromAssemblyResource(this.GetType().Assembly, "resource.id.of.the.css"));
```
or from a `TextReader`:
```
using (var reader = new StringReader(my_css_string))
    myPage.Resources.Add(StyleSheet.FromReader(reader));
```

#### StyleSheet, XamlC and other potential optimizations
At this time, CSS StyleSheets are parsed and evaluated at runtime. That aren't compiled. Every time a StyleSheet is used, it's reparsed again. If parsing time is an issue, enabling caching is trivial, but comes at memory cost.

### Bugs Fixed ###

None. but It'll introduce some new ones

#### Added
Not that much new API...
```
namespace Xamarin.Forms.StyleSheets{
    public sealed class StyleSheet {
        public static StyleSheet FromAssemblyResource(Assembly assembly, string resourceId, IXmlLineInfo lineInfo = null);
        public static StyleSheet FromReader(TextReader reader);
    }
}
namespace Xamarin.Forms {
    public class ResourceDictionary {
        void Add(StyleSheet);
    }

    public class Frame {
        public static readonly BindableProperty BorderColorProperty;
        public Color BorderColor { get; set; }
    }
}
```

#### Obsoleted
```
Frame.OutlineColor { get; set; }
Frame.OutlineColorProperty;
```

### Behavioral Changes ###

/

### PR Checklist ###


.nuspec/Xamarin.Forms.targets
```diff
@@ -92,4 +93,23 @@
...
+
+	<!-- CssG -->
+		<PropertyGroup>
+		<CoreCompileDependsOn>
+			CssG;
+			$(CoreCompileDependsOn);
+		</CoreCompileDependsOn>
+	</PropertyGroup>
+
+	<Target Name="CssG">
```
StephaneDelcroix: the new CssG task is responsible for creating XamlResourceIdAttributes for Css file

Xamarin.Forms.Core.UnitTests/StyleSheets/SelectorTests.cs
```diff
@@ -0,0 +1,110 @@
...
+		[TestCase("page contentview label", false, false, true, false, false, false)]
+		[TestCase("page contentview>label", false, false, true, false, false, false)]
+		[TestCase("page>stacklayout contentview label", false, false, true, false, false, false)]
+		[TestCase("page stacklayout>contentview label", false, false, true, false, false, false)]
+		[TestCase("page stacklayout contentview>label", false, false, true, false, false, false)]
+		[TestCase("page>stacklayout>contentview label", false, false, true, false, false, false)]
+		[TestCase("page>stack/* comment * */layout>contentview label", false, false, true, false, false, false)]
+		[TestCase("page>stacklayout contentview>label", false, false, true, false, false, false)]
+		[TestCase("page stacklayout>contentview>label", false, false, true, false, false, false)]
+		[TestCase("page>stacklayout>contentview>label", false, false, true, false, false, false)]
```
StephaneDelcroix: this is probably the best documentation for the selector we support

Xamarin.Forms.Core/Properties/AssemblyInfo.cs
```diff
@@ -59,4 +60,7 @@
 [assembly: InternalsVisibleTo("Xamarin.Forms.CarouselView")]
 [assembly: Preserve]
 
-[assembly: XmlnsDefinition("http://xamarin.com/schemas/2014/forms", "Xamarin.Forms")]
\ No newline at end of file
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2014/forms", "Xamarin.Forms")]
+
+[assembly: StyleProperty("color", typeof(ITextElement), nameof(TextElement.TextColorProperty))]
+[assembly: StyleProperty("background-color", typeof(VisualElement), nameof(VisualElement.BackgroundColorProperty))]
```
StephaneDelcroix: this is the mapping table to be completed, matching css property names to XF properties

Xamarin.Forms.Core/VisualElement_StyleSheet.cs
```diff
@@ -0,0 +1,103 @@
...
+	{
+		IList<string> IStyleSelectable.Classes
+			=> StyleClass;
+
+		public static readonly BindableProperty StyleSheetProperty =
+			BindableProperty.Create("StyleSheet", typeof(StyleSheet), typeof(VisualElement), default(StyleSheet),
+				propertyChanged: (bp, o, n) => ((VisualElement)bp).OnStyleSheetChanged((StyleSheet)o, (StyleSheet)n));
+
+		[TypeConverter(typeof(StyleSheetConverter))]
+		public StyleSheet StyleSheet {
```
StephaneDelcroix: there's no actual public way to create a StyleSheet, so this expects a string uri in Xaml and the StyleSheet will be generated

Xamarin.Forms.Core/StyleSheets/StylePropertyAttribute.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+
+namespace Xamarin.Forms.StyleSheets
+{
+	[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true, Inherited = true)]
+	sealed class StylePropertyAttribute : Attribute
```
StephaneDelcroix: need to be public

Xamarin.Forms.Core/Properties/AssemblyInfo.cs
```diff
@@ -59,4 +60,17 @@
...
-[assembly: XmlnsDefinition("http://xamarin.com/schemas/2014/forms", "Xamarin.Forms")]
\ No newline at end of file
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2014/forms", "Xamarin.Forms")]
+
+[assembly: StyleProperty("background-color", typeof(VisualElement), nameof(VisualElement.BackgroundColorProperty))]
+[assembly: StyleProperty("background-image", typeof(Page), nameof(Page.BackgroundImage))]
+[assembly: StyleProperty("border-color", typeof(Button), nameof(Button.BorderColor))]
+[assembly: StyleProperty("border-width", typeof(Button), nameof(Button.BorderWidth))]
+[assembly: StyleProperty("color", typeof(ITextElement), nameof(TextElement.TextColorProperty))]
+//[assembly:StyleProperty("direction", .., ..)]
```
StephaneDelcroix: @samhouts your help is needed here

Xamarin.Forms.Core/StyleSheets/StylePropertyAttribute.cs
StephaneDelcroix: no, it does not

.nuspec/Xamarin.Forms.targets
```diff
@@ -78,4 +79,23 @@
...
```
StephaneDelcroix: change it to `BeforeTarget="BeforeCompile"` and remove the property group

Xamarin.Forms.Core.UnitTests/StyleSheets/SelectorTests.cs
```diff
@@ -0,0 +1,114 @@
...
+		[TestCase("visualelement", false, false, false, false, false, false)]
+		[TestCase("^visualelement", true, true, true, true, true, true)]
+		[TestCase("^layout", false, false, false, false, false, true)]
```
StephaneDelcroix: this is how me make CSS support language with inheritance

Xamarin.Forms.Build.Tasks/CompiledConverters/RDSourceTypeConverter.cs
```diff
@@ -66,5 +66,29 @@ public IEnumerable<Instruction> ConvertFromString(string value, ILContext contex
 			//ldloc the stored uri as return value
 			yield return Create(Ldloc, uriVarDef);
 		}
+
+		internal static string GetPathForType(ModuleDefinition module, TypeReference type)
```
StephaneDelcroix: note sure those 2 are used anymore

StephaneDelcroix: that one is used. I removed the other one

Xamarin.Forms.Build.Tasks/SetPropertiesVisitor.cs
```diff
@@ -1156,6 +1156,9 @@ static bool CanAddToResourceDictionary(TypeReference collectionType, IElementNod
 			if (node.XmlType.Name == "ResourceDictionary")
 				return true;
 
+			if (node.XmlType.Name == "StyleSheet")
```
StephaneDelcroix: replace those 3 checks by a single reflection based one (see AddToRD below)

StephaneDelcroix:
thickness in css are top right bottom left, not left top right bottom

tpetrina:
I see that there are individual properties for `padding`, but not for `margin`. Why is that?

Also, can we nest selectors arbitrary many times? Like `grid > grid > label.title`.

StephaneDelcroix:
@tpetrina 
> I see that there are individual properties for padding, but not for margin. Why is that?

nice catch, I'll add them

> Also, can we nest selectors arbitrary many times? Like grid > grid > label.title

yes, you can. see https://github.com/xamarin/Xamarin.Forms/pull/1207/files#diff-73f8da31e9d7780757345a55024ff7eaR99 for a test. The "Selector Combination" part of the PR description should state that, but it's still empty

Xamarin.Forms.Core/StyleSheets/CssReader.cs
```diff
@@ -0,0 +1,97 @@
...
+
+			bool hasStar = false;
+			while (true) {
+				var next = _reader.Read();
+				if (next <= 0)
+					return next;
+				if (unchecked((char)next) == '*')
+					hasStar = true;
+				else if (hasStar && unchecked((char)next) == '/')
+					return Peek(); //recursively call self for comments followign comments
```
samhouts: followign >> following

Xamarin.Forms.Core/StyleSheets/CssReader.cs
```diff
@@ -0,0 +1,97 @@
...
+
+			bool hasStar = false;
+			while (true) {
+				var next = _reader.Read();
+				if (next <= 0)
+					return next;
+				if (unchecked((char)next) == '*')
+					hasStar = true;
+				else if (hasStar && unchecked((char)next) == '/')
+					return Read(); //recursively call self for comments followign comments
```
samhouts: followign >> following

Xamarin.Forms.Platform.Android/Renderers/FrameRenderer.cs
```diff
@@ -210,7 +210,7 @@ void DrawOutline(ACanvas canvas, int width, int height, float cornerRadius)
 			void FrameOnPropertyChanged(object sender, PropertyChangedEventArgs e)
 			{
 				if (e.PropertyName == VisualElement.BackgroundColorProperty.PropertyName ||
-					e.PropertyName == Frame.OutlineColorProperty.PropertyName ||
+				    e.PropertyName == Frame.BorderColorProperty.PropertyName ||
```
samhouts: 👮

Xamarin.Forms.Core/Element_StyleSheets.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Xamarin.Forms
+{
+	public partial class Element : IStyleSelectable
+	{
+		IEnumerable<IStyleSelectable> IStyleSelectable.Children => LogicalChildrenInternal;
+
+		IList<string> IStyleSelectable.Classes => null;
+
+		string IStyleSelectable.Id => StyleId;
```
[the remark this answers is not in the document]
StephaneDelcroix: we could only use `x:Name`. let's bring that in a discussion one of these days

Xamarin.Forms.Core/StyleSheets/CssReader.cs
StephaneDelcroix: don't read the comments

StephaneDelcroix:
@samhouts 
> Are there explicit tests for css precedence when two styles match the same element with the same level of specificity?
we do not support `specificity` yet. All styles with a matching selector are applied in the order in which they're defined
> Do we support `!important`?
we don't, on purpose
> Do we support putting css on an element itself?
Even in the HTML world, this is considered very bad practice (quoting MDN here). But you can set the StyleSheet to inline data to any element.

rmarinho:
build

Xamarin.Forms.Core/Frame.cs
```diff
@@ -5,9 +5,12 @@ namespace Xamarin.Forms
 {
 	[ContentProperty("Content")]
 	[RenderWith(typeof(_FrameRenderer))]
-	public class Frame : ContentView, IElementConfiguration<Frame>, IPaddingElement
+	public class Frame : ContentView, IElementConfiguration<Frame>, IPaddingElement, IBorderElement
 	{
-		public static readonly BindableProperty OutlineColorProperty = BindableProperty.Create("OutlineColor", typeof(Color), typeof(Frame), Color.Default);
+		[Obsolete("OutlineColorProperty is obsolete as of version 3.0.0. Please use BorderColorProperty instead.")]
```
alanmcgovern: I think this change caused #1566

ali-h2010:
I am not sure if someone already asked about this before but if someone did, please guide me to the answer. 

How to add different CSS for tablets like the iPad or android tablets in general?
Currently , if you run the app in a big screen device , everything will be stretched and most of the time will not be acceptable in a professional environment. 

so, I had to do many steps to solve this issue such as:
- bind the padding values and width of some views
- use this condition checking to check if device is a phone or tablet
`if (Device.Idiom == TargetIdiom.Phone)
            {
                HomeIconWidth = 70;
                HomeIconHeight = 70;

                HomeIconFrameWidth = 160;
                HomeIconFrameHeight = 160;

                NumberOfApprovalsLayout = new Rectangle(120, 10, 30, 30);
            }
            else
            {
                HomeIconWidth = 110;
                HomeIconHeight = 110;

                HomeIconFrameWidth = 200;
                HomeIconFrameHeight = 200;

                NumberOfApprovalsLayout = new Rectangle(160, 10, 30, 30);
            }
`

As you can see it's a lot of variables and conditions to set for every page.
Even if I use styles , I couldn't find a way to check the display in Xaml and apply the correct style.

Is this issue still there or does CSS provide a solution to that?
if there is a solution ,from the look of it it won't be like the web development CSS since @media is not supported like the one that bootstrap uses.

mmacneil:
Can styles on an element be changed at runtime?  They are applied as expected on initial load but I seem unable to dynamically change them at runtime based on some event.  For example, in a page code-behind this scenario:
`void ButtonClick()
{
     myPage.MyCoolLabel.StyleClass = new List<string> { "someNewStyle" };
}
`

mmacneil:
Done:  [2678](https://github.com/xamarin/Xamarin.Forms/issues/2678)

Thanks Stephane

shivendramca10:
Will width work in percentage.

PR #1213, opened 2017-10-20
Improve 'ResourceDictionary' class

Makeloft:
### Description of Change ###

Add C# 7 support for old build tools by adding https://www.nuget.org/packages/Microsoft.Net.Compilers/

Improve ResourceDictionary class code:
- use 'dictionary.TryGetValue' method into indexer instead constructions 'if (dictionary.ContainsKey(key)) return dictionary[key];'
- avoid redundant 'ToArray' call when use private method 'OnValuesChanged'
- apply C# 7 code style

### Bugs Fixed ###

none

### API Changes ###

none

### Behavioral Changes ###

'ResourceDictionary' class improvements
- more optimal memory usage in some cases
- best performance of indexer
- new laconic C# 7 code style 

### PR Checklist ###

- [+] Has tests (if omitted, state reason in description)
- [+] Rebased on top of master at time of PR
- [+] Changes adhere to coding standard
- [+] Consolidate commits as makes sense

### DIFF
--- a/Xamarin.Forms.Core/ResourceDictionary.cs
+++ b/Xamarin.Forms.Core/ResourceDictionary.cs
@@ -148,18 +148,9 @@ namespace Xamarin.Forms
 		[IndexerName("Item")]
 		public object this[string index]
 		{
-			get
-			{
-				if (_innerDictionary.ContainsKey(index))
-					return _innerDictionary[index];
-				if (_mergedInstance != null && _mergedInstance.ContainsKey(index))
-					return _mergedInstance[index];
-				if (MergedDictionaries != null)
-					foreach (var dict in MergedDictionaries.Reverse())
-						if (dict.ContainsKey(index))
-							return dict[index];
-				throw new KeyNotFoundException($"The resource '{index}' is not present in the dictionary.");
-			}
+			get => TryGetValue(index, out var value)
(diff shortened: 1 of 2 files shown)


rmarinho:
What problem  does this fix?

Makeloft:
Hi! This fix contains few small improvements (not very serious):
- add C# 7 support for old build tools (for example versions of Visual Studio below 2017)
- best performance of ResourceDictionary indexer (use ‘TryGetValue’ method instead construction ‘if (dic.ContainsKey(key)) return dic[key]’)
- more optimal memory usage when use MergedDictionaries (avoid redundant ‘ToArray’ calls)
- apply C# 7 code style to the class

Makeloft:
Also in some places getting via property 'MergedDictionaries' (for example checks 'MergedDictionaries != null') broke lazy initialization of MergedDictionaries collection. Used backing field '_mergedDictionaries' instead to fix this.

StephaneDelcroix:
Thanks for this PR. We understand you've invested some time in doing this, but we can't accept it as is. Please don't mix your fixes with format changes or refactoring, it makes everything hard to understand.
Please refer to http://tirania.org/blog/archive/2010/Dec-31.html as reference, for example.

PR #1214, opened 2017-10-20
[Internal] Update obsolete messages for 2.5

samhouts:
### Description of Change ###

Just fixing the obsolete messages to say 2.5 instead of 3.0


StephaneDelcroix:
shouldn't this then be targeted at `15-5` ?

samhouts:
Pushed the commits directly to 15-5 and master.

PR #1215, opened 2017-10-20
ResourceDictionary class refactoring

Makeloft:
### Description of Change ###

Refactoring of ResourceDictionary class.

### Bugs Fixed ###

none

### API Changes ###

none

### Behavioral Changes ###

More optimal memory usage and better performance in some cases.

### PR Checklist ###

- [ +] Has tests (if omitted, state reason in description)
- [ +] Rebased on top of master at time of PR
- [ +] Changes adhere to coding standard
- [ +] Consolidate commits as makes sense

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/ResourceDictionaryTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/ResourceDictionaryTests.cs
@@ -481,5 +481,17 @@ namespace Xamarin.Forms.Core.UnitTests
 			rd[commonKey] = "Value";
 			Assert.That(rd[commonKey], Is.EqualTo("Value"));
 		}
+
+		[Test]
+		public void MergedDictionariesDeepSublevels()
+		{
+			var rd0 = new ResourceDictionary();
+			var rd1 = new ResourceDictionary();
+			var rd2 = new ResourceDictionary { { "Key2", "Level2" } };
+			rd1.MergedDictionaries.Add(rd2);
+			rd0.MergedDictionaries.Add(rd1);
+
+			Assert.That(rd0["Key2"], Is.EqualTo("Level2"));
+		}
 	}
 }
--- a/Xamarin.Forms.Core.UnitTests/ResourceDictionaryTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/ResourceDictionaryTests.cs
@@ -476,9 +476,13 @@ namespace Xamarin.Forms.Core.UnitTests
 			var rd1 = new ResourceDictionary { { commonKey, "Value1" } };
 
 			var rd = new ResourceDictionary { MergedDictionaries = { rd0, rd1 } };
+
+			Assert.That(rd0.ContainsKey(commonKey), Is.EqualTo(true));
 			Assert.That(rd[commonKey], Is.EqualTo("Value1"));
 
 			rd[commonKey] = "Value";
+
+			Assert.That(rd0.ContainsKey(commonKey), Is.EqualTo(true));
 			Assert.That(rd[commonKey], Is.EqualTo("Value"));
 		}
 
@@ -491,6 +495,7 @@ namespace Xamarin.Forms.Core.UnitTests
 			rd1.MergedDictionaries.Add(rd2);
 			rd0.MergedDictionaries.Add(rd1);
 
+			Assert.That(rd0.ContainsKey("Key2"), Is.EqualTo(true));
 			Assert.That(rd0["Key2"], Is.EqualTo("Level2"));
 		}
 	}
--- a/Xamarin.Forms.Core.UnitTests/ResourceDictionaryTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/ResourceDictionaryTests.cs
@@ -467,5 +467,19 @@ namespace Xamarin.Forms.Core.UnitTests
 			rd0.Add("foo", "Foo");
 			Assert.That(label.Text, Is.EqualTo("Foo"));
 		}
+
+		[Test]
+		public void MergedDictionariesOverriding()
+		{
+			var commonKey = "CommonKey";
+			var rd0 = new ResourceDictionary { { commonKey, "Value0" } };
+			var rd1 = new ResourceDictionary { { commonKey, "Value1" } };
+
+			var rd = new ResourceDictionary { MergedDictionaries = { rd0, rd1 } };
+			Assert.That(rd[commonKey], Is.EqualTo("Value1"));
+
+			rd[commonKey] = "Value";
+			Assert.That(rd[commonKey], Is.EqualTo("Value"));
+		}
 	}
 }
--- a/Xamarin.Forms.Core/ResourceDictionary.cs
+++ b/Xamarin.Forms.Core/ResourceDictionary.cs
@@ -200,19 +200,12 @@ namespace Xamarin.Forms
 
 		public bool TryGetValue(string key, out object value)
 		{
-			return _innerDictionary.TryGetValue(key, out value)
-				|| (_mergedInstance != null && _mergedInstance.TryGetValue(key, out value))
-				|| (_mergedDictionaries != null && TryGetMergedDictionaryValue(key, out value));
-		}
-
-		bool TryGetMergedDictionaryValue(string key, out object value)
-		{
-			foreach (var dictionary in _mergedDictionaries.Reverse())
-				if (dictionary.TryGetValue(key, out value))
-					return true;
-
-			value = null;
-			return false;
+			object outValue;
+			var containsKey = _innerDictionary.TryGetValue(key, out outValue)
+				|| _mergedInstance != null && _mergedInstance.TryGetValue(key, out outValue)
+				|| _mergedDictionaries != null && _mergedDictionaries.Reverse().Any(d => d.TryGetValue(key, out outValue));
+			value = outValue;
+			return containsKey;
 		}
 
 		event EventHandler<ResourcesChangedEventArgs> IResourceDictionary.ValuesChanged
--- a/Xamarin.Forms.Core/ResourceDictionary.cs
+++ b/Xamarin.Forms.Core/ResourceDictionary.cs
@@ -38,14 +38,13 @@ namespace Xamarin.Forms
 
 		ICollection<ResourceDictionary> _mergedDictionaries;
 		public ICollection<ResourceDictionary> MergedDictionaries {
-			get {
-				if (_mergedDictionaries == null) {
-					var col = new ObservableCollection<ResourceDictionary>();
-					col.CollectionChanged += MergedDictionaries_CollectionChanged;
-					_mergedDictionaries = col;
-				}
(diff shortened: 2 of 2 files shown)


Xamarin.Forms.Core/ResourceDictionary.cs
```diff
@@ -150,14 +149,8 @@ public bool ContainsKey(string key)
...
-				if (_innerDictionary.ContainsKey(index))
-					return _innerDictionary[index];
-				if (_mergedInstance != null && _mergedInstance.ContainsKey(index))
-					return _mergedInstance[index];
-				if (MergedDictionaries != null)
-					foreach (var dict in MergedDictionaries.Reverse())
-						if (dict.ContainsKey(index))
-							return dict[index];
+				object value;
+				if (TryGetValue(index, out value)) return value;
```
StephaneDelcroix: it make sense to only invoke `ContainsKey` to sub directories in `ContainsKey`, and not use `TryGetValue()`

Xamarin.Forms.Core/ResourceDictionary.cs
```diff
@@ -32,20 +32,19 @@ public class ResourceDictionary : IResourceDictionary, IDictionary<string, objec
...
-					var col = new ObservableCollection<ResourceDictionary>();
-					col.CollectionChanged += MergedDictionaries_CollectionChanged;
-					_mergedDictionaries = col;
-				}
-				return _mergedDictionaries;
-			}
+			get { return _mergedDictionaries ?? (_mergedDictionaries = CreateMergedDictionariesCollection()); }
+		}
+
+		private ICollection<ResourceDictionary> CreateMergedDictionariesCollection() {
```
StephaneDelcroix: No point in splitting this in a separate method

Xamarin.Forms.Core.UnitTests/ResourceDictionaryTests.cs
```diff
@@ -467,5 +467,19 @@ public void AddingResourceInMergedRDTriggersValueChanged()
 			rd0.Add("foo", "Foo");
 			Assert.That(label.Text, Is.EqualTo("Foo"));
 		}
+
+		[Test]
+		public void MergedDictionariesOverriding()
```
StephaneDelcroix: was this test failing without your patch, and passing now ?

Xamarin.Forms.Core/ResourceDictionary.cs
```diff
@@ -207,19 +200,12 @@ IEnumerator IEnumerable.GetEnumerator()
...
-					return true;
-
-			value = null;
-			return false;
+			object outValue;
+			var containsKey = _innerDictionary.TryGetValue(key, out outValue)
+				|| _mergedInstance != null && _mergedInstance.TryGetValue(key, out outValue)
+				|| _mergedDictionaries != null && _mergedDictionaries.Reverse().Any(d => d.TryGetValue(key, out outValue));
+			value = outValue;
+			return containsKey;
```
StephaneDelcroix: and here, no real point in merging

Xamarin.Forms.Core/ResourceDictionary.cs
```diff
@@ -248,11 +234,14 @@ void OnValueChanged(string key, object value)
...
-		void OnValuesChanged(params KeyValuePair<string, object>[] values)
+		void OnValuesChanged(KeyValuePair<string, object> pair)
 		{
-			if (values == null || values.Length == 0)
-				return;
-			ValuesChanged?.Invoke(this, new ResourcesChangedEventArgs(values));
+			ValuesChanged?.Invoke(this, new ResourcesChangedEventArgs(new[] { pair }));
+		}
+
+		void OnValuesChanged(IEnumerable<KeyValuePair<string, object>> pairs)
```
StephaneDelcroix: this, and the removal to `.ToArray()` invocation might be worth it, but I'm quite it was done for a reason. I have to evaluate if the reason still stands or if we can accept this part of the patch

Xamarin.Forms.Core/ResourceDictionary.cs
Makeloft: Seems, that there is bug at current master branch.
I don't know why but current implementation of ContainsKey of ResourceDictionary look follow
```javascript
		public bool ContainsKey(string key)
		{
			return _innerDictionary.ContainsKey(key);
		}
```
It is not include merged resources and when we try use ContainKey into indexer we lost recursive merged dictionaries [found values on the first surface level of MergedDictionaries collection].
```javascript
                if (_innerDictionary.ContainsKey(index))
                    return _innerDictionary[index];
                if (_mergedInstance != null && _mergedInstance.ContainsKey(index))
                    return _mergedInstance[index];
                if (_mergedDictionaries != null)
                    foreach (var dict in _mergedDictionaries.Reverse())
                        if (dict.ContainsKey(index))
                            return dict[index];
```
It possible check with new test method [fails on the current master branch]
```javascript
		[Test]
		public void MergedDictionariesDeepSublevels()
		{
			var rd0 = new ResourceDictionary();
			var rd1 = new ResourceDictionary();
			var rd2 = new ResourceDictionary { { "Key2", "Level2" } };
			rd1.MergedDictionaries.Add(rd2);
			rd0.MergedDictionaries.Add(rd1);

			Assert.That(rd0["Key2"], Is.EqualTo("Level2"));
		}
```
Also TryGetValue more fast than construction 'if (dic.ContainsKey(key)) dic[key]' and preffered to use in the current situation. As bonus, allow use C# 7 function code style.

Xamarin.Forms.Core/ResourceDictionary.cs
Makeloft: This splitting allows to use more expressive C# 7 code style.
```javascript
		ICollection<ResourceDictionary> _mergedDictionaries;
		public ICollection<ResourceDictionary> MergedDictionaries
		{
			get
			{
				if (_mergedDictionaries == null)
				{
					var col = new ObservableCollection<ResourceDictionary>();
					col.CollectionChanged += MergedDictionaries_CollectionChanged;
					_mergedDictionaries = col;
				}
				return _mergedDictionaries;
			}
			get { return _mergedDictionaries ?? (_mergedDictionaries = CreateMergedDictionariesCollection()); }
		}
```
Current spitted code may converted at next reformatting step into more clean view
```javascript
		ICollection<ResourceDictionary> _mergedDictionaries;
		public ICollection<ResourceDictionary> MergedDictionaries =>
			_mergedDictionaries ?? (_mergedDictionaries = CreateMergedDictionariesCollection());

		private ICollection<ResourceDictionary> CreateMergedDictionariesCollection() {
			var collection = new ObservableCollection<ResourceDictionary>();
			collection.CollectionChanged += MergedDictionaries_CollectionChanged;
			return collection;
		}
```

Xamarin.Forms.Core/ResourceDictionary.cs
Makeloft: Please, look at the property MergedResources  which uses Linq for similar _mergedDictionaries enumerating.
```javascript
		internal IEnumerable<KeyValuePair<string, object>> MergedResources {
			get {
				if (_mergedDictionaries != null)
					foreach (var r in _mergedDictionaries.Reverse().SelectMany(x => x.MergedResources))
						yield return r;
				if (_mergedInstance != null)
					foreach (var r in _mergedInstance.MergedResources)
						yield return r;
				foreach (var r in _innerDictionary)
					yield return r;
			}
		}
```
Now look at current implementation from master brunch of TryGetValue which uses custom method for enumeration.
```javascript
		public bool TryGetValue(string key, out object value)
		{
			return _innerDictionary.TryGetValue(key, out value)
			       || (_mergedInstance != null && _mergedInstance.TryGetValue(key, out value))
			       || (_mergedDictionaries != null && TryGetMergedDictionaryValue(key, out value));
		}

		bool TryGetMergedDictionaryValue(string key, out object value)
		{
			foreach (var dictionary in _mergedDictionaries.Reverse())
				if (dictionary.TryGetValue(key, out value))
					return true;

			value = null;
			return false;
		}
```

And compare with laconic Linq approach
```javascript
		public bool TryGetValue(string key, out object value)
		{
			var containsKey = _innerDictionary.TryGetValue(key, out var outValue)
				|| _mergedInstance != null && _mergedInstance.TryGetValue(key, out outValue)
				|| _mergedDictionaries != null && _mergedDictionaries.Reverse().Any(d => d.TryGetValue(key, out outValue));
			value = outValue;
			return containsKey;
		}
```

Seems, that second variant of TryGetValue looks more pretty..

Xamarin.Forms.Core/ResourceDictionary.cs
Makeloft: Seems, that ToArray call was really required by signature of method
```javascript
        void OnValuesChanged(params KeyValuePair<string, object>[] values)
         {
            if (values == null || values.Length == 0)
                return;
            ValuesChanged?.Invoke(this, new ResourcesChangedEventArgs(values));
        }
```
which may be safely spitted on two
```javascript
       void OnValuesChanged(KeyValuePair<string, object> pair)
       {
            ValuesChanged?.Invoke(this, new ResourcesChangedEventArgs(new[] { pair }));
       }

       void OnValuesChanged(IEnumerable<KeyValuePair<string, object>> values)
       {
            if (values == null || !values.Any())
                return;
            ValuesChanged?.Invoke(this, new ResourcesChangedEventArgs(values.Select(v => v));
        }
```
We may use projection 'values.Select(v => v)' to avoid sending of reference to original Merged Resource Dictionary into event handler.
Also we can remove checks because it is not necessary. Even if we add empty Resource Dictionary, nothing fear do not happens.

In result we have beauty code
```javascript
       void OnValuesChanged(KeyValuePair<string, object> pair) =>
            ValuesChanged?.Invoke(this, new ResourcesChangedEventArgs(new[] { pair }));

       void OnValuesChanged(IEnumerable<KeyValuePair<string, object>> values) =>
            ValuesChanged?.Invoke(this, new ResourcesChangedEventArgs(values.Select(v => v));
```

Xamarin.Forms.Core/ResourceDictionary.cs
StephaneDelcroix: that's not a bug. it's on purpose

Makeloft: Can you explain why?

How can I use hierarchically merged dictionaries like on WPF or UWP ?

StephaneDelcroix:
so it's all refactoring, no real benefits, no bugs fixed. thanks but no thanks. We do value beautifully laid code, but we do not want to spend any cycle, and potentially loosing part of history, doing just that.

Makeloft:
Stephane, do you listen me? Current implementation contains the real bug. Merged Dictionaries works different on WPF and UWP.

Makeloft:
Reopened here https://github.com/xamarin/Xamarin.Forms/pull/1218

Makeloft:
https://bugzilla.xamarin.com/show_bug.cgi?id=60316

PR #1216, opened 2017-10-21
Simplify row height estimation

kingces95:
### Description of Change ###

Expected row height estimation to be simple. We actually do quite a bit of work to compute the exact height of the first cell (for RetainElement) and use that as the estimate for the remaining cells. This adds a bunch of complexity. For example, life cycle management of the renderer used for estimation had a bug which is causes the unexpected behavior reported here.

This change simply always returns a constant for the row height estimation.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=59813

### API Changes ###

None

### Behavioral Changes ###

Return a constant for row height estimation. Let's see if any tests fail.

### PR Checklist ###


rmarinho:
tire's a uitest failing

kingces95:
Going with this fix instead: https://github.com/xamarin/Xamarin.Forms/pull/1226

PR #1217, opened 2017-10-22
[Android] Updated method "UpdateToolBar" in NavigationPageRenderer

JoergMeier106:
Support of overriding of method "ShouldShowToolbarButton" in MasterDetailPage

### Description of Change ###
Instead of setting toggle.DrawerIndicatorEnabled to true, it is now set to the
value which will returned by MasterDetailPage.ShouldShowToolbarButton();
Describe your changes here.

### Bugs Fixed ###
https://bugzilla.xamarin.com/show_bug.cgi?id=60311

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
@@ -42,6 +42,7 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 		ToolbarTracker _toolbarTracker;
 		DrawerMultiplexedListener _drawerListener;
 		DrawerLayout _drawerLayout;
+		MasterDetailPage _masterDetailPage;
 		bool _toolbarVisible;
 
 		// The following is based on https://android.googlesource.com/platform/frameworks/support/+/refs/heads/master/v4/java/android/support/v4/app/FragmentManager.java#849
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
```diff
@@ -829,9 +828,9 @@ void UpdateToolbar()
 			}
 			else
 			{
-				if (toggle != null)
+				if (toggle != null && _masterDetailPage != null)
```
samhouts: I'm wondering if we should just do the work to find the `MasterDetailPage` again here inside this `if` statement. I don't like the idea of holding a reference to a child page inside of a `NavigationPage`. Seems like we might get a memory leak at best and an `ObjectDisposedException` at worst if the hierarchy changes and that reference doesn't get updated soon enough.

JoergMeier106: But we already do hold a reference of the renderer of `MasterDetailPage` in `_drawerLayout`. If we need to get a fresh reference of `MasterDetailPage`, we have to update `_drawerLayout` and `_drawerToggle`, which is made out of `_drawerlayout`, as well.
So basically, we don't get any advantage of doing the work in `UpdateToolbar` again without also updating `toggle`.

PR #1218, opened 2017-10-23
[Core] Allow Merged Dirctionaries hierarchically merging

Makeloft:
### Description of Change ###

Current implementation of Merged Dictionaries on master can not merge dictionaries hierarchically. Just try to run following test method on master, current brunch and WPF or UWP.

	    public static void MergedDictionariesDeepSublevels()
	    {
		    var rd0 = new ResourceDictionary();
		    var rd1 = new ResourceDictionary { { "Key1", "Level1" } };
		    var rd2 = new ResourceDictionary { { "Key2", "Level2" } };

		    rd0.MergedDictionaries.Add(rd1);
		    rd1.MergedDictionaries.Add(rd2);

		    Assert.That(rd0["Key1"], "Level1");
		    Assert.That(rd0["Key2"], "Level2"); // <= fail on master, work on WPF or UWP

		    //Assert.That(rd0.ContainsKey("Key1"), Is.EqualTo(true));
		    //Assert.That(rd0.ContainsKey("Key2"), Is.EqualTo(true));
	    }

Reopen
https://github.com/xamarin/Xamarin.Forms/pull/1215

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=60316 
fixed by commit https://github.com/xamarin/Xamarin.Forms/pull/1218/commits/e22367ebe90783579fa4a7b3b66da196d41a1960
possible, it needs to fix ContainsKey method too https://github.com/xamarin/Xamarin.Forms/pull/1218/commits/bebf303ffa0f187b769a712273e3b92c35d1c587

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/ResourceDictionaryTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/ResourceDictionaryTests.cs
@@ -481,5 +481,17 @@ namespace Xamarin.Forms.Core.UnitTests
 			rd[commonKey] = "Value";
 			Assert.That(rd[commonKey], Is.EqualTo("Value"));
 		}
+
+		[Test]
+		public void MergedDictionariesDeepSublevels()
+		{
+			var rd0 = new ResourceDictionary();
+			var rd1 = new ResourceDictionary();
+			var rd2 = new ResourceDictionary { { "Key2", "Level2" } };
+			rd1.MergedDictionaries.Add(rd2);
+			rd0.MergedDictionaries.Add(rd1);
+
+			Assert.That(rd0["Key2"], Is.EqualTo("Level2"));
+		}
 	}
 }
--- a/Xamarin.Forms.Core.UnitTests/ResourceDictionaryTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/ResourceDictionaryTests.cs
@@ -476,9 +476,13 @@ namespace Xamarin.Forms.Core.UnitTests
 			var rd1 = new ResourceDictionary { { commonKey, "Value1" } };
 
 			var rd = new ResourceDictionary { MergedDictionaries = { rd0, rd1 } };
+
+			Assert.That(rd0.ContainsKey(commonKey), Is.EqualTo(true));
 			Assert.That(rd[commonKey], Is.EqualTo("Value1"));
 
 			rd[commonKey] = "Value";
+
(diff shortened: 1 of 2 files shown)


bugadani:
If I may comment as an outsider: Instead of reopening rightfully denied pull requests, please understand the reason it got denied and only re-submit parts of the patch that are actually fixing the problem. Refactoring and cleanup is fine on your own projects, but I'm not sure it brings your patch closer to acceptance, rather, the opposite. In your place I would split this patch apart, submit the perf improvements and the test case fix separately (also, you may open a bugzilla ticket just for reference of your bug that you can link to this patch), without any of the LINQ code and other refactorings. C#7 is neat, but spending a patch trying to refactor a tiny part of the code base does not really have any value. "Looks nicer" really is subjective, in this case.

Makeloft:
If you really consider that previous pull request was "rightfully denied" you may fix this bug by yourself.
All my fixes split at separated at logically elaborate commites that not hard to review.

Official reason of closing previous PR
"so it's all refactoring, no real benefits, no bugs fixed. thanks but no thanks."

But in comments I clearly describe which bug is occurs at current master brunch. Seems, that my comments even nobody read and want to close PR as quickly as possible...

Makeloft:
https://bugzilla.xamarin.com/show_bug.cgi?id=60316

Makeloft:
Updated from bug to feature request according Adam’s comment on https://bugzilla.xamarin.com/show_bug.cgi?id=60316

Makeloft:
Placed to the Evolution forum
https://forums.xamarin.com/discussion/105480/allow-merged-dirctionaries-hierarchically-merging

jassmith:
This is by design, and at this time we dont intend to change this behavior. More importantly however a change to this behavior would be a potentially breaking change that even if we wanted to take, we couldn't.

Sorry :/

Makeloft:
@jassmith 
Actually there are not any breaking changes. It is only the functional extension and the small performance optimization which allows to avoid redundant ToArray call.

If you want I can recreate similar PR according with new changes.

Currently I forced to use custom solution described by a link
https://stackoverflow.com/questions/44640549/using-resourcedictionary-mergeddictionaries-in-app-xaml

Sorry but merged dictionaries without support of recursive merging is mostly useless for me. :)

I don't understand why you afraid, all tests have been passed successful.

mmierzwa:
I agree with @Makeloft - without possibility of recursive merging `MergedDictionaries` are not very useful.

PR #1221, opened 2017-10-23
Fix NRE introduced by 59882

kingces95:
### Description of Change ###

Fix NRE introduced by 59882. Expect application can be null before OnStart is called. Actually, while fixing another bug, assumed it was non-null which caused NRE.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=60337

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
@@ -24,6 +24,7 @@ using ARelativeLayout = Android.Widget.RelativeLayout;
 using Xamarin.Forms.Internals;
 using System.Threading.Tasks;
 
+
 #endregion
 
 namespace Xamarin.Forms.Platform.Android
@@ -126,11 +127,13 @@ namespace Xamarin.Forms.Platform.Android
 
 			application.PropertyChanged += AppOnPropertyChanged;
 
-			var iver = Platform.GetRenderer(application.MainPage);
-			if (iver != null)
+			if (application?.MainPage != null)
 			{
-				iver.Dispose();
-				application.MainPage.ClearValue(Platform.RendererProperty);
+				var iver = Platform.GetRenderer(application.MainPage);
+				if (iver != null) {
+					iver.Dispose();
+					application.MainPage.ClearValue(Platform.RendererProperty);
+				}
 			}
 
 			SetMainPage();


Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -126,11 +126,13 @@ protected void LoadApplication(Application application)
 
 			application.PropertyChanged += AppOnPropertyChanged;
 
-			var iver = Platform.GetRenderer(application.MainPage);
-			if (iver != null)
+			if (application != null)
```
bugadani: Shouldn't the event subscription in L127 be also after the check?

jassmith: Yes

kingces95: Huh, L127 that was there before the fix to 59882: https://github.com/xamarin/Xamarin.Forms/pull/1212/files

bugadani: Actually there is already a null check in lines 116-117 so either the issue is a bit more intricate or the fix is unnecessary. Or I'm not seeing something at 10:30PM

shanwayne: Please see updated Bugzilla comment for actual issue.

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -126,11 +126,13 @@ protected void LoadApplication(Application application)
 
 			application.PropertyChanged += AppOnPropertyChanged;
 
-			var iver = Platform.GetRenderer(application.MainPage);
-			if (iver != null)
+			if (application?.MainPage != null)
```
bugadani: The null check of application is unnecessary because it is already done on line 116.

kingces95: True. But we'll let it go anyway.

kingces95:
This change should only impact startup on android. So this XTC failure is likely spurious. 

image

Running CI baseline test here: https://github.com/xamarin/Xamarin.Forms/pull/1160

kingces95:
Failures match CI baseline run: https://github.com/xamarin/Xamarin.Forms/pull/1160

AndersNielsen89:
Has this commit been includede in a release yet?

PR #1222, opened 2017-10-24
[All] Basic Right-To-Left Support

samhouts:
### Description of Change ###

Adds support for right-to-left layouts.

As with UWP, respecting the device's layout direction based on the selected language/locale is the developer's explicit choice and does not happen automatically. Set the `FlowDirection` of the root View/Page to the `Device.FlowDirection` value. All other child views with `MatchParent` will inherit this value. 

All Views default to the `MatchParent` `FlowDirection`. A `View` without a `Parent` will default to `LeftToRight`, and this value will be recursively passed down to the `View`'s children, unless they have an explicitly set `FlowDirection` of their own.

Renderers can access the `IVisualElementController.EffectiveFlowDirection` to get the actual specified direction for a View/Page and whether the value was inherited (implicit) or explicitly set.

Note that you should set the `FlowDirection` once on initial layout. Changing this value causes an expensive relayout and will affect performance.

#### General
- NavigationPage button location, toolbar item location, and transition animation is currently controlled by the device locale, not the FlowDirection setting.
- CarouselPage swipe direction does not flip.
- Images need to mirror when they are explicitly set to right-to-left. May also need to support asset suffixes.
- TextDirection property needs to be added.
- Global app setting for FlowDirection

#### Android
- Searchbar orientation is controlled by the device locale, not the FlowDirection setting.
- ContextAction placement is controlled by the device locale, not the FlowDirection setting.

#### iOS
- Stepper orientation is controlled by the device locale, not the FlowDirection setting.
- EntryCell  text alignment is controlled by the device locale, not the FlowDirection setting.
- ContextAction gesture and alignment are not flipped.

#### UWP
- ContextAction text alignment is controlled by the device locale, not the FlowDirection setting.
- Editor text alignment is controlled by the device locale, not the FlowDirection setting.
- FlowDirection does not appear to be inherited by MasterDetailPage children.

#### macOS
- MasterDetailPage Master page does not pop out from the right in RTL
- Text alignment is controlled by the device locale, not the FlowDirection setting.
- ProgressBar is not controlled by the FlowDirection setting.
- ImageCell, SwitchCell, TextCell are not controlled by the FlowDirection setting.
- Searchbar orientation is controlled by the device locale, not the FlowDirection setting.

#### Android
Users will need to set `android:supportsRtl="true"` and set Android's `MinSDKVersion` to 17 or higher.

To test device-specified RTL on Android, either select a language that is RTL or enable `Force RTL layout direction` in the Developer Options. You will also need to add your language to the list of languages your app supports in the Info.plist. See https://developer.xamarin.com/guides/ios/advanced_topics/localization_and_internationalization/ for more information. If you do change your language and region, you may encounter an exception with `DatePicker`s. See Bug 59077 for more information and a workaround.

#### iOS

There's no easy way to test device-specified RTL on Xamarin.iOS. The `pseudolanguage` options are Xcode only. You will need to change the language _and_ region to an RTL locale in order to test it. Please note that once you do so, `DatePicker`s will throw an exception if you do not include the resources required for alternate calendars. For example, if you're testing in Arabic, be sure to select `mideast` in the `iOSBuild` `Internationalization` section.

#### UWP

Similar to iOS, you will need to change the language _and_ region to an RTL locale in order to test device-specified RTL. You will also need to add the language resource to your `Package.appxmanifest` for your app to support it.

#### Android

> Before
android-before

>After (LTR)
android-after-ltr

>After (RTL Forced)
android-after-rtl

>After (RTL Locale)
android-after-rtl-locale

ios-before

>After (LTR)
ios-after-ltr

>After (RTL Forced)
ios-after-rtl

>After (RTL Locale)
ios-after-rtl-locale

#### UWP

>Before
uwp-before

>After (LTR)
uwp-after-ltr

>After (RTL Forced)
uwp-after-rtl

>After (RTL Locale)
uwp-atfer-rtl-locale

### Bugs Fixed ###

- N/A

### API Changes ###

Added:
- `enum EffectiveFlowDirection`
- `enum FlowDirection`
- `IVisualElementController.EffectiveFlowDirection`
- `VisualElement.FlowDirection`
- `Device.FlowDirection`

#### Android
- For API >=17, we now set the `TextAlignment` instead of the `Gravity` of the `EditText` when we change the `HorizontalTextAlignment` property.

### PR Checklist ###

- [x] Has tests (manual!)

### DIFF
--- a/Xamarin.Forms.ControlGallery.Android/Properties/AndroidManifest.xml
+++ b/Xamarin.Forms.ControlGallery.Android/Properties/AndroidManifest.xml
@@ -16,7 +16,7 @@
 	<uses-permission android:name="android.permission.READ_LOGS" />
 	<uses-permission android:name="android.permission.BIND_NOTIFICATION_LISTENER_SERVICE" />
 	<uses-feature android:glEsVersion="0x00020000" android:required="true" />
-	<application android:label="AndroidControlGallery">
+	<application android:label="AndroidControlGallery" android:supportsRtl="true">
 		<meta-data android:name="com.google.android.gms.version" android:value="@integer/google_play_services_version" />
 	</application>
 </manifest>
--- a/Xamarin.Forms.Controls/ControlGalleryPages/FlowDirectionGallery.cs
+++ b/Xamarin.Forms.Controls/ControlGalleryPages/FlowDirectionGallery.cs
@@ -1,0 +1,429 @@
+﻿using System;
+using System.Linq;
+
+namespace Xamarin.Forms.Controls
+{
+
+	public class FlowDirectionGalleryLandingPage : ContentPage
+	{
+		FlowDirection DeviceDirection => Device.FlowDirection;
+
+		public FlowDirectionGalleryLandingPage()
+		{
+
+			var np = new Button { Text = "Try NavigationPage", Command = new Command(() => { PushNavigationPage(DeviceDirection); }) };
+			var mdp = new Button { Text = "Try MasterDetailPage", Command = new Command(() => { PushMasterDetailPage(DeviceDirection); }) };
+			var crp = new Button { Text = "Try CarouselPage", Command = new Command(() => { PushCarouselPage(DeviceDirection); }) };
+			var tp = new Button { Text = "Try TabbedPage", Command = new Command(() => { PushTabbedPage(DeviceDirection); }) };
+			var cp = new Button { Text = "Try ContentPage", Command = new Command(() => { PushContentPage(DeviceDirection); }) };
+
+			Content = new StackLayout
+			{
+				Children = {
+					new Label { Text = "Click a button below to swap the MainPage of the app." },
+					np,
+					mdp,
+					crp,
+					tp,
+					cp
+				}
+			};
+		}
+
+		public static void PushNavigationPage(FlowDirection direction)
+		{
+			((App)Application.Current).SetMainPage(new FlowDirectionGalleryNP(direction));
+		}
+
+		public static void PushMasterDetailPage(FlowDirection direction)
+		{
+			((App)Application.Current).SetMainPage(new FlowDirectionGalleryMDP(direction));
+		}
+
+		public static void PushCarouselPage(FlowDirection direction)
+		{
+			((App)Application.Current).SetMainPage(new FlowDirectionGalleryCarP(direction));
+		}
+
+		public static void PushTabbedPage(FlowDirection direction)
+		{
+			((App)Application.Current).SetMainPage(new FlowDirectionGalleryTP(direction));
+		}
+
+		public static void PushContentPage(FlowDirection direction)
+		{
+			((App)Application.Current).SetMainPage(new FlowDirectionGalleryCP(direction)
+			{
+				FlowDirection = direction
+			});
+		}
+	}
+
+	public class FlowDirectionGalleryNP : NavigationPage
+	{
+		public FlowDirectionGalleryNP(FlowDirection direction)
+		{
+			FlowDirection = direction;
+			Navigation.PushAsync(new FlowDirectionGalleryCP(direction));
+			Navigation.PushAsync(new FlowDirectionGalleryCP(direction));
+		}
+	}
+
+	public class FlowDirectionGalleryMDP : MasterDetailPage
+	{
+		public FlowDirectionGalleryMDP(FlowDirection direction)
+		{
+			FlowDirection = direction;
+			Master = new FlowDirectionGalleryCP(direction) { Title = "Master", BackgroundColor = Color.Red };
+			Detail = new FlowDirectionGalleryCP(direction) { Title = "Detail" };
+			IsPresented = true;
+		}
+	}
+
+	public class FlowDirectionGalleryCarP : CarouselPage
+	{
+		public FlowDirectionGalleryCarP(FlowDirection direction)
+		{
+			FlowDirection = direction;
+			Children.Add(new FlowDirectionGalleryCP(direction) { Title = "1" });
+			Children.Add(new FlowDirectionGalleryCP(direction) { Title = "2" });
+			Children.Add(new FlowDirectionGalleryCP(direction) { Title = "3" });
+		}
+	}
+
+	public class FlowDirectionGalleryTP : TabbedPage
+	{
+		public FlowDirectionGalleryTP(FlowDirection direction)
+		{
+			FlowDirection = direction;
+			Children.Add(new FlowDirectionGalleryCP(direction) { Title = "1", BackgroundColor = Color.Red });
+			Children.Add(new FlowDirectionGalleryCP(direction) { Title = "2", BackgroundColor = Color.Orange });
+			Children.Add(new FlowDirectionGalleryCP(direction) { Title = "3", BackgroundColor = Color.Yellow });
+		}
+	}
+
+	public class FlowDirectionGalleryCP : ContentPage
+	{
+		FlowDirection DeviceDirection => Device.FlowDirection;
+
+		Page ParentPage => (Parent as Page) ?? this;
+
+		public FlowDirectionGalleryCP(FlowDirection direction)
+		{
+			var item = new ToolbarItem
+			{
+				Icon = "coffee.png",
+				Text = "Item 1",
+			};
+
+			var item2 = new ToolbarItem
+			{
+				Icon = "bank.png",
+				Text = "Item 2",
+			};
+
+			ToolbarItems.Add(item);
+			ToolbarItems.Add(item2);
+
+			Title = "Flow Direction Gallery";
+			NavigationPage.SetHasBackButton(this, true);
+			NavigationPage.SetBackButtonTitle(this, "Back");
+			SetContent(direction);
+		}
+
+		void SetContent(FlowDirection direction)
+		{
+			var hOptions = LayoutOptions.Start;
+
+			var imageCell = new DataTemplate(typeof(ImageCell));
+			imageCell.SetBinding(ImageCell.ImageSourceProperty, ".");
+			imageCell.SetBinding(ImageCell.TextProperty, ".");
+
+			var textCell = new DataTemplate(typeof(TextCell));
+			textCell.SetBinding(TextCell.DetailProperty, ".");
+
+			var entryCell = new DataTemplate(typeof(EntryCell));
+			entryCell.SetBinding(EntryCell.TextProperty, ".");
+
+			var switchCell = new DataTemplate(typeof(SwitchCell));
+			switchCell.SetBinding(SwitchCell.OnProperty, ".");
+			switchCell.SetValue(SwitchCell.TextProperty, "Switch Cell!");
+
+			var vc = new ViewCell
+			{
+				View = new StackLayout
+				{
+					Children = { new Label { HorizontalOptions = hOptions, Text = "View Cell! I have context actions." } }
+				}
+			};
+
+			var a1 = new MenuItem { Text = "First" };
+			vc.ContextActions.Add(a1);
+			var a2 = new MenuItem { Text = "Second" };
+			vc.ContextActions.Add(a2);
+
+			var viewCell = new DataTemplate(() => vc);
+
+			var relayout = new Switch
+			{
+				IsToggled = true,
+				HorizontalOptions = LayoutOptions.End,
+				VerticalOptions = LayoutOptions.Center
+			};
+
+			var flipButton = new Button
+			{
+				Text = direction == FlowDirection.RightToLeft ? "Switch to Left To Right" : "Switch to Right To Left",
+				HorizontalOptions = LayoutOptions.StartAndExpand,
+				VerticalOptions = LayoutOptions.Center
+			};
+
+			flipButton.Clicked += (s, e) =>
+			{
+				FlowDirection newDirection;
+				if (direction == FlowDirection.LeftToRight || direction == FlowDirection.MatchParent)
+					newDirection = FlowDirection.RightToLeft;
+				else
+					newDirection = FlowDirection.LeftToRight;
+
+				if (relayout.IsToggled)
+				{
+					ParentPage.FlowDirection = newDirection;
+
+					direction = newDirection;
+
+					flipButton.Text = direction == FlowDirection.RightToLeft ? "Switch to Left To Right" : "Switch to Right To Left";
+
+					return;
+				}
+
+				if (ParentPage == this)
+				{
+					FlowDirectionGalleryLandingPage.PushContentPage(newDirection);
+					return;
+				}
+				string parentType = ParentPage.GetType().ToString();
+				switch (parentType)
+				{
+					case "Xamarin.Forms.Controls.FlowDirectionGalleryMDP":
+						FlowDirectionGalleryLandingPage.PushMasterDetailPage(newDirection);
+						break;
+					case "Xamarin.Forms.Controls.FlowDirectionGalleryCarP":
+						FlowDirectionGalleryLandingPage.PushCarouselPage(newDirection);
+						break;
+					case "Xamarin.Forms.Controls.FlowDirectionGalleryNP":
+						FlowDirectionGalleryLandingPage.PushNavigationPage(newDirection);
+						break;
+					case "Xamarin.Forms.Controls.FlowDirectionGalleryTP":
+						FlowDirectionGalleryLandingPage.PushTabbedPage(newDirection);
+						break;
+				}
+			};
+
+			var horStack = new StackLayout
+			{
+				Orientation = StackOrientation.Horizontal,
+				Children = { flipButton, new Label { Text = "Relayout", HorizontalOptions = LayoutOptions.End, VerticalOptions = LayoutOptions.Center }, relayout }
+			};
+
+			var grid = new Grid
+			{
+				ColumnDefinitions = {
+					new ColumnDefinition { Width = new GridLength(1, GridUnitType.Star) },
+					new ColumnDefinition { Width = new GridLength(1, GridUnitType.Star) },
+				},
+				RowDefinitions = {
+					new RowDefinition { Height = new GridLength(100, GridUnitType.Absolute) },
+					new RowDefinition { Height = new GridLength(100, GridUnitType.Absolute) },
+					new RowDefinition { Height = new GridLength(100, GridUnitType.Absolute) },
+					new RowDefinition { Height = new GridLength(100, GridUnitType.Absolute) },
+					new RowDefinition { Height = new GridLength(100, GridUnitType.Absolute) },
+					new RowDefinition { Height = new GridLength(100, GridUnitType.Absolute) },
+					new RowDefinition { Height = new GridLength(100, GridUnitType.Absolute) },
+				}
+			};
+
+			int col = 0;
+			int row = 0;
+
+			var ai = AddView<ActivityIndicator>(grid, ref col, ref row);
+			ai.IsRunning = true;
+
+			var box = AddView<BoxView>(grid, ref col, ref row);
+			box.WidthRequest = box.HeightRequest = 20;
+			box.BackgroundColor = Color.Purple;
+
+			var btn = AddView<Button>(grid, ref col, ref row);
+			btn.Text = "Some text";
+
+			var date = AddView<DatePicker>(grid, ref col, ref row, 2);
+
+			var edit = AddView<Editor>(grid, ref col, ref row);
+			edit.WidthRequest = 100;
+			edit.HeightRequest = 100;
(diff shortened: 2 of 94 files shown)


maamountki:
In UWP, Editor text is right-aligned if text already have rtl language value regardless of the selected language/locale, Also if keyboard changed to rtl language while editor text is empty, Editor will align text to right as you typing.

rmarinho:
Needs docs update

samhouts:
@maamountki Yes, sorry, what I meant to say is that I am unable to force the text alignment, but it does properly align itself with the device language/locale. Thank you!

Xamarin.Forms.Platform.iOS/Renderers/TabletMasterDetailRenderer.cs
```diff
@@ -346,6 +354,17 @@ void UpdateControllers()
 			_detailController.AddChildViewController(detail);
 		}
 
+		void UpdateFlowDirection()
+		{
+			if (VisualElementController == null || NativeView == null)
+				return;
+
+			if (VisualElementController.EffectiveFlowDirection.HasFlag(EffectiveFlowDirection.RightToLeft))
```
jassmith: HasFlag is super evil and slow. Like really REALLY slow. Don't use HasFlag.

jassmith:
Approved pending tests and fixing HasFlag

Xamarin.Forms.Controls/ControlGalleryPages/FlowDirectionGallery.cs
```diff
@@ -0,0 +1,429 @@
+﻿using System;
+using System.Linq;
+
+namespace Xamarin.Forms.Controls
+{
+
+	public class FlowDirectionGalleryLandingPage : ContentPage
+	{
+		FlowDirection DeviceDirection => Device.Info.CurrentFlowDirection;
```
StephaneDelcroix: is `Device.Info.CurrentFlowDirection` the recommended way to access the FlowDirection?

if so, it causes 2 problems:
- Device.Info is attributed with `[EditorBrowsable(EditorBrowsableState.Never)]`
- it makes it hard to access from a `XAML` page. the FlowDirection should probably be exposed at the `Device` level, so one can do `<ContentPage FlowDirection="{x:Static Device.FlowDirection}" ... />` (`x:Static` syntax is `Type.Member` and doesn't support `Type.Member.Member`)

Xamarin.Forms.Core/BindableObject.cs
```diff
@@ -605,7 +604,8 @@ enum BindableContextAttributes
 			IsBeingSet = 1 << 1,
 			IsDynamicResource = 1 << 2,
 			IsSetFromStyle = 1 << 3,
-			IsDefaultValue = 1 << 4
+			IsDefaultValue = 1 << 4,
+			IsInherited = 1 << 5,
```
StephaneDelcroix: do we support Inherited BPs ? I don't see this flag used anywhere

Xamarin.Forms.Core/Cells/Cell.cs
```diff
@@ -18,6 +19,24 @@ public abstract class Cell : Element, ICellController
 
 		bool _nextCallToForceUpdateSizeQueued;
 
+		EffectiveFlowDirection _effectiveFlowDirection = EffectiveFlowDirection.LeftToRight | EffectiveFlowDirection.Implicit;
+		EffectiveFlowDirection IFlowDirectionController.EffectiveFlowDirection
```
StephaneDelcroix: as you're implementing this explicitly, I guess this is not meant to be set directly by the user, so there's no point in making this property bindable ?

Xamarin.Forms.Core/Cells/Cell.cs
```diff
@@ -18,6 +19,24 @@ public abstract class Cell : Element, ICellController
...
+					return;
+
+				_effectiveFlowDirection = value;
+
+				var ve = (Parent as VisualElement);
+				ve?.InvalidateMeasureInternal(InvalidationTrigger.Undefined);
+				OnPropertyChanged(VisualElement.FlowDirectionProperty.PropertyName);
+			}
+		}
+		IFlowDirectionController FlowController => this;
```
StephaneDelcroix: Style Police here. You're missing a white line. Please pull over !

(well, fix it only if you have other things to push)

StephaneDelcroix: if I understand this right, you added this property so you don't have to cast to `IFlowDirectionController`everytime you want to invoke an explicitly implemented interface method ? If so, make it a field
```
IFlowDirectionController _flowController = this;
```

Xamarin.Forms.Core/DeviceInfo.cs
```diff
@@ -22,6 +23,18 @@ public DeviceOrientation CurrentOrientation
...
 
+		public FlowDirection CurrentFlowDirection
+		{
+			get { return _currentFlowDirection; }
+			set
+			{
+				if (Equals(_currentFlowDirection, value))
+					return;
+				_currentFlowDirection = value;
+				OnPropertyChanged();
```
StephaneDelcroix: Is this ever expected to change? I might be wrong, but the samples aren't listening to INPC for this. So the proper use would be to bind to this value ?

Xamarin.Forms.Core/EffectiveFlowDirection.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+
+namespace Xamarin.Forms
+{
+	[Flags]
+	public enum EffectiveFlowDirection
+	{
+		LeftToRight = 1 << 0,
+		RightToLeft = 1 << 1,
```
StephaneDelcroix: so the EffectiveFlowDirection can be both LTR and RTL ? and both Implicit and Explicit ?

if not, you might want

```
[Flags]
public enum EffectiveFlowDirection
{
    LTR = 1 << 0;
    Implicit = 1<< 1;
}
```

and add helpers to check mask and evaluate for the negative values. Also, as LTR is the default, I'd give it the `0` value, and `1` to RTL

Xamarin.Forms.Core/EffectiveFlowDirectionExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+			{
+				case FlowDirection.MatchParent:
+					return EffectiveFlowDirection.LeftToRight | EffectiveFlowDirection.Implicit;
+				case FlowDirection.LeftToRight:
+					return EffectiveFlowDirection.LeftToRight | mode;
+				case FlowDirection.RightToLeft:
+					return EffectiveFlowDirection.RightToLeft | mode;
+			}
+
+			throw new InvalidOperationException($"Cannot convert {self} to {nameof(EffectiveFlowDirection)}.");
```
StephaneDelcroix: you mean `default:`

Xamarin.Forms.Core/EffectiveFlowDirectionExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+			else if (self.IsRightToLeft() && !self.IsLeftToRight())
+				return FlowDirection.RightToLeft;
+
+			throw new InvalidOperationException($"Cannot convert {self} to {nameof(FlowDirection)}.");
+		}
+
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public static bool IsRightToLeft(this EffectiveFlowDirection self)
+		{
+			return (self & EffectiveFlowDirection.RightToLeft) != 0;
```
StephaneDelcroix: the canonical way to check a flag is `(self & EffectiveFlowDirection.RightToLeft) == EffectiveFlowDirection.RightToLeft`

it's not important when the flag has only a single bit set, but otherwise, it is

Xamarin.Forms.Core/Layout.cs
```diff
@@ -125,6 +126,18 @@ public sealed override SizeRequest GetSizeRequest(double widthConstraint, double
 
 		public static void LayoutChildIntoBoundingRegion(VisualElement child, Rectangle region)
 		{
+			var parent = child.Parent as IFlowDirectionController;
+			bool isRightToLeft = false;
+			if (parent != null)
+			{
+				isRightToLeft = parent.EffectiveFlowDirection.IsRightToLeft();
+				if (isRightToLeft)
+				{
```
StephaneDelcroix: I'd reduce nesting

```
if (parent != null && (isRightToLeft = parent.EffectiveFlowDirection.IsRightToLeft()))
    region = new Rectangle(parent.Width - region.Right, region.Y, region.Width, region.Height);

```

Xamarin.Forms.Core/VisualElement.cs
```diff
@@ -93,6 +93,33 @@ public partial class VisualElement : Element, IAnimatable, IVisualElementControl
...
+		{
+			get { return _effectiveFlowDirection; }
+			set
+			{
+				if (value == _effectiveFlowDirection)
+					return;
+
+				_effectiveFlowDirection = value;
+				InvalidateMeasureInternal(InvalidationTrigger.Undefined);
+				OnPropertyChanged(FlowDirectionProperty.PropertyName);
```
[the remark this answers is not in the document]
StephaneDelcroix: this one isn't bindable, @jassmith. I don't get the comment

Xamarin.Forms.Core/EffectiveFlowDirection.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+
+namespace Xamarin.Forms
+{
+	[Flags]
+	public enum EffectiveFlowDirection
```
StephaneDelcroix: most of the implementation hide `EffectiveFlowDirection` through explicit interface implementation, so I guess this isn't mean to be used by our users at all ? if so, attribute the type or move it to the Internals namespace

Xamarin.Forms.Core/BindableObject.cs
jassmith: sorry thats leftover from me when I was doing something else with this

Xamarin.Forms.Core/Cells/Cell.cs
jassmith: Correct

Xamarin.Forms.Core/EffectiveFlowDirection.cs
jassmith: I realize its a slight bastardization of a Flags enum, but its never used by users (and should probably go in the internals namespace) and removes any ambiguity if a 3rd flow direction or implicit/explict convention comes up.

jassmith: Sam points out EffectiveFlowDirection needs to be public for custom renderers, so no internals namespace

Xamarin.Forms.Core/EffectiveFlowDirection.cs
samhouts: It should only be set by `Core` code (users set `FlowDirection`), but it should be exposed via `IVisualElementController` so renderers can use it.

Xamarin.Forms.Core/Cells/Cell.cs
samhouts: 👮

samhouts: Can't set a field to `this`.

Xamarin.Forms.Core/DeviceInfo.cs
samhouts: I expected app users to be able to change the language/locale without restarting the app, but that doesn't appear to actually be possible. Android is the closest, but it does restart the activity. I'll remove this.

Xamarin.Forms.Core/EffectiveFlowDirection.cs
StephaneDelcroix: API only used by renderers is usually attributed, or in the internals namespace, or both

Xamarin.Forms.Controls/ControlGalleryPages/FlowDirectionGallery.cs
samhouts: I like it. Trying it out now...

Xamarin.Forms.Core/EffectiveFlowDirection.cs
samhouts: Right, but our users will want to access this for their custom renderers, so it feels like it should be exposed.

Xamarin.Forms.ControlGallery.Android/Properties/AndroidManifest.xml
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionCode="1" android:versionName="1.0" package="AndroidControlGallery.AndroidControlGallery" android:installLocation="auto">
-	<uses-sdk android:minSdkVersion="15" />
+	<uses-sdk android:minSdkVersion="17" />
```
hartez: Doesn't this prevent us from testing things for API 15/16 in the ControlGallery?

Xamarin.Forms.Platform.Android/Cells/SwitchCellRenderer.cs
```diff
@@ -58,6 +63,17 @@ void UpdateHeight()
 			_view.SetRenderHeight(Cell.RenderHeight);
 		}
 
+		void UpdateFlowDirection()
```
hartez: This same UpdateFlowDirection code is repeated several times - seems like a candidate for an extension method.

Xamarin.Forms.Core/Device.cs
```diff
@@ -27,6 +27,7 @@ public static class Device
 		public static void SetIdiom(TargetIdiom value) => Idiom = value;
 		public static TargetIdiom Idiom { get; internal set; }
 
+		//TODO: Why are there two of these? This is never used...?
```
StephaneDelcroix: 🤦‍♂️

Xamarin.Forms.Core/EffectiveFlowDirection.cs
StephaneDelcroix: "It's never used by users" seriously contradicts the "needs to be public for custom renderers". Please fix as this is public API

Xamarin.Forms.Platform.WinRT/SwitchRenderer.cs
```diff
@@ -48,5 +58,16 @@ void OnNativeToggled(object sender, RoutedEventArgs routedEventArgs)
...
+		void UpdateFlowDirection()
+		{
+			if (VisualElementController == null || Control == null)
+				return;
+
+			if (VisualElementController.EffectiveFlowDirection.IsRightToLeft())
+				Control.FlowDirection = WFlowDirection.RightToLeft;
+			else if (VisualElementController.EffectiveFlowDirection.IsLeftToRight())
+				Control.FlowDirection = WFlowDirection.LeftToRight;
+		}
```
kingces95: It's not ur fault but it's a shame our conventions lead us to repeat this code in so many places...

hartez: The conventions don't force the repetition of the stuff inside `UpdateFlowDirection` - all of that could be moved into an extension method.

encrypt0r:
Two questions:
1. Is there anyway currently to make master page to slide from right?
2. Is there any plan to overcome the limitations described here anytime soon?
Btw, thank you for all the work, it's awesome 👍

encrypt0r:
I tried it out and it's a very good start. However I have a couple of questions:

 - Can we set `FlowDirection` for `DisplayAlert`?
 - By default `Label` chooses the `FlowDirection` based on its content (RTL for Arabic and Hebrew and LTR otherwise), how can we force a certain `FlowDirection`?

ali-h2010:
I tested the latest prerelease and allow me to say that you did a great job. 
with just one attribute , I can change the flow of the views! Thank you so much and I can't wait for the awesome release.

I faced a strange interaction though depending on the platform.
On iOS, the views are Right to left when I set them so even if the content is in English. But, in Android, only views in Arabic are RTL directed. Is this intentional ? 
from the images that @samhouts shared, I can see that the flow is changing even if the language is not RTL.
I attached this image for clarification
https://justpaste.it/1jfrn

samhouts:
@encrypt0r I believe Alerts and ActionSheets will be set by the device locale at this time. I didn't investigate whether that was something we could manually set for each platform. We'll take a look. As for forcing a certain `FlowDirection`, you can set `FlowDirection` on any `VisualElement`, including `Label`, to override its inherited values.

@ali-h2010 Are you using the Force RTL Direction dev option in Android, or did you change the language and locale? Did you set `android:supportsRtl="true"` and set `MinSDKVersion` to 17 or higher in your manifest?

ali-h2010:
@samhouts Thanks for the update. I actually didn't set those values. I just updated Xamarin forms to the prerelease and used the `FlowDirection`.

 it worked fine on iOS device with English which is not RTL. I did the same with the android device with the lanague set to English but text in Arabic. The results as you saw in the screenshot.

I should have read your post here since it's very detailed. I just followed the announcement post on Xamarin Forums. 

I will test again and return back to you.

ali-h2010:
I added the required conditions as the following:
`<uses-sdk android:minSdkVersion="17" />
      <application   android:label="SabicAR" 
			     android:supportsRtl="true">
     </application>`
The app is still showing the same behavior in the list view.(last attachement)
On an entry view, the direction is working as intended even if the text is in English.
For now,only the list view is not forced to the right if the text is not RTL which is fine in most cases. 

Many thanks to all who contributed on this great feature and the awesome team of developers.
<uses-sdk android:minSdkVersion="17" />

	<application 	android:label="SabicAR" 
					android:supportsRtl="true">
						
	</application>

encrypt0r:
img

@samhouts I think what I and @ali-h2010 are trying to say is that (As shown by the screenshot above) FlowDirection has no effect on the default direction of Label (Although I referred to text direction as FlowDirection in my original comment).

And I that is the reason the listview items are aligned the way they are. The Label direction is decided by the content of the label, which is very useful in many cases. But there are cases in which you want to control the direction of the text of a label.

Here is a repro:

```xaml
             xmlns:local="clr-namespace:App24"
             x:Class="App24.MainPage"
             Padding="10"
             FlowDirection="RightToLeft">

    <StackLayout>
        <Label Text="English Text" FontSize="Large" />
        <Label Text="نووسینی کوردی" FontSize="Large" />
    </StackLayout>
    
</ContentPage>
```
This is the full project: [TextDirectionRepro.zip](https://github.com/xamarin/Xamarin.Forms/files/1901476/TextDirectionRepro.zip)

ali-h2010:
@encrypt0r In iOS , it's working as expected.

samhouts:
This feedback is very much appreciated! Would you mind opening new issues for suspected bugs or features requests? I don't want them to get lost in this thread. Thanks!

encrypt0r:
@samhouts #2447 #2448 #2449. When do you expect you guys will be able to work on the current limitations (listed in the PR)?

ali-h2010:
@encrypt0r Your post is much more descriptive than mine. Could you please create a new issue with the same content?

smartprogrammer93:
@samhouts 
FlowDirection = rtl on horizental scrollview breaks the scrolling in it

samhouts:
That's not good! Would you mind opening an issue? Thanks!

ali-h2010:
Was someone able to bind the FlowDirection using ViewModels?
I tried but there is no response in the XAML

ali-h2010:
@samhouts 
I am also was not able to use FormattedString in a label. This ignores the FlowDirection 

`<Label TextColor="Black" FlowDirection="MatchParent">
                    <Label.FormattedText>
                        <FormattedString>
                            <Span Text="{Translate:TranslateExtension Text=OldPassword}" />
                            <Span Text="*" ForegroundColor="#CD4A4A" FontAttributes="Bold" />
                        </FormattedString>
                    </Label.FormattedText>
                </Label>`

samhouts:
@ali-h2010 Very sorry to have missed your comment! Please open new issues for any problems you encounter so we don't lose them in this thread. Thank you!

PureWeen:
@ali-h2010  check my notes here as well to see if that resolves the issue for you
https://github.com/xamarin/Xamarin.Forms/issues/3311

huangjinshe:
In the newest version I download today.  I confirm the problem is very different.

The Label flow direction totally based on the content.  
Which mean If you set content to : "English", it will always show as: Left to right.
If you set content to some right to left version font such like arabic: "سىناق". It will always show as Right to left.

Which mean **it will totally based on the font, and Android or Xamarin.Foms auto decide that,**
I really don't like it, it will change the default layout which we already design.(I want it always Left to right, when I wosn't set the flow direction of Label)

PR #1224, opened 2017-10-24
[Xaml] still load Xaml from old assemblies...

StephaneDelcroix:
### Description of Change ###

if the project references an assembly that was generated using a Xamlg version not creating XamlResourceId, we need to fallback to the old logic to load xaml controls from those assemblies.

### Bugs Fixed ###

- not reported yet

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###


Xamarin.Forms.Xaml/XamlLoader.cs
```diff
@@ -154,7 +154,7 @@ static string GetXamlForType(Type type)
 			var assembly = typeInfo.Assembly;
 			var resourceId = typeInfo.GetCustomAttribute<XamlResourceIdAttribute>()?.ResourceId;
 			if (resourceId == null)
-				return null;
+				return LegacyGetXamlForType(type);
```
StephaneDelcroix: this were I should add a `Task.Delay(1000);` to force new adoption

PR #1226, opened 2017-10-24
[Core, iOS] Allocate prototypical view for RetainElement

kingces95:
### Description of Change ###

Expect to be able to re-init ListView content. Actually, when using RetainElement, we experience an expectation due to use of a disposed view. The view in question was disposed by logic that used it to compute an estimated row height. That logic did expect the view to be recycled as part of pooling (resetting the Element). The fix is to allocate a view that does not participate in pooling.   

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=59813

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ classic_obj
 *.log
 *.userprefs
 *.suo
+_docs.xml
 UpgradeLog*
 *UpgradeReport*
 *.user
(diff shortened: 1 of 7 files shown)


bugadani:
Umm, what is that 84k deleted lines?

kingces95:
The 84k lines removed delete a build artifact that got mistakenly checked in.

mohamed-mb:
can you please elaborate on this "The fix is to allocate a view that does not participate in pooling". A sample code would be nice

kingces95:
The fix has no user visible behavior change other than preventing the exception thrown in the bug report. So, the bug report has a sample if you'd like to see the exception.

rmarinho:
So you don't identify any issue with 40704 test? Do you think you can try write it on other way?

PR #1232, opened 2017-10-26
[Xaml] create the [XamlResourceId] at assembly level

StephaneDelcroix:
### Description of Change ###

[Xaml] create the [XamlResourceId] at assembly level

this allows, as before, finding the resource for a type, but also find the type of a resource or from a pack-like uri

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Xaml/XamlResourceIdAttribute.cs
+++ b/Xamarin.Forms.Core/Xaml/XamlResourceIdAttribute.cs
@@ -2,16 +2,18 @@
 
 namespace Xamarin.Forms.Xaml
 {
-	[AttributeUsage(AttributeTargets.Class, Inherited = false, AllowMultiple = false)]
+	[AttributeUsage(AttributeTargets.Assembly, Inherited = false, AllowMultiple = true)]
 	public sealed class XamlResourceIdAttribute : Attribute
 	{
-		public string RootNamespace { get; set; }
 		public string ResourceId { get; set; }
+		public string Path { get; set; }
+		public Type Type { get; set; }
 
-		public XamlResourceIdAttribute(string rootNamespace, string resourceId)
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Core/Xaml/XamlResourceIdAttribute.cs
```diff
@@ -2,16 +2,18 @@
...
+	[AttributeUsage(AttributeTargets.Assembly, Inherited = false, AllowMultiple = true)]
 	public sealed class XamlResourceIdAttribute : Attribute
 	{
-		public string RootNamespace { get; set; }
 		public string ResourceId { get; set; }
+		public string Path { get; set; }
+		public Type Type { get; set; }
 
-		public XamlResourceIdAttribute(string rootNamespace, string resourceId)
+		public XamlResourceIdAttribute(string resourceId, string path, Type type)
```
StephaneDelcroix: the rootNamespace was required to compute a relative path based on a resourceId. as we now keep the path in addition to the resourceId, it's no longer required for the planned future use

Xamarin.Forms.Xaml/XamlLoader.cs
```diff
@@ -152,7 +152,15 @@ static string GetXamlForType(Type type)
 
 			var typeInfo = type.GetTypeInfo();
 			var assembly = typeInfo.Assembly;
-			var resourceId = typeInfo.GetCustomAttribute<XamlResourceIdAttribute>()?.ResourceId;
+
+			string resourceId = null;
+			foreach (var xria in assembly.GetCustomAttributes<XamlResourceIdAttribute>()) {
```
StephaneDelcroix: @jassmith please review this perf-wise, and tell me if you think it's useful to cache that info or not (I'm lending in the *not* direction)

PR #1233, opened 2017-10-26
[Android] Rename CreateRenderer overload for backward compatibility with previewer

hartez:
### Description of Change ###

Having a second public overload of CreateRenderer causes problems for the previewer when it reflects on the method name; to maintain backward compatibility we need to use a different name for the new overload of the method.

### Bugs Fixed ###

- Allows vNext to work with the existing previewer.

### API Changes ###

`public static IVisualElementRenderer CreateRenderer(VisualElement element, Context context)` => `public static IVisualElementRenderer CreateRenderer2(VisualElement element, Context context)`

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/Platform.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/Platform.cs
@@ -264,7 +264,7 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 			if (Android.Platform.GetRenderer(page) != null)
 				return;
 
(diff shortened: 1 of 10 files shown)


adrianknight89:
@hartez It seems somewhat unusual for a method name to be numbered like that. Maybe `CreateRendererWithContext` is a better option?

hartez:
@adrianknight89 Yeah, I think you're right. At least `CreateRendererWithContext` is somewhat descriptive.

hartez:
Going with #1237 for now instead.

PR #1235, opened 2017-10-27
[UWP] Bug 57674 - XF.UWP ListView not honoring INotifyCollectionChanged 

gatm50:
### Description of Change ###

The ListView Renderer for UWP was not managing the OnCollectionChanged event.
Adding a new method and subscribing it to the respective handler to manage the OnCollectionChanged event

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=57674

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #1240, opened 2017-10-30
[UWP] Force SetupContent call on ListView items

pauldipietro:
### Description of Change ###

`ListView`s in UWP are presently not calling `SetupContent` on the initial items; the simplest fix without making any major changes appears to be iterating through the `TemplatedItems` to force a call to `GetOrCreateContent` which then calls `SetupContent`.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=60033

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla60033.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla60033.cs
@@ -1,0 +1,78 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System.Collections.ObjectModel;
+using System.ComponentModel;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Bugzilla, 60033, "[UWP] ListView override void SetupContent not triggered", PlatformAffected.UWP)]
+	public class Bugzilla60033 : TestContentPage
+	{
+		protected override void Init()
+		{
+			var items = new ObservableCollection<string>() { "A", "B", "C" };
+			var lv = new Bugzilla60033ListView()
+			{
+				ItemsSource = items
+			};
+			var setupContentCallCount = new Label();
+			
+			Content = new StackLayout
+			{
+				Children =
+				{
+					new Label
+					{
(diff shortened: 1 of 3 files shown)


samhouts:
@pauldipietro Would you mind looking at #1235 to see if it might actually resolve this issue, too?

pauldipietro:
I'm pretty sure I did a few days back out of curiosity, and it didn't.

rmarinho:
@pauldipietro needs rebase, and a review from @samhouts and @jassmith , i m not sure tbh this is the proper fix.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla60033.cs
```diff
@@ -0,0 +1,78 @@
...
+			}
+
+			protected override void SetupContent(Cell cell, int index)
+			{
+				base.SetupContent(cell, index);
+				SetupContentCallCount++;
+			}
+		}
+	}
+}
```
hartez: Can we get a UI test here? We can run UWP UI tests now (manually), and it'll help us guard against breaking this behavior on other platforms.

Xamarin.Forms.Platform.WinRT/ListViewRenderer.cs
```diff
@@ -85,6 +85,12 @@ protected override void OnElementChanged(ElementChangedEventArgs<ListView> e)
 				// WinRT throws an exception if you set ItemsSource directly to a CVS, so bind it.
 				List.DataContext = new CollectionViewSource { Source = Element.ItemsSource, IsSourceGrouped = Element.IsGroupingEnabled };
 
+				// This is a bit of a hack around the fact that SetupContent is not being called on the initial items
+				// in the ListView; iterating on each of the TemplatedItems gets around this as it then forces
+				// GetOrCreateContent to be called.
+				TemplatedItemsView.TemplatedItems.CollectionChanged += OnTemplatedItemsCollectionChanged;
```
hartez: Is it possible to move this line before the setting of the DataContext above? I'm wondering if we can avoid having to duplicate the handler code on the next line.

rmarinho:
@pauldipietro can you address @ez comment and rebase ? thanks.

pauldipietro:
I think I'm going to close this for the time being since it does feel like a bit of a dirty hack as mentioned.

michaelstonis:
Is there any chance of coming up with a good solution here for UWP? This is definitely still an issue and there is no good way to work around it at the moment.

mjrichards91:
This issue still exists today. Are there any known workarounds to force `SetupContent` to be called for UWP?

PR #1229, opened 2017-10-25
[Xaml[C]] ResourceDictionary.Source

StephaneDelcroix:
### Description of Change ###

Allow setting a Source to a ResourceDictionary. Source is an Uri, relative or absolute, that points to a resource in the same assembly.

Restriction: this can only be used from within Xaml

MergedWith is deprecated (obsolete).

Oh, and you no longer have to create code behind for resource dictionaries ! But if you do, we will use it

TODO:
- compile the RD with XamlC using xml processing instructions (see #1241)

### Bugs Fixed ###

/

### API Changes ###

Added:
 - Uri ResourceDictionary.Source {get; set;}

### Behavioral Changes ###

/

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #1241, opened 2017-10-31
[XamlC] Compile RDs without codebehind

StephaneDelcroix:
### Description of Change ###

PR #1229 allowed the creation of RD without code behind (no `x:Class` and no `.xaml.cs` files). This PR piles on that one (only look at the latest commit while reviewing) and add a `xml` processing instruction to generate a phony codebehind with the right attributes so the XamlC will pick it up.

here's a very simple RD as example:

```
<?xml version="1.0" encoding="UTF-8" ?>
<?xaml-comp compile="true" ?>
<ResourceDictionary xmlns="http://xamarin.com/schemas/2014/forms"
		xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml">
	<Style x:Key="foo" TargetType="Label">
		<Setter Property="TextColor" Value="Orange" />
		<Setter Property="Text" Value="Pumpkin" />
	</Style>
</ResourceDictionary>
```

look at line 2 `<?xaml-comp compile="true" ?>`. This is what instructs XamlG to generate this file
```
//------------------------------------------------------------------------------
// <auto-generated>
//     This code was generated by a tool.
//     Runtime Version:4.0.30319.42000
//
//     Changes to this file may cause incorrect behavior and will be lost if
//     the code is regenerated.
// </auto-generated>
//------------------------------------------------------------------------------

[assembly: global::Xamarin.Forms.Xaml.XamlResourceIdAttribute("Xamarin.Forms.Xaml.UnitTests.RDWithoutCodeBehind.xaml", "RDWithoutCodeBehind.xaml", typeof(global::__XamlGeneratedCode__.@__Type0))]

namespace @__XamlGeneratedCode__ {
    
    
    [global::Xamarin.Forms.Xaml.XamlFilePathAttribute("RDWithoutCodeBehind.xaml")]
    [global::Xamarin.Forms.Xaml.XamlCompilationAttribute(global::Xamarin.Forms.Xaml.XamlCompilationOptions.Compile)]
    [global::System.ComponentModel.EditorBrowsableAttribute(global::System.ComponentModel.EditorBrowsableState.Never)]
    public partial class @__Type0 : global::Xamarin.Forms.ResourceDictionary {
        
        [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Xamarin.Forms.Build.Tasks.XamlG", "0.0.0.0")]
        public @__Type0() {
            this.InitializeComponent();
        }
        
        [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Xamarin.Forms.Build.Tasks.XamlG", "0.0.0.0")]
        private void InitializeComponent() {
            global::Xamarin.Forms.Xaml.Extensions.LoadFromXaml(this, typeof(@__Type0));
        }
    }
}
```
### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/DebugXamlCTask.cs
+++ b/Xamarin.Forms.Build.Tasks/DebugXamlCTask.cs
@@ -56,7 +56,7 @@ namespace Xamarin.Forms.Build.Tasks
 					foreach (var resource in module.Resources.OfType<EmbeddedResource>()) {
 						Logger.LogString(2, "  Resource: {0}... ", resource.Name);
 						string classname;
-						if (!resource.IsXaml(out classname)) {
+						if (!resource.IsXaml(module, out classname)) {
 							Logger.LogLine(2, "skipped.");
 							continue;
 						} else
--- a/Xamarin.Forms.Build.Tasks/Xamarin.Forms.Build.Tasks.csproj
+++ b/Xamarin.Forms.Build.Tasks/Xamarin.Forms.Build.Tasks.csproj
@@ -102,6 +102,7 @@
     <Compile Include="GetTasksAbi.cs" />
     <Compile Include="CompiledConverters\UriTypeConverter.cs" />
     <Compile Include="CompiledConverters\RDSourceTypeConverter.cs" />
+    <Compile Include="XamlGenerator.cs" />
   </ItemGroup>
   <Import Project="$(MSBuildBinPath)\Microsoft.CSharp.targets" />
   <Target Name="AfterBuild">
--- a/Xamarin.Forms.Build.Tasks/XamlCTask.cs
+++ b/Xamarin.Forms.Build.Tasks/XamlCTask.cs
@@ -117,7 +117,7 @@ namespace Xamarin.Forms.Build.Tasks
 					foreach (var resource in module.Resources.OfType<EmbeddedResource>()) {
 						Logger.LogString(2, "  Resource: {0}... ", resource.Name);
 						string classname;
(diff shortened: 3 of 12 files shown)


samhouts:
🎃

wjk:
I have a quick comment on the implementation of this PR: Have you considered making the `__GeneratedTypes__.__Type*` classes internal? IMHO, having these classes public makes the resulting assemblies look extremely messy when used. The example given in @StephaneDelcroix's original post suggests that these classes will be public, but examining the details in 8ad3553 suggests otherwise (I can see no code anywhere in `XamlGenerator.GenerateCode()` that makes the generated type public). Of course, if this would break things, feel free to ignore this suggestion. Thanks!

StephaneDelcroix:
@wjk Of course we could make them internal. But then we would not be able to instantiate them at all...

wjk:
@StephaneDelcroix In that case, would it be possible to give the generated classes/namespaces a more meaningful name? I would suggest renaming `__GeneratedTypes__` to `__XamlGeneratedCodeBehind__` or something similar. I just don't like seeing seemingly-randomly named types littering my IntelliSense. You might also consider adding `[EditorBrowsableAttribute]` to the generated types, like you do to hide public-but-really-internal types and members in Xamarin.Forms proper.

jassmith:
I agree with wjk's assessment of both the names and the [EditorBrowsableAttribute]

StephaneDelcroix:
Can't object to that. I wasn't happy with the names to start with. Will change that

StephaneDelcroix:
Rebase to ease the review

PR #1244, opened 2017-10-31
[iOS] Don't crash when grouped ListView uses ObservableCollection

samhouts:
### Description of Change ###

#948 added an informational exception to help users diagnose indexing issues when adding items. However, the logic throws unnecessarily when the ListView is grouped. This change will simply skip the index check if the ListView is grouped.

Also added a UI test for 56771 to ensure that this doesn't break the original intention of the informational exception.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=59896

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


rmarinho:
build

PR #1247, opened 2017-11-01
Make it possible to customise the window on iOS

martijn00:
### Description of Change ###

This changes the way the window is used in iOS in a backwards compatible way.

### Bugs Fixed ###

It was not possible to swap out the current root for a native Xamarin ViewController, and then navigate back to a Forms one.

Also when changing the MainPage on a custom window it would crash because it would not implement PlatformRenderer. The new check prevents this.

### API Changes ###

Added:
 - Window { get; set; }

### Behavioral Changes ###

This will not change any behaviour but just enable to work easier with Native views, especially in MvvmCross.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/FormsApplicationDelegate.cs
+++ b/Xamarin.Forms.Platform.iOS/FormsApplicationDelegate.cs
@@ -9,20 +9,20 @@ namespace Xamarin.Forms.Platform.iOS
 {
 	public class FormsApplicationDelegate : UIApplicationDelegate
 	{
-        Application _application;
+		Application _application;
 		bool _isSuspended;
 		UIWindow _window;
-        public override UIWindow Window
-        {
-            get
-            {
-                return _window;
-            }
-            set
-            {
-                _window = value;
-            }
-        }
+		public override UIWindow Window
+		{
+			get
+			{
+				return _window;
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.iOS/FormsApplicationDelegate.cs
```diff
@@ -156,20 +168,18 @@ void CheckForAppLink(NSUserActivity userActivity)
...
 
 		void UpdateMainPage()
 		{
 			if (_application.MainPage == null)
 				return;
 
-			var platformRenderer = (PlatformRenderer)_window.RootViewController;
-
-			if (platformRenderer != null)
+			if(Window.RootViewController is PlatformRenderer platformRenderer)
```
samhouts: This is C#7, yeah? Love it, but we're stuck in VS2015 land for now. :( 

Also, if you make other changes to the file, please also add a space between the if and parens. Not a big deal.

martijn00: Hmmm ok, i'll make some changes. There is nothing about this in the Wiki on how to contribute.

martijn00:
@samhouts Implemented requested changes.

PR #1254, opened 2017-11-07
[iOS] Fix NRE when Grouped ListView w/DataTemplateSelector has uneven rows

samhouts:
### Description of Change ###

#1226 improved the estimation for iOS row heights when `HasUnevenRows` is true, but it did not account for grouped `ListView`s using `DataTemplateSelector`s. This change will get the first item in the first group to inflate the `DataTemplate` instead of the first group.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=60524

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
@@ -358,7 +358,7 @@ namespace Xamarin.Forms.Platform.iOS
 				{
 					Control.Layer.RemoveAllAnimations();
 					//iOS11 hack
-					if(Forms.IsiOS11OrNewer)
+					if (Forms.IsiOS11OrNewer)
 					{
 						await Task.Delay(1);
 					}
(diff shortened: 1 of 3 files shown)


kingces95:
Hehe, I also fixed this bug https://github.com/xamarin/Xamarin.Forms/pull/1256! But you have a unit test so let's take yours.

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -663,10 +663,16 @@ internal nfloat GetEstimatedRowHeight(UITableView table)
...
-				var firstCell = templatedItems.ActivateContent(0, templatedItems.ListProxy[0]);
+				Cell firstCell;
+				var isGroupingEnabled = List.IsGroupingEnabled;
+
+				if (isGroupingEnabled)
+					firstCell = templatedItems.ActivateContent(0, templatedItems.GetGroup(0)?.ListProxy[0]);
+				else
+					firstCell = templatedItems.ActivateContent(0, templatedItems.ListProxy[0]);
 
 				// Let's skip this optimization for grouped lists. It will likely cause more trouble than it's worth.
```
kingces95: Not strictly related to this bug, but when I read this comment I expected "skipping" would imply taking the pass less taken; e.g. taking the `if` when grouping was enabled. But we actually take the if if grouping is disabled.

PR #1259, opened 2017-11-09
[XamlG] use random file names while used by UpdateDesignTimeXaml

StephaneDelcroix:
### Description of Change ###

For some reason, `UpdateDesignTimeXaml` target doesn't set all the metadata we rely on for generating the file path. Use a hidden random file name instead. Those files won't anyway end up in the final build artifact.

### Bugs Fixed ###

- the one reported by @hartez 

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/XamlGTask.cs
+++ b/Xamarin.Forms.Build.Tasks/XamlGTask.cs
@@ -32,7 +32,12 @@ namespace Xamarin.Forms.Build.Tasks
 			}
 
 			foreach (var xamlFile in XamlFiles) {
-				var outputFile = Path.Combine(OutputPath, $"{xamlFile.GetMetadata("TargetPath")}.g.cs");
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Build.Tasks/XamlGTask.cs
```diff
@@ -32,7 +32,12 @@ public override bool Execute()
 			}
 
 			foreach (var xamlFile in XamlFiles) {
-				var outputFile = Path.Combine(OutputPath, $"{xamlFile.GetMetadata("TargetPath")}.g.cs");
+				//when invoked from `UpdateDesigntimeXaml` target, the `TargetPath` isn't set, use a random one instead
+				var targetPath = xamlFile.GetMetadata("TargetPath");
+				if (string.IsNullOrWhiteSpace(targetPath))
+					targetPath = $".{Path.GetRandomFileName()}";
```
hartez: Is the "." in front of the path intentional?

PR #1264, opened 2017-11-10
[iOS] Check for iOS8 before changing control flow direction

samhouts:
### Description of Change ###

`UISemanticContentAttribute` was added in iOS9, so we need to avoid calling it in lower versions.

Continuation of #1222.

### Bugs Fixed ###

- Failing UI tests on master

### API Changes ###

None

### Behavioral Changes ###

Native controls on iOS <8 may not fully respect the `FlowDirection` value.

### PR Checklist ###

- [x] Has tests (iOS 8 won't crash anymore)


rmarinho:
@samhouts can you use Forms.IsIos9OrNewer to be consistent

samhouts:
if i HAVE to ;)

PR #467, opened 2016-10-18
Fix out-of-memory crashes on iOS when creating maps

hartez:
### Description of Change

On iOS 10, constructing a new map view leaks memory; after constructing a sufficient number of them, the app will be ejected for memory use. This change pools instances of MKMapView on iOS 10, reusing them to reduce the number of instances created and minimize the leaked memory. 

This change also cleans up the `Dispose` implementations for MapRenderer, ViewRenderer, and MapDelegate to make them more consistent and to avoid some other minor leaks.
- Pool map views on iOS 10 to avoid memory issues
- Clean up disposal implementation for iOS map renderer, view renderer, and map delegate
### Bugs Fixed
- [39489, but for iOS 10](https://bugzilla.xamarin.com/show_bug.cgi?id=39489)
### API Changes

None
### Behavioral Changes

None
### PR Checklist
- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #1265, opened 2017-11-11
[iOS] Disable iOS Map pooling; Add null check

kingces95:
### Description of Change ###

Expected to be able to reset a iOS map view pulled from the recycled pool but actually that threw an exception because recycled map views have null annotation arrays. Added a null check. Also had to disable the pool as the recycled map views failed to render. 

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=60140

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Maps.iOS/MapRenderer.cs
+++ b/Xamarin.Forms.Maps.iOS/MapRenderer.cs
@@ -372,7 +372,8 @@ namespace Xamarin.Forms.Maps.MacOS
 					break;
 				case NotifyCollectionChangedAction.Reset:
 					var mapView = (MKMapView)Control;
-					mapView.RemoveAnnotations(mapView.Annotations);
+					if (mapView.Annotations?.Length > 0)
+						mapView.RemoveAnnotations(mapView.Annotations);
 					AddPins((IList)(Element as Map).Pins);
 					break;
 				case NotifyCollectionChangedAction.Move:


Xamarin.Forms.Maps.iOS/MapRenderer.cs
```diff
@@ -246,7 +246,7 @@ protected override void OnElementChanged(ElementChangedEventArgs<View> e)
 					if (FormsMaps.IsiOs9OrNewer)
 					{
 						// See if we've got an MKMapView available in the pool; if so, use it
-						mapView = MapPool.Get();
+						//mapView = MapPool.Get();
```
hartez: If you're just not going to do anything, remove the `if` statement.

kingces95: @hartez You added the cache a while back. I'm wondering if it ever worked. Was there a UITest for the cache? Do you think it's ok to pull the entire cache/pool?

hartez: @kingces95 According to git, it was added in https://github.com/xamarin/Xamarin.Forms/pull/467. So there is a test (39489). If that test were to pass* with the pooling code commented out, then the underlying leak has probably been fixed and we could remove it.

Alternatively, if you can figure out where the maps leak is coming from and fix it, that'd be awesome.

* It doesn't.

hartez: > Also had to disable the pool as the recycled map views failed to render.

When you say "failed to render", what do you mean? Big blank spot where the map should be? Blank map? Crash?

kingces95: Blank

kingces95:
34561 passed on iOS device locally.

kingces95:
Should have a UITests but gonna push because finally all tests are green.

PR #1238, opened 2017-10-27
[iOS] Fixes for iOS11 

rmarinho:
### Description of Change ###

Fixes and new features for iOS11 support.

Platform specifics for `SafeArea`, `SafeAreaInsets`  and `LargeTitles`.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=60054
- https://bugzilla.xamarin.com/show_bug.cgi?id=59927

### API Changes ###

Added:
Page iOS Platform Specific
 ```
public enum LargeTitleDisplayMode
 {
 	Automatic,
 	Always,
 	Never
 }


public static readonly BindableProperty LargeTitleDisplayProperty = BindableProperty.Create(nameof(LargeTitleDisplay), typeof(LargeTitleDisplayMode), typeof(Page), LargeTitleDisplayMode.Automatic);

public static LargeTitleDisplayMode GetLargeTitleDisplay(BindableObject element)
public static void SetLargeTitleDisplay(BindableObject element, LargeTitleDisplayMode value)
public static IPlatformElementConfiguration<iOS, FormsElement> SetLargeTitleDisplay(this IPlatformElementConfiguration<iOS, FormsElement> config, LargeTitleDisplayMode value)
public static LargeTitleDisplayMode LargeTitleDisplay(this IPlatformElementConfiguration<iOS, FormsElement> config)
public static bool UsingSafeArea(this IPlatformElementConfiguration<iOS, FormsElement> config)



static readonly BindablePropertyKey SafeAreaInsetsPropertyKey = BindableProperty.CreateReadOnly(nameof(SafeAreaInsets), typeof(Thickness), typeof(Page), default(Thickness))

public static readonly BindableProperty SafeAreaInsetsProperty = SafeAreaInsetsPropertyKey.BindableProperty;

public static Thickness GetSafeAreaInsets(BindableObject element)
public static Thickness SafeAreaInsets(this IPlatformElementConfiguration<iOS, FormsElement> config)
public static IPlatformElementConfiguration<iOS, FormsElement> SetSafeAreaInsets(this IPlatformElementConfiguration<iOS, FormsElement> config, Thickness value)

```
NavigationPage iOS Platform Specific
```
public static readonly BindableProperty UseLargeTitlesProperty = BindableProperty.Create("UseLargeTitles", typeof(bool), typeof(Page), false);

public static bool GetPrefersLargeTitles(BindableObject element)
public static void SetPrefersLargeTitles(BindableObject element, bool value)
[...]

PR #1270, opened 2017-11-14
[iOS] Revert iOS11 safe area fix on ViewCell

rmarinho:
### Description of Change ###

Revert change  from #1238 related with ViewCell padding on iPhoneX. 

Users can use safe are from the `Page` to set padding on cells. 

**Before**
simulator screen shot - iphone x - 2017-11-14 at 00 50 03

simulator screen shot - iphone x - 2017-11-14 at 00 51 34

**After**
simulator screen shot - iphone x - 2017-11-14 at 00 53 01

simulator screen shot - iphone x - 2017-11-14 at 00 51 15

### Bugs Fixed ###

- https://forums.xamarin.com/discussion/comment/306004/#Comment_306004

### API Changes ###

None

### Behavioral Changes ###

Cells will not auto adjust to safe area, user is now responsible of handling the padding for now. 

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Cells/ViewCellRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Cells/ViewCellRenderer.cs
@@ -30,16 +30,6 @@ namespace Xamarin.Forms.Platform.iOS
 			return cell;
 		}
 
-		public override void SetBackgroundColor(UITableViewCell tableViewCell, Cell cell, UIColor color)
-		{
-			if (cell is ViewCell && Forms.IsiOS11OrNewer)
-			{
-				color = (cell as ViewCell).View.BackgroundColor == Color.Default ? color : (cell as ViewCell).View.BackgroundColor.ToUIColor();
-				tableViewCell.BackgroundColor = color;
-			}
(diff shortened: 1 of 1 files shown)


MdNaushad:
In which xamarin version will this be available?

nhdanh:
I still get a padding left in viewcell :(. How to remove it

rmarinho:
@nhdanh can you post a screenshot and some sample code?

nhdanh:
This is my screenshot. The viewcell in tableview has shift left. The textcell is normal

nhdanh:
This is my screenshot. The viewcell has shift left. The textcell still normal

nhdanh:
I update XF to newest version and this problem has been solved. Thank you

Sebastian1989101:
@rmarinho is there any example how we can implement this as clean as possible? Because it won't work if the ListView is nested in a Grid and I don't get it quite as nice as you solve this. Also it still seems to use SafeArea even with the latest XF Version 2.5.0.91635 when the TableView or ListView is used and the root of the page (and for my settings page it totally destroyed my page because the offset with the margin seems to set everything a bit off).

PR #1275, opened 2017-11-15
[Core, iOS] Fix 59896; Add empty listView group twice

kingces95:
### Description of Change ###

Expect adding two empty groups to a ListView won't crash app. Actually crashes. Test TryGetValue for null instead of indexing with bad index.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=59896

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/ListProxy.cs
+++ b/Xamarin.Forms.Core/ListProxy.cs
@@ -436,6 +436,9 @@ namespace Xamarin.Forms
 
 		#region IList
 
+		bool IListProxy.TryGetValue(int index, out object value)
+			=> TryGetValue(index, out value);
+
 		object IList.this[int index]
 		{
 			get { return this[index]; }
--- a/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
@@ -666,13 +666,16 @@ namespace Xamarin.Forms.Platform.iOS
 				}
 
 				// We're going to base our estimate off of the first cell
-				Cell firstCell;
 				var isGroupingEnabled = List.IsGroupingEnabled;
 
 				if (isGroupingEnabled)
-					firstCell = templatedItems.ActivateContent(0, templatedItems.GetGroup(0)?.ListProxy[0]);
-				else
-					firstCell = templatedItems.ActivateContent(0, templatedItems.ListProxy[0]);
+					templatedItems = templatedItems.GetGroup(0);
+
+				object item = null;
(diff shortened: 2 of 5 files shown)


Xamarin.Forms.Core/ListProxy.cs
```diff
@@ -436,6 +436,9 @@ public void Reset()
 
 		#region IList
 
+		bool IListProxy.TryGetValue(int index, out object value)
```
StephaneDelcroix: if the explicit implementation forwards to the internal implementation, why not doing a single explicit implementation ?

you will still be able to invoke it, as the interface is public

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -666,13 +666,16 @@ internal nfloat GetEstimatedRowHeight(UITableView table)
...
-				Cell firstCell;
 				var isGroupingEnabled = List.IsGroupingEnabled;
 
 				if (isGroupingEnabled)
-					firstCell = templatedItems.ActivateContent(0, templatedItems.GetGroup(0)?.ListProxy[0]);
-				else
-					firstCell = templatedItems.ActivateContent(0, templatedItems.ListProxy[0]);
+					templatedItems = templatedItems.GetGroup(0);
+
+				object _firstCell = null;
```
StephaneDelcroix: as fucked up as it is, our Style recommend using `_xxx` for instance fields, not local variables.

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -666,13 +666,16 @@ internal nfloat GetEstimatedRowHeight(UITableView table)
...
+				if (templatedItems?.ListProxy.TryGetValue(0, out _firstCell) != true)
```
StephaneDelcroix: I was stuck on this one for a moment, wondering if it was adding a cast to `bool?` or not. turns out that Roslyn does not (not tested on other compilers).

But what is really confusing, is that the behavior of this changes if you change `!= true` to `== false`.

As I'm not smart enough to figure that if I ever have to debug that part due to a rogue refactoring, I recommend changing this to
```
if (templatedItems == null || !templatedItems.ListProxy.TryGetValue(0, out _firstCell))
```

StephaneDelcroix:
Thanks for the changes

kingces95:
Bugzilla 31333 failed once locally and passed once locally. Looks flakey. 
https://devdiv.visualstudio.com/DevDiv/_TestManagement/Runs?_a=resultSummary&runId=787034&resultId=100030

PR #1287, opened 2017-11-18
[Android] AppCompat NavigationBar should handle Transluscent Flags

bentmar:
### Description of Change ###

When using either TransluscentStatus or TransluscentNavigation flags on Android the content will be layed out under the statusbar (not below as it would without them). This PR adds padding and height to NavigationBar if any of those flags are set. The padding is calculated based on the statusbar height.
Now we can finally use the flags mentioned on android to create stunning looking UI.

To use any of the flags just set this in MainActivity OnCreate()

```
	Window.AddFlags(WindowManagerFlags.TranslucentNavigation);
	Window.AddFlags(WindowManagerFlags.TranslucentStatus);
```
Before we added a statusbarunderlay = we couldnt layout stuff under the statusbar even with those flags, but that got removed in 2.3.4 (i think). Now the only thing stopping us is the adjusting padding and height, thats what this PR solves for NavigationBar.

Here are som images:

**BEFORE with transluscentNavigation**
before_trans_nav

**AFTER with transluscentNavigation**
after_trans_nav

**BEFORE with both transluscent statusbar and navigation**
before_trans_both

**After with both transluscent statusbar and navigation**
after_trans_both

**Drawer looks nice on MDP with trascuscentstatusbar!**
after_drawer_trans_status

**Drawer looks nice on MDP with trascuscent flags! (this is KitKat)**
kitkat_drawer_open

**Kitkat with no transluscentflags**
kitkat_noflags

**BEFORE Kitkat with both flags transluscentflags**
kitkat_before_both

**AFTER Kitkat with both flags transluscentflags**
kitkat_after_both

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
@@ -627,6 +627,11 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 			_toolbar = null;
 
 			SetupToolbar();
+			
+			// if the old toolbar had padding from transluscentflags, set it to the new toolbar
+			if (oldToolbar.PaddingTop != 0)
+				_toolbar.SetPadding(0, oldToolbar.PaddingTop, 0, 0);
+			
 			RegisterToolbar();
 			UpdateToolbar();
 			UpdateMenu();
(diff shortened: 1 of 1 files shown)


bentmar:
While debugging the code i see that ResetToolbar is called twice on device orientation. One from propertychanged and one from OnLayout(). 

OnLayout will call resetToolbar() since ActionBarHeight()  returns a different height now that it is in a different orientation => triggering ResetToolbar().

As your comments say in code the call from PropertyChanged  should be removed.

dhaligas:
+1 to this .. would like to see this in there soon

rmarinho:
@bentmar can you rebase to run uitest please.

Thanks

bentmar:
I did this PR "manually" by pasting code in the browser. How does one rebase that?
I can do a new PR (again..) 

@rmarinho

207713:
any word on this when will it be finally pulled? we are in desperate need of this, because we can not refresh our nugets thus it is broken since 2.4 :/

HenKun:
Great and really required fix, thanks. In which release should this fix be included?

PR #1289, opened 2017-11-18
Update Platform Create Renderer

albilaga:
### Description of Change ###

Platform.CreateRenderer(VisualElement element) is marked as obsolete but the replacing method Platform.IVisualElementRenderer CreateRenderer(VisualElement element, Context context) marked as internal so the user can not use it.

### Bugs Fixed ###

- 

### API Changes ###
Changed:
 - internal static internal static IVisualElementRenderer CreateRenderer(VisualElement element, Context context) => public static IVisualElementRenderer CreateRenderer(VisualElement element, Context context)

### Behavioral Changes ###

I don't think there will be any behavioral changes. Just changing the access modifier so the obsolete method can be replaced with the new one.

### PR Checklist ###

- [x ] Rebased on top of master at time of PR
- [x ] Changes adhere to coding standard
- [ x] Consolidate commits as makes sense

### DIFF
--- a/Xamarin.Forms.Platform.Android/Platform.cs
+++ b/Xamarin.Forms.Platform.Android/Platform.cs
@@ -306,7 +306,7 @@ namespace Xamarin.Forms.Platform.Android
 			return CreateRenderer(element, Forms.Context);
 		}
 
-		internal static IVisualElementRenderer CreateRenderer(VisualElement element, Context context)
+		public static IVisualElementRenderer CreateRenderer(VisualElement element, Context context)
 		{
 			IVisualElementRenderer renderer = Registrar.Registered.GetHandler<IVisualElementRenderer>(element.GetType(), context) 
 				?? new DefaultRenderer(context);


adrianknight89:
#1237 changed this to internal to avoid previewer issues. @hartez Why not introduce a breaking change and get rid of the obsolete signature?

albilaga:
Ah I see. Sorry not tracking the prev commit. I think it can be closed now right? Maybe we just can get rid of the obsolete signature and in the release notes too

adrianknight89:
Well, time for a little rant. That would be ideal, but it's up to the team to decide if it can be removed. On a personal level, I'm in favor of breaking changes as they keep codebase clean and lean. :) There are many `Obsolete` warnings going back to v1.2. Perhaps, it's not wise to remove them all, but many of them are small and should have been breaking changes in the first place.

StephaneDelcroix:
see #1303

PR #1302, opened 2017-11-22
[Android] Make new CreateRenderer overload public - fixes #60815

tomgilder:
### Description of Change ###

Old public method has been marked as obsolete, but the replacement method is internal instead of public.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=60815

### API Changes ###

Changed accessibility of Android.Platform.CreateRenderer to public

### DIFF
--- a/Xamarin.Forms.Platform.Android/Platform.cs
+++ b/Xamarin.Forms.Platform.Android/Platform.cs
@@ -308,7 +308,7 @@ namespace Xamarin.Forms.Platform.Android
 			return CreateRenderer(element, Forms.Context);
 		}
 
-		internal static IVisualElementRenderer CreateRenderer(VisualElement element, Context context)
+		public static IVisualElementRenderer CreateRenderer(VisualElement element, Context context)
 		{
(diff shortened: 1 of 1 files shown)


softlion:
+1

hartez:
Unfortunately this is the way it has to be for the time being. For compatibility reasons, we can't make the method public until the next revision of the Forms Previewer.

See #1237 for some more context.

softlion:
This method is the only way for a custom renderer based on reusable views to provides its services, as it needs to associate to each view which is reused, a new renderer (for a new datatemplate item).

hartez:
The more I think about this, the more I think the right answer for now is to remove `Obsolete` from the older version of `CreateRenderer`. The new version of the method exists to support multi-activity scenarios with embedding, which most developers are probably not concerned with.

hartez:
@softlion @tomgilder As an interim solution (until we can make the new CreateRenderer overload public), I'm proposing #1303.

PR #1304, opened 2017-11-23
[Android] Allow Previewer to set a per-page Context

hartez:
### Description of Change ###

Restores the ability for the previewer to set its own Context for each page being rendered.


alanmcgovern:
Will a nuget be linked to this PR before we merge it so we can test that things work now?

rmarinho:
@alanmcgovern  can get them here https://devdiv.visualstudio.com/DevDiv/Xamarin%20Forms/_build/index?buildId=1168936&_a=summary&tab=artifacts

davidortinau:
Tested this PR versus 2.5.0 and confirmed Previewer is no longer throwing an error on SetContext.

PR #1305, opened 2017-11-23
Small refactor on UWP's MasterDetailControl.cs - identation fix

carlosfza:
Just a couple identation fixes on the MasterDetailControl.cs file.

### Behavioral Changes ###

None


StephaneDelcroix:
Thanks for your first contribution.

I'm sorry but we do not take Style change/Style fixes PRs, we instead fix them as we go through them. We prefer a readable history over properly styled sources.

PR #1267, opened 2017-11-13
[Win] Remove Windows 8.1 platforms **Breaking change**

rmarinho:
### Description of Change ###

Remove WinRT , WinRT Phone and WinRT Desktop

### Behavioral Changes ###

Xamarin Forms will stop supporting old Windows platforms and will move to VS2017 support. 

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [ ] Rebased on top of master at time of PR
- [ ] Changes adhere to coding standard
- [ ] Consolidate commits as makes sense

PR #1306, opened 2017-11-24
[All] Move to netstandard2.0 **Breaking change**

rmarinho:
### Description of Change ###

Move to netstandard2.0

Works on top of #1267 

### Behavioral Changes ###

Move to UWP Fall creator's update that supports netstandard2.0

### PR Checklist ###


rmarinho:
Updated @StephaneDelcroix

dotMorten:
@rmarinho Isn't it WAY too early to require Fall Creators Update? I have serious concerns about this, as you're forcing us to stay on an old version of Xamarin.Forms for a long time, which means all of our customers who uses our controls will be stuck on an older version, until FCU is the min supported windows 10 version (which is about 18 months out)

Poppyto:
I agree it's too early, even XBox One isn't ready for that : https://github.com/dotnet/standard/issues/564

rmarinho:
@dotMorten we have the 2.5.0 branch (15-5) with support for older versions, we hope to back port important fixes and release from there. But the future with is netstandard2.0 and that only works for UWP in Fall Creators Update.. there's no way to go around that .

dotMorten:
Ok so you're going to be supporting 2.x and 3.x simultaneously?  Historically you only supported the latest version.
Still though it does put me as a component vendor in an awkward spot

dotMorten:
Also was 2.0 really needed? 1.6 doesn't have enough?

Poppyto:
I'm also curious about 2.0 ? Is there any API need at the moment ?

dotMorten:
Are you planning to actually take advantage of 2.0? Like dropping the Xamarin.Forms color type and move to the .NET Standard 2.0?

dansiegel:
@dotMorten that would be a particularly nasty breaking change. The XF Color type is used everywhere in people's apps. Especially on larger apps that could eat up a lot of time just trying to update to use System.Drawing.Color. I know you may not want to hear this but UWP has such a small following right now to hold up progress for every other platform because some people don't want to install 1709 is crazy. Besides it seems like Windows Update is forcing people to lose time rolling back from 1709  every week anyway.

Poppyto:
The real argument for me is to know if Xamarin has to use new API from .net standard 2.0. 
If not, there's absolutely no interest to drop support *because* the future. Xamarin.Forms has to use the older .net standard until a really need. Otherwise, this is just a way to embrace the Microsoft framework marketing adoption and it's not what is Xamarin.Forms : the best cross-platform Framework.

FYI, with .net standard 2.0, it's the last good bye for Windows Mobile.

rmarinho:
For the time we are not looking at changing any of our API, i think we just changed a internal Enum for FileAccess we had our own version for the .net one from System.iO . 
Also remind there's now a bigger surface area of API's one could use, and imo that's a gain by itself.

CZEMacLeod:
@dotMorten @dansiegel I have to admit I have a heap of casts from/to XF Color already in my code as my data layer/model uses just .NET Standard 2.0 and has no XF references.
What would be really useful is to either inherit XF Color from SD Color and move to SD color so XF Colors can still be used, or (more likely I guess) add implicit casts to XF Color so they can be used 'interchangeably'.
I don't know if, as they are both structs, whether it would be possible to use any of the new Span<T> stuff to prevent having to copy back and forth? I'm not sure what the internal representation of Color is on the different platforms..
Just now I have the rather clunky code
```
private Xamarin.Forms.Color GetColor(System.Drawing.Color color) => Xamarin.Forms.Color.FromRgba(color.R, color.G, color.B, color.A);
private System.Drawing.Color GetColor(Xamarin.Forms.Color color) => System.Drawing.Color.FromArgb((int)color.A, (int)color.R, (int)color.G, (int)color.B);
```

dotMorten:
>  I have to admit I have a heap of casts from/to XF Color already in my code

Me too. Perhaps adding implicit conversion to Xamarin.Forms.Color could be helpful? Actually I'll make a PR for that real quick...

charlesroddie:
In future going the other way round would be better, making System.Drawing.Color primary, depreciating Xamarin.Forms.Color, and implicitly converting from Xamarin.Forms.Color.

dotMorten:
@charlesroddie While I do agree that would be better, changing the property type would be a big breaking change though.

charlesroddie:
Very mild breaking change IMO, Xaml won't break, C# should just require opening System.Drawing for most projects. I admit I'm biased though because F# doens't implicity convert!

dotMorten:
You're thinking about code compatibility. I'm more concerned about binary compatibility.
There's no such thing as mild breaking changes. All other libraries that reference Xamarin.Forms would have to be recompiled for the latest version, which would mean lots of people wouldn't be able to move up to the latest version, because they have to wait for all the 3rd party libs to move. That's a huge amount of 3rd party libs you'd be breaking.

charlesroddie:
That does make sense. Actually I discover that System.Drawing.Color is a legacy color system (8bpc sRGB), but a proper color type and library for .net core/standard is only in the early stages of discussion: (see https://github.com/dotnet/corefx/issues/1563 and https://github.com/dotnet/corefx/issues/2315). So I agree it's best to maintain the status quo (as your PR helps with) until .net works out a better color type. Seems odd to me that .net color is in such a basic state as Windows Color System has been working well for 11 years.

dotMorten:
@rmarinho What was the reason to move the .Design assemblies to 4.7? This generates build warnings for those of us targeting a lower version: https://github.com/xamarin/Xamarin.Forms/issues/2731

PR #1307, opened 2017-11-28
[iOS] ListView should not explicitly remove subviews coming from renderers

samhouts:
### Description of Change ###

#524 added logic on iOS that will remove all `Subview`s of a `ListView` from their parents to prevent memory leaks. On iOS 8, this caused a SIGSEGV for `ActivityIndicator`s. The actual fault appears to be an iOS issue whereby the `UIActivityIndicatorView` is being accessed by the OS after it has been set to `nil` by `RemoveFromSuperview`. 

This change will adjust the aggressive `DisposeSubviews` method to only remove views that are not `IVisualElementRenderer`s, with the assumption that the renderers themselves should handle the disposal and removal of their own views. This effectively prevents the SIGSEGV by not calling `RemoveFromSuperview` on `ActivityIndicatorRenderer` and puts the onus on the renderers to take care of that.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=60563

### API Changes ###

None

### Behavioral Changes ###

May reintroduce some memory leaks if renderers inside of `ListView`s do not properly dispose themselves.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla60563.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla60563.cs
@@ -1,0 +1,92 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System.Collections.Generic;
+
+#if UITEST
+using Xamarin.UITest;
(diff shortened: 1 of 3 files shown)


rmarinho:
need rebase @samhouts

samhouts:
@rmarinho This is against 15-5. It's up to date. Not sure what the build errors are about.

PR #1359, opened 2017-12-08
Add implicit color conversion to .NET Standard's System.Drawing.Color

dotMorten:
### Description of Change ###

This change simplifies working with colors in a .NET Standard way by adding implicit conversion between the two color types:
```
Xamarin.Forms.Color c1 = Xamarin.Forms.Color.FromRgba(1, 1, 1, 1);
System.Drawing.Color c2 = c1;
Xamarin.Forms.Color c3 = c2;
if (c2 == c1)
{
   // Colors are equal
}
```

This came out of this discussion: https://github.com/xamarin/Xamarin.Forms/pull/1306#issuecomment-350355271

### Bugs Fixed ###

- N/A

### API Changes ###

Added:
 - `public static implicit operator System.Drawing.Color(Xamarin.Forms.Color color)`
 - `public static implicit operator Xamarin.Forms.Color(System.Drawing.Color color)`

### Behavioral Changes ###

None.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Color.cs
+++ b/Xamarin.Forms.Core/Color.cs
@@ -394,7 +394,7 @@ namespace Xamarin.Forms
 
 		public static implicit operator System.Drawing.Color(Color color)
 		{
-			return System.Drawing.Color.FromArgb((byte)(color._a * 255), (byte)(color._g * 255), (byte)(color._b * 255), (byte)(color._a * 255));
+			return System.Drawing.Color.FromArgb((byte)(color._a * 255), (byte)(color._r * 255), (byte)(color._g * 255), (byte)(color._b * 255));
 		}
 
 		public static implicit operator Color(System.Drawing.Color color)
--- a/Xamarin.Forms.Core/Color.cs
+++ b/Xamarin.Forms.Core/Color.cs
@@ -392,6 +392,16 @@ namespace Xamarin.Forms
 			return new Color(h, s, l, a, Mode.Hsl);
 		}
 
+		public static implicit operator System.Drawing.Color(Color color)
+		{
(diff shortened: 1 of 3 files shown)


dotMorten:
...One thing to consider: Since there are potential rounding involved, perhaps this should be an explicit rather than implicit conversion? (although I like the simplicity of just being able to use it as shown above)

Xamarin.Forms.Core/Color.cs
```diff
@@ -392,6 +392,16 @@ public static Color FromHsla(double h, double s, double l, double a = 1d)
 			return new Color(h, s, l, a, Mode.Hsl);
 		}
 
+		public static implicit operator System.Drawing.Color(Color color)
+		{
+			return System.Drawing.Color.FromArgb((byte)(color._a * 255), (byte)(color._r * 255), (byte)(color._g * 255), (byte)(color._b * 255));
```
StephaneDelcroix: I'm quite sure this will fail, or get you unexpected results
```
   var sdcolor = (System.Drawing.Color)Xamarin.Forms.Color.Default;
```

StephaneDelcroix: Color.Default has its fields (r, g, b and a) initialized to `-1`. converting to a S.D.Color results of a S.D.Color(1,1,1,1);

That's definitely not what you want.

if you want to match X.F.Color.Default to something, it should converts to S.D.Color.Empty (and back)

dotMorten:
@StephaneDelcroix I've added support for conversion between Color.Empty and Color.Default. From what I can tell, it's not possible to create values outside 0..1 range (all constructors clamp the values), unless you create a Color.Default, so the fix is fairly straight-forward. Added two tests as well to ensure it works.

dotMorten:
@StephaneDelcroix Ping :)

PR #1388, opened 2017-12-12
[Android] Add API condition for initialize LayoutDirection

mohachouch:
### Description of Change ###
The android platform crash when launching in the API15 and API16 versions because the property LayoutDirection is available since API17. 

See Android docs : https://developer.android.com/reference/android/content/res/Configuration.html#getLayoutDirection()

### API Changes ###
None 

### Beehavioral Changes ###
None

### PR Checklist ###


rmarinho:
build

PR #1394, opened 2017-12-13
[Xaml] open the api for loading xaml from string

StephaneDelcroix:
This should never ever be used as a mechanism for loading your application screens.

It might be used if you're writing tools and IDE that generates xaml.

### DIFF
--- a/Xamarin.Forms.Xaml/ViewExtensions.cs
+++ b/Xamarin.Forms.Xaml/ViewExtensions.cs
@@ -37,7 +37,7 @@ namespace Xamarin.Forms.Xaml
 			return view;
 		}
 
-		internal static TXaml LoadFromXaml<TXaml>(this TXaml view, string xaml)
+		public static TXaml LoadFromXaml<TXaml>(this TXaml view, string xaml)
 		{
 			XamlLoader.Load(view, xaml);
 			return view;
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Xaml/ViewExtensions.cs
```diff
@@ -37,7 +37,7 @@ public static TXaml LoadFromXaml<TXaml>(this TXaml view, Type callingType)
 			return view;
 		}
 
-		internal static TXaml LoadFromXaml<TXaml>(this TXaml view, string xaml)
+		public static TXaml LoadFromXaml<TXaml>(this TXaml view, string xaml)
```
StephaneDelcroix: add a 1000ms delay. this is only meant to be used by editing tool anyway

MelbourneDeveloper:
We are so happy about this PR. This proves that your team is listening, and this will greatly enhance the confidence in Xamarin Forms in our team.

nluyan:
why I am using LoadFromXaml method in a constructor Method to get this exception?
System.MethodAccessException: Method `Xamarin.Forms.Xaml.Extensions:LoadFromXaml<App8.MainPage> (App8.MainPage,string)' is inaccessible from method `App8.MainPage:.ctor ()'

PR #1405, opened 2017-12-14
VisualStateManager phase 1

hartez:
### Description of Change ###

Adds the capability to manage properties on Forms VisualElements via declarative states. States can be declared in code or in XAML. 

The `Xamarin.Forms.Controls\GalleryPages\VisualStateManagerGalleries` folder contains example code for using this feature. 

Also adds the Platform Specific `IsLegacyColorModeEnabled` to provide the option of disabling the partial state/color handling introduced in earlier versions of Forms.

### Bugs Fixed ###

-  Incidentally fixes [37621 – SearchBar BackgroundColor behavior inconsistent on Windows platforms](https://bugzilla.xamarin.com/show_bug.cgi?id=37621)

### API Changes ###

**Added:**

VisualStateManager:

`public static Collection<VisualStateGroup> GetVisualStateGroups(VisualElement visualElement)`
`public static void SetVisualStateGroups(VisualElement visualElement, Collection<VisualStateGroup> value)`
`public static bool GoToState(VisualElement visualElement, string name)`

AndroidSpecific, iOSSpecific, WindowsSpecific:

`public static bool GetIsLegacyColorModeEnabled(BindableObject element)`
`public static void SetIsLegacyColorModeEnabled(BindableObject element, bool value)`
`public static bool GetIsLegacyColorModeEnabled(this IPlatformElementConfiguration<Android, VisualElement> config)`
`public static IPlatformElementConfiguration<Android, VisualElement> SetIsLegacyColorModeEnabled(
		    this IPlatformElementConfiguration<Android, VisualElement> config, bool value)`

### PR Checklist ###


hartez:
build

Xamarin.Forms.Build.Tasks/CompiledConverters/BindablePropertyConverter.cs
```diff
@@ -41,6 +42,16 @@ public FieldReference GetBindablePropertyFieldReference(string value, ModuleDefi
 							typeName = (ttnode as ValueNode).Value as string;
 						else if (ttnode is IElementNode)
 							typeName = ((ttnode as IElementNode).CollectionItems.FirstOrDefault() as ValueNode)?.Value as string ?? ((ttnode as IElementNode).Properties [new XmlName("", "TypeName")] as ValueNode)?.Value as string;
+					} else if (parent.XmlType.Name == "VisualState") {
```
StephaneDelcroix: check that the Namespace is XFUri to avoid mismatches

Xamarin.Forms.Build.Tasks/CompiledValueProviders/SetterValueProvider.cs
```diff
@@ -43,5 +47,28 @@ public IEnumerable<Instruction> ProvideValue(VariableDefinitionReference vardefr
 			//set the value
 			yield return Instruction.Create(OpCodes.Callvirt, setValueRef);
 		}
+
+		bool IsSetterCollection(FieldReference bindablePropertyReference, ModuleDefinition module, BaseNode node, ILContext context)
```
StephaneDelcroix: this method doesn't check if this is a Setter at all, I find the name misleading, and I don't understand what this does

Xamarin.Forms.Build.Tasks/SetPropertiesVisitor.cs
```diff
@@ -1350,6 +1381,162 @@ public static TypeReference GetParameterType(ParameterDefinition param)
...
+			// Then push the value to be added onto the stack and call the adder
+			yield return Instruction.Create(OpCodes.Ldloc, varDef);
+			yield return Instruction.Create(OpCodes.Callvirt, adderRef);
+
+			if (adderRef.ReturnType.FullName != "System.Void")
+				yield return Instruction.Create(OpCodes.Pop);
+		}
+
+			static bool IsSetterCollection(VariableDefinition parent, string localName, INode node, ILContext context)
+		{
```
StephaneDelcroix: 👮

Xamarin.Forms.Build.Tasks/SetPropertiesVisitor.cs
```diff
@@ -1350,6 +1381,162 @@ public static TypeReference GetParameterType(ParameterDefinition param)
...
+			var runTimeName = attributes[0].ConstructorArguments[0].Value as string;
+
+			if (string.IsNullOrEmpty(runTimeName)) 
+				return false;
+
+			Context.IL.Append(SetPropertyValue(variableDefinition, new XmlName("", runTimeName), node, Context, node));
+			return true;
+		}
+
+		static bool CanAddToAttachedProperty(FieldReference bpRef, bool attached, INode node, IXmlLineInfo iXmlLineInfo, ILContext context)
```
StephaneDelcroix: I /think/ this is covered already, but I have to check

Xamarin.Forms.Core/BindablePropertyConverter.cs
```diff
@@ -102,5 +106,17 @@ BindableProperty ConvertFrom(Type type, string propertyName, IXmlLineInfo linein
 				throw new XamlParseException($"The PropertyName of {type.Name}.{name} is not {propertyName}", lineinfo);
 			return bp;
 		}
+
+		Type FindTypeForVisualState(List<object> parentObjects)
+		{
+			var obj = parentObjects.Skip(3).Take(1).FirstOrDefault();  // Skip this Setter, VisualState, and VisualStateGroup
```
StephaneDelcroix: you should skip them one by one, and verify their types

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -0,0 +1,161 @@
+﻿using System;
+using System.Collections.ObjectModel;
+
+namespace Xamarin.Forms
+{
+	public static class VisualStateManager
+	{
+		internal class CommonStates
+		{
+			internal const string Normal = "Normal";
```
StephaneDelcroix: should be public, as the inner class is already internal. won't change a thing, but it'll prevent confusion

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -0,0 +1,161 @@
...
+					}
+				}
+
+				// Update the current state
+				group.CurrentState = target;
+
+				// Apply the setters from the new state
+				foreach (Setter setter in target.Setters)
+				{
+					setter.Apply(visualElement);
```
StephaneDelcroix: don't you need to set `fromStyle: true` ? that'll prevent overriding manually set values (I don't know if it's the intention or not)

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -0,0 +1,161 @@
...
+				{
+					continue;
+				}
+
+				// If we've got a new state to transition to, unapply the setters from the current state
+				if (group.CurrentState != null)
+				{
+					foreach (Setter setter in group.CurrentState.Setters)
+					{
+						setter.UnApply(visualElement);
```
StephaneDelcroix: see below for `fromStyle`

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -0,0 +1,161 @@
...
+	public class VisualStateGroup 
+	{
+		public VisualStateGroup()
+		{
+			States = new Collection<VisualState>();
+		}
+
+		public Type TargetType { get; set; }
+		public string Name { get; set; }
+		public Collection<VisualState> States { get; }
```
StephaneDelcroix: `ICollection`. and same, why not `IList` ?

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -0,0 +1,161 @@
...
+					return state;
+				}
+			}
+
+			return null;
+		}
+
+		internal VisualStateGroup Clone()
+		{
+			var clone =  new VisualStateGroup {TargetType = TargetType, Name = Name, CurrentState = CurrentState};
```
StephaneDelcroix: isn't this the right place to set the CurrentState as "Normal" ?

StephaneDelcroix: that seems a right place as well to throw if the name is null (x:Name is mandatory on VSG)

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -0,0 +1,161 @@
...
+	[RuntimeNameProperty("Name")]
+	public class VisualState 
+	{
+		public VisualState()
+		{
+			Setters = new Collection<Setter>();
+		}
+
+		public string Name { get; set; }
+		public Collection<Setter> Setters { get;}
```
StephaneDelcroix: you know...

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -0,0 +1,161 @@
...
+			{
+				clone.States.Add(state.Clone());
+			}
+
+			return clone;
+		}
+	}
+
+	[RuntimeNameProperty("Name")]
+	public class VisualState
```
StephaneDelcroix: no storyboard, no triggers, right ? if we don't have plans to support them in the future, the ContentProperty could be set to "Setters"

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -0,0 +1,161 @@
...
+			foreach (VisualState state in States)
+			{
+				clone.States.Add(state.Clone());
+			}
+
+			return clone;
+		}
+	}
+
+	[RuntimeNameProperty("Name")]
```
StephaneDelcroix: `nameof(Name)`

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -0,0 +1,161 @@
...
+
+				return true;
+			}
+
+			return false;
+		}
+	}
+
+	[RuntimeNameProperty("Name")]
+	[ContentProperty("States")]
```
StephaneDelcroix: nameof for both

Xamarin.Forms.Core/VisualStateManager.cs
hartez: The plan is to support them in the future.

Xamarin.Forms.Core/Setter.cs
```diff
@@ -57,7 +58,12 @@ internal void Apply(BindableObject target, bool fromStyle = false)
 			else if (dynamicResource != null)
 				target.SetDynamicResource(Property, dynamicResource.Key, fromStyle);
 			else
-				target.SetValue(Property, Value, fromStyle);
+			{
+				if (Value is Collection<VisualStateGroup> visualStateGroupCollection)
+					target.SetValue(Property, visualStateGroupCollection.Clone(), fromStyle);
```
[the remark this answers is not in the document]
hartez: We only want the cloning to happen if the property is being set by a Setter; if it's being set directly on the element in Xaml or in code, we don't want to clone it. Is there some way to determine that a Setter is the source from within the BP?

Xamarin.Forms.Core/VisualStateManager.cs
hartez: Fixed.

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -0,0 +1,161 @@
...
+			internal const string Focused = "Focused";
+		}
+
+		public static readonly BindableProperty VisualStateGroupsProperty =
+			BindableProperty.CreateAttached("VisualStateGroups", typeof(Collection<VisualStateGroup>), typeof(VisualElement), 
+				defaultValue: null, propertyChanged: VisualStateGroupsPropertyChanged);
+
+		static void VisualStateGroupsPropertyChanged(BindableObject bindable, object oldValue, object newValue)
+		{
+			if (bindable is VisualElement visualElement)
```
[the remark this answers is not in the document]
hartez: Fixed - now uses a direct cast.

Xamarin.Forms.Build.Tasks/CompiledConverters/BindablePropertyConverter.cs
hartez: Added the check.

Xamarin.Forms.Core/VisualStateManager.cs
hartez: Fixed.

Xamarin.Forms.Core/RuntimeNamePropertyAttribute.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+
+namespace Xamarin.Forms
+{
+	[AttributeUsage(AttributeTargets.Class)]
+	public sealed class RuntimeNamePropertyAttribute : Attribute
```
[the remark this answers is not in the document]
hartez: Fixed the namespace.

Xamarin.Forms.Core/VisualStateManager.cs
hartez: Fixed.

Xamarin.Forms.Core/RuntimeNamePropertyAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+	{
+		public RuntimeNamePropertyAttribute(string name)
+		{
+			Name = name;
+		}
+
+		public string Name { get; private set; }
```
StephaneDelcroix: turn on r#. it'll tell you you don't have to say `private set;`

Xamarin.Forms.Core/VisualElement.cs
```diff
@@ -88,8 +89,8 @@ public partial class VisualElement : Element, IAnimatable, IVisualElementControl
 		public static readonly BindableProperty MinimumHeightRequestProperty = BindableProperty.Create("MinimumHeightRequest", typeof(double), typeof(VisualElement), -1d, propertyChanged: OnRequestChanged);
 
 		[EditorBrowsable(EditorBrowsableState.Never)]
-		public static readonly BindablePropertyKey IsFocusedPropertyKey = BindableProperty.CreateReadOnly("IsFocused", typeof(bool), typeof(VisualElement), default(bool),
-			propertyChanged: OnIsFocusedPropertyChanged);
+		public static readonly BindablePropertyKey IsFocusedPropertyKey = BindableProperty.CreateReadOnly("IsFocused",
```
StephaneDelcroix: I hate that public BP

Xamarin.Forms.Core/VisualElement.cs
```diff
@@ -783,9 +784,26 @@ static void FlowDirectionChanged(BindableObject bindable, object oldValue, objec
...
+
+			VisualStateManager.GoToState(element, isEnabled 
+				? VisualStateManager.CommonStates.Normal 
+				: VisualStateManager.CommonStates.Disabled);
+		}
+
 		static void OnIsFocusedPropertyChanged(BindableObject bindable, object oldvalue, object newvalue)
 		{
-			var element = bindable as VisualElement;
+			if (!(bindable is VisualElement element))
```
StephaneDelcroix: check for GO.GetIsDefault(). element will only be a VE, or null

Xamarin.Forms.Xaml/CollectionExtensions.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+
+namespace Xamarin.Forms.Xaml
+{
+	public static class CollectionExtensions
+	{
+		public static object EnsureCollectionInitialized(this BindableObject bindable, BindableProperty property)
+		{
```
StephaneDelcroix: it's not up to the Xaml to create collection if not explicitly declared. if you want to add to a collection, that collection has to exists (can be lazily created in the property getter).
we're doing this at other places

Xamarin.Forms.Xaml.UnitTests/VisualStateManagerTests.xaml.cs
```diff
@@ -0,0 +1,133 @@
...
+
+				var groups1 = VisualStateManager.GetVisualStateGroups(label1);
+				var groups2 = VisualStateManager.GetVisualStateGroups(label2);
+
+				Assert.AreNotSame(groups1, groups2);
+
+				var currentState1 = groups1[0].CurrentState;
+				var currentState2 = groups2[0].CurrentState;
+
+				Assert.That(currentState1.Name == "Normal");
```
StephaneDelcroix: either use
```
Assert.AreEqual(currentState1.Name, "Normal")
```
or
```
Assert.That(currentState1.Name, IS.EqualTo("Normal")) // actually my preferred form
```
or even
```
Assert.True(currentState1.Name == "Normal")
```

but try to avoid this

Xamarin.Forms.Xaml.UnitTests/VisualStateManagerTests.xaml.cs
```diff
@@ -0,0 +1,133 @@
...
+
+		[TestFixture]
+		public class Tests
+		{
+			[SetUp]
+			public void SetUp ()
+			{
+				Device.PlatformServices = new MockPlatformServices ();
+				Application.Current = new MockApplication ();
+			}
```
StephaneDelcroix: add a TearDown, both of those are static, and could influence the test running next

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -0,0 +1,158 @@
...
+				return true;
+			}
+
+			return false;
+		}
+	}
+
+	[RuntimeNameProperty(nameof(Name))]
+	[ContentProperty(nameof(States))]
+	public class VisualStateGroup
```
StephaneDelcroix: as the `Clone` is internal, this class should be sealed.

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -0,0 +1,158 @@
...
+			{
+				clone.States.Add(state.Clone());
+			}
+
+			return clone;
+		}
+	}
+
+	[RuntimeNameProperty(nameof(Name))]
+	public class VisualState
```
StephaneDelcroix: seal as well

Xamarin.Forms.Core/VisualElement.cs
hartez: It was like that when I got here.  ¯\_(ツ)_/¯

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -0,0 +1,161 @@
...
+	{
+		internal class CommonStates
+		{
+			internal const string Normal = "Normal";
+			internal const string Disabled = "Disabled";
+			internal const string Focused = "Focused";
+		}
+
+		public static readonly BindableProperty VisualStateGroupsProperty =
+			BindableProperty.CreateAttached("VisualStateGroups", typeof(Collection<VisualStateGroup>), typeof(VisualElement),
```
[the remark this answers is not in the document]
hartez: Fixed, using IList<VSG> now.

Xamarin.Forms.Core/VisualStateManager.cs
hartez: Setters aren't the only way to set the VisualStateGroups for an element. They can also be set directly in Xaml or from code. And we don't want to set CurrentState to normal directly on the property - we need it to be set with GoToState so the values for the Normal state are applied to the element.

Xamarin.Forms.Core/RuntimeNamePropertyAttribute.cs
hartez: I have to turn if off when I'm working in your Xaml code, otherwise it formats everything correctly :)

Xamarin.Forms.Core/VisualElement.cs
hartez: Are you saying that GetIsDefault will be faster than the pattern matching for this?

Xamarin.Forms.Core/VisualStateManager.cs
hartez: Fixed, now sets fromStyle.

Xamarin.Forms.Core/VisualStateManager.cs
hartez: Fixed, now sets fromStyle.

Xamarin.Forms.Core/VisualStateManager.cs
hartez: Now using IList.

Xamarin.Forms.Core/VisualStateManager.cs
hartez: Yep. Fixed.

Xamarin.Forms.Core/RuntimeNamePropertyAttribute.cs
hartez: Fixed.

Xamarin.Forms.Xaml.UnitTests/VisualStateManagerTests.xaml.cs
hartez: Added the TearDown.

Xamarin.Forms.Xaml.UnitTests/VisualStateManagerTests.xaml.cs
hartez: Fixed.

Xamarin.Forms.Core/VisualElement.cs
```diff
@@ -783,9 +784,26 @@ static void FlowDirectionChanged(BindableObject bindable, object oldValue, objec
...
+		static void OnIsEnabledPropertyChanged(BindableObject bindable, object oldValue, object newValue)
+		{
+			if (!(bindable is VisualElement element))
+			{
+				return;
+			}
+
+			var isEnabled = (bool)newValue;
+
+			VisualStateManager.GoToState(element, isEnabled
```
nickrandolph: Please don't call the VisualStateManager directly. There should be a method such as "UpdateVisualState" that is called whenever the VisualElement thinks the visual state should be updated. This should be avirtual method so that developers can easily sub-class and override the visual state manager behaviour.

StephaneDelcroix:
I misread. 
Don’t GetIsDefaut. 
Direct cast and null comparison

Xamarin.Forms.Xaml/CollectionExtensions.cs
hartez: Fixed for BindableObject - the collections are now initialized lazily when the value is retrieved.

Xamarin.Forms.Core/VisualStateManager.cs
hartez: Fixed.

Xamarin.Forms.Core/VisualStateManager.cs
hartez: Done.

Xamarin.Forms.Core/VisualElement.cs
hartez: Using direct cast now.

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -0,0 +1,161 @@
...
+		public static void SetVisualStateGroups(VisualElement visualElement, Collection<VisualStateGroup> value)
+		{
+			visualElement.SetValue(VisualStateGroupsProperty, value);
+		}
+
+		public static bool GoToState(VisualElement visualElement, string name)
+		{
+			if (!(visualElement.GetValue(VisualStateGroupsProperty) is Collection<VisualStateGroup> groups))
+			{
+				return false;
```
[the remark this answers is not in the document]
hartez: Now using GetIsDefault.

Xamarin.Forms.Build.Tasks/CompiledValueProviders/SetterValueProvider.cs
hartez: The idea of the method is to check whether the setter is for a collection; renamed the method to `SetterValueIsCollection` to (hopefully) be more clear. It short-circuits ProvideValue if the Setter is for a collection, since the collection will be intialized later when the first item is added.

Currently I have collections in Setters (e.g. `IList<VisualStateGroup>`) being initialized when the first item is added to them (using `EnsureSetterCollectionInitialized`). I suspect that this should instead be happening in `IValueProvider.ProvideValue`/`ICompiledValueProvider.ProvideValue`, but if there's a simple way to do that, I haven't been able to see it yet. 

Any suggestions?

hartez:
> if you do not define your VSM in Style, XamlG will probably try to generate multiple fields with the name "Normal". 

Tested this, and you are absolutely correct.

>The x:Name handling and RuntimeNamePropertyAttribute should be handled in the visitor handling x:Name and namescopes (creading a new node, skipping the x:Name setting)

So in the NameScopeVisitor, should I be creating a new namescope when I've got a VisualStateGroups node which is not part of a Style?

Xamarin.Forms.Core/BindablePropertyConverter.cs
hartez: Added type verification.

hartez:
> if you do not define your VSM in Style, XamlG will probably try to generate multiple fields with the name "Normal".
This should be fixed now.

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -0,0 +1,357 @@
...
+		}
+
+		public Type TargetType { get; set; }
+		public string Name { get; set; }
+		public IList<VisualState> States { get; }
+		public VisualState CurrentState { get; internal set; }
+
+		internal VisualState GetState(string name)
+		{
+			foreach (VisualState state in States)
```
tpetrina: If this method is called often, does it make sense to switch to for loop? Or is this premature optimisation?

Xamarin.Forms.Build.Tasks/SetNamescopesAndRegisterNamesVisitor.cs
```diff
@@ -117,7 +123,7 @@ void SetNameScope(ElementNode node, VariableDefinition ns)
 		void RegisterName(string str, VariableDefinition namescopeVarDef, IList<string> namesInNamescope, VariableDefinition element, INode node)
 		{
 			if (namesInNamescope.Contains(str))
-				throw new XamlParseException($"An element with the name \"{str}\" already exists in this NameScope", node as IXmlLineInfo);
+				throw new XamlParseException($"An element with the name \"{str}\" already exists in this NameScope 1", node as IXmlLineInfo);
```
StephaneDelcroix: ?

hartez: Fixed.

Xamarin.Forms.Core/VisualStateManager.cs
hartez: I didn't think there'd be a significant difference, since the underlying implementation is using List<T>. But just to be certain I wrote up some quick-and-dirty benchmarking tests to see if `for` would make any difference vs `foreach` here. 

For the first batch, I created VisualStateGroupLists with 1 group, and the group had 2 states (which I think will be a pretty common scenario). I then had the test run through 100,000 iterations of moving randomly between the states (using the `GoToState` method). Between the two implementations (`foreach` and `for`), the test runs took the same amount of time on average.

For the second batch, I created an uglier scenario with VisualStateGroupLists with 10 groups, each group having 10 states. Running that over 100,000 iterations, the `foreach` actually ended up being slightly faster. 

Looking at the tests in the profiler, the bulk of the time spent in the method (in either implementation) is the string comparison. The looping method doesn't end up makeing much difference. 

So for the time being I think we'll stick with `foreach` for clarity. If we start running into performance problems with this later, we can revisit it.

rmarinho:
failed test not related

Xamarin.Forms.Core/VisualStateManager.cs
jassmith: The difference isn't in performance but in allocation. foreach generates an iterator which adds some GC pressure. It probably doesn't matter much though.

hartez: The enumerator for List<T> is a struct, so it's pretty low-impact.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/DefaultColorToggleTest.cs
```diff
@@ -6,19 +6,19 @@
 namespace Xamarin.Forms.Controls
 {
 	[Preserve (AllMembers=true)]
-	[Issue (IssueTracker.None, 0, "Default colors toggle test", PlatformAffected.All)]
+	[Issue (IssueTracker.None, 9906753, "Default colors toggle test", PlatformAffected.All)]
```
rmarinho: you can drop the 990 now

hartez: No, I like the number 9906753.

PR #1417, opened 2017-12-18
replaced ACtivityIndicator renderer for UWP to display ProgressRing

mister-giga:
### Description of Change ###

I just changed the UWP renderer for ActivityIndicator from FormsProgressBar to ProgressRing, 
Since the ActivityIndicator has round spinner on Android and same look on iOS, it shall be logical to have ProgressRing on UWP. It will have simillar look and it has even more native look, since most activity indicators are aligned center to whole content, same applies to ProgressRing while ProgressBar is usually aligned at vertically top and horizontally stretched, it makes big UI misunderstanding for Windows users to see ugly horizontal indicator on the center of their screens.

### Bugs Fixed ###

- Provide links to bugs here

### API Changes ###

Added:
 - void ListView.Clear ();

Changed:
 - object ListView.SelectedItem => Cell ListView.SelectedItem

### Behavioral Changes ###

### PR Checklist ###


mister-giga:
Sure, good point!

pauldipietro:
This didn't need to be closed; you could just do a new commit/amended commit with the changes removed and it would still apply to this PR.

mister-giga:
If only I know how to do that :D

mister-giga:
@pauldipietro  does it look better now?

pauldipietro:
Yes, thanks!

mister-giga:
@pauldipietro do you know how is process of merging pull requests with master? Will anyone with write access assign this to himself or how is it happening?

pauldipietro:
This is technically a "breaking" change in terms of what a user might expect; the current indicator is fine and working in the most basic terms, so this sort of change in particular would need to be further reviewed to see if it's something we would move forward with to replace the current implementation. The original request I made was simply to get the PR in order and remove unnecessary changes. I apologize if that confused you in any way.

I should state that we had an [open thread](https://forums.xamarin.com/discussion/100973/activityindicator-rendered-as-a-progressring-in-uwp) about this at one point a number of months ago. There in that thread is a referred to [MSDN Doc](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/progress-controls) which states:

> The ProgressRing only has an indeterminate state, and should be used when any further user interaction is blocked until the operation has completed.

In this case, on the surface, I think we could substitute one for the other, but I'll probably just defer to @jassmith on this one. 🙂

mister-giga:
Can we make a poll? I believe more than 90 UWP developers with forms will like this new change including me.

MarkoBL:
I think the ActivityIndicator should be extended to support both, a progress ring or bar. Or split into RingActivityIndicator and BarActivityInidicator.

Having both types of indicator should be the goal.

mister-giga:
How it may break? It corrects the wrong implementation of renderer on UWP platform

mister-giga:
How can this be an effect? This is a different native control and if i modify name of ActivityIndicator to BarActivityIndicator and add new control such as RingActivityIndicator, it will really break all apps that use Activity Indicator.

mister-giga:
API changes section contains wrong data, i dont know who wrote that. But this equest shall not be closed

jassmith:
Unfortunately this change cannot be accepted as it would completely break users existing apps. I agree with what this PR is doing, but it must remain closed.

mister-giga:
How may it break? 
@jassmith  Displaying what user expects means 'breaking' for you?

pauldipietro:
@mister-giga What a given user sees now is what they are expecting, which is why this PR cannot be taken. Their UI might be designed around the indicator being used in a certain way, and its sudden changing would as such be "breaking" as it is an entirely different control, and not just an adjustment to something such as a bug in the layout. Per earlier conversation and @rmarinho's comment, this would be better served as a Platform Specific or an effect.

Issue #1333, opened 2017-12-06
iOS app crashes occasionally with System.ArgumentOutOfRangeException since Xamarin.Forms 2.5


ssc4gemos:
# Steps to reproduce
Not found yet.

# Expected behavior
No crashes.

# Actual behavior
My iOS app occasionally crashes with System.ArgumentOutOfRangeException since the update from Xamarin.Forms 2.4 to 2.5.

# Supplemental info (logs, images, videos)
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
  at System.ThrowHelper.ThrowArgumentOutOfRangeException (System.ExceptionArgument argument, System.ExceptionResource resource) [0x00018] in /Library/Frameworks/Xamarin.iOS.framework/Versions/11.4.0.214/src/mono/mcs/class/referencesource/mscorlib/system/throwhelper.cs:93 
  at System.ThrowHelper.ThrowArgumentOutOfRangeException () [0x00000] in /Library/Frameworks/Xamarin.iOS.framework/Versions/11.4.0.214/src/mono/mcs/class/referencesource/mscorlib/system/throwhelper.cs:56 
  at System.Collections.Generic.List`1[T].get_Item (System.Int32 index) [0x00009] in <1f5960abbfdf4434802d3b8a03ded4fa>:0 
  at System.Collections.ObjectModel.Collection`1[T].get_Item (System.Int32 index) [0x00000] in <1f5960abbfdf4434802d3b8a03ded4fa>:0 
  at System.Collections.ObjectModel.ReadOnlyCollection`1[T].get_Item (System.Int32 index) [0x00000] in <1f5960abbfdf4434802d3b8a03ded4fa>:0 
  at Xamarin.Forms.Platform.iOS.RendererPool.UpdateRenderers (Xamarin.Forms.Element newElement) [0x0004e] in D:\agent\_work\1\s\Xamarin.Forms.Platform.iOS\RendererPool.cs:150 
  at Xamarin.Forms.Platform.iOS.RendererPool.UpdateNewElement (Xamarin.Forms.VisualElement newElement) [0x00080] in D:\agent\_work\1\s\Xamarin.Forms.Platform.iOS\RendererPool.cs:78 
[...]

PR #1463, opened 2017-12-23
Fix IndexOutOfRangeException for iOS ListView

kingces95:
### Description of Change ###

If iOS requests height for a cell whose item we just removed then when we ask for the item we'll generate an IndexOutOfRangeException. Fix is to verify index before indexing. 

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=59412
https://github.com/xamarin/Xamarin.Forms/issues/1333 (possible, but no repro)

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


dimonkomaron:
What version of Xamarin Forms will contain this fix? Is this fix available in pre-release version?

NPadrutt:
Is here any news when this should be available? I checked the available pre versions but it seems it isn't merged to them.

zoli13:
**Xamarin.Forms.3.0.0.296286-pre2**

Crashing. It used to work fine in Xamarin.Forms.2.5.0.122203. After upgrading, I have this issue on iOS only. Android is OK.
I have a tabbed view with 3 tabs. When 2nd tab is clicked:

```
private void PageABC_Appearing(object sender, EventArgs e)
{
	ViewModel.BuildList();
}
```

```
class ViewModel

public ObservableCollection<PersonPaymentList> PersonPaymentList { get; }
internal void BuildList()
{
	MYList.Clear();

	....
}
```

samhouts:
@zoli13 I think you're experiencing #1333/#2145. Please follow those issues for updates. Thanks.

PR #1465, opened 2017-12-23
[Core] AutoCommandExtension

workgroupengineering:
### Description of Change ###

AutoCommand Extension allows to simplify the use of the Command through conventions, making the ViewModel more readable.

AutoCommand searches on viewModel the methods that have the following signature:

bool Can[CommadName] (object parameter); 

void [CommandName](object parameter)

where [CommandName] is the value of the Commad parameter of AutoCommand

Example:
```xml
  ...
  <StackLayout>
        <Label Text="Enable Button 1" />
	<Switch IsToggled ="{Binding EnableExecute}"/>
	<Label Text="Enable Button 2" />
        <Button Command="{AutoCommand Do}"
    		      Text="Execute"/>
   </StackLayout>
  ...
```

```csharp
class ViewModel : INotifyPropertyChanged
{
	public event PropertyChangedEventHandler PropertyChanged;
	bool  _enableExecute;
	public bool  EnableExecute
	{
		get { return _enableExecute; }
		set
		{
			if (_enableExecute == value)
				return;
			__enableExecute = value;
			OnPropertyChanged();
		}
	}

	[DependedOn(nameof(EnableExecute))]
	bool CanDo(object parameter)
	{
		return EnableExecute;
	}

	void Do(object parameter)
	{

	}
	protected void OnPropertyChanged([CallerMemberName] string propertyName = null)
	{
		PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
	}
}
```
### Bugs Fixed ###

none

### API Changes ###

List all API changes here:

Added:
 -AutoCommandExtension 
 -DependedOn

### Behavioral Changes ###

none

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/ButtonUnitTest.cs
+++ b/Xamarin.Forms.Core.UnitTests/ButtonUnitTest.cs
@@ -35,42 +35,57 @@ namespace Xamarin.Forms.Core.UnitTests
 		}
 
 		[Test]
-		public void TestClickedvent ()
+		[TestCase(true)]
+		[TestCase(false)]
+		public void TestClickedvent (bool isEnabled)
 		{
-			var view = new Button ();
+			var view = new Button()
+			{
+				IsEnabled= isEnabled,
+			};
 
 			bool activated = false;
 			view.Clicked += (sender, e) => activated = true;
 
 			((IButtonController) view).SendClicked ();
 
-			Assert.True (activated);
+			Assert.True (activated == isEnabled ? true : false);
 		}
 
 		[Test]
-		public void TestPressedEvent ()
+		[TestCase(true)]
+		[TestCase(false)]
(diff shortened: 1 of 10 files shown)


workgroupengineering:
In these lines of code ([107](https://github.com/xamarin/Xamarin.Forms/pull/1465/files#diff-4b8377d3500593d80e27dc263220b42fR107),[117](https://github.com/xamarin/Xamarin.Forms/pull/1465/files#diff-4b8377d3500593d80e27dc263220b42fR117)), I have a doubt. Is it better to generate an exception or a warning?

Please review.

rmarinho:
Hi thanks, but we don't think this is needed in the core of Xamarin.Forms. Also new features should be discussed as issues first to get feedback from the team. 

Thanks

PR #1487, opened 2018-01-02
[iOS] Added icon to context menu actions instead of text **Behavior change**

Sebastian1989101:
### Description of Change ###
Changed context action menu of ViewCells to Icons like it is in native apps and on Android instead of Text.

### Bugs Fixed ###

- None

### API Changes ###

- None

### Behavioral Changes ###
Instead of text the icon will appear in context actions like it is on Android.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/ContextActionCell.cs
+++ b/Xamarin.Forms.Platform.iOS/ContextActionCell.cs
@@ -407,9 +407,12 @@ namespace Xamarin.Forms.Platform.iOS
 			else
 				button.SetBackgroundImage(DestructiveBackground, UIControlState.Normal);
 
-			button.SetTitle(item.Text, UIControlState.Normal);
-			button.TitleEdgeInsets = new UIEdgeInsets(0, 15, 0, 15);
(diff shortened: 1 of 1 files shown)


rmarinho:
Won't this break everyone that has a item with icon and text ? 
IF they update the app , they will see a different UI without expecting it. I don't think this could be done like this to be backwards compatible.

rmarinho:
@Sebastian1989101 i will discuss with the team.

jassmith:
Unfortunately this will be a massive breaking change for some of our users. The behavioral change will have to be locked behind a PlatformSpecific to be acceptable. I think we could accept it if written in that format.

Sebastian1989101:
@jassmith so it is a "breaking" change that something works the same way on all platforms? In my opinion it is more a breaking change that it is different across the platforms because I would not expect that. Currently there is no way to handle this with a custom renderer or something. Also I'm sure that this is not 100% compatible with Apples iOS Guidelines for User Interfaces because it is not the same behaviour as it would be with a native implementation.

PR #1488, opened 2018-01-02
[WPF] TimePickerRenderer (and updated WPF nuspecs)

jsuarezruiz:
### Description of Change ###

Added **TimePickerRenderer** (only renderer that was missing) to the WPF backend.
timepickerrenderer wpf

The WPF **nuspec** has also been changed to add the dependency with WPFToolkit.

### Bugs Fixed ###

- None.

### Behavioral Changes ###

Now it is possible to use the TimePicker control from the WPF backend.

### PR Checklist ###


jsuarezruiz:
At the moment there is no renderer for the TimePicker in WPF. In this PR the renderer is included.

rmarinho:
I think we should try not to add extra dependencies, let's open a issue for now about this missing renderer and try to implement a custom control on WPF for this instead of bringing a extra toolkit.

PR #1515, opened 2018-01-06
[iOS]ability to custom iPhone X Home Indicator Auto Hidden behavior

bosxixi:
### Description of Change ###

ability to custom iPhone X Home Indicator Auto Hidden behavior


pauldipietro:
Please refer to this [guide to contributing](https://github.com/xamarin/Xamarin.Forms/blob/master/CONTRIBUTING.md) before submitting a PR with API changes, as such changes must first be made as proposals for review and discussion.

adrianknight89:
Also, the overridden method as well as the method call should be run on iOS 11+ as (I think) they are not available in previous versions.

PR #1521, opened 2018-01-07
Rename ConvertExtensions to ColorExtensions and change access to public

mubold:
### Description of Change ###
Color extensions to convert a Forms Color to a UWP color / brush are not accessible since the class is marked as internal instead of public. I also renamed the class to ColorExtensions to be consistent with all other platforms.

### Bugs Fixed ###
Could not use ToWindowsColor or ToBrush in UWP Forms renderers.

### API Changes ###

Changed:
 - internal static class ConvertExtensions => public static class ColorExtensions

### PR Checklist ###
I don't think this requires tests?


pauldipietro:
Thank you for contribution, but this PR will be closed as we require proposals for all API changes. Please read our [contribution guide](https://github.com/xamarin/Xamarin.Forms/blob/master/CONTRIBUTING.md) for future reference.

PR #1535, opened 2018-01-08
Rename ConvertExtensions to ColorExtensions and change access to public

### LINKED ISSUE #1522: [UWP] Extension methods for Color ToWindowsColor() and ToBrush() not accessible
### Description
The ``ToWindowsColor()`` and `ToBrush()`` extension methods for Color are not available on UWP because the static class ConvertExtensions in which they are found is ``internal`` instead of ``public``
See : https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.UAP/ConvertExtensions.cs#L6

Also, less important but all other implementations of this function on other platforms are in a class named ColorExtensions, I suggest renaming it to be more consistent.

### Expected Behavior

The static class should be public instead of internal, so that those methods can be used outside of Xamarin.Forms, especially useful for renderers.

### Actual Behavior

Extension methods cannot be used.

mubold:
### Description of Change ###
Color extensions to convert a Forms Color to a UWP color / brush are not accessible since the class is marked as internal instead of public. I also renamed the class to ColorExtensions to be consistent with all other platforms.

### Bugs Fixed ###
- fixes #1522

### API Changes ###

Changed:
 - internal static class ConvertExtensions => public static class ColorExtensions

### Behavioral Changes ###
/

### PR Checklist ###
I don't think this requires tests?

### DIFF
--- a/Xamarin.Forms.Platform.UAP/ColorExtensions.cs
+++ b/Xamarin.Forms.Platform.UAP/ColorExtensions.cs
@@ -1,9 +1,8 @@
-﻿using System.Drawing;
-using Windows.UI.Xaml.Media;
+﻿using Windows.UI.Xaml.Media;
 
 namespace Xamarin.Forms.Platform.UWP
 {
-	internal static class ConvertExtensions
+	public static class ColorExtensions
 	{
 		public static Brush ToBrush(this Color color)
 		{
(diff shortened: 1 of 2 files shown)


pauldipietro:
Please don't submit PRs until the proposals are approved. If the original proposal is approved as requested, you may reopen this issue. Thanks!

CZEMacLeod:
@mubold @pauldipietro I notice that ToWindowsColor is now covered with the PR for auto conversion between Color types which also includes various edge case scenarios. I think this code, internal or otherwise, should use the new conversion.

StephaneDelcroix:
I'm revisiting this old item, and we probably should have accepted this back in the day. I took your PR and updated it while maintaining the commit author in #4557. Once #4557 is merged to master, this PR will be closed too. Thanks for your contribution

PR #1570, opened 2018-01-12
[Android] Button Border can be set independent of other properties and will not change the size of the Button. fixes #1436 **behavior change**

### LINKED ISSUE #1436: Button border not drawn on Android without a BorderRadius
Migrated from https://bugzilla.xamarin.com/show_bug.cgi?id=36031

### Description
If a Button has a BorderColor and BorderWidth set, but a 0 BorderRadius, when rendered on Android, the border will not be shown.

via @samhouts: Unfortunately, this fix had unintended visual side effects and must be reverted at this time. We will revisit this issue.

https://github.com/xamarin/Xamarin.Forms/pull/1178 

### Basic Information

- Version with issue: 2.5.0

### Reproduction 

```
using Xamarin.Forms;
using System.Reflection;

namespace xamarinforms2
{
	{
		public App ()
		{
			var assembly = typeof(Button).GetTypeInfo().Assembly;
			var assemblyName = new AssemblyName(assembly.FullName);
			string xamarinFormsDllVersion = assemblyName.Version.Major + "." + assemblyName.Version.Minor + "." + assemblyName.Version.Minor + "." + assemblyName.Version.Build;
			StackLayout stackLayout = new StackLayout {
				VerticalOptions = LayoutOptions.Center,
				Children = {
					new Label {
						Text = "Xamarin.Forms v" + xamarinFormsDllVersion,
					},
					new Button {
						Text = "[default]",
						HorizontalOptions = LayoutOptions.Center,
					},
					new Button {
						Text = "BorderWidth = 1, BorderRadius = [default],",
						HorizontalOptions = LayoutOptions.Center,
						BackgroundColor = Color.Black,
						BorderColor = Color.Red,
						BorderWidth = 1,
					},
					new Button {
						Text = "BorderWidth = 1, BorderRadius = 0,",
						HorizontalOptions = LayoutOptions.Center,
						BackgroundColor = Color.Black,
						BorderColor = Color.Red,
						BorderWidth = 1,
						BorderRadius = 0,
					},
					new Button {
						Text = "BorderWidth = 1, BorderRadius = 1,",
						HorizontalOptions = LayoutOptions.Center,
						BackgroundColor = Color.Black,
						BorderColor = Color.Red,
						BorderWidth = 1,
						BorderRadius = 1,
					},
					new Button {
						Text = "BorderWidth = 1, BorderRadius = Device.OnPlatform<int>(0, 1, 0),",
						HorizontalOptions = LayoutOptions.Center,
						BackgroundColor = Color.Black,
						BorderColor = Color.Red,
						BorderWidth = 1,
						BorderRadius = Device.OnPlatform<int>(0, 1, 0),
					},
				},
			};
			MainPage = new ContentPage {
				Content = stackLayout,
			};
		}
	}
}

```

samhouts:
### Description of Change ###

This fix reverts #1178 and restores the changes made in #941, which will allow Android `Button`s to have a border without a `BorderRadius` specified. 

It goes further to allow a border to be added and/or border radius to be changed without changing the `BackgroundColor` of the `Button`.

It also keeps the size of the `Button` the same as a default Material button, draws a shadow, and preserves the "ripple" effect when the `Button` is pressed.

It does *not* preserve the state list animation that causes the `Button` to slowly fade to the pressed color and increase elevation when long pressed. That can be added in a future PR if necessary

This PR has a potential breaking behavior change. It hardcodes the default color of the button on pre-AppCompat because the theme color does not seem to match the actual color that is used.

We have also marked `Button.BorderRadius` as obsolete in favor of `Button.CornerRadius`. The `BorderRadius` property will forward to the `CornerRadius` property.

Before (AppCompat):
before

Before (pre-AppCompat):
before-preappc

Note: The test description lies on pre-AppCompat, since the default `Button` color is actually darker than the color that the "darker" button is. Apparently, the definition of `Color.Gray` is different.

with #941 (AppCompat):
before-with-941

with #941 (pre-AppCompat):
before-with-941-preappc

After (AppCompat):
after

After (pre-AppCompat):
after-preappc

Note that the shadow is a lighter color of the `BackgroundColor`, matching the pressed color of the `Button`. When a blue background is paired with a red border color, it looks a bit odd, like the `Button` is expanding past the border. However, it looks fine with a sensible color combination, as seen with the gray `Button` with red border.

### Bugs Fixed ###

- fixes #1436 
- https://bugzilla.xamarin.com/show_bug.cgi?id=36031
- https://bugzilla.xamarin.com/show_bug.cgi?id=60736

#### Deprecated (note, marked obsolete but still present and functional):
```csharp
- [Obsolete] Button.BorderRadius
- [Obsolete] Button.BorderRadiusProperty
```

#### Added:
```csharp
- Button.CornerRadius
- Button.CornerRadiusProperty
```

### Behavioral Changes ###

- `Button.BorderRadius` is now obsolete. Please use `Button.CornerRadius` now.
- Default value of `Button.BackgroundColor` is now hardcoded on pre-AppCompat instead of derived from a theme resource.
- Shadow color in Pre-AppCompat is slightly lighter when `BorderRadius`/`BorderColor`/`BorderWidth` are specified.

### PR Checklist ###


StephaneDelcroix:
@samhouts to summarize the discussion we had yesterday, I'd recommend **not** using a defaultValueCreator, but using a default value of `-1` meaning _whatever-the-default-is-for-that-control-on-this-platform_.

to mitigate the breaking change, you could e.g. deprecate `BorderRadius` in favor of `CornerRadius` (shared with `Frame`).

Xamarin.Forms.Core/Button.cs
```diff
@@ -10,6 +10,14 @@ namespace Xamarin.Forms
 	[RenderWith(typeof(_ButtonRenderer))]
 	public class Button : View, IFontElement, ITextElement, IButtonController, IElementConfiguration<Button>
 	{
+		const double DefaultSpacing = 10;
+		const int DefaultBorderRadius = 5;
+
+		public const int DefaultCornerRadius = -1;
+		public const double DefaultBorderWidth = -1;
+		public static Color DefaultBorderColor = Color.Default;
+		public static Color DefaultTextColor = Color.Default;
```
StephaneDelcroix: those 2 aren't magic. they do not require fields.

Xamarin.Forms.Core/Button.cs
```diff
@@ -31,32 +39,42 @@ public class Button : View, IFontElement, ITextElement, IButtonController, IElem
 
 		public static readonly BindableProperty FontAttributesProperty = FontElement.FontAttributesProperty;
 
-		public static readonly BindableProperty BorderWidthProperty = BindableProperty.Create("BorderWidth", typeof(double), typeof(Button), -1d);
+		public static readonly BindableProperty BorderWidthProperty = BindableProperty.Create("BorderWidth", typeof(double), typeof(Button), DefaultBorderWidth);
```
StephaneDelcroix: I don't see the need to introduce a level of indirection here. one can check the default value by doing `BorderWidthProperty.GetDefaultValue()`

Xamarin.Forms.Core/VisualElement.cs
```diff
@@ -7,6 +7,8 @@ namespace Xamarin.Forms
 {
 	public partial class VisualElement : Element, IAnimatable, IVisualElementController, IResourcesProvider
 	{
+		public static Color DefaultBackgroundColor = Color.Default;
```
StephaneDelcroix: same

Xamarin.Forms.Core/Button.cs
```diff
@@ -31,32 +39,42 @@ public class Button : View, IFontElement, ITextElement, IButtonController, IElem
...
 
-		public static readonly BindableProperty BorderColorProperty = BindableProperty.Create("BorderColor", typeof(Color), typeof(Button), Color.Default);
+		public static readonly BindableProperty BorderColorProperty = BindableProperty.Create("BorderColor", typeof(Color), typeof(Button), DefaultBorderColor);
 
-		public static readonly BindableProperty BorderRadiusProperty = BindableProperty.Create("BorderRadius", typeof(int), typeof(Button), 5);
+		[Obsolete("BorderRadiusProperty is obsolete as of 2.5.0. Please use CornerRadius instead.")]
+		public static readonly BindableProperty BorderRadiusProperty = BindableProperty.Create("BorderRadius", typeof(int), typeof(Button), defaultValue: DefaultBorderRadius,
+			propertyChanged: BorderRadiusPropertyChanged);
+
+		public static readonly BindableProperty CornerRadiusProperty = BindableProperty.Create("CornerRadius", typeof(int), typeof(Button), defaultValue: DefaultCornerRadius,
```
StephaneDelcroix: can we promote `CornerRadius` to `BorderElement` (shared with `Frame`) so we could add the `border-radius` CSS property targeting both ?

Xamarin.Forms.Core/Button.cs
```diff
@@ -240,6 +258,43 @@ void OnSourcePropertyChanging(ImageSource oldvalue, ImageSource newvalue)
...
+			var oldVal = (int)bindable.GetValue(Button.BorderRadiusProperty);
+#pragma warning restore
+
+			if (oldVal == val)
+				return;
+
+#pragma warning disable 0618 // retain until BorderRadiusProperty removed
+			bindable.SetValue(Button.BorderRadiusProperty, val);
+#pragma warning restore
+		}
```
StephaneDelcroix: I'm trying to think about all the things that could go wrong here:
- binding => should be fine, unless user binds to the 2 properties and has a poor INPC implementation in VM
- clearvalue => could you please add a unit test for button making sure that when a property is Cleared (using `ClearValue` both are actually cleared ? i.e. IsSet() == false; ?

StephaneDelcroix:
Is this still [breaking] and a **behavior change** ?

Xamarin.Forms.Core/Button.cs
samhouts: I'm targeting 15-5, so that's not in this branch. We can add it when it goes to master. :)

Xamarin.Forms.Core/Button.cs
samhouts: Unit tests added! However, it looks like we don't clear the context when we call `ClearValue`, so `IsSet` returns true in that case. Added a test to `BindableObject` to show that it's not specific to this case.

samhouts:
@StephaneDelcroix Yep, still a breaking/behavior change, as indicated in the updated PR description, although it should be minor.

Xamarin.Forms.Core/Button.cs
StephaneDelcroix: 😢

Xamarin.Forms.Core.UnitTests/BindableObjectUnitTests.cs
```diff
@@ -890,6 +890,23 @@ public void IsSetIsTrueWhenPropSet()
...
+
+			var bindableProperty = BindableProperty.Create("Foo", typeof(string), typeof(MockBindable), defaultValue);
+			var bindable = new MockBindable();
+
+			bindable.SetValue(bindableProperty, newValue);
+
+			bindable.ClearValue(bindableProperty);
+
+			var isSet = bindable.IsSet(bindableProperty);
+			Assert.IsTrue(isSet);
```
StephaneDelcroix: I'll look at fixing this, or investigating why it is that way. if `ClearValue` is dropping the context, you could `ClearValue` the other property when the first is `!IsSet()` in your onPropertyChanged

samhouts:
Since we touched the Core code, we need to ensure that we didn't break iOS or UWP with this change. Here's proof:

iOS: Before
io-before


iOS: After
io-after


UWP Before:
uwp-before


UWP After
uwp-after

bill2004158:
How can I disable shadow for the button which I don't need it?

bill2004158:
it seems the new version added some Padding for the button?

Issue #1423, opened 2017-12-19
Xamarin.Forms UWP ButtonRenderer UpdateContent() ImageOpened event NullReferenceException


jalbertSyncroTech:
I'm getting an intermittent NullReferenceException from the ImageOpened event handler in the UWP ButtonRenderer UpdateContent() function.

The culprit appears to be this line:
Element.InvalidateMeasureNonVirtual(InvalidationTrigger.RendererReady);
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.UAP/ButtonRenderer.cs#L154

It seems that "Element" is getting set to null somewhere between the creation of my button and the ImageOpened event triggering.

An easy fix would be to capture the current value of Element in a temporary variable and use that in the lambda function.

There's a few other comments regarding the issue at these links, but I couldn't find any other direct reports of this bug

Forum › Xamarin.Forms › Xamarin.Forms Releases
Xamarin.Forms 2.3.4.270
https://forums.xamarin.com/discussion/comment/264585/#Comment_264585

alexrainman/CarouselView UWP: crash on window resize #199
https://github.com/alexrainman/CarouselView/issues/199#issuecomment-322115694

rmarinho:
Hi, is it possible to provide a small sample so we can try to reproduce the issue. ?

Thanks

jalbertSyncroTech:
Yes, but it won't be until after the new year and the issue is timing related so it may not be reproducible.

What I'm doing is using a CarouselPage as the MainPage of my application, and then adding all of the pages used in my application to the CarouselPage's children in order to "pre-render" them.  I then swap the MainPage to the real page.  The NullReferenceException happens at this point, but if I put a small Task.Delay (currently 500ms) after "pre-loading" the pages I don't see the issue anymore.

Either way, I think the existing code is incorrect as it's assuming in an asynchronous callback that the Element attached to the renderer hasn't changed between when the image load was requested and when the load finished.

kingces95:
@jalbertSyncroTech Happy New Year! Please attach a reproduction. An intermittent reproduction is sufficient so long as it eventually crashes.

jalbertSyncroTech:
Happy New Year to you too! zip with reproducible case is attached.

[Issue1423.zip](https://github.com/xamarin/Xamarin.Forms/files/1605026/Issue1423.zip)

jalbertSyncroTech:
[...]

PR #1576, opened 2018-01-13
Fix UWP button image rendering crash

breyed:
On UWP, if you click around an app quickly, you can get it crash with this error:
> Xamarin.Forms.Platform.UAP.dll!Xamarin.Forms.Platform.UWP.ButtonRenderer.UpdateContent.AnonymousMethod__0(object sender, Windows.UI.Xaml.RoutedEventArgs args)

Since image loading is deferred, if you navigate quickly (or if you load the image from a slow source), the Xamarin.Forms element may no longer be present when `ImageOpened` is raised. The null check in this PR should fix that.

(I made the change in the GitHub web editor - since it's only one character. I haven't tested. The editor caused the newline change at the bottom of the file. No need to keep that change obviously.)

### DIFF
--- a/Xamarin.Forms.Platform.UAP/ButtonRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/ButtonRenderer.cs
@@ -171,7 +171,7 @@ namespace Xamarin.Forms.Platform.UWP
 			bmp.ImageOpened += (sender, args) => {
 				image.Width = bmp.PixelWidth;
 				image.Height = bmp.PixelHeight;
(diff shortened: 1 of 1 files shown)


jalbertSyncroTech:
@breyed I'm not sure this is the right fix - Element may be null, but it may also be referring to a completely different object from when the image was originally assigned.  An better fix would be to capture the current value of Element in a temporary variable and use that in the lambda function.

I have another bug still open on this issue here:
Xamarin.Forms UWP ButtonRenderer UpdateContent() ImageOpened event NullReferenceException #1423
https://github.com/xamarin/Xamarin.Forms/issues/1423

PR #1578, opened 2018-01-14
[Core] Conditional renderers based on OS version

adrianknight89:
### Description of Change ###

Redoing #619.


Xamarin.Forms.Platform.Android/ExportRendererAttribute.cs
```diff
@@ -1,12 +1,15 @@
...
 
 namespace Xamarin.Forms
 {
 	[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
-	public sealed class ExportRendererAttribute : HandlerAttribute
+	public sealed class ExportRendererAttribute : BaseExportRendererAttribute
 	{
 		public ExportRendererAttribute(Type handler, Type target) : base(handler, target)
 		{
+			MajorVersion = (int)Build.VERSION.SdkInt;
```
adrianknight89: (Just realized) For Android, I opted in for SDK version (e.g. 26) as opposed to OS naming version (8 for Oreo). Let me know if this needs to be the other way around.

Xamarin.Forms.Platform.iOS/ExportRendererAttribute.cs
```diff
@@ -1,16 +1,19 @@
...
 
 namespace Xamarin.Forms
 {
 	[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
-	public sealed class ExportRendererAttribute : HandlerAttribute
+	public sealed class ExportRendererAttribute : BaseExportRendererAttribute
 	{
 #if __MOBILE__
-		public ExportRendererAttribute(Type handler, Type target, UIUserInterfaceIdiom idiom) : base(handler, target)
+		public ExportRendererAttribute(Type handler, Type target, UIUserInterfaceIdiom idiom) : this(handler, target)
```
adrianknight89: Call the other constructor to avoid writing the same line of MajorVersion code twice.

Xamarin.Forms.Platform.Tizen/ExportRendererAttribute.cs
```diff
@@ -1,12 +1,16 @@
 using System;
+using Xamarin.Forms.Core;
+using Information = Tizen.System.Information;
```
adrianknight89: Not sure why, but `Tizen` isn't available for me (even after installing the SDK) unless I short circuit class path like this.

Xamarin.Forms.Platform.Tizen/ExportRendererAttribute.cs
```diff
@@ -1,12 +1,16 @@
...
 
 namespace Xamarin.Forms.Platform.Tizen
 {
 	[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
-	public sealed class ExportRendererAttribute : HandlerAttribute
+	public sealed class ExportRendererAttribute : BaseExportRendererAttribute
 	{
 		public ExportRendererAttribute(Type handler, Type target) : base(handler, target)
 		{
+			if(Information.TryGetValue("tizen.org/feature/platform.version", out string OSVersion))
```
adrianknight89: Since `TryGetValue` is being used to get various kinds of information and the result might possibly fail, should we throw an exception here in that case (i.e. if block isn't executed)? Default for `MajorVersion` is 0.

Xamarin.Forms.Platform.Tizen/ExportRendererAttribute.cs
```diff
@@ -1,12 +1,16 @@
...
+				MajorVersion = Convert.ToInt32(OSVersion.Split('.')[0]);
```
adrianknight89: Here, I assumed version parts would be separated by periods like on other platforms. Not sure if that's the case on Tizen.

Xamarin.Forms.Platform.Tizen/ExportRendererAttribute.cs
rookiejava: API version would be better than platform version in this case. So, it could be `Tizen.Common.DotnetUtil.TizenAPIVersion`.  ([API Ref.](https://developer.tizen.org/dev-guide/csapi/api/Tizen.Common.DotnetUtil.html)).

Xamarin.Forms.Platform.Tizen/ExportRendererAttribute.cs
rookiejava: Yes, platform version is separated by periods something like 4.0.0 on Tizen. However, as I mentioned above, API version is much clear. The API Version in Tizen is almost the same as the API Level in Android. The current/latest version of Tizen API is `4`.

`MajorVersion = Tizen.Common.DotnetUtil.TizenAPIVersion`

Xamarin.Forms.Platform.Tizen/ExportRendererAttribute.cs
adrianknight89: Thanks. I can use API version.

myroot:
I'm unhappy to consume the computing time at launching timing.
It make degression on launching time

How about to delay getting of platform version and caching on  ShouldRegister method.
and If not set the MininumVersion and MaxiumVersion, It can skip the test.

Xamarin.Forms.Core/BaseExportRendererAttribute.cs
```diff
@@ -0,0 +1,58 @@
...
+		#endregion
+
+		protected BaseExportRendererAttribute(Type handler, Type target) : base(handler, target)
+		{
+		}
+
+		public override bool ShouldRegister()
+		{
+			if (MinimumSdkVersion <= MajorVersion && MajorVersion <= MaximumSdkVersion)
+				return base.ShouldRegister();
```
myroot: I think, it does not need to base call
I think, it does not need to base call. . 
Because, 
base class can't change. 
and HandlerAttribute.ShouldRegister always returning true.

adrianknight89: Why do you assume base class can't change? While it is unlikely that it will change, if it happens for whatever reason, then not calling base here would ignore base completely. `ShouldRegister` should perhaps be abstract (or an interface method) and sub-classes should implement it. Right now, base `ShouldRegister` is not doing anything other than returning a boolean result, which is mostly useless, so abstracting it out could be a better choice. What does the team think?

adrianknight89:
I agree that we should cache platform versions and not look them up with each registration (though I'm not sure if looking up the versions is expensive since they are probably cached by underlying platforms already). Perhaps the versions can be lazy singletons?

adrianknight89:
@team, can you recommend a place to store `MajorVersion` that would be initialized once? Every platform seems to be implementing `Forms` and `Platform` classes. Not sure which one would be a better choice.

myroot:
How about change to option variable for MaximumSdkVersion and MiniumSdkVersion?
If did't passing MaximumSdkVersion and MiniumSdkVersion, should nothing be happen. (including getting Platform version)

because, most renderers do not use this feature.

adrianknight89:
CLR restricts what kind of types can be attribute parameters. As far as I know, nullable types are not allowed, hence the primitive integers. I set them to `int.MaxValue` and `int.MinValue` so if users don't need conditional renderers, these won't introduce a breaking change.

https://stackoverflow.com/a/3192882

I removed `MajorVersion` and created an abstract method implemented by each platform that is used only when needed.

myroot:
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/nullable-types/
We can use nullable for value type

Xamarin.Forms.Core/BaseExportRendererAttribute.cs
```diff
@@ -0,0 +1,65 @@
...
+			// Conditional registration is being used
+			var majorVersion = GetMajorVersion();
+
+			if (MinimumSdkVersion <= majorVersion && majorVersion <= MaximumSdkVersion)
+				return base.ShouldRegister();
+
+			return false;
+		}
+
+		public abstract int GetMajorVersion();
```
myroot: How about get only property? It can be also overridden in sub class.

[Choosing Between Properties and Methods](https://msdn.microsoft.com/en-us/library/ms229054(v=vs.100).aspx)

adrianknight89:
I meant nullables for attributes. I don't think the `new` operator (as in `new Nullable<T>`) is allowed. I also don't think nullable types with `?` is allowed as attributes.

So, here, we can't use a nullable when registering a custom renderer:

`[assembly: ExportRenderer(typeof(Bugzilla59457.Bugzilla59457Entry), typeof(_59457CustomRenderer))]`

Again, I might be wrong, but that's what I've assumed for a while.

myroot:
Ok, I now understand, nullable type is invalid for attributes arguments

Xamarin.Forms.Core/BaseExportRendererAttribute.cs
adrianknight89: I'll get back to this later. I think the property would need to be abstract in the base since we'll need polymorphic behavior in the base `ShouldRegister` method.

rmarinho:
@adrianknight89  needs docs update

adrianknight89:
@rmarinho Can you review now?

Xamarin.Forms.Core.UnitTests/Xamarin.Forms.Core.UnitTests.csproj
```diff
@@ -218,7 +219,5 @@
       <LogicalName>Images/crimson.jpg</LogicalName>
     </EmbeddedResource>
   </ItemGroup>
-  <ItemGroup>
-    <Folder Include="StyleSheets\" />
-  </ItemGroup>
-</Project>
```
adrianknight89: @rmarinho  I think this needs to be rebased to include this folder, no?

rmarinho: uhh nice catch. yes please

adrianknight89:
Done.

rmarinho:
build

adrianknight89:
I think `_majorVersion` needs to be moved to the derived classes as a private field since keeping it in base and making it internal at the same time disallows third-party export renderers from setting it. If we do not make it internal, then both `_majorVersion` and `MajorVersion` are exposed which can be confusing and in general only properties should be exposed. The point of `_majorVersion` was to make sure the APIs to get versions are hit only once.

adrianknight89:
Could be reviewed now.

adrianknight89:
Updated docs. Could be reviewed now.

docs/Xamarin.Forms.Core/ns-Xamarin.Forms.Core.xml
```diff
@@ -0,0 +1,6 @@
+<Namespace Name="Xamarin.Forms.Core">
```
rmarinho: i think we can drop this

Xamarin.Forms.Core/BaseExportRendererAttribute.cs
```diff
@@ -0,0 +1,67 @@
...
+
+				if (_minimumSdkVersion > MaximumSdkVersion)
+					throw new ArgumentException("Minimum SDK version must be less than or equal to the maximum SDK version.");
+			}
+		}
+		#endregion
+
+		#region MaximumSdkVersion
+		int _maximumSdkVersion = int.MaxValue;
+		public int MaximumSdkVersion
```
rmarinho: I wonder if we should drop the SDK from here, @jassmith ?

adrianknight89: We should. Since this was a rewrite of an earlier PR, I did not pay attention to this detail. How about system version instead of sdk version? or api version?

docs/Xamarin.Forms.Core/ns-Xamarin.Forms.Core.xml
adrianknight89: You mean the file itself ('docs/Xamarin.Forms.Core/ns-Xamarin.Forms.Core.xml') or the namespace tag? It was auto-generated when I ran docs.

rmarinho: yeah the file itsef

adrianknight89:
@hartez Anything on the naming?

Xamarin.Forms.Core/BaseExportRendererAttribute.cs
hartez: Since "version" means different things on different platforms (OS version, SDK verison, API version), I think we should just go with "version. So `MaximumVersion` and `MinimumVersion`.

hartez:
My concern with this PR is that we're locking ourselves into major version checking on all platforms, when it may not apply. On iOS, for example, we might want to limit a renderer to a minimum version of 9.0 and a max of 9.3; I don't think this allows us to do that. 

Instead of abstracting "version" to a central concept like this, would it make sense to implement an override of ShouldRegister() on each platform, and add min/max version properties which are specific to each platform? That would allow us to express, say, Android version using the BuildVersionCodes enum which Xamarin.Android already provides for us. And we could express iOS versions as version strings (e.g. "11.0" or "9.3.5"), then parse them and use the CheckSystemVersion method which Xamarin.iOS provides (and maintains).

jassmith:
I agree with EZ's comment 100%

rmarinho:
build

rmarinho:
build

rmarinho:
build

rmarinho:
build

rmarinho:
build --uitests

PR #1594, opened 2018-01-17
[Xaml] let the previewer know the asm

StephaneDelcroix:
### Description of Change ###

in addition to the resourcePath, tell the previewer the assembly in
which we're looking for the resource.

it changes an API that's not supposed to be used (yet) but the previewer, but I'd like @alanmcgovern approval on this.

### Bugs Fixed ###

/

### API Changes ###

Changed:
 - `public static Func<AssemblyName, string, string> Xamarin.Forms.Internals.ResourceLoader.ResourceProvider { get; internal set; }`

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/XamlCTask.cs
+++ b/Xamarin.Forms.Build.Tasks/XamlCTask.cs
@@ -6,6 +6,8 @@ using System.Linq;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
 
+using static Mono.Cecil.Cil.OpCodes;
+
 using Xamarin.Forms.Xaml;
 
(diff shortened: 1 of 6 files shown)


alanmcgovern:
Yup, this should bring the API in line with the old one :) perfect! Thanks!

alanmcgovern:
We'll need to ensure this is rolled into the same release that the stylesheet support ships in too :)

alanmcgovern:
loaded the nupkg locally and it works just fine for me! :)

alanmcgovern:
For reference, here's a gif: http://recordit.co/Ar5ZTwPFgG

PR #1645, opened 2018-01-24
[C] fix and speed up binding to BP

StephaneDelcroix:
[C] fix and speed up binding to BP

That code path was no longer hit (name comparison was always false), so
I fixed it. The fix that gets the MethodInfo from the interface instead
of on the type speeds things up by a factor 6 (in my tests)

### DIFF
--- a/Xamarin.Forms.Core/BindingExpression.cs
+++ b/Xamarin.Forms.Core/BindingExpression.cs
@@ -23,9 +23,9 @@ namespace Xamarin.Forms
 		internal BindingExpression(BindingBase binding, string path)
 		{
 			if (binding == null)
-				throw new ArgumentNullException("binding");
+				throw new ArgumentNullException(nameof(binding));
 			if (path == null)
-				throw new ArgumentNullException("path");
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Core/BindingExpression.cs
```diff
@@ -341,19 +341,7 @@ void SetupPart(TypeInfo sourceType, BindingExpressionPart part)
...
-								{
-									ParameterInfo[] parameters = m.GetParameters();
-									if (parameters.Length == 2 && parameters[0].ParameterType == typeof(BindableProperty))
-									{
-										setValueMethod = m;
-										break;
-									}
-								}
-							}
+							MethodInfo setValueMethod = typeof(IElementController).GetMethod("SetValueFromRenderer", new[] { typeof(BindableProperty), typeof(object) });
```
campersau: Why not `nameof(IElementController.SetValueFromRenderer)`?

StephaneDelcroix: that could have been a `nameof()`, indeed

PR #1649, opened 2018-01-24
Add netstandard1.0 support

rmarinho:
### Description of Change ###

Multi-targeting Xamarin.Forms.Core, Xamarin.Forms.Xaml, Xamarin.Forms.Maps to netstandard 1.0 so PCL users can use that without upgrading to netstandard 2.0.

- Cleanup sln by removing old configurations that we don't use.
- Remove confuse script about docs.
- Update nuspec

### Bugs Fixed ###

- Provide links to bugs here

### API Changes ###

Adding back for Netstandard1.0 only 

```
Xamarin.Forms.Internals.FileMode; 
Xamarin.Forms.Internals.FileAccess; 
Xamarin.Forms.Internals.FileShare; 
```

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #1656, opened 2018-01-25
[XamlC] get the Assembly through the TypeInfo

StephaneDelcroix:
### Description of Change ###

[XamlC] get the Assembly through the TypeInfo

when an assembly object is required, XamlC was generating IL equivalent to the following c#
```
var asm = typeof(Foo).Assembly;
```

That API (the `Assembly` property) is not available in PCL profiles, but is always present at runtime. So there was no problem executing the code, but tools like CodeAnalysis were seriously confused.

The fix in this PR changes the generated code to the equivalent of
```
var asm = typeof(Foo).GetTypeInfo().Assembly;
```

Upgrading your projects from PCL to `netstandard2.0` would fix this as well.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=44130

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###


rmarinho:
I wonder if we might keep this around for #1649

StephaneDelcroix:
@rmarinho this will keep working if we merge #1649

PR #1680, opened 2018-01-26
[Android] Redo: Hide Fragments before removing them to allow custom animations

adrianknight89:
Redoing #661. 

Note that `TransitionDuration` must be overriden in a custom renderer to match durations in XML animation resource files so that pages disappear from the screen at the same time animations complete. I'm not sure if there is an API to extract durations from the resource files at runtime for us to sync `TransitionDuration` with those durations without having to set `TransitionDuration` manually. 

Also, another thing to note is it is possible for the page animation and the page removal timer to finish at different times even though they are expected to finish at about the same time since neither is probably guaranteed to finish in the given course of time. To guard for this, it might be a good idea to set `TransitionDuration` slightly higher than durations in resource files. I don't like this kind of coupling, but I can't think of another solution.

**Untested scenario:**

Though small, there is also the possibility that a developer wants different durations for different animations. Not sure if this PR provides support for that, but it might be possible to change `TransitionDuration` before pushes and pops to make this scenario work.

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
@@ -47,6 +47,7 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 		bool _toolbarVisible;
 
 		// The following is based on https://android.googlesource.com/platform/frameworks/support.git/+/4a7e12af4ec095c3a53bb8481d8d92f63157c3b7/v4/java/android/support/v4/app/FragmentManager.java#677
+		// Must be overriden in a custom renderer to match durations in XML animation resource files
 		protected virtual int TransitionDuration { get; set; } = 220;
 
 		public NavigationPageRenderer(Context context) : base(context)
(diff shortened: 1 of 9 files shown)


Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
```diff
@@ -604,7 +605,6 @@ void RemovePage(Page page)
 
 			// Go ahead and take care of the fragment bookkeeping for the page being removed
 			FragmentTransaction transaction = FragmentManager.BeginTransaction();
-			transaction.DisallowAddToBackStack();
```
adrianknight89: Don't think we need this anymore. See #919.

hartez: You are correct - we don't. Feel free to drop it from this PR. Or leave it in and we can address them all at once.

adrianknight89: let's leave it in.

PaulVrugt:
Hello,

maybe this is a stupid question. But is there anywhere I can lookup when this PR will be released in a normal forms release? It is merged in the master branch. Does this mean in will be release in the next Service release?

rmarinho:
@PaulVrugt you can get nuget packages to test from master by checking the status commit on master. If it's on master it will be releaed next time we branch for a Release, service releases are from the 15-5 branch.

PaulVrugt:
@rmarinho thanks for the info. So does that mean it will be in the 2.6 release?  is there any way to know when this new branch is made?

PR #1741, opened 2018-01-30
[Core] Add IsSet to BindableObject

### LINKED ISSUE #1652: Expose BindableObject.GetIsDefault
## Summary
`BindableObject.GetIsDefault` checks whether the value of a bindable object has been set and whether it is the default value, taking into consideration a Default Value Creator. This API is currently internal, so we can't use it in renderers. 

## API Changes

- `public bool BindableObject.IsSet` --> calls `BindableObject.GetIsDefault`

## Intended Use Case
It would be useful in cases such as #1570 where I need to check if the default value of the `Button.BorderRadius` property has changed.

samhouts:
### Description of Change ###

Added new method to `BindableObject` that will allow users to check whether a `BindableProperty` has ever been set (`IsSet()`).

### Bugs Fixed ###

 - fixes #1652 

### API Changes ###

Added:
```csharp
public bool BindableObject.IsSet(BindableProperty targetProperty);
```

### Behavioral Changes ###

None

### PR Checklist ###


Xamarin.Forms.Core/BindableObject.cs
```diff
@@ -47,6 +47,28 @@ public void ClearValue(BindablePropertyKey propertyKey)
...
+		{
+			if (targetProperty == null)
+				throw new ArgumentNullException(nameof(targetProperty));
+
+			return GetContext(targetProperty) != null;
+		}
+
+		public bool IsDefault(BindableProperty targetProperty)
+		{
+			if (targetProperty == null)
```
adrianknight89: Is the null check here needed since Calling `IsSet` is doing the same work? Right now, the argument is being checked for null twice.

StephaneDelcroix:
I'm not sure I understand the use case for `IsDefault()`

samhouts:
@StephaneDelcroix https://github.com/xamarin/Xamarin.Forms/pull/1570/commits/f5f74bcf8fc477e85ba664ec3e64b7ef9108a639#diff-cc10c976ac59dbaad236bd2cbe0d91e2R45 If the `BorderRadius` is the default value, then we'll use the default drawable. If it's anything other than the default drawable, we'll create a new drawable. The default value for `BorderRadius` is different for Android than other platforms, so we need to use a `DefaultValueCreator`. This helps us get all that data in one call.

Xamarin.Forms.Core/BindableObject.cs
```diff
@@ -47,6 +47,25 @@ public void ClearValue(BindablePropertyKey propertyKey)
...
+
+		public bool IsDefault(BindableProperty targetProperty)
+		{
+			if (!IsSet(targetProperty))
+				return true;
+
+			var defaultValue = targetProperty.GetDefaultValue(this);
+			var currentValue = GetValue(targetProperty);
+
+			return defaultValue == currentValue;
```
StephaneDelcroix: will unfortunately never work with defaultValueCreator returning reference types

PR #378, opened 2016-09-25
[Android] Slider should show user-set value on initial load

adrianknight89:
### Description of Change

Android `SeekBar` does not allow Min to be set and `Progress` is an int. This creates a problem when `Slider`'s Min and Max values can be double. For example, when Min is 1 and Max is 10, SeekBar's 1000 ticks need to be distributed over this range. When `Slider` loads initially with a value of 5, Progress will be calculated as an int which results in precision loss. When we read this int back and try to convert it to a double, then the user sees 4.996. To me, it looked like there is no way to fix this problem.

What I did was letting the user see whatever value they set in XF even if this value is slightly off (4.996 is the correct value whereas 5 would be incorrect). However, once they move the slider, then they should begin seeing floating point numbers.

Perhaps, the truly correct thing to do is to use something like a `ValueRequest` property and let the user know their requested value might not always come back the same on initial load.

Also made minor refactoring.

@hartez 
### Bugs Fixed
- https://bugzilla.xamarin.com/show_bug.cgi?id=43429

PR #1756, opened 2018-01-31
[Android] Slider should show correct Value even though SeekBar's Progress is slightly off

### LINKED ISSUE #1625: [Android] Slider value is not changed for the first position change
### Description
Slider value is not changed for the first time its position is changed and `ValueChanged` event is not fired either.
I've been able to confirm this issue on Android. This bug is not present on UWP. Untested on iOS.
This bug has been already reported to Bugzilla [here](https://bugzilla.xamarin.com/show_bug.cgi?id=54887) and the thread points out that it might be linked with commit e0165abbcd1329a3dfd353548f5d75eef0c792de. I decided to report this issue here, because it's an old regression which makes Slider component unusable in practice.
### Steps to Reproduce

1. Open test solution and run under Android
2. Tap (click) anywhere on the slider
3. You can see that the bound label text did not change

### Expected Behavior
Correct value change according to slider position after first tap.
### Actual Behavior
Value does not change after first tap, but slider position does.
### Basic Information
Applies to Android only

- Version with issue: 2.5.0.122203
- IDE:
- Platform Target Frameworks:
  - Android: 8.0
- Android Support Library Version: 25.4.0.2

### Screenshots

qemu-system-i386_2018-01-20_13-44-52

### Reproduction Link

[SliderTest.zip](https://github.com/xamarin/Xamarin.Forms/files/1648809/SliderTest.zip)

adrianknight89:
### Description of Change ###

#378 fixed an issue but also introduced regression. This fixes both #378 and #1625. When the element itself or its properties are changing from the Core side, SeekBar's tracker isn't being manipulated by the user, so we can ignore native changes being propagated back to Core. Please see instructions in the test.

Note that this introduces the problem that Core and Native aren't always in sync as there is a small precision difference. However, not adding this fix is more problematic because users see values that they did not expect.

### Bugs Fixed ###

- https://bugzilla.xamarin.com/show_bug.cgi?id=54887
- fixes #1625

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Github1625.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Github1625.cs
@@ -1,0 +1,45 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 1625, "Slider value is not changed for the first position change", PlatformAffected.Android)]
+	public class Github1625 : TestContentPage // or TestMasterDetailPage, etc ...
+	{
+		protected override void Init()
+		{
+			var slider = new Slider();
+			slider.Maximum = 10;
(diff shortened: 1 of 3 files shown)


DamianMehers:
This change breaks UI Automation of sliders on Android, if Value is databound, the bound value doesn't get updated when the slider is manipulated through  [App.SetSliderValue](https://developer.xamarin.com/api/member/Xamarin.UITest.IApp.SetSliderValue/p/System.String/System.Double/)

PR #1757, opened 2018-01-31
[Android] BackPressed preventing for an each event

rotorgames:
### Description of Change ###

I have [the library for xamarin forms](https://github.com/rotorgames/Rg.Plugins.Popup) and I should catch a back button event in my library and prevent it if it is needed. I have found the BackPressed events in FormsAppCompatActivity and FormsApplicationActivity and if I return `true` in these events the back button will not work. But it works only for the last delegate in the BackPressed event. I have fixed it. Each delegate will be invoked in the invocation list but if one in them returns true that the back button will not pop any pages.

I think that a lot of people would like to use this feature in their libraries.

I have not created any tests because it is simple changes.

I needed to created two CanGoBack methods in FormsAppCompatActivity and FormsApplicationActivity because there are two delegates for the BackPressed events. Maybe you will have better solution.

### Bugs Fixed ###

- 

### API Changes ###

-

Added:
 - private bool FormsAppCompatActivity.CanGoBack();
 - private bool FormsApplicationActivity.CanGoBack();

Changed:
 - void FormsAppCompatActivity.OnBackPressed();
 - void FormsApplicationActivity.OnBackPressed();

### Behavioral Changes ###

No changes

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
@@ -60,7 +60,7 @@ namespace Xamarin.Forms.Platform.Android
 
 		public override void OnBackPressed()
 		{
-			if (BackPressed != null && BackPressed(this, EventArgs.Empty))
+			if (!CanGoBack())
 				return;
 			base.OnBackPressed();
 		}
@@ -274,6 +274,22 @@ namespace Xamarin.Forms.Platform.Android
 			_currentState = AndroidApplicationLifecycleState.OnStop;
 
 			await OnStateChanged();
+		}
+
+		bool CanGoBack()
+		{
+			var backPressed = BackPressed;
+
+			if (backPressed == null)
+				return true;
+
+			var invocationList = backPressed.GetInvocationList().ToList();
+
+			var canGoBack = !invocationList
+				.Select(@delegate => ((BackButtonPressedEventHandler) @delegate).Invoke(this, EventArgs.Empty))
+				.Any(b => b);
+
+			return canGoBack;
 		}
 
 		void AppOnPropertyChanged(object sender, PropertyChangedEventArgs args)
--- a/Xamarin.Forms.Platform.Android/FormsApplicationActivity.cs
+++ b/Xamarin.Forms.Platform.Android/FormsApplicationActivity.cs
@@ -38,7 +38,7 @@ namespace Xamarin.Forms.Platform.Android
 
 		public override void OnBackPressed()
 		{
-			if (BackPressed != null && BackPressed(this, EventArgs.Empty))
+			if (!CanGoBack())
 				return;
 			base.OnBackPressed();
 		}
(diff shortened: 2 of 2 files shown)


Xamarin.Forms.Platform.Android/FormsApplicationActivity.cs
```diff
@@ -193,6 +193,22 @@ protected override void OnStop()
 			OnStateChanged();
 		}
 
+		bool CanGoBack()
+		{
+			var backPressed = BackPressed;
```
adrianknight89: Why do you need a local variable? Can't you use `BackPressed` directly?

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -276,6 +276,22 @@ protected override void OnNewIntent(Intent intent)
...
+			var backPressed = BackPressed;
+
+			if (backPressed == null)
+				return true;
+
+			var invocationList = backPressed.GetInvocationList().ToList();
+
+			var canGoBack = !invocationList
+				.Select(@delegate => ((BackButtonPressedEventHandler) @delegate).Invoke(this, EventArgs.Empty))
+				.Any(b => b);
```
adrianknight89: Doesn't this invoke `OnBackButtonPressed` on EACH page? I don't think this is a good idea as we don't know what kind of custom logic people have in this handler. Naturally, people should expect it to be invoked when its page is at the top of the stack. Invoking handlers deeper down the stack is erroneous in my opinion. Maybe you should create a platform-specific property that allows you to override this logic according to your own use case.

Xamarin.Forms.Platform.Android/FormsApplicationActivity.cs
rotorgames: I can and if it is needed I will change it.

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
rotorgames: @adrianknight89 No event will not be invoked for each pages. Xamarin.Forms uses it only one times. Event will be invoked only for the last page in the stack how it did before.

Xamarin.Forms.Platform.Android/FormsApplicationActivity.cs
adrianknight89: In general, I avoid allocating memory for local variables if I can directly use an existing one. In this scenario, I'm not sure what the compiler does in terms of optimization.

rotorgames: @adrianknight89 This local variable will be released in the end of method with the help of garbage collector.

jassmith:
Looks like some code wasn't properly included.

rotorgames:
@jassmith I have solved my problem without this solution. Thanks.

PR #1761, opened 2018-01-31
[Core] AbsoluteLayout content now fills all available space on Android devices

PureWeen:
### Description of Change ###

Changed AbsoluteLayout.cs to use a platform specific DisplayRound method to allow for cases where a straight Math.Round won't correctly round to the pixel density of the device.  Currently this is only overridden in Android and all the other platforms still just use Math.Round

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=40092
fixes #1128

### API Changes ###

Added:
 - public override double DisplayRound(double value)

Platforms can now customize how rounding works for calculations that pertain to pixels.

### Behavioral Changes ###

None

### PR Checklist ###

- [X ] Has tests (if omitted, state reason in description)
- [ X] Rebased on top of master at time of PR
- [ X] Changes adhere to coding standard
- [ X] Consolidate commits as makes sense

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla40092.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla40092.cs
@@ -1,0 +1,42 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System;
+using System.Threading.Tasks;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+    [Preserve(AllMembers = true)]
+    [Issue(IssueTracker.Bugzilla, 40092, "Ensure android devices with fractional scale factors (3.5) don't have a white line around the border", PlatformAffected.Android)]
+    public class Bugzilla40092 : TestContentPage
+    {
+        protected override void Init()
+        {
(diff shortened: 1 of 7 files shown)


HEFRNicolas:
This problem also appears on iOS using an iPad Air 2 in landscape.
abl_px
I fixed it in a similar way by using a custom AbsoluteLayout and using a DependencyService to do the DeviceRounding.

rmarinho:
@PureWeen do you mind recheck the issue on iOS see if we need to implemente there the same?! thanks

jassmith:
The iPad Air 2 has a scale factor of 2.0. So rounding there should be to half pixels.

PureWeen:
@HEFRNicolas   do you have a sample project I can test?  I ran our test case for this against an iPad air  2 simulator and I don't see any pixel lines on their. It looks like it fills the space without any issues

HEFRNicolas:
@PureWeen 
To reproduce the problem, you will need a navigation bar (just noticed it while doing the sample project) and you might not see the 1 px on a simulator.

sample project: https://github.com/HEFRNicolas/AbsoluteLayoutTest.git

jassmith:
@PureWeen hes absolutely right, integer rounding is wrong here, also for UWP though getting that scaling factor right is going to be uber fun by comparison...

PureWeen:
Thank you for the sample @HEFRNicolas 
I can see the thin red line underneath the box on the ipad air 2 simulator

D4rkC1own187:
This bug still appears on Android API 18. 
Xamarin.Forms: 3.3.0.91240 
Android Library: 27.0.2

PR #1780, opened 2018-02-02
[Android] TimePicker Format should work

### LINKED ISSUE #1461: Cannot change TimePicker format

### Steps to Reproduce
1. Turn the "Use 24-hour format" device setting off on your test device
2. Create a TimePicker
3. Change the format sometime later on (e.g. when the user clicks a button)

### Expected Behavior
TimePicker displays time in new format

### Actual Behavior
TimePicker still displays time in original format

### Basic Information

- Version with issue: Xamarin.Forms 2.5.0.91635

### Proposed Fix
I can see the problem in Xamarin.Forms.Platform.Android.TimePickerRenderer. The format is cached in OnElementChanged then never gets updated. Presumably, the format just needs to be refreshed in OnElementPropertyChanged if FormatProperty has changed.

adrianknight89:
### Description of Change ###

Please see comments as well as instructions in the UI test.

### Bugs Fixed ###

- fixes #1461 

### PR Checklist ###


Xamarin.Forms.Platform.Android/Renderers/TimePickerRenderer.cs
```diff
@@ -2,24 +2,24 @@
...
 	public class TimePickerRenderer : ViewRenderer<TimePicker, EditText>, TimePickerDialog.IOnTimeSetListener
 	{
 		AlertDialog _dialog;
 		TextColorSwitcher _textColorSwitcher;
-		bool _is24HourFormat;
-		string _timeFormat;
+
+		bool Is24HourView
+		{
+			get => (DateFormat.Is24HourFormat(Context) && Element.Format == (string)TimePicker.FormatProperty.DefaultValue) || Element.Format == "HH:mm";
```
adrianknight89: This (now renamed properly and set to a private property) was being used to determine whether the time picker dialog should open up in 24h format or not. Previously, the logic checked system settings only, ignoring user settings. Should Forms make sure system settings override developer settings or vice versa? I assume Google is okay with code overriding system settings because, after all, why does the `TimePickerDialog` API ask for a boolean for the hour format?

Suppose user code is setting format to "HH:mm" but 24h setting is set to off by the device. What's the expectation here? The changes I made allow users to bypass system settings. If the default value, `t`, is used, then we rely exclusively on system settings.

Line 21 is checking two things: default scenario where `Format` is not set and a custom scenario.

Xamarin.Forms.Platform.Android/Renderers/TimePickerRenderer.cs
```diff
@@ -129,7 +125,8 @@ void OnCancelButtonClicked(object sender, EventArgs e)
 
 		void SetTime(TimeSpan time)
 		{
-			Control.Text = DateTime.Today.Add(time).ToString(_timeFormat);
+			var timeFormat = Is24HourView ? "HH:mm" : Element.Format;
+			Control.Text = DateTime.Today.Add(time).ToString(timeFormat);
```
adrianknight89: Instead of caching `timeFormat`, calculate it every time `SetTime` is called.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Xamarin.Forms.Controls.Issues.Shared.projitems
```diff
@@ -257,6 +257,8 @@
     <Compile Include="$(MSBuildThisFileDirectory)Effects.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)FailImageSource.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)GestureBubblingTests.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)Github1461.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)InputTransparentInheritance.cs" />
```
adrianknight89: @rmarinho  `InputTransparentInheritance` shouldn't have been part of the PR. I guess it came with the merge you did?

PR #1791, opened 2018-02-04
[Core] Small refactoring on LogicalChildrenInternal

### LINKED ISSUE #1591: Should the local variable be used instead of the property
A local variable was created for the LogicalChildrenInternal however in line 318 the property is being used instead of the local variable?

https://github.com/xamarin/Xamarin.Forms/blob/f9b27a2edfa2126824f4b41e2712d283eaad50b4/Xamarin.Forms.Core/Layout.cs#L314-L318

I just game a cross this code when looking after multi-thread issue that we are facing in our code.

adrianknight89:
### Bugs Fixed ###

- fixes #1591

### DIFF
--- a/Xamarin.Forms.Core/ContentPresenter.cs
+++ b/Xamarin.Forms.Core/ContentPresenter.cs
@@ -21,9 +21,7 @@ namespace Xamarin.Forms
 		{
 			for (var i = 0; i < LogicalChildrenInternal.Count; i++)
 			{
-				Element element = LogicalChildrenInternal[i];
-				var child = element as View;
-				if (child != null)
+				if (LogicalChildrenInternal[i] is View child)
 					LayoutChildIntoBoundingRegion(child, new Rectangle(x, y, width, height));
 			}
 		}
--- a/Xamarin.Forms.Core/Layout.cs
+++ b/Xamarin.Forms.Core/Layout.cs
@@ -320,9 +320,7 @@ namespace Xamarin.Forms
 
 		internal virtual void OnChildMeasureInvalidated(VisualElement child, InvalidationTrigger trigger)
 		{
-			ReadOnlyCollection<Element> children = LogicalChildrenInternal;
-			int count = children.Count;
-			for (var index = 0; index < count; index++)
+			for (var index = 0; index < LogicalChildrenInternal.Count; index++)
 			{
 				var v = LogicalChildrenInternal[index] as VisualElement;
 				if (v != null && v.IsVisible && (!v.IsPlatformEnabled || !v.IsNativeStateConsistent))
--- a/Xamarin.Forms.Core/TemplatedView.cs
+++ b/Xamarin.Forms.Core/TemplatedView.cs
@@ -20,9 +20,7 @@ namespace Xamarin.Forms
 		{
 			for (var i = 0; i < LogicalChildrenInternal.Count; i++)
 			{
-				Element element = LogicalChildrenInternal[i];
-				var child = element as View;
-				if (child != null)
+				if (LogicalChildrenInternal[i] is View child)
 					LayoutChildIntoBoundingRegion(child, new Rectangle(x, y, width, height));
 			}
 		}


Xamarin.Forms.Core/ContentPresenter.cs
```diff
@@ -21,9 +21,7 @@ protected override void LayoutChildren(double x, double y, double width, double
 		{
 			for (var i = 0; i < LogicalChildrenInternal.Count; i++)
 			{
-				Element element = LogicalChildrenInternal[i];
```
adrianknight89: This and `TemplatedView.cs` isn't doing anything useful with `element` so why not get `child` directly? Also, using C# 7 pattern matching now.

Xamarin.Forms.Core/Layout.cs
```diff
@@ -320,9 +320,7 @@ internal static void LayoutChildIntoBoundingRegion(View child, Rectangle region,
 
 		internal virtual void OnChildMeasureInvalidated(VisualElement child, InvalidationTrigger trigger)
 		{
-			ReadOnlyCollection<Element> children = LogicalChildrenInternal;
-			int count = children.Count;
-			for (var index = 0; index < count; index++)
```
adrianknight89: There are many cases in the source code where `LogicalChildrenInternal` is being used directly instead of being converted to a local variable. Get count directly from the property.

Xamarin.Forms.Core/Layout.cs
```diff
@@ -320,9 +320,7 @@ internal static void LayoutChildIntoBoundingRegion(View child, Rectangle region,
...
+			for (var index = 0; index < LogicalChildrenInternal.Count; index++)
```
StephaneDelcroix: by changing this, `LogicalChildrenInternal.Count` is evaluated at each iteration. please revert this part

adrianknight89: I've noticed that part, but there are many cases `LogicalChildrenInternal.Count` is being used in a for loop in the source code.

https://github.com/xamarin/Xamarin.Forms/blob/210df5f775b2aa682b2c823f1d854d8290e46e27/Xamarin.Forms.Core/StackLayout.cs#L62

https://github.com/xamarin/Xamarin.Forms/blob/49a937fc5c04d6f80d902217c4883d1370c72a4b/Xamarin.Forms.Core/Layout.cs#L260

https://github.com/xamarin/Xamarin.Forms/blob/28b343b2788fa073a3d3a87610275d8f3e756b61/Xamarin.Forms.Core/TemplatedView.cs#L21

https://github.com/xamarin/Xamarin.Forms/blob/5f2796e160f99bad31c1b06a2d1f328b5ed02ba3/Xamarin.Forms.Core/ContentPresenter.cs#L22

https://github.com/xamarin/Xamarin.Forms/blob/485b3450496c91dd7875bbe163a234ccb4c51df8/Xamarin.Forms.Core/Element.cs#L344

Do you still need this change?

hartez: Since `LogicalChildrenInternal` is a `ReadOnlyCollection<Element>`, isn't the `Count` property a fixed `int`? Unless I'm missing something, I don't think there's any re-evaluation happening.

adrianknight89: I was looking up `ReadOnlyCollection<T>` last night. It doesn't seem all that intuitive, but based on what I read, this is a wrapper around an actual list/collection and doesn't allow changes to it. However, changes can still be made to the underlying type (by directly working with it) and they should be reflected back to `ReadOnlyCollection<T>`.

`InternalChildren` is an `ObservableCollection<Element>` and can change.

If the loop needs to read the count property only once, shouldn't all other loops be implemented the same way? Why were they left to read it continuously?

hartez: > shouldn't all other loops be implemented the same way? Why were they left to read it continuously?

Don't discount the possibility that all those other loops are just wrong :)

I thought Stephane was referring to an issue like IEnumerable.Count(). But if the worry is the value of Count changing because the underlying collection changes, then it might make sense to cache it before the loop. Or perhaps he's worried about the speed of accessing the member during the loop.

adrianknight89: I actually prefer counts being cached for all those loops. Since a changing `Count` requires more thought process / planning, I **assume** when the loops were being written, people didn't think much of it, but I agree that Stephane should explain why he prefers it cached.

StephaneDelcroix:
I'm sorry, but the time spent in discussion overweights the potential win of the proposed change by a factor +inf.
I was ready to accept a fix for the original issue, not ready to think about the way we iterate on our logicalchildren.

Thanks again for trying to fix this and make the reporter of #1591 happy.

adrianknight89:
Shouldn't there be a ticket to open a discussion on `LogicalChildrenInternal'?

PR #1808, opened 2018-02-06
[iOS] Fix object disposed exception in ImageCell

kicsiede:
by the time the async image load completes the imagecell could be disposed resulting in an unhandled exception (NSObjectDisposed)

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Cells/ImageCellRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Cells/ImageCellRenderer.cs
@@ -59,7 +59,7 @@ namespace Xamarin.Forms.Platform.iOS
 						target.SetNeedsLayout();
 					}
 					else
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.iOS/Cells/ImageCellRenderer.cs
```diff
@@ -53,12 +53,17 @@ protected override void HandlePropertyChanged(object sender, PropertyChangedEven
...
 				{
-					target.ImageView.Image = uiimage;
-					target.SetNeedsLayout();
+					if (target.Cell != null)
+					{
+						target.ImageView.Image = uiimage;
+						target.SetNeedsLayout();
+					}
+					else
+						uiimage.Dispose();
```
campersau: `uiimage` can be `null` if a `TaskCanceledException` was thrown.

PR #1822, opened 2018-02-06
[macOS] TapGesture assigned on a parent blocks children input (Picker…

### LINKED ISSUE #1821: [macOS] TapGesture assigned on a parent blocks children input controls (Picker) from receiving input.

### Description
Assigning a TapGesture to a parent layout will cause the children to not receive input from the mouse.  You can for instance drag on a picker's button but it will not open normally by clicking.

The idea is to probe it instead of blocking it.

### Steps to Reproduce

1. Add Grid
2. Assign TapGesture of one to Grid.
3. Place multiple controls inside including a picker with data.
4. Click on the drop down button.  It wont receive the click.

### Expected Behavior
All input controls should respond to the click.

### Actual Behavior
Some input controls will ignore the click.

### Basic Information

- Last known good version: None.
- IDE: Visual Studio for Mac.
- Platform Target Frameworks: macOS

BradChase2011:
…) from receiving input.

### Description of Change ###

Assigning a TapGesture to a parent layout will cause the children to not receive input from the mouse. You can for instance drag on a picker's button but it will not open normally by clicking.

The idea is to probe it instead of blocking it.

### Bugs Fixed ###

fixes #1821

### API Changes ###

None.

### Behavioral Changes ###

None.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/EventTracker.cs
+++ b/Xamarin.Forms.Platform.iOS/EventTracker.cs
@@ -95,6 +95,8 @@ namespace Xamarin.Forms.Platform.MacOS
 			var weakRecognizer = new WeakReference(recognizer);
 			var weakEventTracker = new WeakReference(this);
 
+			var tapRecognizer = recognizer as TapGestureRecognizer;
+
 #if !__MOBILE__
 			var clickRecognizer = recognizer as ClickGestureRecognizer;
 			if (clickRecognizer != null)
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.iOS/EventTracker.cs
```diff
@@ -279,22 +301,62 @@ UITapGestureRecognizer CreateTapRecognizer(int numTaps, Action action, int numFi
 				ShouldRecognizeSimultaneously = ShouldRecognizeTapsTogether
 			};
 
+			return result;
+		}		
+#else
+		NativeGestureRecognizer CreateClickRecognizer(int buttonMask, int numberOfClicksRequired, Action returnAction)
```
rmarinho: why are we moving this code?

BradChase2011: I think its more to do with the ShouldRecognizeTapsTogether pushes it down.  It used to be in the #if __MOBILE__ else statement but now is used for the macOS side.

PR #1824, opened 2018-02-07
[UWP] Add SearchBar platform specific for IsSpellCheckEnabled.

### LINKED ISSUE #1691: [Enhancement]Add SearchBar platform specific for IsSpellCheckEnabled
## Rationale

There's no way to set IsSpellCheckEnabled  on the AutoSugestBox, so we need to change the control template, and also expose a new platform specific to change this property inside the control template

## Implementation

We need to create a control template

Allow to enable SpellCheck  for SearchBar
```csharp

SearchBar.On<UWP>().IsSpellCheckEnabled();
SearchBar.On<UWP>().SetSpellCheckEnabled(true/false);
SearchBar.On<UWP>().EnableSpellCheck();
SearchBar.On<UWP>().DisableSpellCheck();

public static readonly BindableProperty IsSpellCheckEnabled Property = BindableProperty.Create("IsSpellCheckEnabled ", typeof(bool), typeof(SearchBar), false);

```

### Android
None

### iOS
None

### UWP
Alter the default control template for AutoSuggestBox and set the property IsSpellCheckEnabled on TextBox 

## Implications for CSS

None

## Backward Compatibility

Third party renderers may need to be updated to ensure that this functionality is supported through the new official mechanism. Further we will need to be careful to code the changes to the renderers in a careful manner to ensure that if someone is already using an effect support this feature that the effect is as best as possible not broken by our changes.

lateralusX:
fixes #1691.

### Description of Change ###

Add support for platform specific property to enable/disable spell checking on UWP SearchBar control.

### API Changes ###

NOTE, added API only apply to UWP.

Added:
 - New bindable property on `Xamarin.Forms.PlatformConfiguration.WindowsSpecific.SearchBar`, "IsSpellCheckEnabled".
 - `SearchBar.On<Windows>().IsSpellCheckEnabled();`
 - `SearchBar.On<Windows>().SetSpellCheckEnabled(true/false);`
 - `SearchBar.On<Windows>().EnableSpellCheck();`
 - `SearchBar.On<Windows>().DisableSpellCheck();`

### Behavioral Changes ###

According to #1691 the default for spell check enabled is set to false. This change set the same default for AutoSuggestBox control template meaning that the native UWP TextBox used by AutoSuggestBox will get a default value of false. This is inline with Win8 defaults of TextBox, but on Win10 the default value for IsSpellCheckEnabled on TextBox is true, see https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.textbox#Windows_UI_Xaml_Controls_TextBox_IsSpellCheckEnabled. If we have customers relying on the default behavior on Win10 then this fix will disable the spell check by default changing that behavior. In order to get back spell checking on a SearchBox instance, user needs to call one of:

```
SearchBar.On<Windows>().SetSpellCheckEnabled(true);
SearchBar.On<Windows>().EnableSpellCheck();
```

### PR Checklist ###


jassmith:
Can you rebase so we can merge?

PR #1827, opened 2018-02-07
[iOS] Fix: disable touch input when user interaction is disabled

kicsiede:
### Description of Change ###

The fix will not generate touch event for TextCell if UserInteractionEnabled is false (due to TextCell.Command.CanExecute returing false) and caching strategy is set RecycleElement / RecycleElementAndDataTemplate.

### Bugs Fixed ###

[https://bugzilla.xamarin.com/show_bug.cgi?id=60045](https://bugzilla.xamarin.com/show_bug.cgi?id=60045)

### API Changes ###

no

### Behavioral Changes ###

Will not generate touch event if UserInteractionEnabled is false.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/ContextActionCell.cs
+++ b/Xamarin.Forms.Platform.iOS/ContextActionCell.cs
@@ -659,7 +659,7 @@ namespace Xamarin.Forms.Platform.iOS
 
 					var cell = table.CellAt(_lastPath) as ContextActionsCell;
 
-					return cell != null;
+					return cell != null && cell.ContentCell.UserInteractionEnabled;
 				};
 			}
 


rmarinho:
I think this might not be the correct solution to fix this, This will mess with the others cells. specially with viewcell

hartez:
This problem is that TextCell still executes the command even though CanExecute returns false. The issue is also present on UWP. Also, I think Rui is right - this might cause some problems for some uses of ViewCell.

Proposing #1889 as an alternative - handle this problem directly in TextCell so it's fixed on all platforms.

jassmith:
Closing as #1889 was merged

PR #1828, opened 2018-02-07
[C] BO.IsSet() returns false after ClearValue()

### LINKED ISSUE #1813: `BindableObject.ClearValue()` does not clear the context of the BindableProperty
### Description

`IsSet()` returns `true` after `ClearValue()` is called on a `BindableProperty`.

### Steps to Reproduce

Unit test can be found at https://github.com/xamarin/Xamarin.Forms/pull/1570/commits/f27e18771cdea0c204d898b90903c2e5eb71fb5b (#1570)

### Expected Behavior

`IsSet()` returns `false` after `ClearValue()` is called on a `BindableProperty`.

### Actual Behavior

`IsSet()` returns `true` after `ClearValue()` is called on a `BindableProperty`.

### Reproduction Link

Unit test can be found at https://github.com/xamarin/Xamarin.Forms/pull/1570/commits/f27e18771cdea0c204d898b90903c2e5eb71fb5b

StephaneDelcroix:
### Description of Change ###

[C] BO.IsSet() returns false after ClearValue()

### Bugs Fixed ###

 - fixes #1813

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/BindableObjectUnitTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/BindableObjectUnitTests.cs
@@ -891,7 +891,7 @@ namespace Xamarin.Forms.Core.UnitTests
 		}
 
 		[Test]
-		public void IsSetIsTrueWhenPropCleared()
+		public void IsSetIsFalseWhenPropCleared()
 		{
 			string defaultValue = "default";
 			string newValue = "new value";
@@ -900,11 +900,9 @@ namespace Xamarin.Forms.Core.UnitTests
 			var bindable = new MockBindable();
 
 			bindable.SetValue(bindableProperty, newValue);
-
 			bindable.ClearValue(bindableProperty);
 
-			var isSet = bindable.IsSet(bindableProperty);
-			Assert.IsTrue(isSet);
+			Assert.That(bindable.IsSet(bindableProperty), Is.False);
 		}
 
 		[Test]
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Core/BindableObject.cs
```diff
@@ -57,7 +57,9 @@ public bool IsSet(BindableProperty targetProperty)
 			if (targetProperty == null)
 				throw new ArgumentNullException(nameof(targetProperty));
 
-			return GetContext(targetProperty) != null;
```
StephaneDelcroix: I think, with the modified behavior, that `IsDefault()` was a better name. It's not too late to make that change...

Xamarin.Forms.Core.UnitTests/BindableObjectUnitTests.cs
```diff
@@ -923,7 +938,7 @@ public void IsSetIsTrueWhenPropSetByDefaultValueCreator()
 			Assert.AreEqual(defaultValueC, created);
 
 			var isSet = bindable.IsSet(bindableProperty);
-			Assert.IsTrue(isSet);
+			Assert.IsFalse(isSet);
```
samhouts: Yay!

Xamarin.Forms.Core/BindableObject.cs
samhouts: When I see a method here that says "IsDefault", I'm going to expect it to return true if the value == the default value, whether that's set by the prop or set by a default value creator or set by my code. This gets us part of that behavior but not all of it, I think. Correct me if I'm wrong.

StephaneDelcroix: I like to think that a v value created with the `defaultValueCreator` is not a default value, as it should be read, and it's value honored.

@jassmith could you weight in one name or the other ?

PR #1831, opened 2018-02-07
[macOS] Alternate picker.  Changed native control to NSPopupButton.

### LINKED ISSUE #1610: Picker only opens when clicking on the arrow, not anywhere else [Mac]
### Description

I have added a Picker to my Xamarin.Forms Mac application, and the only way to get it to open is to click on the small blue square at the right end of it. Clicking on the white region that contains the current selection will not open the Picker. I also can't see a method to have it open programmatically, so I can't solve this by triggering that method on a GestureRecognizer

### Steps to Reproduce

1. Create a Xamarin.Forms Mac application with a Picker
2. Click the white part of the picker where the text goes (not the blue box with the arrow)
3. Observe that the picker does not drop down to reveal its contents.

### Expected Behavior

Clicking anywhere on the picker should open it, so that it behaves the same as other Mac applications (the "Rotation" option in the display preferences on a Mac is a good example).

### Actual Behavior

Clicking anywhere but the blue box does nothing.

### Basic Information

- Version with issue: Xamarin.Forms.2.5.0.121934
- IDE: Rider
- Platform Target Frameworks: 
  - Mac: Xamarin.Mac == 4.0.0.214.

BradChase2011:
### Description of Change ###

Changed the native control in macOS's Picker to the NSPopUpButton from the NSComboBox.

### Bugs Fixed ###

fixes #1610

### API Changes ###

None.

### Behavioral Changes ###

The popup on the control will now size to the elements in the list no longer ellipsising the items.  It also allows for the control to be clicked over the entirety of the control rather than just the drop down button.

### PR Checklist ###

### Screenshots ###
OLD:

screen shot 2018-02-07 at 2 52 19 pm

NEW:
screen shot 2018-02-07 at 2 58 29 pm

### DIFF
--- a/Xamarin.Forms.Platform.MacOS/Renderers/PickerRenderer.cs
+++ b/Xamarin.Forms.Platform.MacOS/Renderers/PickerRenderer.cs
@@ -83,10 +83,7 @@ namespace Xamarin.Forms.Platform.MacOS
 				{
 					_disposed = true;
 					if (Element != null)
-					{
-						//TODO: Implement ObservableList picker source change 
-						//((ObservableList<string>)Element.Items).CollectionChanged -= RowsCollectionChanged;
-					}
+						((INotifyCollectionChanged)Element.Items).CollectionChanged -= RowsCollectionChanged;
 
 					if (Control != null)
 						Control.Activated -= ComboBoxSelectionChanged;
--- a/Xamarin.Forms.Platform.MacOS/Renderers/PickerRenderer.cs
+++ b/Xamarin.Forms.Platform.MacOS/Renderers/PickerRenderer.cs
@@ -2,38 +2,50 @@
 using AppKit;
 using System.ComponentModel;
 using Foundation;
+using System.Collections.Specialized;
+using System.Linq;
 
 namespace Xamarin.Forms.Platform.MacOS
 {
-	public class PickerRenderer : ViewRenderer<Picker, NSComboBox>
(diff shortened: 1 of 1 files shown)


rmarinho:
Can you add some screenshots please

BradChase2011:
@rmarinho 
Screenshots Added.

Xamarin.Forms.Platform.MacOS/Renderers/PickerRenderer.cs
```diff
@@ -77,15 +89,15 @@ protected override void Dispose(bool disposing)
 					}
 
 					if (Control != null)
-						Control.SelectionChanged -= ComboBoxSelectionChanged;
+						Control.Activated -= ComboBoxSelectionChanged;
```
rmarinho: Unsubscribe the CollectionChange too here

BradChase2011: Good find.  I pushed up the changes.  That leads me to some other larger issues at hand.  I am not sure all the work being done in the Dispose method really needs to be happening.  As well as how they are actually being called.  I am seeing minutes worth of time to remove some of our views from scope.  I unfortunately cannot do memory tests right now because the profiler currently cant run on our app.  As soon as we can get that working, I want to run through macOS first and then iOS and verify what is leaking and what isnt and performance test these dispose methods which are again taking minutes to drop one of our views.  I can we we absolutely do have leaks on macOS right now.

PR #1833, opened 2018-02-07
AutoCapitalization support for Entry/Editor

### LINKED ISSUE #1683: [Enhancement] AutoCaptialization support for Entry/Editor
## Rationale

The Entry and Editor Keyboard flags do not represent all the possible platform options for Input Capitalization

## Implementation

Expand Keyboard flags to include Word, Character, and No Capitalized. The keyboard already supports CapitalizeSentence so it just needs to have Words and Characters added to it

Currently
```
[Flags]
	public enum KeyboardFlags
	{
		None = 0,
		CapitalizeSentence = 1,
		Spellcheck = 1 << 1,
		Suggestions = 1 << 2,
		All = ~0
	}
```
Expand To

```
[Flags]
	public enum KeyboardFlags
	{
		None = 0,
		CapitalizeSentence = 1,
		Spellcheck = 1 << 1,
		Suggestions = 1 << 2,
		CapitalizeWords = 1 << 3,
		CapitalizeCharacters = 1 << 4,
		CapitalizeNone = 1 << 5,
		All = ~0
	}
```

Currently None only really has in meaning for iOS

### iOS
Add checks for additional flags to
```
public static void ApplyKeyboard(this IUITextInput textInput, Keyboard keyboard)
```
To set the keyboard settings for iOS

These settings are mutually exclusive so if the user specifies multiple they'll just get one and warning

### Android
Add checks for additional flags to
```
public static InputTypes ToInputType(this Keyboard self)
```
There are currently checks inside this method that verifies valid permutations and warns the user if the settings don't make sense together. The new flags should be figured into these warnings

### UWP
Currently UWP completely lacks any custom keyboard implementation 
```
public static InputScope ToInputScope(this Keyboard self)
```
The WPF implementation has CustomKeyboard settings that mostly all just end up as setting the InputScopeName to Default but it comes with a set of warnings based on non compatible settings which I feel has value
https://msdn.microsoft.com/en-us/library/windows/apps/mt280229.aspx?f=255&MSPPError=-2147217396

Default behavior won't change but user will now get a warning

## Backward Compatibility

The user will see some additional warnings if they setup an invalid configuration on UWP but the behavior will be the same,.  If the user set ALL it would default to Sentence before so just make sure that it still defaults to Sentence if multiple values are set

PureWeen:
### Description of Change ###

The Entry and Editor controls do not have a simple mechanism for controlling the native automatic capitalization features.

Expand Keyboard flags to include Word, Character, and No Capitalized. The keyboard already supports CapitalizeSentence so it just needs to have Words and Characters added to it

Currently
```
[Flags]
	public enum KeyboardFlags
	{
		None = 0,
		CapitalizeSentence = 1,
		Spellcheck = 1 << 1,
		Suggestions = 1 << 2,
		All = ~0
	}
```
Expand To

```
[Flags]
	public enum KeyboardFlags
	{
		None = 0,
		CapitalizeSentence = 1,
		Spellcheck = 1 << 1,
		Suggestions = 1 << 2,
		CapitalizeWords = 1 << 3,
		CapitalizeCharacters = 1 << 4,
		CapitalizeNone = 1 << 5,
		All = ~0
	}
```

Currently None only really has in meaning for iOS

### iOS
Add checks for additional flags to
```
public static void ApplyKeyboard(this IUITextInput textInput, Keyboard keyboard)
```
To set the keyboard settings for iOS

These settings are mutually exclusive so if the user specifies multiple they'll just get one and warning

### Android
Add checks for additional flags to
```
public static InputTypes ToInputType(this Keyboard self)
```
There are currently checks inside this method that verifies valid permutations and warns the user if the settings don't make sense together. The new flags should be figured into these warnings

### UWP
Currently UWP completely lacks any custom keyboard implementation 
```
public static InputScope ToInputScope(this Keyboard self)
```
The WPF implementation has CustomKeyboard settings that mostly all just end up as setting the InputScopeName to Default but it comes with a set of warnings based on non compatible settings which I feel has value
https://msdn.microsoft.com/en-us/library/windows/apps/mt280229.aspx?f=255&MSPPError=-2147217396

Default behavior won't change but user will now get a warning.
You have to use a touch screen keyboard to see any changes. 

### Bugs Fixed ###

fixes #1683

### API Changes ###

Currently
```
[Flags]
	public enum KeyboardFlags
	{
		None = 0,
		CapitalizeSentence = 1,
		Spellcheck = 1 << 1,
		Suggestions = 1 << 2,
		All = ~0
	}
```
Expand To

```
[Flags]
	public enum KeyboardFlags
	{
		None = 0,
		CapitalizeSentence = 1,
		Spellcheck = 1 << 1,
		Suggestions = 1 << 2,
		CapitalizeWords = 1 << 3,
		CapitalizeCharacters = 1 << 4,
		CapitalizeNone = 1 << 5,
		All = ~0
	}
```

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.ControlGallery.Android/CustomRenderers.cs
+++ b/Xamarin.Forms.ControlGallery.Android/CustomRenderers.cs
@@ -30,6 +30,8 @@ using Xamarin.Forms.Controls.Issues;
 
 [assembly: ExportRenderer(typeof(Bugzilla42000._42000NumericEntryNoDecimal), typeof(EntryRendererNoDecimal))]
 [assembly: ExportRenderer(typeof(Bugzilla42000._42000NumericEntryNoNegative), typeof(EntryRendererNoNegative))]
+[assembly: ExportRenderer(typeof(Issue1683.EntryKeyboardFlags), typeof(EntryRendererKeyboardFlags))]
+[assembly: ExportRenderer(typeof(Issue1683.EditorKeyboardFlags), typeof(EditorRendererKeyboardFlags))]
 //[assembly: ExportRenderer(typeof(AndroidHelpText.HintLabel), typeof(HintLabel))]
 [assembly: ExportRenderer(typeof(QuickCollectNavigationPage), typeof(QuickCollectNavigationPageRenderer))]
 
@@ -122,7 +124,7 @@ namespace Xamarin.Forms.ControlGallery.Android
 
 	public class NativeListViewRenderer : ViewRenderer<NativeListView, global::Android.Widget.ListView>
 	{
-#pragma warning disable 618 
+#pragma warning disable 618
 		public NativeListViewRenderer()
 #pragma warning restore 618
 		{
@@ -514,10 +516,12 @@ namespace Xamarin.Forms.ControlGallery.Android
 		}
 	}
 
-#pragma warning disable 618
 	public class CustomButtonRenderer : ButtonRenderer
-#pragma warning restore 618
 	{
+		public CustomButtonRenderer(Context context) : base(context)
+		{
+		}
+
 		protected override AButton CreateNativeControl()
 		{
 			return new CustomNativeButton(Context);
@@ -536,11 +540,12 @@ namespace Xamarin.Forms.ControlGallery.Android
 	}
 
 	// Custom renderers for Bugzilla42000 demonstration purposes
-
-#pragma warning disable 618
 	public class EntryRendererNoNegative : EntryRenderer
-#pragma warning restore 618
 	{
+		public EntryRendererNoNegative(Context context) : base(context)
+		{
+		}
+
 		protected override NumberKeyListener GetDigitsKeyListener(InputTypes inputTypes)
 		{
 			// Disable the NumberFlagSigned bit
@@ -550,16 +555,108 @@ namespace Xamarin.Forms.ControlGallery.Android
 		}
 	}
 
-#pragma warning disable 618
 	public class EntryRendererNoDecimal : EntryRenderer
-#pragma warning restore 618
 	{
+		public EntryRendererNoDecimal(Context context) : base(context)
+		{
+		}
+
 		protected override NumberKeyListener GetDigitsKeyListener(InputTypes inputTypes)
 		{
 			// Disable the NumberFlagDecimal bit
 			inputTypes &= ~InputTypes.NumberFlagDecimal;
 
 			return base.GetDigitsKeyListener(inputTypes);
+		}
+	}
+
+	public class EntryRendererKeyboardFlags : EntryRenderer
+	{
+		public EntryRendererKeyboardFlags(Context context) : base(context)
+		{
+		}
+
+		protected override void OnElementPropertyChanged(object sender, PropertyChangedEventArgs e)
+		{
+			var FlagsToSet = ((Issue1683.EntryKeyboardFlags)Element).FlagsToSet;
+			var FlagsToTestFor = ((Issue1683.EntryKeyboardFlags)Element).FlagsToTestFor;
+
+			base.OnElementPropertyChanged(sender, e);
+
+			Control.SetKeyboardFlags(FlagsToSet);
+			Control.TestKeyboardFlags(FlagsToTestFor);
+		}
+	}
+
+	public class EditorRendererKeyboardFlags : EditorRenderer
+	{
+		public EditorRendererKeyboardFlags(Context context) : base(context)
+		{
+		}
+
+		protected override void OnElementPropertyChanged(object sender, PropertyChangedEventArgs e)
+		{
+			var FlagsToSet = ((Issue1683.EditorKeyboardFlags)Element).FlagsToSet;
+			var FlagsToTestFor = ((Issue1683.EditorKeyboardFlags)Element).FlagsToTestFor;
+			base.OnElementPropertyChanged(sender, e);
+
+			Control.SetKeyboardFlags(FlagsToSet);
+			Control.TestKeyboardFlags(FlagsToTestFor);
+		}
+	}
+
+	public static class KeyboardFlagExtensions
+	{
+		public static void TestKeyboardFlags(this FormsEditText Control, KeyboardFlags? flags)
+		{
+			if (flags == null)
+			{
+				return;
+			}
+			if (flags.Value.HasFlag(KeyboardFlags.CapitalizeSentence))
+			{
+				if (!Control.InputType.HasFlag(InputTypes.TextFlagCapSentences))
+				{
+					throw new Exception("TextFlagCapSentences not correctly set");
+				}
+			}
+			else if (flags.Value.HasFlag(KeyboardFlags.CapitalizeCharacter))
+			{
+				if (!Control.InputType.HasFlag(InputTypes.TextFlagCapCharacters))
+				{
+					throw new Exception("TextFlagCapCharacters not correctly set");
+				}
+			}
+			else if (flags.Value.HasFlag(KeyboardFlags.CapitalizeWord))
+			{
+				if (!Control.InputType.HasFlag(InputTypes.TextFlagCapWords))
+				{
+					throw new Exception("TextFlagCapWords not correctly set");
+				}
+			}
+		}
+
+		public static void SetKeyboardFlags(this FormsEditText Control, KeyboardFlags? flags)
+		{
+			if (flags == null)
+			{
+				return;
+			}
+
+			if (flags.Value.HasFlag(KeyboardFlags.CapitalizeCharacter))
+			{
+				Control.InputType = Control.InputType | InputTypes.TextFlagCapCharacters;
+			}
+
+			if (flags.Value.HasFlag(KeyboardFlags.CapitalizeSentence))
+			{
+				Control.InputType = Control.InputType | InputTypes.TextFlagCapSentences;
+			}
+
+			if (flags.Value.HasFlag(KeyboardFlags.CapitalizeWord))
+			{
+				Control.InputType = Control.InputType | InputTypes.TextFlagCapWords;
+			}
 		}
 	}
 
@@ -573,7 +670,7 @@ namespace Xamarin.Forms.ControlGallery.Android
 	// }
 
 #pragma warning disable CS0618 // Leaving in old constructor so we can verify it works
-	public class NoFlashTestNavigationPage 
+	public class NoFlashTestNavigationPage
 #if FORMS_APPLICATION_ACTIVITY
 		: Xamarin.Forms.Platform.Android.NavigationRenderer
 #else
--- a/Xamarin.Forms.ControlGallery.iOS/CustomRenderers.cs
+++ b/Xamarin.Forms.ControlGallery.iOS/CustomRenderers.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using CoreLocation;
 using Foundation;
 using MapKit;
@@ -21,8 +22,11 @@ using RectangleF = CoreGraphics.CGRect;
 [assembly: ExportRenderer(typeof(Bugzilla43161.AccessoryViewCell), typeof(AccessoryViewCellRenderer))]
 [assembly: ExportRenderer(typeof(Bugzilla36802.AccessoryViewCell), typeof(AccessoryViewCellRenderer))]
 [assembly: ExportRenderer(typeof(Bugzilla52700.NoSelectionViewCell), typeof(NoSelectionViewCellRenderer))]
+[assembly: ExportRenderer(typeof(Issue1683.EntryKeyboardFlags), typeof(EntryRendererKeyboardFlags))]
+[assembly: ExportRenderer(typeof(Issue1683.EditorKeyboardFlags), typeof(EditorRendererKeyboardFlags))]
 namespace Xamarin.Forms.ControlGallery.iOS
 {
+
 	public class CustomIOSMapRenderer : ViewRenderer<CustomMapView, MKMapView>
 	{
 		private MKMapView _mapView;
@@ -391,7 +395,9 @@ namespace Xamarin.Forms.ControlGallery.iOS
 
 		public IEnumerable<string> Items
 		{
-			set { _tableItems = new List<string>(value); 
+			set
+			{
+				_tableItems = new List<string>(value);
 			}
 		}
 
@@ -504,7 +510,8 @@ namespace Xamarin.Forms.ControlGallery.iOS
 		public CollectionViewController(UICollectionViewLayout layout, OnItemSelected onItemSelected) : base(layout)
 		{
 			items = new List<string>();
-			for (int i = 0; i < 20; i++) {
+			for (int i = 0; i < 20; i++)
+			{
 				items.Add($"#{i}");
 			}
 			_onItemSelected = onItemSelected;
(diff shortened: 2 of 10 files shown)


Xamarin.Forms.Platform.iOS/Renderers/EditorRenderer.cs
```diff
@@ -10,6 +10,11 @@ public class EditorRenderer : ViewRenderer<Editor, UITextView>
 		bool _disposed;
 		IEditorController ElementController => Element;
 
+		// Autocapitalization defaults to setences
```
adrianknight89: typo in `setences`

Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
```diff
@@ -23,6 +23,11 @@ public class EntryRenderer : ViewRenderer<Entry, UITextField>
 
 		static readonly int baseHeight = 30;
 		static CGSize initialSize = CGSize.Empty;
+				
+		// Autocapitalization defaults to setences
```
adrianknight89: same here

Xamarin.Forms.Platform.UAP/EntryRenderer.cs
```diff
@@ -204,5 +207,20 @@ void UpdateTextColor()
...
+		{
+			if (Element.IsSet(Xamarin.Forms.InputView.AutoCapitalizationProperty))
+			{
+				Control.AutoCapitalization = Element.AutoCapitalization;
+				switch (Element.AutoCapitalization)
+				{
+					case AutoCapitalization.Sentences:
+					case AutoCapitalization.Words:
+						Internals.Log.Warning(nameof(Entry), $"{Element.AutoCapitalization} is not supported on windows because they are controlled by Settings->Typing");
+						break;
```
adrianknight89: Capitalize the `W` in `windows`. Ironic. :)

Xamarin.Forms.Platform.UAP/FormsTextBox.cs
```diff
@@ -53,10 +57,12 @@ public class FormsTextBox : TextBox
 		public FormsTextBox()
 		{
 			TextChanged += OnTextChanged;
+			TextChanging += OnTextChanging;
```
adrianknight89: I haven't taken a deep look at `FormsTextBox`, but it looks like none of these events are being unsubscribed. Hmm.

adrianknight89:
Just wondering, has this been tested on different platforms when keybord autocapitalization is disabled by the user? Also, the Android keyboard flags don't always seem to work on older API levels. I remember working on bugs a long time that had a similar problem.

Xamarin.Forms.Core/AutoCapitalization.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Text;
+
+namespace Xamarin.Forms
+{
+    public enum AutoCapitalization
+    {
+		/// <summary>
+		/// Don't set anything on the platform component.
+		/// Currently this only affects iOS
+		/// </summary>
```
StephaneDelcroix: we don't do inline comments for public types

Xamarin.Forms.Core/AutoCapitalization.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Xamarin.Forms
+{
+    public enum AutoCapitalization
```
StephaneDelcroix: doesn't it make sense to capitalize it as `Autocapitalization` ? like it is in iOS ?

StephaneDelcroix: could you add a TypeConverter so it converts CSS values https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform to the right enum ?

Xamarin.Forms.Core/AutoCapitalization.cs
```diff
@@ -0,0 +1,31 @@
...
+		Words = 2,
+		/// <summary>
+		/// capitalize the first character of each sentence. 
+		/// </summary>
+		Sentences = 3,
+		/// <summary>
+		///  capitalize all characters. 
+		/// </summary>
+		Characters = 4
+	}
```
StephaneDelcroix: what about turning a sentence to lowercase ? (`lowercase` in CSS)

Xamarin.Forms.Platform.iOS/Renderers/EditorRenderer.cs
```diff
@@ -92,6 +100,43 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
 				UpdateTextAlignment();
 		}
 
+		void UpdateAutoCapitalization()
+		{
+			AutoCapitalization autoCap = AutoCapitalization.Default;
+			if (Element.IsSet(Xamarin.Forms.InputView.AutoCapitalizationProperty))
+			{
+				if (!_hasAutocapitalizationBeenSetBefore)
```
StephaneDelcroix: you can capture that unconditionally, the same way we do it for default colors

Xamarin.Forms.Platform.iOS/Renderers/EditorRenderer.cs
```diff
@@ -92,6 +100,43 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
...
+				{
+					_defaultAutocapitalizationType = Control.AutocapitalizationType;
+					_hasAutocapitalizationBeenSetBefore = true;
+				}
+				autoCap = (AutoCapitalization)Element.GetValue(Xamarin.Forms.InputView.AutoCapitalizationProperty);
+			}
+
+			switch (autoCap)
```
StephaneDelcroix: I'd write an extension method `ToAutocapitalizationType()`, but that's me

Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
```diff
@@ -249,5 +257,43 @@ void UpdateText()
...
+				if (!_hasAutocapitalizationBeenSetBefore)
+				{
+					_defaultAutocapitalizationType = Control.AutocapitalizationType;
+					_hasAutocapitalizationBeenSetBefore = true;
+				}
+				autoCap = (AutoCapitalization)Element.GetValue(Xamarin.Forms.InputView.AutoCapitalizationProperty);
+			}
+
+			switch (autoCap)
+			{
```
StephaneDelcroix: and I'd reuse that extension method here

Xamarin.Forms.Core/AutoCapitalization.cs
PureWeen: I had that thought but this issue seems to be more about the input type settings then it is about applying transformations to the text. The AutoCap settings on iOS/Android all apply to what state they place the keyboard into as the user types. UWP you can enforce rules on the box itself but that seemed to not fit the intentions of the issue. Maybe the enum needs to be changed to CapitalizationInputType?

PureWeen:
> what about turning that on for Label, Buttons

I feel like this issue is more about changing the InputType and not applying transformations to text. If we wanted to apply Words/Sentence transformations to Labels and Buttons I don't think there's a platform version of that so we'd have to roll our own transformations against those 

Which seems like it would be a different setting something like
TextTransformation="AllCaps"

Xamarin.Forms.Core/AutoCapitalization.cs
PureWeen: @StephaneDelcroix  these probably belong to  KeyboardFlags as additional flags

PureWeen:
@adrianknight89  what would be your expected results from those tests be with the setting disabled?  I only see those settings in iOS and when I turn them off then none of this does anything and it just acts like it's been set to None.

As far as APIs at this point I've tested on a physical device that's API 22 and Api 25,27 on emulators

PureWeen:
This implementation is going to be changed to use keyboard flags
https://github.com/xamarin/Xamarin.Forms/issues/1683

Just an fyi :-)

adrianknight89:
@PureWeen On Android, you could install custom keyboards like Google Keyboard, and I remember there being an option to turn on/off keyboard features. I've come across cases where keyboard APIs did not necessarily work. For example, see #396 for `InputType`. That's why I brought up testing this on different APIs and devices.

PureWeen:
@adrianknight89  thanks for referencing that previous issue. That provides some very helpful context and past precedence. 

I tested these changes on an API 15 emulators and they worked as I would expect them to currently.  Right now with this implementation it just boils down to keyboard hints that the keyboard can do what it wants to with.  

Which I can see with those numeric changes wasn't really good enough for relying on Android to do the right thing with.   

UWP really gives you no control over these settings individually :-/

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1683.cs
```diff
@@ -0,0 +1,147 @@
...
+using Xamarin.UITest;
+using NUnit.Framework;
+using System.Diagnostics;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve (AllMembers = true)]
+	[Issue (IssueTracker.Github, 1683, "Auto Capitalization Implementation")]
+	public class Issue1683 : ContentPage
```
samhouts: Can we maybe add a UITest to this that just rotates through all the options and captures screenshots? I'm not sure if it can actually check for the capitalization changes, but I'd like to at least have it run on devices to catch any potential crashes.

Xamarin.Forms.Platform.Android/Renderers/KeyboardExtensions.cs
```diff
@@ -31,9 +31,12 @@ public static InputTypes ToInputType(this Keyboard self)
 			else if (self is CustomKeyboard)
 			{
 				var custom = (CustomKeyboard)self;
-				bool capitalizedSentenceEnabled = (custom.Flags & KeyboardFlags.CapitalizeSentence) == KeyboardFlags.CapitalizeSentence;
-				bool spellcheckEnabled = (custom.Flags & KeyboardFlags.Spellcheck) == KeyboardFlags.Spellcheck;
-				bool suggestionsEnabled = (custom.Flags & KeyboardFlags.Suggestions) == KeyboardFlags.Suggestions;
+				bool capitalizedSentenceEnabled = custom.Flags.HasFlag(KeyboardFlags.CapitalizeSentence);
```
samhouts: We tend to avoid using `HasFlag` because has a substantial performance cost.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1683.cs
samhouts: Also, maybe add to this test a custom renderer that just sets the capitalization. Does the custom renderer still work if we haven't set the `KeyboardFlags`?

Xamarin.Forms.Platform.UAP/KeyboardExtensions.cs
```diff
@@ -42,8 +44,39 @@ public static InputScope ToInputScope(this Keyboard self)
...
-				name.NameValue = InputScopeNameValue.Default;
+				var custom = (CustomKeyboard)self;
+				bool capitalizedSentenceEnabled = custom.Flags.HasFlag(KeyboardFlags.CapitalizeSentence);
+				bool capitalizedWordsEnabled = custom.Flags.HasFlag(KeyboardFlags.CapitalizeWord);
+				bool capitalizedCharacterEnabled = custom.Flags.HasFlag(KeyboardFlags.CapitalizeCharacter);
+				bool spellcheckEnabled = custom.Flags.HasFlag(KeyboardFlags.Spellcheck);
+				bool suggestionsEnabled = custom.Flags.HasFlag(KeyboardFlags.Suggestions);
+				InputScopeNameValue nameValue = InputScopeNameValue.Default;
+				
+				if(capitalizedSentenceEnabled)
```
samhouts: 👮 Noticing inconsistency with the space between `if` and `(`. Just wanted to point it out so you can fix your IDE. :)

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1683.cs
```diff
@@ -0,0 +1,147 @@
+﻿using System;
+
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System.Collections.Generic; 
+
+#if UITEST
```
samhouts: If you opt to not create a UI test, then we should just remove these unnecessary `if` defs. :)

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1683.cs
PureWeen: I started down the path of the UITests but then realized the only way to test these would be to type onto the keyboard which AFAICT you can't really do. But your second point is a good one so I'll just add the UITests I wrote back in to just cycle through the fields and make sure it doesn't crash.

I'll added a custom renderer version of these as well and test through them

Xamarin.Forms.Platform.Android/Renderers/KeyboardExtensions.cs
PureWeen: Good to know. I noticed about 33 places in the code that used it so I just thought maybe it was personal preference.  I'll create a quick issue around that so the standard is established

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1683.cs
```diff
@@ -0,0 +1,202 @@
...
+		}
+
+
+#if UITEST
+		[Test]
+		public void Issue1683Test()
+		{
+			for (int i = 0; i < 6; i++)
+			{
+				RunningApp.WaitForElement("Rotation");
```
samhouts: `RunningApp.WaitForElement(q => q.Marked("Rotation")); ` should fix your UI test errors

.editorconfig
```diff
@@ -1,4 +1,5 @@
 root = true
 
 [*.{cs,xaml}]
-indent_style = tab
\ No newline at end of file
+indent_style = tab
```
hartez: I'm guessing changes to .editorconfig weren't meant to be in the PR.

Xamarin.Forms.ControlGallery.Android/CustomRenderers.cs
```diff
@@ -564,6 +566,95 @@ protected override NumberKeyListener GetDigitsKeyListener(InputTypes inputTypes)
 	}
 
 
+
+#pragma warning disable 618
```
hartez: Instead of disabling the warning, just add the missing constructor for these. That way if people are looking here for example code they don't copy the obsolete renderer format.

Xamarin.Forms.ControlGallery.iOS/CustomRenderers.cs
```diff
@@ -597,5 +604,92 @@ public override UITableViewCell GetCell(Cell item, UITableViewCell reusableCell,
...
+				{
+					throw new Exception("TextFlagCapSentences not correctly set");
+				}
+			}
+
+			else if (flags.Value.HasFlag(KeyboardFlags.CapitalizeCharacter))
+			{
+				if (currentValue != UITextAutocapitalizationType.AllCharacters)
+				{
+                    throw new Exception("CapitalizeCharacter not correctly set");
```
hartez: Spaces -> tabs

Xamarin.Forms.ControlGallery.iOS/CustomRenderers.cs
```diff
@@ -597,5 +604,92 @@ public override UITableViewCell GetCell(Cell item, UITableViewCell reusableCell,
...
+				}
+			}
+			else if (flags.Value.HasFlag(KeyboardFlags.CapitalizeWord))
+			{
+
+				if (currentValue != UITextAutocapitalizationType.Words)
+				{
+                    throw new Exception("CapitalizeWord not correctly set");
```
hartez: Spaces -> tabs

Xamarin.Forms.Platform.UAP/KeyboardExtensions.cs
```diff
@@ -1,5 +1,7 @@
 ﻿using System;
+using System.Diagnostics;
```
hartez: Is this directive being used?

Xamarin.Forms.Platform.UAP/KeyboardExtensions.cs
```diff
@@ -42,8 +44,41 @@ public static InputScope ToInputScope(this Keyboard self)
...
+					{
+						Log.Warning(null, "CapitalizeWord only works when spell check is enabled");
+					}
+
+					nameValue = InputScopeNameValue.NameOrPhoneNumber;
+				}
+
+				if (capitalizedCharacterEnabled)
+				{
+					Log.Warning(null, "UWP doesn't support CapitalizeCharacter");
```
hartez: "doesn't" -> "does not"

.editorconfig
PureWeen: @hartez  I'll remove this specific confusing change but I did intend to modify editor config to include
```
csharp_space_after_keywords_in_control_flow_statements = true
```
After @samhouts  had pointed out some incorrect spacing that I was using.    Is that ok to leave that in for this PR?

hartez: I'd rather have that in a separate commit.

PureWeen:
@hartez  alright rebased and pushed changes

.editorconfig
PureWeen: sounds good created a different PR for that

myroot:
`CapitalizeNone = 1 << 5,` Is it need? 
If CapitalizeSentence, CapitalizeWord and CapitalizeCharacter are not existed in KeyboardFlags, It is absolutely CapitalizeNone.

StephaneDelcroix:
@myroot `None` means "don't change capitalization" (aka default) while `CapitalizeNone` means "everything lowercase".

myroot:
@StephaneDelcroix  But, implementation does not. Same behavior between `None` and `CapitializeNone`.
And, all platform does not support "enforce lowercase"

iOS : https://developer.apple.com/documentation/uikit/uitextautocapitalizationtype
`None : Specifies that there is no automatic text capitalization.`

Android : https://developer.android.com/reference/android/text/InputType.html#TYPE_TEXT_FLAG_CAP_CHARACTERS
 There is no option for CaptializeNone

myroot:
Ok, I now understand
As i understand, 
None means `don't change setting for 'automatic text capitalization'`
CapitalizeNone means `set no automatic text capitalization`

Thanks for explain

JFMG:
I am using Xamarin.Forms 2.5.0.122203 and having issues with the Keyboard...

I got the following code:
```csharp
MyEntry.Keyboard = Keyboard.Create(KeyboardFlags.None);
```
In iOS is the first letter still capitalized. In Android it's not.

When I use an Email keyboard like this:
```xml
<Entry Keyboard="Email" />
```
In iOS is the first letter still capitalized. In Android it's not.

Is that the intended behavior? Can I set the first letter as non capitalized without writing a custom renderer?

PR #1844, opened 2018-02-08
[Core] Add DesignMode.DesignModeEnabled

### LINKED ISSUE #1731: [Enhancement] Add API for programmatically checking if running in design mode
## Rationale

User’s desire a way to specify code that should only run when their code is running in a designer context such as Previewer or Live Player.

## Implementation

```
// In Xamarin.Forms
pubic static class DesignMode
{
    public static bool DesignModeEnabled { get; set; }
}
```

```
// Example usage 
if(Xamarin.Forms.DesignMode.DesignModeEnabled)
{
    // do something special
}
```

The property should be considered to have a public setter so that third-party tools can also mark this as true when they would prefer to. i.e. LiveXaml, GorillaPlayer, etc. may want to set this to true so users can do something special in those contexts. Or, perhaps it's better to make this read only and make some static method for enabling it. i.e. 

`Xamarin.Forms.DesignMode.EnableDesignMode()` 

See [UWP documentation](https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.designmode#Windows_ApplicationModel_DesignMode_DesignModeEnabled). 

## Expected Result

`DesignModeEnabled` would return `true` when code is running in a preview or design-time tool. Otherwise, false. 

### Android

Nothing specific for Android.

### iOS

Nothing specific for iOS.

### UWP

Nothing specific for UWP

## Implications for CSS

None that I am aware of.

## Backward Compatibility

Uncertain, but it seems a new API like this would have little to no backwards compatibility issues.

## Difficulty: Easy

There may be some additional coordination required with the designer team for a few additions for the previewer to work with this. Otherwise, it should be simple to implement.

therealjohn:
### Description of Change ###

Adds a static class and static bool for `DesignModeEnabled` to `Xamarin.Forms.Core`. This is to address the F100 Issue (fixes #1731). This is only intended to address programmatic checks by a developer. It does not cover things like `d:DesignHeight` or other designer aware XAML elements and attributes. 

Tooling would need to toggle the `DesignModeEnabled` property whenever appropriate so that developers can rely on it. 

### Bugs Fixed ###

N/A

### API Changes ###

Added:
 - `public static class DesignMode`
 - `public static bool DesignModeEnabled`

### Behavioral Changes ###

None.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/DesignMode.cs
+++ b/Xamarin.Forms.Core/DesignMode.cs
@@ -1,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Xamarin.Forms
+{
+    public static class DesignMode
+    {
+		public static bool DesignModeEnabled { get; internal set; }
+	}
+}
--- a/Xamarin.Forms.Core/Internals/ResourceLoader.cs
+++ b/Xamarin.Forms.Core/Internals/ResourceLoader.cs
@@ -5,8 +5,17 @@ namespace Xamarin.Forms.Internals
 {
 	public static class ResourceLoader
 	{
+		static Func<string, string> resourceProvider;
+
 		//takes a resource path, returns string content
-		public static Func<AssemblyName, string, string> ResourceProvider { get; internal set; }
+		public static Func<string, string> ResourceProvider {
+			get { return resourceProvider; }
+			internal set {
+				DesignMode.DesignModeEnabled = true;
+				resourceProvider = value;
+			}
+		}
+
 		internal static Action<Exception> ExceptionHandler { get; set; }
 	}
 }
--- a/docs/Xamarin.Forms.Core/index.xml
+++ b/docs/Xamarin.Forms.Core/index.xml
@@ -196,6 +196,7 @@
       <Type Name="DependencyAttribute" Kind="Class" />
       <Type Name="DependencyFetchTarget" Kind="Enumeration" />
       <Type Name="DependencyService" Kind="Class" />
+      <Type Name="DesignMode" Kind="Class" />
       <Type Name="Device" Kind="Class" />
       <Type Name="Device+Styles" Kind="Class" />
       <Type Name="Easing" Kind="Class" />
--- a/Xamarin.Forms.Core.UnitTests/DesignModeTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/DesignModeTests.cs
@@ -1,0 +1,22 @@
+﻿using NUnit;
+using NUnit.Framework;
+
+namespace Xamarin.Forms.Core.UnitTests
+{
+	public class DesignModeTests : BaseTestFixture
+	{
+		[Test]
+		public void DesignModeEnabledIsFalseByDefault()
+		{
+			Assert.IsFalse(DesignMode.DesignModeEnabled);
+		}
+
+		[Test]
+		public void DesignModeEnabledIsTrueWhenSet()
+		{
+			DesignMode.DesignModeEnabled = true;
+
+			Assert.IsTrue(DesignMode.DesignModeEnabled);
+		}
+	}
+}
--- a/Xamarin.Forms.Core.UnitTests/Xamarin.Forms.Core.UnitTests.csproj
+++ b/Xamarin.Forms.Core.UnitTests/Xamarin.Forms.Core.UnitTests.csproj
@@ -73,6 +73,7 @@
     <Compile Include="ColorUnitTests.cs" />
     <Compile Include="CommandSourceTests.cs" />
     <Compile Include="CommandTests.cs" />
+    <Compile Include="DesignModeTests.cs" />
     <Compile Include="EffectiveFlowDirectionExtensions.cs" />
     <Compile Include="FlowDirectionTests.cs" />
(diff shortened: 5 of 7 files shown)


Xamarin.Forms.Core/DesignMode.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Xamarin.Forms.Core
+{
+    public static class DesignMode
+    {
+		/// <summary>
```
StephaneDelcroix: we don't do inline doc. our doc tool can't even extract it from there

therealjohn:
@StephaneDelcroix Tools like the XAML Previewer would need to toggle this when appropriate.

Xamarin.Forms.Core/DesignMode.cs
therealjohn: What's the proper way? I added that for intellisense.

StephaneDelcroix:
> @StephaneDelcroix Tools like the XAML Previewer would need to toggle this when appropriate.

I get that part

I'm looking for the part where it loads designmode xaml elements and attributes...

therealjohn:
@StephaneDelcroix Sorry for the confusion. See my comment [here](https://github.com/xamarin/Xamarin.Forms/issues/1731#issuecomment-364243278).

alanmcgovern:
Basically we souls set the bool to true when we set our delegate in the XamlLoader or ResourceLoader rendering api :)

therealjohn:
@alanmcgovern I wasn't sure if the public setter was awkward or not. The thought behind making it public is to allow third-party tools like Gorilla Player and LiveXAML to take advantage of this too.

StephaneDelcroix:
the problem with a public setter is that users will turn it on and off, in app, at runtime for totally other reasons. supporting this will prevent us from a lot of optimization.

therealjohn:
@StephaneDelcroix I had the same thought. It does feel like an awkward API for that reason. Is there something else we could do to enable other tools to take advantage of this? 

```
public static void EnableDesignMode()
public static void DisableDesignMode()
```

alanmcgovern:
I don't think we want to expose this as public API for anyone to arbitrarily set. If a third party *really* wants to set this then they can use reflection to do so, knowing that they're doing something unsupported. Not even our own tooling is going to directly set this boolean :)

therealjohn:
@StephaneDelcroix @alanmcgovern https://github.com/xamarin/Xamarin.Forms/pull/1844/commits/954fec1467241b866ffe5daa2f162ed3152760a4 adds the internal setter. What else did I mess up that VSTS is failing?

docs/Xamarin.Forms.Core/index.xml
```diff
@@ -456,6 +456,9 @@
       <Type Name="WebViewSource" Kind="Class" />
       <Type Name="WebViewSourceTypeConverter" Kind="Class" />
     </Namespace>
+    <Namespace Name="Xamarin.Forms.Core">
```
StephaneDelcroix: the duplicate namespace is probably what annoys VSTS

alanmcgovern:
We should set the boolean to true whenever the previewer connects to the provided hooks in these two places:
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Xaml/XamlLoader.cs#L46 and https://github.com/xamarin/Xamarin.Forms/blob/38f244aa49fafd104de7d9d30d30bc16f9f8125d/Xamarin.Forms.Core/Internals/ResourceLoader.cs#L9

therealjohn:
@StephaneDelcroix I've removed the namespace. Any new clues to why VSTS is failing?

rmarinho:
@therealjohn is failing because of docs, rebase also with master please. 
Thanks

alanmcgovern:
I'm not sure what happened to this PR, but i rebased it on master, cleaned the history and added the changes we need to support the previewer. It's now 2 commits based on tip of master in this repo. I'll let john take it from here!

https://github.com/therealjohn/Xamarin.Forms/pull/1

rmarinho:
This needs docs update so it can build on CI

Xamarin.Forms.Core/Internals/ResourceLoader.cs
```diff
@@ -5,8 +5,17 @@ namespace Xamarin.Forms.Internals
 {
 	public static class ResourceLoader
 	{
+		static Func<string, string> resourceProvider;
+
 		//takes a resource path, returns string content
-		public static Func<AssemblyName, string, string> ResourceProvider { get; internal set; }
+		public static Func<string, string> ResourceProvider {
```
rmarinho: this change needs docs too @therealjohn

rmarinho:
It's failing to build @therealjohn  
``` 
Xamarin.Forms.Xaml\XamlLoader.cs (171, 56) 
Xamarin.Forms.Xaml\XamlLoader.cs(171,56): Error CS1501: No overload for method 'Invoke' takes 2 arguments
```

alanmcgovern:
Hi - in the interests of sanity can you guys complete and merge this PR? We have burned an awful lot of time trying to fix what is an extremely trivial PR and it's not getting any further along. If it's this hard for us to add such simple features I don't think we'll ever be able to land an actual change.

StephaneDelcroix:
see #1976

PR #1854, opened 2018-02-09
[iOS] ListView full width separators

### LINKED ISSUE #1665: [Enhancement] ListView full width separators on iOS
## Rational

Making the ListView separator on iOS the full width of the screen is sometimes a desireable effect.

## Implementation

```csharp
ListView.On<iOS>().SetSeparatorStyle(SeparatorStyle.FullWidth/Default);

namespace Xamarin.Forms.PlatformConfiguration.iOSSpecific
{

	public enum SeparatorStyle
	{
		Default,
		FullWidth,
	}
}
```

Use an enum so if more styles need to be created in the future, the api is easily extended.

Extra care will have to be taken with iOS 11 to ensure this works properly with the iPhone X as it has funky behavior with the bunny ears.

### Android
No Change

### iOS
Setting the SeparatorStyle to FullWidth should cause the ListView separators to stretch from the leftmost pixel of the ListView to the rightmost. This should apply to header separators as well.

### UWP
No change

## Implications for CSS
Not relevant to CSS

## Backward Compatibility

Third party renderers may need to be updated to ensure that this functionality is supported through the new official mechanism. Further we will need to be careful to code the changes to the renderers in a careful manner to ensure that if someone is already using an effect support this feature that the effect is as best as possible not broken by our changes.

## Difficulty : Easy

The actual modifications to the renderer look fairly straightforward. The impact for the iPhone X is probably the largest curveball.

henricm:
This PR adds a `SeparatorStyle` (with possible values `Default, FullWidth`) property to iOS specific `ListView`. Setting this to `FullWidth` causes the separator between cells to use the full width of the screen.

### Description of Change ###

Added new `enum SeparatorStyle {Default, FullWidth}` which can be set to new property `ListView.SeparatorStyle` on iOS. The `ListViewRenderer` checks if this property is set when cells are rendererd in `ListViewDataSource.GetCell()` and sets `SeparatorInset` and `LayoutMargins` to zero if separator style is `FullWidth`. 

A design decision was made to not support changing separator style back to Default once it has been set to FullWidth. So a reused cell already rendered with FullWidth won't be changed if the separator style is reverted to Default. The reasoning is that you probably only set the `SeparatorStyle` once and you don't want to change it on the fly for an already existing list.

### Bugs Fixed ###

- Fixes #1665 ListView full width separators on iOS

### API Changes ###

Added to `iOSSpecific`:
 - `enum SeparatorStyle {Default, FullWidth}`
 - `SeparatorStyle ListView.SeparatorStyle { get; set; } //Bindable Property`

### Behavioral Changes ###

iOS list view separators can now easily use the full width of the by setting `SeparatorStyle` to `FullWidth`.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/ListView.cs
+++ b/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/ListView.cs
@@ -1,0 +1,30 @@
+﻿namespace Xamarin.Forms.PlatformConfiguration.iOSSpecific
+{
+	using FormsElement = Forms.ListView;
+
+	public static class ListView
+	{
+		public static readonly BindableProperty SeparatorStyleProperty = BindableProperty.Create(nameof(SeparatorStyle), typeof(SeparatorStyle), typeof(FormsElement), SeparatorStyle.Default);
+
+		public static SeparatorStyle GetSeparatorStyle(BindableObject element)
+		{
+			return (SeparatorStyle)element.GetValue(SeparatorStyleProperty);
+		}
+
+		public static void SetSeparatorStyle(BindableObject element, SeparatorStyle value)
+		{
+			element.SetValue(SeparatorStyleProperty, value);
+		}
+
+		public static SeparatorStyle GetSeparatorStyle(this IPlatformElementConfiguration<iOS, FormsElement> config)
+		{
+			return GetSeparatorStyle(config.Element);
+		}
+
+		public static IPlatformElementConfiguration<iOS, FormsElement> SetSeparatorStyle(this IPlatformElementConfiguration<iOS, FormsElement> config, SeparatorStyle value)
+		{
+			SetSeparatorStyle(config.Element, value);
+			return config;
+		}
+	}
+}
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1665.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1665.cs
@@ -1,0 +1,61 @@
+﻿using System;
+using System.Linq;
+using System.Collections.Generic;
+
+using Xamarin.Forms.CustomAttributes;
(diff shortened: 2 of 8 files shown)


rmarinho:
@henricm is missing the platform changes?!

Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/ListView.cs
```diff
@@ -0,0 +1,37 @@
...
+		}
+
+		public static IPlatformElementConfiguration<iOS, FormsElement> SetSeparatorStyle(this IPlatformElementConfiguration<iOS, FormsElement> config, SeparatorStyle value)
+		{
+			SetSeparatorStyle(config.Element, value);
+			return config;
+		}
+	}
+
+	public enum SeparatorStyle
```
jassmith: Make this its own file please :)

Depechie:
Small question, most apps I create we need to set the Separator Color to Transparent to hide it because of design choices.
Can the enum be extended to also include None and that it will set the transparent color for us?

rmarinho:
@Depechie we have SeparatorStyle already for that

Depechie:
@rmarinho right, I think you mean https://developer.xamarin.com/api/field/Xamarin.Forms.ListView.SeparatorVisibilityProperty/ 

Not sure why anymore ( need to test this again ), but could it be that setting this to false also changes a bit how viewcells are rendered?
Pretty sure we always use the Transparent color trick for something ( only not sure anymore why )...

Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/ListView.cs
henricm: Sure! Will fix that, and also include the backend changes. Apologies for missing that :-)

henricm:
Added the renderer code and rebased. There are some test cloud failures, but I'm not sure it's related to changes in this PR.

jassmith:
Looks like the failure was due to the test run exceeding the 6 hour time limit. Oops.

PR #1870, opened 2018-02-13
Allow users to specify resolution method for handlers, effects, and services

### LINKED ISSUE #1739: [Enhancement] Allow hooking into an App's DI Container
## Rationale

Many apps are currently utilizing one DI Container or another. It would be highly beneficial for Xamarin Forms to be able to hook into the app's container (if present) to be able to create types using the app's existing container rather than relying on `Activator.CreateInstance`. This should allow developers to inject services into Renderers, Effects, etc.

## Implementation

The implementation could vary significantly depending on how far the team wants this enhancement to go. Any implementation should meet the following criteria though

- Replace all direct references to `Activator.CreateInstance`
- The "Default" provider should be the only place in the codebase that references `Activator.CreateInstance`
- This should be opt in so as to have no direct effect on apps not using DI or not opting in

### Basic Implementation

A more basic implementation could be something like:

```cs
// In Xamarin Forms
public static class FormsResolver
{
    static Func<Type, object> _containerResolve;

    public static void SetDefaultFactory(Func<Type, object> containerResolve) =>
        _containerResolve = containerResolve;

    public static object Resolve(Type type)
    {
        if (_containerResolve == null)
            return Activator.CreateInstance(type);

        return _containerResolve(type);
    }

    public static T Resolve<T>() => (T)Resolve(typeof(T));
}
```

```cs
// Example Usage
{
    IContainer Container { get; }
    public App()
    {
        Container = new Container();
        FormsResolver.SetDefaultFactory(t => Container.Resolve(t));
    }
}
```

### More Involved Implementation

If the team prefers to use the opportunity to enhance the existing DependencyService and use it as the base service resolver, then a bit more thought and care will need to go into this. The DependencyService currently handles resolving types a bit backwards. With a typical container, the Lifetime of the service's registration is handled at Registration rather than at Resolution, where as the `DependencyService` currently handles this at resolution `DependencyService.Get(DependencyFetchTarget)`. If `DependencyService` is being updated then the following additional requirements should be met:

- Add non-generic Register/Get variants. 

This could look something like:

```cs
public interface IFormsContainerRegistry
{
    void Register(Type serviceType, Type targetType, DependencyLifetime lifetime);
    void RegisterInstance(Type serviceType, object instance);
    object Resolve(Type type);
}

public enum DependencyLifetime
{
    Transient,
    Singleton
}

public class DependencyAttribute : Attribute
{
    public DependencyAttribute(Type implementorType)
        : this(implementorType, DependencyLifetime.Transient)
    {

    }

    public DependencyAttribute(Type implementorType, DependencyLifetime lifetime)
    {
        Implementor = implementorType;
        Lifetime = lifetime;
    }

    internal DependencyLifetime Lifetime { get; private set; }

    internal Type Implementor { get; private set; }
}

public class DependencyService : IFormsContainerRegistry
{
    static Lazy<DependencyService> implementation => 
        new Lazy<DependencyService>(() => new DependencyService());
    static IFormsContainerRegistry _current;
    static IFormsContainerRegistry Current => 
        _current ?? (_current = implementation.Value);

    public static void SetContainerRegistry(IFormsContainerRegistry registry)
    {
        if (_current != null && _current is DependencyService dependencyService)
        {
            foreach (var dependency in dependencyService.DependencyImplementations)
            {
                var serviceType = dependency.Key;
                var data = dependency.Value;
                if (data.Lifetime == DependencyLifetime.Singleton && data.GlobalInstance != null)
                {
                    registry.RegisterInstance(serviceType, data.GlobalInstance);
                }
                else
                {
                    registry.Register(serviceType, data.ImplementorType, data.Lifetime);
                }
            }
        }

        _current = registry;
    }

    public static object Get(Type type) =>
        Current.Resolve(type);

    public static T Get<T>() =>
        (T)Get(typeof(T));

    public static void Register(Type type) =>
         Current.Register(type, type, DependencyLifetime.Singleton);

    public static void Register<T>() =>
        Register(typeof(T));

    public static void Register(Type serviceType, Type targetType)
        Current.Register(serviceType, targetType, DependencyLifetime.Singleton);

    public static void Register<T, TImpl>()
        where T : class
        where TImpl : class, T =>
        Register(typeof(T), typeof(TImpl));

    [Obsolete("Provided for Legacy Compatibility")]
    public static T Get<T>(DependencyFetchTarget fetchTarget)
    {
        if (Current is DependencyServicev2 dependencyService)
        {
            return (T)dependencyService.GetInternal(typeof(T), fetchTarget);
        }

        // When opting into using a container you would need to rely on the Lifetime 
        // it was registered with
        return (T)Current.Resolve(typeof(T));
    }

    DependencyService()
    {
        DependencyImplementations = new Dictionary<Type, DependencyData>();
        Initialize();
    }

    Dictionary<Type, DependencyData> DependencyImplementations { get; }

    void Initialize()
    {
        // Do initialization
    }

    public void Register(Type serviceType, Type targetType, DependencyLifetime lifetime)
    {
        DependencyImplementations.Add(serviceType, new DependencyData
        {
            ImplementorType = targetType,
            Lifetime = lifetime
        });
    }

    public void RegisterInstance(Type serviceType, object instance)
    {
        DependencyImplementations.Add(serviceType, new DependencyData
        {
            ImplementorType = instance.GetType(),
            GlobalInstance = instance,
            Lifetime = DependencyLifetime.Singleton
        });
    }

    public object Resolve(Type serviceType) =>
        GetInternal(serviceType);

    object GetInternal(Type serviceType, DependencyFetchTarget? fetchTarget = null)
    {
        DependencyData dependencyImplementation = DependencyImplementations[serviceType];
        if (dependencyImplementation == null)
            return null;

        if (dependencyImplementation.Lifetime == DependencyLifetime.Singleton
            || fetchTarget == DependencyFetchTarget.GlobalInstance)
        {
            return dependencyImplementation.GlobalInstance ?? 
                   (dependencyImplementation.GlobalInstance = CreateInstance(dependencyImplementation.ImplementorType));
        }

        return CreateInstance(dependencyImplementation.ImplementorType);
    }

    object CreateInstance(Type type) =>
        Activator.CreateInstance(type);

    class DependencyData
    {
        public Type ImplementorType { get; set; }
        public object GlobalInstance { get; set; }
        public DependencyLifetime Lifetime { get; set; }
    }
}
```

```cs
// Example Usage
public class App : Application, IFormsContainerRegistry
{
    IContainer Container { get; }

    public App()
    {
        Container = new Container();
        DependencyService.SetContainerRegistry(this);
    }

    public void Register(Type serviceType, Type targetType, DependencyLifetime lifetime)
    {
        var reuse = lifetime == DependencyLifetime.Singleton 
                                  ? Reuse.Singleton : Reuse.Transient;
        Container.Register(serviceType, targetType, reuse);
    }

    public void RegisterInstance(Type serviceType, object instance) =>
        Container.UseInstance(serviceType, instance);

    public object Resolve(Type type) =>
        Container.Resolve(type);
}
```

## Expected Result

Types created from within Xamarin Forms are able to be resolved using 3rd party containers to help facilitate injecting services into Custom Renderers, Effects etc.

### Android

Nothing specific to Android

### iOS

Nothing specific to iOS

### UWP

Nothing specific to UWP

## Implications for CSS

None as this doesn't touch UI code directly

## Backward Compatibility

This should be able to be done in a way that only enhances the existing API without breaking legacy code, particularly as an Opt-In feature. 

## Difficulty : Easy/Medium

I would tend to believe the difficulty really depends on whether or not the DependencyService is updated as part of this or left alone.

hartez:
### Description of Change ###

Forms provides no option for users to construct and provide their own instances of handlers, effects, or services in a Forms application; instead, users must rely on `Activator.CreateInstance`. This greatly limits user control over the construction and lifetime of components.

This change provides a way for the user to inject their own dependency resolution method into Forms. When Forms needs a component instance of a particular type, the user's resolution method will be given the opportunity to provide the instance.

If no dependency resolution method is provided, Forms falls back to the previous behavior of creating the components as needed. If the dependency resolution method returns `null` for a requested component, Forms falls back to creating the components itself.

When resolving an `Effect` or handler (e.g., a customer renderer), specifying the resolution method is sufficient. If the user wants to resolve a service from the `DependencyService`, they will have to use the new `Resolve<T>()` method. 

[A brief example of using this with an IoC container](https://gist.github.com/hartez/b8075b1bf05ff3494875ee5c9839157f)

### Bugs Fixed ###

- Fixes #1739 

#### Xamarin.Forms.Internals.DependencyResolver (new class)

`public delegate object ResolveDelegate(Type type, params object[] args);`
`public static void ResolveUsing(ResolveDelegate resolveDelegate)`
`public static void ResolveUsing(Func<Type, object> resolveFunc)`

#### Xamarin.Forms.DependencyService

`public static T Resolve<T>(DependencyFetchTarget fallbackFetchTarget = DependencyFetchTarget.GlobalInstance)`

### Behavioral Changes ###

None

### PR Checklist ###


dansiegel:
Will the registrar need to be registered before Forms.Init()? Or will it be safe to register the registrar inside the App in shared code?

brianlagunas:
Could you clarify the usage?  For example, can we assume that this would be the expected implementation.
```
public partial class App : Application
{
    IContainer _someContainer = new SomeContainer();
    public App()
    {
        InitializeComponent();
        Registrar.ResolveUsing(type => _someContainer.Resolve(type));
        MainPage = new MainPage();
    {
}
``

**Comment by brianlagunas** [comment]
By the way, I think the name should be reconsidered.  The term **Registrar** is not an accurate representation of what is happening.  You aren't registering anything.  You are providing a delegate/Action to be used in place of Activator.CreateInstance to create/resolve objects.

I think there are more appropriate names for this class.  Here are some alternatives:
- DependencyResolver
- FormsResolver
- Resolver

I am personally leaning towards DependencyResolver since this more accurately represents the function of the class. It resolves objects and supplies dependencies when using DI.

Really anything other than "Registrar" would be better :)

**Comment by hartez** [comment]
> Will the registrar need to be registered before Forms.Init()? Or will it be safe to register the registrar inside the App in shared code?

There's no explicit requirement to set the resolution method before `Forms.Init`. It can be set any time.

**Comment by hartez** [comment]
> Could you clarify the usage?

The example you give in your comment is a valid one. 

The only constraint is that the Registrar needs to know about your resolution delegate if you want the Registrar to use it. If there are services in your container which your application needs during startup, you will have to notify the Registrar of your delegate early on. If your container is providing a particular type of Effect, the Registrar will need to know about the delegate before you try to create a control using that Effect. You can set the delegate as early or as late as you want, it just has to be available to the Registrar when the Registrar needs an instance of a type.

**Comment by hartez** [comment]
Adding a DNM; I want to think about some API changes.

**Comment by dansiegel** [comment]
> There's no explicit requirement to set the resolution method before Forms.Init. It can be set any time.

@hartez thanks for the clarification, I wasn't 100% sure given the Gist... I would think that as long as you've set this prior to setting MainPage there shouldn't be a Renderer or Effect being created.

I might add, I do agree with @brianlagunas that the naming of `Registrar` feels a bit misleading.

**Comment by hartez** [comment]
To be clear: I didn't choose the name `Registrar`; that class has been around for a long time. It just happens to be (for better or worse) the place where (most) dynamic object creation in Forms occurs. It also does other stuff that makes the name of the class make more sense. 

That said, as I was responding to @brianlagunas about the name, I realized there might be an opportunity here to split `Registrar` into more appropriately-named parts. I'm taking a look at that right now.

**Comment by hartez** [comment]
> Does this PR also enables control/configuration on whether the Registrar will scan all assemblies for all known Effect/Renderer attributes? Still curious about the performance impact of that one.

This PR doesn't address that. 

I actually spent time working up a proof-of-concept on that exact feature a week or so ago. As it turns out, paring down the number of assemblies scanned for custom attributes does very little to affect startup performance. `GetCustomAttributes` is actually pretty fast. Even on the oldest, creakiest Android device I had available the time spent reflecting for the attributes was sub-millisecond for most assemblies. The overall performance gains of the feature ended up being noise even in a project with over 100 assemblies. (I was very disappointed.)

**Comment by rogihee** [comment]
Bummer! I had high hopes for it :-). Thanks for trying it out though!

File: Xamarin.Forms.Core/DependencyResolver.cs [hunk 1]
```diff
@@ -0,0 +1,49 @@
...
+			object ResolveDelegate(Type type, object[] args) => resolveFunc.Invoke(type);
+			Resolver = ResolveDelegate;
+		}
+
+		internal static object Resolve(Type type, params object[] args)
+		{
+			return Resolver?.Invoke(type, args);
+		}
+
+		internal static object ForceResolve(Type type, params object[] args)
```
kingces95: Suggest `ResolveOrActivate`

Xamarin.Forms.Core/DependencyResolver.cs
```diff
@@ -0,0 +1,49 @@
+using System;
+using System.Linq;
+using System.Reflection;
+
+namespace Xamarin.Forms.Internals
+{
+	public static class DependencyResolver
+	{
+		public delegate object ResolveDelegate(Type type, params object[] args);
```
kingces95: Suggest we use `Func<Type, object[], object>`; Suggest we conserve the number of new Types introduced at the expense syntactic sugar at our call site; We can just put the args into a new object[] and save ourselves a bit of public surface area, no?

Xamarin.Forms.Core/DependencyResolver.cs
```diff
@@ -0,0 +1,49 @@
...
+		internal static object Resolve(Type type, params object[] args)
+		{
+			return Resolver?.Invoke(type, args);
+		}
+
+		internal static object ForceResolve(Type type, params object[] args)
+		{
+			var result = Resolve(type, args);
+
+			if (result != null) return result;
```
kingces95: Suggest we throw if Type is not assignable from the result.

Xamarin.Forms.Core/Effect.cs
```diff
@@ -20,11 +20,10 @@ internal Effect()
 
 		public static Effect Resolve(string name)
 		{
-			Type effectType;
 			Effect result = null;
-			if (Internals.Registrar.Effects.TryGetValue(name, out effectType))
+			if (Internals.Registrar.Effects.TryGetValue(name, out Type effectType))
 			{
-				result = (Effect)Activator.CreateInstance(effectType);
+				result = DependencyResolver.ForceResolve(effectType) as Effect;
```
kingces95: Suggest considering cast instead of `as`; The result should either be of `effectType` or null, no?

Xamarin.Forms.Core/DependencyResolver.cs
```diff
@@ -0,0 +1,49 @@
...
+
+			if (args.Length > 0)
```
kingces95: I'm not sure what case the following code handles; Suggest we add unit tests to cover 35:46.

docs/Xamarin.Forms.Core/Xamarin.Forms.Internals/DependencyResolver+ResolveDelegate.xml
```diff
@@ -0,0 +1,31 @@
+<Type Name="DependencyResolver+ResolveDelegate" FullName="Xamarin.Forms.Internals.DependencyResolver+ResolveDelegate">
```
kingces95: We might consider if some dev time now drafting some isense docs might preempt questions/confusion in the forums.

Xamarin.Forms.Core/DependencyResolver.cs
hartez: I went with `ResolveOrCreate`.

Xamarin.Forms.Core/DependencyResolver.cs
hartez: Done.

Xamarin.Forms.Core/DependencyResolver.cs
hartez: Done.

Xamarin.Forms.Core/Effect.cs
hartez: Yep. Done.

dansiegel:
hey guys can we please get this moved from vNext+1 to vNext... would really like to see this in v3.0

jbachelor:
+1 for the suggestion from @dansiegel to move this up to vNext. I know my team could definitely benefit from this.

agriffith:
We could use this sooner rather than later, as well.

PR #1873, opened 2018-02-13
[Android] Allow customizing Android MenuItem

Thetyne:
### Description of Change ###

Allow customizing Android MenuItem with tintcolor and custom drawable.

### API Changes ###
Added:
 - protected virtual void UpdateDrawableMenuItem(Context context, IMenuItem menuItem, ToolbarItem toolBarItem)

Changed:
- void HandleToolbarItemPropertyChanged(object sender, PropertyChangedEventArgs e) -> protected virtual void HandleToolbarItemPropertyChanged(object sender, PropertyChangedEventArgs e)

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
@@ -468,7 +468,7 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 				ResetToolbar();
 		}
 
-		void HandleToolbarItemPropertyChanged(object sender, PropertyChangedEventArgs e)
+		protected virtual void HandleToolbarItemPropertyChanged(object sender, PropertyChangedEventArgs e)
 		{
 			if (e.PropertyName == MenuItem.IsEnabledProperty.PropertyName || e.PropertyName == MenuItem.TextProperty.PropertyName || e.PropertyName == MenuItem.IconProperty.PropertyName)
 				UpdateMenu();
@@ -792,17 +792,35 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 				else
 				{
 					IMenuItem menuItem = menu.Add(item.Text);
-					FileImageSource icon = item.Icon;
-					if (!string.IsNullOrEmpty(icon))
-					{
-						Drawable iconDrawable = context.GetFormsDrawable(icon);
-						if (iconDrawable != null)
-							menuItem.SetIcon(iconDrawable);
-					}
+					UpdateDrawableMenuItem(context, menuItem, item);
 					menuItem.SetEnabled(controller.IsEnabled);
 					menuItem.SetShowAsAction(ShowAsAction.Always);
 					menuItem.SetOnMenuItemClickListener(new GenericMenuClickListener(controller.Activate));
 				}
+			}
+		}
+
+		/// <summary>
+		/// For override specify tint color menu or specify drawable menu
+		/// Sample:
+		/// base.UpdateMenuItem(...)
+		/// menuItem.Icon.Mutate();
+		/// menuItem.Icon.SetColorFilter(this.Resources.GetColor([ResourceColor]), Android.Graphics.PorterDuff.Mode.SrcIn);
+		/// 
+		/// Sample with drawable badge count:
+		/// menuItem.SetIcon(SetDrawableIcon(menuItem.Icon)));
+		/// 
+		/// In this fact, toolBarItem is necessary for access to custom property like badge count
+		/// </summary>
+		/// <param name="menuItem"></param>
+		protected virtual void UpdateDrawableMenuItem(Context context, IMenuItem menuItem, ToolbarItem toolBarItem)
+		{
+			FileImageSource icon = toolBarItem.Icon;
+			if (!string.IsNullOrEmpty(icon))
+			{
+				Drawable iconDrawable = context.GetFormsDrawable(icon);
+				if (iconDrawable != null)
+					menuItem.SetIcon(iconDrawable);
 			}
 		}
 
--- a/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
@@ -807,7 +807,10 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 			{
 				Drawable iconDrawable = context.GetFormsDrawable(icon);
 				if (iconDrawable != null)
+				{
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
```diff
@@ -792,20 +792,38 @@ void UpdateMenu()
...
+		/// menuItem.Icon.Mutate();
+		/// menuItem.Icon.SetColorFilter(this.Resources.GetColor([ResourceColor]), Android.Graphics.PorterDuff.Mode.SrcIn);
+		/// 
+		/// Sample with drawable badge count:
+		/// menuItem.SetIcon(SetDrawableIcon(menuItem.Icon)));
+		/// 
+		/// In this fact, toolBarItem is necessary for access to custom property like badge count
+		/// </summary>
+		/// <param name="menuItem"></param>
+		protected virtual void UpdateDrawableMenuItem(Context context, IMenuItem menuItem, ToolbarItem toolBarItem)
```
hartez: I think this should be renamed to `UpdateMenuItemIcon`.

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
```diff
@@ -468,7 +468,7 @@ void CurrentOnPropertyChanged(object sender, PropertyChangedEventArgs e)
 				ResetToolbar();
 		}
 
-		void HandleToolbarItemPropertyChanged(object sender, PropertyChangedEventArgs e)
+		protected virtual void HandleToolbarItemPropertyChanged(object sender, PropertyChangedEventArgs e)
```
hartez: This doesn't need to be `protected virtual` for the user to customize the behavior of `UpdateDrawableMenuItem`; this should be private.

Thetyne: I need this because if i create my own ToolBarItem with a new property, i need recreate menu.

```
public class CartCountToolBarItem : ToolBarItem
{
static readonly BindableProperty CartCountProperty = .....
}
```

i need call UpdateMenu when this property change for display cart count inside menu item.

```
protected override void HandleToolbarItemPropertyChanged(object sender, PropertyChangedEventArgs e)
		{
base.HandleToolbarItemPropertyChanged(sender, e);

			if (e.PropertyName == CartCountProperty.PropertyName)
				UpdateMenu();
		}
```

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
```diff
@@ -792,20 +792,38 @@ void UpdateMenu()
...
+		/// base.UpdateMenuItem(...)
+		/// menuItem.Icon.Mutate();
+		/// menuItem.Icon.SetColorFilter(this.Resources.GetColor([ResourceColor]), Android.Graphics.PorterDuff.Mode.SrcIn);
+		/// 
+		/// Sample with drawable badge count:
+		/// menuItem.SetIcon(SetDrawableIcon(menuItem.Icon)));
+		/// 
+		/// In this fact, toolBarItem is necessary for access to custom property like badge count
+		/// </summary>
+		/// <param name="menuItem"></param>
```
jassmith: Please don't add inline XML docs, they are (unfortunately) not allowed

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
jassmith: To remain consistent this should be named OnToolbarItemPropertyChanged as our protected virtuals should be.

I am not a huge fan of propagating INPC further but its the hammer we have.

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
```diff
@@ -792,20 +792,38 @@ void UpdateMenu()
...
+		protected virtual void UpdateDrawableMenuItem(Context context, IMenuItem menuItem, ToolbarItem toolBarItem)
+		{
+			FileImageSource icon = toolBarItem.Icon;
+			if (!string.IsNullOrEmpty(icon))
+			{
+				Drawable iconDrawable = context.GetFormsDrawable(icon);
```
jassmith: probably can be disposed after use to clear up GREF

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
hartez: @jassmith So are you cool with this being `protected virtual`? If you're cool with it, then I'm cool with it :)

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
Thetyne: in your code, you don't dispose iconDrawable. Is it necessary?

Xamarin.Forms.Platform.Android/Platform.cs
Xamarin.Forms.Platform.Android/CellAdapter.cs

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
Thetyne: oki  i changed that

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
Thetyne: oki  i removed comment

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
Thetyne: what do you mean GREF?

jassmith: Its an android reference handle to a native java object that is created and tracked under the hood. It is quite expensive to keep them hanging around so for short lived objects disposal is ideal. It basically says to Android "okay Im done with this now, you can hand the object off to Java" instead of waiting for the GC to do that for us.

Making sure we don't creep GREFs is becoming increasingly needed. There are plenty of places where we screw this up for sure, as you already pointed out. It's not needed as you ask, but it does help make sure we don't end up reserving many thousands of GREFs and impact performance quite badly.

Thetyne: ok, i have test and change code.

PR #1875, opened 2018-02-13
[XamlC] Cache Resolve and ImportReference

### LINKED ISSUE #1848: Improve XAMLC Compilation performance
No idea how, stephane can figure that out. The goal is just to profile and improve XAMLC performance.

StephaneDelcroix:
### Description of Change ###

[XamlC] Cache Resolve and ImportReference

ImportReference with System.Reflection based argument is notoriously
slow on .NET. So we cache the results for those.

We do not cache the results for TypeReference, MethodReference or
FieldReference calls, as those are already fast (passthrough if the
reference was already imported), and they aren't valid as dictionary
keys (no concept of equatability).

While we're at it, we shave another few ms from Resolve(TypeReference)
calls as well.

As, on good days, it shaves up to 40% of XamlC time, we can say that it fixes #1848

### Bugs Fixed ###

 - fixes #1848

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/BindablePropertyReferenceExtensions.cs
+++ b/Xamarin.Forms.Build.Tasks/BindablePropertyReferenceExtensions.cs
@@ -26,7 +26,7 @@ namespace Xamarin.Forms.Build.Tasks
 												md.IsStatic &&
 												md.IsPublic &&
 												md.Parameters.Count == 1 &&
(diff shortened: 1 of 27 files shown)


jassmith:
```
1>      167 ms  ResolveAssemblyReference                   1 calls
1>      171 ms  GenerateDepsFile                           1 calls
1>      205 ms  JoinItems                                  8 calls
1>      267 ms  XamlCTask                                  1 calls
1>     1434 ms  Csc                                        1 calls
```

Seeeems good

PR #1889, opened 2018-02-15
Check IsEnabled on TextCell before executing command

hartez:
### Description of Change ###

Adds a check for `IsEnabled` before executing a Command from a TextCell.

### Bugs Fixed ###

* [60045 – ListView with RecycleElement strategy doesn't handle CanExecute of TextCell Command properly (Forms->iOS)](https://bugzilla.xamarin.com/show_bug.cgi?id=60045)

### API Changes ###

None

### Behavioral Changes ###

None 

### PR Checklist ###


kicsiede:
after this change i am still able to select a "disabled" row on iOS (it gets highlighted), although the tap event is not raised indeed.

PR #407, opened 2016-10-01
[iOS] Keyboard can get a little fancy with a customizable toolbar

adrianknight89:
### Description of Change

A platform specific functionality can allow developers to create a keyboard toolbar with any combination of `UIBarButtonSystemItem` buttons provided that space is available. 

See https://developer.apple.com/reference/uikit/uibarbuttonsystemitem

Created an enum to match `UIKit.UIBarButtonSystemItem`. Initially, I wanted to supply a dictionary of buttons, but `FlexibleSpace` can be used multiple times, hence the list. `KeyboardToolbarItem` has three properties, but more can be added if needed. Actions should be executable on the UI thread.

Sample code:

```
void Button_OnClicked(object sender, EventArgs e)
{
    var list = new List<KeyboardToolbarItem>
    {
        new KeyboardToolbarItem(UIBarButtonSystemItem.Action, () => { (sender as Button).BackgroundColor = (sender as Button).BackgroundColor == Color.Red ? Color.Green : Color.Red; }),
        new KeyboardToolbarItem(UIBarButtonSystemItem.FlexibleSpace),
        new KeyboardToolbarItem(UIBarButtonSystemItem.Camera, () => { DisplayAlert("Alert", "Camera tapped", "OK"); }),
        new KeyboardToolbarItem(UIBarButtonSystemItem.FlexibleSpace),
        new KeyboardToolbarItem(UIBarButtonSystemItem.Done, () => { Entry.Unfocus(); }, Color.Green)
    };

    Entry.On<iOS>().SetKeyboardToolbar(_flag ? list : null);

    _flag = !_flag;
}
```

**Initial UI without the toolbar**
img_1772

**Button sets toolbar with 3 items (Notice that Done button is green)**
img_1773

**Action item changes button background color**
img_1774

**Camera item shows an alert**
img_1775

**Done should unfocus the entry.**

Ideally, there should be some sort of unit test that ensures `Xamarin.Forms.PlatformConfiguration.iOSSpecific.UIBarButtonSystemItem` and `UIKit.UIBarButtonSystemItem` have the same values in case Apple decides to add new buttons to the list. There are no unit test projects for individual platforms? I only saw `Xamarin.Forms.Core.UnitTests`.

**Question:**

Should `Editor` have the same functionality? I did not see a renderer for `InputView` so this be a separate add.

PR #1896, opened 2018-02-15
[iOS] Add shadow effect

### LINKED ISSUE #1707: [Enhancement] Drop shadow support for iOS/UWP
## Rational

Drop shadows are fun, but rather platform specific in look/feel.

## API

Drop shadows in Android unfortunately are implemented via Elevation which causes other changes with regards to z-order and does not provide a consistent mechanism for creating a well defined shadow. Unfortunately this makes Android an ungraceful target at best for this feature.

For iOS:
```csharp
View.On<iOS>.SetShadowColor (Color color);
View.On<iOS>.SetShadowRadius (double radius);
View.On<iOS>.SetShadowOffset (Size offset);
View.On<iOS>.SetShadowOpacity (double opacity);
```

for UWP:
```csharp
Layout.On<Windows>.SetShadowColor (Color color);
Layout.On<Windows>.SetShadowRadius (double radius);
Layout.On<Windows>.SetShadowOffset (Size offset);
Layout.On<Windows>.SetShadowOpacity (double opacity);
```

## Expected Result

Setting the drop shadow on a view in iOS should cause that view to have a drop shadow. Setting the draw shadow on a layout in UWP should cause every view in that layout to have a drop shadow.

Ideally both sets of properties would operate on views, but UWP seems to prefer to perform these effects at the panel level.

### Android
No change

### iOS
Should be straightforward setting of CALayer properties.

### UWP
https://docs.microsoft.com/en-us/windows/uwpcommunitytoolkit/controls/dropshadowpanel

We will not want to use the dropshadowpanel itself, just the techniques it uses to produce the effect.

## Implications for CSS

While these properties are relevant to css, I think they should remain unsupported until a general support option can be added for all platforms.

## Backward Compatibility

The renderers on each platform should check BindableObject.GetIsDefault(); the behavior of the renderers should only be altered if the BindableProperty has been explicitly set. This should provide backward compatibility.

## Difficulty : Easy

No obvious implementation issues stick out to me.

jassmith:
### Description of Change ###

Adds the ability to set drop shadows for any arbitrary iOS control. For now I have descoped the UWP implementation as after digging in it appears to be much more involved than initially hoped. It does not seem likely it can be well encapsulated without having to affect renderers not using the new drop shadow API.

This does not mean we shouldn't do UWP at some point, just that the risk profile does not match the goals of the F100 project.

### Bugs Fixed ###

- fixes #1707

### API Changes ###

Added:
```csharp
View.On<iOS>.SetShadowColor (Color color);
View.On<iOS>.SetShadowRadius (double radius);
View.On<iOS>.SetShadowOffset (Size offset);
View.On<iOS>.SetShadowOpacity (double opacity);
```

### Behavioral Changes ###

There are no behavioral changes, just the additions.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1707.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1707.cs
@@ -1,0 +1,32 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using Xamarin.Forms.PlatformConfiguration;
+using Xamarin.Forms.PlatformConfiguration.iOSSpecific;
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 1707, "[Enhancement] Drop shadow support for iOS/UWP", PlatformAffected.iOS | PlatformAffected.UWP)]
+	public class Issue1707 : ContentPage
+	{
+		public Issue1707()
+		{
+			Title = "Shadow Test Page";
+			Padding = 10;
+
+			var layout = new StackLayout();
+			var box = new BoxView();
+			box.Color = Color.Aqua;
+			
+			box.On<iOS>().SetIsShadowEnabled(true);
+			box.On<iOS>().SetShadowOffset(new Size(10, 10));
+			box.On<iOS>().SetShadowColor(Color.Purple);
+
+			box.WidthRequest = box.HeightRequest = 100;
+
+			layout.Children.Add(box);
+
+			Content = layout;
+		}
+	}
+}
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1707.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1707.cs
@@ -6,7 +6,7 @@ using Xamarin.Forms.PlatformConfiguration.iOSSpecific;
 namespace Xamarin.Forms.Controls.Issues
 {
 	[Preserve(AllMembers = true)]
-	[Issue(IssueTracker.Github, 1707, "[Enhancement] Drop shadow support for iOS/UWP", PlatformAffected.iOS | PlatformAffected.UWP)]
+	[Issue(IssueTracker.Github, 1707, "[Enhancement] Drop shadow support for iOS", PlatformAffected.iOS)]
 	public class Issue1707 : ContentPage
 	{
 		public Issue1707()
--- a/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/VisualElement.cs
+++ b/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/VisualElement.cs
@@ -1,6 +1,7 @@
 
 namespace Xamarin.Forms.PlatformConfiguration.iOSSpecific
 {
+	using System;
 	using FormsElement = Forms.VisualElement;
 
 	public static class VisualElement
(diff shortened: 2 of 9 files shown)


jassmith:
and I need to update docs

adrianknight89:
@jassmith Weren't you guys planning to create some kind of effect library instead of baking all of these platform-specific enhancements into the main branch? I worked on #407 and a few others a long time ago, but back then, the team was more adamant about new PS enhancements. What's the new direction now? Accept any platform-specific?

jassmith:
Accept the ones in F100

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1707.cs
```diff
@@ -0,0 +1,32 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using Xamarin.Forms.PlatformConfiguration;
+using Xamarin.Forms.PlatformConfiguration.iOSSpecific;
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 1707, "[Enhancement] Drop shadow support for iOS/UWP", PlatformAffected.iOS | PlatformAffected.UWP)]
```
samhouts: No longer affects UWP

Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/VisualElement.cs
```diff
@@ -30,6 +31,161 @@ public static BlurEffectStyle GetBlurEffect(this IPlatformElementConfiguration<i
...
+		{
+			element.SetValue(ShadowColorProperty, value);
+		}
+
+		public static Color GetShadowColor(this IPlatformElementConfiguration<iOS, Xamarin.Forms.VisualElement> config)
+		{
+			return GetShadowColor(config.Element);
+		}
+
+		public static IPlatformElementConfiguration<iOS, Xamarin.Forms.VisualElement> SetShadowColor(this IPlatformElementConfiguration<iOS, Xamarin.Forms.VisualElement> config, Color value)
```
samhouts: Convention is to use `FormsElement` instead of `Xamarin.Forms.VisualElement` here

jassmith: Thats the problem with using a template :P Will fix.

Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/VisualElement.cs
```diff
@@ -30,6 +31,161 @@ public static BlurEffectStyle GetBlurEffect(this IPlatformElementConfiguration<i
...
+		public class ShadowEffect : RoutingEffect
+		{
+			public ShadowEffect() : base("Xamarin.ShadowEffect") { }
+		}
+
+		public static readonly BindableProperty IsShadowEnabledProperty =
+			BindableProperty.Create("IsShadowEnabled", typeof(bool),
+			typeof(VisualElement), false, propertyChanged: OnIsShadowEnabledChanged);
+
+		private static void OnIsShadowEnabledChanged(BindableObject bindable, object oldValue, object newValue)
```
rmarinho: i would drop private

Xamarin.Forms.Platform.iOS/ShadowEffect.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.ComponentModel;
+using UIKit;
+using PlatformElement = Xamarin.Forms.PlatformConfiguration.iOSSpecific.VisualElement;
+
+
+namespace Xamarin.Forms.Platform.iOS
+{
+	internal class ShadowEffect : PlatformEffect
```
rmarinho: Are all our effects internal?

rmarinho:
IOS failures not related, good to be merged.

Xamarin.Forms.Platform.iOS/ShadowEffect.cs
jassmith: no reason not to make it internal :)

PR #1899, opened 2018-02-16
[XamlC] no longer use any reflection-base ImportReference

StephaneDelcroix:
### Description of Change ###

[XamlC] no longer use any reflection-base ImportReference

reflection-base ImportReference -- we were using MethodBase and Type --
are importing the types present in the reflection context, at the time
of compilation. As the compilation happens on netstandard2.0, and our
assembly can now be netstandard1.0 again, those imported types are
failing to be resolved.

this changes always import references based on the assembly, or the
assembly references.

it might, or might not, give us another speed bump.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/ModuleDefinitionExtensions.cs
+++ b/Xamarin.Forms.Build.Tasks/ModuleDefinitionExtensions.cs
@@ -139,19 +139,34 @@ namespace Xamarin.Forms.Build.Tasks
 												classArguments?.Select(gp => module.GetTypeDefinition((gp.assemblyName, gp.clrNamespace, gp.typeName))).ToArray());
 		}
 
+		static Dictionary<(ModuleDefinition module, (string assemblyName, string clrNamespace, string typeName)), TypeDefinition> typeDefCache
+			= new Dictionary<(ModuleDefinition module, (string assemblyName, string clrNamespace, string typeName)), TypeDefinition>();
+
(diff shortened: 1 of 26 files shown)


jassmith:
Will get you benchmark times tomorrow

PureWeen:
For the latest PR 
I have this version installed
```
<PackageReference Include="Xamarin.Forms" Version="2.6.0.210032" />
```

and with that I'm able to compile netstandard 1.4 with the following build times
```
1>      124 ms  ResolveAssemblyReference                   1 calls
1>      137 ms  JoinItems                                  8 calls
1>      186 ms  Copy                                       2 calls
1>      232 ms  Csc                                        1 calls
1>     9808 ms  XamlCTask                                  1 calls
```

Xamarin.Forms.Build.Tasks/ModuleDefinitionExtensions.cs
```diff
@@ -1,32 +1,182 @@
...
+			if (exportedType != null) {
+				typeDef = exportedType.Resolve();
+				typeDefCache.Add((module, type), typeDef);
+				return typeDef;
+			}
+
+			//I hate you, netstandard
+			if (type.assemblyName == "mscorlib" && type.clrNamespace == "System.Reflection")
+				return module.GetTypeDefinition(("System.Reflection", type.clrNamespace, type.typeName));
+				throw new Exception($"Failed to get typedef for {type}");
```
samhouts: 👮 This LOOKS like this exception happens as part of this `if` clause because of the indentation.

PR #1902, opened 2018-02-16
Remove error XF001 since MSBuild already checks for double imports

kzu:
Importing twice the same targets might be a user authoring error, but it should 
hardly fail the build, unless it causes a real issue when the imported targets are 
actually from different versions, which is checked a few lines below in XF002.

MSBuild will already issue a warning for double imports, so there is already a 
non-silent way for users to fail the build on that case if they want to.

### DIFF
--- a/.nuspec/Xamarin.Forms.targets
+++ b/.nuspec/Xamarin.Forms.targets
@@ -27,11 +27,6 @@
 	</PropertyGroup>
 
 	<Target Name="_ValidateXFTasks" BeforeTargets="_CheckForInvalidConfigurationAndPlatform" Condition="'$(XFDisableTargetsValidation)' != 'True'">
-		<Error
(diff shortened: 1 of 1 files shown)


mhutch:
MSBuild will not warn if you import _multiple different versions_ of the same targets, because it only warns by path. IIRC this was introduced specifically because of cases seen in the wild where users ended up with multiple forms versions being imported into their projects (due to user error, IDE bugs, merge issues, whatever), as otherwise there was no indication of this problem, and the resulting errors were often very confusing.

Note that XF002 checks for the assembly's ABI, not the version of the targets themselves. It provides an actionable error in the case where the parameters of the task don't match the parameters expected by the tasks, which can happen when updating XF or when different projects reference different versions of XF. It doesn't check for an exact version, just for a number representing the last time the task names and parameters were added or changed.

jassmith:
Thank you for the feedback hutch, based on your comment I am going to close this PR.

Issue #1466, opened 2017-12-24
[Internal] Can't clone submodules when checking out Xamarin.Forms


scxp:
### Description

Checking out Xamarin.Forms project is unsuccessful because submodules can't be cloned. This used to work without users having to uncheck the option to clone submodules.

### Steps to Reproduce

1. Checkout Xamarin.Forms on VS for Mac or Windows
2. On Mac, the top level repository gets cloned but submodules can't. Windows fails completely.

scxp:
The issue is that https://github.com/xamarin/Xamarin.Forms.Build is returning 404. Please see the `.gitmodules` file in the solution directory.

kingces95:
Hm. I'm able to clone and build. I assume this is working in CI. 

Asked @rmarinho to pose a few questions to @scxp to see why it's not working locally for him.

rmarinho:
Yeah this will not work like in the past. you should skip submodules to build. nothing we can do here at the time.

scxp:
@rmarinho Can you explain why it won't work like it used to?

rmarinho:
@scxp we moved to have a submodule for internal build, release and ci integration, we had some issues mantaining scripts needed for Release per version with our CI, those scripts live now on submodule and we can have Xamarin.Forms versions pin specific commits of that submodule and CI always work.

PR #1926, opened 2018-02-19
Use relative URL to support recursive checkout in VSTS

kzu:
This is the norm in all other repos building in VSTS and
makes upstream build definitions and scripts much simpler.


scxp:
Why was #1466 closed then without a planned fix? Is this only for VSTS but not Git also?

PR #1941, opened 2018-02-21
Embed toolbox files on design dll

jassmith:
No API changes, just adds metadata to design dll


alanmcgovern:
@jassmith @davidortinau Not sure where's best to put this - but the format of the Xml file is going to be problematic for the toolbox images. VS and VSM use different sized images for their respective toolboxes so we cannot hardcode `-16` inside the image attribute https://github.com/xamarin/Xamarin.Forms/blob/b551753345c7e6cc8cb25bf7031935fb853077a0/Xamarin.Forms.Core.Design/toolbox/Xamarin.Forms.toolbox.xml#L10 ) I don't have a strong opinion on the right way to solve this, but a few options are:

1) Remove the pixel sizing from every image name and from the xml manifest. We will have a 16x16 `view-ActivityIndicator.png` in the win folder and a 48x48 `view-ActivityIndicator-16.png` in the mac folder. The xml file would just require `view-ActivityIndicator-16.png`. On mac we'll use the standard image loading magic to get the `view-ActivityIndicator@2x.png` or `view-ActivityIndicator~dark.png` variants as appropriate.

2) Use one attribute to specify the 'base name' for the image and a second attribute to specify the available sizes:

`<Item Class="Xamarin.Forms.ActivityIndicator" Image="view-ActivityIndicator" Sizes="16,48">`. 

From that we can load the appropriate one from the `win` or `mac` folder depending on which size the IDE prefers, and also which is available. I would expect the windows folder to only ship small icons, and VS will only use those anyway, and the mac folder would only ship large icons and it will only want to use those anyway.

3) Change `Image` to be `Images` and list the full names of all images, including platform. 

`<Item Class="Xamarin.Forms.ActivityIndicator" Images="win/view-ActivityIndicator-16.png,mac/view-ActivityIndicator-48.png">`. 

Then we could easily filter the list so it only contains the images for the platform we're on, and then we could select the appropriately sized image from the remainder (in this case there'll only be 1 to choose from).

vancura:
I don't know how these things work :)

Tell me how to rename the files and I'll do it, no prob.

alanmcgovern:
Let's wait for the final format to be decided on :)

jassmith:
Dont worry about renaming files, we only care about file sizes. I'll make sure they are named correctly.

vancura:
Just to illustrate what I mean:

This is iOS Designer:

1 2x

This is Android Designer:

2 2x

See? It works much better, IMHO.

@jassmith: Could you please show me (with a screenshot) how the Toolbox palette looks in Forms Designer? Of course, I can give you larger icons (I'd need to redesign them, so it would take some time), but if we can have a smaller palette, I'd be glad. Android Designer merely is visually more pleasing.

vancura:
Oh, what did I do? Was it me who closed the PR? I hope not!

jassmith:
Lol the PR was merged a while ago, we are just using this closed PR to do the discussion.

@alanmcgovern please confirm dropping the -16's from the xml is enough to get you guys happy? You can pick up the files by convention from there yes? Basically taking choice 1)

alanmcgovern:
Yes, perfect! This allows this leeway to change it easily in the future if we need icons of differing sizes for different use cases

alanmcgovern:
Just as an examples that might be relevant for forms itself,  we might display some of those icons in the design surface itself, Similar to how we display the icons for iOSExit items and gesture recognisers inside the surface

PR #1952, opened 2018-02-22
[UWP] Icon on TabbedPage

### LINKED ISSUE #1705: [Enhancement] Icon on TabbedPage UWP
## Rationale

Currently Icon are only surface onto iOS and Android and are not shown for the tabs on UWP. Ideally this behavior would work the same across all three platforms

## Implementation

The Forms side of the code is already implemented for this so the main effort is with the UWP TabbedPageRenderer to materialize those icons onto the tabs

### Android
Already works

### iOS
Already works

### UWP
Show icon images

## Backward Compatibility
If users have already implemented a custom renderer against the UWP TabbedPage that adds the icons what's the best way for this to be implemented to reduce that friction?

For UWP this will need to be turned on through Platform specific code
```
layout.On<Windows>()
```
Otherwise icons are just going to show up and surprise people

lateralusX:
### Description of Change ###

Implement icon support for TabbedPage on UWP.

Fixes #1705.

### API Changes ###

Added new Windows platform specific properties to enable/disable icons on tabbed page control as well as possibility to control icon size. 

Added:
- New bindable properties on Xamarin.Forms.PlatformConfiguration.WindowsSpecific.TabbedPage, "HeaderIconsEnabledProperty" and "HeaderIconsSizeProperty"
- TabbedPage.On<Windows>().IsHeaderIconsEnabled();
- TabbedPage.On<Windows>().SetHeaderIconsEnabled(true/false);
- TabbedPage.On<Windows>().GetHeaderIconsEnabled();
- TabbedPage.On<Windows>().EnableHeaderIcons();
- TabbedPage.On<Windows>().DisableHeaderIcons();
- TabbedPage.On<Windows>().SetHeaderIconsSize(true/false);
- TabbedPage.On<Windows>().GetHeaderIconsSize();

### Behavioral Changes ###

By default icons will remain disabled on tabbed page controls so current users shouldn't notice this change unless they explicitly enable it. One thing that does change for all users in TabbedPageStyle.xaml is the change from fixed PivotHeaderItem height to auto. This is needed in order to support different icon sizes without clipping. This shouldn't affect current users since they should already handled the case when headers get clipped, reducing the height to a value below old fixed header item size. If however a user relied on clipping behavior, this change will then affect their use of tabbed page control.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/TabbedPageRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/TabbedPageRenderer.cs
@@ -9,11 +9,20 @@ using Windows.UI.Xaml.Media;
 using Xamarin.Forms.Internals;
 using Xamarin.Forms.PlatformConfiguration.WindowsSpecific;
 using WGrid = Windows.UI.Xaml.Controls.Grid;
+using WTextAlignment = Windows.UI.Xaml.TextAlignment;
+using WHorizontalAlignment = Windows.UI.Xaml.HorizontalAlignment;
+using WVisibility = Windows.UI.Xaml.Visibility;
+using WStackPanel = Windows.UI.Xaml.Controls.StackPanel;
+using WImage = Windows.UI.Xaml.Controls.Image;
+using WTextBlock = Windows.UI.Xaml.Controls.TextBlock;
+using Specifics = Xamarin.Forms.PlatformConfiguration.WindowsSpecific.TabbedPage;
 
 namespace Xamarin.Forms.Platform.UWP
 {
     public class TabbedPageRenderer : IVisualElementRenderer, ITitleProvider, IToolbarProvider
     {
+        const string TabBarHeaderStackPanelName = "TabbedPageHeaderStackPanel";
+        const string TabBarHeaderImageName = "TabbedPageHeaderImage";
         const string TabBarHeaderTextBlockName = "TabbedPageHeaderTextBlock";
         const string TabBarHeaderGridName = "TabbedPageHeaderGrid";
 
@@ -21,6 +30,10 @@ namespace Xamarin.Forms.Platform.UWP
         Color _barTextColor;
         bool _disposed;
         bool _showTitle;
+
+        WTextAlignment _oldBarTextBlockTextAlignment;
+        WHorizontalAlignment _oldBarTextBlockHorinzontalAlignment;
+
         VisualElementTracker<Page, Pivot> _tracker;
 
         ITitleProvider TitleProvider => this;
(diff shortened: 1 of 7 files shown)


lateralusX:
Different screenshots before and after #1705. The first two tabs uses icons from the ControlGallery project while tab 3 and tab 4 test the scenario when no icon have been used for the page.

**Before #1705:**



**With #1705 with collapsed icons (default):**



**With #1705 with visible icons 16x16:**



**With #1705 with visible icons 32x32:**

lateralusX:
@jassmith Added screenshots and made smaller adjustments to make sure MinHeight was set on PivotHeaderItem to default fixed size. As you can see on screenshot 1 and 2 they now look the same.

Xamarin.Forms.Platform.UAP/TabbedPageRenderer.cs
```diff
@@ -333,5 +352,80 @@ void UpdateToolbarPlacement()
...
+                            // Update text block.
+                            var tabBarTextblock = stackPanelItem as WTextBlock;
+                            if (tabBarTextblock.GetValue(FrameworkElement.NameProperty).ToString() == TabBarHeaderTextBlockName)
+                            {
+                                if (headerIconsEnabled)
+                                {
+                                    // Remember old values so we can restore them if icons are collapsed.
+                                    // NOTE, since all Textblock instances in this stack panel comes from the same
+                                    // style, we just keep one copy of the value (since they should be identical).
+                                    _oldBarTextBlockTextAlignment = tabBarTextblock.TextAlignment;
```
PureWeen: One thing I noticed here when messing with your sample is that initially these two *old* values are 
Left and Stretch

Then when you set the size to something larger like (170,170) for the icons they become
Center and Center

And now they are just Center and Center forever and they won't go back to Stretch and Left

You don't really see the result of this because the surrounding stackpanel constrains the elements so these alignments don't really have any effect.  To see the effect if you set the containing stackpanel to a larger width
```
<StackPanel Orientation="Vertical" Name="TabbedPageHeaderStackPanel" Background="Pink" Width="100">
```

Then do
SetSize to some non larger non default (170,170) => Set size back to default => hide icons 
You'll now see the text is centered whereas initially it was left aligned

It's a bit edge case but I wanted to mention it and see what you think

lateralusX:
Thanks for feedback, will address comments. I noticed the previous file TabbedPageRender actually included space instead of tabs that messed things up a bit, but I aligned to what's already in that file and didn't convert it as part of this PR. I can do that overall formatting change as part of this PR as well.

Xamarin.Forms.Platform.UAP/TabbedPageRenderer.cs
lateralusX: Will investigate this edge case, thanks for spotting!

lateralusX:
@PureWeen Good to go?

PureWeen:
Yes :-)

PR #1960, opened 2018-02-22
Remove temporary key PFX files from UWP projects

hartez:
### Description of Change ###

Remove the PFX files from the UWP projects - they aren't required for building and running locally, we don't actually need to create packages from them right now, and they keep expiring and breaking the build.


hartez:
Closing this, need to target a different branch.

PR #1969, opened 2018-02-23
Simplify code when invoking Appearing/Disappearing events

kzu:
This is effectively equivalent to the original code, yet nicer to read :)

This is also how it's being done elsewhere (i.e. fast renderers, 
BindableObject, etc.), probably because that code is newer?


kzu:
Superseded by #1971

PR #1970, opened 2018-02-23
Add Application.PageAppearing/Dissapearing events

kzu:
In scenarios where you want to keep track of pages as they are
appearing on screen, it's useful to be able to subscribe to a
global event from the `Application` to track those pages for
later processing (i.e. showing a timeline for the app, reloading
XAML for some of those pages, maybe replaying the timeline, etc.)

The new events are automatically raised from the Page base class
right after sending the page-level Appearing/Disappearing events.

### DIFF
--- a/Xamarin.Forms.Core/Application.cs
+++ b/Xamarin.Forms.Core/Application.cs
@@ -169,6 +169,10 @@ namespace Xamarin.Forms
 
 		public event EventHandler<ModalPushingEventArgs> ModalPushing;
 
+		public event EventHandler<Page> PageAppearing;
+
+		public event EventHandler<Page> PageDisappearing;
+
 		public async Task SavePropertiesAsync()
 		{
 			if (Device.IsInvokeRequired)
@@ -278,6 +282,12 @@ namespace Xamarin.Forms
 
 			return properties;
 		}
+
+		internal void OnPageAppearing(Page page)
+			=> PageAppearing?.Invoke(this, page);
+
+		internal void OnPageDisappearing(Page page)
+			=> PageDisappearing?.Invoke(this, page);
 
 		void OnModalPopped(Page modalPage)
 		{
--- a/Xamarin.Forms.Core/Page.cs
+++ b/Xamarin.Forms.Core/Page.cs
@@ -317,6 +317,8 @@ namespace Xamarin.Forms
 
 			var pageContainer = this as IPageContainer<Page>;
 			pageContainer?.CurrentPage?.SendAppearing();
+
+			FindApplication(this)?.OnPageAppearing(this);
 		}
 
         [EditorBrowsable(EditorBrowsableState.Never)]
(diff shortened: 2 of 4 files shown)


kzu:
Hm, changing the base actually brought more commits than needed. Fixed now.

Xamarin.Forms.Core/Application.cs
```diff
@@ -169,6 +169,10 @@ public ResourceDictionary Resources
 
 		public event EventHandler<ModalPushingEventArgs> ModalPushing;
 
+		public event EventHandler<Page> PageAppearing;
```
StephaneDelcroix: it those events aren't meant to be consumed by apps, but by tools, make them internal, access then through reflection.

Xamarin.Forms.Core/Page.cs
```diff
@@ -317,6 +317,8 @@ public void SendAppearing()
 
 			var pageContainer = this as IPageContainer<Page>;
 			pageContainer?.CurrentPage?.SendAppearing();
+
+			Application.Current?.OnPageAppearing(this);
```
StephaneDelcroix: it shouldn't be raised on the `Application.Current`, but on the actual `Application` instance on which the page appears (or disappears)

samhouts:
Related to https://forums.xamarin.com/discussion/84510/improved-life-cycle-support

Xamarin.Forms.Core/Application.cs
kzu: agreed with @jassmith that it should be public since other customers have requested such a feature in the past.

Xamarin.Forms.Core/Page.cs
kzu: I don't see a `this.Application` or the like. What's the API I should be using instead?

kzu: Also, since new App() [sets it as current](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/Application.cs#L32) I wonder what's the problem of doing it this way?

Is it even supported to have multiple `Application` instances? None of our templates suggest so...

jassmith: When you have a page and want the application, use Page.parent.parent.parent blah blah blah until you find the Application or strike null. If you strike null there is no applciation because you are either not parented (in which case something really fucky has happened) or you are inside a pre-Application based Forms app in which case you shouldn't trigger those events anyway.

kzu: Fixed! Thanks 👍

jassmith:
waiting for builds :)

Xamarin.Forms.Core/Page.cs
```diff
@@ -337,6 +339,16 @@ public void SendDisappearing()
...
+
+			FindApplication(this)?.OnPageDisappearing(this);
+		}
+
+		Application FindApplication(Element element)
+		{
+			if (element == null)
+				return null;
+
+			return (element.Parent is Application app) ? app : FindApplication(element.Parent);
```
StephaneDelcroix: you can avoid recursion by using a `while` loop

kyle-john: How about using `Application.Current`?

PR #1973, opened 2018-02-26
[Android] attempt to load images from drawables before using bitmap

### LINKED ISSUE #1908: [Android]  Use drawables from resources opposed to creating bitmaps in order to improve performance
### Description
@jonathanpeppers outlines some issues here with how images are loaded with Android

http://jonathanpeppers.com/Blog/xamarin-forms-performance-on-android

> Notice how many Android.Graphics.Bitmaps get created here... If two rows in a ListView use the same image, they each use a copy of the exact same image. If you scroll a cell off screen and bring it back, it will load a new Bitmap object when it is brought back onto the screen. 

On android if we use a Drawable when present instead of the bitmap this helps with reuse of the same image.

Also looking into caching the Resource Ids against the names might be useful as the ResourceManager currently uses reflection

### Steps to Reproduce
Create a large number of images all based on the same Drawable in Android
```
StackLayout listView = new StackLayout();
             
            for (int i = 0; i < 1000; i++)
            {
                listView.Children.Add(new Image(){ Source = "oasis.jpg"});                
            }

            return new ContentPage()
            {
                Content = new ScrollView(){ Content = listView}
            };
```
Currently this will crash with an out of memory exception. 

### Expected Behavior
Use drawables where you can instead of constructing  bitmaps so listviews can handle lots of repetative images.

### Actual Behavior
Crashes with out of memory when using too many images

### Implementation

I've worked up a PoC here
https://github.com/xamarin/Xamarin.Forms/tree/ImagePerf

### Backwards Compatibility
Does this need to be a platform opt in? or will it work just dropped in. What dependencies are there if the ImageView has always been set via a bitmap

If the user isn't using Drawable does the extra check for a Drawable cause them a performance decrease for the lookup of the ID? 

Can the names inside ResourceManager all just be cached against ids? 

Reflection performance vs 
https://developer.android.com/reference/android/content/res/Resources.html#getIdentifier(java.lang.String,%20java.lang.String,%20java.lang.String)

PureWeen:
### Description of Change ###

If the provided source for an image is a FileImageSource then attempt to resolve the image as a drawable via GetResource opposed to creating a bitmap from the file itself.  This requires a lot less memory.  With the included Unit Test I can load 1000 Xamarin Forms Images without causing Out Of Memory Exceptions now. 

### Bugs Fixed ###

fixes #1908

### Behavioral Changes ###

This change was done in such a way that behavior should remain the same outside of some performance gains when an image is used in multtiple places

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
+++ b/Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Threading.Tasks;
 using Android.Graphics;
+using Android.Graphics.Drawables;
 using AImageView = Android.Widget.ImageView;
 
 namespace Xamarin.Forms.Platform.Android
@@ -29,17 +30,27 @@ namespace Xamarin.Forms.Platform.Android
 
 			ImageSource source = newImage?.Source;
 			Bitmap bitmap = null;
+			Drawable drawable = null;
+
 			IImageSourceHandler handler;
 
 			if (source != null && (handler = Internals.Registrar.Registered.GetHandlerForObject<IImageSourceHandler>(source)) != null)
 			{
-				try
+				if (handler is FileImageSourceHandler)
 				{
-					bitmap = await handler.LoadImageAsync(source, imageView.Context);
+					drawable = imageView.Context.GetDrawable((FileImageSource)source);
 				}
-				catch (TaskCanceledException)
+
+				if (drawable == null)
 				{
-					imageController?.SetIsLoading(false);
+					try
+					{
+						bitmap = await handler.LoadImageAsync(source, imageView.Context);
+					}
+					catch (TaskCanceledException)
+					{
+						imageController?.SetIsLoading(false);
+					}
 				}
 			}
 
@@ -51,8 +62,10 @@ namespace Xamarin.Forms.Platform.Android
 
 			if (!imageView.IsDisposed())
 			{
-				if (bitmap == null && source is FileImageSource)
-					imageView.SetImageResource(ResourceManager.GetDrawableByName(((FileImageSource)source).File));
+				if (bitmap == null && drawable != null)
+				{
+					imageView.SetImageDrawable(drawable);
+				}
 				else
 				{
 					imageView.SetImageBitmap(bitmap);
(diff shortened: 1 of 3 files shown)


rmarinho:
You should just that a empty test that takes a screenshot and says manual review, that way we make sure this is tested on UITests

Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
```diff
@@ -29,17 +30,28 @@ internal static class ImageViewExtensions
...
+
 			IImageSourceHandler handler;
 
 			if (source != null && (handler = Internals.Registrar.Registered.GetHandlerForObject<IImageSourceHandler>(source)) != null)
 			{
-				try
+				if (handler is FileImageSourceHandler)
 				{
-					bitmap = await handler.LoadImageAsync(source, imageView.Context);
+					drawable =
```
jassmith: newline seems unneeded...

Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
```diff
@@ -29,17 +30,28 @@ internal static class ImageViewExtensions
...
 			ImageSource source = newImage?.Source;
 			Bitmap bitmap = null;
+			Drawable drawable = null;
+
 			IImageSourceHandler handler;
 
 			if (source != null && (handler = Internals.Registrar.Registered.GetHandlerForObject<IImageSourceHandler>(source)) != null)
 			{
-				try
+				if (handler is FileImageSourceHandler)
```
jassmith: Should check source and be done as
```
if (source is FileImageSource fileSource)
    imageView.Context.GetDrawable(fileSource);
```
This avoid extra cast checks and also if they did something fucky with their handlers we wont get a bad cast.

PureWeen: @jassmith  So the thinking here was for backwards compatibility. FileImageSourceHandler is a sealed class so if the handler is of a different type that means the user registered their own handler to process the FileImageSource and my new code shouldn't be executed here..  Ideally we would just be changing IImageSourceHandler to return a Drawable and move all my code here into FileImageSourceHandler but since that would change the APIs I figured this fix would be a good performance gain for now without burdening the users

jassmith: Good point

PR #1974, opened 2018-02-26
ImageButton

### LINKED ISSUE #1724: [Enhancement] ImageButton
# Rationale

The current Button.Image property is mainly for displaying a small icon next to the text. 

To avoid breaking changes, we should create a new ImageButton control. Both Button and ImageButton should derive from a common ButtonBase.

# Implementation

see Button API, but skip the Label

Implementation on Android, iOS, UWP in renderers.

# Backward Compatibility

# Difficulty: Medium

PureWeen:
### Description of Change ###

Implementation of new Xamarin Forms Control ImageButton

The behavior of the ImageButton emulates all the APIs on the Image and applies the same behavior to the ImageButton. It implements all of the APIs of Button sans anything related to TEXT

This PR is a little heavy handed and parts of it where I apply some generic behaviors to other controls could be moved out to different PRs 

### Issues Resolved ### 
- fixes #1724

#### Code Changes
In this PR I needed to merge the behavior of Image and Button to form ImageButton. I took a composition approach since I needed to merge the behaviors of the two controls.

I started with each control (Button, Image) and refactored out the behavior at the forms and renderer level into into static managers. Once this was done I then just applied the necessary interfaces and managers to ImageButton.

This also will help to centralize some behavior like Commanding and Image Loading.  Image Loading was basically the same code copied between Image and Button Image. Not wanting to just copy it a third place I just centralized it into one of the static managers and then just call it from all three locations.

This would be one area that would make this PR simpler. I could rollback all the changes I made to non ImageButton components and then apply the static managers to those in a separate PR.  As it stands since the behavior is shared a lot of the tests that were written against Button and Image are now transitively applicable to ImageButton

#### Formatting Changes
Because I was touching the files I applied a few sweeping formatting changes which unfortunately will make some changes a bit hard to isolate but since I touched those files I wanted to get the coding standards of said files consistent with the rest of the project

### API Changes ###

There shouldn't be any general API changes.  I added a decent amount of things to the Controller interfaces but I explicitly implemented them all so they don't just show up on the general controls

*ImageButton*  

```c#
public class ImageButton
{

public Color BorderColor {get; set;}

//Once the BoxView radius changes are rolled in it might make sense to just upgrade this
//to the changes there and use that struct and behavior
public int CornerRadius  {get; set;}  //BP
public double BorderWidth {get; set;} //BP
public Aspect Aspect {get; set;} //BP
public bool IsLoading {get; } //BP
public bool IsPressed {get;} //BP
public bool IsOpaque {get; set;} //BP
public ICommand Command {get; set;} //BP
public object CommandParameter {get; set;} //BP
public ImageSource Source {get; set;} //BP
public Color BorderColor {get; set;} //BP

public event EventHandler Clicked;
public event EventHandler Pressed;
public event EventHandler Released;

```

*Button*  

```c#
public bool IsPressed {get;} //BP
public event EventHandler Pressed;
```

The Pressed state was added to the Visual State Manager and implemented for this control and Button

I leveraged the shadow changes that Sam made to the Background Tracker code on android and exposed some platform specific apis to add shadows to ImageButtons
```c#
imageButton.On<Android>().SetShadowColor(Color.Black);
imageButton.On<Android>().SetShadowRadius(1); 
imageButton.On<Android>().SetIsShadowEnabled(true);            
imageButton.On<Android>().SetShadowOffset(new Size(10, 10));
```

~They won't quite work correctly until this PR is merged in
https://github.com/xamarin/Xamarin.Forms/pull/2114~

### Behavioral Changes ###

There should not be any behavior changes and these changes should be backwards compatible with custom renderer's. The change are mostly localized around private and internal methods and classes.

### PR Checklist ###


Xamarin.Forms.Controls/CoreGalleryPages/ImageButtonCoreGalleryPage.cs
```diff
@@ -101,11 +101,12 @@ protected override void Build(StackLayout stackLayout)
 			Add(isLoadingContainer);
 			Add(isOpaqueContainer);
 			Add(sourceContainer);
+
 			Add(borderButtonContainer);
-			Add(borderRadiusContainer);
-			Add(borderWidthContainer);
+            Add(borderWidthContainer);
```
jassmith: Yer whitespace is all screwed up now

Xamarin.Forms.Core.UITests.Shared/Tests/ImageButtonUITests.cs
```diff
@@ -69,22 +69,8 @@ protected override void FixtureTeardown()
...
-			remote.GoTo();
-		}
-
-		[Test]
-		[UiTest(typeof(ImageButton), "CornerRadius")]
-		[UiTestBroken(BrokenReason.CalabashAndroidUnsupported, "Figure out how to get Android Drawables")]
-        public void CornerRadius()
-		{
-            var remote = new ViewContainerRemote(App, Test.ImageButton.CornerRadius, PlatformViewType);
+            var remote = new ViewContainerRemote(App, Test.ImageButton.BorderColor, PlatformViewType);
```
jassmith: same here, gonna stop calling out whitespace now

jassmith:
I dont know where those images came from but you need to replace them with something you make. I dont want to find out that the image gives us copyright issues in the future.

PureWeen:
@jassmith   I just copied them from the other Control Gallery projects. Those files just weren't present in the UWP project

rmarinho:
Failing one test Legacy renderers android

PureWeen:
@rmarinho   I'll check it out.. I had noticed that same test was failing in a different PR so I wasn't sure if that one was just temperamental sometimes

PureWeen:
rebased to master with docs changes and a small Visual Element merge
the three failing unit tests aren't from this PR but are present on master

AndreiMisiukevich:
Hi
You may look through this project

it's something similar with 
https://github.com/AndreiMisiukevich/TouchEffect

PureWeen:
build --uitests

PureWeen:
build --uitests

PureWeen:
build --uitests

PureWeen:
build --uitests

PureWeen:
build --uitests

PureWeen:
build --uitests

samhouts:
@PureWeen Please rebase!

Xamarin.Forms.Core/BindableValueChangedEventArgs.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Xamarin.Forms
+{
+	public class BindableValueChangedEventArgs : EventArgs
```
rmarinho: Is this used somewhere?

Xamarin.Forms.Core/Button.cs
```diff
@@ -46,28 +46,34 @@ public class Button : View, IFontElement, ITextElement, IBorderElement, IButtonC
...
 
-		void IPaddingElement.OnPaddingPropertyChanged (Thickness oldValue, Thickness newValue)
+		void IPaddingElement.OnPaddingPropertyChanged(Thickness oldValue, Thickness newValue)
 		{
 			InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
 		}
 
+
+		internal static readonly BindablePropertyKey IsPressedPropertyKey = BindableProperty.CreateReadOnly(nameof(IsPressed), typeof(bool), typeof(Button), default(bool));
+		public static readonly BindableProperty IsPressedProperty = IsPressedPropertyKey.BindableProperty;
```
rmarinho: This is new on Button no? Maybe add a UITest for it? and we need that on the notes

Xamarin.Forms.Core/Button.cs
```diff
@@ -338,6 +314,58 @@ void IBorderElement.OnBorderColorPropertyChanged(Color oldValue, Color newValue)
 		{
 		}
 
+		private void OnImageSourcesSourceChanged(object sender, EventArgs e) =>
```
rmarinho: Think you need to configure this rule to drop private on your IDE

Xamarin.Forms.Platform.iOS/ViewRenderer.cs
```diff
@@ -36,12 +36,38 @@ public abstract class ViewRenderer<TView, TNativeView> : VisualElementRenderer<T
...
 			return default(TNativeView);
 		}
 
 		public TNativeView Control { get; private set; }
+		NativeView IVisualNativeElementRenderer.Control => Control;
+
+
+		event EventHandler<PropertyChangedEventArgs> IVisualNativeElementRenderer.ElementPropertyChanged
+		{
+			add { _elementPropertyChanged += value; }
```
rmarinho: maybe use WeakEventManager

Xamarin.Forms.Platform.iOS/ViewRenderer.cs
```diff
@@ -36,12 +36,38 @@ public abstract class ViewRenderer<TView, TNativeView> : VisualElementRenderer<T
 #endif
 		NativeColor _defaultColor;
 
+		event EventHandler<PropertyChangedEventArgs> _elementPropertyChanged;
```
rmarinho: Do we really need this can we just use the override? Why ?

Xamarin.Forms.Platform.iOS/Renderers/ImageElementManager.cs
```diff
@@ -0,0 +1,153 @@
...
+			{
+				return;
+			}
+
+			Control.Opaque = imageElement.IsOpaque;
+		}
+
+		public static async Task SetImage(IImageVisualElementRenderer renderer, IImageController imageElement, Image oldElement = null)
+		{
+			var Element = renderer.Element;
```
rmarinho: renderer null check ?

Xamarin.Forms.Platform.iOS/Renderers/ImageButtonRenderer.cs
```diff
@@ -0,0 +1,146 @@
...
+			ImageElementManager.Init(this);
+		}
+
+		public override SizeF SizeThatFits(SizeF size)
+		{
+			var result = base.SizeThatFits(size);
+
+			if (result.Height < _minimumButtonHeight)
+			{
+				result.Height = _minimumButtonHeight;
```
rmarinho: what if  want a really small button?

Xamarin.Forms.Platform.iOS/Renderers/ImageButtonRenderer.cs
```diff
@@ -0,0 +1,146 @@
...
+			if (string.IsNullOrWhiteSpace(elemValue) && Control?.AccessibilityLabel == Control?.Title(UIControlState.Normal))
+				return;
+
+			base.SetAccessibilityLabel();
+		}
+
+		bool IImageVisualElementRenderer.IsDisposed => _isDisposed;
+		void IImageVisualElementRenderer.SetImage(UIImage image)
+		{
+			Control.SetImage(image.ImageWithRenderingMode(UIImageRenderingMode.AlwaysOriginal), UIControlState.Normal);
```
rmarinho: I got a NRE here by use a wrong url , so we might need to check image is null here. Also I think we can pass to clear the Control image ?!

Xamarin.Forms.Core/Image.cs
```diff
@@ -54,122 +54,58 @@ public ImageSource Source
...
-				SetInheritedBindingContext(newvalue, BindingContext);
+				newSource.SourceChanged += image.OnImageSourcesSourceChanged;
 			}
-
-			InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
+			ImageElementManager.ImageSourceChanged(bindable, newSource);
 		}
 
-		static void OnSourcePropertyChanging(BindableObject bindable, object oldvalue, object newvalue)
+		private static void OnImageSourceChanging(BindableObject bindable, object oldValue, object newValue)
```
jassmith: Spurious private

Xamarin.Forms.Core/PlatformConfiguration/AndroidSpecific/ImageButton.cs
```diff
@@ -0,0 +1,109 @@
...
+
+		public static IPlatformElementConfiguration<Android, FormsImageButton> SetShadowColor(this IPlatformElementConfiguration<Android, FormsImageButton> config, Color value)
+		{
+			SetShadowColor(config.Element, value);
+			return config;
+		}
+
+		public static readonly BindableProperty ShadowRadiusProperty =
+			BindableProperty.Create("ShadowRadius", typeof(double),
+			typeof(VisualElement), 10.0);
```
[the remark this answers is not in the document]
jassmith: Im guessing so you dont have to set a radius by defualt

Xamarin.Forms.Platform.Android/Cells/BaseCellView.cs
```diff
@@ -192,7 +192,7 @@ async void UpdateBitmap(ImageSource source, ImageSource previousSource = null)
 		{
 			try
 			{
-				await _imageView.UpdateBitmap(null, source, null, previousSource);
+				await _imageView.UpdateBitmap(source, previousSource).ConfigureAwait(false);
```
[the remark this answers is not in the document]
jassmith: last line executed line in the method, this should prevent a useless marshalling

jassmith:
Seems to work well on testing. Needs rebase. No major issues just some minor cleanup.

jassmith:
Oops meant to approve :P

Xamarin.Forms.Core/PlatformConfiguration/AndroidSpecific/ImageButton.cs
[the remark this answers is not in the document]
PureWeen: TBH this all just came from copying the same properties that are on the iOS side that @jassmith  added when he added these properties on iOS

https://github.com/xamarin/Xamarin.Forms/blob/bd31e1e9fc8b2f9ad94cc99e0c7ab058174821f3/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/VisualElement.cs#L114

Xamarin.Forms.Platform.iOS/Renderers/ImageButtonRenderer.cs
PureWeen: This is what we use for the minimum desired height on iOS buttons as well
https://github.com/xamarin/Xamarin.Forms/blob/bd31e1e9fc8b2f9ad94cc99e0c7ab058174821f3/Xamarin.Forms.Platform.iOS/Renderers/ButtonRenderer.cs#L27

The constraints still get set to really low numbers if you specify them though and that's how they get measured out

Example
image

Xamarin.Forms.Platform.iOS/ViewRenderer.cs
PureWeen: This is the event that the *Managers* hook into

https://github.com/xamarin/Xamarin.Forms/pull/1974/files#diff-f7da8c7c17ec3976ea074bebce797504R12

This way one manager can just hook into this event and then apply the effects of that change back to the same control. This is why this PR removes a lot of the ElementPropertyChanged watching into the managers so the behavior can be generalized

Xamarin.Forms.Platform.iOS/ViewRenderer.cs
PureWeen: I think that might add too much overhead without enough gain.  

This is primarily used internally by the static managers which get unhooked from the dispose. Users could tie into this by casting the renderer to the interface but then it'd be on them to unwire it (which makes sense).  I think WeakEventManagers make more sense at the Core level where we are the ones hooking into events on user owned classes i.e. Commands and INCC on User controlled lists

Xamarin.Forms.Platform.Android/FastRenderers/ImageButtonRenderer.cs
```diff
@@ -0,0 +1,248 @@
...
+		// Button related
+		void IOnClickListener.OnClick(AView v) =>
+			ButtonElementManager.OnClick(Button, Button, v);
+
+		bool IOnTouchListener.OnTouch(AView v, MotionEvent e) =>
+			ButtonElementManager.OnTouch(Button, Button, v, e);
+		// Button related
+
+
+		float IBorderVisualElementRenderer.ShadowRadius => Context.ToPixels(Button.OnThisPlatform().GetShadowRadius());
```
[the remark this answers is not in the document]
PureWeen: Alright I moved these to a method and have it save the value after the first call

PureWeen:
alright @rmarinho  and @jassmith  made the suggested changes

PureWeen:
Actually @rmarinho  I just remembered your additional general notes

> The border on ImageButton on Android is below the image, while on iOS is over the image, I prefer the iOS behaviour.

Yea and UWP is different then both of them as well :-/  I'll need to look at this a bit more and see if I can match them up.  This issue also exists if you use a Button and set the Image on it. For the most part I've been using Android as the source of truth as it's the only platform that actually has an ImageButton. Though I do agree that the iOS behavior would seem more desirable since you might be fine with a bit of the image being covered by a border but just having the border gone is fairly useless

> AspectFill goes over the area it should on Android

It looks like this same issue exists with the Image Control on Android when using Fast renderers. If you use the FastRenderer and check the Aspect Fill in the Control Gallery it's different when you use Fast Renderers vs Not :-/  it's probably something to do with clipping. Since non fast renderers are wrapped the View is clipping the image whereas with fast renderers it isn't

> I was testing Android Button to see everything works and it seems it isn't rendering the border on API27

Good Find!!!  This one I fixed

PureWeen:
> AspectFill goes over the area it should on Android

This is a regression in 3.3 unfortunately and is not caused by ImageButton and will resolve with https://github.com/xamarin/Xamarin.Forms/issues/4133

> The border on ImageButton on Android is below the image, while on iOS is over the image, I prefer the iOS behaviour.

I added code so on Android it will draw the outline as part of the Draw of the ImageButton.  @rmarinho  and @jassmith  let me know what you think about this solution for making the border consistent between ios and android

andreinitescu:
I think ImageButton is superfluous.  Why do we need to copy Android's approach?
Button already supports displaying an image. Why can't one just use the existing Button and leave `Text` property unset?
Any specific Image behavior (IsOpaque) could become properties on the Button (IsImageOpaque) or, to make it extensible for further enhacements on image, have a property like `ImageSettings Image` to allow writing in XAML: ``<Button Image.IsOpaque="true" />`

Regarding some comments on the ticket #1724 for this PR :

Per Stephane Delcroix's initial description
> To avoid breaking changes, we should create a new ImageButton control. Both Button and ImageButton should derive from a common ButtonBase.

Moving the common Button properties up to a new ButtonBase class is **very** important to have. It's the foundation for CheckBox, and RadioButton.

> I feel like this might be better as a new control instead of using Button.Image. That will prevent us from breaking old layouts. 

Having just Button, Is it difficult not to break old layouts?

> Plus, we probably want to use the native ImageButton controls, so we'll have a lot of code switches if we just reuse the same control.

Only Android has a native ImageButton. Al the other platforms don't need it. Android has already a way to set images in buttons, by using setCompoundDrawablesWithIntrinsicBounds

> I started with each control (Button, Image) and refactored out the behavior at the forms and renderer level into into static managers. Once this was done I then just applied the necessary interfaces and managers to ImageButton.

This is good stuff useful for other future enhancements, but it should be part of refactoring, not having to add a new ImageButton .

Ideally, beside the obvious common button properties like `Command`, the `ButtonBase` should behave like the UWP\WPF ButtonBase: have `object Content { get; set; }` and `ContentAlignment` properties.
If you set it to a string, it would display a string (natively, it would use specific native API, i.e. on iOS `setTitle(s)`, on Android `Text =s`, etc.)
If you set it to an `ImageSource`, it would just display an image (natively, it would specific native API, i.e. on iOS `SetImage()`, on Android `SetCompoundDrawablesWithIntrinsicBounds()` etc.)

andreinitescu:
Per Stephane Delcroix's initial description in the ticket #1724
> To avoid breaking changes, we should create a new ImageButton control. Both Button and ImageButton should derive from a common ButtonBase.

I was hoping to see the new ButtonBase class. This would have been very important the the future for adding CheckBox, RadioButton, etc.

PureWeen:
@andreinitescu  

Base classes are just really binding to a certain concept. With the way this PR is structured now everybody can be a button, an image or have a border. 

Ignoring ImageButton for a moment.  We have two controls (Image, Button) that both provided the same behavior against images and now with this PR they both share behavior via *IImageController*

On the renderer side this is really useful for Android because currently in Android we have 3 different button renderers (AppCompat, NonAppCompat, and Fast)

We can't easily create a common base class for those 3 different renderers so the button Behavior is extracted out (much like the border behavior)

This also proved really useful with Images because I was able to homogenize how Images are processed with ImageCells, Image, and ImageButton instead of just having copy pasted code that's not consistent and harder to keep consistent. 

> Why do we need to copy Android's approach?
Button already supports displaying an image. Why can't one just use the existing Button and leave Text property unset?

Another useful purpose for having the ImageButton is that (eventually once the renderers become public) it provides the user with an ImageButtonRenderer on Android that uses an ImageButton control. Plus in general it gives us a new control that's more based around displaying an Image so we can make new layout choices without having to make concessions. For example this control renderers the border differently than a Button with an Image.

andreinitescu:
@PureWeen 

The `ImageButton` duplicates all the properties of the `Button` (CommandProperty, CornerRadiusProperty, CommandParameterProperty, BorderWidthProperty, BorderColorProperty, PaddingProperty, IsPressedProperty) and also events (Clicked, Pressed, Released).

You work on separating different implementation is great, but some a common base class is still necessary in my opinion to share and reuse common behavior. 

Similar to the ButtonBase in UWP\WPF (see it [here](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.primitives.buttonbase)) there should be a ButtonBase class with all those common properties and events related to what a button can do.
This will greatly help when implementing a CheckBox or a RadioButton, because it will just mean subclassing ButtonBase and implement new properties/events. We shouldn't have to copy and paste all those common properties and events from Button class.

Your work on `BorderBackgroundManager` is great, very useful and we should plan to refactor at some point later on other renderers to make use of it (because other controls have a border).

andreinitescu:
Another (great) benefit from having a ButtonBase class is that the Button and ImageButton implementation will become very very small and it will be very clear to see the particular behavior.

andreinitescu:
@PureWeen If you agree with having the ButtonBase, I could help with doing the modifications, and then do a PR on your "ImageButton" branch.

PureWeen:
> The ImageButton duplicates all the properties of the Button (CommandProperty, CornerRadiusProperty, CommandParameterProperty, BorderWidthProperty, BorderColorProperty, PaddingProperty, IsPressedProperty) and also events (Clicked, Pressed, Released).

I don't quite see this as bad duplication because it's just fulfilling an interface contract.  I would say reuse, on its own, doesn't give enough reason for a common base class.  We are typically trying to take more of a compositional approach to things.  I don't feel like it would be correct to burden a ButtonBase class with a property like *BorderColorProperty* because we can't just say a Button has a *BorderColorProperty*. That causes the class to be too thick and opinionated. Then what happens when we have a button where we don't want it to have a BorderColor?    As is *IButtonController* is very specific to what a Button is and can be applied to literally anything and then compositionally it can now work like a button.

andreinitescu:
@PureWeen 
I'd like to share more thoughts. 
I'm aware of challenges with inheritance but I don't understand why all of the sudden it needs to be avoided at all costs. All the properties and methods of the IButtonController are implemented by `Button` and `ImageButton` in exactly same way. Thinking ahead, having to duplicate all that code again for new  buttons does not sound appealing. 
The same issue is with other common properties across other controls. 
There has to be a better way. I saw solutions which use Roslyn compiler to implement interfaces by mixins. 

Another thing: is it necessary that both `IBorderElement` and a `IBorderController` exist? 
Also, I see only `Button` and `ImageButton` implement the new `IBorderController`, but `Frame` doesn't implement it.

PureWeen:
> Another thing: is it necessary that both IBorderElement and a IBorderController exist?

CSS uses IBorderElement and the renderers use IBorderController. 

> Also, I see only Button and ImageButton implement the new IBorderController, but Frame doesn't implement it.

I would need to investigate this one  a bit more.  There's a good argument here for having it implement IBorderController and then fall in line with Buttons so there are the same features


> There has to be a better way. I saw solutions which use Roslyn compiler to implement interfaces by mixins.

I agree. The repeat of properties is annoying and I wasn't super happy with it but I still prefer the freedom from a base class and being able to apply those aspects to things.  For example IBorderController that you brought up about Frame. It'd be nice to just apply that to whatever and have common Border behavior pulled through everything. Maybe once we start using Default Interface methods we can weave those cross cutting concerns into the classes better

If we did do a ButtonBase really the only things I would want to put in there would be around Commanding and things that make a button a button. It can be pressed and it can send commands. Anything stylized would seem to be a bit too opinionated.  Checkbox, for example, when it's first implemented we don't want to just say all these background/borders things will be part of a CheckBox API. I feel like a ButtonBase has limited use

programmation:
@PureWeen IMO things that make a button a button include more than just its ability to respond to a tap by executing a command. We rely also on the visual aspect to know that it's a button (and not a label), which includes, but is not limited to, things like: border (on/off), border colour (perhaps as a gradient), border thickness, border corner radius (perhaps non-uniform), embedded image, title, font, text attributes and so on. I support the idea of an interface to be able to apply button characteristics to other things we want to tap on for a command. But perhaps there's scope for a "functional" button that acts like a button (i.e. supports tap-for-command) vs "visual" button that looks like one as well (i.e. has borders, fill etc)? Alternatively, I think it would be interesting to see if controls can be composed by e.g. making a button out of a composite of (combinations of) `Frame`, `Image` and `Label`. Then separate enhancements to each of the composite parts would improve each of the combinations as well.

andreinitescu:
@PureWeen 

> CSS uses IBorderElement and the renderers use IBorderController.

I see the `IBorderElement` interface is only used here, in `AssemblyInfo.cs`:

       [assembly: StyleProperty("border-color", typeof(IBorderElement), nameof(BorderElement.BorderColorProperty))]

but who uses the `BorderColor` getter defined in the interface?

	interface IBorderElement
	{
		//note to implementor: implement this property publicly
		Color BorderColor { get; }

		//note to implementor: but implement the methods explicitly
		void OnBorderColorPropertyChanged(Color oldValue, Color newValue);
	}

Looking at `StylePropertyAttribute`, it seems to me the stylesheet mechanism uses the given bindable property (last parameter in the StyleProperty c-tor)

		public StylePropertyAttribute(string cssPropertyName, Type targetType, string bindablePropertyName)
		{
			CssPropertyName = cssPropertyName;
			BindablePropertyName = bindablePropertyName;
			TargetType = targetType;
		}

Having the stylesheet mechanism use the bindable property makes sense, because the value defined in a style sheet need to be set on the control, but then how's the BorderColor getter defined in the `IBorderElement` interface used? 
As a test, I actually **removed** the **getter** and all XF code compiled! Then maybe it's used somehow by reflection, but where?

Also, I think `OnBorderColorPropertyChanged` can be just moved to `IBorderController`, it's really not related to stylesheet mechanism I think.

> Maybe once we start using Default Interface methods we can weave those cross cutting concerns into the classes better

That C# 8 feature apparently will only be in .NET Core, see this (https://github.com/dotnet/csharplang/issues/52#issuecomment-420023496)
So it will be a (very) long time until it's available to Xamarin Forms.

> If we did do a ButtonBase really the only things I would want to put in there would be around Commanding and things that make a button a button. It can be pressed and it can send commands. Anything stylized would seem to be a bit too opinionated.

Yes, ButtonBase is what a "button" is and it has its implementation. Classes deriving from it (Button, CheckBox, etc.) automatically get the support for command and the fact it can be pressed.
Albeit thin, it's the foundation for extensibility. If someone wants to create his new "button", it shouldn't require to implement all those properties and methods.

PureWeen:
>  We rely also on the visual aspect to know that it's a button (and not a label), which includes, but is not limited to, things like: border (on/off), border colour (perhaps as a gradient), border thickness, border corner radius (perhaps non-uniform), embedded image, title, font, text attributes and so on. 

Agreed and I feel like this is all embodied with Button so one could inherit from Button. Whether there's a ButtonBase or not doesn't seem to limit this. Which I realize creates the question well why not just have ImageButton inherit Button? Honestly because it's an image and a button so I used interfaces to compose common behavior. If I inherited from Button then I'd have to implement the Image behavior here that already is apart of Image.  It just feels like inheritance is ultimately limiting and runs the risk of getting boxed in by saying this is a ButtonBase. 

> But perhaps there's scope for a "functional" button that acts like a button (i.e. supports tap-for-command) vs "visual" button that looks like one as well (i.e. has borders, fill etc)? 

The functional part I could see as a base class but it feels too thin right now to make a ButtonBase for that. Maybe down the road but it feels premature to just have a base class with only a "Command" for tapping property.  So then what does the static BindableProperty become?  ButtonBase.Command? The whole bindable property part of this also requires a bit more verbosity on the classes otherwise we just end up with the Bindable properties all being on base classes and not consistently apart of the actual class. I wouldn't really want to put any of the static BindableProperties into a base class.

> Alternatively, I think it would be interesting to see if controls can be composed by e.g. making a button out of a composite of (combinations of) Frame, Image and Label. Then separate enhancements to each of the composite parts would improve each of the combinations as well.

Kind of like a decorator? That would be neat. I've wondered this at times as well. Like how WPF/UWP does borders with an actual element or android does CardViews. We tried that a bit with BoxView but people like the APIs just all being on the one element :-) 

In the end these controls are all super boring on the Forms side. They are just value objects that pass through data to the native platforms.

Stubs/Xamarin.Forms.Platform.cs
```diff
@@ -51,6 +51,13 @@ internal class _ImageRenderer { }
 	[RenderWith (typeof (ButtonRenderer))]
 	internal class _ButtonRenderer { }
 
+#if __ANDROID__
+	[RenderWith(typeof(Android.AppCompat.ImageButtonRenderer))]
```
samhouts: Let's just put this in the Android namespace

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1724.cs
```diff
@@ -0,0 +1,167 @@
...
+using Xamarin.Forms.Internals;
+using System;
+using System.Linq;
+using System.Threading;
+using System.Collections.Generic;
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 1724, "[Enhancement] ImageButton ", PlatformAffected.All)]
```
samhouts:
```suggestion
	[Issue(IssueTracker.Github, 1724, "[Enhancement] ImageButton", PlatformAffected.All)]
```

Xamarin.Forms.Platform.Android/AppCompat/ImageButtonRenderer.cs
```diff
@@ -0,0 +1,323 @@
...
+using AColor = Android.Graphics.Color;
+using Xamarin.Forms.PlatformConfiguration.AndroidSpecific;
+using AStateListDrawable = Android.Graphics.Drawables.StateListDrawable;
+using ARect = Android.Graphics.Rect;
+using Android.Graphics.Drawables;
+using Android.Graphics;
+using Xamarin.Forms.Platform.Android.FastRenderers;
+using Android.OS;
+
+namespace Xamarin.Forms.Platform.Android.AppCompat
```
samhouts:
```suggestion
namespace Xamarin.Forms.Platform.Android
```

Xamarin.Forms.Platform.Android/AppCompat/ImageButtonRenderer.cs
```diff
@@ -0,0 +1,323 @@
...
+		static int _selectableItemBackgroundResourceId = 0;
+		public ImageButtonRenderer(Context context) : base(context)
+		{
+			// These set the defaults so visually it matches up with other platforms
+			SetPadding(0, 0, 0, 0);
+			SoundEffectsEnabled = false;
+			SetOnClickListener(this);
+			SetOnTouchListener(this);
+			OnFocusChangeListener = this;
+			//SetAdjustViewBounds(true);
```
samhouts: ?

Xamarin.Forms.Platform.Android/BackgroundManager.cs
```diff
@@ -0,0 +1,64 @@
...
+			if (Element == null || Control == null)
+				return;
+
+			Control.SetBackgroundColor((color ?? Element.BackgroundColor).ToAndroid());
+		}
+
+
+		static void OnElementChanged(object sender, VisualElementChangedEventArgs e)
+		{
+			var reference = Guid.NewGuid().ToString();
```
samhouts: Nope, we don't do this anymore. `Performance.Start` has `reference` as an `out` now.

Xamarin.Forms.Platform.Android/AppCompat/ImageButtonRenderer.cs
```diff
@@ -0,0 +1,323 @@
...
+
+			if (!(element is ImageButton image))
+			{
+				throw new ArgumentException("Element is not of type " + typeof(ImageButton), nameof(element));
+			}
+
+			ImageButton oldElement = ImageButton;
+			ImageButton = image;
+
+			var reference = Guid.NewGuid().ToString();
```
samhouts: Nope

Xamarin.Forms.Platform.Android/FastRenderers/VisualElementRenderer.cs
```diff
@@ -88,6 +81,8 @@ void Dispose(bool disposing)
 
 		void OnElementChanged(object sender, VisualElementChangedEventArgs e)
 		{
+			var reference = Guid.NewGuid().ToString();
```
samhouts: Nope

Xamarin.Forms.Platform.GTK/Renderers/ButtonRenderer.cs
```diff
@@ -32,7 +33,7 @@ protected override void OnElementChanged(ElementChangedEventArgs<Button> e)
 				{
 					// To allow all avalaible options in Xamarin.Forms, a custom control has been created.
```
samhouts: not your typo, but it hurts me. avalaible

Xamarin.Forms.Platform.Android/FastRenderers/VisualElementRenderer.cs
samhouts: I think there might be a fourth one of these somewhere, but my "nope" is missing.

Xamarin.Forms.Platform.Android/AppCompat/ImageButtonRenderer.cs
```diff
@@ -182,23 +191,49 @@ void IVisualElementRenderer.SetElement(VisualElement element)
...
+				var widthRatio = 1f;
+				var heightRatio = 1f;
+
+				switch (ImageButton.Aspect)
 				{
-					bd.DrawOutline(canvas, canvas.Width, canvas.Height);
+					case Aspect.Fill:
+						break;
+					case Aspect.AspectFill:
+					// AspectFill isn't fully supported right now with shadows because we would need to crop the image ourselves
```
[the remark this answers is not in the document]
PureWeen: A warning is good I'll add that

> Could you use this clipping trick to implement this
I don't think so because 4218 is clipping the whole control to the container but in this case we're building a drawable that's already all clipped to the ImageView. So we would need to clip the drawable itself to the border when it's drawn . Drawable doesn't have a clip to bounds property so we'd need to most likely use a mask on top of the image or redraw it clipped

PureWeen:
@kingces95   I added all those test cases as well to the gallery. Thanks for the good notes

@samhouts   
- I applied your changes
- I tested with previewer and it worked fine. The only issue I saw was a crash if you set the source to a URI but that happens even with an Image so this PR doesn't break that more
- I'll create issues around your comments as "known issues"

Issue #1983, opened 2018-02-27
typo issue: string and property name is not same


akgulebubekir:
it shold be "ClassStyles"

https://github.com/xamarin/Xamarin.Forms/blob/185d1ce6adb83dd652552afbca337186a80f2a03/Xamarin.Forms.Core/MergedStyle.cs#L59

StephaneDelcroix:
in this case, it could be "foo" that it wouldn't change a think. that property is only used internally and so the name doesn't matters

PR #1985, opened 2018-02-27
nameof refactorization : all hardcoded property names replaced with n…

akgulebubekir:
…ameof operator.

### Description of Change ###

all hardcoded property names replaced with nameof operator.

### Bugs Fixed ###

- #1983 

### API Changes ###
None

Added:
 None

Changed:
None

### Behavioral Changes ###
None
### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/ActivityIndicator.cs
+++ b/Xamarin.Forms.Core/ActivityIndicator.cs
@@ -6,9 +6,9 @@ namespace Xamarin.Forms
 	[RenderWith(typeof(_ActivityIndicatorRenderer))]
 	public class ActivityIndicator : View, IElementConfiguration<ActivityIndicator>
 	{
-		public static readonly BindableProperty IsRunningProperty = BindableProperty.Create("IsRunning", typeof(bool), typeof(ActivityIndicator), default(bool));
+		public static readonly BindableProperty IsRunningProperty = BindableProperty.Create(nameof(IsRunning), typeof(bool), typeof(ActivityIndicator), default(bool));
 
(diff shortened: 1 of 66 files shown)


StephaneDelcroix:
Thanks for this PR.

We appreciate the effort, but we do not take PR that does refactoring for the sake of it. Instead, we refactor that code when we happen to modify it.

It minimizes the risks, and keep the code history cleaner.

opcodewriter:
So sad. What's the issue with improving the code from time to time? 
It's not like the code in the Xamarin Forms implementation is so perfect it shouldn't or rarely be touched.
Replacing the hard coded values with nameof is actually a pretty legit code change, how in the world you'd not want to do this? 
Who invented these rules?

programmation:
@opcodewriter this is a noble effort and I think your motivation is good. However in a code base of this size (actually in smaller ones too) applying as many changes as you have made to the source all at once is very dangerous, even though it might not look like it from your perspective, mainly because you have touched lots of files that other people are also working on. 
It's true that all these changes need to be made _eventually_ but although your changes are good you haven't in fact fixed anything that was broken. It's better and less dangerous to the health of the project to do this type of code maintenance little by little, in small commits, as @StephaneDelcroix says, so the risks and potential impact are smaller for each change. 
A good time to do this work is when the file(s) are being worked on anyway to fix a real issue. This helps to prevent unintentional merge conflicts and keeps commits for things that actually add features or fix problems.

opcodewriter:
@programmation No, there's no danger with having a commit which is focused on a SINGLE task, which is to replace the hardcoded names (who did that anyway?) with nameof.
The other option is to live in our ivory tower, invent our rules and don't ever have code improved.
And there shouldn't be any merge conflicts because there's very little or no chance at all for someone else to also touch those bindable property definitions.

PR #1987, opened 2018-02-27
Allow custom XF loaders to also access internals of XAML

kzu:
Just like [XF.Core](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/Properties/AssemblyInfo.cs#L26) does, also expose this to custom loaders (i.e. if they need to access anything internal to XAML loading).

### DIFF
--- a/Xamarin.Forms.Xaml/Properties/AssemblyInfo.cs
+++ b/Xamarin.Forms.Xaml/Properties/AssemblyInfo.cs
@@ -5,6 +5,8 @@ using Xamarin.Forms.Internals;
 [assembly: InternalsVisibleTo("Xamarin.Forms.Xaml.UnitTests")]
 [assembly: InternalsVisibleTo("Xamarin.Forms.Build.Tasks")]
 [assembly: InternalsVisibleTo("Xamarin.Forms.Xaml.Design")]
+[assembly: InternalsVisibleTo("Xamarin.Forms.Loader")]
+
 [assembly: Preserve]
 
 [assembly: XmlnsDefinition("http://xamarin.com/schemas/2014/forms", "Xamarin.Forms.Xaml")]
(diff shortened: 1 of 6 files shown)


kzu:
The workaround for now is to name my "interop" assembly `Xamarin.Forms.Build.Tasks` which is not ideal but gives me internal access to both Core and Xaml as needed by my loader.

jassmith:
Update docs please :) InternalsVisibleTo is public API

Xamarin.Forms.Xaml/Properties/AssemblyInfo.cs
```diff
@@ -5,6 +5,7 @@
 [assembly: InternalsVisibleTo("Xamarin.Forms.Xaml.UnitTests")]
 [assembly: InternalsVisibleTo("Xamarin.Forms.Build.Tasks")]
 [assembly: InternalsVisibleTo("Xamarin.Forms.Xaml.Design")]
+[assembly: InternalsVisibleTo("Xamarin.Forms.Loader")]
```
StephaneDelcroix: you shouldn't be using a namespace starting with `Xamarin.Forms`

you shouldn't also poke into internal APIs, as we might change them, and that would break your tool

jassmith:
Just want to make it clear that any internal API you use is NOT considered stable and we are free to change on a whim.

kzu:
That's what internal means, sure.

It's the same thing that anyone could do by using reflection to access
private members even. You can't prevent that from happening, but noone
doing so can come requiring you keep backs compat.

That's totally understood, and the intended usage of this is for tooling
that Xamarin itself will ship, so we can make sure we keep it sync if you
happen to change the internals.
-- 

--
/kzu from mobile

PR #1997, opened 2018-02-28
[flex] better margin support with Wrap

StephaneDelcroix:
### Bugs Fixed ###

/

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###

### DIFF
--- a/Xamarin.Flex/Flex.cs
+++ b/Xamarin.Flex/Flex.cs
@@ -188,7 +188,7 @@ namespace Xamarin.Flex
 	}
 
 	/// <summary>
-	/// Value for <see cref="P:Xamarin.Flex.Item.Wrap" />.
+	/// Value for <see cref="P:Xamarin.Flex.Item.Basis" />.
 	/// </summary>
 	struct Basis {
 		readonly bool _isRelative;
@@ -524,9 +524,7 @@ namespace Xamarin.Flex
 					}
 					else {
 						child.Frame[layout.frame_size2_i] = (layout.vertical ? width : height)
-							- (layout.vertical ? child.MarginLeft : child.MarginTop)
-							- (layout.vertical ? child.MarginRight : child.MarginBottom);
-
+							- child.MarginThickness(!layout.vertical);
 					}
 				}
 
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Core/FlexLayout.cs
```diff
@@ -408,12 +408,12 @@ protected override SizeRequest OnMeasure(double widthConstraint, double heightCo
 			if (double.IsPositiveInfinity(widthConstraint)) {
 				widthConstraint = 0;
 				foreach (var item in _root)
-					widthConstraint = Math.Max(widthConstraint, item.Frame[0] + item.Frame[2]);
+					widthConstraint = Math.Max(widthConstraint, item.Frame[0] + item.Frame[2] + item.MarginRight);
```
StephaneDelcroix: this fixes measuring with margins

Xamarin.Flex/Flex.cs
```diff
@@ -557,7 +555,7 @@ static void layout_item(Item item, float width, float height)
 					float basis = child.Basis.Length;
 					if (child.Basis.IsRelative)
 						basis *= (layout.vertical ? height : width);
-					child.Frame[layout.frame_size_i] = basis;
+					child.Frame[layout.frame_size_i] = basis - child.MarginThickness(layout.vertical);
```
StephaneDelcroix: this removes the margin when using `basis` from the main-axis dimension

Xamarin.Flex/Flex.cs
```diff
@@ -573,8 +571,8 @@ static void layout_item(Item item, float width, float height)
 					}
 
 					float child_size2 = child.Frame[layout.frame_size2_i];
-					if (!float.IsNaN(child_size2) && child_size2 > layout.line_dim) {
-						layout.line_dim = child_size2;
+					if (!float.IsNaN(child_size2) && child_size2 + child.MarginThickness(!layout.vertical) > layout.line_dim) {
+						layout.line_dim = child_size2 + child.MarginThickness(!layout.vertical);
```
StephaneDelcroix: this grows the cross-axis dimension of the lines when using margins, and wrap

Xamarin.Flex/Flex.cs
rmarinho: How will this map to the code ported from flex lib? shouldn't at least we add a note that we mad this change?

StephaneDelcroix: the code never mapped

Issue #1754, opened 2018-01-30
[F100] Rounded Corners


hartez:
## Rationale

Rounded corners are a common visual requirement which Forms does not support without custom renderers. Forms should provide a common interface for defining rounded corners on Views.

## Implementation

We need to add an interface for defining what's required for a View to implement corner rounding (and provide CSS support):

```csharp
interface IRoundedCorners
{
	CornerRadius CornerRadius { get; }
	void CornerRadiusChanged(CornerRadius oldValue, CornerRadius newValue);
}
```

The CornerRadius BindableProperty can be implemented in one place:

```csharp
static class RoundedCornerElement
{
	public static readonly BindableProperty CornerRadiusProperty = BindableProperty.Create(nameof(CornerRadius), typeof(CornerRadius), typeof(IRoundedCorners));
}
```

Renderer implementations should be careful to take [Borders](https://github.com/xamarin/Xamarin.Forms/issues/1695) into consideration.

The following Views are candidates for rounded corners:
BoxView (in progress: https://github.com/xamarin/Xamarin.Forms/issues/1709)
ImageView
StackLayout
Grid
AbsoluteLayout
ScrollView
(probably others)

## Difficulty: Easy
Providing the interface is simple; individual implementations may vary in difficulty.

programmation:
One common requirement is to create an ellipse / circle (depending on geometry of the view). This can't be done easily with a numerical value for border rounding - it's better done as a percentage, or with a flag that says "create an ellipse". Please consider using "special" values for the flag (e.g. -2 for ellipse), or making the values work like `Grid` width/height do.

programmation:
Perhaps this could eventually be extended to allow a `BezierPath` for the border of a view? All the platforms support doing this.

hartez:
Maybe the right option is to have an interface called `IMask` or `IClip`, and allow specifying the method used for clipping/masking the View (e.g., CornerRadius, Path, Ellipse).

StephaneDelcroix:
make sure the default value is `-1` aka RoundedCornerElement.Default

I'm wondering if we couldn't put that property in BorderElement

bentmar:
https://github.com/NAXAM/effects-xamarin-forms
have a look at ViewEffects, Naxam has done a great job on this one, might help to speed things up :)

andreinitescu:
This would be a great enhancement, which would give of a helping hand for making layouts simpler and take less time to render:

[...]

PR #1998, opened 2018-03-01
[Enhancement] Rounded corners on BoxView

### LINKED ISSUE #1709: [Enhancement] Rounded corners on BoxView
## Rationale

Forms does not currently support rounded corners on BoxView.

## Implementation

Add a CornerRadius bindable property to BoxView (using the same struct as in https://github.com/xamarin/Xamarin.Forms/issues/1664).

```	
public static readonly BindableProperty CornerRadiusProperty =
BindableProperty.Create(nameof(CornerRadius), typeof(CornerRadius), default(CornerRadius));
```

On iOS, this will likely require checking for a non-zero radius and switching to a UIBezierPath for drawing. On UWP, it's likely just a matter of applying the RadiusX/RadiusY properties to the underlying Rectangle. On Android, it probably just requires overriding Draw and applying a clipping path.

## Backward Compatibility

This is a new property, so there should not be any backward compatibility issues.

jsuarezruiz:
### Description of Change ###

Add a **CornerRadius** bindable property to **BoxView**.

```
struct CornerRadius
{
     public double TopLeft { get; set; }
     public double TopRight { get; set; }
     public double BottomLeft { get; set; }
     public double BottomRight { get; set; }

      public static implicit operator CornerRadius(double uniformRadius)
      {
            return new CornerRadius(uniformRadius);
      }
 }

public static readonly BindableProperty CornerRadiusProperty =
BindableProperty.Create(nameof(CornerRadius), typeof(CornerRadius), default(CornerRadius));
```

Supported platforms: Android, iOS, UWP, macOS, GTK and WPF.

- Android: Use GradientDrawable and SetCornerRadii method.
- iOS: Use UIBezierPath.
- UWP: Use RadiusX and RadiusY properties from the Rectangle ().
- macOS: Use NSBezierPath.
- GTK: Use Cairo drawing Arcs.
- WPF: Use a VisualBrush in a Rectangle.
- Tizen: It's not possible.

cornerradius boxview

### Bugs Fixed ###

- fixes #1709

### API Changes ###

Added:

- CornerRadius BoxView.CornerRadius{ get; set; } //Bindable Property

### PR Checklist ###


Xamarin.Forms.Core/BoxView.cs
```diff
@@ -8,6 +9,8 @@ public class BoxView : View, IElementConfiguration<BoxView>
 	{
 		public static readonly BindableProperty ColorProperty = BindableProperty.Create("Color", typeof(Color), typeof(BoxView), Color.Default);
 
+		public static readonly BindableProperty CornerRadiusProperty = BindableProperty.Create("CornerRadius", typeof(CornerRadius), typeof(BoxView), default(CornerRadius));
```
StephaneDelcroix: use `nameof(CornerRadius)`

Xamarin.Forms.Core/CornerRadius.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+
+namespace Xamarin.Forms.Core
```
StephaneDelcroix: we don't use that namespace, we use `Xamarin.Forms`

Xamarin.Forms.Core/CornerRadius.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+	public struct CornerRadius
```
StephaneDelcroix: require a `[TypeConverter]` so it can be set from XAML and CSS

Xamarin.Forms.Core/CornerRadius.cs
```diff
@@ -0,0 +1,38 @@
...
+		public double BottomLeft { get; set; }
+		public double BottomRight { get; set; }
+
+		public CornerRadius(double uniformRadius) : this(uniformRadius, uniformRadius, uniformRadius, uniformRadius)
+		{
+		}
+
+		public CornerRadius(double topLeft, double topRight, double bottomLeft, double bottomRight)
+		{
+			if (topLeft < 0 || double.IsNaN(topLeft))
```
StephaneDelcroix: don't do value check in the ctor, but when the property is applied.

Xamarin.Forms.Core/CornerRadius.cs
StephaneDelcroix: a good source of inspiration for this class would be `Thickness`

Xamarin.Forms.Core/CornerRadius.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+
+namespace Xamarin.Forms.Core
+{
+	public struct CornerRadius
+	{
+		public double TopLeft { get; set; }
```
StephaneDelcroix: make it immutable by decorating the setter with `private`

Xamarin.Forms.Core/CornerRadius.cs
```diff
@@ -0,0 +1,38 @@
...
+			TopRight = topRight;
+			BottomLeft = bottomLeft;
+			BottomRight = bottomRight;
+		}
+
+		public static implicit operator CornerRadius(double uniformRadius)
+		{
+			return new CornerRadius(uniformRadius);
+		}
+	}
```
StephaneDelcroix: requires `Equals`, `operator ==`, `operator !=`, `GetHashCode` and `Deconstruct` implementations

StephaneDelcroix: well, `Deconstruct` is optional, but nice to have

Xamarin.Forms.Core/CornerRadius.cs
StephaneDelcroix: As it's a struct, we need a way to figure out if the `CornerRadius` is the default CornerRadius, or if it was created by one of the parameterized `ctor`.

If we ever applied this property to, say, `Button`, having a default CornerRadius would mean "use whatever default on the platform".

There are multiple ways of implementing this, I'd recommend having a private field `isnotdefault` (we can not initialize fields with default values in structs) and set that field to `true` in the parameterized ctors.

That value should then be used in `Equals` and operators, so we can do `if (my box.CorenerRadius == default(CornerRadius)) {}`

Xamarin.Forms.Platform.MacOS/Renderers/BoxViewRenderer.cs
```diff
@@ -32,8 +39,19 @@ protected override void SetBackgroundColor(Color color)
...
+
 			Control.WantsLayer = true;
 			Control.Layer.BackgroundColor = color.ToCGColor();
 		}
+
+		void SetCornerRadius(CornerRadius cornerRadius)
+		{
+			if (Element == null)
+				return;
+
```
StephaneDelcroix: here's a good pattern for setting that property in all renderers:

```
if (cornerRadius == default(CornerRadius)
    RestoreToDefault(...); //you can not just return, as the property might have been set to a value earlier
else
    ApplyCornerRadius(...);
```
might be overkill in this case, as BoxView has no native counterpart, and we know the the default is 0;

andreinitescu:
@StephaneDelcroix About your remark

> I like this very much, to the point I'd wish at some point we have that property on most/all controls. I did the review in that broader usage light, focusing on the Core part.

I am happy Xamarin takes notice of this. Because the way majority of developers use BoxView is to have a rounded background wrapping layout views. Unfortunately the code ends up something like this:

```
<BoxView Padding="10" BackgroundColor="Blue" CornerRadius="7">
   <StackLayout>
        <!-- Entry, Label etc. controls here -->
   </StackLayout
</BoxView>
```
The BoxView is superfluous making the layout more complex than it should be. 

If StackLayout had a CornerRadius property, the layout would be simpler:
```
<StackLayout Padding="10" BackgroundColor="Blue" CornerRadius="7">
        <!-- Entry, Label etc. controls here -->
</BoxView>
```

BoxView is mostly an unnecessary control, unfortunately large majority of  devs, including sometimes myself, had to use it, but it's not doing a great job.

jsuarezruiz:
Thank you very much for the review @StephaneDelcroix . I will apply changes soon!.
@andreinitescu No, all feedback contributes, thank you. Although I think that maybe you would like this enhancement: [https://github.com/xamarin/Xamarin.Forms/issues/1754](https://github.com/xamarin/Xamarin.Forms/issues/1754)

jassmith:
@andreinitescu I very much would like to see CornerRadius brought everywhere, ideally whithout dropping a drawable/mask everywhere on Android except when its actually in use.

If you want to write a Spec and link me to it I will review it and try to get it approved for F100

rmarinho:
build

rmarinho:
build

Xamarin.Forms.Core/BoxView.cs
jsuarezruiz: Changed!. Right now there are property definitions using string name and others with nameof. It would be good to review them all (maybe in a PR?).

Xamarin.Forms.Core/CornerRadius.cs
jsuarezruiz: Added Equals, operator ==, operator !=, GetHashCode and Deconstruct!

Xamarin.Forms.Core/CornerRadius.cs
jsuarezruiz: Done.

Xamarin.Forms.Core/CornerRadius.cs
jsuarezruiz: It is!

Xamarin.Forms.Core/CornerRadius.cs
jsuarezruiz: Changed.

Xamarin.Forms.Core/CornerRadius.cs
jsuarezruiz: It is true!. I have added the Converter.

Xamarin.Forms.Core/CornerRadius.cs
jsuarezruiz: This is important. Changed.

Xamarin.Forms.Core/CornerRadius.cs
jsuarezruiz: Added a IsDefault in CornerRadius class. Is it what is expected?

Xamarin.Forms.Core/BoxView.cs
StephaneDelcroix: we don't do massive refactoring. that cause major merges issues with unrelated branches.

Xamarin.Forms.Core/CornerRadius.cs
StephaneDelcroix: 🎉

Xamarin.Forms.Core/CornerRadius.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System.Diagnostics;
+
+namespace Xamarin.Forms
+{
+	[DebuggerDisplay ("TopLeft={TopLeft}, TopRight={TopRight}, BottomLeft={BottomLeft}, BottomRight={BottomRight}")]
+	[TypeConverter (typeof (CornerRadiusTypeConverter))]
+	public struct CornerRadius
+	{
+		public double TopLeft { get; private set; }
```
StephaneDelcroix: the `private set;` isn't even required, as the property is only set from the `.ctor`

Xamarin.Forms.Core/CornerRadius.cs
```diff
@@ -0,0 +1,78 @@
...
+		{
+			return TopLeft.Equals (other.TopLeft) && TopRight.Equals (other.TopRight) && BottomLeft.Equals (other.BottomLeft) && BottomRight.Equals (other.BottomRight);
+		}
+
+		public override bool Equals (object obj)
+		{
+			if (ReferenceEquals (null, obj))
+				return false;
+
+			return obj is CornerRadius && Equals ((CornerRadius)obj);
```
StephaneDelcroix: `return obj is CornerRadius cornerRadius && Equals(cornerRadius);`

Xamarin.Forms.Core/CornerRadius.cs
```diff
@@ -0,0 +1,78 @@
...
+		}
+
+		public static implicit operator CornerRadius (double uniformRadius)
+		{
+			return new CornerRadius (uniformRadius);
+		}
+
+		bool Equals (CornerRadius other)
+		{
+			return TopLeft.Equals (other.TopLeft) && TopRight.Equals (other.TopRight) && BottomLeft.Equals (other.BottomLeft) && BottomRight.Equals (other.BottomRight);
```
StephaneDelcroix: those are `double`s, it's safe to use `==`

Xamarin.Forms.Core/CornerRadius.cs
```diff
@@ -0,0 +1,78 @@
...
+	public struct CornerRadius
+	{
+		public double TopLeft { get; private set; }
+		public double TopRight { get; private set; }
+		public double BottomLeft { get; private set; }
+		public double BottomRight { get; private set; }
+
+		internal bool IsDefault
+		{
+			get { return TopLeft == 0 && TopRight == 0 && BottomLeft == 0 && BottomRight == 0; }
```
StephaneDelcroix: `(0,0,0,0)` isn't a good default for `CornerRadius`. Some controls (android buttons, e.g.) are rounded by default. if you want to make them with square angles, you will set `button.CornerRadius = new CornerRadius (0,0,0,0)`. the renderer will then interpret this as `Default` and will use the default (which is rounded) for the corners.

as this is a `struct`, you can not have default values for fields (they have to be `0`), so you have to use a flag (`false` by default) that you set to `true` the non-default ctors. `default(CornerRadius)` will return a struct with that field unchanged...

if it's unclear, I can do that for you

Xamarin.Forms.Core/CornerRadius.cs
```diff
@@ -0,0 +1,78 @@
...
+	[DebuggerDisplay ("TopLeft={TopLeft}, TopRight={TopRight}, BottomLeft={BottomLeft}, BottomRight={BottomRight}")]
+	[TypeConverter (typeof (CornerRadiusTypeConverter))]
+	public struct CornerRadius
+	{
+		public double TopLeft { get; private set; }
+		public double TopRight { get; private set; }
+		public double BottomLeft { get; private set; }
+		public double BottomRight { get; private set; }
+
+		internal bool IsDefault
```
StephaneDelcroix: no need to expose this. it's not more expensive to do `if (cornerRadius == default(CornerRadius))`

Xamarin.Forms.Core/CornerRadiusTypeConverter.cs
```diff
@@ -0,0 +1,45 @@
...
+						case 4:
+							if (double.TryParse (cornerRadius[0], NumberStyles.Number, CultureInfo.InvariantCulture, out double topLeft)
+								&& double.TryParse (cornerRadius[1], NumberStyles.Number, CultureInfo.InvariantCulture, out double topRight)
+								&& double.TryParse (cornerRadius[2], NumberStyles.Number, CultureInfo.InvariantCulture, out double bottomLeft)
+								&& double.TryParse (cornerRadius[3], NumberStyles.Number, CultureInfo.InvariantCulture, out double bottomRight))
+								return new CornerRadius (topLeft, topRight, bottomLeft, bottomRight);
+							break;
+					}
+				} else if (value.Contains (" ")) { //CSS
+					var cornerRadius = value.Split(' ');
```
StephaneDelcroix: CSS accepts 1, 2, 3 or 4 arguments: https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius

Xamarin.Forms.Core/CornerRadiusTypeConverter.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+using System.Globalization;
+
+namespace Xamarin.Forms
+{
+	[Xaml.ProvideCompiled ("Xamarin.Forms.Core.XamlC.CornerRadiusTypeConverter")]
```
StephaneDelcroix: you actually do _not_ provide a compiled version of this. you don't have to do it, I will, but remove this attribute.

Xamarin.Forms.Core/CornerRadius.cs
jsuarezruiz: Do you mean the property IsDefault?

Xamarin.Forms.Core/CornerRadiusTypeConverter.cs
jsuarezruiz: Ok.

Xamarin.Forms.Core/CornerRadius.cs
jsuarezruiz: Changed.

Xamarin.Forms.Core/CornerRadius.cs
jsuarezruiz: Changed.

Xamarin.Forms.Core/CornerRadius.cs
jsuarezruiz: Ok!

Xamarin.Forms.Core/CornerRadius.cs
StephaneDelcroix: here's what I meant

```csharp
	public struct CornerRadius
	{
		public double TopLeft { get; }		//as the properties are only set from the ctor, they do not need an internal setter
		public double TopRight { get; }
		public double BottomLeft { get; }
		public double BottomRight { get; }

		bool isParameterized;

		public CornerRadius(double uniformRadius) : this(uniformRadius, uniformRadius, uniformRadius, uniformRadius)
		{
		}

		public CornerRadius(double topLeft, double topRight, double bottomLeft, double bottomRight)
		{
			isParameterized = true;			//not default anymore
			TopLeft = topLeft;
			TopRight = topRight;
			BottomLeft = bottomLeft;
			BottomRight = bottomRight;
		}

		//...

		bool Equals(CornerRadius other)
		{
			if (!isParameterized && !other.isParameterized)
				return true; //default CornerRadius (created with parameterless ctor, or with default(CornerRadius) are Equal
			return TopLeft == other.TopLeft && TopRight == other.TopRight && BottomLeft == other.BottomLeft && BottomRight == other.BottomRight;
		}

		// ...
	}
```

jsuarezruiz: Ahh, now I understood you perfectly. Thank you!

StephaneDelcroix:
Approved the API changes

Xamarin.Forms.Platform.MacOS/Controls/FormsBoxView.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System;
+using AppKit;
+using CoreGraphics;
+
+namespace Xamarin.Forms.Platform.macOS.Controls
+{
+	public class FormsBoxView : NSView
```
rmarinho: make this internal

rmarinho:
@jsuarezruiz  build is failing 

```
Xamarin.Forms.Controls\CoreGalleryPages\CoreBoxViewGalleryPage.cs (36, 138)
Xamarin.Forms.Controls\CoreGalleryPages\CoreBoxViewGalleryPage.cs(36,138): Error CS0246: The type or namespace name 'Core' could not be found (are you missing a using directive or an assembly reference?)
```

Xamarin.Forms.Platform.MacOS/Controls/FormsBoxView.cs
jsuarezruiz: Ok!

jsuarezruiz:
@rmarinho Fixed!

Xamarin.Forms.Platform.MacOS/Controls/FormsBoxView.cs
```diff
@@ -0,0 +1,57 @@
...
+		NSColor _colorToRenderer;
+
+		nfloat _topLeft;
+		nfloat _topRight;
+		nfloat _bottomLeft;
+		nfloat _bottomRight;
+
+		public override void DrawRect (CGRect dirtyRect)
+		{
+			_colorToRenderer.SetFill ();
```
rmarinho: Where is this init? it throws NRE

Xamarin.Forms.Platform.GTK/Renderers/BoxViewRenderer.cs
```diff
@@ -1,85 +1,97 @@
 ﻿using System.ComponentModel;
+using Xamarin.Forms.Core;
```
rmarinho: need to drop this. this namespace doesn't exist

rmarinho: Also about this renderer, I see a white background , can you confirm that if we set the color of background to Transparent it does not show the white background.

screen shot 2018-03-22 at 15 03 47

Xamarin.Forms.Platform.UAP/BoxViewRenderer.cs
```diff
@@ -33,5 +47,11 @@ protected override AutomationPeer OnCreateAutomationPeer()
 
 			return new FrameworkElementAutomationPeer(Control);
 		}
+
+		private void SetCornerRadius(Core.CornerRadius cornerRadius)
```
rmarinho: COre. doesn't exist. this is breaking the build

Xamarin.Forms.Platform.MacOS/Controls/FormsBoxView.cs
jsuarezruiz: I think I should have lost something at some point respect to the original branch: https://github.com/XamarinFormsCommunity/Xamarin.Forms/blob/feature/1709-rounded-boxview/Xamarin.Forms.Platform.MacOS/Renderers/BoxViewRenderer .cs
Fixed.

StephaneDelcroix:
I've rebased and fixed your branch, but I'm unable to push to the original.

it's here https://github.com/xamarin/Xamarin.Forms/tree/jsuarezruiz-feature/1709-rounded-boxview-II

waiting for build results...

petermajor:
Hi there... I think there might be an issue with this corner radius calculation... it doesn't convert the device independent pixels to raw pixels before calling SetCornerRadii

PR #2010, opened 2018-03-04
[Flex] correctly measure images

StephaneDelcroix:
### Description of Change ###

[Flex] correctly measure images

use Aspect.Fill to measure images. also make sure we do not crash if the
Bounds are smaller than the request. Like any other control, we will
draw outside of our bounds, and rely on ClipToBounds for a nice shave.

### PR Checklist ###

### DIFF
--- a/Xamarin.Flex/Flex.cs
+++ b/Xamarin.Flex/Flex.cs
@@ -241,7 +241,7 @@ namespace Xamarin.Flex
 		IList<Item> Children { get; set; }
 		bool ShouldOrderChildren { get; set; }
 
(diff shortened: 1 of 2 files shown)


Xamarin.Flex/Flex.cs
```diff
@@ -925,9 +925,9 @@ public void init(Item item, float width, float height)
 
 				width -= item.PaddingLeft + item.PaddingRight;
 				height -= item.PaddingTop + item.PaddingBottom;
-				if (   width < 0
-				    || height < 0)
-					throw new ArgumentException();
+				//if (   width < 0
+				 //   || height < 0)
+					//throw new ArgumentException();
```
StephaneDelcroix: this is not the final fix

PR #2043, opened 2018-03-07
Rename the icons in the win folder to match the xml

alanmcgovern:
~I still have one question that i need to find the answer to.~

~There are 248 icons in the Mac folder and 32 in the windows folder. I would expect every win icon to have 8 counterparts in the Mac folder (retina/non-retina, light/dark, selected/not-selected). That means we're missing some icons in the Mac folder as we should have 256.~


alanmcgovern:
OOPS! never mind! I also counted the script i used to rename the files :p

PR #2061, opened 2018-03-09
[UWP] AccessKey support on VisualElement and TabbedPage

### LINKED ISSUE #1672: [Enhancement] UWP: Button: AccessKey support
image

## Rationale

Disabled and/or power users find the keyboard easier/faster to use than the mouse. 

## Implementation

A UAP platform specific subset of https://docs.microsoft.com/en-us/windows/uwp/design/input/access-keys as can be exposed via the following:

```csharp
public class View {
	public static readonly BindableProperty AccessKey; // string
	public static readonly BindableProperty AccessKeyPlacement; // AccessKeyPlacement
}

public enum AccessKeyPlacement {
	Auto = 0,
	Top, 
	Bottom, 
	Right, 
	Left, 
	Center, 
}
```

## Expected Result

 https://docs.microsoft.com/en-us/windows/uwp/design/input/access-keys

## Implications for CSS

None.

## Backward Compatibility

None.

## Difficulty : Easy

The test matrix should include all derivations of View typically clicked or tapped.

henricm:
### Description of Change ###

Fixes #1672 by adding AccessKey, AccessKeyPlacement and horizontal
and vertical offset properties to Windows specific VisualElement class.
Also handles AccessKeys for tabbed pages (ContentPage).

### API Changes ###

Added:
 - New `enum AccessKeyPlacement` for specifying placement of the access key tooltip.
 - string WindowsSpecific.VisualElement.AccessKey { get; set; } //Bindable Property
 - AccessKeyPlacement WindowsSpecific.VisualElement.AccessKeyPlacement { get; set; } //Bindable Property
 - double WindowsSpecific.VisualElement.AccessKeyHorizontalOffset{ get; set; } //Bindable Property
 - double WindowsSpecific.VisualElement.AccessKeyVerticalOffset{ get; set; } //Bindable Property
 
### Behavioral Changes ###

When you set AccessKey property button, tab etc will be able to execute its action when user presses the Alt-button. When the Alt-button is pushed once, access key tooltips will be displayed.

### PR Checklist ###


Xamarin.Forms.Platform.UAP/TabbedPageRenderer.cs
```diff
@@ -232,16 +237,48 @@ void OnLoaded(object sender, RoutedEventArgs args)
...
 			e.Apply(Element.Children, Control.Items);
-
+			switch (e.Action)
+			{
+				case NotifyCollectionChangedAction.Add:
+				case NotifyCollectionChangedAction.Remove:
+				case NotifyCollectionChangedAction.Replace:
+					if (e.NewItems != null)
+						foreach (var c in e.NewItems)
+							(c as Page).PropertyChanged += OnChildPagePropertyChanged;
```
jassmith: I wonder if there is a way to not sub/unsub this event and still achieve this effect. Its expensive to sub PropertyChanged as it triggers a fair bit...

henricm: Good point. However - is it that big difference compared to other property change handling.
One thought could be to try to handle this in xaml with converters or something? Haven't looked alot into that yet though.

Xamarin.Forms.Platform.UAP/TabbedPageRenderer.cs
```diff
@@ -232,16 +237,48 @@ void OnLoaded(object sender, RoutedEventArgs args)
...
+		void OnPagesChanged(object sender, NotifyCollectionChangedEventArgs e) {
 			e.Apply(Element.Children, Control.Items);
-
+			switch (e.Action)
+			{
+				case NotifyCollectionChangedAction.Add:
+				case NotifyCollectionChangedAction.Remove:
+				case NotifyCollectionChangedAction.Replace:
+					if (e.NewItems != null)
+						foreach (var c in e.NewItems)
```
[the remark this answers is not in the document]
henricm: Sure. Curious though - why is that preferred? :-)

henricm:
@PureWeen thanks for the feedback!

I haven't looked into implementing access key scope in this PR. I read about how it works in UWP but I reasoned that as a first step we only wanted basic support for access keys. 
However, I could dive more into that if desired? In that case we probably want to add support for `AccessKeyInvoked` event too. 
There is also localization of access keys which natively is handled in XAML and I'm not sure how that would translate in the Xamarin.Forms world.

PureWeen:
@henricm   Yea I looked at the Localization as well seeing what made sense. My thought here is that since this is all settable from forms as a string you can just use the Localization APIs through forms to set things to a Localized value..   If you use the bindings to set these properties then localization can just be achieved that way

As far as the native support for localization that feels like it would just be a different issue as that's just about setting the Uid of the elements right? Opening up a platform API that lets you set the Uid on the native control would just be generally useful it seems but outside the scope of this issue.

As far as *IsAccessKeyScope*. My thoughts would be that if it's easy enough to apply the property at the layout level and then translate this down into UWP go for it but if the devil in the details makes this too tricky then a separate issue would suffice..   Is it as easy as adding a property to *Xamarin.Forms.Layout* then having *LayoutRenderer* on UWP set the property?  Or am I being short sighted?

henricm:
@PureWeen sounds good regarding localization and I will definitely have a look at access key scope. Hopefully later this week.

Xamarin.Forms.Platform.UAP/TabbedPageRenderer.cs
```diff
@@ -429,5 +460,65 @@ void UpdateBarIcons()
 					Control.InvalidateMeasure();
 			}
 		}
+        void UpdateToolbarPlacement()
```
hartez: Spaces -> Tabs

Xamarin.Forms.Platform.UAP/TabbedPageRenderer.cs
```diff
@@ -429,5 +460,65 @@ void UpdateBarIcons()
...
+		}
+
+		void AccessKeyInvokedForTab(UIElement sender, AccessKeyInvokedEventArgs arg)
+		{
+			var tab = sender as TextBlock;
+			if (tab != null && tab.DataContext is Page page)
+				Element.CurrentPage = page;
+		}
+
+		protected void UpdateAccessKey(TextBlock control) {
```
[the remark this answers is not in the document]
henricm: Created AccessKeyHelper which should take care of this now.

henricm:
@PureWeen I wrestled a bit with adding `IsAccessKeyScope` on Layout and then setting the property on the UWP control. However I couldn't get it to work I'm afraid. I also tried enumerate the UWP specific controls (Panels etc) and try to figure out if the property could be set there instead. But I never managed to find a way to get it to work. I'm sure that someone with more UWP and/or Xamarin.Forms knowledge will be able to solve this.

PureWeen:
@henricm   yea :-/  I hoped it'd be just be an easy win but I'm not surprised it wasn't.  Changes look good

henricm:
@hartez the duplicate has been removed and everything is rebased as of today so please take another look when you have time. Thanks!

PR #2083, opened 2018-03-12
Fix VS errors when loading clean solution; prevent VS rewriting Android project files

hartez:
### Description of Change ###

These changes solve two problems. 

#### First:

Right now after a full `git clean`, opening the solution in VSWin results in errors in each project which attempts to import Xamarin.Forms.targets because Xamarin.Forms.Build.Tasks.dll is not yet built. VS doesn't display these errors immediately. If you then build the Tasks project and then attempt to build and run any of the XF projects in the solution (ControlGallery, PagesGallery, or the EmbeddingTestBed), VS will either:

1. Display the "Could not load Xamarin.Forms.targets" errors and refuse to build/run the project or
2. Completely freeze up and have to be killed

Either scenario requires a restart. 

With these changes, VSWin does not load the Xamarin.Forms.targets files if Xamarin.Forms.Build.Tasks.dll is not available, preventing the initial errors. You can simply open the solution after a `git clean`, select a project (e.g. ControlGallery on Android), hit debug, and it will build and run. No restart required.

When MSBuild loads the projects, the conditional on the existence of Xamarin.Forms.Build.Tasks.dll is ignored, preserving the old behavior.

VSMac still requires building the Tasks project, then touching the project/solution to force a reload before building/running one of the Forms projects.

#### Second:

Opening the solution in VS loads each of the Android-targeted projects; after they are loaded, the value of `AndroidUseLatestPlatformSdk` is evaluated and if a later SDK is installed, the value of `TargetFrameworkVersion` is updated to the latest SDK. The project is then unloaded and reloaded. This adds considerable time to the loading of the solution and modifies each Android `.csproj` file; these modifications have to be reverted before committing changes. 

With these changes, the value of `AndroidUseLatestPlatformSdk` is set to `false` when the solution is opened in VS; the projects only load once, and the project files remain unchanged. The old `AndroidUseLatestPlatformSdk` value of `true` is preserved when the projects are loaded by MSBuild.

### DIFF
--- a/PagesGallery/PagesGallery.Droid/PagesGallery.Droid.csproj
+++ b/PagesGallery/PagesGallery.Droid/PagesGallery.Droid.csproj
@@ -212,6 +212,10 @@
     <AndroidResource Include="Resources\values\styles.xml" />
   </ItemGroup>
   <Import Project="$(MSBuildExtensionsPath)\Xamarin\Android\Xamarin.Android.CSharp.targets" />
+  
+  <Import Project="..\..\.nuspec\Xamarin.Forms.targets" Condition="'$(BuildingInsideVisualStudio)' == 'true' AND Exists('..\..\.nuspec\Xamarin.Forms.Build.Tasks.dll')" />
+  <Import Project="..\..\.nuspec\Xamarin.Forms.targets" Condition="'$(BuildingInsideVisualStudio)' != 'true'" />
+  
   <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
     <PropertyGroup>
       <ErrorText>This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is {0}.</ErrorText>
--- a/PagesGallery/PagesGallery.Droid/PagesGallery.Droid.csproj
+++ b/PagesGallery/PagesGallery.Droid/PagesGallery.Droid.csproj
@@ -16,7 +16,6 @@
     <AndroidResgenFile>Resources\Resource.Designer.cs</AndroidResgenFile>
     <GenerateSerializationAssemblies>Off</GenerateSerializationAssemblies>
     <AndroidManifest>Properties\AndroidManifest.xml</AndroidManifest>
-    <AndroidUseLatestPlatformSdk>true</AndroidUseLatestPlatformSdk>
     <TargetFrameworkVersion>v8.0</TargetFrameworkVersion>
     <AndroidSupportedAbis>armeabi,armeabi-v7a,x86</AndroidSupportedAbis>
     <AndroidStoreUncompressedFileExtensions />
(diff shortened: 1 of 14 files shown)


PagesGallery/PagesGallery.Droid/PagesGallery.Droid.csproj
```diff
@@ -205,7 +212,6 @@
     <AndroidResource Include="Resources\values\styles.xml" />
   </ItemGroup>
   <Import Project="$(MSBuildExtensionsPath)\Xamarin\Android\Xamarin.Android.CSharp.targets" />
-  <Import Project="..\..\.nuspec\Xamarin.Forms.targets" />
```
StephaneDelcroix: if this project was referencing the XF nuget, it would import the `.targets` (even if it does not do anything for this project at this moment).

to stay future-proof, and make sure that, when we add a special rule in `.targets` it is applied in our tests just like it would in user's projects, please keep this import in.

this comment applies to every platform head where the Import was removed, obviously

Xamarin.Forms.Controls/Xamarin.Forms.Controls.csproj
```diff
@@ -30,7 +30,9 @@
     <ProjectReference Include="..\Xamarin.Forms.Maps\Xamarin.Forms.Maps.csproj" />
     <ProjectReference Include="..\Xamarin.Forms.Xaml\Xamarin.Forms.Xaml.csproj" />
   </ItemGroup>
-  <Import Project="..\.nuspec\Xamarin.Forms.targets" />
+  
+  <Import Project="..\.nuspec\Xamarin.Forms.targets" Condition="'$(BuildingInsideVisualStudio)' == 'true' AND Exists('..\.nuspec\Xamarin.Forms.Build.Tasks.dll')" />
+  <Import Project="..\.nuspec\Xamarin.Forms.targets" Condition="'$(BuildingInsideVisualStudio)' != 'true'" />
```
StephaneDelcroix: I don't get how the `.targets` will be included if the `Build.Tasks` hasn't been built yet. This csproj has a dependency on the `Build.Tasks` (iirc) and will built it, but that won't retrigger the parsing of the csproj and the reevaluation of the condition. Or will it ?

PagesGallery/PagesGallery.Droid/PagesGallery.Droid.csproj
hartez: Fixed.

Xamarin.Forms.Controls/Xamarin.Forms.Controls.csproj
hartez: To be honest, I'm not sure if VS is re-evaluating the condition or if VS just never considers XF.targets to be "imported". But MSBuild doesn't evaluate that condition until after the dependencies have been built, so the build works fine even if Build.Tasks doesn't exist when it starts.

But if VS considers the project load to have "failed" initially, it won't build the project and it won't change that "failed" status until the project is re-loaded. If it considers it "successful", it will happily hand it off to MSBuild to do the rest.

jassmith:
Build fails :(

hartez:
> Build fails :(

Some sort of VSTS glitch, it's running again now.

PR #2091, opened 2018-03-13
Editor Auto Resize with Text Changes

### LINKED ISSUE #1733: Autoresizable Editor
# Rationale
Currently when using an Editor it will initially render at the height of its content. Once this rendering has occurred the height of the Editor will not flex vertically as the user types thus causing the text to flow off the control or it will just leave large blank space where the control can just decrease in size. If the control occupies a layout where there is available vertical space then it should grow/decrease in size vertically as user types according to it's allowable bounding region.  

## Implementation

```c#

public enum SizeOptions
{
Default,
AutoFitToText
}

public SizeOptions SizeOptions // BP
```

Reviewing the platforms it seems like Android is the only one where the editor will re-flow automatically as the user types. iOS and UWP require watching the textchange and then resizing the control

Because of how Xamarin Forms measures the layout of VisualElements the Measure on the elements will need to be invalidated (efficiently if possible) and recalculated as the content of the Editor Field changes

### Android

- As the content of the Editor field changes the calculated height for the Android Native element should conform to the contained content

### iOS

See Android

### UWP

See Android

## Backward Compatibility

Since this is new behavior it should be fully backwards compatible

## Difficulty : Easy/Medium

The basic idea isn't very difficult to implement. I've tested invalidating the measure as the user types and the editor resizes accordingly.  If possible extra work should be done to ensure that these invalidation's happen efficiently

PureWeen:
### Description of Change ###
fixes #1733

Added a property to Editor that will cause it to grow dynamically as the content of the Editor is changed. Basically as the text is changed the Editors size is recalculated and re-measured

#### Android
Android was the least amount of work and was able to resize itself purely based on having its measure invalidated.

#### iOS
iOS almost worked as seamlessly as Android. 
When a new line is added to the UITextView the vertical growth happens after the view has already scrolled meaning now you had a line of text partially scrolled off screen. I added an event to watch for Frame Changes on the UITextView and if these happened then I forced the ScrollPosition to go back to *(0,0)*  If the Editor is set to AutoSize then the Frame of the UITextView should never be scrolled if the internal content hasn't grown outside the max constraints of the UITextView. Once the UITextView has grown as much as it can then FrameChange events no longer fire so it's able to still scroll without any issues.

#### UWP 
UWP proved to be the most challenging to coerce into resizing to the text as the text changed.

Purely invalidating the layout as text is added to the TextBox won't cause it to expand horizontally.
If the TextBox is set to WordWrap and it's part of the layout it will refuse to Measure itself beyond its established width. Even giving it infinite constraints will cause it to always set its DesiredSize to the same width but with a vertical growth. The only way I was able to grow it was by setting layout renderers width explicitly to some value but then it just set its own Width to that Width which is not helpful. Even vertically it would measure oddly in cases of rapid text changes. Holding down the backspace key or enter key would cause the final result to be not quite right. Both of these issues were fixed by just creating a static TextBox that is not part of the layout which let me just measure the size of the text as it would fit into the TextBox unconstrained and then just return that Size from the GetDesiredSize call.

Before pictures if you added text to an already rendered Editor
image

Setting AutoSize to TextChanges will now resize the Editors to fit the text based on how much the Editor is allowed to grow
image

### API Changes ###

Added:
 - public EditorAutoSizeOption AutoSize {get;set;} //Bindable Property

```csharp
namespace Xamarin.Forms
{
	public enum EditorAutoSizeOption
	{
		Disabled = 0,
		TextChanges = 1
	}
}
```

### Behavioral Changes ###

By default this is disabled so users should notice no behavioral changes unless they opt in for this setting

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/EditorRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/EditorRenderer.cs
@@ -142,14 +142,14 @@ namespace Xamarin.Forms.Platform.UWP
 		}
 
 		/*
-		 * Purely invalidating the layout as text is added to the TextBox won't cause it to expand.
-		 * If the TextBox is set to WordWrap and it's part of the layout it will refuse to Measure itself beyond it's established width.
-		 * Giving it infinite contraints it will always just come back the same width but with a vertical growth.
-		 * The only way I was able to grow it was by setting this.Width to some value but then it just set it's own Width to that Width which isn't helpful.
+		 * Purely invalidating the layout as text is added to the TextBox will not cause it to expand.
+		 * If the TextBox is set to WordWrap and it is part of the layout it will refuse to Measure itself beyond its established width.
+		 * Even giving it infinite constraints will cause it to always set its DesiredSize to the same width but with a vertical growth.
+		 * The only way I was able to grow it was by setting layout renderers width explicitly to some value but then it just set its own Width to that Width which is not helpful.
 		 * Even vertically it would measure oddly in cases of rapid text changes.
 		 * Holding down the backspace key or enter key would cause the final result to be not quite right.
-		 * Both of these issues were fixed by just creating a static TextBox that isn't part of the layout which let me just measure
-		 * the size of the text as it would fit into the TextBox unconstrained.
+		 * Both of these issues were fixed by just creating a static TextBox that is not part of the layout which let me just measure
+		 * the size of the text as it would fit into the TextBox unconstrained and then just return that Size from the GetDesiredSize call.
 		 * */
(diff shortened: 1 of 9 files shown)


Xamarin.Forms.Platform.UAP/EditorRenderer.cs
```diff
@@ -10,25 +11,33 @@ namespace Xamarin.Forms.Platform.UWP
 {
 	public class EditorRenderer : ViewRenderer<Editor, FormsTextBox>
 	{
+		private static FormsTextBox _copyOfTextBox;
```
samhouts: 👮 private keyword

PR #2103, opened 2018-03-14
Mark ctors public

### LINKED ISSUE #1797: Open/Closed principle missing
### Description

I try to answer questions on stackoverflow from time to time. And always think "Ok this is a 5min Task. Just override a Renderer and that's it." But not with Xamarin.Forms. :( `internal` is used all over the place and this makes it impossible to extend native views without rewriting whole renderers. I encountered the following problem back to when Xamarin.Forms was closed source and I had to decompile it just to find out it is often impossible to do some customization.

### Steps to Reproduce (just an example)

1. Imagine you want to override a method in a EditText (Entry) in Android
2. Look up the native Renderer [EntryRenderer](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/Renderers/EntryRenderer.cs). Ok this is creating a `FormsEditText` as native control.
3. Look up the native view [FormsEditText](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/Renderers/FormsEditText.cs)
4. Try to inherit it. Doesn't work, because of an internal constructor.
5. Inherit from `EntryRenderer` and try to override `CreateNativeControl`. Damn! It has to return a `FormsEditText` and you can't inherit it (see 4.)
6. Solution: Copy all the code from `EntryRenderer` and `FormsEditText` prefix them with `My` and add the tiny override you intended to do in step 1.

### Expected Behavior
1. inherit `MyFormsEditText`  from `FormsEditText` override your method
2. inherit `MyEntryRenderer` from `EntryRenderer` return your new `MyFormsEditText` 
3. done.

### Actual Behavior
see steps to reproduce 2. - 6.

### Basic Information

- Version with issue: latest
- IDE: all of them
- Platform Target Frameworks:
  - Android: all 

### General Questions / Remarks
- This is just an example. Why are **internal / non virtual / extension** methods used so widely in the code? **You can't override them**
- Is there a general design decision document where this style was decided for a good reason?
- The problem is: The developer is forced to copy a lot of code from the Xamarin.Forms repo. He will miss bugfixes and gets a lot of unmaintained code :(
- it's **inconsistent** (see [FormsTextView](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/Renderers/FormsTextView.cs) (public/protected ctors) vs  [FormsEditText](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/Renderers/FormsEditText.cs) (internal ctor))

jassmith:
### Description of Change ###

Makes FormsEditText, EntryEditText, and EditorEditText constructors public. Consumption of these classes is generally a bad idea.

### Bugs Fixed ###

- Fixes #1797

### API Changes ###

Description says it all

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/FormsEditText.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/FormsEditText.cs
@@ -11,7 +11,7 @@ namespace Xamarin.Forms.Platform.Android
 	{
 		DescendantFocusToggler _descendantFocusToggler;
 
-		internal FormsEditText(Context context) : base(context)
+		public FormsEditText(Context context) : base(context)
 		{
 			DrawableCompat.Wrap(Background);
 		}
@@ -47,7 +47,7 @@ namespace Xamarin.Forms.Platform.Android
 	[Obsolete("EntryEditText is obsolete as of version 2.4.0. Please use Xamarin.Forms.Platform.Android.FormsEditText instead.")]
 	public class EntryEditText : FormsEditText
 	{
-		internal EntryEditText(Context context) : base(context)
+		public EntryEditText(Context context) : base(context)
 		{
 		}
 	}
@@ -55,7 +55,7 @@ namespace Xamarin.Forms.Platform.Android
 	[Obsolete("EditorEditText is obsolete as of version 2.4.0. Please use Xamarin.Forms.Platform.Android.FormsEditText instead.")]
 	public class EditorEditText : FormsEditText
 	{
-		internal EditorEditText(Context context) : base(context)
+		public EditorEditText(Context context) : base(context)
 		{
 		}
 	}


hartez:
> Consumption of these classes is generally a bad idea.

Consumption of these classes for purposes other than customizing a renderer is generally a bad idea.

Xamarin.Forms.Platform.Android/Renderers/FormsEditText.cs
```diff
@@ -47,15 +47,15 @@ public override bool RequestFocus(FocusSearchDirection direction, Rect previousl
 	[Obsolete("EntryEditText is obsolete as of version 2.4.0. Please use Xamarin.Forms.Platform.Android.FormsEditText instead.")]
 	public class EntryEditText : FormsEditText
 	{
-		internal EntryEditText(Context context) : base(context)
```
samhouts: Even the obsolete ones?

PR #2106, opened 2018-03-15
Try/Catch assembly loading so previewer doesn't have a fit

jassmith:
Fix for previewer, no bug filed.

### DIFF
--- a/Xamarin.Forms.Core/DependencyService.cs
+++ b/Xamarin.Forms.Core/DependencyService.cs
@@ -91,7 +91,17 @@ namespace Xamarin.Forms
 			// Naive implementation can easily take over a second to run
 			foreach (Assembly assembly in assemblies)
 			{
-				Attribute[] attributes = assembly.GetCustomAttributes(targetAttrType).ToArray();
+				Attribute[] attributes;
+				try
+				{
+					attributes = assembly.GetCustomAttributes(targetAttrType).ToArray();
+				}
+				catch (System.IO.FileNotFoundException)
+				{
+					// Sometimes the previewer doesn't actually have everything required for these loads to work
+					continue;
+				}
+				
 				if (attributes.Length == 0)
 					continue;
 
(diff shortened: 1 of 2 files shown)


Xamarin.Forms.Core/DependencyService.cs
```diff
@@ -91,7 +91,17 @@ internal static void Initialize(Assembly[] assemblies)
...
-				Attribute[] attributes = assembly.GetCustomAttributes(targetAttrType).ToArray();
+				Attribute[] attributes;
+				try
+				{
+					attributes = assembly.GetCustomAttributes(targetAttrType).ToArray();
+				}
+				catch (System.IO.FileNotFoundException)
+				{
+					// Sometimes the previewer doesn't actually have everything required for these loads to work
+					continue;
```
hartez: Can we at least log a warning for this so someone who's not using the Previewer can see that a problem is occurring (rather than pulling out their hair trying to determine why their custom renderer isn't loading)?

Xamarin.Forms.Core/Registrar.cs
```diff
@@ -194,7 +194,16 @@ public static void RegisterAll(Type[] attrTypes)
...
-					Attribute[] attributes = assembly.GetCustomAttributes(attrType).ToArray();
+					Attribute[] attributes;
+					try
+					{
+						attributes = assembly.GetCustomAttributes(attrType).ToArray();
+					}
+					catch (System.IO.FileNotFoundException)
+					{
+						// Sometimes the previewer doesn't actually have everything required for these loads to work
+						continue;
```
hartez: Same - need to log this. Or, better yet, have some sort of indication we're running under the Previewer so we know this is an exception we can ignore.

jassmith:
Not waiting for stephane to re-review, we have to get this done. It doesn't really matter, this try/catch will never hit in deployment, only for previewer and other half baked tooling environments.

StephaneDelcroix:
> Not waiting for Stéphane to re-review,

that's the only sane option

jassmith:
I like that you corrected my spelling of your name :P Seriously though we don't need to add a bunch of checks for this, we are bikeshedding. The failure case that causes this can't happen with deployed apps...

jassmith:
Also its like 11pm there, go to bed or at least stop reading work emails. Have some fun :P

PR #2111, opened 2018-03-15
[iOS] Encode url for UIWebView so it doesn't fail with non-ascii characters 

### LINKED ISSUE #1583: WebView fails to load from urlwebviewsource with non-ascii characters (works with Uri)
### Description
When loading a WebView with the 'Source' set as a Uri the webview is quite tolerant.  If the 'Source' is setting using a UrlWebviewSource non-ascii characters cause the view to throw an exception

### Steps to Reproduce

1. create a page with a webview
2. set the Source equal to a url with a non-ascii character

example :
```
<WebView x:Name="WebView" HorizontalOptions="FillAndExpand"
                     VerticalOptions="FillAndExpand">
 </WebView>
```

```
WebView.Source = new UrlWebViewSource { Url = "https://www.google.no/maps/place/Skøyen" };
```

### Expected Behavior
it should escape the url as needed and load without exception. When setting the source with a new Uri object it behaves as expected.

### Actual Behavior
Could not initialize an instance of the type 'Foundation.NSUrl': the native 'initWithString:' method returned nil.
It is possible to ignore this condition by setting MonoTouch.ObjCRuntime.Class.ThrowOnInitFailure to false.

### Basic Information

Currently on Xamarin Forms
2.5.0.121934

rmarinho:
### Description of Change ###

Encode url so it doesn't fail with non-ascii characters

### Bugs Fixed ###

fixes #1583 

### API Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1583_1.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1583_1.cs
@@ -28,7 +28,7 @@ namespace Xamarin.Forms.Controls.Issues
 		public void Issue1583Test ()
 		{
 			RunningApp.WaitForElement (q => q.Marked ("webview"));
-			RunningApp.Screenshot ("I didn'crash and i can see Skøyen");
+			RunningApp.Screenshot ("I didn't crash and i can see Skøyen");
 		}
 #endif
 	}
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1583_1.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1583_1.cs
@@ -1,0 +1,35 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1583_1.cs
```diff
@@ -0,0 +1,35 @@
...
+
+			Content = webview;
+		}
+
+#if UITEST
+		[Test]
+		public void Issue1583Test ()
+		{
+			RunningApp.WaitForElement (q => q.Marked ("webview"));
+			RunningApp.Screenshot ("I didn'crash and i can see Skøyen");
```
samhouts: I didn'crash >> I didn't crash

rmarinho:
build --uitests

Xamarin.Forms.Platform.iOS/Renderers/WebViewRenderer.cs
```diff
@@ -79,7 +79,8 @@ public void LoadHtml(string html, string baseUrl)
 
 		public void LoadUrl(string url)
 		{
-			LoadRequest(new NSUrlRequest(new NSUrl(url)));
+			var encodedStringUrl = new NSString(url).CreateStringByAddingPercentEscapes(NSStringEncoding.UTF8);
+			LoadRequest(new NSUrlRequest(new NSUrl(encodedStringUrl)));
```
amalmo: @rmarinho As a result of this, we now experienced double encoding of our URL´s. A developer on our team also pointed out that his method is marked as deprecated by [Apple](https://developer.apple.com/documentation/foundation/nsstring/1415058-stringbyaddingpercentescapesusin).

We are currently using IdentityModel´s RequestUrl to generate a Authorization URL, which already encode a redirect URL parameter, and as a result of this it is now double encoded.

PR #1935, opened 2018-02-21
[Android] Buttons can be round; remove Padding & Shadow by default

samhouts:
### Description of Change ###

Removing the Padding and Shadow that were added in #1570 to make the X.F. Buttons mimic the style of a default Android `Button`. These can be added back to Buttons using new Platform Specifics.

Also adjusting the padding mismatch that made circle Buttons appear to be slightly rectangular.

Before #1570:
before-1570

After #1570:
before

After this change (note, changed the test case for an additional parameter):
screenshot_20180227-124138

### Bugs Fixed ###

- fixes #1909
- fixes #1954 

### API Changes ###

Added:
- `bool Button.On<Android>().UseDefaultPadding()`
- `void Button.On<Android>().SetUseDefaultPadding(true/false)`
- `bool Button.On<Android>().UseDefaultShadow()`
- `void Button.On<Android>().SetUseDefaultShadow(true/false)`

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #2114, opened 2018-03-15
[Android] Button PaddingLeft now accounts for ShadowDx

PureWeen:
### Description of Change ###

As part of https://github.com/xamarin/Xamarin.Forms/pull/1935, code was added to figure in the padding needed for the background drawable

```csharp
		float PaddingLeft
		{
			get { return (_paddingLeft / 2f); }
			set { _paddingLeft = value; }
		}

		float PaddingTop
		{
			get { return (_paddingTop / 2f) + _shadowDy; }
			set { _paddingTop = value; }
		}
```

But this padding was only added for PaddingTop so the background rectangle won't inset properly relative to the ShadowLayer being added if there's a non zero ShadowX value.

This doesn't really come into play with Button as it always has a zero for the shadowDx but I added some APIs to leverage these more with ImageButton. So the screen shots below are from that

Here it is without the padding
image

Here it is with the padding
```
		float PaddingLeft
		{
			get { return (_paddingLeft / 2f) + _shadowDx; }
			set { _paddingLeft = value; }
		}
```

image

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/ButtonDrawable.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ButtonDrawable.cs
@@ -26,7 +26,7 @@ namespace Xamarin.Forms.Platform.Android
 
 		float PaddingLeft
 		{
-			get { return (_paddingLeft / 2f); }
+			get { return (_paddingLeft / 2f) + _shadowDx; }
 			set { _paddingLeft = value; }
 		}
 


StephaneDelcroix:
Added the do-not-merge until this is retargetted

PureWeen:
This change looks to be compatible with either of those branches

PureWeen:
Confirmed with @jassmith we'd retarget this for 15-5

So I've retargeted and removed the do-not-merge tag

Issue #2104, opened 2018-03-14
If possible remove the use of async void inside FormsAppCompatActivity life cycle methods


PureWeen:
### Description
Currently because of
```
await OnStateChanged();
```
Most of the lifecycle methods inside FormsAppCompatActivity are using async void

This causes any exceptions that are raised in these methods to just be swallowed causing the application to just continue along and eventually crash somewhere else because necessary code never ran

Changing the implementation so these aren't async void would be ideal

### Steps to Reproduce
To reproduce you'd have to manually add an exception into the lifecycle methods.

I came across this because of a different issue that causes Resources to not properly regenerate which leads to the inflation of the Toolbar to fail

```
bar = LayoutInflater.Inflate(ToolbarResource, null).JavaCast<AToolbar>();
```

If that fails then all the code afterwards never runs and then at some random point the app will crash when that initialization code becomes necessary but the root cause has been swallowed

### Expected Behavior
If something throws an exception during the Xamarin Forms life cycle method that exception should not be swallowed or at least be logged in order to give a clear indication of the problem. 

### Actual Behavior
Exception is swallowed

### Implementation Thoughts

At this point the way 
```
OnStateChanged
```
Is executed it's basically fire and forget
There isn't any code that runs afterwards that is dependent on the results of *OnStateChanged* finishing. 

Currently *SendSleepAsync()* is marked as Never Browsable
```
		[EditorBrowsable(EditorBrowsableState.Never)]
		public Task SendSleepAsync()
		{
			OnSleep();
			return SavePropertiesAsync();
		}
```

If that gives us the latitude to make a slight breaking change then we can create a de-async version of SaveProperties
```
		public async Task SavePropertiesAsync()
		{
			if (Device.IsInvokeRequired)
				Device.BeginInvokeOnMainThread(async () => await SetPropertiesAsync());
			else
				await SetPropertiesAsync();
		}

		internal async void SaveProperties()
		{
			try
			{
				await SavePropertiesAsync();
			}
			catch(Exception exc)
			{
				Internals.Log.Warning(nameof(Application), $"Failed to Save Properties: {exc}");
			}
		}
```

And then just de-async upwards from there just allowing for a 
```
void OnStateChanged
```

Another possibility would be if there's a safe way where we can ensure no blocking with a call to
```
_application.SendSleepAsync().Wait()
```

I know with Akavache the recommendation is to override life cycle events and then block the lifecycle event until the cache is flushed

[...]

PR #2122, opened 2018-03-16
Create Sync version of SendSleep so lifecycle methods don't have to be async void

PureWeen:
### Description of Change ###

SaveProperties being async caused a lot of the life cycle methods on the platforms to become *async void* which means that exceptions were just being swallowed causing the Application to fail much later after the initial cause had been swallowed

Outside of UWP none of the life cycle methods depend on the results of the call. They all just called

```
await _application.SendSleepAsync();
```
at the end of the method

I created an internal
```C#
SavePropertiesAsFireAndForget
```

Which can now be used by
```C#
SendSleep 
```

Which causes the call to execute in the same fire and forget fashion, except now it doesn't require a bunch of *async voids* to propagate up the call tree

So in principal it should be the same as it was before except now important exceptions during life cycle events should occur and make it more apparent why an application is failing

The *async voids* on life cycle methods came around from this PR
https://github.com/xamarin/Xamarin.Forms/pull/1075
But even the PR itself just mentions adding those just in case

I added a UI Test with this PR as well and if I comment out the *SemaphoreSLims* that were added as part of PR 1075 then I get the Shared Violation Exceptions PR 1075 is fixing. So the addition of the SemaphoreSlims is all that's really needed to combat the Shared Violation Exceptions and nothing was really gained by making the life cycle methods *async void*

### Bugs Fixed ###

https://github.com/xamarin/Xamarin.Forms/issues/2104

### API Changes ###

Added
```C#
		public void SendSleep()
		{
			OnSleep();
			SavePropertiesAsFireAndForget();
		}
```

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Application.cs
+++ b/Xamarin.Forms.Core/Application.cs
@@ -59,7 +59,7 @@ namespace Xamarin.Forms
 		public static Application Current
 		{
 			get { return s_current; }
-			set 
+			set
 			{
 				if (s_current == value)
 					return;
@@ -138,7 +138,8 @@ namespace Xamarin.Forms
 
 		public ResourceDictionary Resources
 		{
-			get {
+			get
+			{
 				if (_resources != null)
 					return _resources;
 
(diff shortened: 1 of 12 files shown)


rmarinho:
build --uitests

Xamarin.Forms.Core/Application.cs
```diff
@@ -173,12 +174,42 @@ public ResourceDictionary Resources
...
+			{
+				Internals.Log.Warning(nameof(Application), $"Exception while saving Application Properties: {exc}");
+			}
+		}
+
 		public async Task SavePropertiesAsync()
 		{
 			if (Device.IsInvokeRequired)
-				Device.BeginInvokeOnMainThread(async () => await SetPropertiesAsync());
+			{
```
rmarinho: no biggie but I think we don't put bracket in 1 liners

PR #2131, opened 2018-03-19
[C] avoid exception on overriden indexer

StephaneDelcroix:
### Description of Change ###

Binding to a Newtonsoft.Json.Linq.JObject with an indexer throws an
AmbiguousMatchException (because the Indexer is overriden). This fixes
the property finding code so it doesn't thorw, but returns the moste
derived indexer.

### PR Checklist ###


StephaneDelcroix:
not sure it's critical...

PR #2132, opened 2018-03-19
Adding line height to Label and Span

### LINKED ISSUE #1734: Label line-height
# Rationale

Customize vertical appearance of label by setting line-height

# Implementation

Add a LineHeightProperty

# Back-Compat

default should be `-1` so when the value isn't set, the renderers knows that they have to ignore the value

# CSS

`line-height`

# Thoughts

Should this be supported at Span level too, as a TextElement or something ?

henricm:
### Description of Change ###

Fixes #1734 by adding LineHeight property to Label and Span. The unit of LineHeight is relative to the original line height or font size. Also know as *multiplier* on iOS and Android. For UWP the actual line height will be calculated as a multiple of the font size.

Setting LineHeight on Span doesn't work on UWP which doesn't
seem to support that on its `Run` class.

On Android the separate LineHeight on span only has effect if
the text has linebreaks within the string. Then it's possible
to set different line heights for individual lines using Span.
If the lines are automatically split using WordWrap for instance,
it seems like Android will use the LineHeight set on the last Span.

### API Changes ###

Added:
 - double Label.LineHeight { get; set; } //Bindable Property specifying the relative line height.
 - double Span.LineHeight { get; set; } //Bindable Property specifying the relative line height for a segment of the Label. Only works properly on Android currently.
- Added CSS property `line-height` on `Label`.

### Behavioral Changes ###

LineHeight can now be set on labels and spans.

### Before and after screenshots ###

Before:
lineheight-before-capture

After:
span line height after

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/ILineHeightElement.cs
+++ b/Xamarin.Forms.Core/ILineHeightElement.cs
@@ -1,0 +1,12 @@
+using System.ComponentModel;
+
+namespace Xamarin.Forms.Internals
+{
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	interface ILineHeightElement
+	{
+		double LineHeight { get; }
+
+		void OnLineHeightChanged(double oldValue, double newValue);
+	}
+}
--- a/Xamarin.Forms.Core/Label.cs
+++ b/Xamarin.Forms.Core/Label.cs
@@ -11,7 +11,7 @@ namespace Xamarin.Forms
 {
 	[ContentProperty("Text")]
 	[RenderWith(typeof(_LabelRenderer))]
-	public class Label : View, IFontElement, ITextElement, ITextAlignmentElement, IElementConfiguration<Label>
+	public class Label : View, IFontElement, ITextElement, ITextAlignmentElement, ILineHeightElement, IElementConfiguration<Label>
 	{
 		public static readonly BindableProperty HorizontalTextAlignmentProperty = TextAlignmentElement.HorizontalTextAlignmentProperty;
 
@@ -76,7 +76,7 @@ namespace Xamarin.Forms
 		public static readonly BindableProperty LineBreakModeProperty = BindableProperty.Create(nameof(LineBreakMode), typeof(LineBreakMode), typeof(Label), LineBreakMode.WordWrap,
 			propertyChanged: (bindable, oldvalue, newvalue) => ((Label)bindable).InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged));
 
-		public static readonly BindableProperty LineHeightProperty = BindableProperty.Create("LineHeight", typeof(double), typeof(Label), -1.0);
+		public static readonly BindableProperty LineHeightProperty = LineHeightElement.LineHeightProperty;
 
 		readonly Lazy<PlatformConfigurationRegistry<Label>> _platformConfigurationRegistry;
 
@@ -188,6 +188,8 @@ namespace Xamarin.Forms
 
 		void IFontElement.OnFontChanged(Font oldValue, Font newValue) =>
 			 InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
+		void ILineHeightElement.OnLineHeightChanged(double oldValue, double newValue) =>
+			InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
 
 		void OnFormattedTextChanged(object sender, PropertyChangedEventArgs e)
 		{
--- a/Xamarin.Forms.Core/Label.cs
+++ b/Xamarin.Forms.Core/Label.cs
@@ -76,6 +76,8 @@ namespace Xamarin.Forms
 		public static readonly BindableProperty LineBreakModeProperty = BindableProperty.Create(nameof(LineBreakMode), typeof(LineBreakMode), typeof(Label), LineBreakMode.WordWrap,
 			propertyChanged: (bindable, oldvalue, newvalue) => ((Label)bindable).InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged));
 
+		public static readonly BindableProperty LineHeightProperty = BindableProperty.Create("LineHeight", typeof(double), typeof(Label), -1.0);
+
 		readonly Lazy<PlatformConfigurationRegistry<Label>> _platformConfigurationRegistry;
 
 		public Label()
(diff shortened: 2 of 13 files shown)


Xamarin.Forms.Core/Label.cs
```diff
@@ -58,6 +58,8 @@ public class Label : View, IFontElement, ITextElement, ITextAlignmentElement, IE
 		public static readonly BindableProperty LineBreakModeProperty = BindableProperty.Create(nameof(LineBreakMode), typeof(LineBreakMode), typeof(Label), LineBreakMode.WordWrap,
 			propertyChanged: (bindable, oldvalue, newvalue) => ((Label)bindable).InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged));
 
+		public static readonly BindableProperty LineHeightProperty = BindableProperty.Create("LineHeight", typeof(double), typeof(Label), -1.0);
```
StephaneDelcroix: - use `nameof(LineHeight)`
- you need to InvalidateMeasureInternal on line-height change
- I prefer `-1d` to `-1.0`, but that's probably just me

Xamarin.Forms.Core/Properties/AssemblyInfo.cs
```diff
@@ -61,6 +61,7 @@
 [assembly: StyleProperty("text-align", typeof(ITextAlignmentElement), nameof(TextAlignmentElement.HorizontalTextAlignmentProperty), Inherited = true)]
 [assembly: StyleProperty("visibility", typeof(VisualElement), nameof(VisualElement.IsVisibleProperty), Inherited = true)]
 [assembly: StyleProperty("width", typeof(VisualElement), nameof(VisualElement.WidthRequestProperty))]
+[assembly: StyleProperty("line-height", typeof(Label), nameof(Label.LineHeightProperty))]
```
StephaneDelcroix: `line-height` is inherited in CSS, set `Inherited = true`

StephaneDelcroix: that won't set the line-height on Span. both Span and Label have to implement a common interface (look at `IFontElement`)

Xamarin.Forms.Core/Span.cs
```diff
@@ -67,6 +67,8 @@ public string Text
 
 		public static readonly BindableProperty FontAttributesProperty = FontElement.FontAttributesProperty;
 
+		public static readonly BindableProperty LineHeightProperty = Label.LineHeightProperty;
```
StephaneDelcroix: see previous comment, should be something like `LineHeightElement.LineHeightProperty`

henricm: Should we add it to `ITextElement` and `TextElement` or create `ILineHeightElement/LineHeightElement`? Noticed you mentioned TextElement in the issue description.

henricm:
Thanks for the feedback @StephaneDelcroix! I've updated with new commits now.

henricm:
@samhouts of course! Added screenshots now.

StephaneDelcroix:
@henricm 
> Of course! Added screenshots now.

it doesn't seem to work on iOS

Xamarin.Forms.Core/ILineHeightElement.cs
```diff
@@ -0,0 +1,12 @@
+using System.ComponentModel;
+
+namespace Xamarin.Forms.Internals
+{
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	public interface ILineHeightElement
```
StephaneDelcroix: why is this public ?

henricm:
@StephaneDelcroix I described this in the PR description "On iOS the LineHeight set on the first Span will be used. Different LineHeights seems to not be supported."

Xamarin.Forms.Core/Properties/AssemblyInfo.cs
```diff
@@ -61,6 +61,7 @@
 [assembly: StyleProperty("text-align", typeof(ITextAlignmentElement), nameof(TextAlignmentElement.HorizontalTextAlignmentProperty), Inherited = true)]
 [assembly: StyleProperty("visibility", typeof(VisualElement), nameof(VisualElement.IsVisibleProperty), Inherited = true)]
 [assembly: StyleProperty("width", typeof(VisualElement), nameof(VisualElement.WidthRequestProperty))]
+[assembly: StyleProperty("line-height", typeof(Label), nameof(Label.LineHeightProperty), Inherited = true)]
```
StephaneDelcroix: [assembly: StyleProperty("line-height", typeof(ILineHeightElement), nameof(LineHeightElement.LineHeightProperty), Inherited = true)]

Xamarin.Forms.Core/Span.cs
```diff
@@ -115,5 +122,9 @@ void IFontElement.OnFontChanged(Font oldValue, Font newValue)
 		void ITextElement.OnTextColorPropertyChanged(Color oldValue, Color newValue)
 		{
 		}
+
+		public void OnLineHeightChanged(double oldValue, double newValue)
```
StephaneDelcroix: void ILineHeightElement.OnLineHeightChanged

Xamarin.Forms.Core/ILineHeightElement.cs
henricm: Because I'm used to java ;-). Will fix.

henricm:
@StephaneDelcroix I managed to get a basic Xamarin.iOS sample to work with different line heights within the label. Will continue to look if I can get that same approach to work in forms.

henricm:
@rmarinho true. will fix as I'm cleaning up the commits and rebasing now. Will push soonish.

henricm:
Added fix for span line heights on iOS. Setting LineBreakMode before updating the text (where the paragraph style attribute is added) did the trick.
Rebased and updated PR description with new after screenshot for iOS.

henricm:
@StephaneDelcroix rebased as of today. Please have a look when you have time. Thanks!

titonton:
For some reason this is not working on my project. 
``
 <Label x:Name="NameBinding" Text="{Binding Name}" HorizontalTextAlignment="Center" TextColor="#002e62" 
               HorizontalOptions="CenterAndExpand" 
               VerticalOptions="StartAndExpand" 
               FontFamily='{OnPlatform Android="SourceSansPro-SemiBold.ttf#SourceSansPro-SemiBold", iOS="SourceSansPro-SemiBold"}'
               LineHeight="10.0"/>
``

PR #2140, opened 2018-03-20
[Enhancement] Add EvaluateJavaScript method to WebView

### LINKED ISSUE #1699: [Enhancement] Add EvaluateJavaScript method to WebView
## Rationale

Forms does not currently have a simple cross-platform way to call JavaScript methods in a WebView and return the results.

## Implementation

Add a method to WebView, something like:

`public async Task<string> EvaluateJavaScript(string script);`

The renderers should map this to InvokeScriptAsync on UWP, EvaluateJavascript on iOS, and EvaluateJavascript on Android.

## Backward Compatibility

This is a brand new method, so there should not be any backward compatibility issues.

## Difficulty: Moderate

alanag13:
### Description of Change ###

fixes #1699 .

Adds the ability to execute and return the result of a passed javascript string from a platform's webview.

Platform support:

- [X] Android
- [X] iOS
- [X] macOS
- [X] UWP
- [X] WPF
- [ ] Tizen (no native platform support)
- [ ] GTK (no native platform support)

### API Changes ###

Added:
Task<string> WebView.EvaluateJavaScriptAsync

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/WebView.cs
+++ b/Xamarin.Forms.Core/WebView.cs
@@ -3,6 +3,9 @@ using System.ComponentModel;
 using System.Threading.Tasks;
 using Xamarin.Forms.Internals;
 using Xamarin.Forms.Platform;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Collections.Generic;
 
 namespace Xamarin.Forms
 {
@@ -81,7 +84,28 @@ namespace Xamarin.Forms
 		public async Task<string> EvaluateJavaScriptAsync(string script)
 		{
 			EvaluateJavaScriptDelegate handler = EvaluateJavaScriptRequested;
-			return await handler?.Invoke(script);
+
+			//make all the platforms mimic Android's implementation, which is by far the most complete.
+			if (Xamarin.Forms.Device.RuntimePlatform != "Android")
+			{
+				script = EscapeJsString(script);
+				script = "try{JSON.stringify(eval('" + script + "'))}catch(e){'null'};";
+			}
+
+			var result = await handler?.Invoke(script);
+
+			//if the js function errored or returned null/undefined treat it as null
+			if (result == "null")
+				result = null;
+
+			//JSON.stringify wraps the result in literal quotes, we just want the actual returned result
+			//note that if the js function returns the string "null" we will get here and not above
+			else
+				result = result.Trim('"');
+
+
+
+			return result;
 		}
 
 		public void GoBack()
@@ -154,6 +178,39 @@ namespace Xamarin.Forms
 		public IPlatformElementConfiguration<T, WebView> On<T>() where T : IConfigPlatform
 		{
 			return _platformConfigurationRegistry.Value.On<T>();
+		}
+
+		private static string EscapeJsString(string js)
+		{
+			//get every quote in the string along with all the backslashes preceding it
+			var singleQuotes = Regex.Matches(js, @"(\\*?)'");
+			if (singleQuotes.Count == 0)
+				return js;
+
+			var uniqueMatches = new List<string>();
+
+			for (var i=0; i < singleQuotes.Count; i++)
+			{
+				var matchedString = singleQuotes[i].Value;
+				if (!uniqueMatches.Contains(matchedString))
+				{
+					uniqueMatches.Add(matchedString);
+				}
+			}
+
+			uniqueMatches.Sort((x, y) => y.CompareTo(x));
+
+			//escape all quotes from the script as well as add additional escaping to all quotes that were already escaped
+			for (var i=0; i < uniqueMatches.Count; i++)
+			{
+				var match = uniqueMatches[i];
+				var numberOfBackslashes = match.Length - 1;
+				var slashesToAdd = (numberOfBackslashes * 2) + 1;
+				var replacementStr = "'".PadLeft(slashesToAdd + 1, '\\');
+				js = Regex.Replace(js, @"(?<=[^\\])" + Regex.Escape(match), replacementStr);
+			}
+
+			return js;
 		}
 	}
 }
--- a/Xamarin.Forms.Core/WebView.cs
+++ b/Xamarin.Forms.Core/WebView.cs
@@ -9,8 +9,6 @@ namespace Xamarin.Forms
 	[RenderWith(typeof(_WebViewRenderer))]
 	public class WebView : View, IWebViewController, IElementConfiguration<WebView>
 	{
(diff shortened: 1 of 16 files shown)


alanag13:
Question: should `WebView.Eval` be deprecated in favor of this?

Xamarin.Forms.Core/WebView.cs
```diff
@@ -145,5 +179,39 @@ public void SendNavigating(WebNavigatingEventArgs args)
 		{
 			return _platformConfigurationRegistry.Value.On<T>();
 		}
+
+		private static string EscapeJsString(string js)
```
alanag13: Because we and extra `eval` function in the code I added for standardizing platforms, we need to add an extra level of escaping to any single quotes the user may have passed in for their script. I really don't like this but it was all I could come up with to solve this problem. I'm open to suggestions here.

alanag13:
@rmarinho can you give me any insight into why this failed (it looks like only on Windows)?

hartez:
> can you give me any insight into why this failed (it looks like only on Windows)?

@alanag13 Nothing wrong on your end - the Windows build failure was a VSTS hiccup. That build stage should be green now.

Xamarin.Forms.Core/WebView.cs
```diff
@@ -145,5 +179,39 @@ public void SendNavigating(WebNavigatingEventArgs args)
...
+		{
+			if (!js.Contains("'"))
```
alanag13: Because we add an extra eval function in the code I added for standardizing platforms, we need to add an extra level of escaping to any single quotes the user may have passed in for their script. I really don't like this but it was all I could come up with to solve this problem. I'm open to suggestions here.

Xamarin.Forms.Core/WebView.cs
```diff
@@ -184,7 +178,7 @@ public void SendNavigating(WebNavigatingEventArgs args)
 			return _platformConfigurationRegistry.Value.On<T>();
 		}
 
-		private static string EscapeJsString(string js)
+		static string EscapeJsString(string js)
```
alanag13: Because we add an extra eval function in the code I added for standardizing platforms, we need to add an extra level of escaping to any single quotes the user may have passed in for their script. I really don't like this but it was all I could come up with to solve this problem. I'm open to suggestions here.

alanag13:
GitHub is hiding a comment I made on WebView.cs, since I made a change to it after opening this PR,  so just make sure you show outdated or go to the full changes to see it.

Xamarin.Forms.Core/WebView.cs
```diff
@@ -145,5 +177,39 @@ public void SendNavigating(WebNavigatingEventArgs args)
 		{
 			return _platformConfigurationRegistry.Value.On<T>();
 		}
+
+		static string EscapeJsString(string js)
+		{
+			if (!js.Contains("'"))
```
rmarinho: check if this is null or it will throw NRE

Xamarin.Forms.Platform.GTK/Renderers/WebViewRenderer.cs
```diff
@@ -203,6 +206,15 @@ private void OnEvalRequested(object sender, EvalRequested eventArg)
             }
         }
 
+        Task<string> OnEvaluateJavaScriptRequested(string script)
+        {
+	       if (Control != null)
+	       {
+		       Control.ExecuteScript(script);
```
rmarinho: Control?.ExecuteScript(script);

Xamarin.Forms.Platform.GTK/Renderers/WebViewRenderer.cs
```diff
@@ -203,6 +206,15 @@ private void OnEvalRequested(object sender, EvalRequested eventArg)
...
+	       }
+               return null;
```
rmarinho: Should we always return null?

Xamarin.Forms.Platform.UAP/WebViewRenderer.cs
```diff
@@ -140,6 +143,11 @@ void Load()
 			await Control.Dispatcher.RunAsync(CoreDispatcherPriority.Normal, async () => await Control.InvokeScriptAsync("eval", new[] { eventArg.Script }));
 		}
 
+		async Task<string> OnEvaluateJavaScriptRequested(string script)
+		{
+				return await Control.InvokeScriptAsync("eval", new[] { script });
```
rmarinho: Code style issue, extra tabs I think

Xamarin.Forms.Core/WebView.cs
alanag13: Good catch, will do

Xamarin.Forms.Platform.GTK/Renderers/WebViewRenderer.cs
alanag13: Did this per @jassmith's recommendation in the issue thread: https://github.com/xamarin/Xamarin.Forms/issues/1699

alanag13:
@rmarinho updated with your changes

alanag13:
@rmarinho @samhouts are the ui tests failures unrelated? these were all green yesterday before my most recent commit, which seems like it should be pretty harmless . Should I rebase?

jassmith:
Doesn't impact fast renderers, therefor merging :)

alanag13:
@jassmith do we want to deprecate `Eval` in favor of this? To users, the only difference is that this returns a value. Only reason I can think of to not do this is that GTK and Tizen always return null , but it still seems potentially confusing to have both `Eval` and `EvaluateJavaScriptAsync`.

I can open another PR to do that if you think it's the right call, but let me know the version number you want to appear in the deprecation message.

Xamarin.Forms.Core/WebView.cs
```diff
@@ -77,6 +81,33 @@ public void Eval(string script)
...
 
+		public async Task<string> EvaluateJavaScriptAsync(string script)
+		{
+			EvaluateJavaScriptDelegate handler = EvaluateJavaScriptRequested;
+
+			//make all the platforms mimic Android's implementation, which is by far the most complete.
+			if (Xamarin.Forms.Device.RuntimePlatform != "Android")
+			{
+				script = EscapeJsString(script);
+				script = "try{JSON.stringify(eval('" + script + "'))}catch(e){'null'};";
```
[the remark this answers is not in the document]
jerone: Does the use of `eval` affect [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#Unsafe_eval_expressions)?

PR #2143, opened 2018-03-20
Added the ListView.SelectionMode property

### LINKED ISSUE #1801: [Enhancement] Add SelectionMode property to ListView
### Rationale

Allow customization of the `ListView` selection mode.

### Implementation

Add a new bindable property called `SelectionMode`:

```
public class ListView : ...
{
     public ListViewSelectionMode SelectionMode { get; set; }
}
```

```
public enum ListViewSelectionMode
{
     None,
     Single
}
```

The default value of `SelectionMode` property will be `Single` to reflect today's selection mode which is single selection.

`None` will stop the highlight from appearing when a row is tapped.

When `SelectionMode` property is set to `None`, tapping/clicking on items in the `ListView` will no longer select them; the `ItemSelected` event will not fire, and the value of the `SelectedItem` property will remain null.

If the `SelectedItem` property is set when `SelectionMode` transitions to `None`, `SeletedItem` will be set to null and the `ItemSelected` event will not fire.

When `SelectionMode` is None, the `ItemTapped` event will still fire.

### Backward Compatibility

It will work 100% on all platforms.

### Difficulty: Easy

Android: set `listView.ChoiceMode` property to `ChoiceMode.None` or `ChoiceMode.Single`
iOS: set `tableView.AllowsSelection` property to `false` or `true`
UWP: set `listView.SelectionMode` property to `ListViewSelectionMode.None` or `ListViewSelectionMode.Single`

ntherning:
### Description of Change ###

Added the `ListView.SelectionMode` property and the `ListViewSelectionMode` enum. `ListViewSelectionMode.Single` is the default and retains the previous behavior. When `SelectionMode` is set to `ListViewSelectionMode.None` rows cannot be selected and `ItemSelected` will not be fired. `Tapped` events are still fired and the tapped row will be briefly highlighted during the tap. When  a row has been selected and the property is changed from `Single` to `None` the `SelectedItem` property will be set to `null` and a `ItemSelected` event with a `null` item  will be fired.

**NOTE**: The Xamarin Forms APIs already has an enum named `ListViewSelectionMode` in `Xamarin.Forms.PlatformConfiguration.WindowsSpecific`. This is obviously Windows specific and IIUC it's unrelated to the UWP `ListViewSelectionMode` enum which is a superset of the enum I'm adding in this PR. The existence of two enums with the same name in the APIs could be a source of confusion for users.

### Bugs Fixed ###

Fixes #1801 [Enhancement] Add SelectionMode property to ListView

### API Changes ###

Added:
 - `ListViewSelectionMode ListViewSelectionMode { get; set; } //Bindable Property`
 - `enum ListViewSelectionMode { None, Single }`

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/ListViewRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/ListViewRenderer.cs
@@ -73,6 +73,7 @@ namespace Xamarin.Forms.Platform.UWP
 				UpdateHeader();
 				UpdateFooter();
 				UpdateSelectionMode();
+				UpdateWindowsSpecificSelectionMode();
 				ClearSizeEstimate();
 			}
 		}
@@ -115,9 +116,13 @@ namespace Xamarin.Forms.Platform.UWP
 				ClearSizeEstimate();
 				((CollectionViewSource)List.DataContext).Source = Element.ItemsSource;
 			}
-			else if (e.PropertyName == Specifics.SelectionModeProperty.PropertyName)
+			else if (e.PropertyName == ListView.SelectionModeProperty.PropertyName)
 			{
 				UpdateSelectionMode();
+			}
+			else if (e.PropertyName == Specifics.SelectionModeProperty.PropertyName)
+			{
+				UpdateWindowsSpecificSelectionMode();
 			}
 		}
 
@@ -255,6 +260,20 @@ namespace Xamarin.Forms.Platform.UWP
 		}
 
 		void UpdateSelectionMode()
+		{
+			if (Element.SelectionMode == ListViewSelectionMode.None)
+			{
+				List.SelectionMode = Windows.UI.Xaml.Controls.ListViewSelectionMode.None;
+				List.SelectedIndex = -1;
+				Element.SelectedItem = null;
+			}
+			else if (Element.SelectionMode == ListViewSelectionMode.Single)
+			{
+				List.SelectionMode = Windows.UI.Xaml.Controls.ListViewSelectionMode.Single;
+			}
+		}
+
+		void UpdateWindowsSpecificSelectionMode()
 		{
 			if (Element.OnThisPlatform().GetSelectionMode() == PlatformConfiguration.WindowsSpecific.ListViewSelectionMode.Accessible)
 			{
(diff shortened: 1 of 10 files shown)


Xamarin.Forms.Platform.UAP/ListViewRenderer.cs
```diff
@@ -255,6 +260,20 @@ void UpdateGrouping()
 		}
 
 		void UpdateSelectionMode()
+		{
+			if (Element.SelectionMode == ListViewSelectionMode.None)
+			{
+				List.SelectionMode = Windows.UI.Xaml.Controls.ListViewSelectionMode.None;
+				List.SelectedIndex = -1;
```
rmarinho: This should be on core no? so it can be reused for all platforms, like a internal ResetSelectedItem for something , for example iOS is only setting the SelectedItem to null and not the index

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -654,6 +657,18 @@ void UpdateSeparatorVisibility()
 			}
 		}
 
+		void UpdateSelectionMode()
+		{
+			if (Element.SelectionMode == ListViewSelectionMode.None)
+			{
+				Element.SelectedItem = null;
```
rmarinho: For example here you are not setting the index to -1

Xamarin.Forms.Platform.UAP/ListViewRenderer.cs
ntherning: `List` here is the native UWP `Windows.UI.Xaml.Controls.ListView` so this code doesn't generalize to other platforms.

In the iOS `ListViewRenderer.UpdateSelectionMode()` method we call `UITableView.DeselectRow()` which does the equivalent.

On Android there's no need to do anything as the `Xamarin.Forms.Platform.Android.ListViewAdapter` already listens to the `ListView.ItemSelected` event and when the Android `ListViewRenderer.UpdateSelectionMode()` method sets `SelectedItem` to `null` it will deselect the currently selected row in the UI.

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
ntherning: There's no `SelectedIndex` on iOS. The equivalent is done in the next few rows (lines 665-667). Please see my previous comment as well.

Xamarin.Forms.Platform.UAP/ListViewRenderer.cs
rmarinho: Oh Sorry, i was reading this as ListView Forms.

rmarinho:
@ntherning needs rebase to remove the docs, it's not needed anymore

ntherning:
Rebased and removed docs.

PR #2151, opened 2018-03-21
[Android] AppCompat version of EntryRenderer

### LINKED ISSUE #1549: [Feature Request] Add AppCompat version of EntryRenderer
### Summary
There isn't a AppCompat version of the EntryRenderer, there should be.

### API Changes
Create a AppCompat version of the EntryRenderer, so that the native control is set as a AppCompatEditText.

### Intended Use Case
Having a AppCompat version of controls makes it easier to style it the same for all Android Versions

ravinderjangra:
### Description of Change ###

Ported EntryRenderer
fixes #1549

### Bugs Fixed ###

- No Bug Fix

### API Changes ###

Added:
 - void ListView.Clear ();

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/EntryRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/EntryRenderer.cs
@@ -1,0 +1,283 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Android.Support.V7.Widget;
+using Android.App;
+using Android.Content;
+using Android.OS;
+using Android.Runtime;
+using Android.Views;
+using Android.Widget;
+using Android.Text;
+using Android.Views.InputMethods;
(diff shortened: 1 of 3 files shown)


rmarinho:
Can you add some screenshots so we can check the difference ? @jassmith is this something we need? was requested ?

ravinderjangra:
any update on this?

rmarinho:
@ravinderjangra can you put some screenshots? 

Thanks

adrianknight89:
I think this needs to be rebased since the original renderer has more to copy from. Also, we might need to wrap `AppCompatEditText` in `FormsAppCompatEditText` to add the missing code with respect to selection change, keyboard, and focusability stuff, but I suspect the original `FormsEditText` has something to do with #3647.

samhouts:
Since we have an updated PR (#4241), I will close this one for now. Thank you!

PR #2197, opened 2018-03-27
[Core] Centralizing visual state change checks and providing a method to override if users want to add/control state changes themselves

PureWeen:
### Description of Change ###

Currently state changes apply in a very specific/narrow way when a property changes. This doesn't allow users to override and add their own states and it also doesn't take into account that you need to reassess all the properties before setting the state. For example when a control becomes enabled the state might need to go to focused and not directly to Normal

### API Changes ###

Added:
```C#
public partial class VisualElement
{
protected virtual void ChangeVisualState()
```
### Behavioral Changes ###

The only behavior change I can see is if something goes from disabled to enabled the state might go to Focused where as before it would just go to Normal.

### PR Checklist ###


PureWeen:
build --uitests

PR #2202, opened 2018-03-28
[Enhancement] Basic GIF animation features.

### LINKED ISSUE #1704: [F100] Basic GIF animation support
## Rationale
Animated Gifs are popular. Providing the image view with features that allowed it to animate gifs would allow users to not have to use webviews or create other custom renderers

## Implementation

Add an AnimationPlayBehaviorProperty for the Image control which indicates that a gif should start playing once it has loaded or if it should be started through other mechanisms.

Add two methods for starting and stopping the animation

```csharp

public enum ImagePlayBehavior
{
   None,
   OnLoad
}

public static readonly BindableProperty AnimationPlayBehaviorProperty = BindableProperty.Create(nameof(AnimationPlayBehavior), typeof(ImagePlayBehavior), typeof(Image), ImagePlayBehavior.None);

public void StartAnimation();
public void StopAnimation();


```

### Android
If user has specified they want the animation to play onload or they call StartAnimation then start playing the gif animation

### iOS
See Android

### UWP
See Android

## Backward Compatibility
Ensure the way the current Image control processes gifs does not change unless the user opts in for the given properties

## Difficulty : Medium

Work that's been done
https://forums.xamarin.com/discussion/17448/animated-gif-in-image-view
https://github.com/jamesmontemagno/PuppyKittyOverflow/blob/master/PuppyKittyOverflow.Touch/AnimatedImageView.cs
https://github.com/jamesmontemagno/GifImageView-Xamarin.Android
https://github.com/luberda-molinet/FFImageLoading/pull/470/files

lateralusX:
### Description

Adds basic GIF image animation support on UWP/IOS/Android. **NOTE**, according to issue this implementation intends to target small animated GIF's meaning that current implementation does **NOT** include support for caching or streaming. This could of course be added in future if needed and some preparations have been done to support streaming in implemented GIF decoder.

Implementation on UWP uses OS support for animated GIF's included in 1607 (anniversary update) or later. No support added to previous versions. The minimum UWP version supported by Xamarin Forms (master branch) should be 1607.

Implementation on IOS uses IOS GIF parser and key frame animations to present correctly timed key frames to image control.

Implementation on Android includes GIF parser used to parse, decode and create bitmaps from encoded in GIF sources. Implementation uses AnimationDrawable to present correctly timed animation buffers to image control.

- fixes #1704

### API Changes ###

New animation properties and methods on Image:
```C#
public static readonly BindableProperty IsAnimationAutoPlayProperty = BindableProperty.Create(nameof(IsAnimationAutoPlay), typeof(bool), typeof(Image), false);

internal static readonly BindablePropertyKey IsAnimationPlayingPropertyKey = BindableProperty.CreateReadOnly(nameof(IsAnimationPlaying), typeof(bool), typeof(Image), false);

public static readonly BindableProperty IsAnimationPlayingProperty = IsAnimationPlayingPropertyKey.BindableProperty;

public readonly bool IsAnimationPlaying

public readonly bool IsAnimationAutoPlay

public void StartAnimation()

public void StopAnimation()

public event EventHandler AnimationFinishedPlaying;

```
Setting animation auto play to `true `will start to play animation as soon as the animation has been loaded.

Setting animation auto play behavior to `false` will start to play animation ones start method has been called.

Not setting auto play property will not enable animation support for image initially loaded. If start/stop animation API's are called on an image without auto play property set the image will be reloaded as an animated image (if needed).

### Behavioral Changes ###

Nothing, if the animation auto play property has not been set then images will be loaded normal. Trying to use animation support on none GIF image sources will fail the load and fallback trying to load as regular image (Android) or as a 1 frame keyframe animation (IOS).

### PR Checklist ###


lateralusX:
Will add a couple of smaller adjustments to this PR:

* Look at repeat count support on Android.
* Add events when starting/stopping + finished playing (so it could be restarted).
* Validate behavior when loading none GIF's (IOS/Android) and adjust fallback.
* Add a couple of more test cases in issue1704.cs.

rmarinho:
build --uitests

Xamarin.Forms.Core/Image.cs
```diff
@@ -21,6 +21,17 @@ public class Image : View, IImageController, IElementConfiguration<Image>
 
 		public static readonly BindableProperty IsLoadingProperty = IsLoadingPropertyKey.BindableProperty;
 
+		public enum AnimationPlayBehaviorValue
```
rookiejava: how about  `bool IsAutoPlay` instead? Is the `None`  value meanigful?
Even so, naming should be improved. It's a bit ambigous. And `Value` suffix can be ommited.

StephaneDelcroix: I agree with @rookiejava, that enum is probably misnamed, and might not be required at all

Xamarin.Forms.Core/Image.cs
```diff
@@ -21,6 +21,17 @@ public class Image : View, IImageController, IElementConfiguration<Image>
...
+		{
+			None,
+			OnLoad,
+			OnStart
+		};
+
+		public static readonly BindableProperty AnimationPlayBehaviorProperty = BindableProperty.Create(nameof(AnimationPlayBehavior), typeof(AnimationPlayBehaviorValue), typeof(Image), AnimationPlayBehaviorValue.None);
+
+		public static readonly BindableProperty IsAnimationPlayingProperty = BindableProperty.Create (nameof(IsAnimationPlaying), typeof (bool), typeof (Image), false);
```
StephaneDelcroix: I guess you want a read-only property ? not exposing a setter isn't enough. As someone could still set the value from XAML, from a Binding, or by using `SetValue`.

what you want instead is like this:
```
static readonly BindablePropertyKey IsAnimationPlayingPropertyKey = BindableProperty.CreateReadOnly (nameof(IsAnimationPlaying), typeof (bool), typeof (Image), false);
public static readonly BindableProperty IsAnimationPlayingProperty = IsAnimationPlayingPropertyKey.BindableProperty;
```

you can even expose a private setter

```
public bool IsAnimationPlaying
{
  get { return (bool)GetValue(IsAnimationPlayingProperty); }
  private set => SetValue(IsAnimationPlayingPropertyKey, value);
}
```

Xamarin.Forms.Core/Image.cs
```diff
@@ -45,13 +56,36 @@ public bool IsOpaque
...
+		public AnimationPlayBehaviorValue AnimationPlayBehavior
+		{
+			get { return (AnimationPlayBehaviorValue)GetValue(AnimationPlayBehaviorProperty); }
+			set { SetValue(AnimationPlayBehaviorProperty, value); }
+		}
+
+		public void StartAnimation()
+		{
+			if (AnimationPlayBehavior != AnimationPlayBehaviorValue.None)
+				SetValue(IsAnimationPlayingProperty, true);
```
StephaneDelcroix: will stop working when you move to a ro BP

Xamarin.Forms.Core/Internals/ImageParser.cs
```diff
@@ -0,0 +1,819 @@
+﻿using System;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+using System.Diagnostics;
+
+namespace Xamarin.Forms.Internals
+{
+	public class GIFDecoderFormatException : Exception
```
StephaneDelcroix: do you own that code, or was it borrowed from somewhere else ?

lateralusX: Inspiration comes from an MIT project, but this code is a complete rewrite done by myself. The only pieces that still have some similarities are some of the internal parts of the decode and compose methods.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1704.cs
```diff
@@ -0,0 +1,193 @@
...
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 1704, "[Enhancement] Basic gif animation features", PlatformAffected.UWP)]
+	public class Issue1704 : TabbedPage
+	{
```
PureWeen: Can you create a UI Test that just loads this issue? to at least ensure that nothing crashes?

Xamarin.Forms.Core/Image.cs
lateralusX: Agree, enum came from original issue, have switched to an bool auto play property, as suggested.

Xamarin.Forms.Core/Image.cs
lateralusX: Fixed.

Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
```diff
@@ -26,17 +73,18 @@ internal static class ImageViewExtensions
...
 			Bitmap bitmap = null;
+			AnimationDrawable animation = null;
 			Drawable drawable = null;
+			IImageSourceHandlerEx handler;
+			bool useAnimation = newImage.IsSet(Image.AnimationPlayBehaviorProperty) || newImage.IsSet(Image.IsAnimationPlayingProperty);
 
-			IImageSourceHandler handler;
-
-			if (source != null && (handler = Internals.Registrar.Registered.GetHandlerForObject<IImageSourceHandler>(source)) != null)
+			if (source != null && (handler = Internals.Registrar.Registered.GetHandlerForObject<IImageSourceHandlerEx>(source)) != null)
```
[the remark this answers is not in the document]
lateralusX: Fixed.

Xamarin.Forms.Platform.Android/Renderers/FileImageSourceHandler.cs
```diff
@@ -26,10 +31,34 @@ public sealed class FileImageSourceHandler : IImageSourceHandler
...
+			if (!DecodeSynchronously)
+				await BitmapFactory.DecodeResourceAsync(context.Resources, ResourceManager.GetDrawableByName(file), options);
+			else
+				BitmapFactory.DecodeResource(context.Resources, ResourceManager.GetDrawableByName(file), options);
+
+			var decoder = new AndroidGIFImageParser(context, options.InDensity, options.InTargetDensity);
+			using (var stream = context.Resources.OpenRawResource(ResourceManager.GetDrawableByName(file)))
+			{
+				if (!DecodeSynchronously)
+					await decoder.ParseAsync(stream);
```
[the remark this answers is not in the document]
lateralusX: Fixed.

Xamarin.Forms.Platform.Android/Renderers/ImageLoaderSourceHandler.cs
```diff
@@ -26,5 +27,28 @@ public sealed class ImageLoaderSourceHandler : IImageSourceHandler
...
+		public async Task<AnimationDrawable> LoadImageAnimationAsync(ImageSource imagesource, Context context, CancellationToken cancelationToken = default(CancellationToken))
+		{
+			var imageLoader = imagesource as UriImageSource;
+			AnimationDrawable animation = null;
+			if (imageLoader?.Uri != null)
+			{
+				using (Stream imageStream = await imageLoader.GetStreamAsync(cancelationToken).ConfigureAwait(false))
+				{
+					var decoder = new AndroidGIFImageParser(context, 1, 1);
+					await decoder.ParseAsync(imageStream);
```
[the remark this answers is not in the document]
lateralusX: Fixed.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1704.cs
lateralusX: @PureWeen, I have not done a UI test before (just issue tests) so don't know details yet. Should I add a new class into Xamarin.Forms.Core.UITests and hook up my issue class using similar mechanics as ImageUITest and add something into the ControlGalleryPages similar to ImageCoreGalleryPage.cs or is there some easier way just to reuse my issue class? How do I run the UI tests ones added?

lateralusX:
Added support to load image as a regular image, the auto play property not set. If start is later called, control will try to reload image as an animated image (if supported) and start it. If not supported, it will fallback to regular image again.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1704.cs
PureWeen: Here's a sample for a basic UITEST. You just change the structure a little bit of the test and add an IFDEF for the UI Test.  The Controls.ISsues.Shared is part of the UI test project as well

https://github.com/xamarin/Xamarin.Forms/blob/122f0e3e6060711d6c3b2b8523236b8a057434df/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3292.cs

and then I would also add
```C#
[Category(UITestCategories.ManualReview)]
```

Since it'll require someone to physically look at the gifs to make sure they are playing

> How do I run the UI tests once added?
- Deploy the Control Gallery to whatever device you want to run the UI Test on
- Now compile whatever project in the *UITests* folder you want to run 
- at this point the UI Test you added should show up in the text explorer and you should just be able to run it and it will run the UI Test

lateralusX: @PureWeen Added UITest for Issue1704.

Xamarin.Forms.Platform.UAP/ImageRenderer.cs
```diff
@@ -178,6 +191,19 @@ void UpdateAspect()
 				// might have disposed of this Image already.
 				if (Control != null)
 				{
+					if (imagesource is BitmapImage bitmapImage)
+					{
+						if (_nativeAnimationSupport)
+						{
+							bitmapImage.AutoPlay = false;
+							if (Element.IsSet(Image.IsAnimationAutoPlayProperty) || Element.IsSet(Image.IsAnimationPlayingProperty))
+								bitmapImage.AutoPlay = (bool)Element.GetValue(Image.IsAnimationAutoPlayProperty);
```
PureWeen: *Element.IsAnimationAutoPlay*

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1704.cs
```diff
@@ -0,0 +1,485 @@
...
+			Children.Add(_page3);
+			Children.Add(_page4);
+		}
+
+#if UITEST
+		[Test]
+		[Category(UITestCategories.ManualReview)]
+		public void Issue1704Test()
+		{
+			RunningApp.WaitForElement("On Load");
```
PureWeen: Here can you click on the tabs and buttons just to make sure it doesn't crash?  You can just use the text for the tabs to switch between tabs
```C#
RunningApp.Tap(q => q.Marked("ON START"));
RunningApp.WaitForElement(q => q.Marked("Start Animation"));
RunningApp.Tap(q => q.Marked("Start Animation"));
etc...
```

Xamarin.Forms.Platform.iOS/Renderers/ImageRenderer.cs
```diff
@@ -130,42 +255,79 @@ void SetAspect()
...
+					if (uiimage != null)
+					{
+						imageView.Image = uiimage;
+					}
+					else if (animation != null)
+					{
+						imageView.AutoPlay = (bool)Element.GetValue(Image.IsAnimationAutoPlayProperty);
+						imageView.Animation = animation;
+
+						animation.AnimationStopped += OnAnimationStopped;
```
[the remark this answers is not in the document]
lateralusX: Fixed.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1704.cs
lateralusX: Added.

lateralusX:
@PureWeen Thanks for feedback! I have adjusted PR accordingly.

Regarding failure on Android API level 19, turns out there are a couple of bugs in Androids AnimationDrawable on versions before Lollipop + changed defaults on later Android versions for a couple of members (like OneShot) that cause the issues you identified. Added fixes to make sure "defaults" are consistent with older API versions, also added a workaround fix for incorrect animation speed that occurs on older Android versions.

Regarding API selection on Android. I Looked at a couple of larger image libraries on Android before making the implementation and they all seem to implement GIF animation support on their own (not using larger Android components, like Movie API), so I followed that strategy for this PR as well. This will give us control of optimization, image caching etc. not paying for additional resources not needed or out of our control. I expect this to add up, especially when loading several animated GIF's at the same time. The Movie API is deprecated in next version of Android, replaced by a drawable implementation similar to the AnimationDrawable we use in this PR. I have isolated the loaded drawable behind an interface that should make it simple for us to use that implementation on never Android versions.

As clarified in the initial issue, this PR currently focus on small animated GIF's. It doesn't implement image streaming and image caching, but that something we could implement if needed. I'm happy to do that for Android (where it initial will have biggest impact) as a follow-up on this PR, but first I believe it make sense to merge current work before progressing with further optimizations (if we believe that work will be valuable to our users).

PureWeen:
So this is looking great.

@lateralusX 

One thing I noticed is that the gif animation on ios seems to have an issue rendering. Let me know if you see the same thing. I tested this on a physical ipad and the simulator

- in your unit tests I specified "https://metrouk2.files.wordpress.com/2018/03/giphy9.gif" for the source
- click load image
- the space fills like it should have a gif but it's just white
- if i click to a different tab then back the gif loads and starts playing

lateralusX:
@PureWeen  Great catch! When switching layers on a UIView loaded without an animation didn't trigger a redraw on IOS. After adding an explicit call to setNeedsDisplay, the above gif shows up animating when loaded.

PureWeen:
@lateralusX   I'm marking this with *do not merge* because I've been going back and forth about some of the interfaces on this and need a little time (by next week) to put together some thoughts.

Specifically on Android the *IImageSourceHandler* is problematic because it returns a Bitmap and we've had to put together some not so great ways to utilize drawables better for memory reasons. 

I'm a little concerned about introducing another interface *IImageSourceHandlerEx* and then down the road needing something like *IImageSourceHandlerExEx* (are default interfaces a thing yet?)

It'd be nice just make the code that came before you just a wee bit better :-) Hopefully that makes sense.

lateralusX:
OK, NP, if you plan to extend the interface to better handle the regular images we should be extending to the same new interface to make sure we just end up with one new interface (I guess we can't touch IImageSourceHandler since we have external dependencies on that interface?). The interface that I added, IImageSourceHandlerEx abstracted the animated image using IFormsAnimationDrawable, maybe we could use a similar concept for regular images as well.

PureWeen:
@lateralusX  some more conversations going on over here
https://github.com/xamarin/Xamarin.Forms/pull/3045

About how to make the interfaces a bit more focused. So for example in this PR instead of creating the inherited interface we could just create

*ICanLoadAnimation*

I wouldn't make that change yet (don't want to waste your time) but just wanted to make you aware of some additional conversation going on around making the image interfaces more malleable

PureWeen:
Alright @lateralusX  apologies with the delay on this one but just wanted to post an update and ask a couple questions

### Android
Gifs will probably be delayed a bit more because for Android the intention right now is to pull in Glide and use that for gifs
https://github.com/xamarin/Xamarin.Forms/issues/3577

### UWP
Works great :-) I wish it was that few lines of code on the other two platforms

### iOS
Why do this one with AddAnimation and the CALayer? opposed to something like this
https://medium.com/@ldbchan/animate-serial-images-in-uiimageview-using-swift-39f249b25453
https://github.com/luberda-molinet/FFImageLoading/pull/470/files

or setting the AnimatedImages on the UIImageView?

lateralusX:
@PureWeen The IOS implementation was following pointers from original issue #1704. There are also some characteristics around animated GIF's (like different duration time for each frame) that can't be solved using AnimatedImages that needs the additional control available in the CALayer animation support.

For Android, well if lean towards include Glide in Xamarin Forms it make sense to use that library for Gifs, also note that Android API version 28 includes AnimatedImageDrawable with build in support for animated GIF's. A little unfortunate that we didn't had Glide on the map when I initially started this work (just pulling in Glide for this feature would probably been to much overhead and work anyways) but would have saved me a ton of headache… So then the question is how we best proceed with this PR, wait until we have Glide and adapt using that library on Android or exclude Android support from this PR (making the animation support void on Android for now) and merge IOS/UWP first?

BTW, looks like the image interface changes discussed above has been done as well but is it only affecting Android or does this PR need to adapt to changes on UWP/IOS as well?

PureWeen:
@lateralusX  Yea :-/  I looked at a few implementations on iOS and they either just ignore those frame times or do similar to what you have.  https://github.com/Flipboard/FLAnimatedImage uses a CADisplayLink timer to sync to the refresh rate and swaps images out as time moves along... Fun stuff

I've seen some murmuring that AnimatedImageDrawable will make its way to the Support Libraries which would be neat

At this point (I know this is annoying) but just a holding pattern on this one until finishing this gets scheduled into a sprint.

As far as the interfaces. I would say that the way Android does the interfaces represents a pattern to use where we don't have to effect the existing interface by creating IImageSourceHandlerEx

So the idea would be to instead create an interface for animations
IAnimationSourceHandler that contains the logic for LoadImageAnimationAsync

We should probably also condense the IsPlaying/IsPaused bools into an enum like the original ticket.

All that being said. I wouldn't recommend you spend time on this currently as I don't want to waste your time. Once gifs make it back into a sprint then we'll get all the great stuff you've done here finalized and integrated

jsuarezruiz:
@lateralusX Could you rebase?. Thanks.

PureWeen:
Working on the rebase for this now

PureWeen:
Closing this PR in favor of this one https://github.com/xamarin/Xamarin.Forms/pull/7330

Had to heavily rework this to fit with a lot of our new image loading code structures
The core behavior of how this PR plays gifs has been maintained though

PR #2215, opened 2018-03-30
[Android] Second measurement pass in IPlatform.GetNativeSize 

### LINKED ISSUE #1396: Label HorizontalTextAlignment (Center or End) is not kept when navigating back to a page
### Description

The horizontal text alignment of a label is lost on Android when navigating to another page and coming back. It's not updated until the text is updated (via binding). It's the same erroneous behavior whether the Label is in a Grid or a StackLayout. It works as expected on iOS and I haven't tested other platforms.

### Steps to Reproduce

1. Have a label with HorizontalTextAlignment = Center or End
2. Navigate to another page
3. Come back

### Expected Behavior

The text is aligned center or right as configured.

### Actual Behavior

The text is aligned left until the text is updated (via binding).

### Basic Information

- Version with issue: 2.5.0.121934
- IDE: Rider
- Platform Target Frameworks: 
  - Android: 7.1.1 
- Android Support Library Version: 26.1.01
- Affected Devices: Android simulator

### Related

It seems to be the same bug as described here : https://stackoverflow.com/questions/43147929/xamarin-label-losing-text-alignment-when-page-returned-to. Sadly, I can't use the proposed workaround which is to set HorizontalOptions=EndAndExpand or End because the Label flicker when its size changes because of text changes.

### LINKED ISSUE #1415: [Android] HorizontalTextAlignment="Center" loses alignment when scrolled
Moved from https://bugzilla.xamarin.com/show_bug.cgi?id=49311

### Description
I've confirmed this issue. I also confirmed the Label is sized full width, and this issue is still present with Fast Renderers.

From original Bugzilla description:

It's a bit odd, but sometimes the alignment of a Label inside a StackLayout inside a Frame (inside a ViewCell) is messed up.

If you scroll, then sometimes the Label is not centered and sometimes it's centered.

### Expected Behavior
Text should be always centered on the screen

### Actual Behavior
Text occasionally appears left aligned when scrolled.

### Basic Information
- Version with issue: 2.3.3.175, 2.5.0.121934
- Last known good version: ?
- IDE:
- Platform Target Frameworks: 
  - Android: 8.0
- Android Support Library Version: 25.4.0.2

### Screenshots

https://github.com/lhaussknecht/CenterBug/blob/master/assets/centerBug.gif

https://www.dropbox.com/s/z4hzr7rl9o2lvqb/Screenshot%202017-12-18%2009.04.09.png?dl=0

https://www.dropbox.com/s/evk3zhdu0ivfi5p/Screenshot%202017-12-18%2009.23.53.png?dl=0

### Reproduction Link

https://github.com/lhaussknecht/CenterBug

hartez:
### Description of Change ###

Add a `MeasureSpecMode.Exactly` measurement pass in `IPlatform.GetNativeSize` on Android in cases where the layout size is known but the measured size is less than the constrained area.

### Bugs Fixed ###

- fixes #1415 HorizontalTextAlignment="Center" loses alignment when scrolled
- fixes #1396 Label HorizontalTextAlignment (Center or End) is not kept when navigating back to a page

### API Changes ###

Added:
 - public interface  Xamarin.Forms.Platform.Android.IViewRenderer
    * void MeasureExactly();

### Behavioral Changes ###

None

### PR Checklist ###


hartez:
Adding a DNM tag, need to apply this to the Fast Renderers before it can be merged.

PureWeen:
Does IViewRenderer also need to be applied to the ScrollViewRenderer?

Why don't these calls just use Exactly when the Width/Height is constrained?

```C#
SizeRequest IPlatform.GetNativeSize(VisualElement view, double widthConstraint, double heightConstraint)
...

int widthMeasureSpec = widthConstrained
	? MeasureSpecFactory.MakeMeasureSpec((int)widthConstraint, MeasureSpecMode.AtMost)
	: MeasureSpecFactory.MakeMeasureSpec(0, MeasureSpecMode.Unspecified);

int heightMeasureSpec = heightConstrained
	? MeasureSpecFactory.MakeMeasureSpec((int)heightConstraint, MeasureSpecMode.AtMost)
	: MeasureSpecFactory.MakeMeasureSpec(0, MeasureSpecMode.Unspecified);
```

The call in VisualElementTracker that does the actual Layout uses Exactly

The result from using AtMost gets returned which doesn't seem to represent the actual bounds and if I follow up the call chain the AtMost results get wiped out anyway by *GetSizeRequest* which just sets the bounds to the HeighRequest/WidthRequest anyway

hartez:
> Does IViewRenderer also need to be applied to the ScrollViewRenderer?

I don't believe ScrollViewRenderer's measurement goes through GetNativeSize, so the issue this fixes doesn't apply to it. 

> Why don't these calls just use Exactly when the Width/Height is constrained?
Because they are being used to determine the smallest possible native size of the controls for layout. 

Specifically for TextView, if you pass a size larger than the smallest possible bounds for a TextView, AtMost will return the smallest bounds. Exactly will just return the size you passed in. So to determine the minimum amount of space needed for a layout, we have to use AtMost. 

> The result from using AtMost gets returned which doesn't seem to represent the actual bounds and if I follow up the call chain the AtMost results get wiped out anyway by GetSizeRequest which just sets the bounds to the HeighRequest/WidthRequest anyway.

Right, it's the minimum bounds. And assuming that the minimum bounds will fit in the HeightRequest/WidthRequest, we can just go with the HR/WR. The problem we run into is that the last measurement taken determines what Draw actually uses. So if we have a HR/WR of, say, 100 x 200, and we're rendering some short text which actually only needs 100 x 50, then it will try to draw at 100 x 50. We have to perform another measure at Exactly 100 x 200 to ensure that it draws in a space of 100 x 200 (and can therefore center itself horizontally/vertically).

PureWeen:
I guess what I'm not seeing is where the AtMost call causes different behavior to occur then if you use Exact
 
Currently I see the following execution path as I debug through
- Change Text
- VisualElementTracker calls updatelayout and uses Exact for the measurement spec
- Layout is now called on the LabelRenderer which uses the control as it's measured with the Exact measurement specs
- the queue'd UpdateLayout call on Layout is now called which causes the SL to update its children Layouts
- When it gets to the changed label *GetNativeSize* uses AtMost to determine the result that's returned (at this point it does a measure call with AtMost against the label which is what causes the Measure on the Label to break and draw incorrectly from resuming the app)
- Then the MeasureExactly is called (which fixes the issue caused by the previous AtMost call) using the Bounds of the Label that were determined using Exactly for the MeasureSpec. This effectively just resets the Measured values back to what the TextView was before AtMost was used for the MeasureCall
- A SizeRequest that's based off of AtMost gets returned from *GetNativeSize* to *VisualElement.GetSizeRequest* 
- GetSizeRequest doesn't even use this calculation. It just checks the values against the HR/WR and sets them to those values. So if that GetNativeSize request a few steps back just used Exact in the first place the end result seems the same

As it stands there's an AtMost calculated SizeRequest being returned that doesn't correlate to the reality of the Measure on the TextView

If I change the Platform GetNative call to 
```c#
int widthMeasureSpec = widthConstrained
	? MeasureSpecFactory.MakeMeasureSpec((int)widthConstraint, MeasureSpecMode.Exact)
	: MeasureSpecFactory.MakeMeasureSpec(0, MeasureSpecMode.Unspecified);

int heightMeasureSpec = heightConstrained
	? MeasureSpecFactory.MakeMeasureSpec((int)heightConstraint, MeasureSpecMode.Exact)
	: MeasureSpecFactory.MakeMeasureSpec(0, MeasureSpecMode.Unspecified);
```

And comment out the MeasureExact() calls  then all those ui tests still pass.  


> The problem we run into is that the last measurement taken determines what Draw actually uses. So if we have a HR/WR of, say, 100 x 200, and we're rendering some short text which actually only needs 100 x 50, then it will try to draw at 100 x 50.  We have to perform another measure at Exactly 100 x 200 to ensure that it draws in a space of 100 x 200 (and can therefore center itself horizontally/vertically).

Right, so just use Exact in the first place instead of AtMost since the code is going to be re-measuring to Exact and forcing the SizeRequest to match a call to Exact anyway.
 
I'm just still not seeing where the downstream behavior changes if you do those first calls with Exact instead of AtMost since everything downstream just forces the values to match as if the code would have just called Exact in the first place.

hartez:
> If I change the Platform GetNative call to ... And comment out the MeasureExact() calls then all those ui tests still pass.

Using `Exactly` on those first measure calls breaks a lot of other things unrelated to the new tests introduced here. 

`GetNativeSize` is being used to determine what size things need to be on Android so that the Forms layout engine can use those sizes to perform its own layout. It uses `AtMost` because `AtMost` allows the Android View being measured to return [whatever size suits it, up to the specified size](https://developer.android.com/reference/android/view/View.MeasureSpec). What that means for a TextView is "the smallest size I can fit this text into, given the content, typeface, etc.". So the value returned is very useful to the Forms layout engine; that's how the Forms engine knows how much it can compress a Label and still display the text. If we use `Exactly` here, the measure effectively does nothing for Forms; the Android View will just be the size we passed in, and the Forms engine won't get any information.

We do eventually have to measure using `Exactly` for the Android Views, though - if we don't, when `Draw` is called they will only be drawn into that smallest possible area that `AtMost` gave us. For TextViews in particular this is a problem because they'll be drawn to exactly the width of the text rather than the requested width, which kills all the whitespace and makes it seem like the text alignment values aren't being applied. Normally this isn't a problem because one of the `OnLayout` overrides happens after the Forms engine has take its measurements, and `OnLayout` measures using `Exactly`. 

Unfortunately, the two cases referenced here (resuming the application, and some ListView circumstances) are situations where `OnLayout` doesn't get called (because it's not technically necessary; in both cases the layout has already been computed). Since `OnLayout` isn't called, the last measurement of the Android Views was an `AtMost` (the query from the Forms engine), and when `Draw` kicks in it uses those `AtMost` values. 

So this extra `Exactly` measurement call is simply ensuring that if the Forms engine's query was the last thing to measure an Android View, the final measurement was done with an `Exactly`. It's being called 100% for its side effects on `Draw`. 

It's certainly possible that there's a more elegant solution to this, or that the whole problem could be avoided with changes to the interaction between Forms and Android's layout engines. But this solution prevents all instances of this particular _type_ of problem with only a minor performance hit.

PureWeen:
Should this code use the new *MeasureExactly* call from the IViewRenderer interface ?
https://github.com/xamarin/Xamarin.Forms/blob/c18e85ee434d7bacb730215fecf15acc0230917f/Xamarin.Forms.Platform.Android/VisualElementTracker.cs#L92

Since that's the code that sets the measure for the draw that's used? Just to ensure that it is the same call both places?


> with changes to the interaction between Forms and Android's layout engines

The main part here I'm curious about is the order of operations as it pertains to how something like a StackLayout sets the bounds and when the element is drawn on the screen. When the text changes the set of operations that I see are
- TextChanges
- VisualElementTracker calls MeasureExactly/OnLayout
- Android element is drawn on the screen
- Now all Layout calls on the SL resolve. So at this point if SL changes the layout bounds or makes any measure calls then none of this is propagated out to Android (as we see with the bug)

It seems like that second thing *VisualElementTracker calls MeasureExactly/Layout*  should be occur at the end not directly after the TextChanges.  

When the view is first loaded the order of operations goes
- TextChanges
- Now all Layout calls on SL resolve. 
- VisualElementTracker calls MeasureExactly/Layout
- Android element is drawn on the screen

Not saying this is a solution but if you remove *Device.BeginInvokeOnMainThread(() =>* from Layout.cs then this bug doesn't happen and you don't need to add that MeasureExactly() code because the call inside VisualElementTracker resolves last. It seems like the Forms layout system is staying one step behind what gets drawn. 

>  Normally this isn't a problem because one of the OnLayout overrides happens after the Forms engine has take its measurements, and OnLayout measures using Exactly

The violation of this expectation seems to be what's at the root cause right? The fix does indeed  resolve the side effect but I just feel like there's still an edge case out their since the Forms engine is still taking its own measurements post OnLayout and any measurement variations caused by that won't be reflected out since OnLayout was already called and won't be called again.

hartez:
> Now all Layout calls on the SL resolve. So at this point if SL changes the layout bounds or makes any measure calls then none of this is propagated out to Android (as we see with the bug)

If the SL changes the layout bounds, that _will_ trigger another OnLayout and the changes _will_ be propagated to Android. The bugs happen when the SL doesn't _need_ to change the layout bounds. It measures, determines that no changes are needed (the content still fits in the same layout it had before), so OnLayout doesn't get called. The problem is that Android is still holding onto those last `AtMost` measurements, and an Android redraw which is triggered by something that doesn't affect layout (like resuming) will use the `AtMost` values.

PureWeen:
> If the SL changes the layout bounds, that will trigger another OnLayout and the changes will be propagated to Android. 

Alright I see where/how that happens now. I wasn't finding it in the code the other day


> The problem is that Android is still holding onto those last AtMost measurements, and an Android redraw which is triggered by something that doesn't affect layout (like resuming) will use the AtMost values.

This part I get. Regardless of anything that call to GetNativeSize needs to "reset" the measures on the Android element itself . Fixing that fundamental side effect of verifying the AtMost I get. Main thing I'm commenting about at this point are things just involved with the path the layout engine takes.

That being said just two more questions. First question pertains to this PR and second question is more of a general thought that maybe down the road could mean something

1) Should this code use the new *MeasureExactly* call from the IViewRenderer interface ?
https://github.com/xamarin/Xamarin.Forms/blob/c18e85ee434d7bacb730215fecf15acc0230917f/Xamarin.Forms.Platform.Android/VisualElementTracker.cs#L92

Since that's the code that sets the measure for the draw that's used? Just to ensure that it is the same call both places?

2)  
Is there merit in looking at not calling *MeasureExactly* from the VisualElementTracker when text is changed?

The path taken when the text is changed is
- TextChanges
- VisualElementTracker calls MeasureExactly/OnLayout
- Now SL does all its remeasure stuff and if the region has changed then it does a layout

But it seems like that second set of operations could be reduced when the text is changed since the SL will do its checks to see if a Layout is even necessary

hartez:
> Should this code use the new MeasureExactly call from the IViewRenderer interface ?

I don't _think_ so, because MeasureExactly is doing to some checks to see if it's really necessary which are specific to the situation it addresses.

> Is there merit in looking at not calling MeasureExactly from the VisualElementTracker when text is changed?

Maybe? If it saves us a ton of cycles, then yes. But I'm betting that Exactly measure calls don't cost us that much; they don't really have to do a lot. I don't believe there's a ton of computation involved, it's just setting the measured height and width to the values being passed in. 

I could be wrong about that, though - it's something we'd have to profile to know for sure.

PureWeen:
>  is doing to some checks
hmm yea I could see that causing some issues during the initial load phase..

PureWeen:
@hartez  should this be rebased to 3.1 ?

PR #2221, opened 2018-04-01
[Enhancement] Allow underline and strikethrough text decorations on labels and spans

### LINKED ISSUE #1632: [Enhancement] Label underline/strikethrough support
## Rationale

The label currently has no easy mechanism of being underlined. Sometimes applications need the ability to underline text without using an Effect.

## Implementation

Modify the FontAttributes enum to be as follows:

```csharp
public enum FontAttributes
{
	None = 0,
	Bold = 1 << 0,
	Italic = 1 << 1,
	Underline = 1 << 2,
        Strike = 1 << 3
}
```

Renderers and Span parsers will need to understand how to use the new underline and strikethrough flags appropriately.

## Expected Result

On all the platforms the label/span has underline and/or strikethrough applied as expected.

## Implications for CSS

CSS will need to be updated to support the `text-decoration: underline` and `text-decoration:line-through` tags for full compatibility.

## Backward Compatibility

This change is as backwards compatible with existing code as possible. It is possible someone has found some obscure way of depending on there not being any more flags to the FontAttributes enum, but that is unlikely.

Third party renderers will need to be updated to ensure that this functionality is supported through the new official mechanism. Further we will need to be careful to code the changes to the renderers in a careful manner to ensure that if someone is already using an effect to set an underline state on a label that this effect is as best as possible not broken by our changes.

## Difficulty : Easy

This is an easy change and great for a first time contributor. The hardest part will be ensuring the Span can properly parse the new `FontAttributes.Underline` enum values. Some basic defensive coding will be needed to ensure existing code is not broken, ideally the code path when the flags are not set AND have never been set is identical to when they have been. Once a user changes their code to use the flag, the code path can then change for everything.

Even more ideally this would be done without an additional flag for tracking that status; possible using [GetIsDefault](https://github.com/xamarin/Xamarin.Forms/issues/1652).

alanag13:
### Description of Change ###

Fixes #1632.

Adds the ability to apply underline and strikethrough text decorations to Labels and Spans via a new BP, `TextDecorations`. Includes support for the CSS `text-decoration` property.

Note that this implementation is a bit different than what is laid out in #1632, I spoke with @jassmith on this a little bit ago and we agreed on this new approach but I'm happy to answer any questions you might have.

Platform support:
- [X] Android
- [X] iOS
- [X] macOS
- [X] UWP
- [X] WPF
- [X] Tizen
- [ ] GTK

### API Changes ###

enum TextDecorations `None, Underline, Strikethrough`
interface IDecorableText
TextDecorations IDecorableText.TextDecorations //Bindable Property

Label and Span now implement IDecorableText.

### PR Checklist ###


Xamarin.Forms.Platform.UAP/LabelRenderer.cs
```diff
@@ -152,6 +158,26 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
...
+				Control.TextDecorations &= ~Windows.UI.Text.TextDecorations.Underline;
+			else
+				Control.TextDecorations |= Windows.UI.Text.TextDecorations.Underline;
+
+			if ((elementTextDecorations & TextDecorations.Strikethrough) == 0)
+				Control.TextDecorations &= ~Windows.UI.Text.TextDecorations.Strikethrough;
+			else
+				Control.TextDecorations |= Windows.UI.Text.TextDecorations.Strikethrough;
+
+			Control.Text = Control.Text; //TextDecorations are not updated in the UI until the text changes
```
alanag13: This was an oddity I observed on UWP -- it seems that changes to `TextDecorations` do not propagate to the UI until the next time the text changes, as if `TextDecorations` only applies to each *new* string assigned to `Text`.

Xamarin.Forms.Core/DecorableTextElement.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Xamarin.Forms
+{
+	[Flags]
+	[TypeConverter(typeof(TextDecorationConverter))]
+	public enum TextDecorations
+	{
+		None = 0,
+		Strikethrough = 1 << 1,
+		Underline = 1 << 0
```
rookiejava: is this intended order?

Xamarin.Forms.Platform.Android/FastRenderers/LabelRenderer.cs
```diff
@@ -263,6 +263,24 @@ void UpdateFont()
...
 		}
 
+		void UpdateTextDecorations()
+		{
+			if (!Element.IsSet(Label.TextDecorationsProperty))
+				return;
+
+			var textDecorations = Element.TextDecorations;
+
+			if ((textDecorations & TextDecorations.Strikethrough) == 0)
```
rookiejava: how do you feel about using `HasFlag()` instead? (here and there)

Xamarin.Forms.Platform.WPF/Renderers/LabelRenderer.cs
```diff
@@ -38,6 +39,8 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
 
 			if (e.PropertyName == Label.TextProperty.PropertyName || e.PropertyName == Label.FormattedTextProperty.PropertyName)
 				UpdateText();
+			if (e.PropertyName == Label.TextDecorationsProperty.PropertyName)
```
rookiejava: else if

Xamarin.Forms.Core/DecorableTextElement.cs
alanag13: I usually alphabetize them, and the values are to match their values on UWP, but I can shift them if desired.

Xamarin.Forms.Platform.Android/FastRenderers/LabelRenderer.cs
alanag13: It seems like the standard for this project is to avoid using `HasFlag` for performance reasons (although I do see it used in some places) and use bitwise ops instead, so that's why I went that route.

Xamarin.Forms.Platform.WPF/Renderers/LabelRenderer.cs
alanag13: good catch, I'll change this.

Xamarin.Forms.Platform.iOS/Renderers/FormattedStringExtensions.cs
```diff
@@ -61,35 +61,54 @@ internal static NSAttributedString ToAttributed(this Span span, Element owner, C
 			if (text == null)
 				return null;
 
+			var range = new NSRange(0, text.Length);
```
jerone: Unused?

alanag13: Guilty as charged for copying from my iOS LabelRenderer work, will fix this too, thanks.

alanag13:
@hartez it looks like most of the docs got deleted from master, not sure what the reason for that is but let me know if there's anything I should do here

hartez:
@alanag13 We've moved the docs stuff out of the repo entirely. So you can just resolve those conflicts by deleting the conflicting docs folder (`docs/Xamarin.Forms.Core`) in your PR.

alanag13:
@hartez thanks for the clarification, done

Xamarin.Forms.Core/DecorableTextElement.cs
rookiejava: I don't think alphabetical order is required for enum.  There is no such requirement and it is difficult to find such an example. Just like UWP's TextDecorations, it would be better to shift `Strikethrough` and `Underline`.

PureWeen: I'd say shift them. Mainly because all the other [FLAGS]  are in numerical order

Xamarin.Forms.Controls/GalleryPages/LabelGallery.cs
```diff
@@ -29,6 +29,10 @@ public LabelGallery ()
 			var italicfont = new Label { Text = "Custom Italic Font" };
 			var boldfont = new Label { Text = "Custom Bold Font" };
 			var bolditalicfont = new Label { Text = "Custom Bold Italic Font" };
+			var toggleUnderline = new Label { Text = "Tap to toggle Underline", TextDecorations = TextDecorations.Underline };
+			toggleUnderline.GestureRecognizers.Add(new TapGestureRecognizer { Command = new Command(()=> { toggleUnderline.TextDecorations ^= TextDecorations.Underline; }) });
+			var toggleStrike = new Label { Text = "Tap to toggle StrikeThrough", TextDecorations = TextDecorations.Strikethrough };
+			toggleStrike.GestureRecognizers.Add(new TapGestureRecognizer { Command = new Command(() => { toggleStrike.TextDecorations ^= TextDecorations.Strikethrough; }) });
```
PureWeen: Can you add one more here that has both of them set?

Xamarin.Forms.Platform.Android/Renderers/FormattedStringExtensions.cs
```diff
@@ -57,6 +57,9 @@ public static SpannableString ToAttributed(this FormattedString formattedString,
 #pragma warning restore 618
 				else if (defaultFont != Font.Default)
 					spannable.SetSpan(new FontSpan(defaultFont, view), start, end, SpanTypes.InclusiveInclusive);
+
```
PureWeen: Delete extra line

Xamarin.Forms.Platform.Android/FastRenderers/LabelRenderer.cs
PureWeen: > avoid using HasFlag for performance reasons

Xamarin.Forms.Core/IDecorableText.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Xamarin.Forms
+{
+    public interface IDecorableText
```
PureWeen: Spaces => tabs

Xamarin.Forms.Platform.WPF/Extensions/FormattedStringExtensions.cs
```diff
@@ -26,7 +26,17 @@ public static Run ToRun(this Span span)
 #pragma warning disable 618
 				run.ApplyFont(span.Font);
 #pragma warning restore 618
+			if (! span.IsSet(Span.TextDecorationsProperty))
```
PureWeen: remove space

Xamarin.Forms.Core/DecorableTextElement.cs
```diff
@@ -0,0 +1,35 @@
...
+	}
+	static class DecorableTextElement
+	{
+		public static readonly BindableProperty TextDecorationsProperty = BindableProperty.Create("TextDecorations", typeof(TextDecorations), typeof(IDecorableText), TextDecorations.None);
+	}
+
+	[Xaml.TypeConversion(typeof(TextDecorations))]
+	public class TextDecorationConverter : TypeConverter
```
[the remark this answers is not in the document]
alanag13: Font is a css property that takes multiple arguments such as a list of fall back typefaces, size, etc and comma separated values are actually part of its propery syntax. So this all makes sense for Font. You cannot specify multiple text decorations using css,  though there are some hacks using border you can use to achieve this: https://stackoverflow.com/questions/45932420/css-underline-strikethrough-and-overline-with-styles-and-colors-in-one-elem. Devs familiar with css would not expect this property to accept a comma separated list.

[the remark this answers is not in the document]
alanag13: This does mean that it would be impossible to set both underline and strikethrough using only css, but I think thats a pretty niche use case.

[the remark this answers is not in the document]
PureWeen: It seems like CSS just works when doing it

https://jsfiddle.net/L79dsh30/
https://stackoverflow.com/questions/9882443/css-multiple-text-decoration/9882469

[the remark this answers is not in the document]
alanag13: Interesting, I was unaware of that -- goes to show how "helpful" SO can be sometimes.
I'll get this added along with your other suggestions, I should have some time tonight.

rmarinho:
@alanag13 this also needs rebase to pick new certs to run on our CI. thanks

alanag13:
@rmarinho @PureWeen sorry for the delay on this, I'm out of the country until 5/15 but shouldn't take much for me to get this all corrected once I get back.

Xamarin.Forms.Core/DecorableTextElement.cs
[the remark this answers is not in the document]
StephaneDelcroix: most of our TypeConverters accept both CSS and Xaml syntax, and we're not making a big deal if you use one for the other. In both cases, using space or comma as separator, we should be able to set multiple ones at once

Xamarin.Forms.Core/IDecorableText.cs
StephaneDelcroix: use `IDecorableTextElement` for consistency

Xamarin.Forms.Platform.Android/FastRenderers/LabelRenderer.cs
StephaneDelcroix: you're right, we try to avoid HasFlags

also, we prefer handling the `true` case first, like this

```
if ((textDecorations & TextDecorations.Strikethrough) == TextDecoration.StrikeThrough)
```

Xamarin.Forms.Core/DecorableTextElement.cs
myroot: If you want alphabetical order, how about exchange values?
``` C#
 None = 0,
 Strikethrough = 1 << 0,
 Underline = 1 << 1
```

Xamarin.Forms.Platform.Tizen/Renderers/LabelRenderer.cs
```diff
@@ -50,19 +51,31 @@ Native.FormattedString ConvertFormattedText(FormattedString formattedString)
...
 				nativeString.Spans.Add(nativeSpan);
 			}
 
 			return nativeString;
 		}
 
+		void UpdateTextDecorations()
+		{
+			var textDecorations = Element.TextDecorations;
+			Control.Strikethrough = (textDecorations & TextDecorations.Strikethrough) != 0;
```
myroot: Please, wrap it with Control.BatchBegin() and Control.BatchCommit() for performance like the UpdateFontProperties()
and UpdateFormattedText() no need to call

alanag13:
@PureWeen @StephaneDelcroix integrated your suggestions and rebased/retargeted on 3.1.0.

Xamarin.Forms.Core/DecorableTextElement.cs
```diff
@@ -0,0 +1,49 @@
...
+	[TypeConverter(typeof(TextDecorationConverter))]
+	public enum TextDecorations
+	{
+		None = 0,
+		Underline = 1 << 0,
+		Strikethrough = 1 << 1,
+	}
+	static class DecorableTextElement
+	{
+		public static readonly BindableProperty TextDecorationsProperty = BindableProperty.Create("TextDecorations", typeof(TextDecorations), typeof(IDecorableTextElement), TextDecorations.None);
```
StephaneDelcroix: use `name(IDecorableTextElement.TextDecorations)`

Xamarin.Forms.Core/DecorableTextElement.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Xamarin.Forms
+{
+	[Flags]
+	[TypeConverter(typeof(TextDecorationConverter))]
+	public enum TextDecorations
+	{
+		None = 0,
+		Underline = 1 << 0,
+		Strikethrough = 1 << 1,
```
StephaneDelcroix: shouldn't it be `StrikeThrough` ?

alanag13: I actually originally had `StrikeThrough` but this was changed after my initial conversation with @jassmith about this PR -- the corresponding APIs on iOS and UWP refer to this as `Strikethrough`, only Android uses `StrikeThrough`

StephaneDelcroix: whatever, the TypeConverter isn't case sensitive. 👍

StephaneDelcroix:
@alanag13 I re-targeted to master. it's too late in 3.1.0 cycle to add new features

Xamarin.Forms.Core/DecorableTextElement.cs
alanag13: will change this, thanks

alanag13:
@StephaneDelcroix @PureWeen rebased onto master now.

alanag13:
@StephaneDelcroix @PureWeen just checking in on this, I think everything should be ready here but let me know if I missed something.

alanag13:
@PureWeen thanks. also what ui checks are failing for ios 11/10?

Xamarin.Forms.Core/DecorableTextElement.cs
```diff
@@ -0,0 +1,49 @@
...
+			var valueArr = value.Split(',');
+
+			if (valueArr.Length <= 1)
+				valueArr = value.Split(' ');
+
+			foreach (var item in valueArr)
+			{
+				if (Enum.TryParse(item.Trim(), true, out TextDecorations textDecorations))
+					result |= textDecorations;
+				if (item.Equals("line-through", StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
alanag13: good catch on the else if, that was a typo and we shouldnt have to compare to line-through if there was an enum match. will add tests too.

alanag13:
@PureWeen also do you have an issue # or stack trace or anything for that UWP crash?

PureWeen:
> do you have an issue # or stack trace or anything for that UWP crash?

I didn't create an issue # yet... I was hoping to recreate the issue on master outside the scope of this PR but wasn't able to by trying other css things. 

I have a couple other thoughts though of how to recreate on master because I'm guessing the setting of those properties causes the exception so this PR just helped surface it


image

PureWeen:
> thanks. also what ui checks are failing for ios 11/10?
AFAICT these aren't you.

Xamarin.Forms.Platform.UAP/LabelRenderer.cs
```diff
@@ -128,6 +131,7 @@ protected override void OnElementChanged(ElementChangedEventArgs<Label> e)
 				_isInitiallyDefault = Element.IsDefault();
 
 				UpdateText(Control);
+				UpdateTextDecorations();
```
PureWeen: I know this isn't consistent between platforms :-/ but can you change this so Control is passed in here? Since that's at least how all the other updates are here on UWP

PureWeen:
Alright I created this 
https://github.com/xamarin/Xamarin.Forms/issues/2901
Isolated from this PR

After the core tests are added we can probably still just merge this one as the Text Decorations all seem to be properly set so it doesn't seem like that issue should act as a blocker for just getting this merged in 

Though if we don't fix those Span bugs the jubilation of Text Decorations once released will be short :-/

Xamarin.Forms.Platform.UAP/LabelRenderer.cs
PureWeen: So I think this is the code that is causing the exception on UWP.  If the Control TEXT is set via "spans" then this line of code wipes out the inline runs on the control which then causes the exception to happen in *RecalculateSpanPositions* so this code here is going to need to check if text is set via spans and if so then iterate over those instead of setting the Text to itsefl

PureWeen: So I think this is the code that is causing the exception on UWP.  If the Control TEXT is set via "spans" then this line of code wipes out the inline runs on the control which then causes the exception to happen in *RecalculateSpanPositions* so this code here is going to need to check if text is set via spans and if so then iterate over those instead of setting the Text to itsefl

PureWeen:
@alanag13 do you have time to fix the last UWP crash that's happening and rebase? If not let me know and I can make the last set of changes so we can get this one merged in

alanag13:
@PureWeen sorry that I've been absent from this lately. I should be able to handle making those final changes sometime this week.

alanag13:
@PureWeen it looks like your assumption re: the UWP bug was correct. This is resolved now and I added the converter test / other small cleanup points, so this should hopefully be finally good to go 😄

PureWeen:
build --uitests

PureWeen:
Yay!! Once we get through the uitests we'll merge

alanag13:
@PureWeen i can't see the failures but let me know if there's anything else you need from me.

PureWeen:
Yea those don't look like they are related to this

PureWeen:
build --uitests

PureWeen:
build --uitests

PureWeen:
@alanag13  YAY!!!!

Have a good weekend

Goldstrike:
So this feature will be available in the next update?

PR #2224, opened 2018-04-02
[Design] Allow numeric values for FontSize

alanmcgovern:
This ensures that intellisense in Visual Studio
does not report an error when numeric values are
put in this property instead of the 'string' values
like 'Small', 'Large' etc

NOTE: This might need additional niceing up to ensure that actual invalid values are still marked as invalid.

### DIFF
--- a/Xamarin.Forms.Core.Design/Xamarin.Forms.Core.Design.csproj
+++ b/Xamarin.Forms.Core.Design/Xamarin.Forms.Core.Design.csproj
@@ -51,6 +51,7 @@
   <ItemGroup Condition=" '$(OS)' != 'Unix' ">
     <Compile Include="AttributeTableBuilder.cs" />
     <Compile Include="EnumConverter.cs" />
+    <Compile Include="NonExclusiveEnumConverter.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="RegisterMetadata.cs" />
   </ItemGroup>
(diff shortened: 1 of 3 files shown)


jassmith:
Strangely this doesn't build on OSX claiming EnumConverter<> doesn't exist

Xamarin.Forms.Core.Design/Xamarin.Forms.Core.Design.csproj
```diff
@@ -47,6 +47,7 @@
     <Compile Include="..\Xamarin.Forms.Core\Properties\GlobalAssemblyInfo.cs">
       <Link>Properties\GlobalAssemblyInfo.cs</Link>
     </Compile>
+    <Compile Include="NonExclusiveEnumConverter.cs" />
```
jassmith: This needs to be moved into the Windows only ItemGroup

alanmcgovern:
How does the code here ( https://github.com/xamarin/Xamarin.Forms.Design ) relate to the Xamarin.Forms.Design assembly generated in this repo?

rmarinho:
build --uitests

alanmcgovern:
Just to confirm - has this patch been merged to the required branches so the fix will ship with 15.7?

StephaneDelcroix:
@alanmcgovern 
> Just to confirm - has this patch been merged to the required branches so the fix will ship with 15.7?

it'll be in 3.0.0

PR #2323, opened 2018-04-06
[iOS] Grouped ListView will no longer crash when its ItemSource is cleared

jassmith:
### Description of Change ###

Fixes issue where resetting a Grouped ListView on iOS can result in a crash due to an order of operations issue on the TIL. The best fix as always is to move the data into the container of the Cell itself and thus not have to touch the TIL.

### Bugs Fixed ###

Bug reported on Twitter and modified test 45125 to actually cover this case so we still have coverage.

### API Changes ###

None

### Behavioral Changes ###

Less crashy crash

### PR Checklist ###


jassmith:
build --uitests

jassmith:
Failures unrelated :)

PR #2416, opened 2018-04-10
[iOS, UWP] Explicitly dispose of children renderers that belong to Packager

### LINKED ISSUE #2399: [iOS] ViewCell's with nested layout are not garbage collected causing huge memory leak
### Description

ViewCell's with nested layout (for example StackLayout with a ContentView) cause the whole ListView to be never destructed. The garbage collector seems unable to remove the ListView and it's ViewCells even though they are not referenced anymore.

### Steps to Reproduce

Build and run this code:

    {
        public App()
        {
            MainPage = new NavigationPage(new ContentPage
            {
                Content = new Button
                {
                    Text = "Show New ListView",
                    Command = new Command(o => MainPage.Navigation.PushAsync(new ListPage())),
                },
            });
        }
    }
    
    public class ListPage : ContentPage
    {
        public ListPage()
        {
            GC.Collect();
            Content = new ListView
            {
                ItemTemplate = new DataTemplate(typeof(Cell)),
                ItemsSource = new List<string> { "A" },
            };
        }
    }

    class Cell : ViewCell
    {
        string guid = Guid.NewGuid().ToString();

        public Cell()
        {
            Console.WriteLine("Constructor " + guid);
            View = new StackLayout
            {
                Children = {
                    new ContentView
                    {
                        Content = new Label { Text = guid },
                    }
                }
            };
        }

        ~Cell()
        {
            Console.WriteLine("Destructor " + guid);
        }
    }

Open and close the page with the ListView a few times.

### Expected Behavior

Destructor should be called.

### Actual Behavior

ListPage, ListView and the Cells are never removed by the garbage collector causing huge memory leak.

### Basic Information

- Version with issue: Xamarin.Forms 2.5 and 3.0-pre3
- IDE: Visual Studio for Mack
- Platform Target Frameworks:
  - iOS:  11.2

PureWeen:
#### iOS
Added code to VisualElementPackager to dispose of children. Renderers typically create cycle references due to watching the events of the Elements they represent

Renderer => HasElement => Wires it self to events that element owns => Element has Renderer
This causes the Renderers to not finalize because there's a reference cycle

#### UWP
When a ListViewRenderer is disposed nothing was trickling down that dispose operation to the cells/renderers inside the ListView.   Code was added to  Cleanup all the cells of a listview as well as disposing of children in UWP's version of a VisualElementPackager

### Bugs Fixed ###

- fixes #2399

### API Changes ###

None

### Behavioral Changes ###
Dispose is now going to be called on Renderers where it wasn't being called before. If users were  depending on this or have code in their dispose that's wrong then this change will effect those situations. 

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2399.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2399.cs
@@ -1,0 +1,145 @@
+﻿using System;
+using System.Threading.Tasks;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System.Linq;
+using Xamarin.Forms.Controls.Effects;
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+using Xamarin.Forms.Core.UITests;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 2399, "Label Renderer Dispose never called")]
+	public class Issue2399 : TestNavigationPage
+	{
+		static AttachedStateEffectList AttachedStateEffects = new AttachedStateEffectList();
+		Label AllEffectsHaveDetached;
+		protected override void Init()
+		{
+			AllEffectsHaveDetached = new Label()
+			{
+				AutomationId = "AllEventsHaveDetached",
+				Text = "If this text doesn't change then all effects haven't detached"
+			};
+
+			var newPage = new ContentPage
+			{
+				Content =
+					new StackLayout()
+					{
+						Children =
+						{
+							new Button
+							{
+								Text = "Show New ListView",
+								Command = new Command(async o => await Navigation.PushAsync(new ListPage())),
+							},
+
+							new Button
+							{
+								Text = "Garbage Collection Things",
+								Command = new Command(() =>
+								{
+									GC.Collect();
+									GC.WaitForPendingFinalizers();
+									AttachedStateEffects.Clear();
+								}),
+							},
+							AllEffectsHaveDetached
+						}
+					},
+				AutomationId = "ContentPage"
+
+			};
+
+			AttachedStateEffects.AllEventsAttached += OnAllEventsAttached;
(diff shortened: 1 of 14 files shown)


Xamarin.Forms.Platform.iOS/VisualElementPackager.cs
```diff
@@ -54,6 +54,22 @@ protected virtual void Dispose(bool disposing)
 
 			if (disposing)
 			{
+				if (ElementController != null)
+				{
+					for (var z = 0; z < ElementController.LogicalChildren.Count; z++)
```
adrianknight89: Any reason why you used `z` here instead of a typical `i`?

Xamarin.Forms.Platform.iOS/VisualElementPackager.cs
```diff
@@ -54,6 +54,22 @@ protected virtual void Dispose(bool disposing)
...
+					{
+						var child = ElementController.LogicalChildren[z] as VisualElement;
+						if (child == null)
+							continue;
+						var childRenderer = Platform.GetRenderer(child);
+
```
adrianknight89: Could remove the newline here and add it below the first `continue`.

Xamarin.Forms.Platform.iOS/VisualElementPackager.cs
PureWeen: Because I just copy and pasted the structure from lower down in the code where it iterates over the children :-)  Not a great reason but it's a reason

PureWeen:
build --uitests

PureWeen:
build --uitests

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2399.cs
```diff
@@ -0,0 +1,147 @@
...
+	public class Issue2399 : TestNavigationPage
+	{
+		static AttachedStateEffectList AttachedStateEffects = new AttachedStateEffectList();
+		Label AllEffectsHaveDetached;
+		protected override void Init()
+		{
+			AllEffectsHaveDetached = new Label()
+			{
+				AutomationId = "AllEventsHaveDetached",
+				Text = "If this text doesn't change then all effeects haven't detached"
```
samhouts: effeects

PR #2419, opened 2018-04-11
add debug type full property to control gallery netstandard library so break points are hit

PureWeen:
https://developercommunity.visualstudio.com/content/problem/29331/pdbs-are-not-generated-when-building-a-uwp-app-tha.html

Adding debugtype pdbonly or full to the NetStandard library causes the break points to start working in the Control Gallery project.

Reading around I'm still not quite clear whether this is a tooling bug or expected behavior


StephaneDelcroix:
this looks correct

PR #2426, opened 2018-04-11
[iOS, Android, UWP] Added padding to Button control

### LINKED ISSUE #1702: [Enhancement] Padding on Buttons
## Rationale

`Button`s do not currently have a `Padding` property, forcing some users to create hacks, custom renderers, or effects to achieve the desired layout.

## Implementation

Add bindable property to `Button` & make `Button` implement `IPaddingElement`:

```csharp
public static readonly BindableProperty PaddingProperty = PaddingElement.PaddingProperty;
```

Renderers will need to be updated accordingly.

### Android

- `Button` has no padding by default.
- Altering the `Padding` on a `Button` will change the size of the `Button` relative to the position of the `Text`. 

### iOS

See Android

### UWP

See Android

## Implications for CSS

`Button` will now support `padding`.

## Backward Compatibility

Since we are adding new properties, there should be no compatibility problems as long as:

1. We ensure that an `Button`'s padding defaults to 0, as is the expected behavior now,
2. We ensure that any Effects or Custom Renderers that may have been created to produce this behavior take precedence over any values that we set, to the best of our ability.
3. We ensure that this does not conflict with the new `TextAlignment` properties on `Button`s.

Third party renderers will need to be updated to ensure that this functionality is officially supported.

## Difficulty : Medium

paymicro:
### Description of Change ###

Added padding to Button control

**Screencasts**
UWP: http://recordit.co/RO6d9sqjNR
Android: http://recordit.co/mcJqcuVA65
iOS: http://recordit.co/3MjGiRf06X

### Bugs Fixed ###

Fixes #1702

### API Changes ###

Added:

    Thickness Padding { get; set; } //Bindable Property in Button

### Behavioral Changes ###

User will be able to set Padding for Button.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Xamarin.Forms.Controls.Issues.Shared.projitems
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Xamarin.Forms.Controls.Issues.Shared.projitems
@@ -239,6 +239,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)Bugzilla59580.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Effects\AttachedStateEffect.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Effects\AttachedStateEffectList.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)GitHub1702.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)GitHub2598.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)GitHub1878.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Issue1677.cs" />
--- a/Xamarin.Forms.Core/Button.cs
+++ b/Xamarin.Forms.Core/Button.cs
@@ -8,7 +8,7 @@ using Xamarin.Forms.Platform;
 namespace Xamarin.Forms
 {
 	[RenderWith(typeof(_ButtonRenderer))]
-	public class Button : View, IFontElement, ITextElement, IBorderElement, IButtonController, IElementConfiguration<Button>
+	public class Button : View, IFontElement, ITextElement, IBorderElement, IButtonController, IElementConfiguration<Button>, IPaddingElement
 	{
 		const double DefaultSpacing = 10;
 		const int DefaultBorderRadius = 5;
@@ -49,6 +49,24 @@ namespace Xamarin.Forms
 		public static readonly BindableProperty ImageProperty = BindableProperty.Create("Image", typeof(FileImageSource), typeof(Button), default(FileImageSource),
 			propertyChanging: (bindable, oldvalue, newvalue) => ((Button)bindable).OnSourcePropertyChanging((ImageSource)oldvalue, (ImageSource)newvalue),
 			propertyChanged: (bindable, oldvalue, newvalue) => ((Button)bindable).OnSourcePropertyChanged((ImageSource)oldvalue, (ImageSource)newvalue));
+
+		public static readonly BindableProperty PaddingProperty = PaddingElement.PaddingProperty;
+
+		public Thickness Padding
+		{
+			get { return (Thickness)GetValue (PaddingElement.PaddingProperty); }
+			set { SetValue (PaddingElement.PaddingProperty, value); }
+		}
+
+		Thickness IPaddingElement.PaddingDefaultValueCreator ()
+		{
+			return default (Thickness);
+		}
+
+		void IPaddingElement.OnPaddingPropertyChanged (Thickness oldValue, Thickness newValue)
+		{
+			InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
+		}
 
 		readonly Lazy<PlatformConfigurationRegistry<Button>> _platformConfigurationRegistry;
 
--- a/Xamarin.Forms.Platform.Android/FastRenderers/ButtonRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/FastRenderers/ButtonRenderer.cs
@@ -27,6 +27,7 @@ namespace Xamarin.Forms.Platform.Android.FastRenderers
 		readonly EffectControlProvider _effectControlProvider;
 		VisualElementTracker _tracker;
 		ButtonBackgroundTracker _backgroundTracker;
+		Thickness _paddingDeltaPix = new Thickness();
 
 		public event EventHandler<VisualElementChangedEventArgs> ElementChanged;
 		public event EventHandler<PropertyChangedEventArgs> ElementPropertyChanged;
@@ -245,6 +246,7 @@ namespace Xamarin.Forms.Platform.Android.FastRenderers
 				UpdateInputTransparent();
 				UpdateBackgroundColor();
 				UpdateDrawable();
+				UpdatePadding();
 
 				ElevationHelper.SetElevation(this, e.NewElement);
 			}
@@ -282,6 +284,10 @@ namespace Xamarin.Forms.Platform.Android.FastRenderers
 			{
 				UpdateInputTransparent();
 			}
+			else if (e.PropertyName == Button.PaddingProperty.PropertyName)
+			{
+				UpdatePadding();
+			}
 
 			ElementPropertyChanged?.Invoke(this, e);
 		}
@@ -297,9 +303,13 @@ namespace Xamarin.Forms.Platform.Android.FastRenderers
 			{
 				// We've got an image (and no text); it's already centered horizontally,
 				// we just need to adjust the padding so it centers vertically
-				int diff = (b - t - _imageHeight) / 2;
+				var diff = ((b - Context.ToPixels(Button.Padding.Bottom + Button.Padding.Top)) - t - _imageHeight) / 2;
 				diff = Math.Max(diff, 0);
-				SetPadding(0, diff, 0, -diff);
+				UpdateContentEdge(new Thickness(0, diff, 0, -diff));
+			}
+			else
+			{
+				UpdateContentEdge();
 			}
 
 			base.OnLayout(changed, l, t, r, b);
@@ -362,7 +372,7 @@ namespace Xamarin.Forms.Platform.Android.FastRenderers
 				// to handle the vertical centering 
 
 				// Clear any previous padding and set the image as top/center
-				SetPadding(0, 0, 0, 0);
+				UpdateContentEdge();
 				SetCompoundDrawablesWithIntrinsicBounds(null, image, null, null);
 
 				// Keep track of the image height so we can use it in OnLayout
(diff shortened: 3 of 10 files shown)


Xamarin.Forms.Platform.Android/Renderers/ButtonRenderer.cs
```diff
@@ -260,6 +263,24 @@ void UpdateTextColor()
 			_textColorSwitcher?.UpdateTextColor(Control, Element.TextColor);
 		}
 
+		Thickness paddingDelta = new Thickness();
+
+		void UpdatePadding()
```
PureWeen: These changes need to be applied to the AppCompat Renderer and the Fast Renderer 

https://github.com/xamarin/Xamarin.Forms/blob/b96f65bdd9e7e931971338b122a8940e2a35ccc7/Xamarin.Forms.Platform.Android/AppCompat/ButtonRenderer.cs
https://github.com/xamarin/Xamarin.Forms/blob/b96f65bdd9e7e931971338b122a8940e2a35ccc7/Xamarin.Forms.Platform.Android/FastRenderers/ButtonRenderer.cs

Xamarin.Forms.Platform.UAP/ButtonRenderer.cs
```diff
@@ -250,5 +257,15 @@ void UpdateTextColor()
 		{
 			Control.Foreground = Element.TextColor != Color.Default ? Element.TextColor.ToBrush() : (Brush)Windows.UI.Xaml.Application.Current.Resources["DefaultTextForegroundThemeBrush"];
 		}
+
+		void UpdatePadding()
+		{
+			Control.Padding = new WThickness(
```
PureWeen: These arguments are being passed in the incorrect order.

Xamarin.Forms.Platform.Android/Renderers/ButtonRenderer.cs
```diff
@@ -260,6 +263,24 @@ void UpdateTextColor()
...
+		{
+			Control?.SetPadding (
```
PureWeen: These arguments are being passed in the incorrect order.

Xamarin.Forms.Platform.iOS/Renderers/ButtonRenderer.cs
```diff
@@ -268,7 +292,7 @@ void ComputeEdgeInsets(UIButton button, Button.ButtonContentLayout layout)
 			{
 				button.ImageEdgeInsets = new UIEdgeInsets(0, labelWidth + spacing, 0, -labelWidth - spacing);
 				button.TitleEdgeInsets = new UIEdgeInsets(0, -imageWidth - spacing, 0, imageWidth + spacing);
-				button.ContentEdgeInsets = new UIEdgeInsets(0, 2 * spacing, 0, 2 * spacing);
+				UpdateContentEdge (button, new Thickness (0, 2 * spacing, 0, 2 * spacing));
```
PureWeen: Thickness is Left, top, right, bottom
UIEdgeInset is Top, right, bottom, left

So I think all these places where you replaced UIEdgeInsets with Thickness will need to account for this

Xamarin.Forms.Core/Button.cs
```diff
@@ -50,6 +50,29 @@ public class Button : View, IFontElement, ITextElement, IBorderElement, IButtonC
...
+		}
+
+		Thickness IPaddingElement.PaddingDefaultValueCreator ()
+		{
+			return default (Thickness);
+		}
+
+		void IPaddingElement.OnPaddingPropertyChanged (Thickness oldValue, Thickness newValue)
+		{
+			UpdatePadding (newValue);
```
PureWeen: Invalidate Measure
When the padding is changed the measure on the button is going to change so that needs to Invalidate the Measure on the button.  You can see the issue with this if you set the padding after the button has renderered vs before.  If you don't invalidate the measure then it would look different in each case because the height/width won't adjust to the new padding

Xamarin.Forms.Platform.Android/FastRenderers/ButtonRenderer.cs
```diff
@@ -294,7 +299,7 @@ protected override void OnLayout(bool changed, int l, int t, int r, int b)
 				// we just need to adjust the padding so it centers vertically
 				int diff = (b - t - _imageHeight) / 2;
 				diff = Math.Max(diff, 0);
-				SetPadding(0, diff, 0, -diff);
+				UpdateContentEdge(new Thickness (0, diff, 0, -diff));
```
PureWeen: I think the math here is going to need to account for the padding when it sets the content edges. If you create a Button with bottom padding and an image
```C#
new Button()
{
     Image = "coffee.png",
     BackgroundColor = Color.Green,
     Padding = new Thickness(0, 0, 0, 100)
}
```

Then on android it looks like it takes that padding and evens it out between top and bottom ultimately centering the image instead of giving it bottom padding.  If you give it top padding and not bottom padding then it starts to push the image out of the button view

I have the code below and it mostly seems to work though I think it causes the padding to not effect the button evenly since to be perfect we'd have to know the center without the padding and use that for a constant diff. Let me check around and get some additional thoughts on the best way to handle this

```C#
	protected override void OnLayout(bool changed, int l, int t, int r, int b)
		{
			if (_imageHeight > -1)
			{	
				var bottomPadding = Context.ToPixels(Element.Padding.Bottom);
				var topPadding = Context.ToPixels(Element.Padding.Top);
				// We've got an image (and no text); it's already centered horizontally,
				// we just need to adjust the padding so it centers vertically
				var diff = ((b - bottomPadding - topPadding) - t - _imageHeight) / 2;
				diff = Math.Max(diff, 0);
				UpdateContentEdge(new Thickness(0, diff, 0, -diff));
			}

			base.OnLayout(changed, l, t, r, b);
		}
```

paymicro: I think if a padding element is specified at the bottom or top then you need to ignore the diff padding "button" with the image. 
I think it is not necessary to increase the padding, if it was specified with the user.

Xamarin.Forms.Platform.Android/Renderers/ButtonRenderer.cs
```diff
@@ -55,13 +56,13 @@ public override SizeRequest GetDesiredSize(int widthConstraint, int heightConstr
 
 		protected override void OnLayout(bool changed, int l, int t, int r, int b)
 		{
-			if (_imageHeight > -1)
+			if (_imageHeight > -1 && Element.Padding.Bottom == 0 && Element.Padding.Top == 0)
```
PureWeen: So I don't think just opting out like this quite works.  I've had success with this

```C#
protected override void OnLayout(bool changed, int l, int t, int r, int b)
		{
			if (_imageHeight > -1)
			{
				// We've got an image (and no text); it's already centered horizontally,
				// we just need to adjust the padding so it centers vertically
				var diff = ((b - Context.ToPixels(Element.Padding.Bottom + Element.Padding.Top))  - t - _imageHeight) / 2;
				diff = Math.Max(diff, 0);
				UpdateContentEdge(new Thickness(0, diff, 0, -diff));

			}
			else
			{
				UpdateContentEdge();
			}

			base.OnLayout(changed, l, t, r, b);
		}
```

Here's some tester code to try. Using the code you have if I start adding top/bottom padding it causes it to jump to jump as you add padding.

```C#
StackLayout layout = null;
			layout = new StackLayout()
			{
				Children =
					{
						new Button()
						{
							Image = "coffee.png",
							BackgroundColor = Color.Green,
							Padding = new Thickness(),
							HeightRequest = 400,
						},
						new Button()
						{
							Text = "Bottom",
							Command = new Command(() =>
							{
								var button = layout.Children[0] as Button;
								button.Padding = new Thickness(0, button.Padding.Top, 0,button.Padding.Bottom + 10);
								if(!String.IsNullOrWhiteSpace(button.Text))
									button.Text = "Top: " + button.Padding.Top.ToString() + " Bottom: " + button.Padding.Bottom.ToString();
							})
						},
						new Button()
						{
							Text = "Top",
							Command = new Command(() =>
							{
								var button = layout.Children[0] as Button;
								button.Padding = new Thickness(0, button.Padding.Top + 10, 0, button.Padding.Bottom);
								if(!String.IsNullOrWhiteSpace(button.Text))
									button.Text = "Top: " + button.Padding.Top.ToString() + " Bottom: " + button.Padding.Bottom.ToString();
							})
						},
						new Button()
						{
							Text = "Reset",
							Command = new Command(() =>
							{
								var button = layout.Children[0] as Button;
								button.Padding = new Thickness(0, 0, 0,0);
							})
						},
						new Button()
						{
							Text = "Set Text",
							Command = new Command(() =>
							{
								var button = layout.Children[0] as Button;
								button.Text = "some text";
							})
						},
						new Button()
						{
							Text = "Reset Text",
							Command = new Command(() =>
							{
								var button = layout.Children[0] as Button;
								button.Text = String.Empty;
							})
						},
					}
			};

			return new ContentPage()
			{
				Content = layout
			};
```

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Xamarin.Forms.Controls.Issues.Shared.projitems
```diff
@@ -239,6 +239,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)Bugzilla59580.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Effects\AttachedStateEffect.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Effects\AttachedStateEffectList.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)GitHub1702.cs" />
```
PureWeen: Rename to Issue1702.cs

paymicro: I noticed that in `Issue####` classes some other `github` number is specified and different description.
For example `Issue1461.cs` created almost 2 years ago
https://github.com/xamarin/Xamarin.Forms/blob/b794caf7be33e5d3bf8ea39ba719cdbfbc008315/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1461.cs#L30-L31
And `Github1461.cs` created a couple of months ago
https://github.com/xamarin/Xamarin.Forms/blob/b794caf7be33e5d3bf8ea39ba719cdbfbc008315/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Github1461.cs#L13-L14
`Github1461.cs` is correct for this repository https://github.com/xamarin/Xamarin.Forms/issues/1461

I was confused. Is renaming to `Issue####` still valid?

PureWeen: Yea :-/ Right now there's a mix of numbers that relate to the old github repo before the project went public.  For now we're still just creating the Issues as Issue#### when there isn't a number collision

Xamarin.Forms.Core/Button.cs
```diff
@@ -50,6 +50,30 @@ public class Button : View, IFontElement, ITextElement, IBorderElement, IButtonC
...
+			return default (Thickness);
+		}
+
+		void IPaddingElement.OnPaddingPropertyChanged (Thickness oldValue, Thickness newValue)
+		{
+			UpdatePadding(newValue);
+			InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
+		}
+
+		protected virtual void UpdatePadding (Thickness newValue)
```
StephaneDelcroix: this isn't necessary. subclass of Button have other ways of detecting Padding changes.

PureWeen:
Yea they are all from the Scale Matrix error. Once this is rebased with the fix from a few hours ago those should pass.

PureWeen:
build --uitests

czuvich:
I think this change adversely impacted buttons with an Image and Text.  It looks like the Image and Text now hug the boundaries of the button.  I've tried changing the ContentLayout to different options, but it has no effect (at least on Android).  I'll do some more testing and submit a bug report.

nwilliams:
Apologies if this is not the proper place for such a question/comment, but does anybody else find it strange that the `Padding` property on a button doesn't seem to resize the control when using `HorizontalOptions` or `VerticalOptions` set to `Start`, `Center`, or `End`? All I can see it doing is moving the text around _inside_ the button -- and if I use values that exceed the calculated size of the button, the text disappears altogether. I could get around this by instead using `WidthRequest` and `HeightRequest`, but my button text is dynamic so it's not as straight forward, and really seems like a workaround to how the `Padding` property _should_ work.

PR #2431, opened 2018-04-11
[Designer] Do not prefix the toolbox images

alanmcgovern:
The intent is that we can load the toolbox xml file from an
assembly by looking it up by  *exact* name. That means we
should not prefix it with assembly specific info.

Similarly the lookup of the images is *supposed* to be handled
by using *exact* lookups using the format "{platform}.{filename}"
where the filename is the one stored in the toolbox xml file.

This change makes that happen by correctly setting the LogicalName.

### DIFF
--- a/Xamarin.Forms.Core.Design/Xamarin.Forms.Core.Design.csproj
+++ b/Xamarin.Forms.Core.Design/Xamarin.Forms.Core.Design.csproj
@@ -65,302 +65,16 @@
   </ItemGroup>
   <ItemGroup Condition=" '$(OS)' != 'Unix' ">
     <None Include="packages.config" />
-    <None Include="toolbox\icons\mac\layout-FlexLayout-16.png" />
-    <None Include="toolbox\icons\mac\layout-FlexLayout-16%402x.png" />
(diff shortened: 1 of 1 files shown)


alanmcgovern:
This is what the embedded resources look like after the change:  https://gist.github.com/alanmcgovern/e708a2070d07fb4722b932e1cfc0ad88

Which means we don't have to hardcode `private const string MacIconResourceFormatString = "Xamarin.Forms.Core.Design.toolbox.icons.mac.{0}";` in the image loading code.

alanmcgovern:
flexlayout mac: https://gist.github.com/alanmcgovern/e708a2070d07fb4722b932e1cfc0ad88#file-gistfile2-txt-L60-L67

flexlayout win: https://gist.github.com/alanmcgovern/e708a2070d07fb4722b932e1cfc0ad88#file-gistfile2-txt-L267

NOTE: Any new images will Just Work and don't need to be manually added to the csproj

alanmcgovern:
Just a heads up - we're going to assume this will be in the 3.0 release and will base our support on the 'correct' names :)

PR #2433, opened 2018-04-11
[iOS] Picker IsFocused not being set back to false if nothing in picker was selected

### LINKED ISSUE #2398: [iOS] Focus stops working for picker after it's already been focused then unfocused
### Description
Picker.Focus
Picker.Unfocus
Picker.Focus <- does not work

Thinking this might be the cause
https://github.com/xamarin/Xamarin.Forms/pull/1278

Because if you select something in the list then the focus/unfocus works fine

### Steps to Reproduce

Run control gallery test Issue2339 and just try to focus then unfocus then focus again

### Expected Behavior
Picker should keep being able to focus

### Actual Behavior
Picker stops being able to focus

### Basic Information

- Version with issue: 3.0.0
- Last known good version: 2.5.1

### Implementation notes
The original purpose of the Control Gallery test was for windows but it seems from this PR
https://github.com/xamarin/Xamarin.Forms/pull/762
That that's ok

The Issue2339 should be updated to reflect that it only applies to iOS/Android and the behavior isn't consistent on UWP on purpose

PureWeen:
### Description of Change ###

Fixed picker on iOS so when nothing gets selected and it's closed that IsFocused gets correctly set back to false

This is a 2.5.1 to 3.0.0 regression so let me know if you want me to retarget this to 3.0.0

### Bugs Fixed ###

fixes #2398

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
@@ -39,7 +39,6 @@ namespace Xamarin.Forms.ControlGallery.Android
 			//Window.AddFlags(WindowManagerFlags.Fullscreen | WindowManagerFlags.TurnScreenOn);
 
 			base.OnCreate(bundle);
-			
 
 #if TEST_EXPERIMENTAL_RENDERERS
 			Forms.SetFlags("FastRenderers_Experimental");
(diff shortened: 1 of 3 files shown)


rmarinho:
Yeah @PureWeen I think you should target this to 3.0.0

PureWeen:
rebased to 3.0.0

jassmith:
build --uitests

PR #2472, opened 2018-04-16
[UWP] Fixes the search bar query, when you type text quickly

### LINKED ISSUE #2439: [UWP] Searchbar Command Parameter binding happens after the Enter key triggers the command
### Description
When you type into the SearchBar and then hit Enter the Enter key is triggering the command before the SearchBar Text binding resolves. The effect of this is that if you type really fast and then hit enter whatever was present before you started typing is what gets submitted to the Command

### Steps to Reproduce

1.  Run the sample
2.  type in the box quickly and then hit Enter right after typing. The text that's typed into that box gets set to a Label so you should see the text there after you hit Enter. But you will either see nothing or just part of the text (if you didn't type fast enough)
3.  Hit Enter again. And now you will see the text show up

If you type in the box quickly, wait a moment, and then hit enter. All the text gets captured

### Expected Behavior
Everything I typed before hitting enter should be captured into the Bound Search Parameter

### Actual Behavior
Only some of what I typed gets bound

### Basic Information

- Version with issue: 2.5 and 3.0
- Platform Target Frameworks
  - UWP:  16299 

### Reproduction Link
[UwpSearchBar.zip](https://github.com/xamarin/Xamarin.Forms/files/1901000/UwpSearchBar.zip)

paymicro:
### Description of Change ###

Modifies the text of the SearchBar element if it does not match the query.

### Bugs Fixed ###

Fixes #2439

### API Changes ###

--

### Behavioral Changes ###

What will be typed in the search bar will be in the query regardless of the speed of typing.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/SearchBarRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/SearchBarRenderer.cs
@@ -102,6 +102,11 @@ namespace Xamarin.Forms.Platform.UWP
 
 		void OnQuerySubmitted(AutoSuggestBox sender, AutoSuggestBoxQuerySubmittedEventArgs e)
 		{
+			// Modifies the text of the control if it does not match the query.
+			// This is possible because OnTextChanged is fired with a delay
+			if (e.QueryText != Element.Text)
+				Element.SetValueFromRenderer(SearchBar.TextProperty, e.QueryText);
+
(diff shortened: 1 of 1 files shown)


samhouts:
Thank you for the PR! Would you mind rebasing and retargeting it to the 3.0.0 branch?

PR #2476, opened 2018-04-16
Adding support for tapping on iOS Slider to set value

### LINKED ISSUE #1729: [Enhancement] iOS PlatformSpecific for UISlider
## Rationale

There is currently no way to to set the value when tapping a slider on iOS

## Implementation.

Add a PlatformSpecific property for Slider on iOS. Customer renderer changes to subscribe to the tap event, calculate position and set the value. `((IElementController)Element).SetValueFromRenderer(Slider.ValueProperty, Value)

```csharp
Slider.On<iOS>().UpdateOnTap(true/false);
```

### Android

None

### iOS
Value of slider should be updated when tapping

### UWP

None

## Implications for CSS

None

## Backward Compatibility

We should set default to false and make sure all custom renderers continue to work even with our implementation. 
Should respect and have some UITest for InputTransparent and and IsEnabled properties.
Should work with a tap gesture by allowing bubbling of the event to the parent.

henricm:
### Description of Change ###

This PR adds support for tapping a slider on iOS to set the value which is not natively support by a standard `UISlider`. This new behaviour is triggered by setting a new iOS specific property `UpdateOnTap` to `true`.

### Bugs Fixed ###

- fixes #1729 

### API Changes ###

Added:
 - bool Slider.UpdateOnTap { get; set; } //iOS specific Bindable Property

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/SliderRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/SliderRenderer.cs
@@ -2,6 +2,8 @@ using System;
 using System.ComponentModel;
 using UIKit;
 using SizeF = CoreGraphics.CGSize;
+using Xamarin.Forms.PlatformConfiguration.iOSSpecific;
+using Specifics = Xamarin.Forms.PlatformConfiguration.iOSSpecific.Slider;
 
 namespace Xamarin.Forms.Platform.iOS
 {
@@ -9,6 +11,7 @@ namespace Xamarin.Forms.Platform.iOS
 	{
 		SizeF _fitSize;
 		UIColor defaultmintrackcolor, defaultmaxtrackcolor, defaultthumbcolor;
+		UITapGestureRecognizer _sliderTapRecognizer;
 
 		public override SizeF SizeThatFits(SizeF size)
 		{
@@ -18,7 +21,9 @@ namespace Xamarin.Forms.Platform.iOS
 		protected override void Dispose(bool disposing)
 		{
 			if (Control != null)
+			{
 				Control.ValueChanged -= OnControlValueChanged;
+			}
 
 			base.Dispose(disposing);
 		}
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Platform.iOS/Renderers/SliderRenderer.cs
```diff
@@ -152,13 +157,47 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
...
+			{
+				if (Element.OnThisPlatform().GetUpdateOnTap())
+				{
+					if (_sliderTapRecognizer == null)
+					{
+						_sliderTapRecognizer = new UITapGestureRecognizer((recognizer) =>
+						{
+							var control = Control;
+							var tappedLocation = recognizer.LocationInView(control);
+							var val = (tappedLocation.X - control.Frame.X) * control.MaxValue / control.Frame.Size.Width;
```
PureWeen: In the control gallery if you click really fast 5 times it causes the control gallery to reset which then causes a NRE here. It might be an edge case but adding a null check here on *CONTROL* would probably be a safe idea.

Xamarin.Forms.Platform.iOS/Renderers/SliderRenderer.cs
```diff
@@ -18,7 +21,9 @@ public override SizeF SizeThatFits(SizeF size)
 		protected override void Dispose(bool disposing)
 		{
 			if (Control != null)
+			{
 				Control.ValueChanged -= OnControlValueChanged;
+			}
```
PureWeen: Here if the _sliderTapRecognizer is not null can you remove it from the control? That will probably fix the NRE exception I commented about as well but probably good to just remove it anyway to avoid other side effects.

Xamarin.Forms.Platform.iOS/Renderers/SliderRenderer.cs
henricm: @PureWeen Good catch. I've fixed this in new commit by adding null checks to `control` and `tappedLocation`. Seems like this happens because the control has already been removed when the last tap event is handled and this callback is called.

Xamarin.Forms.Platform.iOS/Renderers/SliderRenderer.cs
henricm: @PureWeen I've fixed this now - great you noticed this! Don't think this was the cause to the NRE though.

Xamarin.Forms.Platform.iOS/Renderers/SliderRenderer.cs
```diff
@@ -152,13 +162,51 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
...
 		}
 
 		void OnControlValueChanged(object sender, EventArgs eventArgs)
 		{
 			((IElementController)Element).SetValueFromRenderer(Slider.ValueProperty, Control.Value);
 		}
 
+		void UpdateTapRecognizer()
+		{
+			if (Element != null && Element.IsSet(Specifics.UpdateOnTapProperty))
```
rmarinho: Iwould try avoid nesting  `if (Element == null || Control == null || !Element.IsSet(Specifics.UpdateOnTapProperty)) return; `

and save the check on the bottom on 185

henricm: The check on 185 is for the gesture recognizer callback though. It was needed since tap event was received when the control already has been removed in certain cases.

PR #2432, opened 2018-04-11
[UWP] Multi-window

wachs:
### Description of Change ###

Enable multi-window on UWP by doing the following changes:
- Application: 
   - Single Instance => Multiple instances;
- CoreDispatcher: 
   - Single Dispatcher => Always associated with a WindowId (property);
- BeginInvokeOnMainThread: 
   - Single overload (using a single Dispatcher) => New overloads (with WindowId and BindableObject). Invokes on the corresponding Dispatcher;

Every piece of Xamarin.Forms.Core uses the new Device.BeingInvokeOnMainThread overloads (sending the BindableObject or the WindowId)

### Bugs Fixed ###

- fixes #2229 
- fixes #6189

### API Changes ###

Added:
 - Guid BindableObject.WindowId { get; set; }
 - public interface Xamarin.Forms.Internals.IPlatformServices
    - void BeginInvokeOnMainThread(Action action, Guid windowId);
    - void BeginInvokeOnMainThread(Action action, BindableObject bindableObject);

### Behavioral Changes ###

There is no significant behavioral changes on this version. The only exception is: If executions inside the secondary windows UI Threads are needed, a call to the new Device.BeginInvokeOnMainThread overloads will be necessary (using WindowId or BindableObject params).

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

Issue #2369, opened 2018-04-07
[UWP] Crash when opening multiple Alerts


samhouts:
### Description

When multiple alerts are opened at the same time, UWP crashes with "Only a single ContentDialog can be open at any time."

Related to #881 

### Steps to Reproduce

1. Run B43469 or B45743 on Control Gallery

### Expected Behavior

Test loads

### Actual Behavior

Test crashes

### Basic Information

- Version with issue: 3.0.0
- Last known good version: 2.5.1

ashishm29:
I am facing same issue with 3.0.0.550146 (Latest Version).

Issue #1790, opened 2018-02-03
[UWP] Calling Display Alert from Windows Assigned Access causes app to crash


PureWeen:
### Description
If an app is opened via a share target then Xamarin Forms crashes when checking IsInvokedRequired

```
		public bool IsInvokeRequired => LockApplicationHost.GetForCurrentView() != null
			? !CoreApplication.GetCurrentView().Dispatcher.HasThreadAccess
			: !CoreApplication.MainView.CoreWindow.Dispatcher.HasThreadAccess;
```

See screen shot for exception

### Steps to Reproduce

1.  Run the attached reproduction
2. Close the app
3.  share something of content type "TEXT" to the app .  I was just using the MAP app and sharing from there

The code calls DisplayAlert from the *OnShareTargetActivated* override on uwp.

### Basic Information

- Version with issue: Xamarin Forms: 2.5.0.122203 
- IDE: VS 2017
- Platform Target Frameworks:
  - UWP:  16299 

### Expected Behavior
The app should Display the Alert and not crash

### Actual Behavior

The app crashes

### Screenshots

![image](https://user-images.githubusercontent.com/5375137/35771375-068c30f2-08e9-11e8-933c-317f1e948dd0.png)

### Reproduction Link

[App4.zip](https://github.com/xamarin/Xamarin.Forms/files/1964739/App4.zip)

AlexF12004Roma:
Is there a workaround for this issue?
I cannot publish any UWP app with share option due to this bug?

PureWeen:
I posted some workarounds for some scenarios here
https://github.com/xamarin/Xamarin.Forms/issues/1783

ryanholden8:
FWIW - Currently encountering the same issue

PureWeen:
I'm no longer seeing this issue in 3.0 pre4. @ryanholden8   @AlexF12004Roma   can you test on pre4 see if that fixes your issue?

AlexF12004Roma:
I cannot test on my Project because i need target UWP W10 Anniversary (14393) and XF 3 is not compatible.

PureWeen:
I tested the sample I made in 2.5 and 3.1 and for me this now works in 3.1

Issue #1783, opened 2018-02-02
UWP - OnShareTargetActivated Init Error


AlexF12004Roma:
### Description

When a Forms App starts after a onShareTarget invocation,  Xamarin.Forms.Forms.Init(e) throws an Exception.

### Steps to Reproduce

1.  Add Share Target declaration in UWP (any kind)
2.  Share with the APP
3.  Try to Init Forms

### Expected Behavior
Frame rootFrame initialized

### Actual Behavior
Exception

### Basic Information

- Version with issue: 2.5.0
- IDE: VS2017
- Platform Target Frameworks: 
  - UWP:  Win10 November Update

PureWeen:
This is an interesting (annoying?) one. 

The issue here is that on UWP Xamarin Forms only really works single windowed so you can't re-init the new window opened by OnShareTargetActivated  if you already have a window opened.

To handle a case where the app is opened from OnShareTargetActivated you essentially just have to call the same code in OnLaunched. Both of those args just inherit from *IActivatedEventArgs* so you can pretty much just copy and paste the code from OnLaunched or just move it to a common method

As far as cases where the app is already running you can still control the open window so in this case you can close the new window and use the already running one.

Usually I'll just add logic in and inspect the passed in args to determine if the app is opening new or if it's already running

Not a perfect solution but that's currently how the interactions with UWP work

AlexF12004Roma:
Mmmm.

I  put Forms init in a common method, when the app is already running rootFrame is not null but 
rootFrame.Content is null and if i call "rootFrame.Navigate(typeof(MainPage), files);" i got Exception

````
protected async override void OnShareTargetActivated(ShareTargetActivatedEventArgs args)
        {
            base.OnShareTargetActivated(args);
            Frame rootFrame = InitForms(args);
            //rootFrame.Navigate(typeof(MainPage), files);
            Window.Current.Activate();
            //Chiude L'app
            op.ReportSubmittedBackgroundTask();
        }



`private Frame InitForms(IActivatedEventArgs e)
        {
            Frame rootFrame = Window.Current.Content as Frame;

            // Do not repeat app initialization when the Window already has content,
            // just ensure that the window is active
            if (rootFrame == null)
            {
                // Create a Frame to act as the navigation context and navigate to the first page
                rootFrame = new Frame();

                rootFrame.NavigationFailed += OnNavigationFailed;

                Xamarin.Forms.Forms.Init(e); 

[...]

PR #2526, opened 2018-04-23
[UWP] Fix crashes animation in the thread pool

### LINKED ISSUE #2327: [UWP] ControlGallery crashes running UI test 39821
### Description

The test for 39821 crashes on UWP.

### Steps to Reproduce

1. Navigate to the test.
2. Click the "Animate" button.

### Expected Behavior

The box will be animated.

### Actual Behavior

The application crashes.

paymicro:
### Description of Change ###

Check `HasThreadAccess` instead checking `LockApplicationHost` on `IsInvokeRequired`

### Bugs Fixed ###

Fixes #2327

### API Changes ###

None

### Behavioral Changes ###

`IsInvokeRequired` return true if `_dispatcher.HasThreadAccess` is false

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/WindowsBasePlatformServices.cs
+++ b/Xamarin.Forms.Platform.UAP/WindowsBasePlatformServices.cs
@@ -113,13 +113,7 @@ namespace Xamarin.Forms.Platform.UWP
 			return new WindowsIsolatedStorage(ApplicationData.Current.LocalFolder);
 		}
 
-		// Per https://docs.microsoft.com/en-us/windows-hardware/drivers/partnerapps/create-a-kiosk-app-for-assigned-access:
-		// "Each view or window has its own dispatcher. In assigned access mode, you should not use the MainView dispatcher, 
-		// instead you should use the CurrentView dispatcher." Checking to see if this isn't null (i.e. the current window is
-		// running above lock) calls through GetCurrentView(), and otherwise through MainView.
-		public bool IsInvokeRequired => LockApplicationHost.GetForCurrentView() != null
-			? !CoreApplication.GetCurrentView().Dispatcher.HasThreadAccess
-			: !CoreApplication.MainView.CoreWindow.Dispatcher.HasThreadAccess;
+		public bool IsInvokeRequired => !_dispatcher.HasThreadAccess;
 
 		public string RuntimePlatform => Device.UWP;
 


hartez:
I will review this on Monday - we need to reconcile this with issue #1783 and issue #1790. I need to dig into it to be sure, but I suspect that we are incorrectly using a cached dispatcher for some situations in which it is not appropriate.

@PureWeen you should take a look at this one and also at PR #1534 if you get a chance.

PureWeen:
I'm no longer seeing issue https://github.com/xamarin/Xamarin.Forms/issues/1790 on 3.0 pre4. Not quite sure yet what would have fixed it yet

PureWeen:
I'm guessing the DisplayAlert changes @hartez  made here
https://github.com/xamarin/Xamarin.Forms/issues/2369

Fixed the DisplayAlert Issue in https://github.com/xamarin/Xamarin.Forms/issues/1790

But I don't think this fix resolves the initial exception causing 1790 If I just add an empty IsInvokeRequired Check before requesting the DisplayAlert
```C#
  if(Xamarin.Forms.Device.IsInvokeRequired)
            {
                System.Diagnostics.Debug.WriteLine("No longer required");
            }
```

Then I still get the same exception detailed in 1790

PureWeen:
@hartez  @paymicro   is there any point to have this code at all?
```C#
					return LockApplicationHost.GetForCurrentView() != null
					? !CoreApplication.GetCurrentView().Dispatcher.HasThreadAccess
					: !CoreApplication.MainView.CoreWindow.Dispatcher.HasThreadAccess;
```

Can the property just be 
```
return !_dispatcher.HasThreadAccess;
```

That's the dispatcher the Invoke happens against 
```c#
public void BeginInvokeOnMainThread(Action action)
		{
			_dispatcher.RunAsync(CoreDispatcherPriority.Normal, () => action()).WatchForError();
		}
```

So it doesn't seem like it makes sense to use the static check for IsInvokeRequired and then the local dispatcher for invoke.  

If we just use the local _dispatcher that will also fix 1790 as well as all the issues for this come from accessing those statics

I tested with just using _dispatcher.HasThreadAccess and the UI test this is for passes and https://github.com/xamarin/Xamarin.Forms/pull/1534 passes when running in assigned access mode.

hartez:
@PureWeen That's kind of what I was wondering - if we can just use the cached dispatcher forever. My questions about this are:
1. Will the cached dispatcher work in the kiosk situation?
2. Is the dispatcher we're caching the correct one (Window.Current.Dispatcher is what gets passed in). 

If that is the correct dispatcher for all our purposes, then we probably don't need for it to be injected in the constructor - we can just get rid of that parameter. 

If that dispatcher doesn't work for kiosk situations, then we need some code to detect that we're in kiosk mode so we can return the correct dispatcher.

PureWeen:
> Will the cached dispatcher work in the kiosk situation?

It did for the tests I ran.  I setup Control Gallery so I could set it up with Assigned Access. Remote deployed it to my laptop. Set it up in kiosk mode for a local user and everything worked fine. I ran the UI Test this PR is created to fix and it Animated without crashing.

>  Is the dispatcher we're caching the correct one (Window.Current.Dispatcher is what gets passed in).

https://github.com/xamarin/Xamarin.Forms/blob/3b0a6d243228e26070b752dad004efd4c42c09af/Xamarin.Forms.Platform.UAP/Forms.cs#L39

I think those static methods are more if you're in a multi window/dispatcher environment, since Forms is a one window one dispatcher kind of framework those static methods don't really serve any purpose. 
 
If  just using *_dispatcher* didn't work and we needed these static checks then I feel like BeginInvokeOnMainThread would need to also be rewritten to use those static methods in order to get back the "correct" dispatcher, but right now it just feels like a lot of effort is being made to essentially get access to the same dispatcher that's already stored locally.

hartez:
@PureWeen We also need to reconcile this with the changes in #2432.

PureWeen:
Yea that one sort of invalidates just using _dispatcher :-/  Depending on the timing of https://github.com/xamarin/Xamarin.Forms/pull/2432 might need to just move the discussion over there so that it also accounts for the exception here

I had a variation that I was messing with that went something like

```C#
if((LockApplicationHost.GetForCurrentView() != null)
return !CoreApplication.GetCurrentView().Dispatcher.HasThreadAccess

if(CoreApplication.GetCurrentView() != CoreApplication.MainView.CoreWindow )
return !CoreApplication.GetCurrentView().Dispatcher.HasThreadAccess

return CoreApplication.MainView.CoreWindow.Dispatcher.HasThreadAccess
```

But the problem I hit with this code was that when running the UI TEST with animations
```
CoreApplication.GetCurrentView()
```
Was throwing an exception like
*NRE Element not found exception*

:-/ 

The thing I'm confused about with https://github.com/xamarin/Xamarin.Forms/issues/1790 is how when UWP acts as a share target it seems to behave like assigned access but the *LockApplicationHost.GetForCurrentView()* is null so then *CoreApplication.MainView.CoreWindow* just throws the exception you get when accessing CoreWindow during assigned access :-/ 

Though it seems like in 2432
https://github.com/xamarin/Xamarin.Forms/pull/2432/files?utf8=%E2%9C%93&diff=split#diff-8c5622381de6ca1686f98385f2a68a86R137

It just forces an Invoke if Forms.Dispatchers.Count > 1 so in theory I think that whole middle part there could just be replaced by 
```
Forms.Dispatchers[0].HasThreadAccess
```

PureWeen:
@hartez 
https://github.com/xamarin/Xamarin.Forms/pull/2432/files?utf8=%E2%9C%93&diff=unified#r185910637
They are looking at just making the same change in that PR so might just let that one trump this one

PureWeen:
This bug is really annoying. I'd say we just merge this one in with
```C#
public bool IsInvokeRequired => !_dispatcher.HasThreadAccess
```

And then if https://github.com/xamarin/Xamarin.Forms/pull/2432 ends up replacing it in the same release so be it but it'd be nice to just have this here and merged in sooner rather than later

hartez:
Don't merge it yet, I'm looking at it right now. I'm like 90% sure this fix is not correct.

paymicro:
> @paymicro can you rebase so we fix the OSX build? thanks

@rmarinho sure. Done.

hartez:
> I setup Control Gallery so I could set it up with Assigned Access. Remote deployed it to my laptop. Set it up in kiosk mode for a local user and everything worked fine. 

I was wondering how this was possible, but it makes sense now that I think about it more:

`Device.PlatformServices = new WindowsPlatformServices(Window.Current.Dispatcher);`

In assigned access mode running above the lock screen, `Window.Current.Dispatcher` is already the "view above the lock screen" dispatcher. 

This looks great as-is, and should pretty much be compatible with the multi-window stuff in the other PR.

cschwarz:
Any chance to get that fix into 3.0 or 3.1?

PR #2539, opened 2018-04-24
Controlling Caret position on iOS,Android,UWP and color on iOS

### LINKED ISSUE #1667: [Enhancement] Entry: Position and color of caret
## Rationale

Entry allows text entry into a string. Makes sense to know and set where the text will be entered into the string. Changing the visual indicator of that position is also a nice to have. 

## Implementation

```csharp
public class Entry {
    public static readonly BindableProperty CursorPosition; // int
    public static readonly BindableProperty SelectionLength; // int
    public static readonly BindableProperty CursorColor; // Color
}
```

## Expected Result

`CursorPosition` returns/sets the position at which the next character will be inserted while `CursorColor` sets the Color of the cursor. `SelectionLength` returns the length of text selection within the entry and will be zero if no text is selected. 

## Implications for CSS

None.

## Backward Compatibility

No backwards compatibility concerns. Reasonable defaults exist. 

## Difficulty : Easy

henricm:
### Description of Change ###

This PR adds support for getting and setting
caret position on a text Entry on iOS, Android and UWP.

Also adding support for setting CursorColor but only for iOS since
it seems unsupported on UWP and couldn't figure out how to do it
on Android in an not too complicated manner.

### Bugs Fixed ###

- Fixes #1667 except for cursor color functionality on UWP and Android.

### API Changes ###

Added:
 - `int Entry.CursorPosition { get; set; } //Bindable Property`
 - `int Entry.SelectionLength { get; set; } // Bindable Property`
 - `Color Entry<iOS>.CursorColor { get; set; } // iOS specific Bindable Property`

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1667.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1667.cs
@@ -28,6 +28,14 @@ namespace Xamarin.Forms.Controls.Issues
 			_updateButton = new Button { Text = "Update" };
 			_updateButton.Clicked += UpdateCursor;
 
+			var red = new Button { Text = "Red", TextColor = Color.Red};
+			red.Clicked += (sender, e) => _entry.CursorColor = Color.Red;
+
+			var blue = new Button { Text = "Blue", TextColor = Color.Blue};
+			blue.Clicked += (sender, e) => _entry.CursorColor = Color.Blue;
+
+			var defaultColor = new Button { Text = "Default" };
+			defaultColor.Clicked += (sender, e) => _entry.CursorColor = Color.Default;
 			Content = new StackLayout
 			{
 				Margin = new Thickness(10, 40),
@@ -38,7 +46,10 @@ namespace Xamarin.Forms.Controls.Issues
 					_cursorStartPosition,
 					new Label {Text = "Selection Length:"},
 					_selectionLength,
-					_updateButton
+					_updateButton,
+					red,
+					blue,
+					defaultColor
 				}
 			};
 		}
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1667.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1667.cs
@@ -5,6 +5,11 @@ using System.Linq;
 using Xamarin.Forms.CustomAttributes;
 using Xamarin.Forms.Internals;
 
+#if UITEST
+using Xamarin.Forms.Core.UITests;
+using NUnit.Framework;
+#endif
+
 namespace Xamarin.Forms.Controls.Issues
 {
 	[Preserve (AllMembers = true)]
@@ -19,7 +24,7 @@ namespace Xamarin.Forms.Controls.Issues
 
 		protected override void Init ()
 		{
-			_entry = new Entry {Text = "Enter cursor position below"};
+			_entry = new Entry {Text = "Enter cursor position below", AutomationId = "TextField"};
 			_entry.PropertyChanged += ReadCursor;
 
 			_cursorStartPosition = new Entry();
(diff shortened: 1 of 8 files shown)


samhouts:
Maybe we should convert the CursorColor to an iOS platform specific, since we're not implementing it on other platforms...?

Xamarin.Forms.Platform.Android/Renderers/EntryRenderer.cs
```diff
@@ -279,5 +285,35 @@ void UpdateReturnType()
 			Control.ImeOptions = Element.ReturnType.ToAndroidImeAction();
 			_currentInputImeFlag = Control.ImeOptions;
 		}
+
+		void SelectionChanged(object sender, SelectionChangedEventArgs e)
+		{
+			var control = Control;
```
samhouts: Not sure why we're making a variable for `Control`.

Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
```diff
@@ -302,5 +310,64 @@ void UpdateReturnType()
 			Control.ReturnKeyType = Element.ReturnType.ToUIReturnKeyType();
 		}
 
+		bool _selectedTextRangeIsUpdating;
```
samhouts: Maybe move this with the other fields.

Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
```diff
@@ -302,5 +310,64 @@ void UpdateReturnType()
...
+
+		void UpdateCursorFromControl(NSObservedChange obj)
+		{
+			if (_selectedTextRangeIsUpdating)
```
samhouts: Do we also want to do some null check on `Control`/`Element`?

samhouts:
build --uitests

Xamarin.Forms.Platform.Android/Renderers/EntryRenderer.cs
henricm: My understanding was that the getter for Control makes non-trivial calls to get the native component (however I cannot see that looking at the ViewRenderer code :-)). So if we need to access it more than once it's a good idea for performance to save it in a variable for later use. If this is not how we wish to do it or if I misunderstood just let me know and I'll change it back.

Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
henricm: Not sure if they can be null here, but I'm not sure...

henricm: Good question. I could add them just in case.

henricm:
@samhouts I could move CursorColor to be iOS specific. Just let me know what you prefer.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1667.cs
```diff
@@ -0,0 +1,77 @@
...
+				_entry.SelectionLength = length;
+			}
+		}
+
+		void ReadCursor(object sender, EventArgs args)
+		{
+			_cursorStartPosition.Text = _entry.CursorPosition.ToString();
+			_selectionLength.Text = _entry.SelectionLength.ToString();
+		}
+	}
```
rmarinho: can we add a UITest so we make sure the feature Works / doesn't break.

Xamarin.Forms.Platform.Android/Renderers/EntryRenderer.cs
rmarinho: i would add just a null check to the `control` , and use `ElementController?.SetValueFromRenderer`

Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
rmarinho: yeah i would add it as well

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1667.cs
henricm: Two basic UI tests added.

myroot:
Why don't you support these feature on Editor?

PR #2544, opened 2018-04-25
Fix .css files not excluded from default None items

mrward:
### Description of Change ###

Fixes duplicate CSS files in solution pad in VS Mac.

### Bugs Fixed ###

- Fixes VSTS #606364 - CSS files are duplicated in solution pad

### API Changes ###

None

### PR Checklist ###


jassmith:
Yeah so this absolutely broke people as it tries to reference these dlls in builds.

PR #2556, opened 2018-04-27
Namescope enhancements

StephaneDelcroix:
### Description of Change ###

 - turn Element.FindByName() public
 - reduce the amount of time (both Xaml and XamlC) we need to actually set the NameScope on an object
 - add a IReferenceProvider
   - allows custom markups replicating x:Reference
   - more efficient x:Reference
  

### Bugs Fixed ###

/

### API Changes ###

Added
```
public object Element.FindByName(string name);

public interface Xamarin.Forms.Xaml.IReferenceProvider
{
    object FindByName(string name);
}
```

### Behavioral Changes ###

/

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/SetNamescopesAndRegisterNamesVisitor.cs
+++ b/Xamarin.Forms.Build.Tasks/SetNamescopesAndRegisterNamesVisitor.cs
@@ -48,16 +48,18 @@ namespace Xamarin.Forms.Build.Tasks
 		{
 			VariableDefinition namescopeVarDef;
 			IList<string> namesInNamescope;
+			var setNameScope = false;
 			if (parentNode == null || IsDataTemplate(node, parentNode) || IsStyle(node, parentNode) || IsVisualStateGroupList(node)) {
 				namescopeVarDef = CreateNamescope();
 				namesInNamescope = new List<string>();
+				setNameScope = true;
 			} else {
 				namescopeVarDef = Context.Scopes[parentNode].Item1;
 				namesInNamescope = Context.Scopes[parentNode].Item2;
 			}
-			if (Context.Variables[node].VariableType.InheritsFromOrImplements(Context.Body.Method.Module.ImportReference(("Xamarin.Forms.Core","Xamarin.Forms","BindableObject"))))
+			if (setNameScope && Context.Variables[node].VariableType.InheritsFromOrImplements(Context.Body.Method.Module.ImportReference(("Xamarin.Forms.Core","Xamarin.Forms","BindableObject"))))
 				SetNameScope(node, namescopeVarDef);
-			Context.Scopes[node] = new System.Tuple<VariableDefinition, IList<string>>(namescopeVarDef, namesInNamescope);
+			Context.Scopes[node] = new Tuple<VariableDefinition, IList<string>>(namescopeVarDef, namesInNamescope);
 		}
 	
 		public void Visit(RootNode node, INode parentNode)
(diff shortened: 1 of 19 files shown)


Xamarin.Forms.Build.Tasks/SetNamescopesAndRegisterNamesVisitor.cs
```diff
@@ -48,16 +48,18 @@ public void Visit(ElementNode node, INode parentNode)
...
 			if (parentNode == null || IsDataTemplate(node, parentNode) || IsStyle(node, parentNode) || IsVisualStateGroupList(node)) {
 				namescopeVarDef = CreateNamescope();
 				namesInNamescope = new List<string>();
+				setNameScope = true;
 			} else {
 				namescopeVarDef = Context.Scopes[parentNode].Item1;
 				namesInNamescope = Context.Scopes[parentNode].Item2;
 			}
-			if (Context.Variables[node].VariableType.InheritsFromOrImplements(Context.Body.Method.Module.ImportReference(("Xamarin.Forms.Core","Xamarin.Forms","BindableObject"))))
+			if (setNameScope && Context.Variables[node].VariableType.InheritsFromOrImplements(Context.Body.Method.Module.ImportReference(("Xamarin.Forms.Core","Xamarin.Forms","BindableObject"))))
```
StephaneDelcroix: only set the name scope if it's new (and different from parent)

Xamarin.Forms.Core/Element.cs
```diff
@@ -283,39 +283,22 @@ public bool EffectIsAttached(string name)
...
 		{
-			INameScope namescope = GetNameScope();
+			var namescope = GetNameScope();
 			if (namescope == null)
 				throw new InvalidOperationException("this element is not in a namescope");
 			namescope.RegisterName(name, scopedElement);
 		}
 
-		[Obsolete]
-		void INameScope.RegisterName(string name, object scopedElement, IXmlLineInfo xmlLineInfo)
```
StephaneDelcroix: removing this will prevent code compiled by XamlC 2 years ago to keep working. But the chances for that code to still work are close to nonexistent...

Xamarin.Forms.Core/Element.cs
```diff
@@ -283,39 +283,22 @@ public bool EffectIsAttached(string name)
...
-		{
-			INameScope namescope = GetNameScope();
-			if (namescope == null)
-				throw new InvalidOperationException("this element is not in a namescope");
-			namescope.RegisterName(name, scopedElement, xmlLineInfo);
-		}
-
-		void INameScope.UnregisterName(string name)
```
StephaneDelcroix: that was never used

Xamarin.Forms.Core/Element.cs
```diff
@@ -623,37 +606,32 @@ void EffectsOnCollectionChanged(object sender, NotifyCollectionChangedEventArgs
 			}
 		}
 
-		INameScope GetNameScope()
+		internal INameScope GetNameScope()
```
StephaneDelcroix: for testing

BradChase2011:
@StephaneDelcroix I finally got around to rebasing our build of forms to the latest code and noticed we were crashing everywhere we used templates with x:Reference and where we use FindByName.  Basically it comes down to any template we late load that uses x:Reference will crash out.  I reverted this checkin in our code and everything runs great on the latest master.

So my question is, has anyone else experienced issues with this check in or reported any issues yet?

StephaneDelcroix:
@BradChase2011 could you please open an issue with a reproduction ?

BradChase2011:
@StephaneDelcroix I want to take some time and try to find the culprit before doing so, I just wanted to check if anything was known before I went down that path.  Once I figure that out, I definitely will.

Issue #2552, opened 2018-04-27
"Xamarin.Forms" version="3.0.0.427558-pre4" fails to build for Android API 25 target


richardlawler:
### Description
Attempting to use pre4 build of Xamarin.Forms package gives error when attempting to build Android (API25) target. 
System.IO.FileNotFoundException: Could not load assembly 'Microsoft.Windows.Design.Extensibility, Version=4.3.1.0

### Steps to Reproduce

1. Create a new Blank Xamarin Forms project
2. Update the Xamarin.Forms package for the common, Droid and iOS projects to the pre4 package
3. Try to build the Droid target.

### Expected Behavior

App should build and deploy to device

### Actual Behavior

Error from build.

/Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(2,2): Error: Exception while loading assemblies: System.IO.FileNotFoundException: Could not load assembly 'Microsoft.Windows.Design.Extensibility, Version=4.3.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. Perhaps it doesn't exist in the Mono for Android profile?
File name: 'Microsoft.Windows.Design.Extensibility.dll'
  at Java.Interop.Tools.Cecil.DirectoryAssemblyResolver.Resolve (Mono.Cecil.AssemblyNameReference reference, Mono.Cecil.ReaderParameters parameters) [0x00099] in /Users/builder/data/lanes/5809/a7829590/source/monodroid/external/xamarin-android/external/Java.Interop/src/Java.Interop.Tools.Cecil/Java.Interop.Tools.Cecil/DirectoryAssemblyResolver.cs:229 
  at Java.Interop.Tools.Cecil.DirectoryAssemblyResolver.Resolve (Mono.Cecil.AssemblyNameReference reference) [0x00000] in /Users/builder/data/lanes/5809/a7829590/source/monodroid/external/xamarin-android/external/Java.Interop/src/Java.Interop.Tools.Cecil/Java.Interop.Tools.Cecil/DirectoryAssemblyResolver.cs:179 
[...]

PR #2575, opened 2018-05-01
Use explicit references so VS for Mac doesn't pick up Design files

hartez:
### Description of Change ###

We don't want the files in the Design folder to be added as assembly references when installing the nuget package; to avoid this, we make the list of assembly references explicit.

### Bugs Fixed ###

- #2552

### DIFF
--- a/.nuspec/Xamarin.Forms.nuspec
+++ b/.nuspec/Xamarin.Forms.nuspec
@@ -42,9 +42,49 @@
         <dependency id="Tizen.NET" version="4.0.0"/>
       </group>
     </dependencies>
+	
+	<references>	
+		<group>	
+			<reference file="Xamarin.Forms.Core.dll" />	
+			<reference file="Xamarin.Forms.Platform.dll" />	
+			<reference file="Xamarin.Forms.Xaml.dll" />	
+		</group>
+		<group targetFramework="Xamarin.iOS10">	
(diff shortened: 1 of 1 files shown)


.nuspec/Xamarin.Forms.nuspec
```diff
@@ -105,6 +145,17 @@
     <file src="..\Xamarin.Forms.Core.Design\bin\$Configuration$\Xamarin.Forms.Core.Design.dll" target="lib\netstandard1.0\Design" />
     <file src="..\Xamarin.Forms.Xaml.Design\bin\$Configuration$\Xamarin.Forms.Xaml.Design.dll" target="lib\netstandard1.0\Design" />
 
+	<file src="..\Xamarin.Forms.Core.Design\bin\$Configuration$\Xamarin.Forms.Core.Design.dll" target="lib\MonoAndroid10\Design" />
```
bholmes: white-space error.

.nuspec/Xamarin.Forms.nuspec
```diff
@@ -42,9 +42,49 @@
         <dependency id="Tizen.NET" version="4.0.0"/>
       </group>
     </dependencies>
+	
+	<references>
```
bholmes: Mixing tabs and spaces with the rest of the file.

hartez: Fixed.

PR #2589, opened 2018-05-02
Expose the Source for a StyleSheet

kzu:
This would allow previewer/designer/reloader to determine the source of a
stylesheet and refresh its content when appropriate.

### DIFF
--- a/Xamarin.Forms.Core/StyleSheets/StyleSheet.cs
+++ b/Xamarin.Forms.Core/StyleSheets/StyleSheet.cs
@@ -15,6 +15,8 @@ namespace Xamarin.Forms.StyleSheets
 		{
 		}
 
+		public Uri Source { get; internal set; }
+
 		internal IDictionary<Selector, Style> Styles { get; set; } = new Dictionary<Selector, Style>();
 
 		public static StyleSheet FromAssemblyResource(Assembly assembly, string resourceId, IXmlLineInfo lineInfo = null)
--- a/Xamarin.Forms.Xaml.UnitTests/StyleSheet.xaml.cs
+++ b/Xamarin.Forms.Xaml.UnitTests/StyleSheet.xaml.cs
@@ -1,8 +1,9 @@
 ﻿using System;
+using System.IO;
+using System.Linq;
+using System.Reflection;
 using NUnit.Framework;
-
 using Xamarin.Forms.Core.UnitTests;
-using NUnit.Framework.Constraints;
 
 namespace Xamarin.Forms.Xaml.UnitTests
 {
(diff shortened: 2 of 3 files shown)


kzu:
@StephaneDelcroix yeah, I wonder why XamlC won't just invoke the instantiated `StyleSheetExtension.ProvideValue`, but ¯\_(ツ)_/¯. In any case, when a custom `ResourceLoader` is present (such as in my case and the previewer), the code-path is different, it goes through `LoadFromXaml` and that in turn goes through the markup extension and calls its [IValueProvider.ProvideValue](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Xaml/ApplyPropertiesVisitor.cs#L237), so it's properly set for me. The day XamlC switches to invoking the value provider like everything else, things will "just work". I can file a bug for that if you want me to.

I didn't add a unit test for the `StyleSheetExtension` because I couldn't find any for it:

image

StephaneDelcroix:
@kzu 
> I wonder why XamlC won't just invoke the instantiated `StyleSheetExtension.ProvideValue`

it usually does, on custom ValueProvider. but this one is decorated with `[ProvideCompiled]` meaning the `ProvideValue` is short-circuited and everything is resolved at compile time. see https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Build.Tasks/CompiledValueProviders/StyleSheetProvider.cs

A good place to test would be in 
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Xaml.UnitTests/StyleSheet.xaml
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Xaml.UnitTests/StyleSheet.xaml.cs

kzu:
There you go. Also, proof that it works with XamlC for the scenario I need.

Also, @jassmith @davidortinau if this doesn't ship in 3.0.0, I'd say it's pointless since I'm not going to take a dependency for Live Reload on an unreleased version of XF that will come who knows when. So this property will be unused until such time when vNext ships (although, for such an optimization, I doubt it will be worth forcing people to 3.0+1 even then... dunno).

jassmith:
This wont make 3.0.0, closing as requested

kzu:
Merge it to master then, some day it may be useful ¯\_(ツ)_/¯

kzu:
It's the only code path we care about. Feel free to modify the test as you see fit.

StephaneDelcroix:
it's a public API, it has to work in the cases you do not care about too... or we can turn that into something internal...

kzu:
And what would those scenarios be @StephaneDelcroix? Because there isn't any other besides the one mentioned in the original description.

Feel free to add more, but blocking this for supposed uses it's not intended to solve (or don't exist, even), doesn't make much sense to me.

StephaneDelcroix:
@kzu serializing a visual tree back to xaml ?

kzu:
Who's asking for that? Is that a true feature request for something you want to support in the product? I can't see how it's related to adding this property anyway. Can't you just traverse the visual tree already with current APIs and turn that into a XAML file? Is there even an API we expose to do that or would someone have to do that manually? 

I also don't know enough of how the XamlC does its codegen to fix whatever is broken, so it would be great if you could take a look and make it work the way you prefer.

PR #2615, opened 2018-05-07
Implement OnPlatform/OnIdiom markup extensions

### LINKED ISSUE #2608: OnPlatform/OnIdiom markup extensions
### Description

Sometimes changing a single property value per platform is quite convenient, yet the "full" syntax of `OnPlatform` is overkill for scalar values in attributes. 

For example:

```xml
<Label Text="Hello World">
    <Label.FontSize>
        <OnPlatform x:TypeArguments="x:Double">
            <On Platform="iOS" Value="20" />
            <On Platform="Android" Value="24" />
        </OnPlatform>
    </Label.FontSize>
</Label>
```

vs a markup extension version that could provide:

```xml
<Label Text="Hello World" FontSize="{OnPlatform Android=24, iOS=20}" />
```

Same thing applies for `OnIdiom`, so I think both would make XAML much more concise.

kzu:
Fixes #2608.

image

image

### DIFF
--- a/Xamarin.Forms.Xaml/MarkupExtensionParser.cs
+++ b/Xamarin.Forms.Xaml/MarkupExtensionParser.cs
@@ -20,6 +20,8 @@ namespace Xamarin.Forms.Xaml
 				markupExtension = new StaticResourceExtension();
 			else if (match == "OnPlatform")
 				markupExtension = new OnPlatformExtension();
+			else if (match == "OnIdiom")
+				markupExtension = new OnIdiomExtension();
 			else
 			{
 				if (typeResolver == null)
--- a/Xamarin.Forms.Xaml/MarkupExtensionParser.cs
+++ b/Xamarin.Forms.Xaml/MarkupExtensionParser.cs
@@ -18,6 +18,8 @@ namespace Xamarin.Forms.Xaml
 				markupExtension = new TemplateBindingExtension();
 			else if (match == "StaticResource")
 				markupExtension = new StaticResourceExtension();
+			else if (match == "OnPlatform")
+				markupExtension = new OnPlatformExtension();
 			else
 			{
 				if (typeResolver == null)
@@ -27,7 +29,7 @@ namespace Xamarin.Forms.Xaml
 				//The order of lookup is to look for the Extension-suffixed class name first and then look for the class name without the Extension suffix.
 				if (!typeResolver.TryResolve(match + "Extension", out type) && !typeResolver.TryResolve(match, out type))
 				{
-					var lineInfoProvider = serviceProvider.GetService(typeof (IXmlLineInfoProvider)) as IXmlLineInfoProvider;
+					var lineInfoProvider = serviceProvider.GetService(typeof(IXmlLineInfoProvider)) as IXmlLineInfoProvider;
 					var lineInfo = (lineInfoProvider != null) ? lineInfoProvider.XmlLineInfo : new XmlLineInfo();
 					throw new XamlParseException(String.Format("MarkupExtension not found for {0}", match), lineInfo);
 				}
--- a/Xamarin.Forms.Xaml/MarkupExtensions/OnIdiomExtension.cs
+++ b/Xamarin.Forms.Xaml/MarkupExtensions/OnIdiomExtension.cs
@@ -1,0 +1,46 @@
+﻿using System;
+
+namespace Xamarin.Forms.Xaml
+{
+	public class OnIdiomExtension : IMarkupExtension
+	{
+		// See Device.Idiom
+
+		public object Default { get; set; }
+		public object Unsupported { get; set; }
+		public object Phone { get; set; }
+		public object Tablet { get; set; } 
+		public object Desktop { get; set; }
+		public object TV { get; set; }
+		public object Watch { get; set; }
+
+		public object ProvideValue(IServiceProvider serviceProvider)
+		{
+			if (Default == null && Unsupported == null && Phone == null && 
+				Tablet == null && Desktop == null && TV == null && Watch == null)
+			{
+				var lineInfo = (serviceProvider.GetService(typeof(IXmlLineInfoProvider)) as IXmlLineInfoProvider)?.XmlLineInfo 
+					?? new XmlLineInfo();
+				throw new XamlParseException("OnIdiomExtension requires a value to be specified for at least one idiom or Default.", lineInfo);
+			}
+
+			switch (Device.Idiom)
+			{
+				case TargetIdiom.Unsupported:
+					return Unsupported ?? Default;
+				case TargetIdiom.Phone:
+					return Phone ?? Default;
+				case TargetIdiom.Tablet:
+					return Tablet ?? Default;
+				case TargetIdiom.Desktop:
+					return Desktop ?? Default;
+				case TargetIdiom.TV:
+					return TV ?? Default;
+				case TargetIdiom.Watch:
+					return Watch ?? Default;
+				default:
+					return Default;
+			}
+		}
+	}
+}
--- a/Xamarin.Forms.Xaml/MarkupExtensions/OnPlatformExtension.cs
+++ b/Xamarin.Forms.Xaml/MarkupExtensions/OnPlatformExtension.cs
@@ -5,9 +5,13 @@ namespace Xamarin.Forms.Xaml
 	public class OnPlatformExtension : IMarkupExtension
 	{
 		public object Default { get; set; }
+		public object Android { get; set; }
+		public object GTK { get; set; }
 		public object iOS { get; set; }
-		public object Android { get; set; } 
+		public object macOS { get; set; }
+		public object Tizen { get; set; }
 		public object UWP { get; set; }
+		public object WPF { get; set; }
 		public string Other { get; set; }
 
 
@@ -22,12 +26,20 @@ namespace Xamarin.Forms.Xaml
 
 			switch (Device.RuntimePlatform)
 			{
-				case Device.iOS:
-					return iOS;
 				case Device.Android:
-					return Android;
+					return Android ?? Default;
+				case Device.GTK:
+					return GTK ?? Default;
+				case Device.iOS:
+					return iOS ?? Default;
+				case Device.macOS:
+					return macOS ?? Default;
+				case Device.Tizen:
+					return Tizen ?? Default;
 				case Device.UWP:
-					return UWP;
+					return UWP ?? Default;
+				case Device.WPF:
+					return WPF ?? Default;
 				default:
 					if (string.IsNullOrEmpty(Other))
 						return Default;
--- a/Xamarin.Forms.Xaml/MarkupExtensions/OnPlatformExtension.cs
+++ b/Xamarin.Forms.Xaml/MarkupExtensions/OnPlatformExtension.cs
@@ -1,21 +1,23 @@
 ﻿using System;
-using System.Xml;
 
 namespace Xamarin.Forms.Xaml
 {
 	public class OnPlatformExtension : IMarkupExtension
 	{
+		public object Default { get; set; }
 		public object iOS { get; set; }
 		public object Android { get; set; } 
 		public object UWP { get; set; }
+		public string Other { get; set; }
+
 
 		public object ProvideValue(IServiceProvider serviceProvider)
 		{
-			if (iOS == null && Android == null && UWP == null)
+			if (iOS == null && Android == null && UWP == null && Default == null && string.IsNullOrEmpty(Other))
 			{
-				IXmlLineInfoProvider lineInfoProvider = serviceProvider.GetService(typeof(IXmlLineInfoProvider)) as IXmlLineInfoProvider;
-				IXmlLineInfo lineInfo = (lineInfoProvider != null) ? lineInfoProvider.XmlLineInfo : new XmlLineInfo();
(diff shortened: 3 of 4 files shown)


kzu:
Gotta love having WiFi on the airplane on my way to //build :)

@jassmith all yours!

dansiegel:
`Default` should be added as a `ContentProperty` so that you can have:

```xml
<Label FontSize="{OnPlatform 10, Android=12, UWP=11}" />
```

verses the more verbose:

```xml
<Label FontSize="{OnPlatform Default=10, Android=12, UWP=11}" />
```

Xamarin.Forms.Xaml/MarkupExtensionParser.cs
```diff
@@ -18,6 +18,10 @@ public object Parse(string match, ref string remaining, IServiceProvider service
 				markupExtension = new TemplateBindingExtension();
 			else if (match == "StaticResource")
 				markupExtension = new StaticResourceExtension();
+			else if (match == "OnPlatform")
+				markupExtension = new OnPlatformExtension();
+			else if (match == "OnIdiom")
+				markupExtension = new OnIdiomExtension();
```
StephaneDelcroix: this isn't required, but it allows bypassing constructing the markup through reflection

Xamarin.Forms.Xaml/MarkupExtensions/OnIdiomExtension.cs
```diff
@@ -0,0 +1,46 @@
...
+
+		public object Default { get; set; }
+		public object Unsupported { get; set; }
+		public object Phone { get; set; }
+		public object Tablet { get; set; } 
+		public object Desktop { get; set; }
+		public object TV { get; set; }
+		public object Watch { get; set; }
+
+		public object ProvideValue(IServiceProvider serviceProvider)
```
StephaneDelcroix: the Xaml engine expects from `ProvideValue` a value of the correct type that can be directly assigned to the property. In this case, all the objects are of type string, and not directly assignable to `FontSize`, `TextColor`, `WidthRequest`...

the magic of the runtime Xaml loader compensates for the wrong type and do the type conversion, but
 - I can't guarantee that to work in all cases
 - if fails with XamlC on

Xamarin.Forms.Xaml/MarkupExtensions/OnIdiomExtension.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+
+namespace Xamarin.Forms.Xaml
+{
+	public class OnIdiomExtension : IMarkupExtension
+	{
+		// See Device.Idiom
+
+		public object Default { get; set; }
```
StephaneDelcroix: could just be string in your case...

as the properties aren't typed, you won't get intelissense completion unless special-cased

ChaseFlorell:
@ahoefling that 'other' property is what I was referring to in my previous comment. This is a great addition to Forms :+1:

Xamarin.Forms.Xaml/MarkupExtensions/OnIdiomExtension.cs
kzu: Addressed in latest commits. Thanks!

Xamarin.Forms.Xaml/MarkupExtensions/OnIdiomExtension.cs
kzu: By making these `object`, we can allow further markup extensions to set those values (i.e. `StaticResource`?)

dansiegel:
> it's indeed more concise, unfortunately it doesn't work with XamlC on

This brings up an interesting point. @kzu I imagine there are going to be cases then that we'll need to provide a Converter where the default converters just won't work. Probably will want to add a Converter property as well to handle those cases.

kzu:
If the property is already annotated with a converter, that's already supported ;)

But I guess I could add that too...

kzu:
Ok, added `Converter/ConverterParameter` as seen elsewhere in the codebase. Not sure if the "right" culture to use for the conversion should be `CultureInfo.CurrentUICulture` (as in [Binding](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/Binding.cs#L142)), `CurrentCulture` as I've used, or `InvariantCulture` as the [TypeConversionExtensions](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/Xaml/TypeConversionExtensions.cs#L134) does. 

Thoughts?

Xamarin.Forms.Xaml/MarkupExtensions/OnIdiomExtension.cs
ahoefling: @kzu when I wrote the original change the idea was to make everything an object so we could use this on more than just strings. Since this is a new platform markup extension we should support more than just strings with it. This will open up a lot of flexibility for developers to have let's say different images or objects inserted into their XAML based on the platform or idiom which is a powerful thing.

Xamarin.Forms.Xaml/MarkupExtensions/OnIdiomExtension.cs
```diff
@@ -0,0 +1,78 @@
...
+
+		public object ConverterParameter { get; set; }
+
+		public object ProvideValue(IServiceProvider serviceProvider)
+		{
+			var lineInfo = (serviceProvider.GetService(typeof(IXmlLineInfoProvider)) as IXmlLineInfoProvider)?.XmlLineInfo;
+			if (Default == null && Unsupported == null && Phone == null &&
+				Tablet == null && Desktop == null && TV == null && Watch == null)
+			{
+				throw new XamlParseException("OnIdiomExtension requires a value to be specified for at least one idiom or Default.", lineInfo ?? new XmlLineInfo());
```
StephaneDelcroix: so, doing `{OnIdiom Phone={x:Null}}` will throws even though a value is actually specified ?

I'd change the message requesting one of the value to be non-null

Xamarin.Forms.Xaml/MarkupExtensions/OnIdiomExtension.cs
```diff
@@ -0,0 +1,78 @@
...
+		public object TV { get; set; }
+		public object Watch { get; set; }
+
+		public IValueConverter Converter { get; set; }
+
+		public object ConverterParameter { get; set; }
+
+		public object ProvideValue(IServiceProvider serviceProvider)
+		{
+			var lineInfo = (serviceProvider.GetService(typeof(IXmlLineInfoProvider)) as IXmlLineInfoProvider)?.XmlLineInfo;
```
StephaneDelcroix: I recently added a generic `GetService<T>` you might want to use.

also, do `serviceProvider?.GetService...`

Xamarin.Forms.Xaml/MarkupExtensions/OnIdiomExtension.cs
```diff
@@ -0,0 +1,78 @@
...
+			if (Default == null && Unsupported == null && Phone == null &&
+				Tablet == null && Desktop == null && TV == null && Watch == null)
+			{
+				throw new XamlParseException("OnIdiomExtension requires a value to be specified for at least one idiom or Default.", lineInfo ?? new XmlLineInfo());
+			}
+
+			var valueProvider = serviceProvider.GetService<IProvideValueTarget>() ?? throw new ArgumentException();
```
StephaneDelcroix: `serviceProvider?.`

Xamarin.Forms.Xaml/MarkupExtensions/OnIdiomExtension.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Xamarin.Forms.Xaml
+{
+	[ContentProperty("Default")]
+	public class OnIdiomExtension : IMarkupExtension
+	{
+		// See Device.Idiom
+
+		public object Default { get; set; }
+		public object Unsupported { get; set; }
```
StephaneDelcroix: I'd remove the `Unsupported` property

Xamarin.Forms.Xaml/MarkupExtensions/OnPlatformExtension.cs
```diff
@@ -0,0 +1,96 @@
...
+	{
+		public object Default { get; set; }
+		public object Android { get; set; }
+		public object GTK { get; set; }
+		public object iOS { get; set; }
+		public object macOS { get; set; }
+		public object Tizen { get; set; }
+		public object UWP { get; set; }
+		public object WPF { get; set; }
+		public string Other { get; set; }
```
StephaneDelcroix: Other should't be of type string, as it limits the values to be literals. Unless you plan to parse the potential markups...

I'd remove the `Other` for now, so we can think about a better solution for it (I could, e.g., modify the markup extension parser to add values in a directory instead of assigning properties)

ChaseFlorell: I thought `Other` being of type string was to allow string parsing like this.

```<Button WidthRequest="{OnPlatform Other='Foo:21,Bar25'}" />```

File: Xamarin.Forms.Xaml/MarkupExtensions/OnIdiomExtension.cs (hunk 4 above)
Comment (kzu) [reply Xamarin.Forms.Xaml/MarkupExtensions/OnIdiomExtension.cs]: Makes sense!

File: Xamarin.Forms.Xaml/MarkupExtensions/OnIdiomExtension.cs (hunk 6 above)
Comment (kzu) [reply Xamarin.Forms.Xaml/MarkupExtensions/OnIdiomExtension.cs]: Can the service provider itself be null?

File: Xamarin.Forms.Xaml/MarkupExtensions/OnPlatformExtension.cs (hunk 8 above)
Comment (kzu) [reply Xamarin.Forms.Xaml/MarkupExtensions/OnPlatformExtension.cs]: Yep, indeed. So, @StephaneDelcroix removing it means we don't support custom platforms, but I guess that's fine too for a "v1"?

Comment (ChaseFlorell) [reply Xamarin.Forms.Xaml/MarkupExtensions/OnPlatformExtension.cs]: Alternatively, the extension can be inherited by the user with their custom platform. You abstract away all the shared bits and then implement a new virtual method `HandlePlatform`

```
// pseudo code
public class MyOnPlatformExtension : OnPlatformExtension
{
    public object MyFooPlatform { get; set; }
    public override object HandlePlatform(object value)
    {
        if(MyFooPlatform != null && Device.RuntimePlatform == nameof(MFooPlatform))
        {
            return MyFooPlatform;
        } else {
            return base.HandlePlatform(value);
        }
    }
}
```

Comment (kzu) [reply Xamarin.Forms.Xaml/MarkupExtensions/OnPlatformExtension.cs]: Why wouldn't the `Default` be enough if all the other platforms have their own values? Effectively, the `Default` would become the value for their custom platform...

**Comment by kzu** [comment]
Ok @StephaneDelcroix @jassmith I think this is ready now :)

File: Xamarin.Forms.Xaml/MarkupExtensions/OnPlatformExtension.cs (hunk 8 above)
Comment (ChaseFlorell) [reply Xamarin.Forms.Xaml/MarkupExtensions/OnPlatformExtension.cs]: excepting for the fact that they could potentially have more than one custom platform. Isn't that why Forms moved to stringly typed anyways?

At any rate, if you move the switch logic into a separate virtual method, we'd be able to override easily using my example above.

File: Xamarin.Forms.Xaml/MarkupExtensions/OnPlatformExtension.cs [hunk 9]
```diff
@@ -0,0 +1,81 @@
...
```
StephaneDelcroix: you removed `Other` from the switch, but left the property

Xamarin.Forms.Xaml/MarkupExtensions/OnPlatformExtension.cs
```diff
@@ -0,0 +1,80 @@
...
+		public IValueConverter Converter { get; set; }
+
+		public object ConverterParameter { get; set; }
+
+		public object ProvideValue(IServiceProvider serviceProvider)
+		{
+			var lineInfo = serviceProvider?.GetService<IXmlLineInfoProvider>()?.XmlLineInfo;
+			if (Android == null && GTK == null && iOS == null && 
+				macOS == null && Tizen == null && UWP == null && 
+				WPF == null && Default == null && string.IsNullOrEmpty(Other))
```
StephaneDelcroix: doesn't build. `Other` is no longer defined

kzu: 🤦‍♂️

ghuntley:
this is rad

akamud:
This should be added to the documentation, how are the documentation changes being tracked and linked to the Pull Requests?

EmilAlipiev:
It doesnt seem to be working using like below. I am getting 

No property, bindable property, or event found for 'Default', or mismatching type between value and property.

`      <OnIdiom x:TypeArguments="x:Double" Phone="150" Tablet="250" Default="300" />`

that seems to be working however

    `   HeightRequest="{OnIdiom Phone=200, Tablet=300, Default=350}"`

is that intended?

StephaneDelcroix:
@EmilAlipiev we have currently 2 issues about the `OnPlatform`/`OnIdiom` markup extensions: #3862 and #3920. If your issue relate to any of those, please add a comment. Otherwise, please open a new issue, as we do not follow-up on closed PRs. Thanks.

PR #2579, opened 2018-05-02
[iOS] Added icon to context menu actions instead of text [BROKEN]

Sebastian1989101:
### Description of Change ###

Changed context action menu of ViewCells to Icons like it is in native apps and on Android instead of Text. This only happens if the ListViewRenderer is overwritten and the virtual protected property `ActionCellIconInsteadOfText` is set to `true`. This way this is not a breaking change as you mentioned here: https://github.com/xamarin/Xamarin.Forms/pull/1487

### Bugs Fixed ###

- None

### API Changes ###

 - None

### Behavioral Changes ###

Instead of text the icon will appear in context actions like it is on Android.

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [ ] Rebased on top of the target branch at time of PR
- [ ] Changes adhere to coding standard
- [ ] Consolidate commits as makes sense

PR #2629, opened 2018-05-08
[iOS] Added icon to context menu actions instead of text 

Sebastian1989101:
### Description of Change ###

Changed context action menu of ViewCells to Icons like it is in native apps and on Android instead of Text. This only happens if the ListViewRenderer is overwritten and the virtual protected property `ContextActionDisplayType` is set to `ContextActionCellDisplay.Icon`. This way this is not a breaking change as you mentioned here: https://github.com/xamarin/Xamarin.Forms/pull/1487 

**Note**: This is a updated version from #2579 due to a few problems with the rebase.

### Bugs Fixed ###

- None

### API Changes ###

 - None

### Behavioral Changes ###

Instead of text the icon will appear in context actions like it is on Android.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Cells/CellTableViewCell.cs
+++ b/Xamarin.Forms.Platform.iOS/Cells/CellTableViewCell.cs
@@ -41,7 +41,7 @@ namespace Xamarin.Forms.Platform.iOS
 				PropertyChanged(this, e);
 		}
 
-		internal static UITableViewCell GetNativeCell(UITableView tableView, Cell cell, bool recycleCells = false, string templateId = "")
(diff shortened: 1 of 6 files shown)


StephaneDelcroix:
you probably should add a PlatformSpecific API to the flag can be set from Core, without inheriting ListView

PureWeen:
@Sebastian1989101  can you address @rmarinho's comments? Or do you need us to take this one over?

samhouts:
@Sebastian1989101 We will close this for now, but if you are able to address the comments above, please feel free to reopen it! Thank you!

PR #2684, opened 2018-05-11
[Android] Fix issue #2634. Draw no border if no BorderColor is set

### LINKED ISSUE #2634: Forms 3.0: White border around frames?
In case of my app it looks cool and is a great design idea but I guess it might not be the expected / wanted result that Frames now have a random white border. This appears for me since the 3.0 update of forms.

activa:
### Description of Change ###

This is a fix for issue #2634. On Android, when no BorderColor is set, a white border is always drawn around a Frame. That is a regression bug introduced in 3.0.0

### Bugs Fixed ###

fixes #2634 

### API Changes ###

None

### Behavioral Changes ###

Android only: no border will be drawn around a frame if BorderColor is not set or is set to Color.Default. Previously, a white border was drawn

### PR Checklist ###

Android FrameRenderer doesn't seem to have any tests.


activa:
@StephaneDelcroix  Rebased and target changed to 3.0.0

PR #2685, opened 2018-05-11
[Maps] Handle Pin.PropertyChanged events

### LINKED ISSUE #2121: Maps should listen to Pin property changes
### Description

The Pin class in the Maps project inherits from BindableObject, but the Map does not update on a property change to a Pin.

### Steps to Reproduce

1. Add a Pin to a Map.
2. Change the Position property of the pin.
3. Change the Label property of the pin.
4. Change the Address property of the pin.

### Expected Behavior

2. The pin will move to its new position on the map.
3. When the pin is tapped and its info window is displayed, it will display the new Label on the first line.
4. When the pin is tapped and its info window is displayed, it will display the new Address on the second line.

### Actual Behavior

No property changes are reflected in the UI.

### Basic Information

- Platforms Affected: Android, iOS
- Reproduction: https://gist.github.com/jcmanke/13ebdad7c58f71bf74f4ffc9841f7cea

jcmanke:
### Description of Change ###

Map renderers listen for the PropertyChanged event on each Pin in Map.Pins to update Label, Address, and Position of native pins when changed in cross-platform code.

### Bugs Fixed ###

- Fixes #2121

### API Changes ###

/

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls/GalleryPages/MapGallery.cs
+++ b/Xamarin.Forms.Controls/GalleryPages/MapGallery.cs
@@ -91,7 +91,19 @@ namespace Xamarin.Forms.Controls
 				map.Pins.Add (new Pin { Position = pos, Label = "Rui" });
 				map.MoveToRegion (MapSpan.FromCenterAndRadius (pos, Distance.FromMiles (0.5)));
 			};
-			
+
+			var buttonEditPin = new Button { Text = "Edit Pin" };
+			buttonEditPin.Clicked += (a, e) =>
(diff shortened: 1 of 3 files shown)


StephaneDelcroix:
an UWP implementation would be nice as well, but that shouldn't prevent this branch from being merged

jcmanke:
No UWP implementation required, it already works there as the native PushPin is bound to the Pin.

rmarinho:
Failing tests not related

PR #2696, opened 2018-05-14
[Android] Only set line height value if needed

rmarinho:
### Description of Change ###

We were applying line height of 0 on Fast Renderers 

### Bugs Fixed ###

- Failing tests

### PR Checklist ###


StephaneDelcroix:
as mentioned by @rmarinho, this is not the proper fix, as it prevents animating line-height from 0 to value

StephaneDelcroix:
#2714 is a better fix

PR #2703, opened 2018-05-14
[Android] Added a custom call for calling any manually set OnTouchListeners

### LINKED ISSUE #1942: [Android]   Attached Touch Listener events do not dispatch to immediate parent Grid Renderer View on Android when Child fakes handled
Migrated from https://bugzilla.xamarin.com/show_bug.cgi?id=60823

### Description

Touch listeners attached by a user do not dispatch to the parent view when the child element has "fake" handled an Event

Grid
-- Label

Label fake handles OnTouchEvent which causes the parent view during the DispatchTouchEvents phase to not call OnTouch on Listeners attach via *SetTouchListener* . *OnTouchEvent* is successfully called which is why Xamarin Forms Gestures continue to function properly but if a user has a custom listener they have assigned that will not be called

### Steps to Reproduce

> 1. Tap on the red text "Tap me (inner Grid)!"

### Expected Behavior

> It should become "Was tapped: Inner Grid (red label)"

### Actual Behavior

The above does not occur and the text refers to the green label.

### Basic Information

- Version with issue: Releases after 2.3.4.270
- Last known good version: 2.3.4.270
- Platform Target Frameworks:
  - Android: 8.0
- Android Support Library Version:  25.4.0.2

### Screenshots

N/A

### Reproduction Link

[Original Bugzilla attachment working on 2.3.4.270](https://bugzilla.xamarin.com/attachment.cgi?id=25874)

PureWeen:
### Description of Change ###

For cases when 
- there's a DefaultRenderer in the visual tree
- the child of that DefaultRenderer element being clicked on is one of the Android elements that fakes an OnTouchEvent to short circuit 
- the user manually attaches a Touch Listener to the Default Renderer

This change will cause the Touch Event to correctly propagate to that touch listener

### Bugs Fixed ###

fixes #1942

### Behavioral Changes ###
This shouldn't cause any  behavioral changes. The only path I can follow where a change in behavior would happen is  

- someone setup the scenario above but didn't realize their touch listener wasn't called
- they were returning TRUE from the listener
- They are then expecting OnTouchEvent to still be called

But even in normal Xamarin Android world if you return true from an attached TouchListener the OnTouchEvent of that view won't be called.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.ControlGallery.Android/CustomRenderers.cs
+++ b/Xamarin.Forms.ControlGallery.Android/CustomRenderers.cs
@@ -17,12 +17,15 @@ using Android.Runtime;
 using Android.Util;
 using AButton = Android.Widget.Button;
 using AView = Android.Views.View;
+using AViewGroup = Android.Views.ViewGroup;
 using Android.OS;
 using System.Reflection;
 using Android.Text;
 using Android.Text.Method;
 using Xamarin.Forms.Controls.Issues;
 
+
+[assembly: ExportRenderer(typeof(Issue1942.CustomGrid), typeof(Issue1942GridRenderer))]
 [assembly: ExportRenderer(typeof(Bugzilla31395.CustomContentView), typeof(CustomContentRenderer))]
 [assembly: ExportRenderer(typeof(NativeListView), typeof(NativeListViewRenderer))]
 [assembly: ExportRenderer(typeof(NativeListView2), typeof(NativeAndroidListViewRenderer))]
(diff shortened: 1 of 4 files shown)


PureWeen:
build --uitests

PureWeen:
build --uitests

PureWeen:
build --uitests

PureWeen:
build --uitests

PureWeen:
UI Tests are all passing (checked appcenter) even though above says differently

PR #2715, opened 2018-05-15
[Android] Add null-check to VisualElementPackager.RemoveChild

bruzkovsky:
There can be a NullReferenceException when accessing the field _childPackagers because it is initialized in AddChild and not in ctor.

### Description of Change ###

Add null-check.

### Bugs Fixed ###

None that I am aware of

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


StephaneDelcroix:
> None that I am aware of

so it's not an NRE you've experienced ?

bruzkovsky:
@StephaneDelcroix no I experienced it when calling StackLayout.Children.Clear() on a StackLayout that didn't have any children. From the StackTrace I found the location in the source and noticed this null-check missing. I didn't file a bug report because the change was so small, sorry for the confusion!

PR #2742, opened 2018-05-16
[Binding] FallbackValue and TargetNullValue

### LINKED ISSUE #1803: [Enhancement] Add FallbackValue and TargetNullValue properties to BindingBase
### Rationale

The two binding properties are very handy in situations when the binding is not available or the source is null. It eliminates the need to create value converters or add unnecessary code in the view-model.

`FallbackValue`:  Gets or sets the value to use when **the binding is unable to return a value**.
`TargetNullValue`:  Gets or sets the value that is used in the target when **the value of the source is null**.

First submitted 4 years ago in 2014: https://bugzilla.xamarin.com/show_bug.cgi?id=22720

See their Windows XAML counterparts (_XAML Standard where are thou?_): 
https://msdn.microsoft.com/en-us/library/system.windows.data.bindingbase.fallbackvalue(v=vs.110).aspx
https://msdn.microsoft.com/en-us/library/system.windows.data.bindingbase.targetnullvalue(v=vs.110).aspx

### Implementation

The `BindingBase` class needs to get the two properties. If they are not set, everything will still work like today.

### Backward Compatibility

Everything still works.

StephaneDelcroix:
### Description of Change ###

[Binding] FallbackValue and TargetNullValue

Changes applies to
 - Bindings
 - CompiledBindings
 - {Binding} markup extension
### Bugs Fixed ###

- fixes #1803

### API Changes ###

**Added**
```csharp
public object BindingBase.FallbackValue { get; set; }
public object BindingBase.TargetNullValue { get; set; }
```

#### FallbackValue
Prior to this, when a path (or part of the path) wasn't resolved, the `BindableProperty.DefaultValue` was applied.
Now, if the `Binding.FallbackValue` is non-null, the `FallbackValue` is used.

Converters are never applied.

#### TargetNullValue
Prior to this, a source value of `null` was eventually converted (if a converter was present), eventually formatted (if stringFormat was defined), and the result was then applied.
Now, the value is eventually converted, and if it's still null after the conversion, the TargetNullValue (if non-null) is used. StringFormatting is not applied if TargetNullValue is set.

when those properties aren't used, the previous behavior is unchanged.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/BindingExpression.cs
+++ b/Xamarin.Forms.Core/BindingExpression.cs
@@ -162,7 +162,7 @@ namespace Xamarin.Forms
 					value = Binding.GetSourceValue(value, property.ReturnType);
 				}
 				else
-					value = property.DefaultValue;
+					value = Binding.FallbackValue ?? property.DefaultValue;
 
 				if (!TryConvert(part, ref value, property.ReturnType, true))
 				{
--- a/Xamarin.Forms.Core/TypedBinding.cs
+++ b/Xamarin.Forms.Core/TypedBinding.cs
@@ -199,7 +199,7 @@ namespace Xamarin.Forms.Internals
 				Subscribe((TSource)sourceObject);
 
 			if (needsGetter) {
-				var value = property.DefaultValue;
+				var value = FallbackValue ?? property.DefaultValue;
 				if (isTSource) {
 					try {
 						value = GetSourceValue(_getter((TSource)sourceObject), property.ReturnType);
--- a/Xamarin.Forms.Core.UnitTests/BindingUnitTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/BindingUnitTests.cs
@@ -2211,5 +2211,25 @@ namespace Xamarin.Forms.Core.UnitTests
 			Assert.That(view.GetValue(bp1w), Is.EqualTo("qux"));
 			Assert.That(view.GetValue(bp1t), Is.EqualTo("qux"));
 		}
+
+		[Test]
+		public void FallbackValueWhenSourceIsNull()
+		{
+			var bindable = new MockBindable();
+			var property = BindableProperty.Create("Foo", typeof(string), typeof(MockBindable), "default");
+			bindable.SetBinding(property, new Binding("Foo.Bar") { FallbackValue = "fallback" });
+			Assert.That(bindable.GetValue(property), Is.EqualTo("fallback"));
+		}
+
+		[Test]
+		public void TargetNullValue()
+		{
+			var bindable = new MockBindable();
+			var property = BindableProperty.Create("Foo", typeof(string), typeof(MockBindable), "default");
+			bindable.SetBinding(property, new Binding("Text") { TargetNullValue = "fallback" });
+			Assert.That(bindable.GetValue(property), Is.EqualTo("default"));
+			bindable.BindingContext = new MockViewModel();
+			Assert.That(bindable.GetValue(property), Is.EqualTo("fallback"));
+		}
 	}
 }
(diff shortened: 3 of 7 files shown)


andreinitescu:
The name of the test method `FallbackValueWhenSourceIsNull` [here](https://github.com/xamarin/Xamarin.Forms/blob/eb4ecd8b76e40e94b207d94895c9497d1d3cbfc9/Xamarin.Forms.Core.UnitTests/BindingUnitTests.cs#L2216) is misleading. It sounds as if it's when source is null, which is actually not what the `FallbackValue` is for, `TargetNullValue` is for that.
Based on what I see, a better name should be `FallbackValueWhenSourceCantBeResolved`.

It's not clear to me how exactly `FallbackValue` works in this implementation. 
The standard definition in WPF and UWP is "Gets or sets the value to use when the binding is unable to return a value". 
This includes not only when source can't be resolved, but also:
- The value converter, if any, is not able to convert the resulting value.
- The resulting value is not valid for the binding target (target) property.

See WPF https://msdn.microsoft.com/en-us/library/system.windows.data.bindingbase.fallbackvalue(v=vs.110).aspx
and UWP https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.data.binding.fallbackvalue#Windows_UI_Xaml_Data_Binding_FallbackValue

StephaneDelcroix:
@andreinitescu FallbackValue is used when the source can not be resolved. TargetNullValue is when the source is resolved, but the value is null.

StephaneDelcroix:
@andreinitescu 
> It's not clear to me how exactly FallbackValue works in this implementation.
> The standard definition in WPF and UWP is:
>
> > Gets or sets the value to use when the binding is unable to return a value.
> This includes not only when source can't be resolved, but also:
>
> The value converter, if any, is not able to convert the resulting value.

We don't have the concept of `UnsetValue`, so no

> The resulting value is not valid for the binding target (target) property.

Excellent point. In this case, we do nothing (like we did for `DefaultValue`). As this would change the way `DefaultValue` works as well, could you please open a new issue, so we can evaluate the risk over there ? Thanks

Xamarin.Forms.Core/BindingExpression.cs
```diff
@@ -162,7 +162,7 @@ void ApplyCore(object sourceObject, BindableObject target, BindableProperty prop
 					value = Binding.GetSourceValue(value, property.ReturnType);
 				}
 				else
-					value = property.DefaultValue;
+					value = Binding.FallbackValue ?? property.DefaultValue;
```
jassmith: should this be property.GetDefaultValue();

Xamarin.Forms.Core/TypedBinding.cs
```diff
@@ -199,7 +199,7 @@ internal void ApplyCore(object sourceObject, BindableObject target, BindableProp
 				Subscribe((TSource)sourceObject);
 
 			if (needsGetter) {
-				var value = property.DefaultValue;
+				var value = FallbackValue ?? property.DefaultValue;
```
jassmith: Same question

Xamarin.Forms.Core/BindingExpression.cs
StephaneDelcroix: I think you're right, but it's unrelated to the issue at hand. Also, while setting the default value (or even the fallback ?) we should wet the bpcontext flags accordingly.

StephaneDelcroix: see #2752

ravero:
This is a nice addition. Could you provide a sample of how to use it with {Binding} Markup extension?

Issue #2167, opened 2018-03-22
[Android] Regression HorizontalTextAlignment of Entry control no longer works


AlexeyStarkov:
### Description
HorizontalTextAlignment not working for entry

*Note*
When testing Entry inside the Xamarin.ControlGallery project the HorizontalTextAlignment for entry works fine but when you use HorizontalTextAlignment in a Xamarin Forms project then HorizontalTextAlignment  doesn't work so most likely one of the custom renderers in the ControlGallery project is giving the illusion that this is working

### Steps to Reproduce

1. Add Entry control to page
2. Set "Center" or "End" to HorizontalTextAlignment property
3. Run

### Expected Behavior
Placeholder and text can be placed to Center or End.
### Actual Behavior
Placeholder and text are allways in Start position
### Basic Information

- Version with issue: 3.0.0.296286-pre2
- Last known good version: 2.5.****
- IDE: Visual Studio 2017 (PC)
- Platform Target Frameworks: 
  - Android: Target 8.1
- Android Support Library Version: 27.0.2
- Nuget Packages:
- Affected Devices: Test device: samsung galaxy a5 2017 android 7.0

jassmith:
Works fine for me in testing. Can you give me a repro?

kingnguyen1993:
I have too, May be it happen when use custom renderer.

AlexeyStarkov:
No, I have default Entry control without custom renderer.
<Entry Margin="20,3" Text="{Binding CardNumber, Mode=TwoWay}" Placeholder="0000 0000 0000 0000" HorizontalTextAlignment="Center"  Keyboard="Numeric" PlaceholderColor="{StaticResource PlaceholderColor}"/>

I am trying create new empty project, but I can't because I can't update Xamarin.Android.Support packages (NU1107: Version conflict detected for Xamarin.Android.Support.Compat. Reference the package directly from the project to resolve this issue.) Working with Xamarin is incredibly annoying. All the time something falls off and breaks down. By the way I was forced to do update to 3.0.0.296286-pre2 because after update to last stable version I got the wrong button mapping. After them I tried to do downgrade, but I got any errors too and I could not do it. I had no choice but to update Xamarin to 3.0.0.296286-pre2. (*Facepalm*). Now my buttons are look right, but I've problem with entries (*head to wall*)

PureWeen:
[...]

PR #2743, opened 2018-05-16
Fixes HorizontalTextAlignment not being applied to TextView Elements

PureWeen:
### Description of Change ###
Still set the TextView.Gravity property based on the Horizontal Alignment

In my testing just setting TextAlignment was unreliable. For example in the ControlGallery project it works fine but if you create a new XF project and try to set the HorizontalTextAlignment there on an Entry field it wouldn't work.

In the latter case whenever the View.TextAlignment was set to Center android was just reverting it back to Gravity.

### Bugs Fixed ###
https://github.com/xamarin/Xamarin.Forms/issues/2167

### PR Checklist ###


PureWeen:
build --uitests

PureWeen:
build --uitests

PureWeen:
The tests all passed in app center even though above is not reporting it as such

PR #2765, opened 2018-05-18
[Core] Fix Slider Maximum , Minimum binding issue

### LINKED ISSUE #1943: Slider crashes when Minimum value greater than zero and set before Max value in XAML

### Description
Migrating https://bugzilla.xamarin.com/show_bug.cgi?id=43435 here

The difference between the working and crashing versions is the order of the Minimum and Maximum properties in the XAML. Note that if Minimum is set to zero, the order is no longer an issue.

### Steps to Reproduce

Run app, note that it works fine
Uncomment line #8 in MainPage.xaml
Run app, note it crashes with "System.ArgumentException: Value was an invalid value for Minimum"

### Expected Behavior
App does not crash

### Actual Behavior
App crashes

### Basic Information

- IDE: VS2017
- Platform Target Frameworks: 
  - iOS:  yes
  - Android: yes
  - UWP:  not tested
- Android Support Library Version: 25
- Nuget Packages: just forms
- Affected Devices: tested on Nexus 5x running Android 7, and iPhone 5s simulator running iOS 9.3

### Reproduction Link
https://www.dropbox.com/s/qdweqnvh1hsj4wh/slidercrashsample.zip?dl=0

### LINKED ISSUE #2761: [Core] Slider unable to update Maximum , Minimum by binding
### Description
 Can't update the Maximum/Mnimum using binding
 Becuase validateValue
https://github.com/xamarin/Xamarin.Forms/blob/882bf07917bea57b93d0d9becd768dcab739c26f/Xamarin.Forms.Core/Slider.cs#L10-L14

https://github.com/xamarin/Xamarin.Forms/blob/882bf07917bea57b93d0d9becd768dcab739c26f/Xamarin.Forms.Core/Slider.cs#L21-L24

`validateValue` make hard to update value, according to value, Value update order should be changed. like below
```C#
   if (max > Minimum)
   {
        Maximum = max;
        Minimum = min;
   }
   else
   {
        Minimum = min;
        Maximum = max;
   }
```
But it is impossible in binding

So, I propose remove validateValue handler and update coerceValue like below
``` c#
		public static readonly BindableProperty MinimumProperty = BindableProperty.Create("Minimum", typeof(double), typeof(Slider), 0d, BindingMode.OneWay, coerceValue: (bindable, value) =>
		{
			var slider = (Slider)bindable;
			slider.Value = Math.Max((double)value, slider.Value);
			return value;
		}, propertyChanged:(bindable, oldValue, newValue) =>
		{
			var slider = (Slider)bindable;
			slider.SetValueCore(MaximumProperty, Math.Max(slider.Maximum, (double)newValue));
		});

		public static readonly BindableProperty MaximumProperty = BindableProperty.Create("Maximum", typeof(double), typeof(Slider), 1d, BindingMode.OneWay, coerceValue: (bindable, value) =>
		{
			var slider = (Slider)bindable;
			slider.Value = Math.Min((double)value, slider.Value);
			return value;
		}, propertyChanged:(bindable, oldvalue, newValue) =>
		{
			var slider = (Slider)bindable;
			slider.SetValueCore(MinimumProperty, Math.Min(slider.Minimum, (double)newValue));
		});
```

### Steps to Reproduce

``` c#
    class RangeModel
    {
        public double Max { get; set; }
        public double Min { get; set; }
    }

    Slider slider = new Slider();
    slider.SetBinding(Slider.MaximumProperty, new Binding("Max"));
    slider.SetBinding(Slider.MinimumProperty, new Binding("Min"));
    slider.BindingContext = new RangeModel { Min = -100, Max = -1 }; // ArgumentException was thrown
```

### Expected Behavior
 Update Minimum and Maximum
### Actual Behavior
 ArgumentException was thrown (Value was an invalid value for Maximum)

myroot:
### Description of Change ###
 Remove validateValue handler on Maximum/Minimum properties
 Use propertyChanged for Maximum/Minimum integrity

### Bugs Fixed ###
- fixes #1943
- fixes #2761

### API Changes ###
None

### Behavioral Changes ###
Allow set Minimum value that greater than Maximum value, after Minimum was updated, Maximum value is changed to Max(Minimum, Maximum)

Allow set Maximum value that smaller than Minimum, after Maximum was updated, Minimum value is changed to Min(Minimum, Maximum)

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2761.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2761.cs
@@ -1,0 +1,94 @@
+﻿using System;
+
+using Xamarin.Forms.CustomAttributes;
+using System.Collections.Generic;
+using Xamarin.Forms.Internals;
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 2761, "[Core] Slider unable to update Maximum , Minimum by binding")]
+	public class Issue2761 : TestContentPage
+	{
+		class RangeModel
+		{
+			public double Max { get; set; }
+			public double Min { get; set; }
+		}
+		protected override void Init()
+		{
+			Slider slider = new Slider
+			{
+				HorizontalOptions = LayoutOptions.FillAndExpand,
+			};
+
+			var minLabel = new Label { };
+			var maxLabel = new Label { };
+			var logLabel = new Label { Text = "Log : \n" };
+			minLabel.SetBinding(Label.TextProperty, new Binding("Minimum") { StringFormat = "Min : {0:F2}" });
+			maxLabel.SetBinding(Label.TextProperty, new Binding("Maximum") { StringFormat = "Max : {0:F2}" });
+			minLabel.BindingContext = slider;
+			maxLabel.BindingContext = slider;
+
+			slider.SetBinding(Slider.MaximumProperty, new Binding("Max"));
+			slider.SetBinding(Slider.MinimumProperty, new Binding("Min"));
+
+			System.Console.WriteLine($"Slider.MaximumProperty.DefaultBindingMode = {Slider.MaximumProperty.DefaultBindingMode}");
+
+			RangeModel range1 = new RangeModel { Min = -100, Max = -1 };
+			RangeModel range2 = new RangeModel { Min = 0, Max = 49 };
+			RangeModel range3 = new RangeModel { Min = 50, Max = 100 };
+
+			Padding = new Thickness(20);
+			Content = new StackLayout
+			{
+				Children =
+				{
+					new StackLayout
+					{
+						Orientation = StackOrientation.Horizontal,
+						Children =
+						{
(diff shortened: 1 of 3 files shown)


myroot:
Binding the `Value` is difficult issue to solve. because It twoway binding. 
But I don't think fixed behavior is more wrong than the current behavior

| - | Can binding Maximum/Minimum | Can binding Value |
|---|---|---|
| current behavior |  X | X |
| fixed behavior | O | X |

charlesroddie:
Here is an idea:

Setting Value, Min, Max would set internal mutable variables `reqMin`, `reqVal`, `reqMax`. Getting `Value`, `Min`, `Max` gets the actual values, which are a function of the these req values. If `reqMin < reqVal < reqMax` then the actual values are `(reqMin,reqVal,reqMax)`. If not, then the actual values take some other valid numbers, maybe `(0,0,1)` or `(a,b,c)` where
`a = min(reqMin,reqVal,reqMax)`, `c = max(...)` and `b = min(max(a,reqVal),c)`.

This would not prevent temporary glitches in Value,Min,Max (and so as a purist I would still say the correct fix is to have a single float*float*float property not three float properties) but it would prevent these glitches from having lasting effects (@StephaneDelcroix 's example) and also from crashing the UI (since the UI always gets valid inputs).

PR #2775, opened 2018-05-18
Visual element packager

Clancey:
### Description of Change ###
Fixed a crash with null renderers. Or invalid indexes
Jason Smith should approve!
### PR Checklist ###


Xamarin.Forms.Platform.Android/VisualElementPackager.cs
```diff
@@ -159,7 +160,7 @@ void OnChildAdded(object sender, ElementEventArgs e)
 			if (view != null)
 				AddChild(view);
 
-			if (ElementController.LogicalChildren[ElementController.LogicalChildren.Count - 1] != view)
+			if (ElementController.LogicalChildren.FirstOrDefault() != view)
```
StephaneDelcroix: we try not to use linq expressions

Clancey: I understand. But in the case it is correct. Safety is better.

myroot: Why not **Last**orDefault?

Clancey: You are correct. It should be last or default.

StephaneDelcroix: well, that will enumerate all the children for nothing. probably why we used the indexer before...

myroot: Actually, did not enumerate all children, If it's instance of IList

PR #2779, opened 2018-05-19
[All] Dont call perf calls on normal builds

jassmith:
Just #ifdef out a bunch of things that should have never not been ifdef'd


jassmith:
If you're asking why, Guid.NewGuid() aint cheap

StephaneDelcroix:
we did this at a point to enable our user to get performance reports for their own code.

I have no idea, or no indicator, that it was ever used.

if we're going to remove this, the previous solution using `[Conditional]` was achieving the same effect, but in a cleaner way

StephaneDelcroix:
@samhouts knows about all this

PR #2791, opened 2018-05-21
[Core] Remove Guid.NewGuid() calls from performance tracking

samhouts:
### Description of Change ###

Unless there is a `PerformanceProvider`, we should not waste time creating a reference for the tracker.

### Bugs Fixed ###

- N/A

### API Changes ###

Added parameter to internal `Performance.Start` method

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Performance.cs
+++ b/Xamarin.Forms.Core/Performance.cs
@@ -1,6 +1,7 @@
 ﻿using System.ComponentModel;
 using System.Runtime.CompilerServices;
 using System;
+using System.Threading;
 
 namespace Xamarin.Forms.Internals
 {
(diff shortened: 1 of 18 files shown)


Xamarin.Forms.Core/Performance.cs
```diff
@@ -15,14 +16,29 @@ public interface IPerformanceProvider
...
+		public static void Start(out string reference, string tag = null, [CallerFilePath] string path = null, [CallerMemberName] string member = null)
+		{
+			if (Provider == null)
+			{
+				reference = String.Empty;
+				return;
+			}
+
+			reference = Interlocked.Increment(ref Reference).ToString();
+			Provider?.Start(reference, tag, path, member);
```
StephaneDelcroix: Provider can't be null here

samhouts: oops, yeah, I meant to change that

PR #2792, opened 2018-05-21
Enumerate copy so list can be mutated w\o throwing

### LINKED ISSUE #2576: Regression - Android - Exception thrown when page is popped
PopAsync after a PushAsync throws Unhandled Exception:
System.InvalidOperationException: Collection was modified; enumeration operation may not execute. occurred

Run the attached program.  It will present a list of 'Announcements'.  
1. Tap one of the items to open a second page
2. press the back button to return to the previous page
3. exception is thrown

Page will pop without exception

Exception is thrown

- Version with issue:  2.5.1.444934
- Last known good version:  2.5.0.122203
- IDE: VS2017
- Platform Target Frameworks: 
  - Android: 8.1
- Android Support Library Version: 27.0.2
- Nuget Packages:
- Affected Devices:

[PopException.zip](https://github.com/xamarin/Xamarin.Forms/files/1965013/PopException.zip)

kingces95:
### Description of Change ###

Enumerate copy so list can be mutated w\o throwing. Fixes #2576

### Bugs Fixed ###

fixes #2576

### API Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/EnumerableExtensions.cs
+++ b/Xamarin.Forms.Core/EnumerableExtensions.cs
@@ -33,7 +33,7 @@ namespace Xamarin.Forms.Internals
 			if (predicate == null)
 				predicate = x => true;
 
-			foreach (IGestureRecognizer item in gestures)
+			foreach (IGestureRecognizer item in new List<IGestureRecognizer>(gestures))
 			{
 				var gesture = item as T;
 				if (gesture != null && predicate(gesture))


Xamarin.Forms.Core/EnumerableExtensions.cs
```diff
@@ -33,7 +33,7 @@ public static class EnumerableExtensions
 			if (predicate == null)
 				predicate = x => true;
 
-			foreach (IGestureRecognizer item in gestures)
+			foreach (IGestureRecognizer item in new List<IGestureRecognizer>(gestures))
```
hartez: `GetGesturesFor<T>` is used in a lot of places - allocating a new List every time is going to have an impact on performance. 

If the consumer needs to modify the list, the consumer should be making their own copy.

hartez: > I see yield being used in all the methods in that file. Is it a good choice? yield introduce a lot of state machinery code and still uses heap for allocation. If the idea was to minimize heap, that still happens with yield too. I think for some small collections like gestures, using yield has a negative impact on performance.

The idea is to short-circuit the iteration of the gestures when possible. For instance, a call like `GetGesturesFor<TapGestureRecognizer>(g => g.NumberOfTapsRequired == 1).Any()` doesn't have to iterate over the entire list of gestures; it can stop when it reaches one which matches the criteria. 

We're open to alternative implementations if there's a more performant way to implement this (without changing the method signature, as it's a public API). But we'd need a very compelling analysis or, better yet, some good benchmarks.

> Also, the main point of yield is when collection doesn't change.

I'm not sure what you mean by that. The method is returning an `IEnumerable`; modifying it during enumeration is not allowed. That's fundamental to enumerators - it doesn't have anything to do with `yield`. 

> And last, yield introduces some challenges in code because of its nature.

Again, not sure what you're getting at here. What challenges?

opcodewriter: The yield state machinery created by compiler shouldn't be overlooked.  See this post: https://blogs.msdn.microsoft.com/oldnewthing/20080812-00/?p=21273
The gestures collection is very very little and doesn't justify all the code emitted by compiler.

opcodewriter: yield can sometime add some complexity in code because it is lazy evaluated.

hartez: > The gestures collection is very very small and doesn't justify all the code emitted by compiler.

What alternative to `yield` do you suggest?

hartez: Creating a new List<T> incurs the overhead of allocating and initializing a whole new List<T>. Also, it prevents any short-circuiting. 

Consider a case like `GetGesturesFor<TapGestureRecognizer>(g => g.NumberOfTapsRequired == 1).Any()`. With `yield`, if the very first gesture fits the criteria, no further gestures have to be run through the cast to T, the null check, and the predicate() call. If you return a new List<T>, every gesture has to be run through that logic to create the concrete List, even if you only need to see the first gesture to know the result. 

So just creating a whole new List<T> creates more memory pressure than `yield` and guarantees more operations.

Or are you suggesting that we do something like `gestures.OfType<T>().Where(predicate)`?

kingces95: I'll optimize it!

kingces95: We're already using a `yield` and a `foreach` so that sucker is not all that optimized to begin with.

That said, if we still don't wanna allocate the list then I could simply unroll the loop a few times before the allocation. Chances are there are not more than a few IGestureRecognizer attached at any one time so a 3-4 element loop unroll would likely capture the majority of paths. Would that be sufficient?

kingces95:
Moved to: https://github.com/xamarin/Xamarin.Forms/pull/2828

Issue #1700, opened 2018-01-26
[Enhancement] UWP: TabStop/TabIndex support (for multiple Views)


PureWeen:
## Rationale

Currently there is no way to  influence the tab stop behaviors on UWP. Seeing as how there are a lot of UWP controls that have this property it's very tedious to implement this across all desired elements

## Implementation

Add IsTabStopProperty and TabIndexProperty to all applicable Forms controls.  IsTabStop and TabIndex are properties of Windows.UI.Xaml.Controls.Control  so all XamarinForms elements that map to a Control in UWP will need these properties added

- InputView (covers Entry, Editor)
- DatePicker
- Button
- ComboBox / Picker
- CommandBar? so adding to NavigationPage
- TabbedPage
- ProgresBar
- SearchBar
- Slider
- Stepper
- Switch 
- TextCell
- TimePicker
- Anything I forgot?

```csharp
public static readonly BindableProperty IsTabStopProperty = BindableProperty.Create(nameof(IsTabStop), typeof(bool), typeof(<view adding to>), false);

public static readonly BindableProperty TabIndexProperty = BindableProperty.Create(nameof(TabIndex), typeof(int), typeof(<view adding to>), 0);
```

Possibly surface the TabFocusNavigation property as well
```csharp
    public enum KeyboardNavigationMode
    {
        //
        // Summary:
        //     Tab indexes are considered on the local subtree only inside this container.
        Local = 0,
        //
        // Summary:
        //     Focus returns to the first or the last keyboard navigation stop inside of a container
        //     when the first or last keyboard navigation stop is reached.
        Cycle = 1,
        //
        // Summary:
        //     The container and all of its child elements as a whole receive focus only once.
        Once = 2
    }

public static readonly BindableProperty TabFocusNavigationProperty = BindableProperty.Create(nameof(TabFocusNavigation), typeof(KeyboardNavigationMode), typeof(VisualElement), KeyboardNavigationMode.Local);
```

### Android
Doesn't exist

### iOS
Doesn't exist

### UWP
Map the property from the Forms element to the applicable UWP control

## Backward Compatibility
Third party renderers may need to be updated to ensure that this functionality is supported through the new official mechanism. Further we will need to be careful to code the changes to the renderers in a careful manner to ensure that if someone is already using an effect support this feature that the effect is as best as possible not broken by our changes.  

## Difficulty : Medium
This task overall shouldn't be very difficult. Just a lot of places to get applied into UWP

[...]

Issue #2789, opened 2018-05-21
Improved A11y Support - TabIndex


samhouts:
## Description
As an app developer, I currently am unable to specify an expected focus/tab order for elements in my Xamarin.Forms app that will be used by a screen reader and/or keyboard navigation to improve the usability and accessibility of the app.  When this spec is complete, I can explicitly set the `TabIndex` on any focusable element to have it read to the app user and/or prepare it for input in a particular order.

### Acceptance criteria
* Solution must enable app developer to set an integer value on `Element.TabIndex` with the expectation that focus will cycle through each element in ascending order. 
* Conflicts on `TabIndex` will be resolved by the declaration order. 
* If no value is set in `TabIndex`, the default behavior will be used. 
* If the app developer attempts to set a focus order using a custom renderer, the values set by the custom renderer should take precedence.

#### Per the UWP specification:
* UI elements with `TabIndex` equal to 0 are added to the tab order based on declaration order in XAML or child collections.
* UI elements with `TabIndex` greater than 0 are added to the tab order based on the `TabIndex` value.
* UI elements with `TabIndex` less than 0 are added to the tab order and appear before any zero value.

## API changes
Add: `int Element.TabIndex { get; set; }`

### Description of changes
- UWP: https://docs.microsoft.com/en-us/windows/uwp/accessibility/keyboard-accessibility
- iOS: https://developer.apple.com/documentation/objectivec/nsobject/1615147-accessibilityelements#declarations 
- Android: https://developer.android.com/training/keyboard-input/navigation.html#Tab 
- macOS: https://github.com/xamarin/Xamarin.Forms/pull/1060 (convert from platform specific)

StephaneDelcroix:
for elements with equal TabIndex, declaration order matters, right ?

samhouts:
@StephaneDelcroix  Yep:

> Conflicts on TabIndex will be resolved by the declaration order.

GalaxiaGuy:
Will this have any impact on directional navigation (using a hardware dpad)?

The Android link suggests basic dpad navigation should work by default, but it does not seem to in a Xamarin Forms app:
https://bugzilla.xamarin.com/show_bug.cgi?id=30891

(I haven't yet tried other platforms.)

johnshardman:
[...]

PR #2795, opened 2018-05-22
[Core, UWP, Android] support TabStop/TabIndex

paymicro:
### Description of Change ###

Added support TabStop/TabIndex for `UWP` and `Android` platforms.
PR is expected to be large. It will add implementations for other platforms.

**Screencasts:**
UWP http://recordit.co/InFP4NqQAv
Android http://recordit.co/qwsSribcQ5

### Bugs Fixed ###

Fixes: #1700

### API Changes ###

Added:
 - bool Button.TabStop { get; set; } //Bindable Property
 - int Button.TabIndex { get; set; } // Bindable Property
 - bool DatePicker.TabStop { get; set; } //Bindable Property
 - int DatePicker.TabIndex { get; set; } // Bindable Property
 - bool Editor.TabStop { get; set; } //Bindable Property
 - int Editor.TabIndex { get; set; } // Bindable Property
 - bool Entry.TabStop { get; set; } //Bindable Property
 - int Entry.TabIndex { get; set; } // Bindable Property
 - bool InputView.TabStop { get; set; } //Bindable Property
 - int InputView.TabIndex { get; set; } // Bindable Property
 - bool NavigationPage.TabStop { get; set; } //Bindable Property
 - int NavigationPage.TabIndex { get; set; } // Bindable Property
 - bool Picker.TabStop { get; set; } //Bindable Property
 - int Picker.TabIndex { get; set; } // Bindable Property
 - bool ProgressBar.TabStop { get; set; } //Bindable Property
 - int ProgressBar.TabIndex { get; set; } // Bindable Property
 - bool SearchBar.TabStop { get; set; } //Bindable Property
 - int SearchBar.TabIndex { get; set; } // Bindable Property
 - bool Slider.TabStop { get; set; } //Bindable Property
 - int Slider.TabIndex { get; set; } // Bindable Property
 - bool Stepper.TabStop { get; set; } //Bindable Property
 - int Stepper.TabIndex { get; set; } // Bindable Property
 - bool Switch.TabStop { get; set; } //Bindable Property
 - int Switch.TabIndex { get; set; } // Bindable Property
 - bool TabbedPage.TabStop { get; set; } //Bindable Property
 - int TabbedPage.TabIndex { get; set; } // Bindable Property
 - bool TimePicker.TabStop { get; set; } //Bindable Property
 - int TimePicker.TabIndex { get; set; } // Bindable Property

### Behavioral Changes ###

Added `TabStop` & `TabIndex` properties.
The `IsTabStop` property do not work correctly on the UWP native elements such as `DatePicker`, `TimePicker`, `SearchBar` and `Stepper`.

### PR Checklist ###


StephaneDelcroix:
see #2789

Xamarin.Forms.Core/VisualElement.cs
```diff
@@ -100,6 +100,34 @@ public partial class VisualElement : Element, IAnimatable, IVisualElementControl
 
 		public static readonly BindableProperty FlowDirectionProperty = BindableProperty.Create(nameof(FlowDirection), typeof(FlowDirection), typeof(VisualElement), FlowDirection.MatchParent, propertyChanged: FlowDirectionChanged);
 
+		public static readonly BindableProperty TabIndexProperty =
+			BindableProperty.Create(nameof(TabIndex),
+									typeof(int),
+									typeof(View),
```
samhouts: Did you mean for this to be `VisualElement`, or did you mean to put these properties on `View`? `View` might be a better choice. I don't think it makes sense for `Page`s to have a `TabIndex`/`TabStop`

Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
```diff
@@ -8,6 +8,8 @@
 using Xamarin.Forms.Internals;
 using AView = Android.Views.View;
 using Xamarin.Forms.Platform.Android.FastRenderers;
+using Android.Runtime;
+using System.Linq;
```
samhouts: I wonder if we can do this without Linq. We do try to avoid it.

Hronos2t: Linq is avoided only on Android or anywhere?

Xamarin.Forms.Core/VisualElement.cs
paymicro: Oh. It's mistake. I mean here `VisualElement`.
I originally made for the `View`. But in `UWP` all UI elements have a `TabIndex/TabStop` include Pages.

paymicro:
I want to add an implementation for `iOS` / `MacOS`
There will be something similar on Android implementation.

Xamarin.Forms.Core/Application.cs
```diff
@@ -36,6 +36,33 @@ public Application()
...
+			{
+				foreach (var page in pages)
+				{
+					result.AddRange (WalkChildren (page));
+				}
+			}
+			return result;
+		}
+
+		public IEnumerable<Element> WalkChildren(Element element)
```
samhouts: Does this need to be public?

Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
```diff
@@ -135,6 +134,92 @@ public void UpdateLayout()
...
+			var tabIndexes = new Dictionary<int, List<VisualElement>>();
+			foreach (var ch in allChildrens)
+			{
+				if (ch is VisualElement ve && ve.IsTabStop)
+					childrensWithTabStop.Add(ve);
+			}
+			if (!childrensWithTabStop.Contains(Element))
+				return base.FocusSearch(focused, direction);
+
+			// groupping => childrensWithTabStop.GroupBy(c => c.TabIndex)
```
samhouts: grouping >> grouping

Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
```diff
@@ -135,6 +134,92 @@ public void UpdateLayout()
...
+				// search next element in same TabIndex group
+				var tabGroup = tabIndexes[tabIndex];
+				var nextSubIndex = tabGroup.IndexOf(element) + 1;
+				if (nextSubIndex > 0 && nextSubIndex < tabGroup.Count)
+				{
+					element = tabGroup[nextSubIndex];
+				}
+				else // search next element in next TabIndex group
+				{
+					if (direction.HasFlag(FocusSearchDirection.Backward) ||
```
samhouts: Let's check the flags explicitly instead of using `HasFlag`

Xamarin.Forms.Platform.iOS/VisualElementRenderer.cs
```diff
@@ -150,6 +152,118 @@ public void SetElementSize(Size size)
...
+			var tabIndexes = new Dictionary<int, List<VisualElement>> ();
+			foreach (var ch in allChildrens)
+			{
+				if (ch is VisualElement ve && ve.IsTabStop)
+					childrensWithTabStop.Add (ve);
+			}
+			if (!childrensWithTabStop.Contains (Element))
+				return null;
+
+			// groupping => childrensWithTabStop.GroupBy(c => c.TabIndex)
```
samhouts: grouping >> grouping

paymicro:
Hi @samhouts
I added in the test 2 buttons with `TabIndex` = 90 and 2 buttons with `TabIndex` = 100.
Their behavior corresponds to the description in #2789 - http://recordit.co/N7VsBdKh3K

Xamarin.Forms.Core/Application.cs
```diff
@@ -36,6 +36,20 @@ public Application()
...
 
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public override IEnumerable<T> WalkChildren<T>()
+		{
+			var pages = NavigationProxy?.ModalStack;
+			if (pages == null)
+				return base.WalkChildren<T>();
+
+			var result = new List<T>(base.WalkChildren<T>());
+			foreach (var page in pages)
```
hartez: Will this check all of the pages on the stack? It seems like we should only be worried about the tab indices for items on the top page of the stack.

Xamarin.Forms.Core/Element.cs
```diff
@@ -125,6 +125,19 @@ public string StyleId
 		[EditorBrowsable(EditorBrowsableState.Never)]
 		public ReadOnlyCollection<Element> LogicalChildren => LogicalChildrenInternal;
 
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public virtual IEnumerable<T> WalkChildren<T>() where T : Element
```
hartez: Is there a reason this needs to be depth-first? If not, `Element` already has a method `Descendants()` which iterates breadth-first over its children.

Xamarin.Forms.Platform.iOS/VisualElementRenderer.cs
```diff
@@ -150,6 +152,126 @@ public void SetElementSize(Size size)
...
+								tabIndexesMin = index;
+						}
+						tabIndex = biggerMin != int.MaxValue ? biggerMin : tabIndexesMin;
+						element = tabIndexes[tabIndex][0];
+					}
+				}
+#if __MACOS__
+				var renderer = Platform.GetRenderer(element);
+				// use reflection to get the "Control" property from a specific renderer
+				control = renderer?.GetType ().GetProperty ("Control")?.GetValue (renderer, null) as NativeView;
```
hartez: `renderer.NativeView` will work here without reflection.

Xamarin.Forms.Platform.iOS/VisualElementRenderer.cs
```diff
@@ -150,6 +152,126 @@ public void SetElementSize(Size size)
...
+
+		protected bool TabStop { get; set; } = true;
+
+		protected void UpdateTabStop () => TabStop = Element.IsTabStop;
+
+		protected void UpdateTabIndex () => TabIndex = Element.TabIndex;
+
+		public NativeView FocusSearch (bool forward)
+		{
+			var allChildrens = Application.Current.WalkChildren<VisualElement> ();
```
hartez: Is there any way we could consolidate some of the code in the two versions of `FocusSearch`? The code for grouping the elements by tab index and the code for searching foward/back is all the same; for maintenance purposes, it'd be nice to have it all in one or two methods.

(Not a deal breaker if consolidating it is just too awkward.)

Xamarin.Forms.Core/Application.cs
paymicro: I think so. It will be enough to know the page on which the element is located.

Xamarin.Forms.Core/Element.cs
paymicro: Oh. This method I did not notice. `Descendants()` will be enough. Thanks!

Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
```diff
@@ -135,6 +134,95 @@ public void UpdateLayout()
...
+								tabIndexesMin = index;
+						}
+						tabIndex = biggerMin != int.MaxValue ? biggerMin : tabIndexesMin;
+						element = tabIndexes[tabIndex][0];
+					}
+				}
+
+				var renderer = element.GetRenderer();
+				// use reflection to get the "Control" property from a specific renderer
+				control = renderer.GetType().GetProperty("Control")?.GetValue(renderer, null) as AView;
```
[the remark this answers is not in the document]
paymicro: `renderer.View` returns the `ViewRenderer` again, but we need a native view
http://recordit.co/LiiDK1BaKw

paymicro:
@samhouts accessibility focus order not such a trivial task, perhaps it is necessary to add it in another PR after this?

samhouts:
@paymicro Agreed! Thank you!

@hartez Would you like to re-review? Thanks!!

Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
```diff
@@ -135,6 +134,41 @@ public void UpdateLayout()
...
+				(direction & FocusSearchDirection.Backward) != 0 ||
+				(direction & FocusSearchDirection.Left) != 0 ||
+				(direction & FocusSearchDirection.Up) != 0);
+
+			do
+			{
+				element = element.FindNextElement(forwardDirection, tabIndexes, ref tabIndex);
+				var renderer = element.GetRenderer();
+				// use reflection to get the "Control" property from a specific renderer
+				control = renderer.GetType()?.GetProperty("Control")?.GetValue(renderer, null) as AView;
```
hartez: You should be able to do something like this:

```
var renderer = element.GetRenderer() as ViewRenderer;
control = renderer?.Control;
```

and avoid having to use reflection. Same thing on iOS/macOS.

On Android, if the `ViewRenderer` cast comes back `null`, then it's probably a fast renderer and you can just try casting `renderer` to `AView` and use that result.

Xamarin.Forms.Platform.Android/ViewRenderer.cs
```diff
@@ -14,6 +14,11 @@ public interface IViewRenderer
 		void MeasureExactly();
 	}
 
+	public interface INativeViewRenderer
```
hartez: For the time being, I think this interface can be `internal`.

Xamarin.Forms.Platform.iOS/ViewRenderer.cs
```diff
@@ -18,11 +18,16 @@ namespace Xamarin.Forms.Platform.iOS
 namespace Xamarin.Forms.Platform.MacOS
 #endif
 {
+	public interface INativeViewRenderer
```
hartez: `public` -> `internal`

programmation: I would really prefer that you don't do this. I realise that you're trying to reduce your maintenance surface but at the same time you are cutting off at the knees anyone who wants to use this interface in their own code. If it's at all possible to _not_ make things `internal` please please don't do it. 
I once had to re-implement all of the `WebViewRenderer` functionality in a custom renderer to get around the early rotation-handling issues in the original. Because everything was `internal`, to fix a fundamental bug I had to reproduce almost the entire renderer in my code rather than simply adjusting a small piece of it.
IMO _all_ the methods of a renderer should be `public` and `virtual` or `override`. When we're creating renderers we simply don't know enough of the use cases to lock anything down.

hartez: Well, you're not wrong that leaving it internal would keep custom renderers from being tab stops.

The current interface name (`INativeViewRenderer`) doesn't express what it does (and why it exists) very well; the classes implementing it actually have several 'native' views. What this is _really_ saying is "of all the native  views that make up this renderer, here's the one that makes sense as a tab stop".

So a better name for the interface would be `ITabStop`, with a single `TabStop` property that returns a native view. 

Also, all of the fast renderers on Android are missing this interface right now (the fast Button, Image, etc.) because they don't derive from `ViewRenderer`. 

All of which leads to me wonder if we should just add a `TabStop` property to `IVisualElementRenderer` on Android and iOS.

@samhouts Any thoughts on this? Between `ITabStop` and just adding the property to `IVisualElementRenderer`, do you see a clearly better option? Or a third way that we're missing?

hartez: Now that I think about it some more, we can't add it to `IVisualElementRenderer` without breaking existing implementations. So I think `ITabStop` is the way to go, and it will need to be applied to the fast renderers.

samhouts: I agree. Nice catch!

Xamarin.Forms.Platform.Android/ITabStop.cs
```diff
@@ -0,0 +1,9 @@
+﻿using AView = Android.Views.View;
+
+namespace Xamarin.Forms.Platform.Android
+{
+	internal interface ITabStop
```
hartez: This interface _does_ need to be public (same as iOS); see the discussion with @programmation. My apologies for the confusion.

jassmith:
build --uitests

PureWeen:
@paymicro  can you rebase this one to master?

I tried myself but there were conflicts so figure it's safer for you to do it

Hronos2t:
@PureWeen sure. Done!

PureWeen:
build --uitests

Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
```diff
@@ -135,6 +134,40 @@ public void UpdateLayout()
...
+			int maxAttempts = 0;
+			var tabIndexes = element?.GetTabIndexesOnParentPage(out maxAttempts);
+			if (tabIndexes == null)
+				return null;
+
+			int tabIndex = element.TabIndex;
+			AView control = null;
+			int attempt = 0;
+			bool forwardDirection = !(
+				(direction & FocusSearchDirection.Backward) != 0 ||
```
rookiejava: why aren't `FocusSearchDirection.Right` and `FocusSearchDirection.Down` dealt with here? Did you mean it?

Issue #2826, opened 2018-05-24
[Enhancement] Creating a bindable property with enum backing does not select enum default


adrianknight89:
### Description
When I create a bindable property for an enum, I have to explicitly set a default value in the Bindable.Create() method for that enum otherwise a runtime error will occur. Why is that? In my opinion, by default, the first member of an enum type should be chosen by the Create() method. See attached.

This issue seems to exist for structs as well. For example, if you create a bindable property for `TimeSpan`, you need to explicitly do `new TimeSpan()` or `default(TimeSpan)` while in reality a `TimeSpan` defaults to "00:00:00".

### Reproduction Link

[Demo.zip](https://github.com/xamarin/Xamarin.Forms/files/2036585/Demo.zip)

charlesroddie:
If there is going to be a default, it should be explicit. Placing a value first in an enum does not implly that it is intended to be a default. No need to make bindings even more hacky than they are already.

adrianknight89:
I should have been clear about default enum values. You're right that the first enumeration isn't always the default one. Here's an example:

```
enum MyEnum
{
      A = 1,
      B = 0,
      C = 2
}
```

`default(MyEnum)` would return `B`.

Official [documentation](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/xaml/bindable-properties) for bindable properties mentions the following for the `defaultValue` parameter of `Create()`:

> Providing a default property value that's different from the default for the type of the property.

which, to me, implies that `Create()` is supposed to set the property to its default if `defaultValue` is not provided.

I'd say the documentation should make it clear that for enums and structs, an explicit initialization must be provided or the `Create()` method should be modified to define defaults.

activa:
The documentation is wrong. The default value is of type object and the default is null. When the property type is a value type, the code is trying to cast null to a value type and of course that throws an exception.

So either the code needs to change so it uses default(T) in case the default value is null, or the documentation should make it clear that the provided default value should be something that can be casted the property type.

activa:
Fixing the code to do what the documentation says looks pretty easy.

Adding this to the BindableProperty constructor should do it:
```
if (ReturnType.IsValueType && Nullable.GetUnderlyingType(ReturnType) == null && DefaultValue == null)
    DefaultValue = Activator.CreateInstance(ReturnType);
```

activa:
See PR #2827

PR #2827, opened 2018-05-24
Support default value for value types in BindableProperty

activa:
According to the [documentation](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/xaml/bindable-properties), when no default value is provided in a call to BindableProperty.Create(), it will use the default value of the property type. However, this is not how it behaves for value types. When no explicit default value is provided for a value type property, an exception will be thrown.

### Description of Change ###

When a BindableProperty is created for a value type and the default value is null, the default value will be set to the default for the type.

### API Changes ###

No API changes

### Behavioral Changes ###

BindableProperty.Create() will no longer throw an exception when no default value is provided and the property type is a value type. The default value will be the default value for the property type.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/BindablePropertyUnitTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/BindablePropertyUnitTests.cs
@@ -148,5 +148,49 @@ namespace Xamarin.Forms.Core.UnitTests
 			Assert.AreEqual (1, bindable.GetValue (prop));
 		}
 
+		enum TestEnum
+		{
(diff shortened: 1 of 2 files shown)


adrianknight89:
@activa Thanks for your work! I'd appreciate if the unit test included an enum/struct just to test what is being discussed in #2826. I know you have the `returnType.GetTypeInfo().IsValueType` check, but being safe here... :)

activa:
@adrianknight89 I've added 2 unit tests for both enum and struct properties.

PR #2833, opened 2018-05-25
[Core] Fixes setting to Null Binding Context in Picker

### LINKED ISSUE #2499: Picker.cs Line 215 Throws Argument Out Of Range Exception when Binding Context gets set to Null
### Description

In Picker.cs, when we close a popup dialog that contains a picker control inside the "SelectedIndexChanged" event, the Items collection has 0 elements and a SelectedIndex is still set. (This can occur when some portion of the code sets the BindingContext to null)

Under these conditions, we get an argument out of range exception within the UpdateSelectedItem() method on line 215:

```
void UpdateSelectedItem()
{
	if (SelectedIndex == -1) {
		SelectedItem = null;
		return;
	}
	if (ItemsSource != null) {
		SelectedItem = ItemsSource [SelectedIndex];
		return;
	}
	SelectedItem = Items [SelectedIndex]; //Line 215
}

```
### Steps to Reproduce

1. Set up a Picker control in a popup dialog
2. In the SelectedIndexChanged event, set the BindingContext to null when the last element in the picker is chosen
3. In the SelectedIndexChanged event close the popup when the last element in the Picker is chosen
4. Run the code
5. Select the last element
6. Index out of bounds exception on line 215 of Picker.cs

### Expected Behavior

I would expect there to be a check on line 215 of the Picker.cs that does not allow choosing an item by index value when there aren't any elements in the list.

### Actual Behavior

Throws exception, app bombs

### Reproductions
```C#
Picker picker = new Picker();
			picker.ItemsSource = new List<string> { "cat", "mouse", "rabbit" };
			picker.SelectedIndexChanged += (_, __) =>
			{
				picker.ItemsSource = null;
			};

			return new ContentPage()
			{
				Content = picker
			};
```

https://github.com/xamarin/Xamarin.Forms/files/2036809/ExceptionExample.zip

### LINKED ISSUE #2815: Picker still attempts to set SelectedItem when BindingContext goes to null
### Description
When the BindingContext is set to null and the Picker has an ItemsSource set, then the binding still fires back a change to the source for the SelectedItem. When this occurs the SelectedItem binding causes the bound source to be set to null.

### Steps to Reproduce
1. Add a Picker with a binding to ItemsSource and SelectedItem.
2. Set a valid binding context.
3. Set the binding context back to null.

### Expected Behavior
The SelectedItem binding should not fire back a value to the source. Other controls do not operate this way when their binding context is set to null.

### Actual Behavior
The binding fires a value back to the source and since the ItemsSource has already been cleared at this point the value is always null for SelectedItem.

### Reproduction Link
[PickerBindingContextNull.zip](https://github.com/xamarin/Xamarin.Forms/files/2033682/PickerBindingContextNull.zip)

This is very similar to another issue but it is thought they may not be the same thing: https://github.com/xamarin/Xamarin.Forms/issues/2499

paymicro:
### Description of Change ###

In the `UpdateSelectedItem` sends the real `selectedIndex`

### Bugs Fixed ###

Fixes #2499
Fixes #2815

### API Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Picker.cs
+++ b/Xamarin.Forms.Core/Picker.cs
@@ -150,8 +150,11 @@ namespace Xamarin.Forms
 
 		void OnItemsCollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
 		{
-			SelectedIndex = SelectedIndex.Clamp(-1, Items.Count - 1);
-			UpdateSelectedItem();
+			var oldIndex = SelectedIndex;
+			var newIndex = SelectedIndex = SelectedIndex.Clamp(-1, Items.Count - 1);
+			// If the index has not changed, still need to change the selected item
+			if (newIndex == oldIndex)
+				UpdateSelectedItem(newIndex);
 		}
 
 		static void OnItemsSourceChanged(BindableObject bindable, object oldValue, object newValue)
@@ -214,13 +217,13 @@ namespace Xamarin.Forms
 			((LockableObservableListWrapper)Items).InternalClear();
 			foreach (object item in ItemsSource)
 				((LockableObservableListWrapper)Items).InternalAdd(GetDisplayMember(item));
-			UpdateSelectedItem();
+			UpdateSelectedItem(SelectedIndex);
 		}
 
 		static void OnSelectedIndexChanged(object bindable, object oldValue, object newValue)
 		{
 			var picker = (Picker)bindable;
-			picker.UpdateSelectedItem();
+			picker.UpdateSelectedItem(picker.SelectedIndex);
 			picker.SelectedIndexChanged?.Invoke(bindable, EventArgs.Empty);
 		}
 
@@ -239,19 +242,19 @@ namespace Xamarin.Forms
 			SelectedIndex = Items.IndexOf(selectedItem);
 		}
 
-		void UpdateSelectedItem()
+		void UpdateSelectedItem(int index)
 		{
-			if (SelectedIndex == -1) {
+			if (index == -1) {
 				SelectedItem = null;
 				return;
 			}
 
 			if (ItemsSource != null) {
-				SelectedItem = ItemsSource [SelectedIndex];
+				SelectedItem = ItemsSource [index];
 				return;
 			}
 
-			SelectedItem = Items [SelectedIndex];
+			SelectedItem = Items [index];
 		}
 
 		public IPlatformElementConfiguration<T, Picker> On<T>() where T : IConfigPlatform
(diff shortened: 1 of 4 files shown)


myroot:
As i know, changes of bindable property does not queueing. it is immediately  updated

paymicro:
@myroot unfortunately it is not
https://github.com/xamarin/Xamarin.Forms/blob/1df63c27b9892ce69bf26dadda275b2f7f94772e/Xamarin.Forms.Core/BindableObject.cs#L412

myroot:
@paymicro  Yes, you are right

Xamarin.Forms.Core/Picker.cs
```diff
@@ -150,8 +150,10 @@ void OnItemDisplayBindingChanged(BindingBase oldValue, BindingBase newValue)
 
 		void OnItemsCollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
 		{
-			SelectedIndex = SelectedIndex.Clamp(-1, Items.Count - 1);
-			UpdateSelectedItem();
```
StephaneDelcroix: if the old and the new selectedIndex are different, setting the `SelectedIndex` will trigger the UpdateSelectedItem at the right time.

Xamarin.Forms.Core/Picker.cs
```diff
@@ -150,8 +150,10 @@ void OnItemDisplayBindingChanged(BindingBase oldValue, BindingBase newValue)
...
+			// Changing the SelectedIndex property is sent to a queue that is not instantly executed.
+			// In the UpdateSelectedItem method, we need the actual selectedIndex.
+			var selectedIndex = SelectedIndex = SelectedIndex.Clamp(-1, Items.Count - 1);
+			UpdateSelectedItem(selectedIndex);
```
StephaneDelcroix: so you only have to invoke `UpdateSelectedIndex()` when the source the index hasn't been changed

Xamarin.Forms.Core/Picker.cs
```diff
@@ -239,19 +241,20 @@ void UpdateSelectedIndex(object selectedItem)
 			SelectedIndex = Items.IndexOf(selectedItem);
 		}
 
-		void UpdateSelectedItem()
+		void UpdateSelectedItem(int? actualIndex = null)
```
StephaneDelcroix: I prefer having a mandatory index, instead of an optional

rmarinho:
@paymicro can we retarget this to 3.1.0? thanks

PR #2836, opened 2018-05-26
Use GestureDetector constructor that takes in Context since the one without it is going to be deprecated

PureWeen:
### Description of Change ###

The constructor for *GestureDetector* that doesn't take Context is going to be deprecated. This switches to a constructor that won't be

### PR Checklist ###


PureWeen:
build --uitests

PR #2857, opened 2018-05-29
[Android] Keep track of created ConditionalFocusLayouts so they can all be disposed of

### LINKED ISSUE #2829: [Android] Renderers associated with ListView cells are occasionaly not being disposed of which causes left over events to propagate to disposed views
### Description
A number of issues have popped up related to the renderers of a ListView not all correctly disposing. The use cases are generally the same
- Setup list view for recycling
- Navigate to a page with the listview
- scroll some
- pop the page off
- do this over and over again and at some point an exception will occur related to dispose

This issue is meant to be the root issue for the set of issues people have logged for this problem.

My current theory on this goes as follows 

In certain cases when you scroll a list view and it creates a new view using view recycling this would cause a view to become non visible but still exist in memory.

Because of this when iterating with GetChildAt you would be returned NULL for a given position because that view isn't visible though it and its resources still exist in memory. 

What this means is that DisposeCells now will not be able to access said View 

```C#
void DisposeCells()
		{
			var cellCount = _realListView?.ChildCount ?? 0;
			for (int i = 0; i < cellCount; i++)
			{
				var layout = _realListView.GetChildAt(i) as ConditionalFocusLayout;

				// Headers and footers will be skipped. They are disposed elsewhere.
				if (layout == null || layout.IsDisposed())
					continue;
```

### Steps to Reproduce

https://github.com/xamarin/Xamarin.Forms/issues/2746
https://github.com/xamarin/Xamarin.Forms/issues/2004
https://github.com/xamarin/Xamarin.Forms/issues/2421
https://github.com/xamarin/Xamarin.Forms/issues/2406
https://github.com/xamarin/Xamarin.Forms/issues/2041
https://github.com/xamarin/Xamarin.Forms/issues/2317

### Proposed Fix
https://github.com/xamarin/Xamarin.Forms/tree/fix-gh2829

### Expected Behavior
All created views should be disposed of

### Actual Behavior
Occasionally a non visible view is not recycled

### Basic Information

- Version with issue: 3.0

### Resolution 
I have a branch https://github.com/xamarin/Xamarin.Forms/tree/fix-gh2829  where I've been messing around with some thoughts

PureWeen:
### Description of Change

Currently _DisposeCells_ only iterates using GetChildAt which will only return visible cells. When you would scroll a listview this would cause a cell to not be visible and would thus return NULL for GetChildAt but the ConditionalFocusLayout would still exist in memory. This fix keeps track of created Layouts in an internal list to ensure that all resources get disposed of

I can't say for sure if this is the only cause of https://github.com/xamarin/Xamarin.Forms/issues/2829 but it definitely fixes a case where Dispose wasn't being called for a given cell.  

### Bugs Fixed

fixes #2829

### PR Checklist

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/ListViewAdapter.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ListViewAdapter.cs
@@ -27,6 +27,7 @@ namespace Xamarin.Forms.Platform.Android
 		protected readonly ListView _listView;
 		readonly AListView _realListView;
 		readonly Dictionary<DataTemplate, int> _templateToId = new Dictionary<DataTemplate, int>();
+		readonly List<ConditionalFocusLayout> _layoutsCreated = new List<ConditionalFocusLayout>();
 		int _dataTemplateIncrementer = 2; // lets start at not 0 because ... 
 
 		// We will use _dataTemplateIncrementer to get the proper ViewType key for the item's DataTemplate and store these keys in  _templateToId.
@@ -181,7 +182,7 @@ namespace Xamarin.Forms.Platform.Android
 				_templateToId[itemTemplate] = key;
 			}
 
-			if (key >= ViewTypeCount) 
+			if (key >= ViewTypeCount)
 			{
 				throw new Exception($"ItemTemplate count has exceeded the limit of {ViewTypeCount}" + Environment.NewLine +
 									 "Please make sure to reuse DataTemplate objects");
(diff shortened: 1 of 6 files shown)


Xamarin.Forms.Platform.Android/Renderers/ListViewAdapter.cs
```diff
@@ -474,25 +501,19 @@ void DisposeCells()
...
-
-					renderer?.Dispose();
-					renderer = null;
-				}
+				DisposeOfConditionalFocusLayout(layout);
+			}
 
-				renderedView?.Dispose();
-				renderedView = null;
+			// Having just this loop would probably be sufficient but
```
hartez: It looks like this covers everything the previous loop does (and more); let's just loop once. It'll make the code easier to understand.

PureWeen:
build --uitests

Xamarin.Forms.Platform.Android/Renderers/ListViewAdapter.cs
PureWeen: @hartez fixed

PR #2876, opened 2018-05-30
[UWP,iOS] Enable screen readers to read non-interactive elements

### LINKED ISSUE #1946: the screen reader doesn't read text on screen like label's and other images, It only reads interactive elements like buttons.
### Steps to Reproduce

1. Create a label and provide AutomationProperties
2. Start the narrator for Windows 10 or JAWS 18
3. Run the project on Local Machine.

### Expected Behavior
The narrator should read the text written in the Label, Image or any other non interactive element if AutomationProperties are set in that control.

### Actual Behavior
The narrator is ommiting the text written in the Label, Image or any other non interactive element.

### Basic Information

- Version with issue: 2.5.0.280555, 2.3.4.231, 2.4.0.282 (checked with these three versions)
- Last known good version: none of the provided stable version are working with narrator
- IDE: Visual Studio 2015 update 3
- Platform Target Frameworks: Currently working with UWP only
  - UWP:  Windows  App Target vesion Windows 10(10.0; build 10240)
- Nuget Packages: Xamarin Forms

###Code Snippet used is attached in the file attached
[Accessibility.txt](https://github.com/xamarin/Xamarin.Forms/files/1746927/Accessibility.txt)

samhouts:
### Description of Change ###

We expect screen readers to be able to read content by default. On iOS and UWP, this was not true for non-interactive elements such as Labels.

#### iOS
We create iOS views programmatically, in which case the `IsAccessibilityElement` value defaults to `false`. We expected it to default to `true`, as it states in the documentation, but this apparently is only the case when using storyboards. We are now explicitly setting `IsAccessibilityElement` to true for all controls.

#### UWP
This functionality was originally working, but it was thwarted by a change made later for UI testing. Reverting that change for now until we can find a better alternative.

### Bugs Fixed ###

- fixes #1946

### API Changes ###

None

### Behavioral Changes ###

- UWP UI Tests will probably fail
- Anyone explicitly setting `UIView.IsAccessibilityElement` to `false` in custom renderers should ensure they do so after `SetNativeControl` is called. Otherwise, it will be overridden to `true`.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/LabelRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/LabelRenderer.cs
@@ -35,16 +35,20 @@ namespace Xamarin.Forms.Platform.UWP
 		SizeRequest _perfectSize;
 		bool _perfectSizeValid;
 
-		protected override AutomationPeer OnCreateAutomationPeer()
-		{
-			// We need an automation peer so we can interact with this in automated tests
-			if (Control == null)
-			{
-				return new FrameworkElementAutomationPeer(this);
-			}
+		//TODO: We need to revisit this later when we complete the UI Tests for UWP.
+		// Changing the AutomationPeer here prevents the Narrator from functioning properly.
+		// Oddly, it affects more than just the TextBlocks. It seems to break the entire scan mode.
 
-			return new FrameworkElementAutomationPeer(Control);
-		}
+		//protected override AutomationPeer OnCreateAutomationPeer()
+		//{
+		//	// We need an automation peer so we can interact with this in automated tests
+		//	if (Control == null)
+		//	{
+		//		return new FrameworkElementAutomationPeer(this);
+		//	}
+
+		//	return new TextBlockAutomationPeer(Control);
+		//}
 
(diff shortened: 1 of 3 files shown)


samhouts:
build --uitests

Xamarin.Forms.Platform.UAP/LabelRenderer.cs
```diff
@@ -35,16 +35,20 @@ public class LabelRenderer : ViewRenderer<Label, TextBlock>
...
-		{
-			// We need an automation peer so we can interact with this in automated tests
-			if (Control == null)
-			{
-				return new FrameworkElementAutomationPeer(this);
-			}
-
-			return new FrameworkElementAutomationPeer(Control);
-		}
+		//TODO: We need to revisit this later when we complete the UI Tests for UWP.
```
hartez: Do we have an issue in GH for this problem (so we can track it and someone will address it later)?

samhouts: https://github.com/xamarin/Xamarin.Forms/issues/3022

Issue #2011, opened 2018-03-05
[Android]Null exception throw from ScrollViewRenderer.Dispose(bool disposing) in Xamarin.Forms Android


jessejiang0214:
### Steps to Reproduce

1. Put Scrollview into ListView header
2. Set Scrollview.Orientation="Horizontal" and add some controls into the scrollview
3. pop the page

### Expected Behavior
No exception
### Actual Behavior
Throw null exception, If I catch the null exception from override  Dispose function, it will throw other exception from RemoveAllViews. I found the source code of ScrollViewRenderer 
```
		protected override void Dispose(bool disposing)
		{
			base.Dispose(disposing);

			SetElement(null);

			if (disposing)
			{
				Tracker.Dispose();
				Tracker = null;
				RemoveAllViews();
				_container.Dispose();
				_container = null;
			}
		}
```
It didn't check if Tracker is null.
### Basic Information

- Version with issue: 2.5.0.121934
- Last known good version:
- IDE: Visual Studio for Mac 7.3.3
- Platform Target Frameworks: 

Xamarin.Android
Version: 8.1.5.0 (Visual Studio Community)
Android SDK: /Users/jesse/Library/Developer/Xamarin/android-sdk-macosx
	Supported Android versions:
		6.0 (API level 23)
		7.0 (API level 24)
		7.1 (API level 25)
		8.0 (API level 26)

SDK Tools Version: 26.1.1
SDK Platform Tools Version: 27.0.1
SDK Build Tools Version: 26.0.3

- Nuget Packages: N/A
- Affected Devices: All

hartez:
[Repro](https://github.com/xamarin/Xamarin.Forms/files/1783398/GitHub2011.Repro.zip)

PureWeen:
Bugzilla32462 in the control gallery

rmarinho:
ok I got it we have a fix #3025

PR #2904, opened 2018-06-02
[Android] ignore bugzilla test crashing UI Tests until related issue gets resolved

PureWeen:
### Description of Change ###
bugzilla32462 is currently crashing the UI Test runner. Unfortunately it's testing a regression that was missed but has been regressed for some time and already lives in the wild.

Here's a current issue a user logged related to this bug
https://github.com/xamarin/Xamarin.Forms/issues/2011

Once that bug has been prioritized then we can re-fix but for now it's not providing any benefit to have this test crashing UI Tests

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla32462.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla32462.cs
@@ -49,6 +49,7 @@ namespace Xamarin.Forms.Controls.Issues
 
 #if UITEST
 		[Test]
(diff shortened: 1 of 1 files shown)


PureWeen:
build --uitests

PureWeen:
build --uitests

rmarinho:
I don't think this fixes the issue we need to fix.

rmarinho:
what API was this crashing ?

PureWeen:
It would just completely crash when you navigated away from the page with the scrollview inside the listview header

But it looks like https://github.com/xamarin/Xamarin.Forms/issues/2011 is fixed. If that's the case then I'm just going to close this PR out

PR #2911, opened 2018-06-03
[Core] Report binding failures

### LINKED ISSUE #2639: [Enhancement] Report Binding failures
# Rationale

In Xamarin.Forms, when a {Binding} expression fails, it fails silently, and at runtime. The only indication that the binding has failed is that nothing appears in the UI. Debugging this can be difficult--is it a BindingContext issue? A typo in my property name? Did I accidentally define my property as a member variable? Etc.

# Suggestion

Log binding failures similarly to the way that WPF and UWP do:

>System.Windows.Data Error: 40 : BindingExpression path error: 'NonExistingProperty' property not found on 'object' ''Grid' (Name='pnlMain')'. BindingExpression:Path=NonExistingProperty; DataItem='Grid' (Name='pnlMain'); target element is 'TextBlock' (Name=''); target property is 'Text' (type 'String') 

and print them out to the Debug log.

\~EDIT\~: Looking at [BindingExpression.cs](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/BindingExpression.cs) it looks like the logging code is there, but it never gets printed out to my Debug or Output windows. Could it be a configuration issue on my end?

Running Windows 10, Visual Studio 2017 15.6.7, reproducing this in XF 2.5.1.444934.

pingzing:
### Description of Change ###

 A lot of logging happens internally, for example when a binding fails. In order to make these logs more accessible to the end-user, the `Application.LogWarningsToApplicationOutput` property was introduced. WIth this, logging will also be outputted to the debug console.

### Bugs Fixed ###

- fixes #2639

### API Changes ###

Added: 
```csharp
public static bool Application.LogWarningsToApplicationOutput { get; set; }
```

### Behavioral Changes ###

When `Application.LogWarningsToApplicationOutput` is set to true, a `LogListener` is attached to the internal logging system which outputs the messages through a `Debug.Writeline`. When set to false the `LogListener` is removed again.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Application.cs
+++ b/Xamarin.Forms.Core/Application.cs
@@ -6,6 +6,7 @@ using System.ComponentModel;
 using System.Threading.Tasks;
 using Xamarin.Forms.Internals;
 using Xamarin.Forms.Platform;
+using System.Diagnostics;
 
 namespace Xamarin.Forms
 {
@@ -23,6 +24,27 @@ namespace Xamarin.Forms
 
 		static SemaphoreSlim SaveSemaphore = new SemaphoreSlim(1, 1);
 
+		static Lazy<DelegateLogListener> _applicationOutputListener;
+		static bool _logWarningsToApplicationOutput;
+
+		public static bool LogWarningsToApplicationOutput
+		{
+			get => _logWarningsToApplicationOutput;
+			set
+			{
+				_logWarningsToApplicationOutput = value;
+
+				if (_logWarningsToApplicationOutput)
+				{
+					Log.Listeners.Add(_applicationOutputListener.Value);
+				}
+				else
+				{
+					Log.Listeners.Remove(_applicationOutputListener.Value);
+				}
+			}
+		}
+
 		public Application()
 		{
 			var f = false;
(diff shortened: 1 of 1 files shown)


jfversluis:
Thanks for the extensive feedback @StephaneDelcroix!

Do you have any suggestion on where the public API might live? Is there already a good place to put this?

StephaneDelcroix:
a static property in `Application` class, maybe ?

so you'd have this in your `App.xaml.cs` code behind:

```csharp
public App() {
#if DEBUG
    Application.LogWarningsToApplicationOutput |= true;
#endif
    InitializeComponent();
}
```

But that could be in `Device` as well, and feel free to pick up a better name for the property.

@jassmith thoughts ^^ ?

jfversluis:
Sounds good to me, I feel `Application` would make more sense than `Device` since it applies to the application lifecycle and doesn't really tell you anything about the device.

jfversluis:
@StephaneDelcroix see changes, more like this then?

StephaneDelcroix:
@jfversluis please update PR description to match the new implementation choices

jfversluis:
As much as I want to, I didn't open the initial PR :) @pingzing ?

jfversluis:
I'd propose this, maybe you could replace it @StephaneDelcroix 

### Description of Change ###

In regard to [issue 2639](https://github.com/xamarin/Xamarin.Forms/issues/2639). A lot of logging happens internally, for example when a binding fails. In order to make these logs more accessible to the end-user, the `Application.LogWarningsToApplicationOutput` property was introduced. WIth this, logging will also be outputted to the debug console.

### Bugs Fixed ###

- N/A

### API Changes ###

Added: 
```csharp
public static bool Application.LogWarningsToApplicationOutput
```

### Behavioral Changes ###

When `Application.LogWarningsToApplicationOutput` is set to true, a `LogListener` is attached to the internal logging system which outputs the messages through a `Debug.Writeline`. When set to false the `LogListener` is removed again.

### PR Checklist ###

jfversluis:
@StephaneDelcroix what needs to happen to get this merged? 🙂

PR #2917, opened 2018-06-03
[WPF] Fixed Images from Stream

### LINKED ISSUE #2844: WPF: Image from stream doesn't work
### Description
When trying to using a byte array (which becomes a MemoryStream) as the source of an Image, the Image does not render.

### Steps to Reproduce

1. Load a byte array from a real image. 
2. Turn that byte array into a MemoryStream
3. Turn that into an ImageSource with ImageSource.FromStream(() => memStream)
4. Use that ImageSource on an Image control
5. Image control renders as blank

### Expected Behavior

Image control should render the original image the byte array came from

### Basic Information

I have already researched and fixed the problem, but since I am not experienced in WPF, I am not confident in my fix.  Therefore, I will show my fix here and you guys can put in the correct fix.

The problem is in the ImageRenderer for the WPF platform in the StreamImageSourceHandler class.  It does this:

				using (Stream stream = await ((IStreamImageSource)streamsource).GetStreamAsync(cancelationToken))
				{
					bitmapimage = new BitmapImage()
					{
						StreamSource = stream
					};
				}

I am guessing there are two things wrong here.  The first is that the using statement will GC the stream which we put inside the bitmapimage so the StreamSource of the bitmapimage becomes unusable after it leaves the using block.  The second thing wrong is that you can't just set the StreamSource.  You have to use BeginInit and EndInit (just like you already do for the UriImageSourceHandler in the same source file).  So my fix is the following:

                Stream stream = await ((IStreamImageSource)streamsource).GetStreamAsync(cancelationToken);
                bitmapimage = new BitmapImage();
                bitmapimage.BeginInit();
                bitmapimage.StreamSource = stream;
                bitmapimage.EndInit();

This fixes the rendering of the Image, but perhaps there is a memory leak with the Stream??  Please investigate and commit a proper fix.  Thanks.

jsuarezruiz:
### Description of Change ###

Fixed Images from Stream.
imagen
_NOTE: Image from base64 to byte array, then use a Stream to create the ImageSource._

### Bugs Fixed ###

- fixes #2844

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.WPF/Renderers/ImageRenderer.cs
+++ b/Xamarin.Forms.Platform.WPF/Renderers/ImageRenderer.cs
@@ -141,19 +141,19 @@ namespace Xamarin.Forms.Platform.WPF
 	{
 		public async Task<System.Windows.Media.ImageSource> LoadImageAsync(ImageSource imagesource, CancellationToken cancelationToken = new CancellationToken())
 		{
-			BitmapImage bitmapimage = null;
-			StreamImageSource streamsource = imagesource as StreamImageSource;
-			if (streamsource != null && streamsource.Stream != null)
+			BitmapImage bitmapImage = null;
+
+			if (imagesource is StreamImageSource streamsource && streamsource.Stream != null)
 			{
-				using (Stream stream = await ((IStreamImageSource)streamsource).GetStreamAsync(cancelationToken))
-				{
-					bitmapimage = new BitmapImage()
-					{
-						StreamSource = stream
-					};
-				}
+				Stream stream = await ((IStreamImageSource)streamsource).GetStreamAsync(cancelationToken);
+				bitmapImage = new BitmapImage();
+				bitmapImage.BeginInit();
+				bitmapImage.StreamSource = stream;
+				bitmapImage.CacheOption = BitmapCacheOption.OnLoad;
+				bitmapImage.EndInit();
 			}
-			return bitmapimage;
+
+			return bitmapImage;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.WPF/Renderers/ImageRenderer.cs
```diff
@@ -141,19 +141,19 @@ public sealed class StreamImageSourceHandler : IImageSourceHandler
...
+			if (imagesource is StreamImageSource streamsource && streamsource.Stream != null)
 			{
-				using (Stream stream = await ((IStreamImageSource)streamsource).GetStreamAsync(cancelationToken))
-				{
-					bitmapimage = new BitmapImage()
-					{
-						StreamSource = stream
-					};
-				}
+				Stream stream = await ((IStreamImageSource)streamsource).GetStreamAsync(cancelationToken);
```
VladislavAntonyuk: why do you remove disposing stream. do you need it anywhere else?

legistek: Agreed. Using `BitmapCacheOption.OnLoad` is correct, but then the `Stream` should be disposed. That's the point of using that option - so that you can dispose the stream after the bitmap is created.

jsuarezruiz: Yes, modified. Thanks!

jsuarezruiz:
You are right @rmarinho, changed!

RogerHardiman:
Thank you for the fix

PR #2924, opened 2018-06-04
[UWP] Fixes of calculation of heights in the MasterDetailPage

### LINKED ISSUE #1648: MasterDetailPage throws ArgumentOutOfRangeException
### Description
When adding a MasterDetailPage inside a NavigationPage, it throws an Exception upon loading

### Steps to Reproduce
Start the modified version of the official example at:
https://github.com/TiBall/xamarin-forms-samples/tree/MasterDetailNavigationException/Navigation/MasterDetailPage

### Expected Behavior
MasterDetailPage renders with no Exception

### Actual Behavior
Application terminates because of this error inside the Xamarin Control
"System.ArgumentOutOfRangeException: Non-negative number required.\r\… 
…nParameter name: height\r\n at Windows.Foundation.Size..ctor(Double width, Double height)\r\n at Xamarin.Forms.Platform.UWP.MasterDetailControl.get_MasterSize()\r\n at Xamarin.Forms.Platform.UWP.MasterDetailPageRenderer.UpdateBounds()\r\n at Xamarin.Forms.Platform.UWP.MasterDetailPageRenderer.OnControlLoaded(Object sender, RoutedEventArgs routedEventArgs)"

### Basic Information

- Version with issue:
- Last known good version:
- IDE: VS 2017 v 15.5.3
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:  10.0.10586.0 and higher
- Nuget Packages: "Xamarin.Forms": "2.5.0.91635"
- Affected Devices: Windows

### Reproduction Link
https://github.com/TiBall/xamarin-forms-samples/commit/f79e75114e970f07ddcf7dfc5df85571c48af67b

paymicro:
### Description of Change ###

Added zero `height` check before subtraction from it.

screenshot_3

### Bugs Fixed ###

- Fixes #1648

### API Changes ###

/

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/MasterDetailControl.cs
+++ b/Xamarin.Forms.Platform.UAP/MasterDetailControl.cs
@@ -89,7 +89,7 @@ namespace Xamarin.Forms.Platform.UWP
 				double height = ActualHeight;
 				double width = ActualWidth;
 
-				if (_commandBar != null)
+				if (_commandBar != null && height != 0)
 					height -= _commandBar.ActualHeight;
 
 				if (ShouldShowSplitMode && IsPaneOpen)
@@ -100,7 +100,7 @@ namespace Xamarin.Forms.Platform.UWP
 						width -= _masterPresenter.ActualWidth;
 				}
 
-				return new Windows.Foundation.Size(width >= 0 ? width : 0, height);
+				return new Windows.Foundation.Size(width > 0 ? width : 0, height);
 			}
 		}
 
@@ -143,7 +143,7 @@ namespace Xamarin.Forms.Platform.UWP
 
 				// On first load, the _commandBar will still occupy space by the time this is called.
 				// Check ShouldShowToolbar to make sure the _commandBar will still be there on render.
-				if (_firstLoad && _commandBar != null && ShouldShowToolbar)
+				if (_firstLoad && height != 0 && _commandBar != null && ShouldShowToolbar)
 				{
 					height -= _commandBar.ActualHeight;
 					_firstLoad = false;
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.UAP/MasterDetailControl.cs
```diff
@@ -100,7 +100,7 @@ public Windows.Foundation.Size DetailSize
 						width -= _masterPresenter.ActualWidth;
 				}
 
-				return new Windows.Foundation.Size(width >= 0 ? width : 0, height);
+				return new Windows.Foundation.Size(width > 0 ? width : 0, height);
```
hartez: This could still fail if `ActualHeight` is between zero and `_commandBar.ActualHeight`; height could end up being less than zero. 

What if we reverted line 92 and in 103 did 
`return new Windows.Foundation.Size(width > 0 ? width : 0, height > 0 ? height : 0)`
instead? (Or used `Math.Max(height, 0)`.)

Xamarin.Forms.Platform.UAP/MasterDetailControl.cs
```diff
@@ -143,7 +143,7 @@ public Windows.Foundation.Size MasterSize
 
 				// On first load, the _commandBar will still occupy space by the time this is called.
 				// Check ShouldShowToolbar to make sure the _commandBar will still be there on render.
-				if (_firstLoad && _commandBar != null && ShouldShowToolbar)
+				if (_firstLoad && height != 0 && _commandBar != null && ShouldShowToolbar)
```
hartez: If `height` is zero, then the `_firstLoad` flag won't ever be set to `false`. 
Instead of checking for a zero height here, could we just constrain height to a minimum of zero on line 157 below?

Issue #2943, opened 2018-06-05
[UWP] Last element in grid is rendered with lowest Z-Index


activa:
### Description

PR #2762 breaks rendering of Grid elements. The last element in a grid is rendered with the lowest Z-Index and disappears under all other elements. The last element should have the highest Z-index

This is only broken on UWP. Works fine on Android and iOS

### Steps to Reproduce

1. Add 2 overlapping UI elements to a grid

### Expected Behavior

The last UI element is rendered on top of the first

### Actual Behavior

The first UI element is rendered on top of the second

### Basic Information

- Version with issue:  3.0.0.5308
- Last known good version: 3.0.0.4825
- Affected Devices: UWP

### Screenshots

![image](https://user-images.githubusercontent.com/650710/40998362-302c9e94-68cd-11e8-8e69-b205325c3f96.png)

### Reproduction Link

[UWPGridZBug.zip](https://github.com/xamarin/Xamarin.Forms/files/2073714/UWPGridZBug.zip)

activa:
Can this please get the attention and priority it deserves? This regression bug completely breaks the majority of UWP apps that were built with Xamarin Forms. I would go as far as to suggest pulling both versions 3.0.0.530938 and 3.0.0.550146 from the Nuget repository.

PR #2762, opened 2018-05-18
fixes #2653 on uwp set the Z Index when children aren't added to the end of the stack

PureWeen:
### Description of Change ###

Changes were made in this PR https://github.com/xamarin/Xamarin.Forms/pull/1196 so that less calls were made to set the ZIndex on UWP. This only took into account when RaiseUp and Down were called and didn't factor in when a child element is inserted out of order

### Bugs Fixed ###

Fixes #2653

### Behavioral Changes ###
None

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #1196, opened 2017-10-11
[UWP] Optimized ZIndexing for a thirty-nine percent speed boot on loa…

BradChase2011:
…ding views.

### Description of Change ###

Setting the Z-Index of UI Elements in UWP can be very costly.  I tracked a 39% improvement in speed loading views.  This will only set the Z-Index when the developer explicitly raises or lowers an element rather than every time any element is added to the tree and re-setting all its siblings.

### Bugs Fixed ###

https://bugzilla.xamarin.com/show_bug.cgi?id=60143

### API Changes ###

None.

### Behavioral Changes ###

Views by my calculations (stopwatch) are now loading 39% faster.

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #2945, opened 2018-06-05
Fix issue #2943. [UWP] Set correct Z-Index for last element added to …

activa:
Fix issue #2943 - Set correct Z-Index for last element added to layout

### Description of Change ###

For some reason, the last added element to a layout on UWP would not have the Z-Index set. The result was that the last UI element was rendered **under** all other elements instead of on top.

In PR #2762, a test was done to see if the element being added was the last element, and if it was, the Z-Index recalculation was not performed, which is incorrect. This PR removes that test. It's unclear why that test was there in the first place.

### Bugs Fixed ###

- #2943

### API Changes ###

None

### Behavioral Changes ###

None
### PR Checklist ###

There's no apparent way to test the Z-Index of inserted elements so no test was written.

### DIFF
--- a/Xamarin.Forms.Platform.UAP/VisualElementPackager.cs
+++ b/Xamarin.Forms.Platform.UAP/VisualElementPackager.cs
@@ -119,8 +119,7 @@ namespace Xamarin.Forms.Platform.UWP
 
 			_panel.Children.Add(childRenderer.ContainerElement);
 
-			if (ElementController.LogicalChildren[ElementController.LogicalChildren.Count - 1] != view)
-				EnsureZIndex();
+			EnsureZIndex();
 		}
 
 		void OnChildRemoved(object sender, ElementEventArgs e)
--- a/Xamarin.Forms.Platform.UAP/VisualElementPackager.cs
+++ b/Xamarin.Forms.Platform.UAP/VisualElementPackager.cs
@@ -15,6 +15,7 @@ namespace Xamarin.Forms.Platform.UWP
 		readonly int _rowSpan;
 		bool _disposed;
 		bool _isLoaded;
+		bool _hasZIndexes;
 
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.UAP/VisualElementPackager.cs
```diff
@@ -119,8 +119,7 @@ void OnChildAdded(object sender, ElementEventArgs e)
 
 			_panel.Children.Add(childRenderer.ContainerElement);
 
-			if (ElementController.LogicalChildren[ElementController.LogicalChildren.Count - 1] != view)
-				EnsureZIndex();
+			EnsureZIndex();
```
PureWeen: We don't want to just update the ZIndex's without cause because that will unwind this PR
https://github.com/xamarin/Xamarin.Forms/pull/1196

Which introduced the initial issue that I was fixing. The ideal scenario is that Z-Index is just not set at all unless the user inserts a view somewhere else in the stack or things get moved around (post renderer)

I'm still looking into this one as well because ideally EnzureZIndex isn't called unless it has to be. If that makes sense

activa: Calling EnsureZIndex() will not break the earlier fix because the child elements are already in the correct order at the end of the ChildAdded() event handler.

I'd like to stress that this regression bug breaks **every single UWP app** that uses Xamarin Forms. Apps assume that the last element is always rendered on top of everything else and currently it's the easiest (only?) way to implement popup-like views.

activa:
I've added an extra check to make sure we only call EnsureZIndex when it's necessary. 

It is necessary when inserting an element (not at the end) or when the existing UI elements were already assigned a positive Z-Index

PR #2961, opened 2018-06-06
 Implemented colored refresh indicator for ListView pull-to-refresh

jfversluis:
### Description of Change ###

For a project, I needed to color the spinner which is shown when you pull to refresh on a `ListView`. To my big surprise, this isn't available by default. So, I thought I would add it since I could find any issue or PR for it.

Implemented for iOS and Android only. For other platforms, I did not see a loading indicator at all, or not one that could be colored.

No tests were added since it's a visual only thing.

### Bugs Fixed ###

- N/A

### API Changes ###

Added: 
```csharp
public static readonly BindableProperty ListView.RefreshControlColorProperty;
public Color ListView.RefreshControlColor { get; set; }
```

### Behavioral Changes ###

By default the color of the spinner will be black as is it now, but by setting the new `SpinnerColor` property you can influence the color of the spinner.

### PR Checklist ###


Xamarin.Forms.Core/ListView.cs
```diff
@@ -46,6 +46,8 @@ public class ListView : ItemsView<Cell>, IListViewController, IElementConfigurat
 
 		public static readonly BindableProperty SeparatorColorProperty = BindableProperty.Create("SeparatorColor", typeof(Color), typeof(ListView), Color.Default);
 
+		public static readonly BindableProperty SpinnerColorProperty = BindableProperty.Create(nameof(SpinnerColor), typeof(Color), typeof(ListView), Color.Black);
```
rmarinho: I think you should use the `Color.Default`and in the renderers you need to capture the default color of the spinner when they are initialised , so if the user changes back to the default color you can set the default color of the platform.

Xamarin.Forms.Core/ListView.cs
```diff
@@ -217,6 +219,12 @@ public Color SeparatorColor
 			set { SetValue(SeparatorColorProperty, value); }
 		}
 
+		public Color SpinnerColor
```
rmarinho: RefreshControlColor?  because it could be something different from a spinner in a specific platform.

Xamarin.Forms.Core/ListView.cs
jfversluis: I did `Color.Default`, but then the spinner disappeared for some weird reason, at least on iOS. I'll check it again.

Xamarin.Forms.Core/ListView.cs
jfversluis: Hm makes sense. I was struggling with the name. RefreshControlColor seems to cover it.

jfversluis:
@rmarinho updated naming. `Color.Default` reinstated on the `ListView`, but also added the default black color in the iOS `UpdateRefreshControlColor` method. Seeing this also happens for the [separator color](https://github.com/xamarin/Xamarin.Forms/pull/2961/files#diff-198607f4ce97efcc515a271473524b08L646) this is something that is necessary. If not, the spinner is invisible.

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -673,6 +676,13 @@ void UpdateSelectionMode()
 			}
 		}
 
+		void UpdateSpinnerColor()
+		{
+			var color = Element.RefreshControlColor;
+
+			if (_tableViewController != null)
+				_tableViewController.UpdateRefreshControlColor(color.ToUIColor(UIColor.Black));
```
rmarinho: Instead of `UIColor.Black` you need to find someway to find the initial color of the `RefreshControl`, then on this method when `color == Color.Default` you call `UpdateRefreshControlColor` with that color that you capture initially as the default color of the spinner. If you use Color.Default it's transparent so the spinner disappears, you need to capture the native initial color.

jfversluis: I think I see your point. According to this link, it is nil as a default: https://developer.apple.com/documentation/uikit/uirefreshcontrol/1624847-tintcolor so would that make sense then? Just tested it and set the tint color to null (when the property is Color.Default) makes the spinner black.

I should double check how it is on Android.

jfversluis:
@rmarinho added some more changes for using the default color on iOS

rmarinho:
@jfversluis can you rebase please so I can run uitests on this . thanks

jfversluis:
@rmarinho done and generated documentation for the `ListView`

jfversluis:
@rmarinho where are we on this? 🙂

rookiejava:
It seems to be limited to android and iOS only. Why don't you use PlatformSpecific?

jfversluis:
The platform-specifics are usually when something is really tied to one platform. But to be honest I didn't really think about it and thought that I could implement it on more platforms. I still don't rule that out for the future.

jfversluis:
How are we here?

samhouts:
build --uitests

jfversluis:
Resolved merge conflicts with master

rmarinho:
build --uitests

PR #2964, opened 2018-06-06
Use UpdateOnTap setting for Slider to fix broken gesture bubbling test

hartez:
### Description of Change ###

Sets UpdateOnTap to true for the Slider in the gesture bubbling tests (on iOS).

Also fixes a bug where the initial setting for UpdateOnTap is ignored by the SliderRenderer.


rmarinho:
We need something similar to 3.0.0 but the PS isn't there

PR #2976, opened 2018-06-08
[UWP] Fix listview not rendering when ItemsSource is initially null

activa:
### Description of Change ###

Fix listview not rendering when ItemsSource is initially null

When ItemsSource is initially null, the DataContext is not set, and it is never set when the ItemsSource property is set later. This fix will call ReloadData() when the ItemsSource property is changed on the XF ListView control.

### Bugs Fixed ###

No bug reported yet

### PR Checklist ###


mattrichnz:
@activa Please see issue reported on #2981 and repro project.

PR #2999, opened 2018-06-09
[C] set the conf registry from all ctors

### LINKED ISSUE #2918: Stepper alternate constructor NOT calling this() to initialize platformConfigurationRegistry
If you look at Slider, it has 2 constructors, the default with no arguments and an alternate with min, max, and value.  The default constructor initializes _platformConfigurationRegistry and the alternate constructor extends the default with ": this()".  This way the alternate constructor will still initialize the _platformConfigurationRegistry.

Stepper is NOT doing this.  The alternate constructor does NOT call this() so the _platformConfigurationRegistry is NOT being initialized so any app using the alternate constructor will crash (at least on Android) with NPE.

StephaneDelcroix:
### Description of Change ###

[C] set the conf registry from all ctors

### Bugs Fixed ###

- fixes #2918

### API Changes ###

/

### Behavioral Changes ###

/

### PR Checklist ###


StephaneDelcroix:
dupe of #2994

PR #3011, opened 2018-06-12
[macOS] Entry focus unfocus events

### LINKED ISSUE #3012: macOS Entry Focused/Unfocused fires immediately with click
### Description

When an Entry is clicked, the Focused event is invoked and is immediately followed by the Unfocused event being invoked. 

### Steps to Reproduce

1. Add an Entry to a page and add methods to be called on Focused and Unfocused events of the Entry.
2. Click entry
3. Confirm each event is invoked.

### Expected Behavior
The Focused event should be invoked on first click and when the element was previously focused when the window loses key status and becomes key again.
The Unfocused event should be invoked only when editing has completed, another element / entry gains focus, or when the window loses focus (key status).

### Actual Behavior
The Focused and Unfocused events invoke immediately when the Entry is clicked, and Focused is invoked again once the first character is typed. Neither Focused nor Unfocused get invoked when window focus changes.

### Basic Information

- Version with issue: 
- Last known good version: NA
- IDE: Visual Studio for Mac Version 7.5.2 (build 40)
- Platform Target Frameworks: 
 - macOS: building on 10.13.5 targeting 10.10
- Nuget Packages:
- Affected Devices: MacBook Pro (13-inch, 2017, Two Thunderbolt 3 ports)

### Reproduction Link

The effect can be seen in the Xamarin.Forms.ControlGallery Entry section

jvansickle:
### Description of Change ###

Improve the conditions which an Entry's Focused and Unfocused events fire running on macOS. Previously when the Entry element was clicked, the Focused event would be invoked, and immediately be followed by the Unfocused event being invoked. This was a result of the macOS FirstResponder quirks with NSTextField. 

Reference https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/EventHandlingBasics/EventHandlingBasics.html#//apple_ref/doc/uid/10000060i-CH5-SW23 

### Bugs Fixed ###

- fixes #3012 

### API Changes ###

None

### Behavioral Changes ###

The Entry element on macOS will no longer invoke Focused and Unfocused events when the Entry is clicked. 

The Entry's Focused event is now invoked only when the element obtains FirstResponder or when the Entry was previously focused when the window lost key status and the window has since regained key status.

The Entry's Unfocused event is now invoked when the native NSTextField's DidEndEditing method is called, or when the window loses key status. 

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.MacOS/Renderers/EntryRenderer.cs
+++ b/Xamarin.Forms.Platform.MacOS/Renderers/EntryRenderer.cs
@@ -69,9 +69,7 @@ namespace Xamarin.Forms.Platform.MacOS
 			void HandleWindowDidBecomeKey(object sender, EventArgs args)
 			{
 				if (CurrentEditor == Window.FirstResponder)
-				{
 					FocusChanged?.Invoke(this, new BoolEventArgs(true));
-				}
 			}
 		}
 
@@ -189,7 +187,7 @@ namespace Xamarin.Forms.Platform.MacOS
 		void OnChanged(object sender, EventArgs eventArgs)
 		{
 			UpdateMaxLength();
-			
+
 			ElementController.SetValueFromRenderer(Entry.TextProperty, Control.StringValue);
 		}
 
--- a/Xamarin.Forms.Platform.MacOS/Renderers/EntryRenderer.cs
+++ b/Xamarin.Forms.Platform.MacOS/Renderers/EntryRenderer.cs
@@ -13,6 +13,8 @@ namespace Xamarin.Forms.Platform.MacOS
 
 			bool _windowEventsSet;
 
+			bool _disposed;
+
 			public override bool ResignFirstResponder()
 			{
 				return base.ResignFirstResponder();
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Platform.MacOS/Renderers/EntryRenderer.cs
```diff
@@ -9,15 +10,50 @@ public class EntryRenderer : ViewRenderer<Entry, NSTextField>
...
 			{
 				FocusChanged?.Invoke(this, new BoolEventArgs(true));
-				return base.BecomeFirstResponder();
+
+				var result = base.BecomeFirstResponder();
+
+				if (!_windowEventsSet)
+				{
+					_windowEventsSet = true;
+					Window.DidResignKey += HandleWindowDidResignKey;
```
rmarinho: we need to unsubscribe these events

Xamarin.Forms.Platform.MacOS/Renderers/EntryRenderer.cs
```diff
@@ -9,15 +10,50 @@ public class EntryRenderer : ViewRenderer<Entry, NSTextField>
...
+					Window.DidBecomeKey += HandleWindowDidBecomeKey;
+				}
+
+				return result;
+			}
+
+			public override void DidEndEditing(NSNotification notification)
+			{
+				if (CurrentEditor != Window.FirstResponder)
+				{
```
rmarinho: i would drop the `{` here

Xamarin.Forms.Platform.MacOS/Renderers/EntryRenderer.cs
```diff
@@ -9,15 +10,50 @@ public class EntryRenderer : ViewRenderer<Entry, NSTextField>
...
+					FocusChanged?.Invoke(this, new BoolEventArgs(false));
```
rmarinho: i wonder if we could get in a place that we fire the event multiple times here, and also fire the `FocusChanged` on `HandleWindowDidResignKey`

Xamarin.Forms.Platform.MacOS/Renderers/EntryRenderer.cs
jvansickle: @rmarinho addressed in commit a73caf2

PR #3021, opened 2018-06-12
[Android,IOS] Fix inserting page and popping current

### LINKED ISSUE #2837: Regression - Android - Exception thrown during NavigationPage.Navigation.PopAsync
I am seeing an `ArgumentOutOfRangeException` when executing PopAsync after inserting a page into the navigation stack, which seems to be a regression of [this issue](https://bugzilla.xamarin.com/show_bug.cgi?id=53179) and may be related to #2576 .

The issue can be reproduced by adding a single empty page to an empty template project and then including the following code in an async OnStart() implementation:

```
	var nav = (NavigationPage) MainPage;
	nav.Navigation.InsertPageBefore(new SecondPage(), nav.CurrentPage);
	await nav.Navigation.PopAsync(false);
```

This issue only affects Android projects using Material Design (FormsAppCompatActivity).

- Version with issue: All Xamarin.Forms versions from v2.5.1.444934 onwards
- Last known good version: v2.5.0.280555
- IDE: VS2017 v15.7.1
- Android Target Framework: 7.1, 8.0, 8.1
- Android Support Library Version: v25.4.0.2 and v27.0.2

[StackTrace.txt](https://github.com/xamarin/Xamarin.Forms/files/2041384/StackTrace.txt)
[BugRepro.zip](https://github.com/xamarin/Xamarin.Forms/files/2041385/BugRepro.zip)

rmarinho:
### Description of Change ###

Android
If there was only 1 element on the stack the the índex was trying to remove a element of out bounds. 
IOS
Was removing the top and only controller

### Bugs Fixed ###

- fixes #2837 

### API Changes ###

None

### Behavioral Changes ###

Should not crash when removing the previous page when appearing on Android

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/NavigationRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/NavigationRenderer.cs
@@ -321,9 +321,9 @@ namespace Xamarin.Forms.Platform.iOS
 			if (poppedViewController == null)
 			{
 				//// this happens only when the user does something REALLY dumb like pop right after putting the page as visible.
-				//poppedViewController = TopViewController;
-				//var newControllers = ViewControllers.Remove(poppedViewController);
-				//ViewControllers = newControllers;
+				poppedViewController = TopViewController;
+				var newControllers = ViewControllers.Remove(poppedViewController);
+				ViewControllers = newControllers;
 				actuallyRemoved = true;
 			}
 			else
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Platform.iOS/Renderers/NavigationRenderer.cs
```diff
@@ -320,16 +320,16 @@ protected virtual void OnElementChanged(VisualElementChangedEventArgs e)
 
 			if (poppedViewController == null)
 			{
-				// this happens only when the user does something REALLY dumb like pop right after putting the page as visible.
-				poppedViewController = TopViewController;
-				var newControllers = ViewControllers.Remove(poppedViewController);
-				ViewControllers = newControllers;
+				//// this happens only when the user does something REALLY dumb like pop right after putting the page as visible.
```
hartez: Did you mean to leave these changes in?

Either way, we should remove the "really dumb" from this comment.

kwacks:
Curious when these changes will be released?

PR #3037, opened 2018-06-14
Fix broken 2499 test

StephaneDelcroix:
### Description of Change ###

SelectedIndex being a BP, setting it's value might be delayed, and as
both ios and android PickerRenderer do not update on individual property
changes, but have a single UpdatePicker method, some properties might be
out-of-date, causing a crash.

This doesn't attemps to fix the renderers, but only prevents a crash to
happen.

### API Changes ###

/

### Platforms Affected ###

- iOS
- Android

/

### PR Checklist ###

- [x] Has automated tests 


StephaneDelcroix:
build --uitests

rmarinho:
seems this doesn't fix the test on Android

PR #3041, opened 2018-06-14
Give XFCorePostProcessor a different GUID than Xamarin.Forms.Xaml.Design

PureWeen:
### Description of Change ###

XFCorePostProcessor  was added with the same GUID as Xamarin.Forms.Xaml.Design so VS on Windows kept reassigning it and then adding it to all the build configurations. This commit gives it its own GUID and keeps it removed from all build configurations

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.sln
+++ b/Xamarin.Forms.sln
@@ -174,7 +174,7 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Xamarin.Forms.Maps.Design",
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Xamarin.Forms.Xaml.Design", "Xamarin.Forms.Xaml.Design\Xamarin.Forms.Xaml.Design.csproj", "{65BC4888-CC59-428A-9B75-540CF1C09480}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "XFCorePostProcessor.Tasks", "XFCorePostProcessor.Tasks\XFCorePostProcessor.Tasks.csproj", "{5BBF4A3F-4AD1-47FD-B250-05EA793F939D}"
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.sln
```diff
@@ -1575,6 +1574,18 @@ Global
...
+		{5BBF4A3F-4AD1-47FD-B250-05EA793F939D}.Debug|iPhone.ActiveCfg = Debug|Any CPU
+		{5BBF4A3F-4AD1-47FD-B250-05EA793F939D}.Debug|iPhoneSimulator.ActiveCfg = Debug|Any CPU
+		{5BBF4A3F-4AD1-47FD-B250-05EA793F939D}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{5BBF4A3F-4AD1-47FD-B250-05EA793F939D}.Debug|x86.ActiveCfg = Debug|Any CPU
+		{5BBF4A3F-4AD1-47FD-B250-05EA793F939D}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{5BBF4A3F-4AD1-47FD-B250-05EA793F939D}.Release|ARM.ActiveCfg = Release|Any CPU
+		{5BBF4A3F-4AD1-47FD-B250-05EA793F939D}.Release|iPhone.ActiveCfg = Release|Any CPU
+		{5BBF4A3F-4AD1-47FD-B250-05EA793F939D}.Release|iPhoneSimulator.ActiveCfg = Release|Any CPU
+		{5BBF4A3F-4AD1-47FD-B250-05EA793F939D}.Release|x64.ActiveCfg = Release|Any CPU
+		{5BBF4A3F-4AD1-47FD-B250-05EA793F939D}.Release|x86.ActiveCfg = Release|Any CPU
```
StephaneDelcroix: XFCorePostProcess should never be enabled

PureWeen: Yea the language used here is confusing... It's not actually enabled for those release configurations but VS still needs these lines here

If it was enabled the line would read
```
{5BBF4A3F-4AD1-47FD-B250-05EA793F939D}.Release|x64.ActiveCfg = Release|Any CPU
{5BBF4A3F-4AD1-47FD-B250-05EA793F939D}.Release|x64.Build.0 = Release|Any CPU
```

Like if you just delete all of these entries out VS will just auto add them back in as included into the build configuraitons

PR #3045, opened 2018-06-14
[android] new IImageViewHandler API

jonathanpeppers:
### Description of Change ###

Context:
https://github.com/bumptech/glide
https://github.com/jonathanpeppers/glidex

Currently the way my "proof-of-concept" GlideX library works by
completely bypassing `IImageSourceHandler`. GlideX provides its own
custom `ImageRenderer` and `ImageCellRenderer`. This was required due
to how *opinionated* the Glide library is. Glide's approach is to
never allow a developer access to a `Android.Graphics.Bitmap` or
`Android.Graphics.Drawable` because we would likely do it wrong... and
developers do all the time!

To evolve Xamarin.Forms to where images could be better handled down
the road, I've introduced a new interface:

    public interface IImageViewHandler : IRegisterable
    {
        Task LoadImageAsync(ImageSource imageSource, ImageView imageView, CancellationToken cancellationToken = default(CancellationToken));
    }

The idea is that we can query for `IImageViewHandler` and fall back to
`IImageSourceHandler`. This would allow GlideX to just be an
`IImageViewHandler` and not mess with creating custom renderers.

We can also implement `IImageViewHandler` for `FileImageSource`, to
get some general performance improvements around loading files from
disk:

    string file = ((FileImageSource)imagesource).File;
    if (File.Exists(file))
    {
        //Load with imageView.SetImageURI(Android.Net.Uri)
    }
    else
    {
        //Load with imageView.SetImageResource(int)
    }

I tested this change with new performance tests in `ImageScenarios`:
- Load 100 images using `AndroidResource`
- Load 100 images from disk
- I conveniently prefixed these with `.`, so they appeared first in
  the scenario list

Here are the results from three runs, using `IImageSourceHandler`
versus `IImageViewHandler`, in a HAXM emulator:

`IImageSourceHandler` - 100x `AndroidResource`
- 6059.899
- 3297.885
- 3015.179

`IImageSourceHandler` - 100x from disk
- 12398.71
- 14146.41
- 16060.88

`IImageViewHandler` - 100x `AndroidResource`
- 6748.766
- 2817.975
- 2456.197

`IImageViewHandler` - 100x from disk
- 7326.745
- 4799.001
- 5411.515

There is not going to be as much as an improvement for
`AndroidResource` (maybe not any), since Xamarin.Forms looks for
`Drawable` since: https://github.com/xamarin/Xamarin.Forms/pull/1973

_NOTE: that these scenarios are probably too slow to keep, as it seems
these performance tests are geared to fail if they take longer than
250ms. I can remove these before we merge this PR._

Other changes to make this happen:
- `Registrar` was failing with `InvalidCastException` instead of
  returning `null` when querying for `IImageViewHandler`. I switched
  to using `as` instead of a hard cast. I am not sure why it was not
  already setup this way, since
  `Registrar.Registered.GetHandlerForObject<T>` appears to be called
  and checked for `null` everywhere.

### Issues Slightly Improved ###

Image performance on Android

### API Changes ###

Added:
 - Xamarin.Forms.Platform.Android.IImageViewHandler

### Platforms Affected ###

- Android

If an `IImageViewHandler` is found for a type of `ImageSource` it is used before falling back to a `IImageSourceHandler`.

`Registrar.GetHandler` now returns `null` instead of throwing `InvalidCastException`.

### PR Checklist ###

- [x] Has automated tests (NOTE I expect the new scenarios to fail, they are slow)

### DIFF
--- a/Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
+++ b/Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
@@ -31,27 +31,43 @@ namespace Xamarin.Forms.Platform.Android
 
 			imageView.SetImageResource(global::Android.Resource.Color.Transparent);
 
+			bool setByImageViewHandler = false;
 			Bitmap bitmap = null;
 			Drawable drawable = null;
 
-			IImageSourceHandler handler;
-
-			if (source != null && (handler = Internals.Registrar.Registered.GetHandlerForObject<IImageSourceHandler>(source)) != null)
+			if (source != null)
 			{
-				if (handler is FileImageSourceHandler)
-				{
-					drawable = imageView.Context.GetDrawable((FileImageSource)source);
-				}
-
-				if (drawable == null)
+				var imageViewHandler = Internals.Registrar.Registered.GetHandlerForObject<IImageViewHandler>(source);
+				if (imageViewHandler != null)
 				{
 					try
 					{
-						bitmap = await handler.LoadImageAsync(source, imageView.Context);
+						await imageViewHandler.LoadImageAsync(source, imageView);
+						setByImageViewHandler = true;
 					}
 					catch (TaskCanceledException)
 					{
 						imageController?.SetIsLoading(false);
+					}
+				}
+				else
+				{
+					var imageSourceHandler = Internals.Registrar.Registered.GetHandlerForObject<IImageSourceHandler>(source);
+					if (imageSourceHandler is FileImageSourceHandler)
+					{
+						drawable = imageView.Context.GetDrawable((FileImageSource)source);
+					}
+
+					if (drawable == null)
+					{
+						try
+						{
+							bitmap = await imageSourceHandler.LoadImageAsync(source, imageView.Context);
+						}
+						catch (TaskCanceledException)
+						{
+							imageController?.SetIsLoading(false);
+						}
 					}
 				}
 			}
@@ -63,7 +79,7 @@ namespace Xamarin.Forms.Platform.Android
 				return;
 			}
 
-			if (!imageView.IsDisposed())
+			if (!setByImageViewHandler && !imageView.IsDisposed())
 			{
 				if (bitmap == null && drawable != null)
 				{
--- a/Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
+++ b/Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
@@ -33,7 +33,6 @@ namespace Xamarin.Forms.Platform.Android
 
 			bool setByImageViewHandler = false;
 			Bitmap bitmap = null;
-			Drawable drawable = null;
 
 			if (source != null)
 			{
@@ -53,21 +52,13 @@ namespace Xamarin.Forms.Platform.Android
 				else
 				{
 					var imageSourceHandler = Internals.Registrar.Registered.GetHandlerForObject<IImageSourceHandler>(source);
-					if (imageSourceHandler is FileImageSourceHandler)
+					try
 					{
-						drawable = imageView.Context.GetDrawable((FileImageSource)source);
+						bitmap = await imageSourceHandler.LoadImageAsync(source, imageView.Context);
 					}
-
-					if (drawable == null)
+					catch (TaskCanceledException)
 					{
-						try
-						{
-							bitmap = await imageSourceHandler.LoadImageAsync(source, imageView.Context);
-						}
-						catch (TaskCanceledException)
-						{
-							imageController?.SetIsLoading(false);
-						}
+						imageController?.SetIsLoading(false);
 					}
 				}
 			}
@@ -81,10 +72,8 @@ namespace Xamarin.Forms.Platform.Android
 
 			if (!setByImageViewHandler && !imageView.IsDisposed())
 			{
-				if (bitmap == null && drawable != null)
-				{
-					imageView.SetImageDrawable(drawable);
-				}
+				if (bitmap == null && source is FileImageSource)
+					imageView.SetImageResource(ResourceManager.GetDrawableByName(((FileImageSource)source).File));
 				else
 				{
 					imageView.SetImageBitmap(bitmap);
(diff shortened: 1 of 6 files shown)


StephaneDelcroix:
what's the perf change once you use glide as handler ?

jonathanpeppers:
I ran some original numbers, this was Xamarin.Forms 2.5.0.280555: https://github.com/jonathanpeppers/glidex#comparing-performance (this was before #1973, too)

I should still make a branch where I add in GlideX and try these same perf tests.

The main benefit is that glide reuses images if they are on the screen multiple times--or if you scroll a `ListViewItem` off screen and bring it back. It should have a huge impact for `UriImageSource`.

jonathanpeppers:
@StephaneDelcroix I did a quick test here, using Glide: https://github.com/xamarin/Xamarin.Forms/compare/master...jonathanpeppers:glide-parallel-universe

I created a new performance scenario loading 100x images from a `Uri`.

The results using Xamarin.Forms as-is:
- 4299.866 - only 4 on screen loaded
- 2445.471 - only 2 on screen loaded
- 2189.805 - only 2 on screen loaded

The time taken, appeared to be the number of images that did not throw an exception and properly loaded. There wasn't an error message at the console, but there is probably something else wrong here...

Next, I implemented Glide via `IImageViewHandler`, and re-ran my times:

Glide - 100x `AndroidResource`
- 4352.571
- 2967.391
- 1853.745

Glide - 100x from disk
- 2310.304
- 1506.563
- 1521.049

Glide - 100x from Uri (note that they all loaded)
- 1633.478 - all loaded
- 1238.600 - all loaded
- 1584.175 - all loaded

Glide must be freaking black magic, I don't know how Glide is making improvements in even the `AndroidResource`.

I suspect even `EmbeddedResource` could have an improvement using Glide...

jamesmontemagno:
This is super very much needed and I applaud this work <3

Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
```diff
@@ -63,7 +79,7 @@ internal static class ImageViewExtensions
 				return;
 			}
 
-			if (!imageView.IsDisposed())
+			if (!setByImageViewHandler && !imageView.IsDisposed())
```
PureWeen: We should probably reinstate this code

https://github.com/xamarin/Xamarin.Forms/blob/e8660383b03b4010724b36c6eef471a2e74e7c98/Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs#L55

Because AFAICT anyone that followed your advice on the blog
http://jonathanpeppers.com/Blog/xamarin-forms-performance-on-android

Will end up with broken images after this PR

PureWeen:
Also if you rebase we've fixed that annoying issue with the SLN keeps changing when opened in VS :-)

Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
PureWeen: I had that thought initially as well but I'm pretty sure the export of the handler causes the default one to just get totally replaced 

when I added in 
```C#
[assembly: ExportImageSourceHandler(typeof(FileImageSource), typeof(xfperf.FileImageSourceHandler))]

namespace xfperf
{
    public class FileImageSourceHandler : IImageSourceHandler
    {
        public Task<Bitmap> LoadImageAsync(ImageSource imagesource, Context context, CancellationToken cancelationToken = default(CancellationToken))
        {
            return Task.FromResult<Bitmap>(null);
        }
    }
}
```

It caused this line
```C#
var imageViewHandler = Internals.Registrar.Registered.GetHandlerForObject<IImageViewHandler>(source);
``` 

to return null and fall through to the else statement part and then call into the xfperf handler

jonathanpeppers: I put it back

PureWeen:
build --uitests

jonathanpeppers:
@syed-pelican the only change here is the case of a file from disk; for example,  if you had downloaded a file to disk somewhere and used its path for an `Image`.

To restore old behavior:

You can add your own handler for files that _only_ implements `IImageSourceHandler`, and that should work. It could create an instance of `Xamarin.Forms.Platform.Android.FileImageSourceHandler` and call its `LoadImageAsync` method directly.

PureWeen:
@syed-pelican  what's your use case? Why are you wanting to fall back to the original Image handler?

syed-pelican:
Hi guys,

Sorry about the delay. 

Currently our app behaviour is as follows:

We have `<Image/>` elements among others and the Source property is bound to a public property of type `ImageSource` (lets call it PhotoSource). As the image appears on the view we queue them for download and save them to disk. Once downloaded, we assign PhotoSource as 

`PhotoSource = ImageSource.FromFile(fileName);`

Prior XF 3.2 everything worked fine. However going forward from XF 3.2 the images are getting mixed up as we scroll up and down (please refer to the video files on the link provided).

[Shared Video Link](https://drive.google.com/drive/folders/1rAKadYtilM8LwfhhErDGKHVlW7kwbPM_?usp=sharing)

I'm just trying to rule out if our implementation is incorrect.

Happy to share more information as you need. May be a repo as well but not sure if I can spare enough time off work to cook up all the behaviour we have. Still I hope the video helps.

PureWeen:
@syed-pelican  the issue you're describing sounds like you are affecting the template directly opposed to doing it through bindings?

The ListView you are using what is the Recycle Strategy set to? Is it set to Retain or Recycle?
If you are using Recycle can you set it to Retain and see if that fixes? 

If you are using Recycle I wouldn't recommend setting the PhotoSource directly like you are but instead do it through a binding via a converter.   If you set PhotoSource directly (which I'm guessing you're doing through some event or something?) Then that effects the template that gets reused other places

syed-pelican:
Hi @PureWeen you are right about the use of DataTemplateSelector. Using ValueConverter has actually resolved the issue. However, it took us a while as we had to update every single ListView with same setup. using RetainElement also helps but since our ListView is quite complex and long we prefer to keep RecycleElement. I'm trying to understand what has changed since XF 3.1 that has resulted in this behaviour, as previously we didn't require a ValueConverter.

Hope this helps!

PureWeen:
@syed-pelican  were you only seeing this behavior on Android? or was it doing it for iOS as well when you were using a recycle strategy?

syed-pelican:
Only Android I'm afraid. iOS is fine since we have implemented the views

daniel-luberda:
FFImageLoading implements it too now. Thanks @jonathanpeppers for the idea. Also a one liner: 

`CachedImageRenderer.InitImageViewHandler()` (instead `CachedImageRenderer.InitImageSourceHandler()`)

PR #3101, opened 2018-06-21
Clear all previous animations with the same key when starting a new animation

### LINKED ISSUE #2482: Animating a `View` that is currently animating will throw  `System.InvalidOperationException`:
### Description
Animating a `View` that is currently animating will throw  `System.InvalidOperationException`:

>TaskCompletionSource[TResult].SetResult (TResult result)
System.InvalidOperationException: An attempt was made to transition a task to a final state when it had already completed.

### Steps to Reproduce

1. Download/Clone sample app reproduction: https://github.com/brminnick/AnimationCrashReproductionSample
2. Build/Deploy AnimationCrashReproductionSample to an Android or iOS emulator/device
3. Quickly double-tap the button on the screen to induce the Exception

### Expected Behavior

The animating Task should be canceled gracefully, avoiding the Exception.

### Actual Behavior

Performing an animation on a `View` that is currently animating will throw  `System.InvalidOperationException`:

### Basic Information

- Version with issue: v2.5.1.444934
- Last known good version: Never
- IDE: Visual Studio for Mac
- Platform Target Frameworks: 
  - iOS:  11.2
  - Android: 8.1
  - UWP:  Unknown
- Android Support Library Version: v27.0.2
- Affected Devices: iOS, Android, UWP

### Screenshots

https://github.com/brminnick/Videos/blob/master/AnimationCrashReproductionSample/AnimationCrashReproductionSample.gif

### Reproduction Link

https://github.com/brminnick/AnimationCrashReproductionSample

hartez:
### Description of Change ###

When using the animation extensions to `VisualElement`, queueing up multiple animations can result in a race condition where an animation Task result is set multiple times (throwing an exception) and another animation's result is never set (meaning an `await` on it never returns). 

This change guarantees that all animations with the same key are cancelled before queueing up a new animation with that key, avoiding both scenarios.

### Issues Resolved ###

- fixes #2482 

### API Changes ###

None

### Platforms Affected ###

- Core/XAML (all platforms)

None

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2482.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2482.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.ComponentModel;
 using System.Diagnostics;
 using System.Dynamic;
 using System.Threading;
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2482.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2482.cs
@@ -1,0 +1,83 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Diagnostics;
+using System.Dynamic;
+using System.Threading;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.Forms.Core.UITests;
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if UITEST
+	[Category(UITestCategories.Animation)]
+#endif
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 2482,
+		"Animating a `View` that is currently animating will throw `System.InvalidOperationException`", 
+		PlatformAffected.All)]
+	public class Issue2482 : TestContentPage
+	{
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2482.cs
```diff
@@ -0,0 +1,77 @@
...
+				VerticalOptions = LayoutOptions.Center,
+				AutomationId = ButtonId
+			};
+
+			button.Clicked += async (sender, args) =>
+			{
+				_clicks += 1;
+
+				await button.RotateTo(539, 3000, Easing.CubicOut);
+				await button.RotateTo(0, 3000, Easing.CubicIn);
```
PureWeen: In messing with this I noticed that now it seems to hang on this second await call. If you move the *_clicks += 1;* down below these awaits and then click twice really quickly it never increments the clicks to 2. It should probably still fall through from both clicks yes?

StephaneDelcroix:
(retargeted to master)

hartez:
@rmarinho The API change is in Internals; the API change detector doesn't know it can ignore that namespace.

hartez:
> @hartez what's the api change?

@rmarinho There isn't one; I think the check script is giving us a false positive.

PureWeen:
@hartez 
That looks great!! so much removed redundant code.

I saw in the cleanup you changed a few things to nameof

There are a few more in the AnimateTo calls that could get the same treatment
```
return AnimateTo(view, 0, 1, "LayoutTo", (v, value) => v.Layout(computeBounds(value)), length, easing);
```

Not necessary but just wanted to mention

hartez:
> Not necessary but just wanted to mention

No, it's a good point. Done. 
Bonus: The `AbortAnimations` call was wasting cycles trying to cancel "SizeTo", which doesn't even exist.

PR #3109, opened 2018-06-21
Use GestureDetector constructor that takes in Context 3.0

PureWeen:
This is just to get these changes into an early branch so that once 15.8 drops we'll still be able to compile

https://github.com/xamarin/Xamarin.Forms/pull/2836


PureWeen:
build --uitests

PR #3141, opened 2018-06-25
[UWP] Allow setting StepFrequency of the Slider

### LINKED ISSUE #2955: [UWP] Default Slider has only 10 possible values
Slider intervals are inconsistent accross platforms. The problematic platform is UWP:

The current behavior on UWP is approximately: snap to integers (if there are at least 10 integers in the range); otherwise snap to 10 points in the range.

https://bugzilla.xamarin.com/show_bug.cgi?id=38471
> Charles Petzold 2016-02-04 20:45:48 UTC
> A default WinRT/UWP Slider has its StepFrequency property set to 0.1, meaning that the default slider can only yield values of 0, 0.1, 0.2, ... 0.9, 1.0.
> This is in contrast to the default Android Slider, which is given a maximum value of 1000, yielding values of 0, 0.001, 0.002, ... 0.099, 1.000.
> To make these two Sliders consistent, the WinRT Slider should have its [StepFrequency]( https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.slider.stepfrequency#Windows_UI_Xaml_Controls_Slider_StepFrequency) always set to (Maximum - Minimum) / 1000.

Also on UWP a tooltip is shown by default, unlike on iOS (not sure about Android). For consistency the tooltip should be [disabled](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.slider.isthumbtooltipenabled#Windows_UI_Xaml_Controls_Slider_IsThumbToolTipEnabled) on UWP.

(Ideally there would be Xamarin.Forms properties that allow for setting of slider snap intervals and tooltips, but this would require extra work on iOS since the default slider is [limited](https://stackoverflow.com/questions/40670919/uislider-missing-features-snap-tooltip-and-ticks).)

paymicro:
### Description of Change ###

For the Slider default value of step is set to `1000` to match Android.
Also removed ToolTip with current value.

### Issues Resolved ###

- fixes #2955

### API Changes ###

None

### Platforms Affected ###

- UWP

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.UAP/SliderRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/SliderRenderer.cs
@@ -33,6 +33,7 @@ namespace Xamarin.Forms.Platform.UWP
 					Control.Minimum = e.NewElement.Minimum;
 					Control.Maximum = e.NewElement.Maximum;
 					Control.Value = e.NewElement.Value;
+					Control.IsThumbToolTipEnabled = false;
 
 					slider.ValueChanged += OnNativeValueChanged;
 
@@ -56,7 +57,7 @@ namespace Xamarin.Forms.Platform.UWP
 					}
 				}
 
-				double stepping = Math.Min((e.NewElement.Maximum - e.NewElement.Minimum) / 10, 1);
+				double stepping = Math.Min((e.NewElement.Maximum - e.NewElement.Minimum) / 1000, 1);
 				Control.StepFrequency = stepping;
 				Control.SmallChange = stepping;
 				UpdateFlowDirection();


samhouts:
build

samhouts:
build --uitests

Xamarin.Forms.Platform.UAP/SliderRenderer.cs
```diff
@@ -115,6 +120,14 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
...
 			else if (e.PropertyName == Slider.ThumbImageProperty.PropertyName)
 				UpdateThumbImage();
+			else if (e.PropertyName == Specifics.StepFrequencyProperty.PropertyName)
+				UpdateStepFrequency();
+		}
+
+		void UpdateStepFrequency()
+		{
+			if (Element.IsSet(Specifics.StepFrequencyProperty))
+				Control.StepFrequency = Element.OnThisPlatform().GetStepFrequency();
```
hartez: This needs to update `Control.SmallChange` as well.

hartez: Ideally, the logic in `OnElementChanged` would be moved to `UpdateStepFrequency` and `UpdateStepFrequency` would be called by `OnElementChanged`. 

Regardless, the logic in `UpdateStepFrequency` should handle the situation where step frequency is set back to `-1d` (in case the user wants to revert to the default behavior).

hartez:
@charlesroddie 

> a StepFrequency property (double or Nullable<double>) in Xamarin.Forms

How should that property behave on iOS?

charlesroddie:
> How should that property behave on iOS?

It seems iOS doesn't have that natively. Implementing it is straightforward though:

- Define snap function which snaps a double to the required intervals.
- When asked for the slider position, return the snapped position.
- When the slider has finished dragging, replace the position with the snapped position.

hartez:
@charlesroddie Sorry, I didn't phrase that properly. I'm not worried about _how_ to implement a StepFrequency on iOS. I'm concerned with not breaking current behavior when that property is set. 

If we add a StepFrequency to Slider, in order to maintain the current default behavior on iOS we would have to ignore it. So the basic behavior of StepFrequency would be:

UWP - set native StepFrequency
iOS - do nothing 
Android - set native StepFrequency (technically also a breaking change, but ignore that for the moment)

Now we have a property on Slider that doesn't do anything on one of the platforms. That's ... awkward (and confusing to new users). We could add a PlatformSpecific to allow folks to opt-in to the stepping behavior, but now we're maintaining two different properties.

hartez:
> Other option is to, just like Android does, have 1000 steps by default on UWP, but that's still a breaking behavioral breaking changed, and needs a PlatformSpecific for reverting to the default behavior

The more I think about it, the more I'm convinced this is the correct solution. I think dividing the range of the UWP slider by 10 was a mistake in the first place, and setting it to 1000 (to match Android) would be fixing a bug, not breaking an expected behavior. 

On the off chance anyone is depending on the incorrect behavior, we add a PS for restoring it. 

@jassmith Can we get you to weigh in on this?

charlesroddie:
I would:

- Set the default stepfrequency to null (or 0 or -1 if you want to encode it that way)
- A null stepfrequency corresponds to negligible steps: 1000 on UWP and Android, and no change on iOS. This fixes UWP. There is no need to add a platformspecific to revert behavior. That is the wrong attitude to bug-fixing, and no-one would use it anyway.
- A non-null stepfrequency is implemented as you would expect.

hartez:
I just talked to Jason and he's good with just setting that value to 1000 to match Android, and no platform specific to revert it. So that's what we need out of this PR. 

> A non-null stepfrequency is implemented as you would expect on all platforms.

@charlesroddie Can you post an issue for a StepFrequency property as an enhancement (with your suggestions on how it should behave on iOS)?

paymicro:
@hartez as a result, what changes are expected in this PR? Just change the behavior in UWP as in Android and no platform specific?

hartez:
@paymicro Exactly.

samhouts:
@paymicro Please rebase on master! Thanks!

PR #3164, opened 2018-06-27
[build] require .NET framework >= 4.7

StephaneDelcroix:
### Description of Change ###

Throw a meaningful error on old .NET Framework

### DIFF
--- a/.nuspec/Xamarin.Forms.targets
+++ b/.nuspec/Xamarin.Forms.targets
@@ -54,6 +54,15 @@
 			Condition="'$(IsXBuild)' == 'true'" />
 	</Target>
 
+	<Target Name="_ValidateNETFrameworkVersion"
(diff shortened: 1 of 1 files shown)


.nuspec/Xamarin.Forms.targets
```diff
@@ -54,6 +54,15 @@
 			Condition="'$(IsXBuild)' == 'true'" />
 	</Target>
 
+	<Target Name="_ValidateNETFrameworkVersion"
+			BeforeTargets="_CheckForInvalidConfigurationAndPlatform"
+			Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' AND '$(TargetFrameworkVersion.Substring(1))' &lt; '4.7' ">
```
paymicro: I think that it will be more clear to use `.TrimStart('vV')` instead `.Substring(1)`

PR #3174, opened 2018-06-27
[Android] Don't mess with the `Button.Padding` if the `Padding` property isn't set

samhouts:
### Description of Change ###

#2426 added the `Button.Padding` property, but on Android, it was removing necessary padding when the value wasn't set. This checks to ensure that there is a value set before it adjusts it.

#### Before this change:
screenshot_1530138753

#### After this change: (should match screenshots in #1570)
screenshot_1530138641

### Issues Resolved ###

unreported

### API Changes ###

None

### Platforms Affected ###

- Android

None

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/ButtonRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/ButtonRenderer.cs
@@ -20,6 +20,7 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 		ButtonBackgroundTracker _backgroundTracker;
 		TextColorSwitcher _textColorSwitcher;
 		float _defaultFontSize;
+		Thickness? _defaultPadding;
 		Typeface _defaultTypeface;
 		bool _isDisposed;
 		int _imageHeight = -1;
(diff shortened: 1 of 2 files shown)


samhouts:
build --uitests

Xamarin.Forms.Platform.Android/AppCompat/ButtonRenderer.cs
```diff
@@ -287,6 +287,9 @@ void UpdateTextColor()
 
 		void UpdatePadding()
 		{
+			if (!Element.IsSet(Button.PaddingProperty))
+				return;
+
 			Control?.SetPadding(
 				(int)(Context.ToPixels(Element.Padding.Left) + _paddingDeltaPix.Left),
 				(int)(Context.ToPixels(Element.Padding.Top) + _paddingDeltaPix.Top),
```
StephaneDelcroix: I don't think this covers the case when a `Padding` is set, and then cleared.

paymicro: Is it possible to clear padding? It can not be null, and if set to `new Thickness()`, it will be set to `0`.
but if cleaning is possible then need `SetPadding` to default value 
```C# 
Control?.SetPadding(_defaultPadding.Left, _defaultPadding.Top, _defaultPadding.Right, _defaultPadding.Bottom);
```

samhouts: @paymicro Yep, you can do `button.ClearValue(Button.PaddingProperty)`. You and Stephane are describing the pattern that we generally follow for such property (i.e., store the default value and restore when the value is unset), and I *did* think about it yesterday, but I thought to myself...who would do such a thing? Me being lazy. :D fixing now

samhouts:
build --uitests

PR #3187, opened 2018-06-29
[Android]  Let inflight life cycle events finish up before tearing them down

### LINKED ISSUE #2338: [Android] Crash when switching MainPage from MDP to Page in Init/Appearing/ctor
### Description

When the MasterDetailContainer Dispose runs, it crashes if you're switching the MainPage immediately on a page load.

### Steps to Reproduce

1. Run B40005 test in Control Gallery/
https://github.com/Schinwinkwinsky/DroidFragmentCrash

### Expected Behavior

Test loads

### Actual Behavior

Test immediately crashes

### Basic Information

- Version with issue: 3.0.0,2.5.1
- Last known good version: ?

PureWeen:
### Description of Change ###

OnAppearing for Android fires sometimes during *OnAttached* and sometimes during "OnResume"

- when you swapped out the MainPage it would call RemoveViews mid fragment lifecycle which would then lead to the fragment not found when FragmentManager would try to Resume the fragment

- Added an interface to allow a renderer to be declared MarkedForDeath so that things like adding more fragments or pending operations can just be halted

- Centralized on interactions with Fragments into a single Extension class because that just made it easier to break point or print debug messages for any Fragment modifications.

- Moved all the code for changing out the main page into the MainLooper call so that all "active" life cycle events will just settle before the main page does it's clean up and "realizing"

- Moved the code for when you swap out Details or Master pages it'll use MainLooper if replacing an existing page to ensure life cycle events have concluded

### Issues Resolved ###
 
- fixes #2338

### Platforms Affected ###

- Android

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/Extensions/FragmentManagerExtensions.cs
+++ b/Xamarin.Forms.Platform.Android/Extensions/FragmentManagerExtensions.cs
@@ -7,44 +7,44 @@ namespace Xamarin.Forms.Platform.Android
 	// This is a way to centralize all fragment modifications which makes it a lot easier to debug
 	internal static class FragmentManagerExtensions
 	{
-		public static FragmentTransaction RemoveEx(this FragmentTransaction @this, Fragment fragment)
+		public static FragmentTransaction RemoveEx(this FragmentTransaction fragmentTransaction, Fragment fragment)
 		{
-			return @this.Remove(fragment);
+			return fragmentTransaction.Remove(fragment);
 		}
 
-		public static FragmentTransaction AddEx(this FragmentTransaction @this, int containerViewId, Fragment fragment)
+		public static FragmentTransaction AddEx(this FragmentTransaction fragmentTransaction, int containerViewId, Fragment fragment)
 		{
-			return @this.Add(containerViewId, fragment);
+			return fragmentTransaction.Add(containerViewId, fragment);
 		}
 
-		public static FragmentTransaction HideEx(this FragmentTransaction @this, Fragment fragment)
+		public static FragmentTransaction HideEx(this FragmentTransaction fragmentTransaction, Fragment fragment)
 		{
-			return @this.Hide(fragment);
+			return fragmentTransaction.Hide(fragment);
 		}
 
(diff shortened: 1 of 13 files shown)


PureWeen:
build --uitests

PureWeen:
build --uitests

Xamarin.Forms.Platform.Android/Extensions/FragmentManagerExtensions.cs
```diff
@@ -0,0 +1,50 @@
+﻿using FragmentTransaction = Android.Support.V4.App.FragmentTransaction;
+using Fragment = Android.Support.V4.App.Fragment;
+using FragmentManager = Android.Support.V4.App.FragmentManager;
+
+namespace Xamarin.Forms.Platform.Android
+{
+	// This is a way to centralize all fragment modifications which makes it a lot easier to debug
+	internal static class FragmentManagerExtensions
```
hartez: While this _does_ make the debugging easier, it's also forcing an extra method call on every fragment transaction operation. You could probably achieve the same effect by decorating FragmentTransaction with another class and only using that decorator class when DEBUG is defined. That way you have a central place for debugging but you don't impact the Release version as much.

Something like this:
https://gist.github.com/hartez/a7824a102c595ca096fce8bdca1cec78

PureWeen: I think the extra call get optimized out don't they?

image

I like the idea of the decorator because it also provides a useful place to expand on the debug implementation but I'm concerned about other anomalies of the wrapper or parts of it that can't be overridden like
```c#
public FragmentTransaction SetBreadCrumbShortTitle(string text);
public FragmentTransaction SetBreadCrumbTitle(string text);
```
Though I'm pretty sure these just execute the *ICharSequence* Overrides

Either way I just envision some hard to find inconsistency between debug/release caused by this

hartez: > Either way I just envision some hard to find inconsistency between debug/release caused by this

That's a reasonable fear. I'm fine with the extension method overhead trade-off for the consistency.

PR #3205, opened 2018-07-03
[iOS] Fix ScrollView offset reset on device rotation (Bugzilla #52510)

gmarrot:
### Description of Change ###

I changed OnNativeControlUpdated method of ScrollViewRenderer for iOS to prevent the ContentSize update first to view's bounds and then to content size on device rotation or size change. Now ContentSize will be updated to view's bounds only when the ScrollView's content has an empty size.

### Issues Resolved ###

- fixes [Bugzilla 52510](https://bugzilla.xamarin.com/show_bug.cgi?id=52510)

### API Changes ###

None

### Platforms Affected ###

- iOS

ScrollView will not reset its scroll offset anymore on iOS device rotation.

### PR Checklist ###

- [x] Has automated tests 


PureWeen:
Look at issue
https://github.com/xamarin/Xamarin.Forms/issues/3779

Before resolving this one

gmarrot:
Hi,
I use the app given in issue #3779 to test the behavior. It appears removing the RelativeLayout used as parent of the ListView or ScrollView's container solves the problem.

It seems to be a RelativeLayout issue on device rotation and not a ScrollView's one.

gmarrot:
Pull request #3842 fix this issue too and was merged to master.

PR #3214, opened 2018-07-04
[UWP, Android, iOS] Added better life cycle events

### LINKED ISSUE #2210: [Enhancement] Add better life cycle events
## Rationale

Some lifecycle events are currently unavailable, and others are inconsistently supported across platforms Xamarin.Forms supported. To get to them today developers need to write custom page renderers in the platform projects and surface them in Xamarin.Forms. It's a pain, and we can do better!

### Example or Ordering Issue
Currently, where MainPage is a TabbedPage you get this:

iOS: 
1. MainPage.CurrentPageChanged event handled
2. SchedulePage.OnDisappearing() called
3. SettingsPage.OnAppearing() called

Android:
1. SchedulePage.OnDisappearing() called
2. SettingsPage.OnAppearing() called
3. MainPage.CurrentPageChanged event handled

## Implementation

The comment here I think most closely indicates what we want
https://forums.xamarin.com/discussion/comment/240051/#Comment_240051
https://forums.xamarin.com/discussion/comment/240314/#Comment_240314
> It's worth noting that in the iOS case I am actually suggesting that we do not use ViewWillDisappear here. I do not believe sticking to that particular callback on iOS lends any benefit but having it uniform and predictable as possible in this case is preferable.

I've defined these all on VisualElement as they will now just be symmetrical for View and Page
- VisualElement.Appearing – The view is about to appear. This is a good time to set up update events, etc. Data updates should be enqueued here, but initial data should already be loaded.
- VisualElement.Appeared – The view is visible on screeen, feel free to screenshot it. There is no guarantee that the view only just became visible, just that this happens after Appeared and before DIsappearing with a (strong) bias towards as close to Appearing as possible on the platform. These will most likely be tied to platform events. Just make sure to validate that once the event has fired the VisualElement has indeed appeared
- VisualElement.Disappearing – The view has been removed from the Xamarin.Forms logical hierarchy, either by direct or indirect action by the user. The view MAY still be be visible on screen.
- VisualElement.Disappeared – The view is no longer visible on screen.
- VisualElement.Loaded - Renderer is attached to VisualElement
- VisualElement.Unloaded - Renderer is removed from VisualElement

## Things to consider
- Right now there exists a lot of code that propagates lifecycle events through pages and manually calls appearing/disappearing which is currently causing duplicated propagation https://github.com/xamarin/Xamarin.Forms/commit/c0c007773efdc6f774c3aeed9740adf735ad1e60#diff-dd108d9d76d564cac973db4899b5d797R359
- Modal and Tabs: With modal and tabs there's more of a logical concept for disappearing/disappeared because the pages aren't detached from the visual tree they just get "hidden" by the incoming view
- The most important thing to keep in mind with these events is that they fire consistently. I would play down the importance of having these specifically tied to a platform concept. The importance of the events is to establish the current state and the intent of the state. For example if Appearing fires on android before "attach window" that's fine, Appearing is purely to indicate the intent to appear but it hasn't yet. Appeared will most likely be tied to native but there are some scenarios like dismissing a modal or switching tabs (make sure to test having more than 3 tabs) where it'll be more of a logical thing
- Shell Parts
- ListView and CollectionView

## Expected Result

User can subscribe to additional Page and View life cycle events.

Events are consistently fired across all platforms.

Developer rejoicing is heard in the streets. 😃 🎉  

## Related

https://forums.xamarin.com/discussion/84510/improved-life-cycle-support

paymicro:
### Description of Change ###

Added life-cycle events, such as `Loaded`, `Unloaded`, `BeforeAppearing` for the VisualElement and `Appeared`, `Disappeared` for the Page. 
Also changed the behavior of the `Appearing` and `Disappearing` events for the Page.

### Issues Resolved ###

- fixes #2210

### API Changes ###

Added:
 - event VisualElement.Loaded;
 - event VisualElement.Unloaded;
 - event VisualElement.BeforeAppearing;
 - event Page.Appeared;
 - event Page.Disappeared;

Changed:
 - event Page.Appearing;
 - event Page.Disappearing;

### Platforms Affected ###

- Core (all platforms)
- iOS
- Android
- UWP

Event call order:

| iOS | Android  | UWP  | 
|:----------------------:|:-------------------------:|:------------------------:| 
| image |  image |  image |
|  | LegacyPageEvents |  |
|  image |  image | image |

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/VisualElementTracker.cs
+++ b/Xamarin.Forms.Platform.Android/VisualElementTracker.cs
@@ -53,8 +53,8 @@ namespace Xamarin.Forms.Platform.Android
 			{
 				_element?.SendLoaded();
 
-				if (_element is IPageController page && !Application.Current.UseLegacyPageEvents)
-					page.SendAppeared();
+				if (!Application.Current.UseLegacyPageEvents)
+					(_element as IPageController)?.SendAppeared();
 
 				_loaded = true;
 
@@ -230,8 +230,8 @@ namespace Xamarin.Forms.Platform.Android
 			{
 				_element?.SendUnloaded();
 
-				if (_element is IPageController page && !Application.Current.UseLegacyPageEvents)
-					page.SendDisappeared();
+				if (!Application.Current.UseLegacyPageEvents)
+					(_element as IPageController)?.SendDisappeared();
 			}
 
 			_loaded = false;
@@ -480,21 +480,11 @@ namespace Xamarin.Forms.Platform.Android
 		{
 			public static readonly AttachTracker Instance = new AttachTracker();
 
-			public void OnViewAttachedToWindow(AView attachedView)
-			{
-				if (!(attachedView is IVisualElementRenderer renderer) || renderer.Tracker == null)
-					return;
-
-				renderer.Tracker.HandleViewAttachedToWindow();
-			}
+			public void OnViewAttachedToWindow(AView attachedView) 
+				=> (attachedView as IVisualElementRenderer)?.Tracker?.HandleViewAttachedToWindow();
 
 			public void OnViewDetachedFromWindow(AView detachedView)
-			{
-				if (!(detachedView is IVisualElementRenderer renderer) || renderer.Tracker == null)
-					return;
-
-				renderer.Tracker.HandleViewDetachedFromWindow();
-			}
+				=> (detachedView as IVisualElementRenderer)?.Tracker?.HandleViewDetachedFromWindow();
 		}
 	}
 }
--- a/Xamarin.Forms.Platform.Android/VisualElementTracker.cs
+++ b/Xamarin.Forms.Platform.Android/VisualElementTracker.cs
@@ -19,6 +19,7 @@ namespace Xamarin.Forms.Platform.Android
 		Context _context;
 
 		bool _disposed;
+		bool _loaded;
 
 		VisualElement _element;
 		bool _initialUpdateNeeded = true;
@@ -27,13 +28,11 @@ namespace Xamarin.Forms.Platform.Android
 
 		public VisualElementTracker(IVisualElementRenderer renderer)
 		{
-			if (renderer == null)
-				throw new ArgumentNullException("renderer");
+			_renderer = renderer ?? throw new ArgumentNullException("renderer");
 
 			_batchCommittedHandler = HandleRedrawNeeded;
 			_propertyChangedHandler = HandlePropertyChanged;
 
-			_renderer = renderer;
 			_context = renderer.View.Context;
 			_renderer.ElementChanged += RendererOnElementChanged;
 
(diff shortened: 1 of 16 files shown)


SamuelDebruyn:
Much needed additions, thanks! Now let's hope it gets merged :)

xleon:
Just a note: I made my own live-cycle hack to SendAppearing on iOS `ViewWillAppear` rather than `ViewDidAppear` for ReactiveUI to work better. The side effect is that :

     {
            base.OnAppearing();

            var insets = On<iOS>().SafeAreaInsets(); // <-- returns a wrong value (all insets are 0) 
     }

I´m looking forward for this to be merged but bear in mind that `On<iOS>().SafeAreaInsets();` should be tested.

samhouts:
@paymicro I think I want to take this change for 4.0.0, so please rebase it on 4.0.0. 

Also! Maybe we should add a new property to `Application`: `bool UseLegacyPageEvents`, default `false`. We can check that everywhere we changed the `Appearing`/`Disappearing` events. This would allow developers to opt out of the changed `Appear`ing events but maybe still benefit from the additional events. Thoughts? Is that feasible?

samhouts:
@paymicro This is looking good! The tests are not happy yet, though. Are you looking at that? Thanks!

samhouts:
Hmm, looks like most of the iOS tests are still failing. :(

paymicro:
Thanks @samhouts. Uitests began pass on iOS 👍

paymicro:
it seems that the addition of a VTO broke 4484 test...

Xamarin.Forms.Platform.Android/VisualElementTracker.cs
```diff
@@ -60,8 +78,14 @@ protected virtual void Dispose(bool disposing)
 
 			if (disposing)
 			{
-				SetElement(_element, null);
+				if (_loaded)
+				{
+					_element?.SendUnloaded();
+					if (_element is IPageController page && !Application.Current.UseLegacyPageEvents)
```
PureWeen: this could just call *HandleViewDetachedFromWindow*

it's the same code plus it sets loaded to false whereas the code here doesn't set loaded to false which might lead to double unloaded events?

samhouts:
@paymicro This is outstanding work! Thank you! You implemented this exactly how we asked you to in the original spec. After playing with it and diving into the implementation, @PureWeen and I think that we need to revisit our original spec. The event names could be more consistent (`Appeared` and `Loaded` are very similar events, for example, and could be the same name), and `BeforeAppearing` doesn't seem to offer any more value than `Appearing` does at this time. @PureWeen has some ideas, and he volunteered to work on it next week. Almost there! We'll still get this into 4.0. Thanks!!

PureWeen:
Worked moved over to this branch
https://github.com/xamarin/Xamarin.Forms/tree/fix_gh2210

New pr will be created from there

PR #3245, opened 2018-07-09
1. Android Resource manager now picks up image files stored in mipmap…

### LINKED ISSUE #3244: Android: ImageSource.FromFile does not pick up images if they are stored in a mipmap folder
### Description
If Android images are stored in mipmap folders (see attached screenshot), then ImageSource.FromFile does not pick the images up, since the associated resource lies in the Resources.Mipmap type as opposed to the Resources.Drawable type.

### Steps to Reproduce

1. Create a Xamarin Forms project using File -> New Project
2. Check that image resources are held in mipmap files, as per the attached screenshot. The F# template certainly does this.
3. Load an image using ImageSource.FromFile

### Expected Behavior

The image should be visible.

### Actual Behavior

The image is not visible.
### Basic Information

- Version with issue:
- Last known good version:
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Nuget Packages:
- Affected Devices:

### Screenshots

image

SpiegelSoft:
… directories as well as drawable directories. 2. The workaround for F# resources using the Resources_Drawable class no longer applies and has been removed.

### Issues Resolved ###

- fixes #3244

### Platforms Affected ###

- Core/XAML (all platforms)
- iOS
- Android
- UWP

### PR Checklist ###

- [ ] Has automated tests 


samhouts:
build --uitests

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -114,17 +114,21 @@ public static int GetStyleByName(string name)
 
 		public static void Init(Assembly masterAssembly)
 		{
-			DrawableClass = masterAssembly.GetTypes().FirstOrDefault(x => x.Name == "Drawable" || x.Name == "Resource_Drawable");
+			DrawableClasses = masterAssembly.GetTypes().Where(x => x.Name == "Drawable" || x.Name == "Mipmap").ToArray();
```
PureWeen: - I'm wondering if this should be setup to ensure that Drawable is the first element of the array. If mipmap is first that's going to cause an extra search for most scenarios

- also I'm thinking GetTypes should either be stored locally or looped over to set all of these types up. Each call to GetTypes probably causes some overhead

paymicro: To `Drawable` was always the first element you can sort the enumeration by name `.OrderBy(x => x.Name)`.

PureWeen: My concern there is getting to LINQ'y in code that's going to most likely run during startup 
So trying to optimize this code as much as possible would be ideal

paymicro: I think we can remove LINQ here and add lazy initialization. As well as to cache resource ID, so that each time not to iterate through all the properties.

PureWeen: Yea that looks like what I was envisioning on the Init part which I think is good enough for this PR

The IdHash part I'm wondering about. I would just be scared of that growing too much. 
I've been wanting to look at using an LruCache (https://developer.android.com/reference/android/util/LruCache) for storing bitmaps/drawables in these scenarios to improve with performance and to stay mindful of not creating an unbounded data structure. But I think that part can be a discussion for a different PR

samhouts:
@SpiegelSoft Please feel free to remove the do-not-merge when you submit your next set of changes. Thanks!

samhouts:
@SpiegelSoft Please reopen if you're able to work on this again. Thanks!

PR #3251, opened 2018-07-09
[Android] Improve default Button shadow & size appearance

### LINKED ISSUE #2664: [Android] Buttons with BackgroundColor shadow/size problems.
### Description

Depending on Xamarin.Forms version, Buttons with BackgroundColor set to something different than Default/Transparent differ from material design look.

### Steps to Reproduce

1. Create a project and add `<Button BackgroundColor="BlueViolet" />` to the main page.
2. Install different Xamarin.Forms versions and compare results.

### Expected Behavior

Buttons have material design look.

### Actual Behavior

2.5.0.122203: shadow is clipped because it's too big (similar to #2558)
2.5.0.280555: buttons are too short, shadows are colored.
2.5.1.444934+: buttons are too big, no shadows.

### Basic Information

- Last known good version: none.
- Platform Target Frameworks:
  - Android: 8.1

### Screenshots
_Click to see full size_

**2.5.0.122203**

**2.5.0.280555**

**2.5.1.444934+**

**2.5.0.122203**, **2.5.1.444934+** comparison

samhouts:
### Description of Change ###

When using the `DefaultPadding` and `DefaultShadow` platform specific properties, Button size and shadow now more closely match the default Android size and shadow. Note that this is not intended to be a pixel perfect match.

#### #1436 Before >> After

**WARNING: This is a visual breaking change from 2.5+.**

### Issues Resolved ###

- meant to fix #2664 

### API Changes ###

None

### Platforms Affected ###

- Android

- [Android] Buttons now have a slightly different shadow radius that more closely mimics Android's default Button image. 
- [Android] Button shadow color now defaults to a light gray instead of a shade of the Button color. 
- [Android] Button border is drawn outside of the Button instead of on top of it and can increase the size of the Button. 
- [Android] Button width is slightly increased.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/BorderDrawable.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/BorderDrawable.cs
@@ -36,9 +36,8 @@ namespace Xamarin.Forms.Platform.Android
 			set { _paddingTop = value; }
 		}
 
-
-		double BorderWidth => Math.Max(Button.IsSet(BorderElement.BorderWidthProperty) ? BorderElement.BorderWidth : .25, 0);
-		Color BorderColor => Button.IsSet(BorderElement.BorderColorProperty) && BorderElement.BorderColor != Color.Default ? BorderElement.BorderColor : Color.FromHex("#c5c5c5");
+		double BorderWidth => Math.Max(BorderElement.IsBorderWidthSet() ? BorderElement.BorderWidth : .25, 0);
+		Color BorderColor => BorderElement.IsBorderColorSet() && BorderElement.BorderColor != Color.Default ? BorderElement.BorderColor : Color.FromHex("#c5c5c5");
 
 		public BorderDrawable(Func<double, float> convertToPixels, Color defaultColor, bool drawOutlineWithBackground)
 		{
@@ -232,7 +231,7 @@ namespace Xamarin.Forms.Platform.Android
 			using (var paint = new Paint { AntiAlias = true })
 			using (var path = new Path())
 			{
-				float borderWidth = _convertToPixels(Button.BorderWidth);
+				float borderWidth = _convertToPixels(BorderElement.BorderWidth);
 
 				// adjust border radius so outer edge of stroke is same radius as border radius of background
 				float borderRadius = Math.Max(ConvertCornerRadiusToPixels(), 0);
--- a/Xamarin.Forms.Platform.Android/Renderers/BorderDrawable.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/BorderDrawable.cs
@@ -26,7 +26,7 @@ namespace Xamarin.Forms.Platform.Android
 
 		float PaddingLeft
 		{
-			get { return (_paddingLeft / 2f) + _shadowDx; }
+			get { return (_paddingLeft / 2f) + _shadowDx - _shadowRadius; }
 			set { _paddingLeft = value; }
 		}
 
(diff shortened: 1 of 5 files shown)


samhouts:
build --uitests

Xamarin.Forms.Platform.Android/ButtonBackgroundTracker.cs
```diff
@@ -77,10 +77,10 @@ public void UpdateDrawable()
 
 				var useDefaultShadow = _button.OnThisPlatform().UseDefaultShadow();
 
-				float shadowRadius = useDefaultShadow ? 2 : _nativeButton.ShadowRadius;
+				float shadowRadius = useDefaultShadow ? 5 : _nativeButton.ShadowRadius;
 				float shadowDx = useDefaultShadow ? 0 : _nativeButton.ShadowDx;
-				float shadowDy = useDefaultShadow ? 4 : _nativeButton.ShadowDy;
-				AColor shadowColor = useDefaultShadow ? _backgroundDrawable.PressedBackgroundColor.ToAndroid() : _nativeButton.ShadowColor;
+				float shadowDy = useDefaultShadow ? 1 : _nativeButton.ShadowDy;
+				AColor shadowColor = useDefaultShadow ? Color.FromHex("#c4c4c4").ToAndroid() : _nativeButton.ShadowColor;
```
hartez: Do we have a reference for this color code?

Xamarin.Forms.Platform.Android/Renderers/ButtonDrawable.cs
```diff
@@ -36,6 +36,9 @@ float PaddingTop
 			set { _paddingTop = value; }
 		}
 
+		double BorderWidth => Math.Max(Button.IsSet(Button.BorderWidthProperty) ? Button.BorderWidth : .25, 0);
+		Color BorderColor => Button.IsSet(Button.BorderColorProperty) && Button.BorderColor != Color.Default ? Button.BorderColor : Color.FromHex("#c5c5c5");
```
hartez: Same here - where do these colors come from? Is there a guide or source code we can point to?

samhouts:
build --uitests

samhouts:
build --uitests

Xamarin.Forms.Platform.Android/Renderers/BorderDrawable.cs
```diff
@@ -221,25 +228,21 @@ public RectF GetPaddingBounds(int width, int height)
...
 			{
 				float borderWidth = _convertToPixels(BorderElement.BorderWidth);
-				float inset = borderWidth / 2;
 
 				// adjust border radius so outer edge of stroke is same radius as border radius of background
-				float borderRadius = Math.Max(ConvertCornerRadiusToPixels() - inset, 0);
+				float borderRadius = Math.Max(ConvertCornerRadiusToPixels(), 0);
 
 				RectF rect = new RectF(0, 0, width, height);
-				rect.Inset(inset + PaddingLeft, inset + PaddingTop);
```
rmarinho: @samhouts after some testing , adding this back and subtracting the inset above that we remove, fix the issue with the corner radius that I noticed in my review, at the same time it makes the border width look bigger

PureWeen: I'm seeing the same thing as @rmarinho  on this one
I've been playing with this for a little bit and here's where I am so far

https://gist.github.com/PureWeen/855e7acdffe2087553a75881f95d23f6
Basically I changed the insets a bit because I wasn't quite sure on the intent of the math in the PR

I inset the background just based on padding and shadow
And then I inset the Outline the same but then plus half the borderwidth which just shrinks it down so the outer edge of the stroke lines up with the background

The effect of this is that if you set a really wide stroke you'll still see a shadow

mattleibow:
This is probably something that we want to do: https://github.com/xamarin/Xamarin.Forms/issues/4185

samhouts:
I think this PR introduces more problems than it solves. Closing for now.

PR #3263, opened 2018-07-10
[UWP] Added setting of the directory for the images

### LINKED ISSUE #2235: Improved image organization
## Summary
I would like to propose an improved image asset organization functionality for Xamarin.Forms 3.0.0, especially for UWP, but optionally on other OS as well.

### Default UWP image location change

Xamarin.Forms developers targeting UWP are currently forced to add their image assets directly in the root folder of the UWP project. This quickly becomes very unpleasant experience with tens of images right in the root folder next to files that the developer needs to refer to quickly like `App.xaml` of `Package.appxmanifest` and makes the project look unorganized. While using a shared project to have the images may provide a partial relief of this problem, it is still not a very nice user experience.

Usually in UWP development images are placed in the **Assets** folder. It is not a must-be location like Assets Catalogs in iOS or `drawable` on Android, but by convention most developer place image assets there. Therefore I suggest making this the default location for images instead of the project root. This could be done quite easily by modifying the `Xamarin.Forms.Platform.UWP.FileImageSourceHandler`. Although this is a breaking change, I think it is one worth making in the long run, as it makes the Xamarin.Forms development for UWP much more pleasant.

### (pseudo)Subfolders

Going further, it could be possible to add subfolder support for images as well. This would allow developers to organize their images into subfolders and refer to them as for example:

```
<Image Source="Icons/Help.png" />
```

This works in UWP, but the clear problem is that Android and iOS have the limitation of having images on one level (like all directly in the `drawable` folder) and subfolders are not allowed. For this I propose that the `/` characters in image paths would be replaced by `_` for these platforms - which means another simple modification in the platform-specific `FileImageSourceHandler` classes. Although this still means the images have to be in one single location, we can create a "subfolder-like" structure, which is easier to navigate. In addition - this is an optional feature - the developer who does not want to use this subfolder structure can still use normal filenames, even utilize underscores within them without a problem.

I have been using this approach in my Xamarin.Forms project quite successfully so far (utilizing a custom markup extension to perform the `/` to `_` replacement). Example result of this in the project folders:

**Android**
image

**UWP**
image

## Intended Use Case

The first part of this proposal - changing default image folder to `Assets` folder in UWP is a breaking change that will make the Xamarin.Forms UWP project root folders much less cluttered.

The pseudo-subfolder feature for images is an optional feature which allows developers to use `/` characters when referring to images in XAML, which is then evaluated to subfolders on UWP and underscore-based prefixes on Android and iOS.

### LINKED ISSUE #3228: Feature Request - Default Search Directory for UWP Icons (Platform Specific)
I would really like to put all of my icons/images inside of the `Assets` folder, however this creates some not so great looking code:

```
<Image>
                <Image.Source>
                    <OnPlatform x:TypeArguments="ImageSource">
                        <On Platform="iOS" Value="icon_twitter.png"/>
                        <On Platform="Android" Value="icon_twitter.png"/>
                        <On Platform="UWP" Value="Assets/icon_twitter.png"/>
                    </OnPlatform>
                </Image.Source>
            </Image>
```

Ideally I would like to do:

```xml
<Image Source="icon_twitter.png"/>
```

I could do this if I put it into the UWP root, but that clutters up the project if it is large.

I would like something like:

```
On<UWP>().SetImageSearchDirectory("Assets");
```

Then my nice pretty Image would work.

paymicro:
### Description of Change ###

- For really like to put all of application images inside of the Assets folder, added the platform specific `SetImageDirectory` property. This property set path for all images on the page or on any single image.
- Added support pseudo folders on Android and iOS platforms. 
`<Image Source="assets/logo.png" />` on Android/iOS will actually use the file `assets_logo.png`.

**Screencast:** http://recordit.co/TLna8uvEQM

### Issues Resolved ###

- fixes #3228
- fixes #2235

### API Changes ###

None

### Platforms Affected ###

- UWP
- iOS
- Android

It is possible to change the folder with images using cleaner code in Xaml

### PR Checklist ###

- [ ] Has automated tests 


paymicro:
> what happens when you add an Image to a Page later in the lifecycle? 
Thanks for the excellent comment! I've added support for dynamic adding of images.

jsuarezruiz:
@paymicro LGTM. Although there are some conflicts, could you check it?.

PR #3264, opened 2018-07-10
[UWP] Fix NullRef. in ListViewRenderer.OnCollectionChanged after disposed...

justremotephone:
After changing from 3.0 to 3.1 some UWP pages crashes the app with a NullRef. This is because a new Async-Operation was added in ListViewRenderer.OnCollectionChanged accessing the ListView property which may be null after the ListViewRenderer was disposed.

### Platforms Affected ###

- UWP


rmarinho:
Hi, we will need a issue to look at , your fix could be right, but could be just hidding the underlying problem. 

Open a issue with the detail, or add a test case to your pr so we can test.

Thanks

PR #3268, opened 2018-07-10
[UWP] store alert to local variable for thread safety

### LINKED ISSUE #3250: Intermittent null exception in UWP OnPageAlert()
### Description

I'm getting an intermittent null exception from Xamarin.Forms.Platform.UAP/PlatformUWP.cs OnPageAlert().

The culprit appears to be from the fix for #2607 "[UWP] Deadlock if you rapid fire alerts".

This code is not thread safe, specifically this block:
```
while (s_currentAlert != null)
{
	await s_currentAlert;
}

s_currentAlert = ShowAlert(alertDialog);
options.SetResult(await s_currentAlert.ConfigureAwait(false));
s_currentAlert = null;
```
https://github.com/xamarin/Xamarin.Forms/blob/8c23ad6c893b8e5b7f3fd2573c4ebde0a0441950/Xamarin.Forms.Platform.UAP/PlatformUWP.cs#L107

If two or more alerts are queued up using Device.BeginInvokeOnMainThread() then, because ConfigureAwait(false) continues in a new thread, when the first one finishes it may set s_currentAlert to null in between the null check and the await of the while loop for the second one.

A simple fix may be to:
1. Capture the s_currentAlert using in the while loop in a temporary variable
```
var currentAlert = s_currentAlert;
while (currentAlert != null)
{
  await currentAlert;
  currentAlert = s_currentAlert
}
```
2. Capture the result of ShowAlert()
```
var showAlert = ShowAlert(alertDialog);
s_currentAlert = showAlert;
options.SetResult(await showAlert.ConfigureAwait(false));
s_currentAlert = null;
```

### Steps to Reproduce

1. Queue up two or more alerts
2. Close the first alert, the null exception is intermittent

### Expected Behavior

There should be no null exception.

### Actual Behavior

There is an intermittent null exception.

### Basic Information

- Last known good version: 2.5.0.280555 (likely through 3.0.0.561731)
- IDE: Visual Studio 2017 15.7.3
- Platform Target Frameworks:
  - UWP:  16299

PureWeen:
### Description of Change ###

The while loop on s_currentAlert wasn't thread safe so it could be set to null between the check and the await which meant it would be awaiting on null and throw an exception.

Added text to description but the clicking to cause the issue has to happen too quickly for the UI test to detect

### Issues Resolved ###
- fixes #3250 

### Platforms Affected ###
- UWP

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla43469.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla43469.cs
@@ -14,10 +14,10 @@ namespace Xamarin.Forms.Controls.Issues
 	[Issue(IssueTracker.Bugzilla, 43469, "Calling DisplayAlert twice in WinRT causes a crash", PlatformAffected.WinRT)]
 	public class Bugzilla43469 : TestContentPage
 	{
-		const string kButtonText = "Click to call DisplayAlert six times";
+		const string kButtonText = "Click to call DisplayAlert six times. Click as fast as you can to close them as they popup to ensure it doesn't crash.";
 		protected override void Init()
 		{
-			var button = new Button { Text = "Click to call DisplayAlert six times" };
+			var button = new Button { Text = kButtonText };
 
 			button.Clicked += async (sender, args) =>
 			{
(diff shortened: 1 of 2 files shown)


PureWeen:
build --uitests

PR #3300, opened 2018-07-14
[iOS] Implement `ForceUpdateSize` for `TableView`

### LINKED ISSUE #2842: ViewCell in TableView not adapting to changed size on iOS
### Description
When changing the height of the content in a ViewCell inside a TableView, the cell height is adapted correctly on Android and UWP but not on iOS. Instead the content of the specific cell is larger than the cell itself and appears below the cell's lower border.

### Steps to Reproduce

1. Add a ViewCell to a TableView(HasUnevenRows="True" Intent="Settings").
2. As the cells content set two labels inside a grid or StackLayout below each other. Set the visibility of the first label to true, the second one to false.
3. Add another cell below the first cell in the same table section.
4. Add a gesture recognizer to the Grid/StackLayout that sets the visibility of the second label to true, when the cell is tapped.

### Expected Behavior
The cell adapts its size as on Android or UWP.
### Actual Behavior
The cell content overlaps the lower border of the cell.
### Basic Information

- Version with issue: 3.0.0.482510
- Last known good version: -
- IDE: Visual Studio 2017 Enterprise 15.1
- Platform Target Frameworks: 
  - iOS:  11.3
- Nuget Packages: Just Xamarin.Forms
- Affected Devices: Simulator as well as the physical devices I tested (iPhone 4S and 8)

### Screenshots
Before:
before

After:
after

### Reproduction Link

[TableViewGrowingCellExample.zip](https://github.com/xamarin/Xamarin.Forms/files/2042826/TableViewGrowingCellExample.zip)

samhouts:
### Description of Change ###

`TableView` cells implicitly update size on Android and UWP, but we must force them on iOS. This implements the same pattern used for `ListView`'s `ForceUpdateSize` on `TableView`.

Note that since this drastically reduces performance, you must explicitly call `ForceUpdateSize` on a cell to update its size.

### Issues Resolved ###

- fixes #2842

### API Changes ###

Added:
- iOS `TableViewModelRenderer`  `protected void WireUpForceUpdateSizeRequested`

### Platforms Affected ###

- Core (all platforms)
- iOS

None

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/TableViewModelRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/TableViewModelRenderer.cs
@@ -9,6 +9,8 @@ namespace Xamarin.Forms.Platform.iOS
 	{
 		readonly Dictionary<nint, Cell> _headerCells = new Dictionary<nint, Cell>();
 
+		EventHandler _onForceUpdateSizeRequested;
+
 		protected bool HasBoundGestures;
 		protected UITableView Table;
 
@@ -32,6 +34,9 @@ namespace Xamarin.Forms.Platform.iOS
 			var cell = View.Model.GetCell(indexPath.Section, indexPath.Row);
 
 			var nativeCell = CellTableViewCell.GetNativeCell(tableView, cell);
+
+			WireUpForceUpdateSizeRequested(cell, nativeCell, tableView);
+
 			return nativeCell;
 		}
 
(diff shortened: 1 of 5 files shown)


samhouts:
build --uitests

Xamarin.Forms.Platform.iOS/Renderers/TableViewModelRenderer.cs
```diff
@@ -100,6 +105,20 @@ public override string TitleForHeader(UITableView tableView, nint section)
 			return View.Model.GetSectionTitle((int)section);
 		}
 
+		protected void WireUpForceUpdateSizeRequested(ICellController cell, UITableViewCell nativeCell, UITableView tableView)
```
PureWeen: Is this code addition necessary?

It seems like the code here
https://github.com/xamarin/Xamarin.Forms/blob/14af670f5a200d164284fa4f0b3e4f7ce03b53df/Xamarin.Forms.Platform.iOS/Cells/CellRenderer.cs#L60

Takes care of the Reload.  I commented this addition here out and the UI Test still passed because the ReloadRows inside CellRenderer takes care of it.

The addition to *Cell.cs* seems to be enough

PureWeen:
build --uitests

mattclarkie:
Would this exhibit the issue raised on this https://bugzilla.xamarin.com/show_bug.cgi?id=40432 over 2.5 years ago and marked as "Confirmed" but went nowhere (not even migrated over to GitHub) that means using ForceUpdateSize causes the list to fly allover the place.

Issue #2959, opened 2018-06-06
[UWP] AccessViolationException when adding items with the appearance of the item.


paymicro:
### Description

When added items in ItemSource in `OnItemAppearing` throws `AccessViolationException`

### Steps to Reproduce

1. Create `ListView` with `ItemsSource`
2. Add event `ItemAppearing`
3. In event body add new item to `ItemsSource`

### Expected Behavior

Items are added without errors.

### Actual Behavior

Throw `AccessViolationException`

### Basic Information

- Version with issue: on `master` branch

### Screenshots

https://prnt.sc/jrkdsi

### Reproduction Link

https://github.com/paymicro/Xamarin.Forms/commit/eed5ab74e682a8f9936daa0e78447cefad64a668

Relation bug: #2832

nbevans:
This is still an issue. AccessViolationException's being thrown, intermittently, in certain nuanced scenarios with ListView on UWP.

samhouts:
This issue doesn't seem to have had any activity in a long time. We're working on prioritizing issues and resolving them as quickly as we can. To help us get through the list, we would appreciate an update from you to let us know if this is still affecting you on the latest version of Xamarin.Forms, since it's possible that we may have resolved this as part of another related or duplicate issue. If we don't see any new activity on this issue in the next 30 days, we'll evaluate whether this issue should be closed. Thank you!

samhouts:
Since we haven't heard from you in more than 30 days, we hope this issue is no longer affecting you. If it is, please reopen this issue and provide the requested information so that we can look into it further. Thank you!

PR #2512, opened 2018-04-20
[UWP] Fixed sync of ListView items with ListViewProxy

paymicro:
### Description of Change ###

`ItemsSource` wrapped in an `ObservableCollection`
Added processing for adding and removing items.
Screencast: https://www.screencast.com/t/LrPoEiM8a6v0

### Bugs Fixed ###

Fixes #2485

### API Changes ###

--

### Behavioral Changes ###

Correct Items shown when the collection is changed

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [ ] Changes adhere to coding standard
- [ ] Consolidate commits as makes sense

Issue #2485, opened 2018-04-17
[UWP] Removing items from Listview shows white rows


rmarinho:
### Description
Seems that removing items for the collection leaves a empty white row. 

### Steps to Reproduce

1. Open control gallery issues page
2. Go to issue B56771
3. Click Remove 2

### Expected Behavior
Should not see white space

### Actual Behavior
We see a empty white row.

### Basic Information

- Version with issue: 3.0.0
- Last known good version:
- IDE: VS WIN
- Platform Target Frameworks: 
  - UWP:   

### Screenshots
<img width="1154" alt="screenshot 2018-04-17 12 05 06" src="https://user-images.githubusercontent.com/1235097/38866203-04f4c0e6-4238-11e8-9824-de5306d7cb19.png">

rmarinho:
Needs investigatiion but seems a regression

PR #3323, opened 2018-07-17
[UWP] Fixes CollectionChanged events in ListView

### LINKED ISSUE #3273: UWP Regression: Drag and drop reordering not firing CollectionChanged
### Description
Reordering UWP list view using drag and drop, CollectionChanged events are not fired. This appears to have regressed between 3.0.0.482510 and 3.0.0.530893. In 482510 it was firing both a Remove and an Add event. In 530893 it stopped firing any events on a drag and drop reorder. The latest 3.1 version does not fix the issue.

### Steps to Reproduce
1. Run the attached app using a UWP ListView renderer with drag and drop reordering enabled on build 482510. Note that Add and Remove events fire on a reorder.
2. Run the attached app on build 530893. Note that **_no_** events fire on a reorder.

### Expected Behavior
Add and Remove events should be dispatched on a drag and drop reorder.

### Actual Behavior
No events are dispatched.

### Basic Information
- Version with issue: >= 3.0.0.530893
- Last known good version: 3.0.0.482510
- IDE: VS2017
- Platform Target Frameworks: 
  - UWP:  10.0.16299.309
- Affected Devices: UWP devices

### Reproduction Link
[ReorderNotCallingCollectionChanged.zip](https://github.com/xamarin/Xamarin.Forms/files/2181722/ReorderNotCallingCollectionChanged.zip)

paymicro:
### Description of Change ###

`CollectionChanged` events are fired, when collection changed in user code.

### Issues Resolved ###

- fixes #3273 

### API Changes ###

/

### Platforms Affected ###

- UWP

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
+++ b/Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
@@ -11,6 +11,7 @@ using Xamarin.Forms.Platform.UWP;
 [assembly: ExportRenderer(typeof(Xamarin.Forms.Controls.Bugzilla42602.TextBoxView), typeof(Xamarin.Forms.ControlGallery.WindowsUniversal.TextBoxViewRenderer))]
 [assembly: ExportRenderer(typeof(Issue1683.EntryKeyboardFlags), typeof(EntryRendererKeyboardFlags))]
 [assembly: ExportRenderer(typeof(Issue1683.EditorKeyboardFlags), typeof(EditorRendererKeyboardFlags))]
+[assembly: ExportRenderer(typeof(Issue3273.SortableListView), typeof(SortableListViewRenderer))]
 namespace Xamarin.Forms.ControlGallery.WindowsUniversal
 {
 	public class EntryRendererKeyboardFlags : EntryRenderer
@@ -33,6 +34,24 @@ namespace Xamarin.Forms.ControlGallery.WindowsUniversal
 			Control.TestKeyboardFlags(((Issue1683.EditorKeyboardFlags)Element).FlagsToSet);
 
 
+		}
+	}
+
+	public class SortableListViewRenderer : ListViewRenderer
+	{
+		protected override void OnElementChanged(ElementChangedEventArgs<ListView> e)
+		{
+			base.OnElementChanged(e);
+
+			if (e.NewElement != null)
+			{
+				var control = Control as Windows.UI.Xaml.Controls.ListView;
+
+				control.AllowDrop = true;
+				control.CanDragItems = true;
+				control.CanReorderItems = true;
+				control.ReorderMode = ListViewReorderMode.Enabled;
+			}
 		}
 	}
 
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla56771.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla56771.cs
@@ -39,6 +39,11 @@ namespace Xamarin.Forms.Controls.Issues
 
 		protected override void Init()
 		{
+			data.CollectionChanged += (_, e) =>
+			{
+				var log = $"<{DateTime.Now.ToLongTimeString()}> {e.Action} action fired.";
+				System.Diagnostics.Debug.WriteLine(log);
+			};
 			var label = new Label { Text = "Click the Add 2 button." };
 			var button = new Button
 			{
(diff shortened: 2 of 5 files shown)


samhouts:
build --uitests

paymicro:
Hi @PureWeen. Thanks for review!

> I don't think the added UITest quite reflects the issue reported. The issue reported is using a renderer to enable reordering on the list

I tested the behavior when you turn on drag & drop and it is no different from the call `Move` elements of `SourceItems`. So I think that this test completely reflects the issue.

I refactored the code in PR. And now wrapping in `ObservableCollecion<object>` happens only if it is necessary.

PureWeen:
> I tested the behavior when you turn on drag & drop and it is no different from the call Move elements of SourceItems. So I think that this test completely reflects the issue.

The issue is where the event starts from. When you're moving things around in the UI the propagation is

*Native List View => Moves Items on Its proxy list (which is different than the list the user knows about)=> propagates => and now this change needs to propagate to the list the user knows about*

The UI Test here goes the opposite direction
*User calls Move on the list they know about => changes propagate to proxy => proxy propagates to Native List View*

The other reason I don't feel like the UI test properly reflects the changes is that if I roll back the changes in the PR the UI Test still gives me a "Success" 

I checked in a variation of the unit test here. 
https://github.com/xamarin/Xamarin.Forms/tree/fix-gh3273
I rolled back the changes just to demonstrate how clicking the button will fire the CollectionChanged Event and set the Label to Success. Whereas when dragging the items around the CollectionChanged never fires which is what the original issue is about

That being said your latest changes look like they propagate correctly (I still need to digest them a bit) though it looks like there are some cases a null ItemSource causes NRE Exception in the code.

For example if you just add a blank list view
```C#
			Content = new StackLayout
			{
				Children = {
					statusLabel,
					actionLabel,
					new Button {
						Text = "Move items",
						Command = new Command(() =>
						{
							actionLabel.Text = string.Empty;
							statusLabel.Text = "Failed";
							Items.Move(0, 1); })
					},
					listView,
					new ListView()
				}
			};
```

It causes NRE

paymicro:
@PureWeen thank you for such a comprehensive review and for your UI test! 👍  This helped me a lot. I hope that ListView will work more correctly =)

PureWeen:
@paymicro  This resorting from the UWP side of things has made this an interesting one :-/ 

Does fixing it this way unwind this PR a bit too far?
https://github.com/xamarin/Xamarin.Forms/pull/2512

Which fixed this
https://github.com/xamarin/Xamarin.Forms/issues/2485

I think to stay inline with the PR above and the issues that it fixed this logic has to be inverted
```C#
				if (IsObservableCollection(Element?.ItemsSource.GetType()))
				{
					_collection = (IList)Element?.ItemsSource;
					_collectionIsWrapped = false;
				}
				else
				{
					_collection = new ObservableCollection<object>();
					foreach (var item in Element.ItemsSource)
						_collection.Add(item);
					_collectionIsWrapped = true;
				}
```

And we only wrap collections that implement INotifyCollectionChanged which I realize conflicts with the intent of this PR.

It seems like the only way to make all the issues happy would be to
- wrap an ItemSource if it implements INotifyCollectionChanged
- If the ItemSource  then also IList then changes made to the Wrapper need to propagate to the users list

mattrichnz:
Great work @paymicro @PureWeen!

Love the detection  of whether the source collection is observable vs wrapping it in a proxy collection. Having two disconnected ObservableCollection's didn't really make sense. This makes ListView behave much more like the native control (and as it did before!).

mattrichnz:
@paymicro @PureWeen Just following up - what is the current status on this PR?

paymicro:
@PureWeen this fix correctly works on UItest 56771. ( #2512 #2485 )
UWP ListView does not work correctly with the inheritors of the `ObservableCollection` class, therefore not enough wrap an ItemSource if it implements `INotifyCollectionChanged`. We need to wrap everything that is not an `ObservableCollection`.
I can add synchronization if the original  ItemSource implements `INotifyCollectionChanged`, but is not an `ObservableCollection` class. It necessary to do?

paymicro:
> This change is causing Issue1975 to crash on UWP

This looks like issue https://github.com/xamarin/Xamarin.Forms/issues/2959. I tried to fix it, but without success. Maybe it's crash inside UWP...

mattrichnz:
@paymicro @PureWeen Thanks for all your hard work on this. What is left to do here?

jassmith:
build --uitests

PureWeen:
Alright @paymicro  @samhouts   let me know what you think of the changes I just checked in

I removed the WinRT binding hack that was being used and that fixed the AccessViolationError
From what I can tell going via binding was keeping the WeakEventHandler alive in ListProxy longer then it needed to be

I also added a missing using statement. I do agree that there seems to be some odd issue internally with how CVS interacts with LV.  If I set the ItemsSource on the UWP ListView to null it throws an AccessViolationException still

I ran all the UWP tests marked as ListView and they all passed 

I don't think this is related to #2959

samhouts:
build --uitests

PureWeen:
build --uitests

PureWeen:
rebased to master to fix those android targeting errors

PureWeen:
build --uitests

mattrichnz:
@samhouts Will this make it into 3.2.0 RC?

mattrichnz:
I know this is slated for 3.3.0, but any chance of this making a 3.2.0 service release?

PR #3361, opened 2018-07-20
[iOS] Remove iOS Platform Casts

hartez:
### Description of Change ###

The iOS platform has a few places where `Element.Platform` is cast to the native iOS `Platform` to call helper methods which don't actually need anything from the native `Platform` class. 

This change makes those helper methods static and moves them to their own class. This removes the need for IPlatform in native iOS code.

### Issues Resolved ###

None; laying the groundwork for removing `Platform` from `Element`.

### API Changes ###

None

### Platforms Affected ###

- iOS

None 

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.iOS/DisposeHelpers.cs
+++ b/Xamarin.Forms.Platform.iOS/DisposeHelpers.cs
@@ -1,0 +1,52 @@
+namespace Xamarin.Forms.Platform.iOS
+{
+	internal static class DisposeHelpers
+	{
+		internal static void DisposeModalAndChildRenderers(this Element view)
+		{
+			IVisualElementRenderer renderer;
+
+			foreach (VisualElement child in view.Descendants())
+			{
+				renderer = Platform.GetRenderer(child);
+				child.ClearValue(Platform.RendererProperty);
+
+				if (renderer != null)
+				{
+					renderer.NativeView.RemoveFromSuperview();
+					renderer.Dispose();
+				}
+			}
+
+			renderer = Platform.GetRenderer((VisualElement)view);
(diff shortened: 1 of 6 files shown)


Xamarin.Forms.Platform.iOS/DisposeHelpers.cs
```diff
@@ -0,0 +1,52 @@
+namespace Xamarin.Forms.Platform.iOS
+{
+	internal static class DisposeHelpers
+	{
+		internal static void DisposeModalAndChildRenderers(this Element view)
```
samhouts: Yay for not saying "Model" anymore and confusing the heck out of me.

StephaneDelcroix: the class is `internal` (no need to specify it, btw), so the member can be `public`. it's slightly less confusing...

Xamarin.Forms.Platform.iOS/DisposeHelpers.cs
```diff
@@ -0,0 +1,52 @@
...
+
+				renderer.NativeView.RemoveFromSuperview();
+				renderer.Dispose();
+			}
+			view.ClearValue(Platform.RendererProperty);
+		}
+
+		internal static void DisposeRendererAndChildren(this IVisualElementRenderer rendererToRemove)
+		{
+			if (rendererToRemove == null)
```
StephaneDelcroix: does this really happen ? as it's an extension method, it's like calling (null as IVER).DisposeRendererAndChildren();

it works, but whoever does that deserves a good old NRE instead of returning null

PR #3372, opened 2018-07-22
[WPF] Memory leak when you pop a TabbedPage

### LINKED ISSUE #3267: [WPF] Memory leak when you pop a TabbedPage
### Description

If you use a NavigationPage and push and then pop a TabbedPage, the renderers for elements on the first child of the TabbedPage are instantiated again after it has been removed.

### Steps to Reproduce

1. use a NavigationPage as MainPage
2. PushAsync a TabbedPage
3. go back from the TabbedPage so that it should be popped

### Expected Behavior

All renderers needed by the TabbedPage and any children should be disposed and not instantiated again.

### Actual Behavior

The renderers of the current child of the TabbedPage are first disposed and then the renderers of the first child are instantiated again.

### Basic Information

- IDE: VS 2017 15.8, Preview 3
- Platform Target Frameworks:
  - WPF, .NET Framework 4.6.1

### Reproduction Link

[Repro.WPFTabbedPage.zip](https://github.com/xamarin/Xamarin.Forms/files/2180740/Repro.WPFTabbedPage.zip)
This repro project uses a CustomPage with its own CustomPageRenderer to show the problem. It displays some messages in the lower half of the screen. These show, when a page is pushed/popped by the NavigationPage and which methods are called in the CustomPageRenderer. It all looks fine while you open the TabbedPage and switch between children. But when you go back, the renderers for the first child page are instantiated again.

mohachouch:
### Description of Change ###

This PR fix the memory leak when you pop a TabbedPage on WPF

### Issues Resolved ###

- fixes #3267

### Platforms Affected ###

- WPF

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.WPF/Renderers/ViewRenderer.cs
+++ b/Xamarin.Forms.Platform.WPF/Renderers/ViewRenderer.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Diagnostics;
 using System.Windows;
 using System.Windows.Controls;
 using WControl = System.Windows.Controls.Control;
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.WPF/Renderers/ViewRenderer.cs
```diff
@@ -161,19 +162,30 @@ protected void SetNativeControl(TNativeElement native)
...
 			Control.GotFocus += OnGotFocus;
 			Control.LostFocus += OnLostFocus;
 
 			UpdateBackground();
 			UpdateAlignment();
 		}
+		
+		private void Control_Loaded(object sender, RoutedEventArgs e)
+		{
+			Debug.WriteLine("Control_Loaded : " + this.Control.GetType());
```
rmarinho: Can we remove this Debug info

PR #3388, opened 2018-07-24
[Core] Fixed computing position of children in AbsoluteLayout

### LINKED ISSUE #2368: [UWP] Absolute Layout is not correctly positioning element
### Description

While fixing this issue, please also add instructions to the test case.

### Steps to Reproduce

1. Run Issue1914 in Control Gallery

### Expected Behavior

Bank image should show up in the center of the screen and rotating

### Actual Behavior

The screen is blank

### Basic Information

- Version with issue:3.0.0
- Last known good version:

paymicro:
### Description of Change ###

When fired `OnMeasure` in `AbsoluteLayout` with a proportional bounds original size of children divides on the bounds. It was wrong, because all manipulation with `bounds` already have in the function `LayoutChildren`.
If bounds is equals 1 this issue is not noticeable, but if bounds is far from the standard values it may happen that the boundaries of child will be incorrectly calculated.

### Issues Resolved ###

- fixes #2368

### API Changes ###

/

### Platforms Affected ###

- Core (all platforms)

### Visual Changes ###

Screencasts: 
UWP: http://recordit.co/Mz9KRVoAGa
Android: http://recordit.co/RP7NpKLEnK

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/AbsoluteLayoutTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/AbsoluteLayoutTests.cs
@@ -174,7 +174,7 @@ namespace Xamarin.Forms.Core.UnitTests
 
 			var sizeReq = abs.GetSizeRequest (double.PositiveInfinity, double.PositiveInfinity);
 
-			Assert.AreEqual (new Size (200, 40), sizeReq.Request);
+			Assert.AreEqual (new Size (100, 20), sizeReq.Request);
 			Assert.AreEqual (new Size (0, 0), sizeReq.Minimum);
 		}
 
@@ -196,7 +196,7 @@ namespace Xamarin.Forms.Core.UnitTests
 
 			var sizeReq = abs.GetSizeRequest (double.PositiveInfinity, double.PositiveInfinity);
 
-			Assert.AreEqual (new Size (210, 60), sizeReq.Request);
+			Assert.AreEqual (new Size (110, 40), sizeReq.Request);
 			Assert.AreEqual (new Size (10, 20), sizeReq.Minimum);
 		}
 
--- a/Xamarin.Forms.Core/AbsoluteLayout.cs
+++ b/Xamarin.Forms.Core/AbsoluteLayout.cs
@@ -177,41 +177,29 @@ namespace Xamarin.Forms
 			double minWidth = width;
 			double minHeight = height;
 
-			if (!widthIsProportional && bounds.Width != AutoSize)
+			if (!widthIsProportional)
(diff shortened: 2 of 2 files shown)


samhouts:
build --uitests

samhouts:
@paymicro would you mind rebasing? Thanks!

samhouts:
build --uitests

samhouts:
This change makes me nervous, since the original expectation was in place for about 6 years. I'm going to have to spend some time testing it. Thanks!

Xamarin.Forms.Core.UnitTests/AbsoluteLayoutTests.cs
```diff
@@ -174,7 +174,7 @@ public void SizeRequestWithRelativeChild ()
 
 			var sizeReq = abs.GetSizeRequest (double.PositiveInfinity, double.PositiveInfinity);
 
-			Assert.AreEqual (new Size (200, 40), sizeReq.Request);
+			Assert.AreEqual (new Size (100, 20), sizeReq.Request);
```
StephaneDelcroix: what was wrong with this test ?

Xamarin.Forms.Core.UnitTests/AbsoluteLayoutTests.cs
```diff
@@ -196,7 +196,7 @@ public void SizeRequestWithRelativeSizeChild ()
 
 			var sizeReq = abs.GetSizeRequest (double.PositiveInfinity, double.PositiveInfinity);
 
-			Assert.AreEqual (new Size (210, 60), sizeReq.Request);
+			Assert.AreEqual (new Size (110, 40), sizeReq.Request);
```
StephaneDelcroix: and with this one ?

Xamarin.Forms.Core.UnitTests/AbsoluteLayoutTests.cs
paymicro: When calculating `SizeRequest` no longer divided the size of child element with proportional width/height.

Xamarin.Forms.Core.UnitTests/AbsoluteLayoutTests.cs
paymicro: as above, but because the offset is not proportional, it is added to the size (100 + 10, 20 + 20)

rmarinho:
This is breaking we can't take it. The fix for the issue is just use the correct propotional value.

PR #3400, opened 2018-07-25
[UWP] Fixes call Tapped event in EntryCellTextBox

### LINKED ISSUE #2318: [UWP] EntryCell.Tapped does not fire when within a TableView
### Description

Migrated from https://bugzilla.xamarin.com/show_bug.cgi?id=32040

### Steps to Reproduce

1. Navigate to the UI test for 32040 in the Control Gallery.
2. Tap the "Click Here" Label in the EntryCell.

### Expected Behavior

The text in the Entry should change to "Tapped".

### Actual Behavior

The text in the Entry is selected, but does not change.

### LINKED ISSUE #4901: [UWP] Tapped does not fire on ViewCell
### Description

ViewCell's Tapped event is not firing in a ListView.

This is similar to (possibly a duplicate of) #2318.

### Steps to Reproduce

1. In Control Gallery, navigate to Issue 935. 
2. Tap on "I have been selected:"

### Expected Behavior

Text in the cell changes to "I have been selected:1".

### Actual Behavior

Text does not change.

### Basic Information

- Version with issue: 3.5.0

paymicro:
### Description of Change ###

- Fixed `EntryCell.Tapped` fire when EntryCell within a TableView.
- Fixed `Tapped` event for cell in ListView

### Issues Resolved ###

- fixes #2318
- fixes #4901 

### API Changes ###

/

### Platforms Affected ###

- UWP

### Testing Procedure ###

**Test 1**
* Navigate to the UI test for 32040 in the Control Gallery.
* Tap the "Click Here" Label in the EntryCell.

**Test 2**
* In Control Gallery, navigate to Issue 935.
* Tap on "I have been selected:"

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.UAP/ListViewRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/ListViewRenderer.cs
@@ -722,9 +722,9 @@ namespace Xamarin.Forms.Platform.UWP
 			List.SelectedIndex = index;
 		}
 
-		void OnListItemClicked(int index)
+		void OnListItemClicked(int index, Cell cell = null)
 		{
-			Element.NotifyRowTapped(index, cell: null);
+			Element.NotifyRowTapped(index, cell);
 			_itemWasClicked = true;
 		}
 
@@ -735,7 +735,7 @@ namespace Xamarin.Forms.Platform.UWP
 				var templatedItems = TemplatedItemsView.TemplatedItems;
 				var selectedItemIndex = templatedItems.GetGlobalIndexOfItem(e.ClickedItem);
 
-				OnListItemClicked(selectedItemIndex);
+				OnListItemClicked(selectedItemIndex, e.ClickedItem as Cell);
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


hartez:
Since we have full control over the code involved here, I don't think we need to use reflection to fix this. But leaving that aside, this change doesn't really match the behavior of the other platforms.

On Android and iOS, tapping on the Entry itself does not raise the Tapped event; tapping on the "Click Here" label does raise the event.

On UWP with this change, tapping on the "Click Here" label raises the event, but so does tapping on the Entry.

I think the problem with the original UWP code is that the DataTemplate for EntryCell defines the header label (the "Click Here" part) as a part of the EntryCellTextBox template. So tapping the header label is just as if you were tapping on the text box itself (you can see this if you use a mouse - the cursor changes to an I-beam pointer if you hover over the header label). My guess is that the simplest way to fix this would be to make the header label just ignore taps/clicks - they would go right through, and activate the cell itself (firing the event).

hartez:
It seems like this could be fixed by changing the `EntryCellTextBox`'s `HeaderTemplate` (in `Resources.xaml`) from

```
<uwp:EntryCellTextBox.HeaderTemplate>
	<DataTemplate>
		<TextBlock Text="{Binding Label}" Style="{ThemeResource BaseTextBlockStyle}" Foreground="{Binding LabelColor, Converter={StaticResource ColorConverter}, ConverterParameter=DefaultTextForegroundThemeBrush}" />
	</DataTemplate>
</uwp:EntryCellTextBox.HeaderTemplate>
```

to

```
<uwp:EntryCellTextBox.HeaderTemplate>
	<DataTemplate>
		<TextBlock Text="{Binding Label}" IsHitTestVisible="False" Style="{ThemeResource BaseTextBlockStyle}" Foreground="{Binding LabelColor, Converter={StaticResource ColorConverter}, ConverterParameter=DefaultTextForegroundThemeBrush}" />
	</DataTemplate>
</uwp:EntryCellTextBox.HeaderTemplate>
```

(the change is the `IsHitTestVisible` property)

Does that solve the problem, or am I missing something?

samhouts:
@paymicro Feel free to remove the do-not-merge label when you have updated the PR. Thanks!

PR #2335, opened 2018-04-06
fixed WebView typo, replaced Label icons, added FlexLayout

davidortinau:
### Description of Change ###
Added FlexLayout xml node, snippet, and icons.

### Bugs Fixed ###
Fixed wrong Label icon images
Fixed WebView typo

### API Changes ###
None

### Behavioral Changes ###

None

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Consolidate commits as makes sense

PR #3420, opened 2018-07-27
[UWP] Added Windows Platform Specific to avoid set the default image directory

### LINKED ISSUE #2235: Improved image organization

### LINKED ISSUE #3228: Feature Request - Default Search Directory for UWP Icons (Platform Specific)

jsuarezruiz:
### Description of Change ###

Added **Windows Platform Specific** to avoid set the default image directory.

Captura de pantalla 2019-10-10 a las 11 51 36

### Issues Resolved ###

- fixes #3228
- fixes #2235

### API Changes ###

Added:
 - Windows Application Platform Specific in Xamarin.Forms Core project.

Changed:
 - Changes in UWP ImageRenderer.

`Application.Current.On<Windows>().SetImageSearchDirectory("Assets");`

### Platforms Affected ###
- Core
- UWP

### PR Checklist ###

- [ ] Has automated tests 


Xamarin.Forms.Platform.UAP/ImageRenderer.cs
```diff
@@ -189,5 +195,39 @@ void UpdateAspect()
...
+				return;
+			}
+
+			var imageSearchDirectory = Application.Current.OnThisPlatform().GetImageSearchDirectory();
+
+			if(!string.IsNullOrEmpty(imageSearchDirectory))
+			{
+				var source = Element.Source;
+
+				if (source is FileImageSource)
```
jamesmontemagno: Can we use some C# features here:  if (source is FileImageSource fileSource && fileSource != null)

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3228.xaml.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using Xamarin.Forms.PlatformConfiguration;
+using Xamarin.Forms.PlatformConfiguration.WindowsSpecific;
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
```
samhouts: insert `#if APP` here

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3228.xaml.cs
```diff
@@ -0,0 +1,24 @@
...
+
+			Application.Current.On<Windows>().SetImageSearchDirectory("Assets");
+		}
+
+		protected override void Init()
+		{
+
+		}
+	}
+}
```
samhouts: insert `#endif` here

Xamarin.Forms.Platform.UAP/ImageRenderer.cs
```diff
@@ -189,5 +195,39 @@ void UpdateAspect()
...
+
+					if (fileSource != null)
+					{
+						var filePath = fileSource.File;
+
+						var directory = Path.GetDirectoryName(filePath);
+
+						if (string.IsNullOrEmpty(directory) || !Path.GetFullPath(directory).Equals(Path.GetFullPath(imageSearchDirectory)))
+						{
+							filePath = string.Format("{0}{1}{2}", imageSearchDirectory, Path.DirectorySeparatorChar, filePath);
```
samhouts: Why not `Path.Combine`?

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3228.xaml.cs
jsuarezruiz: Added.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3228.xaml.cs
jsuarezruiz: Added.

Xamarin.Forms.Platform.UAP/ImageRenderer.cs
jsuarezruiz: There is no reason and in fact, it is better. Changed. Thanks!.

Xamarin.Forms.Platform.UAP/ImageRenderer.cs
jsuarezruiz: Sure,. changed.

jsuarezruiz:
Ops, I did not see this [PR ](https://github.com/xamarin/Xamarin.Forms/pull/3263)and it's interesting. It's a good question. Maybe are compatible and both can be added. @davidortinau, what do you think?

davidortinau:
Do we need to go with a platform specific? Could this be a config setting done at or near Xamarin.Forms.Init()?

@jassmith @samhouts

jamesmontemagno:
That was my other idea too @davidortinau looking at it. I only ever need to set it once really IMHO.

samhouts:
For @jamesmontemagno's request, a config setting is perfect. We can probably support @MartinZikmund's request for subfolders (#2335) in a different way altogether, closer to @paymicro's solution in #3263 but with a core property instead of a platform specific.

VladislavAntonyuk:
Will it solve the same problem for wpf?

dotMorten:
Why doesn't XF have something like WPF's pack URIs that works cross-plat and looks up the right way. Similarly we need a xplat build action that does the right thing per platform

SamuelDebruyn:
This line in the ButtonRenderer also has to be modified: https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.UAP/ButtonRenderer.cs#L169

Otherwise the ButtonRenderer will still look in the root for images set as `Image` on a `Button`.

Xamarin.Forms.Platform.UAP/ImageRenderer.cs
```diff
@@ -189,5 +195,34 @@ protected async Task UpdateSource()
...
+			{
+				var source = Element.Source;
+
+				if (source is FileImageSource fileSource && fileSource != null)
+				{
+					var filePath = fileSource.File;
+
+					var directory = Path.GetDirectoryName(filePath);
+
+					if (string.IsNullOrEmpty(directory) || !Path.GetFullPath(directory).Equals(Path.GetFullPath(imageSearchDirectory)))
```
kingces95: So if the file appears to live at the search path then we don't combine the path with the search path. Means we can't do `assets\assets` right? Is there a reason to be backward compatible after someone has set the platform specific? Why not just drop this `if` and always combine? Certainly would be easier to document: 
> We prepend the search path to all asset paths.

MartinZikmund:
Is this still considered? It would be great if it were part of Xamarin.Forms soon :-) .

jsuarezruiz:
This PR has the same objective as: https://github.com/xamarin/Xamarin.Forms/pull/3263
Let's close one and focus on the other.

jsuarezruiz:
Updated branch (resolved conflicts, rebase master, etc).

samhouts:
@jsuarezruiz Please make sure this covered the same scenarios as #3263. Thanks!

hartez:
I'm waaayyyy late to the party here, but the name of the property seems a little strange. It's not really a "search", it's just the location of the images. Something like `SetImageLocation` or `SetImageFolder` or `SetImageDirectory` seems more appropriate.

jsuarezruiz:
@hartez Mnn, you're right that the name could perhaps confuse. I have renamed `ImageSearchDirectory` to `ImageDirectory`. Although, if there is another option that is more appropriate, we can still change it.

jamesmontemagno:
WHOA!!!!!!! This is very exciting!!!!

PR #3433, opened 2018-07-30
[iOS] Fix issue where entry would emit a change notification without a user triggered change

### LINKED ISSUE #3385: [iOS] Entry's TextChanged event is fired on Unfocus even when no text changed
### Description
The Entry's TextChanged event is fired when you unfocus the entry control, with OldTextValue = (null) and NewTextValue = ""

### Steps to Reproduce

1. Place an Entry in a page, register the TextChanged event handler, run the iOS App.
2. Tap on the entry to focus it, the Keyboard will show
3. Tap outside the entry to unfocus it.

### Expected Behavior
No "TextChanged" event should be fired
### Actual Behavior
"TextChanged" event is fired

### Basic Information

- Version with issue: 3.1.0.637273
- IDE: Visual Studio
- Platform Target Frameworks: 
  - iOS:  Tested on iOS 11
  - Android: Not tested 
  - UWP:  Not Test

jassmith:
Basically we were changing from null to ""

### Description of Change ###

iOS no longer updates Forms just to set empty string from a null.

### Issues Resolved ###

- fixes #3385

### API Changes ###

None

### Platforms Affected ###

iOS

None

### PR Checklist ###

- [x] Has automated tests 


Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
```diff
@@ -176,9 +176,14 @@ void OnEditingChanged(object sender, EventArgs eventArgs)
 		void OnEditingEnded(object sender, EventArgs e)
 		{
 			// Typing aid changes don't always raise EditingChanged event
-			if (Control.Text != Element.Text)
+
+			// Normalizing nulls to string.Empty allows us to ensure that a change from null to "" doesn't result in a change event.
+			// While technically this is a difference it serves no functional good.
+			var controlText = Control.Text ?? string.Empty;
```
VladislavAntonyuk: For the other hand, there may be some projects where it is necessary to notify when value changes from null to string empty

jassmith: Thats why I only did this one way from renderer to source. It will notify the renderer when you change it yourself.

samhouts:
build --uitests

PR #3437, opened 2018-07-30
[CSS] allow mapping css prop to multiple BPs

StephaneDelcroix:
### Description of Change ###

To allow mapping a single CSS property to multiple controls, we so far
relied on extracting the BindableProperties to static helper classes,
and implementing an interface for accessors, and eventually change
handlers.

That extraction works well, and is actually a good pattern, as long as:
 - the extracted BPs share the same PropertyName,
 - default value, and
 - return type, obiously

As the CSS `color` property has to map to both `TextColor` and `Color`
BPs, the property extraction wasn't possible.

This change adds the capability to map a single CSS property to
multiples BPS, when extraction isn't possible. Whenever the extraction
to an interface is possible, that behavior is strongly encouraged.

### Issues Resolved ###

/

### API Changes ###

/

### Platforms Affected ###

- Core/XAML (all platforms)

/

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/StyleSheets/IStylableTest.cs
+++ b/Xamarin.Forms.Core.UnitTests/StyleSheets/IStylableTest.cs
@@ -41,6 +41,14 @@ namespace Xamarin.Forms.StyleSheets.UnitTests
 		}
 
 		[TestCase]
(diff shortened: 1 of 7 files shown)


kingces95:
That'll do it, alright.

PR #3482, opened 2018-08-03
[Android,iOS,UWP,WPF] MediaElement control

### LINKED ISSUE #1692: [F100] MediaPlayer View
## Rationale

Forms does not currently have a simple cross-platform way to play audio and video without creating a custom renderer.

## Implementation

This will probably require a `MediaSource` class similar to the `ImageSource` class, with the appropriate subclasses (`FileMediaSource`, `StreamMediaSource`, `UriMediaSource` and their respective type converters).

The actual `MediaPlayerView` should have a bindable property `Source` of `MediaSource`. The renderers for each platform should use native media players. 

The `MediaPlayerView` should have a boolean property `ControlsVisible`; when set to `false`, the transport/playback controls of the native player should be hidden.

The view should have a property `PlaybackState` with possible enum values `Unknown`, `Paused`, and `Playing`.

The view should expose events for 
`PlayRequested`
`PauseRequested`
`PlaybackStateChanged`

Renderers should respond to `PlayRequested` by attempting to play the source media in the native player. They should respond to `PauseRequested` by attempting to pause the media. Changes in native playback state should raise `PlaybackStateChanged` in the `MediaPlayerView`. 

The `MediaPlayerView` should include `SendPlay()` and `SendPause()` methods for raising the `PlayRequested` and `PauseRequested` events.

The initial version of this control will not support background playback for audio. 

## Backward Compatibility

This is a brand new View, so there should not be any backward compatibility issues.

## Difficulty: Very

In addition to playing audio/video, this control also needs to gracefully support incoming call interruption, system-wide audio requests, volume controls (software and hardware), and keeping the screen active during playback. All of these interactions are different on each platform.

peterfoot:
### Description of Change ###

Adds a simple MediaElement control to Xamarin Forms. The control can play online and local content (app package via ms-appx uris and app data folder using ms-appdata) along with file uris. There are full renderers for Android, iOS and UWP and a limited renderer for WPF.
A page has been added to the ControlGallery project to demo the control but there are no automated tests currently.

### Issues Resolved ###

- fixes #1692 

### API Changes ###

Added:
* MediaElement class
  * Aspect property
  * AutoPlay property
  * BufferingProgress property
  * CanSeek property
  * CurrentState property
  * Duration property
  * IsLooping property
  * KeepScreenOn property
  * Position property
  * ShowsPlaybackControls property
  * Source property
  * VideoHeight property
  * VideoWidth property
  * Volume property

  * Play method
  * Pause method
  * Stop method

  * MediaEnded event
  * MediaFailed event
  * MediaOpened event
  * SeekCompleted event

* MediaElementState enum

* MediaSource class
  * FromFile method
  * FromUri method

* FileMediaSource class
  * File property

* UriMediaSource class
  * Uri property

The API is largely unchanged from the original InTheHand.Forms.MediaElement except Stretch is replaced with the existing Aspect enum used by the Image control.

### Platforms Affected ###

- Core/XAML (all platforms)
- iOS
- Android*
- UWP
- WPF**

*Android will not render video stretched to fill (Aspect.Fill) and so will default to the same behaviour as Aspect.AspectFit adding letter-boxing if necessary.
**WPF renderer doesn't currently support all functionality - ShowsPlaybackControls has not equivalent functionality in underlying control. Currently Position change events don't function.

N/A This is a new control

### PR Checklist ###

- [ ] Has automated tests 


Xamarin.Forms.Platform.Android/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,381 @@
...
+        {
+            get
+            {
+                try
+                {
+                    return TimeSpan.FromMilliseconds(_view.CurrentPosition);
+                }
+                catch
+                {
+                    return TimeSpan.Zero;
```
charlesroddie: Is there a reason to for using `TimeSpan.Zero` to represent error or inapplicable cases, and elsewhere in this PR? Wouldn't a `Nullable<TimeSpan>` be better?
For Position, a `TimeSpan.Zero` could indicate an error (e.g. no loaded video), being at the start of a video.
For NaturalDuration, a `TimeSpan.Zero` could indicate an error, or being in a streaming video with unknown length.

peterfoot: It roughly follows the UWP control although that uses a Duration for the NaturalDuration property - which supports an Auto value when it is unknown. I'm updating to make this one nullable with a null value having the same purpose as Duration.Auto. I'll also add a CanSeek property which will return false if no video is loaded.

jfversluis:
Just wanted to say; awesome work!

Xamarin.Forms.Platform.Android/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,381 @@
...
+        void UpdateLayoutParameters()
+        {
+            if (Element.NaturalVideoHeight == 0)
+            {
+                _view.LayoutParameters = new FrameLayout.LayoutParams(Width, Height, GravityFlags.Fill);
+                return;
+            }
+
+            float ratio = (float)Element.NaturalVideoWidth / Element.NaturalVideoHeight;
+            float controlRatio = (float)Width / Height;
```
paymicro: Is it possible that the `Element.NaturalVideoHeight` or `Height` will be zero?

Xamarin.Forms.Platform.UAP/MediaElementRenderer.cs
```diff
@@ -0,0 +1,229 @@
...
+				case Windows.UI.Xaml.Media.MediaElementState.Stopped:
+				case Windows.UI.Xaml.Media.MediaElementState.Closed:
+					if (Element.KeepScreenOn)
+					{
+						_request.RequestRelease();
+					}
+					break;
+			}
+
+			if (Element != null)
```
paymicro: This does not make sense, because the above will already be NRE, in the condition `Element.KeepScreenOn`.
It is better to add the first line of the method:
```C#
if (Element == null)
return;
```

Xamarin.Forms.Controls/GalleryPages/MediaElementDemoPage.cs
```diff
@@ -0,0 +1,79 @@
...
+		{
+			element.Stop();
+			System.Diagnostics.Debug.WriteLine(element.CanSeek);
+		}
+
+		protected override void OnAppearing()
+		{
+			base.OnAppearing();
+
+			element.Source = new Uri("http://sec.ch9.ms/ch9/5d93/a1eab4bf-3288-4faf-81c4-294402a85d93/XamarinShow_mid.mp4");
```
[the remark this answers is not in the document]
peterfoot: Not sure, this is a recent Channel 9 Xamarin Show episode. Happy to replace with another public video uri.

RadwanIbrahim:
wonderful work thank you very much, I have one question can this play audio too

peterfoot:
@RadwanIbrahim yes you can play any audio format supported on the underlying platform e.g. mp3 etc

Xamarin.Forms.ControlGallery.WindowsUniversal/Xamarin.Forms.ControlGallery.WindowsUniversal.csproj
```diff
@@ -129,7 +129,7 @@
     <Compile Include="DisposePageRenderer.cs" />
     <Compile Include="RegistrarValidationService.cs" />
     <Compile Include="SampleNativeControl.cs" />
-    <Compile Include="_2489CustomRenderer.cs" />
+    <None Include="_2489CustomRenderer.cs" />
```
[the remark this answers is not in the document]
peterfoot: Sorry, I made this change locally for the controlgallery app to run correctly but it shouldn't have got committed to GitHub. Will reverse the change.

Xamarin.Forms.Core/MediaElement.cs
```diff
@@ -0,0 +1,267 @@
...
+		static bool ValidatePosition(BindableObject bindable, object value)
+		{
+			MediaElement element = bindable as MediaElement;
+			if (element != null)
+			{
+				if (element._renderer != null)
+				{
+					element._renderer.Seek((TimeSpan)value);
+				}
+			}
```
paymicro: I think this can be written in one line `element?._renderer?.Seek((TimeSpan)value);`

Xamarin.Forms.Core/MediaElement.cs
```diff
@@ -0,0 +1,267 @@
...
+
+			return true;
```
paymicro: Validation result always correct? Perhaps this code can be better called when the property changes, rather than validation.

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,422 @@
...
+
+			base.OnElementPropertyChanged(sender, e);
+		}
+
+		void IMediaElementRenderer.Seek(TimeSpan time)
+		{
+			if (_avPlayerViewController.Player.Status == AVPlayerStatus.ReadyToPlay)
+			{
+				if (_avPlayerViewController.Player.CurrentItem != null)
+				{
```
paymicro: I think it would be better to combine these two conditions into one block and invert it.
```C#
if (_avPlayerViewController.Player.Status != AVPlayerStatus.ReadyToPlay || _avPlayerViewController.Player.CurrentItem == null)
    return;
```

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,422 @@
...
+					NSValue[] ranges = _avPlayerViewController.Player.CurrentItem.SeekableTimeRanges;
+					CMTime seekTo = new CMTime(Convert.ToInt64(time.TotalMilliseconds), 1000);
+					bool canSeek = false;
+					foreach (NSValue v in ranges)
+					{
+						if (seekTo >= v.CMTimeRangeValue.Start && seekTo < (v.CMTimeRangeValue.Start + v.CMTimeRangeValue.Duration))
+						{
+							canSeek = true;
```
paymicro: Is it worth creating `canSeek` variable? It may be easier to immediately call `_avPlayerViewController.Player.Seek` in this line?

Xamarin.Forms.Core/MediaElement.cs
```diff
@@ -0,0 +1,267 @@
...
+			set { SetValue(IsLoopingProperty, value); }
+		}
+		
+		public bool KeepScreenOn
+		{
+			get { return (bool)GetValue(KeepScreenOnProperty); }
+			set { SetValue(KeepScreenOnProperty, value); }
+		}
+
+		public TimeSpan? NaturalDuration
```
kingces95: Could we drop the `Natrual` here and else where? Seems the default case. Later, if there's an abnormal case, then we can append an adjective to that property.

Xamarin.Forms.Core/MediaElement.cs
```diff
@@ -0,0 +1,267 @@
...
+		}
+		
+		[TypeConverter(typeof(UriTypeConverter))]
+		public Uri Source
+		{
+			get { return (Uri)GetValue(SourceProperty); }
+			set { SetValue(SourceProperty, value); }
+		}
+
+        public IDictionary<string, string> HttpHeaders { get; } = new Dictionary<string, string>();
```
kingces95: Can you elaborate on when HttpHeaders this is needed?

kingces95:
Thanks so much for your contribution! Wonderful work! I'm unable to test all the functionality implemented which makes it difficult to sign off on the PR. 

Can we split out the functionality which is able to be tested via the gallery test to a separate PR? Essentially, load via https, `Play`, `Pause`, and `Stop` plus `CurrentState`. Is that a minimum viable feature set?

If not then we'll need to expand the gallery test to include UI to play with those features that encompass a minimum viable set. I've tried to gather the feature set we'd need to test below. 

The test will need UI to set any mutable bindable properties:
  - `AreTransportControlsEnabled`
  - `Position`
  - `AutoPlay`
  - `IsLooping`
  - `KeepScreenOn`
  - `Source`
  - `Aspect`; We'll also need buttons to change the aspect ratio of the screen so that it's greater or less than the ratio of the control. Seems will need this for any minimum viable feature set. 

All properties, not just `CurrentState`, should be displayed on the UI so a tester may verify their values. Beyond the mutable properties these should be displayed:
  - `BufferingProgress`
  - `CanSeek`
  - `NaturalDuration`
  - `NaturalVideoHeight`

The test will need a console to log events as they are raised. Right now we can see that the `CurrentState` changes but we cannot see how it interacts with other events (or even verify they've been raised). We should interleave the events so we may assert they are raised in the proper order. This is especially important for x-plat as it's often the case that different plats will raise events in different order unless care is take to ensure they're the same. The log can simply be a label that appends a new line with a little message indicating that an event has been raised. The following events will need to be logged:
  - `CurrentStateChanged`
  - `MediaEnded`
  - `MediaFailed`
  - `MediaOpened`
  - `SeekCompleted`

We'll need to test each video source scheme on each platform. We'll need examples of each as well as a button to load each example. For example on android we have logic for loading from:
  - `AbsolutePath`
  - `"ms-appx"`, e.g. embedded resource
  - `"ms-appdata"`, e.g. known folders: local, temp, app relative
  - HTTP/HTTPS, with and without headers

Xamarin.Forms.Core/MediaElement.cs
```diff
@@ -0,0 +1,267 @@
...
+using System.Collections.Generic;
+using System.ComponentModel;
+using Xamarin.Forms.Platform;
+
+namespace Xamarin.Forms
+{
+	[RenderWith(typeof(_MediaElementRenderer))]
+	public sealed class MediaElement : View
+	{
+		public static readonly BindableProperty AreTransportControlsEnabledProperty =
```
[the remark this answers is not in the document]
rmarinho: Can we drop the Are ?

Xamarin.Forms.Core/MediaElement.cs
rmarinho: This doesn't follow our normal approach , `Core` should not call a renderer. We need a new approach for this, use a event like `SeekChanged` or something like that that the renderer will subscribe. 

But why don0t we just check PropertyChanged on the renderer for this what validation is done here that can't be done on the renderer ?

Xamarin.Forms.Core/MediaElement.cs
```diff
@@ -0,0 +1,267 @@
...
+			}
+
+			return true;
+		}
+
+		
+		IMediaElementRenderer _renderer = null;
+
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public void SetRenderer(IMediaElementRenderer renderer)
```
rmarinho: We don't use this , renderer is normally an `Attached Property` on the element

Xamarin.Forms.Core/MediaElement.cs
```diff
@@ -0,0 +1,267 @@
...
+		{
+			get { return (bool)GetValue(KeepScreenOnProperty); }
+			set { SetValue(KeepScreenOnProperty, value); }
+		}
+
+		public TimeSpan? NaturalDuration
+		{
+			get
+			{
+				if (_renderer != null)
```
rmarinho: Core should not know about the renderer

Xamarin.Forms.Core/MediaElement.cs
```diff
@@ -0,0 +1,267 @@
...
+				{
+					return _renderer.NaturalDuration;
+				}
+
+				return null;
+			}
+		}
+
+		public int NaturalVideoHeight
```
rmarinho: Let's drop `Natural`

Xamarin.Forms.Core/MediaElement.cs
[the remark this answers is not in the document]
peterfoot: The logic behind this was to following the UWP control. On iOS there is the simpler ShowPlaybackControls and on Android you enable a linked MediaController object. If we want to deviate from UWP ShowPlaybackControls is probably the most logical name.

Xamarin.Forms.Platform.UAP/MediaElementRenderer.cs
```diff
@@ -0,0 +1,236 @@
...
+						}
+					}
+					else
+					{
+						_request.RequestRelease();
+					}
+					break;
+
+				case nameof(MediaElement.Source):
+					Control.Source = Element.Source;
```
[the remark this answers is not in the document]
peterfoot: The control uses uri schemes to load files from local storage or the app package. The mappings are discussed here - https://peterfoot.net/2018/08/09/xamarin-forms-mediaelement/

Xamarin.Forms.Core/MediaElement.cs
```diff
@@ -0,0 +1,267 @@
...
+			}
+		}
+
+		public Aspect Aspect
+		{
+			get { return (Aspect)GetValue(AspectProperty); }
+			set { SetValue(AspectProperty, value); }
+		}
+
+		public event EventHandler CurrentStateChanged;
```
rmarinho: Isn't CUrrrentState a BP if so we don0t need this.

rmarinho:
I tested this PR and it does work ok on iOS and UWP, but i coulnd't get it to play on WPF and Android , but we need to have a test page more complete like @kingces95 said. 
I also think we should add a example for MP3 and no UI (like 0 height) to had the ability to play sound only.

WPF throws a NRE 
Android i see on logs - 
``` 
Couldn't open https://sec.ch9.ms/ch9/5d93/a1eab4bf-3288-4faf-81c4-294402a85d93/XamarinShow_mid.mp4: java.io.FileNotFoundException: No content provider: https://sec.ch9.ms/ch9/5d93/a1eab4bf-3288-4faf-81c4-294402a85d93/XamarinShow_mid.mp4
09-0
```

Xamarin.Forms.Core/MediaElement.cs
```diff
@@ -0,0 +1,267 @@
...
+		  BindableProperty.Create(nameof(BufferingProgress), typeof(double), typeof(MediaElement), 0.0);
+
+		public static readonly BindableProperty IsLoopingProperty =
+		  BindableProperty.Create(nameof(IsLooping), typeof(bool), typeof(MediaElement), false);
+
+		public static readonly BindableProperty KeepScreenOnProperty =
+		  BindableProperty.Create(nameof(KeepScreenOn), typeof(bool), typeof(MediaElement), false);
+
+		public static readonly BindableProperty SourceProperty =
+		  BindableProperty.Create(nameof(Source), typeof(Uri), typeof(MediaElement));
```
rmarinho: This will probably require a MediaSource class similar to the ImageSource class, with the appropriate subclasses (FileMediaSource, StreamMediaSource, UriMediaSource and their respective type converters).

Xamarin.Forms.Platform.Android/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,380 @@
...
+				return;
+			}
+
+			float ratio = (float)Element.NaturalVideoWidth / Element.NaturalVideoHeight;
+			float controlRatio = (float)Width / Height;
+
+			switch (Element.Aspect)
+			{
+				case Aspect.Fill:
+					// TODO: this doesn't stretch like other platforms...
```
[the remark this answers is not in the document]
peterfoot: Have added this text to the PR

Xamarin.Forms.Platform.UAP/MediaElementRenderer.cs
```diff
@@ -0,0 +1,236 @@
...
+					}
+					break;
+			}
+
+			Element.SendCurrentState((MediaElementState)((int)Control.CurrentState));
+		}
+		
+		void BufferingProgressChanged(DependencyObject sender, DependencyProperty dp)
+		{
+			((IElementController)Element).SetValueFromRenderer(MediaElement.BufferingProgressProperty, Control.BufferingProgress);
```
paymicro: throws `NullReferenceException` when click back button in navigation page.
```C#
Xamarin.Forms.Platform.UWP.VisualElementRenderer<TElement, TNativeElement>.Control.get returned null.
```

Xamarin.Forms.Platform.Android/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,380 @@
...
+					break;
+
+				case nameof(MediaElement.CurrentState):
+					switch (Element.CurrentState)
+					{
+						case MediaElementState.Playing:
+							if (!_view.IsPlaying)
+							{
+								_view.Start();
+							}
```
paymicro: After hitting the gallery buttons `Play` > `Stop` > `Play`, the video freezes and cannot be played again.
But If you press `Play` > `Pause` > `Stop` > `Play`, the video continues to play.

Xamarin.Forms.Core/MediaElement.cs
paymicro: I agree, `Core` should not know about the renderer.

Xamarin.Forms.Platform.UAP/MediaElementRenderer.cs
```diff
@@ -0,0 +1,236 @@
...
+		}
+
+		void Control_MediaFailed(object sender, ExceptionRoutedEventArgs e)
+		{
+			Element?.OnMediaFailed();
+		}
+
+		void Control_MediaEnded(object sender, RoutedEventArgs e)
+		{
+			Element?.OnMediaEnded();
```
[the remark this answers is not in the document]
peterfoot: On Windows you can play from a stopped state at the end of media and it will automatically return to the beginning. The Android code was a trick to enable the same however I've noticed that when you return the video to the beginning it starts buffering the media again which may not be desired so I'm going to change the Android version to add a check when playing to see if the position equals the media duration and if so seek to the beginning and then start playing which with give the same behaviour on all without any preloading and will leave the position at the end of the media as expected.

Xamarin.Forms.Core/MediaElement.cs
peterfoot: I'm completely rewriting this to remove the IMediaElementRenderer interface and all the renderer references from the MediaElement. I've implemented instead an IMediaElementController on the MediaElement which the renderer can use to set things in the control without exposing them to the developer just consuming the MediaElement. I've used the WebView as a guide for the architecture.

Xamarin.Forms.Controls/GalleryPages/MediaElementDemoPage.cs
```diff
@@ -49,11 +73,28 @@ public MediaElementDemoPage()
...
+		}
+
+		private void Element_MediaFailed(object sender, EventArgs e)
+		{
+			consoleLabel.Text += "Media failed\r\n";
+		}
+
+		private void Element_MediaEnded(object sender, EventArgs e)
+		{
+			consoleLabel.Text += "Media ended\r\n";
```
paymicro: `\r\n` >> `Environment.NewLine`

Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
```diff
@@ -172,12 +241,197 @@ void OnElementChanged(ElementChangedEventArgs<MediaElement> e)
...
+				if (MediaElement.AutoPlay)
+				{
+					_view.Start();
+					((IMediaElementController)MediaElement).CurrentState = _view.IsPlaying ? MediaElementState.Playing : MediaElementState.Stopped;
+				}
+
+			}
+			else
+			{
+				if(_view.IsPlaying)
```
paymicro: you can be combined `else if (_view.IsPlaying)`

Xamarin.Forms.Platform.UAP/MediaElementRenderer.cs
```diff
@@ -103,18 +72,54 @@ protected override void OnElementChanged(ElementChangedEventArgs<MediaElement> e
 			}
 		}
 
+		private void Element_StateRequested(object sender, StateRequested e)
```
paymicro: We do not use `private` and char `_` in method names

Xamarin.Forms.Platform.UAP/MediaElementRenderer.cs
```diff
@@ -212,7 +231,7 @@ protected override void OnElementPropertyChanged(object sender, System.Component
 					break;
 
 				case nameof(MediaElement.Source):
-					Control.Source = Element.Source;
+					Control.Source = Element.Source;
```
paymicro: Please remove some tab chars at the end of the line.

Xamarin.Forms.Platform.WPF/Renderers/MediaElementRenderer.cs
```diff
@@ -84,22 +47,71 @@ protected override void OnElementChanged(ElementChangedEventArgs<MediaElement> e
...
+					{
+						DisplayRequestRelease();
+					}
+
+					Control.Stop();
+					((IMediaElementController)Element).CurrentState = MediaElementState.Stopped;
+					break;
+			}
+
+			((IMediaElementController)Element).Position = Control.Position;
```
paymicro: `((IMediaElementController)Element)` quite often used. I think it is possible to make private property `IMediaElementController Controller => Element as IMediaElementController` and use it.

Xamarin.Forms.Platform.WPF/Renderers/MediaElementRenderer.cs
```diff
@@ -84,22 +47,71 @@ protected override void OnElementChanged(ElementChangedEventArgs<MediaElement> e
...
+		}
+
+		private void Element_SeekRequested(object sender, SeekRequested e)
```
paymicro: We do not use `private` and char `_` in method names

Xamarin.Forms.Platform.Android/Renderers/FormsVideoView.cs
```diff
@@ -39,7 +39,11 @@ void ExtractMetadata(MediaMetadataRetriever retriever)
 			if (!string.IsNullOrEmpty(durationString))
 			{
 				long durationMS = long.Parse(durationString);
```
paymicro: `durationString` will always be correctly parsed? Perhaps use `TryParse` as above.

Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
```diff
@@ -172,12 +241,197 @@ void OnElementChanged(ElementChangedEventArgs<MediaElement> e)
...
+			if (_mediaPlayer != null)
+			{
+				_mediaPlayer.Dispose();
+				_mediaPlayer = null;
+			}
+		}
+
+		bool MediaPlayer.IOnErrorListener.OnError(MediaPlayer mp, MediaError what, int extra)
+		{
+			throw new NotImplementedException();
```
paymicro: So it should be or forgotten to implement it?

Xamarin.Forms.Platform.UAP/MediaElementRenderer.cs
```diff
@@ -143,16 +148,43 @@ void Control_CurrentStateChanged(object sender, RoutedEventArgs e)
...
+					return MediaElementState.Paused;
+
+				case Windows.UI.Xaml.Media.MediaElementState.Playing:
+					return MediaElementState.Playing;
+
+				case Windows.UI.Xaml.Media.MediaElementState.Stopped:
+					return MediaElementState.Stopped;
+			}
+
+			return 0;
```
paymicro: `0` equals `MediaElementState.Closed`. I think it's more correct to throw an exception.

Xamarin.Forms.ControlGallery.Android/Xamarin.Forms.ControlGallery.Android.csproj
```diff
@@ -197,7 +197,7 @@
     <Compile Include="StringProvider.cs" />
     <Compile Include="TestCloudService.cs" />
     <Compile Include="_1909CustomRenderer.cs" />
-    <Compile Include="_2489CustomRenderer.cs" />
+    <None Include="_2489CustomRenderer.cs" />
```
rmarinho: No need for changes in this file I think

Xamarin.Forms.ControlGallery.Android/Xamarin.Forms.ControlGallery.Android.csproj
```diff
@@ -42,7 +42,7 @@
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
     <OutputPath>bin\Debug\</OutputPath>
-    <DefineConstants>TRACE;DEBUG</DefineConstants>
+    <DefineConstants>TRACE;DEBUG;TEST_EXPERIMENTAL_RENDERERS</DefineConstants>
```
rmarinho: No need to add the Flag, this renderer will be FastRender as from the start so the flag isn't needed.

Xamarin.Forms.Core/MediaElement.cs
```diff
@@ -0,0 +1,232 @@
...
+			get { return (int)GetValue(VideoHeightProperty); }
+		}
+
+		public int VideoWidth
+		{
+			get { return (int)GetValue(VideoWidthProperty); }
+		}
+		
+		[TypeConverter(typeof(UriTypeConverter))]
+		public Uri Source
```
rmarinho: We need a new type of source for media.

Xamarin.Forms.Platform.Android/Renderers/FormsVideoView.cs
```diff
@@ -0,0 +1,103 @@
...
+
+			if (System.IO.File.Exists(path))
+			{
+				MediaMetadataRetriever retriever = new MediaMetadataRetriever();
+				try
+				{
+					retriever.SetDataSource(path);
+					ExtractMetadata(retriever);
+				}
+				catch { }
```
rmarinho: We can't do this and just swallow the error.

Xamarin.Forms.Platform.Android/Renderers/FormsVideoView.cs
```diff
@@ -0,0 +1,103 @@
...
+					retriever.SetDataSource(uri.ToString(), headers);
+				}
+				else
+				{
+					retriever.SetDataSource(Context, uri);
+				}
+
+				ExtractMetadata(retriever);
+			}
+			catch { }
```
rmarinho: same thing here we can't just try catch

peterfoot:
@samhouts Thanks, I've been busy recently and haven't had time to address all the points. I'd particularly welcome some help around the tests because I still haven't got my head around the test app.

Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
```diff
@@ -238,6 +239,13 @@ void OnElementChanged(ElementChangedEventArgs<MediaElement> e)
 			ElementChanged?.Invoke(this, new VisualElementChangedEventArgs(e.OldElement, e.NewElement));
 		}
 
+		void MetadataRetrieved(object sender, EventArgs e)
+		{
+			((IMediaElementController)MediaElement).Duration = _view.DurationTimeSpan;
+			((IMediaElementController)MediaElement).VideoHeight = _view.VideoHeight;
+			((IMediaElementController)MediaElement).VideoWidth = _view.VideoWidth;
```
paymicro: I think that frequent unboxing of `MediaElement` should be replaced to private variable, which will do unboxing once when you `SetElement`.

[boxing and unboxing are computationally expensive process](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/types/boxing-and-unboxing#performance)

peterfoot: This is casting of reference types so doesn't have the boxing overhead but I agree there should be a cleaner way of writing this

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,508 @@
...
+				if (Element.KeepScreenOn)
+				{
+					SetKeepScreenOn(true);
+				}
+
+				_notificationHandle = NSNotificationCenter.DefaultCenter.AddObserver(AVPlayerItem.DidPlayToEndTimeNotification, PlayedToEnd);
+
+				UpdateSource();
+			}
+		}*/
```
paymicro: could you please remove this comment if it's not needed.

Xamarin.Forms.Platform.WPF/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,280 @@
...
+				Control.Width = Element.Width;
+			}
+		}
+
+		protected override void UpdateHeight()
+		{
+			if (Element.Height > 0)
+			{
+				Control.Height = Element.Height;
+			}
```
paymicro: if `Element.Height` is negative value then `Control.Height` nothing will change?
i think better always changes `Height` but with limits. `Control.Height = Math.Max(0, Element.Height)`
also for Width.

Xamarin.Forms.Platform.WPF/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,280 @@
...
+		private void Element_SeekRequested(object sender, SeekRequested e)
+		{
+			Control.Position = e.Position;
+			Controller.Position = Control.Position;
+		}
+
+		void UpdateSource()
+		{
+			if (Element.Source != null)
+			{
```
paymicro: reverse this condition, please.

Xamarin.Forms.Core/MediaElement.cs
```diff
@@ -0,0 +1,237 @@
...
+		public event EventHandler SeekCompleted;
+
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public void OnMediaEnded()
+		{
+			SetValue(CurrentStateProperty, MediaElementState.Stopped);
+			MediaEnded?.Invoke(this, EventArgs.Empty);
+		}
+
+		internal void RaisePropertyChanged(string propertyName)
```
rmarinho: why do we need this ?

Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,488 @@
...
+
+		public VisualElement Element => MediaElement;
+
+		VisualElementTracker IVisualElementRenderer.Tracker => _tracker;
+
+		ViewGroup IVisualElementRenderer.ViewGroup => null;
+
+		AView IVisualElementRenderer.View => this;
+
+		public event EventHandler<VisualElementChangedEventArgs> ElementChanged;
```
rmarinho: Can we order public stuff then protected, then private.. helps reading this code.

Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,488 @@
...
+		SizeRequest IVisualElementRenderer.GetDesiredSize(int widthConstraint, int heightConstraint)
+		{
+			AView view = this;
+			view.Measure(widthConstraint, heightConstraint);
+
+			return new SizeRequest(new Size(MeasuredWidth, MeasuredHeight), new Size());
+		}
+
+		void Initialize()
+		{
```
rmarinho: Can we drop this?

Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,488 @@
...
+
+		void IViewRenderer.MeasureExactly()
+		{
+			ViewRenderer.MeasureExactly(this, Element, Context);
+		}
+
+		void IVisualElementRenderer.SetElement(VisualElement element)
+		{
+			if (element == null)
+			{
```
rmarinho: we could drop the parenthsis here

Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,488 @@
...
+		}
+
+		void UpdateBackgroundColor()
+		{
+			SetBackgroundColor(Element.BackgroundColor.ToAndroid());
+		}
+
+		void IVisualElementRenderer.UpdateLayout()
+		{
+			var lp = _view.LayoutParameters;
```
rmarinho: This doesn't look right..

Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,488 @@
...
+		}
+
+		void OnElementChanged(ElementChangedEventArgs<MediaElement> e)
+		{
+			if (e.OldElement != null)
+			{
+				ReleaseControl();
+			}
+
+			if (e.NewElement != null && !_isDisposed)
```
rmarinho: we shouldn't check _isDisposed we should check maybe if Control (_view)  is null

Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,488 @@
...
+		}
+
+		public void RegisterEffect(Effect effect)
+		{
+			_effectControlProvider.RegisterEffect(effect);
+		}
+
+		void UpdateSource()
+		{
+			if (MediaElement.Source != null)
```
rmarinho: we need to figure a better way to do this, maybe look at how ImageSource works for this.

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,475 @@
...
+			switch (e.State)
+			{
+				case MediaElementState.Playing:
+					var audioSession = AVAudioSession.SharedInstance();
+					NSError err = audioSession.SetCategory(AVAudioSession.CategoryPlayback);
+					audioSession.SetMode(AVAudioSession.ModeMoviePlayback, out err);
+					err = audioSession.SetActive(true);
+
+					_avPlayerViewController.Player.Play();
+					Controller.CurrentState = MediaElementState.Playing;
```
paymicro: this code repeats in line 130-136. I think that better to extract it in method.

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,475 @@
...
+					_avPlayerViewController.Player = new AVPlayer(item);
+					_rateObserver = (NSObject)_avPlayerViewController.Player.AddObserver("rate", NSKeyValueObservingOptions.New, ObserveRate);
+				}
+				
+				if (MediaElement.AutoPlay)
+				{
+					var audioSession = AVAudioSession.SharedInstance();
+					NSError err = audioSession.SetCategory(AVAudioSession.CategoryPlayback);
+					audioSession.SetMode(AVAudioSession.ModeMoviePlayback, out err);
+					err = audioSession.SetActive(true);
```
paymicro: Possible errors are ignored. Perhaps the best solution would be to add an error event to the `MediaElement` class?

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,475 @@
...
+
+		void RemoveStatusObserver()
+		{
+			if (_statusObserver != null)
+			{
+				try
+				{
+					_avPlayerViewController?.Player?.CurrentItem?.RemoveObserver(_statusObserver, "status");
+				}
+				catch { }
```
paymicro: you can delete this line

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,475 @@
...
+				case 1.0f:
+					Controller.CurrentState = MediaElementState.Playing;
+					break;
+			}
+
+			Controller.Position = Position;
+		}
+		void ObserveStatus(NSObservedChange e)
+		{
+			if (e.NewValue != null)
```
paymicro: Is this check required?

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,475 @@
...
+					return ((IVisualElementRenderer)this).NativeView.GetSizeRequest(widthConstraint, (MediaElement.VideoWidth / MediaElement.VideoHeight) * widthConstraint, 44, 44);
+				}
+				else if (double.IsInfinity(widthConstraint))
+				{
+					return ((IVisualElementRenderer)this).NativeView.GetSizeRequest(heightConstraint / (MediaElement.VideoWidth / MediaElement.VideoHeight), heightConstraint, 44, 44);
+				}
+			}
+
+			return ((IVisualElementRenderer)this).NativeView.GetSizeRequest(widthConstraint, 200, 44, 44);
+		}*/
```
paymicro: Please don't forget to delete this comment.

Xamarin.Forms.Platform.UAP/MediaElementRenderer.cs
```diff
@@ -0,0 +1,296 @@
...
+					if (Element.Width > 0)
+					{
+						Width = Element.Width;
+					}
+					break;
+
+				case nameof(MediaElement.Height):
+					if (Element.Height > 0)
+					{
+						Height = Element.Height;
```
paymicro: I think the limit value is better than ignoring set to zero or negative
`Height = Math.Max(0, Element.Height)`

Xamarin.Forms.Platform.Android/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,321 @@
...
+						{
+							_view.SetVideoPath(uriSource.Uri.AbsolutePath);
+						}
+						else
+						{
+							_view.SetVideoURI(global::Android.Net.Uri.Parse(uriSource.Uri.AbsoluteUri));
+						}
+					}
+				}
+				else
```
paymicro: Could you please use this pattern matching -> `else if (Element.Source is FileMediaSource fileSource)`

Xamarin.Forms.Platform.Android/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,321 @@
...
+		{
+			AutoPackage = false;
+		}
+		
+		protected override void Dispose(bool disposing)
+		{
+			if (Control != null)
+			{
+				Control.RemoveAllViews();
+			}
```
paymicro: -> `Control?.RemoveAllViews();`

Xamarin.Forms.Core/MediaElementState.cs
```diff
@@ -0,0 +1,12 @@
...
+{
+	public enum MediaElementState
+	{
+		Closed = 0,
+		Opening = 1,
+		Buffering = 2,
+		Playing = 3,
+		Paused = 4,
+		Stopped = 5,
+	}
```
paymicro: Is it necessary to assign default numbers in this enum?

rmarinho:
Awesome work, we are getting this ready, good comments from @paymicro.

Xamarin.Forms.Platform.Android/Renderers/FormsVideoView.cs
```diff
@@ -0,0 +1,117 @@
...
+			
+			if(!int.TryParse(retriever.ExtractMetadata(MetadataKey.VideoHeight), out _videoHeight))
+			{
+				_videoHeight = 0;
+			}
+
+			long durationMS;
+			string durationString = retriever.ExtractMetadata(MetadataKey.Duration);
+
+			if (!string.IsNullOrEmpty(durationString) && long.TryParse(durationString, out durationMS))
```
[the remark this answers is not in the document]
peterfoot: Yes, ExtractMetadata can return null here if the duration is unavailable for the media type. In this case we want DurationTimeSpan to be null.

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,475 @@
...
+				}
+			}
+			else
+			{
+				if (_idleTimerDisabled)
+				{
+					_idleTimerDisabled = false;
+					UIApplication.SharedApplication.IdleTimerDisabled = false;
+				}
+			}
```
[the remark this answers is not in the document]
peterfoot: _idleTimerDisabled keeps track of when this control requested IdleTimerDisabled and it was not already set. This means that if this value is false but the system-wide setting is true we won't disable it because it has been set elsewhere in the app.

Xamarin.Forms.Core/MediaElementState.cs
peterfoot: In an older version these were to match with the UWP values. It's no longer used like that so I'll remove them.

Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,465 @@
...
+
+		void OnElementChanged(ElementChangedEventArgs<MediaElement> e)
+		{
+			if (e.OldElement != null)
+			{
+				ReleaseControl();
+			}
+
+			if (e.NewElement != null && !(_view is null))
+			{
```
paymicro: This is the only one block where the `_view` value is set. And to get here you need to have the value already been set...

peterfoot: I've changed this as the view can live for the duration of the renderer. I've moved the code to set it up into the constructor and moved the cleanup code into Dispose. Changing the Element then merely needs to reset the properties based on the new element.

peterfoot:
I can't see what the remaining requested changes are, GitHub shows them all as resolved. Can anyone help?

MhAllan:
Can we have memory source please, I mean to play video from Stream or byte array!

Xamarin.Forms.Core/FileMediaSource.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Xamarin.Forms
+{
+	[TypeConverter(typeof(FileImageSourceConverter))]
```
paymicro:
```suggestion
	[TypeConverter(typeof(FileMediaSourceConverter))]
```
Also need to create class `FileMediaSourceConverter`

Xamarin.Forms.Core/MediaSourceConverter.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Xamarin.Forms
+{
+	[Xaml.TypeConversion(typeof(ImageSource))]
```
paymicro:
```suggestion
	[Xaml.TypeConversion(typeof(MediaSource))]
```

peterfoot:
>           Can we have memory source please, I mean to play video from Stream or byte array!

This is a great idea for an update but the priority for now is to get the basic control integrated into Xamarin Forms. I did look at it a while ago and it's easier on some platforms than others but all offer a fairly consistent way of opening Uris and Files and these cover the most common scenarios.

PaulVrugt:
Will this implementation contain support for custom http headers for uris set as source like the original inthehand.forms component did? This is a requirement for us to migrate to the new component

peterfoot:
>           Will this implementation contain support for custom http headers for uris set as source like the original inthehand.forms component did? This is a requirement for us to migrate to the new component

Eventually yes I hope so! The goal with the initial addition was to get the basic functionality first also there was never a UWP implementation for Headers in the standalone component and that would need addressing.

rmarinho:
Build failing 
```
C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\Roslyn\Microsoft.CSharp.Core.targets(52,5): Error MSB3105: The item "Renderers\WkWebViewRenderer.cs" was specified more than once in the "Sources" parameter.  Duplicate items are not supported by the "Sources" parameter.
```

peterfoot:
> ited with

Should be fixed now, I can't build the full solution because of the recurring XamlG error. Hopefully this unblocks it.

samhouts:
build

peterfoot:
>           build

?

StephaneDelcroix:
@peterfoot that's one of the way we can trigger a build of your PR. We can, you can't.

Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,481 @@
...
+			_controller.Visibility = MediaElement.ShowsPlaybackControls ? ViewStates.Visible : ViewStates.Gone;
+		}
+
+		void UpdateSource()
+		{
+			if (MediaElement.Source != null)
+			{
+				if (MediaElement.Source is UriMediaSource uriSource)
+				{
+					if (uriSource.Uri.Scheme == "ms-appx")
```
mattleibow: I am worried that this may cause confusion. This is very UWP-ish and appx has no meaning outside UWP - and the ms- prefix.

Xamarin.Forms.Platform.Android/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,310 @@
...
+using Android.Widget;
+using System;
+using Xamarin.Forms;
+using Xamarin.Forms.Platform.Android;
+using Android.Graphics;
+using Android.Runtime;
+
+namespace Xamarin.Forms.Platform.Android
+{
+	public sealed class MediaElementRenderer : ViewRenderer<MediaElement, FrameLayout>, MediaPlayer.IOnCompletionListener, MediaPlayer.IOnPreparedListener
```
mattleibow: I am wondering if we need this one at all? What is the benefit of the non-fast renderer? We just add an extra view.

Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
mattleibow: I am worried that this may cause confusion. This is very UWP-ish and appx has no meaning outside UWP - and the ms- prefix.

Xamarin.Forms.Platform.Android/Renderers/MediaElementRenderer.cs
mattleibow: I am wondering if we need this one at all? What is the benefit of the non-fast renderer? We just add an extra view to the UI and more code to maintain. Aren't we preferring the fast renderers all over?

Xamarin.Forms.Platform.WPF/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,299 @@
...
+					Control.Source = new Uri(uriSource.Uri.ToString().Replace("ms-appx://", "pack://application:,,,"));
+				}
+				else if (uriSource.Uri.Scheme == "ms-appdata")
+				{
+					string filePath = string.Empty;
+
+					if (uriSource.Uri.LocalPath.StartsWith("/local"))
+					{
+						// WPF doesn't have the concept of an app package local folder so using My Documents as a placeholder
+						filePath = System.IO.Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments), uriSource.Uri.LocalPath.Substring(7));
```
mattleibow: Maybe this should be in the `ApplicationData` path rather? https://docs.microsoft.com/en-us/dotnet/api/system.environment.specialfolder?view=netframework-4.7.2

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,455 @@
...
+						// used for a file embedded in the application package
+						asset = AVAsset.FromUrl(NSUrl.FromFilename(uriSource.Uri.LocalPath.Substring(1)));
+					}
+					else if (uriSource.Uri.Scheme == "ms-appdata")
+					{
+						string filePath = string.Empty;
+
+						if (uriSource.Uri.LocalPath.StartsWith("/local"))
+						{
+							filePath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments), uriSource.Uri.LocalPath.Substring(7));
```
mattleibow: Maybe this should be in the `ApplicationData` path rather? https://docs.microsoft.com/en-us/dotnet/api/system.environment.specialfolder?view=netframework-4.7.2

Xamarin.Forms.Platform.WPF/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,299 @@
...
+					else
+					{
+						throw new ArgumentException("Invalid Uri", "Source");
+					}
+
+					Control.Source = new Uri(filePath);
+				}
+				else if (uriSource.Uri.Scheme == "https")
+				{
+					throw new ArgumentException("HTTPS Not supported on WPF", "Source");
```
mattleibow: Maybe we should mention that HTTP is so that people don't think ALL web sources are bad.

scxp:
I may have missed this while looking through the code, but is there a way to change the volume programmatically?

scxp:
@davidortinau I’d really like to see Xamarin support ExoPlayer on Android. MediaPlayer is device dependent in that it performs poorly on certain devices especially older ones. Any professional developer should favor ExoPlayer instead. I fear that this “basic” video player will disappoint many people eventually. ExoPlayer also performs much better when it comes to streaming. What you’re doing right now is similar to implementing ListView only to end up implementing CollectionView down the road.

mattleibow:
Just an FYI about ExoPlayer: https://github.com/google/ExoPlayer

PaulVrugt:
Maybe we should first just get a basic version out there, and do improvements later. I understand it maybe can be done better, but at this rate, it'll take ages before this feature is released.

quentinR0bert:
@scxp what is the point of your commentary ? if you are not happy about this work, make your own Binding of ExoPlayer (already exist https://github.com/martijn00/ExoPlayerXamarin).

Some other people wait for a incorporated MediaElement, even if its basic, because it will be the official version of the Framework, with a community to support.

Stay strong @peterfoot and thanks for your work.

Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
peterfoot: The original code was built to only use Uri for source, so I replicated the custom ms- schemes across the other platforms. When I reworked it to use typed *MediaSource classes I left it in, but I can see the confusion. It was useful because it allowed a common method to reference files in the app package itself, along with common folders like temp and local data without knowing absolute paths.

Xamarin.Forms.Platform.Android/Renderers/MediaElementRenderer.cs
peterfoot: I'm not sure, but the platform code contains "slow" renderers for all the fast renderers so I'm assuming there are some cases it will fall back to using these? I'd rather we didn't need to maintain both of course!

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
peterfoot: On iOS this will be the local app Documents folder. The other option would be the Library folder which is meant for app data which is never visible to the user and not cloud synced. I'm guessing the latter would be the best option since it avoids accidentally cloud syncing large data files...

Xamarin.Forms.Platform.WPF/Renderers/MediaElementRenderer.cs
peterfoot: I think we'd have to create a folder beneath this with the app name, will investigate further

Xamarin.Forms.Platform.WPF/Renderers/MediaElementRenderer.cs
peterfoot: I've added more detail to the message. The documentation will need to call this out too.

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,471 @@
...
+
+			Performance.Start(out string reference);
+
+			if (oldElement != null)
+			{
+				oldElement.PropertyChanged -= OnElementPropertyChanged;
+				oldElement.SeekRequested -= MediaElementSeekRequested;
+				oldElement.StateRequested -= MediaElementStateRequested;
+				oldElement.PositionRequested -= MediaElementPositionRequested;
+				MediaElement.VolumeRequested -= MediaElementVolumeRequested;
```
rmarinho: oldElement

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,471 @@
...
+using CoreMedia;
+using Foundation;
+using System;
+using System.IO;
+using UIKit;
+using Xamarin.Forms.Internals;
+
+namespace Xamarin.Forms.Platform.iOS
+{
+	public sealed class MediaElementRenderer : UIView, IVisualElementRenderer, IEffectControlProvider
```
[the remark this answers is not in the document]
mattleibow: Can we randomly embed controllers in other views?

Xamarin.Forms.Platform.Android/Renderers/MediaElementRenderer.cs
mattleibow: Where are those default interface members when we need them, eh? https://github.com/dotnet/csharplang/issues/52

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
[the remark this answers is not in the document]
peterfoot: I think there was a reason why this wouldn't work related to the way the controller is created and the sequence of the controller and view creation, but admittedly this was before making it a fast-renderer so I will re-investigate.

Xamarin.Forms.Platform.WPF/Renderers/MediaElementRenderer.cs
peterfoot: This returns (for example) C:\Users\username\AppData\Roaming so we'd have to append either an appname or a company/appname. This has an impact on how the rest of the application data is stored and I'm not sure which convention to settle on?

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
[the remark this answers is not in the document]
peterfoot: I've been testing this approach, fundamentally it works and layout is behaving correctly until the device is rotated and then it screws up completely. I'm not entirely sure why...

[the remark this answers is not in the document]
peterfoot: I think the layout issue is now solved. It was a bit weird when you rotate the device before any media content has loaded.

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,471 @@
...
+
+		AVPlayerViewController _avPlayerViewController = new AVPlayerViewController();
+
+		NSObject _playToEndObserver;
+		NSObject _statusObserver;
+		NSObject _rateObserver;
+
+		VisualElement IVisualElementRenderer.Element => MediaElement;
+
+		UIView IVisualElementRenderer.NativeView => this;
```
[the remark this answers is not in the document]
peterfoot: Done

Xamarin.Forms.Core/MediaElement.cs
```diff
@@ -0,0 +1,300 @@
...
+		{
+			((MediaElement)bindable).OnSourcePropertyChanging((MediaSource)oldvalue, (MediaSource)newvalue);
+		}
+
+		void OnSourcePropertyChanging(MediaSource oldvalue, MediaSource newvalue)
+		{
+			if (oldvalue == null)
+				return;
+
+			oldvalue.SourceChanged -= OnSourceChanged;
```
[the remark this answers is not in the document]
peterfoot: This is notifying the MediaElement if the Source changes internally to trigger the MediaElement (and hence the renderer) to reload the video. This is the same pattern used for the Image control.

Xamarin.Forms.Platform.iOS/Renderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,471 @@
...
+					if (uriSource.Uri.Scheme == "ms-appx")
+					{
+						// used for a file embedded in the application package
+						asset = AVAsset.FromUrl(NSUrl.FromFilename(uriSource.Uri.LocalPath.Substring(1)));
+					}
+					else if (uriSource.Uri.Scheme == "ms-appdata")
+					{
+						string filePath = string.Empty;
+
+						if (uriSource.Uri.LocalPath.StartsWith("/local"))
```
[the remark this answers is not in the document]
peterfoot: I've added a new method there and moved the Uri parsing logic for ms-appdata.

Xamarin.Forms.ControlGallery.iOS/Xamarin.Forms.ControlGallery.iOS.csproj
```diff
@@ -60,7 +60,7 @@
     <WarningLevel>4</WarningLevel>
     <ConsolePause>false</ConsolePause>
     <MtouchArch>ARMv7, ARM64</MtouchArch>
-    <CodesignKey>iPhone Developer</CodesignKey>
+    <CodesignKey>iPhone Developer: Peter Freeman Foot (NFPCJUZSF7)</CodesignKey>
```
modplug: I think you left this in by accident

samhouts:
Thanks for keeping this up to date! We haven't forgotten about this. We're hoping to do a community review of this and CameraView in the next month or so.

rmarinho:
Hi Peter can you update PR description with all the Public API and methods so it can be easier to review, can we also rebase or merge this to get up to date. Thanks.

davidbuckleyni:
How do you play a local mp3 file please is this only in vnext 4.10?.

peterfoot:
> How do you play a local mp3 file please is this only in vnext 4.10?.

This is currently not in a release - you'd have to build the source yourself from this branch.

davidbuckleyni:
Will this allowing playing of any mp3 local or other wise have you an example of code or will it still use androids start stop pause etc of media player and how long before it’s released ?

peterfoot:
> Will this allowing playing of any mp3 local or other wise have you an example of code or will it still use androids start stop pause etc of media player and how long before it’s released ?

You can play both remote and local files. You can either use the native transport controls or programmatically Play/Pause/Stop from your own custom UI. It's currently scheduled for the 4.2 release but I don't know the timeframe for that release.

davidbuckleyni:
> > Will this allowing playing of any mp3 local or other wise have you an example of code or will it still use androids start stop pause etc of media player and how long before it’s released ?
> 
> You can play both remote and local files. You can either use the native transport controls or programmatically Play/Pause/Stop from your own custom UI. It's currently scheduled for the 4.2 release but I don't know the timeframe for that release.

Great thanks though I have seen they're now its been paused from being merged as I no the client of app will want a dam barcode sound when the scan is finished go figure cleints lol.

peterfoot:
@samhouts which branch does it need to target?

rmarinho:
Hum i think the milestone is wrong that s why its market as retarget branch required

jfversluis:
Hey @peterfoot! Just tried to build it, unfortunately it fails on

> Xamarin.Forms.Platform.UAP\ImageExtensions.cs(10,24): Error CS0101: The namespace 'Xamarin.Forms.Platform.UWP' already contains a definition for 'ImageExtensions'

The retarget was on by mistake or maybe something that was intended at that moment, for now this goes into master so that’s all good. If you could fix this issue I will make sure that some people will have another look at this and we should get this in ASAP! Thanks!

peterfoot:
I've renamed the offending class and moved it to the Extensions folder for clarity.

martijn00:
It might be good to make sure the same things work with [MediaManager](https://github.com/martijn00/XamarinMediaManager) in this case, as it does with FFImageLoading for Images. That way people can easily swap out the Forms implementation for a better alternative.

peterfoot:
@martijn00 can you explain in a little more detail what you are requesting?

martijn00:
This: https://github.com/luberda-molinet/FFImageLoading/wiki/Xamarin.Forms-API#iimageviewhandler--iimagesourcehandler but then for media.

rmarinho:
Right, I had some notes about this PR:

- Should Duration, Open, Close be on MediaSource?
-   Can we share media source between 2 views?
- Should use or Registar approach Internals.Registrar.Registered.GetHandlerForObject<IMediaSourceHandler>(source);  
- No need of fast renderer
- Should we had some extensions methods   MediaSource IsVideo?!
-Should have    MediaElement and MediaElementView

peterfoot:
@rmarinho At the moment the MediaSource is a very thin wrapper over a file path or Uri it doesn't handle any of the loading itself so yes you could pass the same one to multiple controls but it would open each separately (no single Open/Close etc actions). 
The control was originally modelled on the UWP control and so the Source property was simply a Uri with the special ms-appx and ms-appdata implement across all the platforms. This was later changed after a code review to a MediaSource but the logic of opening the source remains in the renderer because the way different platforms handle it is very different and this provides a simple consistent approach. So yes it's nothing like ImageSource, but that's how it ended up the way it is today.

Other than looking at the file extension the MediaSource itself would have no idea if the content is a video - this would only be known after it was opened. The control would have to set this based on the current implementation. It doesn't seem to me this would add much value, especially in a first release.

I don't understand "No need for a fast renderer". It removes one of the nested native views for Android even if the subsequent performance benefit is minimal. I was explicitly asked to remove the original renderer and just have the fast renderer months ago.

The concept of a MediaElement and MediaElementView seems to be splitting the control and ending up with something closer to the more full-featured MediaManager which this control is not designed to compete with. I do take the point on-board about making the MediaSource more flexible so that the implementation can be swapped out as required.

samhouts:
@peterfoot Can you give me rights to push to your branch, please? :)

samhouts:
Okay! We've added an experimental flag! One step closer to merge so we can get this tested in the wild! Thanks everyone for your patience, and THANK YOU @peterfoot for keeping this alive!

rmarinho:
Hi @peterfoot can we rebase/merge master? thanks

peterfoot:
@rmarinho will try again to rebase tomorrow as I ran into errors running it today.

Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,480 @@
...
+using Android.Widget;
+using System;
+using System.ComponentModel;
+using Xamarin.Forms.Internals;
+using Xamarin.Forms.Platform.Android.FastRenderers;
+using AView = Android.Views.View;
+
+namespace Xamarin.Forms.Platform.Android
+{
+	public sealed class MediaElementRenderer : FrameLayout, IVisualElementRenderer, IViewRenderer, IEffectControlProvider, MediaPlayer.IOnCompletionListener, MediaPlayer.IOnInfoListener, MediaPlayer.IOnPreparedListener, MediaPlayer.IOnErrorListener
```
PureWeen: @samhouts do we need this to be sealed? I realize it's experimental but seeing as how it's experimental being able to extend the renderer seems even more important because of the possibility of needing to add behavior or work around bugs

Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,480 @@
...
+		{
+			Controller.Duration = _view.DurationTimeSpan;
+			Controller.VideoHeight = _view.VideoHeight;
+			Controller.VideoWidth = _view.VideoWidth;
+
+			Device.BeginInvokeOnMainThread(UpdateLayoutParameters);
+		}
+
+		//todo: make virtual when unsealed
+		protected void OnElementPropertyChanged(object sender, PropertyChangedEventArgs e)
```
PureWeen:
```suggestion
		void OnElementPropertyChanged(object sender, PropertyChangedEventArgs e)
```

Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
```diff
@@ -0,0 +1,480 @@
...
+					if (Platform.GetRenderer(Element) == this)
+						Element.ClearValue(Platform.RendererProperty);
+				}
+			}
+
+			base.Dispose(disposing);
+		}
+
+		//todo: make virtual when unsealed
+		protected void OnElementChanged(ElementChangedEventArgs<MediaElement> e)
```
PureWeen:
```suggestion
		void OnElementChanged(ElementChangedEventArgs<MediaElement> e)
```

PureWeen:
@peterfoot can you give me access to make some changes so I can fix the compile errors?

peterfoot:
@PureWeen I fixed those two but you should have access now if you need to make any other changes

rmarinho:
@peterfoot needs rebase or merge master

samhouts:
Failing tests appear to be unrelated and are passing locally. 

DRUMROLL PLEASE

abdu292:
@peterfoot Congratulations and this is amazing! One question though. How to enable lock screen notifications?
(planning to test my custom implementation of audio player with this one), thanks in advance!

peterfoot:
@abdu292 This isn't supported currently. It will require some investigation to find a cross-platform way of exposing this but it's a good idea for a future update!

abdu292:
thanks @peterfoot. Eagerly waiting for this to be a part. MediaManager plugin has something implemented, which you may have a look at, to get some idea on this.

davidbuckleyni:
@peterfoot Congrats Peter just saw the merge request cant wait get this down and play with it. Planning on making a running app and music is one of its features. Do we no what live version this will go in to or will it be pre-release for now.

martijn00:
@abdu292 it's never going to happen. MediaManager has a dependency on Exoplayer and I don't believe Forms is going to do that. Better use MediaManager for everything media related ;)

abdu292:
thanks @martijn00 . I meant to say to look at the media manager code just to get an idea and not to copy the exact code. Media manager uses Exo for android, that doesn't mean Notifications can't be integrated with simple Media player (as you must know). Player and notifications are two different things, though exo has a nice way of integrating notification.  You/Baseflow have done an excellent job with MediaManager. I have been using MediaManager in the past months. However I ran in to multiple issues - may be my specific use cases - and started writing my own player & notification implementations. Also, when I looked at the history of release it's really scary. It doesn't have a consistent maintaining history, release are not at all consistent (Yes, I understand its open source and releases don't have to be consistent), some comments in the github issues give an impression to the developers that the owner is angry/owner is not interested to maintain the repo any more, etc. 
If a feature is natively available in Xamarin.Forms,  it is unlikely that a developer will go for a third party plugin if he/she gets all the things he/she needs.
thanks for all your work on the MediaManager. I'll definitely look back to this once it hits its first 1.x release. Regards!

davidbuckleyni:
Might be dumb question been away from xammy for a while how does one turn on the feature flag what file ???

jfversluis:
No problem @davidbuckleyni!

You can find that info on the Docs for the actual control. In this case right under the first screenshot [here](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/mediaelement).

I have a somewhat more generic blog post about it [here](https://blog.verslu.is/xamarin/xamarin-forms-xamarin/experimental-flags-xamarin-forms/).

PR #3518, opened 2018-08-08
 [iOS] Unsubscribe CellPropertyChanged when SwitchCellRenderer is disposed

### LINKED ISSUE #3408: [iOS] System.ObjectDisposedException: from SwitchCellRenderer when changing ItemSource
### Description
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'CellTableViewCell'.

The exception happens on this line

https://github.com/xamarin/Xamarin.Forms/blob/d3d59ee4f0b3098457e1debe8d7b03d0d0061a53/Xamarin.Forms.Platform.iOS/Cells/SwitchCellRenderer.cs#L5

### Steps to Reproduce

1. Run the Reproduction
2. click on Tabby and/or sphinx at the top

### Expected Behavior
Not to crash

### Actual Behavior
Crashes

### Basic Information

- Version with issue: 3.2 
- Last known good version:  not sure. I tested back to 2.5 and it has the same issue
- IDE: Visual Studio
- Platform Target Frameworks: 
  - iOS:  11.4
  - Android: 8.1
  - UWP:  16299 

### Reproduction Link

[AzureChallengerMobileApp.zip](https://github.com/xamarin/Xamarin.Forms/files/2233342/AzureChallengerMobileApp.zip)

rmarinho:
### Description of Change ###

SwitchCellRenderer wasn't unsubscribing when disposed. 
The best way would be to just use the event exposed by our `CellTableViewCell` 

We should use this in other cell renderers, this will allow to control the unsubscribing  when cell's are dispose and reused. 

### Issues Resolved ###

- fixes #3408 

### API Changes ###

Changes to `CellTableViewCell`

Add:

`public void HandleCellPropertyChanged(object sender, PropertyChangedEventArgs e)`

`public Action<object, PropertyChangedEventArgs> CellPropertyChanged`

Remove: 

`public Action<object, PropertyChangedEventArgs> PropertyChanged`

### Platforms Affected ###

- iOS

Doesn't crash when SwitchCells go away (are disposed) 

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3408.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3408.cs
@@ -1,0 +1,868 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Collections.Specialized;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using Xamarin.Forms;
+using static Xamarin.Forms.Controls.Issues.Issue3408;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 3408, "System.ObjectDisposedException: from SwitchCellRenderer when changing ItemSource", PlatformAffected.iOS)]
+	public class Issue3408 : TestContentPage
+	{
+		public static List<Recommendation> GetRecommendations(object e)
+		{
+			switch (e)
+			{
+				case List<RecommendationsViewModel> pc: return pc.First().Recommendations;
+				case List<RecommendationsViewModel2> pc: return pc.First().Recommendations;
+				default: return null;
+			}
+		}
+		protected override void Init()
+		{
+
+			var grd = new Grid();
+
+			var aacountListView = new RepeaterSwipeView();
+			aacountListView.ItemTemplate = new AccountDetailsDataTemplateSelector();
+			aacountListView.BindingContext = new List<RecommendationsViewModel> { new RecommendationsViewModel() };
+
+			aacountListView.SetBinding(RepeaterSwipeView.ItemsSourceProperty, ".");
+			var btn = new Button
+			{
+				Text = "Change Source",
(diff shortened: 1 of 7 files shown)


PureWeen:
@rmarinho  yea that's a good idea.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3408.cs
```diff
@@ -0,0 +1,212 @@
...
+				})
+
+			};
+
+			lst.SetBinding(ListView.ItemsSourceProperty, new Binding("Recommendations"));
+			grd.Children.Add(lst);
+			Content = grd;
+		}
+
+		object context = null;
```
PureWeen: Add a reference to this issue
https://github.com/xamarin/Xamarin.Forms/issues/3602

So future people know why this is here

Xamarin.Forms.Platform.iOS/Cells/ImageCellRenderer.cs
```diff
@@ -22,8 +22,8 @@ public override UITableViewCell GetCell(Cell item, UITableViewCell reusableCell,
 
 		protected override void HandlePropertyChanged(object sender, PropertyChangedEventArgs args)
```
PureWeen: I was thinking renaming those override was a good way to go because the type on the sender is now different. 

Now if someone is overriding this and casting sender they are going to get an Invalid Cast Exception at run time.

What about just renaming this one to how it was HandleCellPropertyChanged

And then from there you can call HandlePropertyChanged with the expected sender that way anyone that has overridden this will still get the class they expect

PureWeen: I was thinking renaming the  override was a good way to go because the type on the sender is now different. 

Now if someone is overriding this and casting sender they are going to get an Invalid Cast Exception at run time.

What about just renaming this one to how it was HandleCellPropertyChanged

And then from there you can call HandlePropertyChanged with the expected sender that way anyone that has overridden *HandlePropertyChanged* will still get the class they expect

rmarinho:
Thanks @PureWeen for the feedback most of your references have cross my mind before but I went other way, good to have you double check and get this In better shape so we don't break our users.

Ready for a second review . Thanks

kingces95:
Weird that Android tests are failing..

jassmith:
build --uitests

jassmith:
iOS failure unrelated, android fails completely unrelated and are due to Azure being bad.

PR #3519, opened 2018-08-09
 Implemented SelectedTintColor and UnselectedTintColor for iOS

### LINKED ISSUE #3478: Allow TintColor and UnselectedItemTintColor as platform-specifics
### Description
Now that we have...

```
On<Android>().SetBarItemColor(Color.Red);           // Unselected image+text color
On<Android>().SetBarSelectedItemColor(Color.White); // Selected image+text color
```

...we no longer need a TabbedPage renderer for Android just to set these items. However, it is still required to have an iOS TabbedPage renderer to set the equivalents on iOS side (`TintColor` and `UnselectedItemTintColor`). For feature parity between the two (and for ease of setting these very common properties), I propose the equivalents on iOS such as...

```
On<iOS>().SetTintColor(Color.Red);                 // Unselected image+text color
On<iOS>().SetUnselectedItemTintColor(Color.White); // Selected image+text color
```

### Basic Information

Xamarin.Forms 3.1.0.697729

jfversluis:
### Description of Change ###

Through the PlatformSpecific API it is now possible to set the selected and unselected tint item on iOS.
The unselected item color is only supported on iOS 10 and higher.

- fixes #3478 

### API Changes ###

Added:
 - TabbedPage.On<iOS>().GetUnselectedTintColor();
 - TabbedPage.On<iOS>().GetSelectedTintColor();

### Platforms Affected ###

- iOS

This does not change anything when upgraded as-is. Only when the user leverages this functionality they will see the desired colors in the tab bar on iOS.

### PR Checklist ###

- [ ] Has automated tests 


jfversluis:
I agree. Just after creating this I thought the same thing. Unify the API and make it available for as much of the platforms as possible.

Although it is similar to the PR you mentioned, that is for the ListView, right? The concept is the same but implementation totally different. Or did you just mean in terms of how it should/could work?

rmarinho:
Oh right sorry, when I read the the title I was think this PR was LIstVIew I know see is for TabbedPage

jfversluis:
But other than that. Do we want to get this in de non-platform specific APIs? Seems to make sense.

Xamarin.Forms.Platform.iOS/Renderers/TabbedRenderer.cs
```diff
@@ -363,6 +372,18 @@ void UpdateBarTextColor()
 			TabBar.TintColor = isDefaultColor ? _defaultBarTextColor : barTextColor.ToUIColor();
 		}
 
+		void UpdateTintColor()
+		{
+			if (Tabbed == null || TabBar == null || TabBar.Items == null)
+				return;
+
+			if (SelectedTintColor != Color.Default)
```
rmarinho: what if you want to return to the default color?  you will need to save the default color of the platform and use it when the Color is set to Default. and had change before.

PR #3567, opened 2018-08-14
[Android] Updating Gradle

yeomann:
### Description of Change ###
updating Gradle from old version to Gradle 3+

### Platforms Affected ###
- Android


rmarinho:
We won't be adding this right now as we don't need it we might reopen later.

PR #3568, opened 2018-08-14
[Android] Updating Gradle version

yeomann:
### Description of Change ###
updating Gradle from old version to Gradle 3+

### Platforms Affected ###
- Android

### DIFF
--- a/AndroidNative/build.gradle
+++ b/AndroidNative/build.gradle
@@ -4,7 +4,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.2.0'
+        classpath 'com.android.tools.build:gradle:3.1.1'
     }
 }
 


hartez:
What's the purpose of updating to a newer version of Gradle? What does it give us?

jonathanpeppers:
@hartez your java source is so small you might not even need it.

I think you could use newer java features, maybe they improved build times, etc.

You have only one java class though...

hartez:
> your java source is so small you might not even need it.

Might not need the upgrade, or might not need gradle at all?

Honestly, the gradle stuff is only there as a quick-and-dirty way to make our single `.jar` buildable from the command line on Windows and macOS. If there's a lower-effort way to build that, I'm all for it.

jonathanpeppers:
> Might not need the upgrade, or might not need gradle at all?

I think might not need to update gradle, if you don't need newer Android build features.

Xamarin.Android has places where we setup a custom MSBuild project to run javac: https://github.com/xamarin/xamarin-android/blob/7d32ef3e6ef88eba004f1d2a9f9eb0f42cbf887e/build-tools/scripts/JavaCallableWrappers.targets#L34

Although I don't know if that is worth the work either, if you already have it building. The way you have it now, you can open in Android Studio, right?

hartez:
> Although I don't know if that is worth the work either, if you already have it building. The way you have it now, you can open in Android Studio, right?

Yes. Though the point of setting this up was to avoid having to open Android Studio at all :)

We rarely make changes to FormsViewGroup (maybe once a year), so streamlining the process further isn't a high priority right now. If we decided to bring it into the CI builds at some point, then we might move to using your MSBuild target.

rmarinho:
We won't be adding this right now as we don't need it we might reopen later.

PR #3569, opened 2018-08-14
add project-wide properties as ext block

yeomann:
### Description of Change ###

ref: https://developer.android.com/studio/build/
add better way to get the Gradle based values.
Aslo upgrade to SDK 27

### Issues Resolved ###

additionally, changed the SDK values to 27 as per new changes coming.

Android Target API Level 26 will be required in August 2018.
https://android-developers.googleblog.com/2017/12/improving-app-security-and-performance.html

### DIFF
--- a/AndroidNative/FormsViewGroup/build.gradle
+++ b/AndroidNative/FormsViewGroup/build.gradle
@@ -1,12 +1,21 @@
 apply plugin: 'com.android.library'
 
+ext {
+    buildToolsVersion = "27.0.3"
+    minSdkVersion = 21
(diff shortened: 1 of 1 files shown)


AndroidNative/FormsViewGroup/build.gradle
```diff
@@ -1,12 +1,21 @@
 apply plugin: 'com.android.library'
 
+ext {
+    buildToolsVersion = "27.0.3"
+    minSdkVersion = 21
```
hartez: For the time being, Forms still has to support a minimum of API 15.

PR #3605, opened 2018-08-17
[Core] When building SLN Forms.Core dll doesn't generate

PureWeen:
### Description of Change ###
When building Xamarin.Forms.SLN fresh the build fails because the Xamarin.Forms.Core.dll doesn't get produced.

If you manually build Xamarin.Forms.Core.csproj first then it produces the dll and everything works fine after that.

If you remove the *XFCorePostProcessor UsingTask* then everything compiles fine first go, but this Task shouldn't get in the way of producing "Xamarin.Forms.Core.dll". 

Looking at the build output it looks like msbuild makes the wrong determination about whether it should be running *Csc*

```
1>  Skipping target "CoreCompile" because all output files are up-to-date with respect to the input files.
```

I'll log a separate issue with Visual Studio 

### Platforms Affected ###

- Core/XAML (all platforms)

### PR Checklist ###

- [ ] Has automated tests 


Xamarin.Forms.Core/Xamarin.Forms.Core.csproj
```diff
@@ -21,7 +21,7 @@
 	</ItemGroup>
 	<Import Project="..\Xamarin.Flex\Xamarin.Flex.projitems" Label="Shared" Condition="Exists('..\Xamarin.Flex\Xamarin.Flex.projitems')" />
 	<UsingTask TaskName="XFCorePostProcessor.Tasks.FixXFCoreAssembly" AssemblyFile="..\XFCorePostProcessor.Tasks\bin\Debug\netstandard2.0\XFCorePostProcessor.Tasks.dll" />
-	<Target AfterTargets="AfterCompile" Name="ShameHat">
+	<Target  Condition=" ! ($(DesignTimeBuild) == true OR $(BuildingProject) != true)" AfterTargets="AfterCompile" Name="ShameHat">
```
StephaneDelcroix: that isolated `!` look out of place, couldn't you change it to `!= true` ?

PureWeen: Yea basically I just took what they had here
https://github.com/dotnet/project-system/blob/master/docs/design-time-builds.md#determining-whether-a-target-is-run-in-a-design-time-build

And then added the *!* to the whole thing. Mainly because I've messed up trying to simplify these msbuild conditions down in the past since they can also be ('') so just went the easiest route

That being said I still just checked in a simplified version that should be the same

PR #3618, opened 2018-08-21
Specify minimum UAP version in nuspec

### LINKED ISSUE #2859: UWP build fails when minimum version is Anniversary Update or lower
### Description
Compiling a UWP project that has a minimum version set to Windows 10 Anniversary Update or lower will fail when using Xamarin.Forms 3.0. This is because there is now a dependency on .NET Standard 2.0 for UWP which is only supported when using a minimum version of Fall Creator's Update or higher. 

To prevent this, the UWP target framework version in the nuspec should be updated to 10.0.15138 like in the [Microsoft.NETCore.UniversalWindowsPlatform](https://www.nuget.org/packages/Microsoft.NETCore.UniversalWindowsPlatform/6.1.4) package.

Also reported here: https://developercommunity.visualstudio.com/content/problem/256382/netstandard-compatibility.html

### Steps to Reproduce

1. Create a new Xamarin.Forms project using a shared project
2. In the UWP project's properties, change the Minimum version to Windows 10 Anniversary Update or lower
3. Build the UWP project

### Expected Behavior
The build will complete successfully

### Actual Behavior
The build fails with several errors such as

> The type 'Type' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'
> The type 'Object' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'

### Basic Information

- Version with issue: 3.0+
- Last known good version: 2.5 series
- IDE: VS2017
- Platform Target Frameworks: 
  - UWP:  
    - Min version: Anniversary update, build 14393
    - Target version: April 2018 update, build 17134

samhouts:
### Description of Change ###

Changing the nuspec's targetFramework from uap10.0 to UAP10.0.16299.0 to give users an indication that this package is meant for Fall Creators' Update and above.

### Issues Resolved ###

- related to but does not fix #2859

### API Changes ###

None

### Platforms Affected ###

- UWP

None

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/.nuspec/Xamarin.Forms.nuspec
+++ b/.nuspec/Xamarin.Forms.nuspec
@@ -20,7 +20,7 @@
         <dependency id="Xamarin.Android.Support.v7.CardView" version="27.0.2"/>
         <dependency id="Xamarin.Android.Support.v7.MediaRouter" version="27.0.2"/>
       </group>
-      <group targetFramework="uap10.0">
+      <group targetFramework="UAP10.0.16299.0">
         <dependency id="NETStandard.Library" version="2.0.1"/>
         <dependency id="Microsoft.NETCore.UniversalWindowsPlatform" version="6.0.1" />
       </group>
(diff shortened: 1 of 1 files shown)


rmarinho:
@samhouts or @hartez we need to fix the verify nuspec scripts on the submodule

rmarinho:
@samhouts so didn't you said this actually didn't fixed anything as it will always try to install ? Should it close till we have more feedback from the Nuget team ? 
@hartez can we check what we need to get the script going so we have a package to test also? 

Thanks

PR #3635, opened 2018-08-22
[iOS] Specify a default size for UISearchBar width if needed

### LINKED ISSUE #2139: SearchBar not visible in iOS 10 when using CenterAndExpand

### Description
The Xamarin Forms SearchBar does not appear when using iOS 10. It appears to be fine on iOS 11. Adding HeightRequest does not seem to work
### Steps to Reproduce

1. Add a SearchBar to a Xamarin Forms Content Page
2. Run an iOS simulator or device in iOS 10.3
3. SearchBar is not visible

### Expected Behavior
SearchBar should appear
### Actual Behavior
SearchBar does not appear
### Basic Information

- Version with issue: iOS 10.3
- Last known good version: iOS 11.2
- IDE:
- Platform Target Frameworks: 
  - iOS:  10.3
  - Android:  
  - UWP:   
- Nuget Packages:
- Affected Devices:

### LINKED ISSUE #3413: [iOS] Searchbar in Horizontal Stacklayout doesn't render
### Description
Place a SearchBar control inside a Stacklayout with Orientation Horizontal and it won't render on iOS. It renders on Android and UWP just fine

https://forums.xamarin.com/discussion/95897/search-bar-in-horizontal-stacklayout

### Steps to Reproduce

Create a page with the following XAML

```XAML
            <SearchBar  BackgroundColor="Yellow"  Text="candy mountain" ></SearchBar>
        </StackLayout> 
```

### Expected Behavior
SearchBar should render

### Actual Behavior
SearchBar doesn't render

### Basic Information

- Version with issue: 3.2
- Last known good version: I tested as far back as 2.5.1 and it didn't render there either
- Platform Target Frameworks: 
  - iOS:  11.4

### Screenshots
Android:
image

iOS:
image

### Reproduction Link
 
[StackLayout.zip](https://github.com/xamarin/Xamarin.Forms/files/2233747/StackLayout.zip)

rmarinho:
### Description of Change ###

As of latest versions of iOS11.3 seems that passing the max value to the UISearchBar for sizing breaks.
This only happens when we don't specify any Width request for the SearchBar and the SearchBar is requested to stretch itself. 
We could try to pass a enormous value but I figure this was a best call. 

At the same time we found out iOS10 also wasn't rendering , the issue that is SizeThatFits always is retiring a 0 Width , we override the call and set the width. 

### Issues Resolved ###

- fixes #3413 
- fixes #2139  

### API Changes ###

None

### Platforms Affected ###

- iOS

Should show the Searchbar inside a horizontal StackLayout and blue SearchBar Expanding

Before:

It crashed  and didn't renderer the blue SearchBar with CenterAndExpand

After:

IOS 10
simulator screen shot - iphone 5 - 2018-08-29 at 12 38 38
simulator screen shot - iphone 5 - 2018-08-29 at 12 38 46

iOS11
simulator screen shot - iphone x - 2018-08-29 at 12 20 30
simulator screen shot - iphone x - 2018-08-29 at 12 20 44

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3413.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3413.cs
@@ -1,0 +1,66 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 3413, "[iOS] Searchbar in Horizontal Stacklayout doesn't render", PlatformAffected.iOS)]
+	public class Issue3413 : TestContentPage
+	{
+		protected override void Init()
+		{
+			Padding = new Thickness(20);
+
+			var layout = new StackLayout
+			{
+				Orientation = StackOrientation.Vertical
+			};
+
+			var searchBar = new SearchBar
+			{
+				BackgroundColor = Color.Yellow,
+				Text = "i m on a vertical stacklayout",
+				AutomationId = "srb_vertical"
+			};
+			layout.Children.Add(new Label { Text = "Vertical" });
+			layout.Children.Add(searchBar);
+
+			var layout1 = new StackLayout
+			{
+				Orientation = StackOrientation.Horizontal
+			};
+
+			var searchBar1 = new SearchBar
+			{
+				BackgroundColor = Color.Yellow,
+				Text = "i m on a horizontal stacklayout",
+				AutomationId = "srb_horizontal"
+			};
+
+			layout1.Children.Add(new Label { Text = "Horizontal" });
+			layout1.Children.Add(searchBar1);
+
+			var grid = new Grid();
+			grid.Children.Add(layout);
+			Grid.SetRow(layout, 0);
+			grid.Children.Add(layout1);
+			Grid.SetRow(layout1, 1);
+			Content = grid;
+		}
+
+#if UITEST
+		[Test]
+		public void Issue3413Test ()
+		{
+			RunningApp.WaitForElement (q => q.Marked ("srb_vertical"));
+			RunningApp.WaitForElement (q => q.Marked ("srb_horizontal"));
+			RunningApp.Screenshot ("Please verify we have 2 SearchBar's. One below the label, other side by side with the label");
+		}
+#endif
+	}
+}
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3413.cs
```diff
@@ -0,0 +1,66 @@
...
+			grid.Children.Add(layout);
+			Grid.SetRow(layout, 0);
+			grid.Children.Add(layout1);
+			Grid.SetRow(layout1, 1);
+			Content = grid;
+		}
+
+#if UITEST
+		[Test]
+		public void Issue3413Test ()
```
samhouts: maybe add `ManualReview` test category

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3413.cs
```diff
@@ -0,0 +1,66 @@
...
+		{
+			RunningApp.WaitForElement (q => q.Marked ("srb_vertical"));
+			RunningApp.WaitForElement (q => q.Marked ("srb_horizontal"));
+			RunningApp.Screenshot ("Please verify we have 2 SearchBar's. One below the label, other side by side with the label");
```
samhouts: SearchBars (no apostrophe)

rmarinho:
Hey @PureWeen  thanks for the test on older devices.

I had a look and I got it to work on iOS10, not sure if it worked without this on the pass or not.. but basically SizeToFit returns different values , for example is always return width of 0 even if you pass for example 300 on iOS10.. So the trick is set the with after and not call the base SizeThatFits of the renderer. 

At the same time i see that after rotating is not filling all space on iOS10 (on the right) I might investigate further that to see if I can fix that too. 

iOS 11 left, ios10 right 

screen shot 2018-08-23 at 12 32 33

screen shot 2018-08-23 at 12 52 30

rmarinho:
Found #2139 I bet is related

samhouts:
@rui 
> Found #2139 I bet is related
Does that mean this PR should fix that issue?

rmarinho:
Yes it does @samhouts I updated the screenshots and issue to show the "blue" searchbar that's the one using CenterAndExpand.

@PureWeen I tried another fix, and I think is better , I also updated screenshots. Please review. Thanks

The big issue is SizeThatFits is always broken on iOS10 for the with, passing a value, infinity or 0 will always return values that are not consistent with iOS11.

kingces95:
Well, I tried these modification https://github.com/xamarin/Xamarin.Forms/pull/3686 but I'm seeing odd behavior all around. When I rotate the device the code in question is not always called tho the visual effect is different between my proposed changes and your code. So... I dunno what's going on. Anyway,

I find the trinay statement hard to grock. Tried to split it up.

The comment talks about a zero value so I expected the code to check for zero.

Tried to generalize the version check for all versions. Fewer version checks the better IMHO. 

But like I said, my code works when it's called but it's not always called on rotation for <11 devices.

Issue #2894, opened 2018-06-01
[iOS] Gesture Recognizers added to Span after it's been set to FormattedText don't work and can cause an NRE


PureWeen:
### Description

- Adding a TapGestureRecognizer to a span after it's been set to FormattedText on a label doesn't work
- If you now click that Span really fast you'l get an NRE

### Steps to Reproduce

click on the 
*"I will not fire when clicked"*
text and notice nothing happens
click on it really fast and see an NRE exception

```c#
var label = new Label();

			var s = new FormattedString();
			var span = new Span { Text = "I will fire when clicked", FontAttributes = FontAttributes.Bold };
			var span2 = new Span { Text = "I will not fire when clicked", FontAttributes = FontAttributes.Bold };
			span.GestureRecognizers.Add(new TapGestureRecognizer()
			{
				Command = new Command(async () =>
			   {
				   await MainPage.DisplayAlert("Yay clicked", "yay clicked", "yay clicked");
			   })
			});

			s.Spans.Add(span);
			s.Spans.Add(span2);

			label.FormattedText = s;
			span2.GestureRecognizers.Add(new TapGestureRecognizer()
			{
				Command = new Command(async () =>
			   {
					await MainPage.DisplayAlert("not working", "not working", "not working");
			   })
			});

			return new ContentPage()
			{
				Content = new StackLayout()
				{
					Children = { label },
					Padding = 40
				}
			};
```

### Expected Behavior
Both spans should be clickable and clicking rapidly shouldn't cause a NRE

### Actual Behavior
When clicking on the second span nothing happens.
If you click the span really fast it'll cause a NRE

### Basic Information

- Version with issue: 3.1 nightly
- IDE: VS for Mac
- Platform Target Frameworks:  
  - iOS:  11.2

StephaneDelcroix:
moved to sprint 138 as it doesn't affect 3.1.0

PR #3639, opened 2018-08-22
[Core] fixed #3524 - set binding context on Span Gestures

### LINKED ISSUE #3524: ICommand binding from a TapGestureRecognizer on a Span doesn't work
### Description

I tried out the new gesture recognizer functionality on Spans that's part of 3.2. When binding to an `ICommand` from the `Command` property of a `TapGestureRecognizer`, the `ICommand` isn't executed when a tap occurs. This is true for both XAML and C# implementations.

```
                    <Span Text="default, " Style="{DynamicResource BodyStyle}">
                        <Span.GestureRecognizers>
                            <TapGestureRecognizer Command="{Binding TapCommand}" />
                        </Span.GestureRecognizers>
                    </Span>
```

```
            var span = new Span { Text = "default, " };
            var tapGesture = new TapGestureRecognizer();
            tapGesture.SetBinding(TapGestureRecognizer.CommandProperty, "TapCommand");
            span.GestureRecognizers.Add(tapGesture);
            formattedString.Spans.Add(span);
```

### Steps to Reproduce

1. Run the attached sample. It will launch the LabelPage.
2. Tap the "default" text, in the final Label (this Label comprises three Spans).

### Expected Behavior

The `TapGestureRecognizer` detects the tap and calls the execute delegate of the bound `ICommand`, which displays an alert.

### Actual Behavior

Nothing happens.

### Basic Information

- Version with issue: 3.2.0.637442-pre1
- IDE: VSMac
- Platform Target Frameworks: 
  - iOS:  11.4
  - Android: 8.1
  - UWP:  Not tried. 
- Android Support Library Version: 27.0.2

### Reproduction Link

[Text.zip](https://github.com/xamarin/Xamarin.Forms/files/2274014/Text.zip)

PureWeen:
### Description of Change ###
The BindingContext on the span wasn't being propagated down to the Gestures. I took the code that performs this Propagation with Views and centralized it into an extension method

### Issues Resolved ### 
- fixes #3524

### Platforms Affected ### 
- Core/XAML (all platforms)

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/BindableObjectExtensions.cs
+++ b/Xamarin.Forms.Core/BindableObjectExtensions.cs
@@ -13,20 +13,16 @@ namespace Xamarin.Forms
 
 		internal static void PropagateBindingContext<T>(this BindableObject self, IList<T> children, Action<BindableObject, object> setChildBindingContext)
 		{
-			var gotBindingContext = false;
-			object bc = null;
+			if (children == null || children.Count == 0)
+				return;
+
+			var bc = self.BindingContext;
 
 			for (var i = 0; i < children.Count; i++)
 			{
 				var bo = children[i] as BindableObject;
 				if (bo == null)
 					continue;
-
-				if (!gotBindingContext)
-				{
-					bc = self.BindingContext;
-					gotBindingContext = true;
-				}
 
 				setChildBindingContext(bo, bc);
 			}
--- a/Xamarin.Forms.Core/BindableObjectExtensions.cs
+++ b/Xamarin.Forms.Core/BindableObjectExtensions.cs
@@ -1,10 +1,37 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq.Expressions;
 
 namespace Xamarin.Forms
 {
 	public static class BindableObjectExtensions
 	{
+		internal static void PropagateBindingContext<T>(this BindableObject self, IList<T> children)
+		{
+			PropagateBindingContext(self, children, BindableObject.SetInheritedBindingContext);
+		}
+
+		internal static void PropagateBindingContext<T>(this BindableObject self, IList<T> children, Action<BindableObject, object> setChildBindingContext)
+		{
+			var gotBindingContext = false;
+			object bc = null;
+
+			for (var i = 0; i < children.Count; i++)
+			{
+				var bo = children[i] as BindableObject;
+				if (bo == null)
+					continue;
+
+				if (!gotBindingContext)
+				{
+					bc = self.BindingContext;
+					gotBindingContext = true;
+				}
+
+				setChildBindingContext(bo, bc);
+			}
+		}
+
 		public static void SetBinding(this BindableObject self, BindableProperty targetProperty, string path, BindingMode mode = BindingMode.Default, IValueConverter converter = null,
 									  string stringFormat = null)
 		{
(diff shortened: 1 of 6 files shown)


PureWeen:
@samhouts  so the NRE will actually be fixed by the PR for https://github.com/xamarin/Xamarin.Forms/issues/2894

I was going to package this one with that one but it turns out that 2894 is also broken on UWP/ios :-/ 

So I split this one out  just to get it in

the NRE happens from the change EZ added that causes the control gallery to reset after 5 clicks so those all go null and then the click is processed

Xamarin.Forms.Core/BindableObjectExtensions.cs
```diff
@@ -1,10 +1,37 @@
...
+		}
+
+		internal static void PropagateBindingContext<T>(this BindableObject self, IList<T> children, Action<BindableObject, object> setChildBindingContext)
+		{
+			var gotBindingContext = false;
+			object bc = null;
+
+			for (var i = 0; i < children.Count; i++)
+			{
+				var bo = children[i] as BindableObject;
```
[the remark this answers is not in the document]
PureWeen: So it looks like we can't do this kind of pattern matching yet
https://stackoverflow.com/questions/41436399/expression-of-type-t-cannot-be-handled-by-a-pattern-of-type-x

Because children is a generic list

hartez:
> reset after 5 clicks
@PureWeen To avoid further confusion, maybe I need to modify that to pop up a confirmation dialog after 5 clicks asking if you want to reset the app. Can you open an issue and assign it to me? 

Alternatively, we could make it more clicks, or some other secret gesture :)

PureWeen:
@hartez  will do! Overall that feature has been pretty helpful for finding a few life cycle bugs... It just always takes me a few seconds to correlate what's happening :-)

Xamarin.Forms.Core/BindableObjectExtensions.cs
```diff
@@ -1,10 +1,37 @@
...
+					continue;
+
+				if (!gotBindingContext)
+				{
+					bc = self.BindingContext;
+					gotBindingContext = true;
+				}
+
+				setChildBindingContext(bo, bc);
+			}
```
StephaneDelcroix: can we rewrite this monstrosity (not yours) to get rid of the `gotBindingContext` ?

something like

```csharp
{
    if (children == null || children.Count == 0)
        return;
    var bc = self.BindingContext;
    for (var i = 0; i < children.Count; i++)
    {
        var bo = children[i] as BindableObject;
        if (bo == null)
           continue;
        setChildBindingContext(bo, bc);
 }

File: Xamarin.Forms.Core/Span.cs [hunk 3]
```diff
@@ -95,11 +95,18 @@ public double FontSize
 			set { SetValue(FontElement.FontSizeProperty, value); }
 		}
 
-		public double LineHeight {
+		public double LineHeight
```
StephaneDelcroix: someone is trying to make me mad. but that won:t work. I:m strong.

Xamarin.Forms.Core/BindableObjectExtensions.cs
```diff
@@ -1,10 +1,37 @@
...
 namespace Xamarin.Forms
 {
 	public static class BindableObjectExtensions
 	{
+		internal static void PropagateBindingContext<T>(this BindableObject self, IList<T> children)
+		{
+			PropagateBindingContext(self, children, BindableObject.SetInheritedBindingContext);
+		}
+
+		internal static void PropagateBindingContext<T>(this BindableObject self, IList<T> children, Action<BindableObject, object> setChildBindingContext)
```
StephaneDelcroix: can't you add a type constraint so `T` is `BindableObject` ?

PureWeen:
build --uitest

PR #3676, opened 2018-08-28
[Android] Set initial ScrollView Scroll values when RTL

### LINKED ISSUE #3675: [Android] When set to RTL and api > 17 scroll view doesn't update its values on initial load
### Description
When a scrollview on Android is set to RTL it will automatically start scrolled all the way to the right as its natural position.  The effect this has on the SxrollView on the Forms side of things needs to reflect this behavior so it has accurate values.

### Steps to Reproduce
Run Issue3000 
Notice that Scrolled event never fires and X and Y are never set

### Expected Behavior
Scrolled event should fire so cross platform control is aware that ScrollX and SCrollY changed

### Actual Behavior
nothing happens

### Basic Information

- Version with issue: 3.2

PureWeen:
### Description of Change ###
On the initial layout of a scroll view that is set to RTL the ScrollX position needs to be propagated to the cross platform control. Android will just start the Horizontal scroll all the way to the right without through a scrolled event so the value was never propagating to our control

### Issues Resolved ### 

- fixes #3675 

### Platforms Affected ### 

- Android

When set to RTL the ScrollChanged event will now fire once on load to indicate its non zero X position. This brings it in line when running below API 17 as well

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/ScrollViewRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ScrollViewRenderer.cs
@@ -252,8 +252,13 @@ namespace Xamarin.Forms.Platform.Android
 				_container?.RequestLayout();
 
 			// if the target sdk >= 17 then setting the LayoutDirection on the scroll view natively takes care of the scroll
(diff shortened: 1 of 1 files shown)


PureWeen:
build --uitests

PR #3678, opened 2018-08-29
[Xaml] Fixes including ResourceDictionary from another ResourceDictionary

### LINKED ISSUE #3655: Crash when trying to include resource dictionary from another resource dictionary
### Description

When trying to include resource dictionary from another resource dictionary, exception occurs:
_Xamarin.Forms.Xaml.XamlParseException: Position 6:25. Resource 'GlobalResources.xaml' not found._

Included test app that reproduces issues has following structure:

Resources/GlobalResources.xaml
Resources/LabelStyles.xaml <- includes GlobalResources.xaml  
MainPage.xaml <- includes LabelStyles.xaml

**Workaround**:
Strangely, issue doesn't occur when xml declaration:

`<?xaml-comp compile="true" ?>
`
is included at top of dictionary file that references another dictionary.
_[assembly: XamlCompilation(XamlCompilationOptions.Compile)]_ does not seem to affect bug occurence.

### Steps to Reproduce
1. Include two pure XAML resource dictionaries to forms project
2. Include top dictionary in page or other XAML file

### Expected Behavior
App should not crash
### Actual Behavior
App crashes with resource not found exception

### Basic Information

- Version with issue: 3.1.0.697729
- IDE: Visual Studio 17 15.8
- Platform Target Frameworks: 
  - Android: 8.1
- Affected Devices: All

### Screenshots

image

### Reproduction Link

[DictLink.zip](https://github.com/xamarin/Xamarin.Forms/files/2319366/DictLink.zip)

paymicro:
### Description of Change ###

When you add a resource to an assembly, its key is specified as a relative path from that assembly. So resources may not be found if their path was relative to another resource.
This fix adds a search for a resource in the folder of the parent resource.

### Issues Resolved ### 

- fixes #3655 

### API Changes ###
/

### Platforms Affected ### 

- XAML (all platforms)

Not applicable

### PR Checklist ###

- [ ] Has automated tests 


tscislowicz:
Thanks!

toomasz:
Thanks!

PureWeen:
Build failing 
```
Xamarin.Forms.Platform.Android\AppCompat\PageExtensions.cs(11,17): Error CS0618: 'Fragment' is obsolete: 'This class is obsoleted in this android platform'
Xamarin.Forms.Platform.Android\AppCompat\PageExtensions.cs(53,28): Error CS0618: 'Fragment' is obsolete: 'This class is obsoleted in this android platform'
```
rebasing should fix

Xamarin.Forms.Xaml.UnitTests/ResourceDictionaryWithSource.xaml.cs
```diff
@@ -50,6 +50,8 @@ public void RelativeAndAbsoluteURI(bool useCompiledXaml)
 				Assert.That(((ResourceDictionary)layout.Resources["shortURI"])["sharedfoo"], Is.TypeOf<Style>());
 				Assert.That(((ResourceDictionary)layout.Resources["Colors"])["MediumGrayTextColor"], Is.TypeOf<Color>());
 				Assert.That(((ResourceDictionary)layout.Resources["CompiledColors"])["MediumGrayTextColor"], Is.TypeOf<Color>());
+				Assert.That(((ResourceDictionary)layout.Resources["BaseStyle"])["BaseLabelStyle"], Is.TypeOf<Style>());
+				Assert.That(((ResourceDictionary)layout.Resources["InheritedStyle"])[typeof(Label).FullName], Is.TypeOf<Style>());
```
mattleibow: Do we have a test with multiple layers like this? Base -> Inherited -> Overrides? It probably will work, but a test to confirm and prevent future breaks will be great.

Xamarin.Forms.Core/ResourceDictionary.cs
```diff
@@ -68,7 +71,11 @@ public void SetAndLoadSource(Uri value, string resourcePath, Assembly assembly,
 			if (type != null)
 				_mergedInstance = s_instances.GetValue(type, (key) => (ResourceDictionary)Activator.CreateInstance(key));
 			else
-				_mergedInstance = DependencyService.Get<IResourcesLoader>().CreateFromResource<ResourceDictionary>(resourcePath, assembly, lineInfo);
+			{
+				_assembly = assembly;
+				_path = System.IO.Path.GetDirectoryName(resourcePath);
```
mattleibow: Looking at this, "path" is not entirely correct since it is the root - maybe resourceRoot or directoryPath - something to indicate that is is not the path (which may be a file) but a directory.

Xamarin.Forms.Xaml/ResourcesLoader.cs
```diff
@@ -9,33 +8,18 @@ namespace Xamarin.Forms.Xaml
...
-
-			using (var stream = assembly.GetManifestResourceStream(resourceId)) {
-				if (stream == null)
-					throw new XamlParseException($"No resource found for '{resourceId}'.", lineInfo);
-				using (var reader = new StreamReader(stream)) {
-					var rd = new T();
-					rd.LoadFromXaml(reader.ReadToEnd());
-					return rd;
-				}
-			}
```
mattleibow: Nice! Like the refactor.

Xamarin.Forms.Core/Xaml/IResourcesLoader.cs
```diff
@@ -7,7 +7,7 @@ namespace Xamarin.Forms
 {
 	interface IResourcesLoader
 	{
-		T CreateFromResource<T>(string resourcePath, Assembly assembly, IXmlLineInfo lineInfo) where T : new();
+		T CreateFromResource<T>(string resourcePath, Assembly assembly, IXmlLineInfo lineInfo, ResourceDictionary parentRD) where T : new();
```
mattleibow: Looking at this particular member, I think it is time to rename to a more "resource dictionary" name - maybe `CreateResourceDictionary` or `LoadResourceDictionary`. Right now, we are just using it to load a dictionary - and we have code in there just for dictionaries. If we rename and go full in, we can remove the generics and the casts:

```csharp
ResourceDictionary CreateResourceDictionary(string resourcePath, Assembly assembly, IXmlLineInfo lineInfo, ResourceDictionary parent);
```

```csharp
public ResourceDictionary CreateResourceDictionary(string resourcePath, Assembly assembly, IXmlLineInfo lineInfo, ResourceDictionary parent) {
    var dict = new ResourceDictionary();
    dict.Parent = parent;
    return dict.LoadFromXaml(GetResource(resourcePath, assembly, lineInfo));
}
```

mattleibow: Another alternative is to add a _new_ member that is dictionary specific, and keep the old one:

```csharp
public T CreateFromResource <T>(string resourcePath, Assembly assembly, IXmlLineInfo lineInfo) {
    var res = new T();
    return res.LoadFromXaml(GetResource(resourcePath, assembly, lineInfo));
}
```

mattleibow: I third alternative is to go with option 2 in this file (keep generic), but do the `ResourceDictionary` specific bits in the `ResourceDictionary` type (instead of calling this method to load, just do it there). This is also a good option to be consistent with what we do with the css loading: https://github.com/xamarin/Xamarin.Forms/blob/f2948f1f0cf9829fe6ddafc60df2ac06a6a59ed0/Xamarin.Forms.Xaml/MarkupExtensions/StyleSheetExtension.cs#L35-L36

mattleibow: Of these 3 options - I vote 3 as I see no value with having a generic method that is used in only 1 place for a specific purpose in an unrelated file. The resource loader should be responsible for loading the actual files - the dictionary should be responsible for "parsing" those files (same with the css)

jsuarezruiz:
@paymicro Can you rebase and check the last comments?. Thanks!.

samhouts:
I'm going to close this one for now until we're ready to return to it. Thanks, @paymicro !

PR #3690, opened 2018-08-31
[Core] Added TextTransform property

### LINKED ISSUE #3667: [Enhancement] Add text-transforms to Label
## Summary
Our design-people came up with a design that includes headers that are all uppercase. Right now, we use a converter for this or included it in a markup extension for values that we localize. I think it would be great to have a couple of text transforms (a la CSS) in Forms available by default.

## API Changes

It makes the most sense to implement this on the `Label` control. A new property would be introduced taking in a new enum. You can see them underneath:

- `Label.TextTransform` value of `TextTransform` enum
- `TextTransform` enum: `None`, `Uppercase`, `Lowercase`, `Title`, `None` will be the default to be backwards compatible.

## Intended Use Case
As mentioned already in the summary; you might encouter use-cases where you want to implement text with a certain casing. While this is solvable by entering a string uppercase, writing your own extension, etc. In my opinion this is purely a styling thing and belongs to a property on the control so you don't have to think about this anymore.

paymicro:
### Description of Change ###

Added TextTransform property to the Label

### Issues Resolved ### 

- fixes #3667

### API Changes ###

Added:
 - TextTransform InputView.TextTransform { get; set; } //Bindable Property

```cs
Internals.TextTransform.Uppercase
Internals.TextTransform.Lowercase
Internals.TextTransform.None
```

### Platforms Affected ### 

- Core (all platforms)

Not applicable

### Testing Procedure ###

- run Issue 3667
- click on button "Change TextTransform"
- check result

### PR Checklist ###

- [x] Has automated tests 


Xamarin.Forms.Core/Label.cs
```diff
@@ -127,10 +164,43 @@ public LineBreakMode LineBreakMode
...
+				{
+					sb.Append(char.ToUpperInvariant(c));
+					beginningOfWord = false;
+					continue;
+				}
+				sb.Append(char.ToLowerInvariant(c));
+			}
+			return sb.ToString();
+#else
+			return System.Globalization.CultureInfo.CurrentCulture.TextInfo.ToTitleCase(input.ToLower());
```
StephaneDelcroix: TitleCase is more complex than this, some words shouldn't be capitalized.

StephaneDelcroix: TitleCase is valid for english texts. Different culture capitalize their texts differently

jfversluis:
How are we doing on this? Still feel it would be a valuable addition. @paymicro can you still do it? Else I might take a look :)

paymicro:
@rmarinho on Android text changes on one way... It may be wrong to lose original text.

rmarinho:
@paymicro can you rebase please?

rmarinho:
@paymicro I wonder if we could while we are doing this add a `virtual` method called UpdateFormsText on the Renderers or something, so users could more easily override this behaviour if they want since we are messing with that part of the code.

jsuarezruiz:
@paymicro Can you rebase please?.

jfversluis:
@jsuarezruiz @paymicro I have taken the liberty to do some work on this. Rebased and made it work together with the character spacing stuff!

Please double-check to see if all went well

samhouts:
@rmarinho Issue3667.Issue3667Tests is not passing.

davidortinau:
What does this do? I see @StephaneDelcroix suggesting not to handle TitleCase, but the original description includes no indication of what transforms are supported.

It's the casing transform I'm interesting in, on a button. #5516

jsuarezruiz:
droid3690

rmarinho:
@jsuarezruiz i fixed that issue on Android , can you double check ?! 
THanks

Mikilll94:
In which version of XF this feature is available?

jfversluis:
@Mikilll94 this was merged into master so that would be 4.8

davidortinau:
This doesn't work when using FormattedString or on Span. Tested on Android API29

PR #3700, opened 2018-09-04
[Core] Fixes conversion binding a DateTime to a string if the locale is not the US

### LINKED ISSUE #2049: Bound DateTimes Display in en-US Format Not Local Format
### Description
When you display a bound DateTime on a control like Label, it will always display the DateTime in en-US format. For users outside the USA this is a problem.

Original bug report:
https://bugzilla.xamarin.com/show_bug.cgi?id=58635

**Notes about other platforms**
On Silverlight, UWP, and WPF all controls inherit from FrameworkElement and FrameworkElement has the "Language" property. This is what drives DateTime formatting on these platforms. On Silverlight and WPF there is a bug where the Language property does not default to the local Language. So, by default WPF, and Silverlight have the same bug as Xamarin Forms. But, this can be solved with one line of code:

            Language = XmlLanguage.GetLanguage(Thread.CurrentThread.CurrentCulture.Name);

Here is a WPF sample app that shows this:
https://www.dropbox.com/s/2qcacomtsjweb1o/DateTimeWPF.7z?dl=0

But, on UWP, the Language property correctly defaults to the machine's local language. So, DateTime conversion shows up correctly in UWP by default.

So, Xamarin Forms probably needs a Language property, and this Language property needs to default from the machine's regional settings.

### Steps to Reproduce
1. Switch the culture of your device and Date format to en-AU (day/month/year)
2. Clone this repo: https://ChristianFindlay@bitbucket.org/ChristianFindlay/xamarin-forms-scratch.git
3. Run the UWP, or Android sample, and click "Date Display"
4. Notice that the first date is displayed as "31/1/2000 12:00:00 AM" (This is manually converted from DateTime to string with ToString().
5. Notice that all other displayed dates which are converted using XF's core conversion code are displayed in en-US format as "01/31/2000 00:00:00"
6. I.e. ToString() is giving a different value to the displayed binding. The bug is that binding is not calling ToString() to convert from DateTime to string. It is doing something different and therefore not correctly formatting the date

### Expected Behavior
When a DateTime is converted to a string in binding, it should be converted with a vanilla ToString() which will use the machine's local language settings.

### Actual Behavior
DateTimes are converted to USA format strings regardless of the regional settings.

### Basic Information
- Platform Target Frameworks: 
  - UWP:  16299 
- Nuget Packages:
- Affected Devices:

### Reproduction Link
https://ChristianFindlay@bitbucket.org/ChristianFindlay/xamarin-forms-scratch.git

paymicro:
### Description of Change ###

Fixed converting the date without regard to the installed locale.
You can install the locale using the following code:
```C#
Thread.CurrentThread.CurrentCulture = new System.Globalization.CultureInfo("eu-AU");
```

### Issues Resolved ### 

- fixes #2049

### API Changes ###
 
 None

### Platforms Affected ### 

- Core (all platforms)

Simple types correctly converts to a string from binding context.

**Before**
screenshot_3

**After**
screenshot_2

### Testing Procedure ###
- run UItest 2049
- click on button "Change locale"
- check result
- click on button "Change locale" ...

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2049.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2049.cs
@@ -20,13 +20,30 @@ namespace Xamarin.Forms.Controls.Issues
 		{
 			public DateTime TheDate { get; set; }
 			public float FloatValue { get; set; }
+
+			decimal _decimal;
+			public decimal Decimal
+			{
+				get
+				{
+					return _decimal;
+				}
+				set
+				{
+					_decimal = value;
+					DecimalChange?.Invoke(value);
+				}
+			}
+
+			public static Action<decimal> DecimalChange;
 		}
 
 		DateTime testDate = DateTime.ParseExact("2077-12-31T13:55:56", "yyyy-MM-ddTHH:mm:ss", CultureInfo.InvariantCulture);
 		int _localeIndex = 0;
 		string[] _localeIds = new[] { "en-US", "ru-RU", "en-AU", "zh-Hans" };
 		string _instuctions = $"When you change the locale, the date format must change.{Environment.NewLine}Current locale: ";
-		Label descLabel = new Label(); 
+		Label descLabel = new Label();
+		Label decimalResult = new Label();
 
 		protected override void Init()
 		{
@@ -34,9 +51,16 @@ namespace Xamarin.Forms.Controls.Issues
 
 			var label = new Label();
 			label.SetBinding(Label.TextProperty, nameof(Model.TheDate));
-
 			var labelFloat = new Label();
-			labelFloat.SetBinding(Label.TextProperty, nameof(Model.FloatValue));
+			labelFloat.SetBinding(Label.TextProperty, nameof(Model.Decimal));
+			var entry = new Entry
+			{
+				Keyboard = Keyboard.Numeric
+			};
+			entry.SetBinding(Entry.TextProperty, nameof(Model.Decimal));
+			Model.DecimalChange = (v) => {
+				decimalResult.Text = v.ToString(CultureInfo.CurrentCulture);
+			};
 
 			UpdateDescLabel();
 
@@ -50,7 +74,9 @@ namespace Xamarin.Forms.Controls.Issues
 						Command = new Command(() => ChangeLocale())
 					},
 					label,
-					labelFloat
+					labelFloat,
+					entry,
+					decimalResult
 				}
 			};
 
@@ -62,7 +88,8 @@ namespace Xamarin.Forms.Controls.Issues
 			BindingContext = new Model
 			{
 				TheDate = DateTime.ParseExact("2077-12-31T13:55:56", "yyyy-MM-ddTHH:mm:ss", CultureInfo.InvariantCulture),
-				FloatValue = 123.456f // in some locales, the decimal symbol may not be a dot.
+				FloatValue = 123.456f, // in some locales, the decimal symbol may not be a dot.
+				Decimal = 456.789m
 			};
 		}
 
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2049.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2049.cs
@@ -15,32 +15,9 @@ namespace Xamarin.Forms.Controls.Issues
 	[Issue(IssueTracker.Github, 2049, "Bound DateTimes Display in en-US Format Not Local Format", PlatformAffected.All)]
 	public class Issue2049 : TestContentPage
 	{
-		[Preserve(AllMembers = true)]
-		public class Model
-		{
-			public DateTime TheDate { get; set; }
-			public float FloatValue { get; set; }
-
-			decimal _decimal;
-			public decimal Decimal
-			{
-				get
-				{
-					return _decimal;
-				}
-				set
-				{
-					_decimal = value;
-					DecimalChange?.Invoke(value);
-				}
-			}
-
-			public static Action<decimal> DecimalChange;
-		}
-
-		DateTime testDate = DateTime.ParseExact("2077-12-31T13:55:56", "yyyy-MM-ddTHH:mm:ss", CultureInfo.InvariantCulture);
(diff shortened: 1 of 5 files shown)


jassmith:
build --uitests

StephaneDelcroix:
I will look at this, but I think it's not a matter of using the current locale while converting (there are good reasons why we do not) and more a matter of string formatting.

jassmith:
build --uitests

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2049.cs
```diff
@@ -0,0 +1,123 @@
...
+				{
+					_decimal = value;
+					DecimalChange?.Invoke(value);
+				}
+			}
+
+			public static Action<decimal> DecimalChange;
+		}
+
+		void UpdateDescLabel() => descLabel.Text = $"{_instuctions}{Thread.CurrentThread.CurrentCulture.DisplayName}";
```
samhouts: When I run this on Android, I don't see the current culture on the page.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2049.cs
```diff
@@ -0,0 +1,123 @@
...
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 2049, "Bound DateTimes Display in en-US Format Not Local Format", PlatformAffected.All)]
+	public class Issue2049 : TestContentPage
+	{
+		DateTime _testDate = DateTime.ParseExact("2077-12-31T13:55:56", "yyyy-MM-ddTHH:mm:ss", CultureInfo.InvariantCulture);
+		int _localeIndex = 0;
+		string[] _localeIds = new[] { "en-US", "pt-PT", "ru-RU", "en-AU", "zh-Hans" };
+		string _instuctions = $"When you change the locale, the date format must change.{Environment.NewLine}Current locale: ";
```
samhouts: _instuctions >> _instructions

Xamarin.Forms.Core.UnitTests/BindingUnitTests.cs
```diff
@@ -1897,6 +1897,11 @@ public void ConvertIsCultureInvariant (string culture)
 			slider.Value = 0.9;
 
 			Assert.That (vm.Text, Is.EqualTo ("0.9"));
+
+			// people write numbers using the separator accepted in their locale
+			var ds = System.Threading.Thread.CurrentThread.CurrentCulture.NumberFormat.NumberDecimalSeparator;
+			vm.Text = $"0{ds}7";
+			Assert.That(slider.Value, Is.EqualTo(0.7));
```
kingces95: This test explicitly expects the invariant culture is used when displaying the slider value. This would lead me to believe the original developer thought about Culture and decided to go with the invariant culture everywhere. So this PR is less of a "fix" and more of a feature request; The current behavior is "expected" from the perspective of the original developer. As a feature request, the work item would be to review _all_ renderers and re-establish a new expected behavior wrt culture -- which is to say that it should respect whatever culture the main thread is using.

Xamarin.Forms.Core/BindableProperty.cs
```diff
@@ -317,7 +318,7 @@ internal bool TryConvert(ref object value)
 			TypeConverter typeConverterTo;
 			if (SimpleConvertTypes.TryGetValue(valueType, out convertableTo) && Array.IndexOf(convertableTo, type) != -1)
 			{
-				value = Convert.ChangeType(value, type);
+				value = Convert.ChangeType(value, type, System.Globalization.CultureInfo.CurrentCulture);
```
kingces95: With any luck the feature "Use UI Thread Culture When Preforming Conversions" would only require changes in this function. Who knows, maybe this is the only spot! However, right below, we have a `ConvertFromInvariantString` so we'd have to go and understand if that also needs to be updated to take the UI culture.

Xamarin.Forms.Core/BindingExpression.cs
```diff
@@ -423,19 +423,28 @@ bool TryConvert(BindingExpressionPart part, ref object value, Type convertTo, bo
...
-					throw new FormatException();
 
-				// do not canonicalize "-0"; user will likely enter a period after "-0"
-				if (stringValue == "-0" && DecimalTypes.Contains(convertTo))
-					throw new FormatException();
+				if (DecimalTypes.Contains(convertTo))
+				{
+					// see: https://bugzilla.xamarin.com/show_bug.cgi?id=32871
+					// do not canonicalize "*.[.]"; "1." should not update bound BindableProperty
+					if (stringValue.EndsWith("."))
```
kingces95: This logic should probably be updated to test if the string value ends with the `CurrentCulture` decimal separator instead of the hard coded dot. Even if we stick with invariant culture we should replace the hardcoded dot with the `NumberDecimalSeparator`.

Xamarin.Forms.Core/BindingExpression.cs
```diff
@@ -423,19 +423,28 @@ bool TryConvert(BindingExpressionPart part, ref object value, Type convertTo, bo
...
+						throw new FormatException();
+
+					// do not canonicalize "-0"; user will likely enter a period after "-0"
+					if (stringValue == "-0")
+						throw new FormatException();
+
+					// to change locale if stringValue contains decimal separator for that locale
+					if (stringValue.IndexOf(CultureInfo.CurrentCulture.NumberFormat.NumberDecimalSeparator) != -1)
```
kingces95: This should be removed; We shouldn't infer a culture by testing for culture specific values in the string. If we decide to allow non-invariant culture then we'll simply get the culture from the UI thread.

jassmith:
I agree with Chris here as well, we should be doing conversions with CurrentCulture. I haven't given this a full review but I thought this was supposed to simply make conversions work with comma cultures when one is active.

paymicro:
I leave old "expected" behavior by default, nothing to break. To use current locale for a specific property, you also must enable it. Ex:. `Label.TextProperty.UseCurrentCulture = true`.

StephaneDelcroix:
@paymicro Thanks for this feature. We appreciate the amount of time you spent on it, and understand your will and energy for making Xamarin.Forms easier to use for everyone.

Unfortunately, we need more time to think about this and evaluate all the consequences of those changes, and a pull-request is not the right place to discuss those.

So, I'm closing this PR. It's not a `no`, it's a `not now`.

Feel free to write a spec for your proposed change and we'll happily discuss it in an issue.

PR #3701, opened 2018-09-04
[Android] Fixed TimePicker to support 24H mode

SR84:
The 24h mode for the pickerdialog was not working.

Version 3.1 was working fine. 

Here is an updated file.

### Platforms Affected ### 
- Android


jassmith:
Something went very wrong with API23 Legacy run... I dont think its related to this PR however. Merging

PR #3714, opened 2018-09-05
[iOS] Fix NRE on ContextCell

### LINKED ISSUE #3275: [iOS] For ListView in Recycle mode ScrollTo causes cell leak and in some cases NRE
### Description
When ListView uses Recycle mode and cells have MenuItem and a framework like Prism.Forms is used then ListView.ScrollTo causes NRE.

Root cause - ListView.ScrollTo leaks cells in Recycle mode. Leaked cells with MenuItem still listen to OnMenuItemPropertyChanged and after navigating away from the page Prism.Forms sets BindingContext to null which causes NRE in ContextActionsCell.SetupSelection as GestureRecognizers is already null.

**Another way to get leaked cells is to scroll the list a bit then rotate device and scroll again, go back - crash.**

### Steps to Reproduce

Run App2.iOS project from this solution (can be tricky to build, but possible):
https://github.com/VitalyKnyazev/Leaking-ListView

1. Navigate to Transaction page
2. Click 'Scroll to'
3. Navigate back

### Expected Behavior
No cells are leaked, like when for example UITableView.ScrollRectToVisible is used to implement ListView.ScrollTo instead of UITableView.ScrollToRow.

### Actual Behavior
Recycled cells are leaked when UITableView.ScrollToRow is used, looks like a bug in iOS.

### Basic Information

- IDE: Visual Studio 2017 15.7.5
- Platform Target Frameworks: 
  - iOS:  11.4.1

- Nuget Packages: Xamarin.Forms 3.1.0.583944
- Affected Devices: iPhone/iPad

### Screenshots

N/A

### Reproduction Link

https://github.com/VitalyKnyazev/Leaking-ListView

rmarinho:
### Description of Change ###

Fix a NRE when ContextCell is being disposed

### Issues Resolved ### 
- fixes #3275 

### API Changes ###
 None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###

Go to test 3275 and follow instructions 
 
### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.iOS/ContextActionCell.cs
+++ b/Xamarin.Forms.Platform.iOS/ContextActionCell.cs
@@ -632,7 +632,7 @@ namespace Xamarin.Forms.Platform.iOS
 
 		void SetupSelection(UITableView table)
 		{
-			for (var i = 0; i < table.GestureRecognizers.Length; i++)
+			for (var i = 0; i < table.GestureRecognizers?.Length; i++)
 			{
 				var r = table.GestureRecognizers[i] as SelectGestureRecognizer;
 				if (r != null)
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3275.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3275.cs
@@ -1,0 +1,233 @@
+﻿using System;
+using System.Collections.ObjectModel;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+using System.Windows.Input;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.UITest;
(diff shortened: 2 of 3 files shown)


Xamarin.Forms.Platform.iOS/ContextActionCell.cs
```diff
@@ -632,7 +632,7 @@ UIView SetupButtons(nfloat width, nfloat height)
 
 		void SetupSelection(UITableView table)
 		{
-			for (var i = 0; i < table.GestureRecognizers.Length; i++)
+			for (var i = 0; i < table.GestureRecognizers?.Length; i++)
```
StephaneDelcroix: if this fixes the issue, I'm fine.

but please move the null check out of the `for` loop. This adds complexity to the inner loop that is executed every time

rmarinho:
@PureWeen thanks, my investigation shows tha some Cells go away remvoed by the platform but Dispose is not called so we must call that method to unhook when we are removed from UITableView

rmarinho:
@PureWeen  @StephaneDelcroix can you review again thanks

PR #3751, opened 2018-09-09
[Android Maps] Make markers available to subclasses

LRP-sgravel:
### Description of Change ###
Add method to retrieve native Marker from a pin in Android MapRenderer.  No tests added as it is a simple Extract Method refactoring.

### Issues Resolved ### 
Eases subclassing of MapRenderer to extend pins.  Needed to build a Skia Maps extension Nuget.

### API Changes ###

Added:
protected Marker GetMarkerForPin(Pin pin)

Changed:
N/A
 
 Removed:
N/A
 
### Platforms Affected ### 
- Android

None

N/A

### Testing Procedure ###
N/A

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Maps.Android/MapRenderer.cs
+++ b/Xamarin.Forms.Maps.Android/MapRenderer.cs
@@ -250,7 +250,7 @@ namespace Xamarin.Forms.Maps.Android
 		void PinOnPropertyChanged(object sender, PropertyChangedEventArgs e)
 		{
 			Pin pin = (Pin)sender;
-			Marker marker = _markers.FirstOrDefault(m => m.Id == (string)pin.Id);
+			Marker marker = GetMarkerForPin(pin);
 
 			if (marker == null)
 			{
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Maps.Android/MapRenderer.cs
```diff
@@ -271,6 +271,11 @@ void PinOnPropertyChanged(object sender, PropertyChangedEventArgs e)
 			}
 		}
 
+		protected Marker GetMarkerForPin(Pin pin)
+		{
+			return _markers?.FirstOrDefault(m => m.Id == (string)pin.Id);
```
samhouts: While we're in here, we should remove this usage of `FirstOrDefault` and convert it to a `.Find`. It should be a bit better for performance.

sylvain-toptal: Cool, made me look up the SO thread on performance between both methods!

Done

PR #3762, opened 2018-09-11
WPF - Fix Unable to intercept back button

### LINKED ISSUE #3689: WPF: Unable to intercept back button
### Description

We cannot find any possibility to intercept the back button in a Xamarin Forms WPF project.
I.e. the NavigationPage.OnBackButtonPressed is only called on Android and UWP, but not in WPF.
We have also tried to intercept the back button in the WPF project natively, which is also never called:
System.Windows.Application.Current.Navigating += new NavigatingCancelEventHandler(OnNavigating);

### Steps to Reproduce

1. Add a new WPF Project.
2. Create a page with abutton which pushes another page. Try to intercept the top left back button (back menu) on the second page when going back to the first page.

### Expected Behavior
One should be able to intercept the back button (ind interupting the page popping, if needed) before popping the page. This works in UWP and Android.

### Actual Behavior
No interception available.

### Basic Information

- Version with issue: Xamarin.Forms.Platform.WPF 3.2.0.729530-pre2
- Last known good version:
- IDE: VS 2017 15.8.2
- Platform Target Frameworks: WPF
- Affected Devices: Windows 10

mohachouch:
### Description of Change ###

 Fix  : Unable to intercept back button.

### Issues Resolved ### 

- fixes #3689

### Platforms Affected ### 

- WPF

### DIFF
--- a/Xamarin.Forms.Platform.WPF/Renderers/NavigationPageRenderer.cs
+++ b/Xamarin.Forms.Platform.WPF/Renderers/NavigationPageRenderer.cs
@@ -16,7 +16,10 @@ namespace Xamarin.Forms.Platform.WPF
 
 		public override void OnBackButtonPressed()
 		{
-			if (NavigationPage.CurrentPage?.SendBackButtonPressed() == false)
+			if (NavigationPage?.CurrentPage == null)
+				return;
+
+			if (!NavigationPage.CurrentPage.SendBackButtonPressed())
 				NavigationPage.PopAsync();
 		}
 	}
--- a/Xamarin.Forms.Platform.WPF/Renderers/NavigationPageRenderer.cs
+++ b/Xamarin.Forms.Platform.WPF/Renderers/NavigationPageRenderer.cs
@@ -16,7 +16,8 @@ namespace Xamarin.Forms.Platform.WPF
 
 		public override void OnBackButtonPressed()
 		{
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.WPF/Renderers/NavigationPageRenderer.cs
```diff
@@ -16,7 +16,8 @@ public FormsLightNavigationPage(NavigationPage navigationPage)
 
 		public override void OnBackButtonPressed()
 		{
-			NavigationPage.PopAsync();
+			if (NavigationPage.CurrentPage?.SendBackButtonPressed() == false)
```
rmarinho: !NavigationPage.CurrentPage?.SendBackButtonPressed() ?

mohachouch: I think we cannot use (!) when there is a nullable check (CurrentPage?)

mohachouch: You're right it's not very clean. It would be better to do this:

if(Navigation?.CurrentPage == null)
return ;

if (!NavigationPage.CurrentPage.SendBackButtonPressed())
NavigationPage.PopAsync();

I can modify it like that

PR #3780, opened 2018-09-13
[Android] Fix WebView.Navigating event, make WebViewClient inheritable

### LINKED ISSUE #3778: [Android] WebView Navigating event not always fired
### Description

In the Android WebViewRenderer, the Navigating event is fired from the `ShouldOverrideUrlLoading(AWebView view, string url)` method in the WebClient, which seems to get hit when the original URL is redirected or navigation occurs from the initial URL, but not when the original URL begins loading. (This method was also deprecated in API level 24 and replaced with `ShouldOverrideUrlLoading(AWebView webView, WebResourceRequest request)` .)

The WebClient should instead override the `OnPageStarted` method and fire the Navigating event there, for consistency with firing the Navigated event from `OnPageFinished`. This method is hit for initial loads, redirects, and subsequent navigations.

I also believe that WebClient should be moved from a private inner class of the WebViewRenderer to a public class, so that custom renderers can subclass it to inject custom WebViewClients without having to re-implement firing of all the WebView events.

https://forums.xamarin.com/discussion/138821/navigating-on-android-webview-doesnt-always-fire

### Steps to Reproduce

1. Add a WebView to a page in a Xamarin.Forms project
2. Register for the Navigating event of the WebView
3. Set the source to a UrlWebViewSource that will not redirect (e.g. https://www.google.com)
4. Run on an Android device

[WebViewAndroid.zip](https://github.com/xamarin/Xamarin.Forms/files/2380442/WebViewAndroid.zip)

### Expected Behavior

The Navigating event will be fired when the WebView begins loading google.com.

### Actual Behavior

The Navigating event is not fired.

### Basic Information

- Affected Devices: Android

jcmanke:
### Description of Change ###

- Extracted WebViewRenderer.WebClient to FormsWebViewClient so that it can be inherited from and overwritten when custom renderers require a custom WebViewClient
- Moved firing of WebView.Navigating event from ShouldOverrideUrlLoading to OnPageStarted in FormsWebViewClient so that it is consistently fired every time a URL is navigated to including initial load

### Issues Resolved ### 

- fixes #3778 

### API Changes ###

Added:
 - public override void FormsWebViewClient.OnPageStarted(WebView view, string url, Bitmap favicon)
 - public const string WebViewRenderer.AssetBaseUrl = "file:///android_asset/"
 - protected virtual FormsWebViewClient WebViewRenderer.GetFormsWebViewClient()
 - FormsWebViewClient WebViewRenderer._webViewClient 

Changed:
 - class WebViewRenderer.WebClient => public class FormsWebViewClient
 - bool WebViewRenderer._ignoreSourceChanges => public bool WebViewRenderer.IgnoreSourceChanges { get; set; }
 - IWebViewController WebViewRenderer.ElementController => public IWebViewController WebViewRenderer.ElementController
 - void WebViewRenderer.UpdateCanGoBackForward() => public void WebViewRenderer.UpdateCanGoBackForward()  
 
 Removed:
 - WebClient.ShouldOverrideUrlLoading(WebView view, string url)

### Platforms Affected ### 

- Android

WebView.Navigating will be fired for the initial URL load in addition to redirects and subsequent navigations.

Not applicable

### Testing Procedure ###

 - In the WebViewCoreGalleryPage, add a listener to the element's Navigating event in InitializeElement(). 
 - Change the source URL from http://xamarin.com/ (which gets redirected to https://visualstudio.microsoft.com/xamarin/) to https://www.google.com (which will not be redirected). 
- Run the ControlGallery and verify that the Navigating event is fired for https://www.google.com

### PR Checklist ###

- [ ] Has automated tests

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/WebViewRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/WebViewRenderer.cs
@@ -17,7 +17,7 @@ namespace Xamarin.Forms.Platform.Android
 	{
 		public const string AssetBaseUrl = "file:///android_asset/";
 
-		FormsWebViewClient _webViewClient;
+		WebViewClient _webViewClient;
 		FormsWebChromeClient _webChromeClient;
 
 		public IWebViewController ElementController => Element;
@@ -64,7 +64,7 @@ namespace Xamarin.Forms.Platform.Android
 			base.Dispose(disposing);
 		}
 
-		protected virtual FormsWebViewClient GetFormsWebViewClient()
+		protected virtual WebViewClient GetWebViewClient()
 		{
 			return new FormsWebViewClient(this);
 		}
@@ -95,7 +95,7 @@ namespace Xamarin.Forms.Platform.Android
 				webView.LayoutParameters = new global::Android.Widget.AbsoluteLayout.LayoutParams(LayoutParams.MatchParent, LayoutParams.MatchParent, 0, 0);
 #pragma warning restore 618
 
-				_webViewClient = GetFormsWebViewClient();
+				_webViewClient = GetWebViewClient();
 				webView.SetWebViewClient(_webViewClient);
 
 				_webChromeClient = GetFormsWebChromeClient();
--- a/Xamarin.Forms.Platform.Android/Renderers/WebViewRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/WebViewRenderer.cs
@@ -20,8 +20,8 @@ namespace Xamarin.Forms.Platform.Android
 		WebViewClient _webViewClient;
 		FormsWebChromeClient _webChromeClient;
 
-		public IWebViewController ElementController => Element;
-		public bool IgnoreSourceChanges { get; set; }
+		protected internal IWebViewController ElementController => Element;
+		protected internal bool IgnoreSourceChanges { get; set; }
 
 		public WebViewRenderer(Context context) : base(context)
 		{
@@ -185,7 +185,7 @@ namespace Xamarin.Forms.Platform.Android
 			UpdateCanGoBackForward();
 		}
 
-		public void UpdateCanGoBackForward()
+		protected internal void UpdateCanGoBackForward()
 		{
 			if (Element == null || Control == null)
 				return;
--- a/Xamarin.Forms.Platform.Android/Renderers/WebViewRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/WebViewRenderer.cs
@@ -15,10 +15,13 @@ namespace Xamarin.Forms.Platform.Android
 {
 	public class WebViewRenderer : ViewRenderer<WebView, AWebView>, IWebViewDelegate
 	{
-		bool _ignoreSourceChanges;
+		public const string AssetBaseUrl = "file:///android_asset/";
+
+		FormsWebViewClient _webViewClient;
 		FormsWebChromeClient _webChromeClient;
 
-		IWebViewController ElementController => Element;
+		public IWebViewController ElementController => Element;
+		public bool IgnoreSourceChanges { get; set; }
 
 		public WebViewRenderer(Context context) : base(context)
 		{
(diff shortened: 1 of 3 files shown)


jcmanke:
@samhouts You want me to rebase this onto the 3.3.0 branch?

Xamarin.Forms.Platform.Android/Renderers/WebViewRenderer.cs
```diff
@@ -175,7 +185,7 @@ void OnGoForwardRequested(object sender, EventArgs eventArgs)
 			UpdateCanGoBackForward();
 		}
 
-		void UpdateCanGoBackForward()
```
jassmith: Seems this should be protected at best.

Xamarin.Forms.Platform.Android/Renderers/WebViewRenderer.cs
```diff
@@ -15,10 +15,13 @@ namespace Xamarin.Forms.Platform.Android
...
 	{
-		bool _ignoreSourceChanges;
+		public const string AssetBaseUrl = "file:///android_asset/";
+
+		FormsWebViewClient _webViewClient;
 		FormsWebChromeClient _webChromeClient;
 
-		IWebViewController ElementController => Element;
+		public IWebViewController ElementController => Element;
+		public bool IgnoreSourceChanges { get; set; }
```
jassmith: protected?

Xamarin.Forms.Platform.Android/Renderers/WebViewRenderer.cs
```diff
@@ -47,19 +50,25 @@ protected override void Dispose(bool disposing)
...
 
+					_webViewClient?.Dispose();
 					_webChromeClient?.Dispose();
 				}
 			}
 
 			base.Dispose(disposing);
 		}
 
+		protected virtual FormsWebViewClient GetFormsWebViewClient()
```
jassmith: I haven't looked but can this return the base class instead of FormsWebViewClient?

Xamarin.Forms.Platform.Android/Renderers/WebViewRenderer.cs
jcmanke: I made these public so they can be accessed by the FormsWebViewClient. Could be protected internal instead.

Xamarin.Forms.Platform.Android/Renderers/WebViewRenderer.cs
jcmanke: I did it this way to be consistent with the FormsWebChromeClient.

jassmith:
@jcmanke please rebase on 3.3.0 :)

Xamarin.Forms.Platform.Android/Renderers/WebViewRenderer.cs
jassmith: I would prefer we return base classes whenever possible, keeps as much flexibility there as we can (so long as we dont have to cast it)

jcmanke: I'm split on this because while the renderer could safely treat the client as the base class, the point of opening it up is so that people can subclass the renderer and client without losing any functionality or having to copy & paste a bunch of code. Enforcing that they provide a subclass of FormsWebViewClient keeps users from shooting themselves in the foot and accidentally losing the implementation of the Navigating and Navigated events.

If we want to give them that opportunity I think GetFormsWebChromeClient() should be changed to return the base class as well.

jcmanke:
@jassmith @samhouts Changed the accessors from public to protected internal.

Xamarin.Forms.Platform.Android/Renderers/WebViewRenderer.cs
chrisseed: @jcmanke You may or may not be interested to know that I have just shot myself in the foot by _not_ deriving my custom ```WebViewClient``` from ```FormsWebViewClient``` but thanks to your comment above I didn't spend too much time figuring out why my Navigating events still weren't firing even after upgrading to the latest version of Xamarin.Forms...

PR #3789, opened 2018-09-14
[Android] Fix dispose of Listview renderer

rmarinho:
### Description of Change ###

When navigating back from issue 3408 on the controll Gallery we were getting a crash because of missing constructor on our renderer. 

This seems to be related with this **https://stackoverflow.com/questions/10593022/monodroid-error-when-calling-constructor-of-custom-view-twodscrollview/10603714#comment59366780_10603714**

By  forcing the dispsoe to happen on the Main thread we might fix the issues related with accessing the same object via otherad threads. 

### Issues Resolved ### 

UITests 3408 navigating back

### API Changes ###

 None

### Platforms Affected ### 

- Android

None

Not applicable

### Testing Procedure ###

Go to test 3408 and navigate back , shouldn't crash 

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3408.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3408.cs
@@ -32,7 +32,10 @@ namespace Xamarin.Forms.Controls.Issues
 
 			var grd = new Grid();
 
-			var aacountListView = new ListView();
+			var aacountListView = new ListView
+			{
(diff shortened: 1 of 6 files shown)


kingces95:
@rmarinho Do you have a guess as to why this object is being 'prematurely disposed' re the [SO you found](https://stackoverflow.com/questions/10593022/monodroid-error-when-calling-constructor-of-custom-view-twodscrollview/10603714#comment59366780_10603714)? 

To test if it's a GC bug we could purposely leak all those objects that are being resurrected by stuffing them into a global array... and to test for double dispose we could add counters.

kingces95:
Rui said it didn't work...

PR #3791, opened 2018-09-14
[UITests] wait for element before querying for it

PureWeen:
### Description of Change ###
When attempting to scroll during UI Tests, occasionally the timing of the scroll and querying for an element don't match up. This attempts to query a few times before failing.

### Issues Resolved ### 
This attempts to resolve random failures with calls to *App.ScrollForElement*

### Platforms Affected ### 
UI TESTS

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core.UITests.Shared/Utilities/Gestures.cs
+++ b/Xamarin.Forms.Core.UITests.Shared/Utilities/Gestures.cs
@@ -7,6 +7,24 @@ namespace Xamarin.Forms.Core.UITests
 {
 	internal static class Gestures
 	{
+		static AppRect GetAppRect(this IApp app, Func<AppQuery, AppQuery> elementQuery)
+		{
+			int tryCount = 0;
+			var elements = app.Query(elementQuery);
+			while (elements.Count() == 0 && tryCount < 10)
+			{
+				elements = app.WaitForElement(elementQuery);
+				tryCount++;
+			}
+
+			if(elements.Count() == 0)
+			{
+				throw new Exception("Failed to Retreive Element from elementQuery");
+			}
+
+			return elements.First().Rect;
+		}
+			
 		public static bool ScrollForElement (this IApp app, string query, Drag drag, int maxSteps = 25)
 		{
 			Func<AppQuery, AppQuery> elementQuery = q => q.Raw (query);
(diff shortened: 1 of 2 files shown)


PureWeen:
build --uitests

Xamarin.Forms.Core.UITests.Shared/Utilities/Gestures.cs
```diff
@@ -7,6 +7,24 @@ namespace Xamarin.Forms.Core.UITests
 {
 	internal static class Gestures
 	{
+		static AppRect GetAppRect(this IApp app, Func<AppQuery, AppQuery> elementQuery)
+		{
+			int tryCount = 0;
+			var elements = app.Query(elementQuery);
+			while (elements.Count() == 0 && tryCount < 10)
```
VladislavAntonyuk: elements.Count() == 0 can be replaced with elemnts.Any()

Xamarin.Forms.Core.UITests.Shared/Utilities/Gestures.cs
```diff
@@ -23,7 +41,7 @@ public static bool ScrollForElement (this IApp app, string query, Drag drag, int
 			// check to see if the element is visible already
 			if (app.Query (elementQuery).Length == 1) {
```
kingces95: We expect that because this `app.Query` returned a result then the second `app.Query` to compute `elementDistanceToDragCenter` should return the same result, but actually it doesn't and the fix is to try `n` more times? If so, did we consider reusing the result of the first `app.Query` to also compute `elementDistanceToDragCenter` instead of calling it a second time?

Xamarin.Forms.Core.UITests.Shared/Utilities/Gestures.cs
```diff
@@ -36,8 +54,15 @@ public static bool ScrollForElement (this IApp app, string query, Drag drag, int
 			}
 
 			if (count != maxSteps) {
+				var elements = app.Query(elementQuery);
+				while (elements.Count() == 0)
+				{
+					app.WaitForElement(elementQuery);
+					elements = app.Query(elementQuery);
+				}
```
kingces95: Might this result in a hang if the count continues to return zero for some reason? Might we consider refactoring `GetAppRect` to `QueryNTimes` so it could be used here in addition to the other callsites?

Xamarin.Forms.Core.UITests.Shared/Utilities/Gestures.cs
```diff
@@ -7,6 +7,24 @@ namespace Xamarin.Forms.Core.UITests
 {
 	internal static class Gestures
 	{
+		static AppRect GetAppRect(this IApp app, Func<AppQuery, AppQuery> elementQuery)
```
kingces95: Might consider refactoring to `QueryNTimes` that takes an additional parameter `count`. Then pull `10` into a const that's passed to at each callsite.

PureWeen:
build --uitests

PureWeen:
@kingces95   good points.  I didn't do a very accurate job of maintaining intended behavior

> Might consider refactoring to QueryNTimes

Done

> We expect that because this app.Query returned a result then the second app.Query to compute elementDistanceToDragCenter should return the same result, but actually it doesn't and the fix is to try n more times? If so, did we consider reusing the result of the first app.Query to also compute elementDistanceToDragCenter instead of calling it a second time?

This is the core of the issue and the latest set of changes should resolve.  The calls to *CenterElementInView* also re-queried for the same element so there were cases where the same thing was queried for 3 times.  The code should now be setup to just reuse the result from the first query

please re-review as well @rmarinho

PR #3815, opened 2018-09-18
[CSS] accept 'oblique' for 'font-style'

### LINKED ISSUE #3808: [CSS] font-style doesn't properly support oblique value
This works:

```
.featureHeader {
    font-style: bold;
}
```

However, this doesn't (in Previewer anyway):
```
.featureHeader {
    font-style: oblique;
}
```

According to the W3C spec, the acceptable value for `font-style` are:

- normal
- italic
- oblique
- initial
- inherit

I would expect to be able to use those values. I don't mind that `bold` is working since that's what XF already expects.

Using 3.2.0-pre3

StephaneDelcroix:
### Description of Change ###

[CSS] accept 'oblique' for 'font-style'

### Issues Resolved ### 

- fixes #3808

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/FontAttributes.cs
+++ b/Xamarin.Forms.Core/FontAttributes.cs
@@ -3,10 +3,39 @@ using System;
 namespace Xamarin.Forms
 {
 	[Flags]
+	[TypeConverter(typeof(FontAttributesConverter))]
 	public enum FontAttributes
 	{
 		None = 0,
 		Bold = 1 << 0,
 		Italic = 1 << 1
+	}
+
+	[Xaml.TypeConversion(typeof(FontAttributes))]
+	public sealed class FontAttributesConverter : TypeConverter
+	{
+
+		public override object ConvertFromInvariantString(string value)
+		{
+			if (string.IsNullOrEmpty(value))
+				return FontAttributes.None;
+
+			FontAttributes attributes = FontAttributes.None;
+			for (int i = 0; i < value.Split(',').Length; i++) {
+				var part = value.Split(',')[i].Trim();
+				if (Enum.TryParse(value, true, out FontAttributes attr)) {
+					attributes |= attr;
+					continue;
+				}
+				if (part.Equals("normal", StringComparison.OrdinalIgnoreCase))
+					continue;
+				if (part.Equals("oblique", StringComparison.OrdinalIgnoreCase)) {
+					attributes |= FontAttributes.Italic;
+					continue;
+				}
+				throw new InvalidOperationException(string.Format("Cannot convert \"{0}\" into {1}", value, typeof(FontAttributes)));
+			}
+			return attributes;
+		}
 	}
 }


Xamarin.Forms.Core/FontAttributes.cs
```diff
@@ -3,10 +3,39 @@
...
+	public sealed class FontAttributesConverter : TypeConverter
+	{
+
+		public override object ConvertFromInvariantString(string value)
+		{
+			if (string.IsNullOrEmpty(value))
+				return FontAttributes.None;
+
+			FontAttributes attributes = FontAttributes.None;
+			for (int i = 0; i < value.Split(',').Length; i++) {
```
jassmith: Lets not perform the split 3 different times unless there is a reason to I dont see?

Xamarin.Forms.Core/FontAttributes.cs
```diff
@@ -3,10 +3,39 @@
...
+				var part = value.Split(',')[i].Trim();
+				if (Enum.TryParse(value, true, out FontAttributes attr)) {
```
PureWeen: should this be *part* not value?

StephaneDelcroix: you caught me in full-on laziness mode

Xamarin.Forms.Core/FontAttributes.cs
StephaneDelcroix: auto-refactoring the `foreach` into a `for` did this

Xamarin.Forms.Core/FontAttributes.cs
```diff
@@ -3,10 +3,48 @@
...
+
+		public override object ConvertFromInvariantString(string value)
+		{
+			if (string.IsNullOrEmpty(value))
+				return FontAttributes.None;
+
+			FontAttributes attributes = FontAttributes.None;
+			value = value.Trim();
+			if (value.Contains(",")) { //Xaml
+				foreach (var part in value.Split(','))
```
PureWeen: Should we still stick with not using foreach? Something like this

```C#
			string[] parts = null;
			if (value.Contains(","))
				parts = value.Split(',');
			else
				parts = value.Split(' ');

			for (int i = 0; i < parts.Length; i++)
				attributes |= ParseSingleAttribute(parts[i], value);
```

jassmith: I dont care in this case because its not a hot path

jassmith: Up to you to decide what you want

PureWeen: I'd prefer it because
-consistency
-I like the one for loop applying that logic opposed to the duplication of the for loop

StephaneDelcroix: hot path or not, `foreach` on an array is equally (well, even slightly more) efficient than a `for` loop... https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgBgATIIwG4CwAUMgMyYBM6AwugN6HoOanIAs6AsgBQCUt9jAgG4BDAE7oADmOABndAF50AIgBmAezXoARmKUA6AMoSIYYJwDk6c93wEBA9aICmwgMYALdJxHiJ6MAB2ktIy3Pz29tgAnJwSNuH2jl4+/vKoOGAAPFKisnoAMk4BAObA7hnw8GF2EZFYMTmyANpgALrxNQwAvoRdQA=

PureWeen:
build --uitests

Xamarin.Forms.Core/FontAttributes.cs
PureWeen: That's fine and not something to hold up the PR but I still feel being consistently consistent is more important. If the code base uses *for* everywhere that's not an IEnumerable that's an indicator to contributors. That's something we can put in a coding guideline.  If we just seemingly randomly have foreach vs for then having any guideline about which to use seems pointless and either is always fine

PR #3848, opened 2018-09-19
Ensure that the ListViewAdapter is no longer being used by the ListVi…

### LINKED ISSUE #3603: [Android] Unable to activate instance of type Xamarin.Forms.Platform.Android.ListViewAdapter when using nested ListViews and DataTemplateSelector
### Description

ListView
-DatatemplateSelector
--Listview

When swapping out the ItemSource on the parent ListView you'll get an exception of
*Unable to activate instance of type Xamarin.Forms.Platform.Android.ListViewAdapter*

https://github.com/xamarin/Xamarin.Forms/blob/4d5a1f5482eeb43661d0ac972262e09c53e61117/Xamarin.Forms.Platform.Android/Renderers/ListViewRenderer.cs#L153

### Steps to Reproduce
1. Run Control Gallery for Issue3408
2. Click Change Source
3. Click Change Property
4. Click Change Source
5. Crash

### Expected Behavior
No crash and ItemSource swaps out fine

### Actual Behavior
Crashes

### Basic Information

- Version with issue: 3,2
- Last known good version: Not sure
- IDE:
- Platform Target Frameworks: 
  - Android: 8.0

hartez:
### Description of Change ###

When a `ListViewRenderer` is re-used with a new `ListView` element, the old `ListViewAdapter` is disposed of. But the `ListView` still holds a reference to it and attempts to use it, resulting in a Disposed object access exception.

This change unhooks the `ListViewAdapter` from the `ListView` before disposing of it.

This change also fixes potential double-disposal issues with `ListViewAdapter`.

### Issues Resolved ### 

- fixes problem with the UI Test for 3408 crashing
- fixes #3603 

### API Changes ###

 None

### Platforms Affected ### 

- Android

None

Not applicable

### Testing Procedure ###

Run UI Test 3408. Push a lot of buttons. If the application doesn't crash, there's a chance the bug is fixed.

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
@@ -238,8 +238,6 @@ namespace Xamarin.Forms.Platform.Android
 					}
 				}
 
-				RemoveAllViews();
-
 				if (Element != null)
 				{
 					Element.PropertyChanged -= _propertyChangeHandler;
--- a/Xamarin.Forms.Platform.Android/Renderers/ListViewAdapter.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ListViewAdapter.cs
@@ -15,6 +15,7 @@ namespace Xamarin.Forms.Platform.Android
 {
 	internal class ListViewAdapter : CellAdapter
 	{
+		bool _disposed;
 		static readonly object DefaultItemTypeOrDataTemplate = new object();
 		const int DefaultGroupHeaderTemplateId = 0;
 		const int DefaultItemTemplateId = 1;
@@ -417,6 +418,13 @@ namespace Xamarin.Forms.Platform.Android
 
 		protected override void Dispose(bool disposing)
 		{
+			if (_disposed)
+			{
+				return;
+			}
+
+			_disposed = true;
+
 			if (disposing)
 			{
 				CloseContextActions();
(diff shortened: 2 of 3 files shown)


PureWeen:
It looks like the exception moved up the chain. 
The error with navigating back is with the ListViewRenderer not the ListViewAdapter

Which you click "Change Source" and it loads a new nested ListView it looks like the previous nested ListViewRenderer isn't getting disposed of so when navigating away and RemoveAllViews is called it thows an exception

hartez:
The exception in `RemoveAllViews` is reproducible in API 27, but not in 21 (which is why I didn't encounter it earlier). Not sure exactly where it shows up, but I'll take a look at it later (now that I can repro it).

Regardless, we're looking at two distinct bugs here.

Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
```diff
@@ -238,8 +238,6 @@ protected override void Dispose(bool disposing)
 					}
 				}
 
-				RemoveAllViews();
```
rmarinho: this change is my doubt. al lthe rest seems fine

hartez: Why do we need to call `RemoveAllViews()` if the `ViewGroup` is being disposed? 

If we look at what it does (https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/view/ViewGroup.java#5536) - it's doing a ton of cleanup work to handle removing focus from the child views, cancelling touches, etc. And if forces a layout pass. But we don't need any of that because the whole ViewGroup is going away. I suspect we're just slowing down `Dispose` (and causing weird bugs) by calling this.

Issue #3509, opened 2018-08-08
[iOS] NavigationPage.Popped called twice when Navigation.PopAsync is called


jenscski:
### Steps to Reproduce

1. Create new app with the following code as the app

```
using System;
using Xamarin.Forms;

namespace PagesGallery
{
	public class App : Xamarin.Forms.Application
	{
		public App()
		{
			var page = new NavigationPage(new TestPage(false));
			page.Popped += (s, e) =>
			{
				(e.Page as TestPage).Popped();
			};

			MainPage = page;
		}
	}

	public class TestPage : ContentPage
	{
		bool _popped = false;

		public TestPage(bool showPop)
		{
			Title = "Test page";

			var content = new StackLayout();

			content.Children.Add(new Button
			{
				Text = "Push page",
				Command = new Command(async () => await Navigation.PushAsync(new TestPage(true), false)),
			});

			if (showPop)
			{
				content.Children.Add(new Button
				{
					Text = "Pop page",
					Command = new Command(async () => await Navigation.PopAsync(false)),
				});
			}

			Content = content;
		}

		internal void Popped()
		{
			if (_popped) throw new Exception("Already popped");

			_popped = true;
		}
	}
}

```
2. On first page press Push page, on next page press Pop Page
3. It should crash because of this ```throw new Exception("Already popped");```

### Expected Behavior
NavigationPage.Popped only called once on each page removed from navigation stack
### Actual Behavior
NavigationPage.Popped called twice when Navigation.PopAsync is used to remove page from navigation stack
### Basic Information

- Version with issue: 3.1.0.637273
- Last known good version: 3.0 and 3.2.0.637442-pre1
- IDE: VS 15.7.6
- Platform Target Frameworks:
  - iOS:  11.12.0.4
- Nuget Packages:
- Affected Devices:

TimLariviere:
Encountered the same bug.
But I ran the repro with XF 3.0.0.561731 and XF 3.2.0.637442-pre1, this is not specific to XF 3.1.
The problem was already here and is still here.

This bug is especially bad when using [Elmish.XamarinForms](https://github.com/fsprojects/Elmish.XamarinForms) as we need to manage our own page stack but can't keep in sync with NavigationPage on iOS because of this.

I investigated a little and found that when calling PopAsync, 2 things happened.
1. PopAsync raises the Popped event (this is good)
[...]

PR #3851, opened 2018-09-20
Navigation Page double Popped event Workaround

DerGary:
### Description of Change ###
`NavigationPage.PopAsync()` raises the Popped event twice on iOS. This is due to some event triggering when `EventHandler<NavigationRequestedEventArgs> requestPop = PopRequested;` is executed. To prevent this I added the same workaround that is already present in the `NavigationPageRenderer`, I added a bool that ignores the call to `RemoveAsyncInner `when the Method is executing `PopRequested`.

### Issues Resolved ### 
I don't see this as a fix, it is more like a workaround and this PR is intended more to start a discussion about this problem. Because I'm unable to fix it because I don't have enough experience in the XamarinForms code to understand what exactly happens.

- workaround #3509

### API Changes ###
 None

### Platforms Affected ### 
- iOS

None

Before:
`Popped` was triggered twice when `PopAsync` was called.
After:
`Popped` is called once when `PopAsync` is called.

### Testing Procedure ###

I only tested it on iOS and it does work there. To test it navigate to a page and navigate back from this page with the `PopAsync` method while being subscriped to the `NavigationPage.Popped` event.

### PR Checklist ###

- [x] Has UITest 3509


Xamarin.Forms.Core/NavigationPage.cs
```diff
@@ -282,6 +283,10 @@ public async Task<Page> PopAsyncInner(bool animated, bool fast)
 
 		async Task<Page> INavigationPageController.RemoveAsyncInner(Page page, bool animated, bool fast)
 		{
+			if (_ignorePopCall)
```
PureWeen: I agree this doesn't quite feel like the place to fix.  *RemoveAsyncInner* being called twice on the same page seems like the bug.  I wonder if we can fix this purely inside the iOS *NavigationRenderer* .  

For example returning from 
*NavigationRenderer.UpdateFormsInnerNavigation* if _ignorePopCall true

and then setting _ignorePopCall to true inside

*protected virtual async Task<bool> OnPopViewAsync(Page page, bool animated)*

 I also pushed a UI Test that we can use to test this as well

samhouts:
@PureWeen The status is confusing here. Are you still requesting changes?

Xamarin.Forms.Core/NavigationPage.cs
```diff
@@ -292,10 +299,13 @@ async Task<Page> INavigationPageController.RemoveAsyncInner(Page page, bool anim
 			EventHandler<NavigationRequestedEventArgs> requestPop = PopRequested;
 			if (requestPop != null)
 			{
+				_ignorePopCall = true;
```
PureWeen: I agree this doesn't quite feel like the place to fix.  *RemoveAsyncInner* being called twice on the same page seems like the bug.  I wonder if we can fix this purely inside the iOS *NavigationRenderer* .  

For example returning from 
*NavigationRenderer.UpdateFormsInnerNavigation* if _ignorePopCall true

and then setting _ignorePopCall to true inside

*protected virtual async Task<bool> OnPopViewAsync(Page page, bool animated)*

 I also pushed a UI Test that we can use to test this as well

PureWeen:
@samhouts   yea I pushed changes to the branch with a UI test I created an because of the order of things it outdated my comment

I added another review with the same comment so it's more clear that changes are requested

Xamarin.Forms.Core/NavigationPage.cs
```diff
@@ -47,12 +48,14 @@ public NavigationPage(Page root) : this()
...
 
-		public Color BarBackgroundColor {
+		public Color BarBackgroundColor
+		{
 			get => (Color)GetValue(BarElement.BarBackgroundColorProperty);
 			set => SetValue(BarElement.BarBackgroundColorProperty, value);
 		}
 
-		public Color BarTextColor {
+		public Color BarTextColor
```
StephaneDelcroix: please avoid reformattings when nothing is changed

Xamarin.Forms.Core/NavigationPage.cs
StephaneDelcroix: agreed with @PureWeen

DerGary:
@PureWeen Can you explain to me how I can run this UITest? When I run it through the VS Test Explorer it complains that i don't have an android sdk installed. 

```
[27-Sep-18 15:51:03 Informational] NUnit VS Adapter 2.1.1.0 executing tests is started
[27-Sep-18 15:51:07 Informational] Loading tests from C:\git\Xamarin.FormsGithub\Xamarin.Forms.Core.Android.UITests\bin\Debug\Xamarin.Forms.Core.Android.UITests.dll
[27-Sep-18 15:51:08 Informational] Run started: C:\git\Xamarin.FormsGithub\Xamarin.Forms.Core.Android.UITests\bin\Debug\Xamarin.Forms.Core.Android.UITests.dll
[27-Sep-18 15:51:08 Informational] WARNING: Detected System.Net.Http.dll in bin folder. This may be due to this known mono issue: https://bugzilla.xamarin.com/show_bug.cgi?id=60315#c2 - we recommend you delete this dll before running tests to avoid test failures.
[27-Sep-18 15:51:08 Informational] Full log file: C:\Users\stgera\AppData\Local\Temp\uitest\log-2018-09-27_15-51-08-349.txt
[27-Sep-18 15:51:08 Informational] Potential Android SDK location: (No path) - Not set. [ Source: ANDROID_HOME ]
[27-Sep-18 15:51:08 Error] SetUp failed for test fixture Xamarin.Forms.Controls.Issues.IssuesSetup
[27-Sep-18 15:51:08 Error] SetUp : NUnit.Framework.InconclusiveException : App did not start for some reason: System.Exception: Android SDK not found. Please install it and if it is still not located, please set the ANDROID_HOME environment variable to point to the directory.
```

when I unload the android and uwp uitest projects i get build errors....

Xamarin.Forms.Core/NavigationPage.cs
DerGary: reverted pureweens formatting fix

PureWeen:
@DerGary  are you running VS Windows or Mac?

To run the UI Tests for iOS you have to run those on a Mac which are really the only ones that matter for this. That being said you can just run the Control Gallery project for iOS and navigate to the UI Test

GO TO TESTS CASES => type in 3509 => click test

And do the development that way. The UI Test I added was just to give you a framework inside the context of our Control Gallery to work from.

Xamarin.Forms.Core/NavigationPage.cs
DerGary: @PureWeen your suggestion seems to work, I pushed the changes and tested with the UITest. But I don't think its an elegant solution 🤔

DerGary:
> @DerGary are you running VS Windows or Mac?
> 
> To run the UI Tests for iOS you have to run those on a Mac which are really the only ones that matter for this. That being said you can just run the Control Gallery project for iOS and navigate to the UI Test
> 
> GO TO TESTS CASES => type in 3509 => click test
> 
> And do the development that way. The UI Test I added was just to give you a framework inside the context of our Control Gallery to work from.

Thanks I tried to start the wrong project and didn't know that the Control Gallery is the test project 🤦‍♂️

PureWeen:
build --uitests

Xamarin.Forms.Core/NavigationPage.cs
PureWeen: @DerGary  I agree it's not the most elegant. There's a ping pong that happens unfortunately which I would say is from *RemoveAsyncInner* doing too much.  If the calls of remove start from iOS, via back button or swipe, they should probably call into something more specific on the shared code side. That being said *RemoveAsyncInner* is part of a public interface so we've kind of made our bed, so to speak, and doing it this way gives us our end though the means are slightly inelegant.

I've kicked off the UI Tests because that's where I mainly feel like we'll catch other side effects this might lead to

Xamarin.Forms.Platform.iOS/Renderers/NavigationRenderer.cs
```diff
@@ -843,6 +844,10 @@ internal async Task UpdateFormsInnerNavigation(Page pageBeingRemoved)
 		{
 			if (NavPage == null)
 				return;
+			if (_ignorePopCall) {
+				_ignorePopCall = false;
```
PureWeen: What's the reason for setting the false here? opposed to up inside the OnPopViewAsync call after *actuallyRemoved = (poppedViewController == null) ? true : !await task;*

Is there a use case doing it here that I'm not thinking of?

DerGary: you're right that location is way better. I actually didn't understood what the task did because on first glance it looks like it just subscribes some event handlers. But It waits until the Pop is completed so it shouldn't be a fine location to reset _ignorePopCall.

DerGary: I updated the Pull Request and made sure the popped event works with the UITest.

PureWeen:
build --uitests

PureWeen:
UI Tests all ran fine

PureWeen:
failing tests not related

PR #3879, opened 2018-09-23
[Enhancement] Create a bindable property for AutomationId

### LINKED ISSUE #3845: [Enhancement] Make AutomationID Bindable
## Rationale

See https://xamarin.uservoice.com/forums/144858/suggestions/13865469

## Implementation

TBD

## API

TBD

adrianknight89:
### Description of Change ###

Added a bindable AutomationId property. It's Sunday morning, and I'm half asleep still. Part of me says the work that goes into this isn't as trivial as I made it to be. :)

I believe Xamarin takes care of documentation now.

### Issues Resolved ### 

- fixes #3845

### API Changes ###

Added:
 - string AutomationId { get; set; } //Bindable Property

### Platforms Affected ### 

- Core/XAML (all platforms)

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Element.cs
+++ b/Xamarin.Forms.Core/Element.cs
@@ -4,7 +4,6 @@ using System.Collections.ObjectModel;
 using System.Collections.Specialized;
 using System.ComponentModel;
 using System.Runtime.CompilerServices;
-using System.Xml;
 using Xamarin.Forms.Internals;
 
 namespace Xamarin.Forms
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Core/Element.cs
```diff
@@ -26,9 +25,9 @@ public static void SetMenu(BindableObject bindable, Menu menu)
 
 		internal static readonly ReadOnlyCollection<Element> EmptyChildren = new ReadOnlyCollection<Element>(new Element[0]);
 
-		public static readonly BindableProperty ClassIdProperty = BindableProperty.Create("ClassId", typeof(string), typeof(View), null);
+		public static readonly BindableProperty AutomationIdProperty = BindableProperty.Create(nameof(AutomationId), typeof(string), typeof(View), null);
 
-		string _automationId;
+		public static readonly BindableProperty ClassIdProperty = BindableProperty.Create(nameof(ClassId), typeof(string), typeof(View), null);
```
adrianknight89: Why is ClassId a `View` BP instead of an `Element` BP? I made AutomationId match it, but I think they should belong to `Element`, no?

rmarinho:
Why do we need a BP ?

adrianknight89:
I don't know the use case for ClassId, but that's a bindable property. Why not make AutomationId the same? Tbh, I don't care about this change much, but looks like the OP was following some example where everything was being wired via BPs, and he wanted consistency? Idk.

Xamarin.Forms.Core/Element.cs
hartez: Yes, I think it should be `Element`.

adrianknight89: Done.

hartez:
I've checked all of the UI test failures here, and each one of them is a failure in test 42074 which is being caused by the bug fixed in #3899. So those failures are not a reason to prevent merging this change.

PR #3889, opened 2018-09-24
Fix UI test B43941

### LINKED ISSUE #2376: UI test B43941 needs to be automated
- [ ] Test needs to be automated for all platforms

Viridovics:
### Description of Change ###

* Include all platforms
* Remove non-deterministic test behaviour ``` // At this point, the counter can be any value, but it's most likely not zero. ```

### Issues Resolved ### 

- fixes #2376

### API Changes ###
 
 None

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla43941.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla43941.cs
@@ -33,6 +33,7 @@ namespace Xamarin.Forms.Controls.Issues
 				RunningApp.Back();
 			}
 
+			// At this point, the counter can be any value, but it's most likely not zero.
 			// Invoking GC once is enough to clean up all garbage data and set counter to zero
 			RunningApp.WaitForElement(q => q.Marked("GC"));
 			RunningApp.Tap(q => q.Marked("GC"));
@@ -84,7 +85,7 @@ namespace Xamarin.Forms.Controls.Issues
 				HorizontalTextAlignment = TextAlignment.Center,
 				VerticalTextAlignment = TextAlignment.Center
 			};
-			var childPages = new List<ContentPage43941>();
+
 			Content = new StackLayout
 			{
 				Orientation = StackOrientation.Vertical,
@@ -105,7 +106,6 @@ namespace Xamarin.Forms.Controls.Issues
 						AutomationId = "GC",
 						Command = new Command(o =>
 						{
-							childPages = new List<ContentPage43941>();
 							GC.Collect();
 							GC.WaitForPendingFinalizers();
 							Label.Text = "Counter: " + Counter;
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla43941.cs
```diff
@@ -33,12 +33,10 @@ public void Bugzilla43941Test()
 				RunningApp.Back();
 			}
 
-			// At this point, the counter can be any value, but it's most likely not zero.
 			// Invoking GC once is enough to clean up all garbage data and set counter to zero
 			RunningApp.WaitForElement(q => q.Marked("GC"));
 			RunningApp.Tap(q => q.Marked("GC"));
-
-			RunningApp.WaitForElement(q => q.Marked("Counter: 0"));
+			Assert.AreEqual(0, LandingPage43941.Counter);
```
kingces95: The UITest and the App are different processes and so each get their own copy of `.Counter`; The UITest needs to use `RunningApp` to query the `.Counter`'s value in the other process. 

The app exposes the `.Counter` value by updating the `Label`'s text so that the UITest can query `"Counter: 0"`.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla43941.cs
```diff
@@ -85,7 +83,7 @@ public LandingPage43941()
 				HorizontalTextAlignment = TextAlignment.Center,
 				VerticalTextAlignment = TextAlignment.Center
 			};
-
+			var childPages = new List<ContentPage43941>();
```
kingces95: What is the motivation for storing the pages in the `childPages` list?

Viridovics: If we lose link  on ```ContentPage43941``` then Garbage collector can invoke finalizer of ```ContentPage43941``` at any time. Therefore test behaviour can be different each run. Sometimes counter can be zero after test ```for``` loop.

P.S. This changes are out of issue scope. I just want to remove comment ```// At this point, the counter can be any value, but it's most likely not zero.```

samhouts:
@kingces95 Please re-review

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla43941.cs
hartez: The original issue this test was written for (https://bugzilla.xamarin.com/show_bug.cgi?id=43941) was a memory leak of the ContentPage. The problem was that the counter could _never_ get to zero. 

As long as the instance count is zero at the end of the test, we don't care what the counter values were in the interim. So there's no reason to keep the list around to prevent garbage collection; garbage collection is the thing we _want_ to happen.

Viridovics: I understand that garbage collection is thing we want to happen. Therefore before ```GC.Collect()``` link on ```childPages``` will be removed. Common idea is prevent invoking finalizers at a random time. Finalizers will be started only after _GC button_ click.

hartez:
UI test failures are unrelated.

PR #3928, opened 2018-09-28
Fix UI test 39378

### LINKED ISSUE #2305: UI test for 39378 should _not_ be automated
Test is automatically marked `inconclusive`, does not have a `ManualReview` category. The test should have manual on-screen instructions and the automation should be removed.

Also, the test uses a remote image URL which we do not control. It should be replaced with an image in our repository (see the 32691 test for an example).

adrianknight89:
### Description of Change ###

Removed automated test
Added on-screen instructions
Switched to using another URL for the image
@hartez wanted to see `ManualReview` category added, but this category is set when `#if UITEST` is true. Why set it if the test is not automated? Should this still be added?

### Issues Resolved ### 

- fixes #2305

### PR Checklist ###

- [ ] Has automated tests 


hartez:
> @hartez wanted to see ManualReview category added, but this category is set when #if UITEST is true. 

Sorry, I worded the issue in a confusing way. I meant that _instead of marking it inconclusive_, it should either be marked `ManualReview` or not automated at all.

samhouts:
failing tests not related

PR #3941, opened 2018-09-29
[GTK] Changes in TableView to allow refresh the content (updating/removing TableSections)

### LINKED ISSUE #3725: [GTK] TableView doesn't appear to be updating when you remove and add new Table Sections
### Description
When running the Xamarin Forms Control Gallery for GTK the test cases screen doesn't appear to update it's view as you try to filter down issues

The list of issues is updated by removing and adding a new Section
```C#
	_section = new TableSection("Bug Repro");

				foreach (var issueCell in issueCells) {
					_section.Add (issueCell);
				} 

				Root.Add(_section);
```
But those changes don't seem to propagate visually

### Steps to Reproduce

1.  Run Xamarin.Forms.ControlGallery.GTK
2. Click go to test cases
3. type things into the SearchBar
4. you'll notice that the list of issues doesn't update like it does on the other platforms

### Expected Behavior
the list of issues should filter down based on the input

### Actual Behavior
They don't filter down

### Basic Information

- Version with issue: 3.2
- Last known good version: not sure

### Screenshots

image

vs

image

jsuarezruiz:
### Description of Change ###

Changes in TableView GTK Control and TableViewRenderer to allow refresh the content (updating/removing TableSections).

### Issues Resolved ### 
- fixes #3725 

### Platforms Affected ### 
- GTK

Now, the TableView refresh the content adding or removing TableSections, etc.

### Testing Procedure ###
When running the Xamarin Forms Control Gallery for GTK the test cases screen doesn't appear to update it's view as you try to filter down issues.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.GTK/Controls/TableView.cs
+++ b/Xamarin.Forms.Platform.GTK/Controls/TableView.cs
@@ -28,11 +28,8 @@ namespace Xamarin.Forms.Platform.GTK.Controls
 			}
 			set
 			{
-				if (_source != value)
-				{
-					_source = value;
-					RefreshSource(_source);
-				}
+				_source = value;
+				RefreshSource(_source);
 			}
 		}
 
@@ -59,7 +56,7 @@ namespace Xamarin.Forms.Platform.GTK.Controls
 			}
 		}
 
-		private int GetUnevenRowCellHeight(Gtk.Container cell)
+		int GetUnevenRowCellHeight(Container cell)
 		{
 			int height = -1;
 
@@ -73,7 +70,7 @@ namespace Xamarin.Forms.Platform.GTK.Controls
 			return height;
 		}
 
-		private Cell GetXamarinFormsCell(Container cell)
+		Cell GetXamarinFormsCell(Container cell)
 		{
 			try
 			{
@@ -90,7 +87,7 @@ namespace Xamarin.Forms.Platform.GTK.Controls
 			}
 		}
 
-		private void BuildTableView()
+		void BuildTableView()
 		{
 			CanFocus = true;
 			ShadowType = ShadowType.None;
@@ -100,8 +97,11 @@ namespace Xamarin.Forms.Platform.GTK.Controls
 
 			_root = new VBox(false, 0);
 
-			Viewport viewPort = new Viewport();
-			viewPort.ShadowType = ShadowType.None;
+			Viewport viewPort = new Viewport
+			{
+				ShadowType = ShadowType.None
+			};
+
 			viewPort.Add(_root);
 
 			Add(viewPort);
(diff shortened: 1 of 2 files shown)


Xamarin.Forms.Platform.GTK/Controls/TableView.cs
```diff
@@ -100,20 +97,31 @@ private void BuildTableView()
...
 		}
 
-		private void RefreshSource(TableRoot source)
+		void RefreshSource(TableRoot source)
 		{
+			// Clear
+			_cells.Clear();
+
+			foreach (var child in _root.AllChildren)
+			{
```
PureWeen: Remove brackets for one line block

Xamarin.Forms.Platform.GTK/Renderers/TableViewRenderer.cs
```diff
@@ -34,6 +35,11 @@ protected override void Dispose(bool disposing)
 
 		protected override void OnElementChanged(ElementChangedEventArgs<TableView> e)
 		{
+			if (e.OldElement != null)
+			{
```
PureWeen: Remove for one line block

StephaneDelcroix: @PureWeen just like you, I prefer not to have them, but we have both styles in our code, and we shouldn't block a PR on this.

jsuarezruiz: Out of curiosity, and to take it into account in other PR, which is preferred?

PureWeen: https://github.com/xamarin/Xamarin.Forms/blob/master/CONTRIBUTING.md#bug-fixes
> Follow the style used by the .NET Foundation, with two primary exceptions
https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

The two Primary Exceptions being tabs instead of spaces and not using private

And then from the .NET foundation

> We use Allman style braces, where each brace begins on a new line. A single line statement block can go without braces but the block must be properly indented on its own line and must not be nested in other statement blocks that use braces (See issue 381 for examples). One exception is that a using statement is permitted to be nested within another using statement by starting on the following line at the same indentation level, even if the nested using contains a controlled block.

I like to be consistent because then I can just hit CTRL E, D and rely on the editorconfig settings to ensure my changes match the coding style. In the past VS for Mac was a bit frustrating with indenting properly as well so I found it easier to just do document formats.

jsuarezruiz: Thanks @PureWeen!

PR #3943, opened 2018-09-29
Add a support for an application menu toolbar and make it easier for subclasses to use the FormsWindow

ylatuya:
### Description of Change ###
Add a Gtk menu toolbar to the main window so Gtk+ based application can use an application menu.
Make it easier for applications to change the main window size and subclass it. 

### API Changes ###

 None

### Platforms Affected ### 
- GTK+

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.GTK/FormsWindow.cs
+++ b/Xamarin.Forms.Platform.GTK/FormsWindow.cs
@@ -9,6 +9,9 @@ namespace Xamarin.Forms.Platform.GTK
 	{
 		private Application _application;
 		private Gdk.Size _lastSize;
+		private VBox _content;
+		private MenuBar _menu;
+		private AccelGroup _accelGroup;
 
 		public FormsWindow ()
 			: base (WindowType.Toplevel)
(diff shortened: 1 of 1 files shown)


samhouts:
build

jsuarezruiz:
Is a good idea but, maybe, the best way is to expand the support of menus: https://github.com/xamarin/Xamarin.Forms/pull/1126

In this way, the same syntax between Desktop platforms is maintained, in addition to allowing the creation of shared menus in definition between platforms.

ylatuya:
@jsuarezruiz  That's the final idea, and this would be the starting point of full menus support.
I could change the visibility of the menu/accelerator/content to protected.

jsuarezruiz:
Let me include @rmarinho in the thread. He worked in the menu system in macOS backend.

ylatuya:
This is what we use our selves on Mac to integrate the gtk menu as a native macOS menu https://wiki.gnome.org/Projects/GTK+/OSX/Integration#Gtk-mac-integration.2FGtkOSXApplication. On Windows and Linux you still need the menu bar  in the main window.

rmarinho:
MacOS already supports menus, can you explain what's missing ?

jsuarezruiz:
Sorry, I do not explain well. The Xamarin.Forms menu system is implemented in the macOS backend. @rmarinho, I added you to the thread because you did it. This PR would allow to use menus in GTK but not with the same menu system implemented in macOS.

And my question, is ok to add this for now (until implement the menu system) https://github.com/xamarin/Xamarin.Forms/pull/1126?. To create the menu, specific GTK code would be used.

PR #3946, opened 2018-09-30
[Core] Added IsVisible property to ToolbarItem

### LINKED ISSUE #3838: [Enhancement] Add IsVisible to ToolbarItem
## Rationale

At times you need to check if a toolbar item is visible, and to set it thus.

Related: https://xamarin.uservoice.com/forums/258559/suggestions/15508584

## Implementation

Add new properties to `Toolbaritem` for:
* IsVisible

jsuarezruiz:
### Description of Change ###

Added IsVisible property to ToolbarItem.

### Issues Resolved ### 

- fixes #3838 

### API Changes ###

Added:
 - bool ToolbarItem.IsVisible{ get; set; } //Bindable Property

### Platforms Affected ### 

- Core
- iOS
- Android
- UWP
- macOS
- WPF
- GTK
- Tizen

Some screenshots:

GTK
3838-gtk

WPF
3838-wpf

**NOTE**: In other platforms, the behaviors is the same.

### Testing Procedure ###
Using Core Gallery, open the ToolbarItems sample and press the "Change tb2 visibiltiy" button. Must change the visiblity of the "tb2" ToolbarItem.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/ToolbarItem.cs
+++ b/Xamarin.Forms.Core/ToolbarItem.cs
@@ -21,7 +21,7 @@ namespace Xamarin.Forms
 
 			if (item != null)
 			{
-				var items = ((ContentPage)item.Parent).ToolbarItems;
+				var items = ((Page)item.Parent).ToolbarItems;
 
 				if ((bool)newvalue && !items.Contains(item))
 				{
--- a/Xamarin.Forms.Core/ToolbarItem.cs
+++ b/Xamarin.Forms.Core/ToolbarItem.cs
@@ -16,11 +16,11 @@ namespace Xamarin.Forms
 		{
 			var item = bindable as ToolbarItem;
 
-			if (item != null && item.Parent == null)
-				return;
-
 			if (item != null)
 			{
+				if (item.Parent == null)
+					return;
+
 				var items = ((Page)item.Parent).ToolbarItems;
 
 				if ((bool)newvalue && !items.Contains(item))
--- a/Xamarin.Forms.Core/ToolbarItem.cs
+++ b/Xamarin.Forms.Core/ToolbarItem.cs
@@ -25,11 +25,11 @@ namespace Xamarin.Forms
 
 				if ((bool)newvalue && !items.Contains(item))
 				{
-					items.Add(item);
+					Device.BeginInvokeOnMainThread(() => { items.Add(item); });
 				}
 				else if (!(bool)newvalue && items.Contains(item))
 				{
-					items.Remove(item);
+					Device.BeginInvokeOnMainThread(() => { items.Remove(item); });
 				}
 			}
 		});
--- a/Xamarin.Forms.Core/ToolbarItem.cs
+++ b/Xamarin.Forms.Core/ToolbarItem.cs
@@ -12,11 +12,34 @@ namespace Xamarin.Forms
 
 		static readonly BindableProperty PriorityProperty = BindableProperty.Create("Priority", typeof(int), typeof(ToolbarItem), 0);
 
+		public static readonly BindableProperty IsVisibleProperty = BindableProperty.Create("IsVisible", typeof(bool), typeof(ToolbarItem), true, BindingMode.TwoWay, propertyChanged: (bindable, oldvalue, newvalue) =>
+		{
+			var item = bindable as ToolbarItem;
+
+			if (item != null && item.Parent == null)
+				return;
+
+			if (item != null)
+			{
+				var items = ((ContentPage)item.Parent).ToolbarItems;
+
+				if ((bool)newvalue && !items.Contains(item))
+				{
+					items.Add(item);
+				}
+				else if (!(bool)newvalue && items.Contains(item))
+				{
+					items.Remove(item);
+				}
+			}
+		});
+
 		public ToolbarItem()
 		{
+	
 		}
 
-		public ToolbarItem(string name, string icon, Action activated, ToolbarItemOrder order = ToolbarItemOrder.Default, int priority = 0)
+		public ToolbarItem(string name, string icon, Action activated, ToolbarItemOrder order = ToolbarItemOrder.Default, int priority = 0, bool isVisible = true)
 		{
 			if (activated == null)
 				throw new ArgumentNullException("activated");
@@ -26,6 +49,7 @@ namespace Xamarin.Forms
 			Clicked += (s, e) => activated();
 			Order = order;
 			Priority = priority;
+			IsVisible = isVisible;
 		}
 
 		[Obsolete("Name is obsolete as of version 1.3.0. Please use Text instead.")]
@@ -45,6 +69,12 @@ namespace Xamarin.Forms
 		{
 			get { return (int)GetValue(PriorityProperty); }
 			set { SetValue(PriorityProperty, value); }
+		}
+
+		public bool IsVisible
+		{
+			get { return (bool)GetValue(IsVisibleProperty); }
+			set { SetValue(IsVisibleProperty, value); }
 		}
 
 		[Obsolete("Activated is obsolete as of version 1.3.0. Please use Clicked instead.")]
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Core/ToolbarItem.cs
```diff
@@ -12,11 +12,34 @@ public class ToolbarItem : MenuItem
...
+			var item = bindable as ToolbarItem;
+
+			if (item != null && item.Parent == null)
+				return;
+
+			if (item != null)
+			{
+				var items = ((ContentPage)item.Parent).ToolbarItems;
+
+				if ((bool)newvalue && !items.Contains(item))
```
rookiejava: trivial enhancement
```cs
if (items.Contains(item))
{
     if (!(bool)newValue) 
     {
          items.Remove(item)
     }
}
else
{
    if ((bool)newValue) 
    {
         items.Add(item)
    }
}
```

Xamarin.Forms.Core/ToolbarItem.cs
```diff
@@ -12,11 +12,34 @@ public class ToolbarItem : MenuItem
 
 		static readonly BindableProperty PriorityProperty = BindableProperty.Create("Priority", typeof(int), typeof(ToolbarItem), 0);
 
+		public static readonly BindableProperty IsVisibleProperty = BindableProperty.Create("IsVisible", typeof(bool), typeof(ToolbarItem), true, BindingMode.TwoWay, propertyChanged: (bindable, oldvalue, newvalue) =>
+		{
+			var item = bindable as ToolbarItem;
+
+			if (item != null && item.Parent == null)
```
rookiejava:
```cs
if (item != null)
{
    if (item.Parent == null)
        return;
    
    //...
}
```

Xamarin.Forms.Core/ToolbarItem.cs
```diff
@@ -12,11 +12,34 @@ public class ToolbarItem : MenuItem
...
+				return;
+
+			if (item != null)
+			{
+				var items = ((ContentPage)item.Parent).ToolbarItems;
+
+				if ((bool)newvalue && !items.Contains(item))
+				{
+					items.Add(item);
```
AmrAlSayed0: Any reason why this would not be ```Device.BeginInvokeOnMainThread(() => { items.Add(item); });``` like it is [here](https://forums.xamarin.com/discussion/comment/306901/#Comment_306901) to guarantee it will run on the UI thread.

Xamarin.Forms.Core/ToolbarItem.cs
```diff
@@ -12,11 +12,34 @@ public class ToolbarItem : MenuItem
...
+				}
+				else if (!(bool)newvalue && items.Contains(item))
+				{
+					items.Remove(item);
```
AmrAlSayed0: Same here.

Xamarin.Forms.Core/ToolbarItem.cs
```diff
@@ -12,11 +12,34 @@ public class ToolbarItem : MenuItem
...
+		public static readonly BindableProperty IsVisibleProperty = BindableProperty.Create("IsVisible", typeof(bool), typeof(ToolbarItem), true, BindingMode.TwoWay, propertyChanged: (bindable, oldvalue, newvalue) =>
+		{
+			var item = bindable as ToolbarItem;
+
+			if (item != null && item.Parent == null)
+				return;
+
+			if (item != null)
+			{
+				var items = ((ContentPage)item.Parent).ToolbarItems;
```
AmrAlSayed0: The parent could be any ```Page``` not just ```ContentPage``` so it would be better to cast to the least derived class that has a ```ToolbarItems``` property.

jsuarezruiz: Right, good point.

Xamarin.Forms.Core/ToolbarItem.cs
jsuarezruiz: Changed!

Xamarin.Forms.Core/ToolbarItem.cs
```diff
@@ -12,11 +12,34 @@ public class ToolbarItem : MenuItem
 
 		static readonly BindableProperty PriorityProperty = BindableProperty.Create("Priority", typeof(int), typeof(ToolbarItem), 0);
 
+		public static readonly BindableProperty IsVisibleProperty = BindableProperty.Create("IsVisible", typeof(bool), typeof(ToolbarItem), true, BindingMode.TwoWay, propertyChanged: (bindable, oldvalue, newvalue) =>
```
PureWeen: nameof(IsVisible)

Xamarin.Forms.Core/ToolbarItem.cs
```diff
@@ -12,11 +12,34 @@ public class ToolbarItem : MenuItem
...
+			if (item != null)
+			{
+				if (item.Parent == null)
+					return;
+
+				var items = ((Page)item.Parent).ToolbarItems;
+
+				if ((bool)newvalue && !items.Contains(item))
+				{
+					Device.BeginInvokeOnMainThread(() => { items.Add(item); });
```
PureWeen: I don't think adding/removing the toolbar items from the page is the best way to go about this. That seems like an unexpected side effect of setting visibility. 

Can we maybe use ToolbarTracker and filter out the visible items here?
https://github.com/xamarin/Xamarin.Forms/blob/bd31e1e9fc8b2f9ad94cc99e0c7ab058174821f3/Xamarin.Forms.Core/Internals/ToolbarTracker.cs#L50

Also if we got rid of Linq in that class that would be great as well.

Xamarin.Forms.Core/ToolbarItem.cs
StephaneDelcroix: a few remarks:
- use `nameof()`,
- use the same binding mode as `VisualElement.IsVisibleProperty`,
- when the propertychanged handler is more than a single-line long, extract that to a method, for clarity

and then:
- for css-styling purposes, it'd be nice if the property was actually shared with VisualElement's. But as the ToolBarItems aren't CSS-selectable, I'm fine if we don't do it at this time

Xamarin.Forms.Core/ToolbarItem.cs
StephaneDelcroix: 100% agree with @PureWeen here. handle that closer to the platform

Xamarin.Forms.Core/ToolbarItem.cs
```diff
@@ -12,11 +12,34 @@ public class ToolbarItem : MenuItem
...
+			}
+		});
+
 		public ToolbarItem()
 		{
+	
 		}
 
-		public ToolbarItem(string name, string icon, Action activated, ToolbarItemOrder order = ToolbarItemOrder.Default, int priority = 0)
+		public ToolbarItem(string name, string icon, Action activated, ToolbarItemOrder order = ToolbarItemOrder.Default, int priority = 0, bool isVisible = true)
```
StephaneDelcroix: this is an ABI change and will break on already compiled code. please add an overload with the old API

adammeaney:
I would like to chime in and suggest that this is going to change the toolbar item order if they are all at the same level.

If you had 3 items, all with the same Priority, then toggling the visibility on the second one would move it on screen after you make it visible again.

This could be fixed with setting different priorities, but currently without the ability to toggle visibility none of my apps do that. 

Much simpler to do it in the platforms and leave the toolbar item list the same.

PureWeen:
I'm going to close this one for the time being.

Xamarin.Forms.Core/ToolbarItem.cs
bill2004158: 100% agree with @PureWeen, and we need make sure the position of item is still same as before when item become visible again.

PR #3955, opened 2018-10-01
[iOS, Android, macOS] fixes accessibility of Picker. Support accessibility for macOS,

### LINKED ISSUE #3454: Picker accessibility text is wrong
### Description

Using voiceover/talkback on both Android and iOS, picker control is read as "entry/textfield, double tap to edit",
also adding AutmationProperies.SetName + AutomationProperties.SetHelpText does not prevent device from reading texts in this order:
1. AutomationProperties.SetName 
2. "textfield"
3. AutomationProperties.SetHelpText
4. "double tap to edit"

Additionally, focusing picker (either by method call or by tapping) causes voiceover/talkback to say "character mode, something something"

### Expected Behavior
Read picker as a Button. Picker has no meaning in accessibility (especially on Android). The TextField that's clicked on should read as a "Button" and not a text field because the action a user is going to take on the picker will be the same as a button.  When a user hears that something is an "edit text field" that tells the user the interaction is going to be via a keyboard.

Using SetName and SetHelpText also needs to be more resilient on Android which will probably involve injecting yourself into the delegates that are used to build the help text used.

https://stackoverflow.com/questions/25641081/android-acccessibility-how-do-i-change-the-text-read-out-loud-for-an-edittext-v

Also once something is selected the "Text Field" needs to retain whatever was set for SetName and SetHelpText. 

### Actual Behavior
picker is read as textfield/entry

### Basic Information

- IDE: VS for Mac Comunity 2017 
- Platform Target Frameworks: 
  - iOS: 11.4
  - Android: 27
- Android Support Library Version: v7

paymicro:
### Description of Change ###

[Android, iOS] Changed accessibility of Picker. Now it reads like a `button`. If something is selected in it, this item is also read.
[Mac ] Added support AutomationProperty.

### Issues Resolved ### 

- fixes #3454

### API Changes ###
 
 None

### Platforms Affected ### 

- macOS

None

Before
before

After
after
iOS
iOS

### Testing Procedure ###
- Run UI test 3454 on macOS
- Check accessibility on picker

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3454.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3454.cs
@@ -1,0 +1,32 @@
+﻿using System.Linq;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 3454, "Picker accessibility text is wrong", PlatformAffected.macOS)]
+	public class Issue3454 : TestContentPage
+	{
+		protected override void Init()
+		{
+			var picker = new Picker();
+			var picker2 = new Picker();
+#if !UITEST
+			picker.SetAutomationPropertiesName("pi pi picker");
+			picker.SetAutomationPropertiesHelpText("double tap to edit");
+#endif
+			picker.ItemsSource = picker2.ItemsSource = Enumerable.Range(1, 10).Select(i => $"item {i}").ToList();
+
+			Content = new StackLayout
+			{
+				Children = {
+					new Label { Text = "↓ Picker with setted AutomationProperties ↓" },
+					picker,
+					new Label { Text = "↓ Default picker ↓" },
+					picker2
+				}
+			};
+		}
+	}
+}
--- a/Xamarin.Forms.Platform.Android/AppCompat/PickerRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/PickerRenderer.cs
@@ -1,6 +1,7 @@
 using Android.App;
 using Android.Util;
 using System;
+using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.ComponentModel;
 using System.Linq;
@@ -9,6 +10,8 @@ using Android.Widget;
 using AColor = Android.Graphics.Color;
 using Android.Text;
 using Android.Text.Style;
+using Android.Views;
+using Android.Views.Accessibility;
 
 namespace Xamarin.Forms.Platform.Android.AppCompat
 {
@@ -18,6 +21,11 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 		bool _disposed;
 		TextColorSwitcher _textColorSwitcher;
 		int _originalHintTextColor;
+		PickerAccessibilityDelegate _pickerAccessibilityDelegate;
+
+		HashSet<Keycode> availableKeys = new HashSet<Keycode>(new[] {
+			Keycode.Tab, Keycode.Forward, Keycode.Back, Keycode.DpadDown, Keycode.DpadLeft, Keycode.DpadRight, Keycode.DpadUp
+		});
 
 		public PickerRenderer(Context context) : base(context)
 		{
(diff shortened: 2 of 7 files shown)


PureWeen:
@paymicro  did you reference the correct issue for fixes? The issue you are referencing looks to deal with an issue with the picker control representing itself as a textfield which doesn't seem related to this PR

Xamarin.Forms.Platform.iOS/ViewRenderer.cs
```diff
@@ -141,11 +141,17 @@ protected override void SetAccessibilityHint()
...
+#if __MOBILE__
 			if (_defaultAccessibilityHint == null)
 				_defaultAccessibilityHint = Control.AccessibilityHint;
 
 			Control.AccessibilityHint = (string)Element.GetValue(AutomationProperties.HelpTextProperty) ?? _defaultAccessibilityHint;
+#else
+			if (_defaultAccessibilityHint == null)
+				_defaultAccessibilityHint = Control.AccessibilityTitle;
+
+			Control.AccessibilityTitle = (string)Element.GetValue(AutomationProperties.HelpTextProperty) ?? _defaultAccessibilityHint;
```
PureWeen: The few places in this PR where there is duplicate code like this retrieve the bindable property. Can you move these to a Re-usable property?

Something like
string HelpText => (string)Element.GetValue(AutomationProperties.HelpTextProperty) ?? _defaultAccessibilityHint;

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3454.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Linq;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 3454, "Picker accessibility text is wrong", PlatformAffected.macOS)]
+	public class Issue3454 : TestContentPage
```
PureWeen: Can you run a *format document* over this issue? So all the brackets can be on the same line and the spaces between name and parenthesis can get removed?

paymicro:
> @paymicro did you reference the correct issue for fixes?

Yes. a bit not understood issue. But it's already fixed.

Xamarin.Forms.Platform.iOS/ViewRenderer.cs
```diff
@@ -141,16 +141,18 @@ protected override void SetAccessibilityHint()
 
 			if (Element == null)
 				return;
+
+			var value = (string)Element.GetValue(AutomationProperties.HelpTextProperty) ?? _defaultAccessibilityHint;
```
PureWeen: Hopefully my effort to make these generalized isn't more effort than it's worth :-) If Defs always make things fun 

The issue here is that value is being set prior to the *_defaultAccessibilityHint* being set

I could see it being useful to break that up into its own getter as well in case something else down the road wants to access Control's value.

Something like 
```C#
#if __MOBILE__
string ControlAccessibilityHint => Control.AccessibilityHint;
#else
string ControlAccessibilityHint => Control.AccessibilityTitle;
#endif
```
And then this section could be 

```C#
if (_defaultAccessibilityHint == null)
	_defaultAccessibilityHint = ControlAccessibilityHint ;

var helpTextValue = (string)Element.GetValue(AutomationProperties.HelpTextProperty) ?? _defaultAccessibilityHint;

#if __MOBILE__
			Control.AccessibilityHint = helpTextValue ;
#else
			Control.AccessibilityTitle = helpTextValue ;
#endif
```

And then something similar elsewhere.  It's nice when if defs can be pushed out to methods and then logic code can just not have them

paymicro: No limit to perfection. =)
If I develop this topic, then i can also add a setter. So it will be a little easier and removes some `__MOBILE__` directives.

PureWeen:
@paymicro  did you look at this in relation to Android? Make sure it all works over there?

samhouts:
@rmarinho @PureWeen Please re-review! Thanks

Xamarin.Forms.Platform.Android/AppCompat/PickerRenderer.cs
```diff
@@ -70,10 +68,14 @@ protected override void OnElementChanged(ElementChangedEventArgs<Picker> e)
 					textField.InputType = InputTypes.Null;
 					textField.KeyPress += TextFieldKeyPress;
 					textField.SetOnClickListener(PickerListener.Instance);
+					textField.ContentDescription = nameof(Picker);
```
rmarinho: ContentDescription with be related by AutomationID if that's was set.

rmarinho:
Hi just tested Android, iOS , and MacOs all seem to work as expected

PureWeen:
@paymicro  hold off on working on this one yet. We're still working on defining exactly how we want all these things to work so they are the most a11y compatible.

PureWeen:
alright @paymicro I updated the expected behavior on the original issue for what would be expected for this one.

rmarinho:
Grr i think my merge wasn't good can you double check @paymicro

PR #3963, opened 2018-10-01
InternalsVisibleTo attributes in preparation for CollectionView preview packages

hartez:
Adding the necessary InternalsVisibleTo attributes to allow the CollectionView preview packages to work with 3.3.0 and up while it's in the preview stage.


hartez:
Closing for now until we decide if this is the route we want to take.

PR #4000, opened 2018-10-04
[Android] Amended ScrollViewRenderer with smooth scrolling

AndreiMisiukevich:
### Description of Change ###
Hi
Now Android scroll view has default 1 second animation. This looks a little bit ugly, when we need scroll only a few points (It moves like a turtle)

Perhaps, it would be OK idea to change 1 sec animation to SmoothScrollTo method (like iOS has)

### API Changes ###

Changed:
 - ScrollView ScrollTo implementation

### Platforms Affected ### 
**Android**

Now it work like on iOS, in my opinion

### Testing Procedure ###
Here is code for testing, just try to tap buttons and check animation
It won't take 1 second for every scroll

```csharp
public partial class App : Application
	{
		public App ()
		{
			InitializeComponent ();

			var scroll = new ScrollView();

			var st = new StackLayout
			{
				BackgroundColor = Color.Red,
				HeightRequest = 5000,
				Children = {
					new Button
					{
						Text = "Near",
						BackgroundColor = Color.DarkBlue,
						Command = new Command(() => {
							scroll.ScrollToAsync(0, 10, true);
						})
					},
					new Button
					{
						Text = "Far",
						BackgroundColor = Color.DarkBlue,
						Command = new Command(() => {
							scroll.ScrollToAsync(0, 100, true);
						})
					},
					new Button
					{
						Text = "Very far",
						BackgroundColor = Color.DarkBlue,
						Command = new Command(() => {
							scroll.ScrollToAsync(0, 300, true);
						})
					},
					new Button
					{
						Text = "Start",
						BackgroundColor = Color.DarkBlue,
						Command = new Command(() => {
							scroll.ScrollToAsync(0, 0, true);
						})
					},
					new Button
					{
						Text = "Start",
						BackgroundColor = Color.DarkBlue,
						Command = new Command(() => {
							scroll.ScrollToAsync(0, 0, true);
						})
					},
					new Button
					{
						Text = "Start",
						BackgroundColor = Color.DarkBlue,
						Command = new Command(() => {
							scroll.ScrollToAsync(0, 0, true);
						})
					}
				}
			};

			scroll.Content = st;

			MainPage = new ContentPage
			{
				Content = scroll
			};
		}
	}
```

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/ScrollViewRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ScrollViewRenderer.cs
@@ -4,6 +4,7 @@ using System.Threading.Tasks;
 using Android.Animation;
 using Android.Content;
 using Android.Graphics;
+using Android.Runtime;
 using Android.Views;
 using Android.Widget;
 using Xamarin.Forms.Internals;
@@ -334,7 +335,6 @@ namespace Xamarin.Forms.Platform.Android
 			{
 				return;
 			}
-
 			_isEnabled = Element.IsEnabled;
 		}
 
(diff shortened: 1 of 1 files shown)


samhouts:
Thank you for your submission! Would you please use our pull request template? Thank you! https://github.com/xamarin/Xamarin.Forms/blob/master/PULL_REQUEST_TEMPLATE.md

AndreiMisiukevich:
@samhouts done

AndreiMisiukevich:
Any ideas, how to await **SmoothScrollTo** ?

kingces95:
If the issue is just the time it takes for an animation to complete for short distance, why not just use a shorter time for shorter distances? I believe that would result in a smaller delta. That's important because it's less likely to break anyone to simply change the time it takes for an animation to complete. That said, if there is another reason to use smooth scrolling in addition to the constant than we can consider that.

AndreiMisiukevich:
@kingces95 thanks for feedback
It sounds reasonable, but how can we calculate acceptable time for distance?
What distance does match 1sec ?

kingces95:
Well, we'd have to play with the math. Some type of function that varies from 0 to 1 as a function of the distance could be used as a factor for the duration.

Start with `1 - 1/(1 -x/a)` where `x` is the distance to scroll and `a` is some percentage of the screen size. I tried playing around with different options but it's clear it's gonna take some playing. It's not a binary thing, right? You gotta play with it until it "feels right".

I used [this](https://www.desmos.com/calculator/abra3gmftg) to play with the graphs. 

Or, just start simple, with a linear factor up to the size of the screen then a factor of 1 after that. 

So we'd be happy to play with whatever function you come up with!

kingces95:
Um, well, either way. I'd vote for continuing with this issue but save your old solution somewhere. How about that?

kingces95:
@AndreiMisiukevich Actually, on second thought, please open a new issue for the new branch so the team can compare the solutions! Sorry, should have thought about that. Thanks!!

kingces95:
So we'll just close this one out and you can link it to the next one you open.

AndreiMisiukevich:
@kingces95 i noted, that ListView uses SmoothScroll (Android, sure).

Maybe it's good idea to mark **Task ScrollTo** ScrollView methods as Obsolete and add **void ScrollTo** as for ListView ?

Issue #2709, opened 2018-05-15
[Android] Include basic proguard file so starting Xamarin Forms project can release with proguard enabled


PureWeen:
### Description
The current starter Xamarin Forms Android project will not deploy as is with proguard enabled. 

The minimum configuration currently required is a proguard file with the following entry

```
-keep class android.support.v7.widget.FitWindowsFrameLayout { *; }
-dontwarn android.support.v7.widget.FitWindowsFrameLayout
```

The support libraries all include proguard.txt files as part of their nuget packages so Xamarin Forms could also include the above minimum.

It is debatable whether this is useful since as a user grows an application and uses more things they are going to have to learn how to interact with proguard and modify the file. It is just the inevitable outcome of having proguard enabled. 

Though having the baseline function would probably stop people from creating issues and indicating that it's a defect of Xamarin Forms

### Steps to Reproduce
- Create a new Xamarin Forms Mobile App.
- Under Android options, enabled either "ProGuard" 
- Set Linker to Link SDK Assemblies
- Deploy to a **device.**

### Expected Behavior
App to launch successfully.

### Actual Behavior
Crashes on launch.

### Basic Information

- Version with issue: 3.0.0.446417
- Last known good version: 2.5.1.527436
- IDE: Visual Studio Enterprise 2017 v15.7.1
- Platform Target Frameworks:
  - Android: 8.1

.NET Framework 4.7.2
Android SDK Platform-Tools 27.0.1
Android SDK Build-Tools 27.0.3
Proguard 6.0.3
JDK 8 Update 171
Android NDK R17

### Reproduction Link

[App1.zip](https://github.com/xamarin/Xamarin.Forms/files/1994395/App1.zip)

### RELATED Tickets
https://github.com/xamarin/Xamarin.Forms/issues/2671
https://github.com/xamarin/Xamarin.Forms/issues/2491
https://github.com/xamarin/Xamarin.Forms/issues/2457
https://github.com/xamarin/Xamarin.Forms/issues/2495
#5742

jonathanpeppers:
The way Xamarin.Android works, it creates a rule for any class that generates a Java [ACW](https://developer.xamarin.com/guides/android/advanced_topics/java-integration/android_callable_wrappers/). This is so that any C# type that extends a Java type won't need a proguard rule.

Since the failure here is `android.support.v7.widget.FitWindowsFrameLayout`, Xamarin.Forms is probably not subclassing it, but just using it.

[...]

PR #4003, opened 2018-10-05
[Android] Implement fixed mode for Bottom Navigation Bar Android

### LINKED ISSUE #3083: [Enhancement] Implement fixed mode for Bottom Navigation Bar Android
### Description
Currently if you add more then 3 icons to the bottom tab bar on android it will use an animated shifting mode to display all the tabs.  

Support libraries v28 surface a new setting on BottomNavigationView 
LabelVisibilityMode

Once we take a dependency on v28 with forms then we should surface that enumeration for the platform specifics.

### API Changes

Add an android platform specific api for
```C#
public enum BottomToolBarLabelVisibilityMode
{
      Default;
      Auto ;
      Labeled 
      Selected 
      Unlabeled 
}
SetBottomToolBarLabelVisibilityMode(BottomToolBarLabelVisibilityMode labelVisibilityMode)
SetBottomToolBarItemHorizontalTranslationEnabled(bool)
```
When implementing these settings make sure to just ignore these if they haven't been set and to retain the default if changed

### Notes
https://github.com/xamarin/Xamarin.Forms/issues/1675#issuecomment-394419356
https://montemagno.com/remove-shifting-bottomnavigationview-android/

Viridovics:
### Description of Change ###

Add platform specific api:
```
On<Android>().SetBottomToolbarShiftMode(ToolbarShiftMode.Fixed, ToolbarShiftMode.Fixed);
```

### Issues Resolved ### 

- fixes #3083

### API Changes ###
Added:
 - ToolbarShiftMode GetToolbarShiftMode(BindableObject element)
 - ToolbarShiftMode GetToolbarShiftMode(this IPlatformElementConfiguration<Android, FormsElement> config)
 - ToolbarShiftMode GetToolbarItemShiftMode(BindableObject element)
 - ToolbarShiftMode GetToolbarItemShiftMode(this IPlatformElementConfiguration<Android, FormsElement> config)
 - void SetBottomToolbarShiftMode(BindableObject element, ToolbarShiftMode shiftMode, ToolbarShiftMode itemShiftMode)
 - IPlatformElementConfiguration<Android, FormsElement> SetBottomToolbarShiftMode(this IPlatformElementConfiguration<Android, FormsElement> config, ToolbarShiftMode toolbarShiftMode, ToolbarShiftMode toolbarItemShiftMode)
 
### Platforms Affected ### 
- Android

Not applicable

### Testing Procedure ###
Ui test `BottomNavigationShiftModes`

### PR Checklist ###

- [ ] Has automated tests 


Viridovics:
I'm not sure about this try/catch:
```
                try
                {
                 ...
		}
		catch(Exception e)
		{
			System.Diagnostics.Debug.WriteLine($"Unable to set shift mode: {e}");
		}
```

AmrAlSayed0:
According to this [StackOverflow question](https://stackoverflow.com/questions/40176244/), using reflection will cause problems when ProGaurd is enabled. It will strip the internal non-public fields of the BottomNavigationView and this fix will not work. If this gets merged, this should be atleast documented somewhere.

Viridovics:
Thank you @AmrAlSayed0 !
I think that ProGuard problem can be resolved in #2709

Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs
```diff
@@ -633,6 +640,42 @@ void UpdateBottomNavigationViewIcons()
...
+				return;
+			var bottomNavigationView = _bottomNavigationView;
+
+			try
+			{
+				var menuView = bottomNavigationView.GetChildAt(0) as BottomNavigationMenuView;
+				if (menuView == null)
+					return;
+
+				var shiftMode = menuView.Class.GetDeclaredField("mShiftingMode");
```
rmarinho: What if this was obfuscated by Proguard? we might need to add some rules to it.

Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs
```diff
@@ -633,6 +640,42 @@ void UpdateBottomNavigationViewIcons()
...
+				shiftMode.Accessible = true;
```
rmarinho: I do we set true, and then False?

Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs
```diff
@@ -633,6 +640,42 @@ void UpdateBottomNavigationViewIcons()
...
+				shiftMode.SetBoolean(menuView, IsEnableShiftMode);
+				shiftMode.Accessible = false;
+				shiftMode.Dispose();
```
rmarinho: Do we really need to call Dispose?

Xamarin.Forms.Core/PlatformConfiguration/AndroidSpecific/TabbedPage.cs
```diff
@@ -150,6 +150,48 @@ public static int GetMaxItemCount(this IPlatformElementConfiguration<Android, Fo
 			return GetMaxItemCount(config.Element);
 		}
 
+		public static readonly BindableProperty ToolbarShiftModeProperty =
```
rmarinho: I wonder if we can have the same property for both, the TabbedPage and the Items, instead of having 2 different properties with different naming.

Maybe add ShiftMode to Page ?

Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs
PureWeen: It's just the whacky way you have to do it
https://stackoverflow.com/questions/40176244/how-to-disable-bottomnavigationview-shift-mode

Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs
PureWeen: One thing to note about this is that with API 28 google is grey listing reflection calls

https://developer.android.com/about/versions/pie/restrictions-non-sdk-interfaces

We have an issue to look at this
https://github.com/xamarin/Xamarin.Forms/issues/4118

It looks like Version 28 of the support libraries is going to support setting this so we'll need to determine if we want to add a grey listed api like this until we can upgrade Xamarin Forms to the newer support libraries

Viridovics: [LabelVisibilityMode](https://developer.android.com/reference/com/google/android/material/bottomnavigation/LabelVisibilityMode) has different behaviour than api function ```SetBottomToolbarShiftMode```.
And problems that related with reflection (grey list, Proguard, etc) are pretty unpleasant. 

If we want to wait for new api, we will lose this functionality only for api 26,27 (if I correctly understand how Xamarin forms and support libraries work). 

I think we should wait until Xamarin forms will be upgraded to the newer support libraries.

P.S. And after that I will correct review remarks and suggestions.

PureWeen:
I tested this with v28 of the support libraries and the new setting. It looks like that setting will work for what we need. It's interesting because it looks like the default LabelVisibilityMode with support 28 is the non shifting one. Which is annoying that the behavior is going to just change when updating to v28.

LabelVisibilityMode has 4 different available values and we will probably surface this change by surfacing an API for that ENUM into the platform specifics (I'll modify the original ticket)

I've included a screen shot of how it looks using the new setting in v28

image

PureWeen:
FYI @Viridovics we recently added this 

https://github.com/xamarin/Xamarin.Forms/blob/78385f9fc1fc56dc88bd98e73bf9c8f2f2d0a90a/Xamarin.Forms.Platform.Android/Renderers/BottomNavigationViewUtils.cs

If you feel like updating this PR to use that then we can look at getting this reopended and merged

PR #4025, opened 2018-10-07
Upgrading OnIdiom<T>

### LINKED ISSUE #4006: OnIdiom<T> and {OnIdiom} discrepancies.
### Description

The markup extension `{OnIdiom}` had a `Default` property, where the `<OnIdiom x:TypeArgument="" .../>` doesn't. This is, at best, confusing to our users.

MartinZikmund:
### Description of Change ###

The `OnIdiom<T>` was not matching the format of `OnPlatform<T>`. This PR attempts to upgrade `OnIdiom` to include matching functionality and more friendly XAML interface.

### Issues Resolved ### 

- fixes #4006 

### API Changes ###

Added:
- `T OnIdiom<T>.Default { get; set; }`
- `IList<OnIdiomSetting> OnIdiom.Idioms { get; private set; }`

### Platforms Affected ### 

- Core/XAML (all platforms)

The legacy syntax for `OnIdiom` is still supported and there should be no change required for existing apps.

PR adds support for `Default` value, which is on par with `{OnIdiom}` markup extension and supports even the legacy syntax.

The new syntax is very similar to the one provided by `OnPlatform<T>`:

```
<BoxView WidthRequest="100" HeightRequest="100">
	<BoxView.BackgroundColor>
		<OnIdiom x:TypeArguments="Color" Default="Red">
			<OnIdiomSetting Idiom="Phone" Value="Blue" />
			<OnIdiomSetting Idiom="TV, Tablet" Value="Green" />
		</OnIdiom>
	</BoxView.BackgroundColor>
</BoxView>
```

The only disadvantage is that it is not possible to reuse `<On>` as in `OnPlatform<T>`, because both reside in the same namespace. It would be possible to add `Idiom` property to `On` but that could be confusing to users.

Not applicable

### Testing Procedure ###

I have created demonstration of the `Default` value on `Issue4006` page in controls gallery.

Functionality of `OnIdiom<T>` has been covered by tests in `Xamarin.Forms.Core.UnitTests.OnIdiomTests`.

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/OnIdiom.cs
+++ b/Xamarin.Forms.Core/OnIdiom.cs
@@ -1,33 +1,116 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Xamarin.Forms.Xaml;
+
 namespace Xamarin.Forms
 {
+	[ContentProperty("Idioms")]
 	public class OnIdiom<T>
 	{
-		public T Phone { get; set; }
+		readonly Dictionary<TargetIdiom, T> _legacyValues = new Dictionary<TargetIdiom, T>();
 
-		public T Tablet { get; set; }
+		bool _hasDefault;		
+
+		T _default;
 		
-		public T Desktop { get; set; }
 
-		public T TV { get; set; }
+		public OnIdiom()
+		{
+			Idioms = new List<OnIdiomSetting>();
+		}
 
-		public T Watch { get; set; }
+		public T Phone
+		{
+			get => GetLegacyValue(TargetIdiom.Phone);
+			set => SetLegacyValue(TargetIdiom.Phone, value);
+		}
+
+		public T Tablet
+		{
+			get => GetLegacyValue(TargetIdiom.Tablet);
+			set => SetLegacyValue(TargetIdiom.Tablet, value);
+		}
+
+		public T Desktop
+		{
+			get => GetLegacyValue(TargetIdiom.Desktop);
+			set => SetLegacyValue(TargetIdiom.Desktop, value);
+		}
+
+		public T TV
+		{
+			get => GetLegacyValue(TargetIdiom.TV);
+			set => SetLegacyValue(TargetIdiom.TV, value);
+		}
+
+		public T Watch
+		{
+			get => GetLegacyValue(TargetIdiom.Watch);
+			set => SetLegacyValue(TargetIdiom.Watch, value);
+		}
+
+		public T Default
+		{
+			get { return _default; }
+			set
+			{
+				_hasDefault = true;
+				_default = value;
+			}
+		}
+
+		public IList<OnIdiomSetting> Idioms { get; private set; }
+
+#pragma warning disable RECS0108 // Warns about static fields in generic types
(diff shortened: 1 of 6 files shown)


MartinZikmund:
@StephaneDelcroix Is there a particular reason for this? The original syntax with explicit property per idiom is still supported, but I thought it would be meaningful to have both `OnX` expressions have matching features - including specifying same value for multiple platforms.

The use of `Dictionary` for storing the value for "legacy" syntax is useful to simplify the code - I can quickly check if any of the "idiom-specific" properties has been set, while keeping a common setter code.

Xamarin.Forms.Core/OnIdiom.cs
```diff
@@ -1,33 +1,116 @@
...
-					return onIdiom.Phone;
-				case TargetIdiom.Tablet:
-					return onIdiom.Tablet;
-				case TargetIdiom.Desktop:
-					return onIdiom.Desktop;
-				case TargetIdiom.TV:
-					return onIdiom.TV;
-				case TargetIdiom.Watch:
-					return onIdiom.Watch;
+				if (idiomSetting.Idiom?.FirstOrDefault(idiom =>
```
samhouts:
```suggestion
				if (idiomSetting.Idiom?.Find(idiom =>
```

Xamarin.Forms.Core/OnIdiom.cs
```diff
@@ -1,33 +1,116 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
```
samhouts:
```suggestion
```

Xamarin.Forms.Core/OnIdiom.cs
MartinZikmund: Is there a reason for not using `FirstOrDefault`? The type is `IList` which does not have `Find` method, so the only option would be then to manually write a `foreach` search.

MartinZikmund: I rebased too :-)

samhouts: `FirstOrDefault` and `foreach` are far less performant than `Find` or `for`, especially on Android. We avoid Linq as much as possible because of the enumerator performance hit.

MartinZikmund:
I understand this, but this syntax is much more powerful as it allows to define the same value for multiple idioms at the same time, without possible unnecessary duplication, as shown in the sample. With the existing syntax (which is still available) we can only set the values one by one. I find this more convenient and valuable. But if you don't share this view, I can revert it back.

samhouts:
@MartinZikmund Thank you for this PR! Please take a look at #4225 and let us know if it provides the same flexibility you're going for here.

MartinZikmund:
@samhouts The reasons for the new syntax still hold, as it closely mirrors how `OnPlatform` has evolved and offers more flexibility beyond just offering `Default` value, see the previous comment I have posted. I think the unit test I have added in the PR shows the usefulness as well. But anyway, if you don't see the new feature as useful, it is okay too.

PR #4061, opened 2018-10-11
[XamlC] Resolve parameters in nested generics

akihikodaki:
### Description of Change ###

Generics can be nested; an argument for a parameter of a generic instance type can be another generic
 instance type, which has its own parameters. The parameters in a nested argument must be resolved to specialize such a generic instance type, which `ResolveGenericParameters` basically does.

Consider the following types for example:

```C#
class Foo<T>
{
}

class Bar<T>
{
}

class Quux<T> : Foo<Bar<T>>
{
}
```

Now resolve the base type of `Quux<int>`. The old rash implementation of `ResolveGenericParameters` reads arguments of `Foo`, finds `Bar` is a concrete type, and considers there is nothing to do and returns `Foo<Bar<T>>` as a resolved one. But that is not true. The new implementation checks the argument of `Bar`, finds it is a generic parameter, replaces with a concrete type, `int`, and returns `Foo<Bar<int>>`.

### Issues Resolved ### 
I was too lazy to open an issue.

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [X] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Build.Tasks/TypeReferenceExtensions.cs
+++ b/Xamarin.Forms.Build.Tasks/TypeReferenceExtensions.cs
@@ -334,13 +334,10 @@ namespace Xamarin.Forms.Build.Tasks
 			if (genericdeclType == null)
 				return self;
 
-			if (!genericself.GenericArguments.Any(arg => arg.IsGenericParameter))
-				return self;
-
 			List<TypeReference> args = new List<TypeReference>();
 			for (var i = 0; i < genericself.GenericArguments.Count; i++) {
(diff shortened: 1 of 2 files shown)


samhouts:
build

akihikodaki:
b2993dc351b1ff36635621f72d2f6bfd9d28fea4 does https://github.com/xamarin/Xamarin.Forms/pull/4061#pullrequestreview-165287388.

samhouts:
build

kingces95:
Awaiting @StephaneDelcroix's review...

PR #4068, opened 2018-10-12
Correct minor typo

slang25:
Correcting a very small typo.

Given this is an internal class and never invoked including parameter name explicitly (using named parameters), it should be a safe change.


samhouts:
build

PR #4101, opened 2018-10-16
Remove unnecessary ToArray to reduce number of BindableObject[] allocations

hartez:
### Description of Change ###

Every call to ApplyBindings is copying the properties `List` to an array before iterating over it, which can result in many thousands of extra `BindableObject[]` allocations. This change iterates over the `List` directly to avoid the extra allocations.

### API Changes ###

 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/BindableObject.cs
+++ b/Xamarin.Forms.Core/BindableObject.cs
@@ -434,9 +434,8 @@ namespace Xamarin.Forms
 
 		internal void ApplyBindings(bool skipBindingContext, bool fromBindingContextChanged)
 		{
-			var prop = _properties.ToArray();
-			for (int i = 0, propLength = prop.Length; i < propLength; i++) {
-				BindablePropertyContext context = prop [i];
+			for (int i = 0, propLength = _properties.Count; i < propLength; i++) {
+				BindablePropertyContext context = _properties [i];
 				BindingBase binding = context.Binding;
 				if (binding == null)
 					continue;


StephaneDelcroix:
🌶

Xamarin.Forms.Core/BindableObject.cs
```diff
@@ -434,9 +434,8 @@ internal void SetValueCore(BindableProperty property, object value, SetValueFlag
 
 		internal void ApplyBindings(bool skipBindingContext, bool fromBindingContextChanged)
 		{
-			var prop = _properties.ToArray();
-			for (int i = 0, propLength = prop.Length; i < propLength; i++) {
-				BindablePropertyContext context = prop [i];
+			for (int i = 0, propLength = _properties.Count; i < propLength; i++) {
+				BindablePropertyContext context = _properties [i];
```
StephaneDelcroix: there _was_ a reason for this fix to exists. I wish a test for this exists, and will fail 🤞

hartez: Here's where the ToArray was introduced: https://github.com/xamarin/Duplo/commit/08eae560cc#diff-2e3c6aa6af208011b62f789f3b1300f5L309

kingces95: Huh. Well. I guess we'll find out if it was a _good_ reason or a _bad_ reason soon enough... We could capture the array length and throw if it changes during the loop. That might flush out the issue during testing...

mrts:
Thanks for the great fix! Why do you consider this an API change? It seems natural to me to backport this to 3.4.x as originally intended now that it is clear that it was simply a mistake.

PR #4104, opened 2018-10-16
Implementation of method ForceUpdateSize for Cell on macOS

wachs:
### Description of Change ###

Updated the CellRenderer code with the implementation of ForceUpdateSize for Cell on macOS

### Issues Resolved ### 
None

### API Changes ###
None

Changed:
 - protected void WireUpForceUpdateSizeRequested(ICellController cell, NSView nativeCell, NSTableView tableView)

### Platforms Affected ### 
- macOS

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.MacOS/Cells/CellRenderer.cs
+++ b/Xamarin.Forms.Platform.MacOS/Cells/CellRenderer.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using AppKit;
 using CoreGraphics;
+using Foundation;
 
 namespace Xamarin.Forms.Platform.MacOS
 {
@@ -45,7 +46,15 @@ namespace Xamarin.Forms.Platform.MacOS
 
 			_onForceUpdateSizeRequested = (sender, e) =>
 			{
-				//TODO: Implement ForceUpdateSize
+                var index = tableView?.RowForView(nativeCell);
+                if (index != null)
+                {
+                    NSAnimationContext.BeginGrouping();
+                    NSAnimationContext.CurrentContext.Duration = 0;
+                    var indexSetRow = NSIndexSet.FromIndex(index.Value);
+                    tableView.NoteHeightOfRowsWithIndexesChanged(indexSetRow);
+                    NSAnimationContext.EndGrouping();
+                }
 			};
 
 			cell.ForceUpdateSizeRequested += _onForceUpdateSizeRequested;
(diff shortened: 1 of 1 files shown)


rmarinho:
build

Xamarin.Forms.Platform.MacOS/Cells/CellRenderer.cs
```diff
@@ -45,7 +46,15 @@ protected void WireUpForceUpdateSizeRequested(ICellController cell, NSView nativ
 
 			_onForceUpdateSizeRequested = (sender, e) =>
 			{
-				//TODO: Implement ForceUpdateSize
+                var index = tableView?.RowForView(nativeCell);
```
rmarinho: we will need to fix the styling, we use tabs.

Xamarin.Forms.Platform.MacOS/Cells/CellRenderer.cs
```diff
@@ -45,7 +46,15 @@ protected void WireUpForceUpdateSizeRequested(ICellController cell, NSView nativ
...
+                if (index != null)
+                {
+                    NSAnimationContext.BeginGrouping();
```
rmarinho: Why do we need this ?

rmarinho:
I have tried the PR and it does do what is supposed to, one could use ForceUpdateSize gallery to test.

I have a found a issue that might be or not related with this PR but will need more investigation. 

When you go to the ForceUpdateSize Gallery Page -> ViewCell, then you call ForceUpdate on the cells and while they are updating you navigate back from that page, you will get a crash.

Xamarin.Forms.Platform.MacOS/Cells/CellRenderer.cs
wachs: For NSView-based tables, this method will animate. To remove the animation, according to Apple's documentation, we need to set the duration to 0 in the current context.

https://developer.apple.com/documentation/appkit/nstableview/1532126-noteheightofrows

Xamarin.Forms.Platform.MacOS/Cells/CellRenderer.cs
wachs: Fixed!

wachs:
I tested the same Gallery in master branch and had the same problem. Would you like me to open another issue to fix this?

rmarinho:
@wachs let's open a issue fot that please so we don't forget. 

Yeah I run on master and it also crashed for me.

wachs:
Perfect @rmarinho!
I created the teh issue: https://github.com/xamarin/Xamarin.Forms/issues/4125

PR #4112, opened 2018-10-17
[GTK] Fix several memory leaks in the GTK backend

ylatuya:
### Description of Change ###

This serie of patches fixes several memory leaks in the GTK backend.
~~The first one is the MasterDetailPage that is not cleaning up old pages when the Detail or Menu page is changed.~~

Another issue is how disposal of Gtk objects is handled, wich is now incorrectly done using the Dispose function. Gtk objects should be disposed using the Destroy function that will automatically iterate over the children and destroy them too (https://developer.gnome.org/gtk2/stable/GtkWidget.html#gtk-widget-destroy).

The gtk-sharp bindings discourage the use of Dispose and don't even call the base class, leaving it without effect (https://github.com/mono/gtk-sharp/blob/gtk-sharp-2-12-branch/gtk/Object.custom#L98), so that's why Gtk objects should override the Destroy function. Problably the gtk-sharp bindings should be changed at some point to have a more idiomatic pattern for disposal, but that would break backwards compatibility, so right now the only possible solution is to use Destroy. 

I created a demo application with a MaterDetail that calls the GC on each Detail change. As you can see the number of live objects in master keeps growing while in this branch it's mostly steady going up and down.

Before
screen shot 2018-10-17 at 11 26 54

After
screen shot 2018-10-17 at 11 25 13

### API Changes ###
 None

### Platforms Affected ### 
- Gtk

None

Not applicable

### Testing Procedure ###

Create a sample MasterDetail application with 2 pages and some content inside the pages.
Call a full garbage collection after switching pages that also waits for pending finalizers.
Launch the application in the profiler with automatic snapshots on garbage collection.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.GTK/Renderers/MasterDetailPageRenderer.cs
+++ b/Xamarin.Forms.Platform.GTK/Renderers/MasterDetailPageRenderer.cs
@@ -10,6 +10,9 @@ namespace Xamarin.Forms.Platform.GTK.Renderers
 {
 	public class MasterDetailPageRenderer : AbstractPageRenderer<Controls.MasterDetailPage, MasterDetailPage>
 	{
+		Page _currentMaster;
+		Page _currentDetail;
+
 		public MasterDetailPageRenderer()
 		{
 			MessagingCenter.Subscribe(this, Forms.BarTextColor, (NavigationPage sender, Color color) =>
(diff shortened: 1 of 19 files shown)


rmarinho:
needs rebase ?

ylatuya:
I pushed the wrong branch, fixing it

ylatuya:
Regarding the MasterDetailPage leak, I am not so sure if the solution is the correct one. Should the MasterDetailPage be responsible of disposing the old Detail or should the application be responsible of doing so? I am think of scenarios in which the application would want to reuse the Page and just switch them using the Detail property.

ylatuya:
I just removed the disposal of the Detail page from the MasterDetailPage since it shouldn't be it's responsibility.

ylatuya:
@jsuarezruiz  Looking that reset of components, SearchEntry is using Destroy with the changes in this PR, GLWidget was already using Destroy, so everything should beok

Xamarin.Forms.Platform.GTK/Renderers/MasterDetailPageRenderer.cs
```diff
@@ -10,6 +10,9 @@ namespace Xamarin.Forms.Platform.GTK.Renderers
 {
 	public class MasterDetailPageRenderer : AbstractPageRenderer<Controls.MasterDetailPage, MasterDetailPage>
 	{
+		Page currentMaster;
+		Page currentDetail;
```
PureWeen: Prefix these two with _

PureWeen:
@ylatuya  can you rebase this to master?  There were some errors that got checked into master a while ago and your branch doesn't have the fix so the build is currently failing with 

```
Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Issue3979.xaml.cs(32,4): Error CS0103: The name 'TargetSpan' does not exist in the current context
Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Issue3979.xaml.cs(32,4): Error CS0103: The name 'TargetSpan' does not exist in the current context
Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Issue3979.xaml.cs(32,4): Error CS0103: The name 'TargetSpan' does not exist in the current context
Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Issue3979.xaml.cs(32,4): Error CS0103: The name 'TargetSpan' does not exist in the current context
```

ylatuya:
Rebase and comments processed

Xamarin.Forms.Platform.GTK/Renderers/MasterDetailPageRenderer.cs
ylatuya: Done

PureWeen:
build

PR #4124, opened 2018-10-18
[UWP] fix performance on UWP ListView if ItemsSource is not an ObservableCollection

### LINKED ISSUE #4059: Bad performance on UWP ListView if ItemsSource is not an ObservableCollection
### Description

In my application, I have a ListView with an ItemsSource that implements IEnumerable, IList, ICollection and INotifyCollectionChanged (it is a custom collection).
By default, my list contains only null elements, but on the first access to an element, it is loaded asynchronously (from an sqlite database) and I use INotifyCollectionChanged to notify the load once it has been done.
It works fine on Android and only items displayed on the screen are loaded. It worked well on UWP with Xamarin Forms 2.5.
With Xamarin Forms 3, on UWP, since my list is not an ObservableCollection, the ListViewRenderer creates its own ObservableCollection and accesses all the elements of my source list to copy them causing asynchronous loading and a big performance problem.

### Steps to Reproduce

1. Create a custom collection that implement IEnumerable, IList, IEmptyList, ICollection and INotifyCollectionChanged.
2. Create a collection of this type with 1000 items.
2. Use it as ListView ItemsSource on a page and run the project on UWP

### Expected Behavior

The ListView should only access visible items (or a little more) and it should not copy the list because it implements all necessary interfaces.

### Actual Behavior

The ListView reads the entire collection.

### Basic Information

- Version with issue: 3.2.0.839982
- Last known good version: 2.5
- IDE: Visual Studio 2017
- Platform Target Frameworks: 
  - UWP

paymicro:
### Description of Change ###

If the ItemSource ListView implements the IList collection and INotifyCollectionChanged, then this collection does not wrap into an ObservableCollection.

### Issues Resolved ### 

- fixes #4059 

### API Changes ###

 None

### Platforms Affected ### 

- UWP

None

**Before**
screenshot_2

**After**
screenshot_1

### Testing Procedure ###

- run UI test 3273
- check that less than 1000 items were loaded in the right list
- items can be drag&drop in the right list
- when scrolling the list, more items should be dynamically loaded

### PR Checklist ###

- [ ] Has automated tests 


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3273.cs
```diff
@@ -22,6 +28,128 @@ public class SortableListView : ListView
...
+			}
+
+			public int LoadedItems { get; private set; }
+
+			object SimulateAsyncGetItem(int i)
+			{
+				if (_hash[i] != null)
+					return _hash[i];
+
+				Thread.Sleep(_latency);
```
kingces95: Isn't this blocking the UIThread?

kingces95:
Can you describe the scenario we're optimizing? Why not use the pull to refresh callback to pull data on a non UI thread and on completion swap back to the UI thread to push the new data into an ObservableCollection?

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3273.cs
kingces95: Hm. Would it be fair to say this PR is trying to mitigate the effect of having the UIThread block on network traffic (loading from a slow remote server)? If so then I'd say while adding logic to prevent wrapping into an ObservableCollection may somehow mitigate UIThread blocking I'd suggest swapping to a background thread is a better overall solution. 

Is there a reason why a user couldn't swap to a background thread to do network traffic? Would that approach be slower (or impossible) in some cases and so justify using the optimization this PR is providing?

paymicro: I think the implementation of collection in this test is not so important. The fact is that you only get items from the user collection when they are needed.
Latency has been added to bring the test to real user conditions.

paymicro:
@kingces95 test has been simplified and separated from the old test. I think it will be more clear and understandable.

Xamarin.Forms.Platform.UAP/ListViewRenderer.cs
```diff
@@ -102,7 +102,11 @@ void ReloadData()
 			}
 			else
 			{
-				_collectionIsWrapped = !IsObservableCollection(Element.ItemsSource);
+				// if the collection is not implemented `IList`, then 
+				// each change will reset the entire collection (`Clear` action ListProxy class).
+				var isCorrectCustomCollection = Element.ItemsSource is IList && Element.ItemsSource is INotifyCollectionChanged;
+
+				_collectionIsWrapped = !(isCorrectCustomCollection || IsObservableCollection());
```
hartez: Since `ObservableCollection<T>` implements `INotifyCollectionChanged` and `IList` (via `Collection<T>`), I don't think this needs the `IsObservableCollection` check.

paymicro:
@PureWeen I think the bug in #2485 should not be solved with Xamarin Forms since it is inside the UWP. 
I created a project on pure [UWP (download)](https://github.com/xamarin/Xamarin.Forms/files/2607694/UWP_listview_issue.zip) that reproduces this error. Maybe we should give it to the UWP developers.

PureWeen:
Yea :-/
Can you log the issue and reference the project you created here?
https://developercommunity.visualstudio.com

It seems like the native ListView on UWP just does better when operating against an ObservableCollection

Someone else can weigh in but I feel like leaving the layout issue fixed is better way to go and we've basically hit the fine balance of changes we can do

CollectionView is going to deal with all of this much better

paymicro:
@PureWeen i create report in [developercommunity ](https://developercommunity.visualstudio.com/content/problem/390343/uwp-removing-items-from-listview-shows-white-rows.html)

PureWeen:
Awesome thank you @paymicro 

Unless @kingces95  or @hartez  thinks otherwise. I'd say we just close and not merge this PR

oansquer:
Hi. So the performance problem will not be solved? I am stuck in version 2 because of this regression.
If you do not want to change the current behavior, can you make it possible for me to do it through properties or class inheritance?

PR #4132, opened 2018-10-18
[Android] Add index and null checks to prevent exception when NavigationPage has no root

### LINKED ISSUE #4069: ArgumentOutOfRangeException in Android NavigationPageRenderer
### Description
Our crash reports contain an unhandled `ArgumentOutOfRangeException` in the Android `NavigationPageRenderer`. Unfortunately I could never reproduce the crash myself.

[Stacktrace.txt](https://github.com/xamarin/Xamarin.Forms/files/2472795/Stacktrace.txt)

I guess this exception occurs here in [line 621](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs#L621) because there is no check if the collection is empty. My suggestion is to use `FirstOrDefault()` with a null check.

### Steps to Reproduce
Unknown

### Expected Behavior
No unhandled exception causing a crash.

### Actual Behavior
App crashes.

### Basic Information

- Version with issue: 3.2.0.839982
- IDE: Visual Studio 2017
- Platform Target Frameworks:
  - Android: 7.0
- Affected Devices: Multiple vendors and devices, Android 6 - 8

hartez:
### Description of Change ###

If a NavigationPage is created using the parameterless constructor and does have a root page pushed onto the stack before the NavigationPageRenderer is attached to the window, the NavigationPageRenderer will throw an ArgumentOutOfRangeException while attempting to update the toolbar.

This change adds checks to prevent that crash and subsequent null checks for other root-page-dependent methods.

### Issues Resolved ### 

- fixes #4069 

### API Changes ###

 None

### Platforms Affected ### 

- Android

None

Not applicable

### PR Checklist ###

- [x] Has automated tests 


PureWeen:
@hartez  rebase and it looks like the tests are failing on ios

PR #4144, opened 2018-10-19
Adding instructions and UI tests for Bugzilla44461

### LINKED ISSUE #2380: UI test B44461 needs to be automated
- [ ] Test needs to be automated
- [ ] Manual test needs instructions

mikescandy:
### Description of Change ###

Added instructions and UI test for test Bugzilla44461

### Issues Resolved ### 

- fixes #2380

### API Changes ###

 None

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS

None

Not applicable

### PR Checklist ###

- [X] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla44461.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla44461.cs
@@ -26,7 +26,8 @@ namespace Xamarin.Forms.Controls.Issues
 
 			var instructions = new Label
 			{
-				Text = @"Tap the first button (Button0). The button should be aligned with the left side of the screen. If it's not, the test failed."
+				Text = @"Tap the first button (Button0). The button should be aligned with the left side of the screen "
+				+ "and should not move. If it's not, the test failed."
 			};
 
 			grid.Children.Add(instructions);
@@ -68,8 +69,8 @@ namespace Xamarin.Forms.Controls.Issues
 		public void Bugzilla44461Test()
 		{
 			var positions = TapButton(0);
-			Assert.AreEqual(positions.initialPosition, positions.finalPosition);
-			Assert.AreEqual(0, positions.finalPosition.X);
+			Assert.AreEqual(positions.initialPosition.X, positions.finalPosition.X);
+			Assert.LessOrEqual(positions.finalPosition.X, 1);
 			RunningApp.Screenshot ("Button0 is aligned with the left side of the screen");
 		}
 
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla44461.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla44461.cs
@@ -9,7 +9,7 @@ using Xamarin.UITest.Queries;
 using NUnit.Framework;
 #endif
 
-namespace Xamarin.Forms.Controls
+namespace Xamarin.Forms.Controls.Issues
 {
 	[Preserve(AllMembers = true)]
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla44461.cs
```diff
@@ -1,19 +1,36 @@
...
+using Xamarin.Forms.CustomAttributes;
 using Xamarin.Forms.Internals;
 
+#if UITEST
+using Xamarin.UITest;
+using Xamarin.UITest.Queries;
+using NUnit.Framework;
+#endif
+
 namespace Xamarin.Forms.Controls
```
PureWeen:
```suggestion
namespace Xamarin.Forms.Controls.Issues
```

The UI Test are failing because if this namespace isn't *Xamarin.Forms.Controls.Issues* then the startup code doesn't run for NUnit

mikescandy:
Nice...

PureWeen:
@mikescandy  I think those are just UI Tests quirks
I retriggered those two and will check back in a few

PureWeen:
Three engineers are riding in a car.

One is a mechanical engineer, one is an electrical engineer, and one is a computer engineer.

The car breaks down and coasts to the side of the road.

"Hang on," says the mechanical engineer. "The problem is probably the engine, let me have a look at it and I'll have us on the road again in no time."

"Wait," says the electrical engineer. "The way it just stopped like that, I think it's the electrical system. Let me have a look and I'll get us going again in a minute or two."

"Hold on," says the computer engineer. "Why don't we all just get out of the car and get in again, and then see if it starts?"

andreinitescu:
@PureWeen 

> CSS uses IBorderElement and the renderers use IBorderController.

I see the `IBorderElement` interface is only used here, in `AssemblyInfo.cs`:

       [assembly: StyleProperty("border-color", typeof(IBorderElement), nameof(BorderElement.BorderColorProperty))]

but who uses the `BorderColor` getter defined in the interface?

	interface IBorderElement
	{
		//note to implementor: implement this property publicly
		Color BorderColor { get; }

		//note to implementor: but implement the methods explicitly
		void OnBorderColorPropertyChanged(Color oldValue, Color newValue);
	}

Looking at `StylePropertyAttribute`, it seems to me the stylesheet mechanism uses the given bindable property (last parameter in the StyleProperty c-tor)

		public StylePropertyAttribute(string cssPropertyName, Type targetType, string bindablePropertyName)
		{
			CssPropertyName = cssPropertyName;
			BindablePropertyName = bindablePropertyName;
			TargetType = targetType;
		}

Having the stylesheet mechanism use the bindable property makes sense, because the value defined in a style sheet need to be set on the control, but then how's the BorderColor getter defined in the `IBorderElement` interface used? 
As a test, I actually removed the property and all XF code compiled! Then maybe it's used somehow by reflection, but where?

Also, I think `OnBorderColorPropertyChanged` can be just moved to `IBorderController`, it's really not related to stylesheet mechanism I think.

> Maybe once we start using Default Interface methods we can weave those cross cutting concerns into the classes better

That C# 8 feature apparently will only be in .NET Core, see this (https://github.com/dotnet/csharplang/issues/52#issuecomment-420023496)
So it will be a (very) long time until it's available to Xamarin Forms.

> If we did do a ButtonBase really the only things I would want to put in there would be around Commanding and things that make a button a button. It can be pressed and it can send commands. Anything stylized would seem to be a bit too opinionated.

Yes, ButtonBase is what a "button" is and it has its implementation. Classes deriving from it (Button, CheckBox, etc.) automatically get the support for command and the fact it can be pressed.
Albeit thin, it's the foundation for extensibility. If someone wants to create his new "button", it shouldn't require to implement all those properties and methods.

andreinitescu:
Sorry for posting comment here, wrong place, I got distracted by @PureWeen 's joke 😆

PR #4157, opened 2018-10-20
[Android] Crash at Dependency service call possible fix

### LINKED ISSUE #4097: (Android) Crash at Dependency service call
Checking with android devices. In offline, on kill and relaunch app was crashing and the error was below

"Index was outside the bounds of the array. at System.Collections.Generic.Dictionary`2[TKey,TValue].TryInsert (TKey key, TValue value, System.Collections.Generic.InsertionBehavior behavior) [0x00147] in <cd3db6fb7af04deba6ed8e11978795f3>:0 at System.Collections.Generic.Dictionary`2[TKey,TValue].set_Item (TKey key, TValue value) [0x00000] in <cd3db6fb7af04deba6ed8e11978795f3>:0 at Xamarin.Forms.DependencyService.Get[T] (Xamarin.Forms.DependencyFetchTarget fetchTarget)"

### Steps to Reproduce
1. Use dependency service in app
2. Go to offline
3. Use the app, kill and relaunch the app

### Expected Behavior
App should work properly

### Actual Behavior
Crash is happening ( 70 to 80% times)

- Version with issue: 3.0.0.561731
- Last known good version: 3.0.0.561731
- IDE:visual studio
- Platform Target Frameworks: 
 - Android: all versions above 5.0
- Nuget Packages: Xamarin Forms 3.0.0.561731
- Affected Devices: Android

AndreiMisiukevich:
### Description of Change ###

Added lock sections into DependencyService class

### Issues Resolved ### 

- fixes #4097 

https://github.com/xamarin/Xamarin.Forms/issues/4097#issuecomment-431542383

### API Changes ###

Added lock sections for accessing dictionary

### Platforms Affected ### 
- Core

Hope it will help to prevent crashing

Not applicable

### Testing Procedure ###
I didn't have "luck" to reproduce that issue, but hope it will help


Xamarin.Forms.Core/DependencyService.cs
```diff
@@ -24,15 +25,18 @@ public static class DependencyService
 		{
 			Initialize();
 
-			Type targetType = typeof(T);
-
-			if (!DependencyImplementations.ContainsKey(targetType))
+			DependencyData dependencyImplementation;
+			lock (Locker)
```
PureWeen: Instead of creating a new type we can probably just lock on *DependencyImplementations* since it's private

Precedence
https://github.com/xamarin/Xamarin.Forms/blob/bd31e1e9fc8b2f9ad94cc99e0c7ab058174821f3/Xamarin.Forms.Core/Interactivity/AttachedCollection.cs#L64

Xamarin.Forms.Core/DependencyService.cs
```diff
@@ -61,7 +65,10 @@ public static class DependencyService
 			if (!DependencyTypes.Contains(targetType))
 				DependencyTypes.Add(targetType);
 
-			DependencyImplementations[targetType] = new DependencyData { ImplementorType = implementorType };
+			lock (Locker)
+			{
```
PureWeen: Remove brackets

Xamarin.Forms.Core/DependencyService.cs
hartez: Locking on an object created for the purpose is clearer and easier for maintenance.

PureWeen: That's fine (and I agree). I mainly suggested it because of how *SynchronizedList* and *AttachedCollection* lock on the internal lists so it was more a suggestion of consistency. That being said there are a couple places that use static objects which kills my argument.

PureWeen:
> I didn't have "luck" to reproduce that issue, but hope it will help

Did you try firing off multiple background tasks to access the service? 

Like you could create a bunch of tasks that will run on the thread pool then do something like

```C#
await Tasks.WhenAll(Tasks).ConfigureAwait(false)
```

Xamarin.Forms.Core/DependencyService.cs
hartez: What's wrong with the braces?

AndreiMisiukevich:
> > I didn't have "luck" to reproduce that issue, but hope it will help
> 
> Did you try firing off multiple background tasks to access the service?
> 
> Like you could create a bunch of tasks that will run on the thread pool then do something like
> 
> ```cs
> await Tasks.WhenAll(Tasks).ConfigureAwait(false)
> ```

Yes, i tried to run several tasks for resolving the same service. 
Ok, i will try again with ConfigureAwait(false)

Xamarin.Forms.Core/DependencyService.cs
AndreiMisiukevich: @PureWeen 
I can remove it, sure. But this file uses "wrong" code style in other places as well :) 
image

PureWeen: I was always told to remove them per our coding conventions (I think @rmarinho was poking me about it) but if that's not our convention then we can just leave them

Technically the guideline is arbitrary

> We use Allman style braces, where each brace begins on a new line. A single line statement block can go without braces but the block must be properly indented on its own line and must not be nested in other statement blocks that use braces (See issue 381 for examples). One exception is that a using statement is permitted to be nested within another using statement by starting on the following line at the same indentation level, even if the nested using contains a controlled block.

So I'll just stop harassing people about it

PureWeen:
the existing ui tests all passed github is just reporting them wrong
FYI

AndreiMisiukevich:
@PureWeen could you please clarify regarding to convention, how should i set names for static fields?
Somewhere i see **s_name**, somewhere **_name** and **Name** too.

Thanks

PureWeen:
Which I realize is kind of silly since even in that file you have two conventions

```C#
static bool s_initialized;
		static readonly List<Type> DependencyTypes = new List<Type>();
		static readonly Dictionary<Type, DependencyData> DependencyImplementations = new Dictionary<Type, DependencyData>();
```

AndreiMisiukevich:
@PureWeen @hartez i still cannot reproduce the crash
everything works fine with and without lock-section (but i tried on simulator).

I've pushed current state, maybe you have any ides...

AndreiMisiukevich:
I'm going to reproduce crash with new another scenario
I will try to register the same type in multiple tasks

AndreiMisiukevich:
Unfortunately, i can't reproduce it on GenyMotion simulator (Google Nexus 5, Android 6.0.0.)

PureWeen:
The difficulty to reproduce makes sense and it seems like the locks resolved the issue for the original issue creator. I know it's not ideal but it seems like a tricky edge case to reproduce. Any objections to merging if all UI Tests pass @hartez  ?

AndreiMisiukevich:
@PureWeen @hartez hey guys, i wanted to check it on real phone, but with latest update of Visual Studio and Android sdk I started to get next error

image

Unfortunately i'm not able to test android at all until realize how to solve the error.
*I use VS for Mac*

Maybe it will be reproduced, if check issue with my test case and original service code.

Thx

PureWeen:
some things to try

- git clean -xdf
- delete bin/objs
- delete all the contents of Resource.designer.cs in the head project and rebuild

AndreiMisiukevich:
@PureWeen thanks, *git clean -xdf* solved my issue.
Tried my test case on android 6.0 Xiaomi Redmi 4 - works fine, no crash..

PR #4181, opened 2018-10-23
[UWP] Changing IsPassword property changed InputScope

### LINKED ISSUE #4093: When changing IsPassword property of an entry in runtime into true, the keyboard setting of entry is changed from numeric to default
### Description
My requirement is to use Numeric keyboard for an entry which IsPassword property is changed into true at run time through a button click but when the button is clicked, the keyboard setting is changed to default only in UWP platform
### Steps to Reproduce

1. Set keyboard as numeric for an entry
2. Through button click changes **IsPassword** property of entry into true
3. The keyboard is changed into default in UWP

### Expected Behavior
The Keyboard do not change from numeric after changing IsPassword into true
### Actual Behavior
The Keyboard  changes from numeric into default after changing IsPassword into true

### Screenshots
image
After changing **IsPassword** into true
image

### Reproduction Link

[DataForm_Entry.zip](https://github.com/xamarin/Xamarin.Forms/files/2481738/DataForm_Entry.zip)

jfversluis:
### Description of Change ###

Removed touching the InputScope whenever the IsPassword property is toggled. This was done before to prevent dictionary suggestions, now it caused the keyboard to be reset to default while it should be numeric or similar.

### Issues Resolved ### 
- fixes #4093

### API Changes ###
None

### Platforms Affected ### 

- UWP

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 


PureWeen:
@jfversluis 
```
Xamarin.Forms.Platform.UAP\FormsTextBox.cs(45,8): Error CS0649: Field 'FormsTextBox._cachedPredictionsSetting' is never assigned to, and will always have its default value false
```

jfversluis:
Removed a bit too much, should be good now

Xamarin.Forms.Platform.UAP/FormsTextBox.cs
```diff
@@ -337,16 +318,13 @@ void UpdateInputScope()
 		{
 			if (IsPassword)
 			{
-				_cachedInputScope = InputScope;
 				_cachedSpellCheckSetting = IsSpellCheckEnabled;
 				_cachedPredictionsSetting = IsTextPredictionEnabled;
-				InputScope = PasswordInputScope; // Change to default input scope so we don't have suggestions, etc.
```
kingces95: Hm, as the comment says, the PasswordInputScope has a purpose. If we remove it, then whoever we added for is gonna be mad. So... we we'll need to devise a way to fix this bug without removing the PasswordInputScope.

jfversluis: Looking at #4093 it seems the original author doesn't know and agrees that it seems kind of redundant with all the other disabling of text prediction going on. I don't really see any point in doing this anymore

kingces95: Nice. Will continue this discussion on that thread...

kingces95:
Waiting for v2 of changes that doesn't remove the scope (and the nice comment about not auto-completing passwords).

kingces95:
Just closing out for accounting reasons. We'd still like to merge this change. Please re-open once you've implemented the change without removing the scope.

jfversluis:
Added new code for v2. I can't seem to reopen this myself though

Xamarin.Forms.Platform.UAP/FormsTextBox.cs
```diff
@@ -340,7 +359,16 @@ void UpdateInputScope()
 				_cachedInputScope = InputScope;
 				_cachedSpellCheckSetting = IsSpellCheckEnabled;
 				_cachedPredictionsSetting = IsTextPredictionEnabled;
-				InputScope = PasswordInputScope; // Change to default input scope so we don't have suggestions, etc.
+
+				if (InputScope.Names.Any(i => i.NameValue == InputScopeNameValue.Number))
```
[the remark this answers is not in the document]
jfversluis: Should be fixed :)

Xamarin.Forms.Controls/CoreGalleryPages/EntryCoreGalleryPage.cs
```diff
@@ -66,6 +66,26 @@ protected override void Build (StackLayout stackLayout)
 
 			var maxLengthContainer = new ViewContainer<Entry>(Test.Entry.MaxLength,	new Entry { MaxLength = 3 });
 
+			var isPasswordInputScopeContainer = new ViewContainer<Entry>(Test.Entry.IsPasswordNumeric,	new Entry { Keyboard = Keyboard.Numeric });
```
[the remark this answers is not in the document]
jfversluis: Should be fixed :)

PR #4229, opened 2018-10-26
[iOS] fix for SearchBar background color

### LINKED ISSUE #4194: SearchBar changes color if initial state is invisible
### Steps to Reproduce

1. I add 2 search bars with the same color. One is visible, second is not:
  ```
<SearchBar Text="Visible" BackgroundColor="Lime" Margin="0, 100, 0, 0"></SearchBar>
<SearchBar x:Name="SearchBar1" IsVisible="false" BackgroundColor="Lime"></SearchBar>
<Button Text="Show" Clicked="Handle_Clicked" HorizontalOptions="Center" VerticalOptions="CenterAndExpand"></Button>

```
2. Make second visible on button click:
```
async void Handle_Clicked(object sender, System.EventArgs e)
{
    SearchBar1.IsVisible = true;
}

```
3. As a result I have the second search bar in dark green [https://stackoverflow.com/questions/52933166/searchbar-changes-color-if-initial-state-is-invisible](https://stackoverflow.com/questions/52933166/searchbar-changes-color-if-initial-state-is-invisible)
The issue is only appears on iOS ,Android works fine.

### Basic Information

- Version with issue:
- Last known good version:
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Nuget Packages:
- Affected Devices:

### PR DESCRIPTION
**Pull request by andreinitescu**
### Description of Change ###

On iOS, if a `SearchBar` is created with `IsVisible="false"` and a `BackgroundColor`, after it's made later made visible by setting `IsVisible` to `true`, the background color is darkened.

What's interesting is that if the `SearchBar` is put right as a child of the `Page`, the issue doesn't appear. Issue only appears when `SearchBar` is inside a container like `Grid` or `StackLayout`

This fix makes ensures the background image is always cleared so that the color set by BarTintColor is always reflected in the `SearchBar` look. Per my investigation, without always setting the background image, it looks like there are some issues sometimes with UISearchBar. Not sure if it's completely related but I stumbled on [this](https://stackoverflow.com/questions/42477779/uisearchbar-briefly-displays-gray-background-when-returning-to-title-from-a-show)

### Issues Resolved ### 

- fixes #4194 

### API Changes ###

None

### Platforms Affected ### 

- iOS

See description and screenshots

Before:
image

After:
image

### Testing Procedure ###

I added a test page for this issue with instructions.

image

### PR Checklist ###

- [ ] Has automated tests 

### CODE REVIEW

**Comment by PureWeen** [comment]
Can you rebase to fix conflict please?

**Comment by andreinitescu** [comment]
@PureWeen  Sure, done!

File: Xamarin.Forms.Platform.iOS/Renderers/SearchBarRenderer.cs [hunk 1]
```diff
@@ -120,8 +120,7 @@ protected override void SetBackgroundColor(Color color)
 			
 			Control.BarTintColor = color.ToUIColor(_defaultTintColor);
 
-			if (color.A < 1)
```
hartez: @rmarinho Do you remember why this `if` statement was here? It came from https://github.com/xamarin/Duplo/pull/3265 and https://github.com/xamarin/Duplo/pull/3304 (fixes for https://bugzilla.xamarin.com/show_bug.cgi?id=31669).

rmarinho: Sorry I don't recall tbh.

andreinitescu: @hartez My take on the logic for the `if` is this: if there's any transparency set on the color, clear the background image so that the tint color actually uses that color with alpha, otherwise it won't work. I think the idea was the clearing the background image is not necessary otherwise, just setting the `BarTintColor` should had worked, but it looks like it doesn't. Not sure if this helps.

rmarinho:
build

PR #4236, opened 2018-10-27
[Xaml] Support correctly custom resource providers

ylatuya:
### Description of Change ###

When having a resources provider set, we should return the resource as soon as possible without having to read the original one from assembly.
This fixes live-reload scenarios in which reloading of code is done creating new types in dynamic assemblies, where GetManifestResourceStream is not available.

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 


kzu:
Makes sense to me :+1:

StephaneDelcroix:
could you please add a test ? thanks

PR #4238, opened 2018-10-27
[XamlC] Always enable implicit casting and boxing and unboxing

akihikodaki:
### Description of Change ###

XamlC often uses single instruction `ldloc` to load a local variable to stack to use as an argument for a method, but the instruction does not do necessary type conversion. For example, the following XAML code does not work because it does not box value type where `collections` is `clr-namespace:System.Collections;assembly=mscorlib`:

```XAML
<collections:List>
  <x:Single>0</x:Single>
</collections:List>
```

Some code emitting `ldloc` emits implicit casting code and boxing/unboxing instructions as well, but such code are duplicated over the codebase and the behavior is inconsistent.

This change introduces an internal method `Xamarin.Forms.Build.Tasks.VariableDefinitionExtensions.LoadAs` rather than duplicating such type conversion code even more after all `ldloc` emission and checking they are all consistent.

It has grown big because of the comprehensive change, so I decided to limit the scope it deals with. Specifically, it does __not__:

* add type conversion after `ldfld`, `call`, `callvirt`,
* validate the type conversion, or
* add code to load pointer of value type for `this` parameter (`ldloca` should be used instead of `ldloc` for such case.)

In future, you may add such features to `Xamarin.Forms.Build.Tasks.VariableDefinitionExtensions.LoadAs` or add a similar method.

### Issues Resolved ### 

I didn't find any issues involved, but maybe there are.

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [X] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Build.Tasks/TypeReferenceExtensions.cs
+++ b/Xamarin.Forms.Build.Tasks/TypeReferenceExtensions.cs
@@ -333,6 +333,38 @@ namespace Xamarin.Forms.Build.Tasks
 			return null;
 		}
 
+		public static TypeReference ResolveGenericParameters(this TypeReference self, MethodReference declaringMethodReference)
+		{
+			var genericself = self as GenericParameter;
+			if (genericself != null) {
+				IGenericInstance instance;
+
+				switch (genericself.Type) {
+				case GenericParameterType.Method:
+					instance = (IGenericInstance)declaringMethodReference;
+					break;
+
+				case GenericParameterType.Type:
+					instance = (IGenericInstance)declaringMethodReference.DeclaringType;
+					break;
+
+				default:
+					throw new Exception("unknown generic parameter type");
+				}
+
+				return instance.GenericArguments[genericself.Position];
+			}
+
+			var genericInstanceSelf = self as GenericInstanceType;
+			if (genericInstanceSelf != null) {
+				var genericArguments = genericInstanceSelf.GenericArguments;
+				var arguments = genericArguments.Select(argument => argument.ResolveGenericParameters(declaringMethodReference));
+				return genericInstanceSelf.ElementType.MakeGenericInstanceType(arguments.ToArray());
+			}
+
+			return self;
+		}
+
 		public static TypeReference ResolveGenericParameters(this TypeReference self, TypeReference declaringTypeReference)
 		{
 			var genericself = self as GenericInstanceType;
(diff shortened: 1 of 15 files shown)


StephaneDelcroix:
Thanks. I'll look at this soon (and your other PRs)

Xamarin.Forms.Build.Tasks/VariableDefinitionExtensions.cs
```diff
@@ -0,0 +1,23 @@
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+using System.Collections.Generic;
+
+namespace Xamarin.Forms.Build.Tasks
+{
+	static class VariableDefinitionExtensions
+	{
+		public static IEnumerable<Instruction> LoadAs(this VariableDefinition self, TypeReference type, ModuleDefinition module)
+		{
```
StephaneDelcroix: I like the idea of having this in a separate extension method, but I'd keep the `LdLoc` out of it so it could be used on any object on the stack.

Xamarin.Forms.Build.Tasks/TypeReferenceExtensions.cs
```diff
@@ -324,6 +324,38 @@ public static MethodReference GetImplicitOperatorTo(this TypeReference fromType,
 			return null;
 		}
 
+		public static TypeReference ResolveGenericParameters(this TypeReference self, MethodReference declaringMethodReference)
```
StephaneDelcroix: don't we have this already ?

akihikodaki: What method do you think can be alternative for this?

Xamarin.Forms.Build.Tasks/VariableDefinitionExtensions.cs
akihikodaki: I agree that the type conversion can be used without `ldloc`, and I also note the possibility in the first comment of this pull request.

However, I decided to create this convenient method because I saw combinations of the type conversion and `ldloc` so frequently. I think we may have a method dedicated for type conversion and call it from this `LoadAs` method, or something for `Ldfld`, or even directly. I have not made such a generic type conversion method yet because I didn't create something like `Xamarin.Forms.Build.Tasks.FieldExtensions.LoadAs` nor find an appropriate situation to call the type conversion method without knowing an instruction to load an object to the stack.

I can make a change accordingly if you still think the type conversion logic should have a dedicated method in this pull request, so please tell me in that case.

akihikodaki:
476e12b25740d308099096ef7b6e513f182bdc47 adds a test with XAML.

StephaneDelcroix:
could you please rebase this on 3.5.0 ? thanks

akihikodaki:
08a8e2117290f1269c8ad0e6c182dc65a83ca563: rebased to `3.5.0`.

PR #4241, opened 2018-10-28
[Android] AppCompat version of EntryRenderer

adrianknight89:
Since there hasn't been any traction with #2151 since March, I wanted to redo this. The code added here is an exact copy of the previous renderer with the inclusion of `FormsAppCompatEditText` in addition. I ran the Entry Gallery and did not see any visible change compared to the previous renderer. Please run UI tests and let me know if this breaks anything.


adrianknight89:
@rmarinho if you want both renderers to share code, we could do that, but it's a design decision. I can't tell to what degree the renderers will diverge in the future if at all.

samhouts:
Could we use the FastRenderer pattern for this?

PR #4248, opened 2018-10-29
[Android] Made HandleKeyboardOnFocus protected to avoid using reflection

### LINKED ISSUE #3647: Two keyboards appear when editing the TextColumn after the NumericColumn with Samsung  Android version 6 devices only on Input type of Samsung Keyboard
### Description

Hi Team, 

We at Syncfusion develop data visualization controls for Xamarin and we have more than 90 controls spread across all platforms of Xamarin. Please refer the below link for a description about us.
UserGuide: https://help.syncfusion.com/xamarin/sfdatagrid/overview
FeatureTour: https://www.syncfusion.com/products/xamarin

Below mentioned issue is only on Samsung  devices with "**Android version 6 Marshmallow and API 23**" with Input type as only on "**Samsung Keyboard**"

We would like to bring to your notice a bug that we encountered recently when editing the first column (Numeric column) alphanumeric Keyboard appears which is fine, after that while editing second column (Text Column), same alpha numeric keyboard appears first, next text keyboard comes to the view which is issue. Issue has replicated on 3.2.0 version too.

Also we have already crated bug report related to this issue. Bug report link : https://github.com/xamarin/Xamarin.Forms/issues/3582, But you have validated wrongly as by deploying Android_6 emulators and stated keyboard working fine. So, Do check on Samsung Android version 6 devices with only **Samsung keyboard**. Also we would like to mention here while giving input type as google keyboard on same Samsung Android version 6 devices keyboard working fine.

### Steps to Reproduce

1.Run the attached sample in Samsung  devices with "**Android version 6 Marshmallow and API 23**" with Input type as only on "**Samsung Keyboard**"
2.Enters in to edit mode by double tabbing
3.Edit any cell in the first column (Numeric column)
4.Then tab any where to collapse the keyboard
5.Now edit the second column (Text column)
6.Able to see first "Alphanumeric keyboard" comes to the view after that "Text keyboard" appears, which is mentioned issue

### Expected Behavior

Only Text keyboard should appear while edit the Text column after editing of Numeric column

### Actual Behavior

After editing Numeric column if editing happens for Text column, first Alphanumeric keyboard comes to the view then Text keyboard appears.

### Basic Information

- Version with issue:  2.4.0.280
- Last known good version: 3.2.0.729530-pre 2
- IDE: 
- Platform Target Frameworks:
  - Android: Android: 6 Marshmallow API 23
- Nuget Packages: 3.2.0.729530-pre 2
- Affected Devices: Android version 6 Marshmallow API 23 devices

### VideoLink

(https://github.com/xamarin/Xamarin.Forms/files/2293076/TwoKeyboardVideo.zip)

### Reproduction Sample link 

[BugReport.zip](https://github.com/xamarin/Xamarin.Forms/files/2317134/BugReport.zip)

Note: In case if SfDatagrid packages are not restored in the sample, you can download the package Syncfusion.Xamarin.SfDataGrid from nuget.org.

adrianknight89:
### Description of Change ###

This is an internal property of `ViewRenderer` which shows/hides the soft keyboard during focus change events. This property is being overridden in derived classes such as `EntryRenderer`, `EditorRenderer`, and `SearchBarRenderer`. These renderers all set `HandleKeyboardOnFocus` to true so that when the element focus is changed programmatically, we can show or hide the keyboard. However, by setting this property to true and allowing `ViewRenderer` to control the soft keyboard, we introduce #3647. I've given more details [here](https://github.com/xamarin/Xamarin.Forms/issues/3647#issuecomment-427539020).

The way I fixed the above issue (for my own use case) is like so:

- create a custom `EntryRenderer`
- set `HandleKeyboardOnFocus` to false in `OnElementChanged` after calling base.
- Handle `IsFocused` changes 

```
{

	if(e.PropertyName == VisualElement.IsFocusedProperty.PropertyName)
	{
		if (!Control.IsFocused)
		        HideKeyboard();
		else
			ShowKeyboard();
	}
}
```

I think a proper fix would have to move the following code into `OnElementPropertyChanged` of `ViewRenderer`:

```
//handles keyboard on focus for Editor, Entry and SearchBar
if (HandleKeyboardOnFocus)
{
	if (e.Focus)
		Control.ShowKeyboard();
	else
		Control.HideKeyboard();
}
```
so that it wouldn't be executed inside the event handler for focus change. The derived types will need to set the property to true before calling base. 

I won't look at @DeepikaBalaSubramaniyan's reproduction, but I hope this can be useful when you find a fix for the issue.

### Issues Resolved ### 

- fixes #3647 (Not really. This is not a proper fix, but it allows us to come up with a workaround without using reflection to modify the internal property).

### API Changes ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/ViewRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/ViewRenderer.cs
@@ -52,7 +52,7 @@ namespace Xamarin.Forms.Platform.Android
 
 		SoftInput _startingInputMode;
 
-		internal bool HandleKeyboardOnFocus;
+		protected bool HandleKeyboardOnFocus;
 
 		public TNativeView Control { get; private set; }
 


hartez:
@adrianknight89 I think you're correct that developers should be able to override this behavior, but I think HandleOnKeyboardFocus was always an awkward way to do this. Take a look at #4384 as an alternative and see what you think.

adrianknight89:
Thanks. I agree that `HandleOnKeyboardFocus` isn't the ideal way to do this. Looking forward to trying out your fix.

Issue #2952, opened 2018-06-06
[iOS] WebView not render/update HTML source after second call


raphibolliger:
### Description
WebView does not render HTML content (except first time call) when BindingContext is changed at the page on iOS. The same example works fine on Android. The page is always the same and BindingContext changes before showing with navigation. For details see the attached gifs.

### Steps to Reproduce
Run the attached WebViewTest. (Actually the project name is UserDialogs, because of an other test.)

1. Press Button "HtmlWebView1" -> HTML shows in WebView properly
2. Press Back
3. Press Button "HtmlWebView2" -> Empty WebView
4. Press Back
5. Select another HtmlWebView -> Always empty
6. All further calls to WebViews will be empty

### Implementation Details
```XAML
<WebView HorizontalOptions="FillAndExpand" VerticalOptions="FillAndExpand"
    <WebView.Source>
        <HtmlWebViewSource Html="{Binding HtmlSource}"></HtmlWebViewSource>
    </WebView.Source>
</WebView>
```
```C#
// html source
const string HTML1 = "<h1>PAGE 1</h1><p>This is Page 1</p>";

// page is always created
_htmlWebView = (Page)Activator.CreateInstance(typeof(HtmlWebView));
...
// setting the binding context when pushing a new view (context will be removed on destroy)
_htmlWebView.BindingContext = new HtmlViewModel(HTML1);
await Navigation.PushAsync(_htmlWebView);
```

### Expected Behavior
It should show the changed HTML source in the WebView. On Android the behaviour is correct. On iOS it works only the first time on a view call.

### Actual Behavior
After a BindingContext change it will never show any HTML content in the WebView on iOS after first call.
### Basic Information

- Version with issue: Xamarin Forms 3.0.0.482510
- Last known good version: never
- IDE: Microsoft Visual Studio Professional 2017 (15.7.3)
- Platform Target Frameworks: 
  - iOS:  11.4
  - Android: 8.1 
- Nuget Packages: Xamarin Forms 3.0.0.482510
- Affected Devices: iPhone

### Screenshots / Videos
![ios1](https://user-images.githubusercontent.com/6289764/41024974-f9067058-6970-11e8-8782-7b516ef8627a.gif)
![ios2](https://user-images.githubusercontent.com/6289764/41024980-fdffe454-6970-11e8-911e-4c8171fa7b4d.gif)
[...]

PR #4254, opened 2018-10-30
[iOS] Add JavaScript dialog delegate to WkWebView

### LINKED ISSUE #4253: Javascript alert dialogs don't work when using WKWebView
### Description
The `alert()`, `confirm()` and `prompt()` Javascript functions normally produce the appropriate native dialogs for user interaction, but don't if the WebView renderer is set to `WkWebViewRenderer`. 

### Steps to Reproduce

1.  Add `[assembly: ExportRenderer(typeof(WebView),typeof(Xamarin.Forms.Platform.iOS.WkWebViewRenderer))]` to **Xamarin.Forms.ControlGallery.iOS**
2. Launch **Xamarin.Forms.ControlGallery.iOS**
3. Navigate to **WebView Gallery -> JavaScriptAlert View**
4. Click the button

### Expected Behavior
A native dialog box should appear as it does when using the default `UIWebView`.

### Actual Behavior
Nothing happens.

### Basic Information

- Version with issue: current master
- IDE: VS 2017 15.8.6
- Platform Target Frameworks:
  - iOS:  12.0

alanag13:
### Description of Change ###
The default renderer for `WebView` on iOS uses `UIWebView`, which handles Javascript-based user dialogs automatically. With `WKWebView`, these dialogs need to be set up manually with a delate.

This will allow users choosing to "upgrade" to `WkWebView` to retain the functionality of any javascript dialogs their web source may be using.

### Issues Resolved ### 
- Fixes #4253

### Platforms Affected ### 
- iOS

Without this change, dialogs created via Javascript in the `WKWebView` would never appear, now they will appear at the same times as they do when using `UIWebView`.

### Testing Procedure ###
Use the repro steps mentioned in #4253 and the updated UI test to ensure the dialogs appear and behave correctly (confirm logs a bool, prompt logs an input string, etc).

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.ControlGallery.iOS/Properties/AssemblyInfo.cs
+++ b/Xamarin.Forms.ControlGallery.iOS/Properties/AssemblyInfo.cs
@@ -42,3 +42,4 @@ using Xamarin.Forms.Controls;
 // Deliberately broken image source and handler so we can test handling of image loading errors
 [assembly: ExportImageSourceHandler(typeof(FailImageSource), typeof(BrokenImageSourceHandler))]
 [assembly: ExportRenderer(typeof(WkWebView), typeof(Xamarin.Forms.Platform.iOS.WkWebViewRenderer))]
+
--- a/Xamarin.Forms.ControlGallery.iOS/Xamarin.Forms.ControlGallery.iOS.csproj
+++ b/Xamarin.Forms.ControlGallery.iOS/Xamarin.Forms.ControlGallery.iOS.csproj
@@ -26,7 +26,8 @@
     <MtouchArch>i386, x86_64</MtouchArch>
     <MtouchLink>SdkOnly</MtouchLink>
     <MtouchDebug>True</MtouchDebug>
-    <CodesignEntitlements>Entitlements.plist</CodesignEntitlements>
+    <CodesignEntitlements>
+    </CodesignEntitlements>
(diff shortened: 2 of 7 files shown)


alanag13:
I'm not sure whether or not the native dialogs that show up are targetable via automation, but I tried to do what I could re: UI tests. Let me know if you have any questions.

alanag13:
Thanks, I like the dedupe changes a lot better 😄

alanag13:
@PureWeen the difference in button size is due to WkWebView not applying a default `DOCTYPE`, and possibly some other things (default zoom level, etc) that `UIWebView` does for you. In short, the difference in button size here is basically any implementation detail of the two WebViews. Most sites developed for the actual web **should** have these set in the html already in order to account for browser differences that happen without it, but what we *could* do is simply add those changes to the gallery code so that the galleries look the same.  Let me know your thoughts.

PureWeen:
@alanag13  if it's easy enough to add to the gallery code to make them look the same I'd say do it. Having that in there I feel would be useful. Plus that button is almost hard to even see small :-)

alanag13:
@PureWeen done. As an aside, I did notice that I can reproduce #2952 by simply flipping through the gallery views that use `HtmlWebViewSource`.

PR #4263, opened 2018-10-30
[iOS] ControlsGallery, shake gesture to navigate to the root page

### LINKED ISSUE #3724: [iOS] Make the 5 click reset control gallery feature of iOS more explicit
### Description

Currently on the control gallery in ios if you tap 5 times quick enough it causes the control gallery to reset. This feature is extremely helpful but can also be very confusing when testing things like gestures or just clicking buttons in general.

It would be useful to slightly modify this feature to lower the probability of the action matching up with testing procedures

Adding a way to disable this feature, pop open a dialog instead of auto restart, or  possibly isolate the clicking to a specific region.

AndreiMisiukevich:
### Description of Change ###
I propose a new approach for navigating to the root page. **ctr**+**cmd**+**z** for simulator (shake gesture)

### Issues Resolved ### 
- fixes #3724 

### API Changes ###
Added:
Shake gesture to navigate to the root page
Removed:
5 taps to navigate to the root page

### Platforms Affected ### 
- iOS sample project

None

Not applicable

### Testing Procedure ###
Start ControlGallery.iOS, go to any screen and try to shake real device or select shake gesture (CTRL + CMD + Z) for simulator.


Xamarin.Forms.ControlGallery.iOS/AppDelegate.cs
```diff
@@ -169,12 +169,13 @@ public override bool FinishedLaunching(UIApplication uiApplication, NSDictionary
 					//	e.NativeView.AccessibilityIdentifier = e.View.StyleId;
 				}
 
-				if (e.NativeView != null)
-				{
-					var view = e.NativeView;
-					var tapGestureRecognizer = new UITapGestureRecognizer(() => Reset("")) { NumberOfTapsRequired = 5 };
-					view.AddGestureRecognizer(tapGestureRecognizer);
-				}
+				// Used shakegesture insted https://github.com/xamarin/Xamarin.Forms/issues/3724
```
PureWeen: just delete this code instead of commenting it out

Xamarin.Forms.ControlGallery.iOS/Main.cs
```diff
@@ -8,7 +8,7 @@ public class Application
 	{
 		static void Main(string[] args)
 		{
-			UIApplication.Main(args, null, "AppDelegate");
+			UIApplication.Main(args, "Application", "AppDelegate");
```
PureWeen: Can we use this instead
```
UIApplication.Main(args, typeof(CustomApplication), typeof(AppDelegate));
```

Just makes me feel better than having strings. Then we can also remove the Register on CustomApplication

Xamarin.Forms.ControlGallery.iOS/AppDelegate.cs
AndreiMisiukevich: done

Xamarin.Forms.ControlGallery.iOS/Main.cs
AndreiMisiukevich: done)

AndreiMisiukevich:
@PureWeen please re-review )

PR #4265, opened 2018-10-31
[Enhancement] Add CollectionView progress behind feature flag

hartez:
### Description of Change ###

Adds progress on #3172 behind the CollectionView_Experimental flag.

### Issues Resolved ### 

- #3172, partially.

### API Changes ###

See #3172, though not all of those changes are complete yet.

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###

- [x] Has automated tests 


Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
```diff
@@ -43,7 +43,9 @@ protected override void OnCreate(Bundle bundle)
 #if TEST_EXPERIMENTAL_RENDERERS
 			Forms.SetFlags("FastRenderers_Experimental");
 #else
-			Forms.SetFlags("Fake_Flag"); // So we can test for flag initialization issues
+			// Fake_Flag is here so we can test for flag initialization issues
+			// CollectionView lets us test CollectionView stuff until it's officially released
+			Forms.SetFlags("Fake_Flag", "CollectionView_Experimental");
```
adrianknight89: Why are these two grouped together as opposed to being separate statements where the latter part goes under `#if TEST_EXPERIMENTAL_RENDERERS`?

Xamarin.Forms.Core/Items/ItemsControl.cs
```diff
@@ -0,0 +1,104 @@
...
+	public class ListItemsLayout : ItemsLayout
+	{
+		// TODO hartez 2018/08/29 17:28:42 Consider changing this name to LinearItemsLayout; not everything using it is a list (e.g., Carousel)	
+		public ListItemsLayout(ItemsLayoutOrientation orientation) : base(orientation)
+		{
+		}
+
+		// TODO hartez 2018/05/31 15:56:23 Should these just be called Vertical and Horizontal (without List)?	
+		public static readonly IItemsLayout VerticalList = new ListItemsLayout(ItemsLayoutOrientation.Vertical); 
+		public static readonly IItemsLayout HorizontalList = new ListItemsLayout(ItemsLayoutOrientation.Horizontal);
```
adrianknight89: Should these be lazy loaded?

Xamarin.Forms.Core/Items/ItemsControl.cs
```diff
@@ -0,0 +1,104 @@
...
+			base(orientation)
+		{
+			Span = span;
+		}
+
+		public override string ToString()
+		{
+			var orientation = Orientation == ItemsLayoutOrientation.Horizontal ? "Horizontal Grid" : "Vertical Grid";
+			
+			return $"{orientation}, {nameof(Span)} {Span}";
```
adrianknight89: Should we put a `:` after `{nameof(Span)}`?

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -0,0 +1,134 @@
...
+			BindableProperty.Create(nameof(EmptyView), typeof(object), typeof(ItemsView), null);
+
+		public object EmptyView
+		{
+			get => GetValue(EmptyViewProperty);
+			set => SetValue(EmptyViewProperty, value);
+		}
+
+		public static readonly BindableProperty EmptyViewTemplateProperty =
+			BindableProperty.Create(nameof(EmptyView), typeof(DataTemplate), typeof(ItemsView), null);
```
adrianknight89: `nameof` should use `EmptyViewTemplate`

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -0,0 +1,134 @@
...
+			};
+		}
+	}
+
+	public class ItemsView : View
+	{
+		// TODO hartez 2018/06/24 11:37:00 Give DisplayMemberPath some thought	
+
+		public static readonly BindableProperty EmptyViewProperty =
+			BindableProperty.Create(nameof(EmptyView), typeof(object), typeof(ItemsView), null);
```
adrianknight89: Why is this an object instead of `View`?

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -0,0 +1,134 @@
...
+			get => (IEnumerable)GetValue(ItemsSourceProperty);
+			set => SetValue(ItemsSourceProperty, value);
+		}
+
+		// TODO hartez 2018/08/29 17:35:10 Should ItemsView be abstract? With ItemsLayout as an interface?
+		// Trying to come up with a reasonable way to restrict CarouselView to ListItemsLayout(LinearLayout) 
+		// ((because setting Carousel to grid is ... weird? And by default it just won't do anything.))
+		// And allow CollectionView to use a broader set of Layout options
+		// So the Bindable property only exists at the CarouselView/CollectionView (i.e., concrete class) level
+		// but some version of IItemsLayout is still here?
```
adrianknight89: This could also be enforced through exceptions to restrict usage, but I'd use it as a last resort.

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -0,0 +1,134 @@
...
+
+		public event EventHandler<ScrollToRequestEventArgs> ScrollToRequested;
+
+		protected override SizeRequest OnMeasure(double widthConstraint, double heightConstraint)
+		{
+			// TODO hartez 2018-05-22 05:04 PM This 40,40 is what LV1 does; can we come up with something less arbitrary?
+			var minimumSize = new Size(40, 40);
+
+			var maxWidth = Math.Min(Device.Info.ScaledScreenSize.Width, widthConstraint);
+			var maxHeight = Math.Min(Device.Info.ScaledScreenSize.Height, heightConstraint);
```
adrianknight89: Somewhat confused by the naming of these variables. Should they be named `minWidth` and `minHeight` instead or `maxAllowedWidth` and `maxAllowedHeight`?

Xamarin.Forms.Core/Items/ScrollToRequestEventArgs.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+
+namespace Xamarin.Forms
+{
+	public class ScrollToRequestEventArgs : EventArgs
+	{
+		public ScrollToMode Mode { get; }
+
+		public ScrollToPosition ScrollToPosition { get; }
+		public bool Animate { get; }
```
adrianknight89: Verbs are best for method names. I'd prefer `Animated` or `IsAnimated`.

Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
hartez: Good point; I've added CollectionView to the "experimental" list.

Xamarin.Forms.Core/Items/ItemsControl.cs
hartez: I doubt these constructors are more expensive than a `Lazy<T>`.

Xamarin.Forms.Core/Items/ItemsView.cs
hartez: Good catch! Fixed.

Xamarin.Forms.Core/Items/ItemsView.cs
hartez: Because EmptyView can be a View (when EmptyViewTemplate is not set), or it can be the binding context (when EmptyViewTemplate _is_ set).

Xamarin.Forms.Core/Items/ItemsView.cs
hartez: I don't see why they are confusing. The maximum possible width of the control is the width of the screen, unless a smaller width constraint is given. Same for the height.

Xamarin.Forms.Core/Items/ScrollToRequestEventArgs.cs
hartez: Agreed. Renaming to `IsAnimated`.

Xamarin.Forms.Platform.iOS/CollectionView/TemplatedCell.cs
```diff
@@ -0,0 +1,73 @@
...
+		protected nfloat ConstrainedDimension;
+
+		[Export("initWithFrame:")]
+		protected TemplatedCell(CGRect frame) : base(frame)
+		{
+		}
+
+		public IVisualElementRenderer VisualElementRenderer { get; private set; }
+
+		public override void Constrain(nfloat constant)
```
adrianknight89: Maybe needs to be renamed to `ConstrainTo` like an earlier example.

Xamarin.Forms.Platform.iOS/CollectionView/ListSource.cs
```diff
@@ -0,0 +1,25 @@
...
+		public ListSource(IEnumerable<object> enumerable) : base(enumerable)
+		{
+			
+		}
+
+		public ListSource(IEnumerable enumerable)
+		{
+			foreach (object item in enumerable)
+			{
+				Add(item);
```
adrianknight89: No way to leverage `AddRange`?

Xamarin.Forms.Core/Items/ItemsView.cs
adrianknight89: The confusing part for me was seeing `max` in variable names and then seeing `Math.Min` in the initialization. In a totally unrelated example, we could do `var maxWidth = Math.Max(Device.Info.ScaledScreenSize.Width, widthConstraint);` and the variable name would still make sense. Not a huge deal though, esp. after seeing the init logic. :)

Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
adrianknight89: Would `CollectionView_Experimental` still be part of `Forms.SetFlags("Fake_Flag", "CollectionView_Experimental");`?

hartez: Yes.  `TEST_EXPERIMENTAL_RENDERERS` is just there to allow our CI to build and UI test a version of Control Gallery with the fast renderers (a better name for it would have been `TEST_FAST_RENDERERS`).  CollectionView should be available for testing in both versions of Control Gallery.

Xamarin.Forms.Platform.iOS/CollectionView/ListSource.cs
hartez: Your comment made me curious about the performance of AddRange/InsertRange here; using either one would require doing something like `.Cast<object>` from LINQ, but the internal storage growth might be more efficient. So I whipped up a quick test: https://github.com/hartez/ListSourceTest

From my quick-and-dirty testing, adding the items one-by-one in a `foreach` took about 3/4 the time of AddRange/InsertRange. Feel free to clone that repo and look at my testing; I might have missed something obvious. But it looks like this is the fastest option (which surprises me).

Disclaimer: testing was done on a desktop using StopWatch, so this is pseudo-scientific at best.

samhouts:
Not a blocker for this phase, but it does crash on iOS9.3
image

Xamarin.Forms.Platform.iOS/CollectionView/ListSource.cs
adrianknight89: I ran your project, and I'm getting similar results. Looks like casting `IEnumerable` to `IEnumerable<T>` via LINQ is costly. 

I tried replacing the `foreach` loop with a `MoveNext()/Current` logic, but there was no performance benefit. Also, your last sample is the most efficient partly because the other two run beforehand and the underlying data source does not change. If you comment them out and leave the last one, it will take much longer to finish. There is some initialization overhead which gets gobbled up by the other tests.

Having said that, I think `foreach` here is the right choice.

hartez:
> Not a blocker for this phase, but it does crash on iOS9.3 (clicked the first button)
> image

Fixed with https://github.com/xamarin/Xamarin.Forms/pull/4265/commits/16d2b199817ce4bb82294c16ceedf9152a385b5f

Xamarin.Forms.Core/Application.cs
```diff
@@ -206,6 +219,19 @@ public ResourceDictionary Resources
 
 		public event EventHandler<Page> PageDisappearing;
 
+		public static void SetFlags(params string[] flags)
```
StephaneDelcroix: we already have the flags in Core at the Device level. Application, or Device: pick one

Xamarin.Forms.Core/Flags.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Linq;
+using static System.String;
+
+namespace Xamarin.Forms
+{
+	public static class CoreFlags
+	{
+		internal const string CollectionViewExperimental = "CollectionView_Experimental";
```
StephaneDelcroix: so, the magic string is internal, but the helper is public ?

I'd drop this class, and move those into the CollectionView class. Keep the string internal, or we won't be able to drop it in the future, and turn the helper internal or at least not EditorBrowseable

Xamarin.Forms.Core/Items/ItemsControl.cs
```diff
@@ -0,0 +1,103 @@
+﻿using System.ComponentModel;
+
+namespace Xamarin.Forms
+{
+	// TODO hartez 2018/05/31 11:49:49 Move the stuff in ItemsControl.cs to the appropriate files
+	public interface IItemsLayout : INotifyPropertyChanged {}
```
StephaneDelcroix: image

Xamarin.Forms.Core/Items/ItemsControl.cs
```diff
@@ -0,0 +1,103 @@
...
+		public SnapPointsType SnapPointsType
+		{
+			get => (SnapPointsType)GetValue(SnapPointsTypeProperty);
+			set => SetValue(SnapPointsTypeProperty, value);
+		}
+	}
+
+	public class ListItemsLayout : ItemsLayout
+	{
+		// TODO hartez 2018/08/29 17:28:42 Consider changing this name to LinearItemsLayout; not everything using it is a list (e.g., Carousel)
```
StephaneDelcroix: 👍

Xamarin.Forms.Core/Items/ItemsControl.cs
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	public class ListItemsLayout : ItemsLayout
+	{
+		// TODO hartez 2018/08/29 17:28:42 Consider changing this name to LinearItemsLayout; not everything using it is a list (e.g., Carousel)	
+		public ListItemsLayout(ItemsLayoutOrientation orientation) : base(orientation)
+		{
+		}
+
+		// TODO hartez 2018/05/31 15:56:23 Should these just be called Vertical and Horizontal (without List)?
```
StephaneDelcroix: 👍

Xamarin.Forms.Core/Items/ItemsControl.cs
```diff
@@ -0,0 +1,103 @@
...
+		{
+		}
+
+		// TODO hartez 2018/05/31 15:56:23 Should these just be called Vertical and Horizontal (without List)?	
+		public static readonly IItemsLayout VerticalList = new ListItemsLayout(ItemsLayoutOrientation.Vertical); 
+		public static readonly IItemsLayout HorizontalList = new ListItemsLayout(ItemsLayoutOrientation.Horizontal);
+
+		// TODO hartez 2018/08/29 20:31:54 Need something like these previous two, but as a carousel default	
+
+		public override string ToString()
```
StephaneDelcroix: this was probably there for debugging purposes. it's time to get rid of it (if we ship strings, we need to consider internationalization)

Xamarin.Forms.Core/Items/ItemsControl.cs
```diff
@@ -0,0 +1,103 @@
...
+			set => SetValue(SpanProperty, value);
+		}
+
+		public GridItemsLayout([Parameter("Span")] int span, [Parameter("Orientation")] ItemsLayoutOrientation orientation) :
+			base(orientation)
+		{
+			Span = span;
+		}
+
+		public override string ToString()
```
StephaneDelcroix: drop

Xamarin.Forms.Core/Items/ItemsControl.cs
```diff
@@ -0,0 +1,103 @@
...
+			BindableProperty.Create(nameof(Span), typeof(int), typeof(GridItemsLayout), 1, 
+				validateValue: (bindable, value) => (int)value >= 1);
+
+		public int Span
+		{
+			get => (int)GetValue(SpanProperty);
+			set => SetValue(SpanProperty, value);
+		}
+
+		public GridItemsLayout([Parameter("Span")] int span, [Parameter("Orientation")] ItemsLayoutOrientation orientation) :
```
StephaneDelcroix: why are you doing the `[Parameter]` thing ?

Xamarin.Forms.Core/Items/ItemsControl.cs
```diff
@@ -0,0 +1,103 @@
...
+
+		public override string ToString()
+		{
+			return Orientation == ItemsLayoutOrientation.Horizontal ? "Horizontal List" : "Vertical List";
+		}
+	}
+
+	public class GridItemsLayout : ItemsLayout
+	{
+		public static readonly BindableProperty SpanProperty =
```
StephaneDelcroix: shouldn't Span be an attached BP? or am I missing something ?

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -0,0 +1,142 @@
...
+using Xamarin.Forms.Platform;
+
+namespace Xamarin.Forms
+{
+	// TODO hartez 2018/06/23 13:42:22 Trying this out for a nicer read in OnElementPropertyChanged, not sure if I like it yet	
+	public static class PropertyChangedEventArgsExtensions
+	{
+		public static bool Is(this PropertyChangedEventArgs args, BindableProperty property)
+		{
+			CoreFlags.VerifyCollectionViewFlagEnabled();
```
StephaneDelcroix: you can't have both this as an extension method AND contains CV specific check  ¯\_(ツ)_/¯

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Collections;
+using System.ComponentModel;
+using System.Diagnostics;
+using System.Text.RegularExpressions;
+using Xamarin.Forms.Platform;
+
+namespace Xamarin.Forms
+{
+	// TODO hartez 2018/06/23 13:42:22 Trying this out for a nicer read in OnElementPropertyChanged, not sure if I like it yet	
+	public static class PropertyChangedEventArgsExtensions
```
StephaneDelcroix: change this to internal while we are unsure

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -0,0 +1,142 @@
...
+		public static bool Is(this PropertyChangedEventArgs args, BindableProperty property)
+		{
+			CoreFlags.VerifyCollectionViewFlagEnabled();
+			return args.PropertyName == property.PropertyName;
+		}
+
+		public static bool IsOneOf(this PropertyChangedEventArgs args, params BindableProperty[] properties)
+		{
+			CoreFlags.VerifyCollectionViewFlagEnabled();
+			for (int n = 0; n < properties.Length; n++)
```
StephaneDelcroix: properties is an array, `foreach` is free

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -0,0 +1,142 @@
...
+					return true;
+				}
+			}
+
+			return false;
+		}
+	}
+
+	[RenderWith(typeof(_CarouselViewRenderer))]
+	public class CarouselView : ItemsView
```
StephaneDelcroix: move to it's own file

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	public class ItemsView : View
+	{
+		protected internal ItemsView()
+		{
+			CoreFlags.VerifyCollectionViewFlagEnabled(constructorHint: nameof(ItemsView));
+		}
+
+		// TODO hartez 2018/06/24 11:37:00 Give DisplayMemberPath some thought
```
StephaneDelcroix: not now, drop the //TODO

Xamarin.Forms.Core/Application.cs
hartez: D'oh. Somehow I missed that we already had them in Device.

Xamarin.Forms.Core/Items/ItemsView.cs
hartez: Why not? What if I make it not EditorBrowseable?

Xamarin.Forms.Core/Items/ItemsView.cs
hartez: I can't, the renderers need visibility to it.

Xamarin.Forms.Core/Items/ItemsView.cs
hartez: Done.

Xamarin.Forms.Core/Items/ItemsView.cs
hartez: Fixed.

Xamarin.Forms.Core/Flags.cs
hartez: Done.

Xamarin.Forms.Core/Application.cs
hartez: Yeah, that was dumb. Fixed.

Xamarin.Forms.Core/Items/ItemsView.cs
StephaneDelcroix: your class is `PropertyChangedEventArgsExtension`. given the name and the API I might want to use it for any property changed. but it won't, unless the (unrelated to the class) flag is set. 👎 

also, writing `if (args.Is(Label.TextProperty))` is a bit odd, due to the already existing meaning of `is`. So I read this as `if (args is Label.TextProperty)`...

Xamarin.Forms.Core/Items/ItemsView.cs
StephaneDelcroix: If this is a helper specific to renderers but that we don't want to expose publicly, we can copy the helper class in every platform project.

hartez:
@StephaneDelcroix 

> there are still //TODO left

That's because there are still things left to do. It's a work-in-progress, that's why it's behind the flag. 
Or do you just want the TODO comments themselves removed?

> we don't want ToString()

Why don't we want to implement ToString()?

Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
```diff
@@ -41,9 +41,12 @@ protected override void OnCreate(Bundle bundle)
...
 
 #if TEST_EXPERIMENTAL_RENDERERS
-			Forms.SetFlags("FastRenderers_Experimental");
+			// CollectionView lets us test CollectionView stuff until it's officially released
+			Forms.SetFlags("FastRenderers_Experimental", "CollectionView_Experimental");
 #else
-			Forms.SetFlags("Fake_Flag"); // So we can test for flag initialization issues
+			// Fake_Flag is here so we can test for flag initialization issues
+			// CollectionView lets us test CollectionView stuff until it's officially released
+			Forms.SetFlags("Fake_Flag", "CollectionView_Experimental");
```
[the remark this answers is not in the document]
hartez: Done.

Xamarin.Forms.Core/Application.cs
```diff
@@ -7,6 +7,7 @@
 using Xamarin.Forms.Internals;
 using Xamarin.Forms.Platform;
 using System.Diagnostics;
+using System.Linq;
```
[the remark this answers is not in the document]
hartez: Removed.

Xamarin.Forms.Platform.UAP/ListViewRenderer.cs
```diff
@@ -17,6 +17,7 @@
 using WApp = Windows.UI.Xaml.Application;
 using Xamarin.Forms.Internals;
 using Xamarin.Forms.PlatformConfiguration.WindowsSpecific;
+using WItemsControl = Windows.UI.Xaml.Controls.ItemsControl;
```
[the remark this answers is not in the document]
hartez: Removed.

Xamarin.Forms.Core/Items/ItemsView.cs
hartez: Okay, moved to internal helper classes on each platform.

Xamarin.Forms.Core/Items/ItemsView.cs
hartez: Moved this class to an internal helper on each platform, so not accidentally discoverable and no need for the flag.

PureWeen:
@hartez   rebase to fix conflicts then let's run UI tests again with flags off

Xamarin.Forms.Core/Items/GridItemsLayout.cs
```diff
@@ -0,0 +1,28 @@
...
+			BindableProperty.Create(nameof(Span), typeof(int), typeof(GridItemsLayout), 1, 
+				validateValue: (bindable, value) => (int)value >= 1);
+
+		public int Span
+		{
+			get => (int)GetValue(SpanProperty);
+			set => SetValue(SpanProperty, value);
+		}
+
+		public GridItemsLayout([Parameter("Span")] int span, [Parameter("Orientation")] ItemsLayoutOrientation orientation) :
```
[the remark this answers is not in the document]
hartez: My bad, that's left over from before we decided to allow spans to be changed. Fixed now.

Xamarin.Forms.Core/Items/ItemsControl.cs
hartez: Leftover from when Span wasn't mutable. Fixed now.

Xamarin.Forms.Core/Items/ListItemsLayout.cs
```diff
@@ -0,0 +1,21 @@
...
+		{
+		}
+
+		// TODO hartez 2018/05/31 15:56:23 Should these just be called Vertical and Horizontal (without List)?	
+		public static readonly IItemsLayout VerticalList = new ListItemsLayout(ItemsLayoutOrientation.Vertical); 
+		public static readonly IItemsLayout HorizontalList = new ListItemsLayout(ItemsLayoutOrientation.Horizontal);
+
+		// TODO hartez 2018/08/29 20:31:54 Need something like these previous two, but as a carousel default	
+
+		public override string ToString()
```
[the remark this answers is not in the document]
hartez: Dropped.

Xamarin.Forms.Core/Items/GridItemsLayout.cs
```diff
@@ -0,0 +1,28 @@
...
+			base(orientation)
+		{
+			Span = span;
+		}
+
+		public override string ToString()
```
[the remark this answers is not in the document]
hartez: Dropped.

hartez:
@PureWeen 
> Can we get a public API diff caused by this PR? 

Here are all the new public classes/interfaces:

Core, Added:

- public interface IItemsLayout
- public enum SnapPointsType
- public enum SnapPointsAlignment
- public enum ItemsLayoutOrientation
- public class ScrollToRequestEventArgs
- public abstract class ItemsLayout
- public class ListItemsLayout
- public class GridItemsLayout 
- public class ItemsView
- public class CollectionView
- public class CarouselView

UWP, Added:

- public class ItemContentControl
- public class CollectionViewRenderer

iOS, Added:

- public abstract class ItemsViewCell 
- public abstract class DefaultCell 
- public abstract class TemplatedCell 
- public abstract class ItemsViewLayout 
- public class CarouselViewRenderer
- public class CollectionViewRenderer 

Android, Added:

- public class SnapManager 
- public class PositionalSmoothScroller 
- public class ItemsViewRenderer 
- public class ItemsViewAdapter 	
- public class EmptyViewAdapter 
- public class CollectionViewRenderer 
- public class CarouselViewRenderer

Issue #1718, opened 2018-01-26
[Enhancement] Bindable Repeater control


hartez:
## Rationale

Forms does not currently contain a control which displays items from a bound items source in a non-scrollable container.

## Implementation

Attached properties targeting `Layout` will be added to support an `ItemsSource` and `ItemsTemplate`. Setting these properties will add the templated items to the target layout. 

```
public static class BindableLayout
{
	public static readonly BindableProperty ItemsSourceProperty =
		BindableProperty.Create("ItemsSource", typeof(IList), typeof(Layout), default(IList));

	public static readonly BindableProperty ItemTemplateProperty =
		BindableProperty.Create("ItemTemplate", typeof(DataTemplate), typeof(Layout);
}
```

The ItemTemplate is responsible for converting the objects in the ItemsSource into `View` objects. 

Setting the `ItemsSource` will clear out any existing children and update the `Children` collection to match the source. 

## Backward Compatibility

These are new properties, so backward compatibility is likely not an issue.

## Difficulty: Moderate

bmacombe:
I pretty much have this control implemented already as defined.  It will likely just need some refinement, clean up and best practice review.

It supports DataTemplate or DataTemplateSelector.  It can also handle a ItemSource which implements INotifyCollectionChanged to add/remove items as needed instead of rebuilding all the children

bmacombe:
I've been looking at the Forms code base to see how to best modify my code for style, convention, etc.

I thought looking at ListView would be a good example, which lead me to ItemsView and TemplatedItemsList.  Which I don't fully understand how they work yet.  But my question is, does this feature need all that complication or simply add a new view to the stack layout for each item in the bound collection based on the DataTemplate?

My current solution does the simple method, which doesn't make it perform very well for large lists of items, which isn't my use case for it.

Thoughts?

GalaxiaGuy:
There is an alternative route that could be taken.

I implemented `ItemsSource` and `ItemTemplate` as attached properties that could be applied to any `Layout`:

https://github.com/GalaxiaGuy/MobileLibraries/blob/master/XamarinForms.Layout/Layout.cs

Technically that version doesn't support `DataTemplateSelector` but I have implemented that in version elsewhere.

(It gets complicated after line 67 just to support `INotifyCollectionChanged`).

andreinitescu:
I also think the idea of a separate "Repeater" control is not enough and we can have something better than that.

bmacombe:
[...]

PR #4269, opened 2018-10-31
[Enhancement] Map with ItemsSource and ItemTemplate

### LINKED ISSUE #1708: [Enhancement] PinsSource on Forms.Map
## Rationale

Forms.Maps does not currently have a way to bind a list of Pins on a Map.

## Implementation

Add a bindable property PinsSource to Map:

```
public static readonly BindableProperty PinsSourceProperty =
	BindableProperty.Create(nameof(PinsSource), typeof(IList), typeof(Map), default(IList));

public IList PinsSource
{
	get { return (IList)GetValue(PinsSourceProperty); }
	set { SetValue(PinsSourceProperty, value); }
}	
```

And a PinTemplate property:

```
public static readonly BindableProperty PinTemplateProperty = BindableProperty.Create("PinTemplate", typeof(DataTemplate), typeof(Map);

public DataTemplate PinTemplate
{
	get { return (DataTemplate)GetValue(PinTemplateProperty); }
	set { SetValue(PinTemplateProperty, value); }
}
```

The PinTemplate is responsible for converting the objects in the PinsSource into `Xamarin.Forms.Maps.Pin` objects.

Setting the `PinsSource` will clear out any existing pins and update the pin collection to match the source. 
	
## Backward Compatibility

This is a new property, so there should not be any backward compatibility issues.

andreinitescu:
### Description of Change ###

It allows `Map` view to generate its pins by binding to an items source and set a template to create the pins:
```xaml
<map:Map ItemsSource="{Binding Places}">
     <map:Map.ItemTemplate>
          <DataTemplate>
                    <map:Pin Position="{Binding Position}"
                             Address="{Binding Address}"
                             Label="{Binding Description}" />
          </DataTemplate>
     </map:Map.ItemTemplate>
</map:Map>
```
Setting the `ItemsSource` removes any existing pins. 

### Issues Resolved ### 

- fixes #1708

### API Changes ###

New `Map` members:
 - IEnumerable ItemsSource { get; set; } //Bindable Property
 - DataTemplate ItemTemplate { get; set; } // Bindable Property

`Pin` now derives from `Element` in order to allow it to be put in the `DataTemplate`.

### Platforms Affected ### 

- Core/XAML (all platforms)

There are no visual changes

### Testing Procedure ###

I added a gallery page "Map With ItemsSource Gallery" which helps with testing all operations on a collection (screenshot below).

Most of the buttons should be self explanatory. "Update" button updates position of all pins to have the longitude of map's last pin. "Replace" button replaces the map's last pin with a new pin.

image

### PR Checklist ###

- [x] Has automated tests 


andreinitescu:
I'm working on some updates and I'll push very soon, so you might want to hold off with review, thank you.

andreinitescu:
Changes complete, please review, thank you.

jcmanke:
Why not support `DataTemplateSelector` for the `ItemTemplate`?

rmarinho:
@andreinitescu can't you reuse the BindableLayout helpers here? I think it makes more sense to reuse some of that code. Maybe even improve the the helpers so we can use it here and in more places.

Thanks

andreinitescu:
@rmarinho My first thought was also about reusability, but BindableLayout works with actual views on the Layout<View>. So I think since the Map view is not actually a control which has child views like Layout<View> does, there are different patterns in my opinions. 
The commonality is handling collection changes by INotifyCollectionChanged.OnCollectionChange, but I don't think it's worth abstracting/make that reusable.

Could you please be more specific?

andreinitescu:
I don't see a reason to close this since it's work intended to implement an approved feature. 

If there's something which needs to be changed in the existing work, it makes sense to first understand what should the changes be so I can make the changes right on this one instead of spending time on another PR.

StephaneDelcroix:
Hi @andreinitescu,

It's still our goal to provide templating capabilities to all controls exposing collections (as described in #1708). But we would like to avoid
1. exposing templating directly (prefer attached BPs)
2. duplicating the INPC/INCC code everywhere.

This is the reason why I closed this PR, because we find it kinder to close it now than ignoring it for months, then closing it.

The best approach, in my opinion, would be -- for maintaining purposes --  to generify the BindableLayout code, and then exposing a non-generic BindableLayout and PinRepeater (whatever the name) classes. You're welcome to propose something.

Thanks for your time, and contributions.

andreinitescu:
@StephaneDelcroix  Sure, like I said, I'm open for discussion and eager to contribute more. I just want to avoid working in vain and with your help by simple planning it can be avoided. I just need to know how you want it.

What's the best way to open a discussion for this? 

1. Create a separate issue just for that.

2. Maybe you reopen the existing #1718. Which would make more sense since it seems you guys are not actually happy with it even if it was merged into master. So I think reopen the original issue? 

3. Add more comments here. But I think you said you guys don't watch activity on closed issues, not sure.

Xamarin.Forms.Maps/Pin.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Xamarin.Forms.Maps
 {
-	public class Pin : BindableObject
+	public class Pin : Element
```
StephaneDelcroix: why did you had to change this ?

andreinitescu: Per my understanding, `Pin` has to be an `Element` in order to be the root of `DataTemplate`. Please correct me if I'm wrong, I think the technical way to describe is, it's part of the logical tree but not the visual tree. I think same pattern exists in ListView? (the cells)

StephaneDelcroix: 👍

jcmanke:
I still think DataTemplateSelectors should be supported for the ItemTemplate property before this is merged. Saying that it isn't needed because the only type that can be used is Pin is like saying ListViews don't need it because the ItemTemplate must return a Cell. Once upon a time, the Pin class was sealed and that argument might have been valid, but that hasn't been the case for over a year.

Suppose I have a `List<Address>` I want to bind to the Map, and have subclassed the `Address` class as `House` and `Business`. I want to bind different properties to the Pin's Address and Label properties depending on if it's a House or a Business. How would I do that without a DataTemplateSelector?

rmarinho:
Failing tests not related

andreinitescu:
> Saying that it isn't needed because the only type that can be used is Pin is like saying ListViews don't need it because the ItemTemplate must return a Cell

Not exactly. The ListView has the ViewCell which takes **any** View as a template to render the data item. Currently the Map.ItemTemplate only expects the Pin.  

> Suppose I have a List<Address> I want to bind to the Map, and have subclassed the Address class as House and Business. I want to bind different properties to the Pin's Address and Label properties depending on if it's a House or a Business. How would I do that without a DataTemplateSelector?

It's not impossible, if you're in hurry, just have an interface like IAddress which your House and Business classes implement? 

I had already started working on a PR to add the `DataTemplateSelector`. It's a simple feature, but a PR to the framework XF takes some effort and time, it's not just about adding your custom renderer in you're app and you're done.

kalyanmungi:
Enhancement suggestion : When clicked on pin, its good to have callout view immediately on top of pin.  Please see the image
MapView

gabrielbunselmeyer:
With the Tag property being an object, how do we bind an int to it? As it's an object its type becomes a {Xamarin.Form.Binding}. Only thing we can get out of it is the path of the binding instead of its value.

image

PR #4301, opened 2018-11-05
[XamlC] Explicitly cast values for ldc.i4

akihikodaki:
### Description of Change ###

This is a bug fix. Mono.Cecil accepts only `System.Int32` as an operand for `OpCodes.Ldc_I4` and casting is required for this case.

### Issues Resolved ### 

I didn't find any relevant issue.

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###

Just try to set a `System.Byte` property with XamlC. For example, the following XAML will cause an exception where `Foo.Bar` is a property whose type is `System.Byte`.

```XAML
<Foo Bar="42" />
```

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Build.Tasks/NodeILExtensions.cs
+++ b/Xamarin.Forms.Build.Tasks/NodeILExtensions.cs
@@ -161,17 +161,17 @@ namespace Xamarin.Forms.Build.Tasks
 			if (targetTypeRef.ResolveCached().BaseType != null && targetTypeRef.ResolveCached().BaseType.FullName == "System.Enum")
 				yield return PushParsedEnum(targetTypeRef, str, node);
 			else if (targetTypeRef.FullName == "System.Char")
-				yield return Instruction.Create(OpCodes.Ldc_I4, Char.Parse(str));
+				yield return Instruction.Create(OpCodes.Ldc_I4, unchecked((int)Char.Parse(str)));
 			else if (targetTypeRef.FullName == "System.SByte")
-				yield return Instruction.Create(OpCodes.Ldc_I4, SByte.Parse(str, CultureInfo.InvariantCulture));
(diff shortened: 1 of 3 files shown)


StephaneDelcroix:
This looks right. It requires a test, e.g. next to the test for `UShort` in `SetValue.xaml[.cs]`

Xamarin.Forms.Build.Tasks/NodeILExtensions.cs
```diff
@@ -161,17 +161,17 @@ public static bool CanConvertValue(this ValueNode node, ILContext context, TypeR
 			if (targetTypeRef.ResolveCached().BaseType != null && targetTypeRef.ResolveCached().BaseType.FullName == "System.Enum")
 				yield return PushParsedEnum(targetTypeRef, str, node);
 			else if (targetTypeRef.FullName == "System.Char")
-				yield return Instruction.Create(OpCodes.Ldc_I4, Char.Parse(str));
+				yield return Instruction.Create(OpCodes.Ldc_I4, (int)Char.Parse(str));
```
StephaneDelcroix: those casts can be `unchecked`

akihikodaki: You mean to wrap them with `unchecked`? That is right, but I wonder if it makes difference.

akihikodaki:
ac17be60d0637a3c9736048302b01eadc2096b9b adds tests and wraps casting with `unchecked`.

Xamarin.Forms.Build.Tasks/NodeILExtensions.cs
StephaneDelcroix: you're right, it doesn't change a thing. my wrong.

StephaneDelcroix:
build

PR #4313, opened 2018-11-05
[iOS] make button renderer constructor public

PureWeen:
### Description of Change ###

Having a protected constructor on a renderer causes the previewer to break

### Platforms Affected ### 
- iOS

### PR Checklist ###

- [ ] Has automated tests 


alanmcgovern:
Hola! I just wanted to update this with a screencast and a question!

This screencast shows the ABI break that the original commit introduced: http://recordit.co/CHvaZq8Zen .

It was great that we caught this with the normal regression test-suite, but I'm wondering if we can add a tool to detect these breaks before we PRs are merged. If you don't have such a tool and want one, I think monodevelop already has one that could be re-used :)

Issue #4312, opened 2018-11-05
[UWP] Skipping setting "default values" for element properties in renderers is problematic


RemusVasii:
https://github.com/xamarin/Xamarin.Forms/blob/ee91ac443941133a9163c62787d831ad16aaa156/Xamarin.Forms.Platform.UAP/ButtonRenderer.cs#L58

Is checking of default values really necessary in renderers?:

If ... Element.Padding != Button.PaddingProperty.DefaultValue ... UpdatePadding();

The default padding value for a Xamarin button is (0,0,0,0), while the default padding for a UWP button is (8,4,8,4). (https://msdn.microsoft.com/en-us/library/windows/apps/mt299109.aspx)

This means that if I set padding (0,0,0,0) on a Xamarin button, and I really want it to be (0,0,0,0) on UWP, the renderer will skip setting the padding property. And my UWP button will have a (8,4,8,4) padding, which is not my intent!

Does checking and skipping default property values in renderers have any meaningful purpose? It seems to me that it creates more problems than it solves.

RemusVasii:
Well, there is also a TODO coment in there, which I did not see initially:

https://github.com/xamarin/Xamarin.Forms/blob/ee91ac443941133a9163c62787d831ad16aaa156/Xamarin.Forms.Platform.UAP/ButtonRenderer.cs#L42

But anyway, why not solve the problem now?

PR #4330, opened 2018-11-07
[All] Fix TextDecorations & DataTigger interaction

### LINKED ISSUE #4121: Forms 3.3: TextDecorations aren't unset by a DataTrigger
### Description

In Forms 3.3, `TextDecorations` aren't unset by a DataTrigger, unlike `FontAttributes`, which are.

### Steps to Reproduce

1. Run the attached sample.
2. Browse to the "Basic Switch Bindings" page.
3. Toggle the italic/bold switches on/off and the text appearance changes as expected.
4. Toggle the underline or strikethrough switches on/off.

### Expected Behavior

The text is underlined/receives strikethrough, and when toggling the switches to off the text decoration is undone.

### Actual Behavior

The text is underlined/receives strikethrough, but the text decoration remains when toggling the switches to off.

### Basic Information

- Version with issue: 3.3 -pre3.
- IDE: VSMac
- Platform Target Frameworks: 
  - iOS:  12.0
  - Android: API 27.
  - UWP:  Not tried.

### Screenshots

Underline switch toggled on:

simulator screen shot - iphone 8 - 2018-10-18 at 10 26 07

Underline switch toggled off:

simulator screen shot - iphone 8 - 2018-10-18 at 10 26 12

### Reproduction Link

[SwitchDemo.zip](https://github.com/xamarin/Xamarin.Forms/files/2491188/SwitchDemo.zip)

### LINKED ISSUE #8272: [Bug] TextDecorations don't work
# Description

Tested on iOS. I set underline or strikethrough on a label and nothing renders using 4.3.0 stable. 

### Basic Information

- Version with issue: 4.3.0
- Last known good version: ?
- Platform Target Frameworks: 
  - iOS:  12.2, 13.1

### Screenshots

Screenshot 2019-10-27 20 29 05

### Reproduction Link

https://github.com/davidortinau/fabriculous/blob/bug/decorations/Fabriculous/Pages/ButtonsPage.xaml#L42

alanag13:
### Description of Change ###
Fixes an issue that makes it impossible to remove a TextDecoration that is set by a DataTrigger once it has been set.

More accurately, this change makes `BindableObject.ClearValue` behave the way it is supposed to for `TextDecorations`.

### Issues Resolved ### 
- fix #4121
- fix #8272 

### Platforms Affected ### 
 - iOS
- macOS
- Android
- UWP
- WPF

Calls to `ClearValue` on Elements with `TextDecorations` will now correctly display their pre-datatrigger appearance.

### Testing Procedure ###
Run the included UITest. The switch should cause to alternately appear and disappear.

### PR Checklist ###

- [X] Has automated tests 


alanag13:
**Background:**
In many recent changes to XF, a check to `BindableObject.IsSet` has been added to the render logic in order to prevent any XF-established defaults from clobbering any values that may have been set by custom renderers or effects.

As seen here, a drawback to this approach is that `IsSet` will be false when `ClearValue` is called, so in this example, DataTriggers could set a new value, but never reset it. I noticed that some other properties try to work around this problem by also checking to see if new value is the same as the default value, like this:
https://github.com/xamarin/Xamarin.Forms/blob/ee91ac443941133a9163c62787d831ad16aaa156/Xamarin.Forms.Platform.UAP/ButtonRenderer.cs#L42

However, as the comment notes, this has the drawback that it is impossible to set the property back to the default value. See #4312.

I *believe* that the approach I'm taking here resolves this more high-level issue while also still serving the purpose of protecting the values in the case of custom renderers and effects, but am looking for feedback/confirmation. If the team agrees this approach looks good, I can similarly file a PR that would address #4312.

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
```diff
@@ -161,7 +161,7 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
 				UpdateTextDecorations();
```
PureWeen:
```suggestion
				UpdateTextDecorations(false);
```

Xamarin.Forms.Platform.Android/FastRenderers/LabelRenderer.cs
```diff
@@ -292,9 +292,9 @@ void UpdateFont()
 			}
 		}
 
-		void UpdateTextDecorations()
+		void UpdateTextDecorations(bool isNewElement)
```
PureWeen: I wonder if instead of adding this property we just do an if check inside *OnElementChanged* to not call UpdateTextDecorations since that's really the only scenario this bool is accounting for

Xamarin.Forms.Platform.Android/FastRenderers/LabelRenderer.cs
```diff
@@ -292,9 +292,9 @@ void UpdateFont()
...
 		{
-			if (!Element.IsSet(Label.TextDecorationsProperty))
+			if (!Element.IsSet(Label.TextDecorationsProperty) && isNewElement)
```
PureWeen: I'm thinking that checking for IsSet isn't the most reliable way to check for the TextDecorationsProperty. 
Wouldn't checking *if Label.TextDecoration == TextDecorations.None*
Be more accurate?

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
alanag13: Am I missing something here? True should be used if it's a new element, false if it's an updated property on an existing one.

Xamarin.Forms.Platform.Android/FastRenderers/LabelRenderer.cs
alanag13: That would prevent you from explicitly removing the text decorations after they had been set for the first time.

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
PureWeen: Nothing was specified so it was causing a compile error. 

This UpdateTextDecorations is part of the ElementPropertyChanged right?

https://github.com/xamarin/Xamarin.Forms/blob/221f43d689393bc58811d602c93fc7a74f348a5c/Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs#L161

Xamarin.Forms.Platform.Android/FastRenderers/LabelRenderer.cs
alanag13: Yeah, we could just check for !isSet in that call instead of passing the bool, I can change it to work that way

alanag13:
@PureWeen had a couple questions and responses about some of your points. Also, let me know about my background in the pot above. This problem exists in a lot of there's properties that use IsSet and I wanted to check to see if this approach  (with whatever tweaks we land  on) was good for addressing other properties too.

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
alanag13: Oh oops, that was added in a PR separate from the one I did this feature in so I missed it. I could have sworn I ran this on all platforms, that's embarrassing.

Xamarin.Forms.Platform.Android/FastRenderers/LabelRenderer.cs
alanag13: actually wait, no it wouldn't haha. I've locked myself into the line of thinking that I had before adding isNewElement. I can change this too.

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
alanag13: Ah, I think I tested this before that change went in and then rebased with out building again before I filed this.

PureWeen:
@alanag13 so I'm not the biggest fan of checking IsSet mainly because if you define a defaultValueCreator on the bindable property then IsSet doesn't really do anything useful. I consider IsSet to have side effects that I don't want to deal with. For example if I add a defaultValueCreator to something it'll cause all code that uses IsSet to now behave differently.

The core idea with these types of checks is basically
- Did the user set anything? no? ok just use the platform default
- Did the user set something? yes? ok let's store the default so if they want to go back to the default we can

With Enumerations that's typically done by having a "Default" and then with non enumerations we've sometimes used IsSet because then (in theory) the user can use ClearValue at which point we return back to the native property.

Example here https://github.com/xamarin/Xamarin.Forms/pull/3174/files  with Button Padding. The code in that PR doesn't work fully as intended because the bindable property for padding has a DefaultValueCreator. This means that when you clear the bindable property IsSet remains true so the user can't get back to "Default". The padding code needs to be rewritten to have some sort of "default value" representing not set. I'm thinking something like *new Thickness(double.NegativeInfinity)*

(Hopefully I'm still answering your question and not going off on a tangent)

So (hypothetically) with TextDecorations if we had stuck to the usual pattern  "None" would always be actionable and then Default would be what is used to say just don't do anything please or please return me to my defaults. This is why we are now having to add some extra code. Because we've overloaded what "None" means

For example we now kind of have two different things happening based on the "None" setting. If you have a new control don't do anything but if you set it back to None then do things that "represent" none.  In theory they are the same so it doesn't matter but it is a bit atypical to our pattern. 

Because
None == Default then I would say *!IsSet* can't be trusted as meaning Default because what if someone changes the Default on the bindable property? Now the code breaks.

That's why I was thinking instead of using *!IsSet* we would check if the property != None
```C#
		protected override void OnElementChanged(ElementChangedEventArgs<Label> e)
		{
			if (e.NewElement != null)
			{
				if (Control == null)
				{
					SetNativeControl(new NativeLabel(RectangleF.Empty));
#if !__MOBILE__
					Control.Editable = false;
					Control.Bezeled = false;
					Control.DrawsBackground = false;
#endif
				}

				UpdateLineBreakMode();
				UpdateAlignment();
				UpdateText();
				if(Element.TextDecorations != TextDecorations.None)
					UpdateTextDecorations();
				UpdateTextColor();
				UpdateFont();
				UpdateMaxLines();
			}

			base.OnElementChanged(e);
		}
```

alanag13:
@PureWeen thanks for the detailed response! That all makes sense. Do you think the other renderers that use isSet could be improved by a combo of checking  for the default value and checking if it's an update, or do you think this solution is only applicable in this case?

PureWeen:
@alanag13  It's been a minute and I still kind of have holiday brain but I think I'm following you.

In these cases we usually save the native thing being changed to a local default value so the state can be "reset" to whatever it was natively.

I think this handles the cases you bring up 
1) if it's the first time applied it just saves the current state of PaintFlags to *paintflagsDefault*
2) If anything causes the TextDecorations to flip back to None then just apply the *paintflagsDefault*

```C#
		int paintflagsDefault = -1;
		void UpdateTextDecorations()
		{
			var textDecorations = Element.TextDecorations;

			if (paintflagsDefault == -1)
				paintflagsDefault = (int)PaintFlags;

			if (textDecorations == TextDecorations.None)
				PaintFlags = (PaintFlags)paintflagsDefault;
			else
			{
				if ((textDecorations & TextDecorations.Strikethrough) == 0)
					PaintFlags &= ~PaintFlags.StrikeThruText;
				else
					PaintFlags |= PaintFlags.StrikeThruText;

				if ((textDecorations & TextDecorations.Underline) == 0)
					PaintFlags &= ~PaintFlags.UnderlineText;
				else
					PaintFlags |= PaintFlags.UnderlineText;
			}
		}
```

It's similar to how the typeface code does it
https://github.com/xamarin/Xamarin.Forms/blob/5716f8d75704328a8a73a1bfdc83e37e55d5d662/Xamarin.Forms.Platform.Android/Renderers/FontExtensions.cs#L62

samhouts:
@alanag13 Feel free to remove the do-not-merge label when you've submitted your changes. Thanks!

samhouts:
@alanag13 Are you able to address the feedback? Thanks!

alanag13:
hey @samhouts I've been meaning to come back to this and clean up the final feedback, I'll try to get to it soon. Sorry to keep y'all waiting.

GitInfo.txt
```diff
@@ -1 +1 @@
-4.1.0
\ No newline at end of file
+4.1.0
```
alanag13: I merged master into my branch to get caught up instead of 4.1.0, and this is what happened when I tried to change it back 🤷‍♂

alanag13:
@PureWeen @rmarinho @samhouts I'm back from the dead and this is ready to be reviewed again. Unfortunately it doesn't look like I have permission to edit labels on issues, so I can't remove the `DO NOT MERGE` label like was previously requested.

alanag13:
@PureWeen @rmarinho @samhouts are the UITest failures related? Any guidance on what to be looking at if they are?

alanag13:
@samhouts @PureWeen just wanted to check on this again and see what i can do, if anything. Are the UITests related?

rmarinho:
@alanag13 sorry, let's rebase again, see if tests feel better.

GitInfo.txt
rmarinho: encoding?

alanag13:
@rmarinho rebased onto 4.1

GitInfo.txt
alanag13: reverted via `git checkout`, seems good now

alanag13:
@rmarinho test failures related? Should I rebase again?

PureWeen:
Yes please @alanag13 ! Can you rebase on 4.2.0?

Xamarin.Forms.Platform.UAP/LabelRenderer.cs
```diff
@@ -136,7 +137,9 @@ protected override void OnElementChanged(ElementChangedEventArgs<Label> e)
 				_isInitiallyDefault = Element.IsDefault();
 
 				UpdateText(Control);
-				UpdateTextDecorations(Control);
+
+				if (e.OldElement?.TextDecorations != e.NewElement.TextDecorations)
```
rmarinho:
```suggestion
				if (e.OldElement?.TextDecorations != e.NewElement?.TextDecorations)
```

Xamarin.Forms.Platform.UAP/LabelRenderer.cs
```diff
@@ -177,20 +180,25 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
...
-			if (!Element.IsSet(Label.TextDecorationsProperty))
-				return;
-
 			var elementTextDecorations = Element.TextDecorations;
 
-			if ((elementTextDecorations & TextDecorations.Underline) == 0)
-				textBlock.TextDecorations &= ~Windows.UI.Text.TextDecorations.Underline;
-			else
-				textBlock.TextDecorations |= Windows.UI.Text.TextDecorations.Underline;
+			if (_textDecorationsDefault == 1)
```
rmarinho:
```suggestion
			if (Control == null)
			    return;
			if (_textDecorationsDefault == 1)
```

Xamarin.Forms.Platform.WPF/Renderers/LabelRenderer.cs
```diff
@@ -25,7 +26,8 @@ protected override void OnElementChanged(ElementChangedEventArgs<Label> e)
 
 				// Update control property 
 				UpdateText();
-				UpdateTextDecorations();
+				if (e.OldElement?.TextDecorations != e.NewElement.TextDecorations)
```
rmarinho:
```suggestion
				if (e.OldElement?.TextDecorations != e.NewElement?.TextDecorations)
```

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
```diff
@@ -164,8 +166,13 @@ protected override void OnElementChanged(ElementChangedEventArgs<Label> e)
...
 				UpdateLineBreakMode();
 				UpdateAlignment();
-				UpdateText();
-				UpdateTextDecorations();
+				if (e.OldElement?.Text != e.NewElement?.Text)
+				{
+					UpdateText();
+					UpdateTextDecorations();
+				}
+				if (e.OldElement?.TextDecorations != e.NewElement.TextDecorations)
```
rmarinho:
```suggestion
				if (e.OldElement?.TextDecorations != e.NewElement?.TextDecorations)
```

rmarinho:
@alanag13 can you pick this up soon, or should we take over? thanks

alanag13:
hey @rmarinho sorry for the slow movement here, I just moved and am still in the process of getting all my things in order, but I'm almost done. I'll try to get this rebased again soon.

alanag13:
@rmarinho retargeted/rebased to master.

Xamarin.Forms.Platform.UAP/LabelRenderer.cs
```diff
@@ -143,7 +144,9 @@ protected override void OnElementChanged(ElementChangedEventArgs<Label> e)
 				_isInitiallyDefault = Element.IsDefault();
 
 				UpdateText(Control);
-				UpdateTextDecorations(Control);
+
+				if (e.OldElement?.TextDecorations != e.NewElement?.TextDecorations)
```
PureWeen: Do we need this check?

Xamarin.Forms.Platform.WPF/Renderers/LabelRenderer.cs
```diff
@@ -26,7 +27,8 @@ protected override void OnElementChanged(ElementChangedEventArgs<Label> e)
 
 				// Update control property 
 				UpdateText();
-				UpdateTextDecorations();
+				if (e.OldElement?.TextDecorations != e.NewElement?.TextDecorations)
```
PureWeen: Do we need this check?

alanag13:
@PureWeen ugh, I must have messed up something in the merge, I admittedly cheated and didn't test it over again. I'll check it out :-|

alanag13:
@PureWeen It appears that https://github.com/xamarin/Xamarin.Forms/pull/6944/files#diff-381c6c37e826f38028e8841495e4f78e inadvertently made text decorations stop working on iOS entirely, I found that the normal gallery views also no longer displayed correctly.

PureWeen:
@alanag13 that's unfortunate :-/

 in your explorations did you see the issue? Is it an easy fix along side this PR?

alanag13:
@PureWeen I was able to make the TextDecorations behave as normal once again and pushed that change, but for some reason now the the text size seems to change when I toggle the switch, I'm pausing here for now but can take a closer look again in a bit, unless you have any ideas?

PureWeen:
@alanag13 can you log an issue expanding on these comments?

https://github.com/xamarin/Xamarin.Forms/pull/4330#issuecomment-545731143

Just indicate what gallery you are running? Then we can get it marked as a regression and get that fixed quickly

alanag13:
@PureWeen sorry I didn't see this, it looks like @davidortinau made an issue though. I'm just running the Label Gallery, it has underline and strikethrough pages

alanag13:
@PureWeen both issues should be fixed now.
At the time that I first implemented TextDecorations, (i think) it was the only prop that made use of `AttributedText`. Now other props use that, so `_defaultTextDecorations` was saving TextAttributes that were added from unrelated props; this is what was causing my size change issue. The fact that text decorations werent working at all was just caused by a line that was accidentally deleted.

alanag13:
@rmarinho Your merge of master into this fix branch undid my change that fixed this on iOS:
https://github.com/xamarin/Xamarin.Forms/pull/4330/files/5d21503bf1e7e278962f3eb8d96ed242dd141355..582e9a1c3699ec66ca0dceed71db1ad8db4369fa

I don't know for certain but I suspect that the UWP problem you're describing is also an issue in master, can you verify this?

I apologize but I don't currently have much time to look at this, if it isn't too much to ask i'd like MSFT to close this one out.

rmarinho:
Master works fine on UWP . 

Ok we will be taking over this one.  

Thanks

akamud:
Is this planned for 4.3 SR or 4.4? Since this causes a regression on iOS (#8272) that is quite impactful (due to it also being a visual bug), I'm kind of worried that there is no activity on this issue anymore.

Mikilll94:
Any update on this? This causes a regression on iOS - #8272, so this should have high priority.

jsuarezruiz:
@alanag13 Could you rebase branch 4.5.0 to fix the conflicts?. Thanks!

samhouts:
I'm going to close this because we've fixed #8272, and we'll likely need to reevaluate the fix for #4121 now that it's done. @alanag13 If you're able to pick that work up again, we'd be happy to work with you! Thanks!!

PR #4333, opened 2018-11-07
Make ToolbarItem constructor match properties

breyed:
This changes the `ToolbarItem` constructor's `name` parameter to `text`, to match the non-obsolete `Text` that gets the value. Without the change, it's confusing when tracing data flow related to `Text`, since it doesn't appear at first that it's being set.

### Issues Resolved ### 

- fixes #

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/ToolbarItem.cs
+++ b/Xamarin.Forms.Core/ToolbarItem.cs
@@ -16,12 +16,12 @@ namespace Xamarin.Forms
 		{
 		}
 
-		public ToolbarItem(string name, string icon, Action activated, ToolbarItemOrder order = ToolbarItemOrder.Default, int priority = 0)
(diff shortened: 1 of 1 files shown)


StephaneDelcroix:
Thanks for this first PR, @breyed. We usually do not accept that kind of PRs (or style-changing ones), as the cost/benefit is quite high. Aim for something bigger next time, but make sure there is a "next time".

breyed:
IMHO, improvements to usage of the library are generally more important than internal-only improvements, such as your example of style changes. Since the parameter name shows up in users' Intellisense and impacts how easy the library is to work with, I'd say even such a small change is valuable enough to make. APIs that have their tidiness neglected are subject to death by a thousand papercuts.

Thanks for understanding and pulling in this edit. If I was aware of other changes to roll in, I would have included them, and will continue to keep an eye out.

PR #4369, opened 2018-11-10
WIP - CheckBox & CheckBoxCell Implementation

### LINKED ISSUE #1728: [Enhancement] Checkbox Control
## Rationale
Currently the only type of toggle control is a switch. Checkboxes can be useful for indicating action. For example agreeing to something and there are cases where people prefer a checkbox to a switch for their UI.

## Implementation
Create a new Control and set of platform renderers for a CheckBox control and a CheckBoxCell

```csharp
public class CheckBox : IButtonController
{
        public bool IsChecked {get; set; } // BP
}
```

VisualStateManager additions: Checked, Pressed

### Android
https://developer.android.com/guide/topics/ui/controls/checkbox.html

### iOS
Draw it

### UWP
https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/checkbox

## Implications for CSS

```css
checkbox {
}
```
## Backward Compatibility

## Difficulty : Medium

Not much work for Android and UWP but the iOS implementation still needs to be fleshed out a bit

jamesmontemagno:
### Description of Change ###

Start Implementation of CheckBox and CheckBoxCell. Initial feedback is welcome.

CheckBox Implementations:
* [x] Android
* [x] iOS
* [x] UWP
* [x] macOS
* [x] WPF

CheckBoxCell Implementations:
* [x] Android
* [x] iOS
* [x] UWP
* [x] macOS
* [x] WPF

Properties:
* [x] IsChecked
* [x] CheckedColor 
* [x] UncheckedColor 

### Issues Resolved ### 
- fixes #1728

### API Changes ###
Added:
 - CheckBox

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS
- Android
- UWP
- WPF

None

Android:
image

iOS:
image

WPF:
image

UWP:
image

### PR Checklist ###

- [ ] Has automated tests 


PureWeen:
```
Xamarin.Forms.Platform.WPF\Renderers\SwitchRenderer.cs(11,15): Error CS0311: The type 'Xamarin.Forms.CheckBox' cannot be used as type parameter 'TNativeElement' in the generic type or method 'ViewRenderer<TElement, TNativeElement>'. There is no implicit reference conversion from 'Xamarin.Forms.CheckBox' to 'System.Windows.FrameworkElement'.
```

Xamarin.Forms.Platform.Android/AppCompat/CheckBoxRenderer.cs
```diff
@@ -0,0 +1,100 @@
+using System;
+using System.ComponentModel;
+using Android.Content;
+using Android.Support.V7.Widget;
+using Android.Widget;
+using Xamarin.Forms.Platform.Android.FastRenderers;
+
+namespace Xamarin.Forms.Platform.Android.AppCompat
+{
+	public class CheckBoxRenderer : ViewRenderer<CheckBox, AppCompatCheckBox>, CompoundButton.IOnCheckedChangeListener
```
PureWeen: For new controls on android we just use the FastRenderer structure to eliminate the unnecessary ViewGroup in the hierarchy.  A slimmed down version of ButtonRenderer should work
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/FastRenderers/ButtonRenderer.cs

PureWeen: Also we just put them into the Android namespace for the new controls as well

https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/AppCompat/ImageButtonRenderer.cs#L15

Xamarin.Forms.Platform.Android/Renderers/CheckBoxRenderer.cs
```diff
@@ -0,0 +1,92 @@
+using System;
+using System.ComponentModel;
+using Android.Content;
+using Android.Widget;
+using ACheckBox = Android.Widget.CheckBox;
+
+namespace Xamarin.Forms.Platform.Android
+{
+	public class CheckBoxRenderer : ViewRenderer<CheckBox, ACheckBox>, CompoundButton.IOnCheckedChangeListener
```
PureWeen: Probably won't include this one. We're not really adding new controls behavior to non appcompat. It's mainly just in maintenance mode

CasNouwens:
Fingers crossed that this control makes it into base Xamarin.Forms soon 😄

Xamarin.Forms.Platform.Android/AppCompat/CheckBoxRenderer.cs
jamesmontemagno: This is now addressed!

Xamarin.Forms.Platform.Android/Renderers/CheckBoxRenderer.cs
jamesmontemagno: Now addressed!

jamesmontemagno:
@PureWeen take a look at the latest android code and let me know what you think!

nawidy:
Looking forward to getting this in for iOS & Android

andreinitescu:
@jamesmontemagno Few things:
1. It's not supposed to have any text? 
2. The Xamarin.Forms.CheckBox doesn't have Clicked or Pressed events. Why not keep these from Button? CheckBox is actually a button too.

jamesmontemagno:
> @jamesmontemagno Few things:
> 
> 1. It's not supposed to have any text? i.e. checkBox.Text = "SomeText";
> 2. The Xamarin.Forms.CheckBox doesn't have Clicked or Pressed events. Why not keep these from Button? CheckBox is actually a button too.

1. Every platform it is just a checkbox no text with it at all. Just use a stacklayout for that really like the toggle. 

2. There is a CheckedChanged event.

PureWeen:
I agree on leaving the text property out. Android does have a text property that shows up if you set it but getting iOS to match will be annoying plus I feel like that's going to require adding too many other "text" based apis to this. Font, color etc.. All of which you can just achieve with a Label.

I could see adding a Pressed event and also a "Pressed" state for Visual State Manager (much like Button and ImageButton has) as the user might want to visually change something while the users finger is on the check box. Also an "Checked" VSM state

The part of this I'm wondering about are the *CheckedColor* *UncheckedColor* apis

I'm wondering if CheckBox should implement *IBorderElement* and how hard it'd be to achieve those. CornerRadius could be left for later but *BorderColor* and *BackgroundColor* seem relevant.

The behavior of changing colors baed on Checked and Unchecked would thus be achieved through VSM. With the added "Pressed" state you could define the whole visual experience with VSM

andreinitescu:
From a practical standpoint, a check-box needs the label. Except iOS which doesn't have a check-box, **on all platforms (Android/UWP/MacOS/WPF/Tizen) check-box has label**. 

I don't see why Xamarin Forms should get it with no label, and having to wrap it by a StackLayout and add a Label. 

From an implementation standpoint, on iOS, implementing it based on the UIKit.UIButton which already comes with text too, shouldn't be difficult. Render it to an UIButton makes it easy to reuse existing button renderer implementation. Code duplication can be avoided with little refactoring. 

Please take a look to my implementation here:  https://github.com/andreinitescu/Xamarin.Forms/tree/fix-1728-Checkbox
It's not complete (I stopped once James said he will work on it), but it shows for iOS what I described above.  Also, see it in action here https://www.youtube.com/watch?v=0aoB0ca6AZc&feature=youtu.be
image

For iOS, note I extracted the text coloring implementation used by Button to the [`TextButtonColorTracker<TButton>`](https://github.com/andreinitescu/Xamarin.Forms/blob/fix-1728-Checkbox/Xamarin.Forms.Platform.iOS/Renderers/TextButtonColorTracker.cs) class so it can be used by both Button and CheckBox. This makes it reusable across any TButton ([see the interfaces](https://github.com/andreinitescu/Xamarin.Forms/blob/fix-1728-Checkbox/Xamarin.Forms.Platform.iOS/Renderers/TextButtonColorTracker.cs) ). There are also some new interfaces/classes [IToggleButtonElement.cs](https://github.com/andreinitescu/Xamarin.Forms/blob/fix-1728-Checkbox/Xamarin.Forms.Core/ToggleButtonElement.cs) and [ToggleButtonElement.cs](https://github.com/andreinitescu/Xamarin.Forms/blob/fix-1728-Checkbox/Xamarin.Forms.Core/ToggleButtonElement.cs) which lay the foundation for toggle buttons (CheckBox, RadioButton).

I hope you can take a look and give label a chance so we don't have to add two more controls every time :)

andreinitescu:
> I think that a Text property is a novel idea
Not sure what you mean by "novel" idea. Please see my comment above, I mentioned that except iOS, where check-box doesn't exist, ALL platforms have check-box, and ALL have label too.

> the text isn't always on the right side, maybe left, maybe below or above
The most common scenario is having it on the right or left side (depending on the RTL setting).

> Having a default text option is nice, but what about scale, font properties, text properties, so much overlap in general. 
I mentioned in my comment that, many of the implementation can be reused easily. Please take a look.

> but I look at Switch and there is no Text property there and really CheckBox and Switch are the exact same control IMHO.
Let's think it another way. Apple doesn't seem to want on iOS a check-box control as we know it, the iOS Human Interface Guidelines doesn't mention it anywhere, and UIKit.UISwitch is the iOS control for an on/off widget. 
We definitely want a check-box, even native apps implement it. But a check-box inherently has a label. Like I said, all platforms which have check-box, have label.

jamesmontemagno:
Yeah, I mean I can go either way on this and I think everyone will have a different opinion. I just want the checkbox. I think I have to leave it up to the Forms team to decide what the next steps are here. I can either implement a bunch of font stuff based on what you did or not.

PureWeen:
> I think everyone will have a different opinion. 

Agreed

>  I just want the checkbox.

👍 

> I can either implement a bunch of font stuff based on what you did or not.

I'd say after the *Just a checkbox* is finished and released if there is obviously a big desire for text we can look at adding it, but I'd rather just get the checkbox done and released first. Delaying the PR because of text doesn't seem valid.

andreinitescu:
Thank you. I guess we can always change things later as long as it doesn't break things. 
From my point of view, a check-box without a label makes no sense. Having to add two more controls (StackLayout and Label) is very ugly and I don't see the point. We can do it right.

andreinitescu:
> Delaying the PR because of text doesn't seem valid.

I agree but with little effort this can be done right now. The need for CheckBox is long overdue (just look to the refereneced ticket, it's 1 year old). Adding without a label is not cool in my opinion.  With a bit more effort we can do it better.

jamesmontemagno:
I like the idea of getting it in and then opening up a new PR with some further refactorings that you did and adding a Label later on. 

Seem like a go? I think i just need to add unit tests?

Xamarin.Forms.Core/Cells/CheckBoxCell.cs
```diff
@@ -0,0 +1,44 @@
+using System;
+
+namespace Xamarin.Forms
+{
+	public class CheckBoxCell : Cell
```
hartez: Drop the new Cell types; folks should be using ViewCells with DataTemplates.

PureWeen:
I think part of the comments here still need some discussion.
https://github.com/xamarin/Xamarin.Forms/pull/4369#issuecomment-447592939

### Pressed and Visual State Manager

I could see adding a Pressed event and also a "Pressed" state for Visual State Manager (much like Button and ImageButton has) as the user might want to visually change something while the users finger is on the check box. Also a "Checked" VSM state


### CheckedColor and Unchecked Color changes
I feel like Checked and unchecked color are a bit too vague and we can give a bit more fine grained control over the coloring that matches other coloring apis

I'm wondering if CheckBox should implement IBorderElement and how hard it'd be to achieve those. CornerRadius could be left for later but BorderColor and BackgroundColor seem relevant. This would also align it with CSS properties

The behavior of changing colors baed on Checked and Unchecked would thus be achieved through VSM. With the added "Pressed" state you could define the whole visual experience with VSM

hartez:
> Also a "Checked" VSM state

Yep. State-based colors should be handled by the VSM.

hartez:
> I think part of the comments here still need some discussion.
> [#4369 (comment)](https://github.com/xamarin/Xamarin.Forms/pull/4369#issuecomment-447592939)

I'm of the opinion that a CheckBox without a Content/Text property is incomplete. That said, I'm sympathetic to the "just give us a damn checkbox already" point-of-view, and I can't think of any reason why we couldn't add the Content/Text property in a separate effort. So I'm okay if we do it in two parts.

Xamarin.Forms.Core/Cells/CheckBoxCell.cs
jamesmontemagno: <3 <3 no? All about that TableView :) i can drop it though if it is a requirement.

andreinitescu:
> I like the idea of getting it in and then opening up a new PR with some further refactorings that you did and adding a Label later on.

Sure, use anything you think it's helpful.

andreinitescu:
What's blocking this one? Any help needed?

jamesmontemagno:
The ToDo'sand asks are:

* [ ] Remove all new Cells
* [ ] Add Unit Tests

Anything else @PureWeen and @hartez ?

jamesmontemagno:
@PureWeen Yes, coloring is only available on iOS/Android as they only support it. UWP would need a full XAML item template to accomplish it, which doesn't seem like a good idea.

I am still a bit confused on the Pressed state, but will give it a go.

PureWeen:
> Yes, coloring is only available on iOS/Android as they only support it. UWP would need a full XAML item template to accomplish it, which doesn't seem like a good idea.

Why? full XAML templates are what make UWP far superior to iOS and Android :-)

We do this a few places for example the SliderStyle.xaml, FormsTextBoxStyle.xaml, FormsProgressBarStyle.xaml, etc...

> I am still a bit confused on the Pressed state, but will give it a go.

All of our other "Button" implementations have a pressed so you can use those for reference.  The idea is being able to add style while the touch is on the component but before releasing. For example adding a subtle border change color or something similar to give the user a visual queue they've made contact. It's a useful state to have for the VSM

jamesmontemagno:
Alright, I think we are good now.

I have a single color called "TintColor". Setting this is the overall color of the control regardless of checked or unchecked state.

I added an "IsChecked" state that can be used in the VSM to change the color based on the state. 

Let me know what you think :)

syedhoque89:
This whole comment thread reminds me of current British PM Mrs. Theresa May. 
Stubborn AF. Like @jamesmontemagno said. Lets have it first and get it into the wild.

PureWeen:
- I think instead of TintColor we should just use BackgroundColor unless there's a compelling reason not to that I'm missing. 

> UWP would need a full XAML item template to accomplish it, which doesn't seem like a good idea.

- That's fine we do this for a number of UWP controls

- If you're feeling adventurous can you make a Material checkbox for iOS to match Android :-) if you don't have time we can just do that in a second PR

- can you rebase this to master? rebase instead of merge?

jamesmontemagno:
So, I guess I just think of BackgroundColor as the main background of what is behind the control. It should be similar to the Switch BackgroundColor or else it is inconsistent as it has a BackgroundColor and OnColor.

So I would recommend to keep it TintColor as it represents the border and fill color. Or something similar.

I don't have too much knowledge of UWP to do this so not sure if I feel comfortable adding it.

I think a Material button would be great :) <3 Maybe a new PR :)

Fixed any conflicts.

PureWeen:
> So, I guess I just think of BackgroundColor as the main background of what is behind the control. It should be similar to the Switch BackgroundColor or else it is inconsistent as it has a BackgroundColor and OnColor.

Makes sense. The one caveat though is that BackGroundColor on Android doesn't do anything since it's a fastrenderer and it's not nested inside a ViewGroup. Not sure off hand if there's a way to get that to match up with iOS

Also background across the controls isn't super consistent. Even with switch 

image

Xamarin.Forms.Platform.Android/AppCompat/CheckBoxRenderer.cs
```diff
@@ -0,0 +1,258 @@
...
+		}
+
+
+		AppCompatCheckBox Control => this;
+		public CheckBoxRenderer(Context context) : base(context)
+		{
+			// These set the defaults so visually it matches up with other platforms
+			SetPadding(0, 0, 0, 0);
+			SoundEffectsEnabled = false;
+			SetOnCheckedChangeListener(this);
```
PureWeen: Null this out in dispose

Xamarin.Forms.Platform.Android/AppCompat/CheckBoxRenderer.cs
```diff
@@ -0,0 +1,258 @@
...
+			{
+				CompoundButtonCompat.SetButtonTintList(Control, list);
+				CompoundButtonCompat.SetButtonTintMode(Control, mode);
+			}
+				
+		}
+
+
+		// general state related
+		void IOnFocusChangeListener.OnFocusChange(AView v, bool hasFocus)
```
PureWeen: I think you'll need this in the constructor
```C#
OnFocusChangeListener = this;
```

in order for this to run, and then make sure to null it out in the dispose

Xamarin.Forms.Core.UnitTests/CheckBoxUnitTests.cs
```diff
@@ -0,0 +1,48 @@
...
+
+using NUnit.Framework;
+
+namespace Xamarin.Forms.Core.UnitTests
+{
+	[TestFixture]
+	public class CheckBoxUnitTests : BaseTestFixture
+	{
+		[Test]
+		public void TestConstructor ()
```
hartez: Formatting - drop the space between the method name and the opening parenthesis.

Xamarin.Forms.Platform.Android/AppCompat/CheckBoxRenderer.cs
```diff
@@ -0,0 +1,258 @@
...
+			{
+				oldElement.PropertyChanged -= OnElementPropertyChanged;
+			}
+
+			element.PropertyChanged += OnElementPropertyChanged;
+
+			if (_tracker == null)
+			{
+				_tracker = new VisualElementTracker(this);
+
```
hartez: Formatting - drop the extra blank line.

Xamarin.Forms.Platform.Android/AppCompat/CheckBoxRenderer.cs
```diff
@@ -0,0 +1,258 @@
...
+			((IElementController)CheckBox).SetValueFromRenderer(CheckBox.IsCheckedProperty, isChecked);
+		}
+
+		void UpdateOnColor()
+		{		
+
+			if (CheckBox == null || Control == null)
+				return;
+
+
```
hartez: Formatting - extra blank line.

Xamarin.Forms.Platform.Android/AppCompat/CheckBoxRenderer.cs
```diff
@@ -0,0 +1,258 @@
...
+			var mode = PorterDuff.Mode.SrcIn;
+
+			var stateChecked = global::Android.Resource.Attribute.StateChecked;
+			var stateEnabled = global::Android.Resource.Attribute.StateEnabled;
+			var statePressed = global::Android.Resource.Attribute.StatePressed;
+
+			//Need to find a way to get this color out of Android somehow.
```
hartez: Is this comment still accurate? Is there a color we need to pull from the theme here?

Xamarin.Forms.Platform.Android/AppCompat/CheckBoxRenderer.cs
```diff
@@ -0,0 +1,258 @@
...
+			}
+			else
+			{
+				CompoundButtonCompat.SetButtonTintList(Control, list);
+				CompoundButtonCompat.SetButtonTintMode(Control, mode);
+			}
+				
+		}
+
+
```
hartez: Drop the extra blank lines.

Xamarin.Forms.Platform.Android/AppCompat/CheckBoxRenderer.cs
```diff
@@ -0,0 +1,258 @@
...
+					new int[]
+					{
+						disabledColor,
+						tintColor,
+						tintColor,
+						tintColor,
+						disabledColor,
+					});
+				
+
```
hartez: Drop the extra blank lines.

Xamarin.Forms.Platform.Android/AppCompat/CheckBoxRenderer.cs
```diff
@@ -0,0 +1,258 @@
...
+
+		// general state related
+		void IOnFocusChangeListener.OnFocusChange(AView v, bool hasFocus)
+		{
+			((IElementController)CheckBox).SetValueFromRenderer(VisualElement.IsFocusedPropertyKey, hasFocus);
+		}
+		// general state related
+
+
```
hartez: Drop the extra blank lines.

Xamarin.Forms.Platform.UAP/CheckBoxRenderer.cs
```diff
@@ -0,0 +1,58 @@
...
+			base.OnElementChanged(e);
+
+			if (e.NewElement != null)
+			{
+				if (Control == null)
+				{
+					var control = new WindowsCheckbox();
+					control.Checked += OnNativeChecked;
+					control.Unchecked += OnNativeChecked;
+					//control.ClearValue(WindowsCheckbox.IsCheckedProperty);
```
hartez: Is this intentionally commented out? If so, just remove it; if not, uncomment.

Xamarin.Forms.Platform.WPF/Renderers/CheckBoxRenderer.cs
```diff
@@ -0,0 +1,74 @@
...
+		protected override void OnElementChanged(ElementChangedEventArgs<CheckBox> e)
+		{
+			if (e.NewElement != null)
+			{
+				if (Control == null) // construct and SetNativeControl and suscribe control event
+				{
+					SetNativeControl(new WPFCheckBox());
+					Control.Checked += OnNativeChecked;
+					Control.Unchecked += OnNativeChecked;
+
```
hartez: Drop the blank lines.

Xamarin.Forms.Platform.iOS/Renderers/CheckBoxRenderer.cs
```diff
@@ -0,0 +1,273 @@
...
+			if (sizeConstraint.Request.Height == 0)
+			{
+				if (heightConstraint <= 0 || double.IsInfinity(heightConstraint))
+				{
+					height = DefaultSize;
+					set = true;
+				}
+			}
+
+
```
hartez: Extra blank lines.

Xamarin.Forms.Platform.iOS/Renderers/CheckBoxRenderer.cs
```diff
@@ -0,0 +1,273 @@
...
+				if (heightConstraint <= 0 || double.IsInfinity(heightConstraint))
+				{
+					height = DefaultSize;
+					set = true;
+				}
+			}
+
+			
+
+			if(set)
```
hartez: Formatting - missing space.

Xamarin.Forms.Platform.iOS/Renderers/CheckBoxRenderer.cs
```diff
@@ -0,0 +1,273 @@
...
+
+			if (e.PropertyName == CheckBox.TintColorProperty.PropertyName)
+				UpdateTintColor();
+			else if (e.PropertyName == CheckBox.IsEnabledProperty.PropertyName)
+				Control.IsEnabled = Element.IsEnabled;
+		}
+	}
+
+
+	public class XFCheckBox : UIControl
```
hartez: Move this to its own .cs file. Also, for consistency this should be named `FormsCheckBox` unless there's some conflict or other compelling reason to break with the pattern.

Xamarin.Forms.Platform.iOS/Renderers/CheckBoxRenderer.cs
```diff
@@ -0,0 +1,273 @@
...
+					return;
+
+				_tintColor = value;
+				SetNeedsDisplay();
+			}
+		}
+
+		public override void Draw(CGRect rect)
+		{
+			//base.Draw(rect);
```
hartez: If this is intentional, remove the commented code; otherwise, uncomment.

StephaneDelcroix:
@jamesmontemagno is this still WIP ?

PureWeen:
@StephaneDelcroix yea I've taken it over for now

PureWeen:
rebased, applied @hartez changes, added material, and fixed a few other quirks

Also moved the PR over here
https://github.com/xamarin/Xamarin.Forms/pull/6232

I didn't want to break this one with the rebase and it just makes it easier to merge it into a feature branch

andreinitescu:
@hartez
> I'm of the opinion that a CheckBox without a Content/Text property is incomplete. That said, I'm sympathetic to the "just give us a damn checkbox already" point-of-view, and **I can't think of any reason why we couldn't add the Content/Text property in a separate effort.** So I'm okay if we do it in two parts.

Are we sure adding Text property later is doable? Once it gets released, it would be hard to change any public behavior of the class.

andreinitescu:
Other issues I see with releasing CheckBox without label:
- Accesibility (important)
- A separate `Label` control near the `CheckBox` makes the `Label` look disconnected. There's no visual feedback on tap.

samhouts:
@andreinitescu I think we can safely add that in later. It would be an option to show or hide it. I've seen designs as recently as today that don't require a text/content property. Your concern about the visual feedback is valid, and I'd also expect to be able to toggle the checked state when clicking the associated label. For accessibility purposes, we do already have `LabeledBy`, so we can get by with that for now.

We can continue this conversation on #6232 and/or a new issue specifically for this feature.

Thanks!

andreinitescu:
>  I've seen designs as recently as today that don't require a text/content property

Thanks for the feedback!
I think it's easier to have a Text property and just leave it empty for the scenarios you mentioned, than not having a Text property and when you need it, being forced to add two more controls, a horizontal StackLayout and a Label.

NickeManarin:
Please, add a Text property. Also, add for a Radio button too. It does not makes sense to have a checkbox without any text.

PR #4375, opened 2018-11-11
RelativeSource Binding

### LINKED ISSUE #3847: [F100] Add support for RelativeSource binding
## Rationale

https://forums.xamarin.com/discussion/20179/databinding-to-a-command-inside-a-datatemplate#latest

http://xamarin.uservoice.com/forums/258559/suggestions/6451625

## Implementation

Like UWP? https://docs.microsoft.com/en-us/windows/uwp/xaml-platform/relativesource-markup-extension

legistek:
### Description of Change ###

Implements `RelativeSource` binding mechanism, intended to mimic WPF/UWP behavior and syntax. `Self`, `TemplatedParent`, and `FindAncestor` modes are implemented. (`PreviousData` is much trickier and is left for another day.)

### Issues Resolved ### 
- fixes #3847 

### API Changes ###
Added:

New items:
- `class RelativeBindingSource` (equivalent to WPF/UWP `RelativeSource` class
- `enum RelativeBindingSourceMode` (equivalent to WPF/UWP `RelativeSourceMode` enum
- `class RelativeSourceExtension`

- in `Binding`:
- `public RelativeBindingSource RelativeSource`

- in `BindingExtension`:
- `public RelativeBindingSource RelativeSource`

Changed:
- `TemplateBinding` is now derived from `Binding` and uses `RelativeSource` mechanism rather than directly deriving from `BindingBase`

### Platforms Affected ### 
- Core/XAML (all platforms)

There should be no changes to existing applications. Developers can now use `RelativeSource` binding in new applications.

Not applicable

### Testing Procedure ###
Unit tests are provided.

### PR Checklist ###

- [X] Has automated tests 


StephaneDelcroix:
I haven't reviewed this yet (and won't be this week), but would it make sense to have `Source={RelativeSource ...}` or does Source and RelativeSource can exists together ?

hartez:
Looks like some of the unit tests are failing:

`BindingBaseUnitTests.ReuseBindingInstance` and `NativeBindingTests.ThrowsOnMissingProperty`

legistek:
@hartez  I figured out what's going on. I had to declare `Binding.Apply` as `async void` in order to implement `TemplatedParent` and `FindAncestor` modes (just like `TemplateBinding` did before). The exceptions were getting thrown correctly but apparently not in the same thread so the unit tests were getting tricked. I'm working on a fix but it's definitely a false positive at this point.

legistek:
@hartez  ok that should be better now. The `async` part only gets invoked if `RelativeSource` is used so those unit tests are now triggering the exception in the main thread and getting caught by the unit test as before.

I also tried out @StephaneDelcroix's idea for allowing `Source={RelativeSource...}`. So that works interchangeably with `RelativeSource={RelativeSource...}` now.

Finally, a colleague of mine had what I think is an awesome idea to have the `FindAncestor` mode match the ancestor's `BindingContext` type OR the ancestor's own type. If it matches by `BindingContext` type the source resolves to the `BindingContext` value. That seems even more useful than matching to the view/control type and very convenient for MVVM. A typical use case would be when you have a `DataTemplate` for items in a `ListView`, the `BindingContext` is of course the individual item in the list, but you could use `FindAncestor AncestorType={x:Type vm:MyParentVIewModelType}` to bind to properties/commands in the parent view model rather than the item view model, without having to know or care what the parent's view type is. LMK what you guys think.

Xamarin.Forms.Core/RelativeBindingSourceMode.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Xamarin.Forms
+{
+	public enum RelativeBindingSourceMode
+	{
+		// TODO? (much tricker than the other three)
+		// PreviousData = 0,
```
hartez: Is this intentionally commented out?

StephaneDelcroix:
if `Source` and `RelativeSource` are equivalent semantically, drop the `RelativeSource`

Xamarin.Forms.Core/RelativeBindingSourceMode.cs
legistek: Yes I meant it as a placeholder in case someone wanted to tackle that down the road.

legistek:
@StephaneDelcroix  I certainly can. My thinking was just to stay semantically consistent with WPF and UWP, making it easier to add this to XAML Standard, etc.,  But it's up to you.

Xamarin.Forms.Core/TemplateBinding.cs
```diff
@@ -3,20 +3,17 @@
 
 namespace Xamarin.Forms
 {
-	public class TemplateBinding : BindingBase
+	// Theoretically this class shouldn't be necessary given
+	// Binding.RelativeSource and TemplateBindingExtension 
+	// but it's here to avoid breaking changes.
+	public sealed class TemplateBinding : Binding
```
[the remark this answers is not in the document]
legistek: @StephaneDelcroix yeah I don't really love it myself, but I love redundant code less. :) Alternatively, we could just leave `TemplateBinding` alone and mark it as `Obsolete`. There's really no need for it since you have the `TemplateBindingExtension` for XAML and in code you can use regular `Binding` with `RelativeSource`. But of course we don't know if it's been used in the field.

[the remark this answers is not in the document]
StephaneDelcroix: I agree, let's [Obsolete]

legistek:
@StephaneDelcroix  ok please take a look at the newest commit. I made `Binding` `sealed` again and obsoleted `TemplateBinding` (also changed anywhere it was being used to use the new format).

I also discovered that `FindAncestor` bindings weren't getting re-applied when a distant ancestor was changed but the `BindingContext` wasn't, and so had to make a few more code changes to accomodate that. While doing that I also saw that `TemplatedParent` and `FindAncestor` bindings were getting re-applied way more often than necessary (neither should be affected by a `BindingContext` change, and `FindAncestor` only needs to be reapplied when there's an ancestor change). So I addressed both those issues. Let me know if you want any more changes or don't like the new changes.

Xamarin.Forms.Core/Binding.cs
```diff
@@ -113,23 +116,80 @@ internal override void Apply(bool fromTarget)
...
 			var isApplied = IsApplied;
 
 			base.Apply(src ?? context, bindObj, targetProperty, fromBindingContextChanged: fromBindingContextChanged);
 
-			if (src != null && isApplied && fromBindingContextChanged)
+			if (isApplied && !ShouldReapply(fromBindingContextChanged, fromAncestorChanged))
 				return;
 
-			object bindingContext = src ?? Context ?? context;
+			if ( this.Source is RelativeBindingSource )
```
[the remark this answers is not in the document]
StephaneDelcroix: this is weird that the type check (`is`) is located here, and the conversion is in `ResolveRelativeSource`. I would drop the check here, and check in `ResolveRelativeSource` and return is Source isn't a RelativeSource

Xamarin.Forms.Core/Binding.cs
```diff
@@ -113,23 +116,80 @@ internal override void Apply(bool fromTarget)
...
+									break;
+								}
+							}
+							parent = await TemplateUtilities.GetRealParentAsync(parent);
+							currentLevel++;
+						}
+						break;
+					};
+				default:
+					throw new NotImplementedException();
```
StephaneDelcroix: what is the part left unimplemented ? shouldn't you throw InvalidOperationException if the enum is out of the range ?

Xamarin.Forms.Core/Binding.cs
```diff
@@ -113,23 +116,80 @@ internal override void Apply(bool fromTarget)
...
+			{
+				case RelativeBindingSourceMode.Self:
+					bindingContext = bindObj;
+					break;
+				case RelativeBindingSourceMode.TemplatedParent:
+					{
+						var view = bindObj as Element;
+						if (view == null)
+							throw new InvalidOperationException();
+						bindingContext = await TemplateUtilities.FindTemplatedParentAsync(view);
```
StephaneDelcroix: can you explain why this is async ? and if it's right to have it async, you'll need `ConfigureAwait(false)`

Xamarin.Forms.Core/Binding.cs
```diff
@@ -113,23 +116,80 @@ internal override void Apply(bool fromTarget)
...
+						break;
+					}
+				case RelativeBindingSourceMode.FindAncestor:
+					{
+						if (!(bindObj is Element elem))
+							throw new InvalidOperationException();
+						Element parent = await TemplateUtilities.GetRealParentAsync(elem);
```
StephaneDelcroix: same, why async ?

Xamarin.Forms.Core/Binding.cs
```diff
@@ -113,23 +116,80 @@ internal override void Apply(bool fromTarget)
...
+			object bindingContext = null;
+			switch (relSource.Mode)
+			{
+				case RelativeBindingSourceMode.Self:
+					bindingContext = bindObj;
+					break;
+				case RelativeBindingSourceMode.TemplatedParent:
+					{
+						var view = bindObj as Element;
+						if (view == null)
```
StephaneDelcroix: use pattern matching

Xamarin.Forms.Core/Binding.cs
```diff
@@ -113,23 +116,80 @@ internal override void Apply(bool fromTarget)
...
 
-			if (src != null && isApplied && fromBindingContextChanged)
+			if (isApplied && !ShouldReapply(fromBindingContextChanged, fromAncestorChanged))
 				return;
 
-			object bindingContext = src ?? Context ?? context;
+			if ( this.Source is RelativeBindingSource )
+				ResolveRelativeSource(bindObj, targetProperty);
+			else
+				CompleteApplyBinding(src ?? Context ?? context, bindObj, targetProperty);
```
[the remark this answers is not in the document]
StephaneDelcroix: one more reason to make this not async

Xamarin.Forms.Core/Binding.cs
```diff
@@ -113,23 +116,80 @@ internal override void Apply(bool fromTarget)
...
+						while (parent != null)
+						{
+							if (currentLevel >= relSource.AncestorLevel)
+							{
+								if (relSource.AncestorType.IsInstanceOfType(parent))
+								{
+									bindingContext = parent;
+									break;
+								}
+								else if (relSource.AncestorType.IsInstanceOfType(parent.BindingContext))
```
StephaneDelcroix: is this how it behaves in other platforms ? I understand it's convenient, but is it the least surprising way ?

Xamarin.Forms.Core/Binding.cs
```diff
@@ -113,23 +116,80 @@ internal override void Apply(bool fromTarget)
...
+								{
+									bindingContext = parent.BindingContext;
+									break;
+								}
+							}
+							parent = await TemplateUtilities.GetRealParentAsync(parent);
+							currentLevel++;
+						}
+						break;
+					};
```
StephaneDelcroix: what if the binding context isn't found ? should we log ?

Xamarin.Forms.Core/Binding.cs
```diff
@@ -139,7 +199,7 @@ internal override BindingBase Clone()
 				Source = Source,
 				UpdateSourceEventName = UpdateSourceEventName,
 				TargetNullValue = TargetNullValue,
-				FallbackValue = FallbackValue,
+				FallbackValue = FallbackValue
```
StephaneDelcroix: nope, that was correct :) and on purpose

Xamarin.Forms.Core/Binding.cs
```diff
@@ -159,17 +219,42 @@ internal override object GetTargetValue(object value, Type sourcePropertyType)
 			return base.GetTargetValue(value, sourcePropertyType);
 		}
 
-		internal override void Unapply(bool fromBindingContextChanged = false)
+		internal override void Unapply(bool fromBindingContextChanged = false, bool fromAncestorChange = false)
 		{
-			if (Source != null && fromBindingContextChanged && IsApplied)
+			if ( this.IsApplied && !ShouldReapply(fromBindingContextChanged, fromAncestorChange))
```
StephaneDelcroix:
```suggestion
			if (IsApplied && !ShouldReapply(fromBindingContextChanged, fromAncestorChange))
```

StephaneDelcroix: move the `IsApplied` check into `ShouldReapply`

Xamarin.Forms.Core/RelativeBindingSource.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Xamarin.Forms
+{
+	public class RelativeBindingSource
```
StephaneDelcroix:
```suggestion
	public sealed class RelativeBindingSource
```

Xamarin.Forms.Core/RelativeBindingSource.cs
```diff
@@ -0,0 +1,67 @@
...
+		{
+			get;
+			set;
+		}
+
+		public Type AncestorType
+		{
+			get => _ancestorType;
+			set
+			{
```
StephaneDelcroix: setter shouldn't modify 2 properties

Xamarin.Forms.Core/RelativeBindingSource.cs
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		public int AncestorLevel
+		{
+			get => _ancestorLevel;
+			set
+			{
+				_ancestorLevel = value;
+				if (_ancestorLevel > 0)
+					this.Mode = RelativeBindingSourceMode.FindAncestor;
```
StephaneDelcroix: same

Xamarin.Forms.Xaml/MarkupExtensions/BindingExtension.cs
```diff
@@ -18,6 +18,13 @@ public sealed class BindingExtension : IMarkupExtension<BindingBase>
 		public object FallbackValue { get; set; }
 		public TypedBindingBase TypedBinding { get; set; }
 
+		// Here to maintain parity with other XAML flavors.
+		public RelativeBindingSource RelativeSource
+		{
+			get => this.Source as RelativeBindingSource;
+			set => this.Source = value;
+		}
+
```
StephaneDelcroix: please remove this

Xamarin.Forms.Core/Binding.cs
legistek: Yeah so this was how it was being done before, just inside `TemplateBinding`. `TemplateBinding.Apply` was an `async void`. The ultimate culprit is `TemplateUtilities.GetRealParentAsync` (which `FindAncestor` also needs) and that is using a `TaskCompletionSource` and an event hook in order to `await` the setting of the parent, presumably because the binding might be applied before the parent is set. 

I'm assuming there are other ways around this, but I was trying to change as little as necessary. I'll give it a shot if you think it's worth trying to eliminate, but it looks like template binding has worked this way for a few years now so I don't know if you want me to mess with it.

Xamarin.Forms.Core/Binding.cs
legistek: Good question, and no it's not. This would be an enhancement over WPF/UWP. But I too was starting to have second thoughts about the "surprising" nature of it and so was thinking maybe adding another type of 'RelativeBindingSourceMode' in its own right, e.g., `FindAncestorBindingContext`. Then there's no ambiguity that you're looking to match by the parent binding context rather than the parent's view type. Would that be better?

Xamarin.Forms.Core/Binding.cs
legistek: Oops inadvertent :)

Xamarin.Forms.Xaml/MarkupExtensions/BindingExtension.cs
legistek: Ok you're the boss, boss, But I just want to make sure we're 100% clear that this would deviate from the WPF/UWP syntax. If it's the redundancy that bothers you, as an XForms consumer I'd rather require the use of `RelativeSource=` than make `Source=` and `RelativeSource=` interchangeable, just to minimize differences between the platforms. Please confirm?

legistek:
Thanks @StephaneDelcroix  please see a few places where I replied, particularly about `async`. It might be possible to eliminate it but I'm using the same `async` method as the existing `TemplateBinding` does to find the parent, so to get rid of it would need a number of additional changes that might introduce subtle issues.

legistek:
@StephaneDelcroix sorry for the delay (and that I couldn't finish it in time for 3.5). Eliminating the async was a little tricky but I think I got it. 
Since these are new APIs what if anything do you want me to do about documentation? Believe it or not I actually like writing it. :)

legistek:
Ok @StephaneDelcroix please review again at your convenience. The only thing I want to alert you about is the `async` issue. It turns out getting rid of it is probably not practical, but after exploring alternatives I think it's actually the most efficient approach.

The fundamental problem is that bindings get applied before children are assigned to parents. This has always been a problem for `TemplateBinding` and also winds up being a problem for `RelativeSource`. I thought of 3 ways to address this: 
1) Using `async` and `await`, as `TemplateBinding` has always worked (discussed more below)
2) Propagating parent assignments down the entire visual tree every time one occurs, and re-applying bindings as necessary. This does work, but the problem is that as the tree is being built, you will have an exponentially growing number of recursive operations that would grind large applications to a halt. Preventing excessive and unnecessary re-applications of bindings would also be a challenge.
3) Do not apply any `RelativeSource` bindings until the visual tree is built. This would require me to mess around in a lot more places than I think you want, and I'm not even sure is possible without pretty much re-writing the entire visual tree generation process.

The really interesting thing about the `async` approach is that it still winds up all executing on one thread. As the visual tree is built, elements are sending out `ParentSet` event notifications to whomever is listening. Those of course get executed synchronously. `TemplateUtilities.GetRealParentAsync` works by subscribing to the `ParentSet` event on the desired element, and `await`ing the triggering of that event using a `TaskCompletionSource`. When that happens execution temporarily jumps back to `GetRealParentAsync`, which ultimately returns you to `async void Binding.ApplyRelativeSourceAsync`. When that completes, execution returns to building the visual tree. So the execution path does jump around but it does so on one thread and in the same way that it would under method (2), except that the notifications only go out only to whoever is listening, rather than propagating down the whole tree every time.

Thus I think this is actually the cleanest and most efficient way of doing this given the rest of the code base. If we were totally rewriting the way the visual tree is built, we could perhaps do better, but I do think this is the best we can do with what we've got now. Please let me know what you think about this and the rest of the PR. Thanks!

andreinitescu:
I wish this got more love...

@legistek Could you please resolve conflict and squash commits so it's easier to check

ddobrev:
Fellows, I thank all of you for this effort. This issue can often be worked around with {x:Reference} but it's quite inconvenient. I have therefore been expecting this feature for a long time now. Would you have an idea when it's going to be released?

legistek:
@StephaneDelcroix I figured out a way to eliminate any reliance on `async` or `await`. In the process I made `TemplatedParent` binding a lot more efficient than before.  All the unit tests are passing for me so please review and let me know if you want any changes.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Github3847.xaml.cs
```diff
@@ -0,0 +1,78 @@
...
+		public string FullName => $"{FirstName} {LastName}";
+	}
+
+	public class CompanyViewModel
+	{
+		public string Name { get; set; }
+
+		public ObservableCollection<PersonViewModel> Employees { get; set; }
+
+		Command _DeleteEmployeeCommand;
```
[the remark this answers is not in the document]
legistek: Done thanks.

legistek:
Just an FYI this needs to be merged in first before I can PR #4565 because making `RelativeSource` binding work for the children of the `MultiBinding` is gonna take some special voodoo magic.

Xamarin.Forms.Core/Binding.cs
```diff
@@ -124,11 +129,117 @@ internal override void Apply(object context, BindableObject bindObj, BindablePro
...
+						targetProperty, 
+						currentElement, 
+						currentLevel,
+						chain);
+				};
+				currentElement.ParentSet += onElementParentSet;
+			}			
+		}
+
+		private bool ElementFitsAncestorTypeAndLevel(Element element, int level)
```
StephaneDelcroix:
```suggestion
		bool ElementFitsAncestorTypeAndLevel(Element element, int level)
```

Xamarin.Forms.Core/BindingExpression.cs
```diff
@@ -458,6 +469,105 @@ internal static bool TryConvert(ref object value, BindableProperty targetPropert
...
 		}
 
+		internal void SubscribeToTemplatedParentChanges(Element target, BindableProperty targetProperty)
+		{
+			_targetProperty = targetProperty;
+			target.TemplatedParentChanged += OnTargetTemplatedParentChanged;
+			_trackingTemplatedParent = true;
+		}
+
+		private void OnTargetTemplatedParentChanged(object sender, EventArgs e)
```
StephaneDelcroix:
```suggestion
		void OnTargetTemplatedParentChanged(object sender, EventArgs e)
```

Xamarin.Forms.Core/BindingExpression.cs
```diff
@@ -458,6 +469,105 @@ internal static bool TryConvert(ref object value, BindableProperty targetPropert
...
+				return;
+			_ancestryChange = new List<WeakReference<Element>>();
+			foreach (var elem in chain)
+			{
+				elem.ParentSet += OnElementParentSet;
+				_ancestryChange.Add(new WeakReference<Element>(elem));
+			}
+		}
+
+		private void ClearAncestryChangeSubscriptions(int beginningWith = 0)
```
StephaneDelcroix:
```suggestion
		private void ClearAncestryChangeSubscriptions(int beginningWith = 0)
```

Xamarin.Forms.Core/BindingExpression.cs
```diff
@@ -458,6 +469,105 @@ internal static bool TryConvert(ref object value, BindableProperty targetPropert
...
+				var weakElement = _ancestryChange.Last();
+				if (weakElement.TryGetTarget(out elem))
+					elem.ParentSet -= OnElementParentSet;
+				_ancestryChange.RemoveAt(_ancestryChange.Count - 1);
+			}
+		}
+
+		// Returns -1 if the member is not in the chain or the
+		// chain is no longer valid.
+		private int FindAncestryIndex(Element elem)
```
StephaneDelcroix:
```suggestion
		private int FindAncestryIndex(Element elem)
```

Xamarin.Forms.Core/BindingExpression.cs
```diff
@@ -458,6 +469,105 @@ internal static bool TryConvert(ref object value, BindableProperty targetPropert
...
+				Element chainMember = null;
+				if (!weak.TryGetTarget(out chainMember))
+					return -1;
+				else if (object.Equals(elem, chainMember))
+					return i;
+			}
+			return -1;
+		}
+
+		private void OnElementParentSet(object sender, EventArgs e)
```
StephaneDelcroix:
```suggestion
		void OnElementParentSet(object sender, EventArgs e)
```

Xamarin.Forms.Core/Element.cs
```diff
@@ -208,9 +208,37 @@ public Element Parent
...
+					this.TemplatedParent = value;
 			}
 		}
 
+		internal event EventHandler TemplatedParentChanged;
+		
+		BindableObject _templatedParent;
+		public BindableObject TemplatedParent
+		{
+			get => _templatedParent ?? (_templatedParent = this.Parent?.TemplatedParent);
```
StephaneDelcroix: this will not behave well with reparenting. I don't think this value should be cached when computed
```suggestion
			get => _templatedParent ?? (Parent?.TemplatedParent);
```

Xamarin.Forms.Core/Element.cs
```diff
@@ -208,9 +208,37 @@ public Element Parent
 				OnParentSet();
 
 				OnPropertyChanged();
+
+				if (this.IsTemplateRoot)
+					this.TemplatedParent = value;
 			}
 		}
 
+		internal event EventHandler TemplatedParentChanged;
```
StephaneDelcroix: why not use PropertyChanged ?
```suggestion
```

Xamarin.Forms.Core/Element.cs
```diff
@@ -208,9 +208,37 @@ public Element Parent
...
+		BindableObject _templatedParent;
+		public BindableObject TemplatedParent
+		{
+			get => _templatedParent ?? (_templatedParent = this.Parent?.TemplatedParent);
+			private set
+			{
+				if (_templatedParent == value)
+					return;
+				_templatedParent = value;
+				TemplatedParentChanged?.Invoke(this, null);
```
StephaneDelcroix:
```suggestion
				OnPropertyChanged();
```

Xamarin.Forms.Core/Element.cs
```diff
@@ -208,9 +208,37 @@ public Element Parent
...
+				foreach (var element in this.Descendants())
+				{
+					if (value == null)
+						// Item has been detached from its templated 
+						// parent, so reset all the stored values
+						element._templatedParent = null;
+					element.TemplatedParentChanged?.Invoke(element, null);
```
StephaneDelcroix:
```suggestion
					element.OnPropertyChanged();
```
need to doublecheck this one ^^

Xamarin.Forms.Core/TemplateBinding.cs
```diff
@@ -3,7 +3,8 @@
 
 namespace Xamarin.Forms
 {
-	public class TemplateBinding : BindingBase
+	[Obsolete("Use Binding.Source=RelativeBindingSource.TemplatedParent")]
+	public sealed class TemplateBinding : BindingBase
```
StephaneDelcroix: ❤️

Xamarin.Forms.Xaml/MarkupExtensions/TemplateBindingExtension.cs
```diff
@@ -24,7 +24,15 @@ public TemplateBindingExtension()
 
 		BindingBase IMarkupExtension<BindingBase>.ProvideValue(IServiceProvider serviceProvider)
 		{
-			return new TemplateBinding(Path, Mode, Converter, ConverterParameter, StringFormat);
+			return new Binding
```
StephaneDelcroix: shouldn't we obsolete TemplateBindingExtension ?

Xamarin.Forms.Xaml/MarkupExtensions/RelativeSourceExtension.cs
```diff
@@ -0,0 +1,83 @@
...
+		{
+			get;
+			set;
+		}
+
+		RelativeBindingSource IMarkupExtension<RelativeBindingSource>.ProvideValue(IServiceProvider serviceProvider)
+		{
+			if (AncestorType != null)
+			{
+				RelativeBindingSourceMode mode;
```
StephaneDelcroix: rename to `actualMode`, or `computedMode`, or something

legistek:
Great! I'll finish it up and rebase this week.

> next step once this is committed is to allow compilation of bindings with relative binding sources

Do you mean in XamlC, identifying the relativesource parent at compile time instead of only looking for it at runtime when the binding is applied?

Xamarin.Forms.Core/Element.cs
legistek: I have it clearing the cached template parent for all the `Descendants()` when the actual parent changes, so that should cover reparenting, no? There's one or two older unit tests that test for this too and I made sure they passed.

The reason I wanted to cache it was if folks (like me!) want to make more use of templated controls over native controls it would be a lot more efficient to apply bindings on 100s or 1000s of controls if we didn't have to crawl the tree every time we want to get the templated parent.

Xamarin.Forms.Core/Element.cs
```diff
@@ -208,9 +208,37 @@ public Element Parent
...
+
+				if (this.IsTemplateRoot)
+					this.TemplatedParent = value;
 			}
 		}
 
+		internal event EventHandler TemplatedParentChanged;
+		
+		BindableObject _templatedParent;
+		public BindableObject TemplatedParent
```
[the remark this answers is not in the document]
legistek: No because `TemplatedParent` is populated for every `Element` created by a `ControlTemplate`. It's what `Binding.Apply` uses to locate the binding source in `TemplateBinding`s. So every `Element` potentially needs this member in order to eliminate the previous `async void` mechanism of crawling up the tree looking for the `IControlTemplated` every time a binding is applied.

And frankly if we were being thorough this would be populated for `Element`s made by `DataTemplate`s as well but I didn't want to go outside the scope of this issue.

Xamarin.Forms.Core/Element.cs
```diff
@@ -208,9 +208,37 @@ public Element Parent
...
+					if (value == null)
+						// Item has been detached from its templated 
+						// parent, so reset all the stored values
+						element._templatedParent = null;
+					element.TemplatedParentChanged?.Invoke(element, null);
+				}
+			}
+		}
+
+		internal bool IsTemplateRoot { get; set; }
```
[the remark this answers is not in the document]
legistek: `IsTemplateRoot` tells us whether this particular element is the `TemplateRoot` of the control. It's used to determine whether to propagate `TemplatedParent` changes down the tree.

Or more precisely:
`IControlTemplated.TemplateRoot.IsTemplateRoot === true`

legistek:
@samhouts Please don't merge until I finish addressing @StephaneDelcroix 's comments - thanks!

Xamarin.Forms.Core/Element.cs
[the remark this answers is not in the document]
andreinitescu: Yes, it does same thing with what `IControlTemplate`, that's my point. Also, not all elements are templated, hence I don't see why Element should have those properties.

Xamarin.Forms.Core/Element.cs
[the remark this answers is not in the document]
andreinitescu: Oh, I see, that's great! The existing mechanism has been very "itchy" to me since I first saw it. Thanks for reply.

Xamarin.Forms.Core/Element.cs
legistek: We could. I thought it would be more efficient this way because it's a highly specific event that is usually only going to be triggered once per `Element`, and will have a relatively small number of listeners compared to `PropertyChanged`.

Xamarin.Forms.Xaml/MarkupExtensions/TemplateBindingExtension.cs
legistek: > shouldn't we obsolete TemplateBindingExtension ?
@StephaneDelcroix we're obsoleting `TemplateBinding` in Core for use in code, but still want to keep the `Property="{TemplateBinding Path}"` syntax in XAML don't we?

legistek:
Ok all done. @StephaneDelcroix  please let me know if you disagree with any of my responses to your comments. I did change the way `TemplatedParent` was computed just a bit to make sure re-parenting behaved correctly, and added unit test cases for that. I would keep the special `internal` event `TemplatedParentChanged` for `BindingExpression` to use efficiently but did add the `OnPropertyChanged` notification too in case other subscribers want to know if that property's changed - of course this is all an extreme edge case anyway.

Xamarin.Forms.Core/Element.cs
StephaneDelcroix: ok

Xamarin.Forms.Xaml/MarkupExtensions/TemplateBindingExtension.cs
StephaneDelcroix: if you think so, I'm fine

vincentcastagna:
Is this supposed to behave the same as `{Binding BindingContext.IsBusy, Source={x:Reference This}}` ? Where `This` is the page containing the `ControlTemplate` for example ?

legistek:
@vincentcastagna no, I believe `x:Reference` is used to refer to specific named elements in the same name scope (like the same `ContentPage`). `RelativeSource` binding is about locating a binding source anywhere in the visual tree at or above the item being bound to.

`Self` mode would set the binding source to be the element receiving the binding itself. For instance if you wanted to design a control that automatically linked two properties like `BorderColor` and `BackgroundColor`, you could make a `Style` that included `<Setter Property="BorderColor" Value="{Binding BackgroundColor, Source={RelativeSource Self}"/>`

`FindAncestor` sets the binding source to be some element in the visual tree that's an ancestor of the target element, based on the ancestor type. This is where you could potentially say `{Binding BindingContext.IsBusy, Source={RelativeSource FindAncestor}, AncestorType={x:Type ContentPage}}` and that would get you what you're going for in your example. And in an enhancement to the WPF version, we also added the ability to select that ancestor based on the `BindingContext` type of the ancestor, rather than the type of visual element, which makes it easier to do pure MVVM in parent-child situations.

And finally `TemplatedParent` is identical to `TemplateBinding` because the latter is now just a markup extension as opposed to being its own `BindingBase`-derivative.

IoTFier:
Hi @legistek , this is a great feature. thank you.  Thanks to @PureWeen . He pointed out this to me yesterday. 

**Is there a sample repo that demonstrates every options provided in this PR?**

...........................................................
I am trying to use this in my App, as I write this note. I am using XF 4.3 and Prism 7.2
This is the hierarchy of my Page and contentviews.
MainPage.xaml (contentpage) | ViewModel - MainPageViewModel.cs
--> SfListViewTemplate.xaml (contentview) | uses syncfusion listview
    --> DataListItemTemplate.xaml (contentview) | used for SfListView's ItemTemplate
        --> BusinessRecordLine1.xaml (contentview)
        --> BusinessRecordLine2.xaml (contentview)

Using prism Dialog service I want to show a dialog when an ImageButton / Label on BusinessRecordLine1.xaml is clicked / Tapped by the user. I have defined a DelegateCommand in the viewmodel of the MainPage.xaml and have been trying to attach it to an ImageButton and a Label defined in the BusinessRecordLine1.xaml. 

But, the clicking of that ImageButton is not getting wired to the DelegateCommand when wired it using EventToCommandBehavior . 
**That's where I am trying to use this feature of XF 4.3 from this PR**
I would truly appreciate your thoughts. 
Thank you.

legistek:
@IoTFier to answer your first question there are examples in the unit tests and control gallery page for the original issue (3847). For instance `GH3847.xaml` in `Xamarin.Forms.Xaml.UnitTests > Issues` has several examples.  (You don't need the `x:` in `x:RelativeSource` - just say `RelativeSource` - but I don't think it is harmful).

As for your issue, is the actual type name of your view model `MainPg` or `MainPageViewModel`? The `AncestorType` parameter takes the CLR type of the view model, not the `x:Name`. So I think you want to say `AncestorType={x:Type local:MainPageViewModel}`. (Where `local` is your XML namespace prefix).

IoTFier:
@legistek , the name of the viewmodel is MainPageViewModel.

In the ContentView (BusinessRecordLine1.xaml) even after declaring the XML namespace for the view model, the intellisense fails to recognize that type. I still provided that name, but even then it didn't wire up that DelegateCommand. Apparently, my app didn't crash on either iOS or Android Simulators 

            <Label.GestureRecognizers>
                <TapGestureRecognizer
                    Command="{Binding Path=DisplayDialogCommand, Source={x:RelativeSource Mode=FindAncestorBindingContext, AncestorType={x:Type pgVM:MainPageViewModel}}}"
                    CommandParameter="{Binding .}"
                    NumberOfTapsRequired="1" />
            </Label.GestureRecognizers>

legistek:
Your XAML looks correct now. Unfortunately I'm not familiar with Syncfusion's controls so I can't say for sure what the problem might be. 

If you want to post a project with the *minimum* content necessary to reproduce I can try to take a look.

IoTFier:
attn - @davidbritch , @samhouts , @StephaneDelcroix , @hartez , @andreinitescu 
............................................................
Hi @legistek - I have confirmed where the bug is. 

In XF 4.3, the RelativeSource Binding fails to identify/detect ancestor types, if they have a default constructor with parameter(s) (and wired to IoC containers). 

Reference - [Bind from a ControlTemplate to a ViewModel with Xamarin.Forms](https://www.davidbritch.com/2019/11/bind-from-controltemplate-to-viewmodel.html)

_This could also be a cause why the inherited BindingContext is not available and accessible to all ViewElements across the hierarchy below the ContentPage._

I am using prism v7.2 (MVVM framework) to register my ContentPages and Corresponding ViewModels. There is a base ViewModel which takes a parameter for INavigationService and IDialogService. As a result, RelativeSource binding is able to detect only those View Models having a default constructor with no parameter although my prism based view models exist in the same xml namespace. 

Same thing applies to getting a handle on the ContentPage when I set the Mode = FindAncestorBindingContext & try to invoke a command method in the code behind of a ContentPage that can use the DisplayAlert method on that page to display an alert message.

Would like to hear from @davidbritch and/or any other core member from Xamarin Forms family who can validate this before I file a github issue.

Thank you.

legistek:
@IoTFier what makes you believe the number of parameters in the view model constructor is the issue?

If you look in [Binding.cs](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/Binding.cs) line 243, we use `IsAssignableFrom` to test whether the type of the `BindingContext` matches the `AncestorType`. I'm not aware of any reason (and none is given in the .NET docs) why the number of constructors in either type would matter to `IsAssignableFrom`. 

Perhaps you could construct a minimal project without using third party controls that replicates the issue just by using view models without parameterless constructors?

If you're right though then the workaround should be to just use the `FindAncestor` mode rather than `FindAncestorBindingContext` mode. Set the `AncestorType` to your `MainPage` and add `BindingContext.` at the beginning of the path. That should be equivalent (and it would also tell us whether it's an issue with the way we're navigating up the parent tree).

legistek:
@IoTFier also is it possible the binding context is assigned after the visual tree is built, or changed at runtime? I'm looking back at the code here and I don't see anything that would catch when the `BindingContext` of the ancestor changes. So in your case it might be looking for the `MainPageViewModel`, initially not finding anything with that binding context in the visual tree. Later when the view model is assigned nothing would happen because we're only listening for _parent_ changes, not parent _BindingContext._ changes. If that's the case, then the workaround I proposed a few minutes ago should work (and also we do seem to have a bug).

IoTFier:
> @IoTFier what makes you believe the number of parameters in the view model constructor is the issue?
> 
> If you look in [Binding.cs](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/Binding.cs) line 243, we use `IsAssignableFrom` to test whether the type of the `BindingContext` matches the `AncestorType`. I'm not aware of any reason (and none is given in the .NET docs) why the number of constructors in either type would matter to `IsAssignableFrom`.
> 
> Perhaps you could construct a minimal project without using third party controls that replicates the issue just by using view models without parameterless constructors?
> 
> If you're right though then the workaround should be to just use the `FindAncestor` mode rather than `FindAncestorBindingContext` mode. Set the `AncestorType` to your `MainPage` and add `BindingContext.` at the beginning of the path. That should be equivalent (and it would also tell us whether it's an issue with the way we're navigating up the parent tree).

@legistek - I have confirmed the bug on my end against the test case(s) that I mentioned. Will try to create a repro (even using the sample that @davidbritch has posted).  I am not changing / resetting / assigning BindingContext anywhere besides setting them in App.cs as mentioned in the para below. I have been digging into RelativeSource Binding for over more than a week including the [twitter - contentpage based popups](https://twitter.com/BritchDavid/status/1188752399193825280), and I assure you this is a solid bug or you can call it a behavior / design which is failing to recognize those ViewModel and ContentPage types having default constructors with one or more parameters.

**I am not very clear on the suggested workaround "Set the AncestorType to your MainPage and add BindingContext. at the beginning of the path."**  I don't want to change the BindingContext of my app's ContentPage(s) that are declared in the App.cs (in the method protected override void RegisterTypes(IContainerRegistry containerRegistry)

Happy to have the capability of RelativeSource binding - but in greater reality - my core need is to show a popup based on a click / tap of ViewElement (RadioButton in this case). Such a long route to get that working ... still struggling ..

I am happy to demo my App, code to show that bug with Microsoft team as I have NDA with Microsoft.

Thank you,

legistek:
>  I have confirmed the bug on my end against the test case(s) that I mentioned. 

That's great, but doesn't help us. How have you confirmed it exactly? 

> I am not changing / resetting / assigning BindingContext anywhere besides setting them in App.cs as mentioned in the para below.

We need to know *when* in the app lifecycle the `BindingContext` is being assigned to `MainPage`. If this occurs after the visual tree is finished building, then based on the current code I don't think `FindAncestorBindingContext` will work properly.

Again, there is no reason the number of arguments in the view model constructor directly matters to relative source binding. However, having a parameterless constructor _is_ required to instantiate a view model through XAML. If you instead have to instantiate your view model in code and assign it after the visual tree has been built, that might indeed impact relative source binding. The issue is not the parameterless constructor or lack thereof - the issue is whether the viewmodel is instantiated and assigned during the construction of the visual tree or later. That's why we need to understand what exactly you did to test your theory that the number of parameters in the constructor matters.

> I am not very clear on the suggested workaround

Change

`Command="{Binding Path=DisplayDialogCommand, Source={x:RelativeSource Mode=FindAncestorBindingContext, AncestorType={x:Type pgVM:MainPageViewModel}}}"`

to

`Command="{Binding Path=BindingContext.DisplayDialogCommand, Source={x:RelativeSource Mode=FindAncestor, AncestorType={x:Type **prefix**:MainPage}}}"`

> I am happy to demo my App, code to show that bug with Microsoft team as I have NDA with Microsoft.

Well I'm willing and trying to help you now and have no desire or need to see your application or its code, but it's up to you.

IoTFier:
> > I have confirmed the bug on my end against the test case(s) that I mentioned.
> 
> That's great, but doesn't help us. How have you confirmed it exactly? 
As offered, I will get you the repro (preferably using Prism v7.2). It will be great if you are open for a screenshare session for me to show you the behavior in the repro solution ? If yes, let me know your email address and I will send you those details after we find a workable time slot. You need
> 
> > I am not changing / resetting / assigning BindingContext anywhere besides setting them in App.cs as mentioned in the para below.
> 
> We need to know _when_ in the app lifecycle the `BindingContext` is being assigned to `MainPage`. If this occurs after the visual tree is finished building, then based on the current code I don't think `FindAncestorBindingContext` will work properly.
> 
> Again, there is no reason the number of arguments in the view model constructor directly matters to relative source binding. However, having a parameterless constructor _is_ required to instantiate a view model through XAML. If you instead have to instantiate your view model in code and assign it after the visual tree has been built, that might indeed impact relative source binding. The issue is not the parameterless constructor or lack thereof - the issue is whether the viewmodel is instantiated and assigned during the construction of the visual tree or later. That's why we need to understand what exactly you did to test your theory that the number of parameters in the constructor matters.
> 
> > I am not very clear on the suggested workaround
> 
> Change
> 
> `Command="{Binding Path=DisplayDialogCommand, Source={x:RelativeSource Mode=FindAncestorBindingContext, AncestorType={x:Type pgVM:MainPageViewModel}}}"`
> 
> to
> 
> `Command="{Binding Path=BindingContext.DisplayDialogCommand, Source={x:RelativeSource Mode=FindAncestor, AncestorType={x:Type **prefix**:MainPage}}}"`
I had tried this option, and the command doesn't get wired up at runtime.  Per my initial note in this thread - the ContentView having this ViewButton wired to the VM DelegateCommand is 3-4 levels below the ContentPage. It is used dyanmically to form a contentview that provides ItemTemplate to the Syncfusion ListView ( Syncfusion listview has nothing to do with this issue).

> 
> > I am happy to demo my App, code to show that bug with Microsoft team as I have NDA with Microsoft.
> 
> Well I'm willing and trying to help you now and have no desire or need to see your application or its code, but it's up to you.

legistek:
> As offered, I will get you the repro (preferably using Prism v7.2). It will be great if you are open for a screenshare session for me to show you the behavior in the repro solution ? If yes, let me know your email address and I will send you those details after we find a workable time slot. 

It really needs to be a repro that doesn't rely on third party components because I'll need to run it myself against the debugger with the Xamarin Forms source. If you could create that and post it here I'll take a look and then happy to do a screen share if I need clarification on anything.

StephaneDelcroix:
@IoTFier please go ahead and file a new isolated issue, that does not depends on Prism, SyncFusion or anything, so we can get a better understanding at what you're trying to do, figure out if it's a bug or not, and provide a solution.

Conversations on closed threads like this aren't tracked.

Thanks

PR #4377, opened 2018-11-12
[iOS] Fixed keyboard type for cells in TableView

### LINKED ISSUE #1386: EntryCell within TableView using wrong keyboard

### Description
I have 3 EntryCell's on a screen. The 2nd and 3rd cells are visibly toggled by pressing the first (remove and add section).

If EntryCell 2 has a Numeric keyboard, then EntryCell 3 will also have a numeric keyboard, even though I set it to Plain.

### Steps to Reproduce

See sample code in forum:
https://forums.xamarin.com/discussion/112000/tableview-textcell-keyboard-wrongly-set-as-numeric

### Expected Behavior
Entry Cell keyboard should reflect the desired keyboard (Numeric, Plain etc..)

### Actual Behavior
Entry Cell keyboard seems to fix to Numeric when the sections are added/removed from a table view.

### Basic Information

This affects iOS only.

### Reproduction Link

See sample code: https://forums.xamarin.com/discussion/112000/tableview-textcell-keyboard-wrongly-set-as-numeric

AndreiMisiukevich:
### Description of Change ###

Reset KeyboardType for table view cell each time it is updated.

### Issues Resolved ### 

- fixes #1386

### API Changes ###
None

### Platforms Affected ### 
- iOS

Entry cell has correct keyboard type

Not applicable

### Testing Procedure ###
Follow steps described in test case 1386

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Extensions/Extensions.cs
+++ b/Xamarin.Forms.Platform.iOS/Extensions/Extensions.cs
@@ -10,13 +10,13 @@ namespace Xamarin.Forms.Platform.iOS
 			textInput.AutocapitalizationType = UITextAutocapitalizationType.None;
 			textInput.AutocorrectionType = UITextAutocorrectionType.No;
 			textInput.SpellCheckingType = UITextSpellCheckingType.No;
+			textInput.KeyboardType = UIKeyboardType.Default;
 
 			if (keyboard == Keyboard.Default)
 			{
 				textInput.AutocapitalizationType = UITextAutocapitalizationType.Sentences;
 				textInput.AutocorrectionType = UITextAutocorrectionType.Default;
 				textInput.SpellCheckingType = UITextSpellCheckingType.Default;
-				textInput.KeyboardType = UIKeyboardType.Default;
 			}
 			else if (keyboard == Keyboard.Chat)
 			{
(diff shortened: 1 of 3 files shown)


malsamaxyz:
Is this bug related to this problem?:
https://stackoverflow.com/questions/53298649/xamarin-forms-keyboard-closing-on-entry-after-every-key-stroke

AndreiMisiukevich:
> Is this bug related to this problem?:
> https://stackoverflow.com/questions/53298649/xamarin-forms-keyboard-closing-on-entry-after-every-key-stroke

Don't think so

Xamarin.Forms.Platform.iOS/Extensions/Extensions.cs
```diff
@@ -10,13 +10,13 @@ public static void ApplyKeyboard(this IUITextInput textInput, Keyboard keyboard)
 			textInput.AutocapitalizationType = UITextAutocapitalizationType.None;
 			textInput.AutocorrectionType = UITextAutocorrectionType.No;
 			textInput.SpellCheckingType = UITextSpellCheckingType.No;
+			textInput.KeyboardType = UIKeyboardType.Default;
```
kingces95: This affect `Chat` and `Text` as well as `CustomKeyboard`. I'd assume they all expect to modify a `UIKeyboardType.Default` keyboard. So this fixes the chat and text cases as well. Happy day!

AndreiMisiukevich: Yeah, that's right

PR #4384, opened 2018-11-12
[Android] Fix brief display of incorrect keyboard

### LINKED ISSUE #3647: Two keyboards appear when editing the TextColumn after the NumericColumn with Samsung  Android version 6 devices only on Input type of Samsung Keyboard

hartez:
### Description of Change ###

Proposing this as an alternative to #4248 and a solution to #3647.

Rather than making `HandleKeyboardOnFocus` available to custom renderers, this change makes the `OnFocusChangeRequested` handler available and implements overrides in the appropriate classes for showing/hiding the keyboard (rather than handling it in `ViewRenderer`). 

This change also posts requests to show the keyboard to the main looper queue, thus matching them up with the posted requests for control focus; this fixes the "wrong keyboard briefly displayed" bug. 

Previously, the focus requests were delayed but the keyboard display happened within `OnFocusChangeRequested`; this meant that `ShowKeyboard` was called when the target control was not actually focused, and `ShowSoftInput` always failed, requiring a call to `ToggleSoftInput` to force the keyboard open. The `ToggleSoftInput` call caused the last or default keyboard to be displayed briefly until the main looper processed the focus request and displayed the correct keyboard for the input.

By posting the "show keyboard" requests to the queue after the "focus" requests, they happen in the correct order and `ShowKeyboard` is called with the correct control having focus; `ShowSoftInput` then runs successfully.

### Issues Resolved ### 

- fixes #3647 

### API Changes ###

Android, ViewRenderer.cs:

`internal virtual void OnFocusChangeRequested(object sender, VisualElement.FocusRequestArgs e)` ->
`protected virtual void OnFocusChangeRequested(object sender, VisualElement.FocusRequestArgs e)`

### Platforms Affected ### 

- Android

None

Not applicable

### Testing Procedure ###

Programmatically switch focus between an Entry/Editor/SearchBar with a Numeric keyboard and an Entry/Editor/SearchBar with an Email keyboard several dozen times; verify that the incorrect keyboard does not flash briefly onscreen.

### PR Checklist ###

- [x] Has automated tests 


Xamarin.Forms.Platform.Android/Renderers/EditorRenderer.cs
```diff
@@ -56,12 +56,28 @@ protected override FormsEditText CreateNativeControl()
...
+			}
+
+			base.OnFocusChangeRequested(sender, e);
+
+			if (e.Focus)
+			{
+				// Post this to the main looper queue so it doesn't happen until the other focus stuff has resolved
+				// Otherwise, ShowKeyboard will be called before this control is truly focused, and we will potentially
+				// be displaying the wrong keyboard
+				Device.BeginInvokeOnMainThread(() => Control.ShowKeyboard());
```
adrianknight89: The code in `OnFocusChangeRequested` seems to be repeated across renderers. Should it be refactored into a separate file/method?

hartez: We'd have to either create two shared methods for the code on either side of `base.OnFocusChangeRequested(sender, e);` or pass the call to `base` in as a delegate. I don't love the first option because it creates two very small methods that don't do much and doesn't reduce the actual code duplication by much. And the second option introduces a new and possibly confusing pattern here. 

If we were talking about more code (say, 50 lines instead of 11 (and mostly braces/comments)), I'd say yes. But I'm not sure the maintenance gain is worth the complexity penalty here.

PureWeen:
@hartez  can you rebase this? I was going to just do it but it looks like there were some additional changes inside ViewRenderer related to popups and some disposed checks so it wasn't quite as straight forward as I'd of hoped

hartez:
@PureWeen I added your test as a manual test (Issue4384). And fixed KeyboardManager.cs to handle SearchBar appropriately.

hartez:
Android UI test failures are on 39821 (not caused by this change, fixed by https://github.com/xamarin/Xamarin.Forms/pull/4866) and 41424 on API 19 (not caused by this change, fixed by https://github.com/xamarin/Xamarin.Forms/commit/e9274c05ec5178d2234ae3d354c0bf603dad4b6b). So they can be ignored.

berlamont:
What version is this going to be in, I don't see it in any of the nightly builds so far?

PR #4402, opened 2018-11-14
Change known colors to public properties

kzu:
The XAML completion only picks up public properties, not fields, 
so change the known color fields to properties.

Before:

image

After:

image


StephaneDelcroix:
Thanks for this @kzu. Unfortunately this is a major breaking change that we can't take.

kzu:
Would you care to elaborate? I'm sure I'm missing something, because unless someone is doing 
useless reflection (hardcoding to filter out for *fields only*, in addition) to get at these values, 
this should totally work without changing customer's code.

@mgoertz-msft is there an easy fix to light up the XAML completion also for static fields?

StephaneDelcroix:
customer code will compile just fine. Already compiled code against previous version won't work without recompiling against this

kzu:
So, given that customers explicitly opt-in to newer versions by bumping a nuget package, wouldn't that be an acceptable change for (say) a major version? i.e. do it in v4?

I find it hard to believe that we're not going to even evaluate improvements even if they are source compatible, like, ever.

PR #4414, opened 2018-11-15
[iOS, Android] Implement Snap alignment for CollectionView

hartez:
### Description of Change ###

Add handling of `SnapPointsAlignment` (`Start`, `Center`, and `End`) and `SnapPointsType` (`None`, `Mandatory`, and `MandatorySingle`) to the CollectionView on the iOS platform.

This also implements the missing `End`/`MandatorySingle` combination on Android.

### API Changes ###

 None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###

Run Control Gallery, navigate to CollectionView Gallery -> Snap Points Galleries and run the examples there.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewLayout.cs
+++ b/Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewLayout.cs
@@ -215,5 +215,68 @@ namespace Xamarin.Forms.Platform.iOS
 		{
 			_needCellSizeUpdate = true;
 		}
+		
+		public override CGPoint TargetContentOffset(CGPoint proposedContentOffset, CGPoint scrollingVelocity)
+		{
+			var snapPointsType = _itemsLayout.SnapPointsType;
+
+			if (snapPointsType == SnapPointsType.None)
+			{
+				// Nothing to do here; fall back to the default
+				return base.TargetContentOffset(proposedContentOffset, scrollingVelocity);
+			}
+
+			// Get the viewport of the UICollectionView
+			var viewport = new CGRect(proposedContentOffset, CollectionView.Bounds.Size);
+
+			// And find all the elements currently visible in the viewport
+			var visibleElements = LayoutAttributesForElementsInRect(viewport);
+
+			if (visibleElements.Length == 0)
+			{
+				// Nothing to see here; fall back to the default
+				return base.TargetContentOffset(proposedContentOffset, scrollingVelocity);
+			}
+
+			var alignment = _itemsLayout.SnapPointsAlignment;
+
+			if (visibleElements.Length == 1)
+			{
+				// If there is only one item in the viewport and snapping is mandatory,
+				// then we need to align the viewport with it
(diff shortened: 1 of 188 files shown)


hartez:
@PureWeen 

> Also it seems like the Update Item Count doesn't work on iOS. If I change it and click update it always just sits at 50 items and the source doesn't seem to update

Can you file an issue for the item count thing? It's very possible that something's broken with ItemsSource on iOS, but that's unrelated to the snap behavior.

Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewLayout.cs
```diff
@@ -197,5 +197,68 @@ void UpdateConstraints(CGSize size)
 			ConstrainTo(size);
 			UpdateCellConstraints();
 		}
+
+		public override CGPoint TargetContentOffset(CGPoint proposedContentOffset, CGPoint scrollingVelocity)
+		{
+						var snapPointsType = _itemsLayout.SnapPointsType;
```
samhouts:
```suggestion
		var snapPointsType = _itemsLayout.SnapPointsType;
```

PureWeen:
@samhouts  I'm pretty sure @hartez is still working on the issue I brought up with this one.

Is that correct @hartez ?

hartez:
> @samhouts I'm pretty sure @hartez is still working on the issue I brought up with this one.
> 
> Is that correct @hartez ?

Should be ready for another review now. I added the missing MandatorySingle stuff for iOS, plus the overlooked End/MandatorySingle combination for Android.

PureWeen:
@hartez  I'm trying to run the gallery and I'm getting this exception when clicking on the List Samples


image

hartez:
@PureWeen This is ready for another review.

PureWeen:
@hartez  the action on the snaps is looking good. Two things I'm noticing

1) On Android when you set it to Mandatory Snap Type the Grid will move to match the SnapPointsAlignment each time you change it.  But that's the only scenario it does that.  If you set it to Mandatory Single or do the same thing on iOS it doesn't update to match SnapPointsAlignment 

collectionview


2) The Horizontal Grid test layout looks different on iOS vs Android. Is this a bug? or just something that has to do with the device or particular images used?

image

hartez:
> @hartez the action on the snaps is looking good. Two things I'm noticing
 
> On Android when you set it to Mandatory Snap Type the Grid will move to match the SnapPointsAlignment each time you change it.  But that's the only scenario it does that.  If you set it to Mandatory Single or do the same thing on iOS it doesn't update to match SnapPointsAlignment

> The Horizontal Grid test layout looks different on iOS vs Android. Is this a bug? or just something that has to do with the device or particular images used?
 
@pureween Any objections to merging this as-is and filing issues for those two bugs? It'd be nice to stop rebasing this PR and to work on those issues one at a time, independently.

PR #4427, opened 2018-11-15
Show the desired API we'd like to use

alanmcgovern:
What would need to be done to make the tests pass without us having to parse the xaml ourselves? On the face of it it looks like the existing parser should be able to fill in this information as already parses the full xaml and has all of this information.


StephaneDelcroix:
thanks for this. I get your intend. but a PR is not the right place to discuss this (and we discussed this last week)

PR #4431, opened 2018-11-15
[Enhancement][Core] ListView should report index of item on all events

### LINKED ISSUE #2811: ListView should report index of item on all events
## Summary
On `ListView`, add a `SelectedIndex` property. Also, add `Index` (of item) property to `ItemTappedEventArgs`, `SelectedItemChangedEventArgs` and `ItemVisibilityEventArgs`, used by `ItemTapped`,  `ItemSelected`, `ItemAppearing` and `ItemDisappearing` respectively.

## Intended Use Case
Today we need to call `IndexOf(item)` to get the index in the list, which is inefficient for most collection types. This can be avoided since the listview must track this information already.

jfversluis:
### Description of Change ###

All `EventArgs` classes that are invoked by the `ListView` now report the corresponding index of the item.

### Issues Resolved ### 

- fixes #2811

### API Changes ###

Added:
 - int ItemTappedEventArgs.ItemIndex { get; private set; }
 - int SelectedItemChangedEventArgs.SelectedItemIndex { get; private set; }
 - int ItemVisibilityEventArgs.ItemIndex { get; private set; }

Changed:
 - ctor: ItemTappedEventArgs(object group, object item) => ItemTappedEventArgs(object group, object item, int itemIndex)
 - ctor: ItemVisibilityEventArgs(object item) => ItemVisibilityEventArgs(object item, int itemIndex)
 - ctor: SelectedItemChangedEventArgs(object selectedItem) => SelectedItemChangedEventArgs(object selectedItem, int selectedItemIndex)
 
### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 


Xamarin.Forms.Core/ItemTappedEventArgs.cs
```diff
@@ -4,14 +4,17 @@ namespace Xamarin.Forms
 {
 	public class ItemTappedEventArgs : EventArgs
 	{
-		public ItemTappedEventArgs(object group, object item)
+		public ItemTappedEventArgs(object group, object item, int itemIndex)
```
StephaneDelcroix: it's unlikely that external code has a reference to this constructor, but in case it does, this is a breaking change. you need to keep the existing ctor (attribute it with `[Obsolete]`) that calls the new one with `-1` as `itemIndex` argument.

Xamarin.Forms.Core/ItemVisibilityEventArgs.cs
```diff
@@ -4,11 +4,14 @@ namespace Xamarin.Forms
 {
 	public sealed class ItemVisibilityEventArgs : EventArgs
 	{
-		public ItemVisibilityEventArgs(object item)
+		public ItemVisibilityEventArgs(object item, int itemIndex)
```
StephaneDelcroix: same

Xamarin.Forms.Core/SelectedItemChangedEventArgs.cs
```diff
@@ -4,11 +4,15 @@ namespace Xamarin.Forms
 {
 	public class SelectedItemChangedEventArgs : EventArgs
 	{
-		public SelectedItemChangedEventArgs(object selectedItem)
+		public SelectedItemChangedEventArgs(object selectedItem, int selectedItemIndex)
```
StephaneDelcroix: same

Xamarin.Forms.Core/ItemTappedEventArgs.cs
jfversluis: Good call, will fix it now

jfversluis:
Fixed!

Xamarin.Forms.Core/ListView.cs
```diff
@@ -271,11 +271,11 @@ public Element HeaderElement
 
 		[EditorBrowsable(EditorBrowsableState.Never)]
 		public void SendCellAppearing(Cell cell)
-			=> ItemAppearing?.Invoke(this, new ItemVisibilityEventArgs(cell.BindingContext));
+			=> ItemAppearing?.Invoke(this, new ItemVisibilityEventArgs(cell.BindingContext, TemplatedItems.GetGlobalIndexOfItem(cell?.BindingContext)));
```
StephaneDelcroix: what's the cost of GetIndex ?

rmarinho:
Build fails :

```

Xamarin.Forms.Platform.UAP\ListViewRenderer.cs(77,34): Error CS0618: 'SelectedItemChangedEventArgs.SelectedItemChangedEventArgs(object)' is obsolete: 'Please use the constructor that reports the items index'

```

jfversluis:
Build should be fixed now

Xamarin.Forms.Core/ListView.cs
StephaneDelcroix: please figure that out. if the complexity is too high, we'll have to close this PR

StephaneDelcroix:
build

Xamarin.Forms.Core/ListView.cs
jfversluis: When not grouped it should be O(N), it goes to a proxy object and just gets the index.

    public int GetGlobalIndexOfItem(object item)
    {
	if (!IsGroupingEnabled)
		return ListProxy.IndexOf(item);
        ....
    }

When grouped, it loops once for sure, but shouldn't be different from the `IndexOf` and then loops again recursively. Making it O(N^2).

    if (_groupedItems != null)
    {
        foreach (TemplatedItemsList<TView, TItem> children in _groupedItems.Values)
        {
	        count++;

	        int index = children.GetGlobalIndexOfItem(item);
	        if (index != -1)
	            return count + index;

	        count += children.GetDescendantCount();
        }
    }

I also went ahead and added a stopwatch and tested with appearing, dissapearing and selecting which are all sub 1ms or even sub .1ms. Full version: https://www.dropbox.com/s/hfnvtdvn8mt96br/StopwatchDemo.mov?dl=0

Stopwatch test

But maybe I'm overlooking something?

StephaneDelcroix: is it worth trying to expose this as `Lazy`? @hartez, as you've spent quite a lot of time with LV lately, could you weight in one way or the other ?

hartez: > is it worth trying to expose this as `Lazy`? @hartez, as you've spent quite a lot of time with LV lately, could you weight in one way or the other ?

@StephaneDelcroix 

Which property are you proposing to make `Lazy`? Or are you asking if we should cache the index in the `Cell`?

hartez: @jfversluis Have you checked the performance characteristics on Android? That platform would be much more of a concern.

jfversluis: Valid point. Tried the same on Android: https://www.dropbox.com/s/vdafery4lulgeat/StopwatchDemoAndroid.mov?dl=0

Times are a bit higher but overall no more than 3ms with some exceptions going to 26ms (end of video). It's a bit hard to see because the bindings seem to be broken on Android for some reason (as far as I can tell not related to this change). Also, I didn't do anything to enable fast renderers. Not sure if they are enabled for the gallery app or if it should have any impact at all.

hartez: You can enable the fast renderers in the Control Gallery by building it with the TEST_EXPERIMENTAL_RENDERERS symbol defined. But I don't think they should have any impact on the time spent finding indexes.

jfversluis: Tried, doesn't have impact indeed. As far as I'm concerned it's totally up to you guys. I can see how the performance is a crucial thing for the ListView and I don't think this "feature" doesn't outweigh any performance issues that this might cause. At least now we know :)

Xamarin.Forms.Core/SelectedItemChangedEventArgs.cs
```diff
@@ -4,11 +4,22 @@ namespace Xamarin.Forms
 {
 	public class SelectedItemChangedEventArgs : EventArgs
 	{
+		[Obsolete("Please use the constructor that reports the items index")]
```
hartez: We need to make these obsolete messages consistent with the rest of Forms. Something like
`[Obsolete("This constructor is obsolete as of version 3.5. Please use SelectedItemChangedEventArgs(object selectedItem, int selectedItemIndex) instead.")]`

samhouts:
build --uitests

jfversluis:
I think it's safe to say these UI tests are stuck :P

hartez:
None of these test failures have anything to do with this PR.

PR #4448, opened 2018-11-16
[Core]Call invalidate measure after the FormattedText is set

### LINKED ISSUE #4026: Bindable Span height Issue
### Description
When trying update span text using binding the text will cut off , if the new label text needs more rows to display
### Steps to Reproduce

1. Open the sample project
2. Click on the button (to make the bindable span text longer)
3. Note that the text will cut off

### Expected Behavior
The label height should be update and the text shouldn't cut off
### Actual Behavior
The bindable span text cut off
### Basic Information

- IDE: Microsoft visual studio 2017 15.8.5
- Platform Target Frameworks: 
  - iOS:  12 simulator
  - Android: 8 device and simulator
### Screenshots

Android
android issue
IOS
bindable-ios-span-issue

### Reproduction Link

Sample Code
[Bindable-Span-Issue.zip](https://github.com/xamarin/Xamarin.Forms/files/2453962/Bindable-Span-Issue.zip)

rmarinho:
### Description of Change ###

When the FormattedText was changed we invalidate measure before the Text was update on the control. 

[iOS]
Hook the PropertyChanging event to clear the perfectSize cache

### Issues Resolved ### 

- fixes #4026 

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

The invalidate measure is called after PropertyChanged.

Not applicable

### Testing Procedure ###

Go to issue 4026 on the gallery, add tab "Add More Text to Span", make sure all text is shown. Click it more times to see new lines are added and it always shows all new texto added.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Label.cs
+++ b/Xamarin.Forms.Core/Label.cs
@@ -48,6 +48,7 @@ namespace Xamarin.Forms
 
 					formattedString.SpansCollectionChanged -= label.Span_CollectionChanged;
 					formattedString.PropertyChanged -= label.OnFormattedTextChanged;
+					formattedString.PropertyChanging -= label.OnFormattedTextChanging;
 					formattedString.Parent = null;
 					label.RemoveSpans(formattedString.Spans);
 				}
@@ -59,6 +60,7 @@ namespace Xamarin.Forms
 				{
 					var formattedString = (FormattedString)newvalue;
 					formattedString.Parent = label;
+					formattedString.PropertyChanging += label.OnFormattedTextChanging;
 					formattedString.PropertyChanged += label.OnFormattedTextChanged;
 					formattedString.SpansCollectionChanged += label.Span_CollectionChanged;
 					label.SetupSpans(formattedString.Spans);
@@ -208,6 +210,10 @@ namespace Xamarin.Forms
 		void ILineHeightElement.OnLineHeightChanged(double oldValue, double newValue) =>
 			InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
 
+		void OnFormattedTextChanging(object sender, PropertyChangingEventArgs e)
+		{
+			OnPropertyChanging("FormattedText");
+		}
 		void OnFormattedTextChanged(object sender, PropertyChangedEventArgs e)
 		{
 			InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
--- a/Xamarin.Forms.Core/Label.cs
+++ b/Xamarin.Forms.Core/Label.cs
@@ -216,8 +216,8 @@ namespace Xamarin.Forms
 		}
 		void OnFormattedTextChanged(object sender, PropertyChangedEventArgs e)
 		{
-			InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
 			OnPropertyChanged("FormattedText");
+			InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
 		}
 
 		void SetupSpans(System.Collections.IEnumerable spans)
--- a/Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
@@ -164,7 +164,8 @@ namespace Xamarin.Forms.Platform.MacOS
 				|| e.PropertyName == Label.FormattedTextProperty.PropertyName
 				|| e.PropertyName == Label.LineBreakModeProperty.PropertyName
 				|| e.PropertyName == Label.LineHeightProperty.PropertyName
-				|| e.PropertyName == Label.TextColorProperty.PropertyName)
+				|| e.PropertyName == Label.TextColorProperty.PropertyName
+				|| e.PropertyName == Label.FontProperty.PropertyName)
 			{
 				_perfectSizeValid = false;
 			}
--- a/Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
@@ -3,6 +3,7 @@ using System.ComponentModel;
 using RectangleF = CoreGraphics.CGRect;
 using SizeF = CoreGraphics.CGSize;
 using Foundation;
+using System.Collections.Generic;
 
 #if __MOBILE__
 using UIKit;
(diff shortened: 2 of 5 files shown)


Xamarin.Forms.Core/Label.cs
```diff
@@ -208,10 +210,14 @@ public int MaxLines
...
 
+		void OnFormattedTextChanging(object sender, PropertyChangingEventArgs e)
+		{
+			OnPropertyChanging("FormattedText");
+		}
 		void OnFormattedTextChanged(object sender, PropertyChangedEventArgs e)
 		{
-			InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
 			OnPropertyChanged("FormattedText");
+			InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
```
rmarinho: This is the most important change.

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
```diff
@@ -143,6 +158,18 @@ protected override void OnElementChanged(ElementChangedEventArgs<Label> e)
 			base.OnElementChanged(e);
 		}
 
+		void ElementPropertyChanging(object sender, PropertyChangingEventArgs e)
+		{
+			if (e.PropertyName == Label.TextProperty.PropertyName
+				|| e.PropertyName == Label.FormattedTextProperty.PropertyName
+				|| e.PropertyName == Label.LineBreakModeProperty.PropertyName
+				|| e.PropertyName == Label.LineHeightProperty.PropertyName
+				|| e.PropertyName == Label.TextColorProperty.PropertyName)
```
rmarinho:
```suggestion
				|| e.PropertyName == Label.TextColorProperty.PropertyName
				|| e.PropertyName == Label.FontProperty.PropertyName)
```

samhouts:
build --uitests

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
```diff
@@ -143,6 +158,19 @@ protected override void OnElementChanged(ElementChangedEventArgs<Label> e)
...
 		}
 
+		void ElementPropertyChanging(object sender, PropertyChangingEventArgs e)
+		{
+			if (e.PropertyName == Label.TextProperty.PropertyName
+				|| e.PropertyName == Label.FormattedTextProperty.PropertyName
+				|| e.PropertyName == Label.LineBreakModeProperty.PropertyName
+				|| e.PropertyName == Label.LineHeightProperty.PropertyName
+				|| e.PropertyName == Label.TextColorProperty.PropertyName
+				|| e.PropertyName == Label.FontProperty.PropertyName)
```
paymicro: Maybe it would be better to prepare HashSet of `PropertyName`?

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
```diff
@@ -117,12 +117,27 @@ public override void Layout()
 
 		}
 
+		protected override void Dispose(bool disposing)
+		{
+			base.Dispose(disposing);
+			if (disposing)
+			{
+				Element.PropertyChanging -= ElementPropertyChanging;
```
samhouts: null check

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
```diff
@@ -117,12 +117,27 @@ public override void Layout()
...
+			}
+		}
+
 		protected override void OnElementChanged(ElementChangedEventArgs<Label> e)
 		{
+			if (e.OldElement != null)
+			{
+				Element.PropertyChanging -= ElementPropertyChanging;
```
samhouts: null check

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
```diff
@@ -117,12 +117,27 @@ public override void Layout()
...
+			}
+
 			if (e.NewElement != null)
 			{
 				if (Control == null)
 				{
+					Element.PropertyChanging += ElementPropertyChanging;
```
samhouts: null check

rmarinho:
great catch @samhouts

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
```diff
@@ -174,6 +203,12 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
 				UpdateMaxLines();
 		}
 
+		void ElementPropertyChanging(object sender, PropertyChangingEventArgs e)
+		{
+			if (s_perfectSizeSet.Contains(e.PropertyName))
```
PureWeen: Could we get rid of all the *Changing* code in this pr and just move this statement up to

https://github.com/xamarin/Xamarin.Forms/pull/4448/files#diff-381c6c37e826f38028e8841495e4f78eR176

I tested it a little bit and it seems to work. Is there a scenario where this code needs to be part of *Changing* opposed to at the top of *Changed*  ?

rmarinho: That will be a little different from what it was before. I could break back the Changing. but in my head makes sense to invalidate before so we make sure  when any re layout trigged on PropertyChanged the flag is false

samhouts: > That will be a little different from what it was before.

In what way?

rmarinho: https://github.com/xamarin/Xamarin.Forms/pull/4448/files#diff-381c6c37e826f38028e8841495e4f78eL256

is inside some methods it's not invalidated always.

samhouts: Can you explain why setting it on `Changing` is different from/better than setting it on `Changed`? We can still use the HashSet to limit which properties invalidates, right?

rmarinho: Right we can still use. But imagine other parts of the Native Layout on the platform are also hooked to Changed events, and they will trigger the native `GetDesiredSize` to fire, this will try prevent that case.

Issue #4435, opened 2018-11-15
Spec: Visual


davidortinau:
## Rationale

Customers desire to more easily create Xamarin.Forms applications that look mostly or exactly the same by default on iOS and Android. In examining the style of design that is most typically implemented, Material design is the most beneficial starting point. 

Developers should be able to express their design for this starting point at the top level of their Xamarin.Forms application, and have that preference reflected throughout.

Developers should be able to also control this preference at the ContentPage and control levell.

In the future, additional Visual styles may be supported, such as Cupertino, Fluent, or any other design styling that someone would wish to implement themselves.

## Goals

- Visually identical rendering
  - Not to be confused with “pixel perfect”. Visually identical means the naked eye can’t tell the difference at a glance.
- High performance
  - Switching
  - Rendering
  - Inflation
- [Linker friendly](https://github.com/xamarin/Xamarin.Forms/issues/5020)
- Dynamic switching between visuals
- Target controls
  - [ActivityIndicator](https://github.com/xamarin/Xamarin.Forms/issues/5007)
  - [Button](
https://github.com/xamarin/Xamarin.Forms/issues/4951)
  - [DatePicker](https://github.com/xamarin/Xamarin.Forms/issues/5010)
  - [Entry](
https://github.com/xamarin/Xamarin.Forms/issues/4949)
  - [Editor](https://github.com/xamarin/Xamarin.Forms/issues/5009)
  - ListView
    - Done with just some default property setting really
  - [Picker](https://github.com/xamarin/Xamarin.Forms/issues/5010)
  - [ProgressBar](https://github.com/xamarin/Xamarin.Forms/issues/5008)
  - [Slider](https://github.com/xamarin/Xamarin.Forms/issues/5018)
  - [Stepper](https://github.com/xamarin/Xamarin.Forms/issues/5011)
  - [Switch](https://github.com/xamarin/Xamarin.Forms/issues/5017)
  - [TimePicker](https://github.com/xamarin/Xamarin.Forms/issues/5010)
## Stretch Goals
- User control of rendering
  - In renderer
  - In cross-platform code
- [Support of Material-only properties](https://github.com/xamarin/Xamarin.Forms/issues/5005)
  - Elevation in particular would be neat
- Implement a material switch into iOS
- SearchBar

## Limitations

[...]

PR #4449, opened 2018-11-16
[Enhancement] Add Visual progress behind feature flag 

PureWeen:
### Description of Change ###
Add a Visual property to VisualElement that allows us to swap out different renderers based on a Visual Type property

https://github.com/xamarin/Xamarin.Forms/issues/4435

This PR also adds a starting set of Material Renderers as the first use case for this behavior
- Button
- Entry
- Frame (CardView)
- ProgressBar (iOS)

### API Changes ###

Added:

```C#
public class VisualElement
{
		public IVisual Visual
		{
			get { return (IVisual)GetValue(VisualProperty); }
			set { SetValue(VisualProperty, value); }
		}
}

```
```C#
public sealed class ExportRendererAttribute : HandlerAttribute
{
		public ExportRendererAttribute(Type handler, Type target) : this(handler, target, null)
		{
		}
}
// Example Usage
[assembly: ExportRenderer(typeof(Xamarin.Forms.Button), typeof(Xamarin.Forms.Platform.iOS.Material.MaterialButtonRenderer), new[] { typeof(VisualRendererMarker.Material) })]
namespace Xamarin.Forms.Platform.iOS.Material
{
	public class MaterialButtonRenderer : ViewRenderer<Button, MButton>

```

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

There should be no visual changes to standard Default Renderers 
Here is the view when using Material Visuals

image

image

### Testing Procedure ###
- Make sure Visuals are hidden behind the experimental flag correctly
- Make sure all current renderer behavior has been maintained (lots of changes to registrar) so that this change isn't breaking

### PR Checklist ###

- [X] Has automated tests 

### DIFF
--- a/Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
@@ -42,13 +42,12 @@ namespace Xamarin.Forms.ControlGallery.Android
 
 #if TEST_EXPERIMENTAL_RENDERERS
 			// CollectionView lets us test CollectionView stuff until it's officially released
-			Forms.SetFlags("FastRenderers_Experimental");
+			Forms.SetFlags("FastRenderers_Experimental"/*, "CollectionView_Experimental", "Visual_Experimental"*/); 
 #else
 			// Fake_Flag is here so we can test for flag initialization issues
 			// CollectionView lets us test CollectionView stuff until it's officially released
-			Forms.SetFlags("Fake_Flag"); 
+			Forms.SetFlags("Fake_Flag"/*, "CollectionView_Experimental", "Visual_Experimental"*/); 
 #endif
-
 			Forms.Init(this, bundle);
 
 			FormsMaps.Init(this, bundle);
--- a/Xamarin.Forms.ControlGallery.iOS/AppDelegate.cs
+++ b/Xamarin.Forms.ControlGallery.iOS/AppDelegate.cs
@@ -159,7 +159,7 @@ namespace Xamarin.Forms.ControlGallery.iOS
 			App.IOSVersion = int.Parse(versionPart[0]);
 
 			Xamarin.Calabash.Start();
-			
+			// Forms.SetFlags("CollectionView_Experimental", "Visual_Experimental");
 			Forms.Init();
 			FormsMaps.Init();
 			Forms.ViewInitialized += (object sender, ViewInitializedEventArgs e) =>
--- a/Xamarin.Forms.Controls/Xamarin.Forms.Controls.csproj
+++ b/Xamarin.Forms.Controls/Xamarin.Forms.Controls.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
   </PropertyGroup>
--- a/Xamarin.Forms.Core.UnitTests/VisualTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/VisualTests.cs
@@ -417,17 +417,6 @@ namespace Xamarin.Forms.Core.UnitTests
 			Assume.That(child.Visual == Forms.VisualMarker.MatchParent, "child view Visual should be MatchParent");
 		}
 
-		//static void AddImplicitToDefaultScrollView(ScrollView parent, View child)
-		//{
-		//	parent.Content = child;
-
-		//	IVisualController controller = child;
-
-		//	Assume.That(controller.EffectiveVisual.IsImplicit(), "child view EffectiveVisual should be Implicit");
-		//	Assume.That(controller.EffectiveVisual.IsDefault(), "child view EffectiveVisual should be Default");
-		//	Assume.That(child.Visual == Visual.MatchParent, "child view Visual should be MatchParent");
-		//}
-
 		static void AddImplicitToMaterial(StackLayout parent, View child)
 		{
 			parent.Children.Add(child);
@@ -448,18 +437,6 @@ namespace Xamarin.Forms.Core.UnitTests
 			Assume.That(controller.EffectiveVisual == Forms.VisualMarker.Material, "child view EffectiveVisual should be Material");
 			Assume.That(child.Visual == Forms.VisualMarker.MatchParent, "child view Visual should be MatchParent");
 		}
-
-		//static ScrollView ExplicitDefaultScrollView()
-		//{
-		//	var layout = new ScrollView { Visual = Visual.Default };
-
-		//	IVisualController controller = layout;
-
-		//	Assume.That(controller.EffectiveVisual.IsExplicit(), "Explicit Default view EffectiveVisual should be Explicit");
-		//	Assume.That(controller.EffectiveVisual.IsDefault(), "Explicit Default view EffectiveVisual should be Default");
-		//	Assume.That(layout.Visual == Visual.Default, "Explicit Default view Visual should be Default");
-		//	return layout;
-		//}
 
 		static StackLayout ExplicitDefaultLayout()
 		{
@@ -507,19 +484,6 @@ namespace Xamarin.Forms.Core.UnitTests
 
 			return layout;
 		}
-
-		//static View ExplicitMaterialView()
-		//{
-		//	var view = new View { Visual = Visual.Material };
-
-		//	IVisualController controller = view;
-
-		//	Assume.That(controller.EffectiveVisual.IsExplicit(), "Explicit RTL view EffectiveVisual should be Explicit");
-		//	Assume.That(controller.EffectiveVisual.IsMaterial(), "Explicit RTL view EffectiveVisual should be Material");
-		//	Assume.That(((View)view).Visual == Visual.Material, "Explicit RTL view Visual should be Material");
-
-		//	return view;
-		//}
 
 		static ScrollView ImplicitDefaultScrollView()
 		{
--- a/Xamarin.Forms.Core.UnitTests/VisualTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/VisualTests.cs
@@ -1,0 +1,573 @@
+using NUnit.Framework;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using Xamarin.Forms.Internals;
+
+namespace Xamarin.Forms.Core.UnitTests
+{
+	[TestFixture]
+	public class VisualTests : BaseTestFixture
+	{
+		[SetUp]
+		public override void Setup()
+		{
+			Device.SetFlags(new List<string> { ExperimentalFlags.VisualExperimental });
+			base.Setup();
+			var mockDeviceInfo = new TestDeviceInfo();
+			Device.Info = mockDeviceInfo;
+		}
+
+		[Test]
+		public void ListViewVisualIsInheritedByViewCells()
+		{
+			var lv = new ListView { Visual = Forms.VisualMarker.Material, ItemTemplate = new DataTemplate(() => new ViewCell { View = new View() }) };
+
+			lv.ItemsSource = Enumerable.Range(0, 10);
+
+			ViewCell cell = lv.TemplatedItems[0] as ViewCell;
+			IVisualController target = cell.View;
+			Assert.AreEqual(Forms.VisualMarker.Material, target.EffectiveVisual, "ViewCell View is not Material");
+		}
+
+		[Test]
+		public void ListViewVisualIsInheritedByImageInViewCells()
+		{
+			var lv = new ListView { Visual = Forms.VisualMarker.Material, ItemTemplate = new DataTemplate(() => new ViewCell { View = new Label() }) };
+
+			lv.ItemsSource = Enumerable.Range(0, 10);
+
+			ViewCell cell = lv.TemplatedItems[0] as ViewCell;
+			IVisualController target = cell.View;
+			Assert.AreEqual(Forms.VisualMarker.Material, target.EffectiveVisual, "ViewCell View is not Material");
+		}
+
+		[Test]
+		public void ScrollViewSetsVisualAndGrandchildMaintainsParentExplicitValue()
+		{
+			var layout = ImplicitDefaultScrollView();
+			var layout2 = ExplicitMaterialScrollView();
+			IVisualController view = ImplicitDefaultView();
+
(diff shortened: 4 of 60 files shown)


Xamarin.Forms.Core/Element.cs
```diff
@@ -561,7 +580,8 @@ void EffectsOnCollectionChanged(object sender, NotifyCollectionChangedEventArgs
 		internal INameScope GetNameScope()
 		{
 			var element = this;
-			do {
+			do
+			{
```
StephaneDelcroix: 🙍‍♂️

Xamarin.Forms.Core/ExperimentalFlags.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace Xamarin.Forms
+{
+	internal static class ExperimentalFlags
+	{
```
StephaneDelcroix: I'd prefer if you could keep this logic in some Visual class, and leave the CollectionView where it is too. we do not need to generify this

Xamarin.Forms.Core/IPropertyPropagationController.cs
```diff
@@ -0,0 +1,7 @@
+namespace Xamarin.Forms
+{
+	internal interface IPropertyPropagationController
```
StephaneDelcroix: `internal` is implicit

Xamarin.Forms.Core/Items/CollectionView.cs
```diff
@@ -24,28 +24,7 @@ public CollectionView()
 			string constructorHint = null,
 			[CallerMemberName] string memberName = "")
 		{
-			if (!Device.Flags.Contains(CollectionViewExperimental))
```
StephaneDelcroix: see upper comment

Xamarin.Forms.Core/Properties/AssemblyInfo.cs
```diff
@@ -92,16 +92,17 @@
...
+[assembly: StyleProperty("-xf-orientation", typeof(ScrollView), nameof(ScrollView.OrientationProperty))]
+[assembly: StyleProperty("-xf-orientation", typeof(StackLayout), nameof(StackLayout.OrientationProperty))]
+[assembly: StyleProperty("-xf-placeholder", typeof(IPlaceholderElement), nameof(PlaceholderElement.PlaceholderProperty))]
+[assembly: StyleProperty("-xf-placeholder-color", typeof(IPlaceholderElement), nameof(PlaceholderElement.PlaceholderColorProperty))]
 [assembly: StyleProperty("-xf-spacing", typeof(StackLayout), nameof(StackLayout.SpacingProperty))]
-[assembly: StyleProperty("-xf-orientation", typeof(StackLayout), nameof(StackLayout.OrientationProperty))]
\ No newline at end of file
+[assembly: StyleProperty("-xf-thumb-color", typeof(Slider), nameof(Slider.ThumbColorProperty))]
+[assembly: StyleProperty("-xf-vertical-scroll-bar-visibility", typeof(ScrollView), nameof(ScrollView.VerticalScrollBarVisibilityProperty))]
+[assembly: StyleProperty("-xf-visual", typeof(VisualElement), nameof(VisualElement.VisualProperty))]
```
StephaneDelcroix: what's happening here ?

Xamarin.Forms.Core/VisualElement.cs
```diff
@@ -56,9 +57,34 @@ public partial class VisualElement : Element, IAnimatable, IVisualElementControl
...
+
+				_effectiveVisual = value;
+				OnPropertyChanged(VisualProperty.PropertyName);
+			}
+		}
+
 		static void OnTransformChanged(BindableObject bindable, object oldValue, object newValue)
 		{
-			if ((string)newValue == "none") {
+			if ((string)newValue == "none")
```
StephaneDelcroix: please do not reformat

Xamarin.Forms.Platform.iOS/Material/MaterialButtonRenderer.cs
```diff
@@ -0,0 +1,336 @@
...
+using UIKit;
+using Xamarin.Forms.PlatformConfiguration.iOSSpecific;
+using SizeF = CoreGraphics.CGSize;
+using MButton = MaterialComponents.Button;
+using Xamarin.Forms;
+
+[assembly: ExportRenderer(typeof(Xamarin.Forms.Button), typeof(Xamarin.Forms.Platform.iOS.Material.MaterialButtonRenderer), new[] { typeof(VisualRendererMarker.Material) })]
+namespace Xamarin.Forms.Platform.iOS.Material
+{
+	public class MaterialButtonRenderer : ViewRenderer<Button, MButton>
```
StephaneDelcroix: I'm curious why, as `MButton` inherits from `UIButton`, you didn't make the MaterialButtonRenderer inherits from ButtonRenderer ?
I get it's easier to start from a new clean slate, or at least more fun, but I'm afraid that doing a whole new set of renderers will make resurface bugs we solved in the past. It's also 2 codepaths to maintain in the future, and this isn't a situation like AppCompat where there's a plan to deprecate one set...

Xamarin.Forms.Core/Properties/AssemblyInfo.cs
PureWeen: rebase snafu

Xamarin.Forms.Platform.iOS/Material/MaterialButtonRenderer.cs
PureWeen: This will be refactored to inherit. Probably today or tomorrow. It just became too difficult while working on this to differentiate API changes. I built the material control up in a predictable way and now that I have a strong idea of where the two need to have split logic I can just base class the ButtonRenderer. On the android side as well I'm planning on moving out similar logic to static classes

Xamarin.Forms.Core/ExperimentalFlags.cs
PureWeen: Why? I generalized it to ensure the exception messaging and behavior stays consistent between the features. Also once Shell is rolled in behind a flag it can just use this static method as well and thus maintain consistent flag failing behavior

Xamarin.Forms.ControlGallery.iOS/AppDelegate.cs
```diff
@@ -159,7 +159,7 @@ public override bool FinishedLaunching(UIApplication uiApplication, NSDictionary
 			App.IOSVersion = int.Parse(versionPart[0]);
 
 			Xamarin.Calabash.Start();
-			
+			// Forms.SetFlags("CollectionView_Experimental", "Visual_Experimental");
```
samhouts: Should we `#if` this instead of commenting it out?

Xamarin.Forms.Core.UnitTests/VisualTests.cs
```diff
@@ -0,0 +1,573 @@
...
+		static void AddImplicitToDefault(StackLayout parent, View child)
+		{
+			parent.Children.Add(child);
+
+			IVisualController controller = child;
+
+			Assume.That(child.Visual == Forms.VisualMarker.MatchParent, "child view Visual should be MatchParent");
+		}
+
+		//static void AddImplicitToDefaultScrollView(ScrollView parent, View child)
```
samhouts: Should we just remove this?

Xamarin.Forms.Core.UnitTests/VisualTests.cs
```diff
@@ -0,0 +1,573 @@
...
+		{
+			parent.Content = child;
+
+			IVisualController controller = child;
+
+			Assume.That(controller.EffectiveVisual == Forms.VisualMarker.Material, "child view EffectiveVisual should be Material");
+			Assume.That(child.Visual == Forms.VisualMarker.MatchParent, "child view Visual should be MatchParent");
+		}
+
+		//static ScrollView ExplicitDefaultScrollView()
```
samhouts: And this?

Xamarin.Forms.Core.UnitTests/VisualTests.cs
```diff
@@ -0,0 +1,573 @@
...
+
+			IVisualController controller = layout;
+
+			Assume.That(controller.EffectiveVisual.IsMaterial(), "Explicit RTL view EffectiveVisual should be Material");
+			Assume.That(layout.Visual == Forms.VisualMarker.Material, "Explicit RTL view Visual should be Material");
+
+			return layout;
+		}
+
+		//static View ExplicitMaterialView()
```
samhouts: And this?

Xamarin.Forms.Core/Visual.cs
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	[Xaml.TypeConversion(typeof(IVisual))]
+	public class VisualTypeConverter : TypeConverter
+	{
+		public override object ConvertFromInvariantString(string value)
+		{
+			if (value != null)
+			{
+				switch (value.Trim().ToLowerInvariant())
```
samhouts: I like the concise switch statement, but I wonder if we should use the same pattern as https://github.com/xamarin/Xamarin.Forms/pull/4501/files#diff-27bd44696d3c1c400c11e4507cababd6R22

Xamarin.Forms.ControlGallery.iOS/AppDelegate.cs
PureWeen: I mainly added this so it's just easier for anyone using ControlGallery to uncomment these and test the experimental.

Xamarin.Forms.Core/Visual.cs
PureWeen: @StephaneDelcroix  added this code

@StephaneDelcroix  thoughts?

Xamarin.Forms.Core.UnitTests/VisualTests.cs
PureWeen: @samhouts 
Yea

On a side note these remained commented out because the analogous methods inside FlowDirectionTests are never called

```
AddExplicitLTRToScrollView
 AddImplicitToLTRScrollView
ExplicitLeftToRightScrollView
```

Should those be removed from the FlowDirectionTests? Or are there some tests that still should be added that use those?

Xamarin.Forms.ControlGallery.iOS/AppDelegate.cs
StephaneDelcroix: 👍

Xamarin.Forms.Controls/Xamarin.Forms.Controls.csproj
```diff
@@ -115,6 +115,9 @@
     <EmbeddedResource Update="ControlGalleryPages\LayoutAddPerformance.xaml">
       <Generator>MSBuild:UpdateDesignTimeXaml</Generator>
     </EmbeddedResource>
+    <EmbeddedResource Update="ControlGalleryPages\VisualGallery.xaml">
+      <Generator>MSBuild:UpdateDesignTimeXaml</Generator>
+    </EmbeddedResource>
```
StephaneDelcroix: I thought that project got moved to sdk-style, with DefaultItems

StephaneDelcroix: not a blocker

Xamarin.Forms.Core/Visual.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Xamarin.Forms
+{
+	public static class VisualMarker
+	{
+		public static IVisual MatchParent { get; } = new VisualRendererMarker.MatchParent();
+		public static IVisual Default { get; } = new VisualRendererMarker.Default();
+		public static IVisual Material { get; } = new VisualRendererMarker.Material();
+	}
+
+	public static class VisualRendererMarker
```
StephaneDelcroix: I think this one should't be public. or if they are, the ctor for the nested classes should be internal to prevent instantiation

Xamarin.Forms.Core/Visual.cs
StephaneDelcroix: ToLowerInvariant is more expensive than Equals with a StringComparison
also, use `nameof()`

I agree with @samhouts , but this is not a blocker

PureWeen: changed

Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
```diff
@@ -42,13 +42,12 @@ protected override void OnCreate(Bundle bundle)
 
 #if TEST_EXPERIMENTAL_RENDERERS
 			// CollectionView lets us test CollectionView stuff until it's officially released
-			Forms.SetFlags("FastRenderers_Experimental");
+			Forms.SetFlags("FastRenderers_Experimental"/*, "CollectionView_Experimental", "Visual_Experimental"*/);
```
rmarinho: let's add the flag on always no ? So we can write UITests for it ?

Xamarin.Forms.Core/VisualElement.cs
```diff
@@ -779,7 +805,7 @@ protected override void OnParentSet()
 			}
 #pragma warning restore 0618
 
-			FlowController.NotifyFlowDirectionChanged();
+			PropertyPropagationController.PropagatePropertyChanged(null);
```
rmarinho:
```suggestion
			PropertyPropagationController.PropagatePropertyChanged(VisualElement.FlowDirectionProperty.PropertyName);
```

Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
PureWeen: The main reason we left these out for these PRs was due to wanting to verify proper behavior with the flags disabled since these are experimental.  This way we verify if someone installs a new package and doesn't set these it doesn't blow up for them. 

The next evolution we can probably just leave the flags on

Xamarin.Forms.Core/VisualElement.cs
PureWeen: The purpose of the null is saying hey some event has happened at the parent level so all children you need to re-evaluate yourselves.

Visual and FlowDirection both propagate to the children so the children use the NULL to know that they need to check both properties of the parent

https://github.com/xamarin/Xamarin.Forms/pull/4449/files#diff-b2408192f3f7437c9eed488a33f903e8R1059

https://github.com/xamarin/Xamarin.Forms/pull/4449/files#diff-b2408192f3f7437c9eed488a33f903e8R1062

PR #4495, opened 2018-11-21
[Android] add Element and Element Property Changed hooks for ImageButton

PureWeen:
### Description of Change ###
ImageButton Renderer for Android didn't provide overrides that are consistent with other renderers for hooking into element and element property change events

### API Changes ###

Added the following methods to the ImageButtonRenderer
```C#
protected virtual void OnElementChanged(ElementChangedEventArgs<ImageButton> e);
protected virtual void OnElementPropertyChanged(object sender, PropertyChangedEventArgs e);
```

### Platforms Affected ### 
- Android

### Testing Procedure ###
Add a custom ImageButtonRenderer for Android

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/ImageButtonRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/ImageButtonRenderer.cs
@@ -172,8 +172,13 @@ namespace Xamarin.Forms.Platform.Android
 			UpdateInputTransparent();
 			UpdatePadding();
 
-			ElementChanged?.Invoke(this, new VisualElementChangedEventArgs(oldElement, ImageButton));
+			OnElementChanged(new ElementChangedEventArgs<ImageButton>(oldElement, ImageButton));
 			ImageButton?.SendViewInitialized(Control);
+		}
+
+		protected virtual void OnElementChanged(ElementChangedEventArgs<ImageButton> e)
+		{
+			ElementChanged?.Invoke(this, new VisualElementChangedEventArgs(e.OldElement, e.NewElement));
 		}
 
 		public override void Draw(Canvas canvas)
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.Android/AppCompat/ImageButtonRenderer.cs
```diff
@@ -172,10 +172,15 @@ void IVisualElementRenderer.SetElement(VisualElement element)
 			UpdateInputTransparent();
 			UpdatePadding();
 
-			ElementChanged?.Invoke(this, new VisualElementChangedEventArgs(oldElement, ImageButton));
+			OnElementChanged(new ElementChangedEventArgs<ImageButton>(oldElement, ImageButton));
 			ImageButton?.SendViewInitialized(Control);
 		}
 
+		protected virtual void OnElementChanged(ElementChangedEventArgs<ImageButton> e)
```
StephaneDelcroix: it's better to keep the `virtual void OnElementChanged` empty and invokes it after firing the event

PureWeen: I mainly did it like this because this is how VisualElementRenderer does it
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/VisualElementRenderer.cs#L305

So I was just trying to mimic that one for the sake of consistency

PureWeen: For the record I do agree with @StephaneDelcroix 

But  I did it this way because this is the same way we do it literally everywhere that I could find. So I feel like at this point consistency should be the winner

https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/WkWebViewRenderer.cs#L126

samhouts: Yeah, agreed. Consistency is best. :/

StephaneDelcroix: > Yeah, agreed. Consistency is best. :/

disagreed, but I'm not going to die on this hill

PR #4527, opened 2018-11-23
Make Label display HTML from a string

### LINKED ISSUE #4504: [F100] Make Label display HTML from a string
### Rationale

Make Label able to display HTML from HTML string. 

Sure, solutions exist from worse ones (WebView) to better ones (renderers). But it maybe time after so many years to have such a simple feature built in into the framework?

### Implementation
 Every platform has a way to render HTML:

**iOS**: UILabel.AttributedText = new NSAttributedString(...);
**Android**: TextView = Html.fromHtml(html);
**UWP**: TextBlock.Text = Windows.Data.Html.HtmlUtilities.ConvertToText(html);

I don't know exactly what is best way to add this, so it's simple and without adding breaking changes.
Maybe add a new property to Label, something like `Html`:

```
public class Label : ..
{ 
     public string Html { get; set; }  // bindable property
}
```

Setting it would clear the content of the existing `FormattedText` property. Same happens the other way, setting `FormattedText` clears out the `Html` property. Normally there's no logical reason for someone to use both properties, but even if he\she does it, it will work OK by both properties being mutually exclusive.

This is backward compatible, it's simple to use and implement.

jfversluis:
### Description of Change ###

With the new `TextType` property you can now set HTML formatted content on the `Label`. In the future other types might be added that use pure text formatting like Markdown. Functionality around `FormattedText` is unchanged.

In the discussion below we have been exploring the possibility to solve this with different spans. This is possible on iOS, but from what I can tell not on Android (and thus did not explore other platforms). While talking to @PureWeen we decided this was the best alternative.

Although we do allow HTML (through the native handlers for that on iOS and Android) not all tags are supported. We are limited by the support on the different platforms. I feel that might be something that is going to come up in the future.

### Issues Resolved ### 

- fixes #4504

### API Changes ###
Added:
 - public enum TextType
 - TextType Label.TextType { get; set; } //Bindable Property

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP
- Mac OS

None by default, but the user is now able to show HTML formatted content by setting the `Label`s `TextType` property to `TextType.Html`. Default is `TextType.PlainText` which changes nothing in the default behavior as it is today

HTML Content on iOS

HTML Content on Android

HTML Content on UWP

### Testing Procedure ###

Test cases were added in the gallery app

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/LabelRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/LabelRenderer.cs
@@ -122,7 +122,6 @@ namespace Xamarin.Forms.Platform.Android
 				UpdateLineHeight();
 				UpdateGravity();
 				UpdateMaxLines();
-				UpdateTextType();
 			}
 			else
 			{
@@ -136,8 +135,6 @@ namespace Xamarin.Forms.Platform.Android
 					UpdateMaxLines();
 				if (e.OldElement.CharacterSpacing != e.NewElement.CharacterSpacing)
 					UpdateCharacterSpacing();
-				if (e.OldElement.TextType != e.NewElement.TextType)
-					UpdateTextType();
 			}
 			UpdateTextDecorations();
 			UpdatePadding();
@@ -160,7 +157,8 @@ namespace Xamarin.Forms.Platform.Android
 				UpdateLineBreakMode();
 			else if (e.PropertyName == Label.TextDecorationsProperty.PropertyName)
 				UpdateTextDecorations();
-			else if (e.PropertyName == Label.TextProperty.PropertyName || e.PropertyName == Label.FormattedTextProperty.PropertyName)
+			else if (e.PropertyName == Label.TextProperty.PropertyName || e.PropertyName == Label.FormattedTextProperty.PropertyName
+				|| e.PropertyName == Label.TextTypeProperty.PropertyName)
 				UpdateText();
 			else if (e.PropertyName == Label.LineHeightProperty.PropertyName)
 				UpdateLineHeight();
@@ -168,8 +166,6 @@ namespace Xamarin.Forms.Platform.Android
 				UpdateMaxLines();
 			else if (e.PropertyName == Label.PaddingProperty.PropertyName)
 				UpdatePadding();
-			else if (e.PropertyName == Label.TextTypeProperty.PropertyName)
-				UpdateTextType();
 		}
 
 		void UpdateColor()
@@ -246,7 +242,6 @@ namespace Xamarin.Forms.Platform.Android
 			}
 		}
 
-
 		void UpdateLineHeight()
 		{
 			_lastSizeRequest = null;
@@ -278,7 +273,25 @@ namespace Xamarin.Forms.Platform.Android
 					_view.SetTextColor(_labelTextColorDefault);
 					_lastUpdateColor = Color.Default;
 				}
-				_view.Text = Element.Text;
+
+				switch (Element.TextType)
+				{
+
+					case TextType.Html:
+						if (Forms.IsNougatOrNewer)
+							Control.SetText(Html.FromHtml(Element.Text, FromHtmlOptions.ModeCompact), TextView.BufferType.Spannable);
+						else
+#pragma warning disable CS0618 // Type or member is obsolete
+							Control.SetText(Html.FromHtml(Element.Text), TextView.BufferType.Spannable);
+#pragma warning restore CS0618 // Type or member is obsolete
+						break;
+
+					default:
+						_view.Text = Element.Text;
+
+						break;
+				}
+
 				UpdateColor();
 				UpdateFont();
 
@@ -297,31 +310,6 @@ namespace Xamarin.Forms.Platform.Android
 				(int)Context.ToPixels(Element.Padding.Bottom));
 
 			_lastSizeRequest = null;
-		}
-		
-		void UpdateTextType()
-		{
-			if (Element.Text == null)
-				return;
-
-			switch (Element.TextType)
-			{
-
-				case TextType.Html:
-					if (Forms.IsNougatOrNewer)
-						Control.SetText(Html.FromHtml(Element.Text, FromHtmlOptions.ModeCompact), TextView.BufferType.Spannable);
-					else
-#pragma warning disable CS0618 // Type or member is obsolete
-						Control.SetText(Html.FromHtml(Element.Text), TextView.BufferType.Spannable);
-#pragma warning restore CS0618 // Type or member is obsolete
-
-					_lastSizeRequest = null;
-					break;
-
-				default:
-					UpdateText();
-					break;
-			}
 		}
 
 		public override bool OnTouchEvent(MotionEvent e)
--- a/Xamarin.Forms.Platform.Android/Renderers/LabelRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/LabelRenderer.cs
@@ -168,7 +168,7 @@ namespace Xamarin.Forms.Platform.Android
 				UpdateMaxLines();
 			else if (e.PropertyName == Label.PaddingProperty.PropertyName)
 				UpdatePadding();
-			else if (e.PropertyName == Label.HtmlProperty.PropertyName)
+			else if (e.PropertyName == Label.TextTypeProperty.PropertyName)
 				UpdateTextType();
 		}
 
@@ -301,15 +301,27 @@ namespace Xamarin.Forms.Platform.Android
 		
 		void UpdateTextType()
 		{
-			if (string.IsNullOrWhiteSpace(Element.Text))
+			if (Element.Text == null)
 				return;
 
-			if (Forms.IsNougatOrNewer)
-				Control.SetText(Html.FromHtml(Element.Text, FromHtmlOptions.ModeCompact), TextView.BufferType.Spannable);
-			else
+			switch (Element.TextType)
+			{
+
+				case TextType.Html:
+					if (Forms.IsNougatOrNewer)
+						Control.SetText(Html.FromHtml(Element.Text, FromHtmlOptions.ModeCompact), TextView.BufferType.Spannable);
+					else
 #pragma warning disable CS0618 // Type or member is obsolete
-				Control.SetText(Html.FromHtml(Element.Text), TextView.BufferType.Spannable);
+						Control.SetText(Html.FromHtml(Element.Text), TextView.BufferType.Spannable);
 #pragma warning restore CS0618 // Type or member is obsolete
+
+					_lastSizeRequest = null;
+					break;
+
+				default:
+					UpdateText();
+					break;
+			}
 		}
 
 		public override bool OnTouchEvent(MotionEvent e)
--- a/Xamarin.Forms.Platform.Android/Renderers/LabelRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/LabelRenderer.cs
@@ -122,6 +122,7 @@ namespace Xamarin.Forms.Platform.Android
 				UpdateLineHeight();
 				UpdateGravity();
 				UpdateMaxLines();
+				UpdateTextType();
 			}
 			else
 			{
@@ -135,7 +136,8 @@ namespace Xamarin.Forms.Platform.Android
 					UpdateMaxLines();
 				if (e.OldElement.CharacterSpacing != e.NewElement.CharacterSpacing)
 					UpdateCharacterSpacing();
-
+				if (e.OldElement.TextType != e.NewElement.TextType)
+					UpdateTextType();
 			}
 			UpdateTextDecorations();
 			UpdatePadding();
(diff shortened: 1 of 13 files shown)


jfversluis:
Implemented UWP with the code from @matteobortolazzo. Left out the `<a>` tag since I don't think it works on other platforms as well. If anything more is needed let me know :)

jfversluis:
Regarding the separate `Html` property; I totally agree. If we could come up with a way to have it in the existing `Text` property, that would be fine by me. I mostly implemented it this way because of the spec and that it makes sense to not mess with the existing functionality too much.

You would prefer to simply put the value in `Text`, lose the `Html` and have an additional content-type related property then?

Users would even be able to have some kind of HTML/Source view when you toggle the content-type property but not the text, that might be nice.

For the switching text/HTML, I can see this happening on iOS, the height of the `Label` becomes 0 when switching back to HTML the second time. On Android, the height looks good, but the HTML still doesn't show. Can't really put my finger on why this is happening, but it seems to be something in the Core rather than platform-specific.

Anyway, might be good to agree on the actual spec before moving forward on this.

StephaneDelcroix:
> What are your thoughts on adding a ContentType or TextType property where you set it as HTML opposed to adding an HTML property?

if it supports `FormattedText` too, I'm in. otherwise, not sure if it's worth the pain

jfversluis:
Then _if_ we would do that, it would look like this.

```
public enum ContentType
{
    PlainText,
    Html,
    FormattedText
}
```

I think `(Plain)Text` and `Html` speak for themselves, but how would you see `FormattedText` working? This needs a physically different object than the former two options. So, `FormattedText` would still need a separate property to take in its value, while the other two are plain strings.

PureWeen:
I don't think we can rope *FormattedText* in for the reasons you've outlined.  If we had something like an *object Content* where it could parade as multiple things maybe. I think of it like settings Text vs Run elements on UWP at this point

The other downside with ContentType is that you have to set more than one thing to achieve HTML but it just doesn't feel right to add additionally mutually exclusive content setters.  What about when people really want a markdown view? 

I'm thinking *TextType* for the name as well since that aligns better with the Label APIs

At this point I'd still hold off on making changes just because it'll be annoying to do so before solidifying. I'll get some more buy in and then we'll get changes applied.

PureWeen:
@CZEMacLeod  it absolutely would!! Which is why I like adding a TextType enum we can grow opposed to just adding additional "content type" apis.  A mark down TextType would be pretty awesome!

mattleibow:
I saw the words markdown and label and thought of this:
https://docs.microsoft.com/en-us/windows/communitytoolkit/controls/markdowntextblock

UWP is done. 😸

PureWeen:
Yea though on some level I do wonder if we are overloading what "label" is for and if things like this would be better suited for their own controls <HtmlViewer/> <MarkdownTextBlock/> etc... 

That way the APIs of the control could reflect the intent more specifically

mattleibow:
There are always 2 sides... A rich text label is great for more specific implementation. But then we have 2 types of labels and all logic might have to be duplicated. 😭

PureWeen:
Alright chatted with @mattleibow about this one a bit more and we're thinking about leveraging different types of spans for this. 

```C#
label.FormattedText = new FormattedString {
    Spans = {
        new HtmlSpan("<b>bold</b>"),
        new MarkdownSpan("## Heading 2"),
        new Span("plain text")
    }
}
````

I still need to spike the idea a bit to see if this is possible but I just wanted to drop a note about the direction we're thinking.

Apologies for not having this baked more before you started :-/
Once we solidify the APIs if you don't have the time we can just rearrange since you've already done all the hard work

**Comment by CZEMacLeod** [comment]
@mattleibow  @PureWeen Yeah - I remember seeing the toolkit markdown control and was thinking of it.
I know that there is a separate parser which is netstandard 2.0 and a base class for renderers.
[Markdown Parser](https://docs.microsoft.com/en-gb/windows/communitytoolkit/parsers/markdownparser)
It would be really good to be able to leverage this work in Xamarin cross platform; it always seemed like this had been written with that in mind.
[Microsoft.Toolkit.Parsers](https://www.nuget.org/packages/Microsoft.Toolkit.Parsers/)
I think there is a [WPF version](https://www.nuget.org/packages/MarkdownViewer/) of the renderer as well as the [UWP version](https://www.nuget.org/packages/Microsoft.Toolkit.Uwp.UI.Controls/)

**Comment by jfversluis** [comment]
> Apologies for not having this baked more before you started :-/
Once we solidify the APIs if you don't have the time we can just rearrange since you've already done all the hard work

Don't worry about it! Let's just call it agile 😉 

I think it's for the better. You guys raised some very good points and the `Span` approach looks pretty good. Doesn't mess with the core functionality of the `Label` too much and allows for all kinds of extensibility. 

I'm happy to look into it, just focussing on the HTML part for now then I'd think and make Markdown a separate one just to not do too much at once. But @PureWeen you mentioned you want to flesh it out some more? Let me know when you're done, or maybe if you're up for it we can have a quick chat about it whenever you're ready.

**Comment by mattleibow** [comment]
@jfversluis Totally ignore the the markdown for now - that was just a look into the future.

With regards to getting started, you can also just have a go. It may just work with only a few extra lines in the existing label. If you hit blocks, then just add comments and we can talk some more.

**Comment by PureWeen** [comment]
@jfversluis the main thing is that I just didn't want to waste your time. The aspects of this I'm curious to flesh out is seeing how possible mixing and matching spans like that will be on the native platform. Labels map to fairly basic content elements on the native platforms so does it make sense to be able to express that level of variation on the spans? Is that even possible?

Or are we setting ourselves up for failure? 

If you want to dive into it a bit like @mattleibow said that would be awesome. If you want to chat or maybe we could live share mess around with it let me know and we can figure out a time for that :-)

**Comment by KSemenenko** [comment]
any updates @jfversluis?

**Comment by jfversluis** [comment]
@KSemenenko not yet, haven't found the spare time to pick this up yet, sorry.

@PureWeen it makes total sense, but I think it's a matter of trying mostly. I think the native APIs also translates it into spans ultimately. The thing I am still a bit worried about is not supporting every tag. For instance, you can see this happening for UWP where I just implemented the most markup things, but I can foresee that people also expect their links to just show up and work as expected. How far do we want to go with that?

In any case, my time is limited at the moment, I will look at it ASAP, but if you want to step on it, by all means! I'm still happy to have a look together then, maybe we should just go ahead and plan that to get things in motion

**Comment by PureWeen** [comment]
@jfversluis can you rebase to fix conflicts when you get a chance?

File: Xamarin.Forms.Platform.Android/Renderers/LabelRenderer.cs [hunk 1]
```diff
@@ -164,8 +166,10 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
 				UpdateLineHeight();
 			else if (e.PropertyName == Label.MaxLinesProperty.PropertyName)
 				UpdateMaxLines();
-			else if (e.PropertyName == ImageButton.PaddingProperty.PropertyName)
```
jfversluis: I noticed this coming in from another commit and although it probably works because of the same string value, it felt wrong.

jfversluis:
@Pureween done

Xamarin.Forms.Platform.Android/Renderers/LabelRenderer.cs
```diff
@@ -294,6 +298,31 @@ void UpdatePadding()
...
+		{
+			if (Element.Text == null)
+				return;
+
+			switch (Element.TextType)
+			{
+
+				case TextType.Html:
+					if (Forms.IsNougatOrNewer)
+						Control.SetText(Html.FromHtml(Element.Text, FromHtmlOptions.ModeCompact), TextView.BufferType.Spannable);
```
PureWeen: - Can this code be centralized into FormsTextView so it's not duplicated between the fast and slow renderer?
- Also, I don't think setting the text here is correct. Shouldn't this just call *UpdateText*? And then inside UpdateText you figure out whether the user is wanting to set HTML or PlainText? The way it's currently implemented if the user sets Element.Text it will always set the text as if TextType == PlainText

```C#
        void Button_Clicked(object sender, EventArgs e)
        {
            lblTextChanges.TextType = TextType.Html;
            lblTextChanges.Text = "<b>" + DateTime.Now.ToString() + "</b>";
        }
```

The above code will not display as HTML but it should

jfversluis:
You're totally right! Moved it into the `UpdateText` altogether and eliminated the need for a separate `UpdateTextType` call.

However, with this change do we still want to pull this up to the `FormsTextView`? I think we can do it, but IMO we should probably look into moving all of `UpdateText` and probably others.

jfversluis:
There are like 100 tests failing on each run, that seems... Odd. I've requested another run.

jfversluis:
build --uitests

jfversluis:
@PureWeen tests much better now! The ones that are still failing are unrelated. Retested locally.

jfversluis:
@PureWeen ALL TESTS GREEN!!! Quickly merge it! :D

jfversluis:
@jsuarezruiz fixed the issue! Because of the refactorings the font was now applied after the HTML which apparently just removes all HTML styling. Added a `TextType == Text` check to a couple of methods to prevent weird behavior.

@PureWeen just need one more bit of green in the form of an approval check? 😬

jsuarezruiz:
@jfversluis Now, it works fine on Android. :)

jfversluis:
Failing tests do seem to be unrelated. Can't seem to trigger a new run. I've retested them locally (with FR on) and they seem to succeed. I think it's good enough!

davidortinau:
Can we not do HTML in XAML?

```
                <Label>
                    <Label.Text>
                        <![CDATA[
                        First Line<br/>SecondLine
                        ]]>
                    </Label.Text>
                </Label>
```

jfversluis:
Hm I missed the CDATA variant. For some reason when you use that the Text property is null. I will see if that is something we can fix. Maybe @StephaneDelcroix has any ideas about that?

It does work when you encode the string like this: 

```
<Label TextType="Html">
    <Label.Text>
        &lt;h1&gt;Hello World!&lt;/h1&gt;&lt;br/&gt;SecondLine
    </Label.Text>
</Label>
```

StephaneDelcroix:
@jfversluis @davidortinau the `<Label.Text>` is redundant, as `Text` is the `ContentProperty` for `Label`

StephaneDelcroix:
@jfversluis @davidortinau 

this works:
```xaml
        <Label x:Name="label2"
                TextType="Html">
            <![CDATA[
                <h1>Hello World!</h1><br/>SecondLine
            ]]>
        </Label>
```

but, as you reported, this assign a null value to `Text`
```xaml
        <Label x:Name="label4"
                TextType="Html">
            <Label.Text>
                <![CDATA[
                    <h1>Hello World!</h1><br/>SecondLine
                ]]>
            </Label.Text>
        </Label>
```

Nanouz:
Hi,

How does this work? I don't understand. I need basic html markup to show but nothing seems to be working for me.

acaliaro:
I thinks it has a very "basic" implementation. I am not been able to use "style". It seems (for Android) that you can use only tags like <b>

TingtingAn:
While  TextType="Html" and Text is Chinese symbols,  it will show messy code on iOS.

acaliaro:
> While  TextType="Html" and Text is Chinese symbols,  it will show messy code on iOS.

Open an issue

marciordonez:
How can I use this? Is it native? Didn't work here :(

jfversluis:
What version are you using @marciordonez? And please define "not working".

tiseidel:
Hi, if I have some HMTL like `<p>German text with some <i>Umlaute</i> like ä ö ü Ä Ö Ü</p>` Android displays them corretly (has an encoding that supports the characters) but iOS can't display them properly. Is there a way I can change the default encoding/ how is the encoding handled?

marciordonez:
> Hi, if I have some HMTL like `<p>German text with some <i>Umlaute</i> like ä ö ü Ä Ö Ü</p>` Android displays them corretly (has an encoding that supports the characters) but iOS can't display them properly. Is there a way I can change the default encoding/ how is the encoding handled?

Try to use native xamarin collection view. This is a recent version.

https://devblogs.microsoft.com/xamarin/xamarin-forms-4-3-introducing-collectionview/

jfversluis:
Hi @tiseidel, there is an issue logged for that and from the looks of it also a PR to fix it. See #8156 for the progress on this.

marciordonez:
With the release of xamarin forms 4.3 this functionality has been fixed.

jfversluis:
That is great to hear! Thank you for letting us know!

PrvnKmr:
Hi,

Will this just allow display of HTML text?
Because when I try using **`<a href>`** tags, even though it shows a link, the url doesn't open in browser as it is supposed to.

Thanks, 

Praveen Kumar

tiseidel:
> 
> 
> Hi,
> 
> Will this just allow display of HTML text?
> Because when I try using **`<a href>`** tags, even though it shows a link, the url doesn't open in browser as it is supposed to.
> 
> Thanks,
> 
> Praveen Kumar

Hi, as you can see here: https://blog.verslu.is/xamarin/xamarin-forms-xamarin/html-in-your-xamarin-forms-label/ links are not working.

You could implement a workaround where you listen for taps/clicks on the Label and then parse the content for any links to open in a browser.

PR #4538, opened 2018-11-27
Fix edge case in unsubscribing weak events

vguzmanp:
When an event has two subscriptions to the same handler,
unsubscribing cleared all the handlers, not only one.

### Description of Change ###
I added a break to stop unsubscribing after the first successful one.

### Issues Resolved ### 
I think this was not reported before.

### API Changes ###
None

### Platforms Affected ### 
Core

Subscribing to a weak event twice and unsubscribing once will unsubscribe only once.

Not applicable

### Testing Procedure ###
I found the issue in a unit test, where doing a swap of two elements in a modified ObservableCollection made one of them lose all the PropertyChanged subscriptions.

### PR Checklist ###

- [X] Has automated tests

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/WeakEventManagerTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/WeakEventManagerTests.cs
@@ -190,6 +190,21 @@ namespace Xamarin.Forms.Core.UnitTests
 		}
 
 		[Test]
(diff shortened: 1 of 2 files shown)


hartez:
> I don't think we should allow subscribing multiple time

Regular events in C# allow subscribing multiple times.

PR #4559, opened 2018-11-28
[C] Seal our attributes

### LINKED ISSUE #1788: [Enhancement] Seal public attributes
## Rationale 

Some of the attributes in Forms not follow the guidance for [CA1813](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1813-avoid-unsealed-attributes). These include:

- `ExportEffectAttribute`
- `ResolutionGroupNameAttribute`
- `DependencyAttribute`

Sealing these attributes would give us a minor performance boost during application startup (as we search through assemblies for Effects and service implementations). 

We should also seal `PreserveAttribute`; while it is not reflected at runtime, it is currently in the `Internals` namespace which is meant to convey that third parties should not be directly utilizing it. Marking it as `sealed` would allow the compiler to help enforce that intent.

## Difficulty: Trivial

Literally adding four words.

## Risks

Since these attributes have been public and unsealed for a while, it's possible that sealing them would be a breaking change for anyone who has subclassed them.

StephaneDelcroix:
### Description of Change ###

Follow CA1813 guidance. And seal our attributes.

### Issues Resolved ### 

- fixes #1788

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Before
seal

After
seal-2

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/DependencyAttribute.cs
+++ b/Xamarin.Forms.Core/DependencyAttribute.cs
@@ -3,7 +3,7 @@ using System;
 namespace Xamarin.Forms
 {
 	[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
-	public class DependencyAttribute : Attribute
+	public sealed class DependencyAttribute : Attribute
 	{
 		public DependencyAttribute(Type implementorType)
 		{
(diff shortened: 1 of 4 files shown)


StephaneDelcroix:
this is a potentially breaking change, but I can't figure a **sane** scenario in which we actually break anything

PR #4571, opened 2018-11-28
update support and sync up nuspecs

PureWeen:
### Description of Change ###
- update android support libraries to preview8
- split applinks and maps into Monodroid81 and Monodroid90 so we have dlls built against correct version otherwise user gets warnings when targeting 8.1 and using maps and applinks

### Testing Procedure ###
Pull down nugets creates and install applinks/maps and make sure it all workss

### PR Checklist ###

- [ ] Has automated tests 


.nuspec/Xamarin.Forms.nuspec
```diff
@@ -34,8 +34,10 @@
       <group targetFramework="tizen40">
         <dependency id="Tizen.NET" version="4.0.0"/>
       </group>
+      <group targetFramework="Xamarin.iOS10">
+        <dependency id="Xamarin.iOS.MaterialComponents" version="60.1.0"/>
```
bholmes: We want to avoid this long term, correct?  Forms Core should not need this dep.

.nuspec/Xamarin.Forms.AppLinks.nuspec
```diff
@@ -16,14 +16,21 @@
       <group targetFramework="MonoAndroid81">
         <dependency id="Xamarin.GooglePlayServices.AppIndexing" version="60.1142.1"/>
       </group>
+      <group targetFramework="MonoAndroid90">
+        <dependency id="Xamarin.GooglePlayServices.AppIndexing" version="60.1142.1"/>
```
bholmes: This seems redundant.  I am guessing that the MonoAndroid81 TF will not be assumed, and that is why this is listed explicitly?

.nuspec/Xamarin.Forms.nuspec
PureWeen: I don't think we are avoiding it.  It's required for the Material Visual components on iOS and the only way to not have the dep would be to create a separate Xamarin.Forms.Material package or just not use material components on iOS.

bholmes: I would like to find a solution where this dep is optional and users that use Material will need to add this dep explicitly.  Need to think about the user experience. 

At the very least we will need to ensure that it does link out when no in use.

.nuspec/Xamarin.Forms.AppLinks.nuspec
PureWeen: I'd need to test if it picks up MonoAndroid81 (it should) 

My other thinking had to do with the nuget listing and that it might be preferred to have it listed under a MonoAndroid90 section

.nuspec/Xamarin.Forms.nuspec
rmarinho: I think it won't be easy, if the dll was built using this there's no way they can build after instal the Nuget without installing this irc. 
Only way is we ship the dll and everything needed on our package. 

But in the UX field, users will just update a package that will install this package, there will be more stuff there but the update experience should just work. and Linker should do the rest of keeping the app small.

PR #4575, opened 2018-11-29
Added the style from .NET Foundation plus variances

mattleibow:
### Description of Change ###

Added the .NET Foundation code style:
 - https://github.com/dotnet/corefx/blob/master/.editorconfig
 - https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

Also added the few variances:
 - https://github.com/xamarin/Xamarin.Forms/blob/master/.github/CONTRIBUTING.md

Variances:
```
[*]
# no newline at the end of files
insert_final_newline = false

[*.cs]
# use tabs in C#
indent_style = tab
# don't use private
dotnet_style_require_accessibility_modifiers = never:suggestion
# use var in most cases
csharp_style_var_for_built_in_types = true:none
csharp_style_var_when_type_is_apparent = true:none
csharp_style_var_elsewhere = false:none
# static is PascalCase, and not s_camelCase
dotnet_naming_style.static_prefix_style.capitalization = pascal_case 
[*.{xaml}]
# XAML uses 2 space indents
indent_size = 2
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,169 @@
+# https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference
+# https://editorconfig.org
+
+# top-most EditorConfig file
 root = true
 
-[*.{cs,xaml}]
+# Default settings:
+# A newline ending every file
+# Use 4 spaces as indentation
+[*]
+insert_final_newline = false
+indent_style = space
+indent_size = 4
+
+[project.json]
+indent_size = 2
+
+# C# files
+[*.cs]
 indent_style = tab
+
+# Modifier preferences
+dotnet_style_require_accessibility_modifiers = never:suggestion
+
+# New line preferences
+csharp_new_line_before_open_brace = all
+csharp_new_line_before_else = true
+csharp_new_line_before_catch = true
+csharp_new_line_before_finally = true
+csharp_new_line_before_members_in_object_initializers = true
+csharp_new_line_before_members_in_anonymous_types = true
+csharp_new_line_between_query_expression_clauses = true
+
+# Indentation preferences
+csharp_indent_block_contents = true
+csharp_indent_braces = false
+csharp_indent_case_contents = true
+csharp_indent_switch_labels = true
+csharp_indent_labels = one_less_than_current
+
+# avoid this. unless absolutely necessary
+dotnet_style_qualification_for_field = false:suggestion
+dotnet_style_qualification_for_property = false:suggestion
+dotnet_style_qualification_for_method = false:suggestion
+dotnet_style_qualification_for_event = false:suggestion
+
+# only use var when it's obvious what the variable type is
+csharp_style_var_for_built_in_types = true:none
+csharp_style_var_when_type_is_apparent = true:none
+csharp_style_var_elsewhere = false:none
+
+# use language keywords instead of BCL types
+dotnet_style_predefined_type_for_locals_parameters_members = true:suggestion
+dotnet_style_predefined_type_for_member_access = true:suggestion
+
+# name all constant fields using PascalCase
+dotnet_naming_rule.constant_fields_should_be_pascal_case.severity = suggestion
+dotnet_naming_rule.constant_fields_should_be_pascal_case.symbols  = constant_fields
+dotnet_naming_rule.constant_fields_should_be_pascal_case.style    = pascal_case_style
+
+dotnet_naming_symbols.constant_fields.applicable_kinds   = field
+dotnet_naming_symbols.constant_fields.required_modifiers = const
+
+dotnet_naming_style.pascal_case_style.capitalization = pascal_case
+
+# static fields should be PascalCase
+dotnet_naming_rule.static_fields_should_have_prefix.severity = suggestion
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -1,5 +1,169 @@
...
+[*.cs]
 indent_style = tab
-csharp_space_after_keywords_in_control_flow_statements = true
\ No newline at end of file
+
+# Modifier preferences
+dotnet_style_require_accessibility_modifiers = never:suggestion
+
+# New line preferences
+csharp_new_line_before_open_brace = all
```
StephaneDelcroix: meh

.editorconfig
```diff
@@ -1,5 +1,169 @@
...
+csharp_new_line_before_finally = true
+csharp_new_line_before_members_in_object_initializers = true
+csharp_new_line_before_members_in_anonymous_types = true
+csharp_new_line_between_query_expression_clauses = true
+
+# Indentation preferences
+csharp_indent_block_contents = true
+csharp_indent_braces = false
+csharp_indent_case_contents = true
+csharp_indent_switch_labels = true
```
StephaneDelcroix: that makes 2 level of indentation for a switch, where it's logically one

.editorconfig
```diff
@@ -1,5 +1,169 @@
...
+
+dotnet_naming_symbols.constant_fields.applicable_kinds   = field
+dotnet_naming_symbols.constant_fields.required_modifiers = const
+
+dotnet_naming_style.pascal_case_style.capitalization = pascal_case
+
+# static fields should be PascalCase
+dotnet_naming_rule.static_fields_should_have_prefix.severity = suggestion
+dotnet_naming_rule.static_fields_should_have_prefix.symbols  = static_fields
+dotnet_naming_rule.static_fields_should_have_prefix.style    = static_prefix_style
```
StephaneDelcroix: I think we use camelCase

.editorconfig
```diff
@@ -1,5 +1,169 @@
...
+
+dotnet_naming_symbols.static_fields.applicable_kinds   = field
+dotnet_naming_symbols.static_fields.required_modifiers = static
+
+dotnet_naming_style.static_prefix_style.capitalization = pascal_case 
+
+# internal and private fields should be _camelCase
+dotnet_naming_rule.camel_case_for_private_internal_fields.severity = suggestion
+dotnet_naming_rule.camel_case_for_private_internal_fields.symbols  = private_internal_fields
+dotnet_naming_rule.camel_case_for_private_internal_fields.style    = camel_case_underscore_style
```
StephaneDelcroix: that's probably the worst decision we ever agreed on

.editorconfig
mattleibow: This is the current style in the code:

```csharp
switch (test)
{
    case 1:
        break;
}

Comment (mattleibow) [reply .editorconfig]: If we change the rules, then it will be a totally different style, unused before.

File: .editorconfig (hunk 1 above)
Comment (mattleibow) [reply .editorconfig]: we can put whatever we want here - but what do we want?
https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference?view=vs-2017#newline

File: .editorconfig (hunk 4 above)
Comment (mattleibow) [reply .editorconfig]: People love it though!

File: .editorconfig (hunk 3 above)
Comment (mattleibow) [reply .editorconfig]: we actuallu have a wide variety:
 - `s_checksum` (old code)
 - `IsFreshProperty (binding propertys and const-like
 - `_counter` (oldish code)
 - `something` (I spotted one or two)

Comment (samhouts) [reply .editorconfig]: I think this was one rule we couldn't agree on so we just ignored it and went on with life.

File: .editorconfig (hunk 1 above)
Comment (samhouts) [reply .editorconfig]: It should be newline in all projects except for the Xamarin.Forms.Xaml projects.

File: .editorconfig (hunk 2 above)
Comment (StephaneDelcroix) [reply .editorconfig]: I prefer this style, used in some places:
```csharp
switch(foo) {
case 1:
    break;
}
```

as it only make a single level of indentation for a single conditional. Oh wait, it require the opening brace to be on the same line ? Now my world start to make some sense...

.editorconfig
StephaneDelcroix: and surprisingly, that's the one I care the less about

.editorconfig
```diff
@@ -1,5 +1,169 @@
...
+# Expression-level preferences
+dotnet_style_object_initializer = true:suggestion
+dotnet_style_collection_initializer = true:suggestion
+dotnet_style_explicit_tuple_names = true:suggestion
+dotnet_style_coalesce_expression = true:suggestion
+dotnet_style_null_propagation = true:suggestion
+
+# Expression-bodied members
+csharp_style_expression_bodied_methods = false:none
+csharp_style_expression_bodied_constructors = false:none
```
StephaneDelcroix: does this means we won't get suggestions for ctors and methods ?

mattleibow: We will still get suggestions if the press Ctrl+. or click the lightbulb - just not a squiggly line. We can change this.

.editorconfig
mattleibow: We can override this for that project...

PR #4598, opened 2018-11-30
[WIP] Updating the Material Visual

mattleibow:
### Description of Change ###

The PR to show the changes for the Material controls.

### Issues Resolved ### 

- fixes #

### API Changes ###
 
| Forms | Android | iOS |
| ------ | -------- | --- |
| Frame | MaterialCardView | Card |
| Button | MaterialButton | Button |
| ProgressBar | tweaked native | ProgressView | 
| Slider | n/a | Slider |
| Entry | TextInputLayout | TextField |

 - `Frame` is done on both, but due to a bug/limitation on iOS, the contents are only clipped on v70+ of material components
 - `Button` is done on both
 - `ProgressBar` is done on iOS, Android is just a tweaked version of the native renderer (new control is pending Google code)
 - `Slider` is done on iOS, Android uses native control
 - `Entry` is almost done, @PureWeen is looking into some issues and will update accordingly :)

### Platforms Affected ### 

- iOS
- Android

 - New Material renderers

Not applicable

### PR Checklist ###

- [ ] Has automated tests 


mattleibow:
I am still working on these renderers, so this is just the initial code. But, there appears to be an issue with layouts on both iOS and Android.

## iOS

iOS does not appear to crop the contents to the cardview, so it hangs over the edge. The way Google does this is to create a special view that has a mask: https://github.com/material-components/material-components-ios/blob/v70.1.0/components/Cards/examples/CardExampleViewController.swift#L74

But this is new code, the "old" way - the version that we currently build against does something more complex: https://github.com/material-components/material-components-ios/blob/v60.1.0/components/Cards/examples/CardExampleViewController.swift#L85

Here is a screenshot, note the sharp corners on the top card and the picture outside the red border on the bottom:
screen shot 2018-11-30 at 19 00 56


## Android

Android does something different for the v28 support and the new AndroidX components, so this may change.

For the v28 support, the cardview is modified so that the padding/contentpadding are not the default functionality. The card no longer applies padding, but rather content padding based on the stroke width. If there is no stroke, then there is no padding - the child must supply it's own.

Based on preliminary research, similar layout logic for the ButtonRenderer needs to be applied.

Screenshot (the blue sticking out of the red was a bug that is fixed in the AndroidX component):

screen shot 2018-11-30 at 19 01 46

Xamarin.Forms.Platform.Android/Material/MaterialFrameRenderer.cs
```diff
@@ -180,129 +123,179 @@ protected virtual void OnElementChanged(ElementChangedEventArgs<Frame> e)
...
+				StrokeWidth = _defaultStrokeWidth;
 			}
 			else
 			{
 				StrokeColor = borderColor.ToAndroid();
-				StrokeWidth = 25;
-			}
 
+				// TODO: this should be 1, but it is high here to demonstrate the layout issue
+				StrokeWidth = (int)Context.ToPixels(10);
```
mattleibow: The Android layout needs to be fixed, and to demonstrate the issue, I have left this a large number.

Xamarin.Forms.Platform.iOS/Material/MaterialFrameRenderer.cs
```diff
@@ -1,51 +1,65 @@
...
 using UIKit;
 using Xamarin.Forms;
 using MCard = MaterialComponents.Card;
 
+// this won't go here permanently it's just for testing at this point
 [assembly: ExportRenderer(typeof(Xamarin.Forms.Frame), typeof(Xamarin.Forms.Platform.iOS.Material.MaterialFrameRenderer), new[] { typeof(VisualRendererMarker.Material) })]
+
 namespace Xamarin.Forms.Platform.iOS.Material
 {
 	public class MaterialFrameRenderer : MCard, IVisualElementRenderer
```
mattleibow: iOS needs some research to determine what the best way to crop the contents to fit the curve.

Xamarin.Forms.Core/Button.cs
```diff
@@ -390,26 +390,5 @@ public override object ConvertFromInvariantString(string value)
 				return new ButtonContentLayout(position, spacing);
 			}
 		}
-
-
-		protected override void OnPropertyChanged([CallerMemberName] string propertyName = null)
```
mattleibow: This is now set via the renderers

Xamarin.Forms.Core/Frame.cs
```diff
@@ -56,6 +58,29 @@ public float CornerRadius
 
 		int IBorderElement.CornerRadius => (int)CornerRadius;
 
+		void IFrameController.SetContentPadding(Thickness value)
```
mattleibow: This is mainly for Android which has additional padding due to the border.

Xamarin.Forms.Platform.Android/Material/MaterialButtonRenderer.cs
```diff
@@ -444,87 +286,139 @@ void UpdateBitmap()
...
+			}
+		}
+
+		void UpdateBorder()
+		{
+			if (_disposed || Element == null)
+				return;
+
+			var cornerRadius = Element.CornerRadius;
+			if (cornerRadius >= 0f || _defaultCornerRadius >= 0f)
```
mattleibow: As with most of the properties, first check to see if we are setting default values to defaults. This reduces the number of times we actually talk via JNI. And, no need to set the same values.

Xamarin.Forms.Platform.Android/Material/MaterialFrameRenderer.cs
```diff
@@ -180,129 +128,178 @@ protected virtual void OnElementChanged(ElementChangedEventArgs<Frame> e)
...
 			{
 				StrokeColor = borderColor.ToAndroid();
-				StrokeWidth = 25;
+				StrokeWidth = (int)Context.ToPixels(1);
 			}
 
+			// update the native and forms view with the border
+			SetContentPadding(StrokeWidth, StrokeWidth, StrokeWidth, StrokeWidth);
+			if (Element is IFrameController frame)
+				frame.SetContentPadding(Context.FromPixels(StrokeWidth));
```
mattleibow: This chunk of code is probably going to go away with AndroidX as they have moved the contents into a separate view with it's own padding.

Xamarin.Forms.Platform.iOS/Material/MaterialButtonRenderer.cs
```diff
@@ -94,9 +86,20 @@ protected override void OnElementChanged(ElementChangedEventArgs<Button> e)
...
 		}
 
+		protected virtual ButtonScheme CreateButtonScheme()
+		{
+			return new ButtonScheme
+			{
+				TypographyScheme = new TypographyScheme(),
+				ColorScheme = MaterialColors.Light.CreateColorScheme()
+			};
+		}
```
mattleibow: The iOS Material design is set via the Themer applying a Scheme. In this case, we are using the "light" theme with values copied from Android.

Xamarin.Forms.Platform.iOS/Material/MaterialButtonRenderer.cs
```diff
@@ -94,9 +86,20 @@ protected override void OnElementChanged(ElementChangedEventArgs<Button> e)
...
+
 		protected override MButton CreateNativeControl()
 		{
-			return new MButton();
+			var button = new MButton();
+			ContainedButtonThemer.ApplyScheme(CreateButtonScheme(), button);
```
mattleibow: This `ContainedButtonThemer` represents the "normal" button of a solid background. There are variants, no background/text, outline button, FAB. These are no longer individual/separate controls.

Xamarin.Forms.Platform.iOS/Material/MaterialButtonRenderer.cs
```diff
@@ -132,54 +135,101 @@ protected override void SetAccessibilityLabel()
 			base.SetAccessibilityLabel();
 		}
 
-		void SetControlPropertiesFromProxy()
-		{
-			foreach (UIControlState uiControlState in s_controlStates)
-			{
-				Control.SetTitleColor(UIButton.Appearance.TitleColor(uiControlState), uiControlState); // if new values are null, old values are preserved.
-				Control.SetTitleShadowColor(UIButton.Appearance.TitleShadowColor(uiControlState), uiControlState);
-				Control.SetBackgroundImage(UIButton.Appearance.BackgroundImageForState(uiControlState), uiControlState);
```
mattleibow: Don't use the UIAppearance types - i don't think.

Xamarin.Forms.Platform.iOS/Material/MaterialColors.cs
```diff
@@ -0,0 +1,84 @@
+﻿using MaterialComponents;
+using UIKit;
+
+namespace Xamarin.Forms.Platform.iOS.Material
+{
+	// colors from material-components-android
+	// https://github.com/material-components/material-components-android/blob/3637c23078afc909e42833fd1c5fd47bb3271b5f/lib/java/com/google/android/material/color/res/values/colors.xml
+	internal static class MaterialColors
```
mattleibow: This is the set of "Material" colors that we override the defaults to match Android

Xamarin.Forms.Platform.iOS/Material/MaterialFrameRenderer.cs
```diff
@@ -54,112 +77,122 @@ public void SetElement(VisualElement element)
 
 			if (element != null)
 			{
-				element.PropertyChanged += OnElementPropertyChanged;
+				CardThemer.ApplyScheme(CreateCardScheme(), this);
```
mattleibow: Theme is applied here.

Xamarin.Forms.Platform.Android/Resources/values/styles.xml
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="UTF-8" ?>
 <resources>
-  <style name="XamarinFormsMaterialTheme" parent="Theme.MaterialComponents">
+  <style name="XamarinFormsMaterialTheme" parent="Theme.MaterialComponents.Light.DarkActionBar">
```
mattleibow: Switched this to the template default.

Xamarin.Forms.Platform.iOS/Material/MaterialSliderRenderer.cs
```diff
@@ -0,0 +1,183 @@
...
+				Control.SetThumbColor(color.ToUIColor(), UIControlState.Normal);
+		}
+
+		void OnControlValueChanged(object sender, EventArgs eventArgs)
+		{
+			Element.SetValueFromRenderer(Slider.ValueProperty, Control.Value);
+		}
+
+		// TODO: decide if we want to just by default like Android does
+		//[Export("slider:shouldJumpToValue:")]
```
mattleibow: The native iOS renderer has a touch gesture detector to handle taps. Done via the platform-specifics. For Material, this is already implemented by default. However, we may want to DISABLE this feature.

Xamarin.Forms.Platform.iOS/Renderers/ImageRenderer.cs
```diff
@@ -120,6 +120,27 @@ public interface IImageSourceHandler : IRegisterable
 		Task<UIImage> LoadImageAsync(ImageSource imagesource, CancellationToken cancelationToken = default(CancellationToken), float scale = 1);
 	}
 
+	internal static class ImageSourceExtensions
+	{
+		public static async Task<UIImage> GetNativeImageAsync(this ImageSource source)
```
mattleibow: I extracted the simple logic of taking an ImageSource and getting a UIImage - this chunk is duplicated in at least 20 places. Hopefully, moving forward, we can use this instead. This is mostly for image buttons, icons, etc. The `<Image />` element has more functionality for loading and progress tracking.

Xamarin.Forms.Platform.iOS/Material/MaterialSliderRenderer.cs
```diff
@@ -0,0 +1,183 @@
...
+using Xamarin.Forms.PlatformConfiguration.iOSSpecific;
+using MSlider = MaterialComponents.Slider;
+using Specifics = Xamarin.Forms.PlatformConfiguration.iOSSpecific.Slider;
+
+[assembly: ExportRenderer(typeof(Xamarin.Forms.Slider), typeof(Xamarin.Forms.Platform.iOS.Material.MaterialSliderRenderer), new[] { typeof(VisualRendererMarker.Material) })]
+
+namespace Xamarin.Forms.Platform.iOS.Material
+{
+	public class MaterialSliderRenderer : ViewRenderer<Slider, MSlider>,
+		ISliderDelegate
```
mattleibow: This should actually be removed if we don't want to disable the touch/jump action.

Xamarin.Forms.Core/ProgressBar.cs
```diff
@@ -47,30 +47,5 @@ public Task<bool> ProgressTo(double value, uint length, Easing easing)
 		{
 			return _platformConfigurationRegistry.Value.On<T>();
 		}
-
-
-		protected override void OnPropertyChanged([CallerMemberName] string propertyName = null)
```
mattleibow: This is now in the renderers

Xamarin.Forms.Platform.iOS/Material/MaterialFrameRenderer.cs
```diff
@@ -1,165 +1,235 @@
...
 		}
 
-		public event EventHandler<VisualElementChangedEventArgs> ElementChanged;
+		public override void LayoutSubviews()
+		{
+			base.LayoutSubviews();
+
+			// if the card's shadow has a path/shape, then use that to clip the contents
+			var content = Element?.Content;
+			if (content != null && Layer is ShapedShadowLayer shadowLayer && shadowLayer.ShapeLayer.Path is CGPath shapePath)
```
mattleibow: This code will only run when the components is updated to 72.1 - there is a change somewhere between 60 and 70 that actually sets the shape on the layer. The code is exactly the same, just that with the new implementation of the card, it gets set.

Right, now, the contents will hang out, but after updating the content is clipped.

Xamarin.Forms.Platform.iOS/Material/MaterialFrameRenderer.cs
```diff
@@ -1,165 +1,235 @@
...
 		}
 
-		public UIViewController ViewController
+		public override void WillRemoveSubview(UIView uiview)
 		{
-			get { return null; }
+			var content = Element?.Content;
+			if (content != null && uiview == Platform.GetRenderer(content))
+			{
+				uiview.Layer.Mask = null;
```
mattleibow: This may not be needed, but I want to make sure that we unhook everything that we hooked. This just empties the card's mask when the element is removed from the card.

Xamarin.Forms.Platform.iOS/Material/MaterialFrameRenderer.cs
mattleibow: @PureWeen This is here so that we don't have to change code later when we update.

mattleibow:
The actual changes from your review to now is: https://github.com/xamarin/Xamarin.Forms/compare/a699f1d6d208f66e53a848dda40e9bba433f4cd1...b894ab6c1a3f579bda14dbbbdf3b282390cb899f

Xamarin.Forms.Platform.iOS/Material/MaterialFrameRenderer.cs
```diff
@@ -1,165 +1,235 @@
...
-		public SizeRequest GetDesiredSize(double widthConstraint, double heightConstraint)
+		void UpdateBackgroundColor()
 		{
-			return NativeView.GetSizeRequest(widthConstraint, heightConstraint, 44, 44);
+			if (_defaultBackgroundColor == null)
+				_defaultBackgroundColor = BackgroundColor;
+
+			var bgColor = Element.BackgroundColor;
+			if (bgColor.IsDefault)
+				BackgroundColor = _defaultBackgroundColor;
```
PureWeen: Is this the right thing to set? Or do we need to set it on the scheme and then reapply the scheme?

Xamarin.Forms.Controls/ControlGalleryPages/VisualGallery.xaml.cs
```diff
@@ -1,13 +1,11 @@
...
-using System.Collections.Generic;
-
-using Xamarin.Forms;
 
 namespace Xamarin.Forms.Controls
 {
 	public partial class VisualGallery : ContentPage
 	{
 		bool isVisible = false;
+
```
StephaneDelcroix: I get this one

Xamarin.Forms.Controls/ControlGalleryPages/VisualGallery.xaml.cs
```diff
@@ -16,22 +14,32 @@ public VisualGallery()
 		protected override void OnAppearing()
 		{
 			isVisible = true;
+
 			base.OnAppearing();
+
```
StephaneDelcroix: but I don't get those. please avoid reformatting code you do not change

StephaneDelcroix: Sorry, this is gallery code, I actually do not care, do whatever you want :)

Xamarin.Forms.Core/Button.cs
StephaneDelcroix: as it should be

Xamarin.Forms.Core/Frame.cs
StephaneDelcroix: I have a very bad feeling about this, but it's kinda private API, so let's go with it if we can't infer that information from the Measuring API

Xamarin.Forms.Platform.Android/Resources/values/styles.xml
```diff
@@ -1,7 +1,7 @@
...
   </style>
-  <style name="XamarinFormsMaterialProgressBarHorizontal" parent="android:Widget.ProgressBar.Horizontal">
+  <style name="XamarinFormsMaterialProgressBarHorizontal" parent="Widget.AppCompat.ProgressBar.Horizontal">
```
PureWeen: The reason here that I was using *android:Widget.ProgressBar.Horizontal* vs *Widget.AppCompat.ProgressBar.Horizontal*

Is that the I couldn't get the height of the progress bar to grow when it's set to *Widget.AppCompat.ProgressBar.Horizontal*

image

Xamarin.Forms.Core/Frame.cs
mattleibow: I believe this is temporary... The AndroidX controls are not using this method anymore, so things will be grand! I hope...

Xamarin.Forms.Platform.Android/Resources/values/styles.xml
mattleibow: I believe it has to do with this: 
https://github.com/aosp-mirror/platform_frameworks_base/blob/e80b45506501815061b079dcb10bf87443bd385d/core/res/res/values/styles_material.xml#L778-L783

We can probably override these values in our style if need be.

mattleibow:
Closing this PR as the changes have been merged into https://github.com/xamarin/Xamarin.Forms/pull/4733

PR #4612, opened 2018-12-03
[Android] Fixes keyboard navigation in custom view

### LINKED ISSUE #4561: Android: Keyboard navigation does not work after updating to 3.3
### Description
After updating to 3.3 the keyboard navigation that I have in my native Android View which is rendered in Xamarin Forms no longer works. 

The FocusSearch method that is overridden in the VisualElementRenderer seems to causes the issue. It is expected to have more complex native Android controls that could be rendered in Xamarin Forms and breaking the keyboard navigation (respecting only the TabIndex) is not correct. 

### Steps to Reproduce

1. Run the attached project.
2. Focus the first EditText.
3. Try to navigate with the keyboard to the next EditText.

### Expected Behavior
You should be able to see a Button in the keyboard through which you can perform the navigation.

### Actual Behavior
Can no longer navigate using the keyboard.
### Basic Information

- Version with issue: 3.3.0
- Last known good version:3.2.0
- IDE: Visual Studio 2017
- Platform Target Frameworks: 
  - Android: 8.1

### Reproduction Link

[TabIndexBug.zip](https://github.com/xamarin/Xamarin.Forms/files/2623632/TabIndexBug.zip)

paymicro:
### Description of Change ###

In our implementation of `TabStop` in Android dropped the keyboard navigation via native settings `NextFocus<direction>Id`. For fix it added check if `NextFocus<direction>Id` property is set on focused view then returns the native function result. So, `NextFocus<direction>Id` takes precedence over the implementation of `TabStop`.

### Issues Resolved ### 

- fixes #4561 

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

Controls with custom renderer can use `NextFocus<direction>Id` to customize keyboard navigation.

Not applicable

### Testing Procedure ###

- run UItest 4561
- select any textbox and try using the TAB key navigation
- focus should change from one textbox to another

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.ControlGallery.Android/CustomRenderers.cs
+++ b/Xamarin.Forms.ControlGallery.Android/CustomRenderers.cs
@@ -40,7 +40,7 @@ using Xamarin.Forms.Controls.Issues;
 //[assembly: ExportRenderer(typeof(AndroidHelpText.HintLabel), typeof(HintLabel))]
 [assembly: ExportRenderer(typeof(QuickCollectNavigationPage), typeof(QuickCollectNavigationPageRenderer))]
 [assembly: ExportRenderer(typeof(Issue4782.Issue4782ImageButton), typeof(Issue4782ImageButtonImageButtonRenderer))]
-
+[assembly: ExportRenderer(typeof(Issue4561.CustomView), typeof(Issue4561CustomViewRenderer))]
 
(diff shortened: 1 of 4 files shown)


APopatanasov:
@paymicro  @mattleibow When this will be merged? It is very important for my application and currently I am not able to update to the latest version of Xamarin Forms because of it.

PR #4614, opened 2018-12-03
[iOS] Fixed ListView cannot reset selected status

### LINKED ISSUE #4394: [iOS] ListView cannot reset selected status
### Description
ListView cannot reset selected status
[UnselectedListView.zip](https://github.com/xamarin/Xamarin.Forms/files/2577065/UnselectedListView.zip)

### Steps to Reproduce
as in my simple demo

1. click to select a item, which will become selected, and it shows SelectedBackgroundView 
2. and click **same** item again,  it try to clear selected status by `ItemsListView.SelectedItem = null;`
3. then it should clear SelectedBackgroundView, however, it doesn't

### Basic Information

- Version with issue: 3.3.0.967583
- IDE: VS
- Platform Target Frameworks: 
  - iOS:  

### Reproduction Link

[UnselectedListView.zip](https://github.com/xamarin/Xamarin.Forms/files/2577065/UnselectedListView.zip)

### Reason:

1.  ListViewDataSource will set _selectionFromNative to true, when clicked on an item.
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs#L1124
2. and reset it to false when SelectedItem Changed.
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs#L1064
3. however, if click on **same** item,  SelectedItem is not changed, and it doesn't have chance to reset _selectionFromNative to false.
4.  if ItemsListView.SelectedItem = null then, because of _selectionFromNative is still true, it will prevent _uiTableView.DeselectRow() to executed in the same method OnItemSelected.
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs#L1092
5. thus, it cannot reset selected status, even SelectedItem is set to null.

masonyc:
### Description of Change ###
Added _lastSelected variable in listview renderer to support unselect action

### Issues Resolved ###
fixes #4394 

### API Changes ###
None

### Platforms Affected ###
iOS

In iOS, now you can deselect the last selected item in the list view.

## Before ##
before

## After ##
after

### Testing Procedure ###
Start ControlGallery.iOS, Search and go to G4394 Page in tests screen, you should be able to unselect the selected item now


rmarinho:
This will be a breaking change , I wonder if we should add an enhancement for SelectionMode.

masonyc:
> This will be a breaking change , I wonder if we should add an enhancement for SelectionMode.

But the old behavior does not exist in android. This just makes them act in a same way so I think we dont need to add an option in selection mode?

mattleibow:
This is a feature we wrote for Android specifically:
https://github.com/xamarin/Xamarin.Forms/blob/505a855cad39322c8b235281d71c35b70c9df129/Xamarin.Forms.Platform.Android/Renderers/ListViewAdapter.cs#L627-L648

So yes, technically we are fixing a bug... But, this is a breaking change for iOS - suddenly double taps are deselected. And, how does this work on UWP? If this is just for Android, then this is a feature that might need to go behind some flag/property. 

Finally, this is really a feature that should work with `ListViewSelectionMode`. What happens when that is set to `None` and then with `Single`?

We could an another mode here to be `Toggle`?

 - `None` - never select
 - `Single` - select and never deselect
 - `Toggle` - select and deselect

PureWeen:
I agree with @mattleibow 
We need a way to fix this but enable the developer to define the behavior so it's not breaking

PR #4595, opened 2018-11-30
[Shell] allow missing trailing slash on uri

StephaneDelcroix:
### Description of Change ###

do not crash anymore while navigation to `app:///home/foo` (without a trailing slash)

### Issues Resolved ### 

/

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

### Behavioral/Visual Changes ###

None

### Before/After Screenshots ### 

Not applicable

### PR Checklist ###

- [ ] Has automated tests 
- [x] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard

PR #4615, opened 2018-12-03
[Shell] Allow queryParams navigation

### LINKED ISSUE #4596: [Shell] queryParams in navigation not handled
### Description

while navigating in shell

```csharp
await shell.GoToAsync("app:///sdxshell/about?welcome=sdx");
```

the queryParams are not handled.

They should:
1. check is the page.VM implements IQueryAttributes
2. check if the page implements IQueryAttributes
3. look for a mapped property in the page VM
4. look for a mapped property on the page itself

StephaneDelcroix:
### Description of Change ###

~~NOTE: please merge #4595 first, and rebase this on top~~

Allow uri navigation with queryParams. queryParams can be applied
to:
- the ShellContent (if inherited)
- the Content (or inflated ContentTemplate) of the ShellContent
- the BindingContext of the Content

those 3 can inherit an interface and handle the queryParams, or
the types can be decorated with some [QueryProperty] attributes.
Attributes are handled on the 3 typesL ShellContent, actual content,
or bindingcontext.

### Issues Resolved ### 

- fixes #4596

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Shell/Shell.cs
+++ b/Xamarin.Forms.Core/Shell/Shell.cs
@@ -402,32 +402,31 @@ namespace Xamarin.Forms
 				prefix = route + ".";
 			}
 
-			var type = element.GetType();
-			var typeInfo = type.GetTypeInfo();
-#if NETSTANDARD1_0
-			var effectAttributes = typeInfo.GetCustomAttributes(typeof(QueryPropertyAttribute), true).ToArray();
-#else
-			var effectAttributes = typeInfo.GetCustomAttributes(typeof(QueryPropertyAttribute), true);
-#endif
+			//if the lastItem is implicitly wrapped, get the actual ShellContent
+			if (isLastItem) {
+				if (element is ShellItem shellitem && shellitem.Items.FirstOrDefault() is ShellSection section)
+					element = section;
+				if (element is ShellSection shellsection && shellsection.Items.FirstOrDefault() is ShellContent content)
+					element = content;
+				if (element is ShellContent shellcontent && shellcontent.Content is Element e)
+					element = e;
+			}
 
-			if (effectAttributes.Length == 0)
+			if (!(element is BaseShellItem baseShellItem))
 				return;
 
-			foreach (var a in effectAttributes)
-			{
-				if (a is QueryPropertyAttribute attrib)
-				{
-					if (query.TryGetValue(prefix + attrib.QueryId, out var value))
-					{
-						PropertyInfo prop = type.GetRuntimeProperty(attrib.Name);
-
-						if (prop != null && prop.CanWrite && prop.SetMethod.IsPublic)
-						{
-							prop.SetValue(element, value);
-						}
-					}
-				}
+			//filter the query to only apply the keys with matching prefix
+			var filteredQuery = new Dictionary<string, string>(query.Count);
+			foreach (var q in query) {
+				if (!q.Key.StartsWith(prefix, StringComparison.Ordinal))
+					continue;
+				var key = q.Key.Substring(prefix.Length);
+				if (key.Contains('.'))
+					continue;
+				filteredQuery.Add(key, q.Value);
 			}
+
+			baseShellItem.ApplyQueryAttributes(filteredQuery);
 		}
 
 		static string GenerateQueryString(Dictionary<string, string> queryData)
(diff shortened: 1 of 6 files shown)


Xamarin.Forms.Core/Shell/Shell.cs
```diff
@@ -450,41 +449,41 @@ static string GenerateQueryString(Dictionary<string, string> queryData)
 
 		static void GetQueryStringData(Element element, bool isLastItem, Dictionary<string, string> result)
 		{
-			string prefix = string.Empty;
```
StephaneDelcroix: I think we do not need this.

Xamarin.Forms.Core/Shell/ShellGroupItem.cs
```diff
@@ -5,10 +5,9 @@ public class ShellGroupItem : BaseShellItem
 		public static readonly BindableProperty FlyoutDisplayOptionsProperty =
 			BindableProperty.Create(nameof(FlyoutDisplayOptions), typeof(FlyoutDisplayOptions), typeof(ShellItem), FlyoutDisplayOptions.AsSingleItem, BindingMode.OneTime);
 
-		public FlyoutDisplayOptions FlyoutDisplayOptions
-		{
-			get { return (FlyoutDisplayOptions)GetValue(FlyoutDisplayOptionsProperty); }
-			set { SetValue(FlyoutDisplayOptionsProperty, value); }
+		public FlyoutDisplayOptions FlyoutDisplayOptions {
```
hartez: Revert this change so the history isn't polluted.

Xamarin.Forms.Core/Shell/Shell.cs
```diff
@@ -402,32 +402,31 @@ internal static void ApplyQueryAttributes(Element element, IDictionary<string, s
...
-			var effectAttributes = typeInfo.GetCustomAttributes(typeof(QueryPropertyAttribute), true).ToArray();
-#else
-			var effectAttributes = typeInfo.GetCustomAttributes(typeof(QueryPropertyAttribute), true);
-#endif
-
-			if (effectAttributes.Length == 0)
-				return;
+			//if the lastItem is implicitly wrapped, get the actual ShellContent
+			if (isLastItem) {
+				if (element is ShellItem shellitem && shellitem.Items.FirstOrDefault() is ShellSection section)
```
kingces95: "Implicitly wrapped" implies to me a single element in the list. So, instead of `FirstOrDefault` I'd expect `SingleOrDefault`. Of course, I also live in grass shack and commute in a canoe, so...

Xamarin.Forms.Core/Shell/Shell.cs
```diff
@@ -402,32 +402,31 @@ internal static void ApplyQueryAttributes(Element element, IDictionary<string, s
...
-						if (prop != null && prop.CanWrite && prop.SetMethod.IsPublic)
-						{
-							prop.SetValue(element, value);
-						}
-					}
-				}
+			//filter the query to only apply the keys with matching prefix
+			var filteredQuery = new Dictionary<string, string>(query.Count);
+			foreach (var q in query) {
+				if (!q.Key.StartsWith(prefix, StringComparison.Ordinal))
```
kingces95: So the query parameters must contain the prefix or they'll be filtered out. I was expecting there'd be an implied prefix inferred from the URI. As it stands right now, if the prefix if 'foo.bar.baz' then all they query parameter must have that prefix to be included right? Is that not verbose? I guess I'd need to see an example. Anyway, it'll work.

kingces95: Oh. I sse. So, I able to whack values onto any part of the segment leading up to the target? Wonderful!

Xamarin.Forms.Core/Shell/ShellContent.cs
```diff
@@ -40,6 +61,12 @@ Page IShellContentController.GetOrCreateContent()
 			if (result != null && result.Parent != this)
 				OnChildAdded(result);
 
+
+			if (_delayedQueryParams != null && result  != null) {
```
kingces95: Any chance this get's applied sometime long after the navigation occurs?

PR #4002, opened 2018-10-04
[iOS] Change default background color of Cell to UIColor.Clear

AndreiMisiukevich:
### Description of Change ###
Hi
I suggest to change default background color of iOS cell to UIColor.Clear;

If developer wants to change BackgroudColor he will be able to set color of View in his ViewCell
Don't see any reasons to force set white/gray background as default. Maybe I missed something?

### Issues Resolved ### 
- fixes #3450

### API Changes ###
Changed:
 - [iOS] Default cell color to Transparent

### Platforms Affected ### 
**iOS**

### Behavioral/Visual Changes ###
Create possibility to have transparent group header (for listView)

[BEFORE]

[AFTER]

### Testing Procedure ###
Sample
```csharp
	public partial class App : Application
	{
		public App ()
		{
			InitializeComponent ();

			MainPage = new ContentPage
			{
				BackgroundColor = Color.Black,
				Content = new ListView
				{
					BackgroundColor = Color.Transparent,
					SeparatorVisibility = SeparatorVisibility.None,
					IsGroupingEnabled = true,
					GroupHeaderTemplate = new DataTemplate(() => {
						return new ViewCell
						{
							Height = 60,
							View = new ContentView
							{
								Content = new Label
								{
									TextColor = Color.White,
									Text = "TEST LABEL"
								},
								HeightRequest = 60,
								BackgroundColor = Color.Transparent
							}
						};
					}),
					ItemTemplate = new DataTemplate(() => {
						return new ViewCell
						{
							Height = 80,
							View = new ContentView
							{
								BackgroundColor = Color.Yellow
							}
						};
					}),
					ItemsSource = new List<List<string>>
					{
						new List<string>
						{
							"Test"
						}
					}
				}
			};
		}
	}
```

PR #4616, opened 2018-12-03
[Enhancement] Added GroupHeaderStyle property for iOS specific ListView class

AndreiMisiukevich:
### Description of Change ###
Created possibility to manage group header style from shared code. (Currently it's impossible at all as related class is incapsulated)

### Issues Resolved ### 
fixes https://github.com/xamarin/Xamarin.Forms/pull/4002#issuecomment-431049817
So, it will allow to set Grouped style and avoid such issues

### API Changes ###

Added: ( Xamarin.Forms.PlatformConfiguration.iOSSpecific.ListView)
- GroupHeaderStyleProperty
- SetGroupHeaderStyle (and 1 override)
- GetGroupHeaderStyle

Added: GroupHeaderStyle enum in the same namespace

### Platforms Affected ### 
- Core
- iOS

image

### Testing Procedure ###

Add next line of code to Bugzilla34912 after list declaration and see that Grouped style is applied (like on Android)

```csharp
Xamarin.Forms.PlatformConfiguration.iOSSpecific.ListView.SetGroupHeaderStyle(list, PlatformConfiguration.iOSSpecific.GroupHeaderStyle.Grouped);
```

### DIFF
--- a/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/ListView.cs
+++ b/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/ListView.cs
@@ -26,5 +26,29 @@
 			SetSeparatorStyle(config.Element, value);
 			return config;
 		}
+
+		public static readonly BindableProperty GroupHeaderStyleProperty = BindableProperty.Create(nameof(GroupHeaderStyle), typeof(GroupHeaderStyle), typeof(FormsElement), GroupHeaderStyle.Plain);
+
+		public static GroupHeaderStyle GetGroupHeaderStyle(BindableObject element)
+		{
+			return (GroupHeaderStyle)element.GetValue(GroupHeaderStyleProperty);
+		}
+
+		public static void SetGroupHeaderStyle(BindableObject element, GroupHeaderStyle value)
+		{
+			element.SetValue(GroupHeaderStyleProperty, value);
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/ListView.cs
```diff
@@ -26,5 +26,28 @@ public static SeparatorStyle GetSeparatorStyle(this IPlatformElementConfiguratio
 			SetSeparatorStyle(config.Element, value);
 			return config;
 		}
+
+		public static readonly BindableProperty GroupHeaderStyleProperty = BindableProperty.Create(nameof(GroupHeaderStyle), typeof(GroupHeaderStyle), typeof(FormsElement), GroupHeaderStyle.Plain);
```
rmarinho: We should make a note this will only work when setting once, since we are not handling if this property changes on the iOS platform. I do think it doesn't make sense to change this after it's being applied the first time.

AndreiMisiukevich: @rmarinho agreed.
What is your proposition? I thought about it, but have no idea, how to make it more user-friendly.
Just leave a doc-comment something like

```csharp
        /// <summary>
        /// You should change this value during setting up a list. Changing this value won't affect style after list rendering.
        /// </summary>
```

PR #4626, opened 2018-12-04
[Shell] expose CurrentShell

### LINKED ISSUE #4625: [Shell] statically expose the current shell
## Summary
Getting a reference to the current shell is tedious and repetitive. As the shell has to be the App.MainPage, we could expose a static version of `GoToAsync()` that tries to get the shell from there. If the shell is not found, an exception will be thrown.

If the user want to navigate on a shell that's not currently App.MainPage (multiple shell app, anyone ?) he can still keep a reference to that shell and use the non-static GoToAsync

this is required for #4594

## API Changes

```csharp
public Task static Shell.GoToAsync(ShellNavigationState state, bool animate = true);
```
## Intended Use Case

replace this usage
```csharp
var shell = Application.Current.MainPage as Shell;
await shell.GoToAsync("app:///s/about?welcome=you");
```

by
```csharp
await Shell.GoToAsync("app:///s/about?welcome=you");
```

StephaneDelcroix:
### Description of Change ###

[Shell] expose CurrentShell

### Issues Resolved ### 

- fixes #4625

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Shell/Shell.cs
+++ b/Xamarin.Forms.Core/Shell/Shell.cs
@@ -322,7 +322,7 @@ namespace Xamarin.Forms
 			OnNavigated(new ShellNavigatedEventArgs(oldState, CurrentState, source));
 		}
 
-		public static Shell CurrentShell => Application.Current?.MainPage as Shell ?? throw new InvalidOperationException("Application.Current.MainPage is not a Shell");
+		public static Shell CurrentShell => Application.Current?.MainPage as Shell;
 
 		public async Task GoToAsync(ShellNavigationState state, bool animate = true)
 		{
--- a/Xamarin.Forms.Core/Shell/Shell.cs
+++ b/Xamarin.Forms.Core/Shell/Shell.cs
@@ -162,8 +162,6 @@ namespace Xamarin.Forms
 
 		static readonly BindablePropertyKey MenuItemsPropertyKey =
 			BindableProperty.CreateReadOnly(nameof(MenuItems), typeof(MenuItemCollection), typeof(Shell), null, defaultValueCreator: bo => new MenuItemCollection());
-			
-#region IShellController
 
 		List<(IAppearanceObserver Observer, Element Pivot)> _appearanceObservers = new List<(IAppearanceObserver Observer, Element Pivot)>();
 		List<IFlyoutBehaviorObserver> _flyoutBehaviorObservers = new List<IFlyoutBehaviorObserver>();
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Core/Shell/Shell.cs
```diff
@@ -324,9 +322,7 @@ void IShellController.UpdateCurrentState(ShellNavigationSource source)
 			OnNavigated(new ShellNavigatedEventArgs(oldState, CurrentState, source));
 		}
 
-#endregion IShellController
-
-#region URI Navigation
+		public static Shell CurrentShell => Application.Current?.MainPage as Shell ?? throw new InvalidOperationException("Application.Current.MainPage is not a Shell");
```
paymicro: is it necessary to throw an exception here?

Xamarin.Forms.Core/Shell/Shell.cs
```diff
@@ -162,8 +162,6 @@ static void OnShellColorValueChanged(BindableObject bindable, object oldValue, o
 
 		static readonly BindablePropertyKey MenuItemsPropertyKey =
 			BindableProperty.CreateReadOnly(nameof(MenuItems), typeof(MenuItemCollection), typeof(Shell), null, defaultValueCreator: bo => new MenuItemCollection());
-			
-#region IShellController
```
hartez: Why drop the regions?

Xamarin.Forms.Core/Shell/Shell.cs
hartez: I'm wondering the same thing. Do we have any mechanism for a component to determine if it's running in a Shell app (as opposed to a classic app)? If not, does it make sense to allow a check for `CurrentShell == null`?

Xamarin.Forms.Core/Shell/Shell.cs
StephaneDelcroix: for a few reasons:
- I don't like them
- they're not consistent with the rest of the code, or even the rest of the shell code
- they enforce some weird code ordering (this one had a region for BP, then a region for interface implementation. those 2, for example, can collide)
- they're out of date faster than code comments

now, I get why one might like them: being able to collapse region. I know a thing or two about where to place opening braces, that saves SO MUCH vertical space that collapsing is no longer required AT ALL.

Xamarin.Forms.Core/Shell/Shell.cs
StephaneDelcroix: go one
```suggestion
		public static Shell CurrentShell => Application.Current?.MainPage as Shell;
```

Xamarin.Forms.Core/Shell/Shell.cs
hartez: Fair enough. I didn't realize they weren't being applied consistently. 

> I know a thing or two about where to place opening braces, that saves SO MUCH vertical space that collapsing is no longer required AT ALL.
 
:)

PR #4636, opened 2018-12-04
Issue2963 needs to just have it's UI test removed since it needs to be manually verified 

### LINKED ISSUE #2312: Issue2963 needs to just have it's UI test removed since it needs to be manually verified
The action of clicking on the text box and then typing text can't be automated the UI Test

AndreiMisiukevich:
### Description of Change ###

Removed UI test as was requested in original ticket

### Issues Resolved ### 

- fixes #2312


PureWeen:
Looking at the UI Test again I feel like I missed the behavior the UI Test is providing with checking the IsFocused property. It clicks the editor and then ensures that Focused doesn't change.  I looked to see if there's another test that covers this case but didn't find one so we'll keep this one around.

Thanks for the commit but my statement in that issue that it's all just manual was incorrect

PR #4641, opened 2018-12-05
[XamlC] Compile the {DataTemplate} markup

### LINKED ISSUE #4592: [Shell] compile the DataTemplateExtension
the `{DataTemplate}` markup extension is not compiled by XamlC, and could be to:
- catch exception at compile time
- speed up runtime

StephaneDelcroix:
### Description of Change ###

[XamlC] Compile the {DataTemplate} markup

### Issues Resolved ### 

- fixes #4592

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Build.Tasks/CompiledMarkupExtensions/DataTemplateExtension.cs
+++ b/Xamarin.Forms.Build.Tasks/CompiledMarkupExtensions/DataTemplateExtension.cs
@@ -1,0 +1,39 @@
+﻿using System.Linq;
+using System.Xml;
+using System.Collections.Generic;
+
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+
+using Xamarin.Forms.Xaml;
+
+using static Mono.Cecil.Cil.Instruction;
+using static Mono.Cecil.Cil.OpCodes;
+
+namespace Xamarin.Forms.Build.Tasks
+{
+	class DataTemplateExtension : ICompiledMarkupExtension
+	{
+		public IEnumerable<Instruction> ProvideValue(IElementNode node, ModuleDefinition module, ILContext context, out TypeReference typeRef)
+		{
+			typeRef = module.ImportReference(("Xamarin.Forms.Core", "Xamarin.Forms", "DataTemplate"));
+			var name = new XmlName("", "TypeName");
+
+			if (!node.Properties.TryGetValue(name, out INode typeNameNode) && node.CollectionItems.Any())
+				typeNameNode = node.CollectionItems[0];
+
+			if (!(typeNameNode is ValueNode valueNode))
+				throw new XamlParseException("TypeName isn't set.", node as XmlLineInfo);
+
+			var contentTypeRef = module.ImportReference(XmlTypeExtensions.GetTypeReference(valueNode.Value as string, module, node as BaseNode))
+				?? throw new XamlParseException($"Can't resolve type `{valueNode.Value}'.", node as IXmlLineInfo);
+
+			var dataTemplateCtor = module.ImportCtorReference(typeRef, new[] { module.ImportReference(("mscorlib", "System", "Type")) });
+			return new List<Instruction> {
+				Create(Ldtoken, module.ImportReference(contentTypeRef)),
+				Create(Call, module.ImportMethodReference(("mscorlib", "System", "Type"), methodName: "GetTypeFromHandle", parameterTypes: new[] { ("mscorlib", "System", "RuntimeTypeHandle") }, isStatic: true)),
+				Create(Newobj, dataTemplateCtor),
+			};
+		}
+	}
+}
--- a/Xamarin.Forms.Build.Tasks/CompiledMarkupExtensions/TypeExtension.cs
+++ b/Xamarin.Forms.Build.Tasks/CompiledMarkupExtensions/TypeExtension.cs
@@ -10,7 +10,6 @@ using Xamarin.Forms.Xaml;
 using static Mono.Cecil.Cil.Instruction;
 using static Mono.Cecil.Cil.OpCodes;
 
-
 namespace Xamarin.Forms.Build.Tasks
 {
 	class TypeExtension : ICompiledMarkupExtension
@@ -18,14 +17,12 @@ namespace Xamarin.Forms.Build.Tasks
 		public IEnumerable<Instruction> ProvideValue(IElementNode node, ModuleDefinition module, ILContext context, out TypeReference memberRef)
 		{
 			memberRef = module.ImportReference(("mscorlib", "System", "Type"));
-			INode typeNameNode;
-
 			var name = new XmlName("", "TypeName");
-			if (!node.Properties.TryGetValue(name, out typeNameNode) && node.CollectionItems.Any())
+
+			if (!node.Properties.TryGetValue(name, out INode typeNameNode) && node.CollectionItems.Any())
 				typeNameNode = node.CollectionItems[0];
 
-			var valueNode = typeNameNode as ValueNode;
-			if (valueNode == null)
+			if (!(typeNameNode is ValueNode valueNode))
 				throw new XamlParseException("TypeName isn't set.", node as XmlLineInfo);
 
 			if (!node.Properties.ContainsKey(name)) {
--- a/Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
+++ b/Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
@@ -1,0 +1,33 @@
+﻿using NUnit.Framework;
+
+using Xamarin.Forms.Core.UnitTests;
+
+namespace Xamarin.Forms.Xaml.UnitTests
+{
+	public partial class DataTemplateExtension : ContentPage
+	{
+		public DataTemplateExtension() => InitializeComponent();
+		public DataTemplateExtension(bool useCompiledXaml)
+		{
+			//this stub will be replaced at compile time
+		}
+
+		[TestFixture]
+		class Tests
+		{
+			[SetUp]
+			public void Setup() => Device.PlatformServices = new MockPlatformServices();
+			[TearDown] public void TearDown() => Device.PlatformServices = null;
+
+			[TestCase(true), TestCase(false)]
+			public void DataTemplateExtension(bool useCompiledXaml)
+			{
+				var layout = new DataTemplateExtension(useCompiledXaml);
+				var content = layout.Resources["content"] as ShellContent;
+				var template = content.ContentTemplate;
+				var obj = template.CreateContent();
+				Assert.That(obj, Is.TypeOf<DataTemplateExtension>());
+			}
+		}
+	}
+}
--- a/Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
+++ b/Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
@@ -15,8 +15,7 @@ namespace Xamarin.Forms.Xaml.UnitTests
 		[TestFixture]
 		class Tests
 		{
-			[SetUp]
-			public void Setup() => Device.PlatformServices = new MockPlatformServices();
+			[SetUp] public void Setup() => Device.PlatformServices = new MockPlatformServices();
 			[TearDown] public void TearDown() => Device.PlatformServices = null;
 
 			[TestCase(true), TestCase(false)]
--- a/Xamarin.Forms.Xaml/MarkupExtensions/DataTemplateExtension.cs
+++ b/Xamarin.Forms.Xaml/MarkupExtensions/DataTemplateExtension.cs
@@ -2,17 +2,26 @@
 
 namespace Xamarin.Forms.Xaml
 {
-	[ContentProperty("Type")]
+	[ContentProperty("TypeName")]
+	[ProvideCompiled("Xamarin.Forms.Build.Tasks.DataTemplateExtension")]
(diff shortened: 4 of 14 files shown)


Xamarin.Forms.Build.Tasks/CompiledMarkupExtensions/TypeExtension.cs
```diff
@@ -10,22 +10,19 @@
...
 
-
 namespace Xamarin.Forms.Build.Tasks
 {
 	class TypeExtension : ICompiledMarkupExtension
 	{
 		public IEnumerable<Instruction> ProvideValue(IElementNode node, ModuleDefinition module, ILContext context, out TypeReference memberRef)
 		{
 			memberRef = module.ImportReference(("mscorlib", "System", "Type"));
-			INode typeNameNode;
```
StephaneDelcroix: refactored the `TypeExtension` while reading through, as both show similarities

Xamarin.Forms.Xaml/MarkupExtensions/DataTemplateExtension.cs
```diff
@@ -2,17 +2,26 @@
 
 namespace Xamarin.Forms.Xaml
 {
-	[ContentProperty("Type")]
+	[ContentProperty("TypeName")]
+	[ProvideCompiled("Xamarin.Forms.Build.Tasks.DataTemplateExtension")]
 	public sealed class DataTemplateExtension : IMarkupExtension<DataTemplate>
 	{
-		public string Type { get; set; }
+		public string TypeName { get; set; }
```
StephaneDelcroix: that's the API-Change

Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
```diff
@@ -0,0 +1,33 @@
...
+	public partial class DataTemplateExtension : ContentPage
+	{
+		public DataTemplateExtension() => InitializeComponent();
+		public DataTemplateExtension(bool useCompiledXaml)
+		{
+			//this stub will be replaced at compile time
+		}
+
+		[TestFixture]
+		class Tests
```
mattleibow: Is this something we like? Nested test classes?

Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
```diff
@@ -0,0 +1,33 @@
...
+		{
+			[SetUp]
+			public void Setup() => Device.PlatformServices = new MockPlatformServices();
+			[TearDown] public void TearDown() => Device.PlatformServices = null;
```
mattleibow: It might be better to be consistent with attributes on a separate line.

Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
```diff
@@ -0,0 +1,33 @@
...
+
+			[TestCase(true), TestCase(false)]
```
mattleibow: It might be better to indicate multiple test with separate lines for each test case.

Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
```diff
@@ -0,0 +1,33 @@
...
+using Xamarin.Forms.Core.UnitTests;
+
+namespace Xamarin.Forms.Xaml.UnitTests
+{
+	public partial class DataTemplateExtension : ContentPage
+	{
+		public DataTemplateExtension() => InitializeComponent();
+		public DataTemplateExtension(bool useCompiledXaml)
+		{
+			//this stub will be replaced at compile time
```
mattleibow: I am not sure I understand this comment... Does this compilation actually modify the type?

Xamarin.Forms.Xaml/MarkupExtensions/DataTemplateExtension.cs
mattleibow: Why did we change this? It looks like we are talking about the type... Although the other extensions are using TypeName it appears... So I am assuming that?

Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
```diff
@@ -0,0 +1,33 @@
...
+
+		[TestFixture]
+		class Tests
+		{
+			[SetUp]
+			public void Setup() => Device.PlatformServices = new MockPlatformServices();
+			[TearDown] public void TearDown() => Device.PlatformServices = null;
+
+			[TestCase(true), TestCase(false)]
+			public void DataTemplateExtension(bool useCompiledXaml)
```
mattleibow: How does this test whether the markup extension was the compile one or the dynamic one?

Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
```diff
@@ -0,0 +1,33 @@
+﻿using NUnit.Framework;
+
+using Xamarin.Forms.Core.UnitTests;
+
+namespace Xamarin.Forms.Xaml.UnitTests
+{
+	public partial class DataTemplateExtension : ContentPage
```
mattleibow: This is just me, but there are too many things called `DataTemplateExtension` here... both the library types, the page, the test.

Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
StephaneDelcroix: 1/ it's something I like
2/ it's the most convenient way in this type of cases, so we can access `layout.label` which is a private field created by x:Name
3/ all xaml unit tests uses the same schema
4/ and now, for longer answer. in the distant pasts I can recall working professionally with unit tests (20 years ago, java environment), unit tests were always placed in nested classes of the class being tested. That allowed to make real unit tests for the private part of the class. Years passed, and we tend to move unit tests in their own assemblies. And those new kind of unit tests no longer test the way a class works, but they test more the public API of it. I'm not nostalgic here, but we lost some finer grained testing in the transition. And we're using `internal` and `InternalsVisibleTo` when we want that agility back...

Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
StephaneDelcroix: I agree for consistency. this is boilerplate, and i like the boilerplate to
1. take as few lines as possible
2. be of a recognizable shape

Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
StephaneDelcroix: when you can enumerate all the values of a bool in 2 conditions, I read this as "a single test, for all the possible values of bool" vs multiple tests

Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
StephaneDelcroix: OH YEAH

We have a special task that takes this assembly and replace the content of this ctor with a conditional that either goes to InitializeComponent (which was itself replaced by the XamlC task) or a call to inflate the xaml dynamically at runtime.

Again, this is boilerplate, and you can ignore what it does or the way it does it, as it's impossible to have a mental model of all the abstractions going on with the Xaml Compiler...

Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
StephaneDelcroix: too many, but no confusion

Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
StephaneDelcroix: this assumes that, in the case when `useCompiledXaml` is `true`, the `ctor` in
```csharp
var layout = new DataTemplateExtension(useCompiledXaml);
```
will return the compiled version of the Xaml.

And this test that the compiled version behave as it should.

There's no easy way to test that the markup as actually been compiled, as it has no side effects, and leave lo artifacts. But let me show you that it actually works:

```il
	IL_005a:  nop 
	IL_005b:  ldtoken Xamarin.Forms.Xaml.UnitTests.DataTemplateExtension
	IL_0060:  call class [mscorlib]System.Type class [mscorlib]System.Type::GetTypeFromHandle(valuetype [mscorlib]System.RuntimeTypeHandle)
	IL_0065:  newobj instance void class [Xamarin.Forms.Core]Xamarin.Forms.DataTemplate::'.ctor'(class [mscorlib]System.Type)
	IL_006a:  stloc.0 
```

and then, further down the `il`

```il
	IL_0080:  ldloc.1 
	IL_0081:  ldsfld class [Xamarin.Forms.Core]Xamarin.Forms.BindableProperty [Xamarin.Forms.Core]Xamarin.Forms.ShellContent::ContentTemplateProperty
	IL_0086:  ldloc.0 
	IL_0087:  callvirt instance void class [Xamarin.Forms.Core]Xamarin.Forms.BindableObject::SetValue(class [Xamarin.Forms.Core]Xamarin.Forms.BindableProperty, object)
```

(local `V_1` is the `ShellContent` instance)

so this does shows that the markup extension is not evaluated at runtime, but actually compiled to the equiv code of
```csharp
new DataTemplate(typeof(Foo));
```

Xamarin.Forms.Xaml/MarkupExtensions/DataTemplateExtension.cs
StephaneDelcroix: 1. to be aligned with the `{x:Type}` markup
2. to avoid confusion, and require a type name, not an actual type, so this is not valid: `{DataTemplate {Type local:foo}}`

Xamarin.Forms.Xaml.UnitTests/DataTemplateExtension.xaml.cs
StephaneDelcroix: fixed

Xamarin.Forms.Xaml/MarkupExtensions/TypeExtension.cs
```diff
@@ -10,13 +10,14 @@ public class TypeExtension : IMarkupExtension<Type>
...
-			if (string.IsNullOrEmpty(TypeName))
-				throw new InvalidOperationException("TypeName isn't set.");
 			if (serviceProvider == null)
 				throw new ArgumentNullException(nameof(serviceProvider));
-			var typeResolver = serviceProvider.GetService(typeof (IXamlTypeResolver)) as IXamlTypeResolver;
-			if (typeResolver == null)
+			if (!(serviceProvider.GetService(typeof(IXamlTypeResolver)) is IXamlTypeResolver typeResolver))
 				throw new ArgumentException("No IXamlTypeResolver in IServiceProvider");
+			if (string.IsNullOrEmpty(TypeName)) {
+				var li = (serviceProvider.GetService(typeof(IXmlLineInfoProvider)) is IXmlLineInfoProvider lip) ? lip.XmlLineInfo : new XmlLineInfo();
```
paymicro: This feature repeats 3 times in this PR (6 times in Xaml project). Perhaps to allocate a separate static method?
```C#
XmlLineInfo GetLineInfo(IServiceProvider serviceProvider) 
=> (serviceProvider.GetService(typeof(IXmlLineInfoProvider)) is IXmlLineInfoProvider lip) ? lip.XmlLineInfo : new XmlLineInfo()
```

StephaneDelcroix: rationalizing that usage is part of #4582.

Xamarin.Forms.Build.Tasks/CompiledMarkupExtensions/DataTemplateExtension.cs
```diff
@@ -0,0 +1,39 @@
...
+using static Mono.Cecil.Cil.OpCodes;
+
+namespace Xamarin.Forms.Build.Tasks
+{
+	class DataTemplateExtension : ICompiledMarkupExtension
+	{
+		public IEnumerable<Instruction> ProvideValue(IElementNode node, ModuleDefinition module, ILContext context, out TypeReference typeRef)
+		{
+			typeRef = module.ImportReference(("Xamarin.Forms.Core", "Xamarin.Forms", "DataTemplate"));
+			var name = new XmlName("", "TypeName");
```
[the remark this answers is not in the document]
StephaneDelcroix: y, maybe

Xamarin.Forms.Build.Tasks/CompiledMarkupExtensions/DataTemplateExtension.cs
```diff
@@ -0,0 +1,39 @@
...
+using Mono.Cecil.Cil;
+
+using Xamarin.Forms.Xaml;
+
+using static Mono.Cecil.Cil.Instruction;
+using static Mono.Cecil.Cil.OpCodes;
+
+namespace Xamarin.Forms.Build.Tasks
+{
+	class DataTemplateExtension : ICompiledMarkupExtension
```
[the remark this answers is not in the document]
StephaneDelcroix: no, it's not, as explained to @mattleibow

PR #4569, opened 2018-11-28
[Android] ItemContentView should use all available space

hartez:
### Description of Change ###

`ItemContentView` should take up the available space given to it by the layout manager.
Also fixes some inconsistent/inaccurate class naming for cells/views.

### Issues Resolved ### 

- fixes #4406

### API Changes ###

 None

### Platforms Affected ### 

- Android

### Behavioral/Visual Changes ###

CollectionView items on Android now layout like they do on iOS.

### Before/After Screenshots ### 

Before: 
screenshot_1543363225

After:
screenshot_1543422703

### Testing Procedure ###

Navigate to Control Gallery -> CollectionView Gallery -> DataTemplate Galleries -> Vertical List (Code); if the items are centered, the test has passed. If they are left-aligned, the test has failed.

### PR Checklist ###

- [ ] Has automated tests 
- [x] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard

PR #4664, opened 2018-12-06
[CollectionView] Fixes for ObservableCollection usage on Android and iOS

### LINKED ISSUE #4600: [iOS] CollectionView crash with empty ObservableCollection
### Description

If the ObservableCollection is empty it will crash on iOS

### Steps to Reproduce

1. Try run attached sample

### Expected Behavior

Should not crash, when items are added I should see them appear on the screen

### Basic Information

- Version with issue: 4.0.0-pre1
- Platform Target Frameworks: 
  - iOS:  

### Screenshots

screenshot 2018-11-30 at 23 51 09

### Reproduction Link

[CollectionViewSample.zip](https://github.com/xamarin/Xamarin.Forms/files/2635336/CollectionViewSample.zip)

rmarinho:
### Description of Change ###

On Android and iOS navigating from a page that had a ObservableCollection as source was causing a crash because the iNotifyPropertyChanged event wasn't unsubscribed. Implemented Disposable on IItemsViewSource, (this was also in the TODO list from @hartez ). 

### API Changes ###

Changed:
 - internal interface IItemsViewSource - > internal interface IItemsViewSource : IDisposable
Added 

### Platforms Affected ### 

- Core/XAML (all platforms)
- Android

Should add items to the collection view , and allow to navigate back and to another gallery

### Testing Procedure ###

Visit ObservableCodeCollectionViewGallery with empty source.

### PR Checklist ###

- [ ] Has automated tests 


Xamarin.Forms.Platform.Android/CollectionView/ItemsViewRenderer.cs
```diff
@@ -295,8 +295,8 @@ void TearDownOldElement(ItemsView oldElement)
 
 			if (adapter != null)
 			{
-				adapter.Dispose();
```
rmarinho: We need to un hood our adapter before dispose it.

Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewLayout.cs
```diff
@@ -120,13 +122,18 @@ public override bool ShouldInvalidateLayoutForBoundsChange(CGRect newBounds)
...
 
-		protected void DetermineCellSize()
+		internal protected void DetermineCellSize()
 		{
 			if (GetPrototype == null)
 			{
 				return;
 			}
 
+			if (!(GetPrototype() is ItemsViewCell prototype))
```
rmarinho: At this time we could not have a prototype because we don't have cells.

Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs
```diff
@@ -10,17 +10,21 @@ internal class ObservableItemsSource : IItemsViewSource
...
 			_collectionView = collectionView;
 			_itemsSource = (IList)itemSource;
 
 			((INotifyCollectionChanged)itemSource).CollectionChanged += CollectionChanged;
+			_isEmpty = !_itemsSource.GetEnumerator().MoveNext();
 		}
 
 		void CollectionChanged(object sender, NotifyCollectionChangedEventArgs args)
 		{
+			ShouldDetermineCellSize();
```
rmarinho: We are adding our first cell, the collection was empty and now we will need to DetermineCellSize

Xamarin.Forms.Platform.iOS/CollectionView/CollectionViewController.cs
```diff
@@ -42,7 +42,14 @@ public override UICollectionViewCell GetCell(UICollectionView collectionView, NS
 
 		public override nint GetItemsCount(UICollectionView collectionView, nint section)
 		{
-			return _itemsSource.Count;
+			var count = _itemsSource.Count;
+			if (_layout.IsDeterminingCellSize && count == 1)
```
rmarinho: I would like a better idea for this I don't have one right now. 
So basically when you set EstimateItemSize on `DetermineCellSize` It will queue a call to GetItemsCount, since the source already has 1 item, but we called this method before adding a NativeCell to the UICollectionView, so to return the correct count we need to return 0.

mattleibow: I may not be understanding this, but what happens if there are 2 items in the source? Why do we check specifically for `1` item? What if there is no items? Does this mean the size might be wrong? If the size is going to be incorrect, can we postpone the measuring until we have an item?

Xamarin.Forms.Platform.iOS/CollectionView/CollectionViewController.cs
```diff
@@ -42,7 +42,14 @@ public override UICollectionViewCell GetCell(UICollectionView collectionView, NS
...
+			{
+				_layout.FinishedDeterminingCellSize();
```
mattleibow: There appears to be a disconnect here and the layout. It is `FinishedDeterminingCellSize` here, but `FinishDetermineCellSize` in the layout. Maybe this?

```suggestion
				_layout.FinishDeterminingCellSize();
```

Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewLayout.cs
```diff
@@ -155,7 +157,10 @@ protected void DetermineCellSize()
 			{
 				EstimatedItemSize = measure;
 			}
+		}
 
+		internal void FinishDetermineCellSize()
+		{
```
mattleibow: Here we have split the size calculation into 2 methods, but we never seem to call this. There is a possible connection and type on `FinishedDeterminingCellSize` in `CollectionViewController.GetItemsCount`, but this appears to be for just 1 type. How are these two methods supposed to be used?

If there are 2 methods that need to be called, do we have a way to test that both are bing called? For me, the split gets scary as we now have forced a second method that can be missed. And by looking at the current code, it will only ever run on iOS when there is a SINGLE item in the source.

Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs
```diff
@@ -10,17 +10,21 @@ internal class ObservableItemsSource : IItemsViewSource
 	{
 		readonly UICollectionView _collectionView;
 		readonly IList _itemsSource;
+		bool _isEmpty;
+
 
 		public ObservableItemsSource(IEnumerable itemSource, UICollectionView collectionView)
 		{
 			_collectionView = collectionView;
 			_itemsSource = (IList)itemSource;
```
mattleibow: Is this safe? I just noticed that we just cast. It might be better to change the constructor parameter to `IList` - this way we move the responsibility to the caller.

Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs
```diff
@@ -10,17 +10,21 @@ internal class ObservableItemsSource : IItemsViewSource
...
 
 			((INotifyCollectionChanged)itemSource).CollectionChanged += CollectionChanged;
+			_isEmpty = !_itemsSource.GetEnumerator().MoveNext();
```
mattleibow: If this really is an `IList`, cant we rather do `Count`? This should be better.

Xamarin.Forms.Platform.Android/CollectionView/ItemsViewAdapter.cs
```diff
@@ -16,9 +17,10 @@ namespace Xamarin.Forms.Platform.Android
 	public class ItemsViewAdapter : RecyclerView.Adapter
 	{
 		protected readonly ItemsView ItemsView;
-		readonly Context _context;
-		readonly Func<IVisualElementRenderer, Context, AView> _createView;
-		readonly IItemsViewSource _itemsSource;
+		Context _context;
```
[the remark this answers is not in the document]
hartez: See https://github.com/xamarin/Xamarin.Forms/pull/4569#pullrequestreview-179602894

Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewLayout.cs
rmarinho: Fixed the call, it should be always called, but you are correct that a better way will be preferred. Need to see what @hartez thinks about it.

Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs
rmarinho: Done

Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs
```diff
@@ -11,10 +11,10 @@ internal class ObservableItemsSource : IItemsViewSource
 		readonly UICollectionView _collectionView;
 		readonly IList _itemsSource;
 
-		public ObservableItemsSource(IEnumerable itemSource, UICollectionView collectionView)
+		public ObservableItemsSource(IList itemSource, UICollectionView collectionView)
```
hartez: Why was this changed to IList?

Xamarin.Forms.Platform.Android/CollectionView/ObservableItemsSource.cs
```diff
@@ -23,6 +22,28 @@ public ObservableItemsSource(IEnumerable itemSource, RecyclerView.Adapter adapte
...
+
+		protected virtual void Dispose(bool disposing)
+		{
+			if (!_disposed)
+			{
+				if (disposing)
+				{
+					((INotifyCollectionChanged)_itemsSource).CollectionChanged -= CollectionChanged;
+					_itemsSource = null;
+					_adapter = null;
```
hartez: No point in setting `_adapter` and `_itemsSource` to `null` here, this instance is going away anyway.

Xamarin.Forms.Platform.Android/CollectionView/ObservableItemsSource.cs
```diff
@@ -7,9 +7,8 @@ namespace Xamarin.Forms.Platform.Android
 {
 	internal class ObservableItemsSource : IItemsViewSource
 	{
-		// TODO hartez 2018/07/30 14:40:11 We may need to implement IDisposable to make sure this all gets cleaned up	
-		readonly RecyclerView.Adapter _adapter;
-		readonly IList _itemsSource;
+		RecyclerView.Adapter _adapter;
```
hartez: Since we don't need to `null` these out, they don't actually change and we can leave these as `readonly`.

Xamarin.Forms.Platform.Android/CollectionView/ListSource.cs
```diff
@@ -21,5 +23,24 @@ public ListSource(IEnumerable enumerable)
 				Add(item);
 			}
 		}
+
+		public void Dispose()
```
hartez: Since we aren't actually inheriting from this anywhere right now, we could just marked it `sealed` and then we don't have to implement the `protected virtual void Dispose(bool disposing)` pattern at all.

Xamarin.Forms.Platform.Android/CollectionView/ItemsViewAdapter.cs
```diff
@@ -66,6 +65,22 @@ public override RecyclerView.ViewHolder OnCreateViewHolder(ViewGroup parent, int
...
 		}
 
+		protected override void Dispose(bool disposing)
+		{
+			if (!_disposed)
+			{
+				if (disposing)
+				{
+					_itemsSource?.Dispose();
+					_itemsSource = null;
```
hartez: Setting `_itemsSource` and `_createView` to null doesn't do anything here, these can be removed.

Xamarin.Forms.Platform.Android/CollectionView/ItemsViewAdapter.cs
```diff
@@ -1,23 +1,22 @@
...
+	// TODO hartez 2018/07/25 14:39:29 Split up CollectionViewAdapter into one for templates, one for text
+	// TODO hartez 2018/07/25 14:43:04 Experiment with an ItemSource property change as _adapter.notifyDataSetChanged
 	// TODO hartez 2018/07/25 14:44:15 Template property changed should do a whole new adapter; and that way we can cache the template
 
 	public class ItemsViewAdapter : RecyclerView.Adapter
 	{
 		protected readonly ItemsView ItemsView;
-		readonly Func<IVisualElementRenderer, Context, AView> _createView;
-		readonly IItemsViewSource _itemsSource;
+		Func<IVisualElementRenderer, Context, AView> _createView;
```
hartez: Since these don't need to be modified (see comments below), the `readonly` can be reinstated.

Xamarin.Forms.Platform.Android/CollectionView/ItemsViewAdapter.cs
```diff
@@ -66,6 +65,22 @@ public override RecyclerView.ViewHolder OnCreateViewHolder(ViewGroup parent, int
 			return new TemplatedItemViewHolder(itemContentControl, templateElement);
 		}
 
+		protected override void Dispose(bool disposing)
+		{
+			if (!_disposed)
```
hartez: If this is already disposed, it shouldn't call `base.Dispose()` again.

Xamarin.Forms.Platform.Android/CollectionView/EmptySource.cs
```diff
@@ -7,5 +7,25 @@ internal class EmptySource : IItemsViewSource
...
 		public object this[int index] => throw new IndexOutOfRangeException("IItemsViewSource is empty");
+
+		public void Dispose()
+		{
+			Dispose(true);
+		}
+
+		bool _disposed;
+
+		protected virtual void Dispose(bool disposing)
```
hartez: We can just mark this as `sealed` and then we don't have to implement the `protected virtual void Dispose(bool disposing)` pattern.

Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs
mattleibow: This was changed to `IList` because this source only understands lists, not any arb collection type. For example:
`public object this[int index]` and `public int Count`. What was happening was the constructor was just doing a straight cast, but this could cause issues in the future. Since this was just for observable collections, they are `IList`. In addition the factory has it defined as: `case IList _ when itemsSource is INotifyCollectionChanged:` - basically `IList` again.

I foresee evil where Bob comes along and says: "ooh, look. I can wrap my queue with this and magic will happen." and the only magic is a major 💥

rmarinho:
build

Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs
hartez: Well, it's `internal`, so Bob would have to slip it by us, but okay. I'm cool with changing it to `IList`, but for consistency it should also be changed in the Android version.

samhouts:
@rmarinho Please remove the do-not-merge when you've updated the PR. Thanks!

rmarinho:
I rebased and reworked a little bit this issue, I also saw the same issue using the sample I added with the timer "Add Items with timer to Empty Collection" on iOS and added the fix.

PR #4672, opened 2018-12-07
[build] add a Directory.Build.props

StephaneDelcroix:
### Description of Change ###

Copying `<PropertyGroup>`s on (almost) every .csproj, and get the
relative paths just right is tedious. Starting msbuild 15 (vs2017)
we can use a Directory.Build.props (and Directory.Build.targets)
so I see no reason why we shouldn't.

see https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2017

### Issues Resolved ### 

/

### API Changes ###

 
 None

### Platforms Affected ### 

/

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,7 @@
+<Project>
+  <PropertyGroup>
+    <__XFBuildTasksLocation>$(_XFBuildTasksLocation)</__XFBuildTasksLocation>
+    <__XFBuildTasksLocation Condition="'$(__XFBuildTasksLocation)' == '' AND '$(MSBuildRuntimeType)' == 'Core'">.nuspec\netstandard2.0\</__XFBuildTasksLocation>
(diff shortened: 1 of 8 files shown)


gabor-nemeth:
Is it just me who can't build the modified projects anyomre? Directory.Build.props has relative paths to its folder, and those are not valid from the projects' perspective.

Applying the path to `$(SolutionDir)` can be one of the possible fixes.
`<__XFBuildTasksLocation Condition="'$(__XFBuildTasksLocation)' == ''">$(SolutionDir)\.nuspec\net461\</__XFBuildTasksLocation>
  </PropertyGroup>`

StephaneDelcroix:
@gabor-nemeth the patch is supposed to be, after merges, `.nuspec/net46/` but if you have any problem, please report them

PR #4688, opened 2018-12-08
[iOS] Update cell size estimates after empty data source

### LINKED ISSUE #4600: [iOS] CollectionView crash with empty ObservableCollection

hartez:
### Description of Change ###

Normally the CollectionView on iOS uses the first available cell as a prototype to determine the estimated size (or exact size) of subsequent cells. But if there is no first available cell when the CollectionView first appears, the estimates remain zero and new cells are added, but are not visible. 

These changes allow the CollectionView to use the first available cell even if there is no data when the CollectionView first appears.

### Issues Resolved ### 

- fixes #4600

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###

Added an automated test; for manual testing, in Control Gallery navigate to CollectionView Gallery -> Observable Collection Galleries -> Add/Remove Items (Grid, Initially Empty) and click the Insert button; an item should appear. Click the button again; another item should appear, and both items should be the same size.

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4600.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4600.cs
@@ -1,0 +1,38 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.Forms.Core.UITests;
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if UITEST
+	[Category(UITestCategories.ManualReview)]
+	[Explicit] // Marked explicit until we have a lane for CollectionView tests
+#endif
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 4600, "[iOS] CollectionView crash with empty ObservableCollection", PlatformAffected.iOS)]
+	public class Issue4600 : TestNavigationPage
+	{
+		protected override void Init()
+		{
+#if APP
+			PushAsync(
+				new GalleryPages.CollectionViewGalleries.ObservableCodeCollectionViewGallery(initialItems: 0));
+#endif
+		}
+
+#if UITEST
+		[Test]
+		public void InitiallyEmptySourceDisplaysAddedItem()
+		{
+			RunningApp.WaitForElement("Insert");
+			RunningApp.Tap("Insert");
+			RunningApp.WaitForElement("Inserted");
+		}
+#endif
+	}
+}
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4600.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4600.cs
@@ -11,7 +11,7 @@ namespace Xamarin.Forms.Controls.Issues
 {
 #if UITEST
 	[Category(UITestCategories.ManualReview)]
-	[Explicit] // Marked explicit until we have a lane for CollectionView tests
+	[Ignore("Ignoring until we have a lane to run CollectionView test (or CV is not behind a flag")] 
 #endif
 	[Preserve(AllMembers = true)]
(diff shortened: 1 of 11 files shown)


Xamarin.Forms.Platform.iOS/CollectionView/CollectionViewController.cs
```diff
@@ -12,6 +12,7 @@ internal class CollectionViewController : UICollectionViewController
 		readonly ItemsView _itemsView;
 		readonly ItemsViewLayout _layout;
 		bool _initialConstraintsSet;
+		int _previousCount = -1;
```
StephaneDelcroix: do we need to reset this to -1 when the collection is cleared ?

Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewLayout.cs
```diff
@@ -67,24 +66,32 @@ protected virtual void HandlePropertyChanged(PropertyChangedEventArgs  propertyC
...
+			if (_needCellSizeUpdate)
+			{
+				// Our cell size/estimate is out of date, probably because we moved from zero to one item; update it
+				_needCellSizeUpdate = false;
+				DetermineCellSize();
+			}
+		}
+
 		[Export("collectionView:layout:insetForSectionAtIndex:")]
-		[CompilerGenerated]
```
StephaneDelcroix: there's a history behind those... probably a way to make sure r# wasn't raising warnings...

hartez: Doesn't seem to bother R#er to leave them out. At least not with the latest version.

Xamarin.Forms.Platform.iOS/CollectionView/CollectionViewController.cs
hartez: No. The -1 is only important on the very first run of `GetItemsCount`, when we are first setting up the CollectionView. We don't want the size update to run at that point because it will already be running as a result of the bounds of the CollectionView changing (to whatever the initial size is). After that, we can rely on the actual count.

mattleibow: Could we not change this to a bool? like `hasRunCellLayout`. Then, in the `GetItemsCount` method, we check `if (!hasRunCellLayout && count > 0) hasRunCellLayout = true; ...`

Just like in `ItemsViewLayout` with `_needCellSizeUpdate`.

Xamarin.Forms.Platform.iOS/CollectionView/TemplatedVerticalListCell.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
mattleibow: This is probably causing the conflict :sob:

samhouts:
build --uitests

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4600.cs
```diff
@@ -0,0 +1,38 @@
...
+using Xamarin.Forms.Core.UITests;
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if UITEST
+	[Category(UITestCategories.ManualReview)]
+	[Explicit] // Marked explicit until we have a lane for CollectionView tests
```
rmarinho: this should keep the test of running on AppCenter right? it isn't .. it's failing the test.

hartez: It should have, but it seems UITests is doing something weird. Changed it to Ignore, we'll hope that works.

mattleibow:
build --uitests

PR #4699, opened 2018-12-11
Properly detect SDK style projects

kzu:
The `$(MSBuildSDKsPath)` is actually *always* set whenever there are MSBuild SDKs 
*installed*, not necessarily when the current project is an SDK-style project itself. 

This means that we're potentially adding duplicate items and otherwise disrupting 
normal operation of the non-SDK style projects with this import. 

So use the well-known properties that are set by the .NET [Sdk.props](https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.props#L22-L27) file instead, 
which is the first imported target even before nuget-provided .props files.

### DIFF
--- a/.nuspec/Xamarin.Forms.props
+++ b/.nuspec/Xamarin.Forms.props
@@ -1,5 +1,10 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+	<PropertyGroup>
+		<!-- Sdk.props imports Microsoft.NET.Sdk.props which defines Configurations and Platforms, which is the SDK-style way -->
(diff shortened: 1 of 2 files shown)


kzu:
To be squashed... I wish GitHub allowed editing more than one file per commit from the website ;)

mhutch:
Kinda wonder whether conditioning it on `EnableDefaultItems` might express intent better. "sdk-style projects" is a loosely defined grab-bag of things.

StephaneDelcroix:
@kzu 
> To be squashed
we always squash

could you rebase and target this to 3.5.0 ? thanks

kzu:
Maybe @mhutch was right and doing the condition on `$(EnableDefaultItems)` would have been better. @StephaneDelcroix

StephaneDelcroix:
I reverted this. @kzu @mhutch please come up with better solutions for this. thanks

PR #4704, opened 2018-12-11
[WIP] Updating the Material Visual to v72.1

mattleibow:
Requires: https://jenkins.mono-project.com/job/Components-OpenSource-PR/56/Azure/

PR #4705, opened 2018-12-11
[WIP] Updating the Material Visual to v72.2.0 + Alerts

mattleibow:
### Description

The main point of this PR is to start a discussion on how to style non-element things. A button is fine with a `Visual` property. Different ones on a page will look weird, but still just an appearance.

Popups, alerts, etc will look bad if each button pops up a different style depending on which page or which control launched it.

**Options**
 - ~~new alert overload for material~~ (pre-crossed as this is evil, pure evil)
 - page-based visual: material pages show material popups
    - page passes current visual to alert arguments
    - alert implementation looks up page visual
 - app-based/global visual: the app decides on the style
    - alert implementation looks up visual
 - more...

Reopened: https://github.com/xamarin/Xamarin.Forms/pull/4704


Xamarin.Forms.Platform.iOS/Platform.cs
```diff
@@ -318,6 +318,38 @@ void PresentAlert(AlertArguments arguments)
...
+				AlertControllerThemer.ApplyScheme(scheme, malert);
+
+				window.RootViewController = new UIViewController();
+				window.RootViewController.View.BackgroundColor = Color.Transparent.ToUIColor();
+				window.WindowLevel = UIWindowLevel.Alert + 1;
+				window.MakeKeyAndVisible();
+				window.RootViewController.PresentViewController(malert, true, null);
+
+				return;
+			}
```
mattleibow: This is just for material alerts on iOS, and is a hack right now. Do not ever write code like this!

Xamarin.Forms.Core/AlertArguments.cs
```diff
@@ -6,13 +6,14 @@ namespace Xamarin.Forms.Internals
 	[EditorBrowsable(EditorBrowsableState.Never)]
 	public class AlertArguments
 	{
-		public AlertArguments(string title, string message, string accept, string cancel)
+		public AlertArguments(string title, string message, string accept, string cancel, IVisual visual = null)
```
mattleibow: I am not sure if we want to do something like this here... we may actually want to query the page at the point we show the alert. I still need to see if that is possible.

Xamarin.Forms.Core/Page.cs
```diff
@@ -146,7 +146,7 @@ public Task<bool> DisplayAlert(string title, string message, string accept, stri
 			if (string.IsNullOrEmpty(cancel))
 				throw new ArgumentNullException("cancel");
 
-			var args = new AlertArguments(title, message, accept, cancel);
+			var args = new AlertArguments(title, message, accept, cancel, Visual);
```
mattleibow: I don't think this is good here, visual should be a "global" thing, not per request. This also lends itself to having different popups on different pages - which I believe is a very bad thing.

PureWeen:
https://github.com/xamarin/XamarinComponents/pull/495

Xamarin.Forms.Platform.iOS/Platform.cs
```diff
@@ -303,100 +303,180 @@ bool PageIsChildOfPlatform(Page page)
...
-		UIAlertAction CreateActionWithWindowHide(string text, UIAlertActionStyle style, Action setResult, UIWindow window)
+		void PresentAlert(AlertArguments arguments)
 		{
-			return UIAlertAction.Create(text, style,
-					a =>
+			PresentPopUp(false, (window, onComplete) =>
+			{
+				UIViewController alertViewController = null;
+
+				if (arguments.Visual.IsMaterial())
```
PureWeen: My first thought about this is that we probably need to think of a way to make a hook for this so people can provide there own implementations.  The idea with Visual is that anyone can hook in a "Visual" so we need a way of asking for an implementation based on the Visual. Maybe something that's registered into Dependency Service? 

Which also plays into your comment
*// TODO: move this out of the main assembly for linking*

Because we'll need some mechanism for that right?

PureWeen: Though we can probably get that implemented later on so that we can merge this PR

PureWeen: This is usually done through the IRegisterable interface and then creating a connection between two types. The cross platform and then the native

I was thinking maybe we create a new empty class that's just something like
*Class Alertview* or we anchor the registerable to *ActionSheetArguments* which is what is passed to *Internals.Registrar.Registered.GetHandler* to retrieve some interface that's used to pass the arguments to

Xamarin.Forms.Core/Page.cs
PureWeen: The code here looks to have a reference to the sender page so you can just send that to PresentAlert opposed to adding it onto the arguments

https://github.com/xamarin/Xamarin.Forms/blob/b1241170b61b9cef6c0963c5b5725d0e7bf443d8/Xamarin.Forms.Platform.iOS/Platform.cs#L41

Xamarin.Forms.Platform.iOS/Platform.cs
mattleibow: Yeah. I think we are on the same page here. We need to have some "registerable" something that can be looked up for "system" concepts - such as alerts or, if we ever go this route, local notifications. Not quite "renderers", more "logic chunks".

PureWeen: Yea this is how things like the ImageHandlers work. They aren't renderers they just get registered against the ImageSources like

https://github.com/xamarin/Xamarin.Forms/blob/release-3.4.0-sr1/Xamarin.Forms.Platform.Android/Properties/AssemblyInfo.cs#L64

So I'm thinking something similar to that

Xamarin.Forms.Core/Page.cs
mattleibow: That is probably a way better thing to do!

samhouts:
@PureWeen Please reopen when we're ready to work on this again. Thanks!

PR #4707, opened 2018-12-11
[Android] Fix old root page dispose when main page change

### LINKED ISSUE #4325: Unable to activate instance of type Xamarin.Forms.Platform.Android.AppCompat.TabbedPageRenderer from native handle
### Description

The below stack trace appeared in App Center Crashes recently from users who downloaded my app from the Google Play Store. I haven't been able to reproduce it yet, but wanted to share the stack trace with the team.

It is happening on Xamarin.Forms v3.3.0.912540.

### Steps to Reproduce

Unknown

### Basic Information

- Version with issue: Unknown
- IDE: Visual Studio Mac
- Platform Target Frameworks: 
  - Android: Android 8.1
- Android Support Library Version: v27.0.2.1
- Nuget Packages: Xamarin.Forms v3.3.0.912540
- Affected Devices: (Screen shot below)

### Screenshots

None

### Reproduction Link

None

### Stack Trace

```
Package: com.minnick.mondaypundayapp
Version Code: 157
Version Name: 3.4.4
Android: 8.0.0
Android Build: OPR1.170623.032
Manufacturer: LGE
Model: LG-H871
CrashReporter Key: bdb9c0db-4709-40c8-a607-34fd6f5ad9a2
Start Date: 2018-11-02T02:19:59.878Z
Date: 2018-11-02T16:56:47.097Z

Xamarin Exception Stack:
System.NotSupportedException: Unable to activate instance of type Xamarin.Forms.Platform.Android.AppCompat.TabbedPageRenderer from native handle 0x79 (key_handle 0x5f389d2).
  at Java.Interop.TypeManager.CreateInstance (System.IntPtr handle, Android.Runtime.JniHandleOwnership transfer, System.Type targetType) [0x00182] in <52cda36f06644c5abf922ea558a8f95b>:0
  at Java.Lang.Object.GetObject (System.IntPtr handle, Android.Runtime.JniHandleOwnership transfer, System.Type type) [0x000c1] in <52cda36f06644c5abf922ea558a8f95b>:0
  at Java.Lang.Object._GetObject[T] (System.IntPtr handle, Android.Runtime.JniHandleOwnership transfer) [0x00017] in <52cda36f06644c5abf922ea558a8f95b>:0
  at Java.Lang.Object.GetObject[T] (System.IntPtr handle, Android.Runtime.JniHandleOwnership transfer) [0x00000] in <52cda36f06644c5abf922ea558a8f95b>:0
  at Android.Views.ViewGroup.GetChildAt (System.Int32 index) [0x00036] in <52cda36f06644c5abf922ea558a8f95b>:0
  at Xamarin.Forms.Platform.Android.AppCompat.Platform.SetPageInternal (Xamarin.Forms.Page newRoot) [0x0001a] in <bf1be5de9f774184903c6001067cd76a>:0
  at Xamarin.Forms.Platform.Android.AppCompat.Platform+<>c__DisplayClass36_0.<SetPage>b__0 () [0x0001f] in <bf1be5de9f774184903c6001067cd76a>:0
  at Java.Lang.Thread+RunnableImplementor.Run () [0x00008] in <52cda36f06644c5abf922ea558a8f95b>:0
  at Java.Lang.IRunnableInvoker.n_Run (System.IntPtr jnienv, System.IntPtr native__this) [0x00009] in <52cda36f06644c5abf922ea558a8f95b>:0
  at (wrapper dynamic-method) System.Object.28(intptr,intptr)
--- End of inner exception stack trace ---
  System.MissingMethodException: No constructor found for Xamarin.Forms.Platform.Android.AppCompat.TabbedPageRenderer::.ctor(System.IntPtr, Android.Runtime.JniHandleOwnership)
    at Java.Interop.TypeManager.CreateProxy (System.Type type, System.IntPtr handle, Android.Runtime.JniHandleOwnership transfer) [0x00055] in <52cda36f06644c5abf922ea558a8f95b>:0
    at Java.Interop.TypeManager.CreateInstance (System.IntPtr handle, Android.Runtime.JniHandleOwnership transfer, System.Type targetType) [0x00116] in <52cda36f06644c5abf922ea558a8f95b>:0
  --- End of inner exception stack trace ---
    Java.Interop.JavaLocationException: Exception of type 'Java.Interop.JavaLocationException' was thrown.
      at Java.Lang.Error: Exception of type 'Java.Lang.Error' was thrown.
      at java.lang.Error: Java callstack:
      at mono.java.lang.RunnableImplementor.n_run(Native Method)
      at mono.java.lang.RunnableImplementor.run(RunnableImplementor.java:30)
      at android.os.Handler.handleCallback(Handler.java:789)
      at android.os.Handler.dispatchMessage(Handler.java:98)
      at android.os.Looper.loop(Looper.java:164)
      at android.app.ActivityThread.main(ActivityThread.java:6710)
      at java.lang.reflect.Method.invoke(Native Method)
      at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:770)

Thread 2:
0   dalvik.system.VMStack.getThreadStackTrace(VMStack.java:-2)
1   java.lang.Thread.getStackTrace(Thread.java:1536)
2   java.lang.Thread.getAllStackTraces(Thread.java:1586)
3   com.microsoft.appcenter.crashes.Crashes.saveUncaughtException(Crashes.java:923)
4   com.microsoft.appcenter.crashes.WrapperSdkExceptionManager.saveWrapperException(WrapperSdkExceptionManager.java:50)
5   mono.java.lang.RunnableImplementor.n_run(RunnableImplementor.java:-2)
6   mono.java.lang.RunnableImplementor.run(RunnableImplementor.java:30)
7   android.os.Handler.handleCallback(Handler.java:789)
8   android.os.Handler.dispatchMessage(Handler.java:98)
9   android.os.Looper.loop(Looper.java:164)
10  android.app.ActivityThread.main(ActivityThread.java:6710)
11  java.lang.reflect.Method.invoke(Method.java:-2)
12  com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
13  com.android.internal.os.ZygoteInit.main(ZygoteInit.java:770)

Thread 866:
0   java.lang.Object.wait(Object.java:-2)
1   java.lang.Daemons$ReferenceQueueDaemon.runInternal(Daemons.java:178)
2   java.lang.Daemons$Daemon.run(Daemons.java:103)
3   java.lang.Thread.run(Thread.java:764)

Thread 867:
0   java.lang.Object.wait(Object.java:-2)
1   java.lang.Object.wait(Object.java:422)
2   java.lang.ref.ReferenceQueue.remove(ReferenceQueue.java:188)
3   java.lang.ref.ReferenceQueue.remove(ReferenceQueue.java:209)
4   java.lang.Daemons$FinalizerDaemon.runInternal(Daemons.java:232)
5   java.lang.Daemons$Daemon.run(Daemons.java:103)
6   java.lang.Thread.run(Thread.java:764)

Thread 868:
0   java.lang.Object.wait(Object.java:-2)
1   java.lang.Daemons$FinalizerWatchdogDaemon.sleepUntilNeeded(Daemons.java:297)
2   java.lang.Daemons$FinalizerWatchdogDaemon.runInternal(Daemons.java:277)
3   java.lang.Daemons$Daemon.run(Daemons.java:103)
4   java.lang.Thread.run(Thread.java:764)

Thread 885:
0   android.os.MessageQueue.nativePollOnce(MessageQueue.java:-2)
1   android.os.MessageQueue.next(MessageQueue.java:325)
2   android.os.Looper.loop(Looper.java:142)
3   android.os.HandlerThread.run(HandlerThread.java:65)

Thread 887:
0   android.os.MessageQueue.nativePollOnce(MessageQueue.java:-2)
1   android.os.MessageQueue.next(MessageQueue.java:325)
2   android.os.Looper.loop(Looper.java:142)
3   android.os.HandlerThread.run(HandlerThread.java:65)

Thread 888:
0   android.os.MessageQueue.nativePollOnce(MessageQueue.java:-2)
1   android.os.MessageQueue.next(MessageQueue.java:325)
2   android.os.Looper.loop(Looper.java:142)
3   android.os.HandlerThread.run(HandlerThread.java:65)

Thread 901:
0   java.lang.Object.wait(Object.java:-2)
1   com.android.okhttp.okio.AsyncTimeout.awaitTimeout(AsyncTimeout.java:318)
2   com.android.okhttp.okio.AsyncTimeout.-wrap0
3   com.android.okhttp.okio.AsyncTimeout$Watchdog.run(AsyncTimeout.java:293)
```

### Affected Devices / OS's

image

### LINKED ISSUE #4706: Android crash on main page change
### Description
After migrated to Xamarin.Forms 3.4.0 from Xamarin 2.5.0, I got several exceptions in AppCenter of type **JavaLocationException**/**NotSupportedException**.

image

Before that, my application has no crash reported.

After some tests, impossible for me to reproduce it.

However, the information that can be deduced from the stacktrace is that the crash is during the main page affectation, and that the renderer seem to have been disposed.
So I cloned the repository and go to **SetPageInternal** method, the only incoming call was from the **LoadApplication** method in the activity. In this method was made the dispose of the renderer when the page was replaced. For me this case is possible, if you use a splashscreen (which is my case) or if the method LoadApplication is called a second time, when Android recreate the activity for exemple.

So to confirm that it's the case, I tried to call LoadApplication method twice and indeed the exception occured.

After some debug in the code, I found that the dispose of the renderer was also done later in the Cleanup method called by the method SetPageInternal inside the Platform class.
So I removed the dispose call in the Activity but after that another exception occured : **IllegalArgumentException: No view found for id 0x1 (unknown)**.
This time it was the ClearValue of  the page renderer which caused that, so I moved this code in the platform cleanup code near the dispose.

With those modifications I can now call the LoadApplication method twice without crash and with the application working smoothly.

I think that it will fix the crash encountered by the customers which use splashscreen in their apps or the case when application are restored from background.

I will made a PullRequest that fix the issue.

### Steps to Reproduce

1.  Create a new project from the Template
2.  Call LoadApplication twice in the MainActivity (this code is obviously wrong but must not crash the application)

### Expected Behavior
No crash

### Actual Behavior
Crash

### Basic Information

- Version with issue: All after 2.5.0
- Last known good version: 2.5.0
- IDE: Visual Studio 2017 15.9.3
- Platform Target Frameworks:
  - Android

kvpt:
### Description of Change ###
Don't do anything in SetPage if the page as not changed, do not destroy the previous one in this case.

### Issues Resolved ### 
- Fixes #4706
- Fixes #4325

### Platforms Affected ### 
- Android

### PR Checklist ###
- [ ] Has automated tests => Only manual test is possible

### DIFF
--- a/Xamarin.Forms.Core/Application.cs
+++ b/Xamarin.Forms.Core/Application.cs
@@ -5,8 +5,6 @@ using System.Threading;
 using System.ComponentModel;
 using System.Threading.Tasks;
 using Xamarin.Forms.Internals;
-using Xamarin.Forms.Platform;
-using System.Diagnostics;
 
 namespace Xamarin.Forms
 {
@@ -31,9 +29,6 @@ namespace Xamarin.Forms
 
 		public Application()
 		{
-			var f = false;
-			if (f)
-				Loader.Load();
 			NavigationProxy = new NavigationImpl(this);
 			SetCurrentApplication(this);
 
--- a/Xamarin.Forms.Core/Application.cs
+++ b/Xamarin.Forms.Core/Application.cs
@@ -5,6 +5,8 @@ using System.Threading;
 using System.ComponentModel;
 using System.Threading.Tasks;
 using Xamarin.Forms.Internals;
+using Xamarin.Forms.Platform;
+using System.Diagnostics;
 
 namespace Xamarin.Forms
 {
@@ -29,6 +31,9 @@ namespace Xamarin.Forms
 
 		public Application()
 		{
+			var f = false;
+			if (f)
+				Loader.Load();
 			NavigationProxy = new NavigationImpl(this);
 			SetCurrentApplication(this);
 
--- a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
@@ -18,7 +18,6 @@ using AToolbar = Android.Support.V7.Widget.Toolbar;
 using AColor = Android.Graphics.Color;
 using ARelativeLayout = Android.Widget.RelativeLayout;
 using Xamarin.Forms.Internals;
-using System.Threading.Tasks;
 
 #endregion
 
@@ -88,7 +87,7 @@ namespace Xamarin.Forms.Platform.Android
 			{
 			    throw new InvalidOperationException("Activity OnCreate was not called prior to loading the application. Did you forget a base.OnCreate call?");
 			}
-			
+
 			if (!_renderersAdded)
 			{
 				RegisterHandlerForDefaultRenderer(typeof(NavigationPage), typeof(NavigationPageRenderer), typeof(NavigationRenderer));
@@ -115,7 +114,7 @@ namespace Xamarin.Forms.Platform.Android
 			}
 
 			_application = application ?? throw new ArgumentNullException(nameof(application));
-			(application as IApplicationController)?.SetAppIndexingProvider(new AndroidAppIndexProvider(this));
+			(application as IApplicationController).SetAppIndexingProvider(new AndroidAppIndexProvider(this));
 			Xamarin.Forms.Application.SetCurrentApplication(application);
 
 			if (Xamarin.Forms.Application.Current.OnThisPlatform().GetWindowSoftInputModeAdjust() != WindowSoftInputModeAdjust.Unspecified)
@@ -123,17 +122,8 @@ namespace Xamarin.Forms.Platform.Android
 
 			CheckForAppLink(Intent);
 
+			application.PropertyChanged -= AppOnPropertyChanged;
 			application.PropertyChanged += AppOnPropertyChanged;
-
-			if (application?.MainPage != null)
-			{
-				var iver = Android.Platform.GetRenderer(application.MainPage);
-				if (iver != null)
-				{
-					iver.Dispose();
-					application.MainPage.ClearValue(Android.Platform.RendererProperty);
-				}
-			}
 
 			SetMainPage();
 		}
@@ -187,7 +177,7 @@ namespace Xamarin.Forms.Platform.Android
 
 			if (Forms.IsLollipopOrNewer)
 			{
-				// Listen for the device going into power save mode so we can handle animations being disabled	
+				// Listen for the device going into power save mode so we can handle animations being disabled
 				_powerSaveModeBroadcastReceiver = new PowerSaveModeBroadcastReceiver();
 			}
 		}
@@ -326,7 +316,7 @@ namespace Xamarin.Forms.Platform.Android
 			PopupManager.ResetBusyCount(this);
 
 			Platform = new AppCompat.Platform(this);
-			
+
 			Platform.SetPage(page);
 			_layout.AddView(Platform);
 			_layout.BringToFront();
--- a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
@@ -127,15 +127,6 @@ namespace Xamarin.Forms.Platform.Android
 
 			application.PropertyChanged += AppOnPropertyChanged;
 
-			if (application.MainPage != null)
-			{
-				var oldPageRenderer = Android.Platform.GetRenderer(application.MainPage);
-				if (oldPageRenderer != null)
-				{
-					application.MainPage.ClearValue(Android.Platform.RendererProperty);
-				}
-			}
-
 			SetMainPage();
 		}
 
--- a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
@@ -114,7 +114,7 @@ namespace Xamarin.Forms.Platform.Android
 			}
 
 			_application = application ?? throw new ArgumentNullException(nameof(application));
-			(application as IApplicationController).SetAppIndexingProvider(new AndroidAppIndexProvider(this));
+			((IApplicationController)application).SetAppIndexingProvider(new AndroidAppIndexProvider(this));
 			Xamarin.Forms.Application.SetCurrentApplication(application);
 
 			if (Xamarin.Forms.Application.Current.OnThisPlatform().GetWindowSoftInputModeAdjust() != WindowSoftInputModeAdjust.Unspecified)
--- a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
@@ -127,6 +127,15 @@ namespace Xamarin.Forms.Platform.Android
 
 			application.PropertyChanged += AppOnPropertyChanged;
 
+			if (application.MainPage != null)
+			{
+				var oldPageRenderer = Android.Platform.GetRenderer(application.MainPage);
+				if (oldPageRenderer != null)
+				{
+					application.MainPage.ClearValue(Android.Platform.RendererProperty);
+				}
+			}
+
 			SetMainPage();
 		}
 
--- a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
@@ -291,6 +291,12 @@ namespace Xamarin.Forms.Platform.Android
 
 		void AppOnPropertyChanged(object sender, PropertyChangedEventArgs args)
 		{
+			// Activity in pause must not react to application changes
+			if (_currentState >= AndroidApplicationLifecycleState.OnPause)
+			{
+				return;
+			}
+
 			if (args.PropertyName == "MainPage")
 				InternalSetPage(_application.MainPage);
 			if (args.PropertyName == PlatformConfiguration.AndroidSpecific.Application.WindowSoftInputModeAdjustProperty.PropertyName)
--- a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
@@ -113,6 +113,9 @@ namespace Xamarin.Forms.Platform.Android
 				_renderersAdded = true;
 			}
 
+			if (_application != null)
+				_application.PropertyChanged -= AppOnPropertyChanged;
+
 			_application = application ?? throw new ArgumentNullException(nameof(application));
 			((IApplicationController)application).SetAppIndexingProvider(new AndroidAppIndexProvider(this));
 			Xamarin.Forms.Application.SetCurrentApplication(application);
@@ -122,7 +125,6 @@ namespace Xamarin.Forms.Platform.Android
 
 			CheckForAppLink(Intent);
 
-			application.PropertyChanged -= AppOnPropertyChanged;
 			application.PropertyChanged += AppOnPropertyChanged;
 
 			SetMainPage();
--- a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
@@ -83,9 +83,9 @@ namespace Xamarin.Forms.Platform.Android
 
 		protected void LoadApplication(Application application)
 		{
-			if(!_activityCreated)
+			if (!_activityCreated)
 			{
-			    throw new InvalidOperationException("Activity OnCreate was not called prior to loading the application. Did you forget a base.OnCreate call?");
+				throw new InvalidOperationException("Activity OnCreate was not called prior to loading the application. Did you forget a base.OnCreate call?");
 			}
 
 			if (!_renderersAdded)
@@ -186,7 +186,11 @@ namespace Xamarin.Forms.Platform.Android
 
 		protected override void OnDestroy()
 		{
+			if (_application != null)
+				_application.PropertyChanged -= AppOnPropertyChanged;
+
 			PopupManager.Unsubscribe(this);
+
 			Platform?.Dispose();
 
 			// call at the end to avoid race conditions with Platform dispose
--- a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
@@ -297,7 +297,7 @@ namespace Xamarin.Forms.Platform.Android
 				return;
 			}
 
-			if (args.PropertyName == "MainPage")
+			if (args.PropertyName == nameof(_application.MainPage))
 				InternalSetPage(_application.MainPage);
 			if (args.PropertyName == PlatformConfiguration.AndroidSpecific.Application.WindowSoftInputModeAdjustProperty.PropertyName)
 				SetSoftInputMode();
--- a/Xamarin.Forms.Platform.Android/AppCompat/Platform.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/Platform.cs
@@ -6,7 +6,6 @@ using Android.Content;
 using Android.OS;
 using Android.Views;
 using Android.Views.Animations;
-using ARelativeLayout = Android.Widget.RelativeLayout;
 using AView = Android.Views.View;
 using Xamarin.Forms.Internals;
 
@@ -302,9 +301,6 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 
 				foreach (var root in _navModel.Roots)
 					renderersToDispose.Add(Android.Platform.GetRenderer(root));
-
-				if (Android.Platform.GetRenderer(Page) != null)
-					Page.ClearValue(Android.Platform.RendererProperty);
 
 				_navModel = new NavigationModel();
 
--- a/Xamarin.Forms.Platform.Android/AppCompat/Platform.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/Platform.cs
@@ -250,6 +250,11 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 
 		internal void SetPage(Page newRoot)
 		{
+			if (Page == newRoot)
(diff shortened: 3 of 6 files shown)


kvpt:
I also cleaning the import and removed dead code inside Application class that I came across.

Xamarin.Forms.Core/Application.cs
```diff
@@ -31,9 +29,6 @@ public class Application : Element, IResourcesProvider, IApplicationController,
 
 		public Application()
 		{
-			var f = false;
```
samhouts: As silly as this looks, this code serves a purpose. Please do not remove it. Thank you!

kvpt: Ok, no problem.

The code is here from the first commit on github, so it don't provide any hint about its meaning but as the variable is local to the function and I do not see any way to change it.
I thought the change was without impact, but since it's not directly related to this PR I leave that to a future "house keeping" commit.

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -115,26 +114,17 @@ protected void LoadApplication(Application application)
 			}
 
 			_application = application ?? throw new ArgumentNullException(nameof(application));
-			(application as IApplicationController)?.SetAppIndexingProvider(new AndroidAppIndexProvider(this));
+			(application as IApplicationController).SetAppIndexingProvider(new AndroidAppIndexProvider(this));
```
StephaneDelcroix: direct cast would be ok too
```suggestion
			((IApplicationController)application).SetAppIndexingProvider(new AndroidAppIndexProvider(this));
```

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -115,26 +114,17 @@ protected void LoadApplication(Application application)
...
 			Xamarin.Forms.Application.SetCurrentApplication(application);
 
 			if (Xamarin.Forms.Application.Current.OnThisPlatform().GetWindowSoftInputModeAdjust() != WindowSoftInputModeAdjust.Unspecified)
 				SetSoftInputMode();
 
 			CheckForAppLink(Intent);
 
+			application.PropertyChanged -= AppOnPropertyChanged;
```
StephaneDelcroix: you want to remove the handler on the _old_ application, which is, if not null, `_application` before L116

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -115,26 +114,17 @@ protected void LoadApplication(Application application)
...
 			application.PropertyChanged += AppOnPropertyChanged;
 
-			if (application?.MainPage != null)
```
StephaneDelcroix: I think we still need to dispose the old main page renderer, on `_application`, before L116

Xamarin.Forms.Platform.Android/AppCompat/Platform.cs
```diff
@@ -294,6 +294,9 @@ void SetPageInternal(Page newRoot)
 				foreach (var root in _navModel.Roots)
 					renderersToDispose.Add(Android.Platform.GetRenderer(root));
 
+				if (Android.Platform.GetRenderer(Page) != null)
+					Page.ClearValue(Android.Platform.RendererProperty);
```
StephaneDelcroix: and you might want to dispose the renderer too

Xamarin.Forms.Platform.Android/FormsApplicationActivity.cs
```diff
@@ -84,27 +80,18 @@ public void SetPage(Page page)
...
 
-			(application as IApplicationController)?.SetAppIndexingProvider(new AndroidAppIndexProvider(this));
+			(application as IApplicationController).SetAppIndexingProvider(new AndroidAppIndexProvider(this));
 
 			_application = application;
 			Xamarin.Forms.Application.SetCurrentApplication(application);
 
 			SetSoftInputMode();
 
+			application.PropertyChanged -= AppOnPropertyChanged;
```
StephaneDelcroix: see my earlier comment

Xamarin.Forms.Platform.Android/FormsApplicationActivity.cs
```diff
@@ -84,27 +80,18 @@ public void SetPage(Page page)
...
 			application.PropertyChanged += AppOnPropertyChanged;
 
-			if (application?.MainPage != null)
```
StephaneDelcroix: see ^^^

Xamarin.Forms.Platform.Android/AppCompat/Platform.cs
kvpt: The renderer is disposed by the Cleanup method called later in this method.

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
kvpt: Good point, I tested with the same application object for the two calls but I also need to test with a different application object.
I will test that case.

kvpt: I tested the two cases, with the same application or a different it work fine in all case.

I tried to dispose the renderer on _application but it's too early.

With dispose + clear :
image

With clear only :
image

That the same issue that the current code.

Xamarin.Forms.Core/Application.cs
samhouts: Yep, it definitely deserves a comment. This code is here to prevent the iOS linker from completely removing the Xamarin.Forms.Platform DLL.

kvpt:
I can provide a reproduction case to illustrate the crash if needed.

kvpt:
Here the repro project : [issue4706.zip](https://github.com/xamarin/Xamarin.Forms/files/2692710/issue4706.zip)
I made a new project from the VS template and replaced the nugets with the source code of the master branch of the framework, it's easier to debug like that.

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
kvpt: I deployed the fix in the staging version of my application.
I dont have anymore report of NotSupportedException but I have now reports of NullReferenceException inside the layout method from some devices.

I did further investigations and found the origin of the layout call.
The layout is triggered by the affectation of the ToolbarVisibility [here](https://github.com/xamarin/Xamarin.Forms/blob/ceb143a2a5f970a70561c53c14f8c9a3328999c4/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs#L136).
For testing I removed the call and now I didn't have the layout anymore time and can clear the renderer earlier whitout the NRE.

After that I tested the IsLayoutRequested property and it was true, so I added a check to this condition before request another layout pass.

It fix the issue in my reproduction code but I dont know if it fix all the cases.
This layout request call seems out of place compared to others.
What's your opinion on that ?

kvpt:
We have two problems here.
The dispose which was called twice, previously fixed.
And the layout call from the toolbar which cause NRE during layout depending on when the renderer is cleared. I'm unsure how to proceed on this part, I made a try.
I will deploy this new version of the fix tomorrow in staging and see if it fix the problem or not.

kvpt:
I finally found the root cause of this bug.
When the application is reused, the MainPage dont change, it's the same instance.
So when the SetPage method is called, the renderer of the oldPage is disposed but it's also the renderer of the newPage, so when the page try to render it crash.

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -114,8 +113,11 @@ protected void LoadApplication(Application application)
 				_renderersAdded = true;
 			}
 
+			if (_application != null)
+				_application.PropertyChanged -= AppOnPropertyChanged;
```
mattleibow: Good catch.

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -114,8 +113,11 @@ protected void LoadApplication(Application application)
...
+
 			_application = application ?? throw new ArgumentNullException(nameof(application));
-			(application as IApplicationController)?.SetAppIndexingProvider(new AndroidAppIndexProvider(this));
+			((IApplicationController)application).SetAppIndexingProvider(new AndroidAppIndexProvider(this));
```
mattleibow: Why this change? If anything, just call `SetAppIndexingProvider` directly as it is a public member of `Application` (but marked as not visible to the editor)

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
mattleibow: Not sure if I am following, but could it be that you are disposing the same page? ie: is this the same application object with the same page?
Also, that last null exception - is that the old page that is throwing, or the new one? If it is the old page, then why is it trying to lay it out?

Xamarin.Forms.Platform.Android/FormsApplicationActivity.cs
```diff
@@ -83,28 +79,17 @@ public void SetPage(Page page)
...
-			if (application == null)
-				throw new ArgumentNullException("application");
+			if (_application != null)
+				_application.PropertyChanged -= AppOnPropertyChanged;
 
-			(application as IApplicationController)?.SetAppIndexingProvider(new AndroidAppIndexProvider(this));
-
-			_application = application;
+			_application = application ?? throw new ArgumentNullException(nameof(application));
+			((IApplicationController)application).SetAppIndexingProvider(new AndroidAppIndexProvider(this));
```
mattleibow: Same as above. why the change? and maybe use the method directly.

Xamarin.Forms.Platform.Android/FormsApplicationActivity.cs
mattleibow: You nuked the disposing. What is there that is done to replace this logic?

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
kvpt: My change was to remove the **?** because it was useless, then StephaneDelcroix suggered to replace the safe cast with a direct cast.

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
kvpt: Yes the dispose problem occurs because it's the same page with the same app.
The null reference is throwed because of the layout tirggered by setting the toolbar visibility and if the page have already been disposed like in this case, it crash.

Xamarin.Forms.Platform.Android/FormsApplicationActivity.cs
kvpt: The dispose is already done later in the Platform class.
This code was added later by PR #1200 and #1212 but the fix was wrong..
If you not dispose the page in the case it's the same page the issue that PR tried to fix doesn't exist anymore.

Xamarin.Forms.Platform.Android/FormsApplicationActivity.cs
kvpt: This change was only to be uniform between the two platforms.

mreuben:
Whats the fix for this issue?

kvpt:
The fix is this PR, it fix the case where you reuse the same application object between two calls of OnCreate method.

It correct two things :
- It add a verification inside the method SetPage which check if it's not the same page that is setted again because in this case the page is disposed and cant be reused after that.
- It also fix the PropertyChanged event handler on the application object which wasn't unscribed.

mreuben:
I am getting this error while navigating to previous page.
Sorry to ask. What is meant by PR? Xamarin.forms version is 
3.0.0.550146
Please mention the fix.

This is the code in Mainactivity.cs
 public class MainActivity : global::Xamarin.Forms.Platform.Android.FormsAppCompatActivity
    {
        protected override void OnCreate(Bundle bundle)
        {
            TabLayoutResource = Resource.Layout.Tabbar;
            ToolbarResource = Resource.Layout.Toolbar;

            base.OnCreate(bundle);

            global::Xamarin.Forms.Forms.Init(this, bundle);
            try
            {
                LoadApplication(new App());
            }
            catch(Exception ex)
            {
                throw ex;
            }
        }
        public override void OnRequestPermissionsResult(int requestCode, string[] permissions, Android.Content.PM.Permission[] grantResults)
        {
            PermissionsImplementation.Current.OnRequestPermissionsResult(requestCode, permissions, grantResults);
        }

kvpt:
@mreuben, PR means [Pull Request](https://help.github.com/articles/about-pull-requests/), here the link to the github documentation.
The fix here is on the Framework code (Xamarin Forms) not the User code (Your code), the only way to apply it is to made a custom nuget package from the source.
But from your example, I can tell that this PR will dont fix your issue, the issue here, is caused by the App object reusing which is not your case because you create a new App each time.
I advise you to open a new [issue ](https://help.github.com/articles/about-issues/) with a complete stacktrace and provide a reproduction project. This is the first step that will allow developers to fix the issue.

kvpt:
After deploying this patch, the number of crashes form app decreased but have not completely disappeared.
After many tries I discovered the remaining issue.
The activity was not unsubscribed from the application on destroy.
So if you have use a static application like me, when the activity is destroyed and a new is recreated on application restore. 
The AppOnPropertyChanged handler is called from the old activity and the also the new.
The consequence is the SetPage method is called twice, that result in either white screen or rendered  disposed (aka JavaLocationException) and also a memory leak because the if the application is static it prevent the old activity to be disposed.

This kind of problem is critical and affect my app for almost two years (an others as well).

At this time, if you use a static application object, Xamarin.Forms do not survive after a destroy.

I think that this PR need to be merged ASAP, what do you think ?

kvpt:
My last patch doesn't fix the problem completely, I have one remaning crash since the last fix.
After some more research I found that it's because the lifecycle is more complex than I thought, thanks to [this post](https://medium.com/androiddevelopers/the-android-lifecycle-cheat-sheet-part-ii-multiple-activities-a411fd139f24). The destroy of the old activity is called after the call to create of the new activity.
It works most of the time because the start method is slow but sometime if the destroy is called too late, the problem come back.
So I added a check in the handler to ignore the event if the activity state is paused.
I hope it's the last one this time.

PR #4709, opened 2018-12-12
[iOS] Added RowAnimationsEnabled platform specific to control row animations

### LINKED ISSUE #4199: [iOS] Add ability to turn off animations in ListView during data load
Currently, there is no reliable way to disable animations in `ListView` when its items collection is being updated. The existing `UITableViewRowAnimation` properties do not always disable animations particularly when `ListView` is being shown for the first time. A more robust way to do this is to perform `beginUpdates()` and `endUpdates()` operations like so:

```
UIView.performWithoutAnimation {
    tableView.beginUpdates()
    tableView.endUpdates()   
}
```

A platform specific property would go in [here](https://github.com/xamarin/Xamarin.Forms/blob/76287374645ca1cf65e4c15ebbff21a935f583a0/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs#L543), [here](https://github.com/xamarin/Xamarin.Forms/blob/76287374645ca1cf65e4c15ebbff21a935f583a0/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs#L1141), [here](https://github.com/xamarin/Xamarin.Forms/blob/76287374645ca1cf65e4c15ebbff21a935f583a0/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs#L269), and [here](https://github.com/xamarin/Xamarin.Forms/blob/76287374645ca1cf65e4c15ebbff21a935f583a0/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs#L1191) so that row update animations can be turned off.

Also, I noticed that `ReloadSectionsAnimation` is not being used, and it should either be removed or utilized somewhere in the renderer.

adrianknight89:
### Description of Change ###

An explanation as to why this is needed is discussed in the issue ticket. Essentially, this allows the developer to turn off row animations on demand instead of relying on `UITableViewRowAnimation` which does not work in certain cases. Currently, I'm setting `UIView.AnimationsEnabled` to false as a workaround, but this has very bad side effects like turning off status bar animations as well as soft keyboard animations.

PS: I'd like to be able to test this in my project, but I can't create a nuget package with `.create-nuget.bat` since it's looking for `..\Xamarin.Forms.Platform.Android\bin\debug\MonoAndroid81` and I have Android 9 installed. Does your script need to be updated?

You could try testing this in `60699` by turning off the platform specific.

### Issues Resolved ### 

- fixes #4199

### API Changes ###

Added:

bool RowAnimationsEnabled platform specific

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/ListView.cs
+++ b/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/ListView.cs
@@ -50,7 +50,6 @@
 			return config;
 		}
 
-		#region RowAnimationsEnabled
 		public static readonly BindableProperty RowAnimationsEnabledProperty = BindableProperty.Create(nameof(RowAnimationsEnabled), typeof(bool), typeof(ListView), true);
 
 		public static bool GetRowAnimationsEnabled(BindableObject element)
@@ -73,6 +72,5 @@
 		{
 			return GetRowAnimationsEnabled(config.Element);
 		}
-		#endregion
 	}
 }
--- a/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
@@ -40,6 +40,11 @@ namespace Xamarin.Forms.Platform.iOS
 		protected UITableViewRowAnimation InsertRowsAnimation { get; set; } = UITableViewRowAnimation.Automatic;
 		protected UITableViewRowAnimation DeleteRowsAnimation { get; set; } = UITableViewRowAnimation.Automatic;
 		protected UITableViewRowAnimation ReloadRowsAnimation { get; set; } = UITableViewRowAnimation.Automatic;
+		protected UITableViewRowAnimation ReloadSectionsAnimation
+		{
+			get { return _dataSource.ReloadSectionsAnimation; }
+			set { _dataSource.ReloadSectionsAnimation = value; }
+		}
 
 		public override SizeRequest GetDesiredSize(double widthConstraint, double heightConstraint)
 		{
@@ -971,6 +976,7 @@ namespace Xamarin.Forms.Platform.iOS
 			bool _isDragging;
 			bool _selectionFromNative;
 			bool _disposed;
+			public UITableViewRowAnimation ReloadSectionsAnimation { get; set; } = UITableViewRowAnimation.Automatic;
 
 			public ListViewDataSource(ListViewDataSource source)
 			{


Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -41,11 +40,6 @@ public class ListViewRenderer : ViewRenderer<ListView, UITableView>
 		protected UITableViewRowAnimation InsertRowsAnimation { get; set; } = UITableViewRowAnimation.Automatic;
 		protected UITableViewRowAnimation DeleteRowsAnimation { get; set; } = UITableViewRowAnimation.Automatic;
 		protected UITableViewRowAnimation ReloadRowsAnimation { get; set; } = UITableViewRowAnimation.Automatic;
-		protected UITableViewRowAnimation ReloadSectionsAnimation
```
adrianknight89: This is not being used anywhere.

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -607,23 +583,101 @@ void UpdateItems(NotifyCollectionChangedEventArgs e, int section, bool resetWhen
...
 
 				case NotifyCollectionChangedAction.Reset:
 					InvalidateCellCache();
-					Control.ReloadData();
+					ReloadData();
 					return;
 			}
 		}
 
+		void InsertRows(NotifyCollectionChangedEventArgs e, int section)
```
rmarinho: can we pass just e.NewStartingIndex, e.NewItems.Count  ?

adrianknight89: Done.

Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/ListView.cs
```diff
@@ -49,5 +49,30 @@ public static GroupHeaderStyle GetGroupHeaderStyle(this IPlatformElementConfigur
 			SetGroupHeaderStyle(config.Element, value);
 			return config;
 		}
+
+		#region RowAnimationsEnabled
```
StephaneDelcroix: don't introduce regions

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
StephaneDelcroix: keep it. it might be there so it can be accessed by subclasses and custom renderers

adrianknight89:
Done

PR #4737, opened 2018-12-14
[macOS] Only the first Navigating event of WebView can be triggered

zhuxinghan:
### Description of Change ###

Fix the problem that only the first Navigating event of WebView can be triggered

### Platforms Affected ### 

- macOS


rmarinho:
build

rmarinho:
build --uitests

PR #4742, opened 2018-12-14
Condition default items on the right property for SDK projects

kzu:
See https://github.com/xamarin/Xamarin.Forms/pull/4699#issuecomment-446082593

### DIFF
--- a/.nuspec/Xamarin.Forms.props
+++ b/.nuspec/Xamarin.Forms.props
@@ -1,10 +1,5 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-	<PropertyGroup>
-		<!-- Sdk.props imports Microsoft.NET.Sdk.props which defines Configurations and Platforms, which is the SDK-style way -->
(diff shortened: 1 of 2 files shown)


.nuspec/Xamarin.Forms.props
```diff
@@ -15,6 +10,6 @@
 	The actual item groups are in a separate conditionally-imported file as they use constructs that
 	are not compatible with older MSBuild versions.
 	-->
-	<Import Project="$(MSBuildThisFileDirectory)Xamarin.Forms.DefaultItems.props" Condition="'$(IsSdkStyle)' == 'true'" />
+	<Import Project="$(MSBuildThisFileDirectory)Xamarin.Forms.DefaultItems.props" Condition="'$(EnableDefaultItems)' == 'true'" />
```
PureWeen: It looks like the EnableDefaultItems property hasn't been set by this point.

If I change the condition to 
```
<Import Project="$(MSBuildThisFileDirectory)Xamarin.Forms.DefaultItems.props" Condition="'$(EnableDefaultItems)' == ''" />
```

Then it pulls the props file in and it compiles

StephaneDelcroix:
this is no longer valid, as I reverted #4699

PR #4743, opened 2018-12-14
[X] Design time properties

StephaneDelcroix:
An alternate xaml resource file provider can request the XamlLoader to
_not_ ignore normally ignored properties in prebuilt XF design xmlns, as
in the following snippet:

```xaml
<ContentPage
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    xmlns:d="http://xamarin.com/schemas/2014/forms/design"
    mc:Ignorable="d">
  <Label d:Text="Bar" Text="Foo" x:Name="label" />
</ContentPage>
```

The `d:` should be the prefix used by default for this, but any other
prefix will do too.

The `d:Text` property maps to the exact same property as `Text`, as the
XmllnsDefinitionAttributes are identical (that's convenient from a Intelisense
point of view), but, when (and only when) instructed by a provided Xaml resource
loader, the `d:Text` will override the `Text` property.

This works with virtually all properties defined on built-in Xamarin.Forms
controls, but it doesn't mean it's a sane idea to try to assign design value
to all existing properties.

The API for setting the ResourceLoader had to change, and instead of taking
pre-defined arguments, it accepts and returns query and response types. This
is slightly less convenient to invoke through reflection, but way more easy
to extend in the future.

As a side note, I'm quite proud, and thankful, of the work done by that 5-people
2-hours brainstorming-meeting that came up with the new name for the now
deprecated `ResourceProvider`.

### DIFF
--- a/Xamarin.Forms.Core/Internals/ResourceLoader.cs
+++ b/Xamarin.Forms.Core/Internals/ResourceLoader.cs
@@ -1,19 +1,54 @@
 ﻿using System;
+using System.ComponentModel;
 using System.Reflection;
 
 namespace Xamarin.Forms.Internals
 {
+	[EditorBrowsable(EditorBrowsableState.Never)]
 	public static class ResourceLoader
 	{
 		static Func<AssemblyName, string, string> resourceProvider;
 
+		[Obsolete("You shouldn't have used this one to begin with, don't use the other one either")]
 		//takes a resource path, returns string content
 		public static Func<AssemblyName, string, string> ResourceProvider {
 			get => resourceProvider;
 			internal set {
-				DesignMode.IsDesignModeEnabled = true;
 				resourceProvider = value;
+				if (value != null)
+					ResourceProvider2 = rlq => new ResourceLoadingResponse { ResourceContent = value(rlq.AssemblyName, rlq.ResourcePath) };
+				else
+					ResourceProvider2 = null;
 			}
+		}
+
+		static Func<ResourceLoadingQuery, ResourceLoadingResponse> _resourceProvider2;
+		public static Func<ResourceLoadingQuery, ResourceLoadingResponse> ResourceProvider2 {
+			get => _resourceProvider2;
+			internal set {
+				DesignMode.IsDesignModeEnabled = value != null;
+				_resourceProvider2 = value;
+			}
+		}
+
+		[Obsolete("Can't touch this")]
+		public static bool CanProvideContentFor(ResourceLoadingQuery rlq)
+		{
+			if (_resourceProvider2 == null)
+				return false;
+			return _resourceProvider2(rlq) != null;
+		}
+
+		public class ResourceLoadingQuery
+		{
+			public AssemblyName AssemblyName { get; set; }
+			public string ResourcePath { get; set; }
+		}
+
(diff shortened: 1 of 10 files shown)


rmarinho:
build

Xamarin.Forms.Core/Internals/ResourceLoader.cs
```diff
@@ -1,21 +1,56 @@
...
 using System.Reflection;
 
 namespace Xamarin.Forms.Internals
 {
+	[EditorBrowsable(EditorBrowsableState.Never)]
 	public static class ResourceLoader
 	{
 		static Func<AssemblyName, string, string> resourceProvider;
 
+		[Obsolete("You shouldn't have used this one to begin with, don't use the other one either")]
```
mattleibow: nice 😄

Xamarin.Forms.Xaml/PruneIgnoredNodesVisitor.cs
```diff
@@ -17,12 +20,21 @@ public void Visit(ElementNode node, INode parentNode)
...
 			{
 				var propertyName = propertyKvp.Key;
-				var propertyValue = (propertyKvp.Value as ValueNode)?.Value as string;
-				if (propertyValue == null)
+				if (!((propertyKvp.Value as ValueNode)?.Value is string propertyValue))
 					continue;
 				if (!propertyName.Equals(XamlParser.McUri, "Ignorable"))
 					continue;
-				(parentNode.IgnorablePrefixes ?? (parentNode.IgnorablePrefixes = new List<string>())).AddRange(propertyValue.Split(','));
+				var prefixes = propertyValue.Split(',').ToList();
```
mattleibow: The `,` is the incorrect splitter here.
https://docs.microsoft.com/en-us/dotnet/framework/wpf/advanced/mc-ignorable-attribute
> Multiple prefixes can be specified in the attribute, using one or more white-space characters as the separator, for example: mc:Ignorable="ignore1 ignore2".

StephaneDelcroix: good catch

Xamarin.Forms.Xaml/PruneIgnoredNodesVisitor.cs
```diff
@@ -17,12 +20,21 @@ public void Visit(ElementNode node, INode parentNode)
...
+				if (UseDesignProperties) {//if we're in design mode for this file
+					var prefixIndexToRemove = -1;
+					for (var i = 0; i < prefixes.Count; i++)
+						if (node.NamespaceResolver.LookupNamespace(prefixes[i]) == XamlParser.XFDesignUri)
+							prefixIndexToRemove = i;
+					if (prefixIndexToRemove != -1)
+						prefixes.RemoveAt(prefixIndexToRemove);
+				}
```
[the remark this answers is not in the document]
StephaneDelcroix: this doesn't remove the design properties, but the normal ones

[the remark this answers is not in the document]
StephaneDelcroix: thanks for caring about the crazy dude

PR #4745, opened 2018-12-14
[UItests] Fix UITest 41424

rmarinho:
### Description of Change ###

Fix UITest 41424

### Issues Resolved ### 

- fixes #

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla41424.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla41424.cs
@@ -67,6 +67,7 @@ namespace Xamarin.Forms.Controls.Issues
 		public void DatePickerCancelShouldUnfocus()
 		{
 			RunningApp.Tap(q => q.Marked(DatePicker));
+			RunningApp.WaitForElement(q => q.Marked("OK"));
 			RunningApp.Back();
 
 			RunningApp.WaitForElement(q => q.Marked("Click to view focus state"));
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla41424.cs
```diff
@@ -67,19 +68,37 @@ protected override void Init()
...
-			RunningApp.Back();
+			RunningApp.WaitForElement(q => q.Marked("OK"));
+			Assert.IsTrue(DialogIsOpened(),"Call Focus Picker");
 
 			RunningApp.WaitForElement(q => q.Marked("Click to view focus state"));
 			RunningApp.Tap(q => q.Marked("Click to view focus state"));
 			RunningApp.WaitForElement(q => q.Marked("unfocused"));
 		}
+
+		bool DialogIsOpened()
```
hartez: A couple of requests:
1. Can we make this an extension method that takes `this App` (i.e., RunningApp.DialogIsOpened()`) so we can use it in other places? It seems like a generally useful method.
2. Can we add a timeout parameter (int timeout = 1500)?
3. Can we have it poll every few hundred milliseconds until the timeout? That way it _could_end ealier than 1.5 seconds, if the dialog is already open.

Not required for getting this merged, just a nice-to-have.

mattleibow:
I agree with @hartez to make this end sooner. Otherwise, this test will always take a minimum of 3 seconds - and it could be done in 200ms. And, an extension method will help with many of the other dialog/popup/picker tests that we have and soon will have.

PR #4746, opened 2018-12-15
Condition default items on the UsingMicrosoftNETSdk property for SDK projects 

PureWeen:
https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props#L26

If someone only has pre 2.0 sdk installed then this property doesn't work
Not sure if that means this won't work

Compiling works on VSMac

Here's the binlog from compiling a PCL library
image


StephaneDelcroix:
please revisit as we had to revert #4699

PR #4747, opened 2018-12-15
Min viable FontImageSource

### LINKED ISSUE #3189: Xamarin.Forms.FontIconSource Spec
# Font Icons

The goal of this spec is to support the use of font icons (e.g., the [Segoe MDL 2 icons](https://docs.microsoft.com/en-us/windows/uwp/design/style/segoe-ui-symbol-font) or [Font Awesome](https://fontawesome.com/)) as image sources. 

_Note: Nothing in this specification is guaranteed to be final; all features, implementations, and interfaces are subject to change._

# API

## IFontElement

Implementing classes support setting various font properties. 

_Note: this is not a new interface; it's already part of Forms. It is included here only for reference purposes._

```csharp
public interface IFontElement
{
	FontAttributes FontAttributes { get; }
	string FontFamily { get; }
	double FontSize { get; }
}
```

## FontIconSource

Supports using glyph icons.

```csharp
public class FontIconSource : ImageSource, IFontElement
{
	public static readonly BindableProperty GlyphProperty;

	public static readonly BindableProperty NameProperty;
}
```

### Properties

| API | Description |
| Glyph | Gets or sets the charactor code to be displayed. |
| Color | Gets or sets the color of the glyph.  |
| Name | Gets or sets the 'friendly' name of the glyph. If this property is set, the value of Glyph will be retrieved from the `FontIconRegistry`. | 

## LayeredFontIconSource

Supports layering glyph icons (e.g., see: https://docs.microsoft.com/en-us/windows/uwp/design/style/segoe-ui-symbol-font#layering-and-mirroring).

```csharp
[ContentProperty("Layers")]
public class LayeredFontIconSource : ImageSource, IFontElement
{
	public IList<FontIconSource> Layers { get; }
}
```

### Properties

| API | Description |
| Layers | Gets the set of icons to be displayed. The icons are stacked in the order they are defined, last one on top. |

If a `FontIconSource` in the `Layers` list does not specify a font property, that property's value is inherited from the `LayeredFontIconSource`.

## FontIconRegistry

Supports registering 'friendly' names for icon glyphs. A Forms application has a single instance of this class.

```csharp
public class FontIconRegistry
{
	public void SetDefaultIconFontFamily(string fontFamily);
	public void Clear();
	public void SetGlyphName(string fontFamily, string glyph, string name);
	public void SetGlyphName(string glyph, string name);
	public string GetGlyph(string name);
	public string GetGlyph(string fontFamily, string name);
}
```

### Methods

| API | Description |
| SetDefaultIconFontFamily | Sets the default font family to use when registering and resolving glyphs. If this value is set, it will be used by any `FontIconSource` which does not specify a `FontFamily`. |
| Clear | Removes all entries from the regsitry. |
| SetGlyphName(string fontFamily, string glyph, string name) | Registers a friendly name for a glyph with a specific font family. |
| SetGlyphName(string glyph, string name) | Registers a friendly name for a glyph with the default font family. Throws an exception if a default font family is not set. |
| GetGlyph(string name) | Retrieves the glyph for the specified name using the default font family. Throws an exception if the default font family is not set or the name is not found. |
| GetGlyph(string name) | Retrieves the glyph for the specified name and font family. Throws an exception if either the font family or name are not found. |

## FontIconExtension

Supports specifying a `FontIconSource` inline in an `Image` tag's `Source` property. For example,

```
<Image Source="{FontIcon Name=share,Color={StaticResource primaryColor}}" />
``` 

```csharp
public sealed class FontIconExtension : IMarkupExtension<ImageSource>
{
...
}
```

kingces95:
### Description of Change ###

`FontImageSource` derives from `ImageSource` and accepts a glyph (character) from a font plus a size and color which it'll use to create a `.png` which can then be displayed as an image. Unfortunately, many places where you'd like to use this currently accept a `FileImageSource` but that'll be a separate work item. 

### Issues Resolved ### 

- fixes #3189 tho that spec has aspirations of unifying the metadata needed to identify the font such that the XAML is simplified to a single word. That's a great idea but is beyond what's necessary for a minimum viable feature. And anyway, abstracting resources across platforms is a larger issue.

### API Changes ###

Added:
`FontImageSource` 
see https://github.com/xamarin/Xamarin.Forms/compare/font-image-source?expand=1#diff-6e54598e403bbd75200d983f2182df8fR564

snag_cad069f

### Testing Procedure ###
Run `Font ImageSource - Legacy`


charlesroddie:
Why the intermediate png?

Xamarin.Forms.Core/FontImageSource.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Xamarin.Forms.Internals;
+using ValidateValueDelegate = Xamarin.Forms.BindableProperty.ValidateValueDelegate;
+
+namespace Xamarin.Forms
+{
+	public partial class FontImageSource : ImageSource
+	{
+		private static BindableProperty CreateBindableProperty<T>(
```
StephaneDelcroix: I'd avoid trying to be smart here. it makes the code harder to understand in the big picture

Xamarin.Forms.Core/FontImageSource.cs
```diff
@@ -0,0 +1,61 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Xamarin.Forms.Internals;
+using ValidateValueDelegate = Xamarin.Forms.BindableProperty.ValidateValueDelegate;
+
+namespace Xamarin.Forms
+{
+	public partial class FontImageSource : ImageSource
```
StephaneDelcroix: ImageSourceConverter needs to be extended with a new format to create a FontImageSource easily from XAML. please propose something

paymicro:
I have an implementation for UWP. Do we need to add this?
screenshot_1

samhouts:
build --uitests

Xamarin.Forms.Core/FontImageSource.cs
```diff
@@ -0,0 +1,61 @@
...
+				declaringType: typeof(FontImageSource),
+				defaultValue: defaultValue,
+				validateValue: validate
+			);
+		}
+
+		public double Size { get => (double)GetValue(SizeProperty); set => SetValue(SizeProperty, value); }
+		public static readonly BindableProperty SizeProperty = CreateBindableProperty(nameof(Size), 30d);
+
+		public char Glyph { get => (char)GetValue(GlyphProperty); set => SetValue(GlyphProperty, value); }
```
mattleibow: The one issue I have with this is the fact that UTF-32 (such as unicode emojis) take up 2 characters, so this might be better as a string...

kingces95:
> Why the intermediate png?

To allow emoji to be used everywhere an `ImageSource` is allowed as opposed to just where a label is allowed. At least, that's the dream. We still need to change `FileImageSource` to `ImageSource` all over the place to realize the dream. And yes, the dream does come at the cost of drawing a png...

mattleibow:
If we cache the image somehow, then this really is an optimization - in a sense - because we don't have to redraw the emoji each time.

Xamarin.Forms.Platform.UAP/FontImageSourceHandler.cs
```diff
@@ -0,0 +1,41 @@
...
+		float _minimumDpi = 300;
+
+		public Task<Windows.UI.Xaml.Media.ImageSource> LoadImageAsync(ImageSource imagesource,
+			CancellationToken cancelationToken = default(CancellationToken))
+		{
+			if (!(imagesource is FontImageSource fontsource))
+				return null;
+
+			var device = CanvasDevice.GetSharedDevice();
+			var dpi = Math.Max(_minimumDpi, Windows.Graphics.Display.DisplayInformation.GetForCurrentView().LogicalDpi);
```
mattleibow: Just a note about DPI on UWP: it can change during the lifetime of an app and even the control. The density can be changed via settings, but also by just dragging the app to a new monitor. I have a 200% surface and a 150% externbal display.

This might also change if accessibility is used - possibly on iOS and Android as well.

Xamarin.Forms.Platform.UAP/ImageRenderer.cs
```diff
@@ -28,9 +29,28 @@ public override SizeRequest GetDesiredSize(double widthConstraint, double height
...
+					new Size
+					{
+						Width = canvas.SizeInPixels.Width,
+						Height = canvas.SizeInPixels.Height
+					});
+			}
+			else
+			{
+				throw new InvalidCastException($"\"{Control.Source.GetType().FullName}\" is not supported.");
+			}
```
mattleibow: I would move this into an extension method because it probably needs to be used on other image-bearing controls - such as image buttons or page icons.

Xamarin.Forms.Platform.UAP/FontImageSourceHandler.cs
kingces95: Truth! But should fixing that use case block the PR? Kinda sounds like a corner case on a corner platform, no? Otherwise, we'll have to detect that the DPI changed and then run around and update glyphs? Sounds icky (or is it simple)? Do we think this is the only place we have that issue? Which is to say, if there are other places we cache DPI specific resources then how about we open an enhancement, subject it to prioritization, and fix 'em all at once.

I'm outta chaff.

paymicro: > we'll have to detect that the DPI changed and then run around and update glyphs?

it simple.

mattleibow: I agree this is not a blocking issue at this time, but just wanted to make sure we all knew about this possibility. And, as you mention, we possibly have the same issue in other areas - maybe we need another issue...

kingces95: Nice. So we'll open a separate "Fix multimon DPI change on UAP" and fixing `FontImageSource` can be task one?

mattleibow: Yeah. We probably can actually make this a public concept - the app may want to adjust to things like this. I was doing this for macOS for Xamarin.Essentials - a way to track when the resolution/scale changes. If the OS does not support it, then the event just never fires.

**macOS**
```csharp
NSNotificationCenter.DefaultCenter.AddObserver(NSApplication.DidChangeScreenParametersNotification, OnDidChangeScreenParameters);
```

**UWP:**
```csharp
var di = DisplayInformation.GetForCurrentView();
di.DpiChanged += OnDisplayInformationChanged;
```

samhouts:
Failing tests are unrelated to this PR

minaairsupport:
Nice feature but I think it will be better if you have a helper class like that as no one will memorize all this Glyph

https://github.com/neilkennedy/FontAwesome.Xamarin/blob/master/FontAwesome.Xamarin/FontAwesome.cs

andreinitescu:
Is it really necessary to keep an array in memory, even if it's just 4 items, having to iterate over it using a loop with temporary variables?
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/FontImageSource.cs#L47
Why not just use a `switch` like how all the other code in the framework uses

Also having `CreateBindableProperty` just in that class seems odd to me, at least why not put it in `BindableObject` or an utility class...

PR #4773, opened 2018-12-17
Add Device.IsVirtual property

kzu:
Sometimes (especially for tooling and/or telemetry purposes) it's
useful to know if your code is running on a physical device or an
emulator/simulator. Xamarin.Essentials provides this information
via the DeviceInfo class, but it requires its own separate initialization
in addition to what XF already does.

For a library targeting XF, depending on Xamarin.Essentials to provide
this information would mean taking a dependency on its package.
Consumers of the library would need to now know to initialize Essentials
properly or otherwise the library will fail. Since it would be a
transitive dependency, it wouldn't be easy to figure out what's
required to make it work.

So in order to make this information available in the easiest
possible way, a new `IsVirtual` property is added to the `Device`
class, which is set for Android and iOS in the same way Essentials does,
but using a simple boolean instead to reduce the added API surface.

### API Changes ###
Added:
 - bool IsVirtual { get; } 

### Platforms Affected ### 
- iOS
- Android

None

Not applicable

### Testing Procedure ###
Since this code is already tested as part of Xamarin.Essentials, likely not needed again here?

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Device.cs
+++ b/Xamarin.Forms.Core/Device.cs
@@ -72,6 +72,10 @@ namespace Xamarin.Forms
 		public static FlowDirection FlowDirection { get; internal set; }
 
 		[EditorBrowsable(EditorBrowsableState.Never)]
+		public static void SetIsVirtual(bool value) => IsVirtual = value;
+		public static bool IsVirtual { get; internal set; }
+
+		[EditorBrowsable(EditorBrowsableState.Never)]
 		public static bool IsInvokeRequired
 		{
 			get { return PlatformServices.IsInvokeRequired; }
--- a/Xamarin.Forms.Platform.Android/Forms.cs
+++ b/Xamarin.Forms.Platform.Android/Forms.cs
@@ -14,6 +14,7 @@ using Android.App;
 using Android.Content;
 using Android.Content.Res;
 using Android.OS;
+using Android.Provider;
 using Android.Util;
 using Android.Views;
 using Xamarin.Forms.Internals;
@@ -182,6 +183,20 @@ namespace Xamarin.Forms
 			// This could change as a result of a config change, so we need to check it every time
 			int minWidthDp = activity.Resources.Configuration.SmallestScreenWidthDp;
 			Device.SetIdiom(minWidthDp >= TabletCrossover ? TargetIdiom.Tablet : TargetIdiom.Phone);
+
+			// From https://github.com/xamarin/Essentials/blob/master/Xamarin.Essentials/DeviceInfo/DeviceInfo.android.cs
+			var isEmulator =
+				Build.Fingerprint.StartsWith("generic", StringComparison.InvariantCulture) ||
+				Build.Fingerprint.StartsWith("unknown", StringComparison.InvariantCulture) ||
+				Build.Model.Contains("google_sdk") ||
+				Build.Model.Contains("Emulator") ||
+				Build.Model.Contains("Android SDK built for x86") ||
+				Build.Manufacturer.Contains("Genymotion") ||
+				(Build.Brand.StartsWith("generic", StringComparison.InvariantCulture) && Build.Device.StartsWith("generic", StringComparison.InvariantCulture)) ||
+				Build.Product.Equals("google_sdk", StringComparison.InvariantCulture);
+
+			if (isEmulator)
+				Device.SetIsVirtual(true);
 
 			if (Build.VERSION.SdkInt >= BuildVersionCodes.JellyBeanMr1)
 				Device.SetFlowDirection(activity.Resources.Configuration.LayoutDirection.ToFlowDirection());
--- a/Xamarin.Forms.Platform.iOS/Forms.cs
+++ b/Xamarin.Forms.Platform.iOS/Forms.cs
@@ -13,6 +13,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Xamarin.Forms.Internals;
 using Foundation;
+using ObjCRuntime;
 #if __MOBILE__
 using UIKit;
 using Xamarin.Forms.Platform.iOS;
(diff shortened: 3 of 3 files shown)


StephaneDelcroix:
can you elaborate on the use case ?

kzu:
Same use cases for the Essentials one. @jamesmontemagno?--

--
@kzu from mobile

kzu:
Just thought of one: your library could use reflection.emit except when running on iOS non-virtual device, in which case it would need to fallback to slower approaches

jamesmontemagno:
I do not feel as though Xamarin.Forms should be duplicating Xamarin.Essentials functionality as the core idea is that Xamarin.Essentials is installed and used in every single application. This is why Essentials has some of the original Forms APIs for device information. It also is completely linker optimized, which is a nice perk.

In VS 2019 all templates will have Essentials installed in it.

kzu:
Yet Essentials requires additional init for Android, according to the documentation. So someone adding a package that transitively depends on Essentials would be in a broken state right after installation. 

I'm fine not duplicating, as long as XF init implicitly implies Essentials init somehow. Since we're shipping both packages, we should figure out a way for both to be initialized together and not require anything beyond the XF init (I dislike that never ending list of `Init`s that is typical in XF apps :( ). 

Also, if no duplication is the goal, maybe XF should take a dependency on Essentials and do away with the *already* duplicated code it has for detecting device properties 👍 . I could send a PR that does that instead!

StephaneDelcroix:
I tend to agree with @jamesmontemagno here, and think duplication isn't good. at some point the capability of both code will diverge and it'll be disturbing for the user

kzu:
So no answer to my last comment, at all? :-1:

StephaneDelcroix:
thanks for all the thumbs. we discussed this quickly internally, and agreed on closing it quickly instead of keeping the PR open for ages, then closing

StephaneDelcroix:
> I'm fine not duplicating, as long as XF init implicitly implies Essentials init somehow.
not possible, unless we take a dependency on Essentials, and we don't want that.

the templates could, and IMO should, include both Forms and Essentials, and Init() both

kzu:
> thanks for all the thumbs. we discussed this quickly internally, and agreed on closing it quickly instead of keeping the PR open for ages, then closing

How does that make the decision transparent at all? 

Templates do not answer the concern I raised about libs depending transitively on Essentials, and it only works for people doing File | New. 

What about the proposal to de-duplicate and have XF take a dep on Essentials? Another unanswered question. ¯\_(ツ)_/¯

I see PRs open for weeks from contributors, with back & forth conversations going for quite a bit, what's the rush here? :-1:

kzu:
I'll keep it open so it doesn't go unnoticed and forgotten.

kzu:
Ok, so de-duplicating this with Essentials [DeviceInfo](https://github.com/xamarin/Essentials/tree/master/Xamarin.Essentials/DeviceInfo) won't work since it's iOS/Android/UWP only. Since XF targets a wider set of platforms, no library author would be able to rely on Essentials to get the `IsVirtual` value without restricting himself unnecessarily to those three platforms.

So, back to the original contribution: XF would be duplicating a little bit (more, since there's already quite a bit of that in `Device.cs`) but it would benefit the whole breath of XF platforms. Going Essentials to get whether the app is running on a virtual device is a no-go for a library author. His only option would be to multi-target and do it himself, which would be way overkill IMO.

jamesmontemagno:
> Yet Essentials requires additional init for Android, according to the documentation. So someone adding a package that transitively depends on Essentials would be in a broken state right after installation.
> 
> I'm fine not duplicating, as long as XF init implicitly implies Essentials init somehow. Since we're shipping both packages, we should figure out a way for both to be initialized together and not require anything beyond the XF init (I dislike that never ending list of `Init`s that is typical in XF apps :( ).
> 
> Also, if no duplication is the goal, maybe XF should take a dependency on Essentials and do away with the _already_ duplicated code it has for detecting device properties 👍 . I could send a PR that does that instead!

For this I think:

1.) We bubble up a proper exception and only do so when the Init is needed on Android. It is not needed on any other platform and we use the Application context where we can. We heavily document the getting started and open a readme.txt when installing Xamarin.Essentials. Xamarin developers are pretty used to this Init call as it is nothing super new. In the templates in VS 2019 it adds all of the code including Init and permissions for the dev.

I think I see where you are coming from as far as a library creator writing a unit test and wanting to know this information, but easily enough that code could be written or added to the unit test itself. 

2. I think we may be able to work on something more streamlined here and expose something inside of an AppCompat library instead of having the init (although would need it for permissions probably...) @Redth is looking into this for long term.

3. I don't think XF should depend on XE or vice versa. The issue with duplication comes with naming collisions when devs are using both and we did our absolute hardest to avoid that.

kzu:
So, going back a little bit @jamesmontemagno: would you agree that knowing a device is virtual is a generally useful thing to know? (if not, may I ask why we expose it in Essentials? ;)).

Given that, and @migueldeicaza's comment that a tiny bit of duplication isn't the worst thing in the world (specially if we avoid taking a dependency on the whole Essentials just for a Boolean property), what would be a good reason *not* to have a XF-only equivalent of said useful property?

dansiegel:
Thinking on this.. having an IsVirtual property would be helpful in a lot of scenarios where it doesn’t make sense to take a dependency on XE, but there is already an inherent (transitive) dependency on XF. As an example this would be helpful when it comes to determining whether do I use a real service that might require Bluetooth or a Mock that emulates the connection.

kzu:
That's a great example @dansiegel!

jamesmontemagno:
I think the original PR conflated a few issues. The question isn't about Essentials and the Init call at all and shouldn't really be part of the discussion. It is really if having this functionality in Xamarin.Forms will add value to developers using it and how they will use it compared to having other dependencies or writing the code themselves. There will always be overlap between Xamarin.Forms and other libraries so the question is about does this API fit Xamarin.Forms and if so how should that API be constructed.

For instance this API submitted today would add a third option:
* DevicePlugin: IsDevice
* This PR: IsVirtual
* Essentials: DeviceType

samhouts:
Do our other supported Xamarin.Forms platforms have a concept of "virtual" that we can take advantage of? If not, maybe this should be implemented as a platform specific instead of a core property.

samhouts:
As far as I can tell, UWP doesn't have this functionality. So can we rework this to a set of platform specifics for Android and iOS? Thanks!

jamesmontemagno:
Xamarin.Essentials has this functionality today and is part of the default template. I think it is best to keep it there as the API could change for these devices and it wouldn't force the user to upgrade Xamarin.Forms for it.

PR #4783, opened 2018-12-18
[Android] Check if Drawable is null in Draw override on ImageButton

### LINKED ISSUE #4782: [Android] 3rd party libraries that set the ImageButton drawable to null will cause ImageButtonRenderer to crash
### Description
https://github.com/jonathanpeppers/glidex/issues/16#issuecomment-447610273

Need to just check if the Drawable is null

PureWeen:
### Description of Change ###
Forms just sets the default Drawable to "Transparent" so the Drawable is never null. 3rd party libraries that implement their own handlers won't do this though so need to check if Drawable is null. 

### Issues Resolved ### 
- fixes #4782

### Platforms Affected ### 
- Android

### Testing Procedure ###
There aren't any.  I added a custom renderer UI test to replicate setting Drawable to null but there's no way to manually test outside of just making sure changes don't appear to cause more breaks.
You can check the ImageButtonGallery just to makes sure all seems correct

### PR Checklist ###

- [x] Has automated tests 


rmarinho:
build --uitests

PR #4817, opened 2018-12-18
[UWP] Min viable FontImageSource

### LINKED ISSUE #3189: Xamarin.Forms.FontIconSource Spec

paymicro:
### Description of Change ###

see https://github.com/xamarin/Xamarin.Forms/pull/4747

### Issues Resolved ### 

- fixes #3189 for UWP

### API Changes ###

 
 None

### Platforms Affected ### 

- UWP

screenshot_1

### Testing Procedure ###

Run Font ImageSource - Legacy

### PR Checklist ###

- [ ] Has automated tests 


Xamarin.Forms.Platform.UAP/FontImageSourceHandler.cs
```diff
@@ -0,0 +1,57 @@
...
+		}
+
+		string ToWindowsFontFamily(string fontFamily)
+		{
+			var ff = fontFamily.Split("#");
+			var fontFile = ff[0];
+			var fontName = string.IsNullOrEmpty(ff[1]) || ff.Length == 1 
+				? System.IO.Path.GetFileNameWithoutExtension(fontFile) 
+				: ff[1];
+			return $"/Assets/{fontFile}#{fontName}";
```
mattleibow: This appears to suggest that it must be in the `Assets` folder? I never actually use that for app assets, rather for app icons and splash screens. Often you just create another folder for Resources or Images or Content.

Xamarin.Forms.Platform.UAP/FontImageSourceHandler.cs
```diff
@@ -0,0 +1,57 @@
...
+	{
+		public async Task<Windows.UI.Xaml.Media.ImageSource> LoadImageAsync(ImageSource imagesource,
+			CancellationToken cancelationToken = default(CancellationToken))
+		{
+			BitmapSource image = null;
+			if (imagesource is FontImageSource fontsource)
+			{
+				var iconcolor = fontsource.Color != Color.Default ? fontsource.Color : Color.White;
+				var device = CanvasDevice.GetSharedDevice();
+				var localDpi = Windows.Graphics.Display.DisplayInformation.GetForCurrentView().LogicalDpi;
```
mattleibow: Android and iOS are not so much, but UWP supports multi-scale and multi-monitor. The common case would be to move the app from one screen to another, potentially changing the resolution. We might want to handle this case to avoid distorted images.

paymicro: Ideally, yes. But here we return the bitmap. And if it is redrawn at each DPI measurement, then this greatly affects the performance. With a large number of icons, the application freezes for a couple of seconds. Therefore, I added a minimum dpi equal to 300. This should be enough so that the icon does not look soapy in most cases.

Xamarin.Forms.Platform.UAP/ImageRenderer.cs
```diff
@@ -28,9 +29,28 @@ public override SizeRequest GetDesiredSize(double widthConstraint, double height
...
+					new Size
+					{
+						Width = canvas.SizeInPixels.Width,
+						Height = canvas.SizeInPixels.Height
+					});
+			}
+			else
+			{
+				throw new InvalidCastException($"\"{Control.Source.GetType().FullName}\" is not supported.");
+			}
```
mattleibow: I would maybe put this into an extension method because I don't think this is the only place we might use this. I am thinking image buttons and other bits that use images.

kingces95:
Long live @paymicro!

samhouts:
It looks like the Win2D.uwp didn't get installed automatically for this customer. https://github.com/xamarin/Xamarin.Forms/issues/5188. Let's investigate why.

PR #4820, opened 2018-12-19
Adds some tests around different ways instantiating a type can throw

TimBarham:
### Description of Change ###

@StephaneDelcroix ... For the Forms Previewer, we'll need someway to handle this scenario. Perhaps another callback that lets us provide a replacement type when instantiating a type fails?

However, simply providing a replacement type won't be enough to sort this. For example:

1. The scenario covered by the `FormsFallbackWithUnsupportedNonDefaultConstructor` test: If we have provided a fallback type in response to the `FallbackTypeResolver` callback, it is very likely that instantiation of that fallback type will fail in non-default constructor scenarios (for example, using `x:FactoryMethod` or `x:Arguments`). In such a case, if we have a callback to handle instantiation throwing an exception, what should it return? Ultimately we probably just instantiate the fallback type using its default constructor.

2. Similarly if we allow a custom type to be created, and it throws an exception while being instantiated, and `x:FactoryMethod` or `x:Arguments` was used to instantiate it, simply returning providing a replacement type in response to that failure will very likely fail again. I can't add a test for this scenario until we have a callback that handles instantiation failures.

The possibility also exists, of course, that instantiating the replacement type will work in the above scenario (for example, the user type might derive from a Forms type and have a constructor parameter that it bases to the base type constructor). So perhaps the best approach would be that the Previewer provides Forms with a replacement type (when instantiation fails) and Forms tries to instantiate that replacement type using the standard mechanism... then if *that* still fails, simply try to instantiate using the default constructor (if that wasn't already what we were doing). And if *that* fails, I think we'd simply throw 😄 

FYI @alanmcgovern, @drewgillies

### DIFF
--- a/Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
+++ b/Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
@@ -355,7 +355,14 @@ namespace Xamarin.Forms.Xaml.UnitTests
 					<local:SettingPropertyThrows TestValue=""Test"" />
 				</ContentPage>";
 
+			var exceptions = new List<Exception>();
+			Xamarin.Forms.Internals.ResourceLoader.ExceptionHandler = exceptions.Add;
+
 			Assert.DoesNotThrow(() => XamlLoader.Create(xaml, true));
+
+			// Assuming we use ResourceLoader.ExceptionHandler to handle exceptions like this... if we use a different approach,
+			// we'll need to change this assertion...
+			Assert.That(exceptions.Count, Is.EqualTo(1));
 		}
 
 		[Test]
--- a/Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
+++ b/Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
@@ -303,7 +303,7 @@ namespace Xamarin.Forms.Xaml.UnitTests
 		}
 
 		[Test]
-		public void InstantiateWithDefaultConstructorThrows()
+		public void CanReplaceTypeWhenDefaultConstructorThrows()
 		{
 			var xaml = @"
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
```diff
@@ -303,6 +302,81 @@ public void UnknownMarkupExtensionKnownType()
...
+		{
+			var xaml = @"
+				<ContentPage xmlns=""http://xamarin.com/schemas/2014/forms""
+					xmlns:x=""http://schemas.microsoft.com/winfx/2009/xaml""
+					xmlns:local=""clr-namespace:Xamarin.Forms.Xaml.UnitTests;assembly=Xamarin.Forms.Xaml.UnitTests"">
+					<local:InstantiateThrows />
+				</ContentPage>";
+
+			XamlLoader.Create(xaml, true);
+		}
```
StephaneDelcroix: those aren't tests

StephaneDelcroix:
I'll see if can give you a hook on object creation

Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
```diff
@@ -451,4 +525,27 @@ public void CssStyleTargetingMissingTypeNotAppliedToFallbackType()
...
 	}
+
+	public class InstantiateThrows
+	{
+		public InstantiateThrows()
+		{
+			throw new InvalidOperationException();
+		}
+
+		public InstantiateThrows([Parameter("Value")] string value)
```
StephaneDelcroix: this is internal API, no one is supposed to use it

TimBarham: Oh, hah, I thought I remembered seeing it documented somewhere previously. Guess we can delete this one.

StephaneDelcroix:
those tests are now covered by #5099 and #5101

Issue #4451, opened 2018-11-17
Span's Text Property are not Changing when Binding


moustafashaban:
### Description

When I tried to bind text property to a string property in the view model, it is working like OneWayToSource, which means if I tried to notify property from the view model, nothing is changing in the view

### Steps to Reproduce

1. Add Entry control to a page
2. Bind TextProperty or PlaceHolderProperty to any string Property in your view model class
3. Change property value in view model and notify the changed (nothing will happen)

### Expected Behavior
Update the Text or PlaceHolderText value if I changed it from the view model class

### Actual Behavior
Text property or Placeholder property are not changed 

### Basic Information
Xamarin Forms Lib: 3.4.0.1
OS: this occurred in Android 8.1, not tested IOS yet

I noticed this issue in Span's Text (inside label control) also

rmarinho:
Hi , so can you add a sample showing how you are setting up the Bindings I m not being able to reproduce this.  Here's the code I used on iOS and Android 

```
	var master = new ContentPage
			{
				Title = "Master",
				BackgroundColor = Color.SkyBlue,
				Icon = "menuIcon"
			};
			var layout = new StackLayout { BackgroundColor = Color.Red };
		
			var entry = new Entry();

			entry.SetBinding(Entry.TextProperty, new Binding("Title"));
			entry.SetBinding(Entry.PlaceholderProperty, new Binding("Title"));
			entry.BindingContext = master;


			layout.Children.Add(entry);
			layout.Children.Add(new Button
			{
				Text = "Change image",
				Command = new Command(() =>
				{
					master.Title = "I changed";
				})
			});

			master.Content = layout;
```

moustafashaban:
Thanks for working on this, and sorry for my late reply
In terms of Entry I found the issue, it was in my Entry renderer I forgor the base.OnElementPropertyChanged(sender, e); in my OnElementPropertyChanged function

but the span issue is still there, 

```
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             xmlns:local="clr-namespace:XF_SpanBindTest"
             x:Class="XF_SpanBindTest.MainPage">

    <StackLayout VerticalOptions="Center" HorizontalOptions="Center">
        <Label Text="{Binding Text1}" TextColor="Green"/>
        <Label TextColor="Blue">
            <Label.FormattedText >
                <FormattedString>
[...]

Issue #2177, opened 2018-03-24
Span Binding not updating even after OnPropertyChanged


smartprogrammer93:
### Description
Span Binding not updating even after OnPropertyChanged
### Steps to Reproduce

1. ViewModel with INotifyPropertyChanged
2. Bind Span inside formated text inside lable to a property
3. change the value of the property and call OnPropertyChanged

### Expected Behavior
the value updates
### Actual Behavior
the value stays the same
### Basic Information

- Version with issue: 3.1.0.309646
- Last known good version:
- IDE: visual studio

@adamped Please check this one out since you did the #1850 PR for this

Thanks

adamped:
@smartprogrammer93 - by default, the Span Text has OneTime Binding. Set the Binding Mode=OneWay and it will work.

It was recommended to default that way based on the premise that Span's are not likely to be updated, but I am not really convinced that would be the case anymore, or there are going to be a lot of support cases or issues around developers not knowing this @jassmith

smartprogrammer93:
@adamped 

Thank you for your fast reply. I admit, i didn't try setting the binding Mode. but i do believe that having the default mode to be one time is a bit confusing. as the binding for a label is OneWay, and the span is just part of the label.

But that is just my personal opinion, I am sure you have your reasons to making it OneTime.

I will leave the issue open in case you would like to discuss this further. if not, please feel free to close it.

davidsaldana:
Yep, the OneWay fixed it for me...

bnserig:
+1
Default should be OneWay.

rbrian:
+1 oneway would help

limefrogyank:
Funny, this just got me, too.  I think OneWay would be better... but as long as this issue is searchable, it's an easy fix.

TheTrainor:
This has bitten me so many times. OneWay should be the default.

andreinitescu:
`Text` property has the default to `OneWay`, so I also think from a practical point of view, the default should be to `OneWay` too. The majority of scenarios for using binding on Span is when data is not available when Label is created.

But like a wise man has said many times _The ship has sailed_, I'm not sure if the default can be changed.
I don't think it would be a (much) breaking change however to change default from `OneTime` to `OneWay`, it would break few scenarios where value is not expected to be updated by the Span the 2nd time after set.
 An alternate solution would be to have a static default value which can be modified: `Span.TextDefaultBindingMode = BindingMode.OneWay`

PR #4850, opened 2018-12-23
Changed Span default binding to OneWay

### LINKED ISSUE #3167: Proposal: Change default binding mode of Span Text property to OneWay
Xamarin.Forms [3.1](https://developer.xamarin.com/releases/xamarin-forms/xamarin-forms-3.1/3.1.0/) [introduced BindableSpan](https://github.com/xamarin/Xamarin.Forms/pull/1850) 👍.  
As part of that work is was [suggested](https://github.com/xamarin/Xamarin.Forms/pull/1850#issuecomment-366139567) that the default BindingMode be set to OneTime because OneTime bindings are cheaper. 

I would like to start a discussion, and propose switching the default binding mode to be OneWay for the [Span.TextProperty](https://github.com/xamarin/Xamarin.Forms/blob/5112f0928fc377550a5dd58213b520a0f6cae518/Xamarin.Forms.Core/Span.cs#L53-L54). There are also the other `Color` properties on Span as well that are currently defaulting to OneTime. I would also suggest that these be changed as well; however that is a much smaller concern.

I do realize that I can set the Mode on all of my Span's Bindings to be OneWay and get the behavior I am describing. However, this issue is not about saving a few keystrokes. It is about lost development time, trying to debug binding issues that arise from unexpected behavior.
I am also aware that there was at least [one other issue](https://github.com/xamarin/Xamarin.Forms/issues/2177) raised, where this unexpected default caused a problem. 

## The Common Case 
First I would like to make the case, that in general if I am applying a `{Binding}` to a Span OneWay is the most common use case. Broadly speaking there are only two common sources for the `{Binding}`. 
- Bind to the property on another control. I can come up with many examples where this property would be expected to change. Binding to the text of an Entry field. Binding to the selected item in a ListView. Binding to the current position of the CaroselView. Etc. I struggle to come up with very many cases where I am using a Binding to effectively link two pieces of text. For most of those situations, I would merely put the shared text as a resource (referenced with StaticResource) and avoid a Binding altogether.
- Bind to a property on a BindingContext (ViewModel). For the OneTime binding to be useful here, it would require that the property is populated when the object is bound, and never change. I can imagine a few use cases binding to POCO (no INotifyPropertyChanged - INPC) where this might be the case. Such as binding to model straight out of Entity Framework. There is also the possibility that the Span is used as part of a DataTemplate, in which case the object is likely populated when it is bound, and replacing the entire instance with a new one, effectively causes the view to update. However, I would propose that this is the exception case and not the rule. Binding to a POCO is nice, but the times where merely binding to its properties once are few and far between. Data needs to be updated. Data is commonly retrieved asynchronously. Simply updating a view model and raising INotifyPropertyChanged is what is expected.

## Performance
Currently it looks as if the only performance gains from using BindingMode.OneTime come [when compiling your XAML](https://github.com/xamarin/Xamarin.Forms/blob/130dfdf8439a0e1f5e4b9762f21207c6af8c13e2/Xamarin.Forms.Build.Tasks/SetPropertiesVisitor.cs#L413-L415). I do not want to downplay the importance of compiled XAML or the performance gains from not needing to create and emit those handlers; compiled XAML is the default and I typically see it as the default starting point. However, unless I am missing something, at run time (assuming non-compiled XAML or items created in code) there is [little difference between a OneWay and OneTime binding](https://github.com/xamarin/Xamarin.Forms/blob/11b1496db5a0a6b95afaf3b50c66b6bfddfd74c0/Xamarin.Forms.Core/BindingExpression.cs#L109). From this perspective, the [performance gains](https://www.xkcd.com/221/) do not seem worth the loss of the time and functionality. Have I missed anything here?

## Familiarity with other XAML frameworks
For what it is people familiar with other XAML frameworks will also likely expect OneWay (minimum) binding as a default. I am hesitant to use the words "XAML Standard" but the spirit of it is here. Changing the default binding mode to OneWay will align with what is already done in the other frameworks.

- [WPF Run](https://docs.microsoft.com/en-us/dotnet/api/system.windows.documents.run.text?view=netframework-4.7.2#System_Windows_Documents_Run_Text): Defaults to [TwoWay](https://referencesource.microsoft.com/#PresentationFramework/src/Framework/System/Windows/Documents/Run.cs,105). When using within a TextBlock (the most common case) this deviates from the normal WPF behavior of TwoWay defaults for properties that can change due to user interaction, and OneWay for everything else. 
- [UPW Run](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.documents.run): In general UWP defaults `{Binding}` to [OneWay](https://docs.microsoft.com/en-us/windows/uwp/data-binding/data-binding-in-depth#xbind-and-binding-feature-comparison), except in the case the [TextProperty](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.data.bindingmode#remarks). 
- Xamarin.Forms: Prior to the new bindable Span, the only other property in the framework that defaults to OneTime is [Entry.IsTextPredictionEnabledProperty](https://github.com/xamarin/Xamarin.Forms/blob/cd6b01832c0e58aa02f7d107026982593683feea/Xamarin.Forms.Core/Entry.cs#L36:43). I believe this deviation from the norm in the rest of the the framework, is also contributing to the confusion.

## Comparison with `x:Bind`
`x:Bind` [is coming](https://github.com/xamarin/Xamarin.Forms/projects/7) for Xamarin.Forms. With it, I would suggest, comes the opportunity to reset developer's Binding expectations. In [UWP](https://docs.microsoft.com/en-us/windows/uwp/data-binding/data-binding-in-depth#xbind-and-binding-feature-comparison) "`{x:Bind}` defaults to OneTime; `{Binding}` defaults to OneWay". Those already familiar with UWP XAML know this, and I would suggest that keeping Xamarin.Forms inline with this pattern will help meet developer's expectations. 

## Relevant parties/stakeholders 
@adamped 
@jassmith

andreinitescu:
### Description of Change ###

Many have been bitten, including the undersigned, because of Span's default binding to OneTime:
See #2177, but also https://github.com/xamarin/Xamarin.Forms/issues/4451#issuecomment-449577935

A very good and lengthy description is in #3167

### Issues Resolved ### 

- fixes #3167

### API Changes ###
Changed TextProperty default binding mode from OneTime to OneWay.

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Span.cs
+++ b/Xamarin.Forms.Core/Span.cs
@@ -15,7 +15,7 @@ namespace Xamarin.Forms
 		}
 
 		public static readonly BindableProperty StyleProperty = BindableProperty.Create(nameof(Style), typeof(Style), typeof(Span), default(Style),
-			propertyChanged: (bindable, oldvalue, newvalue) => ((Span)bindable)._mergedStyle.Style = (Style)newvalue, defaultBindingMode: BindingMode.OneTime);
+			propertyChanged: (bindable, oldvalue, newvalue) => ((Span)bindable)._mergedStyle.Style = (Style)newvalue, defaultBindingMode: BindingMode.OneWay);
 
 		public static readonly BindableProperty TextDecorationsProperty = DecorableTextElement.TextDecorationsProperty;
 
@@ -26,7 +26,7 @@ namespace Xamarin.Forms
 		}
 
 		public static readonly BindableProperty BackgroundColorProperty
-			= BindableProperty.Create(nameof(BackgroundColor), typeof(Color), typeof(Span), default(Color), defaultBindingMode: BindingMode.OneTime);
+			= BindableProperty.Create(nameof(BackgroundColor), typeof(Color), typeof(Span), default(Color), defaultBindingMode: BindingMode.OneWay);
 
 		public Color BackgroundColor
 		{
@@ -55,7 +55,7 @@ namespace Xamarin.Forms
 #pragma warning restore 618
 
 		public static readonly BindableProperty TextProperty
-			= BindableProperty.Create(nameof(Text), typeof(string), typeof(Span), "", defaultBindingMode: BindingMode.OneTime);
+			= BindableProperty.Create(nameof(Text), typeof(string), typeof(Span), "", defaultBindingMode: BindingMode.OneWay);
 
 		public string Text
 		{


StephaneDelcroix:
this has a huge perf implication

andreinitescu:
Changes requested implemented.

PR #4697, opened 2018-12-10
[UITests] CollectionView

rmarinho:
### Description of Change ###

Add set of cross platform UITests tests for CollectionView galleries 

### PR Checklist ###

- [ ] Has automated tests 
- [ ] Rebased on top of the target branch at time of PR
- [ ] Changes adhere to coding standard

PR #4865, opened 2018-12-26
[iOS] CollectionView item replacement on iOS

hartez:
### Description of Change ###

Replace items on iOS

### Issues Resolved ### 

- implements item replacement from #3172

### API Changes ###

 None

### Platforms Affected ### 

- iOS

### Testing Procedure ###

Automated tests will be available once #4697 is merged.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs
+++ b/Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs
@@ -30,6 +30,7 @@ namespace Xamarin.Forms.Platform.iOS
 					Remove(args);
 					break;
 				case NotifyCollectionChangedAction.Replace:
+					Replace(args);
 					break;
 				case NotifyCollectionChangedAction.Move:
 					Move(args);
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs
```diff
@@ -41,6 +41,14 @@ void CollectionChanged(object sender, NotifyCollectionChangedEventArgs args)
 			}
 		}
 
+		private void Replace(NotifyCollectionChangedEventArgs args)
+		{
+			var startIndex = args.NewStartingIndex > -1 ? args.NewStartingIndex : _itemsSource.IndexOf(args.NewItems[0]);
+			var count = args.NewItems.Count;
```
paymicro: This code repeats in `Add` and `Remove` methods. Perhaps it be placed in `CreateIndexesFrom` func or make another one?

Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs
```diff
@@ -48,6 +49,14 @@ void Move(NotifyCollectionChangedEventArgs args)
 
 			_collectionView.MoveItem(oldPath, newPath);
 		}
+		
+		private void Replace(NotifyCollectionChangedEventArgs args)
+		{
+			var startIndex = args.NewStartingIndex > -1 ? args.NewStartingIndex : _itemsSource.IndexOf(args.NewItems[0]);
```
kingces95: Huh. Why don't we throw here? What does a negative index mean? What does it mean if the `NewStartingIndex` is not the index of the first element of `NewItems`? Isn't that what it means to replace? That the index is the same for the old and new item? Anyway, I'm sure there's a good reason but for future spelunkers a note would be nice.

PR #4177, opened 2018-10-22
[Android] Java.Lang.RuntimeException: Animators may only be run on Lo…

JTOne123:
[Android] Java.Lang.RuntimeException: Animators may only be run on Looper threads on Battery Saver Mode

### Description of Change ###
Animators may only be run on Looper threads

### Issues Resolved ### 
- fixes #4176 

### Platforms Affected ### 
- Android

### Behavioral/Visual Changes ###
None

PR #4866, opened 2018-12-27
[Android] Only post DisableTimer to the queue if it's not on the UI thread

hartez:
### Description of Change ###

PR #4177 attempts to fix an issue with disabling the Android timer on a non-UI thread by using BeginInvokeOnMainThread, which posts the action to the main looper to be run later on the UI thread. 

But if the call to DisableTimer is happening because the Ticker itself has finished all of the pending actions, DisableTimer needs to be called immediately in order to end the animations so that awaited animation calls can return. Since the animations are running on the UI thread to begin with, these calls to DisableTimer are _already_ on the UI thread; they don't _need_ to be posted there for later invocation.

This change adds a check in DisableTimer to see if it's already on the UI thread; if so, it is executed immediately. Otherwise, it is posted to the UI thread.

### Issues Resolved ### 

- fixes bug where awaited animations never finish (UI Test 39821)

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

None

Not applicable

### Testing Procedure ###

Run UI Test 39821; without the fix, it will never finish (i.e., the "Success" label will never be displayed).

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/AndroidTicker.cs
+++ b/Xamarin.Forms.Platform.Android/AndroidTicker.cs
@@ -58,10 +58,17 @@ namespace Xamarin.Forms.Platform.Android
 
 		protected override void DisableTimer()
 		{
-			Device.BeginInvokeOnMainThread(new Action(() =>
+			if (Device.IsInvokeRequired)
+			{
+				Device.BeginInvokeOnMainThread(new Action(() =>
+				{
+					_val?.Cancel();
+				}));
+			}
+			else
 			{
 				_val?.Cancel();
-			}));
+			}
 		}
 
 		protected override void EnableTimer()


hartez:
> would it be better to put this logic inside the begin invoke? 

Which bit do you mean when you say "this logic"?

mattleibow:
The logic for if on main then run else use invoke. Basically the if check.

hartez:
> The logic for if on main then run else use invoke. Basically the if check.

Oh, I get what you're saying. So changing the Android implementation of `BeginInvokeOnMainThread` to something like

```
if(IsInvokeRequired) { 
    s_handler.Post(action); 
} else {
    action(); // we're already on the main thread, just go ahead a run it instead of posting to the queue
}
```

That would definitely break a few things. Most notably around [layout](https://github.com/xamarin/Xamarin.Forms/blob/c3da7dbb34a16e702bcd3b186646a88c7fb7b75c/Xamarin.Forms.Core/Layout.cs#L362), where there's logic which depends on putting the full layout pass onto queue. There are a couple of other places in the Android renderers which rely on that behavior (letting current operations finish before continuing). 

It would also make Android behave very differently from iOS and UWP (and presuimably the other platforms); they also post the action to the queue to be run later.

PR #4883, opened 2019-01-02
[UWP] Implementation of Switch.OnColor

### LINKED ISSUE #4460: [UWP] Switch color is not set
### Description

3.1.0 introduced the ability to set the color of a switch. This isn't working for UWP.

### Steps to Reproduce
```
<Switch OnColor="Purple" IsToggled="true" />
```

### Expected Behavior

Switch color is purple.

### Actual Behavior

Switch color is blue.

### Basic Information

### Screenshots

image

Android showing the purple:

image

### Reproduction Link

https://github.com/davidortinau/TheLittleThingsPlayground/blob/master/TheLittleThingsPlayground/Views/ThreeOnePage.xaml#L56

jfversluis:
### Description of Change ###

Implements the `Switch.OnColor` for UWP which was reported as a bug to be not working

### Issues Resolved ### 

- fixes #4460

### API Changes ###

 
 None

### Platforms Affected ### 

- UWP

`Switch.OnColor` now actually works on UWP!

After: 

screenshot 2019-01-02 15 38 41

### Testing Procedure ###

Added a case for this in the Switch Gallery app

untitled

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
@@ -16,7 +16,6 @@ namespace Xamarin.Forms.Platform.UWP
 		protected override void OnElementChanged(ElementChangedEventArgs<Switch> e)
 		{
 			base.OnElementChanged(e);
-			
 			if (e.NewElement != null)
 			{
 				if (Control == null)
--- a/Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
@@ -10,8 +10,8 @@ namespace Xamarin.Forms.Platform.UWP
 {
 	public class SwitchRenderer : ViewRenderer<Switch, ToggleSwitch>
 	{
-		private object _originalOnColor;
-		private Brush _originalOnColorBrush;
+		object _originalOnColor;
+		Brush _originalOnColorBrush;
 
 		protected override void OnElementChanged(ElementChangedEventArgs<Switch> e)
 		{
@@ -23,7 +23,7 @@ namespace Xamarin.Forms.Platform.UWP
 				{
 					var control = new ToggleSwitch();
 					control.Toggled += OnNativeToggled;
-					control.Loaded += Control_Loaded;
+					control.Loaded += OnControlLoaded;
 					control.ClearValue(ToggleSwitch.OnContentProperty);
 					control.ClearValue(ToggleSwitch.OffContentProperty);
 
@@ -34,12 +34,6 @@ namespace Xamarin.Forms.Platform.UWP
 
 				UpdateFlowDirection();
 			}
-		}
-
-		private void Control_Loaded(object sender, RoutedEventArgs e)
-		{
-			UpdateOnColor();
-			Control.Loaded -= Control_Loaded;
 		}
 
 		protected override void OnElementPropertyChanged(object sender, PropertyChangedEventArgs e)
@@ -59,6 +53,12 @@ namespace Xamarin.Forms.Platform.UWP
 		}
 
 		protected override bool PreventGestureBubbling { get; set; } = true;
+
+		void OnControlLoaded(object sender, RoutedEventArgs e)
+		{
+			UpdateOnColor();
+			Control.Loaded -= OnControlLoaded;
+		}
 
 		void OnNativeToggled(object sender, RoutedEventArgs routedEventArgs)
 		{
@@ -98,7 +98,7 @@ namespace Xamarin.Forms.Platform.UWP
 							if (_originalOnColor == null)
 								_originalOnColor = frame.Value;
 
-							if (Element.IsSet(Switch.OnColorProperty) && !Element.OnColor.IsDefault)
+							if (!Element.OnColor.IsDefault)
 								frame.Value = new SolidColorBrush(Element.OnColor.ToWindowsColor()) { Opacity = .7 };
 							else
 								frame.Value = _originalOnColor;
@@ -108,12 +108,12 @@ namespace Xamarin.Forms.Platform.UWP
 				}
 			}
 
-			var rect = Control.GetDescendantsByName<Windows.UI.Xaml.Shapes.Rectangle>("SwitchKnobBounds").FirstOrDefault();
+			var rect = Control.GetDescendantsByName<Windows.UI.Xaml.Shapes.Rectangle>("SwitchKnobBounds").First();
 
 			if (_originalOnColorBrush == null)
 				_originalOnColorBrush = rect.Fill;
 
-			if (Element.IsSet(Switch.OnColorProperty) && !Element.OnColor.IsDefault)
+			if (!Element.OnColor.IsDefault)
 				rect.Fill = new SolidColorBrush(Element.OnColor.ToWindowsColor());
 			else
 				rect.Fill = _originalOnColorBrush;
--- a/Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
@@ -10,7 +10,7 @@ namespace Xamarin.Forms.Platform.UWP
 {
 	public class SwitchRenderer : ViewRenderer<Switch, ToggleSwitch>
 	{
-		object _originalOnColor;
+		Brush _originalOnHoverColor;
 		Brush _originalOnColorBrush;
 
 		protected override void OnElementChanged(ElementChangedEventArgs<Switch> e)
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
```diff
@@ -53,5 +69,54 @@ void UpdateFlowDirection()
...
+
+					foreach (var timeline in state.Storyboard.Children.OfType<ObjectAnimationUsingKeyFrames>())
+					{
+						var property = Storyboard.GetTargetProperty(timeline);
+						var target = Storyboard.GetTargetName(timeline);
+						if ((target == "SwitchKnobBounds") && (property == "Fill"))
+						{
+							var frame = timeline.KeyFrames.First();
+
+							if (_originalOnColor == null)
```
jfversluis: Had to save the original value(s), because `ClearValue` would cause the control to disappear

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
```diff
@@ -53,5 +69,54 @@ void UpdateFlowDirection()
...
+								_originalOnColor = frame.Value;
+
+							if (Element.IsSet(Switch.OnColorProperty) && !Element.OnColor.IsDefault)
```
jfversluis: The additional `IsDefault` check is needed (unfortunately) since just using `IsSet` when setting `Color.Default` as a value, would still return `true` causing the control to disappear

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
```diff
@@ -28,6 +36,12 @@ protected override void OnElementChanged(ElementChangedEventArgs<Switch> e)
 			}
 		}
 
+		private void Control_Loaded(object sender, RoutedEventArgs e)
```
jfversluis: This has to be set when the control is fully loaded, else it will crash/will not have any effect

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
```diff
@@ -28,6 +36,12 @@ protected override void OnElementChanged(ElementChangedEventArgs<Switch> e)
...
+		{
+			UpdateOnColor();
+			Control.Loaded -= Control_Loaded;
```
StephaneDelcroix: not sure this is required

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
StephaneDelcroix: that's how it's done elsewhere

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
StephaneDelcroix: the `color.IsDefault` is probably enough

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
```diff
@@ -53,5 +69,54 @@ void UpdateFlowDirection()
...
+						var target = Storyboard.GetTargetName(timeline);
+						if ((target == "SwitchKnobBounds") && (property == "Fill"))
+						{
+							var frame = timeline.KeyFrames.First();
+
+							if (_originalOnColor == null)
+								_originalOnColor = frame.Value;
+
+							if (Element.IsSet(Switch.OnColorProperty) && !Element.OnColor.IsDefault)
+								frame.Value = new SolidColorBrush(Element.OnColor.ToWindowsColor()) { Opacity = .7 };
```
paymicro: why 70% opacity?

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
```diff
@@ -53,5 +69,54 @@ void UpdateFlowDirection()
...
+							break;
+						}
+					}
+				}
+			}
+
+			var rect = Control.GetDescendantsByName<Windows.UI.Xaml.Shapes.Rectangle>("SwitchKnobBounds").FirstOrDefault();
+
+			if (_originalOnColorBrush == null)
+				_originalOnColorBrush = rect.Fill;
```
paymicro: if `rect` cannot be null, then you should call `First` instead `FirstOrDefault` in LINQ

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
paymicro: Could yo please drop `private` and rename to `OnControlLoaded`.

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
```diff
@@ -1,21 +1,29 @@
...
+using Windows.UI.Xaml.Media;
+using Windows.UI.Xaml.Media.Animation;
+using Windows.UI.Xaml.Shapes;
 
 namespace Xamarin.Forms.Platform.UWP
 {
 	public class SwitchRenderer : ViewRenderer<Switch, ToggleSwitch>
 	{
+		private object _originalOnColor;
+		private Brush _originalOnColorBrush;
```
paymicro: drop `private`

jfversluis:
Seen the comments, need to find some time to process them... Will do that ASAP! Thanks!

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
jfversluis: Then I'm confused. In another place, I had to use `IsSet`. What is the use-case for one or the other?

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
jfversluis: whoops... Done

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
jfversluis: You mean using the `Loaded` event? I see in the `SwitchCell` (https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.UAP/CellControl.cs#L106) that it happens in the measure override. Would you rather have it there?

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
jfversluis: I have taken this (with permission) from the Xamarin Community Toolkit (https://github.com/xamarin/XamarinCommunityToolkit/blob/master/Toolkit/Effects/Switch/SwitchChangeColor.uwp.cs).

This is probably because this is the default. @Depechie ?

I have another PR with the selected or unselected tab colors (https://github.com/xamarin/Xamarin.Forms/pull/4899/files#diff-a3c5e562833b067b956c34f89bd0ebbbR581), maybe implement it more like that? Capturing the default unselected color first?

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
jfversluis: Done

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
jfversluis: Done

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
Depechie: Good question, actually not sure what the default look & feel is for UWP.
If you are indeed able to get the setting from somewhere, I guess that will be better than using a fixed value.

samhouts:
@jfversluis Do you have more changes pending, or is this ready for re-review? Thanks!

jfversluis:
Removed hardcoded opacity, now also taken from the original color. Ready for re-review now!

Xamarin.Forms.Platform.UAP/SwitchRenderer.cs
```diff
@@ -40,10 +47,18 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
 			{
 				UpdateFlowDirection();
 			}
+			else if (e.PropertyName == Switch.OnColorProperty.PropertyName)
+				UpdateOnColor();
 		}
 
 		protected override bool PreventGestureBubbling { get; set; } = true;
 
+		void OnControlLoaded(object sender, RoutedEventArgs e)
```
yrest: Hi guys, just wondering if somebody can help me to figure out what am I doing wrong getting this exception after update?

image

paymicro: It seems that `Control.GetChildren<Windows.UI.Xaml.Controls.Grid>().FirstOrDefault()` returns `null` for some reason.

yrest: @paymicro thanks for prompt reply! This is a piece of xaml I'm using (first Switch is used to activate the second)
            < StackLayout Orientation="Horizontal" >
                < Switch x:Name="FirstSwitch" IsToggled="{Binding **FirstToggled**}" / >
            < /StackLayout >
            < StackLayout Orientation="Horizontal" IsVisible="{Binding **FirstToggled**}" >
                < Switch x:Name="SecondSwitch" IsToggled="{Binding SecondToggled}" / >
            < /StackLayout >

It seems if I use the same variable, used as IsToggled binding, somewhere else, in this case in the IsVisible binding, it breaks...
Have just created brand new XF App1, updated from nuget to the latest version from yesterday, added same xaml and it crashes.
Actually only rolling back to 3.5.0.169047 works for me, for the two 3.6 in nuget it seems to crash for such cases.

Issue #2118, opened 2018-03-16
Toolbar icon not shown on App startup


TheSprigg:
Bug report best practices: https://github.com/xamarin/Xamarin.Forms/wiki/Submitting-Issues

### Description
"SetSupportActionBar()" has to be set to be able to intercept "OnOptionsItemSelected()". When not setting "SetSupportActionBar()" the function "OnOptionsItemSelected()" in MainActivity will not be triggered.

Unfortunately after adding this line of code the toolbar item is not shown at the first page after startup (MainPage). The toolbar item will appear after navigating to a different page and back to MainPage.

### Steps to Reproduce
1. Start App. 
Expected: Toolbar item "About" should be present 
Actual: No toolbar item is present

2. Navigate to second page via button

3. Navigate back via toolbar. This works and "OnOptionsItemSelected()" is triggered. 
Expected and Actual: Toolbar item "About" is now shown on MainPage and is working.

### Basic Information
- Version with issue: 2.5.0.280555

- Last known good version:
- IDE:
- Platform Target Frameworks:
  - Android: 8.1 (but bug was in earlier releases, too
- Android Support Library Version: 27.0.2
- Affected Devices: Tested on Samsung Galaxy Tab S2 (Android 7.0), Google Pixel (Android 8.0)

### Reproduction Link
https://github.com/TheSprigg/Toolbarbutton_not_shown/

davidW83:
I have a pseudosplashscreen for that matter as a workaround. 
Funny thing is i'm doing som heavy workload with it, so it is actually good to have it now.

But this even happens on very rare occaions with the xaml ToolbarItem aswell, i never could reproduce it though...
And to be honest, alot issues i've encountered with Xamarin are posted here as bugs months later. :)

Paul-Brenner-Tangoe:
Just hit this one as well. @davidW83 can you expand on how you implemented your workaround?

Yves-Be:
Encountered the same problem as well.

Worked around it by calling this in the OnAppearing of the first page:
```
/// <summary>
/// Workaround for ToolbarItems not apearing on android (https://github.com/xamarin/Xamarin.Forms/issues/2118)
/// </summary>
private void FixToolBarItemsNotAppearingOnAndroid()
{
    if (Device.RuntimePlatform == Device.Android)
    {
        Task.Run(async () =>
        {
            await Task.Delay(1000);
            var items = ToolbarItems.ToList();
[...]

Issue #1944, opened 2018-02-22
Can't Intercept Software Back Button


AceCoderLaura:
### Description
Software back button doesn't use the same code paths as the hardware back button. This leads to unexpected behaviour on Android (and from what I've read, iOS).

### Steps to Reproduce

1. Override the "OnBackButtonPressed()" method on MasterDetailPage.
2. Run application on Android and navigate to a detail page in the MasterDetailPage.
3. Press the "Back" arrow on the command bar.

### Expected Behavior
The "OnBackButtonPressed()" method should be invoked, same as when the hardware "back" button is pressed.

### Actual Behavior
The method is not invoked.

### Basic Information

- Version with issue:
- Last known good version: Unknown
- IDE: Visual Studio 2017
- Platform Target Frameworks:
  - Android: 8.0 Oreo
  - UWP: UWP 16299 (Works fine!)
- Android Support Library Version: v7/v4
- Nuget Packages:

> NETStandard.Library                 {2.0.1}                                  SQLite.Net.Standard                                                                 
> System.ServiceModel.Primitives      {4.4.1}                                  Adapt.Model.Helpdesk                                                                
> NETStandard.Library                 {2.0.1}                                  Adapt.Model.Helpdesk                                                                
> System.Runtime.Serialization.Pri... {4.3.0}                                  Adapt.Model.Helpdesk                                                                
> System.ServiceModel.Http            {4.4.1}                                  Adapt.XivicClient.Standard                                                          
> System.Reflection.TypeExtensions    {4.4.0}                                  Adapt.XivicClient.Standard                                                          
> System.ServiceModel.Primitives      {4.4.1}                                  Adapt.XivicClient.Standard                                                          
> NETStandard.Library                 {2.0.1}                                  Adapt.XivicClient.Standard                                                          
> System.Runtime.Serialization.Pri... {4.3.0}                                  Adapt.XivicClient.Standard                                                          
> System.ServiceModel.Primitives      {4.4.1}                                  Adapt.Model.Common.Standard                                                         
> NETStandard.Library                 {2.0.1}                                  Adapt.Model.Common.Standard                                                         
> System.Runtime.Serialization.Pri... {4.3.0}                                  Adapt.Model.Common.Standard                                                         
> NETStandard.Library                 {2.0.1}                                  Adapt.Presentation.Standard                                                         
> Xamarin.Forms                       {2.5.0.280555}                           Adapt.Presentation.Standard                                                         
[...]

Issue #1413, opened 2017-12-18
[Android] Issue for triggering an event when the ActionBar's back button is being tapped


toussam:
### Description

I am trying to trigger an event when the user is tapping on the ActionBar's back button. After some research, I found out that I have to overwrite the method OnOptionsItemSelected into my MainActivity, but when the MainPage is set with a MasterDetailPage, this last one is not being called.

For information, I am also overriding the OnPostCreate method to call SetSupportActionBar since that I am using FormsAppCompatActivity and that the loaded toolbar was null if it was loaded into the OnCreate one.

### Steps to Reproduce

1. Set a breakpoint into MainActivity, line 36
2. Launch the application
3. Click on the "CLICK HERE" button to be redirect to the SecondPage
4. Click on the ActionBar's back button

### Expected Behavior

The breakpoint shoud be fired.

### Actual Behavior

The breakpoint is not being fired when doing so.

### Basic Information

- Version with issue: 2.4.0.280
- IDE: Visual Studio 2017
- Android Target Framework: minSdkVersion is the 15th and the used one is the 26th
- Android Support Library Version: v25.4.0.2

### Reproduction Link

https://github.com/toussam/BackButtonIssueSample

toussam:
Hi Samantha, hi Jason,

Is there any news about this enhancement ? I saw that it was added to the backlog before being removed from it.

Thank you in advance for your response and wish you a great day !
Best regards,
Mathias

jfversluis:
Hey there! So sorry this never received a response, that is not really great... It seems that there isn't a lot of traction here and together with the fact that Xamarin.Forms is now in maintenance mode, I will be closing this one. Hope you get a chance to check out .NET MAUI!

PR #4893, opened 2019-01-04
[Android] wire up the correct toolbar to SetSupportActionBar

### LINKED ISSUE #4827: OnOptionsItemSelected() not called after a configuration/orientation change.
### Description

In my application I have to intercept the software back button click and it works fine as long as I don't change the orientation. After changing the orientation from portait to landscape the method OnOptionsItemSelected() is never called again.

```
[Activity(Label = "TestApp", Icon = "@mipmap/icon", Theme = "@style/MainTheme", MainLauncher = true, ConfigurationChanges = ConfigChanges.ScreenSize | ConfigChanges.Orientation)]
    public class MainActivity : global::Xamarin.Forms.Platform.Android.FormsAppCompatActivity
{
     {
            TabLayoutResource = Resource.Layout.Tabbar;
            ToolbarResource = Resource.Layout.Toolbar;
            base.OnCreate(savedInstanceState);

            global::Xamarin.Forms.Forms.Init(this, savedInstanceState);
            LoadApplication(new MyApp());
     }
     protected override void OnPostCreate(Bundle savedInstanceState)
     {
            base.OnPostCreate(savedInstanceState);
            Android.Support.V7.Widget.Toolbar toolbar = this.FindViewById<Android.Support.V7.Widget.Toolbar>(Resource.Id.toolbar);
            SetSupportActionBar(toolbar);
      }

      public override bool OnOptionsItemSelected(IMenuItem item)
      {
            return base.OnOptionsItemSelected(item);
      }
}
```

styles.xml
```
<resources>

  <style name="MainTheme" parent="MainTheme.Base">
  </style>
  <!-- Base theme applied no matter what API -->
  <style name="MainTheme.Base" parent="Theme.AppCompat.Light.DarkActionBar">
    <!--If you are using revision 22.1 please use just windowNoTitle. Without android:-->
    <item name="windowNoTitle">true</item>
    <!--We will be using the toolbar so no need to show ActionBar-->
    <item name="windowActionBar">true</item>
    <!-- Set theme colors from http://www.google.com/design/spec/style/color.html#color-color-palette -->
    <!-- colorPrimary is used for the default action bar background -->
    <item name="colorPrimary">#2196F3</item>
    <!-- colorPrimaryDark is used for the status bar -->
    <item name="colorPrimaryDark">#1976D2</item>
    <!-- colorAccent is used as the default value for colorControlActivated
         which is used to tint widgets -->
    <item name="colorAccent">#FF4081</item>
    <!-- You can also set colorControlNormal, colorControlActivated
         colorControlHighlight and colorSwitchThumbNormal. -->
    <item name="windowActionModeOverlay">true</item>

    <item name="android:datePickerDialogTheme">@style/AppCompatDialogStyle</item>
  </style>

  <style name="AppCompatDialogStyle" parent="Theme.AppCompat.Light.Dialog">
    <item name="colorAccent">#FF4081</item>
  </style>
</resources>
```

toolbar.axml

```
<android.support.v7.widget.Toolbar
    xmlns:android="http://schemas.android.com/apk/res/android"
    android:id="@+id/toolbar"
    android:layout_width="match_parent"
    android:layout_height="wrap_content"
    android:background="?attr/colorPrimary"
    android:theme="@style/ThemeOverlay.AppCompat.Dark.ActionBar"
    android:popupTheme="@style/ThemeOverlay.AppCompat.Light" />
```

### Steps to Reproduce

1. Start the app
2. Change orientation
3. OnOptionsItemSelected() not called an backbutton 

### Basic Information
Microsoft Visual Studio Enterprise 2017 
Version 15.9.4

Xamarin   4.12.3.77 
Xamarin.Android SDK   9.1.4.2 

- Version with issue:
- IDE:
- Platform Target Frameworks: 
  - Android: 7.0
- Nuget Packages: 
Xamarin.Forms 3.4.0.1008975
Xamarin.Android.Support.... 28.0.0
- Affected Devices: Samsung Galaxy S5

PureWeen:
### Description of Change ###
I finished up this PR knowing it would be slightly breaking but then realized it causes the back button on Android to operate differently than iOS which maybe was intentional?

is the purpose of this code
https://github.com/xamarin/Xamarin.Forms/blob/78385f9fc1fc56dc88bd98e73bf9c8f2f2d0a90a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs#L160-L170

To make it so that the software back button doesn't fire the *OnBackButtonPressed* of the contentpage on Android?

The Visible toolbar is created on the NavigationRenderer
https://github.com/xamarin/Xamarin.Forms/blob/ceb25378059c35fcec55e18bbbbe9c58d71b01e6/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs#L744-L759

So SetSupportActionBar(bar); isn't using the right toolbar which means this code never runs
https://github.com/xamarin/Xamarin.Forms/blob/78385f9fc1fc56dc88bd98e73bf9c8f2f2d0a90a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs#L69

I came across this while investigating
https://github.com/xamarin/Xamarin.Forms/issues/4827

I'll fill out the rest if this PR seems valid

### Issues Resolved ### 

- fixes #4827 

Possibly fixes
https://github.com/xamarin/Xamarin.Forms/issues/2118
Need to research a bit further

### Platforms Affected ### 
- Android

Clicking the software backbutton on Android will now fire the *OnBackButtonPressed* on Android

Also now on startup we're only inflating the toolbar view once opposed to twice

### Testing Procedure ###
Run the included Issue

### PR Checklist ###

- [X] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4827.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4827.cs
@@ -14,16 +14,21 @@ using Xamarin.Forms.Core.UITests;
 
 namespace Xamarin.Forms.Controls.Issues
 {
+	// This test only passes because of an override in FormsAppCompatActivity passes the 
+	// software back button press to the OnBackButtonPressed Command
 	[Preserve(AllMembers = true)]
-	[Issue(IssueTracker.Github, 4827, "[Android] Software backbutton doesn't work",
+	[Issue(IssueTracker.Github, 4827, "[Android] Software backbutton doesn't continue on device rotation",
 		PlatformAffected.iOS)]
 #if UITEST
 	[NUnit.Framework.Category(UITestCategories.Navigation)]
 #endif
 	public class Issue4827 : TestNavigationPage
 	{
+
+
 		Label results = new Label();
 		const string _success = "Success";
+		const string _pushPageButton = "Push Page";
 
 		public void Success()
 		{
@@ -43,7 +48,7 @@ namespace Xamarin.Forms.Controls.Issues
 						results,
 						new Button()
 						{
-							Text = "Push Page",
+							Text = _pushPageButton,
 							Command = new Command(PushSecondPage)
 						}
 					}
@@ -76,7 +81,7 @@ namespace Xamarin.Forms.Controls.Issues
 					{
 						new Label()
 						{
-							Text = "Click back. If you see text that says Success then test has passed"
+							Text = "Click back. If you see text that says Success then this test has passed"
 						}
 					}
 				};
@@ -92,9 +97,17 @@ namespace Xamarin.Forms.Controls.Issues
 
 #if UITEST && __ANDROID__
 		[Test]
-		public void UpdatingSourceOfDisposedListViewDoesNotCrash()
+		public void TestForSoftwareBackButtonTriggeringOnBackButtonPressed()
 		{
-
+			RunningApp.WaitForElement(_pushPageButton);
+			RunningApp.Tap(_pushPageButton);
+			RunningApp.WaitForElement(_success);
+			RunningApp.SetOrientationLandscape();
+			RunningApp.Tap(_pushPageButton);
+			RunningApp.WaitForElement(_success);
+			RunningApp.SetOrientationPortrait();
+			RunningApp.Tap(_pushPageButton);
+			RunningApp.WaitForElement(_success);
 		}
 #endif
 	}
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4827.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4827.cs
@@ -105,15 +105,15 @@ namespace Xamarin.Forms.Controls.Issues
 
 			var softwareBackButton = RunningApp.Query(app => app.Marked("toolbar").Descendant()).Where(x=> x.Class.Contains("AppCompatImageButton")).FirstOrDefault();
 
-			RunningApp.TapCoordinates(softwareBackButton.Rect.X, softwareBackButton.Rect.Y);
+			RunningApp.Tap(PlatformQueries.NavigationBarBackButton);
 			RunningApp.WaitForElement(_success);
 			RunningApp.SetOrientationLandscape();
 			RunningApp.Tap(_pushPageButton);
-			RunningApp.TapCoordinates(softwareBackButton.Rect.X, softwareBackButton.Rect.Y);
+			RunningApp.Tap(PlatformQueries.NavigationBarBackButton);
 			RunningApp.WaitForElement(_success);
 			RunningApp.SetOrientationPortrait();
 			RunningApp.Tap(_pushPageButton);
-			RunningApp.TapCoordinates(softwareBackButton.Rect.X, softwareBackButton.Rect.Y);
+			RunningApp.Tap(PlatformQueries.NavigationBarBackButton);
 			RunningApp.WaitForElement(_success);
 		}
 #endif
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4827.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4827.cs
@@ -4,7 +4,7 @@ using System.Collections.ObjectModel;
 using System.Text;
 using Xamarin.Forms.CustomAttributes;
 using Xamarin.Forms.Internals;
-
+using System.Linq;
 
 #if UITEST
 using Xamarin.UITest;
@@ -29,6 +29,7 @@ namespace Xamarin.Forms.Controls.Issues
 		Label results = new Label();
 		const string _success = "Success";
 		const string _pushPageButton = "Push Page";
+		const string _instructionsLabel = "_instructionsLabel";
 
 		public void Success()
 		{
@@ -81,7 +82,8 @@ namespace Xamarin.Forms.Controls.Issues
 					{
 						new Label()
 						{
-							Text = "Click back. If you see text that says Success then this test has passed"
+							Text = "Click back. If you see text that says Success then this test has passed",
+							AutomationId = _instructionsLabel
 						}
 					}
 				};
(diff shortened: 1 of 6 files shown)


hartez:
I doubt that the purpose of setting up the toolbar outside of the NativationPageRenderer was to mess with the back button. That's a lot of effort to go to just to stop the back button from working. 

So what you are saying is that your change fixes the issue, but it also causes the back button to behave differently? Can you elaborate on how it's different (and how it compares to iOS?)

PureWeen:
@hartez so right now the OnBackButtonPressed override on Page only executes for the hardware button on Android and on UWP. If you click the software BackButton (from a navigation page) on iOS or android that method doesn't fire

This PR fixes it on Android but it doesn't on iOS. I know there was a lot of talk around iOS and * OnBackButtonPressed* but I think that was mainly around not tying swipe to *OnBackButtonPressed*

If this PR seems valid then we should probably mark this one DNM and I should also make it so * OnBackButtonPressed* fires for the software button on iOS

hartez:
> @hartez so right now the OnBackButtonPressed override on Page only executes for the hardware button on Android and on UWP. If you click the software BackButton (from a navigation page) on iOS or android that method doesn't fire

Yes. And that's the correct behavior, at least according to the documentation. `OnBackButtonPressed` only applies to the hardware button, and only on platforms which have one (i.e., not iOS):
https://docs.microsoft.com/en-us/dotnet/api/xamarin.forms.navigationpage.onbackbuttonpressed?view=xamarin-forms

> This PR fixes it on Android but it doesn't on iOS. I know there was a lot of talk around iOS and _OnBackButtonPressed_ but I think that was mainly around not tying swipe to _OnBackButtonPressed_

So this PR makes the hardware back button on Android work for NavigationPage after the app resumes?

> If this PR seems valid then we should probably mark this one DNM and I should also make it so _OnBackButtonPressed_ fires for the software button on iOS
> 
> Or Merge this one so it gets applied and then schedule an iOS fix for the near future

If this fixes the after-resume-hardware-back-button on Android, we should get it merged and have the iOS discussion (if any) elsewhere. 

Also, why is this marked with the `breaking` label?

PureWeen:
To summarize this a bit first
- None of this has anything to do with the hardware back button
- right now we call *SetSupportActionBar* with a ToolBar that never sees the light of day. The ToolBar that's created in the NavigationRenderer is what gets renderered but this never gets passed to *SetSupportActionBar*
- We have code in our base class that looks like it was put there so that the Software Back Button would call OnBackButtonPressed https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/FormsApplicationActivity.cs#L64  as is this code is never called and if it was removed from the current core code there would be no behavior change
- The issue that prompted this fix #4827 was the user using a popular workaround to get the *OnBackButtonPressed* firing for the software button but that workaround fails if the NavigationRenderer is recreated during an orientation change or a resume that has to recreate the NAvigationRenderer because then NAvigationRenderer  renderers a new ToolBar that doesn't get passed to *SetSupportActionBar*
- Another way to fix #4827 would be to add an api to *NavigationPage* telling it to call that method for a software *OnBackButtonPressed* https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/FormsApplicationActivity.cs#L64 then we won't break default behavior but will provide something lots of people want :-) 


> Yes. And that's the correct behavior, at least according to the documentation. OnBackButtonPressed only applies to the hardware button, and only on platforms which have one (i.e., not iOS):
https://docs.microsoft.com/en-us/dotnet/api/xamarin.forms.navigationpage.onbackbuttonpressed?view=xamarin-forms

I'm just curious if that documentation was written because it's always been a bug or the intention initially was for it to work like this?

> So this PR makes the hardware back button on Android work for NavigationPage after the app resumes?

No it makes the *software* back button work after the app resumes or the orientation changes. Really anything that causes the NavigationRenderer to recreate itself.  The hardware back button always stays working but the original issues is about how they are trying to make the Software back button work https://github.com/xamarin/Xamarin.Forms/issues/4827.  The workaround in that issue is found a number of places where people suggest how to wire up the OnBackButtonPressed to the Software button. But that workaround breaks whenever the *NavigationRenderer* is recreated

Since our NavigationRenderer creates a new support bar on Orientation change then the code they have wired up stops working as our current NavigationRenderer doesn't call *SetSupportActionBar* all we do in our code is set the action bar to some dummy ActionBar that never sees the light of day

> Also, why is this marked with the breaking label?
Because it makes it so the software back button now fires *OnBackButtonPressed*

If we want to honor the original documentation and just never enable the SoftwareBackButton then we should just close #1944, #1413, and #4827 and say it's by design 

Though it seems like adding an API to enable the Software Back Button to fire *OnBackButtonPressed* on all three platforms would be super helpful

hartez:
> I'm just curious if that documentation was written because it's always been a bug or the intention initially was for it to work like this?

AFAIK, it has always worked this way. And that makes sense - the hardware back button is not the same thing as "navigating back". The hardware back button also serves functions like dismissing keyboards, nav drawers, and other stuff. The "back button" provided by the NavigationPage is a distinct UI concept from any hardware buttons provided on Android or UWP. 

But looking at the comments in #1944 and #1413, it's clear that the current API (and years of workaround cruft) is causing a bunch of confusion. We need to fix it, but not as part of this PR.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4827.cs
```diff
@@ -0,0 +1,101 @@
...
+			protected override bool OnBackButtonPressed()
+			{
+				_issue4827.Success();
+				return base.OnBackButtonPressed();
+			}
+		}
+
+#if UITEST && __ANDROID__
+		[Test]
+		public void UpdatingSourceOfDisposedListViewDoesNotCrash()
```
hartez: I don't understand this test name.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4827.cs
```diff
@@ -0,0 +1,101 @@
...
+
+			public BackTestContentPage(Issue4827 issue4827)
+			{
+				Content = new StackLayout()
+				{
+					Children =
+					{
+						new Label()
+						{
+							Text = "Click back. If you see text that says Success then test has passed"
```
hartez: "then test" -> "then this test"

PureWeen:
> the hardware back button is not the same thing as "navigating back". The hardware back button also serves functions like dismissing keyboards, nav drawers, and other stuff. The "back button" provided by the NavigationPage is a distinct UI concept from any hardware buttons provided on Android or UWP.

Only thing is that UWP calls *OnBackButtonPressed* when using the navigate back button.

But yea we should probably provide a specific *OnNavigatingBack* or similar that ties the navigation concept more directly to intent. As is people keep overriding  *OnBackButtonPressed* and then some permutation of what they test stops the navigation so they think they're good to go.

Maybe even adding a context args to *OnBackButtonPressed* would help so they know exactly what action is happening they want to respond to.

PureWeen:
build --uitests

PureWeen:
alright @hartez @kingces95 ready for review

PureWeen:
@kingces95 just wait for now

I ran into some additional issues with this PR that will need to be addressed first which is probably why the tests are failing

pauldiston:
Having upgraded from XF 3.3 to XF 4.0 yesterday I am experiencing a number of issues, one being OnOptionsItemSelected isn't called any longer, which I presume is a result of the incorrect toolbar being wired up, hence this PR.
For clarification, this comment "samhouts removed this from Closed in vNext (4.1.0) 9 days ago", does this mean that this issue will be resolved in XF 4.1?
I have tried the setting of the toolbar "manually" from (https://forums.xamarin.com/discussion/67854/android-menu-and-back-button-not-working) but I don't seem to be able to get this working. If I have to wait until XF 4.1 for this to be resolved then so be it, as I am already waiting for issues to be resolved with the CarouselView which is planned for XF 4.1 as well.

pauldiston:
@samhouts Would you be able to provide any information regarding the question I had about the comment?

samhouts:
@pauldiston This was not merged, so it will not be resolved yet. There was a workaround described in https://github.com/xamarin/Xamarin.Forms/issues/4827#issuecomment-480153471 that may help you.

@PureWeen fyi

pauldiston:
@samhouts Having parked this issue back in June 2019, I am back again with the same issue in version 4.5.0.430. Having tried a work around I found here (https://theconfuzedsourcecode.wordpress.com/2017/03/02/formsappcompatactivity-is-not-calling-onoptionsitemselected-xamarin-android/) I am now running into this issue (https://github.com/xamarin/Xamarin.Forms/issues/2118).
There must be a way of handling the OnOptionsItemSelected without having all these problems.

pauldiston:
@samhouts After further investigation, is seems that everytime I want to set the toolbar items for a page I need to call SetSupportActionBar, doing it in the OnCreate is just not enough. Time to add my SetToolbar method call to each page that has toolbar items.

Issue #4597, opened 2018-11-30
[Android] With the introduction of ImageButton in 3.4.0, the behaviour of setting Source to null on an Image has changed in Android


jorisvergeer:
### Description

I cannot clear an Image by setting the source to null. 
This behavior has changed in 3.4.0 with the introduction of ImageButton.

This only affects Android.

The issue is probably causes by this change in Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs

![image](https://user-images.githubusercontent.com/1906639/49288355-cd3be780-f4a0-11e8-8b1a-663f7f8b582f.png)

### Steps to Reproduce

1. Create an image view
2. Set source to an image resource
3. Set source back to null

### Expected Behavior
The Image is goes back to its initial state without an image to show.

### Actual Behavior
The old image is still visible.

### Basic Information

- Version with issue: 3.4.0
- Last known good version: 3.1

PureWeen:
In testing it looks like this behavior changed from 3.1 to 3.2
I tested with 3.2 and 3.2 exhibits the behavior you are talking about and it also causes it to act differently than Button when you set that source to null

jorisvergeer:
It could be wrong in an earlier version. It got noticed in our app when we updated to 3.4.

PR #4915, opened 2019-01-07
[WIP] Unify the image handling

### LINKED ISSUE #3207: Replace all occurences of FileImageSource with ImageSource
### Description
In order to be able to use other image sources such as SVG based ones all Core controls and renderers should use `ImageSource ` instead of `FileImageSource`.

When the Core is updated, renderers should also be updated to never assume a `FileImageSource`.

These are the occurrences I could track down in Core and other Platform renderers:

### Core
1. `Button.Image`
2. `MenuItem.Icon`
3. `NavigationPage.TitleIcon`
4. `Page.Icon`
5. `Slider.ThumbImage`

### Android
1. NavigationPageRenderer toolbar icon
2. TabbedPageRendererer tab icon
3. ImageViewExtensions that is used by BaseCellView
4. Fastrenderers Button.Image
5. Platform title Icon

### iOS
1. ButtonRenderer image
2. NagivationRenderer title image
3. SliderRenderer thumb image

### UWP
None.

See also:  #1955

### LINKED ISSUE #4689: [Android] StackLayout.LayoutChildren null ref XF 3.4
First time reporting a bug. I hope this is an acceptable format. I have provided as much detail as I can, with a video of the issue, with timestamps, and a proposed solution at the end of this post. I use my own private source control server for my projects, so I'm also new to GitHub.

### Description
When modifying a StackLayout, a null reference exception is raised in Xamarin.Forms.Core.StackLayout.cs:66. The Plots and Requests properties of the layoutInformationCopy (and _layoutInformation) are null. This occurs on Android, when the app is minimized. I don't think this exception has been raised when the app is maximized. I guess it may have something to do with the activity being suspended.

I determined this by forking the Xamarin.Forms source and debugging it, thanks to this very helpful post by Jon Goldberger: https://blog.xamarin.com/debugging-xamarin-forms/

I followed his instructions, and used my own fork to figure out the above crash point. Prior to following the guide, I only knew that there was a null ref in the StackLayout.LayoutChildren method.

Looking through the code, you may wonder how these properties can be null, because in my testing, CalculateLayout is ALWAYS called before this exception is raised, and CalculateLayout sets these properties to non-null. I have found that if I set a conditional breakpoint before the call to CalculateLayout, for when _layoutInformation.Plots is null, and if I then F11 into CalculateLayout and F10 over everything in there, then the Plots and Requests properties are ALWAYS non-null. However, if I hit that break point, and then F5 over the method, there is a very good chance that the properties will still be null after executing CalculateLayout, leading to a null ref on line 66.

Perhaps you guys who know more about the framework will know more about how this could happen. I've not been looking at the code for very long, but I cannot see how these properties could be null. It's like there's a glitch at a deeper level.

As it's such an odd issue, I decided to make a video. Some parts are sped up, and some are slowed down, to shorten the video and make some points more clear. I also blurred a few things for privacy: https://www.youtube.com/watch?v=88mGjcyWvWg&feature=youtu.be

In the video, you can see my desktop, running VS 2017, and the WPF version of my Clipboard Everywhere app. This WPF version doesn't show the problem, but is used to demonstrate the problem on Android by sending a file to my phone, one of my new features I wanted to release soon. WPF app sends a file to the Android app, which is currently being debugged on my Note 9. The app loads, initializes, and then I minimize it. I then send the file from WPF, and if I F5 over the CalculateLayout method, the Plots and Requests properties have a good chance to be null, crashing the app. When the file is sent, this causes the layout to update. It seems the issue is close to the visibility of child items in the main StackLayout for the page. The points where the crash occur are near where these child items are made visible and invisible, but I'm not sure how to narrow it down further.

**Timestamps:** 
* 0 to 00:23, demonstrating the null ref, I actually didn't expect this part to crash. It's a bit unstable.
* 00:28 to 02:07, demonstrating that, when stepping through the method, the Plots and Requests properties are never null
* 02:10 to end, demonstrating that the properties can be null when skipping the method with F5

The "child" always seems to be a StackLayout containing two children, which are both custom controls I made, however that's only what "child" refers to because it can't progress through the loop, due to the null properties. Here is the full Layout of the page. It's hosted in a TabbedPage, inside a NavigationPage: https://pastebin.com/YyrBMywr

The StackLayout with the two children is the one that's very near the top, with Orientation="Horizontal". I use the CustomEntry and ImageButton (my own version, made it a bit back, it's actually not displaying images but FontAwesome icons), frequently through my apps, and haven't seen this issue anywhere else.

The layout_SendingFiles, scrollView_CurrentPath and layout_Files all contain dynamic content. Controls are added at runtime but I don't delete them, I found it better performance-wise to cache them for re-use, so they have their IsVisible set to False when not in use. When being used again, I set their properties as appropriate for the new content they are displaying. However, I have to manually set the heights of each control in the Files layouts, and the heights of the Files layouts themselves, each time the layouts are updated, as the controls seem to inexplicably double in height every time I re-use them. Perhaps this relates to a different bug in the framework. I've had to redesign a layout to avoid that issue before.

### Steps to Reproduce
I can't share my code, and I'm not sure if I will be able to make another project to reproduce this issue. I don't have time travel debugging, which I feel could be useful here, so I'm not sure how to investigate this further.

Reproducing this issue currently depends on having the Android app, the WPF app, and the web service running.

```C#
class CustomLabel : Label
{
	protected override SizeRequest OnMeasure(double widthConstraint, double heightConstraint)
	{
		var returnValue = base.OnMeasure(widthConstraint, heightConstraint);
		return returnValue;
	}
}

public Page CreateDefaultMainPage()
{
	return new ContentPage()
	{
		Content = new StackLayout()
		{
			Children = 
			{
				new CustomLabel()
			}
		}
	};
}
```

### Expected Behavior
Not crashing.

### Actual Behavior
Crashing due to null ref

### Basic Information

- Version with issue: 3.4
- Last known good version: Unknown. Looks like StackLayout is not changed very frequently.
- IDE: VS 15.9.2
- Platform Target Frameworks:
  - Android: v9, API 28
- Android Support Library Version: v27.0.2.1
- Nuget Packages: Lots. Can I get a list somewhere from VS?
- Affected Devices: Galaxy Note 9 Exynos, unsure if it would happen on others. I would like to test the emulators but they don't use bridged networking so the particular software feature I've developed that raises this problem doesn't work as the emulators don't get their own IP.

### StackTrace
{"Source":"Xamarin.Forms.Core","StackTrace":" at Xamarin.Forms.StackLayout.LayoutChildren (System.Double x, System.Double y, System.Double width, System.Double height) [0x000b4] in C:\\Repos\\Xamarin.Forms-Forked\\Xamarin.Forms.Core\\StackLayout.cs:66 <br/> at Xamarin.Forms.Layout.UpdateChildrenLayout () [0x00166] in C:\\Repos\\Xamarin.Forms-Forked\\Xamarin.Forms.Core\\Layout.cs:263 <br/> at Xamarin.Forms.Layout.OnSizeAllocated (System.Double width, System.Double height) [0x00011] in C:\\Repos\\Xamarin.Forms-Forked\\Xamarin.Forms.Core\\Layout.cs:223 <br/> at Xamarin.Forms.VisualElement.SizeAllocated (System.Double width, System.Double height) [0x00001] in C:\\Repos\\Xamarin.Forms-Forked\\Xamarin.Forms.Core\\VisualElement.cs:801 <br/> at Xamarin.Forms.VisualElement.SetSize (System.Double width, System.Double height) [0x0002c] in C:\\Repos\\Xamarin.Forms-Forked\\Xamarin.Forms.Core\\VisualElement.cs:1033 <br/> at Xamarin.Forms.VisualElement.set_Bounds (Xamarin.Forms.Rectangle value) [0x00069] in C:\\Repos\\Xamarin.Forms-Forked\\Xamarin.Forms.Core\\VisualElement.cs:277 <br/> at Xamarin.Forms.VisualElement.Layout (Xamarin.Forms.Rectangle bounds) [0x00001] in C:\\Repos\\Xamarin.Forms-Forked\\Xamarin.Forms.Core\\VisualElement.cs:687 <br/> at Xamarin.Forms.Layout.LayoutChildIntoBoundingRegion (Xamarin.Forms.VisualElement child, Xamarin.Forms.Rectangle region) [0x00211] in C:\\Repos\\Xamarin.Forms-Forked\\Xamarin.Forms.Core\\Layout.cs:177 <br/> at Xamarin.Forms.ScrollView.LayoutChildren (System.Double x, System.Double y, System.Double width, System.Double height) [0x00095] in C:\\Repos\\Xamarin.Forms-Forked\\Xamarin.Forms.Core\\ScrollView.cs:201 <br/> at Xamarin.Forms.Layout.UpdateChildrenLayout () [0x00166] in C:\\Repos\\Xamarin.Forms-Forked\\Xamarin.Forms.Core\\Layout.cs:263 <br/> at Xamarin.Forms.Layout.OnSizeAllocated (System.Double width, System.Double height) [0x00011] in C:\\Repos\\Xamarin.Forms-Forked\\Xamarin.Forms.Core\\Layout.cs:223 <br/> at Xamarin.Forms.VisualElement.SizeAllocated (System.Double width, System.Double height) [0x00001] in C:\\Repos\\Xamarin.Forms-Forked\\Xamarin.Forms.Core\\VisualElement.cs:801 <br/> at Xamarin.Forms.Layout+<>c.<OnChildMeasureInvalidated>b__45_0 () [0x00094] in C:\\Repos\\Xamarin.Forms-Forked\\Xamarin.Forms.Core\\Layout.cs:380 <br/> at Java.Lang.Thread+RunnableImplementor.Run () [0x00008] in <2530aa2e412d4754af3ae7e1a8c80b6d>:0 <br/> at Java.Lang.IRunnableInvoker.n_Run (System.IntPtr jnienv, System.IntPtr native__this) [0x00009] in <2530aa2e412d4754af3ae7e1a8c80b6d>:0 <br/> at (wrapper dynamic-method) System.Object.23(intptr,intptr)","InnerException":null,"Message":"Object reference not set to an instance of an object."} 

### Screenshots
This image shows an attempted fix, by including a null check, but the check passed and the null got through.

The three breakpoints are conditionals for when _layoutInformation.Plots is null. The first was hit, the second two were not. Looks like Plots was set to null after leaving the else block.
stacklayoutbug_nullcheckpassed

### Proposed Solution
The best solution I have found so far is to try/catch the LayoutChildIntoBoundingRegion method. This seems to work without any negative effects that I can see, so far. It doesn't seem to cause any layout issues. Here are the changes I would like to propose for the StackLayout.LayoutChildren method: https://github.com/Liam2349/Xamarin.Forms/commit/4fe7f9215ae65b1ea65aaf8303464fd600a4595c

In addition to my try/catch, I removed the layoutInformationCopy variable, as it doesn't seem to be a copy. Perhaps someone mixed up their reference and value-types when writing that in?

It's 3AM here in the UK, and I've been at this for a few hours now, so I'll see how it goes tomorrow. I look forward to your responses. Please let me know if I've missed anything.

Thanks for reading and I look forward to your responses.

mattleibow:
### Description of Change ###

This is now a big PR that changes everything to use `ImageSource` instead of strings, files or other image-y types.

**NOTE: This is a set of breaking changes as the property types have changed!** 
__NOTE: The changes to the public API can be reverted temporarily as it is going down a level. As a result, all the code will still be valid.__

### Issues Resolved ### 

 - Fixes the inconsistency with image source types.
 - Fixes #3207 
- Fixes #4689

### Known Issues ###
https://github.com/xamarin/Xamarin.Forms/issues/6004
https://github.com/xamarin/Xamarin.Forms/issues/6010
regresses #4597 on API 19

### API Changes ###

Added:

```
class ImageSource {
    // Returns true if this image source is actually not going to return anything
    // It is not going to determine if the result will actually have a valid image, but 
    // rather it will indicate if there is no file path, or if the stream is null, or if 
    / the uri is empty
    public virtual bool IsEmpty { get; }
}
```

Changed the property types to `ImageSource` for the following properties:

```csharp
Button.Image // from FileImageSource
MenuItem.Icon // from FileImageSource
NavigationPage.TitleIcon // from FileImageSource (this also had the static getter/setter method for the attached property)
Page.BackgroundImage // from string
Page.Icon // from FileImageSource
Slider.ThumbImage // from FileImageSource
```

As a result of things changing, I also changed the name of `ImageElement.FileImageProperty` to `ImageProperty`. This should not affect anyone as it was internal.

### Platforms Affected ### 

All the platforms. Everything.

None.

Not applicable

### Testing Procedure ###

Enable the `SetMainPage(new ImageSourcesGallery());` in the gallery app and make sure all the buttons do what they are supposed to do.

### PR Checklist ###

- [ ] Has automated tests 


samhouts:
This is prep for #3207, right?

mattleibow:
A big yes to #3207. Once this PR is merged, the switch to ImageSource will be "trivial", but then we have to make all image loading async because we could be loading an image from a remote url. And, this is probably for the better because loading a large image into a button/toolbar (which can be done) already blocks the UI.

rogihee:
Great work!

dansiegel:
would really love to know what the hold up on this is. This would actually make it a lot of scenarios a lot easier... not the least of which would be to use the new FontImageSource for Page Icons... Not to mention would probably help flatten a lot of UI's by allowing me to specify an Image Source for a Background rather than having to use some layout trickery.

StephaneDelcroix:
@dansiegel this introduce a breaking change. that's why it's still holding

groege:
@StephaneDelcroix Can you give us a timeline when we can hope for a release (or prelease) with the changes? This would also be nice for the issue #3207.
We're waiting for those changes before we continue development on our app, so if we'd have at least a guess I could forward it to my boss on when he can expect us to continue with that project.

toomasz:
When this will be merged? It would make it possible to use FontAwasome for regular button which will make it look pretty nice with Visual=Material

jamiewest:
I asked @davidortinau whether he thought this would make the 4.0 release and he said it would be a slim change. Just FYI

[Twitter Link](https://twitter.com/Jamiewst/status/1105214077247209474)

toomasz:
@jamiewest So I guess it's unlikely it will make into the release soon :(

dansiegel:
@StephaneDelcroix I understand this represents breaking changes. 4.0 is also the best possible time to introduce such a change as it is a MAJOR version change.

PureWeen:
Xamarin.Forms.Controls.Issues.Bugzilla41038

Looks like it's failing on all the iOS platforms

Xamarin.Forms.Platform.iOS/Renderers/NavigationRenderer.cs
```diff
@@ -743,35 +743,36 @@ internal static async void SetMasterLeftBarButton(UIViewController containerCont
...
+
+				if(icon == null || containerController.NavigationItem.LeftBarButtonItem == null)
 				{
-					// Throws Exception otherwise would catch more specific exception type
-					shouldUseIcon = false;
+					containerController.NavigationItem.LeftBarButtonItem = new UIBarButtonItem(masterDetailPage.Master.Title, UIBarButtonItemStyle.Plain, OnItemTapped);
 				}
-			}
 
-			if (!shouldUseIcon)
```
PureWeen: @mattleibow 
I moved all of this code up into the callback

https://github.com/xamarin/Xamarin.Forms/pull/4915/commits/e8d23bf746fcfb6c009c166618f7cee0571ad6ce

Otherwise it was just never setting the title if the icon wasn't specified

Xamarin.Forms.Platform.Android/IButtonLayoutRenderer.cs
```diff
@@ -1,12 +1,9 @@
...
 
 namespace Xamarin.Forms.Platform.Android
 {
-	public interface IButtonLayoutRenderer
+	public interface IButtonLayoutRenderer : IVisualElementRenderer
 	{
-		AppCompatButton View { get; }
-		Button Element { get; }
-		event EventHandler<VisualElementChangedEventArgs> ElementChanged;
+		new AppCompatButton View { get; }
```
PureWeen: @mattleibow @samhouts I had to add this back because the IVisualElementRenderer.View is just going to return the renderer itself for renderers that aren't fast renderers. So all Text was showing up blank in the Control Gallery

PureWeen:
known issue
https://github.com/xamarin/Xamarin.Forms/issues/6004

jamiewest:
This is a great set of changes, really impressed you all got through for 4.0.0. Excited to see this in the near future 👍

EmilAlipiev:
are those changes documented somewhere? I dont find anything regarding this in official Documentation

PR #4916, opened 2019-01-08
[Android] Allow the page background to be reset

mattleibow:
### Description of Change ###

Once a background is set on the page, it can never be unset.

### Issues Resolved ### 

- fixes #

### API Changes ###

 None

### Platforms Affected ### 

- Android

Setting a page background to an image, and then setting it to null will now actually set the page to no image. Previously, the image stayed.

Not applicable

### Testing Procedure ###

 - Create a page
 - Set the background to a valid image
 - _observe the image_
 - Set the background to null
 - _observe the blank page_

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/MasterDetailPageRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/MasterDetailPageRenderer.cs
@@ -379,6 +379,8 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 			string backgroundImage = view.BackgroundImage;
 			if (!string.IsNullOrEmpty(backgroundImage))
 				this.SetBackground(Context.GetDrawable(backgroundImage));
+			else
+				this.SetBackground(null);
 		}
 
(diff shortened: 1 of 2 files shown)


samhouts:
In order to stabilize 3.5.0, this was reverted. The fix will be reapplied by #5066 in master for the next version.

taublast:
I can see this as one of the points that were "fixed" in [4.0.0.94569-pre3].. Well on all versions up to [4.0.0.62955-pre2] i had no issues and in the latest 4.0.0.94569-pre3 on android my pages backgrounds now are just plain white. And the flyout drawer menu became silly transparent..

andreinitescu:
@taublast see @samhouts 's prevous comment

@samhouts If a nightly build (which per my understanding is based on the master branch) could be done, maybe @taublast could test with it and see if it fixes it for him

samhouts:
You should no longer see the missing background color issue in 3.5.0 stable and 4.0.0-pre4. Thanks!

PR #4944, opened 2019-01-09
[tests] UWP requires file extensions

mattleibow:
This should also be fine on iOS/Android as the extension is used in the other tests.


samhouts:
build --uitests

samhouts:
failures unrelated

PR #4953, opened 2019-01-10
Makes iOS ImageButton opaque when pressed

### LINKED ISSUE #4947: [iOS] ImageButton is half transparent when pressed
### Description
On iOS, when ImageButton is in pressed state it's translucent. It should behave like on other platforms, and have a solid color.

My workaround (and probably fix):
Set UIButtonType to Custom
Set AdjustsImageWhenHighlighted to false

Code:
```
using Gemini.Changes.iOS.Renderers;
using UIKit;
using Xamarin.Forms;
using Xamarin.Forms.Platform.iOS;

[assembly: ExportRenderer(typeof(ImageButton), typeof(CustomImageButtonRenderer))]
namespace Gemini.Changes.iOS.Renderers
{
    public class CustomImageButtonRenderer : ImageButtonRenderer
    {
        protected override void OnElementChanged(ElementChangedEventArgs<ImageButton> e)
        {
            if (Control == null)
            {
                SetNativeControl(UIButton.FromType(UIButtonType.Custom));
                Control.AdjustsImageWhenHighlighted = false;
            }

            base.OnElementChanged(e);
        }
    }
}
```

### Steps to Reproduce

1. Use any ImageButton (for  example from https://github.com/davidortinau/TheLittleThingsPlayground)
2. Press ImageButton on iOS
3. 

### Expected Behavior
Button should be smaller, but color should stay the same

### Actual Behavior
Button is half transparent

### Basic Information
- Version with issue: 3.4.0.1009999, 3.4.0.1009999
- Last known good version:
- IDE: Visual Studio 15.9.4
- Platform Target Frameworks: 
  - iOS:  12.1
- Affected Devices: iPhone

kingces95:
### Description of Change ###

Makes iOS ImageButton opaque when pressed.

### Issues Resolved ### 
- fixes #4947 
### Platforms Affected ### 
- iOS
### Testing
Use update `.create-nuget.bat` to create private nuget to run AllLittleThings sample.
image

### DIFF
--- a/.create-nuget.bat
+++ b/.create-nuget.bat
@@ -9,7 +9,8 @@ set NUGET_EXE=%NUGET_DIR%NuGet.exe
 
 mkdir Xamarin.Forms.Platform.MacOS\bin\debug\
 mkdir Xamarin.Forms.Platform.Tizen\bin\debug\tizen40\
-mkdir Xamarin.Forms.Platform.MacOS\bin\debug\
+mkdir Xamarin.Forms.Maps.Tizen\bin\debug\Tizen40
+mkdir Xamarin.Forms.Maps.MacOS\bin\debug
 mkdir Xamarin.Forms.Platform.UAP\bin\debug\
 mkdir Xamarin.Forms.Platform.ios\bin\debug\
 mkdir Stubs\Xamarin.Forms.Platform.iOS\bin\iPhone\debug\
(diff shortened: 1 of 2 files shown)


mattleibow:
Is this what we really want to do? All the other buttons on iOS go lighter. If this is a button, how do we indicate that it is being pressed? If we don't want any change, then maybe just use an image view?

Another thing, if we are trying to remove the "go light" effect, then maybe this should be a platform specific? All buttons on iOS, and only iOS, go light. It may be useful to disable this effect on any button. For normal buttons, someone may be using a visual state manager, so this lightning may also be undesired.

samhouts:
Closing. See https://github.com/xamarin/Xamarin.Forms/issues/4947#issuecomment-462916933

PR #4985, opened 2019-01-14
[iOS] CollectionView single selection

hartez:
### Description of Change ###

Single selection for CollectionView on iOS

### Issues Resolved ### 

- Partially implements #3172 

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###

In Control Gallery, navigate to CollectionView Gallery -> Selection Galleries 

### PR Checklist ###

- [ ] Has automated tests 


Xamarin.Forms.Platform.iOS/CollectionView/UICollectionViewDelegateForwarder.cs
```diff
@@ -0,0 +1,60 @@
...
+
+		public override UIEdgeInsets GetInsetForSection(UICollectionView collectionView, UICollectionViewLayout layout,
+			nint section)
+		{
+			if (ItemsViewLayout == null)
+			{
+				return default(UIEdgeInsets);
+			}
+
+			return ItemsViewLayout.GetInsetForSection(collectionView, layout, section);
```
paymicro: more question marks - less code =)
```suggestion
			return ItemsViewLayout?.GetInsetForSection(collectionView, layout, section) ?? default(UIEdgeInsets);
```

rmarinho:
build

Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewController.cs
```diff
@@ -21,14 +21,23 @@ internal class CollectionViewController : UICollectionViewController
...
 			_itemsSource = ItemsSourceFactory.Create(_itemsView.ItemsSource, CollectionView);
 			_layout = layout;
 
 			_layout.GetPrototype = GetPrototype;
 			_layout.UniformSize = false; // todo hartez Link this to ItemsView.ItemSizingStrategy hint
+
+			Delegator = new UICollectionViewDelegator
+			{
+				ItemsViewLayout = _layout
+			};
```
mattleibow: Just mentioning here - if this is required, maybe make it a constructor arg?

Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewRenderer.cs
```diff
@@ -0,0 +1,144 @@
...
+		}
+
+		protected virtual ItemsViewLayout SelectLayout(IItemsLayout layoutSpecification)
+		{
+			if (layoutSpecification is GridItemsLayout gridItemsLayout)
+			{
+				return new GridViewLayout(gridItemsLayout);
+			}
+
+			if (layoutSpecification is ListItemsLayout listItemsLayout)
```
mattleibow: maybe make this an else/if - feels better that way. #nitpicking

Xamarin.Forms.Platform.iOS/CollectionView/SelectableItemsViewRenderer.cs
```diff
@@ -0,0 +1,78 @@
...
+
+				// If the mode is Multiple and SelectedItem is set to null, don't do anything
+				return;
+			}
+
+			if (mode != SelectionMode.Multiple)
+			{
+				ClearSelection();
+				SelectableItemsViewController.SelectItem(selectedItem);
+			}
```
mattleibow: This code looks iffy... we do a check for selected item == null, then inside that, we check for none/single. and then return.
Over here, we do a check for !multiple - which is the same as none || single.

They look to be the same type of action, but need confirmation as to the desired actions.

You might be wanting:

```csharp
if (mode != SelectionMode.Multiple) {
    ClearSelection();
    if (selectedItem != null) {
        SelectableItemsViewController.SelectItem(selectedItem);
    }
} else {
    // handle multiple selection...
}
```

Xamarin.Forms.Platform.iOS/CollectionView/SelectableItemsViewRenderer.cs
```diff
@@ -0,0 +1,78 @@
...
+				SelectableItemsViewController.UpdateSelectionMode();
+			}
+		}
+
+		protected override void SetUpNewElement(ItemsView newElement)
+		{
+			if (newElement != null && !(newElement is SelectableItemsView))
+			{
+				throw new ArgumentException($"{nameof(newElement)} must be of type {typeof(SelectableItemsView).Name}");
+			}
```
mattleibow: Not for this PR, but I think we need to create an extension method to do this:

`newElement.EnsureTypeOrNull<SelectableItemsView>()`

I know I have been using this chunk of code for material...

Xamarin.Forms.Platform.iOS/CollectionView/TemplatedCell.cs
```diff
@@ -69,5 +69,21 @@ void ClearSubviews()
...
+			get => base.Selected;
+			set
+			{
+				base.Selected = value;
+
+				var element = VisualElementRenderer?.Element;
+
+				if (element != null)
+				{
+					VisualStateManager.GoToState(VisualElementRenderer.Element, value ? "Selected" : "Normal");
```
mattleibow: Probably better to use the variable. And, we probably need to start thinking about these magic strings - they should be constants.

```suggestion
					VisualStateManager.GoToState(element, value ? "Selected" : "Normal");
```

Xamarin.Forms.Platform.iOS/CollectionView/TemplatedCell.cs
```diff
@@ -69,5 +69,21 @@ void ClearSubviews()
...
+		public override bool Selected
+		{
+			get => base.Selected;
+			set
+			{
+				base.Selected = value;
+
+				var element = VisualElementRenderer?.Element;
+
+				if (element != null)
```
mattleibow: Do we have a check (if it is possible to get to this state) that when the element is set, it goes to the correct state?

Xamarin.Forms.Platform.iOS/CollectionView/UICollectionViewDelegateForwarder.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System;
+using Foundation;
+using UIKit;
+
+namespace Xamarin.Forms.Platform.iOS
+{
+	public class UICollectionViewDelegator : UICollectionViewDelegateFlowLayout
```
mattleibow: The type name and file name do not match #nitpicking

Xamarin.Forms.Platform.iOS/CollectionView/CarouselViewRenderer.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Xamarin.Forms.Platform.iOS
+{
+	public class CarouselViewRenderer
+	{
+		public CarouselViewRenderer()
```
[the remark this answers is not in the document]
hartez: You assume correctly.

Xamarin.Forms.Platform.iOS/CollectionView/CollectionViewRenderer.cs
```diff
@@ -1,202 +1,4 @@
...
-					return UICollectionViewScrollPosition.Top;
-				case ScrollToPosition.End:
-					return UICollectionViewScrollPosition.Bottom;
-				case ScrollToPosition.Center:
-				default:
-					return UICollectionViewScrollPosition.CenteredVertically;
-			}
-		}
-	}
+	public class CollectionViewRenderer : SelectableItemsViewRenderer { }
```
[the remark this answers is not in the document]
hartez: No? I don't understand the question.

Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewCell.cs
```diff
@@ -12,6 +12,13 @@ protected ItemsViewCell(CGRect frame) : base(frame)
 		{
 			CollectionView.VerifyCollectionViewFlagEnabled(nameof(ItemsViewCell));
 			ContentView.BackgroundColor = UIColor.Clear;
+
+			var selectedBackgroundView = new UIView
+			{
+				BackgroundColor = UIColor.Gray
+			};
+
+			SelectedBackgroundView = selectedBackgroundView;
```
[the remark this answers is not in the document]
hartez: Android already handles it (default platform-normal selection handling, with the option to override via the VSM). 
It's already on my TODO list, but here you go in case I get hit by a bus: https://github.com/xamarin/Xamarin.Forms/issues/5050

Xamarin.Forms.Platform.iOS/CollectionView/SelectableItemsViewRenderer.cs
hartez: >  but need confirmation as to the desired actions

What needs confirmation?

hartez: Simplified.

Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewController.cs
hartez: Done.

Xamarin.Forms.Platform.iOS/CollectionView/TemplatedCell.cs
hartez: I don't understand the question.

Xamarin.Forms.Platform.iOS/CollectionView/CollectionViewRenderer.cs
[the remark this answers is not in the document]
mattleibow: After our little chat on Thurs, I now understand :)

Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewCell.cs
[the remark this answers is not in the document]
mattleibow: I was more thinking that you might move on to bigger, better things and forget about these lowly "issues". But, hey, go big bus, or take it home, right?

Xamarin.Forms.Platform.iOS/CollectionView/TemplatedCell.cs
mattleibow: If at some later stage, we go `cell.Element = something`, will it go to the correct state? It looks like we only change state when selection changes. So if we are selected and then get an element, will we still end up in the right state?

hartez: Oh, the situation you're describing doesn't happen. All the element creation is managed by the controller; there's no place for a user to later set the element of a cell. There isn't even a cell.Element to set.

Issue #3749, opened 2018-09-08
[Core] Isn't DataTemplateSelector the wrong name?


adrianknight89:
This class is inheriting from `DataTemplate` which technically makes it another data template. A selector seems to imply a different kind of class that provides functionality around the object it works with. Shouldn't it be actually called something like `MultiDataTemplate`? I'm wondering if this class can be marked obsolete in favor of something that serves its purpose. Maybe the name is fine, but it shouldn't inherit from that base class.

andreinitescu:
I can't understand why `DataTemplateSelector` derives from `DataTemplate`. 
I wonder for example what happens if I have a `DataTemplateSelector` instance and I call `CreateContent()` ?
Following other XAML technologies, the `DataTemplateSelector` in Xamarin Forms should not derive from any other class.

GalaxiaGuy:
This happens: https://github.com/xamarin/Xamarin.Forms/issues/3544

andreinitescu:
@GalaxiaGuy So much confusion in that thread, and it shows that having DataTemplateSelector derived from DataTemplate creates a lot of confusion. I am not sure why the issue is marked with "needs-repro" and "needs-info".

The method `CreateContent` actually guards against using `DataTemplateSelector` in place of `DataTemplate`, because really a `DataTemplateSelector` is not what a `DataTemplate` is meant for:
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/ElementTemplate.cs#L79

I think what happened is, in Xamarin Forms someone decided to fuse the roles of DataTemplate and DataTemplateSelector, in order to make it _easier™_ to work with `ListView`. 
Instead, it should really be a separate specialized property of `DataTemplateSelector` type, just how the XAML frameworks (WPF, UWP) do it (See [ItemTemplateSelector ](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.itemscontrol.itemtemplateselector)in UWP)

GalaxiaGuy:
I assume the tags are left there from the initial post which didn't have a concrete example of the problem.

One thing to note is it not just `ListView` using `DataTemplate` these days. The new Shell stuff uses it as well as a handful of third party libraries - including one I wrote which is why I have familiarity with it.

[...]

Issue #3544, opened 2018-08-10
[Core] Fix CreateContent usage on DataTemplatesSelector


numb3r:
### Description

-it makes sense to me to open DataTemplateExtension (change the namespace, remove the attribute)
-we should review our own parameterless CreateContent(), I'm thinking about LV.Header and Footer templates.

I agree with @StephaneDelcroix 's recommendations
CreateContent is the public api and it doesn't work so it needs to be fixed or made obsolete for a DataTemplateSelector and the only functional way to do it should be moved out of internals

### Original Description

Using DataTemplate, when it tries to create content for ItemTemplate it throw Invalid Operation exception saying "LoadTemplate should not be null".

Version with issue Xamarin 4.10.10.2:

Platform Target Frameworks: 
Android: Android SDK 8.3.3.2 
Affected Devices: Samsung SM-T377P

### Reproduction
[DataTemplateDemo.zip](https://github.com/xamarin/Xamarin.Forms/files/2376740/DataTemplateDemo.zip)

samhouts:
Can you attach a small project that reproduces this issue? Which version of the Xamarin.Forms nuget package are you using? Thanks!

GalaxiaGuy:
The message seems to be generated here:

https://github.com/xamarin/Xamarin.Forms/blob/d4480b2e48e4ec85bf0f20b19e3b7b9d6cf1744a/Xamarin.Forms.Core/ElementTemplate.cs#L77

I believe when I saw this problem a while ago, it was because the `DataTemplate` I was trying to use was really a `DataTemplateSelector`. If that is the case, then there is already an extension method that checks for this situation and does the right thing:

https://github.com/xamarin/Xamarin.Forms/blob/e3b6e4755eda0e7611694929d6b0529ee2c40d4b/Xamarin.Forms.Core/DataTemplateExtensions.cs#L17

To improve the situation, I think the two checks at the beginning of `ElementTemplate.CreateContent` should be the other way round (since in the case of a `DataTemplateSelector`, `LoadContent` is always null).

StephaneDelcroix:
I really wonder what you guys are doing to end up with a null there. Please attach a repro so we can fix it.

numb3r:
@samhouts @StephaneDelcroix 
Here's a link to my repo. 
Please uncomment section of code where i am using DataTemplate in view. 
https://github.com/numb3r/Page-dot-indicator-repo

StephaneDelcroix:
[...]

PR #4994, opened 2019-01-15
Removed redundant ItemTemplateSelector prop from BindableLayout

AndreiMisiukevich:
### Description of Change ###

Hi guys. From my point of view ItemTemplateSelectorProperty will confuse developers as it is not clear what it the priority of choosing ItemTemplate and ItemTemplateSelector.
Besides **DataTemplateSelector** inherits **DataTemplate** and definitely there is no need in ItemTemplateSelectorProperty (Like we have in **ListView** control)

Since **BindableLayout** is fresh class, i suggest to mark this property as obsolete and remove in further releases.

### API Changes ###
Suggest to remove BindableLayout.**ItemTemplateSelectorProperty**

### Platforms Affected ### 
- Core

None

Not applicable

### DIFF
--- a/Xamarin.Forms.Core/BindableLayout.cs
+++ b/Xamarin.Forms.Core/BindableLayout.cs
@@ -14,9 +14,8 @@ namespace Xamarin.Forms
 			BindableProperty.CreateAttached("ItemTemplate", typeof(DataTemplate), typeof(Layout<View>), default(DataTemplate),
 				propertyChanged: (b, o, n) => { GetBindableLayoutController(b).ItemTemplate = (DataTemplate)n; });
 
-		public static readonly BindableProperty ItemTemplateSelectorProperty =
-			BindableProperty.CreateAttached("ItemTemplateSelector", typeof(DataTemplateSelector), typeof(Layout<View>), default(DataTemplateSelector),
-				propertyChanged: (b, o, n) => { GetBindableLayoutController(b).ItemTemplateSelector = (DataTemplateSelector)n; });
+		[Obsolete("ItemTemplateSelectorProperty is obsolete. Please use ItemTemplateProperty instead.")]
+		public static readonly BindableProperty ItemTemplateSelectorProperty = ItemTemplateProperty;
 
 		static readonly BindableProperty BindableLayoutControllerProperty =
 			 BindableProperty.CreateAttached("BindableLayoutController", typeof(BindableLayoutController), typeof(Layout<View>), default(BindableLayoutController),
@@ -43,14 +42,16 @@ namespace Xamarin.Forms
 			return (DataTemplate)b.GetValue(ItemTemplateProperty);
 		}
 
+		[Obsolete("SetItemTemplateSelector is obsolete. Please use SetItemTemplate instead.")]
 		public static void SetItemTemplateSelector(BindableObject b, DataTemplateSelector value)
 		{
-			b.SetValue(ItemTemplateSelectorProperty, value);
+			b.SetValue(ItemTemplateProperty, value);
 		}
 
+		[Obsolete("GetItemTemplateSelector is obsolete. Please use GetItemTemplate instead.")]
 		public static DataTemplateSelector GetItemTemplateSelector(BindableObject b)
 		{
-			return (DataTemplateSelector)b.GetValue(ItemTemplateSelectorProperty);
+			return (DataTemplateSelector)b.GetValue(ItemTemplateProperty);
 		}
 
 		static BindableLayoutController GetBindableLayoutController(BindableObject b)
@@ -78,7 +79,6 @@ namespace Xamarin.Forms
 			newC.StartBatchUpdate();
 			newC.ItemsSource = GetItemsSource(b);
 			newC.ItemTemplate = GetItemTemplate(b);
-			newC.ItemTemplateSelector = GetItemTemplateSelector(b);
 			newC.EndBatchUpdate();
 		}
 	}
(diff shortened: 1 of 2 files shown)


andreinitescu:
>  ItemTemplateSelectorProperty will confuse developers as it is not clear what it the priority of choosing ItemTemplate and ItemTemplateSelector.

It shouldn't be confusing, the documentation should state it. The priority is clear and implemented accordingly, it's not random. ItemTemplate takes precedence over ItemTemplateSelector, like how it works on WPF and UWP. This is an old pattern in Windows XAML.

> Besides DataTemplateSelector inherits DataTemplate and definitely there is no need in ItemTemplateSelectorProperty (Like we have in ListView control)

You can't set a DataTemplateSelector instance to ItemTemplate, it will throw an exception. That's what the ItemTemplateSelector is for.
The fact that DataTemplateSelector inherits from DataTemplate is very unfortunate, and it has been discussed in other tickets, and it's the actual source of confusion.
However that doesn't mean it must to be perpetuated everywhere, and every new functionality should inherit same wrong behavior.
My suggestion is to encourage devs to move away from using DataTemplateSelector in place of DataTemplate, by adding a warning (see my comment https://github.com/xamarin/Xamarin.Forms/issues/3749#issuecomment-427277340)

AndreiMisiukevich:
@andreinitescu thanks for reaction. You are right re WPF.
But we should understand that Xamarin.Forms isn't WPF and has own infrastructure (like DataTemplateSelector inheritance or DependencyProperty/BindableProperty etc.)

Also regarding to @StephaneDelcroix  comment https://github.com/xamarin/Xamarin.Forms/issues/3749#issuecomment-442394925
we can realize that inheritance won't be change.
I think would be better to follow XF initial idea.

andreinitescu:
> But we should keep in mind, that there are dozens of 3rd party libraries, which were created with idea that DataTemplateSelector is subclass of DataTemplate.

Yes, but please note I didn't say to change the inheritance.

StephaneDelcroix:
image

BindableLayout hasn't been released yet, so it's fine to remove the API, no need to Obsolete it

Xamarin.Forms.Core/BindableLayout.cs
```diff
@@ -14,9 +14,8 @@ public static class BindableLayout
 			BindableProperty.CreateAttached("ItemTemplate", typeof(DataTemplate), typeof(Layout<View>), default(DataTemplate),
 				propertyChanged: (b, o, n) => { GetBindableLayoutController(b).ItemTemplate = (DataTemplate)n; });
 
-		public static readonly BindableProperty ItemTemplateSelectorProperty =
-			BindableProperty.CreateAttached("ItemTemplateSelector", typeof(DataTemplateSelector), typeof(Layout<View>), default(DataTemplateSelector),
-				propertyChanged: (b, o, n) => { GetBindableLayoutController(b).ItemTemplateSelector = (DataTemplateSelector)n; });
+		[Obsolete("ItemTemplateSelectorProperty is obsolete. Please use ItemTemplateProperty instead.")]
+		public static readonly BindableProperty ItemTemplateSelectorProperty = ItemTemplateProperty;
```
StephaneDelcroix: drop it

StephaneDelcroix:
> @StephaneDelcroix Care to share more thoughts? Thanks!

I'm all for consistency. DTS in XF is a kind of DT. we can disagree, we can choose to change that later on, but at this time, this PR makes sense.

andreinitescu:
> we can choose to change that later on

Once feature is released, we know it can't be changed...
I don't understand the sudden reason for a change.

> DTS in XF is a kind of DT

I'm sorry if I sound like a scratched record, but it can't be.
It's breaking Liskov substitution principle: DataTemplateSelector derives from DataTemplate but it can't be used in place of a DataTemplate. 
The implementation throws exception: https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/ElementTemplate.cs#L79

This created confusion and still does, some wonder why it doesn't work. See #3544 as just an example. The issue is still open since August last year, after several comments, and the reason is simple. 

But you guys are owners and know better, I'm here to help and provide feedback.

andreinitescu:
It's OK to remove ItemTemplateSelector if you want, as long as DataTemplateSelector is made to work as a DataTemplate, please see my comment https://github.com/xamarin/Xamarin.Forms/issues/3544#issuecomment-454689484

AndreiMisiukevich:
@andreinitescu I think it can make sense to obsolete **CreateContent()** method and replace it with **CreateContent (object context)**
Not sure.

andreinitescu:
the `CreateContent(context)` overload kind of makes sense, although, if you think about it why would all template classes need a context? 
But it's OK I guess... However it must be on the base ElementTemplate class, and it would require everyone to update their code to use it.

AndreiMisiukevich:
@StephaneDelcroix please re-review

andreinitescu:
In my opinion, this change should not be implemented before #3544 is figured out exactly.
Otherwise, the confusion which exists today with DataTemplateSelector will spread to BindableLayout and CollectionView too, and any other new controls using both the item template (DataTemplate) and item template selector (DataTemplateSelector).

mattleibow:
Based on the discussion here and a chat with @andreinitescu, I created this PR to hopefully move the discussion there and to solve some of the issues raised: https://github.com/xamarin/Xamarin.Forms/pull/5115

Feel free to hack that PR apart and give suggestions, comments and ask questions. One we resolve that, we can come back to this. That PR agrees with this PR, but in a slightly different way.

AndreiMisiukevich:
@StephaneDelcroix seems, bindable layout has jus been released

https://blog.xamarin.com/5-things-youll-love-xamarin-forms-3-5/

AndreiMisiukevich:
> image
> 
> BindableLayout hasn't been released yet, so it's fine to remove the API, no need to Obsolete it

seems, i need revert my last commit as BindableLayout is released.
Or, perhaps, this PR has no sense and we will live with two properties 😅

AndreiMisiukevich:
Hey guys @mattleibow @StephaneDelcroix 
what did we decide?

AndreiMisiukevich:
@mattleibow @StephaneDelcroix @andreinitescu 

So, guys, what is the our accepted approach? Do we preserve DTS and DT both?

samhouts:
Well, I guess we keep this as is. Thanks, @AndreiMisiukevich !

PR #5026, opened 2019-01-19
[iOS] Move Material iOS components to a separate nuget

### LINKED ISSUE #5020: [Material] Move Material renderers out of iOS default project
# Move the iOS material renderers to a separate assembly

Currently the user takes the hit for the material dll on ios even if they don't want to use material. Moving the renderers out to a separate dll will require users to opt in but for those that aren't going to use material they won't take the size hit to their application

Android material is all built into the design support libraries that are already included for other forms features so this doesn't apply to android

# API

```C#
public override bool FinishedLaunching(UIApplication uiApplication, NSDictionary launchOptions)
{
     Xamarin.Forms.Init();
     Xamarin.Forms.Material.Init();
```

### Properties

| API | Description |
| Xamarin.Forms.Material.Init() | This will cause the material dll to not be removed and it will setup the  material rendererers |

# Difficulty : medium

PureWeen:
### Description of Change ###

Move material iOS renderers out to a separate dll and added a required an init call from the iOS platform. The effect of this is that if you don't want to use material components then the material Nuget dll and the material renderers dll won't get included into the app package. If the user forgets the init call then they should see a warning about needing to add the Init call for material

### Issues Resolved ### 

- fixes #5020

### API Changes ###

Added required call to enable material on ios
```C#
FormsMaterial.Init()
```

Added:
 - FormsMaterial.Init()

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS

### Testing Procedure ###
- Remove the FormsMaterial.Init() call from the app delegate on control gallery
- verify that application still runs and material related dlls don't get added into the project

### PR Checklist ###

- [ ] Has automated tests 


PureWeen:
build --uitests

Xamarin.Forms.ControlGallery.iOS/AppDelegate.cs
```diff
@@ -162,6 +162,8 @@ public override bool FinishedLaunching(UIApplication uiApplication, NSDictionary
 			// Forms.SetFlags("CollectionView_Experimental", "Visual_Experimental", "Shell_Experimental");
 			Forms.Init();
 			FormsMaps.Init();
+			FormsMaterial.Init();
```
mattleibow: We might want to do this for all the other app projects as well - the pages and embedded I think

Xamarin.Forms.ControlGallery.iOS/AppDelegate.cs
```diff
@@ -454,4 +456,4 @@ public object ConvertBack(object value, Type targetType, object parameter, Cultu
 		}
 	}
 }
-#endif
+#endif
```
mattleibow: We probably don't need to do this.

Xamarin.Forms.ControlGallery.iOS/Xamarin.Forms.ControlGallery.iOS.csproj
```diff
@@ -149,6 +149,10 @@
       <IsAppExtension>false</IsAppExtension>
       <IsWatchApp>false</IsWatchApp>
     </ProjectReference>
+    <ProjectReference Include="..\Xamarin.Forms.Material.iOS\Xamarin.Forms.Material.iOS.csproj">
+      <Project>{8A75B1DC-CEED-4B1B-8675-A7DFFD1E6DE4}</Project>
+      <Name>Xamarin.Forms.Material.iOS</Name>
+    </ProjectReference>
```
mattleibow: See my comment above about the other test projects

Xamarin.Forms.Core/Properties/AssemblyInfo.cs
```diff
@@ -122,4 +123,4 @@
 [assembly: StyleProperty("-xf-shell-tabbar-title", typeof(Element), nameof(Shell.ShellTabBarTitleColorProperty), PropertyOwnerType = typeof(Shell))]
 [assembly: StyleProperty("-xf-shell-tabbar-unselected", typeof(Element), nameof(Shell.ShellTabBarUnselectedColorProperty), PropertyOwnerType = typeof(Shell))]
 [assembly: StyleProperty("-xf-shell-title", typeof(Element), nameof(Shell.ShellTitleColorProperty), PropertyOwnerType = typeof(Shell))]
-[assembly: StyleProperty("-xf-shell-unselected", typeof(Element), nameof(Shell.ShellUnselectedColorProperty), PropertyOwnerType = typeof(Shell))]
+[assembly: StyleProperty("-xf-shell-unselected", typeof(Element), nameof(Shell.ShellUnselectedColorProperty), PropertyOwnerType = typeof(Shell))]
```
mattleibow: We probably don't need to do this.

Xamarin.Forms.Core/Registrar.cs
```diff
@@ -328,4 +332,4 @@ public static void RegisterAll(Type[] attrTypes)
 			DependencyService.Initialize(assemblies);
 		}
 	}
-}
+}
```
mattleibow: We probably don't need to do this.

Xamarin.Forms.Core/Visual.cs
```diff
@@ -4,9 +4,20 @@ namespace Xamarin.Forms
...
+		static bool _isMaterialRegistered = false;
+
 		public static IVisual MatchParent { get; } = new VisualRendererMarker.MatchParent();
 		public static IVisual Default { get; } = new VisualRendererMarker.Default();
 		public static IVisual Material { get; } = new VisualRendererMarker.Material();
+
+		internal static void RegisterMaterial() => _isMaterialRegistered = true;
+		internal static void MaterialCheck()
+		{
+			if (_isMaterialRegistered)
```
mattleibow: How does this check work on other platforms? This looks like it will always be false on Android as there is no set from anywhere but iOS.

We could have the platform register this if it supports material...

Xamarin.Forms.Core/Visual.cs
```diff
@@ -4,9 +4,20 @@ namespace Xamarin.Forms
...
+				return;
+
+			Internals.Log.Warning("Visual", "Material needs to be registered on iOS by calling FormsMaterial.Init() after Forms.Init()");
```
mattleibow: What does this look like on Android or UWP? Maybe we should wrap this in a platform check...

Because we know that iOS needs a NuGet, we can have this message, but for the other platforms, we probably need another message - such as "Material is not supported on this platform." or something along those lines.

With regards to my comment above, I wouldn't actual write any `if (android) then don't log` code, but rather have the android platform actually call init. That way, we don't do anything special in here.

Xamarin.Forms.Material.iOS/Xamarin.Forms.Material.iOS.csproj
```diff
@@ -0,0 +1,76 @@
...
+    <MtouchVerbosity></MtouchVerbosity>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="System" />
+    <Reference Include="System.Xml" />
+    <Reference Include="System.Core" />
+    <Reference Include="Xamarin.iOS" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Xamarin.iOS.MaterialComponents" Version="72.2.0" />
```
mattleibow: We probably need to bump this (and the other cases) to the latest 72.2.0.1 as that has correct package information and the license check.

Xamarin.Forms.Core/Visual.cs
PureWeen: I opted to still just check for the platform in this code because we want the message to be different on iOS vs other platforms. We could add some hooks so iOS can register its own message but that seems like overkill

PureWeen:
build

rmarinho:
build

PureWeen:
@bholmes @Redth  assigned to you because of nuspec changes

Redth:
The dependency version for Xamarin.iOS.MaterialComponents LGTM in terms of license/attribution.

StephaneDelcroix:
shouldn't this be targeted at 3.5.0 ?

PureWeen:
alright @Redth @bholmes @davidortinau @mattleibow 

Separated this out into it's own nuget
- Nuget is called Xamarin.Forms.Visual.Material
- I setup dummy `_._` files in Monodroid and netstandard targets so the package installs into those which I feel is a better user experiencea

please review nuspec updates and let me know

PureWeen:
build

PureWeen:
build

PureWeen:
build

PR #5000, opened 2019-01-17
[Android] fixes TabStop on Stepper Control

paymicro:
### Issues Resolved ### 

partially fixes #4956 [Android]

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

### Behavioral/Visual Changes ###

Stepper don't skipped from with TabStop

### Testing Procedure ###

- Run Issue3872
- Tab Over controls
- You're able to tab to stepper

### PR Checklist ###

- [ ] Has automated tests 
- [x] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard

PR #5027, opened 2019-01-19
[Material] [Android, iOS] Materializing the stepper

### LINKED ISSUE #5011: [Material] Stepper Spec
# Material Contained Button Spec

There's no specific guidance for a stepper control but the stepper control is really just two buttons next to each other. The Android implementation is our own custom implementation doing just that. The material implementation will provide

- the buttons can just use Outlined Styled Material Buttons which looks like they should be able to match up with Android easily. If it proves too difficult to style the iOS control to look like Android then we can always create our own stepper control on iOS that's basically to material buttons next to each other. Since we've already done the work to make the Material Buttons look the same

Existing comparison
image
 

### Expectations 

- **Visually Identical**

The stepper control doesn't have any specific styling APIs currently that need to be considered.

# API
```XAML
<Stepper Visual="Material" />
```

```C#
new Stepper() { Visual = VisualMarker.Material }
```

# Difficulty : hopefully medium but could be high depending how hard it is to match our android to the iOS implementation

mattleibow:
### Description of Change ###

Adding a renderer for the material stepper:
 - refactored the actual stepper logic out so that all the controls can just implement it
 - added a new renderer for material

### Issues Resolved ### 

- fixes #5011 
- fixes #5000 

### API Changes ###
 
 - a new `MaterialStepperRenderer`
 - a new `StepperRendererManager` to manage all stepper logic

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android

None

**Material (with full-width characters / new characters)**

| Android | iOS |
| :-: | :-: |
|  | <image src="https://user-images.githubusercontent.com/1096616/52025060-b3079f80-250b-11e9-9d08-d61860f79ebb.png" width="350" /> |

**Default (with full-width characters)**

| Android | iOS |
| :-: | :-: |
|  | <image src="https://user-images.githubusercontent.com/1096616/52025059-b26f0900-250b-11e9-98e4-3c8bbc8ab931.png" width="350" /> |

**Material (with plus and hyphen / old characters)**

| Android | iOS |
| :-: | :-: |
|  | <image src="https://user-images.githubusercontent.com/1096616/51565371-8532b300-1e9a-11e9-8744-b2a39a154d7f.png" width="350" /> |

### Testing Procedure ###

Should just be a visual style change when using Material. Functionality should be exactly the same.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/StepperRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/StepperRenderer.cs
@@ -47,20 +47,22 @@ namespace Xamarin.Forms.Platform.Android
 				SetNativeControl(layout);
 			}
 
-			StepperRendererManager.UpdateButtons(this, null, _downButton, _upButton);
+			StepperRendererManager.UpdateButtons(this, _downButton, _upButton);
 		}
 
 		protected override void OnElementPropertyChanged(object sender, PropertyChangedEventArgs e)
 		{
 			base.OnElementPropertyChanged(sender, e);
-			StepperRendererManager.UpdateButtons(this, e, _downButton, _upButton);
+			StepperRendererManager.UpdateButtons(this, _downButton, _upButton, e);
 		}
 
 		Stepper IStepperRenderer.Element => Element;
 
-		AButton IStepperRenderer.GetButton(bool upButton) => upButton ? _upButton : _downButton;
+		AButton IStepperRenderer.UpButton => _upButton;
 
-		AButton IStepperRenderer.CreateButton(bool isUpButton)
+		AButton IStepperRenderer.DownButton => _downButton;
+
+		AButton IStepperRenderer.CreateButton()
 		{
 			var button = new AButton(Context);
 			button.SetHeight((int)Context.ToPixels(10.0));
--- a/Xamarin.Forms.Platform.Android/Renderers/StepperRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/StepperRenderer.cs
@@ -2,6 +2,7 @@ using System;
 using System.ComponentModel;
 using Android.Content;
 using Android.Widget;
+using Android.Views;
 using AButton = Android.Widget.Button;
 
 namespace Xamarin.Forms.Platform.Android
@@ -25,7 +26,12 @@ namespace Xamarin.Forms.Platform.Android
 
 		protected override LinearLayout CreateNativeControl()
 		{
-			return new LinearLayout(Context) { Orientation = Orientation.Horizontal };
+			return new LinearLayout(Context)
+			{
+				Orientation = Orientation.Horizontal,
+				Focusable = true,
+				DescendantFocusability = DescendantFocusability.AfterDescendants
+			};
 		}
 
 		protected override void OnElementChanged(ElementChangedEventArgs<Stepper> e)
--- a/Xamarin.Forms.Platform.Android/Renderers/StepperRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/StepperRenderer.cs
@@ -1,14 +1,12 @@
 using System;
 using System.ComponentModel;
 using Android.Content;
-using Android.Views;
 using Android.Widget;
 using AButton = Android.Widget.Button;
-using Object = Java.Lang.Object;
 
 namespace Xamarin.Forms.Platform.Android
 {
-	public class StepperRenderer : ViewRenderer<Stepper, LinearLayout>
+	public class StepperRenderer : ViewRenderer<Stepper, LinearLayout>, IStepperRenderer
 	{
 		AButton _downButton;
 		AButton _upButton;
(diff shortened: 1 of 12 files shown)


mattleibow:
When testing the tabstops fix in #5000 (which I have included here because I modified the renderer code), I noticed that the Material stepper doesn't seem to be working with tabs. I am not yet sure what the cause may be, but I do see a lot of extra code in the base renderers that we may need to re-implement.

mattleibow:
I reverted the changes to use a "Fast" renderer as it doesn't seem to work with multiple children views. I probably need to investigate this further, but for now steppers aren't the most important control.

mattleibow:
Thanks for the feedback, will get on that. In the meantime, I added the iOS one. Not sure if creating a new "control" is cool for iOS - or where I should put it.

This is not exactly what I wanted to do, since there is no need for the control - I can just do the layout in the renderer. But, I can't seem to find the concept of a no-native-view renderer. I am also thinking that this might also be good for Android.

Xamarin.Forms.Platform.Android/StepperRendererManager.cs
```diff
@@ -0,0 +1,68 @@
...
+			downButton.NextFocusForwardId = upButton.Id;
+		}
+
+		public static void UpdateButtons<TButton>(IStepperRenderer renderer, PropertyChangedEventArgs e, TButton downButton, TButton upButton)
+			where TButton : AButton
+		{
+			if (!(renderer?.Element is Stepper stepper))
+				return;
+
+			if (e == null ||
```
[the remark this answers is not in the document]
mattleibow: The check is not your traditional null check - if null, then update, otherwise check the properties. I use this same method (and bass null explicitly) to set up the buttons after creation.

Xamarin.Forms.Platform.iOS/Material/MaterialStepperRenderer.cs
```diff
@@ -0,0 +1,154 @@
...
+				var increment = stepper.Increment;
+				if (button == Control.DecrementButton)
+					increment = -increment;
+
+				stepper.SetValueFromRenderer(Stepper.ValueProperty, stepper.Value + increment);
+			}
+		}
+	}
+
+	public class MaterialStepper : UIView
```
mattleibow: I am not sure where we want to put this - this is very this-renderer specific and has no use elsewhere. Basically - just for layout.

Xamarin.Forms.Platform.iOS/Material/MaterialStepperRenderer.cs
```diff
@@ -0,0 +1,154 @@
...
+			if (e.PropertyName == Stepper.MinimumProperty.PropertyName ||
+				e.PropertyName == Stepper.MaximumProperty.PropertyName ||
+				e.PropertyName == Stepper.ValueProperty.PropertyName ||
+				e.PropertyName == VisualElement.IsEnabledProperty.PropertyName)
+			{
+				UpdateButtons();
+			}
+		}
+
+		private void UpdateButtons()
```
paymicro: Drop `private`. Also 2 times below.

mattleibow:
This comment is still up for discussion:
https://github.com/xamarin/Xamarin.Forms/pull/5027#pullrequestreview-195230863

Xamarin.Forms.Platform.Android/Renderers/StepperRenderer.cs
```diff
@@ -36,74 +40,33 @@ protected override void OnElementChanged(ElementChangedEventArgs<Stepper> e)
...
 
-				if (v == renderer._upButton)
-					((IElementController)stepper).SetValueFromRenderer(Stepper.ValueProperty, stepper.Value + stepper.Increment);
-				else if (v == renderer._downButton)
-					((IElementController)stepper).SetValueFromRenderer(Stepper.ValueProperty, stepper.Value - stepper.Increment);
-			}
+		AButton IStepperRenderer.CreateButton()
+		{
+			var button = new AButton(Context);
+			button.SetHeight((int)Context.ToPixels(10.0));
```
hartez: Where does this 10 come from?

Xamarin.Forms.Platform.Android/StepperRendererManager.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+	public static class StepperRendererManager
+	{
+		public static void CreateStepperButtons<TButton>(IStepperRenderer renderer, out TButton downButton, out TButton upButton)
+			where TButton : AButton
+		{
+			downButton = (TButton)renderer.CreateButton();
+			downButton.Id = Platform.GenerateViewId();
+			downButton.Focusable = true;
+			downButton.Text = "-";
```
hartez: Should we be using the minus sign (U+2212) here instead of the hyphen-minus (U+002D) here?
That is, should we be using "−" (minus sign) instead of "-" (the one you get from your keyboard)?

Xamarin.Forms.Platform.iOS/Material/MaterialStepperRenderer.cs
```diff
@@ -0,0 +1,154 @@
...
+		}
+
+		protected override void Dispose(bool disposing)
+		{
+			if (Control is MaterialStepper control)
+			{
+				control.DecrementButton.TouchUpInside -= OnStep;
+				control.IncrementButton.TouchUpInside -= OnStep;
+			}
+
```
hartez: Should this also be disposing _buttonScheme?

Xamarin.Forms.Platform.Android/Renderers/StepperRenderer.cs
mattleibow: That was always there: https://github.com/xamarin/Xamarin.Forms/blob/0dd61b71ec33c566ddcdf0fcb3aa2ec177b67399/Xamarin.Forms.Platform.Android/Renderers/StepperRenderer.cs#L40

Xamarin.Forms.Platform.Android/StepperRendererManager.cs
mattleibow: I like that.

Xamarin.Forms.Platform.iOS/Material/MaterialStepperRenderer.cs
mattleibow: Fixed.

Xamarin.Forms.Platform.Android/StepperRendererManager.cs
mattleibow: Going to try full-width characters ( ＋ /  － )

hartez: I'm down with the full-width hyphen-minus.

Xamarin.Forms.Platform.Android/Renderers/StepperRenderer.cs
hartez: Fair enough.

Xamarin.Forms.Platform.Android/StepperRendererManager.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+	public static class StepperRendererManager
+	{
+		public static void CreateStepperButtons<TButton>(IStepperRenderer renderer, out TButton downButton, out TButton upButton)
+			where TButton : AButton
+		{
+			downButton = (TButton)renderer.CreateButton();
+			downButton.Id = Platform.GenerateViewId();
+			downButton.Focusable = true;
+			downButton.Text = "－";
```
paymicro: I have concerns that the default font will not always be the same. May not even contain this symbol. Is it possible in the future to use icons instead of text?

mattleibow: Right now, no. We can add the ability to set icons, and that is probably a better option - especially with the new FontIconSource. That is the dream. But, do people actually use this control? :) It is basically 2 buttons with zero customisability. The only reason for this PR is to make the material demos better - setting a page to material, the stepper was the only buttons that weren't material - and there is absolutely no way to customize the font/size/color/anything. I am sure people just create their own - 2 buttons at 36x36, with a 4 space. Boom, you have a stepper, lol.

hartez: Could we explicitly set the font on these two buttons? Then we wouldn't have to worry about default font changes changing the appearance. (Scaling might still be an issue.)

mattleibow: That might be a breaking change as the initial stepper did not have text styles at all.

PureWeen:
build --uitests

PureWeen:
@mattleibow can you retarget to 3.6 as well please

mattleibow:
@PureWeen Thanks for spotting that. 

Now... I have a big question... Why would we respond to height requests and not width requests?
The old was is just pure random, but with material, maybe we should change that to fill the allocated space?

If I want it to fit, then I can use start/center/end depending, but if I want the buttons to be bigger/stretched, then I should respect the user's wishes. Only responding to the height is weird and probably the bad option here.

I think it might be better to just increate the button size to the size of the overall control. #VotingTime

PureWeen:
Oh yea the width also :-p

> I think it might be better to just increate the button size to the size of the overall control.

That would be my vote. Instead of just shaping the containing view. This way people can just make huge steppers.

PureWeen:
build --uitests

PureWeen:
@hartez or @paymicro  can you rereview Matt's latest changes

mattleibow:
Thanks fir the feedback, will get on it.

mattleibow:
@paymicro @PureWeen I had a look at the background color of the steppers, and I think we should hold off for right now. As these buttons are outline buttons - and theoretically should _never_ have a background. As such, there is no real gude for ignoring this behavior. I would rather add a "feature" later on to add the background color than to get it wrong here.

I would rather first do the correct styling for outline buttons in general, then just apply that here. In the meantime, I disable setting the background color.

mattleibow:
I jst want to note that the iOS accessibility is not working - you can't select the individual buttons.

Issue #2801, opened 2018-05-22
Android: ShowAlert returns Task which never completes when Activity is destroyed


andreinitescu:
### Description

Page.ShowAlert and other similar methods on Page return Task.
On Android, the Task never returns when Activity is destroyed.
This can be seen easily by using Developer options and have the option to recreate the activity when navigating away from it.
Android even gives a warning in logcat that the app leaks the Activity.

This issue was described in 2014 in the Xamarin Forms forum: https://forums.xamarin.com/discussion/18010/dialogs-leaking-memory-on-android

### Steps to Reproduce

```
class MainPage : Page
{ 
   public MainPage()
  {
       Do();
  }
  
  async void Do() {
      await ShowAlert("A", "B", "Cancel");  // This will never return even when Activity is recreated
  }
}
```

### Expected Behavior

Task should complete (or maybe throw TaskCanceledException?)

### Actual Behavior

Task never completes.

### More info 

The TaskCompletionSource used [here](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/AlertArguments.cs#L15) in AlertArguments class should be cancelled when Activity is destroyed or recreated such that code returns.
The Page.DisplayAlert returns the Task of the TaskCompletionSource [here ](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/Page.cs#L149)

```
public Task<bool> DisplayAlert(string title, string message, string accept, string cancel)
{
	if (string.IsNullOrEmpty(cancel))
		throw new ArgumentNullException("cancel");

	var args = new AlertArguments(title, message, accept, cancel);a
	MessagingCenter.Send(this, AlertSignalName, args);
	return args.Result.Task;
}
```

The TaskCompletionSource has result set [here](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/PopupManager.cs#L116) however in the case when Activity is destroyed or recreated, nor a result nor an exception is thrown to somehow tell the client awaiting on the ShowAlert to continue.

I think a partial solution is that in PopupRequestHelper:Dispose [here](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/PopupManager.cs#L56)  to have a way to cancel all the active TaskCompletionSource's.

I think this can apply to other kinds of dialogs\popups as well and I think there's a similar problem on UWP and iOS too.

[...]

PR #5028, opened 2019-01-20
[Core] Add some Device.InvokeOnMainThread helpers

knocte:
These helpers keep popping up on forums/stackoverflow[1][2],
would be nice to include them out of the box.

[1] https://forums.xamarin.com/discussion/51720/wait-for-device-begininvokeonmainthread
[2] https://stackoverflow.com/a/47941859

### DIFF
--- a/Xamarin.Forms.Core/Device.cs
+++ b/Xamarin.Forms.Core/Device.cs
@@ -104,6 +104,57 @@ namespace Xamarin.Forms
 			PlatformServices.BeginInvokeOnMainThread(action);
 		}
 
+		public static Task<T> InvokeOnMainThreadAsync<T>(Func<T> func)
+		{
+			var tcs = new TaskCompletionSource<T>();
+			BeginInvokeOnMainThread(() =>
+			{
+				try
+				{
+					var result = func();
+					tcs.SetResult(result);
+				}
+				catch (Exception ex)
+				{
+					tcs.SetException(ex);
+				}
+			});
+			return tcs.Task;
+		}
+
+		public static Task InvokeOnMainThreadAsync(Action action)
+		{
+			Func<object> dummyFunc = () => { action(); return null; };
+			return InvokeOnMainThreadAsync(dummyFunc);
+		}
+
+		public static Task<T> InvokeOnMainThreadAsync<T>(Func<Task<T>> funcTask)
+		{
+			var tcs = new TaskCompletionSource<T>();
+			BeginInvokeOnMainThread(
+				async() =>
+				{
+					try
+					{
+						var ret = await funcTask().ConfigureAwait(false);
+						tcs.SetResult(ret);
+					}
+					catch (Exception e)
+					{
+						tcs.SetException(e);
+					}
+				}
+			);
+
+			return tcs.Task;
+		}
+
+		public static Task InvokeOnMainThreadAsync(Func<Task> funcTask)
+		{
+			Func<Task<object>> dummyFunc = () => { funcTask(); return null; };
+			return InvokeOnMainThreadAsync(dummyFunc);
+		}
+
 		public static double GetNamedSize(NamedSize size, Element targetElement)
 		{
 			return GetNamedSize(size, targetElement.GetType());
--- a/Xamarin.Forms.Core/Device.cs
+++ b/Xamarin.Forms.Core/Device.cs
@@ -155,6 +155,15 @@ namespace Xamarin.Forms
 			return InvokeOnMainThreadAsync(dummyFunc);
 		}
 
+		public static async Task<SynchronizationContext> GetMainThreadSynchronizationContextAsync()
+		{
+			SynchronizationContext ret = null;
+			await InvokeOnMainThreadAsync(() =>
+				ret = SynchronizationContext.Current
+			).ConfigureAwait(false);
+			return ret;
+		}
+
 		public static double GetNamedSize(NamedSize size, Element targetElement)
 		{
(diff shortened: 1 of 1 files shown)


kvpt:
In this case I think that the **Begin** prefix is confusing because with **BeginInvoke** you just launch the action, it's a fire forget, but in this case, after the await of the call the action is terminated.

knocte:
@kvpt thanks, renamed

knocte:
@StephaneDelcroix fixed thanks

Xamarin.Forms.Core/Device.cs
```diff
@@ -104,6 +104,85 @@ public static void BeginInvokeOnMainThread(Action action)
 			PlatformServices.BeginInvokeOnMainThread(action);
 		}
 
+		public static Task InvokeOnMainThreadAsync(Action action)
+		{
+			var tcs = new TaskCompletionSource<object>();
```
paymicro: Perhaps `action` can reuse `Func` code
```C#
return InvokeOnMainThreadAsync(() => { action(); return true; });
```

Xamarin.Forms.Core/Device.cs
```diff
@@ -104,6 +104,85 @@ public static void BeginInvokeOnMainThread(Action action)
...
+				catch (Exception e)
+				{
+					tcs.SetException(e);
+				}
+			});
+
+			return tcs.Task;
+		}
+
+		public static Task InvokeOnMainThreadAsync(Task actionTask)
```
campersau: The task is only awaited on the main thread but not started on it. Is this the intended behaviour?

samhouts:
build

knocte:
> Perhaps `action` can reuse `Func` code

Done, thanks.

> The task is only awaited on the main thread but not started on it. Is this the intended behaviour?

Oh damn, that's a good point. I copied the implementation from here: https://stackoverflow.com/a/47941859 ; and didn't realise about it. I asked the author of the answer a question around this, and the answer doesn't sound very convincing to me. I guess I should remove the two overloads that receive Task objects? Thoughts?

AmrAlSayed0:
[xamarin/Essentials/Xamarin.Essentials/MainThread](https://github.com/xamarin/Essentials/tree/master/Xamarin.Essentials/MainThread) There are some here but they don't cover all of what is in this PR.

knocte:
> There are some here

I only see 2 methods there, same ones as we already have in Xamarin.Forms, no asynchronous API.

andreinitescu:
In my opinion, any new implementation which creates Task should make sure it works properly when Activity is recreated, see #2801

knocte:
> any new implementation

Not really new implementations, just wrappers around what's already there.

knocte:
> Oh damn, that's a good point. I copied the implementation from here: 
> https://stackoverflow.com/a/47941859 ; and didn't realise about it. I asked the author of the answer
> a question around this, and the answer doesn't sound very convincing to me. I guess I should
> remove the two overloads that receive Task objects? 

Replaced the overloads that received Task objects, with overloads that receive Func<Task> objects,
so that the action/function that generates the task is executed on the main thread. I think this should be good now.

knocte:
(Sorry, didn't mean to close. Just reopened.)

Xamarin.Forms.Core/Device.cs
```diff
@@ -104,6 +104,65 @@ public static void BeginInvokeOnMainThread(Action action)
...
+		{
+			return InvokeOnMainThreadAsync(funcTask);
+		}
+
+		public static async Task<SynchronizationContext> GetMainThreadSynchronizationContextAsync()
+		{
+			SynchronizationContext ret = null;
+			await InvokeOnMainThreadAsync(() =>
+				ret = SynchronizationContext.Current
+			);
```
StephaneDelcroix:
```suggestion
			).ConfigureAwait(false);
```

rmarinho:
Hey @knocte  can you rebase please ? thanks

knocte:
Why is a rebase needed if there are no conflicts?

StephaneDelcroix:
build

knocte:
BTW I cannot inspect what's wrong with CI, the "OSX Phase" links me to somewhere that ends up showing a 401 error.

StephaneDelcroix:
@knocte I'm rebuilding to figure that out

StephaneDelcroix:
@knocte the build issue on the CI is some provisioning problem on iOS. that's why @rmarinho requested a rebase. It'd be nice if you could do it.

knocte:
rebased

StephaneDelcroix:
build

StephaneDelcroix:
@rmarinho please figure out what's still wrong with provisioning and rebased PR.

Xamarin.Forms.Core/Device.cs
```diff
@@ -104,6 +104,65 @@ public static void BeginInvokeOnMainThread(Action action)
...
+					}
+				}
+			);
+
+			return tcs.Task;
+		}
+
+		public static Task InvokeOnMainThreadAsync(Func<Task> funcTask)
+		{
+			return InvokeOnMainThreadAsync(funcTask);
```
campersau: This calls itself. Do something like this:

```cs
Func<Task<object>> dummyFunc = () => { funcTask(); return null; };
return InvokeOnMainThreadAsync(dummyFunc);
```

StephaneDelcroix: good catch

knocte: Thanks, fixed.

PR #5034, opened 2019-01-22
Added the ability to Target a bindable object for Setters

### LINKED ISSUE #4924: [F100] Add "Target" support to VisualStateManager
## Summary
This probably should be in this issue (https://github.com/xamarin/Xamarin.Forms/issues/4232), but I thought I would create a new one as this might be a smaller task.

I was porting my UWP app to Xamarin.Forms and I noticed that the VisualStateManager does not allow states to affect child controls. Take this pseudo xaml:

```xaml
<Grid>
    <VisualStateManager>
        <State Name="Big">
            <!-- New feature can change children -->
            <!-- UWP style -->
            <Setter Target="myButton.Color" Value="Red" />
            <!-- WPF style -->
            <Setter TargetName="myButton" Property="Color" Value="Red" />

            <!-- Current feature only changes the current element -->
           <Setter Target="BackgroundColor" Value="Red" />
        </State>
    </VisualStateManager>

    <Button x:Name="myButton" Color="Blue" />

</Grid>
```

This XAML allows the VSM to modify properties on the children, not just the current element.
https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.setter.target#Windows_UI_Xaml_Setter_Target
https://docs.microsoft.com/en-us/dotnet/api/system.windows.setter.targetname?view=netframework-4.7.2#System_Windows_Setter_TargetName

## API Changes

I am not sure if the UWP or the WPF route is better, but we might just be able to add a `TargetName` property to `Setter` and then instead of a `this.SetProperty(PropertyName, Value)` we would `this.FindElement(TargetName).SetProperty(PropertyName, Value)`.

## Intended Use Case

This will solve the issue of multiple VSM elements in a single page that responds to a single event, such as a state for a wide screen or a narrow screen, or for device orientation. One VSM can control the entire page.

The use case that I was working on was on a narrow screen, I would have 2 columns in my grid, but on a wide screen, I would have 4 and show a big header with more features.

### LINKED ISSUE #5622: CollectionView SelectedItem Color
### Description
When no BackgroundColor is set to ItemTemplate, the SelectedItem shows a Orange background color, when you set a BackgroundColor in ItemTemplate, then SelectedItem has the same color than the rest of the items and you could not Identify the SelectedItem.
It would be appreciated also being able to set the BackgroundColor / BorderColor of selectedItem instead of just a orange color.
### Steps to Reproduce

Having this XAML
```
<ContentPage>
        <Grid>
            <Grid.ColumnDefinitions>
                <ColumnDefinition />
                <ColumnDefinition Width="auto" />
            </Grid.ColumnDefinitions>
            <CollectionView Grid.Column="0" 
                            SelectionMode="Single"
                            SelectedItem="{Binding SelectedItem, Mode=TwoWay}"
                            ItemsSource="{Binding ObservableCollectionTest}">
                <CollectionView.ItemsLayout>
                    <GridItemsLayout Orientation="Vertical" />
                </CollectionView.ItemsLayout>
                <CollectionView.ItemTemplate>
                    <DataTemplate>
                        <Grid BackgroundColor="Wheat">
                            <Label Text="{Binding}"></Label>
                        </Grid>
                    </DataTemplate>
                
            </CollectionView>
            <Label Grid.Column="1"
                   Text="{Binding SelectedItem}" />
        </Grid>
    </ContentPage>
```
With this data in the Model
```
public static string SelectedItemPropertyName() { return nameof(SelectedItem); }
        private string _SelectedItem;
        public string SelectedItem
        {
            get
            {
                return _SelectedItem;
            }
            set
            {
                _SelectedItem = value;
                RaisePropertyChanged();
            }
        }
        public static string ObservableCollectionTestPropertyName() { return nameof(ObservableCollectionTest); }
        private ObservableCollection<string> _ObservableCollectionTest = new ObservableCollection<string>(new List<string>() { "Test1", "test2", "test3" } );
        public ObservableCollection<string> ObservableCollectionTest
        {
            get
            {
                return _ObservableCollectionTest;
            }
            set
            {
                _ObservableCollectionTest = value;
                RaisePropertyChanged();
            }
        }
SelectedItem = ObservableCollectionTest[1];
```
### Expected Behavior
SelectedItem having a different background color or at least a border color
### Actual Behavior
It looks like the rest of the items
### Basic Information

- Platform Target Frameworks: 
  - Android: 8.1

brainoffline:
### Description of Change ###

Adds the ability to target a child object for a Setter for visual state management or Styles.
See VisualStateManagerTests.xaml for an example.

```
		<VisualStateGroup Name="ColorStates">
			<VisualState Name="Red">
				<VisualState.Setters>
					<Setter TargetName="TargetLabel1" Property="Label.BackgroundColor" Value="#340002" />
					<Setter TargetName="TargetLabel1" Property="Label.TextColor" Value="#920e0c" />
					<Setter TargetName="TargetLabel1" Property="Label.Text" Value="Red" />
				</VisualState.Setters>
			</VisualState>

			<VisualState Name="Blue">
				<VisualState.Setters>
					<Setter TargetName="TargetLabel1" Property="Label.BackgroundColor" Value="#4a707a" />
					<Setter TargetName="TargetLabel1" Property="Label.TextColor" Value="#94b0b7" />
					<Setter TargetName="TargetLabel1" Property="Label.Text" Value="Blue" />
				</VisualState.Setters>
			</VisualState>
		</VisualStateGroup>
```

### Issues Resolved ### 

- closes #4924 [Enhancement] Add "Target" support to VisualStateManager
- closes #5622

### API Changes ###

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [x] Has automated tests 


mattleibow:
I have another question... How does this work with styles? One of the things that I used this part of the VSM is to apply a style to all instances of a custom control. Then, I would change the state, and the manager would update sub views. Similar thing with templates - we don't have access to the actual instance at the time of creation.

Not sure what this all means, but is a lookup for the element by name better than a reference? If this all works, then this is fine.

brainoffline:
@mattleibow That is not the separation of concerns that I would normally take.
My approach would be that the LabelledEntry would be self-contained and expose maybe a `HasError` property.  Based on that property, it would be the responsibility of the `LabelledEntry` control to change it's own visual state.  
Therefor, there is no problem with referring to an internal control (x:Name="TextEntry") from the page's code-behind.  It will also enable binding to the `HasError` property.

mattleibow:
How does this work with the VSM added via Style? The same issue occurs, you don't have a reference to any objects - but you want to add a state.

You _could_ add the state to the actual control, but you loose the ability to have different state styles.

The case I actually noticed this was with different screen sizes/orientation. When I was in a wide screen, I moved some elements around (which can use references), but then changed the layout of items in a list (which was added via data binding/templates). I _could_ have created a control to wrap the items and their layout, but that means code. The whole point of XAML is to not write code for visual things.

For example, here is a sample xaml file that defines some control and uses the visual state manager and styles for a certain look. When the screen goes wide, it changes:
https://gist.github.com/mattleibow/2f2f33c353be0b33e31bf9007095b105

This example is a little weird in that I use n `AdaptiveTrigger` which is related to the window size, but that is just for demo purposes. It could very well be a custom trigger that just decides based on device type (phone / desktop / tv). Using this forn of styling, You can define your UI once and then just move things around based on screen size or app type. There are alternatives to actually re-create the entire template for each style, but that is a lot of duplication of XAML.

brainoffline:
Great feedback.  Thank you @mattleibow 
I have added additional flexibility when selecting the target to apply the values to. 

```
<Setter Target="{x:Reference TargetLabel1}" Property="Label.BackgroundColor" Value="#4a707a" />
<Setter TargetName="TargetLabel1" Property="Label.TextColor" Value="#94b0b7" />
```

To select which object to update, the search starts where the VSM is defined.  i.e. Page level, control level, ...
If Target is present, then that object will be selected.
If TargetName is present, it will search under the currently selected object (VSM parent or Target) to find the correct BindableObject to select.
Values are then applied to the Property of the selected object.

This may be a little more complex to document, but it should work for more cases.

Xamarin.Forms.Core/Setter.cs
```diff
@@ -101,5 +108,24 @@ internal void UnApply(BindableObject target, bool fromStyle = false)
 			else
 				target.ClearValue(Property);
 		}
+
+		BindableObject FindTargetObject(BindableObject root, BindableObject target, string targetName)
+		{
+			root = target ?? root;
+			if (root == null)
+				return null;
+			if (!string.IsNullOrWhiteSpace(targetName))
```
StephaneDelcroix: invert the condition

Xamarin.Forms.Core/Setter.cs
```diff
@@ -14,6 +14,10 @@ public sealed class Setter : IValueProvider
 	{
 		readonly ConditionalWeakTable<BindableObject, object> _originalValues = new ConditionalWeakTable<BindableObject, object>();
 
+		public BindableObject Target { get; set; }
+
+		public string TargetName { get; set; }
```
StephaneDelcroix: I find weird that you can use either Target or TargetName, as they do not look mutually exclusive

StephaneDelcroix:
needs to be targeted to `master`

brainoffline:
Just slapped it with the `Simple` stick.  Less complexity = less problems.

I have removed Target from the Setter.
This leaves TargetName as a way of targeting child elements.  This is the same model used by WPF.

If TargetName is present, then it will search the NameScope of the applied bindable object (basically child elements).  
This technique applies to both Style Setters as well as VSM Setters.  They are essentially the same thing.

mattleibow:
I also liked the fact that a reference could be set... That might have been a performance boost - not requiring a search through the views.

But then things get complicated for the user... 

But, we can also be extra clever and make the type object... If it is a string, we do a lookup. If it is a bind able object, we use that directly. Or, we can go the route of creating a new type that implicitly converts from both:

```csharp
class BindableTargetObject {
    public string TargetName;
    public BindableObject Target;
    public BindableObject GetTarget(BindableObject scope) =>
        Target ?? scope?.FindObject(TargetName) ?? scope;
} 
```

A bit more complex, but actually quite simple - especially since this works with Styles.

brainoffline:
More awesome feedback.
I am not sure how introducing a `BindableTargetObject` would look in the Xaml.  I have gone with the simpler approach of using an object.
I have changed the property name back to `Target` and the developer can either using a x:Reference or name of object.  Reference is quicker and name has more flexibility around templates and stuff.

Xamarin.Forms.Core/Setter.cs
```diff
@@ -47,59 +60,76 @@ object IValueProvider.ProvideValue(IServiceProvider serviceProvider)
...
 			}
 			else
-				target.ClearValue(Property);
+				targetObject.ClearValue(Property);
+		}
+
+		BindableObject FindTargetObject(BindableObject scope, object target)
+		{
+			if (scope == null || target == null)
+				return scope;
```
mattleibow: Is this necessary? I think the next 2 if statements will catch any nulls. The only case that looks to change is if scope is null, but target is BindableObject. But, we are returning the object directly.

Xamarin.Forms.Core/Setter.cs
```diff
@@ -14,6 +14,19 @@ public sealed class Setter : IValueProvider
 	{
 		readonly ConditionalWeakTable<BindableObject, object> _originalValues = new ConditionalWeakTable<BindableObject, object>();
 
+		/// <summary>
```
rmarinho: We don't use inline comments

StephaneDelcroix:
> Just slapped it with the Simple stick. Less complexity = less problems.
>
> I have removed Target from the Setter.
> This leaves TargetName as a way of targeting child elements. This is the same model used by WPF.

I would be ok with this, but that's not what the code looks like right now

samhouts:
@brainoffline Are you able to address the comments above? Let us know if we should take this over. Thanks!

andreinitescu:
This is a powerful and much needed feature, which has the potential to save a lot of custom code.

@brainoffline Could you please say if you're still able to work on this? If not, I could try give it a shot please. Please let me know, thanks!

jfversluis:
Hey @brainoffline thanks for your awesome work! We would like to get this in soon. Since you haven’t responded in some time, I’m going to assume that you do not have the possibility to finish it.

@andreinitescu you’ve offered. Are you able to make some progress on this? Else we will take it ourselves

andreinitescu:
Yes, I did offer, but I got no reply if I could actually start the work on it. 
But, since you seem eager to work on it, sure, please go ahead, I don't want to slow releasing new features to the framework...

andreinitescu:
Most of the work implementing `string TargetName` was done already by @brainoffline, my suggestion is it would be good to support `object Target` like in UWP.

samhouts:
@andreinitescu It looks like you may be able to get to this before us! Still willing? Thanks!

bezysoftware:
Was this pushed to v4.4.0? https://github.com/xamarin/Xamarin.Forms/wiki/Feature-Roadmap still indicated v4.3.0

jfversluis:
Thanks for asking @bezysoftware!

The thing I can say for sure is that it won't be in 4.3 unfortunately. I will do my absolute best to get it into 4.4. The PR is mostly ready, so I think that should be possible :)

samhouts:
closing in favor of #8144

PR #5054, opened 2019-01-24
Android Perf Builds

kingces95:
Produce a release android nuget package like this:
```
> .create-nuget.bat rdroid
```


.create-nuget.bat
```diff
@@ -7,147 +7,46 @@ rem This is not ideal, but it's better than nothing, and it usually works fine.
...
 )
+if "%1" == "rdroid" (
+   set CONFIG=release
+   call .create-stubs.bat
+   %NUGET_EXE% restore .xamarin.forms.android.nuget.sln
+   msbuild /v:m /p:configuration=release /p:platform="any cpu" /p:WarningLevel=0 /p:CreateAllAndroidTargets=true .xamarin.forms.android.nuget.sln
+)
+if "%1" == "pdroid" (
+   set CONFIG=release
+   msbuild /v:m /p:configuration=release /p:platform="anyCpu" /p:WarningLevel=0 /p:CreateAllAndroidTargets=true Xamarin.Forms.Platform.Android\Xamarin.Forms.Platform.Android.csproj
```
mattleibow: I see this has a `anyCpu` instead of `any cpu` - is that intended?

PR #5059, opened 2019-01-25
Using nameof instead of string literal

Thomas-S-B:
### Description of Change ###

Cleaned code: Using nameof instead of string literal


StephaneDelcroix:
Thanks for this. But unfortunately we do not do cosmetic changes if we don't have to. Not because we like our codebase dirty, but because it hurts history and mergeability.

PR #5074, opened 2019-01-27
Cake build script

PureWeen:
### Description of Change ###
Set of cake targets for easily building/packaging/deploying Forms

### Targets
- Clean for removing bin/objs
- Build for just building SLN
- NugetPack for packing into nugets 
  - ./build.ps1 -Target NugetPack -ScriptArgs '-packageVersion="9.9.9-custom"'
  - Currently only works on windows but could create a nuspec for ios or could convert nuspec files to nuspecsettings in cake
  - ./build.ps1 -Target NugetPack  uses GitVersion.CommandLine to generate the version from the tag or the branch name if you haven't specified a package version at the command line
- VSMAC builds the correct multi targeted projects and opens the SLN in vmac.
  - Currently there's a multi targeting bug with VSM that doesn't let it be opened
- DeployAndroid builds an apk and deploys it to a connected device and starts the gallery. 
  - useful for putting up a new apk when running ui tests

#### Windows

```ps1
# Generates nuget based on branch
 ./build.ps1 -Target NugetPack

# Deploy to whatever android emulator you have running
 ./build.ps1 -Target DeployAndroid

# Deletes all the bin/obj folders
 ./build.ps1 -Target Clean
```

#### Mac

```sh
# Builds necessary projects to run CG on mac and opens sln on vsmac
 ./build.sh1 --target vsmac
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,4 @@ AndroidNative/gradlew.bat
 *.swp
 tools/
 !tools/mdoc/**/*
+caketools/
--- a/.gitignore
+++ b/.gitignore
@@ -63,4 +63,4 @@ AndroidNative/gradlew.bat
 *.pdb
 # temporary vim files
 *.swp
-tools/
+caketools/
--- a/.gitignore
+++ b/.gitignore
@@ -63,4 +63,5 @@ AndroidNative/gradlew.bat
 *.pdb
 # temporary vim files
 *.swp
-caketools/
+tools/
+!tools/mdoc/**/*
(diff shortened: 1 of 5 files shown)


PureWeen:
@rmarinho  thoughts on @kingces95 comment? Do you think running our builds through these scripts would be useful?  I like the idea because then we have a script that consistent with our CI

If the script breaks CI breaks and we have to fix script which means we maintain a working script for the people

rmarinho:
@PureWeen we could use in the future no problem, but we need to first migrate to yaml, then we update yaml files for this new stuff.

PR #5079, opened 2019-01-28
[Android Material] Linear Progress Indicator

### LINKED ISSUE #5008: [Material] Linear Progress Indicator
# Material Linear Progress Indicator

https://material.io/develop/ios/components/progress-indicators/progress-views/

- **Visually Identical**

ProgressBar doesn't expose very many apis so the expectation here is that the default would size and color itself the same on iOS and Android

The hopefully not too tricky part here will be that there is only a Material ProgressBar iOS. On the android one it's still just using the original control so material visual on Android will just apply default styles or just inherit from the existing renderer

#### Current state
image

# Scenarios

```XAML
<ProgressBar Visual="Material" />
```

# Difficulty : medium

paymicro:
### Description of Change ###

fixes Linear Progress Indicator

### Issues Resolved ### 

- fixes #5008

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

Not applicable

### Testing Procedure ###

- run [Material] ProgressBar Gallery
- check Visual Gallery

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/Material/MaterialProgressBarRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Material/MaterialProgressBarRenderer.cs
@@ -13,13 +13,14 @@ using AColor = Android.Graphics.Color;
 using AProgressBar = Android.Widget.ProgressBar;
 using AView = Android.Views.View;
 
-[assembly: ExportRenderer(typeof(Xamarin.Forms.ProgressBar), typeof(MaterialProgressBarRenderer), new[] { typeof(VisualRendererMarker.Material) })]
+[assembly: ExportRenderer(typeof(ProgressBar), typeof(MaterialProgressBarRenderer), new[] { typeof(VisualRendererMarker.Material) })]
 
 namespace Xamarin.Forms.Platform.Android.Material
 {
 	public class MaterialProgressBarRenderer : AProgressBar,
 		IVisualElementRenderer, IViewRenderer, ITabStop
 	{
+		const float BackgroundAlpha = 0.6f;
 		const int MaximumValue = 10000;
 
 		int? _defaultLabelFor;
@@ -27,6 +28,8 @@ namespace Xamarin.Forms.Platform.Android.Material
 		bool _disposed;
 
 		ProgressBar _element;
+
+		AColor _defaultPrimaryColor => MaterialColors.Light.PrimaryColor;
 
 		VisualElementTracker _visualElementTracker;
 		VisualElementRenderer _visualElementRenderer;
@@ -122,7 +125,6 @@ namespace Xamarin.Forms.Platform.Android.Material
 		protected virtual void OnElementPropertyChanged(object sender, PropertyChangedEventArgs e)
 		{
 			ElementPropertyChanged?.Invoke(this, e);
-
 			if (e.PropertyName == ProgressBar.ProgressProperty.PropertyName)
 				UpdateProgress();
 			else if (e.PropertyName == ProgressBar.ProgressColorProperty.PropertyName || e.PropertyName == VisualElement.BackgroundColorProperty.PropertyName)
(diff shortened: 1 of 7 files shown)


PureWeen:
@paymicro can you rebase this to master when you get a chance?

PureWeen:
build --uitests

Xamarin.Forms.Platform.Android/Material/MaterialProgressBarRenderer.cs
```diff
@@ -13,13 +14,14 @@
...
 
-[assembly: ExportRenderer(typeof(Xamarin.Forms.ProgressBar), typeof(MaterialProgressBarRenderer), new[] { typeof(VisualRendererMarker.Material) })]
+[assembly: ExportRenderer(typeof(ProgressBar), typeof(MaterialProgressBarRenderer), new[] { typeof(VisualRendererMarker.Material) })]
 
 namespace Xamarin.Forms.Platform.Android.Material
 {
 	public class MaterialProgressBarRenderer : AProgressBar,
 		IVisualElementRenderer, IViewRenderer, ITabStop
 	{
+		const float BackgroundAlpha = 0.3f;
```
mattleibow: Where did we get this value from?

Xamarin.Forms.Platform.Android/Resources/values/styles.xml
```diff
@@ -2,7 +2,10 @@
 <resources>
   <style name="XamarinFormsMaterialTheme" parent="Theme.MaterialComponents.Light.DarkActionBar">
   </style>
-  <style name="XamarinFormsMaterialProgressBarHorizontal" parent="Widget.AppCompat.ProgressBar.Horizontal">
+  <style name="XamarinFormsMaterialProgressBarHorizontal" parent="android:Widget.ProgressBar.Horizontal">
+    <item name="android:indeterminateOnly">false</item>
+    <item name="android:minHeight">4dp</item>
+    <item name="android:progressDrawable">@drawable/MaterialProgressBar</item>
```
mattleibow: Why did you opt to use the base control and a custom drawing?

Xamarin.Forms.Platform.Android/Material/MaterialProgressBarRenderer.cs
```diff
@@ -142,51 +146,11 @@ void UpdateColors()
...
-				ProgressBackgroundTintList = ColorStateList.ValueOf(background);
-
-				// TODO: Potentially override alpha to match material design.
-				ProgressBackgroundTintMode = PorterDuff.Mode.SrcOver;
-			}
+			var primary = Element.ProgressColor.IsDefault ? _defaultPrimaryColor : Element.ProgressColor.ToAndroid();
+			ProgressTintList = ColorStateList.ValueOf(primary);
+			var background = Element.BackgroundColor.IsDefault ? _defaultBackgroundColor : Element.BackgroundColor.ToAndroid();
+			ProgressBackgroundTintList = ColorStateList.ValueOf(background);
+			ProgressBackgroundTintMode = PorterDuff.Mode.SrcIn;
```
mattleibow: I think we need to have a fallback to follow material and if only the progress color is set, then we have the lighter background of that color - unless specifically set. Right now, it looks that if I set the color to purple, only the main color is set, the track remains black - immediately breaking the material design. If we want to allow the user to break the guides, then we need to let them do the work to get there.

Xamarin.Forms.Platform.Android/Resources/values/styles.xml
mattleibow: Just got reminded by Shane, Google has a _HARD_ set for 4dp. Of course, this is absolutely not true for iOS, but hey, why follow your own guides.

mattleibow:
@paymicro Could you attach screenshots of before/after to the issue for comparison and records (for when Google breaks the controls and the style suddenly changes under us)?

Xamarin.Forms.Platform.Android/Resources/drawable/MaterialProgressBar.xml
```diff
@@ -0,0 +1,51 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- 
+From original drawable https://android.googlesource.com/platform/frameworks/base/+/master/core/res/res/drawable/progress_horizontal.xml 
+with removed rounded corners.
```
mattleibow: Maybe use the AppCompat - if possible - and use this style: https://github.com/aosp-mirror/platform_frameworks_base/blob/master/core/res/res/drawable/progress_horizontal_material.xml

Xamarin.Forms.Platform.Android/Material/MaterialProgressBarRenderer.cs
mattleibow: Yay. Google. This is typical for Material - neither platform matches the spec.

Xamarin.Forms.Platform.Android/Material/MaterialProgressBarRenderer.cs
```diff
@@ -13,13 +13,14 @@
...
 
-[assembly: ExportRenderer(typeof(Xamarin.Forms.ProgressBar), typeof(MaterialProgressBarRenderer), new[] { typeof(VisualRendererMarker.Material) })]
+[assembly: ExportRenderer(typeof(ProgressBar), typeof(MaterialProgressBarRenderer), new[] { typeof(VisualRendererMarker.Material) })]
 
 namespace Xamarin.Forms.Platform.Android.Material
 {
 	public class MaterialProgressBarRenderer : AProgressBar,
 		IVisualElementRenderer, IViewRenderer, ITabStop
 	{
+		const float BackgroundAlpha = 0.6f;
```
mattleibow: I think this constant should be moved to the `MaterialColors` type (as the rest of the values like this are there)

mattleibow:
Merged this into the material branch for the slider / progress as they are the same control and I want to reuse some of the coloring code. (and the nice gallery)

PR #5080, opened 2019-01-28
Update support dependencies to latest 28.0.0.1 version

PureWeen:
### Description of Change ###
Latest version of the support libraries help to mitigate an issue that causes VS to freeze when there is a mismatch of support versions from transient dependencies


PureWeen:
build --uitests

PR #5082, opened 2019-01-29
Correctly load addresses of intermediate values in compiled binding paths

akihikodaki:
### Description of Change ###

The intermediate values in compiled binding paths must be loaded as addresses when calling the subsequent property getter. However, the preceding property getters directly return the values if they are value types. This change adds a code to convert those values into addresses.

### Issues Resolved ### 

None

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [X] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Build.Tasks/SetPropertiesVisitor.cs
+++ b/Xamarin.Forms.Build.Tasks/SetPropertiesVisitor.cs
@@ -526,6 +526,8 @@ namespace Xamarin.Forms.Build.Tasks
 				il.Emit(Ret);
 			}
 			else {
+				var locs = new Dictionary<TypeReference, VariableDefinition>();
+
 				if (tSourceRef.IsValueType)
 					il.Emit(Ldarga_S, (byte)0);
 				else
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Build.Tasks/SetPropertiesVisitor.cs
```diff
@@ -526,6 +526,8 @@ static IEnumerable<Instruction> CompiledBindingGetGetter(TypeReference tSourceRe
 				il.Emit(Ret);
 			}
 			else {
+				var locs = new Dictionary<TypeReference, VariableDefinition>();
```
StephaneDelcroix: if have a feeling that storing vardef using the TypeRef as key might cause problems if the same typeref is found more than once in the binding path...

`{Binding foo.bar.baz.qux}` when foo and baz have the same (value)type, hence the same TypeRef...

akihikodaki: Such uses of those local variables have exclusive lifetime. The use of local variable which begins with the instruction emitted at line 549 always ends with the instruction emitted at line 558, 593, or 595, and the variable is not reused in the range.

PR #5088, opened 2019-01-29
[Android iOS Material] ActivityIndicator

### LINKED ISSUE #5007: [Material] Circular Progress Indicator
# Material Circular Progress Indicator

https://material.io/design/components/progress-indicators.html#circular-progress-indicators

- **Visually Identical**

ActivityIndicator doesn't expose very many apis so the expectation here is that the default would size and color itself the same on iOS and Android

The hopefully not too tricky part here will be that there is only a Material Activity Indicator on iOS. On the android one it's still just using the original control so material visual on Android will just apply default styles or just inherit from the existing renderer

#### Current state of implementation
image

 
# Scenarios

```XAML
<ActivityIndicator Visual="Material"  />
```

# Difficulty : medium

paymicro:
### Issues Resolved ### 

- fixes #5007

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS
- Android

None

**Before**
image

**After**
screenshot_13

### Testing Procedure ###

- run Visual Gallery

### PR Checklist ###

- [ ] Has automated tests 


Xamarin.Forms.Material.iOS/MaterialActivityIndicatorRenderer.cs
```diff
@@ -32,6 +34,14 @@ protected override void OnElementChanged(ElementChangedEventArgs<ActivityIndicat
...
 
 					SetNativeControl(CreateNativeControl());
+
+					_backgroundLayer = new CAShapeLayer
+					{
+						LineWidth = 4,
+						FillColor = UIColor.Clear.CGColor,
+						Hidden = true
+					};
+					Control.Layer.InsertSublayer(_backgroundLayer, 0);
```
mattleibow: Nice! Like the fact we can just put our own background in here.

Xamarin.Forms.Material.iOS/MaterialActivityIndicatorRenderer.cs
```diff
@@ -80,36 +98,17 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
...
-			//       For now, just show/hide the track.
-
-			if (color.IsDefault)
-				Control.TrackEnabled = false;
-			else
-				Control.TrackEnabled = true;
-
-			// handle the actual background color next to the main color
-			UpdateColor();
+			_backgroundLayer.Path = UIBezierPath.FromArc(Control.Center, Control.Radius - Control.StrokeWidth / 2, 0, 360, true).CGPath;
```
mattleibow: Is it possible to set this path only when the radius/stroke width changes? We are creating a new native object and a managed object here. Also, is it possible to have this layer "scale"? Not sure of iOS's powers, but can we create a circle layer and let iOS resize it? I don't think the actual indictor has properties for the radius and stroke width, so we might just have to update when the size changes.

Xamarin.Forms.Platform.Android/Material/MaterialActivityIndicatorRenderer.cs
```diff
@@ -136,32 +135,32 @@ public override bool OnTouchEvent(MotionEvent e)
...
+					: Element.Color.ToAndroid();
 				_control.IndeterminateTintList = ColorStateList.ValueOf(progress);
 			}
 			else
 			{
-				var progress = progressColor.ToAndroid();
-				_control.IndeterminateTintList = ColorStateList.ValueOf(progress);
+				_control.Visibility = Element.BackgroundColor.IsDefault 
+					? ViewStates.Gone 
+					: ViewStates.Visible;
```
mattleibow: Just checking @PureWeen, is this something we want? If I am understanding correctly, the track will remain visible if a color is set, but not running. I think this is the correct action, but I believe it differs from the existing implementations. We are doing new controls, so we can do what we want here.

Xamarin.Forms.Platform.Android/Material/MaterialActivityIndicatorRenderer.cs
```diff
@@ -179,7 +178,7 @@ SizeRequest IVisualElementRenderer.GetDesiredSize(int widthConstraint, int heigh
 		}
 
 		void IVisualElementRenderer.SetElement(VisualElement element) =>
-			Element = (element as ActivityIndicator) ?? throw new ArgumentException("Element must be of type ActivityIndicator.");
+			Element = (element as ActivityIndicator) ?? throw new ArgumentException($"{nameof(element)} must be of type {typeof(ActivityIndicator).Name}.");
```
mattleibow: Why is this `typeof().Name` instead of just `nameof()`?

Xamarin.Forms.Platform.Android/Resources/values/styles.xml
```diff
@@ -5,7 +5,8 @@
   <style name="XamarinFormsMaterialProgressBarHorizontal" parent="Widget.AppCompat.ProgressBar.Horizontal">
   </style>
   <style name="XamarinFormsMaterialProgressBarCircular" parent="Widget.AppCompat.ProgressBar">
+    <item name="android:background">@drawable/MaterialActivityIndicatorBackgroud</item>
```
mattleibow: There is a typo in the name "backgroud"

Xamarin.Forms.Platform.Android/Material/MaterialActivityIndicatorRenderer.cs
kingces95: Hm. I think the message should include the name of the type that element is because it's that type the user assigned which needs to be an ActivityIndicator.

Xamarin.Forms.Platform.Android/Material/MaterialActivityIndicatorRenderer.cs
paymicro: This need for show background circle if backgroud color is set.

Xamarin.Forms.Material.iOS/MaterialActivityIndicatorRenderer.cs
paymicro: The code is based on the material ActivityIndicator component
https://github.com/material-components/material-components-ios/blob/stable/components/ActivityIndicator/src/MDCActivityIndicator.m#L570
in `MDCActivityIndicator` this path recreates every time whenever you change any properties. Therefore, when drawing layers, we also recalculate this path. This allows to correctly display it when you change sizes or colors.

PureWeen:
build --uitests

Issue #4826, opened 2018-12-20
[Enhancement] Make CollectionView ItemTemplate work with DataTemplateSelector


adrianknight89:
Currently, setting the `ItemTemplate` property to a `DataTemplateSelector` throws an exception because `LoadTemplate` in `ElementTemplate` is null. We should be able to provide different data templates based on item state when using `CollectionView`.

matteopiccioni:
I think the following exception is related DataTemplateSelector

XF version: 4.0.0.304370-pre8
With this version of XF DataTemplateSelector are working  great on Android

Using iOS I have the following exception:
`Objective-C exception thrown.  Name: NSInternalInconsistencyException Reason: Invalid update: invalid number of items in section 0.  The number of items contained in an existing section after the update (1) must be equal to the number of items contained in that section before the update (1), plus or minus the number of items inserted or deleted from that section (1 inserted, 0 deleted) and plus or minus the number of items moved into or out of that section (0 moved in, 0 moved out).
Native stack trace:
	0   CoreFoundation                      0x00000001ba281ebc <redacted> + 252
	1   libobjc.A.dylib                     0x00000001b9451a50 objc_exception_throw + 56
	2   CoreFoundation                      0x00000001ba197a2c <redacted> + 0
	3   Foundation                          0x00000001bac861d0 <redacted> + 112
	4   UIKitCore                           0x00000001e6e50d04 <redacted> + 12840
	5   UIKitCore                           0x00000001e6e4d760 <redacted> + 548
	6   UIKitCore                           0x00000001e6e4d804 <redacted> + 68
	7   UserApp.iOS                         0x0000000105c6edb0 _ZN7plcrash2MS5async24dwarf_cfa_state_iteratorIyxE4nextEPjPNS1_28plcrash_dwarf_cfa_reg_rule_tEPy + 50642384`

Henri2h:
I don't know if it supposed to be fixed, but I still experience this issue using XF 4.0.0.394984-pre10.

TheBaileyBrew:
@matteopiccioni -- I had a similar issue, and it was related to how I was refreshing my CollectionView. 

Are you by chance, trying to empty the ObservableCollection with .Clear() or by a different method?

I was trying to change the Collection manually (by clearing out data) rather than just reinstantiating the OC with the new data that was being returned from my API call. by doing this instead, I stopped getting that error.

[...]

PR #5115, opened 2019-02-02
Some improvements to DataTemplateSelector and DataTemplate

### LINKED ISSUE #3407: ListView's GroupHeaderTemplate do not work with DataTemplateSelector on Android
Xamarin.Forms 3.1.0.583944
I want to use the DataTemplateSelector for the GroupHeaderTemplate.

``` xml
<ListView Grid.Row="1" ItemsSource="{Binding PickerGroups}" IsGroupingEnabled="True" ItemTemplate="{StaticResource PickerDataTemplateSelector}"
                  GroupHeaderTemplate="{StaticResource PickerDataTemplateSelector}"  CachingStrategy="RecycleElementAndDataTemplate">
</ListView>
```
I get here such a crash:
```
ListProxy.get_Item (System.Int32 index)
D:\a\1\s\Xamarin.Forms.Core\ListProxy.cs - line 127
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. Parameter name: index

1 ListProxy.get_Item (System.Int32 index) D:\a\1\s\Xamarin.Forms.Core\ListProxy.cs:127
2 IList.get_Item (System.Int32 index) D:\a\1\s\Xamarin.Forms.Core\ListProxy.cs:444
3 ListViewAdapter.GetItemViewType (System.Int32 position) D:\a\1\s\Xamarin.Forms.Platform.Android\Renderers\ListViewAdapter.cs:172
4 BaseAdapter.n_GetItemViewType_I (System.IntPtr jnienv, System.IntPtr native__this, System.Int32 position)
5 (wrapper dynamic-method) System.Object.cbacde75-f4e8-4fe2-bec7-81edead3c9fe(intptr,intptr,int)
```

### LINKED ISSUE #3544: [Core] Fix CreateContent usage on DataTemplatesSelector

mattleibow:
### Description of Change ###

There are a few issues with the DataTemplateSelector, nameley that it is weird. It is entirely separate from DataTemplate, yet derived directly from it. As a result, none of the base methods are relevant or work. 

I understand _why_ it is like it it, but I hope tis PR makes the usage of the selector much easier via the base members, and now it can be used directly instead of having to first check the type.

The hope is that moving forward, we only call the new `CreateContent` method instead of the parameterless one. This will allow the user to use either type and we don't have to write any checking code.

### Issues Resolved ### 

- prepares for #4994
- fixes #3407
- fixes #3544
- sets up for #4826
- addresses #3749

### API Changes ###

Added:

 - `public object ElementTemplate.CreateContent(object item, BindableObject container);`

### Platforms Affected ### 

- Core/XAML (all platforms)

Using a DataTemplateSelector with a ListView Header/Footer now no longer throws, but works as expected.

Not applicable

### Testing Procedure ###

Added some unit tests to check for changes.

### PR Checklist ###

- [x] Has automated tests


Xamarin.Forms.Core/DataTemplateExtensions.cs
```diff
@@ -7,16 +7,10 @@ public static class DataTemplateExtensions
...
-			var selector = self as DataTemplateSelector;
-			if (selector == null)
-				return self;
+			if (self is DataTemplateSelector selector)
+				return selector.SelectTemplate(item, container);
 
-			return selector.SelectTemplate(item, container);
-		}
-
-		public static object CreateContent(this DataTemplate self, object item, BindableObject container)
```
mattleibow: Besides the slight refactor to use new code above, this member is now merged with `ElementTemplate` to provide an overload that can be used everywhere, regardless of the `DataTemplate` type.

Xamarin.Forms.Core/BindableLayout.cs
```diff
@@ -181,14 +181,14 @@ void CreateChildren()
...
-			return CreateItemView(item, index, _itemTemplate ?? _itemTemplateSelector?.SelectTemplate(item, null));
+			return CreateItemView(item, index, _itemTemplate ?? _itemTemplateSelector);
 		}
 
 		View CreateItemView(object item, int index, DataTemplate dataTemplate)
 		{
 			if (dataTemplate != null)
 			{
-				var view = (View)dataTemplate.CreateContent();
+				var view = (View)dataTemplate.CreateContent(item, null);
```
mattleibow: This is just one instance, but we can now call this method - regardless if it is a selector or a template.

Xamarin.Forms.Core/DataTemplateSelector.cs
```diff
@@ -5,18 +5,20 @@ namespace Xamarin.Forms
 {
 	public abstract class DataTemplateSelector : DataTemplate
 	{
-		Dictionary<Type, DataTemplate> _dataTemplates = new Dictionary<Type, DataTemplate>();
+		readonly Dictionary<Type, DataTemplate> _dataTemplates = new Dictionary<Type, DataTemplate>();
+
+		// NOTE: we don't call base as this is a proxy object that does not need to be bound
+		//       and we do not actually make use of any of the base functionality
+		internal override object OnCreateContent(object item, BindableObject container) =>
+			SelectTemplate(item, container).CreateContent();
```
mattleibow: This method overrides the base behavior to make the selector work as expected.

Xamarin.Forms.Core/DataTemplate.cs
```diff
@@ -49,10 +49,14 @@ public void SetValue(BindableProperty property, object value)
...
 
-		internal override void SetupContent(object item)
+		internal override object OnCreateContent(object item, BindableObject container)
 		{
-			ApplyBindings(item);
-			ApplyValues(item);
+			var content = base.OnCreateContent(item, container);
+
+			ApplyBindings(content);
+			ApplyValues(content);
```
mattleibow: I removed the internal `SetupContent` from the base and added it here. No need to have that method there just for this type.

Xamarin.Forms.Core/ElementTemplate.cs
```diff
@@ -72,21 +70,18 @@ void IElement.RemoveResourcesChangedListener(Action<object, ResourcesChangedEven
...
+		public object CreateContent(object item, BindableObject container) =>
+			OnCreateContent(item, container);
 
-		internal virtual void SetupContent(object item)
+		internal virtual object OnCreateContent(object item, BindableObject container)
 		{
+			if (_loadTemplate == null)
+				throw new InvalidOperationException("LoadTemplate should not be null");
+
+			return _loadTemplate();
```
mattleibow: This is the default implementation - just instantiate.

Xamarin.Forms.Platform.UAP/CellControl.cs
```diff
@@ -314,14 +314,9 @@ void SetCell(object newContext)
...
 
-				if (template is DataTemplateSelector)
-				{
-					template = ((DataTemplateSelector)template).SelectTemplate(bindingContext, lv);
-				}
-
 				if (template != null)
 				{
-					cell = template.CreateContent() as Cell;
+					cell = template.CreateContent(bindingContext, lv) as Cell;
```
mattleibow: Another example of usage, no longer the check before using the method.

andreinitescu:
This can work for sure. 

However my hope was to try come with a better solution, since I think now with CollectionView coming it's a good opportunity. DataTemplateSelector is only currently being used by ListView which is being replaced by CollectionView. So if handling item selector is fixed properly, all new controls after the fix, CollectionView (but also bindable layouts, Map with Pin item template and selector, etc.) will handle item selector better.

What is the better solution? 
I don't think the "item selector" should derive from DataTemplate, it's not using any of the logic and data from it. 
Second, it makes much every DataTemplateSelector instance much heavier than necessary, every DataTemplate uses `Interlock.Increment` on c-tor:

		public DataTemplate()
		{
			_id = Interlocked.Increment(ref idCounter);
			_idString = GetType().FullName + _id;
		}

		public DataTemplate(Type type) : base(type)
		{
			_id = Interlocked.Increment(ref idCounter);
			_idString = type.FullName;
		}

Also, every DataTemplate instance allocates two dictionaries (Bindings, Values):

		public IDictionary<BindableProperty, BindingBase> Bindings { get; } = new Dictionary<BindableProperty, BindingBase>();

		public IDictionary<BindableProperty, object> Values { get; } = new Dictionary<BindableProperty, object>();

DataTemplateSelector is a simple strategy to return a DataTemplate, but I can't see why it has to be a DataTemplate itself.

I think I can understand why inheritance was chosen, I guess try make implementation within the XF easier by having selector as a template, but I don't think it works well.

mattleibow:
I agree with @andreinitescu quite a bit, and I think he has a better solution than this. This PR can be merged regardless as it helps with current implementations.

With regards to the inheritance, we can't change that, but maybe we should start looking for something because the base does do thinks that are entirely unnecessary. We really need to look at that.

I'll be honest, I am not sure I like the idea of an interface for the selector as that prevents us from adding methods or functionality. We can make use of extension methods, but we have to make sure we get the API right. I don't think we can change it again. 

It would make me happy to have an interface for both, and we use that going forward, but that might not be practical.

Andrei is right, now is the time to decide and we should look at splitting this. 

A am new to the coadebase, so I think we need to wait for the old hands to weigh is with actual facts.

mattleibow:
Just had a brilliant (relative I know) idea... Scrap this conversation and fix a bigger issue.

We have this concept of DataTemplate and ControlTemplate.  Then we have a DataTemplateSelector. What about the control?

Is it too crazy to have a base TemplateSelector that returns the base ElementSelector so we can make use of a selector for controls. 

With shell, we could totally make use of this ability to pick content based on the data item. 

Is this maybe a better option?

andreinitescu:
> I am not sure I like the idea of an interface for the selector as that prevents us from adding methods or functionality

I don't disagree, but I just wonder what kind of functionality would be needed? The only concern of selector is to return a DataTemplate instance. If more functionality is needed that should be something pretty specific, for specific purpose, and for that we can always have a base class, or, just use the default interface methods in C# 8 if possible, but fundamentally I think a selector should just return a DataTemplate instance, nothing more. Which sounds like a contract (interface).

> We have this concept of DataTemplate and ControlTemplate. Then we have a DataTemplateSelector. What about the control?
> Is it too crazy to have a base TemplateSelector that returns the base ElementSelector so we can make use of a selector for controls.

I'm not sure I follow you, are you wondering why there isn't any ControlTemplateSelector?
I think the idea is DataTemplate are meant to render data items, ControlTemplate is for another purpose, it's for templatable views, it represents their structure. Sounds similar but they are for different purposes.
I'm not sure I understand what would a selector for ControlTemplate do.

> Again I just thought, why is the list view header and footer a data template? It should be a control template! It has no data and effectively binds to null or the same data source as the list view.

I think HeaderTemplate and FooterTemplate as DataTemplate makes sense, it works in conjunction with Header and Footer properties which are of object type. 

----------------------

In order to have a more concrete discussion, I did some very quick work (it's incomplete but very close to completion) for the idea of adding a new IDataTemplateSelector and having separate property for selector, you can check it out here https://github.com/andreinitescu/Xamarin.Forms/tree/IDataTemplateSelector

If you think it's something it's worth pursuing I could finish it.

Xamarin.Forms.Core/DataTemplateSelector.cs
```diff
@@ -5,18 +5,20 @@ namespace Xamarin.Forms
...
-			var recycle = listView == null ? false :
-				(listView.CachingStrategy & ListViewCachingStrategy.RecycleElementAndDataTemplate) ==
-					ListViewCachingStrategy.RecycleElementAndDataTemplate;
+			var recycle = false;
+			if (container is ListView listView)
+				recycle = listView.CachingStrategy.HasFlag(ListViewCachingStrategy.RecycleElementAndDataTemplate);
 
-			DataTemplate dataTemplate = null;
-			if (recycle && _dataTemplates.TryGetValue(item.GetType(), out dataTemplate))
+			if (recycle && _dataTemplates.TryGetValue(item?.GetType(), out var dataTemplate))
```
hartez: These null checks on `item` don't really help us, because the result of `item?.GetType()` if `item` is `null` will be `null`, and we'll just get an ArgumentNullException instead of a NullReferenceException. There should probably be a `null` check and exception on `item` at the beginning of the method so it's really obvious why things are failing.

mattleibow:
```suggestion
			if (recycle && _dataTemplates.TryGetValue(item.GetType(), out var dataTemplate))
```

Xamarin.Forms.Core/DataTemplateSelector.cs
```diff
@@ -30,7 +32,7 @@ public DataTemplate SelectTemplate(object item, BindableObject container)
 					throw new NotSupportedException(
 						"RecycleElementAndDataTemplate requires DataTemplate activated with ctor taking a type.");
 
-				_dataTemplates[item.GetType()] = dataTemplate;
+				_dataTemplates[item?.GetType()] = dataTemplate;
```
mattleibow:
```suggestion
				_dataTemplates[item.GetType()] = dataTemplate;
```

samhouts:
@mattleibow Please rebase to resolve conflicts. Thanks!

mattleibow:
On it!

Xamarin.Forms.Core/DataTemplateSelector.cs
```diff
@@ -5,18 +5,20 @@ namespace Xamarin.Forms
 {
 	public abstract class DataTemplateSelector : DataTemplate
 	{
-		Dictionary<Type, DataTemplate> _dataTemplates = new Dictionary<Type, DataTemplate>();
+		readonly Dictionary<Type, DataTemplate> _dataTemplates = new Dictionary<Type, DataTemplate>();
+
+		// NOTE: we don't call base as this is a proxy object that does not need to be bound
+		//       and we do not actually make use of any of the base functionality
+		internal override object OnCreateContent(object item, BindableObject container) =>
+			SelectTemplate(item, container)?.CreateContent(item, container);
```
mattleibow: Pass the item and container to both the template selector and the template. If different data is needed for some reason, the caller can still use the `SelectTemplate` method as always.

KSemenenko:
@mattleibow are there any updates here?

rmarinho:
Unit tests are failing @mattleibow  , can you take a look ?

KSemenenko:
any updates?

samhouts:
I'm going to close this one for now until we're ready to return to it. Thanks, @mattleibow !

PR #5123, opened 2019-02-04
Fluent layout support

AndreiMisiukevich:
### Description of Change ###

I prefer code over xaml for creating UI. I presume, for those guys, who like code UI more than XAML, it would be convenient/comfortable to use fluent layout extensions methods.

Let's check on simple sample (we want to create button and label and bind their properties to any view model (without details of VM.. it is abstraction)

NOW:
```csharp
public class TestPage : ContentPage
{
	public TestPage()
	{
		var button = new Button
		{
			Text = "Button"
		};
		button.SetBinding(Button.CommandProperty, "TapCommand");

		var label = new Label();
		label.SetBinding(Label.TextProperty, "LabelText");
		label.SetBinding(Label.TextColorProperty, "LabelColor");

		Content = new StackLayout
		{
			Children =
			{
				button,
				label
			}
		};
	}
}
```

FLUENT:
```csharp
public class TestPage : ContentPage
{
	public TestPage()
 	=> Content = new StackLayout().Children(
		new Button
	 	{
		 	Text = "Button"
	 	}.Command("TapCommand"),
		new Label()
			.Bind(Label.TextProperty, "LabelText")
			.Bind(Label.TextColorProperty, "LabelColor"));
}	
```

You needn't have to create local temp variables for bindings etc.
There are base methods only, but we can improve and extend them easily.

Any ideas?

### API Changes ###
Added new extensions methods

### Platforms Affected ### 
- Core

None

Not applicable

### Testing Procedure ###
Xamarin.Forms.Controls project, App.cs file:
uncomment ```//MainPage = new RootPages.RootFluentPage(); // fluent test ``` and run.
See RootFluentPage class. Layout is fluent


Xamarin.Forms.Controls/RootPages/RootFluentPage.cs
```diff
@@ -0,0 +1,227 @@
...
+						{
+							BackgroundColor = Color.Gold,
+							Content = new Label
+							{
+								FontSize = 20,
+								Text = "Click Me (3)"
+							}
+						}.Tap(nameof(RootFluentViewModel.ClickCommand), "(3) Clicked")
+					}
+			}.Fluent(p => BindingContext = new RootFluentViewModel());
```
StephaneDelcroix: why don't you set the context like any other property ?

Xamarin.Forms.Core/FluentExtensions.cs
```diff
@@ -0,0 +1,128 @@
+﻿using System;
+
+namespace Xamarin.Forms.Fluent
+{
+	public static class FluentExtensions
+	{
+		#region BindableObject section
+
+		public static TView Fluent<TView>(this TView self,
```
StephaneDelcroix: there is very little gain to be had over here, I'd remove this call

Xamarin.Forms.Core/FluentExtensions.cs
```diff
@@ -0,0 +1,128 @@
...
+				throw new ArgumentNullException(nameof(self));
+			if (targetProperty == null)
+				throw new ArgumentNullException(nameof(targetProperty));
+
+			self.SetBinding(targetProperty, bindingBase);
+
+			return self;
+		}
+
+		public static TView Bind<TView>(this TView self,
```
StephaneDelcroix: this will get out of sync in no time. either provide the same subset of params as BindableObjectExtension.SetBinding does, or drop this.

Xamarin.Forms.Controls/RootPages/RootFluentPage.cs
AndreiMisiukevich: @StephaneDelcroix i did it for demonstrating all current possibilities of fluent layout.

it can be useful for subscribing on events etc.

``` csharp
...
.Fluent(view => view.PropertyChanged += (s, e) => Debug.WriteLine(e.PropertyName)); ```

File: Xamarin.Forms.Core/FluentExtensions.cs (hunk 2 above)
Comment (AndreiMisiukevich) [reply Xamarin.Forms.Core/FluentExtensions.cs]: @StephaneDelcroix as i mentioned above. It can be useful for event subscriptions..

File: Xamarin.Forms.Core/FluentExtensions.cs (hunk 3 above)
Comment (AndreiMisiukevich) [reply Xamarin.Forms.Core/FluentExtensions.cs]: @StephaneDelcroix did you mean method below?

```csharp
public static TView Bind<TView>(this TView self,
			BindableProperty targetProperty,
			string path,
			BindingMode mode = BindingMode.Default,
			IValueConverter converter = null,
			object converterParameter = null,
			string stringFormat = null,
			object source = null) where TView : BindableObject
			=> Bind(self, targetProperty, new Binding
			{
				Path = path,
				Mode = mode,
				Converter = converter,
				ConverterParameter = converterParameter,
				StringFormat = stringFormat,
				Source = source
			});
```

Seems, it provides more options than BindableObjectExtension.SetBinding (addition **source** parameter)

AndreiMisiukevich:
> Thanks for doing this. I understand it might be helpful when you create your UI from code. But, as
> 
> * it covers only a subset of the API,
> * it will be a pain to keep in sync with API changes,
> 
> I would recommend not to merge this in the Core code, but you should definitely publish that as a separate nuget.

Yeah, it makes sense.

But maybe it will be useful to have little possibility of this feature from box?

For example bindings only.

instead of returning **void**, we can return current element in **SetBinding** method.

```csharp
new Label().SetBinding(Label.TextProperty, "Text").SetBinding(VisualElement.BackgroundColor, "Color");
```

Ideas?

Thx for review 👍

AndreiMisiukevich:
@StephaneDelcroix @jamesmontemagno 

Probably, it would be useful for code-lovers to have chance for setting bindings fluently like i did it here (as example): https://github.com/xamarin/Xamarin.Forms/blob/40f38196ec273daa6348d1fe281e52d6048ab20b/Xamarin.Forms.Controls/GalleryPages/BoundContentPage.cs#L73-L75

@StephaneDelcroix RE challenging to cover/support all api. I agree with that point. So, i suggest to update SetBinding methods for now only. After i am going to create separate nuget for achieving maximum fluency goal (as you suggested).

Could you please re-review the PR?
Thx

AndreiMisiukevich:
Also i think, it will be super convenient to have **Children** method

```csharp

var st = new StackLayout
{
    Children =
    {
        new FirstView(),
        new SecondView()
    }
};

if(Device.RuntimePlatform == Device.Android)
{
    st.Children.Add(new AndroidSpecificView());
}
```

VS

```csharp
new StackLayout().Children(
        new FirstView(),
        new SecondView(),
        Device.RuntimePlatform == Device.Android ? new AndroidSpecificView() : null
```


Extension method for all **Layout<View>** subclasses. Maybe we can find generic scope of methods and include in in box?

AndreiMisiukevich:
Also setting some of attached properties like Absolute Layout flags and bounds.

mattleibow:
I also think that even if we _just_ do the binding, it will be a big help.

For the full fluent, handling of events is also a nice thing.

AndreiMisiukevich:
> I also think that even if we _just_ do the binding, it will be a big help.
> 
> For the full fluent, handling of events is also a nice thing.
@mattleibow 
1) SetBinding
2) SetChildren (for all Layout<View>) with possibility to avoid null elements 
3) SetEventHandlers (for those who prefer event handlers over commands)

```csharp
new Button()
    SetEventHandlers(b => b.Clicked += OnButtonClicked);
```

mattleibow:
I was just thinking some more and it might be good to actually create this entire fluent api as a set of extension methods:

element.AddBinding().AddBinding()
layout.AddChild().AddChild() - this is a case where the children require different args per layout type (Grid, Stack, etc)
view.AddScrolledEvent().AddClickedEvent() - this again is enhanced per view
view.SetGridColumn().SetAbsoluteBounds()

For the event, we _could_ have the one that takes the delegate as a fallback.

The advantage of a proper extension-based fluent API is that we can keep the core "pure" and then have all the cool overloads and other specific features separate.

AndreiMisiukevich:
> I was just thinking some more and it might be good to actually create this entire fluent api as a set of extension methods:
> 
> element.AddBinding().AddBinding()
> layout.AddChild().AddChild() - this is a case where the children require different args per layout type (Grid, Stack, etc)
> view.AddScrolledEvent().AddClickedEvent() - this again is enhanced per view
> view.SetGridColumn().SetAbsoluteBounds()
> 
> For the event, we _could_ have the one that takes the delegate as a fallback.
> 
> The advantage of a proper extension-based fluent API is that we can keep the core "pure" and then have all the cool overloads and other specific features separate.

Do you think, it would be better to create separate nuget or put something in XF box?

StephaneDelcroix:
@AndreiMisiukevichI would be ok with the latest version of the PR if this wasn't a breaking change. I'm gonna close this for now, as nothing prevents you to have this as extension in your own project or in a separate nuget

PR #5136, opened 2019-02-05
Span.LineBreak

StephaneDelcroix:
### Description of Change ###

```xaml
<LineBreak/>
```

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Span.cs
+++ b/Xamarin.Forms.Core/Span.cs
@@ -163,6 +163,6 @@ namespace Xamarin.Forms
 
 	public class NewLine : Span
 	{
-		public NewLine() => Text = Environment.NewLine;
+		public LineBreak() => Text = Environment.NewLine;
 	}
 }
--- a/Xamarin.Forms.Core/Span.cs
+++ b/Xamarin.Forms.Core/Span.cs
@@ -161,7 +161,7 @@ namespace Xamarin.Forms
 		}
 	}
 
-	public class NewLine : Span
+	public class LineBreak : Span
 	{
 		public LineBreak() => Text = Environment.NewLine;
 	}
--- a/Xamarin.Forms.Core/Span.cs
+++ b/Xamarin.Forms.Core/Span.cs
@@ -55,7 +55,8 @@ namespace Xamarin.Forms
 #pragma warning restore 618
 
 		public static readonly BindableProperty TextProperty
-			= BindableProperty.Create(nameof(Text), typeof(string), typeof(Span), "", defaultBindingMode: BindingMode.OneWay);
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Core/Span.cs
```diff
@@ -55,7 +55,8 @@ public Color ForegroundColor
 #pragma warning restore 618
 
 		public static readonly BindableProperty TextProperty
-			= BindableProperty.Create(nameof(Text), typeof(string), typeof(Span), "", defaultBindingMode: BindingMode.OneWay);
+			= BindableProperty.Create(nameof(Text), typeof(string), typeof(Span), "", defaultBindingMode: BindingMode.OneWay,
+				validateValue: (b, v) => !(b is NewLine));
```
StephaneDelcroix:
```suggestion
				validateValue: (b, v) => !(b is LineBreak));
```

Xamarin.Forms.Core/Span.cs
```diff
@@ -159,4 +160,9 @@ void ILineHeightElement.OnLineHeightChanged(double oldValue, double newValue)
 		{
 		}
 	}
+
+	public class NewLine : Span
```
StephaneDelcroix:
```suggestion
	public class LineBreak : Span
```

Xamarin.Forms.Core/Span.cs
```diff
@@ -159,4 +160,9 @@ void ILineHeightElement.OnLineHeightChanged(double oldValue, double newValue)
...
+	{
+		public NewLine() => Text = Environment.NewLine;
```
StephaneDelcroix:
```suggestion
		public LineBreak() => Text = Environment.NewLine;
```

dotMorten:
> This was a quick and dirty implementation for david, and it ended up being mostly dirty...

At least you're honest with yourself :-) Should Xamarin.Forms be adding dirty stuff though? I mean the above implementation anyone could do in their own library. IMHO the Xamarin.Forms implementation should be solid for the long run.

Btw I missed you making it not modifiable with that validate value callback. Very clever trick/hack! :-)

StephaneDelcroix:
@dotMorten I'm not going to merge this as is

KSemenenko:
Maybe add something like <Space /> which will insert " " character?

adrianknight89:
@StephaneDelcroix I don't see a problem with making this a special `Span` as long as the validation is in place. Creating two different types could be a breaking change. For example, `FormattedText.Spans` collection will need to be renamed and we can't add a `LineBreak` to it unless the underlying data type for the collection is changed. This is too much work for this tiny element.

My only issue is that line breaks are often not semantic. See [here](https://stackoverflow.com/questions/1726073/is-it-sometimes-bad-to-use-br).

PR #5137, opened 2019-02-05
[Mac] Added support AutomationProperty

paymicro:
### Description of Change ###

Added support `AutomationProperty` for _regular renderer_'s, _cells_ and _toolbar_ buttons.
This is part of PR #3955

### Issues Resolved ### 

none

### API Changes ###
 
 None

### Platforms Affected ### 

- Mac

None

Before
before

After
after

### Testing Procedure ###

- check AutomationPropertyGallery

### PR Checklist ###

- [ ] Has automated tests 


rmarinho:
I think there was a scenario we need the defaults? @samhouts you remember ?

samhouts:
We use the defaults to make sure that we don't overwrite or lose any values that the customer may have set via custom renderer or effects.

rmarinho:
Why you are removing the default @paymicro ? like @samhouts said users could have set that in there own renderers.

rmarinho:
@PureWeen after you activate voice over there's a special instructions for navigating a table, I think is command + option plus arrows , I got it to tell "option a"

PR #5145, opened 2019-02-06
[iOS] fix accessibility of Picker

### LINKED ISSUE #3454: Picker accessibility text is wrong

paymicro:
### Description of Change ###

Pickers reads like a Button. This is part of PR #3955

### Issues Resolved ### 

- partially fixes #3454

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS

None

Before
screenshot_20

After
screenshot_19

### Testing Procedure ###

- start `PickerGallery - Legacy` in ControlGallery
- run `Accessibility Inspector` or `VoiceOver`
- picker type should be like a button

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls/GalleryPages/PickerGallery.cs
+++ b/Xamarin.Forms.Controls/GalleryPages/PickerGallery.cs
@@ -22,6 +22,7 @@ namespace Xamarin.Forms.Controls
 			p1.SelectedIndexChanged += (sender, e) => {
 				testLabel.Text = "Selected Index Changed";
 			};
+			p1.SetAutomationPropertiesName("Title picker");
 
 			Content = new ScrollView { 
 				Content = new StackLayout {
--- a/Xamarin.Forms.Platform.iOS/Renderers/DatePickerRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/DatePickerRenderer.cs
@@ -81,6 +81,8 @@ namespace Xamarin.Forms.Platform.iOS
 
 				_useLegacyColorManagement = e.NewElement.UseLegacyColorManagement();
 
+				entry.AccessibilityTraits = UIAccessibilityTrait.Button;
+
 				SetNativeControl(entry);
 			}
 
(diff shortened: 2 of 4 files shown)


PureWeen:
@paymicro can we make it so it doesn't say textfield at all when you click on the picker?

Can you copy over the *run UItest 3454* from the android part you broke this out from so I can test those features as well?

samhouts:
@paymicro See @PureWeen's comment. Thanks!!

paymicro:
@PureWeen i rebased to 4.2.0. UItest 3454 already became available for inspection. 

> can we make it so it doesn't say textfield at all when you click on the picker?

The type of all pickers is changed to a `button`. Nowhere does textfield speak.

PR #5155, opened 2019-02-07
[previewer] more info to the callback

StephaneDelcroix:
### Description of Change ###

[previewer] more info to the callback

### Issues Resolved ### 

/

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Xaml/XamlLoader.cs
+++ b/Xamarin.Forms.Xaml/XamlLoader.cs
@@ -317,6 +317,6 @@ namespace Xamarin.Forms.Xaml
 
 		internal static Func<IList<FallbackTypeInfo>, Type, Type> FallbackTypeResolver { get; set; }
 		internal static Action<CallbackTypeInfo, object> ValueCreatedCallback  { get; set; }
-		internal static Func<CallbackTypeInfo, object> InstantiationFailedCallback { get; set; }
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Xaml/XamlLoader.cs
```diff
@@ -317,6 +317,6 @@ public struct CallbackTypeInfo
 
 		internal static Func<IList<FallbackTypeInfo>, Type, Type> FallbackTypeResolver { get; set; }
 		internal static Action<CallbackTypeInfo, object> ValueCreatedCallback  { get; set; }
-		internal static Func<CallbackTypeInfo, object> InstantiationFailedCallback { get; set; }
+		internal static Func<CallbackTypeInfo, Type, Exception, object> InstantiationFailedCallback { get; set; }
```
TimBarham: Instead of adding parameters, can we instead define a new type for our parameter (say `InstantiationFailedInfo`) that has the same properties as `CallbackTypeInfo`, plus `Type` and `Exception`?

That way, we can add more info later if we find we need it without breaking our existing reflection handling.

StephaneDelcroix: I think it's fine

alanmcgovern: 👍

PR #5171, opened 2019-02-08
Add a test for when we want to silently ignore

alanmcgovern:
Sometimes we cannot replace a type with a placeholder, and in this
scenario we wish to silently ignore the xaml we can't handle.

### Issues Resolved ### 

- fixes #

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

> VS bug [#790713](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/790713)

### DIFF
--- a/Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
+++ b/Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
@@ -51,7 +51,7 @@ namespace Xamarin.Forms.Xaml.UnitTests
 		}
 
 		[Test]
-		public void ContentPageWithMissingType()
+		public void ContentPageWithMissingType_MockviewReplacement()
 		{
 			XamlLoader.FallbackTypeResolver = (p, type) => type ?? typeof(MockView);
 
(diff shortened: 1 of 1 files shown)


alanmcgovern:
/cc @StephaneDelcroix @TimBarham 

This test is reproducing the same issue we had in our builds. I reverted the change on our side which made us return `null` instead of a `placeholder`. We should review this a bit further before making a decision. I appreciate it's already Tim's weekend, so can we sync up on Monday?

samhouts:
We'll close this for now and revisit if it's still an issue. Thanks!

alanmcgovern:
it's still an issue :)

alanmcgovern:
The fundamental problem is that users can (and will) type invalid namespace declarations, which reference types which do not, and will never, exist. In this scenario we need Forms to silently ignore that so we can do a best effort render, and we'll rely on intellisense to point out the error.

The simplest repro is take any xaml file and put a deliberate typo in the type name.

    `xmlns:vm="clr-Mine.ViewModels"` -> `xmlns:vm="clr-Mine.ViewModels.WILLNEVEREXIST"`

In this scenario we should not completely fail due to types prefixed with `vm:` not being resolvable.

TimBarham:
So my question is: _why_ should we not "completely fail" in this scenario?

**Current behavior:**

If the file has already been rendered at least once and the user introduces such an error (for example, they're in the middle of editing a tag and it is currently invalid), Forms will throw, and we won't update the current render. As soon as they finish typing the tag, Forms will render successfully and we will update the Previewer.

If the user opens a XAML file that already contains such an error, we won't render, but we will display the exception so there is clear guidance what the problem is.

This current behavior matches the behavior of the Windows XAML designer.

**If Forms silently ignores this issue:**

If the file has already been rendered, and the user introduces such an error like in the above example, the control will disappear from the previewer until the user finishes typing, at which point it will re-appear. 

If the file is opened with the error already in place, it will render, but without the control and with no guidance as to why that control is not rendering (apart from, hopefully, intellisense).

Personally I favor the current behavior because it seems most useful to the user in most circumstances, and also because it matches the behavior of the Windows XAML designer.

alanmcgovern:
We had a discussion about this and the current thinking is that these kind of 'the user has just done it wrong' scenario already has a pretty clearly defined UX in the WPF/UWP designer and we should aim to match that. They do error out when users write invalid xaml like this, so the plan is that we will continue to error out, and (hopefully) the user will understand and address the issue in their xaml.

PR #5182, opened 2019-02-09
DetectPlatform Fixed for UWP

### LINKED ISSUE #4852: [Enhancement] TargetIdiom as Xbox or Console
## Summary
Currently, if i run a UWP, on xbox or desktop, TargetIdiom is always Desktop. This can be extended as we know from the Device.Family as below

AnalyticsInfo.VersionInfo.DeviceFamily == "Windows.Desktop"
AnalyticsInfo.VersionInfo.DeviceFamily == "Windows.Xbox"

## API Changes
 
TargetIdiom enum has new value as "Xbox" or "Console"
 
## Intended Use Case
It helps to adjust functionality and UI on Xbox to separate from Desktop

EmilAlipiev:
…onInfo.DeviceFamily

UWP-Xbox sets TV...Surface tablets sets Desktop intead of Phone

### Description of Change ###

fixes #4852

### Issues Resolved ### 

Using previous code, it was missing platforms like Xbox which should be set as TV.
Beside that Surface tablet PCs were getting Idiom as phone. Because ApiInformation.IsTypePresent("Windows.Phone.UI.Input.HardwareButtons")  returns false for Surface. It supposed to be Desktop not Phone as confirmed in the msdn question below. Using DeviceFamily returns Desktop. 

https://social.msdn.microsoft.com/Forums/windowsapps/en-US/0b75ea16-1d2a-4997-bfa2-8cb6f73ff1e8/uwp-what-device-family-is-a-surface-tablet?forum=wpdevelop#39c5d17a-4706-4c24-b3c0-9b17ca60dd4f

- fixes #

### API Changes ###
Running UWP app on Xbox returns Idiom as TV, Surface returns as Desktop
 
### Platforms Affected ### 

- UWP

### DIFF
--- a/Xamarin.Forms.Platform.UAP/Forms.cs
+++ b/Xamarin.Forms.Platform.UAP/Forms.cs
@@ -40,18 +40,22 @@ namespace Xamarin.Forms
 			Device.SetFlags(s_flags);
 			Device.Info = new WindowsDeviceInfo();
 
-			switch (DetectPlatform())
+			switch (Windows.System.Profile.AnalyticsInfo.VersionInfo.DeviceFamily)
 			{
-				case Windows.Foundation.Metadata.Platform.Windows:
+				case "Windows.Desktop":
 					Device.SetIdiom(TargetIdiom.Desktop);
 					break;
-				case Windows.Foundation.Metadata.Platform.WindowsPhone:
+				case "Windows.Mobile":
 					Device.SetIdiom(TargetIdiom.Phone);
+					break;
+				case "Windows.Xbox":
+					Device.SetIdiom(TargetIdiom.TV);
 					break;
 				default:
 					Device.SetIdiom(TargetIdiom.Tablet);
 					break;
 			}
+
 			ExpressionSearch.Default = new WindowsExpressionSearch();
 
 			Registrar.ExtraAssemblies = rendererAssemblies?.ToArray();
@@ -63,16 +67,7 @@ namespace Xamarin.Forms
 
 			Platform.UWP.Platform.SubscribeAlertsAndActionSheets();
 		}
-
-		static Windows.Foundation.Metadata.Platform DetectPlatform()
-		{
-			bool isHardwareButtonsAPIPresent = ApiInformation.IsTypePresent("Windows.Phone.UI.Input.HardwareButtons");
-
-			if (isHardwareButtonsAPIPresent)
-				return Windows.Foundation.Metadata.Platform.WindowsPhone;
-			return Windows.Foundation.Metadata.Platform.Windows;
-		}
-
+		 
 		static FlowDirection GetFlowDirection()
 		{
 			string resourceFlowDirection = ResourceContext.GetForCurrentView().QualifierValues["LayoutDirection"];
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.UAP/Forms.cs
```diff
@@ -40,18 +40,22 @@ public static void Init(IActivatedEventArgs launchActivatedEventArgs, IEnumerabl
 			Device.SetFlags(s_flags);
 			Device.Info = new WindowsDeviceInfo();
 
-			switch (DetectPlatform())
+			switch (Windows.System.Profile.AnalyticsInfo.VersionInfo.DeviceFamily)
 			{
-				case Windows.Foundation.Metadata.Platform.Windows:
+				case "Windows.Desktop":
 					Device.SetIdiom(TargetIdiom.Desktop);
```
paymicro: I think we need to separate the `Surface` tablet from the desktop
```suggestion
					if (Windows.UI.ViewManagement.UIViewSettings.GetForCurrentView().UserInteractionMode ==
						Windows.UI.ViewManagement.UserInteractionMode.Touch)
						Device.SetIdiom(TargetIdiom.Tablet);
					else
						Device.SetIdiom(TargetIdiom.Desktop);
```

Xamarin.Forms.Platform.UAP/Forms.cs
```diff
@@ -40,18 +40,22 @@ public static void Init(IActivatedEventArgs launchActivatedEventArgs, IEnumerabl
...
 					break;
-				case Windows.Foundation.Metadata.Platform.WindowsPhone:
+				case "Windows.Mobile":
 					Device.SetIdiom(TargetIdiom.Phone);
 					break;
+				case "Windows.Xbox":
+					Device.SetIdiom(TargetIdiom.TV);
+					break;
 				default:
 					Device.SetIdiom(TargetIdiom.Tablet);
```
paymicro: I'm not sure that the tablet is the correct device by default. The number of device types (`Windows.Holographic`, `Windows.Team`, `Windows.IoT`) may increase and not all of them will be compatible with the tablet.

What you think @samhouts ? Perhaps set these types of devices as `TargetIdiom.Unsupported`?

samhouts: Hmm. Yeah, I think that sounds like a decent idea.

paymicro:
```suggestion
					Device.SetIdiom(TargetIdiom.Unsupported);
```

Xamarin.Forms.Platform.UAP/Forms.cs
EmilAlipiev: does that UserInteractionMode  work only when you Switch to tablet mode on touch capable devices? Because if not Surface but any other Notebook with touch Screen will fall into this category also.

paymicro: In fact, it will depend on the system settings. [Which can be changed at any time.](https://docs.microsoft.com/en-us/uwp/api/windows.ui.viewmanagement.userinteractionmode#remarks)

> Some devices (PC, laptop, tablet) support both a Desktop (mouse-optimized) and a Tablet (touch-optimized) mode.
Users can switch between running in Tablet mode and Desktop mode by going to Settings > System > Tablet mode and setting Make Windows more touch-friendly when using your device as a tablet.

For tablets the default will be `Touch` mode, for desktops `Mouse` mode. 
Touch mode indicates that the user has tablet mode enabled.
image

PR #5185, opened 2019-02-09
[Enhancement] Added NavigationPage.IconColorProperty

### LINKED ISSUE #5177: [Enhancement] Change NavigationPage back button color on Android
## Summary
I cannot change NavigationPage back button color on Android. On iOS color of back button is the same as color of text - so I can change it using BarTextColor property. This is not working on Android and I believe that this should behave the same. For now the only solution is to use custom renderer.

I think that it should be changed somewhere here https://github.com/xamarin/Xamarin.Forms/blob/7cc9a282bdeb76405c793574ebe0d096072f4228/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs#L935 but I'm unable do it myself.

AndreiMisiukevich:
### Description of Change ###

It makes sense to apply BarTextColor to title AND navigation icon (hamburger, back etc.)

### Issues Resolved ### 

- fixes #5177

### API Changes ###

Added NavigationPage.IconColorProperty

### Platforms Affected ### 

- Android
- iOS

BarTextColor will manage colors for Title text and Icon

Try to set BarTextColor to Red

BEFORE:

AFTER:

### Testing Procedure ###

Set BarTextColor to Red to any navigation page and run issues gallery.
Icon should be Red as well as Title


rmarinho:
This is a breaking change, we might need to add as a PS

AndreiMisiukevich:
@rmarinho i can redesign it to be a platform specific setting :)

But don't you think, that it is a bug?

AndreiMisiukevich:
Updated

rmarinho:
Hey @AndreiMisiukevich I was talking with @PureWeen  and he suggested we might be better by adding a property to set the icon color and implement that in all 3 platforms. What do you think?

AndreiMisiukevich:
> Hey @AndreiMisiukevich I was talking with @PureWeen and he suggested we might be better by adding a property to set the icon color and implement that in all 3 platforms. What do you think?

Hm makes sense.

But we should keep in mind that currently iOS uses BarTextColor as color for icon.

"On iOS color of back button is the same as color of text - so I can change it using BarTextColor property." - from original ticket

will have to do smth like
**iOS pseudocode**
```
if not iconColor.IsDefault
    then [use iconColor]
else [use textColor]
```

for android leave as is (just will change to use instance's  property)

...

UWP ?

rmarinho:
@AndreiMisiukevich yap sounds like a good plan, we will look ump after.

AndreiMisiukevich:
@rmarinho updated

Xamarin.Forms.Core/NavigationPage.cs
```diff
@@ -26,6 +26,8 @@ public class NavigationPage : Page, IPageContainer<Page>, IBarElement, INavigati
 
 		public static readonly BindableProperty BarTextColorProperty = BarElement.BarTextColorProperty;
 
+		public static readonly BindableProperty IconColorProperty = BindableProperty.Create(nameof(IconColor), typeof(Color), typeof(NavigationPage), Color.Default);
```
StephaneDelcroix: why is TitleIcon attached (and applied to the Page inside of a NavPage) and IconColor a normal BP, applied to the NavPage ?

Xamarin.Forms.Core/NavigationPage.cs
```diff
@@ -58,6 +60,12 @@ public NavigationPage(Page root) : this()
 			set => SetValue(BarElement.BarTextColorProperty, value);
 		}
 
+		public Color IconColor
+		{
+			get => (Color)GetValue(IconColorProperty);
+			set => SetValue(IconColorProperty, value);
+		}
```
StephaneDelcroix: see ^^^

Xamarin.Forms.Core/NavigationPage.cs
AndreiMisiukevich: @StephaneDelcroix nice point
i will update this PR with attached property

AndreiMisiukevich:
updated

matteopiccioni:
I follow with high interest this PR
In my project I have some tab pages with different bar color
In iOS all is working great because if I set BarTextColor, the Icon color is changed in the same way
In Android unfortunately the icon color is not modified, so in Android project, I had to use a theme with a fixed icon color that is always visibile, but the final appears is worse than iOS

After the update, using dynamic resources as my following case, the icon in Android will be of the right color?

    `<Style ApplyToDerivedTypes="true" TargetType="NavigationPage">`  
        `<Setter Property="BarTextColor" Value="{DynamicResource Key=NavigationBarTextColorDynamic}"/>  <!--  colore che cambia in base al default o ai tab   -->`
        `<Setter Property="BarBackgroundColor" Value="{DynamicResource Key=NavigationBarBackgroundColorDynamic}"/>  <!--  colore che cambia in base al default o ai tab   -->`
    `</Style>`
and in my OnCurrentPageChanged of the TabPage:

```
        static IDictionary<Type, Color> colorMap = new Dictionary<Type, Color>
        {
            { typeof(TabSecurityPage), (Color)Xamarin.Forms.Application.Current.Resources["tabSecurityFontColor"]},
            { typeof(SmartHomePage), (Color)Xamarin.Forms.Application.Current.Resources["tabSmartHomeFontColor"] },
            { typeof(TabClimaPage), (Color)Xamarin.Forms.Application.Current.Resources["tabClimaFontColor"] },
            { typeof(TabSistemaPage), (Color)Xamarin.Forms.Application.Current.Resources["tabSistemaFontColor"] }
        };

        static IDictionary<Type, String> titleMap = new Dictionary<Type, String>
        {
            { typeof(TabSecurityPage), "Sicurezza" },
            { typeof(SmartHomePage), "Smart home" },
            { typeof(TabClimaPage), "Clima" },
            { typeof(TabSistemaPage), "Sistema" }
        };
        protected override void OnCurrentPageChanged()
        {
            base.OnCurrentPageChanged();

            if (CurrentPage != null)
            {
                SelectedTabColor = colorMap[CurrentPage.GetType()];
                Title = titleMap[CurrentPage.GetType()];

                Xamarin.Forms.Application.Current.Resources["NavigationBarTextColorDynamic"] = colorMap[CurrentPage.GetType()];
                Xamarin.Forms.Application.Current.Resources["NavigationBarBackgroundColorDynamic"] = Xamarin.Forms.Application.Current.Resources["NavigationBarWhiteColor"];
            }
        }

```
Thanks


Hope this new property will be released soon
Thanks

AndreiMisiukevich:
rebased on the top of master branch

sharpwood:
mark

Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
```diff
@@ -1003,6 +1005,10 @@ void UpdateToolbar()
 			if (!textColor.IsDefault)
 				bar.SetTitleTextColor(textColor.ToAndroid().ToArgb());
 
+			Color navIconColor = NavigationPage.GetIconColor(Current);
+			if(!navIconColor.IsDefault)
```
samhouts: Looks like you won't be able to set the color back to default after setting it to an explicit color.

AndreiMisiukevich: but it's a generic issue of all platform-specific stuff.

CursorColorProperty from iOS-specific Entry has exactly the same issue. (and a lof of other)

samhouts: We miss it a lot, sure. We have an opportunity to not miss it here! The pattern we try to use is to cache the first value that was already set in a field (like `_defaultNavIconColor`); if the navIconColor is then set to `Default`, we set it to the field's value.

AndreiMisiukevich: @samhouts makes sense) but how would we handle cases, when a user changes something via custom renderer (properties in controls directly). What is "real" default value?

samhouts: By the time this code is run, we hope that the custom renderer has already set the value of the color filter, so whatever we get from the current color filter is what they've set....theoretically! That's why I try to always test new changes against a custom renderer that's already set the value I'm affecting.

AndreiMisiukevich: @samhouts great. now is clear) thx

samhouts:
@AndreiMisiukevich Would you mind rebasing on master to fix the build? Thanks!

PR #5186, opened 2019-02-09
[Xaml[C]] Accept prefixed property names of markup extensions

akihikodaki:
### Description of Change ###

[MS-XAML-2009] (page 79) says a markup extension accepts prefixed member names. This change supports the syntax. It is essential to use a generic markup extension, and this includes a test case for such a situation.

### Issues Resolved ### 

None

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [X] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Build.Tasks/ExpandMarkupsVisitor.cs
+++ b/Xamarin.Forms.Build.Tasks/ExpandMarkupsVisitor.cs
@@ -127,27 +127,68 @@ namespace Xamarin.Forms.Build.Tasks
 				if (split.Length > 2)
 					throw new ArgumentException();
 
-				string prefix, name;
-				if (split.Length == 2) {
-					prefix = split[0];
-					name = split[1];
-				}
-				else {
-					prefix = "";
-					name = split[0];
-				}
+				var (prefix, name) = ParseName(match);
 
 				var namespaceuri = nsResolver.LookupNamespace(prefix) ?? "";
 				if (!string.IsNullOrEmpty(prefix) && string.IsNullOrEmpty(namespaceuri))
 					throw new XamlParseException($"Undeclared xmlns prefix '{prefix}'", xmlLineInfo);
+
+				IList<XmlType> typeArguments = null;
+				var childnodes = new List<(XmlName, INode)>();
+				var contentname = new XmlName(null, null);
+
+				if (remaining.StartsWith("}", StringComparison.Ordinal))
+				{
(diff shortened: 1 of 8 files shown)


samhouts:
@StephaneDelcroix ?

StephaneDelcroix:
> @StephaneDelcroix ?

I'm not comfortable with this one"
 - the test `P="{scg:List x:TypeArguments=x:String}` kinda assumes there's a ListExtension;
 - it's unclear if this adds the ability for markup extensions to be generic, or more than this;
 - it bypasses the virtual HandleProperty/SetPropertyValue methods, making it hard to evolve in the future

@akihikodaki I'd really need a better understanding of what you're trying to achieve here

akihikodaki:
> the test P="{scg:List x:TypeArguments=x:String} kinda assumes there's a ListExtension;

I used `System.Collections.Generic.List` because it is an universally available generic type and is expected to be reliable, which makes it suitable for a unit test. There is no intention to look for `ListExtension` or something kind of. Such a use of the type is conforming to [MS-XAML-2009] and works fine on Xamarin.Forms.

> it's unclear if this adds the ability for markup extensions to be generic, or more than this;

It allows any prefixed property names including `x:TypeArguments` but not limited to. An attached property will also work for example. Also, the specification compliance implies better compatibility with other XAML implementations and may reduce surprise for those who are familiar with them.

> it bypasses the virtual HandleProperty/SetPropertyValue methods, making it hard to evolve in the future

`HandleProperty` is not virtual. `SetPropertyValue` was indeed virtual and used to be called in `HandleProperty`. However that is not true any more with this change; the call was extracted from `HandleProperty`, and now the caller can determine how to handle properties. The added extensibility is necessary to handle `x:TypeArguments`, which needs different handling for the compiler and the interpreter.

samhouts:
@StephaneDelcroix Are we still considering this PR?

akihikodaki:
4078b99070a7ba7fbedbd3c9a5378fbb379f2044: rebased to 84ebf4884396e83c37ca31ad84a198d379d08ea7.

StephaneDelcroix:
could you rebase on master to fix the conflicts ? thx

akihikodaki:
e506eb53934c4a77ba69787673de877aa88d9250: rebased to 6eafaca536fd7184fbfceda33257b01880f734db.

PR #5192, opened 2019-02-10
[Core] Incorrect increments in Stepper

### LINKED ISSUE #5168: [Core] Incorrect increments in Stepper
### Description

Stepper control with a small increment may sets incorrect value.

### Steps to Reproduce

1. run Control Gallery > Stepper Gallery Legasy
2. click 3 times on latest `+` button
3. click 3 times on `-` button

### Expected Behavior

Value should be the same as it was originally = 0

### Actual Behavior

Value = 1.38777878078145E-17

### Basic Information

- Version with issue:
- Last known good version:
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Nuget Packages:
- Affected Devices:

### Screenshots

screenshot_21

AndreiMisiukevich:
### Description of Change ###

Stepper control with a small increment may sets incorrect value.
So we need to avoid values which aren't in STEP (Increment) sequence.

### Issues Resolved ### 

- fixes #5168 

### API Changes ###

None

### Platforms Affected ### 
- Core

Stepper works properly with small Increment value

Not applicable

### Testing Procedure ###

* run Control Gallery > Stepper Gallery - Legacy
* click 3 times on latest `+` button
* click 3 times on `-` button

### DIFF
--- a/Xamarin.Forms.Core/Stepper.cs
+++ b/Xamarin.Forms.Core/Stepper.cs
@@ -34,7 +34,7 @@ namespace Xamarin.Forms
 			var stepper = (Stepper)bindable;
 			var clampedValue = ((double)value).Clamp(stepper.Minimum, stepper.Maximum);
 			var decimalPlaces = stepper.DecimalPlaces;
-			return decimalPlaces >= 0 && decimalPlaces <= 15
+			return decimalPlaces >= 0
 				? Math.Round(clampedValue, decimalPlaces)
 				: clampedValue;
 		}, propertyChanged: (bindable, oldValue, newValue) =>
--- a/Xamarin.Forms.Core/Stepper.cs
+++ b/Xamarin.Forms.Core/Stepper.cs
@@ -32,7 +32,11 @@ namespace Xamarin.Forms
 		public static readonly BindableProperty ValueProperty = BindableProperty.Create("Value", typeof(double), typeof(Stepper), 0.0, BindingMode.TwoWay, coerceValue: (bindable, value) =>
 		{
 			var stepper = (Stepper)bindable;
-			return Math.Round(((double)value).Clamp(stepper.Minimum, stepper.Maximum), stepper.ValueRoundDigits);
+			var clampedValue = ((double)value).Clamp(stepper.Minimum, stepper.Maximum);
+			var decimalPlaces = stepper.DecimalPlaces;
+			return decimalPlaces >= 0 && decimalPlaces <= 15
+				? Math.Round(clampedValue, decimalPlaces)
+				: clampedValue;
 		}, propertyChanged: (bindable, oldValue, newValue) =>
 		{
 			var stepper = (Stepper)bindable;
@@ -43,7 +47,7 @@ namespace Xamarin.Forms
 
 		public static readonly BindableProperty IncrementProperty = BindableProperty.Create("Increment", typeof(double), typeof(Stepper), 1.0);
 
-		public static readonly BindableProperty ValueRoundDigitsProperty = BindableProperty.Create("ValueRoundDigits", typeof(int), typeof(Stepper), 10);
+		public static readonly BindableProperty DecimalPlacesProperty = BindableProperty.Create("DecimalPlaces", typeof(int), typeof(Stepper), -1);
 
 		readonly Lazy<PlatformConfigurationRegistry<Stepper>> _platformConfigurationRegistry;
 
@@ -77,10 +81,10 @@ namespace Xamarin.Forms
 			set { SetValue(IncrementProperty, value); }
 		}
 
-		public int ValueRoundDigits
+		public int DecimalPlaces
 		{
-			get { return (int)GetValue(ValueRoundDigitsProperty); }
-			set { SetValue(ValueRoundDigitsProperty, value); }
+			get { return (int)GetValue(DecimalPlacesProperty); }
+			set { SetValue(DecimalPlacesProperty, value); }
 		}
 
 		public double Maximum
--- a/Xamarin.Forms.Core/Stepper.cs
+++ b/Xamarin.Forms.Core/Stepper.cs
@@ -35,7 +35,7 @@ namespace Xamarin.Forms
 			var clampedValue = ((double)value).Clamp(stepper.Minimum, stepper.Maximum);
 			var decimalPlaces = stepper.DecimalPlaces;
 			return decimalPlaces >= 0
-				? Math.Round(clampedValue, decimalPlaces)
+				? Math.Round(clampedValue, Math.Min(decimalPlaces, 15)) // double has 15 decimal places. But we don't want to throw exception even though user set DecimalPlaces > 15
 				: clampedValue;
 		}, propertyChanged: (bindable, oldValue, newValue) =>
 		{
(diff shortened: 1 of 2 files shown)


mattleibow:
Thanks for catching this.

Xamarin.Forms.Core/Stepper.cs
```diff
@@ -32,7 +32,11 @@ public class Stepper : View, IElementConfiguration<Stepper>
 		public static readonly BindableProperty ValueProperty = BindableProperty.Create("Value", typeof(double), typeof(Stepper), 0.0, BindingMode.TwoWay, coerceValue: (bindable, value) =>
 		{
 			var stepper = (Stepper)bindable;
-			return ((double)value).Clamp(stepper.Minimum, stepper.Maximum);
+			var result = ((double)value).Clamp(stepper.Minimum, stepper.Maximum);
+			var increment = stepper.Increment;
+			return Math.Abs(increment) > double.Epsilon
+				? Math.Round(result / increment) * increment
+				: result;
```
mattleibow: Maybe we should actually do this the other way around to avoid the division if it has reached the bounds:

```csharp
if (value >= max)
  return max;
else if (value <= min)
  return min;
else 
  return clamp & round;
```

AndreiMisiukevich: @mattleibow sure, we can, but in this case we needn't have to call Clamp method here (max and min cases would be handled with if-else statements

mattleibow: You are correct. Release the clamps and let the beast free!

AndreiMisiukevich: @mattleibow oh.. my code it totally wrong

For example: MinValue: 1, MaxValue : 7, Step: 2, Value: 3

Round(3 / 2) * 2 = 4 .... I need to rewrite it

AndreiMisiukevich:
@mattleibow i have just update PR
I think, it would be the most right approach -- to add something like Epsilon (ValueRoundDigits)

Probably it makes sense to reach an agreement with property name.. )

mattleibow:
I think you are correct. But, a better name might be `DecimalPlaces`. I got inspiration from WinForms: https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.numericupdown

```csharp
public int DecimalPlaces { get; set; }
```

But, now we are changing existing behavior. Previously, I could set Value=1.12345678901234567890 😄 and now you are rounding to 1.1234567890. There is probably no one actually doing this, but there is always the case that "Bob-from-Accounting™️" is. We might have to set a default of `-1` which means that anything < 0 is NO rounding. For 0+, we round.

Another case to consider is that Increment=0.01 and DecimalPlaces=1. The buttons will never do anything and no one will know why. Or possible worse, Increment=0.25 and DecimalPlaces=1. Regardless of what happens, the increment will be 0.2. This is something that we can document as a "feature", but in the case of WinForms, it is purely for rendering purposes. The actual value is the real deal.

What we could do is add an additional property that is the ActualValue and that is the unrounded version. So if you set Value=x, then ActualValue=x, Value=Round(x). And if you set ActualValue=y, then ActualValue=y but Value=Round(y). The buttons then will set the ActualValue instead of Value, and ActualValue would set Value. In effect, we work with ActualValue, and that is then is passed on to Value.

Now this is more complex, but it may be useful. Or, maybe not. Any takers?

AndreiMisiukevich:
@mattleibow 
**DecimalPlaces** looks good, thanks

The 2nd approach (DecimalPlaces, Value and ActualValue) looks best from my point of view.
So, we preserve real value for those, who need it, but solve rendering issue.
Also, probably, it makes sense to change default value of DecimalPlaces to **-1**.

AndreiMisiukevich:
@mattleibow i updated PR.
So, we don't change default behavior

Xamarin.Forms.Core/Stepper.cs
```diff
@@ -43,6 +47,8 @@ public class Stepper : View, IElementConfiguration<Stepper>
 
 		public static readonly BindableProperty IncrementProperty = BindableProperty.Create("Increment", typeof(double), typeof(Stepper), 1.0);
 
+		public static readonly BindableProperty DecimalPlacesProperty = BindableProperty.Create("DecimalPlaces", typeof(int), typeof(Stepper), -1);
```
StephaneDelcroix:
```suggestion
		public static readonly BindableProperty DecimalPlacesProperty = BindableProperty.Create(nameof(DecimalPlaces), typeof(int), typeof(Stepper), -1);
```

Xamarin.Forms.Core/Stepper.cs
```diff
@@ -32,7 +32,11 @@ public class Stepper : View, IElementConfiguration<Stepper>
 		public static readonly BindableProperty ValueProperty = BindableProperty.Create("Value", typeof(double), typeof(Stepper), 0.0, BindingMode.TwoWay, coerceValue: (bindable, value) =>
 		{
 			var stepper = (Stepper)bindable;
-			return ((double)value).Clamp(stepper.Minimum, stepper.Maximum);
+			var clampedValue = ((double)value).Clamp(stepper.Minimum, stepper.Maximum);
+			var decimalPlaces = stepper.DecimalPlaces;
+			return decimalPlaces >= 0 && decimalPlaces <= 15
```
StephaneDelcroix: why the `<=15` ?

AndreiMisiukevich: @StephaneDelcroix **double** has 15 decimal digits of precision

AndreiMisiukevich: @StephaneDelcroix do you think it is unused?

StephaneDelcroix: I think it's better to validate the value of the DecimalPlacesProperty

AndreiMisiukevich: Makes sense

AndreiMisiukevich:
@StephaneDelcroix done

AndreiMisiukevich:
@PureWeen well, may we change type of Value property from **double** to **decimal**.
Won't it be breaking ?

PureWeen:
I don't think we'd have to change it from double to decimal (though it probably should have been a decimal from the start like the numeric up down control) 

I'm guessing there are definitely nuances to what I'm proposing but I'm just saying I don't quite understand why someone would use *DecimalPlaces* 

If I have a stepper that I set to 0.001 then I will expect that to increment by 0.001 each time. Whatever decimal place precision my increment has will be equivalent to what I want the decimal places precision to be.

The DecimalPlaces (as it's used on windows forms) makes sense because it's just visual. If you set the increment to 0.001 you could set decimal places to 3 because you want it to display as 0.300 but it doesn't change the internal value.

So for example the stepper could just have an internal integer counter keeping track of how many times they've clicked up and down.  If they clicked up 12 times and down 6 times then the counter would be 6. If the increment is 0.001 then inside coerceValue you could just do

6 * 0.001 = 0.006

Or we can do something similar to Tizen
https://github.com/xamarin/Xamarin.Forms/blob/bd31e1e9fc8b2f9ad94cc99e0c7ab058174821f3/Xamarin.Forms.Platform.Tizen/Renderers/StepperRenderer.cs#L49

The other part I'm wondering about (which also comes up with how this PR is done) is how this should apply to the renderers. Right now the value surfaced out is equivalent to the native control so if we coerce it at the cross platform level there's now a disconnect between those values. 

Can we coerce these values at the native level? Android and UWP we have more control over so that's not hard and iOS we can create our own stepper and override Value.

That'd be my thinking to try first. Would be to coerce at the native level (like Tizen) in a way that makes sure precision is maintained which will then in turn surface those out at the cross platform level consistently.

AndreiMisiukevich:
@PureWeen i will leave it for someone else, if you don't mind
Probably my PR and our conversation will help somebody else to provide fitting solution

StephaneDelcroix:
@PureWeen using counter * increment to get the value assumes the increment never changes...

PR #5194, opened 2019-02-10
[Visual] Material Pickers

### LINKED ISSUE #5010: [Material] Pickers
# Material Contained Picker Spec
Currently all of the picker fields themselves are basically just Entry controls. For the first pass of Material we will style the text field based on Material Entry but we won't create material versions of the pop ups themselves. For example the DatePicker will still popup natively on iOS and Android.

TimePicker, DatePicker, and the Picker control will all be included with this update

The control itself will just be a Filled Textfield (#4949) using a transparent background

image

PureWeen:
### Description of Change ###
Visual material implementation for the entry fields that are used for pickers. The pickers themselves are still native but this PR now makes the *textfields* use the same material design between iOS and Android

### Issues Resolved ### 
- fixes #5010 

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android

Android Material vs Default
image

iOS Material vs Default
image

### Testing Procedure ###
Run Visual Gallery and test out the pickers.
Compare iOS and Android

```C#
using System;
using System.Text;

namespace Xamarin.Forms.Sandbox
{
	public partial class App 
	{
		// This code is called from the App Constructor so just initialize the main page of the application here
		void InitializeMainPage()
		{
			var picker = new Picker() { ItemsSource = new List<string> { "cat" } };
			var datePicker = new DatePicker();
			var timePicker = new TimePicker();

			Material.Picker.SetPlaceholder(picker, "Click to Picker");
			Material.DatePicker.SetPlaceholder(datePicker, "Click to date Picker");
			Material.TimePicker.SetPlaceholder(timePicker, "Click to time Picker");


			Material.Picker.SetPlaceholderColor(picker, Color.Green);
			Material.DatePicker.SetPlaceholderColor(datePicker, Color.Pink);
			Material.TimePicker.SetPlaceholderColor(timePicker, Color.Yellow);

			MainPage = new ContentPage()
			{
				Content = CreateStackLayout(new View[] { picker, datePicker, timePicker  })
			};

			MainPage.Visual = VisualMarker.Material;
		}
	}
}
```

### PR Checklist ###

- [ ] Has automated tests 


PureWeen:
The failing tests aren't related to this PR

rmarinho:
Can we have just the picker changes ?

PureWeen:
> Can we have just the picker changes ?

@rmarinho 
All gone 😢

mattleibow:
> > Can we have just the picker changes ?
> 
> @rmarinho
> Removed sandbox from this PR 😢

Hehe, now that the sandbox has been merged into 3.6.0, if we update this branch, we get the sandbox back 😸

mattleibow:
Still going through this, but some observations as I run through it...

* There seems to be an "issue" with the underline on Android - this might be an issue with the native implementation, but might not:
  android-entry

* I also notice that the pickers look "empty" without the Title. Looking at the iOS and Android implementations, we might be able to get away with adding the Title property to all of them, and just implement it on the material. It already exists for the vanilla "picker", so that is fine. For the dat and time pickers, the default implementation has no place for it (or rather the place is occupied by an always-selected value). Material, on the other hand, looks like it needs the title.

* When navigating back after typing this comment, I noticed the default picker did not launch on Android:
  android-pickers

PureWeen:
@mattleibow 

>     * I also notice that the pickers look "empty" without the Title.

I added a material attached property to set the place holders. Some work also needs to be done on Entry when there is no placeholder so the height just shrinks to the text but that will be a different PR.

>      * There seems to be an "issue" with the underline on Android

Unfortunately this is the effect of my workaround to center the PH text and I'm not sure if we can fix now. Android X fixes the placeholder positioning so it is centered

>     * When navigating back after typing this comment, I noticed the default picker did not launch on

Yea I inherited from the non appcompat pickerrenderer by accident. I've updated the correct base class and now this should be all working

>     * The Material picker looks very different to the non-Material and has 2 buttons,

We'll address the popups in a later PR :-/ Maybe with your Alerts PR that you started? We can make some material popups that look the same

PR #5209, opened 2019-02-12
[Material] [Android, iOS] Added Slider and ProgressBar

### LINKED ISSUE #5008: [Material] Linear Progress Indicator

### LINKED ISSUE #5018: [Material] [Android, iOS] Slider Spec
# Material Slider 

https://material.io/design/components/sliders.html

- **Visually Identical**

- **Performance**
Material renderer should be modeled after the fast renderer

Slider doesn't expose very many apis so the expectation here is that the default would size and color itself the same on iOS and Android

The hopefully not too tricky part will be that there is only a Material slider on iOS. On the android one it's still using the original control so material visual on Android will just apply default styles

#### Current state
image

# Scenarios

```XAML
<Slider Visual="Material" />
```

# Difficulty : medium possibly difficult

mattleibow:
### Description of Change ###

Added the Material versions of the ProgressBar and Slider.

### Issues Resolved ### 

 - fixes #5008 
 - fixes #5079 
 - fixes #5018 

### API Changes ###

Added the Material renderers for Slider and ProgressBar for Android and iOS.

The rules for the coloring are as follows:

 - All default colors, then use the defaults
 - A single color for `ProgressBar.ProgressColor` or `Slider.MinimumTrackColor`, then follow material: a single color throughout (including the slider thumb), but the background has an alpha
 - A single background color for `ProgressBar.BackgroundColor` or `Slider.MaximumTrackColor`, then just set the background color
 - Both progress and background, then throw material out and set colors explicitly
 - A default slider thumb, then fall back to the primary color
 - A specific thumb, then use that explicitly

### Platforms Affected ### 

- iOS
- Android

New Material renderers.

|   |  Android  |  iOS  |
|-:| :-: | :-: |
| Defaults |  |  |
| Height of 50 |  |  |
| Primary color of Red |  |  |
| Primary and Background |  |  |
| All |  |  |
| All with 50% alpha |  |  |

### Testing Procedure ###

Loads of buttons and switches on the `MaterialProgressBarGallery` gallery or "[Material] ProgressBar & Slider Gallery".

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls/Controls/ColorPicker.cs
+++ b/Xamarin.Forms.Controls/Controls/ColorPicker.cs
@@ -1,0 +1,171 @@
+using System;
+
+namespace Xamarin.Forms.Controls
+{
+	public class ColorPicker : ContentView
+	{
+		public static readonly BindableProperty UseDefaultProperty = BindableProperty.Create(nameof(UseDefault), typeof(bool), typeof(ColorPicker), false,
+			propertyChanged: OnColorChanged);
+
+		public static readonly BindableProperty ColorProperty = BindableProperty.Create(nameof(Color), typeof(Color), typeof(ColorPicker), Color.Default,
+			propertyChanged: OnColorChanged);
+
+		public static readonly BindableProperty TitleProperty = BindableProperty.Create(nameof(Title), typeof(string), typeof(ColorPicker), "Pick a color:",
+			propertyChanged: OnTitleChanged);
+
+		static readonly string[] _components = { "R", "G", "B", "A" };
+
+		Label _titleLabel;
+		Slider[] _sliders;
+		BoxView _box;
+		Label _hexLabel;
+		Switch _useDefault;
+
+		public ColorPicker()
+		{
+			var grid = new Grid
+			{
+				Padding = 0,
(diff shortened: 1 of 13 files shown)


PureWeen:
Failing tests unrelated

paymicro:
[Android] All with 50% alpha
image
I think the lines should be without space.

mattleibow:
@paymicro thanks for pointing out the weird thing. I for got to mention it on this PR.
As far as I can see, this is not something that most people would do, and if they do, the shouldn't 😄Unless we are going to make time for a custom slider... in addition to the progress bar.

PR #5212, opened 2019-02-12
Add DesignerInterface

### LINKED ISSUE #4429: Future proof API for design time rendering
We have been bitten several times by API removals in forms. This is the most recent occurrence of the problem https://github.com/xamarin/Xamarin.Forms/pull/4355 .

Rather than invoking APIs directly as we have been doing in the past, would it be possible to provide a static class which encompasses every invocation we need? This insulates us from changes to APIs and allows Forms a bit more leeway in making internal changes. It could also make it a little easier to make versioned changes to the API.

chkn:
### Description of Change ###

Adds an API contract for the Forms previewer.

For Core and Xaml assemblies, members have been added to the `DesignerInterface` class. If some API surface changes in Forms, we can catch it and update the `DesignerInterface` implementation without changing its API surface, allowing older versions of the previewer to work with newer versions of Forms. If we cannot keep the `DesignerInterface` backward compatible, there is now a version that we can bump, so older versions of the previewer can fail gracefully with a helpful error message.

For the platform-specific assemblies, it was not possible to reference their types directly here, so a list is added at the end for reference.

fixes #4429

### API Changes ###

Added:
 -  Xamarin.Forms.Xaml.DesignerInterface

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

> VS bug [#790714](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/790714)

### DIFF
--- a/Xamarin.Forms.Xaml/DesignerInterface.cs
+++ b/Xamarin.Forms.Xaml/DesignerInterface.cs
@@ -1,0 +1,120 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+using Xamarin.Forms.Internals;
+
+namespace Xamarin.Forms.Xaml
+{
+	// IMPORTANT!!
+	//  READ BEFORE MODIFYING!!
+	// This class represents the API that the Forms previewer will use
+	//  via reflection. Adding members or changing implementations is fine, but
+	//  DO NOT break this API unless you know what you're doing and/or are
+	//  coordinating with the designer team.
+	// Thanks!
+	static class DesignerInterface
+	{
+		const int LatestSupportedVersion = 1;
+		static int Version => LatestSupportedVersion;
+
+		// These types are used via reflection for type checks..
+		static Type IMarkupExtensionType => typeof (IMarkupExtension);
+		static Type ViewType => typeof (Xamarin.Forms.View);
+		static Type BoxViewType => typeof (Xamarin.Forms.BoxView);
+		static Type VisualElementType => typeof (Xamarin.Forms.VisualElement);
+		static Type ApplicationType => typeof (Xamarin.Forms.Application);
+
+		static Func<ResourceLoader.ResourceLoadingQuery, ResourceLoader.ResourceLoadingResponse> ResourceProvider2 {
+			get => ResourceLoader.ResourceProvider2;
+			set => ResourceLoader.ResourceProvider2 = value;
+		}
+
+		static Action<Exception> ExceptionHandler {
+			get => ResourceLoader.ExceptionHandler;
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Xaml/DesignerInterface.cs
```diff
@@ -0,0 +1,120 @@
...
+
+	// Android:
+	//
+	//  static bool Forms.IsInitialized { get; }
+	//  static void ResourceManager.Init(Assembly masterAssembly)
+	//  static bool FileImageSourceHandler.DecodeSynchronously { set; }
+	//  static void Platform.SetPageContext(BindableObject bindable, Context context)
+	//  void IVisualElementRenderer.UpdateLayout()
+	//  ViewGroup IVisualElementRenderer.ViewGroup { get; }*
+	//  static void Forms.Init(Context activity, Bundle bundle)*
```
chkn: The above 2 starred members appear to be deprecated. Should we take this opportunity to switch to the preferred APIs?

Xamarin.Forms.Xaml/DesignerInterface.cs
```diff
@@ -0,0 +1,120 @@
...
+
+		static Func<IList<XamlLoader.FallbackTypeInfo>, Type, Type> FallbackTypeResolver {
+			get => XamlLoader.FallbackTypeResolver;
+			set => XamlLoader.FallbackTypeResolver = value;
+		}
+
+		static Action<XamlLoader.CallbackTypeInfo, object> ValueCreatedCallback  {
+			get => XamlLoader.ValueCreatedCallback;
+			set => XamlLoader.ValueCreatedCallback = value;
+		}
```
chkn: Can we also make `FallbackTypeInfo` and `CallbackTypeInfo` classes instead of structs?

StephaneDelcroix:
Thanks @chkn for coming up with this list. I had a different idea on how to achieve this, here's how I think this should be done:

1/ Create a separate assembly (and nuget, eventually) `Xamarin.Forms.Loader` that contains all the API used by the previewer, but that doesn't have a dependency on XF
2/ Move all the reflection code to this assembly, and have the previewer take a dep on it
3/ Version and update this assembly on it's own schedule, not tied to XF's

With that, we share the contract together (previewer and xf) and we have a way to test the contract.

what are your thoughts ?

chkn:
@StephaneDelcroix It sounds like that provides a weaker guarantee (in this case, `DesignerInterface` can't break due to Forms API change, because Forms wouldn't compile) and has more moving parts, but I'm not sure I fully understand your proposal.

Perhaps we should have a call with @alanmcgovern? Having an interface like this was originally his brainchild

PR #5223, opened 2019-02-13
[iOS] Frame.CornerRadius faulty

### LINKED ISSUE #5199: iOS - Frame.CornerRadius faulty
### Description
The corner radius is faulty on iOS

### Steps to Reproduce
Code-Snipped:
`                            <Frame CornerRadius="80" Grid.Row="0" Padding="2.5" BackgroundColor="Transparent" HeightRequest="80" WidthRequest="80"  HorizontalOptions="Center" BorderColor="{StaticResource LightBackgroundColor}">
                                <fontAwesome1:FontAwesomeIcon  Text="fas fa-store" TextColor="{StaticResource LightTextColor}" HorizontalOptions="Center" VerticalOptions="Center"  FontSize="35"   />
                            </Frame>
   `

### Expected Behavior
Round Frame
### Actual Behavior
Square Frame with edges
### Basic Information

- Version with issue: XF 3.4.0.1009999
- IDE: Visual Studio Enterprise 2017
- Platform Target Frameworks: 
  - iOS:  12.1
- Affected Devices: iPhone 7

### Screenshots

image

AndreiMisiukevich:
### Description of Change ###

Prevent setting corner radius greater than Width/Height (As described in original issue)

### Issues Resolved ### 

- fixes #5199

### API Changes ###
None

### Platforms Affected ### 
- iOS

No "broken" shapes

**BEFORE:**

before1
before2

**AFTER:**
after1
after2

### Testing Procedure ###
Small code snippet for testing

```csharp
MainPage = new ContentPage
{
	Content = new StackLayout
	{
		Children =
		{
			new Frame
			{
				VerticalOptions = LayoutOptions.CenterAndExpand,
				HorizontalOptions = LayoutOptions.CenterAndExpand,
				WidthRequest = 100,
				HeightRequest = 100,
				CornerRadius = 100,
				BackgroundColor = Color.Red,
			}
		}
	}
};
```


programmation:
Please don't do this. Sometimes we _want_ to have CornerRadius > Height or > Width to achieve a required look.

AndreiMisiukevich:
> Please don't do this. Sometimes we _want_ to have CornerRadius > Height or > Width to achieve a required look.

how do you handle it on android?

programmation:
I don’t do anything platform-specific. 

If I want a round button (or elliptical, or whatever), I use the `OnPropertyChanged` callback to adjust the `CornerRadius` when either the `Width` or the `Height` changes. 

IMO the original issue as reported is an incorrect usage of `CornerRadius` rather than a bug in XF as such. The fix for the original issue is to set the `CornerRadius` to the right value.

It would be interesting to see if `CornerRadius` (or any other XAML property) could be specified as some expression derived from another property (e.g. Min(Height, Width) / 2) in the XAML. Perhaps with a MarkupExtension, along the lines of StringFormat?

Failing that the `OnPropertyChanged` callback works well enough.

The current implementation is fine AFAICT.

AndreiMisiukevich:
@programmation but currently platforms have different behavior :)

programmation:
I did not know that! What's the difference?

programmation:
I just tried a `Frame` and a `Button` and they both behave the same (as far as CornerRadius goes, anyway) on both Android and iOS. This is with XF 3.5. Perhaps an earlier version had some differences?

AndreiMisiukevich:
@programmation hm on Forms 3.1 android will draw proper circle if you try to set HeightRequest=WidthRequest=CornerRadius

programmation:
That suggests to me that there's a fundamental misunderstanding (or a bug in implementation) of what "CornerRadius" actually means. How can "corner radius" ever equal "height" and make a circle? It makes no sense mathematically. So I infer that either Android or XF is limiting the CornerRadius when it shouldn't be. 
I guess it may not be the most important thing in the world, but if we can avoid creating artificial limitations to cope with things that are actual errors in usage I think that would enable many more useful things to be done with the framework.

samhouts:
I think this needs some discussion, based on the comments here and on the issue.

samhouts:
@AndreiMisiukevich I'm going to go ahead and close this PR to keep it from getting stale, but let's continue the discussion on the issue. Thanks!!

PR #5235, opened 2019-02-13
[UWP] improve performance

paymicro:
### Description of Change ###

Improve performance.

### Issues Resolved ### 

none

### API Changes ###

Changed:
```C# 
public interface IElementConfiguration<out TElement>
```
to
```C#
public interface IElementConfiguration<TElement>
```

### Platforms Affected ### 

- Core (all platforms)
- UWP

None

Not applicable

### Testing Procedure ###

- run any UItest

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/AbsoluteLayout.cs
+++ b/Xamarin.Forms.Core/AbsoluteLayout.cs
@@ -6,7 +6,7 @@ using Xamarin.Forms.Internals;
 
 namespace Xamarin.Forms
 {
-	public class AbsoluteLayout : Layout<View>, IElementConfiguration<AbsoluteLayout>
+	public class AbsoluteLayout : Layout<View>, IFastElementConfiguration<AbsoluteLayout>
 	{
 		public static readonly BindableProperty LayoutFlagsProperty = BindableProperty.CreateAttached("LayoutFlags", typeof(AbsoluteLayoutFlags), typeof(AbsoluteLayout), AbsoluteLayoutFlags.None);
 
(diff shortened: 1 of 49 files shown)


PureWeen:
Oh I see because in that commit it got added to the Cell
Hmmm I still don't think we can remove the out on IElementConfiguration unfortunately as that will change too much. Can we achieve the same thing @AndreiMisiukevich achieved in that PR another way? Maybe create an IElementConfiguration that doesn't have the out parameter? or something else?

AndreiMisiukevich:
> Oh I see because in that commit it got added to the Cell
> Hmmm I still don't think we can remove the out on IElementConfiguration unfortunately as that will change too much. Can we achieve the same thing @AndreiMisiukevich achieved in that PR another way? Maybe create an IElementConfiguration that doesn't have the out parameter? or something else?

Hi
Do you mean this PR https://github.com/xamarin/Xamarin.Forms/pull/4002/files ?

paymicro:
> * Can we make this change? It's a public interface so it'll probably break people right?

Right. Probably breaking change, but i think that such a chance is extremely small.

> * This has had an our parameter since 2016 right? It was an out generic in 3.5 when the performance was fine so it seems like there's a different cause we can look to in order to fix the

In my opinion, this initially reduced productivity, but the number of elements did not reach a critical mass.

> Maybe create an IElementConfiguration that doesn't have the out parameter? or something else?

Hm. Maybe create `IFastElementConfiguration`... or `IElementConfiguration2`. And mark `IElementConfiguration` as obsolete... I'll see what I can do.

paymicro:
@AndreiMisiukevich yes. https://github.com/xamarin/Xamarin.Forms/pull/4002 markedly reduces performance on UWP.

StephaneDelcroix:
any measures ?

samhouts:
@hartez Yeah, it was like three years ago, but do you remember if we had a compelling reason to make `IElementConfiguration` covariant?

samhouts:
This is currently blocking #2432, so we need to address it quickly.

PureWeen:
closed because of
https://github.com/xamarin/Xamarin.Forms/pull/5449

PR #5248, opened 2019-02-14
[Performance] Allow manual initialization of expensive operations to potentially decrease startup time

kingces95:
To generate this profile you need to build a private nuget package with this branch and then build https://github.com/kingces95/droidperf.
profile

Goal is to reduce startup time by making renderer registration explicit and allowing for feature opt out and to build in a profiler to allow adding regression tests.
NO BREAKING CHANGES -- All features are opt in: A new `Init` call + profiler logic lives behind a define.

To generate the nuget:
> set NUGET_DIR=x:\nuget\4.4.1\
> .create-nuget.bat rdroid
> X:\git\forms\.nuspec\Xamarin.Forms.9.9.46.nupkg

To build the droidperf:
* Install VS Ent -- only that supports AOT
* Config = Release
* Build/Deploy XFFastUp.Android

Profiler API:
The profiler API is built in but hidden behind a PROFILE define. The define is set in this branch to ease development. Will remove before accepting PR. To inject sampling points simply put:
> Profile.Push("Timing Partition A")
> ... // logic to time
> Profile.PopPush("Next Partition")
> ... // more logic to time
> Profile.Pop()
To report on the samples use the droidperf app linked above.

Explicit Registration:
This is an example of using explicit renderer registration and optional opt-out of features. 
* `Handlers` is an array of `ExportRendererAttribute`
* `Flags` is a list of features to opt out of using
Taken from: https://github.com/kingces95/droidperf/blob/23f8a333c08fd9f346ef3f4ef93ebfbbe68bc23a/XFMinUp/XFFastUp.Android/MainActivity.cs#L41
```C#
var activation = new ActivationOptions()
{
	Activity = this,
	Bundle = savedInstanceState,
	ResourceAssembly = Assembly.GetExecutingAssembly(),
	Handlers = Exports.Handlers,
	Flags = ActivationFlags.NoCss
};
Forms.Init(activation);
//Forms.Init(this, savedInstanceState, Assembly.GetExecutingAssembly());
```


mattleibow:
Nice!

KSemenenko:
any updates?

Xamarin.Forms.Core/Registrar.cs
```diff
@@ -251,9 +259,57 @@ static Registrar()
...
+			var length = attributes.Length;
+			for (var i = 0; i < length; i++)
+			{
+				var attribute = attributes[i];
+				if (attribute.ShouldRegister())
+					Registered.Register(attribute.HandlerType, attribute.TargetType, attribute.SupportedVisuals);
+			}
+		}
+
+		public static void RegisterSytlesheets()
```
hartez: RegisterSytlesheets -> RegisterStyleSheets

Xamarin.Forms.Core/Profiler.cs
```diff
@@ -0,0 +1,123 @@
...
+			public int Line;
+		}
+		public static List<Datum> Data = new List<Datum>(Capacity);
+
+		static Stack<Profile> s_stack = new Stack<Profile>(Capacity);
+		static int s_depth = 0;
+		static bool s_stopped = false;
+		static Stopwatch s_stopwatch = new Stopwatch();
+
+		readonly long start;
```
hartez: Any reason you're not using the `_` prefix for these?

Xamarin.Forms.Core/Profiler.cs
```diff
@@ -0,0 +1,123 @@
...
+
+			s_depth++;
+		}
+		public void Dispose()
+		{
+			if (s_stopped && start == 0)
+				return;
+
+			var ticks = s_stopwatch.ElapsedTicks - start;
+			var depth = --s_depth;
```
hartez: `depth` doesn't appear to be used here.

Xamarin.Forms.Core/Profiler.cs
```diff
@@ -0,0 +1,123 @@
...
+
+		public static void Stop()
+		{
+			// unwind stack
+			s_stopped = true;
+			while (s_stack.Count > 0)
+				s_stack.Pop();
+		}
+		[Conditional(PROFILE)]
+		public static void Push(
```
hartez: I like the way this is designed, but I'm wondering about calling the methods `Push`, `Pop`, and `PopPush`. That sort of requires that the user knowing that the profiler is a stack, internally. If I want to start profiling something, I push it onto the stack, if I want to stop profiling it I pop it off the stack - would it be easier for a future user if the methods were just called `Start`, `Stop`, and something like `Checkpoint` for `PopPush`? 

Not a deal breaker, since everyone using this will be be, by definition, a programmer who should know what a stack is. But it might be easier for new folks to pick up (and for use to read when reviewing).

kingces95: Oh, well, sure. If it's `start` then there will be many nested `start` calls which seems odd.. Sucker is taking samples inside stack frames so `push` and `pop` seemed reasonable. I could get behind changing `PopPush` to `Sample`. How about that?

Xamarin.Forms.Core/Profiler.cs
kingces95: Habit. I'll refresh the conventions to be consistent. Thx!

kingces95:
Got the go ahead to check this in behind a PROFILE define. So I'll address the concerns, double check ensure everything is behind the PROFILE define, then get a final review, and merge it in... that's my plan.

Xamarin.Forms.Core/Profiler.cs
hartez: I can live with `PopPush` -> `Sample` and leaving the others as-is.

kingces95: How about `Start` and `Stop` to enable the profiler. Without calling `Start` all the public entry points are noops and are small enough so they should be in-lined so that their cost is minimal and we're able to simply ship the profiler disabled in the product.

Then `Push`, `Pop` and `PopPush` become `SampleStart`, `SampleEnd` and `SamplePartition`.

kingces95:
Eh, changed my plan. Instead I'm not gonna hide behind a PROFILE! HaHA! Bait-and-switch!

Instead, I'll add `Start`. Without calling `Start` all the public entry points are noops and are small enough so they should be in-lined so that their cost will minimal and allow us to simply ship the profiler disabled in the product.

kingces95:
Build

kingces95:
Gonna push this one instead of the split one. The feature is the explicit init while the profiler is the test and the test required some infrastructure. So to test the first one, I needed the second one. So the code was verified as a pair.. Going forward, other perf work may look similar -- a feature + some injected profiling sampling points needed to expose the target timings and verify they go away.

Xamarin.Forms.Core/Profiler.cs
rmarinho: @kingces95 can you address this

kingces95: Done.

Xamarin.Forms.Core/Profiler.cs
kingces95: Removed

Xamarin.Forms.Core/Profiler.cs
kingces95: I thought he was talking about the statics. I'll switch the statics to caps and instance to underbars.

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -127,7 +128,11 @@ protected void LoadApplication(Application application)
 
 			application.PropertyChanged += AppOnPropertyChanged;
 
+			//Profile.FramePartition(nameof(SetMainPage));
```
hartez: Why are all these Profile calls commented out?

PR #5264, opened 2019-02-14
Setup sandbox app inside our project

PureWeen:
### Description of Change ###

This PR also adds a Sandbox project which greatly improves the development loop. It deploys to android in about 14 seconds vs the 40 seconds required for Control Gallery so it's much quicker to iterate with.

It's also a more friendly place for people to tinker with the source code

It's a more natural place for people to start when opening the Gallery and we can add some specific methods for helping people out. For example I added some simple methods for creating a stack layout page with safe padding.

People could also use this for logging issues. They can just paste the contents of the App.starthere.cs file into an issue instead of attaching a project


mattleibow:
This probably should be rebased on 3.6.0.

PureWeen:
rebased :-)

mattleibow:
I just pushed a fix for UWP because there were clashes with the `Application` and `Frame` types - seems they exist in UWP and Forms.

PR #5277, opened 2019-02-15
Entry ClearButtonMode

andreinitescu:
### Description of Change ###

Added `ClearButtonMode` property to Entry. For now it only has two values: `Never` and `WhileEditing` (shows clear button only when Entry has focus and text is set).

iOS has the feature built-in. 
Android: I had to build this
UWP: has it bult-in. 

It's a very common requirement in the UI, It's tiresome to keep adding an effect in every app for this.

### API Changes ###

Added:

`Entry.ClearButtonMode` bindable property

### Platforms Affected ### 

- Core
- iOS
- Android

See screenshots below

Android:
image

iOS:
image

### PR Checklist ###

- [ ] Has automated tests 


mattleibow:
I think you can do this with UWP. But, I believe there is a control in the community toolkit with even more options. Maybe it is time to have a look at that - or at least the new WinUI library and see if the control in there can do it.

andreinitescu:
@mattleibow 
> I think you can do this with UW

Yes it can be done, but what I meant is there's no built-in way to make the close button not to appear. I wonder if it's even worth doing it, even if it sacrifices consistency.

> But, I believe there is a control in the community toolkit with even more options

Just so I understand what you mean exactly, which feature/control are you referring to?  

The question I have is, should we have a property on the Entry instead of platform specifics?

mattleibow:
Oops, forgot for a second that UWP automatically adds the clear button. I was thinking actual buttons.
I think it should be OK to not have the ability to hide the clear button (such as UWP) or not have a button at all (Tizen).

I did see this: https://social.msdn.microsoft.com/Forums/en-US/f522a1f4-e171-49b2-be8e-7a4346978ff0/uwphow-to-restrict-the-clear-button-to-clear-the-values-in-the-textbox-in-uwp
But, that appears to be an anti-pattern, and probably not worth the effort.

rookiejava:
> I'll ask the Tizen team if there is a way to have a clear button.

Yes, we can. 😉

mattleibow:
@andreinitescu Seems that a property on the control is a good thing, maybe for v1 we can just let UWP ignore it and always show. WPF/macOS probably are not _that_ essential at the moment.

rookiejava:
I can help you implement it on Tizen. Or if you have any problem setting up the Tizen development environment, I can do it myself. :)

andreinitescu:
@rookiejava Awesome, thanks! I don't have the Tizen dev tools installed, so if you'd like, you can open a separate ticket for Tizen impl and go for it.

but first, @mattleibow are we sure a property makes more sense than platform specifics? Property is definitely more easier to use, just want to be sure if we're not missing anything, so effort and time is not wasted with refactoring to property.

mattleibow:
@andreinitescu I think a property is fine. We are doing a lot of work just for Android and iOS. And, it works on UWP (it can be removed either with the code in the SO post or a custom control template). And it works on Tizen. 

I see no need to hold back those platforms just for macOS.

andreinitescu:
I changed to using a property.

Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
```diff
@@ -488,9 +493,16 @@ void SetSelectionLengthFromRenderer(int selectionLength)
...
-    }
+		void UpdateIsReadOnly()
+		{
+			Control.UserInteractionEnabled = !Element.IsReadOnly;
+		}
+
+		void UpdateClearButtonVisibility(bool isInitializating)
+		{
+			bool showClearButton = Element.ClearButtonMode == InputViewClearButtonMode.WhileEditing;
+			if ((isInitializating && showClearButton) || !isInitializating)
```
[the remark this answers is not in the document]
andreinitescu: Thanks for the review! 
I think the logic is correct albeit the condition could be simplified to: 
`if(!isInitializing || showClearButon)`

Here's the logic I implemented:
My goal is to call the native API (`Control.ClearButtonMode`) when it's really needed, which is in two situations:
- either XF `ClearButtomMode` property has changed, and in this case native API must be set to reflect the XF property change
- or the renderer is initializing and the XF property is set to `WhileEditing` so native must be called to reflect the mode (on iOS the default mode for UITexField is not to show the clear button)

Further more, Instead of `isInitializing` and negation, I changed it to `forceUpdate`, I pushed the changes, it should be clearer.

Or am I missing something, what's the bug you're seeing?

Xamarin.Forms.Platform.Android/Renderers/EntryRenderer.cs
```diff
@@ -477,4 +485,85 @@ protected virtual void UpdateIsReadOnly()
 			Control.Focusable = isReadOnly;
 		}
 	}
+
+	// Entry clear button management
+	public abstract partial class EntryRendererBase<TControl>
```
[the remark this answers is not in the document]
andreinitescu: I'm not decided. I initially wanted to have the code which deals with clear button separately. 

Using a static class like `EntryClearButtonManager` could be better but I wanted to have a way to override the drawable in case someone wants to: 
`protected virtual Drawable GetCloseButtonDrawable()`.
 Another option is to have this virtual method in the EntryRendererBase class and pass the drawable to the `EntryClearButtonManager`.

Or I could just give up on this idea and put everything int the EntryRendererBase..
What do you think?

andreinitescu:
> Also, what happens when you use a Material text box

I added to the "Entry Material Demos" gallery page a button to toggle the ClearButtonMode on all entries.
On iOS, it works great.
On Android, there's an issue with the material Entry, but I am not if it's related to my code. If you tap/click on the Entry, the `ViewRenderer.OnNativeFocusChanged` is not called and I rely on it to show the clear button.  If you hit Tab key in emulator, the `ViewRenderer.OnNativeFocusChanged` and clear button appears.

Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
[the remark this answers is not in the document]
mattleibow: Ah, I again thought that result of the if statement was the value set. My bad.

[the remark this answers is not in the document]
andreinitescu: No worries. Let me know if you think there's something which could improve readability.

Xamarin.Forms.Platform.Android/Renderers/EntryRenderer.cs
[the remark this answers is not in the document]
mattleibow: The manager might be a good idea. You can also use an interface so that the manager can talk back to the entry.

mattleibow:
That souls like a bug. I know @samhouts was working on focus and tabstops, but I am not sure if this is something she came across and fixed.

Xamarin.Forms.Platform.Android/Renderers/EntryRenderer.cs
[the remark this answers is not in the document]
andreinitescu: I can do that, sure. Yet, the partial class is simple and it doesn't need additional classes/interfaces... 
I wonder if it would help if I just move it to a separate file like EntryBaseRenderer.ClearButton.cs

Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
```diff
@@ -488,9 +491,16 @@ void SetSelectionLengthFromRenderer(int selectionLength)
...
-        {
-            Control.UserInteractionEnabled = !Element.IsReadOnly;
-        }
-    }
+		void UpdateIsReadOnly()
+		{
+			Control.UserInteractionEnabled = !Element.IsReadOnly;
+		}
+
+		void UpdateClearButtonVisibility(bool forceUpdate)
```
adrianknight89: Can you explain why we need this `forceUpdate` flag?

andreinitescu: Sure, take a look [here](https://github.com/xamarin/Xamarin.Forms/pull/5277#discussion_r257570333)

rookiejava: What do you think about below (make `UpdateClearButtonVisiblilty()` simple) ? 


```cs
protected override void OnElementChanged(ElementChangedEventArgs<Entry> e)
{
 //...
    UpdateCursorColor();
    UpdateIsReadOnly();
    if (Element.ClearButtonMode == InputViewClearButtonMode.WhileEditing) // or if (Element.IsSet(Entry.ClearButtonModeProperty))
        UpdateClearButtonVisibility();
// ...
}

void UpdateClearButtonVisibility()
{
    Control.ClearButtonMode =  Element.ClearButtonMode == InputViewClearButtonMode.WhileEditing ? UITextFieldViewMode.WhileEditing : UITextFieldViewMode.Never;
}
```

adrianknight89: This should be the preferred approach, imo. Setting Control.ClearButtonMode to its default value during initialization should not cost a performance hit, so there is no need to introduce an extra flag.

andreinitescu: @rookiejava Why would calling the `ClearButtonMode` bindable property on `ElementChanged` be better? A simple flag is much much less costly than reading bindable property.

rookiejava: This is simply why `ClearButtonMode bindable property is always used. (regardless of flag)

```cs
bool showClearButton = Element.ClearButtonMode == InputViewClearButtonMode.WhileEditing;  //<--- this line is always executed.
```

andreinitescu: Maybe I wasn't clear. `Element.ClearButtonMode` actually calls `BindableObject.GetValue(ClearButtonModeProperty)` under the hood, which goes into BindableObject implementation and tries get the bindable property value. 
Doing this on each and every Entry initialization on a Page can be avoided by a simple flag per my implementation above.

rookiejava: The flag you are talking (`forceUpdate`) is used to determine whether to call `Control.ClearButtonMode`, not `Element.ClearButtonMode`. In other words, regardless of whether the flag (`forceUpdate`) value is `true` or `false`, `Element.ClearButtonMode` is always called in your suggestion.

andreinitescu: Guys, thanks for your feedback, I implemented the changes. I removed the flags. 
I initially didn't like the idea of adding any more logic right in the `OnElementChanged()`, there's already quite some logic already. But on a 2nd thought I guess it's better to do that instead of having the logic of those flags...

andreinitescu:
The issue still remains with material on Android.

samhouts:
@andreinitescu Would you mind reporting the Focus issue separately? Thanks!

Xamarin.Forms.Core/InputView.cs
```diff
@@ -38,4 +38,10 @@ public bool IsReadOnly
 			set { SetValue(IsReadOnlyProperty, value); }
 		}
 	}
+
+	public enum InputViewClearButtonMode
```
StephaneDelcroix:
```suggestion
	public enum ClearButtonMode
```

andreinitescu:
@StephaneDelcroix iOS using exactly same naming: 
https://docs.microsoft.com/en-us/dotnet/api/uikit.uitextfield.clearbuttonmode?view=xamarin-ios-sdk-12

```
[get: Foundation.Export("clearButtonMode")]
[set: Foundation.Export("setClearButtonMode:")]
public virtual UIKit.UITextFieldViewMode ClearButtonMode { get; set; }
```

StephaneDelcroix:
> iOS using exactly same naming

the naming of the property type indicates this is a _view_ or _visibility_ thing, which is a bit better.



> it's just a matter of reading the documentation

:P who does that ??? :P

andreinitescu:
> the naming of the property type indicates this is a view or visibility thing, which is a bit better.

Yes, that's why I wanted it to be named  `enum InputViewClearButtonMode` . But I see you suggested `enum ClearButtonMode`. Sorry, I'm confused of what you're suggesting

StephaneDelcroix:
`InputViewClearButtonMode` is `Class` + `Property`, that's why I suggested to drop the `Class` part.

Xamarin.Forms.Core/InputView.cs
```diff
@@ -38,4 +38,10 @@ public bool IsReadOnly
...
+	{
+		Never,
+		WhileEditing
```
rmarinho: Maybe add a `Default`

andreinitescu:
Naming is always hard. In my opinion I would not make it general, but keep it related to the InputView class name (e.g. `InputViewClearButtonMode`) and the enum values related to how it's displayed when the input view has focus (never show the button or show it only while editing).

But you're the stake holders, so let me know which one you want, because I see there are couple different suggestions..

jcmanke:
If you want the enum to be related to the InputView class to the point of defining it in the same file, I wonder if the property should also be placed on the InputView and not just one of its subclasses?

andreinitescu:
It would be great if this could get merged sooner, it's been 4 months. I'd like to avoid having to write renderers again for this purpose.

rmarinho:
Sure @andreinitescu let's start with a rebase to run tests :) 

I will review again and talk with the team internally to decide on the naming thing think is whats blocking this. 

Thanks

andreinitescu:
@rmarinho You mean rebase again against master, on another branch?

rmarinho:
> @rmarinho You mean rebase again against master, on another branch?

Master

andreinitescu:
@rmarinho 

1. Rebased against master.
2. Added UI test to toggle the clear button mode.
3. Moving ClearButtonMode to InputView requires more work especially on iOS where the native control used to render the Editor doesn't support showing a close button. It's doable but it requires more work. 
Can we please get this feature for Entry first and once it's merged and released, I can work on adding it for Editor and Shell too? Otherwise it can take another 4 months. Thanks.

Please let me know what you think.

andreinitescu:
@rmarinho 

1. Rebased against master.
2. Removed "InputMode" naming as you suggested.
3. Added UI test to toggle the clear button mode.
4. Moving ClearButtonMode to InputView requires more work especially on iOS where the native control used to render the Editor doesn't support showing a close button. It's doable but it requires more work. 
Can we please get this feature for Entry first and once it's merged and released, I can work on adding it for Editor and Shell too? Otherwise it can take another 4 months. Thanks.

Please let me know what you think.

andreinitescu:
@rmarinho 
> Seems we need a Default, or it will default to Never and break existing Apps .

Not sure I understand, the default of the property is set to Never already, which is the current native control  behavior (native controls don't display the close button by default on any platform):

https://github.com/xamarin/Xamarin.Forms/pull/5277/commits/e14a63c1c15733f223369b4b0fdd227518c2e793#diff-06ffb37c6df41841088dbdaf9466ba8dR42

Xamarin.Forms.Core/InputView.cs
```diff
@@ -38,4 +38,10 @@ public bool IsReadOnly
 			set { SetValue(IsReadOnlyProperty, value); }
 		}
 	}
+
+	public enum ClearButtonMode
```
StephaneDelcroix: Shouldn't the enum, and the property, be named `ClearButtonVisibility` ?

Xamarin.Forms.Platform.Android/Renderers/EntryRenderer.cs
```diff
@@ -61,7 +63,7 @@ protected override void UpdateTextColor(Color color)
 		}
 	}
 
-	public abstract class EntryRendererBase<TControl> : ViewRenderer<Entry, TControl>, ITextWatcher, TextView.IOnEditorActionListener
+	public abstract partial class EntryRendererBase<TControl> : ViewRenderer<Entry, TControl>, ITextWatcher, TextView.IOnEditorActionListener
```
StephaneDelcroix: why does this needs to be partial ?

StephaneDelcroix: ok, got it

Xamarin.Forms.Platform.Android/Renderers/EntryRenderer.cs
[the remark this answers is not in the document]
StephaneDelcroix: looks like a strange way of not doing `#region`s

Xamarin.Forms.Core/Entry.cs
```diff
@@ -106,7 +108,7 @@ public bool IsTextPredictionEnabled
 			get { return (bool)GetValue(IsTextPredictionEnabledProperty); }
 			set { SetValue(IsTextPredictionEnabledProperty, value); }
 		}
-    
+
```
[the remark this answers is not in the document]
andreinitescu: @StephaneDelcroix Most probably that comes from automatic document formatting. However, Visual Studio shows no difference to me:

image

Xamarin.Forms.Core/InputView.cs
andreinitescu: @StephaneDelcroix Some reasons I think worth considering:
1. "Visibility" usually makes me think of true/false. This works more like a 'mode'. It depends how you see it I guess.
2. It's easier for developers which are used with iOS API to instantly know what is it about: https://developer.apple.com/documentation/uikit/uitextfieldviewmode?language=objc
3. ClearButtonMode is shorter than ClearButtonVisibility.

andreinitescu: Obviously these are suggestions, I don't emotionally attach to this stuff usually, you (team) make the  last call

samhouts: Visibility makes more sense to me. We have many other uses of Visibility that are mapped to enum values, too.

andreinitescu: OK, I renamed to `CloseButtonVisibility`, see the commit

Xamarin.Forms.Core/Entry.cs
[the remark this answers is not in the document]
andreinitescu: @StephaneDelcroix OK, I spent some time on this to see where this came from. Someone added spaces when that property was implemented, when in fact all the code in there uses tabs. After I revered to that, Visual Studio actually tells me to refactor:

image

I wish the formatting becomes an error so people can't commit without unformatting the code.

PR #5283, opened 2019-02-16
[Android iOS Visual] fixes background in ActivityIndicator

### LINKED ISSUE #5261: [Visual] Activity Indicators aren't responding to height requests correctly
### Description

on iOS height is totally ignored and on Android it's drawing multiple circles

```XAML
            <Label Text="Custom Background Color" Margin="0,0,0,-10" />
                <ActivityIndicator HeightRequest="200"  BackgroundColor="{StaticResource SecondaryColor}" IsRunning="true" HorizontalOptions="FillAndExpand" />
                <ActivityIndicator HeightRequest="200"  BackgroundColor="{StaticResource SecondaryColor}" IsRunning="false" HorizontalOptions="FillAndExpand" />
            </StackLayout>

            <Label Text="Custom" Margin="0,0,0,-10" />
                <ActivityIndicator HeightRequest="200"  Color="{StaticResource PrimaryColor}" BackgroundColor="{StaticResource SecondaryColor}" IsRunning="true" HorizontalOptions="FillAndExpand" />
                <ActivityIndicator HeightRequest="200"  Color="{StaticResource PrimaryColor}" BackgroundColor="{StaticResource SecondaryColor}" IsRunning="false" HorizontalOptions="FillAndExpand" />
            </StackLayout>
```

paymicro:
### Description of Change ###

Fixed the back circle of the activation indicator when resizing. The `ThicknessRatio` of the activator indicator is tied to the width so that the width and height are the same, it is added to the square layout.
image

### Issues Resolved ### 

- fixes #5261

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

After
**Android**
screenshot_2

**iOS**
screenshot_3 |

### Testing Procedure ###

- run VisualGallery

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Material.iOS/MaterialActivityIndicatorRenderer.cs
+++ b/Xamarin.Forms.Material.iOS/MaterialActivityIndicatorRenderer.cs
@@ -1,3 +1,4 @@
+using System;
 using System.ComponentModel;
 using CoreAnimation;
 using CoreGraphics;
@@ -39,7 +40,7 @@ namespace Xamarin.Forms.Platform.iOS.Material
 
 					_backgroundLayer = new CAShapeLayer
 					{
-						LineWidth = 4,
+						LineWidth = Control.StrokeWidth,
 						FillColor = UIColor.Clear.CGColor,
 						Hidden = true
 					};
@@ -80,9 +81,22 @@ namespace Xamarin.Forms.Platform.iOS.Material
 			if (_center != Control.Center)
 			{
 				_center = Control.Center;
-				_backgroundLayer.Path = UIBezierPath.FromArc(_center, Control.Radius - 2, 0, 360, true).CGPath;
+				_backgroundLayer.LineWidth = Control.StrokeWidth;
+				var pathRadius = Control.Radius - Control.StrokeWidth / 2;
+				_backgroundLayer.Path = UIBezierPath.FromArc(_center, pathRadius, 0, 360, true).CGPath;
 			}
 			SetBackgroundColor(Element.BackgroundColor);
+		}
+
+		public override CGSize SizeThatFits(CGSize size)
+		{
+			var radius = NMath.Min(size.Width, size.Height) / 2;
+			if (!nfloat.IsInfinity(radius))
+			{
+				Control.Radius = radius;
+				Control.StrokeWidth = Control.Radius / 6;
+			}
+			return base.SizeThatFits(size);
 		}
 
 		protected override void OnElementPropertyChanged(object sender, PropertyChangedEventArgs e)
--- a/Xamarin.Forms.Material.iOS/MaterialActivityIndicatorRenderer.cs
+++ b/Xamarin.Forms.Material.iOS/MaterialActivityIndicatorRenderer.cs
@@ -12,11 +12,15 @@ namespace Xamarin.Forms.Platform.iOS.Material
 {
 	public class MaterialActivityIndicatorRenderer : ViewRenderer<ActivityIndicator, MActivityIndicator>
 	{
+		const float _minimumSize = 12;
+		const float _defaultRadius = 24;
+		const float _defaultStrokeWidth = 4;
+		const float _defaultSize = 2 * _defaultRadius + _defaultStrokeWidth;
+
 		SemanticColorScheme _defaultColorScheme;
 		SemanticColorScheme _colorScheme;
 
 		CAShapeLayer _backgroundLayer;
-		CGPoint _center;
 
 		public MaterialActivityIndicatorRenderer()
 		{
@@ -49,6 +53,7 @@ namespace Xamarin.Forms.Platform.iOS.Material
 
 				UpdateColor();
 				UpdateIsRunning();
+				SetBackgroundColor(Element.BackgroundColor);
 
 				ApplyTheme();
 			}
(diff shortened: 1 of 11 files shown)


Xamarin.Forms.Material.iOS/MaterialActivityIndicatorRenderer.cs
```diff
@@ -80,11 +81,24 @@ public override void LayoutSubviews()
...
 			SetBackgroundColor(Element.BackgroundColor);
 		}
 
+		public override CGSize SizeThatFits(CGSize size)
+		{
+			var radius = NMath.Min(size.Width, size.Height) / 2;
+			if (!nfloat.IsInfinity(radius))
+			{
+				Control.Radius = radius;
+				Control.StrokeWidth = Control.Radius / 6;
```
mattleibow: Maybe move this 6 into a constant so that it has a human name for future reference.

Xamarin.Forms.Platform.Android/Material/MaterialActivityIndicatorRenderer.cs
```diff
@@ -40,9 +40,16 @@ public MaterialActivityIndicatorRenderer(Context context)
...
-			_control = new AProgressBar(new ContextThemeWrapper(context, Resource.Style.XamarinFormsMaterialProgressBarCircular), null, Resource.Style.XamarinFormsMaterialProgressBarCircular);
-			_control.Indeterminate = true;
-			AddView(_control);
+			_control = new AProgressBar(new ContextThemeWrapper(context, Resource.Style.XamarinFormsMaterialProgressBarCircular), null, Resource.Style.XamarinFormsMaterialProgressBarCircular)
+			{
+				Indeterminate = true
+			};
+			var squareLayout = new SquareLayout(context)
+			{
+				LayoutParameters = new LayoutParams(LayoutParams.MatchParent, LayoutParams.MatchParent, GravityFlags.Center)
```
mattleibow: Maybe instead of making a new layout, make this renderer inherit from the square layout. I am not sure if that will work, but then we can further reduce nesting.
Or, instead of using AProgress directly here, make a SquareProgress control.

Xamarin.Forms.Material.iOS/MaterialActivityIndicatorRenderer.cs
paymicro: I'm not sure that this is needed, this is a computed constant only needed here.
By spec `StrokeWidth` is equal 4 with diameter is 48... For scaling `StrokeWidth` = diameter / 12 = radius / 6. As in Android.

Xamarin.Forms.Platform.Android/Material/MaterialActivityIndicatorRenderer.cs
paymicro: done.

Xamarin.Forms.Platform.Android/Renderers/CircularProgress.cs
```diff
@@ -0,0 +1,51 @@
...
+			base.Dispose(disposing);
+			if (disposing && !this.IsDisposed())
+				ViewTreeObserver.RemoveOnPreDrawListener(this);
+		}
+
+		protected override void OnMeasure(int widthMeasureSpec, int heightMeasureSpec)
+		{
+			base.OnMeasure(widthMeasureSpec, heightMeasureSpec);
+
+			if (IsInEditMode)
```
mattleibow: I have never used this before, but what does this do? If I think it does what it looks like, then this code only runs in the designer? Or is edit mode something different?

paymicro: Yes, it seems it was unnecessary. I've simplified this class, should work at least faster.

mattleibow:
Things look right, but for some reason the sizes don't appear to match... And the thickness. Seems weird since we are using the same values:

| Android | iOS |
| :-: | :-: |
|  |  |

Could it be that the ratios are differing from the outside/inside of the circle? Is there a padding?

I did some measuring at different sizes, and it appears that Android has a totalwidth/thickness scale of 9.7 and iOS has 11.5. And that after accounting for device scaling, iOS is almost perfect with just a few pixels out, but Android seems to have a scaling padding of a few pixels - when the width is 100dp, then the padding appears to be 8dp. As a result, the default size of iOS is about 50dp and Android is about 40dp.

I am not sure how much effort to put into this control, but maybe have a look and see if we can remove padding. The button had all sorts of issues until we removed the margin and then everything was "perfect". @PureWeen What do you think?

paymicro:
@mattleibow It looks like bug in Android `IndeterminateDrawable`. Its size is less than the doubled stroke width.
When i fixes background ring size: image 
I can not change the size `IndeterminateDrawable`. At least it's not as easy as I would like.

paymicro:
Thickness issue solved by changing the strike width in iOS. This deviates slightly from the specification, but the results are very close.
Also removed padding from Android.

PureWeen:
I think I'm still slightly seeing the issue @paymicro  referred to 

### Android API 23 Nexus 5 device

```
new ActivityIndicator(){BackgroundColor = Color.Blue,IsRunning = true, Color = Color.Red,Visual = VisualMarker.Material},
```

image

PR #5289, opened 2019-02-18
[visual] Add extra parameter to android visual constructors

PureWeen:
### Description of Change ###
Material Android renderers need to know the style they are being set to on creation so they can be set to the correct style id.  The style on an android element isn't settable at runtime post inception.

The idea is being able to do this via the constructor
```C#
static int getThemeId(BindableObject element)
		{
			var style = Xamarin.Forms.Material.Button.GetStyle(element);
			int themeId = Resource.Style.XamarinFormsMaterialTheme;
			switch (style)
			{
				case Xamarin.Forms.Material.Style.Outline:
					themeId = Resource.Style.XamarinFormsMaterialOutlinedButtonTheme;
					break;
				case Xamarin.Forms.Material.Style.Text:
					themeId = Resource.Style.XamarinFormsMaterialTextButtonTheme;
					break;
			}

			return themeId;
		}

		public MaterialButtonRenderer(Context context, BindableObject element)
			: base(new ContextThemeWrapper(context, getThemeId(element)))
		{
			VisualElement.VerifyVisualFlagEnabled();
```
 
### Platforms Affected ### 

- Core/XAML (all platforms
- Android

If a constructor isn't found with two arguments it will just default back to the previous behavior which just uses the constructor that takes a context

### Testing Procedure ###
Unit Test Added

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/DependencyResolver.cs
+++ b/Xamarin.Forms.Core/DependencyResolver.cs
@@ -33,7 +33,9 @@ namespace Xamarin.Forms.Internals
 			return result;
 		}
 
-		internal static object ResolveOrCreate(Type type, params object[] args)
+		internal static object ResolveOrCreate(Type type) => ResolveOrCreate(type, null, null);
+
+		internal static object ResolveOrCreate(Type type, object source, Type visualType, params object[] args)
 		{
 			var result = Resolve(type, args);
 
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Core/DependencyResolver.cs
```diff
@@ -33,14 +33,20 @@ internal static object Resolve(Type type, params object[] args)
...
+		internal static object ResolveOrCreate(Type type, object source, Type visualType, params object[] args)
 		{
 			var result = Resolve(type, args);
 
 			if (result != null) return result;
 
 			if (args.Length > 0)
 			{
+				if(visualType != typeof(VisualRendererMarker.Default))
+					if (type.GetTypeInfo().DeclaredConstructors.Any(info => info.GetParameters().Length == 2))
```
hartez: Won't we end up at this line if the visualType parameter is `null`?

PR #5293, opened 2019-02-18
[WIP] A "demo" for a custom Visual implementation

mattleibow:
**This is not a thing. This is not a start of a thing.**

This is just a test to make sure the bits are in place when a third party wants to create a custom Visual implementation.

So far, everything seems to work. The only issue is that the visual can't be selected in XAML as it falls back to the default.


Xamarin.Forms.Sandbox.iOS/SkiaSharpButtonRenderer.Shared.cs
```diff
@@ -0,0 +1,166 @@
...
+
+#if __ANDROID__
+using SkiaSharp.Views.Android;
+using SKPaintSurfaceEventArgs = SkiaSharp.Views.Android.SKPaintSurfaceEventArgs;
+#elif __IOS__
+using SkiaSharp.Views.iOS;
+using SKPaintSurfaceEventArgs = SkiaSharp.Views.iOS.SKPaintSurfaceEventArgs;
+#endif
+
+[assembly: ExportRenderer(typeof(Button), typeof(SkiaSharpVisual.SkiaSharpButtonRenderer), new[] { typeof(SkiaSharpVisual.SkiaSharp) })]
```
mattleibow: It appears that the only extra thing to use visual is the last parameter:
```csharp
new[] { typeof(SkiaSharpVisual.SkiaSharp) }
```

Which is how this was all designed. So, everything is working as intended.

Xamarin.Forms.Sandbox/Visual.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Xamarin.Forms;
+
+namespace SkiaSharpVisual
+{
+	public sealed class SkiaSharp : IVisual
+	{
+		public static SkiaSharp Instance { get; } = new SkiaSharp();
```
mattleibow: I am not sure how we would want third parties to set up their visual. We might want to have a look at creating a patter to keep everyone consistent. But, this appears to work - even though I am creating a conflict with the `SkiaSharp` namespace...

PR #5298, opened 2019-02-18
Update BottomNavigationViewUtils.cs: removing redundant Dispose() calls

PawKanarek:
We should remove Dispose() and using statement from this util as we don't destroy items here, we are only updating their properties, that's wrong place to dispose items. 

Casting objects to classes that implements IDisposable interface dosen't mean we need call Dispose() on them, and surely we don't want call it here.

### Description of Change ###

Removed redundant using statement and Dispose() call

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

None

Not applicable


rmarinho:
I actually found some similar issues with Shell, where using was not the right way to do it.

rmarinho:
We have a failing tests on android I think is related, can you double check it works @PawKanarek ? thanks

Xamarin.Forms.Controls.Issues.RestartAppTest.ForcingRestartDoesNotCauseCrash_google_pixel_2-9

PawKanarek:
@rmarinho  Firstly:  i don't have access to details of failed tests because im not authorized.
"TF400813: The user 'Windows Live ID\mymail' is not authorized to access this resource."
 
Secondly: I don't know why there is any problem with LegacyRenderes/FastRenderers tests. As far as i know, my change was only affecting SHELL navigation because that's only place where this util was used. And for sake, this is a 100% good practice to not dispose objects when we are only casting to IDisposable interface, correct me if im wrong.

Can you authorize my Microsoft account, or paste details of failed tests? 

Regards!

Issue #5300, opened 2019-02-18
Referencing images which don't exist causes Forms to crash


alanmcgovern:
Forms should be able to gracefully ignore missing images when doing design time rendering. Missing images could be as simple as a typo, or just an incomplete string.

In Android the null reference is because the 4 fields populated from the `ResourceManager.Init` call are null when in no-compile-rendering mode. We'd like this scenario to result in a no-op, instead of a crash, for now. We'll file a separate piece of work to make this scenario work as expected.

For iOS we already support loading images appropriately as forms uses NSBundle methods to locate the image, and we hook into those and provide the data. Therefore all we want iOS to do is not crash when the image doesn't exist.

We should review the rest of the codebase to ensure other places with similar behaviours are updated.

Sample xaml:
```
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
			 xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
			 x:Class="TestFormsNoCompile2341.Views.ItemDetailPage" 
			 Title="{Binding Title}"
			 BackgroundImage="tab_aabout.png">
```

Android trace:
```
System.NullReferenceException: Object reference not set to an instance of an object
Xamarin.Forms.Platform.Android.ResourceManager.GetId (System.Type type, System.String memberName)
Xamarin.Forms.Platform.Android.ResourceManager.IdFromTitle (System.String title, System.Type type)
Xamarin.Forms.Platform.Android.ResourceManager.GetDrawable (Android.Content.Context context, System.String name)
Xamarin.Forms.Platform.Android.PageRenderer.UpdateBackgroundImage (Xamarin.Forms.Page view)
Xamarin.Forms.Platform.Android.PageRenderer.OnElementChanged (Xamarin.Forms.Platform.Android.ElementChangedEventArgs`1[TElement] e)
Xamarin.Forms.Platform.Android.VisualElementRenderer`1[TElement].SetElement (TElement element)
Xamarin.Forms.Platform.Android.VisualElementRenderer`1[TElement].Xamarin.Forms.Platform.Android.IVisualElementRenderer.SetElement (Xamarin.Forms.VisualElement element)
Xamarin.Forms.Platform.Android.Platform.CreateRenderer (Xamarin.Forms.VisualElement element, Android.Content.Context context)
Xamarin.Forms.Platform.Android.Platform.CreateRenderer (Xamarin.Forms.VisualElement element)
<unknown method>
[...]

PR #5305, opened 2019-02-19
Fix NRE in Android ResourceManager when reference resource in Previewer

TimBarham:
### Description of Change ###

When running in the Forms Previewer, `Xamarin.Forms.Platform.Android.ResourceManager.Init()` won't find the expected types if the project has not been built, and `GetId()` will fail with an NRE.

Add a null check to `GetId()` and return 0.

### Issues Resolved ### 

Fixes part of #5300 

### API Changes ###
 
None

### Platforms Affected ### 

- Android

None

A sample app sets `ContentPage.BackgroundImage` in a `Style` in `App.xaml`. If the project is not built, currently every page fails to render in the Previewer (Android) as follows:

image

With this fix in place, all pages render as expected, for example...

image

### Testing Procedure ###

Verified Forms XAML pages that failed to render in Previewer now render successfully.

### PR Checklist ###

- [ ] Has automated tests - we will add tests on the previewer side to cover this scenario

### DIFF
--- a/Xamarin.Forms.Platform.Android/ResourceManager.cs
+++ b/Xamarin.Forms.Platform.Android/ResourceManager.cs
@@ -138,6 +138,10 @@ namespace Xamarin.Forms.Platform.Android
 
 		static int GetId(Type type, string memberName)
 		{
+			// This may legitimately be null in designer scenarios
+			if (type == null)
+				return 0;
+
 			object value = null;
 			var fields = type.GetFields();
 			for (int i = 0; i < fields.Length; i++)


TimBarham:

TimBarham:
@davidortinau - this is the change we need to allow pages in the _ConferenceVision_ app (and presumably many apps) to render in Android without building.

PureWeen:
build --uitests

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -138,6 +138,10 @@ internal static int IdFromTitle(string title, Type type)
 
 		static int GetId(Type type, string memberName)
 		{
+			// This may legitimately be null in designer scenarios
+			if (type == null)
```
PureWeen: The behavior of the previewer when the state of the project is not compiled is to not display any images? Users will still need to compile that app in order for drawables to show up? Just validating the expectations of the viewer when app is in a non compiled state

TimBarham: At this stage, yes, we just want to fix the NRE as it will break many common scenarios, and it is too close to shipping Dev 16 to do more.

Post 16.0 , however, we'd like to investigate options to actually display the images.

PureWeen:
I'm wondering if it might make sense to just set all the types to something so all the cases of NRE are covered and all cases will just fail silently.  

```C#
public static void Init(Assembly masterAssembly)
{
	DrawableClass = masterAssembly.GetTypes().FirstOrDefault(x => x.Name == "Drawable" || x.Name == "Resource_Drawable");
	ResourceClass = masterAssembly.GetTypes().FirstOrDefault(x => x.Name == "Id" || x.Name == "Resource_Id");
	StyleClass = masterAssembly.GetTypes().FirstOrDefault(x => x.Name == "Style" || x.Name == "Resource_Style");
	LayoutClass = masterAssembly.GetTypes().FirstOrDefault(x => x.Name == "Layout" || x.Name == "Resource_Layout");

        // set classes to something so designer doesn't crash with an NRE 
	DrawableClass = DrawableClass ?? typeof(object);
	ResourceClass = ResourceClass ?? typeof(object);
	StyleClass = StyleClass ?? typeof(object);
	LayoutClass = LayoutClass ?? typeof(object);
}
```

alanmcgovern:
https://github.com/xamarin/Xamarin.Forms/issues/5300

This is the corresponding bug

TimBarham:
@PureWeen - it appears it does get called, as your suggested change also fixes the NRE. Of course, I'd be happier if we included _both_ changes 😄

Xamarin.Forms.Platform.Android/ResourceManager.cs
alanmcgovern: Yeah - the expectation is that images should (and will) just work. That is a separate piece of work, as Tim said. We're not even looking at it currently :)

TimBarham:
Are the test failures significant?

rmarinho:
@TimBarham no that's ok

PR #5307, opened 2019-02-19
Use a better way of finding the Android root assembly

alanmcgovern:
If you want a foolproof way of getting a reference to the assembly from
the Android/iOS/whatever project, then just use the EntryAssembly.

### DIFF
--- a/Xamarin.Forms.Platform.Android/Forms.cs
+++ b/Xamarin.Forms.Platform.Android/Forms.cs
@@ -70,7 +70,7 @@ namespace Xamarin.Forms
 		// Why is bundle a param if never used?
 		public static void Init(Context activity, Bundle bundle)
 		{
-			Assembly resourceAssembly = Assembly.GetCallingAssembly();
+			Assembly resourceAssembly = Assembly.GetEntryAssembly ();
 			SetupInit(activity, resourceAssembly);
 		}
 


Xamarin.Forms.Platform.Android/Forms.cs
```diff
@@ -70,7 +70,7 @@ public static Color GetColorButtonNormal(Context context)
 		// Why is bundle a param if never used?
 		public static void Init(Context activity, Bundle bundle)
 		{
-			Assembly resourceAssembly = Assembly.GetCallingAssembly();
+			Assembly resourceAssembly = Assembly.GetEntryAssembly ();
```
PureWeen: Does this work? I tested this yesterday briefly when trying to find a better approach and in my tests this returns null

image

PureWeen: https://bugzilla.xamarin.com/show_bug.cgi?id=21097

```
Assembly.GetEntryAssembly() is supposed to return the _executable_ (.exe file) that was used to launch the app. Xamarin.Android apps don't contain .exe files; they consist of .dll files.

http://msdn.microsoft.com/en-us/library/system.reflection.assembly.getentryassembly(v=vs.110).aspx
> Gets the process executable in the default application domain.

Consequently, Xamarin.Android apps don't have an entry assembly, and Assembly.GetEntryAssembly() returns `null`.
```

alanmcgovern: 👍 

That's kinda surprising. I assumed whichever assembly contained the `Main` method would've been returned :)

PR #5309, opened 2019-02-19
[XamlC] Identify generic instance types correctly when importing ctor

akihikodaki:
### Description of Change ###

The old implementation fails to identify generic instance types for typed binding getter, whose type is:
System.Func`2<TSource, ValueTuple<TProperty, bool>>

This fixes the issue by taking generic arguments into account.

### Issues Resolved ### 

None

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###

Run the automated test.

### PR Checklist ###

- [X] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Build.Tasks/ModuleDefinitionExtensions.cs
+++ b/Xamarin.Forms.Build.Tasks/ModuleDefinitionExtensions.cs
@@ -47,7 +47,7 @@ namespace Xamarin.Forms.Build.Tasks
 
 		public static MethodReference ImportCtorReference(this ModuleDefinition module, TypeReference type, TypeReference[] parameterTypes)
 		{
-			var ctorKey = $"{type}.ctor({(parameterTypes == null ? "" : string.Join(",", parameterTypes.Select(tr => (tr.Scope.Name, tr.Namespace, tr.Name))))})";
+			var ctorKey = $"{type}.ctor({(parameterTypes == null ? "" : string.Join(",", parameterTypes.Select(SerializeTypeReference)))})";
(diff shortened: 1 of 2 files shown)


StephaneDelcroix:
@akihikodaki one of these days you'll have to explain me what you exactly do to find those dark spots in the Xaml Compiler :)

akihikodaki:
Someday I may write an article describing the details, but here I briefly explain it. There are three common patterns when I find a bug in XamlC:
1. Using the latest features in a real world application
I don't think typed binding and even XamlC are so popular today. (Who cares `x:DataType` and `XamlCompilationAttribute`? An application will work even without them.) That may let several bugs which you are likely to encounter when writing an application remain hidden.
2. Writing my own components
I write and use a wrapper of Unity's UI system with Xamarin.Forms. Some components are generic or value types, and revealed bugs around type consistency.
3. Reading the source code
For example, prior to #4238, you could `grep` `ldloc` and see many inconsistencies in type conversion. You may still find similar problems if you search for instructions such as `call` / `callvirt`. Type conversion will be a redundant operation in this kind of code generator, and should be de-duplicated by creating pairs of stack-pushing instructions and stack-popping instructions, which will eventually create a tree.

StephaneDelcroix:
@akihikodaki thanks. When we introduced XamlC and compiled bindings, we knew there was some uncovered cases, especially for value types and generics, but they usually don't show up in common applications. I'm glad your use cases are uncommon and you're reporting and fixing those dark holes

PR #4477, opened 2018-11-19
[Core] Move TextProperty, ITextElement, and IPlaceholderElement to InputView

jcmanke:
### Description of Change ###

The Entry, Editor, and SearchBar classes are all subclasses of InputView, but each separately implements the interfaces ITextElement and IPlaceholderElement and a Text property. This PR moves these properties into InputView while maintaining ABI backwards compatibility.

### Issues Resolved ### 

- N/A

### API Changes ###

Changed:
- Moved `Text` property, `TextChanged` event, `OnTextChanged` method, `IPlaceholderElement` implementation, `ITextElement` implementation from `Entry`, `Editor`, and `SearchBar` to `InputView`

### Platforms Affected ### 

- Core/XAML (all platforms)

### Behavioral/Visual Changes ###

None

### Before/After Screenshots ### 

N/A

### Testing Procedure ###

N/A

### PR Checklist ###

- [ ] Has automated tests 
- [x] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard

PR #3248, opened 2018-07-09
Change SearchBar base class to InputView

jcmanke:
### Description of Change ###

Change the base class of SearchBar from View to InputView to give it Keyboard, MaxLength, and IsSpellCheckEnabled properties.

### Issues Resolved ###

- fixes #2902

### API Changes ###

Added:
 - Android SearchBarRenderer
   - void UpdateMaxLength()
   - void UpdateInputType()
   - protected virtual NumberKeyListener  GetDigitsKeyListener(InputTypes inputTypes)
 - iOS SearchBarRenderer
   - void UpdateMaxLength()
   - void UpdateKeyboard()
   - bool ShouldChangeText(UISearchBar searchBar, NSRange range, string text) 
 - UWP SearchBarRenderer
   - void UpdateMaxLength()
   - void UpdateInputScope()

Changed:
 - Core: public class SearchBar : View --> public class SearchBar : InputView
 - iOS: public static void Extensions.ApplyKeyboard(this IUITextInput textInput, Keyboard keyboard) --> public static void Extensions.ApplyKeyboard(this IUITextInputTraits textInput, Keyboard keyboard)

### Platforms Affected ###

- Core/XAML (all platforms)
- iOS
- Android
- UWP

### Behavioral/Visual Changes ###

Potential conflicts between UWP platform specific for SearchBar.IsSpellCheckEnabled and inherited InputView.IsSpellCheckEnabled

### PR Checklist ###

- [x] Has automated tests
- [x] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard

PR #5320, opened 2019-02-20
Added ReturnType to Editor

### LINKED ISSUE #4832: Editor crashes when my app was updated to latest xamarin.forms 3.4 .0.1008975
### Description

My App crashes by using Editor, when I updated xamarin.forms from 3.2 to3.4 .

It crashes only when  my smartphone's orientation is landscape.  

It happens only by Andriod. There is no such a  bug by iOS.

### Steps to Reproduce

1. Andriod, uses portable project, one of tagets is .NET Framework 4.5
2. smartphone's orientation is landscape ( portrait is ok)
3. Input some text for Editor then press "DONE"
4. Crashes:

Java.Lang.IllegalStateException: focus search returned a view that wasn't able to take focus!

### Basic Information

- Version with issue:
- Last known good version:
- IDE: 
Visual Studio 2017
- Platform Target Frameworks:  
   
  - Android: <  7.1 -->

andreinitescu:
### Description of Change ###

Added support for ReturnType property to Editor. Moved ReturnType from Entry to base class InputView.
It also fixes several reported issues.

Probably ReturnCommand property should also be moved to InputView, something which will be handled in a future ticket.

### API Changes ###

- Core/XAML (all platforms)
- iOS
- Android
- UWP
- Tizen

None

Not applicable

### PR Checklist ###

- [x] Has automated tests 


adrianknight89:
@andreinitescu Regarding the future ticket, ideally I'd like to see `ReturnCommand` moved into `InputView` at the same time as `ReturnType`. Would it be too much work to do that in the same PR or do you expect to submit another PR for this soon? Thanks for the work.

Off-topic: It'd be neat to have a keyboard type for posting long text for social apps which includes the `@` and `#` symbols. iOS has a built-in [keyboard type](https://stackoverflow.com/questions/11751673/how-does-instagram-customize-the-uikeyboard-returnkey-to-an-and-a) for this. Not sure about Android though.

jcmanke:
@StephaneDelcroix So you've decided moving public properties and fields to the base class doesn't constitute an unacceptable ABI break? (In relation to #4477)

jcmanke:
Another thing, #3248 made `SearchBar` another subclass of `InputView`, do we really want `ReturnType` for that?

adammeaney:
Why would we want to have a return type on an Editor?

I am just asking for the use case for this? I can't imagine wanting my editors to ever do things besides add a new line in the editor or close it, potentially. Any other thing would just be weird for an editor.

StephaneDelcroix:
@jcmanke moving a public property should be fine, as property accessors are invoked using `callvirt`. moving the actual BP (the field) is a breaking change. thanks for your sharps eyes.

jcmanke:
@rmarinho The ImeAction is shown on Android when you have the phone in landscape.

Screenshot_1553180558

samhouts:
I've seen a couple of comments here asking for a use case for ReturnType on Editor/SearchBar. Do we have an answer for that? Thanks!

andreinitescu:
My initial goal was to fix crash issue, but since I was on it, I wanted to do a bit extra and surface the native feature to allow setting the return key for Editor just like Entry has it. No good deed goes unpunished 😆

samhouts:
@andreinitescu Thank you for the work! Let's table this one for now, since we don't have a clear need for it. We appreciate the time you spent here!!!

PR #5331, opened 2019-02-21
Fix NRE in Android ResourceManager when reference resource in Previewer

TimBarham:
Backport from fix in 3.6.0

(cherry picked from commit fb59591580ea474e7f48e5941854fd790dc5a8cb)


TimBarham:
@alanmcgovern FYI

PR #5332, opened 2019-02-21
[Material] rename iOS assembly to be non platform specific

PureWeen:
### Description of Change ###
It doesn't make any sense to have platform specific dlls
If we create Material dlls for netstandard or android then we should just bait and switch them not create multiple dlls

### Platforms Affected ### 
- iOS


PureWeen:
Downloaded and tested the nuget for this over here
https://github.com/PureWeen/visualtesting

it's also checked into that project

PR #5338, opened 2019-02-21
Fix for Issue 3886, UWP Master Detail, incorrect master height after window resize

### LINKED ISSUE #3886: UWP Master Detail Page Height Incorrect After Resizing App Window
### Description
If you resize the app window height, the master detail page will incorrectly display the first time.

### Steps to Reproduce

0453a16e-fdbd-4fd7-8fd6-cb8d4a653f7e

1. Start a UWP master detail page template app
2. Open hamburger menu
3. Resize window height
3. Open hamburger menu
4. Notice the incorrect height
5. Close hamburger menu
6. Open hamburger menu
6. Height is now correct

### Expected Behavior

After resizing the window I would expect the master detail page to have the correct height.

### Actual Behavior

After resizing the window the master detail page's height is incorrect until closing and re-opening.

### Basic Information

- Version with issue: <= 3.2.0.839982
- Last known good version: Never/Unknown, I tested all the way back to XF 2.5.1 and all versions had the same issue.
- IDE: VS 15.8.4
- Platform Target Frameworks:
  - UWP:  16299

### Reproduction Link

https://drive.google.com/open?id=1sjSZEB7eC3rLAOyG9plHbRNK469Fjnbd

bmacombe:
### Description of Change ###
Modifies the code below by removing the first assignment of height to the master presenters actual height.  When the MasterDetailControl is in the popover mode, the master presenter actual height is always the previous size on the presenter, which causes the master pane to be sized incorrectly if the window size has changed since it's last display.

// Use the ActualHeight of the _masterPresenter to automatically adjust for the Master Title
double height = _masterPresenter?.ActualHeight ?? 0;

// If there's no content, use the height of the control to make sure the background color expands.
if (height == 0)
height = ActualHeight;

### Issues Resolved ### 
- fixes #3886 

### API Changes ###
None

### Platforms Affected ### 
- UWP

## Before
before

## After
after

### PR Checklist ###

- [ ] Has automated tests 


bmacombe:
@samhouts Digging into the history of that file, the behavior I'm changing was made by a commit of yours in PR #983 referencing a [Bugzilla issue 54036](https://bugzilla.xamarin.com/show_bug.cgi?id=54036).

I don't believe the change I made would affect what you were trying to resolve in #983 but would appreciate your review and thoughts.

Thanks!

bmacombe:
@paymicro  Thanks, I'll look into the padding to see what is up. Just make sure I understand you correctly on the to much code has changed.  You want me to remove the test page I created and just update 1461 one?

paymicro:
@bmacombe Yes exactly. I think that a small visual change in test 1461 is enough to verify this fix.

bmacombe:
Looked into the padding issue.  This change didn't break it, it just didn't fully resolve the issue.  The padding is correctly show on the second opening on the popover.  So the change I made partially resolved the issue.  I'll continue to look into this.

bmacombe:
The real issue appears to be with MasterSize being called prior to the master being displayed.

It seems what is "first load" is different between the various Master Behaviors.  The current method of setting _firstLoad = false when MasterSize is first called works correctly for all modes but Popover.  Looking at ways to alter when _firstLoad is set.

Xamarin.Forms.Platform.UAP/MasterDetailControl.cs
```diff
@@ -161,18 +161,12 @@ public Windows.Foundation.Size MasterSize
...
 			get
 			{
-				// Use the ActualHeight of the _masterPresenter to automatically adjust for the Master Title
-				double height = _masterPresenter?.ActualHeight ?? 0;
-
-				// If there's no content, use the height of the control to make sure the background color expands.
-				if (height == 0)
-					height = ActualHeight;
-
+				double height = ActualHeight;
```
bmacombe: So I kept the original change of always starting with the actual height vs the last master height

Xamarin.Forms.Platform.UAP/MasterDetailControl.cs
```diff
@@ -161,18 +161,12 @@ public Windows.Foundation.Size MasterSize
...
 				double width = 0;
 
 				// On first load, the _commandBar will still occupy space by the time this is called.
 				// Check ShouldShowToolbar to make sure the _commandBar will still be there on render.
-				if (_firstLoad && _commandBar != null && ShouldShowToolbar)
+				if (_firstLoad || !ShouldShowSplitMode && _commandBar != null && ShouldShowToolbar)
```
bmacombe: And then added this so that when the control is in popover mode it will always subtract the command bar height but if we are in split view it is still only applied on the initial load.

bmacombe:
@paymicro would you please review again?  I updated issue 1461 as you requested and believe I have the issue corrected now.  I left a couple of notes to explain the changes.

Thanks

Xamarin.Forms.Platform.UAP/MasterDetailControl.cs
paymicro:
```suggestion
				if ((_firstLoad || !ShouldShowSplitMode) && _commandBar != null && ShouldShowToolbar)
```

PR #5342, opened 2019-02-21
previewer; PageRenderer trap for missing bg image

kingces95:
Previewer team opened https://github.com/xamarin/Xamarin.Forms/issues/5300 which asks we scrub for all places where we throw if an image is not found. So, these change simply ignore missing images.

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Extensions/ColorExtensions.cs
+++ b/Xamarin.Forms.Platform.iOS/Extensions/ColorExtensions.cs
@@ -32,6 +32,15 @@ namespace Xamarin.Forms.Platform.MacOS
 #endif
 		}
 
+		public static UIColor FromPatternImageFromBundle(string bgImage)
+		{
+			var image = UIImage.FromBundle(bgImage);
+			if (image == null)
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Platform.iOS/Extensions/ColorExtensions.cs
```diff
@@ -32,6 +32,15 @@ public static CGColor ToCGColor(this Color color)
 #endif
 		}
 
+		public static UIColor FromPatternImageFromBundle(string bgImage)
+		{
+			var image = UIImage.FromBundle(bgImage);
```
samhouts: Doesn't build on MacOS. Need to make an alias for the MacOS type. `NSImage`, I presume?

```  
/Users/vsts/agent/2.147.1/work/1/s/Xamarin.Forms.Platform.iOS/Extensions/ColorExtensions.cs(37,16): error CS0103: The name 'UIImage' does not exist in the current context 
```

kingces95:
Oh, I'll fix that alright!

PR #5345, opened 2019-02-22
[Visual] Material Editor

### LINKED ISSUE #5009: [Material] Filled Editor Spec
### Description
This is just a copy of #4949 except for the editor control which uses the same native control. Once #4949 has been completed and merged then Editor can be worked on

https://material.io/design/components/text-fields.html#filled-text-field

image

### Expectations 

- **Visually Identical**

- **States**
  Make sure these states are the same between ios/android and are true to the material guidelines (Inactive, active, focused, Activated, Hover, Disabled). Error messages will be added at a later date. Currently the color of underline and placeholder just uses the Material theme defaults. The material style use a predefined Alpha when the control is in a given state so this is currently what we hard code.  We will expose this through an attached property at a later date. The renderers are constructed in such a way that the user can override the styling if they want to customize these color states

# Scenarios
```XAML
<Editor Visual="Material" />
```

```C#
new Editor() { Visual = VisualMarker.Material }
```

PureWeen:
### Description of Change ###

Material Editor

### Issues Resolved ### 

- fixes #5009 

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android

### Testing Procedure ###
- Check Editors on Visual Gallery

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Material.iOS/MaterialEditorRenderer.cs
+++ b/Xamarin.Forms.Material.iOS/MaterialEditorRenderer.cs
@@ -1,9 +1,7 @@
-﻿using CoreGraphics;
-using UIKit;
+﻿using UIKit;
 using MaterialComponents;
-using System.Threading.Tasks;
-using Foundation;
 using System;
+using CoreGraphics;
 
 namespace Xamarin.Forms.Platform.iOS.Material
 {
@@ -72,9 +70,22 @@ namespace Xamarin.Forms.Platform.iOS.Material
 		protected internal override void UpdateAutoSizeOption()
 		{
 			base.UpdateAutoSizeOption();
-			Control.ExpandsOnOverflow = Element.AutoSize == EditorAutoSizeOption.TextChanges;
+			Control.AutoSizeWithChanges = Element.AutoSize == EditorAutoSizeOption.TextChanges;
+
+			if(!Control.ExpandsOnOverflow)
+				Control.ExpandsOnOverflow = Element.AutoSize == EditorAutoSizeOption.TextChanges;
 		}
 
+
+		public override CGRect Frame
+		{
+			get => base.Frame; 
+			set
+			{
+				base.Frame = value;
+
+			}
+		}
 
 		// this is required to force the placeholder to size correctly if it starts out prefilled
 		void InitialPlaceholderSetupHack()
--- a/Xamarin.Forms.Material.iOS/MaterialEditorRenderer.cs
+++ b/Xamarin.Forms.Material.iOS/MaterialEditorRenderer.cs
@@ -1,25 +1,25 @@
 ﻿using CoreGraphics;
 using UIKit;
+using MaterialComponents;
 
 namespace Xamarin.Forms.Platform.iOS.Material
 {
-	public class MaterialEntryRenderer : EntryRendererBase<MaterialTextField>, IMaterialEntryRenderer
+	public class MaterialEditorRenderer : EditorRendererBase<MaterialMultilineTextField>, IMaterialEntryRenderer
 	{
-
-		public MaterialEntryRenderer()
+		public MaterialEditorRenderer()
 		{
 			VisualElement.VerifyVisualFlagEnabled();
+			ClipsToBounds = true;
 		}
 
-		protected override MaterialTextField CreateNativeControl()
-		{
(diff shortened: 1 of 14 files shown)


PureWeen:
@paymicro  good finds. 

iOS should be fixed now

Xamarin.Forms.Material.iOS/MaterialEditorRenderer.cs
```diff
@@ -0,0 +1,116 @@
...
+
+		protected internal override void UpdateAutoSizeOption()
+		{
+			base.UpdateAutoSizeOption();
+			Control.ExpandsOnOverflow = Element.AutoSize == EditorAutoSizeOption.TextChanges;
+		}
+
+
+		// this is required to force the placeholder to size correctly if it starts out prefilled
+		void InitialPlaceholderSetupHack()
```
PureWeen: If you set the placeholder and text of the control before focused the placeholder floats to the top left but doesn't resize. The following code was the only way I could figure out how to make it work. Once it's fixed upstream we can remove this code.

mattleibow: Has this been fixed? We are a few versions behind.

PureWeen: Maybe? I found this https://github.com/material-components/material-components-ios/commit/99871be6cd2ca9b365d99d46a6c28a93ac22e802

Once we get 3.6.0 out we can get updated to latest material and then go from there

Xamarin.Forms.Material.iOS/MaterialMultiLineTextField.cs
```diff
@@ -0,0 +1,108 @@
...
+			else
+				_contentSize = TextView.ContentSize;
+
+			if (!expandTurnedBackOn)
+				UpdateExpandsOnOverflow();
+
+			return result;
+		}
+
+		bool NumberOfLinesCheck()
```
paymicro: The method name does not match the content, instead of just checking it changes `ExpandsOnOverflow` variable.

Xamarin.Forms.Material.iOS/MaterialMultiLineTextField.cs
```diff
@@ -0,0 +1,108 @@
...
+				return false;
+
+			return (((NumberOfLines + 1) * TextView.Font.LineHeight) > Frame.Height);
+		}
+
+		void UpdateExpandsOnOverflow()
+		{
+			if (!NumberOfLinesCheck() && !AutoSizeWithChanges && ExpandsOnOverflow && Frame.Height > 0 && TextView?.Font != null) 
+			{
+				if (shouldRestrainSize())
```
paymicro: I think it is worth to simplify

paymicro: I think it is worth to simplify checks
```C#
if (ExpandsOnOverflow && !AutoSizeWithChanges && ShouldRestrainSize())
ExpandsOnOverflow = false;
```

paymicro: I think it is worth to simplify

Xamarin.Forms.Material.iOS/MaterialMultiLineTextField.cs
```diff
@@ -0,0 +1,108 @@
...
+				{
+					ExpandsOnOverflow = false;
+				}
+			}
+		}
+
+		public override CGRect Frame
+		{
+			get => base.Frame; set
```
paymicro: formating

Xamarin.Forms.Material.iOS/MaterialMultiLineTextField.cs
```diff
@@ -0,0 +1,108 @@
...
+				base.Frame = value;
+				UpdateExpandsOnOverflow();
+			}
+		}
+
+		int NumberOfLines
+		{
+			get
+			{
+				if (TextView?.ContentSize == null || TextView?.Font == null || TextView.Font.LineHeight == 0)
```
paymicro:
```suggestion
				if (TextView?.ContentSize == null || TextView.Font == null || TextView.Font.LineHeight == 0)
```

PR #5349, opened 2019-02-22
RadioButton

### LINKED ISSUE #2404: [F100] RadioButton
## Implementation

```C#
public class RadioButton : IToggleButtonElement, ToggleButtonElement, ...
{
      EventHandler<CheckedChangedEventArgs> CheckedChanged;
      bool IsChecked {get; set; }  //BP
      string GroupName { get; set; } //BP
      Color Color //BP
	public Color BorderColor //BP
	public int CornerRadius //BP
	public double BorderWidth //BP
	public ICommand Command //BP
	public object CommandParameter //BP
	public Font Font //BP
	public string Text //BP
	public Color TextColor //BP
}
```
## Rationale

The only way to achieve the `RadioButton` behavior in Xamarin.Forms out the box is by using a drop-down. Otherwise we need to use a set of `Switch`es having to enable and disable them on demand.
The problem gets even more complicated when using MVVM.
We have to generate data items for every element on the list with a bindable/notifiable property indicating state of current item.

The reason a drop-down is not enough, is because in many apps, we need the users to be able to make up their choice with a single sight and action, especially for small quantity list drop-downs (e.g. Male/Female/Unknown). A drop-down requires _a_. clicking the drop-down to open it, _b_. browsing the available options, _c_. selecting the desired item; whereas `RadioButton` is not only achieved with a single click, but the users can conclude what they are gonna select before even approaching the field. Furthermore, for speed-critical app (i.e. medic or customer-service related) when UX is top priority, a `RadioButton` is a must so the users already know where the choice they'll select is located, so they can click its position without even reading its content each time. All that is impossible to achieve with a traditional drop-down. Regarding this point, `Switch` is also not always the preferred option because for most platforms it takes up x3 the space a `RadioButton` does.

#### Android
https://developer.android.com/guide/topics/ui/controls/radiobutton.html

#### iOS
Implementation is fairly simple with setting the background images according the current state.

#### UWP
https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/radio-button

### Implications for CSS

```css
radiobutton {
}
```

### Backward Compatibility

### Difficulty : Medium
Not much work for Android and UWP but the iOS implementation still needs to be fleshed out a bit, and there are numerous working samples in XLabs etc. to start from.

andreinitescu:
### Description of Change ###

RadioButton implementation.  

### Issues Resolved ### 

- fixes #2404 

### API Changes ###

Added:

```
public class RadioButton : IToggleButtonElement, ToggleButtonElement, ...
{
      public event EventHandler Checked;
      public event EventHandler Unchecked;

      public bool IsChecked {get; set; }  //BP
      public string GroupName { get; set; } //BP
}
```

```
interface IToggleButtonElement
{
      event EventHandler Checked;
      event EventHandler Unchecked;
}
```

```
static class ToggleButtonElement
{
     public static bool IsChecked { get; set; }  // BP
}
```

 

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP
- MacOS

#### GroupName property

The "RadioButton group Gallery - Legacy" has **three** tabs which show group behavior. 
Radio-buttons with group name null are grouped by their parent.
Radio-buttons with same group name, no matter the hierarchy, are grouped at page level.

Video: https://www.youtube.com/watch?v=fwYSmvjCfbw

image

### PR Checklist ###

- [ ] Has automated tests 


andreinitescu:
Some thoughts:

On Android, I think RadioButtonRenderer could derive from ButtonFastRenderer because android.widget.RadioButton indirectly derives from android.widget.Button. Right now there's a lot of duplication in the RadioButtonRenderer.
Also, I'd like to use ButtonLayoutManager for padding and text, but it's meant to deal with Button's ContentLayout. I'd have to refactor ButtonLayoutManager, maybe a BaseTextButtonLayoutManager which deals with padding and text?

On iOS, it's a similar situation, much of the code in RadioButtonRenderer and ButtonRenderer is duplicated. I think it makes sense here to have a BaseButtonRenderer too.

Waiting for your input, thanks.

rookiejava:
I'll add the Tizen implementation through the subsequent PR. 😉

samhouts:
We also have #4369, and as you mentioned, there can be some shared code there. I wonder how we do that without causing a tangled mess.

Xamarin.Forms.Core/PlatformConfiguration/AndroidSpecific/RadioButton.cs
```diff
@@ -0,0 +1,57 @@
...
+		}
+
+		public static IPlatformElementConfiguration<Android, FormsElement> SetUseDefaultPadding(this IPlatformElementConfiguration<Android, FormsElement> config, bool value)
+		{
+			SetUseDefaultPadding(config.Element, value);
+			return config;
+		}
+		#endregion
+
+		#region UseDefaultShadow
```
StephaneDelcroix: no regions, thanks

Xamarin.Forms.Core/ITextButtonElement.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.ComponentModel;
+
+namespace Xamarin.Forms
+{
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	public interface ITextButtonElement
```
StephaneDelcroix: drop this. add a Text property to ITextElement

Xamarin.Forms.Core/ITextElement.cs
```diff
@@ -1,8 +1,10 @@
 using System;
+using System.ComponentModel;
 
 namespace Xamarin.Forms
 {
-	interface ITextElement
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	public interface ITextElement
```
StephaneDelcroix: you shouldn't be using this properties from renderers, but using SetValue. no need to be public

Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -0,0 +1,425 @@
...
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Windows.Input;
+using Xamarin.Forms.Internals;
+using Xamarin.Forms.Platform;
+
+namespace Xamarin.Forms
+{
+	[RenderWith(typeof(_RadioButtonRenderer))]
+	public partial class RadioButton : IToggleButtonElement, IToggleButtonController, ITextButtonElement
```
StephaneDelcroix: no need for a partial class

StephaneDelcroix: please share code, logic, and naming with `Switch`. a `RadioButton` is a `Switch` with a `Text` and a `GroupName`

Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -0,0 +1,425 @@
...
+	{
+		static Dictionary<string, List<WeakReference<RadioButton>>> _groupNameToElements;
```
StephaneDelcroix: couldn't you use a `ConditionalWeakTable` ?https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2?view=netframework-4.7.1

Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -0,0 +1,425 @@
...
+			get { return (string)GetValue(GroupNameProperty); }
+			set { SetValue(GroupNameProperty, value); }
+		}
+
+		public event EventHandler Checked;
+		public event EventHandler Unchecked;
+
+		void IToggleButtonController.SetIsChecked(bool isChecked)
+		{
+			IsChecked = isChecked;
```
StephaneDelcroix: this will drop existing bindings, the proper way is to use SetValueFromRenderer

Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -0,0 +1,425 @@
...
+		}
+
+		void IToggleButtonElement.RaiseCheckedEvent(bool isChecked)
+		{
+			if (isChecked)
+			{
+				Checked?.Invoke(this, EventArgs.Empty);
```
StephaneDelcroix: use a single `Toggled` event. like it's done in `Switch`

Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -0,0 +1,425 @@
...
+					WeakReference<RadioButton> weakRef = elements[i];
+					if (weakRef.TryGetTarget(out RadioButton rb))
+					{
+						// Uncheck all checked RadioButtons different from the current one
+						if (rb != this && (rb.IsChecked == true) && rootScope == GetVisualRoot(rb))
+						{
+							rb.UncheckRadioButton();
+						}
+
+						++i;
```
StephaneDelcroix: we prefer `i++` to `++i`

Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -0,0 +1,425 @@
...
+namespace Xamarin.Forms
+{
+	[RenderWith(typeof(_RadioButtonRenderer))]
+	public partial class RadioButton : IToggleButtonElement, IToggleButtonController, ITextButtonElement
+	{
+		static Dictionary<string, List<WeakReference<RadioButton>>> _groupNameToElements;
+
+		public static readonly BindableProperty IsCheckedProperty = ToggleButtonElement.IsCheckedProperty;
+
+		public static readonly BindableProperty GroupNameProperty = BindableProperty.Create(nameof(GroupName), typeof(string), typeof(RadioButton), null,
```
StephaneDelcroix: what's the scope of GroupName ? it looks global, which will prevent having a custom control embedding a group displayed twice in the same app

Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -0,0 +1,425 @@
...
+			return parent;
+		}
+	}
+
+	public partial class RadioButton :
+		View,
+		IViewController,
+		IElementConfiguration<RadioButton>,
+		IButtonElement,
+		IButtonController,
```
StephaneDelcroix: I don't think this should be a ButtonElement, and we shouldn't be implementing Pressed and Released. please discuss

Xamarin.Forms.Core/ToggleButton.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+using System.ComponentModel;
+
```
StephaneDelcroix: should be shared with Switch. it's fine to have multiple interface in a single file, but rename it to `ToggleButtonElement` then

Xamarin.Forms.Core/ToggleButton.cs
```diff
@@ -0,0 +1,37 @@
...
+	public interface IToggleButtonController
+	{
+		void SetIsChecked(bool isChecked);
+	}
+
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	interface IToggleButtonElement
+	{
+		event EventHandler Checked;
+		event EventHandler Unchecked;
```
StephaneDelcroix: `Toggled`

Xamarin.Forms.Core/ToggleButton.cs
```diff
@@ -0,0 +1,37 @@
...
+
+		bool IsChecked { get; set; }
+
+		void RaiseCheckedEvent(bool isChecked);
```
StephaneDelcroix: I don't think this is required

Xamarin.Forms.Core/ToggleButton.cs
```diff
@@ -0,0 +1,37 @@
...
+
+		void OnIsCheckedPropertyChanged(bool oldValue, bool newValue);
+	}
+
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	static class ToggleButtonElement
+	{
+		public static readonly BindableProperty IsCheckedProperty = BindableProperty.Create(nameof(IToggleButtonElement.IsChecked), typeof(bool), typeof(IToggleButtonElement), false, propertyChanged: (b, o, n) => OnIsCheckedChanged((IToggleButtonElement)b, (bool)o, (bool)n));
```
StephaneDelcroix: should be `Toggled`, and binding mode should be 2Way. see Switch

andreinitescu:
This is very different than how the CheckBox works. I wish there was feedback on the original issue.

I'm not sure how to continue with this now.

PureWeen:
> This is very different than how the CheckBox works. 

Can you elaborate on this a bit? Are you saying that your approach of an implementation is different than how we did CheckBox? Or that fundamentally the two don't align so nothing should be shared betewen them?


>     ToggleButtonElement can be used as a base class for all kind of toggle buttons, RadioButton, CheckBox, or any custom drawn toggle buttons, etc.

I feel like the mileage here of a common base class wouldn't give us too much. I've found that just using interfaces to illustrate shared property is enough. 

>    On Android, I think RadioButtonRenderer could derive from ButtonFastRenderer because android.widget.RadioButton indirectly derives from android.widget.Button. Right now there's a lot of code duplication across the RadioButtonRenderer and ButtonFastRenderer.
>    Also, I'd like to use ButtonLayoutManager for padding and text, but it's meant to deal with Button's ContentLayout. I'd have to refactor ButtonLayoutManager, maybe a BaseTextButtonLayoutManager which deals with padding and text?

Don't we want to add something like (or just exactly) ContentLayout to Checkbox and Radiobutton as well so you can position the text to the right or left of the checkbox and also effect the spacing?  I hadn't dove into it yet but my plan for checkbox was to just pass the checkbox into buttonlayoutmanager, initialize it the same way it's initialized on material, and then just hope it all works :-) 

Checkbox, RadioButton, and Button with an Image Source set
Are all basically the same thing right? The only difference being that CB and RB have state so the Drawable gets toggled based on that state. Otherwise they all just implement button on the native platforms and it's all just the same code we already use for Button + ImageSource + Text

>     On iOS, it's a similar situation, much of the code in RadioButtonRenderer and ButtonRenderer is duplicated. I think it makes sense here to have a BaseButtonRenderer too.

The implementation of ButtonRenderer on ios is super thin. It's more just passing things to Managers.  I think just following the same strategy with the checkbox renderer of passing to the managers first is a good idea. Then once it's done and working we can look at refactoring to a base class. Either way I feel like this is a minor detail as most of the core shared code here was already refactored to managers

Xamarin.Forms.Core/RadioButton.cs
PureWeen: Pressed and Released are valid for VSM 

People might want to trigger some shadow or background effect when the user has touched the element

andreinitescu:
> Can you elaborate on this a bit? Are you saying that your approach of an implementation is different than how we did CheckBox? 

My RadioButton impl. is very different than the CheckBox impl. I wanted the RadioButton to be an actual button, with text, pressed states.

> Or that fundamentally the two don't align so nothing should be shared betewen them?

They should be aligned I think, they are both toggle buttons and share similar properties (same IsChecked property, same CheckChanged event, and same visual states, same border properties), just with a different UI, and the RadioButton has some extra logic for grouping.

> I feel like the mileage here of a common base class wouldn't give us too much. I've found that just using interfaces to illustrate shared property is enough and then just reusing the bindable properties via the classes like ButtonElement

A toggle button base class would have important advantages in my opinion:
1. Makes it very clear the responsibility of derived classes.
    The implementation for CheckBox and RadioButton would become very short and concise, which is much easier to understand and maintain than having a lot of code duplicated.
    I'm aware adding a base class must always be planned carefully because of single inheritance in C#, but for this specific purpose of a toggle button I think it should be OK.
 
2. Makes it extensible for custom controls.
    If I want to create a custom drawn toggle button, I shouldn't have to create a class from scratch (IsChecked property, CheckedChanged event, border properties, visual state logic, etc.).

However, if you still think it doesn't make sense to have a common class, that's OK for me.

If you'd like me to still work on this control, please let me know feedback on the existing PR.
@StephaneDelcroix already placed feedback, but I'm confused. He is suggesting to have Switch as a base class. But the CheckBox doesn't have Switch as base class... Shouldn't CheckBox and RadioButton be aligned? Also, I'm not sure about having Switch as base class would be OK.
With little planning and guidance, I think I can work on refactoring the RadioButton to align with how you want it.

Thanks!

PureWeen:
> My RadioButton impl. is very different than the CheckBox impl. I wanted the RadioButton to be a button , with text, pressed states. After all on native platforms which have a CheckBox (and all have it except iOS), the native check-box control is actually a button.

I feel like *very different* is a bit of an overstatement :-) Currently I see CheckBox as a subset to your vision.  CheckBox and RadioButton are effectively the same controls just with different images and different state behavior when you click on them. So I see whatever you are doing for RB as something that's just going to be symmetrically applied to CB by whatever means of sharing we come up with 

> The sharing at the cross platform level

Still not sure. Switch has IsToggled and CheckBox has IsChecked. Checkbox hasn't been released so should we call it IsToggled instead and create common base class? In a perfect world I'd say we go the android route and just create a CompoundButton that inherits from Button then just make switch, radio, checkbox all inherit from that because even at the platform level they are all just buttons with states so literally all the code we have that operates on buttons should just be applied to cb, rb, switch.  Let me bounce this off a few people then get back to you.

jamesmontemagno:
Is there a color property like CheckBox has? Perhaps we can match that?

jamesmontemagno:
My feedback is there should only be one event:

public event EventHandler CheckedChanged;

PureWeen:
@andreinitescu how much of the renderer level work do you think can be shifted to the ButtonLayoutManager? 

For example this code is all handled inside the button layout manager
https://github.com/xamarin/Xamarin.Forms/pull/5349/files#diff-e72f13e6476e7d488467cb66436c78eaR385

I feel like most of the renderers should just match the button renderers pretty closely. 
They could even match it down to the ImageSource being exposed right? Because technically users could just supply the ImageSource they want used.

Also I updated the original ticket to match your APIs more closely and then also pulled in @jamesmontemagno comment there of the eventing.

As far as the base classes at the Core level I think we're still leaning towards just inheriting from VisualElement but overall I don't want discussions around that to hold up your work. I think there's still some homogenization at the renderer level that can happen which we will just apply to checkbox once RB is merged and it'd be nice to just have those renderer generalized a bit more

andreinitescu:
> @andreinitescu how much of the renderer level work do you think can be shifted to the ButtonLayoutManager?

You mean `RadioButton` should have a `ButtonContentLayout ContentLayout { get; set; }` property?

PureWeen:
@andreinitescu 

Is this the only reason you didn't use the ButtonLayoutManager

> I'd like to use ButtonLayoutManager for padding and text, but it's meant to deal with Button's ContentLayout. I'd have to refactor ButtonLayoutManager, maybe a BaseTextButtonLayoutManager which deals with padding and text?

That's more of the problem I'd like to figure out so we can just reuse as much code as possible on the renderer side. For example the iOS ButtonLayoutManager I feel like could just be used. If we just figured out a way to refactor to use the ButtonLayoutManagers on the respective platforms. Also it'd be nice if all the padding just passed through the logic in side BLM

andreinitescu:
I see that the ButtonLayoutManager manages a ButtonLayout. It happens that the ButtonLayoutManager also sets the padding, but it doesn't sound to me like RadioButton should use ButtonLayoutManager, unless RadioButton gets ContentLayout property too.

Maybe a better idea is to extract managing the padding out of the ButtonLayout (i.e. PaddingElementManager), because after all, the native padding on Android is really not a button characteristic, native padding can be set on any native Android view.

andreinitescu:
On a second thought, I think maybe having ContentLayout property on a radio button would actually be a good idea for several reasons.
From an implementation point of view, it would allow reusing the existing ButtonLayoutManager.

andreinitescu:
@PureWeen It's still not entirely clear to me the direction you want me to follow.

If we want the RadioButton to have the ContentLayout property, this means RadioButton has ImageSource property as well. 
Shouldn't maybe RadioButton just derive from Button?
On the pro side, we won't need to manage a lot of duplicated properties and methods in both Button and RadioButton.

PureWeen:
> If we want the RadioButton to have the ContentLayout property, this means RadioButton has ImageSource property as well.

Yea! Because technically we could just get rid of all the platform code we created for checkbox and RadioButton and then just create a total cross platform control that's just a button with an image we embed.  

```
<Button ImageSource="checkbox" />
```


So the APIs for checkbox/button/radiobutton should be mostly symmetrical and then CB and RB just have some built in state and images so life is made a littler bit easier. 

@davidortinau 's first question about Checkbox was how he could use a custom image :-) 

> Shouldn't maybe RadioButton just derive from Button?
> On the pro side, we won't need to manage a lot of duplicated properties and methods in both Button and RadioButton.

I'm still not super convinced at the benefit here. The controls at the cross platform level are just property bags. Plus Button has properties like *BorderRadiusProperty* which we don't really want to come along for the ride on Checkbox and Radiobutton

The only place I see where inheritance gets us anything is with the use of ButtonLayoutManager but I think that's easily solved by interfaces.

In my perfect world minds eye :-)  The ButtonRenderer on the platforms would look almost identical to the RadioButton and Checkbox Renderers. The only difference will be that RB and CB have some added behavior around state "checked/groups" etc...

That never goes as smoothly as we'd like it to but in my happy place it'd just be as simple as copying the button renderer and changing the base class

andreinitescu:
> @davidortinau 's first question about Checkbox was how he could use a custom image :-)

Even if I add the ImageSource property to RadioButton, if you want it to be used to draw the tick, this would require modifying how the native control draws the tick. For example, on Android, it would require setting the Android style to a different style so it won't display the default tick anymore.

If you're now looking for as much flexibility as possible: 
What if someone really wants the native look and feel on those platforms where RadioButton exists? (and in fact, all platforms except iOS have a RadioButton). After all, isn't this the classic pattern in Xamarin Forms, to render to native controls?
Or, what if someone wants a RadioButton with tick + custom image? Or a RadioButton with all 3 (tick + custom image + text)? 
All these questions apply to CheckBox too.

So here's what I am thinking:

1. The Core RadioButton is the same with Button (exact same implementation) and some specific logic for the radio button (grouping). This means RadioButton has ImageSource as well but its purpose is not to draw the tick but to draw an image beside the tick, just like the Button.  In other words, the RadioButton will be capable to display tick + text + image.
   About reusing the ButtonLayoutManager:  On Android, if I'm not mistaken, it should just work. On iOS, it can be reused easily, but it would require some tweaking to take into account the tick.

2. About having a custom image for the tick. Either:
  a. Have an `ImageSource TickSource` property. This will be beside the `ImageSource` property. If it's null, the renderers will use the default tick (on iOS, the default tick drawn in the  XamarinForms.Platform.iOS)
  b. Developers will have to rely on native: on Android using styles. On iOS, using a custom CheckBox renderer (we need to make sure a custom image can be given).

What do you think?

Xamarin.Forms.Core/PlatformConfiguration/AndroidSpecific/RadioButton.cs
```diff
@@ -0,0 +1,57 @@
+﻿namespace Xamarin.Forms.PlatformConfiguration.AndroidSpecific
```
PureWeen: Do we need any of these platform specifics? I think these really only existed for Button so that we could change how these worked in a non breaking way. With RadioButton we can just use the defaults

PureWeen:
> Even if I add the ImageSource property to RadioButton, if you want it to be used to draw the tick, this would require modifying how the native control draws the tick. For example, on Android, it would require setting the Android style to a different style so it won't display the default tick anymore.

Agreed :-) The idea definitely needs to be spiked a little bit more and I'm not too sure how tricky this will be. I'm thinking it's doable because of all the work @mattleibow did to get the Material Button positioning the image correctly with the layout manager. But Android can be super finicky :-/ if exposing the ContentLayout API just proves to be too tricky then it would be ok to delay that for a later PR. Just get it working well for the default case and with a swappable image. The Android *CompoundButton* has a method called *setButtonDrawable* so 🤞 that it's easy. Then people can just use our VSM to swap the image based on state

> Or, what if someone wants a RadioButton with tick + custom image? Or a RadioButton with all 3 (tick + custom image + text)?

I would leave this for an enhancement after we get RadioButton finished. 

So for the first version I would say make Button and RadioButton APIs mostly the same (we can ignore ContentLayout for the first pass unless you find that it's trivial to implement).

I feel you've mostly done this on the cross platform side but just need to add ImageSource and then on the platform level reuse ButtonLayoutManager for padding and insets. 

If adding ImageSource proves to be difficult as well then we can punt on that for another enhancement.

andreinitescu:
Just checking to be sure I understand correctly how you see it workning.
What will the default value of ImageSource for CheckBox be? Is it `null` and in this case the default native tick is drawn? Otherwise, it will draw the tick with the given image.
Is this what how you're seeing it too?

> I would leave this for an enhancement after we get RadioButton finished.

I think the name of the property needs to be different than the `ImageSource` (maybe `TickSource`) so that if later we can add `ImageSource` to RadionButotn to support drawing a custom image beside the tick and text.

PureWeen:
> What will the default value of ImageSource for RadionButotn be? Is it null and in this case the default native tick is drawn? Otherwise, it will draw the tick with the given image.
Is this what how you're seeing it too?

I think that'll work.  The only scenario I'm wondering about is if we want to enable the ability to just not have an image. Like if they just want the text to indicate state. Possible solutions to this I see are
- users can just use a 0, 0 image or transparent (kind of a gross solution but workable)
- differentiate between setting the ImageSource to NULL and Clearing the ImageSource property
  - if you clear a bindable property the assumption is that it will revert to "Default" state so in this case default state is a checkbox. Where as setting it to null would be an indication that you didn't want a drawable. I'm not super sure on what this implementation would look like
- If/when we add ContentLayout maybe adding an additional ImagePosition property to the enum of "hide" which just indicates you want the image hidden
  - so far I like this one the most

> I think the name of the property needs to be different than the ImageSource (maybe TickSource) so that if later we can add ImageSource to RadionButotn to support drawing a custom image beside the tick and text.

I think this makes sense. I'm just trying to think through the name. I'm worried if we call it something like CheckMarkSource or TickSource then that is too specific to the state of being "checked". Ideally the name just indicates the currently displayed image. Which keeps taking my brain back to just calling it ImageSource

andreinitescu:
@PureWeen Please checkout changes for Core and Android only (if you'd like to run, unload the other platform projects please as it won't compile I think ).

Here's my approach:
```
public class RadioButton : Button
{
      public const string IsCheckedVisualState = "IsChecked";
 
      public event EventHandler<CheckedChangedEventArgs> CheckedChanged;

      public bool IsChecked { get; set; } // BP
      public string GroupName { get; set; } // BP
      public ImageSource ButtonSource { get; set; } // BP
}
```

I chose `ButtonSource` being inspired by Android's `setButtonDrawable(Drawable d)` on the `CompoundButton` class.
Deriving from Button makes ButtonLayoutManager just work.

On Android, I copied and pasted the fast ButtonRenderer to RadioButtonRenderer and only implemented the checked listener and  updating the button image.

Here's how it looks:
image

Instead of duplicating all the `ButtonRenderer` code into the `RadioButtonRenderer`, I'm thinking I could move the logic from `ButtoinRenderer` to a `ButtonRendererManager`. Both `ButtonRenderer` and `RadioButtonRenderer` could forward to an inner `ButtonRendererManager` instance.

What do you think?

I could do the same thing for iOS, the existing ButtonLayoutManager will work, but it needs to take into account the button check image. For this, I could either subclass ButtonLayoutManager (RadioButtonLayoutManager or just adapt a bit the ButtonLayoutManager).

Xamarin.Forms.Core/RadioButton.cs
andreinitescu: @StephaneDelcroix 
Thanks for bringing the idea of `ConditionalWeakTable`. How do you see using it instead of the `_groupNameToElements` dictionary and how would it better?  I need to get all the radio buttons which are part of the same group (a string).

Xamarin.Forms.Core/RadioButton.cs
andreinitescu: @StephaneDelcroix  I tried to describe the scope of GroupName in the description. Does it make sense?

andreinitescu:
@samhouts Just wondering why's this tagged with a/shell ? 

@StephaneDelcroix Large majority of your comments don't apply anymore because implementation changed a lot based on feedback I got from Shane. However, there are still a couple which makes sense and I replied.

Xamarin.Forms.Core/ITextButtonElement.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.ComponentModel;
+
+namespace Xamarin.Forms
+{
+	//[EditorBrowsable(EditorBrowsableState.Never)]
+	//public interface ITextButtonElement
```
StephaneDelcroix: ??

Xamarin.Forms.Core/PlatformConfiguration/AndroidSpecific/RadioButton.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+	//using FormsElement = Forms.RadioButton;
```
StephaneDelcroix: ??

Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -0,0 +1,173 @@
...
+
+		public event EventHandler<CheckedChangedEventArgs> CheckedChanged;
+
+		public bool IsChecked
+		{
+			get { return (bool)GetValue(IsCheckedProperty); }
+			set { SetValue(IsCheckedProperty, value); }
+		}
+
+		public string GroupName
```
StephaneDelcroix: It looks like the scope of the group name is global, meaning you can't have 2 instances of the same custom control alive at once.

How is this done in other platforms ?

Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -0,0 +1,173 @@
...
+		public static readonly BindableProperty IsCheckedProperty = BindableProperty.Create(
+			nameof(IsChecked), typeof(bool), typeof(RadioButton), false, propertyChanged: (b, o, n) => ((RadioButton)b).OnIsCheckedPropertyChanged((bool)n));
+
+		public static readonly BindableProperty GroupNameProperty = BindableProperty.Create(
+			nameof(GroupName), typeof(string), typeof(RadioButton), null, propertyChanged: (b, o, n) => ((RadioButton)b).OnGroupNamePropertyChanged((string)o, (string)n));
+
+		public static readonly BindableProperty ButtonSourceProperty = BindableProperty.Create(
+			nameof(ButtonSource), typeof(ImageSource), typeof(RadioButton), null);
+
+		public event EventHandler<CheckedChangedEventArgs> CheckedChanged;
```
StephaneDelcroix: I prefer `Toggled`, which means just that

Xamarin.Forms.Core/ITextButtonElement.cs
andreinitescu: Fixed

Xamarin.Forms.Core/PlatformConfiguration/AndroidSpecific/RadioButton.cs
andreinitescu: Fixed

Xamarin.Forms.Core/RadioButton.cs
andreinitescu: I don't understand what you mean by it's "global". You definitely can have two instances with the same group name, but it depends if the two have the same parent or not. You mentioned something similar in a previous comment, and I replied to that too. I explained the group behavior in the description, there's also a gallery page showing the  group name behavior in different scenarios, and I made a video with the page gallery.

Xamarin.Forms.Core/RadioButton.cs
andreinitescu: The CheckBox has the Checked event too... Based on discussions with Shane I thought this is the preferred approach?

andreinitescu:
- what's up with the empty files ?
  Fixed. There have been so much refactoring and work I missed that. 
 
- unsure about naming of event
  I provided feedback on this in your comment

- we probably want a Command too
  Command for what exactly? Check command? Can we add that later? CheckBox doesn't have Command, Text, Image for check, and much more
 
- unsure about the scope of GroupName, global scope seems too broad
  I provided feedback on this in your comment

StephaneDelcroix:
sorry about the Command thing. I meant: there's is a Command without a CommandParameter

Xamarin.Forms.Core/RadioButton.cs
StephaneDelcroix: Checked would be fine, at least better than CheckedChanged. Switch has `Toggled`, so  ¯\_(ツ)_/¯

Xamarin.Forms.Core/RadioButton.cs
StephaneDelcroix: ok, so the scope is name + parent ?

StephaneDelcroix: so what if the individual radio button are nested within frames ?

andreinitescu: > ok, so the scope is name + parent ?

yes

> so what if the individual radio button are nested within frames ?

what about it?

andreinitescu:
> sorry about the Command thing. I meant: there's is a Command without a CommandParameter

I'm not following you, my RadioButton has no Command.

Xamarin.Forms.Core/RadioButton.cs
StephaneDelcroix: you have
```
StackLayout
  Frame
    RadioButton
  Frame
    RadioButton
```

so the 2 radio buttons aren't in the same group ?

Xamarin.Forms.Core/RadioButton.cs
StephaneDelcroix: I don't see how this is resolved

Xamarin.Forms.Core/RadioButton.cs
StephaneDelcroix: the problem with this approach is the List keeps getting populated, and so the performance degrades over time

Xamarin.Forms.Core/RadioButton.cs
StephaneDelcroix: also, you say the scope is the parent, but looking at GetVisualRoot, the scope is the page. what if I have this

```
Page
  StackLayout
    YesNoMaybeControl
    YesNoMaybeControl
    YesNoMaybeControl
```

Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -0,0 +1,173 @@
...
+			for (int i = 0; i < elements.Count;)
+			{
+				if (elements[i].TryGetTarget(out RadioButton rb) && rb == elementToRemove)
+					elements.RemoveAt(i);
+				else
+					i++;
+			}
+		}
+
+		static Element GetVisualRoot(Element element)
```
StephaneDelcroix: 👎

Xamarin.Forms.Core/RadioButton.cs
StephaneDelcroix: I've seen `PurgeDeads`, but we probably can do better

Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -0,0 +1,173 @@
...
+
+		public static readonly BindableProperty GroupNameProperty = BindableProperty.Create(
+			nameof(GroupName), typeof(string), typeof(RadioButton), null, propertyChanged: (b, o, n) => ((RadioButton)b).OnGroupNamePropertyChanged((string)o, (string)n));
+
+		public static readonly BindableProperty ButtonSourceProperty = BindableProperty.Create(
+			nameof(ButtonSource), typeof(ImageSource), typeof(RadioButton), null);
+
+		public event EventHandler<CheckedChangedEventArgs> CheckedChanged;
+
+		public bool IsChecked
```
StephaneDelcroix: `Checked` is probably a better name. We try to avoid property names containing multiple verbs. The `IsXxxx` construct is valid when Xxxx is an adjective. We can debate that in this case `Checked` is used as an adjective or not, I still think the name is overloaded

StephaneDelcroix: this is not a big deal, I know we have `IsEnabled` already, and might want to reserve the `Checked` name for the event. so it's fine to leave it as is, and rename the event

Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -0,0 +1,173 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using Xamarin.Forms.Platform;
+
+namespace Xamarin.Forms
+{
+	[RenderWith(typeof(_RadioButtonRenderer))]
+	public partial class RadioButton : Button
```
StephaneDelcroix: drop the partial. where is the `IToggleElement` interface you use to have ?

Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -0,0 +1,173 @@
...
+	{
+		static Dictionary<string, List<WeakReference<RadioButton>>> _groupNameToElements;
+
+		public const string IsCheckedVisualState = "IsChecked";
+
+		public static readonly BindableProperty IsCheckedProperty = BindableProperty.Create(
```
StephaneDelcroix: this isn't 2-Way, but should

Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -0,0 +1,173 @@
...
+			nameof(IsChecked), typeof(bool), typeof(RadioButton), false, propertyChanged: (b, o, n) => ((RadioButton)b).OnIsCheckedPropertyChanged((bool)n));
+
+		public static readonly BindableProperty GroupNameProperty = BindableProperty.Create(
+			nameof(GroupName), typeof(string), typeof(RadioButton), null, propertyChanged: (b, o, n) => ((RadioButton)b).OnGroupNamePropertyChanged((string)o, (string)n));
+
+		public static readonly BindableProperty ButtonSourceProperty = BindableProperty.Create(
```
StephaneDelcroix: how does this works ? is this a spriteMap ? don't we need a source for both on and off ?

jsuarezruiz:
@andreinitescu There is some feedback (and also conflicts). Let us know if we can help you with something ;)

SagarPanwala:
How end user can differentiate between RadioButton and Checkbox in iOS ?
Current implementation of checkbox in iOS is in Round Shape, which identically says user that if there are group of checkboxes, they can select only 1.
Can we make Square shape for checkbox in iOS as well or give a property for shape?

PR #5350, opened 2019-02-22
[Android] add build error if TargetFramework is too low

PureWeen:
### Description of Change ###
Currently because of net standard targets you can install Xamarin.Forms into any Android project regardless of how low the Target Framework is set. This will just cause the Android platform DLL to not get added to the android project and nothing will compile. This error gives the user more information about why their particular project isn't compiling.

### Issues Resolved ### 
http://feedback.devdiv.io/458514

Trying to compile an Android target with Target Framework lower than 8.1 will throw an actionable error message

### Testing Procedure ###
- add nuget into a project with android target framework < 8.1 and you should see a compile error

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/.nuspec/Xamarin.Forms.Android.targets
+++ b/.nuspec/Xamarin.Forms.Android.targets
@@ -2,7 +2,7 @@
   <ItemGroup>
     <ProguardConfiguration Include="$(MSBuildThisFileDirectory)proguard.cfg" />
   </ItemGroup>
-  <Target Name="ValidateTargetFrameworkVersionForForms" BeforeTargets="CoreCompile">
+  <Target Name="ValidateTargetFrameworkVersionForForms" BeforeTargets="CoreCompile" Condition="'$(XFDisableTargetFrameworkValidation)' != 'True'">
     <PropertyGroup>
       <MinTargetFrameworkVersionForForms>8.1</MinTargetFrameworkVersionForForms>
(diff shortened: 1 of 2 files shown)


.nuspec/Xamarin.Forms.Android.targets
```diff
@@ -2,4 +2,12 @@
   <ItemGroup>
     <ProguardConfiguration Include="$(MSBuildThisFileDirectory)proguard.cfg" />
   </ItemGroup>
+  <Target Name="ValidateTargetFrameworkVersionForForms" BeforeTargets="CoreCompile">
```
StephaneDelcroix: needs the condition so we can disable the validation (see other checks in the same file)

.nuspec/Xamarin.Forms.Visual.Material.targets
```diff
@@ -1,5 +1,5 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-    <Target Name="ValidateTargetFrameworkVersionForMaterial" BeforeTargets="BeforeCompile">
+    <Target Name="ValidateTargetFrameworkVersionForMaterial" BeforeTargets="CoreCompile">
```
StephaneDelcroix: needs the condition so we can disable the validation

mattleibow:
Ooh, yeah. Forgot about that. We always need a way to disable the checks...
Good catch!

PR #5369, opened 2019-02-25
[previewer] resilience on createvalue

StephaneDelcroix:
### Description of Change ###

Up to now, we were resilient on missing properties(value assignment) for the previewer, This adds supports for resilience on unsupported types (object creation).

### Issues Resolved ### 

- fixes #5171

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [x] Has automated tests 

> VS bug [#802254](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/802254)

### DIFF
--- a/Xamarin.Forms.Xaml/ApplyPropertiesVisitor.cs
+++ b/Xamarin.Forms.Xaml/ApplyPropertiesVisitor.cs
@@ -36,13 +36,15 @@ namespace Xamarin.Forms.Xaml
 		public bool StopOnResourceDictionary { get; }
 		public bool VisitNodeOnDataTemplate => true;
 		public bool SkipChildren(INode node, INode parentNode) => false;
-		public bool IsResourceDictionary(ElementNode node) => typeof(ResourceDictionary).IsAssignableFrom(Context.Types[node]);
+		public bool IsResourceDictionary(ElementNode node) => Context.Types.TryGetValue(node, out var type) && typeof(ResourceDictionary).IsAssignableFrom(type);
 
 		public void Visit(ValueNode node, INode parentNode)
 		{
 			var parentElement = parentNode as IElementNode;
 			var value = Values [node];
-			var source = Values [parentNode];
+			if (!Values.TryGetValue(parentNode, out var source) && Context.ExceptionHandler != null)
+				return;
+
 			XmlName propertyName;
 
 			if (TryGetPropertyName(node, parentNode, out propertyName)) {
@@ -95,7 +97,8 @@ namespace Xamarin.Forms.Xaml
 				parentElement = parentNode as IElementNode;
 			}
 
-			var value = Values[node];
+			if (!Values.TryGetValue(node, out var value) && Context.ExceptionHandler != null)
+				return;
 
 			if (propertyName != XmlName.Empty || TryGetPropertyName(node, parentNode, out propertyName)) {
 				if (Skips.Contains(propertyName))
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Xaml/ApplyPropertiesVisitor.cs
```diff
@@ -36,13 +36,15 @@ public ApplyPropertiesVisitor(HydrationContext context, bool stopOnResourceDicti
 		public bool StopOnResourceDictionary { get; }
 		public bool VisitNodeOnDataTemplate => true;
 		public bool SkipChildren(INode node, INode parentNode) => false;
-		public bool IsResourceDictionary(ElementNode node) => typeof(ResourceDictionary).IsAssignableFrom(Context.Types[node]);
+		public bool IsResourceDictionary(ElementNode node) => Context.Types.TryGetValue(node, out var type) && typeof(ResourceDictionary).IsAssignableFrom(type);
 
 		public void Visit(ValueNode node, INode parentNode)
 		{
 			var parentElement = parentNode as IElementNode;
 			var value = Values [node];
```
mattleibow: Shouldn't this also be in a TryGetValue? We do that for the one in the next method: 
https://github.com/xamarin/Xamarin.Forms/pull/5369/files#diff-370e53f89e3c43f7073b0f2f85dcd6b2R100

Xamarin.Forms.Xaml/ApplyPropertiesVisitor.cs
```diff
@@ -36,13 +36,15 @@ public ApplyPropertiesVisitor(HydrationContext context, bool stopOnResourceDicti
...
-			var source = Values [parentNode];
+			if (!Values.TryGetValue(parentNode, out var source) && Context.ExceptionHandler != null)
```
mattleibow: Isn't adding the `Context.ExceptionHandler != null` changing the behaviour? Is that what we want?

Xamarin.Forms.Xaml/ApplyPropertiesVisitor.cs
```diff
@@ -95,20 +97,23 @@ public void Visit(ElementNode node, INode parentNode)
 				parentElement = parentNode as IElementNode;
 			}
 
-			var value = Values[node];
+			if (!Values.TryGetValue(node, out var value) && Context.ExceptionHandler != null)
```
mattleibow: Are we happy with the `var s0 = Values[parentNode];` in the chunk above?

Xamarin.Forms.Xaml/ApplyPropertiesVisitor.cs
StephaneDelcroix: it's very unlikely for a ValueNode (usually created from a string literal) to fail construction, so this is probably fine

Xamarin.Forms.Xaml/ApplyPropertiesVisitor.cs
StephaneDelcroix: we don't want to hide potential bugs, so we only return if the value isn't found AND the previewer is plugged in. If the previewer is not there, we no longer fail with a KeyNotFound, but probably with some NRE later on

none of this should ever happen. if it happens, it indicates a bug in the loader or an error (that we don't handle) in the xaml. In both cases, I want this code to keep failing, the issue to be reported so we can either fix the loader, or throw a meaningful exception

Xamarin.Forms.Xaml/ApplyPropertiesVisitor.cs
StephaneDelcroix: TryGetPropertyName will probably fail on unregistered parentNode, so it looks safe. again, I want this code to keep throwing on anything going wrong

PR #5373, opened 2019-02-25
[iOS] Load specified Font Family or PostScript Font

### LINKED ISSUE #5372: FontImageSourceHandler on iOS improperly loads Font
### Description

The iOS implementation of the FontImageSourceHandler improperly loads the FontFamily. For example when using Font Awesome Pro you would have the Font Family "Font Awesome 5 Pro" which would have the following fonts:

- FontAwesome5Pro-Light
- FontAwesome5Pro-Regular
- FontAwesome5Pro-Solid

The lookup logic requires the use of the Font Family Name instead of allowing the PostScript name meaning that you have no control over which font gets returned, as a result you will always get the Regular font. 

### Expected Behavior

It would be expected that this would produce an icon from the Font Awesome Solid font for the User glyph.

```xml
<FontImageSource Glyph="{x:Static local:FontAwesomeSolid.User}"
              FontFamily="FontAwesome5Pro-Solid" />
```

### Actual Behavior

As shown the current lookup logic is only looking at the overall Font Family:

```cs
var hasFontFamily = fontsource.FontFamily != null && UIFont.FamilyNames.Contains(fontsource.FontFamily);
var font = hasFontFamily ?
    UIFont.FromName(fontsource.FontFamily, (float)fontsource.Size) :
    UIFont.SystemFontOfSize((float)fontsource.Size);
```

This really should be more like:

```cs
var font = UIFont.FromName(fontsource.FontFamily ?? string.Empty, (float)fontsource.Size) ??
    UIFont.SystemFontOfSize((float)fontsource.Size);
```

This will allow specifying Icon Fonts by their PostScript name or getting the default font for a particular Font Family.

### Basic Information

- Version with issue: 3.6+

### Repro

[Issue5372.zip](https://github.com/xamarin/Xamarin.Forms/files/2903427/Issue5372.zip)

dansiegel:
### Description of Change ###

Fixes the iOS FontImageSourceHandler to use any valid UIFont Name which could be the Font Family Name or the Post Script name thereby allowing users to specify the exact font when using multiple Icon fonts within the same Font Family i.e. Font Awesome.

### Issues Resolved ### 

- fixes #5372 

### API Changes ###

 None

### Platforms Affected ### 

- iOS

The FontImageSourceHandler will now locate valid Fonts by PostScript name and not just use the default font for a specified Font Family.

### Testing Procedure ###

Visual Studio nearly crashed my system just loading Xamarin Forms... I did test the UIFont loading in a separate project though and validated that this will work without throwing exceptions. The `fontsource.FontFamily` is guarded from providing a null value which would throw a NullReferenceException rather than simply providing a null UIFont.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/ImageRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/ImageRenderer.cs
@@ -202,9 +202,7 @@ namespace Xamarin.Forms.Platform.iOS
 			{
 				var iconcolor = fontsource.Color != Color.Default ? fontsource.Color : Color.White;
 				var imagesize = new SizeF((float)fontsource.Size, (float)fontsource.Size);
-				var hasFontFamily = fontsource.FontFamily != null && UIFont.FamilyNames.Contains(fontsource.FontFamily);
-				var font = hasFontFamily ?
-					UIFont.FromName(fontsource.FontFamily, (float)fontsource.Size) :
+				var font = UIFont.FromName(fontsource.FontFamily ?? string.Empty, (float)fontsource.Size) ??
 					UIFont.SystemFontOfSize((float)fontsource.Size);
 
 				UIGraphics.BeginImageContextWithOptions(imagesize, false, 0f);


samhouts:
@dansiegel Can you rebase to 3.6.0, please? Thanks!

dansiegel:
@samhouts done...

hartez:
@dansiegel Could you post a repro of the problem here or in #5372 so we have something to verify this with?

A UI test (even a manual one) would be even better, but I realize that might be difficult since you'd need to include the fonts in the Control Gallery project.

dansiegel:
@hartez no problem... when i get my machine back up I'll throw together a sample

Xamarin.Forms.Platform.iOS/Renderers/ImageRenderer.cs
```diff
@@ -202,9 +202,7 @@ public sealed class FontImageSourceHandler : IImageSourceHandler
 			{
 				var iconcolor = fontsource.Color != Color.Default ? fontsource.Color : Color.White;
 				var imagesize = new SizeF((float)fontsource.Size, (float)fontsource.Size);
-				var hasFontFamily = fontsource.FontFamily != null && UIFont.FamilyNames.Contains(fontsource.FontFamily);
-				var font = hasFontFamily ?
-					UIFont.FromName(fontsource.FontFamily, (float)fontsource.Size) :
+				var font = UIFont.FromName(fontsource.FontFamily ?? string.Empty, (float)fontsource.Size) ??
```
mattleibow: As long as this does not throw for an invalid name, we should be good.

mattleibow: Confirmed: `UIFont.FromName("carrots", 12) == null`

dansiegel: @mattleibow passing in a null for the font name throws a NullReferenceException hence the null guard where we pass in string.Empty if the FontFamily is null... if the Font is not located it returns a default (null) UIFont which is guarded to provide the default system font.

mattleibow: @dansiegel ✅ Merge!

PR #5397, opened 2019-02-27
Catch exceptions thrown by TypeConverter when setting properties

TimBarham:
### Description of Change ###

In `ApplyPropertiesVisitor.TrySetValue()`, catch and record any exception thrown by the TypeConverter and return false, rather than letting it throw.

The exception will later be thrown by `ApplyPropertiesVisitor.SetPropertyValue()`, but this change means `HydrationContext.ExceptionHandler` will be called if it has been set (rather than throwing the exception).

### Issues Resolved ### 

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/802696

### API Changes ###

None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###

* Manually tested exception is now passed to ExceptionHandler when running in the Previewer
* Added a test to verify the above scenario
* Manually tested exception still throws in non-Previewer scenarios.

### PR Checklist ###

- [x] Has automated tests


TimBarham:
@samhouts and @StephaneDelcroix - this PR is created against 3.5.0 - is that appropriate?

@alanmcgovern and @mattleibow - FYI.

PR #5398, opened 2019-02-27
[Android] Remove changes to IVisualElementRenderer so it remains backwards compatible

PureWeen:
### Description of Change ###
Added a new interface to handle LayoutChanges so that IVisualElementRenderer can stay the same and not break older libraries 

### Platforms Affected ### 
- Android

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/ImageButtonRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/ImageButtonRenderer.cs
@@ -21,7 +21,8 @@ namespace Xamarin.Forms.Platform.Android
 		IImageRendererController,
 		AView.IOnFocusChangeListener,
 		AView.IOnClickListener,
-		AView.IOnTouchListener
+		AView.IOnTouchListener,
+		ILayoutChanges
 	{
 		bool _inputTransparent;
 		bool _disposed;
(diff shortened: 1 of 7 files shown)


samhouts:
Master is where we introduced this problem? Or do we need to target a release branch?

PureWeen:
@samhouts 3.4.0 is where it was introduced

I just targeted master because we hadn't seen issues about it and just didn't want to complicate the 3.6.0 release

But I can retarget

samhouts:
Actually, I think if we do this now, we'll break them again. https://github.com/luberda-molinet/FFImageLoading/issues/1154

PureWeen:
> Actually, I think if we do this now, we'll break them again. [luberda-molinet/FFImageLoading#1154](https://github.com/luberda-molinet/FFImageLoading/issues/1154)

So based on that I don't think we can merge this in

daniel-luberda:
Regarding FFImageLoading, there are currently two versions: 

- stable nuget: old `IVisualElementRenderer` behaviour 
- prerelease nuget: new `IVisualElementRenderer` behaviour 

I'll adapt code to whichever implementation you choose.

PureWeen:
@daniel-luberda  the plan is to roll it back to 

>     * stable nuget: old `IVisualElementRenderer` behaviour

There are probably other libraries out there that also broke from this change so we'd prefer to roll back changes

PureWeen:
Alright @daniel-luberda  it's merged in so the fix should be available with the first release of 3.6

We're working on putting some better checks in place so we can avoid these unfortunate scenarios. That being said, if this does happen again please let us know and we'll get a service release out ASAP

PR #5407, opened 2019-02-27
Adds test covering case where markup extension throws an exception

TimBarham:
When rendering in the previewer, this should be caught and passed to `ExceptionHandler`.

@alanmcgovern FYI

### DIFF
--- a/Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
+++ b/Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
@@ -550,6 +550,25 @@ namespace Xamarin.Forms.Xaml.UnitTests
 			Assert.DoesNotThrow(() => XamlLoader.Create(xaml, true));
 			Assert.That(exceptions.Count, Is.EqualTo(1));
 		}
+
+		[Test]
+		public void IgnoreMarkupExtensionException()
+		{
+			// The user scenario is a user type with a static member referenced in an x:Static markup extension. The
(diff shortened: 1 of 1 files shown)


TimBarham:
Original bug report: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/802437

Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
```diff
@@ -550,6 +550,25 @@ public void IgnoreConverterException()
 			Assert.DoesNotThrow(() => XamlLoader.Create(xaml, true));
 			Assert.That(exceptions.Count, Is.EqualTo(1));
 		}
+
+		[Test]
+		public void IgnoreMarkupExtensionException()
+		{
+			// The user scenario is a user type with a static member referenced in an x:Static markup extension. The
+			// previewer replaces the user type, so the static member does not exist. However we should handle general
+			// cases where x:Static and other markup extensions fail.
```
StephaneDelcroix: actually, in this case, your x:Static syntax is wrong, but whatever, the point is having a markup that throws...

TimBarham: Yes, I know... but either way it ultimately throws in the same method, so it didn't seem important whether I had `{x:Static Foo}` or `{x:Static Foo.Bar}`

TimBarham: Happy to include both 😁

PR #5408, opened 2019-02-28
Revert "[previewer] resilience on createvalue"

samhouts:
Reverts xamarin/Xamarin.Forms#5369 per https://github.com/xamarin/Xamarin.Forms/pull/5171#issuecomment-467671840

> VS bug [#803779](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/803779)

### DIFF
--- a/Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
+++ b/Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
@@ -51,7 +51,7 @@ namespace Xamarin.Forms.Xaml.UnitTests
 		}
 
 		[Test]
-		public void ContentPageWithMissingTypeMockviewReplacement()
+		public void ContentPageWithMissingType()
 		{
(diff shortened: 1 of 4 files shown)


TimBarham:
Don't merge this yet, @StephaneDelcroix ... I going to experiment a bit with your original fix in place before we revert it, and I want to chat to @alanmcgovern some more (@samhouts will tag her tomorrow AM, so we have until then to make a final final decision).

While with current Previewer functionality I think throwing with invalid types gives us the best behavior, going forwards I do like the idea that, when `ExceptionHandler` is provided, Forms catches any errors it possibly can and passes them to `ExceptionHandler` rather than throwing... so Forms itself is as resilient as possible, and we can decide on the Previewer side what to do with those exceptions.

But we currently don't have anything in place for that - all calls to `ExceptionHandler` are logged, but otherwise ignored.

So I'm tempted to say let's not revert this (sigh... after I spent an hour convincing @alanmcgovern we should).

StephaneDelcroix:
This does not make any sense, but I appreciate your effort :)

PR #5410, opened 2019-02-28
Adds a test to verify root node FallbackTypeInfos

TimBarham:
If the root tag is something like `<local:Foo>`, Forms doesn't generate any `FallbackTypeInfos`, so the previewer has no way to process the type.

This seems to be because of [these lines](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Xaml/XamlParser.cs#L422) in `XamlParser.GetTypeReference()`...
```
	if (defaultAssemblyName == null)
		return null;
```
When determining the type of the root tag, `defaultAssemblyName` is null because `Context.RootElement` (referenced in `CreateValuesVisitor.Visit()` to get the root assembly) has not yet been set.


TimBarham:
Here is the original bug report: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/801087

PR #5413, opened 2019-02-28
[X] allow previewer fallback on rootnode

StephaneDelcroix:
### Description of Change ###

[X] allow previewer fallback on rootnode

### Issues Resolved ### 

- fixes #5410

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [x] Has automated tests 


TimBarham:
Thanks @StephaneDelcroix.

I tried this out. While this makes the test pass, it will still fail if the xmlns doesn't specify an assembly name. Is there a way we can make that work? (I get that Forms won't be able to resolve in that situation, but we will be able to, because we know on the previewer side what the root element's assembly is)

TimBarham:
LGTM, thanks!

StephaneDelcroix:
do not merge yet

StephaneDelcroix:
I had to check how the change was behaving without the previewer plugged in for this sample. It throws as it should:

> Xamarin.Forms.Xaml.XamlParseException : Type local:MissingType not found in xmlns using:my.namespace

PR #5426, opened 2019-03-01
fix changed apis

### LINKED ISSUE #4644: [iOS] Additional parameter on ToolbarItemExtensions.ToUIBarButtonItem in XF 4 is a breaking change
As part of the Shell work (according to the history) an extra optional parameter was added to a method:

https://github.com/xamarin/Xamarin.Forms/blob/0a67f8b0a105fbf56b6ac704cfc283840a189817/Xamarin.Forms.Platform.iOS/Extensions/ToolbarItemExtensions.cs#L12

This was an added as an optional parameter, but that doesn't prevent libraries built against older versions of Xamarin Forms from breaking.

This change means any library using the method `ToolbarItemExtensions.ToUIBarButtonItem` will have to synchronize a source change with a Forms dependency change (which will be a problem if there are any libraries using this method that are no longer maintained).

Although I suspect the number of libraries this could affect is quite low, this just came up here: https://github.com/jsmarcus/Iconize/issues/115

It should be possible to add the new behavior as an overload (without optional parameters).

### LINKED ISSUE #5170: Crash on launch after upgrading to 3.5
### Steps to Reproduce

1. Upgrade project from 3.4.0.1029999 to 3.5.0.129452

### Expected Behavior

There are no documented API/breaking changes, so project should be compatible.

### Actual Behavior

Crash on launch, both Android and UWP, see screenshot below.

xamforms_missingfieldexception

### Basic Information

- Version with issue: 3.5.0.129452
- Last known good version: 3.4.0.1029999
- IDE: Visual Studio 2017 15.9.6
- Platform Target Frameworks:
  - Android: 9.0
  - UWP: 17763

### Reproduction Link

Managed to reproduce a similar exception, simply by taking a new Xamarin.Forms Master-Detail project and updating all nuget packages. See below.

xamforms_entrypointnotfoundexception

[App1.zip](https://github.com/xamarin/Xamarin.Forms/files/2845264/App1.zip)

PureWeen:
### Description of Change ###
- Fix API changes

### Issues Resolved ### 
- fixes #4644
- fixes #5170

### API Changes ###
- added back default handler for HandlerAttribute
- set UpdateTextColor, UpdatePlaceHolderText, UpdateFont, UpdatePlaceholderColor, UpdateColor on the EntryRenderers to protected so they are exposed to the world
- add back *public PickerEditText(Context context, IPickerRenderer pickerRenderer) : this(context)*
- add back *public static void ApplyKeyboard(this IUITextInput textInput, Keyboard keyboard)*
- add back *public static UIKit.UIBarButtonItem ToUIBarButtonItem(this Xamarin.Forms.ToolbarItem item, bool forceName)*

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS
- Android

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
@@ -239,7 +239,7 @@ namespace Xamarin.Forms.Platform.iOS
 			Control.TextAlignment = Element.HorizontalTextAlignment.ToNativeTextAlignment(((IVisualElementController)Element).EffectiveFlowDirection);
 		}
 
-		protected internal virtual void UpdateColor()
+		private protected virtual void UpdateColor()
 		{
 			var textColor = Element.TextColor;
 
@@ -258,7 +258,7 @@ namespace Xamarin.Forms.Platform.iOS
 			Control.AdjustsFontSizeToFitWidth = Element.OnThisPlatform().AdjustsFontSizeToFitWidth();
 		}
 
-		protected internal virtual void UpdateFont()
+		private protected virtual void UpdateFont()
 		{
 			if (initialSize == CGSize.Empty)
 			{
@@ -306,7 +306,7 @@ namespace Xamarin.Forms.Platform.iOS
 				Control.SecureTextEntry = Element.IsPassword;
 		}
 
-		protected internal virtual void UpdatePlaceholder()
+		private protected virtual void UpdatePlaceholder()
 		{
 			var formatted = (FormattedString)Element.Placeholder;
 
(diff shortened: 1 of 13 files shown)


Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
```diff
@@ -239,7 +239,7 @@ void UpdateAlignment()
 			Control.TextAlignment = Element.HorizontalTextAlignment.ToNativeTextAlignment(((IVisualElementController)Element).EffectiveFlowDirection);
 		}
 
-		protected internal virtual void UpdateColor()
+		private protected virtual void UpdateColor()
```
kingces95: Just need to sanity check that this member was not exposed publicly in a previous release.

PR #5432, opened 2019-03-03
[iOS] Check new element before creating placeholder label

hartez:
### Description of Change ###

CreatePlaceholderLabel is being called even if the new Element is null, which causes a crash on on dispose.

### Issues Resolved ### 

- fixes crash when disposing Editor

### API Changes ###

 None

### Platforms Affected ### 

- iOS

None

Not applicable

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/EditorRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/EditorRenderer.cs
@@ -31,11 +31,14 @@ namespace Xamarin.Forms.Platform.iOS
 
 		protected override void OnElementChanged(ElementChangedEventArgs<Editor> e)
 		{
-			// create label so it can get updated during the initial setup loop
-			_placeholderLabel = new UILabel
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.iOS/Renderers/EditorRenderer.cs
```diff
@@ -39,7 +39,10 @@ protected override void OnElementChanged(ElementChangedEventArgs<Editor> e)
 
 			base.OnElementChanged(e);
 
-			CreatePlaceholderLabel();
+			if (e.NewElement != null)
```
PureWeen: Maybe add this check above where it creates the label as well? or just check if the label is not null?

PureWeen:
UI Tests for this passed on 3.6 and when merged to 4.0

https://devdiv.visualstudio.com/DevDiv/_releaseProgress?releaseId=269614&_a=release-pipeline-progress

PR #5448, opened 2019-03-04
[Android]Fix for frame renderer's not disposing drawable bug #5446

### LINKED ISSUE #5446: [Android]FrameRenderer is creating new FrameDrawable every time BackgroundColor changes without disposing
### Description
When the BackgroundColor or CornerRadius properties of a Frame are changed, the FrameRenderer (not the FastRenderer), creates a new instance of the FrameDrawable without disposing the previous instance.

For my project, we have multiple frames that have blinking background colors and over a couple minutes, the blinking visibly slows down, and if left long enough, the app crashes. Because the FrameDrawable isn't being disposed, the Bitmap is still using memory and the listener for the Frame's property changes is still listening and still drawing the background to the bitmap.

### Steps to Reproduce

1. Ensure the Android Activity is not AppCompat
     - AppCompat is using the FastRenderer which doesn't have this issue
2. Add a couple frames to a page
3. Set a timer to fire every 250ms and to change the background color every time it fires

### Expected Behavior
The frames should keep blinking at the same frequency (every 250ms)

### Actual Behavior
After a couple minutes, the blinking of the frames slows down. The Android Choreographer will start to log that the UI thread is doing too much and that frames are being skipped. Eventually the app will crash due to running out of memory

### Basic Information

- Version with issue: 3.4.0 through 3.5.0
- Last known good version: 3.3.0.912540 (it seems this issue was introduced with pull request #3933)
- IDE: Visual Studio 2017 (15.9.5)
- Platform Target Frameworks:
  - Android: 8.1
- Affected Devices: Emulator

### Reproduction Link
[App.zip](https://github.com/xamarin/Xamarin.Forms/files/2928346/App.zip)

sstragier:
### Description of Change ###
Modified the Android FrameRenderer to not create a new FrameDrawable every time BackgroundColor and CornerRadius changes. This is done by storing the current FrameDrawable so that it can be disposed or re-set as the Background.

In some case, the base class changes the Background so that it is not longer the FrameDrawable, this was causing the corner radius to disappear. Now, if the base class changes the Background, the renderer just sets it back to the current FrameDrawable.

I could not identify a way to automate the testing for this fix.

### Issues Resolved ### 
- fixes #5446

### API Changes ###
 None

### Platforms Affected ### 
- Android

None

Not applicable

### Testing Procedure ###
Use the attached project in #5446 and update the Xamarin.Forms nuget reference to use this fix. Run the project on an Android device and emulator. After running for a few minutes, the blinking did not slow and the app did not crash.

### PR Checklist ###

- [ ] Has automated tests

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/FrameRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/FrameRenderer.cs
@@ -11,7 +11,7 @@ namespace Xamarin.Forms.Platform.Android
 	public class FrameRenderer : VisualElementRenderer<Frame>
 	{
 		bool _disposed;
-		private FrameDrawable _drawable;
+		FrameDrawable _drawable;
 		readonly MotionEventHelper _motionEventHelper = new MotionEventHelper();
 
 		public FrameRenderer(Context context) : base(context)
@@ -54,7 +54,7 @@ namespace Xamarin.Forms.Platform.Android
 		{
 			base.OnElementChanged(e);
 
-			if (e.NewElement != null && e.OldElement == null)
+			if (e.NewElement != null && e.NewElement != e.OldElement)
 			{
 				_motionEventHelper.UpdateElement(e.NewElement);
 
--- a/Xamarin.Forms.Platform.Android/Renderers/FrameRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/FrameRenderer.cs
@@ -4,15 +4,14 @@ using Android.Content;
 using Android.Graphics;
 using Android.Graphics.Drawables;
 using Android.Views;
-using AButton = Android.Widget.Button;
 using ACanvas = Android.Graphics.Canvas;
-using GlobalResource = Android.Resource;
 
 namespace Xamarin.Forms.Platform.Android
 {
 	public class FrameRenderer : VisualElementRenderer<Frame>
 	{
 		bool _disposed;
+		private FrameDrawable _drawable;
 		readonly MotionEventHelper _motionEventHelper = new MotionEventHelper();
 
 		public FrameRenderer(Context context) : base(context)
@@ -32,6 +31,12 @@ namespace Xamarin.Forms.Platform.Android
 
 			if (disposing && !_disposed)
 			{
+				// The background may or may not match the drawable, so ensure that it is only disposed once
+				if (Background != _drawable)
+				{
+					_drawable?.Dispose();
+				}
+
 				Background.Dispose();
 				_disposed = true;
 			}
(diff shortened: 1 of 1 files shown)


sstragier:
The FrameRender's OnElementChanged method has the following code: `if (e.NewElement != null && e.OldElement == null)`
which is updating the motion event helper and setting the drawable. I was wondering if it should be changed to:
`if (e.NewElement != null)` or `if (e.NewElement != null && e.NewElement != e.OldElement)`
that way, if the element is changed without being set to null first, the motion event helper and drawable would be updated correctly.

I didn't want to just change it since I'm not that familiar with how this method is being called and what the motion event helper is doing to know if there would be an issue with the suggested changes above.

Xamarin.Forms.Platform.Android/Renderers/FrameRenderer.cs
```diff
@@ -4,15 +4,14 @@
...
-using AButton = Android.Widget.Button;
 using ACanvas = Android.Graphics.Canvas;
-using GlobalResource = Android.Resource;
 
 namespace Xamarin.Forms.Platform.Android
 {
 	public class FrameRenderer : VisualElementRenderer<Frame>
 	{
 		bool _disposed;
+		private FrameDrawable _drawable;
```
samhouts:
```suggestion
		FrameDrawable _drawable;
```

samhouts:
@sstragier That sounds reasonable. Give it a try, and we'll see how the UI tests react to it. Thanks!

Xamarin.Forms.Platform.Android/Renderers/FrameRenderer.cs
sstragier: Thanks for catching that, fixed!

kingces95:
> In some case, the base class changes the Background so that it is not longer the FrameDrawable

And when this happens the Background is the same as the `BackgroundColor` remains the same as before just without the `CornerRadius`? So that's why we can just whack the `Background`?

sstragier:
@kingces95 Yes, it's my understanding that the `Background` being set by the base class just has the `BackgroundColor` without the `CornerRadius`. If you're asking if the `Background` that's being replaced needs to be disposed, that's a good question that I don't know the answer to.

I wasn't sure if there would be any side-effects of disposing the `Background` being replaced (e.g. if the base class still has a reference to the drawable). So I decided to leave that as-is, since the previous code was not disposing the replaced `Background`.

Let me know if you'd like me to update the PR to dispose the `Background` before replacing it.

kingces95:
Let's see if this alternative change will pass test... https://github.com/xamarin/Xamarin.Forms/pull/5617

sstragier:
@kingces95 Sounds good. Also, I think the `CornerRadius` property check in `FrameRenderer.OnElementPropertyChanged` could be removed since this property is specific to `Frame` and won't be on the base class. Meaning the `Background` would only need to be overwritten when the `BackgroundColor` is changed. The `FrameRenderer` is already listening to changes to `BackgroundColor`, `BorderColor`, and `CornerRadius` in order to update the drawable as needed.

rmarinho:
Can we check if #5617 fixed this issue?

rmarinho:
We don't need this #5617 fixed this i test with nightly on #5446 

Xamarin.Forms.4.0.1.346818-nightly.

PR #5449, opened 2019-03-04
remove covariant out on IElementConfiguration

PureWeen:
### Description of Change ###
Currently there appears to be an issue with UWP where a DataContext that implements a covariant generic constrained interface performs very poorly

If you create a native UWP app and use the following ViewModel structure
```C#
	public class ViewModel : IElementConfiguration<Element>
	{
		public string CompositionName { get; set; } = "test class";
	}

	public class Element
	{

	}

	// Remove the out on the generic or the constraint on the generic and rerun the application
	public interface IElementConfiguration<out TElement> where TElement : Element
	{
	}


	uwpListView.ItemsSource = Enumerable.Range(0,1000).Select(x=> new ViewModel()).ToList();
```

You will see the same performance degradation

- Closes #5235

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS
- Android
- UWP

### Testing Procedure ###
- Verify you can run UWP UI Tests

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/IElementConfiguration.cs
+++ b/Xamarin.Forms.Core/IElementConfiguration.cs
@@ -1,7 +1,8 @@
 
 namespace Xamarin.Forms
 {
-	public interface IElementConfiguration<out TElement> where TElement : Element
+	// Don't make this generic covariant as it causes UWP performance to tank
+	public interface IElementConfiguration<TElement> where TElement : Element
 	{
 		IPlatformElementConfiguration<T, TElement> On<T>() where T : IConfigPlatform;
 	}


PureWeen:
From what I can tell this isn't an ABI break but more of a compiler break
The IL difference between the interfaces with covariance is

Covariant
```
.class interface nested public auto ansi abstract IElementConfiguration`1<+T>
{
} // end of class IElementConfiguration`1
```

Not Covariant
```
.class interface nested public auto ansi abstract IElementConfigurationNotOut`1<T>
{
} // end of class IElementConfigurationNotOut`1
```

I ran a few tests by creating a library based on the covariant version and then performed some casting operations that would only work 

I checked the IL for the concrete implementations of the interfaces and the objects look the same. It doesn't look like the covariance doesn't cause anything additional to exist on the class itself

Once the tests pass this should be fine to merge into 4.0.0

The main place people would hit issues is if they were using the covariance feature in their code at which point their code won't compile and they'll have to adjust but this doesn't look like a runtime break

PureWeen:
pinging @kingces95  in case he has any additional thoughts

samhouts:
Should we point this to 3.6.0?

PureWeen:
@samhouts the last we all chatted about this we were saying to just target 4.0.0 since this is somewhat breaking

samhouts:
@PureWeen You're right! 

I haven't yet seen customer reports indicating that the performance degradation has been a problem. My hope is that those affected customers are comfortable on 3.5 until 4.0 is released. Unfortunately, in order to fix this issue on 3.6, we have two bad options:

1. Backport this fix to 3.6.0 and accept the possible compile break (seems low risk but does violate our policy on no breaking changes in service releases), or
2. Pull #4002 out of 3.6.0 and put it into 4.0.0 with this fix in place. Since this is removing a feature from 3.6.0, this is a more serious breaking change, and I don't think it's a viable option.

samhouts:
@PureWeen Oh, but we still need to retarget to 4.0.0. :)

PureWeen:
I don't think the issue needs to be fixed in 3.6 

The main area where this causes major issues is with UI Tests and large lists. From a usability standpoint it just results in a slight delay of renderering if you scroll really fast on the list. It doesn't actually affect how fast the user can scroll. So I think the user impact of this issue is low enough to where we can just leave it for 4.0.0

rebased

rmarinho:
Did the UITests on UWP worked fine @PureWeen ?

PureWeen:
@rmarinho  yes
Anytime I run UWP UI tests now I just apply this fix before running them. So I've ran through the tests a good amount of times now with this fix applied

PR #5460, opened 2019-03-05
Fix [Android] WebView Navigating Cancel property not working with custom scheme

### LINKED ISSUE #4891: [Android] WebView Navigating Cancel property not working with custom scheme
### Description

We utilize custom URL schemes heavily within our app for app communication with an external website's javascript, so that the website can invoke behavior in-app and vice versa.

Since Xamarin Forms 3.3.0, the `Navigating`/`Navigated` event behavior has changed such that this is broken due to several issues found.

This is a very similar issue to #4223 but is different because their issue is occurring on first load, not on navigation events.

### Steps to Reproduce

See example repository. This same code was tested and working with 3.2.0, and when updating to 3.3.0, behavior is broken

1. Create a WebView - https://github.com/kensykora/xamarin-forms-4223/blob/master/XamarinFormsIssue_4223/MainPage.xaml
2. Bind the `WebView`'s starting content to some HTML that contains a link to a custom URL scheme. - https://github.com/kensykora/xamarin-forms-4223/blob/master/XamarinFormsIssue_4223/MainPage.xaml.cs#L20-L23
2. Bind events to `Navigating` / `Navigated`. In the `Navigating` event, look for a custom scheme and cancel the request if navigation occurs with the custom scheme, such that you can intercept the URL and perform some behavior in-app. https://github.com/kensykora/xamarin-forms-4223/blob/master/XamarinFormsIssue_4223/MainPage.xaml.cs#L26-L35
4. Run the app
5. Tap on the link with the custom scheme

### Expected Behavior

Navigation does not occur. App is able to get a `Navigating` callback and successfully cancel the event. If `Navigated` event does fire, the underlying webview does not actually navigate, since `Cancel` was invoked (in 3.2.0, `Navigated` does not fire)

### Actual Behavior

Sporadic: as you can see in the screen capture, 3 things occur incorrectly:

1) Webview shows android error saying it's an unknown scheme. A navigation does occur despite us trying to cancel it.
2) Multiple `Navigated` events (failure) occur. I would expect just 1, if trying to stay backwards compatible (which using semver, a minor version 3.2 -> 3.3 should!!!) there should be no `Navigated` events called to maintain backward compatibility
3) in some cases the underlying `WebView` seems to stop responding to taps.

### Basic Information

- Version with issue: 3.3.0.967583
- Last known good version: 3.2.0.871581
- IDE: Xamarin 
- Platform Target Frameworks:
  - Android: 8.1 (API 27)
- Nuget Packages: none other than Xamarin Forms and Support Packages
- Affected Devices: Tested on a simulator, but observed on all android devices in another app

### Screenshots

3.2.0 Behavior - Working as expected:

3 2 0-behavior

3.3.0 Behavior - Broken

3 3 0-behavior

### Reproduction Link

https://github.com/kensykora/xamarin-forms-4223

[XamForms4223-3.2.0.zip](https://github.com/xamarin/Xamarin.Forms/files/2724791/XamForms4223-3.2.0.zip)

[XamForms4223-3.3.0.zip](https://github.com/xamarin/Xamarin.Forms/files/2724795/XamForms4223-3.3.0.zip)

Edit: Clarified expectations to maintain backward compatibility

### LINKED ISSUE #4919: Cancel Webview Navigation using "e.cancel = true" not working
Have two separate Android apps running, both using a webview.  

If the webview receives a URL which contains a "target=blank" parameter, this would open the link in the device's browser using Device.OpenUri(uri), and would cancel the navigation within the webview in the app.   This was working fine using version 3.2.0 (have just rolled back and can confirm this still works), but after upgrading to version 3.4.0, in both apps, the link does open in the external browser, BUT also navigates within the app too. Therefore, it appears the "e.cancel = true;" seems to have ceased functioning here.

### Steps to Reproduce

1.   Build a webview with code as follows:

```
//get the global webview set at app start
WebView WV = new WebView();

//create a webview and fill screen with it
WV.Source = url;
WV.HorizontalOptions = LayoutOptions.Fill;
WV.VerticalOptions = LayoutOptions.FillAndExpand;

//set the webview so that any links containing a querystring parameter of target=blank, then open these in the device's external browser, not in the webview
//    (NOTE - this NOT the target attribute of the <a> link, it needs to be in the querystring of the URL
WV.Navigating += (object sender, WebNavigatingEventArgs e) =>
    {
        if (e.Url.ToLower().Contains("target=blank") || e.Url.ToLower().Contains("target=_blank") || e.Url.ToLower().Contains("webapp/login"))
            {
                var uri = new Uri(e.Url);                            
                Device.OpenUri(uri);
                e.Cancel = true;
            }
                else
            {
                e.Cancel = false;
            };
        };
```

2.  Try two links (called from JS) from a remote website: eg - 

```
<button onclick="window.location.assign('https://www.google.com/?target=blank'); return false;">Google Out</button>

<button onclick="window.location.assign('https://www.google.com/'); return false;">Google In</button>
```

### Expected Behavior
The first link will open in the external browser and navigation stops in the app.  The second link will open in the app.

### Actual Behavior
in 3.2, it works fine.  In 3.4, the first opens both in the external browser, but also in the app. Navigation is not cancelled.

### Basic Information

- Version with issue:  3.4 and 3.3
- Last known good version:  3.2 
- IDE:   Visual Studio 2017 v. 15.9.4
- Platform Target Frameworks: 
  - Android:  9.0

brackert:
Updates Xamarin.Forms.Platform.Android.FormsWebViewClient

This was broken with release 3.3.0.

Notes:
Depending on the API version, Android's StopLoading does not work as expected/consistently which is used to cancel the load in OnPageStarted. This is why 3.3.0 created this 'custom scheme' problem.

In addition, Android's ShouldOverrideUrlLoading is NOT executed when the WebView.Source is explicitly set. This is why 3.3.0 fixed the problem of Navigating not firing when the Source is set.

### Issues Resolved ###
fixes #4891 
fixes #4919

### API Changes ###
None

### Platforms Affected ###
Android

None

Not applicable

### Testing Procedure ###
see Issue4891.cs

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4891.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4891.cs
@@ -1,0 +1,105 @@
+﻿using System;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.Forms.Core.UITests;
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if UITEST
+	[Category(UITestCategories.ManualReview)]
+#endif
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 4891, "[Android] WebView Navigating Cancel property not working with custom scheme", PlatformAffected.Android)]
+	public class Issue4891 : TestContentPage // or TestMasterDetailPage, etc ...
+	{
+		Button _back;
+		WebView _myWebView;
+		Label _log;
+		ScrollView _logScrollView;
+		
+		protected override void Init()
+		{
+
+			_back = new Button
+			{
+				Text = "Back"
+			};
+			_back.Clicked += Back_Clicked;
+
+			_myWebView = new WebView
+			{
+				HorizontalOptions = LayoutOptions.StartAndExpand,
+				VerticalOptions = LayoutOptions.Start,
+				HeightRequest = 240
+			};
+			_myWebView.Navigating += MyWebView_Navigating;
+			_myWebView.Navigated += MyWebView_Navigated;
+			_myWebView.Source = new HtmlWebViewSource()
+			{
+				Html = "<html><body>Click on the link below. Expected results:<br/>1. Navigating event logged.<br/>2. Navigated event NOT logged.<br/>3. This page stays loaded in the WebView control.<br/><br/><a href='xamforms4223://custom'>Navigate to Custom xamforms4223 scheme</a></body></html>"
+			};
+
+			_log = new Label
+			{
+				VerticalOptions = LayoutOptions.StartAndExpand,
+				HorizontalOptions = LayoutOptions.FillAndExpand,
+				Text = ""
+			};
+
+			_logScrollView = new ScrollView
+			{
+				VerticalOptions = LayoutOptions.StartAndExpand,
+				HorizontalOptions = LayoutOptions.FillAndExpand,
+				Content = _log
+			};
+
+			Content = new StackLayout
+			{
+				Children =
+				{
+					_myWebView,
+					_back,
+					_logScrollView
+				}
+			};
+		}
+
+
+		void MyWebView_Navigating(object sender, WebNavigatingEventArgs e)
+		{
+			LogToScreen($"Navigating: {e.Url}");
+			if (e.Url.StartsWith("xamforms4223", StringComparison.OrdinalIgnoreCase))
+			{
+				LogToScreen("Caught custom scheme, cancelling navigation.");
+				e.Cancel = true;
+			}
+		}
+
+		void MyWebView_Navigated(object sender, WebNavigatedEventArgs e)
+		{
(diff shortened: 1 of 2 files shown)


TopsKing:
@samhouts Is this problem solved in the latest version 3.6?

samhouts:
@TopsKing No, the PRs are still pending review. Thanks for your patience!

simontocknell:
Any news, this is stopping me being able to launch my app on Android!

paymicro:
I took all the best of this PR and added to https://github.com/xamarin/Xamarin.Forms/pull/5458 Thanks @brackert excellent work 👍 
Also [fixed work on api 23](http://recordit.co/kBaeHLAFmH) and less.

paymicro:
Changes made to PR #5458

SebastianKruseWago:
Is this already fixed in the latest 3.6 NuGet package? Because I still get this error all the time.

paymicro:
@SebastianKruseWago this is fixed in `4.0.0.304370-pre8` nuget package.

SebastianKruseWago:
@paymicro Do you know it will make it into a future 3.6 release? I cannot update to 4.0 yet because some libraries I use require 3.6.

NaveedShaikh:
@brackert : I tried to create a custom renderer and triggered `ShouldOverrideUrlLoading` but the issue is, it doesn't delegate back to the previous controller, wherein I have to perform a certain task. Kindly help. Thanks.

PR #5462, opened 2019-03-05
[Android] ensure scrollbar has been initialized

### LINKED ISSUE #5461: [Android] ScrollView crashes when setting ScrollbarFadingEnabled to false in Custom Renderer
### Description

If an Android View sets *ScrollbarFadingEnabled* to false then the drawable on a ScrollView won't get initialized which will cause a crash during Draw

From what I can tell looking at the android source two scenarios will initialize the scrollbars on a view

1) if the view is inflated with scrollbars set to anything but NONE
https://android.googlesource.com/platform/frameworks/base/+/a175a5b/core/java/android/view/View.java#2956
2) if the user calls awakenScrollBars
https://android.googlesource.com/platform/frameworks/base/+/a175a5b/core/java/android/view/View.java#8384

Since we don't inflate the views with attributes we need to call awakenScrollBars so the drawable initializes

I've found that in some cases this crash will happen without a Custom Renderer but I haven't been able to reproduce that scenario in our Control Gallery. 

This issue didn't happen when we inherited from ScrollView and not NestedScrollView. NestedScrollView is a totally different implementation than ScrollView so my guess is that the loading behavior or how those drawables are used changed

### Steps to Reproduce

1. Create a Custom Renderer that inherits from ScrollViewRenderer
2.  set ScrollbarFadingEnabled in the constructor or in the override of OnElementChanged
3. App will now crash when using that scrollview

### Expected Behavior
Shouldn't crash

### Actual Behavior
Crashes

### Basic Information

- Version with issue: 3.5+
- Last known good version: 3.4

PureWeen:
### Description of Change ###
Added code to force the underlying scrollbars on a NestedScrollView control to initialize to avoid crashes during the draw phase 

### Issues Resolved ### 

- fixes #5461

### Platforms Affected ### 
- Android

### Testing Procedure ###
- there are automated tests

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.ControlGallery.Android/CustomRenderers.cs
+++ b/Xamarin.Forms.ControlGallery.Android/CustomRenderers.cs
@@ -63,7 +63,7 @@ namespace Xamarin.Forms.ControlGallery.Android
 	{
 		public ScrollbarFadingEnabledFalseScrollViewRenderer(Context context) : base(context)
 		{
-			// I do a cast here just so this will fail just to be sure we don't chance the base types
+			// I do a cast here just so this will fail just to be sure we don't change the base types
 			var castingTest = (global::Android.Support.V4.Widget.NestedScrollView)this;
 			castingTest.ScrollbarFadingEnabled = false;
 		}
--- a/Xamarin.Forms.ControlGallery.Android/CustomRenderers.cs
+++ b/Xamarin.Forms.ControlGallery.Android/CustomRenderers.cs
@@ -25,6 +25,7 @@ using Android.Text.Method;
 using Xamarin.Forms.Controls.Issues;
 
 
+[assembly: ExportRenderer(typeof(Issue5461.ScrollbarFadingEnabledFalseScrollView), typeof(ScrollbarFadingEnabledFalseScrollViewRenderer))]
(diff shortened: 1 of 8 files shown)


Xamarin.Forms.ControlGallery.Android/CustomRenderers.cs
```diff
@@ -58,6 +59,16 @@ public NonAppCompatSwitchRenderer(Context context) : base(context)
 		}
 	}
 
+	public class ScrollbarFadingEnabledFalseScrollViewRenderer : ScrollViewRenderer
+	{
+		public ScrollbarFadingEnabledFalseScrollViewRenderer(Context context) : base(context)
+		{
+			// I do a cast here just so this will fail just to be sure we don't chance the base types
```
samhouts: 👍

hartez: chance -> change

PureWeen:
```suggestion
			// I do a cast here just so this will fail just to be sure we don't change the base types
```

Issue #5465, opened 2019-03-06
[UWP] Issue2927 failing and Tapped is passing the wrong value in for Item


PureWeen:
### Description
This is a little tricky to explain or I just confused myself while working on this

https://github.com/xamarin/Xamarin.Forms/pull/3400

Appears to have fixed https://github.com/xamarin/Xamarin.Forms/issues/4901 because it made it so that two (maybe more) wrongs made a right

If you run *Issue2927* on master it will crash because PR https://github.com/xamarin/Xamarin.Forms/pull/3400 made it so the BindingContext gets passed into the ItemTapped event args but what should be passed into ItemTapped is the View not the BindingContext

The reason this seems to have worked is that there is a bug on UWP where if the BindingContext implements a Cell then the BindingContext is used as the template rather than the template.  *Issue 935* masks this behavior because the ItemSource is using the same cell type as the template.  

If you run *Issue 935* with pre PR *3400* changes the Tapped event does in fact fire. The problem is that it's firing against the incorrectly renderered Cell Control so the changes don't take effect on the view.

From what I can tell this happens because the CellControl on UWP doesn't use the TemplatedItems list to create controls it just goes right the template on the list View

https://github.com/xamarin/Xamarin.Forms/blob/5e133758ab6cbf9b953ebc159eb581b3907cee3d/Xamarin.Forms.Platform.UAP/CellControl.cs#L314

Where as the other platforms go via the TemplatedItems List
https://github.com/xamarin/Xamarin.Forms/blob/4cfa82569dd74858849eff6bd30cdf1b6ce7552c/Xamarin.Forms.Platform.Android/Renderers/ListViewAdapter.cs#L539

I have a branch I've started working on over here
https://github.com/xamarin/Xamarin.Forms/tree/fix_uwp_cell

Some of the ideas are rough and I don't think it works with grouping or TableViews yet
But it does fix the issues presented by  *Issue2927* and  *Issue 935* 

### Steps to Reproduce

*Issue2927* crashes
if you fix the crash on *Issue2927* 
the make sure *Issue 935* still works

### Expected Behavior
*Issue 935* and *Issue2927*  should be able to simultaneously work

### Actual Behavior
they don't

paymicro:
It seems that in UWP to work on both Android and iOS we need a lot of changes in ListViewRenderer and CellControl. There is a possibility that even more errors will appear in the aftermath...
[...]

PR #5476, opened 2019-03-06
[UWP] Roll back PR 3400 so the correct context is passed into Tapped

PureWeen:
### Description of Change ###

The description of why is summarized here
https://github.com/xamarin/Xamarin.Forms/issues/5465

### Platforms Affected ### 
- UWP

### Testing Procedure ###
- run Issue2927Test and it will no longer crash

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.UAP/ListViewRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/ListViewRenderer.cs
@@ -722,9 +722,9 @@ namespace Xamarin.Forms.Platform.UWP
 			List.SelectedIndex = index;
 		}
 
-		void OnListItemClicked(int index, Cell cell = null)
+		void OnListItemClicked(int index)
 		{
-			Element.NotifyRowTapped(index, cell);
+			Element.NotifyRowTapped(index);
 			_itemWasClicked = true;
 		}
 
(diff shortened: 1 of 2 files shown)


Xamarin.Forms.Platform.UAP/ListViewRenderer.cs
```diff
@@ -735,7 +735,7 @@ void OnListItemClicked(object sender, ItemClickEventArgs e)
 				var templatedItems = TemplatedItemsView.TemplatedItems;
 				var selectedItemIndex = templatedItems.GetGlobalIndexOfItem(e.ClickedItem);
 
-				OnListItemClicked(selectedItemIndex, e.ClickedItem as Cell);
```
PureWeen: ClickedItem contains the BindingContext not the Cell.  It's only a cell if the BindingContext is also a Cell which isn't a case where we would still want to pass it in as the cell argument. *NotifyRowTapped* needs to just locate the Cell itself where the event fires. Which is also broken on UWP but this change makes it less broken.

This scenario was fairly edge case anyway because since ClickedItem is the BindingContext it will almost never be a Cell unless it's binding to itself or the BindingContext for some reason implements Cell (which it normally won't). So in almost all cases this cast will just be null and in the cases where it's not null then it's passing the wrong Cell into *NotifyRowTapped*

samhouts:
Should we point this at 3.6.0?

PureWeen:
@samhouts  no these changes are only in master

PureWeen:
@samhouts  wait but we do have a 4.0 branch now
let me target that

PureWeen:
@samhouts  alright retargeted

Issue #5473, opened 2019-03-06
Missing markup extensions cause parsing to fail


alanmcgovern:
In no-compile rendering mode we would expect this construct to not cause parsing to fail:

```
<Image>
    <Image.Source>
        <OnPlatform x:TypeArguments="ImageSource">
            <On Platform="iOS" Value="{local:Missing Test}" />
        </OnPlatform>
    </Image.Source>
</Image>
```

Currently we get a MarkupExtensionNotFound exception if the type really does not exist in the project.

If you do go ahead and create the class in your project:
```
	public class MissingExtension : IMarkupExtension
	{
		public object ProvideValue(IServiceProvider serviceProvider)
			=> throw new NotImplementedException();
	}
```

then instead you'll get this exception at runtime. My guess is that we've subclassed System.Object and it does not implement `IMarkupExtension` which is why we get the invalid cast:

```
[2019-03-06 13:44:22.1] ERROR: : iOSXamlRenderer had a rendering error: System.InvalidCastException: Specified cast is not valid.
<unknown method>
Xamarin.Forms.OnPlatform`1[T].op_Implicit (Xamarin.Forms.OnPlatform`1[T] onPlatform) in D:\a\1\s\Xamarin.Forms.Core\OnPlatform.cs:75
<unknown method>
System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) in /Library/Frameworks/Xamarin.iOS.framework/Versions/12.7.0.185/src/Xamarin.iOS/mcs/class/corlib/System.Reflection/MonoMethod.cs:305
```

We should gracefully handle both of these scenarios. /cc @TimBarham

> VS bug [#812516](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/812516), VS bug [#812888](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/812888)

TimBarham:
I'm adding a test for the first scenario. I can't repro the second scenario... when unbuilt, I get no error. When built, I hit the `NotImplementedException` in the `ProvideValue()` method, as expected. This was on Android... I'll double check iOS shortly.

TimBarham:
Although I guess we also want to fail silently when `ProvideValue()` throws?

TimBarham:
See #5484 for the test.

[...]

PR #5484, opened 2019-03-07
Adds a test scenario for another Previewer Markup Extension issue

TimBarham:
See #5473


TimBarham:
FYI @alanmcgovern @drewgillies @samhouts

StephaneDelcroix:
that's not a markup extension issue, but a type-finding one

samhouts:
@TimBarham Can you verify for me that #5485 fixes the issue? Thanks!

TimBarham:
@samhouts - @StephaneDelcroix's PR contains the same test case from this PR, so I'm confident it does.

PR #5496, opened 2019-03-08
[Android] NamedSize now returns defaultValue instead of -1 if parsing from resource fails

TimBarham:
### Description of Change ###

In the previewer, `TryGetTextAppearance()` returns `-0.5` for all values. Return default value if we get a size of 0 or less.

### Issues Resolved ### 

- fixes https://devdiv.visualstudio.com/DevDiv/_queries/edit/763877

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

None

Before:

image

After:

image

### Testing Procedure ###
Manual testing in previewer.

### PR Checklist ###

- [ ] Has automated tests (we will test this in the previewer)

### DIFF
--- a/Xamarin.Forms.Platform.Android/Forms.cs
+++ b/Xamarin.Forms.Platform.Android/Forms.cs
@@ -549,9 +549,9 @@ namespace Xamarin.Forms
 			{
 				double myValue;
 
-				if (TryGetTextAppearance(themeDefault, out myValue))
+				if (TryGetTextAppearance(themeDefault, out myValue) && myValue > 0)
 					return myValue;
-				if (TryGetTextAppearance(deviceDefault, out myValue))
+				if (TryGetTextAppearance(deviceDefault, out myValue) && myValue > 0)
 					return myValue;
 				return defaultValue;
 			}


TimBarham:
@alanmcgovern @pierceboggan FYI

alanmcgovern:
LGTM!

My understanding is that the -0.5 is because the API returned -1 for error and we did a pixels/dpi conversion on it to make it -0.5.

I don't know if we want to create a separate task to delve into why it's returning -1 for failure in the first place.

TimBarham:
Yeah, I was thinking this would be "good enough" for the Previewer for now, but would also be good to figure out if we can get real values. I'm not set up for debugging into Forms code, however... I'll chat to @PureWeen (or whoever is appropriate) about this early next week.

alanmcgovern:
yup - the fallback path clearly works when we treat errors as errors :)

StephaneDelcroix:
approving this, but I think it shouldn't be merged in 3.6.0. there's a risk of breaking applications. @samhouts: your call

alanmcgovern:
I'd like to understand the risk of breaking applications.

From my reading of the patch I think it can only have an affect in the case where a negative font size is being returned. Is there a scenario where giving an application a 'normal' font size instead of a negative font size would break them? Or, in other words, is there a scenario where we expect negative font sizes to be returned?

samhouts:
Hmm, let me take a few minutes to look into this to see what the implications may be. I'd rather get the fix into 3.6 if we can, but I need to evaluate if it's actually breaking or not. As a policy, we don't allow breaking behavior changes into SRs if we can help it! Thanks.

samhouts:
I'm going to say that the `-1` is an old, old bug that I caused. It was supposed to fall back to the default value if it was unable to determine the font size, but instead it allows the -1 error to flow through. Few customers may be depending on this to fail now, but I hope that they are just frustrated that it didn't just work in the first place and welcome this fix.

TimBarham:
Thanks @samhouts!

TimBarham:
@samhouts - are those test failures something we care about? Can we merge?

samhouts:
@TimBarham They seem to be new, so we'll need to take a look. :/

samhouts:
Failed tests aren't related or even reproducible locally. 👍

PR #5512, opened 2019-03-11
[nuspec] Start using the 28.x versions of Android support

mattleibow:
### Description of Change ###

Use the 28.x versions of the AndroidSupport packages everywhere.
These packages will work on all Android versions down to Android 6.0. This means that there will no longer be package version conflicts - regardless of the target framework.

### Issues Resolved ### 

- fixes NuGet issues

### API Changes ###

 None

### Platforms Affected ### 

- Android

None

Not applicable

### Testing Procedure ###

The packages should work as always. No visible changes other than it should always install the 28.x regardless of the framework version.

### PR Checklist ###

- [ ] Has automated tests 


PureWeen:
As long as this matches up with whatever the latest public stable release templates are set to then this should be fine

PR #5550, opened 2019-03-14
[X] Allow empty property xaml-elements

StephaneDelcroix:
### Description of Change ###

Allow empty properties.

NOTE: before merging this, we need to check if UWP and WPF allows this construct.

### Issues Resolved ### 

- fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/817710
- reverts #4582

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 


TimBarham:
UWP and WPF both accept an empty element tag as equivalent to an element with no content.

StephaneDelcroix:
> you actually get error dialogs pop up, which I have never seen before.

looks like a `Debug.Assert()` failing

StephaneDelcroix:
> UWP and WPF both accept an empty element tag as equivalent to an element with no content.

what example have you tried, @TimBarham ? empty properties, or empty objects ? I want to be super sure here...

samhouts:
@StephaneDelcroix I can do this in a UWP project:

```
<Page
    x:Class="EmptyObj.MainPage"
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    xmlns:local="using:EmptyObj"
    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
    mc:Ignorable="d"
    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">

    <Page.Resources />

    <Grid >

    </Grid>
</Page>
```

This seems analogous to what we're doing here. 

And if I'm not totally nuts, the WPF XAML parser also allows both props and objs to be self-terminating and/or empty. 

https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Parser/XamlScanner.cs#L265
https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Parser/XamlScanner.cs#L408

So I think this is a good change!

TimBarham:
I specifically tried the Resources property, like @samhouts' example. Other properties (like the `Content` property of a `Button`, if I recall) rejected both `<Element.Property />` and `<Element.Property></Element.Property>` styles, but that was because the property itself rejected an empty value - both constructs were still treated identically (because semantically they mean exactly the same thing).

PR #5562, opened 2019-03-14
[Android] Fix issue when Message is null on Android API19

rmarinho:
### Description of Change ###

Fix issue when Message is null on Android API19 introduced on PR #5462 

### API Changes ###

 
 None

### Platforms Affected ### 

- Android

None

Not applicable

### Testing Procedure ###

Run gallery and go to test #5461 or a test with a Scrollview on API19 

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/Extensions/ScrollViewExtensions.cs
+++ b/Xamarin.Forms.Platform.Android/Extensions/ScrollViewExtensions.cs
@@ -41,5 +41,13 @@ namespace Xamarin.Forms.Platform.Android
 
 			scrollView.ScrollBarsInitialized = true;
 		}
+
+		internal static bool HandleDrawException(this Java.Lang.NullPointerException npe)
+		{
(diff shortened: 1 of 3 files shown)


PureWeen:
I need to play with this a little bit on API 19

The point of the catch is to fix it so the exception only gets thrown that first time and then it fixes the underlying cause but API 19 just perpetually throws the exception from the Draw method 

In theory the exception should never be thrown based on code I added so need to just dive in a little bit on API 19 see if we can fix for this condition

PureWeen:
@samhouts @rmarinho   It looks like calling Awaken on KitKat puts the NestedScrollView in some state where it perpetually throws exceptions.

I tested this on API 21 and API 26 as well just to make sure after KitKat my fix still applies and it looks like it does.

I just removed the When condition completely though because I noticed as you move through the apis the exception message is different :-/  With some it throws an exception during the SetTint call with others it's during the mutate call

PR #5572, opened 2019-03-15
Ignore FindByName exception when running in previewer

TimBarham:
### Description of Change ###

When running in the previewer, some types are replaced.

If a replaced type has an `x:Name` attribute, this will result in an `InvalidCastException` when `FindByName()` tries to cast it to the original type. So return `default(T)` in this scenario (this may just result in an exception later trying to use the result of this call, but it is at least less likely to be in the constructor).

### Issues Resolved ### 

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/821129

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###
Tested in previewer. Added unit test that covers the scenario.

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/NameScopeExtensions.cs
+++ b/Xamarin.Forms.Core/NameScopeExtensions.cs
@@ -1,3 +1,4 @@
+using System;
 using Xamarin.Forms.Internals;
 
 namespace Xamarin.Forms
@@ -5,7 +6,16 @@ namespace Xamarin.Forms
 	public static class NameScopeExtensions
 	{
 		public static T FindByName<T>(this Element element, string name)
-		=> (T)element.FindByName(name);
+		{
+			try {
+				return (T)element.FindByName(name);
+			}
+			catch (InvalidCastException) {
+				if (DesignMode.IsDesignModeEnabled)
+					return default(T);
+				throw;
+			}
+		}
 
 		internal static T FindByName<T>(this INameScope namescope, string name)
 			=> (T)namescope.FindByName(name);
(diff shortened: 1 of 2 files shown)


TimBarham:

Xamarin.Forms.Core/NameScopeExtensions.cs
```diff
@@ -1,11 +1,21 @@
...
 	public static class NameScopeExtensions
 	{
 		public static T FindByName<T>(this Element element, string name)
-		=> (T)element.FindByName(name);
+		{
+			try {
+				return (T)element.FindByName(name);
+			}
+			catch (InvalidCastException) {
+				if (DesignMode.IsDesignModeEnabled)
```
StephaneDelcroix: `DesignMode` doesn't mean the previewer is plugged in

TimBarham: Do you mean... `IsDesignModeEnabled` could be true without the previewer, or `IsDesignModeEnabled` could be false with the previewer?

It is only set if `XamlFileProvider` or `ResourceProvider2` are set. Are there non-previewer scenarios these are used?

TimBarham:
Can we get traction on this? This is a bad bug, and I can't see any way to fix it on the previewer side. @StephaneDelcroix can you elaborate on your concerns on using `DesignMode.IsDesignModeEnabled`?

@samhouts and @alanmcgovern thoughts?

samhouts:
Stephane is going to revisit this one. He'll have more info soon! Thanks!

StephaneDelcroix:
This is a really annoying case.

We catch (or try to catch ) exceptions in the context of Create(xaml, true), but here the exception is thrown outside of it.

The only valuable global we could use is `ResourceLoader.ExceptionHandler`

PR #5582, opened 2019-03-15
Fix BindableLayout's incorrect layout.children.add -> use insert

### LINKED ISSUE #5579: BindableLayout displays items in the wrong order
### Description
BindableLayout displays items in the wrong order if you insert items in a collection that implements INotifyCollectionChanged instead of just adding the items.

### Steps to Reproduce

1.  Create a view that uses a bindableLayout
2. Bind the bindablelayout to a collection that implements INotifyCollectionChanged
3. Add 5 items to the collection
4. Insert an item in the collection at position 2

### Expected Behavior
The newly created template for the item inserted at position 2 to appear on position 2

### Actual Behavior
The template for item 2 is shown at position 6 (end of list)

### Basic Information

- Version with issue: 3.6.0
- Affected Devices: All

csteeg:
### Description of Change ###

In the BindableLayout, when bound to a collection that implements INotifyCollectionChanged,
item inserts are always handled as Add

There is an extension method in Xamarin.Forms.Internals.NotifyCollectionChangedEventArgsExtensions that handles all the events correctly when you pass the callbacks.
I changed BindableLayout.cs to use this extension method when the collection changes.

Also, I removed passing the index to CreateView, since createview doesn't do anything with that.
And, in the reset, I replaced layout.children.clear() with CreateChildren().
CreateChildren clears the layout.children and then checks if it needs to create new ones.
Some libraries (eg DynamicData/ReactiveUI) trigger the reset after a batch update of the collection. It can be disputed if this correct behaviour, but Xamarin's listviews on all platforms support this behavior also, so I thought it would be fair to have BindableLayout support it as well. (It's a real small extra check)

### Issues Resolved ### 

- fixes #5579

### API Changes ###
None

### Platforms Affected ### 
All

Items will appear on the correct place instead of always at the end of the list

Not applicable

### Testing Procedure ###
Modified BindableLayoutTests.TracksInsert to simulate this situation

### PR Checklist ###

- [X] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/BindableLayoutTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/BindableLayoutTests.cs
@@ -49,10 +49,10 @@ namespace Xamarin.Forms.Core.UnitTests
 				IsPlatformEnabled = true,
 			};
 
-			var itemsSource = new ObservableCollection<int>();
+			var itemsSource = new ObservableCollection<int>() { 0, 1, 2, 3, 4 };
 			BindableLayout.SetItemsSource(layout, itemsSource);
 
-			itemsSource.Insert(0, 1);
+			itemsSource.Insert(2, 5);
 			Assert.IsTrue(IsLayoutWithItemsSource(itemsSource, layout));
 		}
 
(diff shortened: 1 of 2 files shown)


StephaneDelcroix:
you might want to rebase and retarget to 4.0.0

PR #5585, opened 2019-03-15
[Android] fall back to using native get identifier if reflection fails to retrieve drawable

PureWeen:
### Description of Change ###

This will try to retrieve a resource id by using the native GetIdentifier instead of reflection

### Platforms Affected ### 
- Android

### Testing Procedure ###
- Test any image code to make sure there are no breaking changes
- Install nuget into android project that references an image that's part of the Android project and test that it loads inside the previewer without a build

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/Platform.cs
+++ b/Xamarin.Forms.Platform.Android/Platform.cs
@@ -63,9 +63,9 @@ namespace Xamarin.Forms.Platform.Android
 
 		internal Platform(Context context, bool embedded)
 		{
-			PackageName = context.PackageName;
 			_embedded = embedded;
 			_context = context ?? throw new ArgumentNullException(nameof(context), "Somehow we're getting a null context passed in");
+			PackageName = context.PackageName;
 			_activity = context as Activity;
 
 			if (!embedded)
--- a/Xamarin.Forms.Platform.Android/Platform.cs
+++ b/Xamarin.Forms.Platform.Android/Platform.cs
@@ -26,6 +26,8 @@ namespace Xamarin.Forms.Platform.Android
 		, IPlatform
 #pragma warning restore
 	{
+
+		internal static string PackageName { get; private set; }
 		internal const string CloseContextActionsSignalName = "Xamarin.CloseContextActions";
 
 		internal static readonly BindableProperty RendererProperty = BindableProperty.CreateAttached("Renderer", typeof(IVisualElementRenderer), typeof(Platform), default(IVisualElementRenderer),
@@ -61,6 +63,7 @@ namespace Xamarin.Forms.Platform.Android
 
 		internal Platform(Context context, bool embedded)
 		{
+			PackageName = context.PackageName;
 			_embedded = embedded;
 			_context = context ?? throw new ArgumentNullException(nameof(context), "Somehow we're getting a null context passed in");
 			_activity = context as Activity;
--- a/Xamarin.Forms.Platform.Android/ResourceManager.cs
+++ b/Xamarin.Forms.Platform.Android/ResourceManager.cs
@@ -10,8 +10,6 @@ using Path = System.IO.Path;
 using Xamarin.Forms.Internals;
 using AndroidAppCompat = Android.Support.V7.Content.Res.AppCompatResources;
 using System.ComponentModel;
-using System.Collections.Generic;
-using System.IO;
 
 namespace Xamarin.Forms.Platform.Android
 {
--- a/Xamarin.Forms.Platform.Android/ResourceManager.cs
+++ b/Xamarin.Forms.Platform.Android/ResourceManager.cs
@@ -152,42 +152,34 @@ namespace Xamarin.Forms.Platform.Android
 
 		static int IdFromTitle(string title, Type type, Resources resource)
 		{
-			int id = 0;
-
-			if (title == null)
-				return id;
-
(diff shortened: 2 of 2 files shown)


Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -66,20 +68,20 @@ public static Bitmap GetBitmap(this Resources resource, FileImageSource fileImag
...
+			return BitmapFactory.DecodeResourceAsync(resource, IdFromTitle(name, DrawableClass, resource));
 		}
 
 		[Obsolete("GetDrawable(this Resources, string) is obsolete as of version 2.5. "
 			+ "Please use GetDrawable(this Context, string) instead.")]
 		[EditorBrowsable(EditorBrowsableState.Never)]
 		public static Drawable GetDrawable(this Resources resource, string name)
-		{
-			int id = IdFromTitle(name, DrawableClass);
+		{
```
samhouts:
```suggestion
		{
```

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -127,15 +130,46 @@ public static void Init(Assembly masterAssembly)
...
 			LayoutClass = masterAssembly.GetTypes().FirstOrDefault(x => x.Name == "Layout" || x.Name == "Resource_Layout");
+
+			_resourceTypeNames.Add(DrawableClass, "drawable");
+			_resourceTypeNames.Add(ResourceClass, "id");
+			_resourceTypeNames.Add(StyleClass, "style");
+			_resourceTypeNames.Add(LayoutClass, "layout");
 		}
 
-		internal static int IdFromTitle(string title, Type type)
+		static int IdFromTitle(string title, Type type)
```
samhouts: Weird. Why did he add the internal? https://github.com/xamarin/Duplo/commit/7e971911eebffd2d2de516f86bafc4cd54b6a5d5#diff-6d2fedc564b28324406a6ff722fad0e9R53

PureWeen: yea nothing uses it so I figured just remove it :-/

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -127,15 +130,46 @@ public static void Init(Assembly masterAssembly)
...
+			if(_resourceTypeNames.TryGetValue(type, out string value))
+				id = resource.GetIdentifier(name, value, Platform.PackageName);
+
+			if (id > 0)
+				return id;
+
+			return GetId(type, name);
+		}
+
+		static int IdFromTitle(string title, Type type, Context context)
```
garuma: Based on other comment this one can be simplified to call the other.

Xamarin.Forms.Platform.Android/Platform.cs
```diff
@@ -61,6 +63,7 @@ public class Platform : BindableObject, INavigation, IDisposable, IPlatformLayou
 
 		internal Platform(Context context, bool embedded)
 		{
+			PackageName = context.PackageName;
```
garuma: For our case here I don't think we need this tracked and you can safely pass `null` for the third argument of `GetIdentifier`. This will default to look for resources in the project (and not the framework) which is what the reflection code does as well.

BretJohnson: @garuma And I assume this all relies on the design time build being run, to copy the image files to the right place? Not that that's a problem, as I believe the previewer always does that, but I wanted to confirm the mechanism.

PureWeen: When running the app that doesn't seem to work

```c#
// returns the resource id
context.Resources.GetIdentifier(name, value, context.PackageName);

// this just returns a 0
context.Resources.GetIdentifier(name, value, null);
```

garuma: Just checked the difference between designer and native Android and you are right, the designer re-implementation (in Google's layoutlib) looks to be more lenient and doesn't require a package argument. The framework version however will bail out if it's not given.

Looking at the latest changes I think your fallback approach makes sense. If there is a NuGet package that's produced from this we can double check it works on Monday.

@BretJohnson yes this will use the resources from the generated build output same as what the Android designer does for rendering.

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -89,9 +92,17 @@ public static Drawable GetDrawable(this Resources resource, string name)
 			return AndroidAppCompat.GetDrawable(Forms.Context, id);
 		}
 
+		public static void LogInfo(string message)
```
PureWeen: I just left this in so we can easily reference it for logging previewer messages

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -89,9 +90,17 @@ public static Drawable GetDrawable(this Resources resource, string name)
...
```
samhouts:
```suggestion
		[EditorBrowsable(EditorBrowsableState.Never)]
		public static void LogInfo(string message)
```

PureWeen: oops :-/

If no other functional changes are requested we'll make this change before merging so the build/uitests can finish

PR #5587, opened 2019-03-16
Man init

kingces95:
Ala cart forms initialization.

Disregard bat file stuff. 

closes https://github.com/xamarin/Xamarin.Forms/pull/5248

### DIFF
--- a/Xamarin.Forms.Platform.Android/Forms.cs
+++ b/Xamarin.Forms.Platform.Android/Forms.cs
@@ -244,7 +244,7 @@ namespace Xamarin.Forms
 
 			if (!IsInitialized)
 			{
-				if (maybeOptions != null)
+				if (maybeOptions.HasValue)
 				{
 					var options = maybeOptions.Value;
 					var handlers = options.Handlers;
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Platform.Android/Forms.cs
```diff
@@ -189,8 +225,40 @@ static void SetupInit(Context activity, Assembly resourceAssembly)
 
 			if (!IsInitialized)
 			{
-				// Only need to do this once
-				Registrar.RegisterAll(new[] { typeof(ExportRendererAttribute), typeof(ExportCellAttribute), typeof(ExportImageSourceHandlerAttribute) });
+				if (maybeOptions != null)
```
paymicro:
```suggestion
				if (maybeOptions.HasValue)
```
I think it will be more clear

Xamarin.Forms.Platform.Android/Renderers/OpenGLViewRenderer.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Xamarin.Forms.Platform.Android
 {
-	internal class OpenGLViewRenderer : ViewRenderer<OpenGLView, GLSurfaceView>
+	public class OpenGLViewRenderer : ViewRenderer<OpenGLView, GLSurfaceView>
```
paymicro: does this change make sense?

kingces95: To explicitly register the renderer it needs to be public.

PR #5589, opened 2019-03-16
Call InstantiationFailedCallback for any exception

TimBarham:
### Description of Change ###

If `Activator.CreateInstance()` fails, call `InstantiationFailedCallback` (if set) regardless of the exception. For example, in the repro case, the exception is MissingMethodException because there is no default constructor. But there doesn't seem to be any good reason to limit which exceptions we capture here.

### Issues Resolved ### 

- fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/817764

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###
Manual testing in previewer.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Xaml/CreateValuesVisitor.cs
+++ b/Xamarin.Forms.Xaml/CreateValuesVisitor.cs
@@ -94,8 +94,8 @@ namespace Xamarin.Forms.Xaml
 						try {
 							value = Activator.CreateInstance(type);
 						}
-						catch (TargetInvocationException tie) {
-							value = XamlLoader.InstantiationFailedCallback?.Invoke(new XamlLoader.CallbackTypeInfo { XmlNamespace = node.XmlType.NamespaceUri, XmlTypeName = node.XmlType.Name }, type, tie) ?? throw tie;
+						catch (Exception e) {
+							value = XamlLoader.InstantiationFailedCallback?.Invoke(new XamlLoader.CallbackTypeInfo { XmlNamespace = node.XmlType.NamespaceUri, XmlTypeName = node.XmlType.Name }, type, e) ?? throw e;
 						}
 					}
 				}


TimBarham:

TimBarham:
Hmmm... All we're doing in this `try { }` is calling `Activator.CreateInstance()`. Why do we care what the exception is? What exceptions might Forms not be able to recover from?

Anyway... In this specific case, it is an abstract class, and the exception is a `MissingMethodException`. But a protected default constructor in a concrete class will also repro. So the following test will repro:
``` cs
[Test]
public void IgnoreInaccessibleConstructorException()
{
	var xaml = @"
			<ContentPage xmlns=""http://xamarin.com/schemas/2014/forms""
				xmlns:x=""http://schemas.microsoft.com/winfx/2009/xaml""
				xmlns:local=""clr-namespace:Xamarin.Forms.Xaml.UnitTests;assembly=Xamarin.Forms.Xaml.UnitTests"">
				<local:InaccessibleDefaultConstructor />
				<local:AbstractClass />
		</ContentPage>";
	var exceptions = new List<Exception>();
	Xamarin.Forms.Internals.ResourceLoader.ExceptionHandler = exceptions.Add;
	Assert.DoesNotThrow(() => XamlLoader.Create(xaml, true));
}
```
With the following classes...
``` cs
public class InaccessibleDefaultConstructor
{
	protected InaccessibleDefaultConstructor() { }
}

public abstract class AbstractClass
{
	protected AbstractClass() { }
}
```

StephaneDelcroix:
let's add a catch for `MemberAccessException` then

StephaneDelcroix:
I'll take care of this

alanmcgovern:
```
we do not want to catch too many exception, especially cases we can not recover from. please provide an example in which this is required, and we'll add a case for it
```
There are few exceptions which *can* be thrown so we're probably fine with the new PR. The things we care about are the ones wrapped by the TIE.

( https://docs.microsoft.com/en-us/dotnet/api/system.activator.createinstance?view=netframework-4.7.2 )

PR #5604, opened 2019-03-18
[iOS] Fixes memory leak when EntryCell is used

### LINKED ISSUE #5555: [iOS] EntryCell causes memoryleak
### Description

Having an EntryCell on a TableView causes the whole page to stay in memory. Any other cell types are cleaned up correctly.

this issue may be related to #5560

### Steps to Reproduce

1. Push a page with an EntryCell on it
2. Pop the page
3. GC.Collect() 

### Expected Behavior

The page should be garbage collected after being popped.

### Actual Behavior

The page stays in memory indefinitely.

### Basic Information

- Version with issue: forms 3.6.0.220655
- Last known good version: ?
- IDE: VS2017 15.9.8
- Platform Target Frameworks: 
  - iOS:  12.1.4

### Reproduction Link

[FormsTest.zip](https://github.com/xamarin/Xamarin.Forms/files/2966324/FormsTest.zip)

kicsiede:
### Description of Change ###

Fixes leak when EntryCell is used.
The handler was keeping the renderer alive.

### Issues Resolved ### 

- fixes #5555

### API Changes ###

 None

### Platforms Affected ### 

- iOS

### Testing Procedure ###

see test in #5555

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Cells/EntryCellRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Cells/EntryCellRenderer.cs
@@ -154,22 +154,35 @@ namespace Xamarin.Forms.Platform.iOS
 
 			public event EventHandler TextFieldTextChanged;
 
-			bool OnShouldReturn(UITextField view)
+			static bool OnShouldReturn(UITextField view)
 			{
-				var handler = KeyboardDoneButtonPressed;
-				if (handler != null)
-					handler(this, EventArgs.Empty);
+                var realCell = GetRealCell<EntryCellTableViewCell>(view);
+                var handler = realCell?.KeyboardDoneButtonPressed;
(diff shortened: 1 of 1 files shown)


kingces95:
Delegate allocated here

image

kingces95:
Prob still leaks a bit. Would have to store the delegate in a static too. Nice catch!

image

kicsiede:
@paymicro see #5655

kicsiede:
@paymicro & @kingces95 as i think this is a recurring issue it would be nice to have a more generic solution, for example every gui-test should check for leakage as well. could be somewhat automatized like:
1. navigate to a page with tested control (with weakref on page)
2. gc.collect
3. throw error if weakref is not null

PR #5624, opened 2019-03-20
[C] alias shell names

StephaneDelcroix:
### API Additions ###

Tab alias for a ShellSection
TabContent alias for a ShellContent
FlyoutItem alias for a ShellItem

Shell has a property now called Flyout that maps to Items

### DIFF
--- a/Xamarin.Forms.Core/Shell/ShellContent.cs
+++ b/Xamarin.Forms.Core/Shell/ShellContent.cs
@@ -16,7 +16,7 @@ namespace Xamarin.Forms
 	{
 	}
 
-	[ContentProperty("Content")]
+	[ContentProperty(nameof(Content))]
 	public class ShellContent : BaseShellItem, IShellContentController
 	{
 		static readonly BindablePropertyKey MenuItemsPropertyKey =
--- a/Xamarin.Forms.Core/Shell/ShellContent.cs
+++ b/Xamarin.Forms.Core/Shell/ShellContent.cs
@@ -12,6 +12,10 @@ using Xamarin.Forms.Internals;
 
 namespace Xamarin.Forms
 {
+	public class TabContent : ShellContent
+	{
+	}
+
 	[ContentProperty("Content")]
 	public class ShellContent : BaseShellItem, IShellContentController
 	{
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Core/Shell/ShellSection.cs
```diff
@@ -8,7 +8,12 @@
 
 namespace Xamarin.Forms
 {
-	[ContentProperty("Items")]
+	public class ShellTab : ShellSection
```
davidortinau: Can we do `Tab` instead?

Xamarin.Forms.Core/Shell/ShellContent.cs
```diff
@@ -12,6 +12,10 @@
 
 namespace Xamarin.Forms
 {
+	public class TabContent : ShellContent
+	{
+	}
+
 	[ContentProperty("Content")]
 	public class ShellContent : BaseShellItem, IShellContentController
```
davidortinau: Could we have a Content alias, removing the `Shell` here as well?

StephaneDelcroix: I would not, as ShellContent has a property Content. I'm fine with an alias, be we need a better name, something like "Page", but not that

Xamarin.Forms.Core/Shell/ShellSection.cs
StephaneDelcroix: if the mid-term goal is to deprecate NavigationPage, TabbedPage, and MDP, I'm fine if we have some overlap in names. otherwise, it's better to avoid confusion

Xamarin.Forms.Core/Shell/ShellContent.cs
PureWeen: we'll probably merge as is for pre8 testing

But would there be a way to achieve what David is asking with simplifying down the names so we could have

```
<Shell>
        <Tab>
            <Content>
                
            </Content>
        </Tab>
</Shell>
```

it does seem confusing to now have something called
*TabContent* inside the Xamarin.Forms namespace that only belongs to Shell

PR #5671, opened 2019-03-22
[Android] don't clear shell content because native page isn't visible

### LINKED ISSUE #4684: Xamarin 4.0.0.8055-pre1 Shell Crashes when navigating too quickly
### Steps to Reproduce

1.  Create AppShell Application
2.  Create 3 ContentPages
3.  Setup shell like so:
```xml
<Shell xmlns="http://xamarin.com/schemas/2014/forms"
             xmlns:local="clr-namespace:Example"
             RouteScheme="app" 
             RouteHost="example.com"
             Route="main"
             x:Class="Example.AppShell">
    <ShellItem Title="Home" Route="Home">
        <ShellSection>
            <ShellContent>
                <local:HomePage />
            </ShellContent>
        </ShellSection>
    </ShellItem>
    <ShellItem Title="Connect" Route="connect">
        <ShellSection>
            <ShellContent Title="Connect">
                <local:ConnectPage />
            </ShellContent>
            <ShellContent Title="Control">
                <local:ControlPage />
            </ShellContent>
        </ShellSection>
    </ShellItem>
</Shell>
```

### Expected Behavior

Since all the pages are basically just the defaults, we expect to have all the navigation be quick and simple.

### Actual Behavior

If you navigate back and forth between pages with tabs and the pages without and move back and forth fairly quick the app crashes when you click on the hamburger with a NullReferenceException without much information to help on the first exception.  It may not happen on the first click of the hamburger icon but if you go back and forth and toggle between pages in the tabs it should be reproducable.

image

The second unhandled exception does have a clue:
image

I also verified that it isn't related to the other issue I submitted at https://github.com/xamarin/Xamarin.Forms/issues/4683 by removing the styles and just letting it run by default themes

### Basic Information

- Version with issue: 4.0.0.8055-pre1
- IDE: Visual Studio 2017
- Platform Target Frameworks: 
  - Android: 9.0
- Android Support Library Version: 28.0.0-preview8
- Affected Devices: OnePlus 6, Emulator

PureWeen:
### Description of Change ###

On android when you navigated away from a page it calls recycle on Shell Content but there's not really a reason to remove the shells internal content page is there?

iOS doesn't have this behavior when navigating away from pages. 

When shell starts up it instantiates all of the content pages that are part of the shell contents so it doesn't seem like this value should ever get set to null 

### Issues Resolved ### 

- fixes #4684

### Platforms Affected ### 
- Android

### Testing Procedure ###
- if you have a shell with two shell items
- toggle between them
- and now click on one of the tabs in a shell item

HEre's the xaml that I used to test
```
<ShellItem Title="Home" Route="Home">
        <ShellSection>
            <ShellContent>
                <ContentPage />
            </ShellContent>
        </ShellSection>
    </ShellItem>
    <ShellItem Title="Connect" Route="connect">
        <ShellSection>
            <ShellContent Title="Connect">
                <ContentPage  Title="Connect"/>
            </ShellContent>
            <ShellContent Title="Control">
                <ContentPage Title="Control" />
            </ShellContent>
        </ShellSection>
    </ShellItem>
```

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Shell/ShellContent.cs
+++ b/Xamarin.Forms.Core/Shell/ShellContent.cs
@@ -74,11 +74,6 @@ namespace Xamarin.Forms
 
 		void IShellContentController.RecyclePage(Page page)
 		{
-			if (ContentCache == page)
-			{
-				OnChildRemoved(page);
-				ContentCache = null;
-			}
 		}
 
 		Page _contentCache;
--- a/Xamarin.Forms.Platform.Android/Renderers/ShellSectionRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ShellSectionRenderer.cs
@@ -48,7 +48,11 @@ namespace Xamarin.Forms.Platform.Android
 			{
 				ShellSection.SetValueFromRenderer(ShellSection.CurrentItemProperty, shellContent);
 
-				_toolbarTracker.Page = ((IShellContentController)shellContent).Page;
+				var page = ((IShellContentController)shellContent).Page;
+				if(page == null)
+					throw new ArgumentNullException(nameof(page), "Shell Content Page is Null");
+
+				_toolbarTracker.Page = page;
 			}
 			else
 			{
--- a/Xamarin.Forms.Platform.Android/Renderers/ShellSectionRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ShellSectionRenderer.cs
@@ -41,16 +41,16 @@ namespace Xamarin.Forms.Platform.Android
 				return;
 
 			var stack = shellSection.Stack.ToList();
-			bool result = ((IShellController)_shellContext.Shell).ProposeNavigation(ShellNavigationSource.ShellContentChanged, 
+			bool result = ((IShellController)_shellContext.Shell).ProposeNavigation(ShellNavigationSource.ShellContentChanged,
 				(ShellItem)shellSection.Parent, shellSection, shellContent, stack, true);
 
 			if (result)
 			{
-				ShellSection.SetValueFromRenderer(ShellSection.CurrentItemProperty, shellContent);
-
 				var page = ((IShellContentController)shellContent).Page;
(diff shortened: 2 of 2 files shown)


Xamarin.Forms.Core/Shell/ShellContent.cs
```diff
@@ -74,11 +74,6 @@ Page IShellContentController.GetOrCreateContent()
 
 		void IShellContentController.RecyclePage(Page page)
 		{
-			if (ContentCache == page)
```
PureWeen: do we even need the recycle call anymore? iOS doesn't use this call and it's kind of confusing

Xamarin.Forms.Platform.Android/Renderers/ShellSectionRenderer.cs
```diff
@@ -48,7 +48,11 @@ void ViewPager.IOnPageChangeListener.OnPageSelected(int position)
 			{
 				ShellSection.SetValueFromRenderer(ShellSection.CurrentItemProperty, shellContent);
 
-				_toolbarTracker.Page = ((IShellContentController)shellContent).Page;
+				var page = ((IShellContentController)shellContent).Page;
+				if(page == null)
+					throw new ArgumentNullException(nameof(page), "Shell Content Page is Null");
```
PureWeen: is this a valid check? This bug was annoying to track backwards to here so I put this here. ToolbarTracker can't really function with page set to null

hartez: Should it be earlier? If it's going to fail anyway, should ProposeNavigation even be called?

PureWeen: I moved it above the *SetValueFromRenderer* call but I think it's still fine there. That page is only used in that branch of the if statement so it seems excessive to have it always do that check in a case where the result is going to be false

Issue #2520, opened 2018-04-21
[Android] Effects don't detach and OnElementchanged(null) is never called


PureWeen:
### Description

On Android when the Renderer disposes it doesn't call SetElement(null) it only sets the property so it doesn't propagate anywhere.  This behavior isn't consistent with how it works on iOS and UWP.  In the documentation it typically talks about using *OnDetached* and *if(e.OldElement != null)* to unhooks your code.  As it stands on Android it's probably best to write a general method that unhooks everything and then call that from both Dispose and *if(e.OldElement != null)*  as that's the only reliable way your custom bits will be disposed of.

I tested calling *SetElement(null)* but it didn't "just work" 

### Steps to Reproduce
- run the sample
- the sample has a custom effect and it's own custom button renderer
- click the button on the first page to navigate forward
- click the button on the second page to navigate back
- notice on UWP/iOS that OnDetached and OnElementChanged are called when the page is popped off the stack but Android that does not happen

### Expected Behavior
- OnElementChanged gets called when the Element is set to null
- Effects get Detached when the Element is set to null

### Actual Behavior
- OnElementChanged never gets called when the Element is set to null
- Effects don't Detached when the Element is set to null

### Basic Information
- Version with issue: 3.0 
 
### Reproduction Link
[EffectsAndRenderers.zip](https://github.com/xamarin/Xamarin.Forms/files/1934109/EffectsAndRenderers.zip)

samhouts:
TODO: Clean up test added in #2857 when this is resolved.

kvpt:
I don't know if it requires a separate issue, but the same apply to behavior.

samhouts:
This was partially resolved for legacy renderers, but fast renderers still suffer from it.

chaosifier:
Any updates on the issue?

dersia:
Is anyone working on this?

dersia:
I started working on this issue...
From what I can see so far, the fix would go through all Heads. There are places in the GTK head that are also involved. 
@PureWeen do you think this should be fixed in all Heads?
What bothers me, is that when it comes to Effects, other Heads seem to use other methods to attach Effects, this makes it hard to refactor and make sure it works everywhere as intended.

Or should I just provide a fix for Android and we'll open issues for the other Heads?

PureWeen:
[...]

PR #5685, opened 2019-03-26
[C] Add EventToCommandBehavior

davidortinau:
### Description of Change ###

This is the same class from our sample and documentation. I see this all over the place, and ListView needs it to really do MVVM. 

I added namespace `Xamarin.Forms.Behaviors`. Is `Xamarin.Forms` better? It seems like most things are in that space, but it's a bit crowded. Is that for XAML simplicity?

### API Changes ###

None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ none ] Has automated tests

### DIFF
--- a/Xamarin.Forms.Core/Behaviors/BehaviorBase.cs
+++ b/Xamarin.Forms.Core/Behaviors/BehaviorBase.cs
@@ -1,0 +1,40 @@
+using System;
+using Xamarin.Forms;
+
+namespace Xamarin.Forms.Behaviors
+{
+    public class BehaviorBase<T> : Behavior<T> where T : BindableObject
+	{
+		public T AssociatedObject { get; private set; }
+
+		protected override void OnAttachedTo (T bindable)
+		{
+			base.OnAttachedTo (bindable);
+			AssociatedObject = bindable;
+
+			if (bindable.BindingContext != null) {
+				BindingContext = bindable.BindingContext;
+			}
+
+			bindable.BindingContextChanged += OnBindingContextChanged;
+		}
+
+		protected override void OnDetachingFrom (T bindable)
+		{
+			base.OnDetachingFrom (bindable);
+			bindable.BindingContextChanged -= OnBindingContextChanged;
+			AssociatedObject = null;
+		}
+
+		void OnBindingContextChanged (object sender, EventArgs e)
+		{
+			OnBindingContextChanged ();
+		}
+
+		protected override void OnBindingContextChanged ()
+		{
+			base.OnBindingContextChanged ();
+			BindingContext = AssociatedObject.BindingContext;
+		}
+	}
+}
--- a/Xamarin.Forms.Core/Behaviors/EventToCommandBehavior.cs
+++ b/Xamarin.Forms.Core/Behaviors/EventToCommandBehavior.cs
@@ -1,0 +1,115 @@
+using System;
+using System.Reflection;
+using System.Windows.Input;
+using Xamarin.Forms;
+
+namespace Xamarin.Forms.Behaviors
+{
+	public class EventToCommandBehavior : BehaviorBase<View>
+	{
+		Delegate eventHandler;
+
+		public static readonly BindableProperty EventNameProperty = BindableProperty.Create ("EventName", typeof(string), typeof(EventToCommandBehavior), null, propertyChanged: OnEventNameChanged);
+		public static readonly BindableProperty CommandProperty = BindableProperty.Create ("Command", typeof(ICommand), typeof(EventToCommandBehavior), null);
+		public static readonly BindableProperty CommandParameterProperty = BindableProperty.Create ("CommandParameter", typeof(object), typeof(EventToCommandBehavior), null);
+		public static readonly BindableProperty InputConverterProperty = BindableProperty.Create ("Converter", typeof(IValueConverter), typeof(EventToCommandBehavior), null);
+
+		public string EventName {
+			get { return (string)GetValue (EventNameProperty); }
+			set { SetValue (EventNameProperty, value); }
+		}
+
+		public ICommand Command {
+			get { return (ICommand)GetValue (CommandProperty); }
+			set { SetValue (CommandProperty, value); }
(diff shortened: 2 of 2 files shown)


davidortinau:
Should we also include the event args converter in this behavior? I'm thinking baked right in. Thoughts? Or is that converter useful beyond this behavior?

davidortinau:
ok, on second thought (third thought?) should we abandon this altogether and just add the command to ListView? Let's stop playing around and get right to the heart of this, right?!

davidortinau:
Fourth thought (too much coffee?), this is itself useful beyond ListView. So....let's keep it as is and consider ListView another topic.

andreinitescu:
@davidortinau My suggestion is to just use Prism's implementation:
https://github.com/PrismLibrary/Prism/blob/master/Source/Xamarin/Prism.Forms/Behaviors/EventToCommandBehavior.cs

Beside the obvious properties like EventName, Command and CommandParameter, it also has the very useful parameters:
- EventArgsConverter
- EventArgsConverterParameter
- EventArgsParameterPath

Prism has this implementation for a very long time and it works well, so I strongly suggest to "adopt" it.
@brianlagunas Thoughts? 😃

brianlagunas:
Keep the `Xamarin.Forms.Behaviors` namespace.

You are more than welcome to take the Prism version of the EventToCommand (with unit tests). You can just copy/paste it into this PR, or I can submit another PR to replace this one with unit tests included.

adrianknight89:
I'd like the team to test this against memory leaks when it's being added to the codebase. In the past, I've had cases where behaviors were not being removed when pages were removed from the stack.

brianlagunas:
@adrianknight89 I believe this is still the case. When pages are popped off a stack, behaviors are not cleaned up automatically. Devs are still responsible for cleaning these up.

andreinitescu:
> When pages are popped off a stack, behaviors are not cleaned up automatically. 

Has someone filled an issued for this? Seems like a serious problem to me.

adrianknight89:
I vaguely remember seeing it on the forums or Bugzilla but never on here. I think the original issue was reported under another title, but the root cause seemed to be leaky behaviors.

davidortinau:
@adrianknight89 I went looking for it and couldn't find it. If you locate it, let me know.

Thanks @brianlagunas, we may do that. I haven't had a chance to look at it just yet.

kvpt:
> Has someone filled an issued for this? Seems like a serious problem to me.

This issue is mainly about effects but it's the same problem with the behaviors : https://github.com/xamarin/Xamarin.Forms/issues/2520

wcoder:
Hi everyone!

I suggest you add an additional method (like `bool CheckConditionOfEventArgs(object eventArgs)`) for the ability for extending of `EventToCommandBehavior` to implement other behaviors with commands?

## Example 

### InfiniteScrollBehavior

```cs
public class InfiniteScrollBehavior : EventToCommandBehavior
{
    public InfiniteScrollBehavior()
    {
        EventName = "ItemAppearing";
    }

    protected override bool CheckConditionOfEventArgs(object eventArgs)
    {
        var args = (ItemVisibilityEventArgs)eventArgs;
        var listView = (ListView)AssociatedObject;

        if (listView.ItemsSource is IList list)
        {
            if (list.Count == 0)
                return false;

            //hit bottom!
            if (args.Item == list[list.Count - 1])
            {
                return true;
            }
        }

        return false;
    }
}
```
Using:
```
<ListView.Behaviors>
    <behaviors:InfiniteScrollBehavior Command="{Binding LoadMoreDataCommand}" />
</ListView.Behaviors>
```

What do you think about it?

andreinitescu:
@wcoder Prism has OnEventRaised virtual so you can override it in any way you want:
```
 protected virtual void OnEventRaised(object sender, EventArgs eventArgs)
```

StephaneDelcroix:
> I added namespace Xamarin.Forms.Behaviors. Is Xamarin.Forms better? It seems like most things are in that space, but it's a bit crowded. Is that for XAML simplicity?

yes, XF is better

Xamarin.Forms.Core/Behaviors/BehaviorBase.cs
```diff
@@ -0,0 +1,40 @@
+using System;
+using Xamarin.Forms;
+
+namespace Xamarin.Forms.Behaviors
+{
+    public class BehaviorBase<T> : Behavior<T> where T : BindableObject
+	{
+		public T AssociatedObject { get; private set; }
```
StephaneDelcroix: this is all wrong. Behaviors are meant to be declared in Styles, and have shared instances. So your behavior should keep track of an AssociatedObject, but an array of AssociatedObjects.

and then you need to keep a kind of proxy for the bindable context thing, so the bindings are applied.

I'm guilty of this, I'm the one who made Behavior inherit from BindableObject

andreinitescu: @StephaneDelcroix In my opinion, that's not an issue.
Isn't the same problem with any other object type which is not meant to work in Style? Developers should be aware the Style is shared and doesn't work as a template. For example, do triggers work when put in a Style? 

In WPF there's a [x:Shared](https://docs.microsoft.com/en-us/dotnet/framework/xaml-services/x-shared-attribute) attribute , see https://stackoverflow.com/questions/22321966/interaction-triggers-in-style-in-resourcedictionary-wpf/22329342
Also See this interesting thread for using Behavior in Style in WPF: https://stackoverflow.com/questions/1647815/how-to-add-a-blend-behavior-in-a-style-setter

So I think having Behavior<T> like how David implemented it is OK.

StephaneDelcroix: Triggers and Behaviors work when set in a Style, yes.

We could support `x:Shared`, please open a enhancement request for it.

samhouts:
I believe we're in agreement that this is not the approach we'll take at this time, so I will close this PR. Thank you!!

PR #5689, opened 2019-03-26
Refactoring Map.ItemsSource to use NotifyCollectionChangedEventArgsEx…

andreinitescu:
### Description of Change ###

Refactoring Map.ItemsSource to use NotifyCollectionChangedEventArgsExtensions instead

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Maps/Map.cs
+++ b/Xamarin.Forms.Maps/Map.cs
@@ -5,6 +5,7 @@ using System.Collections.ObjectModel;
 using System.Collections.Specialized;
 using System.ComponentModel;
 using System.Linq;
+using Xamarin.Forms.Internals;
 
 namespace Xamarin.Forms.Maps
 {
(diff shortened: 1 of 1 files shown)


andreinitescu:
I strongly recommend merging this as soon as possible. Thanks.

Xamarin.Forms.Maps/Map.cs
```diff
@@ -155,37 +156,10 @@ void OnItemTemplatePropertyChanged(DataTemplate oldItemTemplate, DataTemplate ne
...
-				case NotifyCollectionChangedAction.Remove:
-					if (e.OldStartingIndex == -1)
-						goto case NotifyCollectionChangedAction.Reset;
-					foreach (object item in e.OldItems)
-						RemovePin(item);
-					break;
-				case NotifyCollectionChangedAction.Replace:
-					if (e.OldStartingIndex == -1)
-						goto case NotifyCollectionChangedAction.Reset;
-					foreach (object item in e.OldItems)
```
hartez: With the extensions, a Replace with different item counts won't work the same here. That is, if you are replacing 3 items with 2 items, the old version would work; the new version will just jump to Reset. Is that something we should be concerned about?

andreinitescu: Well, I see #5582 got merged, so I guess if it was OK for BindableLayout, it should be OK for Map too? It makes sense to have INotifyCollectionChanged implementation consistent across controls. 
Maybe the extensions could be changed (a future PR?) so that it doesn't reset collection when new and old items differ in length.

hartez: > I guess if it was OK for BindableLayout, it should be OK for Map too?

Either that, or we made a terrible mistake merging 5582. That implementation of Replace in `Apply` is very suspect.

It looks like Map.Pins is an `ObservableCollection<Pin>`, so while I don't like this implementation of `Apply`, it should be fine for this purpose. It looks like the implementation of `ObservableCollection` only uses `NotifyCollectionChangedAction.Replace` for a single item being set. So at least for this specific application, it shouldn't cause any immediate problems.

andreinitescu: Yes, most apps use `ObservableCollection` so it's fine, but for apps which which do something more advanced and throw custom `NotifyCollectionChangedAction` it will be less efficient

samhouts:
Faiing tests not related. Thanks!

PR #5695, opened 2019-03-26
[iOS] Fix: memory leak when Navigation.RemovePage is used

kicsiede:
The leak occurs in a MasterDetailPage.Detail = NavigationPage(ContentPage) scenario if the ContentPage has ToolBarItems. (ToolBarItems create a strong dependency therefore the ViewController of the NavigationPage must be explicitly disposed)

### Description of Change ###

disposing viewcontroller in RemovePage.

### Platforms Affected ### 

- iOS

### PR Checklist ###

- [x] Has automated tests 


andreinitescu:
@kicsiede Just a suggestion, adding an unit test for this would be great...

rmarinho:
Hey do you think you can you add a UITest ?! 

thanks

kicsiede:
@rmarinho please check

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5695.cs
```diff
@@ -0,0 +1,96 @@
...
+#if UITEST
+using NUnit.Framework;
+using Xamarin.UITest;
+using Xamarin.Forms.Core.UITests;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 5555, "Memory leak when SwitchCell or EntryCell", PlatformAffected.iOS)]
```
paymicro: copy/paste detected =)
I'll think about something more universal for any controls when checking for memory leaks.

rmarinho: I agree, I was gonna say we had a test for most of the renderers where we pushed a page with a control, and make sure it was collected, I wonder if we can do that in the dynamic gallery test you created Pavel

kicsiede: @paymicro @rmarinho fyi here it is not the control causing the leak but the RemovePage method.

PR #5698, opened 2019-03-26
[Shell] ignored case in routing

paymicro:
### Description of Change ###

Routing is supported in uppercase, case is ignored.
Also added support digits and symbols `+` `-` `@` `:` `%` `.` `_`  `~` `#` `=`
The length of one part of the route is limited to 1 to 100 characters.

### Issues Resolved ### 

none

### API Changes ###

 
 None

### Platforms Affected ### 

- Core

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Routing.cs
+++ b/Xamarin.Forms.Core/Routing.cs
@@ -7,7 +7,12 @@ namespace Xamarin.Forms
 	public static class Routing
 	{
 		static int s_routeCount = 0;
+
+#if NETSTANDARD1_0
 		static Dictionary<string, RouteFactory> s_routes = new Dictionary<string, RouteFactory>();
+#else
+		static Dictionary<string, RouteFactory> s_routes = new Dictionary<string, RouteFactory>(StringComparer.InvariantCultureIgnoreCase);
+#endif
 
 		internal const string ImplicitPrefix = "IMPL_";
 
@@ -41,7 +46,9 @@ namespace Xamarin.Forms
 		public static Element GetOrCreateContent(string route)
 		{
 			Element result = null;
-
+#if NETSTANDARD1_0
+			route = route.ToLowerInvariant();
+#endif
 			if (s_routes.TryGetValue(route, out var content))
 				result = content.GetOrCreate();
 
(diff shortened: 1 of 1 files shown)


rmarinho:
Needs, rebase, can we add a unit test ?

Xamarin.Forms.Core/Routing.cs
```diff
@@ -69,16 +79,22 @@ public static string GetRoute(Element obj)
 		public static void RegisterRoute(string route, RouteFactory factory)
 		{
 			if (!ValidateRoute(route))
-				throw new ArgumentException("Route must contain only lowercase letters");
+				throw new ArgumentException("Route contain invalid letters");
```
hartez: "Route contain invalid letters" -> "Route contains invalid characters"

Xamarin.Forms.Core/Routing.cs
```diff
@@ -88,13 +104,7 @@ public static void SetRoute(Element obj, string value)
...
 
 		static bool ValidateRoute(string route)
-		{
-			// Honestly this could probably be expanded to allow any URI allowable character
-			// I just dont want to figure out what that validation looks like.
-			// It does however need to be lowercase since uri case sensitivity is a bit touchy
-			Regex r = new Regex(@"^[a-z|\/]*$");
-			return r.IsMatch(route);
-		}
+			=> new Regex(@"^[-a-zA-Z0-9@:%._\+~=\/]{1,100}$").IsMatch(route);
```
hartez: Instead of having each RegisterRoute overload check the result of ValidateRoute and throw an exception, perhaps ValidateRoute should return void and throw an exception if the route is invalid.

Then ValidateRoute can throw detail exception messages.

PureWeen: So we're thinking just change the route check to instantiate a URI 
if it succeeds it's valid if it doesn't then it's not valid. Instead of trying to use a REGEX

Xamarin.Forms.Core/Routing.cs
```diff
@@ -68,17 +78,21 @@ public static string GetRoute(Element obj)
 
 		public static void RegisterRoute(string route, RouteFactory factory)
 		{
-			if (!ValidateRoute(route))
-				throw new ArgumentException("Route must contain only lowercase letters");
+			ValidateRoute(route);
 
+#if NETSTANDARD1_0
+			route = route.ToLowerInvariant();
```
PureWeen: Why to lower invariant on netstandard 1?

Xamarin.Forms.Core/Routing.cs
```diff
@@ -7,25 +7,30 @@ namespace Xamarin.Forms
 	public static class Routing
 	{
 		static int s_routeCount = 0;
+
+#if NETSTANDARD1_0
 		static Dictionary<string, RouteFactory> s_routes = new Dictionary<string, RouteFactory>();
+#else
+		static Dictionary<string, RouteFactory> s_routes = new Dictionary<string, RouteFactory>(StringComparer.InvariantCultureIgnoreCase);
```
PureWeen: So for this one we're actually making it all case sensitive so any to lowers or ignore cases can be removed.

Xamarin.Forms.Core/Routing.cs
StephaneDelcroix: because the non-ns10 has a StringComparer, see ^^^^

PureWeen: I don't think it should anymore though. I'm pretty sure those just got added when I was proposing casing shouldn't matter but then EZ opened my eyes to why it has to matter 👀

PureWeen:
copied these changes slightly modified over into https://github.com/xamarin/Xamarin.Forms/tree/shell_uri

PR #5727, opened 2019-03-28
README.md: Xamarin.Forms is not specific for C#

knocte:
It can be used from any .NET language (in fact I built an F# Xamarin.Forms app myself).

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 # Xamarin.Forms #
 
-Xamarin.Forms provides a way to quickly build native apps for iOS, Android, Windows and macOS, completely in C#.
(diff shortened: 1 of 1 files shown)


StephaneDelcroix:
it is correct, per se, but I'm afraid this will make us support scenario we're not ready to support. @davidortinau: your call

samhouts:
@knocte Thank you! Yes, it is technically possible to build an app in any .NET language, and we're excited that you're doing that! As Stephane mentioned, we actively test and support C# at this time, so that's all we're ready to commit to at this point. But please do let us know if you run into any issues with F#!

PR #5740, opened 2019-03-28
Extend Control Gallery to allow for Shell UI Tests

PureWeen:
### Description of Change ###

- added UI test for Issue4684 to demonstrate UI tests
- Added an internal constructor to shell that lets us disable the flag check. Since all our UI tests have internals visible this makes it easy to circumvent the flag check without changing the flags
- added a TestShell base class to mimic other Test<Pages>. Added some helper methods here for opening flyout and clicking on menus
- Added Store shell to top list of gallery so it can be swapped in for main page
- Changed the base class on store shell to test shell. In theory this should make it so you can create an Issue that just inherits from StoreShell and then write a UI test against StoreShell
- Changed the AccessibilityIdentifier on the app bar button on iOS to be OK like it is on android which is probably wrong in both cases but now it's at least consistently wrong. This will need to be updated to a more correct AccessibilityIdentifier
- 

### API Changes ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4684.xaml.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4684.xaml.cs
@@ -14,7 +14,7 @@ namespace Xamarin.Forms.Controls.Issues
 	[Issue(IssueTracker.Github, 4684, "[Android] don't clear shell content because native page isn't visible",
 		PlatformAffected.Android)]
 #if UITEST
-	[NUnit.Framework.Category(UITestCategories.ListView)]
+	[NUnit.Framework.Category(UITestCategories.Shell)]
 #endif
 	public sealed partial class Issue4684 : TestShell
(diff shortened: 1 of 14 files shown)


PureWeen:
Android UI Test failures look to all be related to Fast Renderers which are already fixed

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4684.xaml.cs
```diff
@@ -0,0 +1,50 @@
...
+using Xamarin.Forms.Core.UITests;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 4684, "[Android] don't clear shell content because native page isn't visible",
+		PlatformAffected.Android)]
+#if UITEST
+	[NUnit.Framework.Category(UITestCategories.ListView)]
```
rmarinho: this should be Shell

PR #5730, opened 2019-03-28
[Shell] propagate BindingContext to SearchHandler

StephaneDelcroix:
### Issues Resolved ### 

- fixes #5705
### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

### Behavioral/Visual Changes ###

None

### Before/After Screenshots ### 

Not applicable

### PR Checklist ###

- [x] Has automated tests 
- [x] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard

PR #5743, opened 2019-03-29
[Shell] propagate BC to BackButtonBehavior

### LINKED ISSUE #5734: [Shell] BackButtonBehavior BindingContext not inherited from page
### Description

`BackButtonBehavior` can bind to commands. However, while I can bind to a ViewModel ICommand, this is only possible if I set the `BackButtonBehavior.BindingContext` directly. Leaving the `BindingContext` unset results in the `Command` binding not resolving. I'd expect `BackButtonBehavior` to inherit the `BindingContext` from the page it's on.

`BackCommand` unresolved:

```
    <Shell.BackButtonBehavior>
        <BackButtonBehavior Command="{Binding BackCommand}"
                            TextOverride="back"
                            IsEnabled="true" />    
    </Shell.BackButtonBehavior>
```

`BackCommand` resolved:

```
    <Shell.BackButtonBehavior>
        <BackButtonBehavior Command="{Binding BackCommand}"
                            BindingContext="{Binding Source={x:Reference self}, Path=BindingContext}"
                            TextOverride="back"
                            IsEnabled="true" />    
    </Shell.BackButtonBehavior>
```

### Steps to Reproduce

1. Run the attached sample.
2. Browse to the Monkeys page (bottom tab).
3. Select a monkey to move to the MonkeyDetailsPage.
4. Click the back (text) button.

### Expected Behavior

`MonkeyDetailViewModel.BackCommand` is resolved, and the `NavigateBack` method is executed.

### Actual Behavior

MonkeyDetailViewModel.BackCommand remains unresolved.

### Basic Information

- Version with issue: 4.0-pre7
- IDE: VSMac 7.8.3

### Reproduction Link

[Xaminals.zip](https://github.com/xamarin/Xamarin.Forms/files/3019128/Xaminals.zip)

StephaneDelcroix:
### Description of Change ###

no test, as it's straightforward, and mimics #5730

### Issues Resolved ### 

- fixes #5734

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 


Xamarin.Forms.Core/BindableObject.cs
```diff
@@ -194,7 +194,12 @@ public static void SetInheritedBindingContext(BindableObject bindable, object va
 
 		protected void ApplyBindings() => ApplyBindings(skipBindingContext: false, fromBindingContextChanged: false);
 
-		protected virtual void OnBindingContextChanged() => BindingContextChanged?.Invoke(this, EventArgs.Empty);
+		protected virtual void OnBindingContextChanged()
+		{
+			BindingContextChanged?.Invoke(this, EventArgs.Empty);
+			if (Shell.GetBackButtonBehavior(this) is BackButtonBehavior buttonBehavior)
```
PureWeen: Should we generalize this into a method on Shell? There's a few things it seems like this should apply to?

`Shell.PropagateBindingContext`

Then it can do this same thing here to
`SetTitleView`
`SetSearchHandler`

I think that's the rest for now

Then it can also check if it's even running inside a shell before trying to propagate any of these values at all.

StephaneDelcroix: @PureWeen this is the first time we're declaring attached BP of type BO, and I choose not to generalize for now to see:
- if it's needed,
- what pattern emerges from the use cases

As I foresee it, the generalization will probably happen in BP/BO, and not shell, making sure the context is propagated when a property is added (or attached), and when the BC changes. This is not a change I'd like to see in 4.0.0, so I think this should be left as-is.

We shouldn't check if it's running inside the Shell for the following reasons:
- checking if a Page is inside a Shell is more expensive than propagating the BC
- applying the BC when a page is parented to a Shell is a nightmare
- SearchHandler, e.g., could be reused outside the context of a Shell (provided the renderers are implemented)
- this would break expectations
- we still hope all app will end up being Shell-based, so no need to check :P

PureWeen: Sounds good. SetTitleView needs the same treatment right? and I think that'll be the last one.

Loop that in with this PR? or make a new one?

PR #5754, opened 2019-03-29
[Android/iOS] RemainingItemsThreshold and Scrolled implementation for CollectionView

### LINKED ISSUE #4654: CarouselView notify when scrolled
## Summary
In CarouselView, the only way to change the visible item is by using the ```ScrollTo``` method of ```ItemsView```. This is very useful so you can change the visible item(or selectedindex) from the Xamarin.Forms. But, when the user swipes the Abstraction Layer never gets notified. From a little workaround, in Android the ```ItemsViewRenderer``` could add a ```OnScrollListener``` and inside the ```OnScrolled``` by the ```FindFirstVisibleItemPosition``` or ```FindFirstCompletelyVisibleItemPosition``` that exists on both ```LinearLayoutManager``` and ```GridLayoutManager``` the Renderer could notify the Abstraction Layer.

## API Changes
Maybe add a property like ```SelectedIndex``` so the developer will know when the visible item is changed, or maybe just a ScrolledEvent with EventArgs that containes the new Index. This would a CarouselView specific property or could apply to the CollectionView/ItemsView.

## Intended Use Case
This API will be very important when using the ```CarouselView``` with ```SnapPointsType``` to ```MandatorySingle```, so the developer will know when the user moved to the next item. This one can extend the functionality of the CarouselView and it will give us the opportunity of using it as a TabLayout.

EDIT2: On iOS There is the Scrolled Event on the UICollectionViewController. I could make a PR if the proposal gets accepted :)

EDIT: A dummy RVScrollListener which i got a working sample is the following:
The RVScrollListener added OnElementChanged and removed on Dispose, like the following
```cs
_scrollListener = new RVScrollListener(e.NewElement as CustomCarouselView);
AddOnScrollListener(_scrollListener); 
```
```cs
class RVScrollListener : OnScrollListener
{
	private CustomCarouselView _carouselView;

	public RVScrollListener(CustomCarouselView carouselView)
	{
		_carouselView = carouselView;
	}

	public override void OnScrolled(RecyclerView recyclerView, int dx, int dy)
	{
		base.OnScrolled(recyclerView, dx, dy);

		var position = -1;

		var layoutManager = recyclerView.GetLayoutManager();

		switch (layoutManager)
		{
			case GridLayoutManager gridLayoutManager:
				position = gridLayoutManager.FindFirstCompletelyVisibleItemPosition();
				break;
			case LinearLayoutManager linearLayoutManager:
				position = linearLayoutManager.FindFirstCompletelyVisibleItemPosition();
				break;
		}

		if (position != -1)
		{
			_carouselView?.OnScrollChanged(position);
		}
	}
}
```

Of course there must be more checks because the OnScrolled may be called multiple times, if for example the user goes back and forth the same item, so it will fire the event only when the position has been changed.

### LINKED ISSUE #5623: CollectionView with Incremental Collection (RemainingItemsThreshold)
### Description
I don't see any way of working with an incremental collection because I dont kwow when should I call LoadMoreItems method, with a ListView there is an ItemAppearing event, in there you could figure it out when you're reaching the last items so you could load more, in a scrollview we have scrolled event, so you could figure it out when you're reaching the bottom of the list so you could load more items, but with CollectionView I dont see the way to do it.

- Platform Target Frameworks: 

  - Android: 8.1
  - UWP:  Windows 10 version 1803 Build 17134

### LINKED ISSUE #5716: [Enhancement] Add Scrolled event to CollectionView
## Description
CollectionView aims to be more flexible by provide us the ability to know when CollectionView has been scrolled.

Many people have been asking for such an event, and a quick web search for 'xamarin forms listview scroll event' shows it. to make parallax headers, make some views dis/appearing when scroll up/down or other stuff using the scroll event.

### API Event
| API | Description |
| Scrolled  | Occurs when user interact with the CollectionView|

## Implementation
this is exact same signature of ScrollView.Scrolled event .

```public event EventHandler<Xamarin.Forms.ScrolledEventArgs> Scrolled;```

## Backward Compatibility
New feature, so it does not have any impact.

## Difficulty
Hmmm...
CollectionView inherit from ScrollView, it should not be a big deal.

### LINKED ISSUE #6812: implement itemAppearing event in CollectionView
## Summary
want to implement item appearing event in collection view to enable work with pagination with collection view .

### PR DESCRIPTION
**Pull request by adrianknight89**
### Description of Change ###

Not sure if this is the perfect implementation, but here's my take on "load on demand" scrolling for `CollectionView` as well as scroll event notifications. Here's a [demo](https://1drv.ms/v/s!AjlbPgOcTyP2dP_tvMTJjru423E) on Android. Implementation for UWP and Tizen is left out for others.

giphy (2)

**Note:** You should review [UICollectionViewDataSourcePrefetching](https://developer.apple.com/documentation/uikit/uicollectionviewdatasourceprefetching) that was made available with iOS 10 before merging this code. I briefly looked at it but could not figure out how `RemainingItemsThreshold` could be used with prefetching.

### Issues Resolved ### 

- fixes #5623
- fixes #5716 
- fixes #4654
- fixes #6812

### API Changes ###
Added:
- RemainingItemsThreshold
- RemainingItemsThresholdReached (event)
- RemainingItemsThresholdCommand
- Scrolled (event)
- ScrolledEventArgs

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android

### Testing Procedure ###
Added an automated test but could not get it to run due to workspace issues with the UI test projects. For example, despite setting ANDROID_HOME variable, the UI test framework still seems to be confused about the location of this variable. Please run the test and if it doesn't work, please modify it to make sure it does.

### PR Checklist ###

- [x] Has automated tests 

### CODE REVIEW

File: Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Github5623.xaml [hunk 1]
```diff
@@ -0,0 +1,27 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
```
adrianknight89: I had to submit this XAML file because I could not wire up bindings in code behind. @hartez Not sure if it was me, but there might be an issue with `SetBinding` call for `ItemsSource`. Would be great to have tests in place for this.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Github5623.xaml.cs
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		protected override void Init()
+		{
+
+		}
+
+		private async void CollectionView_RemainingItemsThresholdReached(object sender, System.EventArgs e)
+		{
+			await SemaphoreSlim.WaitAsync();
```
adrianknight89: The event/command will be called multiple times if the user scrolls down like a mad cow (which also depends on cell size, web service latency, and handler processing time). We can serialize event handlers and use a control variable (such as continuation token being null after the last web service call) to stop processing unwanted invocations.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Github5623.xaml.cs
```diff
@@ -0,0 +1,168 @@
...
+					return;
+				}
+
+				var nextSet = await GetNextSetAsync();
+
+				Device.BeginInvokeOnMainThread(() =>
+				{
+					foreach (var item in nextSet)
+					{
+						itemsSource.Add(item);
```
adrianknight89: While the web service is async, items must be added to the underlying data source on the main thread. Sometimes there is a small lag when that happens especially when scrolling is continuing. Is there any way to prevent this?

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Github5623.xaml.cs
```diff
@@ -0,0 +1,168 @@
...
+			{
+				RunningApp.ScrollDown();
+
+				elementAvailable = elementQuery.Any();
+				if (elementAvailable)
+					break;
+
+				++attempt;
+
+				if (attempt == 30)
```
adrianknight89: If this is wasting too much time, please tweak `_pageSize` and `_maximumItemCount` to make it more suitable for automation.

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -44,6 +43,23 @@ public IEnumerable ItemsSource
 			set => SetValue(ItemsSourceProperty, value);
 		}
 
+		public static readonly BindableProperty RemainingItemsThresholdReachedCommandProperty = BindableProperty.Create(nameof(RemainingItemsThresholdReachedCommand), typeof(ICommand), typeof(ItemsView), null);
```
adrianknight89: `null` is probably redundant here, y/n?

Xamarin.Forms.Platform.Android/CollectionView/RecyclerViewScrollListener.cs
```diff
@@ -0,0 +1,58 @@
...
+		public override void OnScrollStateChanged(RecyclerView recyclerView, int newState)
+		{
+			base.OnScrollStateChanged(recyclerView, newState);
+		}
+
+		public override void OnScrolled(RecyclerView recyclerView, int dx, int dy)
+		{
+			base.OnScrolled(recyclerView, dx, dy);
+
+			var linearLayoutManager = recyclerView.GetLayoutManager() as LinearLayoutManager;
```
adrianknight89: I assumed that every layout manager is a base `LinearLayoutManager`. Do you expect this to be false in the future?

Xamarin.Forms.Platform.iOS/CollectionView/UICollectionViewDelegator.cs
```diff
@@ -21,6 +21,20 @@ public UICollectionViewDelegator(ItemsViewLayout itemsViewLayout, ItemsViewContr
 
 		public override void WillDisplayCell(UICollectionView collectionView, UICollectionViewCell cell, NSIndexPath path)
 		{
+			switch (ItemsViewController.ItemsView.RemainingItemsThreshold)
+			{
+				case -1:
+					return;
+				case 0:
+					if (path.Row == ItemsViewController.ItemsViewSource.Count - 1)
```
adrianknight89: @hartez If I'm not mistaken, you were using `indexPath.Row` throughout `CollectionView` implementation, but I believe the correct property is `indexPath.Item` for collections. Here's a [discussion](https://stackoverflow.com/a/14765781/246392) on that. I left this to be `Row` for now, but we should consider switching to `Item`.

samhouts:
@adrianknight89 Would you mind rebasing to 4.0.0? Thanks!!

adrianknight89:
@samhouts These rebases are a nightmare. :) I think it's done.

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -44,6 +43,23 @@ public IEnumerable ItemsSource
...
+
+		public ICommand RemainingItemsThresholdReachedCommand
+		{
+			get { return (ICommand)GetValue(RemainingItemsThresholdReachedCommandProperty); }
+			set { SetValue(RemainingItemsThresholdReachedCommandProperty, value); }
+		}
+
+		public static readonly BindableProperty RemainingItemsThresholdProperty =
```
adrianknight89: Should these properties be added to `CollectionView` directly instead of the base class? I guess there is a possibility that somebody might want to lazy load `CarouselView` items.

adrianknight89:
@hartez @StephaneDelcroix I'd like to have an event to know which items are visible on the screen. This event can allow people to create their own custom `infinite` scroll, but it also lets people do interesting things like play an animation or video once the items are visible and pause/stop them when they are not. I do wonder if it makes sense for us to expose this event instead of implementing `RemainingItemsThreshold`. Or, should we have two events, one for `RemainingItemsThreshold` and one for item visibility and let people subscribe to them at will?

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -114,8 +130,15 @@ public ItemSizingStrategy ItemSizingStrategy
 			OnScrollToRequested(new ScrollToRequestEventArgs(item, group, position, animate));
 		}
 
+		public void SendRemainingItemsThresholdReached()
```
adrianknight89: Does this one need `[EditorBrowsable(EditorBrowsableState.Never)]`?

Xamarin.Forms.Platform.iOS/CollectionView/UICollectionViewDelegator.cs
```diff
@@ -19,6 +22,45 @@ public UICollectionViewDelegator(ItemsViewLayout itemsViewLayout, ItemsViewContr
...
+
+		public override void DraggingEnded(UIScrollView scrollView, bool willDecelerate)
+		{
+			_horizontalOffset = 0;
+			_verticalOffset = 0;
+		}
+
+		public override void Scrolled(UIScrollView scrollView)
+		{
+			var indexPathsForVisibleItems = ItemsViewController.CollectionView.IndexPathsForVisibleItems.OrderBy(x => x.Row);
```
adrianknight89: Index paths do not arrive sorted, so we sort them here before reporting them to `Core`.

Xamarin.Forms.Platform.Android/CollectionView/ItemsViewRenderer.cs
```diff
@@ -483,5 +492,77 @@ internal void UpdateEmptyViewVisibility()
...
+
+		public class CustomGridLayoutManager : GridLayoutManager
+		{
+			public CustomGridLayoutManager(Context context, int spanCount, int orientation, bool reverseLayout) : base(context, spanCount, orientation, reverseLayout)
+			{
+
+			}
+		}
+
+		public class CustomLinearLayoutManager : LinearLayoutManager
```
adrianknight89: Extended the default layout managers because Android calculates horizontal/vertical scroll offsets by first calculating an average row height instead of looking at every row height. Unlike iOS, `RecyclerView` does not report actual offsets. See [here](https://stackoverflow.com/a/32863117/246392).

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Github5623.xaml.cs
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	[Preserve(AllMembers = true)]
+	public class ViewModel5623
+	{
+		public ObservableCollection<Model5623> Items { get; set; }
+
+		public Command RemainingItemsThresholdReachedCommand { get; set; }
+
+		public ItemSizingStrategy ItemSizingStrategy { get; set; } = ItemSizingStrategy.MeasureAllItems;
```
adrianknight89: This strategy doesn't work well on iOS (see the first item as it isn't measured/laid out properly). Not sure if this is a known bug. When testing on iOS, you might want to change this to first item measure so every row gets the same size.

Xamarin.Forms.Core/Items/ScrolledEventArgs.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+
+namespace Xamarin.Forms.Core.Items
+{
+	public class ScrolledEventArgs : EventArgs
```
adrianknight89: Should this extend the existing `ScrolledEventArgs` (which `CarouselView` spec seems to indicate)?

Xamarin.Forms.Platform.Android/CollectionView/RecyclerViewScrollListener.cs
hartez: We expect that someone will eventually want to use a LayoutManager that is _not_ derived from LinearLayoutManager. So we try our best to do something reasonable when this happens. For now, I'd suggest just not firing the event if the LayoutManager is not a LinearLayoutManager.

Ideally, we'd add some sort of interface that other LayoutManagers could implement so we could interrogate them about the remaining items.

Xamarin.Forms.Core/Items/ItemsView.cs
hartez: No.

Xamarin.Forms.Platform.Android/CollectionView/RecyclerViewScrollListener.cs
adrianknight89: When the layout manager is not LinearLayoutManager, then we send -1 for the item indices. This can be documented for devs.

Xamarin.Forms.Core/Items/ItemsView.cs
StephaneDelcroix: no, it's fine. defaultValue parameter shouldn't have been made optional.

StephaneDelcroix: a `Command`, but no `CommandParameter` ?

StephaneDelcroix:
```suggestion
		public static readonly BindableProperty RemainingItemsThresholdReachedCommandProperty =
			BindableProperty.Create(nameof(RemainingItemsThresholdReachedCommand), typeof(ICommand), typeof(ItemsView), null);
```

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -44,6 +43,23 @@ public IEnumerable ItemsSource
 			set => SetValue(ItemsSourceProperty, value);
 		}
 
+		public static readonly BindableProperty RemainingItemsThresholdReachedCommandProperty = BindableProperty.Create(nameof(RemainingItemsThresholdReachedCommand), typeof(ICommand), typeof(ItemsView), null);
+
+		public ICommand RemainingItemsThresholdReachedCommand
+		{
+			get { return (ICommand)GetValue(RemainingItemsThresholdReachedCommandProperty); }
```
StephaneDelcroix:
```suggestion
			get => (ICommand)GetValue(RemainingItemsThresholdReachedCommandProperty);
```

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -44,6 +43,23 @@ public IEnumerable ItemsSource
...
+			set { SetValue(RemainingItemsThresholdReachedCommandProperty, value); }
```
StephaneDelcroix:
```suggestion
			set => SetValue(RemainingItemsThresholdReachedCommandProperty, value);
```

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -44,6 +43,23 @@ public IEnumerable ItemsSource
...
+		}
+
+		public static readonly BindableProperty RemainingItemsThresholdProperty =
+			BindableProperty.Create(nameof(RemainingItemsThreshold), typeof(int), typeof(ItemsView), -1, validateValue: (bindable, value) => (int)value >= -1);
```
StephaneDelcroix: setting the value to -1 shouldn't be allowed. `ClearValue` is there for that

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -133,5 +163,18 @@ protected virtual void OnScrollToRequested(ScrollToRequestEventArgs e)
 		{
 			ScrollToRequested?.Invoke(this, e);
 		}
+
+		protected virtual void OnRemainingItemsThresholdReached()
+		{
+			RemainingItemsThresholdReached?.Invoke(this, EventArgs.Empty);
```
StephaneDelcroix: don't fire events from a virtual method, but fire them from SendRemainingItemsThresholdReached(). same for triggering the Command.

then, it's fine to keep the virtual method with an empty body

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -133,5 +163,18 @@ protected virtual void OnScrollToRequested(ScrollToRequestEventArgs e)
...
+
+			if(RemainingItemsThresholdReachedCommand?.CanExecute(null) == true)
+				RemainingItemsThresholdReachedCommand?.Execute(null);
+		}
+
+		protected virtual void OnScrolled(Core.Items.ScrolledEventArgs e)
+		{
+			Scrolled?.Invoke(this, e);
```
StephaneDelcroix: see ^^^

Xamarin.Forms.Core/Items/ScrolledEventArgs.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+
+namespace Xamarin.Forms.Core.Items
```
StephaneDelcroix: why a new namespace ? please avoid

Xamarin.Forms.Core/Items/ScrolledEventArgs.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+	public class ScrolledEventArgs : EventArgs
+	{
+		public double HorizontalDelta { get; private set; }
```
StephaneDelcroix:
```suggestion
		public double HorizontalDelta { get; }
```

same for other auto-properties

Xamarin.Forms.Core/Items/ScrolledEventArgs.cs
```diff
@@ -0,0 +1,32 @@
...
+
+		public double VerticalOffset { get; private set; }
+
+		public int FirstVisibleItemIndex { get; private set; }
+
+		public int CenterItemIndex { get; set; }
+
+		public int LastVisibleItemIndex { get; private set; }
+
+		public ScrolledEventArgs(double horizontalDelta, double verticalDelta, double horizontalOffset, double verticalOffset, int firstVisibleItemIndex, int centerItemIndex, int lastVisibleItemIndex)
```
StephaneDelcroix: that's a hell of a signature ! shouldn't the calling code set the properties instead ?

Xamarin.Forms.Core/Items/ScrolledEventArgs.cs
adrianknight89: How should this be handled? Extend `Xamarin.Forms.ScrolledEventArgs` with a new class in the same namespace or add new properties and ctor to it?

Xamarin.Forms.Core/Items/ItemsView.cs
adrianknight89: What would a typical `CommandParameter` look like in this case?

Xamarin.Forms.Core/Items/ScrolledEventArgs.cs
adrianknight89: I'm fine with this. I was trying to match the existing constructor in the other class since it didn't make public setters available.

Xamarin.Forms.Core/Items/ItemsView.cs
adrianknight89: There are many examples in the code where setting the value to default is allowed. Do you still want me to change this?

Phenek:
It would be awesome if the ScrolledEventArgs could provide us some GestureStatus,
similar to [PanUpdatedEventArgs](https://docs.microsoft.com/fr-fr/dotnet/api/xamarin.forms.panupdatedeventargs?view=xamarin-forms).StatusType Property

```public enum GestureStatus``` 

| Enum | Int |Description|
| ------------- | ------------- |------------- |
|Canceled | 3 | The gesture was canceled|
|Completed | 2 | The gesture completed|
|Running | 1 | The gesture is still being recognized|
|Started | 0 | The gesture started|

regards,

File: Xamarin.Forms.Core/Items/ItemsView.cs (hunk 8 above)
Comment (hartez) [reply Xamarin.Forms.Core/Items/ItemsView.cs]: Yes, these should be on ItemsView so "infinite" CarouselViews are possible.

**Comment by adrianknight89** [comment]
@StephaneDelcroix can this be reviewed again?

**Comment by ahoefling** [comment]
@adrianknight89 this is an AWESOME PR! Great job on the hard work you did here. Do you have any samples of using this as a custom renderer while we wait for this to get merged into the platform?

To Reviews:
Is there any updates on the timeline for this? I am currently working on an app and was hoping to use the new `CollectionView` and `CarouselView`. I need to change the UI as the items are `Scrolled` and without this, it creates a barrier as I may need use one of the open source plugins instead of the new platform control.

**Comment by adrianknight89** [comment]
@ahoefling Thanks Andrew. I don't have custom renderers set up for this. You could copy the code and try it yourself though I think you might have an issue on iOS since the code changes the accessibility modifiers of two properties (not sure if you could use reflection as a workaround). I agree that not having scroll functionality is a showstopper to many people. I'm hoping the team will review the code as soon as they can / want to.

**Comment by jleopore** [comment]
I'm also hoping this gets into a release soon.

File: Xamarin.Forms.Platform.Android/CollectionView/RecyclerViewScrollListener.cs [hunk 22]
```diff
@@ -0,0 +1,116 @@
...
+			var centerItemIndex = -1;
+
+			if (recyclerView.GetLayoutManager() is LinearLayoutManager linearLayoutManager)
+			{
+				firstVisibleItemIndex = linearLayoutManager.FindFirstVisibleItemPosition();
+				lastVisibleItemIndex = linearLayoutManager.FindLastVisibleItemPosition();
+				centerItemIndex = CalculateCenterItemIndex(firstVisibleItemIndex, lastVisibleItemIndex, linearLayoutManager);
+			}
+
+			var itemsViewScrolledEventArgs = new ItemsViewScrolledEventArgs
```
ahoefling: If I am understanding the code and intent correctly here it is retrieving the index of the item in the center. Given 3 items on screen:
```
[0] [1] [2]
```
The `centerItemIndex` will be 1. Then if the screen continues on showing
```
[2] [3] [4]
```
The `centerItemIndex` will be 3. After performing the calculation it creates the event `ItemsViewScrolledEventArgs`.

It appears that we are sending the event back up the stack with every cycle of the `OnScrolled` leaving the Shared Code to handle when the new item has been scrolled to. With the integration I am attempting this makes it more difficult to handle the `Scrolled` event when the shared code only cares about when the new center item has changed.

@adrianknight89 if everything I stated above is correct, I would propose the following change to this as it is the exact modification I am making to your PR to get through this limitation on my project.

# API Changes
+ Add event to `RecyclerViewScrollListener` - `IndexChanged`
+ Add event to `ItemsView` - `ItemChanged`

# Usage
In the shared code the CarouselView will then add an event listener for `ItemChanged` to perform any logic
```c#
var carouselView = new CarouselView();
carouselView.ItemChanged += (s,e) =>
{
    // business rules for when the item changes
};
```

# Code Snippets
IndexChangedEventArgs
```c#
public class IndexChangedEventArgs : EventArgs
{
    public int Index { get; set; }
    public IndexChangedEventArgs(int index)
    {
        Index = index;
    }
}
```

Change line 45 to this
```c#
var centerItemIndex = CalculateCenterItemIndex(firstVisibleItemIndex, lastVisibleItemIndex, linearLayoutManager);

if (_currentIndex != centerItemIndex)
{
    _currentIndex = centerItemIndex;
    ItemChange?.Invoke(this, new IndexChangedEventArgs(_currentIndex));
}
```
Change line 295 of `ItemsViewRenderer`
```c#
_recycleViewScrollListener = new RecyclerViewScrollListener(ItemsView, ItemsViewAdapter);
_recycleViewScrollListener.ItemChange += (s, e) =>
{
    var index = 0;
    foreach (var item in ItemsView.ItemsSource)
    {
        if (index == e.Index && ItemsView is IItemChanged itemChanged)
        {
            itemChanged.OnItemChanged(new SelectedItemChangedEventArgs(item, index));   
            break;
        }
        index++;
    }
}
AddOnScrollListener(_recycleViewScrollListener);
```

Add the event and create a helper method for invoking the event
```c#
public class ItemsView : View, IItemChanged
{
    // …

    public event EventHandler<SelectedItemChangedEventArgs> ItemChanged;
    void IItemChanged.OnItemChanged(SelectedItemChangedEventArgs eventArgs)
    {
        ItemChanged.Invoke(this, eventArgs);
    }
}
```

@adrianknight89 I understand your PR is trying to accomplish 2 items but 
In my case I added an IItemChanged interface to simplify the invocation in the renderer
```c#
public interface IItemChanged
{
    void OnItemChanged(SelectedItemChangedEventArgs eventArgs);
}
```

## Explanation
The idea here is to only throw an event when the `centerItemIndex` is changed. Once that event is handled in the `ItemsViewRenderer` it can then retrieve the current item. The renderer can then fire an event that includes both 
- IndexChanged
- ItemChanged

This now makes the usage in the shared code very simple

ahoefling:
@adrianknight89 I took your code and got it working in a custom renderer. It didn't do exactly what I would expect and thus have added a very verbose comment in the code with a proposal to make the API work how I would expect. 

Again you did a lot of great work in this PR, I can't stress this enough. I think my recommendation will simplify any integration to have an event fire only when the centered item is changed. Looking forward to discussing more in the thread.

https://github.com/xamarin/Xamarin.Forms/pull/5754#discussion_r296432557

Xamarin.Forms.Platform.Android/CollectionView/RecyclerViewScrollListener.cs
adrianknight89: @ahoefling I remember thinking about a similar thing at the time of writing the PR, but I couldn't decide if the team would like to see multiple events or a single one to track changes. Remember that `ItemsViewScrolledEventArgs` has multiple properties, and it is very likely for horizontal or vertical offsets to change while visible item indexes remain the same, hence the event firing during every scroll change.

As for your last comment with respect to throwing an event only when `centerItemIndex` changes, I'm not sure if this is the right approach. What happens if the row count changes dynamically such that the center index stays the same while others change? For example, if we have items 0 1 2 & `_currentIndex` is 1, and we remove the last item and scroll the collection, current index should stay the same while last index should be set to 1. What I'm trying to get at is just because `_currentIndex`  remains the same does not mean other indexes have not changed.

adrianknight89: That said, we could create an `ItemChanged` event and leave offset changes separate and fire `ItemChanged` only when any of the three indexes changes instead of firing it with every offset change. I'd like to hear what the team thinks before moving forward.

gsgou:
@ahoefling could you pls share your custom renderer if possible?
I would like to use it in an active project else i will have
to switch to FastGrid until the feature is done.

samhouts:
@adrianknight89 Would you mind rebasing?

@hartez Can you weigh in on @ahoefling's suggestion?

Thanks!!

adrianknight89:
@samhouts Should be done.

Xamarin.Forms.Platform.Android/CollectionView/RecyclerViewScrollListener.cs
ahoefling: @adrianknight89 thanks for the detailed clarification, it is helpful for me to understand the intent of the change.

I was only thinking in terms of a CarouselView or horizontal sliding view. In that case you only really need to deal with 1 row of items at a time. On the project that I modified your code I added the event to handle when the center item of a CarouselView has changed.

Perhaps my question/proposal isn't suited for this PR but for a follow-up item that will target the CarouselView specifically (or any 1 row horizontally CollectionView).

I'm curious to hear when the Xamarin Team and others have to say on this as well

samhouts: >Perhaps my question/proposal isn't suited for this PR but for a follow-up item that will target the CarouselView specifically (or any 1 row horizontally CollectionView).

@ahoefling That sounds perfect to me! Please open a new issue. We'll go ahead and merge this one and then the  team and community will be better prepared to discuss your suggestions. Thanks!!!

samhouts:
@adrianknight89 Uh oh! Build is failing...

```

  | Xamarin.Forms.Platform.Android\CollectionView\ItemsViewRenderer.cs(39,4): Error CS0234: The type or namespace name 'VerifyCollectionViewFlagEnabled' does not exist in the namespace 'Xamarin.Forms.Platform.Android.CollectionView' (are you missing an assembly reference?)
-- | --
  | Xamarin.Forms.Platform.Android\CollectionView\PositionalSmoothScroller.cs(12,4): Error CS0234: The type or namespace name 'VerifyCollectionViewFlagEnabled' does not exist in the namespace 'Xamarin.Forms.Platform.Android.CollectionView' (are you missing an assembly reference?)
  | Xamarin.Forms.Platform.Android\CollectionView\SnapManager.cs(14,4): Error CS0234: The type or namespace name 'VerifyCollectionViewFlagEnabled' does not exist in the namespace 'Xamarin.Forms.Platform.Android.CollectionView' (are you missing an assembly reference?)
  | Xamarin.Forms.Platform.Android\CollectionView\ItemsViewAdapter.cs(24,4): Error CS0234: The type or namespace name 'VerifyCollectionViewFlagEnabled' does not exist in the namespace 'Xamarin.Forms.Platform.Android.CollectionView' (are you missing an assembly reference?)
  | Xamarin.Forms.Platform.Android\CollectionView\EmptyViewAdapter.cs(46,4): Error CS0234: The type or namespace name 'VerifyCollectionViewFlagEnabled' does not exist in the namespace 'Xamarin.Forms.Platform.Android.CollectionView' (are you missing an assembly reference?)
  | Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Github5623.xaml.cs(84,4): Error CS0103: The name 'Label1' does not exist in the current context
  | Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Github5623.xaml.cs(85,4): Error CS0103: The name 'Label2' does not exist in the current context
  | Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Github5623.xaml.cs(86,4): Error CS0103: The name 'Label3' does not exist in the current context
  | Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Github5623.xaml.cs(87,4): Error CS0103: The name 'Label4' does not exist in the current context
  | Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Github5623.xaml.cs(88,4): Error CS0103: The name 'Label5' does not exist in the current context


```

adrianknight89:
@samhouts All those issues should be related to the same problem with `VerifyCollectionViewFlagEnabled`. I couldn't get the solution to compile so I commented out the lines where the verification is done. I thought your setup wouldn't have this issue. Not sure if adding `using Xamarin.Forms.Platform.Android.CollectionView;` is messing up with anything, but let me look at this in a few hours.

Xamarin.Forms.Platform.Android/CollectionView/RecyclerViewScrollListener.cs
ahoefling: Perfect! I'll reference the PR in the new issue

adrianknight89: @ahoefling If there is only one item on the screen, then it makes sense to track center item index changes. However, there will be cases when people want to use CarouselView with multiple visible items on the screen because each item is smaller than the width of the screen.

ahoefling: @samhouts New issue created: #6756

adrianknight89:
@samhouts Can you build again?

Also, I can't verify the UI test due to issues with my workspace. Can you please verify that it works? If not, you'll need to make adjustments to it. Thanks!

samhouts:
@adrianknight89 Now it's:

```

Xamarin.Forms.Platform.iOS\CollectionView\UICollectionViewDelegator.cs(87,12): Error CS8107: Feature 'default literal' is not available in C# 7.0. Please use language version 7.1 or greater.
--
  | Xamarin.Forms.Platform.iOS\CollectionView\UICollectionViewDelegator.cs(98,12): Error CS8107: Feature 'default literal' is not available in C# 7.0. Please use language version 7.1 or greater.
  | Xamarin.Forms.Platform.iOS\CollectionView\UICollectionViewDelegator.cs(109,12): Error CS8107: Feature 'default literal' is not available in C# 7.0. Please use language version 7.1 or greater.
  | Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Github5623.xaml.cs(84,4): Error CS0103: The name 'Label1' does not exist in the current context
  | Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Github5623.xaml.cs(85,4): Error CS0103: The name 'Label2' does not exist in the current context
  | Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Github5623.xaml.cs(86,4): Error CS0103: The name 'Label3' does not exist in the current context
  | Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Github5623.xaml.cs(87,4): Error CS0103: The name 'Label4' does not exist in the current context
  | Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Github5623.xaml.cs(88,4): Error CS0103: The name 'Label5' does not exist in the current context
  | Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Github5623.xaml.cs(89,4): Error CS0103: The name 'Label6' does not exist in the current context
  | Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Github5623.xaml.cs(90,4): Error CS0103: The name 'Label7' does not exist in the current context


```

adrianknight89:
@samhouts reverted `default` changes. The labels are defined in XAML and I can run the Gallery on my device. What's the trick to getting them recognized by the build process?

adrianknight89:
@samhouts I rebased again. It looks like the sample code is failing on iOS, especially when we scroll down rapidly after the view is loaded initially. Unfortunately, between March and now, a lot of code was added, which seems to have broken the work done here.

When fast scroll occurs, I see white screen (and noticed the index paths for visible items return nothing). I don't remember observing this back in March. @hartez Was something added to make sure this is the case now?

I added count check [here](https://github.com/xamarin/Xamarin.Forms/pull/5754/files#diff-c0546c1c6ae62790675b0a2e01c9fb85R41) to help fix this issue.

adrianknight89:
You could build now.

adrianknight89:
@hartez Another thing I see is that items resize as scrolling happens and new items are being fetched. You could observe this when 5623 is loaded initially and you scroll down just a bit to trigger the next load. Again, I don't recall seeing this issue in the past. 

Items should not resize during scrolling (at least this would be the feature I want). I think this might be due to the built-in row animations that make items spring back and forth to make room for the new ones.

Xamarin.Forms.Platform.Android/CollectionView/RecyclerViewScrollListener.cs
```diff
@@ -0,0 +1,116 @@
...
+					break;
+			}
+		}
+
+		static int CalculateCenterItemIndex(int firstVisibleItemIndex, int lastVisibleItemIndex, LinearLayoutManager linearLayoutManager)
+		{
+			var keyValuePairs = new Dictionary<int, int>();
+			for (var i = firstVisibleItemIndex; i <= lastVisibleItemIndex; i++)
+			{
+				var view = linearLayoutManager.FindViewByPosition(i);
```
ahoefling: When a `CollectionView` is reset during the app execution, consider the state of the app changes and it requires a `Clear()` on the `ObservableCollection<T>` that is bound to y our `CollectionView.ItemsSource`. 

In this scenario the `firstVisibleItemIndex` can be less than `0`. Typically the index would be reset to `-1`. The code as provided will throw a `NullReferenceException` at line 84 where it attempts to use the returned `view`.

My solution was to put a check at the top of the function:

```c#
if (firstVisibleItemIndext < 0)
    return -1;
```

@adrianknight89 what are your thoughts on this?

adrianknight89: We shouldn't expect to see -1 at all. What you're seeing is probably due to the fact that a layout pass has not occurred yet. I'm not sure if this issue needs to be addressed somewhere else. I ran into a similar issue on iOS today and made a similar fix. I'll update the Android side as well. Thanks!

ahoefling: No problem, I saw it and wanted to make sure you were aware. Perhaps, we should put some check or validation higher up the call stack so it never happens in the calculation method.

adrianknight89:
@samhouts Sorry that you've had to build several times. If you guys could tell me how to resolve the build issues with variable names, I'll make another change, but as I said, my workspace is fine.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Github5623.xaml.cs
```diff
@@ -0,0 +1,201 @@
...
+			}
+			finally
+			{
+				SemaphoreSlim.Release();
+			}
+		}
+
+		void CollectionView_OnScrolled(object sender, ItemsViewScrolledEventArgs e)
+		{
+			Label1.Text = "HorizontalDelta: " + e.HorizontalDelta;
```
jfversluis: Wrap this in a `#if APP` directive for the (UITests projects) build to succeed

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Github5623.xaml.cs
```diff
@@ -0,0 +1,201 @@
...
+
+				return collection;
+			});
+		}
+
+#if UITEST
+		[Test]
+		public void CollectionViewInfiniteScroll()
+		{
+			RunningApp.WaitForElement (q => q.Marked ("CollectionView"));
```
jfversluis: Locally this times out (on Android). No clue why, adding an explicit `AutomationId` to the `CollectionView` seems to help. Personally I like to avoid the object names as identifiers (i.e. `x:Name="CollectionView"`) but not sure if that has anything to do with it

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Github5623.xaml.cs
```diff
@@ -0,0 +1,201 @@
...
+
+				elementAvailable = elementQuery.Any();
+				if (elementAvailable)
+					break;
+
+				++attempt;
+
+				if (attempt == 30)
+				{
+					Assert.Fail("Failed to find the last element");
```
jfversluis: After adding the above, it then fails on this line. It seems the UI test needs some work :)

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Github5623.xaml.cs
adrianknight89: Appended `5623` to the name. Let's see if it fixes the issue. Looking up objects by their name should in theory work without the need for an explicit automation id.

jfversluis: I agree! That should definitely work. I added a `Repl` call in there and this query came up empty. Since the `AutomationId` is meant to be used for this purpose I tried that and it worked. I didn't try this variation though :)

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Github5623.xaml.cs
adrianknight89: As I've told the team before, I haven't been able to run the UI tests due to errors with my setup. In the past, I've set the `ANDROID_HOME` variable, but the test library wasn't able to find it. I've also seen dlls constantly locked on the C drive along with symbols not loaded. It'd be great if Xamarin can provide clear instructions on how to run the UI tests in this solution.

jfversluis: My apologies, I must have overlooked that. I'm assuming you're on Windows then, since you're not mentioning iOS?

adrianknight89: That said, if the UI test continues to fail, I'll let the team take it from here and fix it for us all. :)

adrianknight89: Correct. Win 10.

jfversluis: Are you on the Xamarin Chat slack? Or have some other way to reach you? Or worst case can you reach out to gerald.versluis@microsoft.com? Let's see if we can get it to work and from there I'll create a little writeup to make sure this doesn't happen to other people

adrianknight89: Emailed.

adrianknight89:
@jfversluis I moved private variables into the `APP` scope. Can you trigger another build?

jfversluis:
OK, fixed the test that you have included. Idea is the same, implementation a bit different :)

I do feel there is more that we can automate with testing here, no?

adrianknight89:
@jfversluis We can also scroll back to the top after seeing index 99, so reverse the automation maybe? :) Also, we can play around with drag coordinates. At every iteration, we can decide to drag by a random amount on the y coordinate instead of a fixed amount. Another one could be mixing drags with item scrolling (i.e. drag on the y coordinate and `ScrollTo` an item / index) to see how the two play together.

jfversluis:
Ideally we would want to check the values that are changing like the delta, offset, centered index etc. and see if all that checks out. Not sure if that is possible without too much hassle with a UI test though.

jfversluis:
I do need to tweak this test though. Now it triggers the iOS panel 🙈 

image

adrianknight89:
@jfversluis I just rebased to resolve conflicts after the newly-added scrollbar visibility changes, but it looks like I lost your changes. Would you mind pushing them again? Sorry for the inconvenience. :(

jfversluis:
Failing test is not related to this change

jfversluis:
@StephaneDelcroix could you give this another look? @hartez still OK with this?

hartez:
@jfversluis Yeah, I'm good.

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -158,5 +198,15 @@ protected virtual void OnScrollToRequested(ScrollToRequestEventArgs e)
 		{
 			ScrollToRequested?.Invoke(this, e);
 		}
+
+		protected virtual void OnRemainingItemsThresholdReached(EventArgs e)
```
StephaneDelcroix: there's probably no value in passing the EventArgs here

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -139,8 +156,31 @@ public ItemSizingStrategy ItemSizingStrategy
...
+			Scrolled?.Invoke(this, e);
+
+			OnScrolled(e);
+		}
+
 		public event EventHandler<ScrollToRequestEventArgs> ScrollToRequested;
 
+		public event EventHandler<ItemsViewScrolledEventArgs> Scrolled;
+
+		public event EventHandler<EventArgs> RemainingItemsThresholdReached;
```
StephaneDelcroix: could use the non-generic one

Xamarin.Forms.Core/Items/ItemsView.cs
```diff
@@ -44,6 +43,15 @@ public IEnumerable ItemsSource
...
 		}
 
+		public static readonly BindableProperty RemainingItemsThresholdReachedCommandProperty = 
+			BindableProperty.Create(nameof(RemainingItemsThresholdReachedCommand), typeof(ICommand), typeof(ItemsView), null);
+
+		public ICommand RemainingItemsThresholdReachedCommand
+		{
+			get => (ICommand)GetValue(RemainingItemsThresholdReachedCommandProperty);
+			set => SetValue(RemainingItemsThresholdReachedCommandProperty, value);
+		}
```
StephaneDelcroix: every command usually has it's CommandParameter property too

Xamarin.Forms.Core/Items/ItemsView.cs
StephaneDelcroix: do we really need an event, a command, and a virtual method ?

adrianknight89:
@jfversluis Can you integrate changes in [this gist](https://gist.github.com/adrianknight89/b00773294a68b87748296adcd6a3df23) into `ItemsView`? I want to avoid all that merge / rebase stuff by pushing a change on top of yours. :) This should address most of @StephaneDelcroix's review.

jfversluis:
Of course I want to do it for you! But you’re peeking my interest… Why aren’t you able to do it yourself? Is there something I can help you with in that regard?

adrianknight89:
Nevermind. I thought it'd be like the last time due to your changes. I hate going through Vim and VS has a tendency to convert everything into a giant merge, but looks like all is done at this point.

clergym3n:
Can we see this merged soon? We're still waiting to be able to switch to CollectionView, but this along with RefreshView has been a roadblock.

mbalous:
Is there an ETA for CarouselView on UWP? Thank you

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Github5623.xaml.cs
Dids: @adrianknight89 Did this ever get resolved? Been trying to optimize CarouselView content loading for a while now, especially when loading/unloading content on the fly, but I still have a split second of blocking UI thread lag.

PR #5771, opened 2019-04-01
Add covariance back to IElementConfiguration

PureWeen:
### Description of Change ###
Fast Renderer UI tests broke because IElementConfiguration<T> can no longer be covariantly cast which makes us realize that removing the out parameter will be too breaking. It'll also be breaking in such a way that's not obvious. 

Reverts #5449

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS
- Android
- UWP

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/IElementConfiguration.cs
+++ b/Xamarin.Forms.Core/IElementConfiguration.cs
@@ -2,7 +2,7 @@
 namespace Xamarin.Forms
 {
 	// Don't make this generic covariant as it causes UWP performance to tank
-	public interface IElementConfiguration<TElement> where TElement : Element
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Core/IElementConfiguration.cs
```diff
@@ -2,7 +2,7 @@
 namespace Xamarin.Forms
 {
 	// Don't make this generic covariant as it causes UWP performance to tank
```
samhouts: Leaving this comment here will make future devs' heads explode.

PureWeen: haha oh yea
Once UI Tests pass I'll remove before merging

PR #5775, opened 2019-04-01
[C] SearchHandler.SelectedItem

### LINKED ISSUE #5712: [Shell]SearchHandler should include SelectedItem property
### Description

for the cases when you do not want to inherit from it...
### Steps to Reproduce

1. 
2. 
3. 

### Basic Information

- Version with issue:
- Last known good version:
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Nuget Packages:
- Affected Devices:

StephaneDelcroix:
### Description of Change ###

Add a selectedItem property for the case where you do not want to subclass SearchHandler

### Issues Resolved ### 

- fixes #5712

### API Changes ###

Added:
 - public static readonly BindableProperty SearchHandler.SelectedItemProperty;
 - public SearchHandler.SelectedItem { get; }
 

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Shell/SearchHandler.cs
+++ b/Xamarin.Forms.Core/Shell/SearchHandler.cs
@@ -7,8 +7,6 @@ namespace Xamarin.Forms
 {
 	public class SearchHandler : BindableObject, ISearchHandlerController
 	{
-		#region ISearchHandlerController
-
 		event EventHandler<ListProxyChangedEventArgs> ISearchHandlerController.ListProxyChanged
 		{
 			add { _listProxyChanged += value; }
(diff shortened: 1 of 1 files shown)


StephaneDelcroix:
> Needs a SelectedItemChanged event as well

there is INPC, there's a virtual OnItemSelected(), I don't think we need an event. Events aren't MVVM-friendly, and for each event we'd need a Command and CommandParameter... Events make sense in case like Button.Clicked, but I'm not sure there's a need here... What do you think ?

PureWeen:
> there is INPC, there's a virtual OnItemSelected(), I don't think we need an event. Events aren't MVVM-friendly, and for each event we'd need a Command and CommandParameter... Events make sense in case like Button.Clicked, but I'm not sure there's a need here... What do you think ?

That's fine

My only additional thought would be that for ItemsSource/List based structures this makes our apis a bit inconsistent. 

- ListView - Has an ItemSelected event but no Command
- CollectionView - Has an selection event and command
- SearchHandler - has just INPC

So I get that INPC is enough for now but precedence would make me look for an event for ItemSelected

StephaneDelcroix:
> Most use cases the user picks a option the sets the SelectedItem, but wouldn't devs wan't to set this to highlight a selection for example?

we prevent this by using a `robp`

PR #5804, opened 2019-04-03
XamlLoader should not ignore x:TypeArguments on root node

TimBarham:
### Description of Change ###

`XamlLoader.Create()` does not resolve root type if it is a generic. Also, it provides the wrong type name to `FallbackTypeResolver` (it provides the base type name, rather than, for example, ``TypeName`1``). This is because it ignores the `x:TypeArguments` attribute when determining the type name.

### Issues Resolved ### 

- fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/815082

### API Changes ###

None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###
Manually tested in previewer. Added test to `DesignTimeLoaderTests`.

### PR Checklist ###

- [x] Has automated tests

### DIFF
--- a/Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
+++ b/Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
@@ -607,7 +607,6 @@ namespace Xamarin.Forms.Xaml.UnitTests
 			string clrNamespace = null;
 			string typeName = null;
 
-
 			XamlLoader.FallbackTypeResolver = (fallbackTypeInfos, type) =>
 			{
 				assemblyName = fallbackTypeInfos?[1].AssemblyName;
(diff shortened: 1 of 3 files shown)


StephaneDelcroix:
👎 from a XAML inflating point of view, this is not necessary, so this is augmenting the XAML only for the purpose of the previewer. (for reference, see last week discussion about root-level elements).

Even if we add support for x:TypeArguments we're left with the following issue:
- do they map to the same type arguments defined in code-behind ? if they don't, which one is right ?
- as this is not enforced (not required for it to be valid xaml), users will not declare x:TypeArguments, so this feature will never be used
- even if we're able to report the right open generic type, that doesn't help instantiating the type at all

TimBarham:
I'm not sure I understand what you are saying. The following scenario is entirely supported (but it fails in the previewer if we're not rendering `MainPage`):

I have the following XAML:
```
<local:BaseContentPage xmlns="http://xamarin.com/schemas/2014/forms"
    xmlns:local="clr-namespace:MyApp"
    x:TypeArguments="x:Object"
    x:Class="MyApp.MainPage" />
```
With the following code behind:
```
public partial class MainPage : BaseContentPage<object>
{
    public MainPage() => InitializeComponent();
}
```
And the following implementation of `BaseContentPage`:
```
public abstract class BaseContentPage<T> : ContentPage {}
```
To answer your questions...

* The `x:TypeArguments` attribute is indeed required if `BaseContentPage` is a generic type. The project won't compile if it isn't there, or if it doesn't specify the same type as the code behind.

* It is enforced. The project won't compile without it.

* It does help instantiating the type, because `XamlParser.GetElementType()` processes the `x:TypeArguments` attribute, and calls `Type.MakeGenericType()` to create the constructed generic type.

In the previewer, if `MainPage` is a type we are going to instantiate, then there is no problem. We simply instantiate it and everything works.

But if it isn't, we inflate the XAML by calling `XamlLoader.Create()`, and Forms calls the `FallbackTypeResolver` asking us to resolve the type with name `BaseContentPage`. We can't find any information about that type, because it should be asking us to resolve the type with name ``BaseContentPage`1``. So the previewer.

But this change doesn't only fix the previewer... it simply fixes calls `XamlLoader.Create()` (I will remove the change in `XamlLoader.Load()` - it is harmless, but pointless). For XamlLoader.Create(), the existing logic is simply wrong. It will try to find a non-generic type, which will either fail, or find the wrong type if there happens to be one with the same name. With my change, it finds the correct type (if it exists) and correctly applies the generic type arguments. And, it makes the previewer work.

StephaneDelcroix:
thanks for your input, I didn't recalled we had support for that in XamlG. if we do, and it's enforced, I'll consider integrating this 👍

TimBarham:
Thanks @StephaneDelcroix 😄

programmation:
Am I right in thinking the `<object>` in the example code is the BindingContext for the page? If so, MvvmCross and other frameworks have `ViewModel<TInitializer>` and `ViewModel<TInitializer, TResponse>` generics that turn the page into `Page<ViewModel<TInitializer, TResponse>>` type and I've never been clear how to set that up properly. Will this kind of generic BindingContext also now work?

StephaneDelcroix:
please target this to 4.0.0

PR #5847, opened 2019-04-08
[Android] add another gc.collect for Android to ensure all is collected

PureWeen:
### Description of Change ###
Fix Issue5555 so it no longer fails on android

### Testing Procedure ###
Run issue5555 on android and watch it succeed
let all the automated tests finish

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5555.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5555.cs
@@ -59,8 +59,8 @@ namespace Xamarin.Forms.Controls.Issues
 						result.Text = "Click 'Push page' again";
 						return;
 					}
-					GC.Collect();
-					GC.WaitForPendingFinalizers();
+
+					GarbageCollectionHelper.Collect();
 					result.Text = list[list.Count - 2].IsAlive ? "Failed" : "Success";
 				})
 			};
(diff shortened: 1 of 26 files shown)


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5555.cs
```diff
@@ -59,6 +59,8 @@ protected override void Init ()
 						result.Text = "Click 'Push page' again";
 						return;
 					}
+					GC.Collect();
+					GC.WaitForPendingFinalizers();
```
adrianknight89: In general, we should call `GC.Collect();` after waiting for pending finalizers so that we can collect any dead objects which are created when calling the finalizers, so I
d add another `GC.Collect();` here..

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5555.cs
```diff
@@ -79,6 +81,7 @@ protected override void Init ()
 							GC.Collect();
 							GC.WaitForPendingFinalizers();
 							GC.Collect();
+							GC.WaitForPendingFinalizers();
```
adrianknight89: And I'm not sure about the point of this line. Why wait for the finalizers again? Never seen it this way before.

samhouts:
None of the tests you touched failed!! 🎉

PR #5850, opened 2019-04-08
Move IElementConfiguration to an internal structure on Cell

PureWeen:
### Description of Change ###
Another attempt to resolve the issue with UWP covariance performance. Another option we could go for would be to just convert the UI Test list in control gallery to a collectionview

Though right now it's a TableView

Basically I just move the PlatformConfiguration to an internal class on Cell to proxy the On calls to

### Testing Procedure ###
- make sure UI Tests work on UWP
- make sure you can change the default background color of a cell on iOS using the platform specific

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Cells/Cell.cs
+++ b/Xamarin.Forms.Core/Cells/Cell.cs
@@ -268,7 +268,7 @@ namespace Xamarin.Forms
 			return _elementConfiguration.Value;
 		}
 
-		public class ElementConfiguration : IElementConfiguration<Cell>
+		class ElementConfiguration : IElementConfiguration<Cell>
 		{
 			readonly Lazy<PlatformConfigurationRegistry<Cell>> _platformConfigurationRegistry;
 			public ElementConfiguration(Cell cell)
--- a/Xamarin.Forms.Core/Cells/Cell.cs
+++ b/Xamarin.Forms.Core/Cells/Cell.cs
@@ -8,22 +8,22 @@ using Xamarin.Forms.Internals;
 
 namespace Xamarin.Forms
 {
-	public abstract class Cell : Element, ICellController, IElementConfiguration<Cell>, IFlowDirectionController, IPropertyPropagationController, IVisualController
+	// Don't add IElementConfiguration<Cell> because it kills performance on UWP structures that use Cells
+	public abstract class Cell : Element, ICellController, IFlowDirectionController, IPropertyPropagationController, IVisualController
 	{
 		public const int DefaultCellHeight = 40;
 		public static readonly BindableProperty IsEnabledProperty = BindableProperty.Create("IsEnabled", typeof(bool), typeof(Cell), true, propertyChanged: OnIsEnabledPropertyChanged);
 
 		ObservableCollection<MenuItem> _contextActions;
+		readonly Lazy<ElementConfiguration> _elementConfiguration;
 
 		double _height = -1;
 
 		bool _nextCallToForceUpdateSizeQueued;
 
-		readonly Lazy<PlatformConfigurationRegistry<Cell>> _platformConfigurationRegistry;
-
 		public Cell()
 		{
-			_platformConfigurationRegistry = new Lazy<PlatformConfigurationRegistry<Cell>>(() => new PlatformConfigurationRegistry<Cell>(this));
+			_elementConfiguration = new Lazy<ElementConfiguration>(() => new ElementConfiguration(this));
 		}
 
 		EffectiveFlowDirection _effectiveFlowDirection = default(EffectiveFlowDirection);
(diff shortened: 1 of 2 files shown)


samhouts:
@PureWeen The Android 28 LR failures look new.

PureWeen:
I reran LR for API 28 and the failed UI tests this time around are different
I ran all the ones that failed in that first run locally and they were all green for me :-/ 
Let me know what you think of the new run

https://appcenter.ms/orgs/xtc-Xamarin-Forms/apps/AndroidControlGallery-1/test/series/pull-5850/runs/b79b2c73-7a48-4373-b4e3-7446b79d45af

Xamarin.Forms.Core/Cells/Cell.cs
```diff
@@ -257,5 +252,39 @@ void OnParentPropertyChanging(object sender, PropertyChangingEventArgs e)
 			if (e.PropertyName == "RowHeight")
 				OnPropertyChanging("RenderHeight");
 		}
+
+
+		#region internal element configuration
```
StephaneDelcroix: 🤦‍♂️ oh, no, a `#region`

Xamarin.Forms.Core/Cells/Cell.cs
```diff
@@ -8,22 +8,22 @@
 
 namespace Xamarin.Forms
 {
-	public abstract class Cell : Element, ICellController, IElementConfiguration<Cell>, IFlowDirectionController, IPropertyPropagationController, IVisualController
+	// Don't add IElementConfiguration<Cell> because it kills performance on UWP structures that use Cells
+	public abstract class Cell : Element, ICellController, IFlowDirectionController, IPropertyPropagationController, IVisualController
```
StephaneDelcroix: so, this is technically a breaking change ? can you investigate if we should care about that or not ?

Xamarin.Forms.Core/Cells/Cell.cs
```diff
@@ -8,22 +8,22 @@
 
 namespace Xamarin.Forms
 {
-	public abstract class Cell : Element, ICellController, IElementConfiguration<Cell>, IFlowDirectionController, IPropertyPropagationController, IVisualController
+	// Don't add IElementConfiguration<Cell> because it kills performance on UWP structures that use Cells
```
StephaneDelcroix: this is not a code comment, because as soon as the patch is applied it will be out of date and refer to something that's no longer there :)

but that's a good PR comment to add, for history purposes

Xamarin.Forms.Core/Cells/Cell.cs
```diff
@@ -257,5 +252,39 @@ void OnParentPropertyChanging(object sender, PropertyChangingEventArgs e)
 			if (e.PropertyName == "RowHeight")
 				OnPropertyChanging("RenderHeight");
 		}
+
+
+		#region internal element configuration
+		// This creates an internal class to keep track of IElementConfiguration because adding 
+		// IElementConfiguration<T> to the Cell itself kills performance on UWP
+		// Issue has been logged with UWP
+		public IPlatformElementConfiguration<T, Cell> On<T>() where T : IConfigPlatform
```
StephaneDelcroix: see, the region says `internal element configuration`, and this is 1/public, and 2/platform configuration. regions are out of date more quickly than code comment does

Xamarin.Forms.Core/Cells/Cell.cs
PureWeen: @StephaneDelcroix what are the scenarios where you would see this cause breaking? I just want to make sure I've considered them all.

The main case I know of would be if in user code someone casted Cell as IElementConfiguration<Cell> and is using that instance to apply the platform specifics. 

I tested if it would cause breaking if someone inherited from Cell
```C#
   // this is compiled against a cell with the interface
    public class ShaneCell : Cell
    {
    }
```

I then used the compiled library in a project that has the interface removed. 
I also tested visa versa. 

This fixes performance on UWP specifically with listview and since it seems like a minor 4.0 breaking change then I think we're ok

Xamarin.Forms.Core/Cells/Cell.cs
PureWeen: Yea I should use the word nested or inner not internal

Xamarin.Forms.Core/Cells/Cell.cs
PureWeen: @StephaneDelcroix  I don't quite follow. The main purpose of the comment is so that future generations won't add the interface to Cell and run into this same issue.  Would it make more sense if I left a commented out version of IElementConfiguration<cell> so that there is more context?

PureWeen:
rebased to 4.0.0 api 28 LR is passing

API 28 FR, ios10, ios12 tests aren't related

PR #5853, opened 2019-04-09
Create UI test template for XAML pages

adrianknight89:
It's time-consuming to create XAML test files from scratch. This template will help speed things up. Also updated existing template to add an empty view model & model.


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/_Template.cs
```diff
@@ -14,27 +14,47 @@ namespace Xamarin.Forms.Controls.Issues
 #endif
 	[Preserve(AllMembers = true)]
 	[Issue(IssueTracker.Github, 1, "Issue Description", PlatformAffected.Default)]
-	public class Issue1 : TestContentPage // or TestMasterDetailPage, etc ...
+	public class Github1 : TestContentPage // or TestMasterDetailPage, etc ...
```
hartez: Historically, we've used `Issue` instead of `GitHub` for issues tracked in GitHub. (GitHub was the original issue tracker, before things were moved to Bugzilla.) For consistency, I'm inclined to say we should stick with `Issue` in the template. 

That said, I'm not strongly against the more logical `GitHub` prefix, but if we're going to use it, we should case it correctly (`GitHub` instead of `Github`).

hartez: Okay, having talked with @samhouts on this, I think we should stick with `Issue` for GitHub issues. Otherwise, we'll end up with weird ordering in the IDEs for tests, which makes it inconvenient to find the ones you're looking for.

adrianknight89: I'll revert this back to `Issue` later today. Thanks.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Xamarin.Forms.Controls.Issues.Shared.projitems
```diff
@@ -1157,4 +1160,15 @@
...
   </ItemGroup>
-</Project>
+  <ItemGroup>
+    <EmbeddedResource Include="$(MSBuildThisFileDirectory)_TemplateMarkup.xaml">
+      <SubType>Designer</SubType>
+      <Generator>MSBuild:Compile</Generator>
+    </EmbeddedResource>
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Update="C:\g\Xamarin.Forms\Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\_TemplateMarkup.xaml.cs">
```
StephaneDelcroix: ain't gonna fly (absolute path)

StephaneDelcroix: also, why does this require an `Update` ?

adrianknight89: Not sure why this got added in the project file. Is the team still on VS2017? I recently switched to 2019.

adrianknight89: Seems like I can't set the custom tool to `MSBuild:UpdateDesignTimeXaml` and VS is adding an `Update` with the absolute path by default for any new XAML file added to the shared project.

PR #5866, opened 2019-04-09
[Maps] Execute Map.Pins CollectionChanged handling on the main thread

### LINKED ISSUE #5865: [Maps] Exception when adding a Pin off the main thread
### Description

If a `Pin` is added to a `Map` off the main thread, an exception is thrown. This is likely primarily going to affect users using the recently-added item templating features, when the bound `ItemsSource` is updated on a background thread.

### Steps to Reproduce

Add a Pin to a Map from off of the main thread (such as by doing it in a `Task.Run()`)

### Expected Behavior

The pin is added to the map.

### Actual Behavior

An exception is thrown.

### Basic Information

- Version with issue: 3.6.0.293080
- Last known good version: None
- IDE: Visual Studio 2017
- Platform Target Frameworks:
  - iOS:  12.2
  - Android: 9.0
  - UWP:  16299
- Nuget Packages: Xamarin.Forms.Maps 3.6.0.293080

### Stack Traces

Android:
```
Java.Lang.IllegalStateException: Not on the main thread
  at Java.Interop.JniEnvironment+InstanceMethods.CallObjectMethod (Java.Interop.JniObjectReference instance, Java.Interop.JniMethodInfo method, Java.Interop.JniArgumentValue* args) [0x00069] in <286213b9e14c442ba8d8d94cc9dbec8e>:0 
  at Android.Runtime.JNIEnv.CallObjectMethod (System.IntPtr jobject, System.IntPtr jmethod, Android.Runtime.JValue* parms) [0x0000e] in <95d915721ad748108ff096dce8f8f32c>:0 
  at Android.Gms.Maps.GoogleMap.AddMarker (Android.Gms.Maps.Model.MarkerOptions options) [0x00048] in <0ef98d0b7df944fb8f20d9e078937bf5>:0 
  at Xamarin.Forms.Maps.Android.MapRenderer+<>c__DisplayClass20_0.<AddPins>b__0 (Xamarin.Forms.Maps.Pin p) [0x0000f] in <979ddbf58b814973bfd6665371ba6422>:0 
  at System.Linq.Enumerable+SelectEnumerableIterator`2[TSource,TResult].MoveNext () [0x00048] in <715c2ff6913942e6aa8535593b3ef35a>:0 
  at System.Collections.Generic.List`1[T].AddEnumerable (System.Collections.Generic.IEnumerable`1[T] enumerable) [0x00059] in <d4a23bbd2f544c30a48c44dd622ce09f>:0 
  at System.Collections.Generic.List`1[T].InsertRange (System.Int32 index, System.Collections.Generic.IEnumerable`1[T] collection) [0x000f4] in <d4a23bbd2f544c30a48c44dd622ce09f>:0 
  at System.Collections.Generic.List`1[T].AddRange (System.Collections.Generic.IEnumerable`1[T] collection) [0x00000] in <d4a23bbd2f544c30a48c44dd622ce09f>:0 
  at Xamarin.Forms.Maps.Android.MapRenderer.AddPins (System.Collections.IList pins) [0x00052] in <979ddbf58b814973bfd6665371ba6422>:0 
  at Xamarin.Forms.Maps.Android.MapRenderer.OnCollectionChanged (System.Object sender, System.Collections.Specialized.NotifyCollectionChangedEventArgs notifyCollectionChangedEventArgs) [0x00029] in <979ddbf58b814973bfd6665371ba6422>:0 
  at (wrapper delegate-invoke) <Module>.invoke_void_object_NotifyCollectionChangedEventArgs(object,System.Collections.Specialized.NotifyCollectionChangedEventArgs)
  at System.Collections.ObjectModel.ObservableCollection`1[T].OnCollectionChanged (System.Collections.Specialized.NotifyCollectionChangedEventArgs e) [0x00018] in <fb6d78e506844b3b96d5b35aa047fbbd>:0 
  at System.Collections.ObjectModel.ObservableCollection`1[T].OnCollectionChanged (System.Collections.Specialized.NotifyCollectionChangedAction action, System.Object item, System.Int32 index) [0x00009] in <fb6d78e506844b3b96d5b35aa047fbbd>:0 
  at System.Collections.ObjectModel.ObservableCollection`1[T].InsertItem (System.Int32 index, T item) [0x0001a] in <fb6d78e506844b3b96d5b35aa047fbbd>:0 
  at System.Collections.ObjectModel.Collection`1[T].Add (T item) [0x00020] in <d4a23bbd2f544c30a48c44dd622ce09f>:0 
  at MapAsyncPins.MainPage.<AddPinClicked>b__1_0 () [0x00001] in C:\Users\joem\Documents\Visual Studio 2017\Projects\MapAsyncPins\MapAsyncPins\MapAsyncPins\MainPage.xaml.cs:24 
  at System.Threading.Tasks.Task.InnerInvoke () [0x0000f] in <d4a23bbd2f544c30a48c44dd622ce09f>:0 
  at System.Threading.Tasks.Task.Execute () [0x00000] in <d4a23bbd2f544c30a48c44dd622ce09f>:0 
--- End of stack trace from previous location where exception was thrown ---
  at MapAsyncPins.MainPage+<AddPinClicked>d__1.MoveNext () [0x00034] in C:\Users\joem\Documents\Visual Studio 2017\Projects\MapAsyncPins\MapAsyncPins\MapAsyncPins\MainPage.xaml.cs:22 
  --- End of managed Java.Lang.IllegalStateException stack trace ---
com.google.maps.api.android.lib6.common.apiexception.c: Not on the main thread
	at com.google.maps.api.android.lib6.common.j.b(:com.google.android.gms.dynamite_dynamitemodulesb@12521025@12.5.21 (040800-189987672):11)
	at com.google.maps.api.android.lib6.common.o.a(:com.google.android.gms.dynamite_dynamitemodulesb@12521025@12.5.21 (040800-189987672):5)
	at com.google.maps.api.android.lib6.impl.bc.a(:com.google.android.gms.dynamite_dynamitemodulesb@12521025@12.5.21 (040800-189987672):487)
	at com.google.android.gms.maps.internal.l.onTransact(:com.google.android.gms.dynamite_dynamitemodulesb@12521025@12.5.21 (040800-189987672):94)
	at android.os.Binder.transact(Binder.java:604)
	at com.google.android.gms.internal.zzeb.zza(Unknown Source:7)
	at com.google.android.gms.maps.internal.zzg.addMarker(Unknown Source:9)
	at com.google.android.gms.maps.GoogleMap.addMarker(Unknown Source:2)
```

iOS:
```
UIKit.UIKitThreadAccessException: UIKit Consistency error: you are calling a UIKit method that can only be invoked from the UI thread.
  at UIKit.UIApplication.EnsureUIThread () [0x00020] in /Library/Frameworks/Xamarin.iOS.framework/Versions/12.4.0.64/src/Xamarin.iOS/UIKit/UIApplication.cs:89 
  at MapKit.MKShape.set_Title (System.String value) [0x00000] in /Library/Frameworks/Xamarin.iOS.framework/Versions/12.4.0.64/src/Xamarin.iOS/MapKit/MKShape.g.cs:172 
  at Xamarin.Forms.Maps.iOS.MapRenderer.CreateAnnotation (Xamarin.Forms.Maps.Pin pin) [0x0000c] in <cfc9f794aff84397ac644065fbccf083>:0 
  at Xamarin.Forms.Maps.iOS.MapRenderer.AddPins (System.Collections.IList pins) [0x00027] in <cfc9f794aff84397ac644065fbccf083>:0 
  at Xamarin.Forms.Maps.iOS.MapRenderer.OnCollectionChanged (System.Object sender, System.Collections.Specialized.NotifyCollectionChangedEventArgs notifyCollectionChangedEventArgs) [0x00029] in <cfc9f794aff84397ac644065fbccf083>:0 
  at (wrapper delegate-invoke) <Module>.invoke_void_object_NotifyCollectionChangedEventArgs(object,System.Collections.Specialized.NotifyCollectionChangedEventArgs)
  at System.Collections.ObjectModel.ObservableCollection`1[T].OnCollectionChanged (System.Collections.Specialized.NotifyCollectionChangedEventArgs e) [0x00018] in <4748941ef4974ccdbea6f7bf1893b672>:0 
  at System.Collections.ObjectModel.ObservableCollection`1[T].OnCollectionChanged (System.Collections.Specialized.NotifyCollectionChangedAction action, System.Object item, System.Int32 index) [0x00009] in <4748941ef4974ccdbea6f7bf1893b672>:0 
  at System.Collections.ObjectModel.ObservableCollection`1[T].InsertItem (System.Int32 index, T item) [0x0001a] in <4748941ef4974ccdbea6f7bf1893b672>:0 
  at System.Collections.ObjectModel.Collection`1[T].Add (T item) [0x00020] in <debab4bc68ca4ca1b79d0ded11987ada>:0 
  at MapAsyncPins.MainPage.<AddPinClicked>b__1_0 () [0x00001] in C:\Users\joem\Documents\Visual Studio 2017\Projects\MapAsyncPins\MapAsyncPins\MapAsyncPins\MainPage.xaml.cs:24 
  at System.Threading.Tasks.Task.InnerInvoke () [0x0000f] in /Library/Frameworks/Xamarin.iOS.framework/Versions/12.4.0.64/src/Xamarin.iOS/external/corert/src/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs:2501 
  at System.Threading.Tasks.Task.Execute () [0x00000] in /Library/Frameworks/Xamarin.iOS.framework/Versions/12.4.0.64/src/Xamarin.iOS/external/corert/src/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs:2344 
--- End of stack trace from previous location where exception was thrown ---
  at MapAsyncPins.MainPage+<AddPinClicked>d__1.MoveNext () [0x00034] in C:\Users\joem\Documents\Visual Studio 2017\Projects\MapAsyncPins\MapAsyncPins\MapAsyncPins\MainPage.xaml.cs:22 
```

UWP:
```
System.Exception: The application called an interface that was marshalled for a different thread. (Exception from HRESULT: 0x8001010E (RPC_E_WRONG_THREAD))
   at Windows.UI.Xaml.Controls.Maps.MapControl.get_Children()
   at Xamarin.Forms.Maps.UWP.MapRenderer.LoadPin(Pin pin)
   at Xamarin.Forms.Maps.UWP.MapRenderer.OnCollectionChanged(Object sender, NotifyCollectionChangedEventArgs e)
   at System.Collections.Specialized.NotifyCollectionChangedEventHandler.Invoke(Object sender, NotifyCollectionChangedEventArgs e)
   at System.Collections.ObjectModel.ObservableCollection`1.OnCollectionChanged(NotifyCollectionChangedEventArgs e)
   at System.Collections.ObjectModel.ObservableCollection`1.InsertItem(Int32 index, T item)
   at System.Collections.ObjectModel.Collection`1.Add(T item)
   at MapAsyncPins.MainPage.<AddPinClicked>b__1_0()
   at System.Threading.Tasks.Task.InnerInvoke()
   at System.Threading.Tasks.Task.<>c.<.cctor>b__276_1(Object obj)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at MapAsyncPins.MainPage.<AddPinClicked>d__1.MoveNext()
```

### Reproduction Link

[MapAsyncPins.zip](https://github.com/xamarin/Xamarin.Forms/files/3060806/MapAsyncPins.zip)

jcmanke:
### Description of Change ###

Handle the CollectionChanged event from Map.Pins on the main thread in renderers.

### Issues Resolved ### 

- fixes #5865

### API Changes ###

 None

### Platforms Affected ### 

- iOS
- Android
- UWP

May experience a performance hit when adding/removing multiple pins from the map.

Not applicable

### Testing Procedure ###

1. Run the Control Gallery app.
2. Go to page "Map With ItemsSource Gallery - Legacy"
3. Test all of the buttons at the bottom of the page.

### PR Checklist ###

- [ ] Has automated tests


adammeaney:
Personally, I would advocate that if one platform or another needs events handled on the UI thread, that the UI layer should be handling it.

Why should my ViewModels be forced to know that this random list needs to happen on the UI thread when that is a View only concern?

andreinitescu:
I don't agree with these changes. As @kingces95 pointed out, it's the app layer which need to handle this if that's what's needed.

adammeaney:
@andreinitescu I guess I have not heard any reasons why that would be other than that's how a few things work currently.

Not everything, as Joe rightly points out that ListViews don't need that.

How would you properly even handle that in the application layer? Do you implement a way to call Device.BeginInvokeOnMainThread in your ViewModel for every time you need to edit your pin list?

That just feels like it should violate every MVVM principle I know, introducing code to the View Model that is only there because of your choice of View. It shouldn't matter if I show my VM property of List<DataObject> in a map or a list, I should trust that the framework will handle updating the UI on the appropriate thread.

What reason would I not want that?

andreinitescu:
I apologize if it sounded like I'm just jumping into the discussion for no reason. I only commented because it happens I've done some work on that part of the code.

The issue #5865 was wrongly tagged as a bug, it's not a bug. Those exceptions are normal, meaning, those are thrown by native UI framework because you are updating the UI (the map native control) on a different thread than the main thread. It's not a Xamarin.Forms bug.

Even if it's not a bug and you're asking for a change in the XF framework, I don't think adding asynchronicity makes sense because, like with other UI elements, updating elements/view is supposed to run on the main thread. XF is an UI abstraction framework but it makes to follow the native UI frameworks which requires using the main thread when updating the UI.
Also, if someone wants to add a pin, immediately after the call returns it's expected that the pin collection to contain that pin, which is no longer the case with this change.

If you want to keep creating the Pins on a separate thread, you can do that, but in the end, the native forces you to add the pins to the native map on the main thread.

adammeaney:
I guess we have a very different understanding of what we are looking at then.

Because the point of this PR is to make sure that adding pins happens on the main thread, to support the way the platforms handle the pins.

Its that View Model code rarely happens on the main thread, property changes can come from other places, and we need to make sure that even if I have a code block like 

```
public async Task GetData()
{
    var pinModels = await _DataService.RetrieveData().ConfigureAwait(false);
    PinList = pinModels.Select(pinModel => PinModel.ToXFPin());
}
```

and that executes from a background thread or anything, that we can still see the pins bound in the pin list.

jcmanke:
@andreinitescu I feel like you're misreading my changes, or we're otherwise talking past each other. Whether you directly add to `Map.Pins` or the list that `Map.ItemsSource` is bound to, `Map.Pins` will immediately contain the new `Pin` object. It is only the rendering of that `Pin` into the native `Marker`/`IMKAnnotation`/`PushPin` that I am marshalling onto the UI thread.

If the Xamarin team wants to relabel #5865 as an enhancement rather than a bug that is fine by me.

andreinitescu:
>  Map.Pins will immediately contain the new Pin object. It is only the rendering of that Pin into the native Marker/IMKAnnotation/PushPin that I am marshalling onto the UI thread.

Yes, you're correct, my bad. I see what you mean. But after adding to the pin collection returns, the native pin collection will not have that native pin added. The two collections will be out of sync at this point. I'm not sure what are the implications around that, if they are any. I tend to think more in the native terms and I feel like it would be simpler if native pins are added synchronously, but maybe I'm wrong.

hartez:
After thinking about this some more, I'm afraid I have to agree with @kingces and @andreinitescu - this is the responsibility of the application, not the framework.

@kingces nails it here:

> And why just for OnCollectionChanged? Why not extend that precedent to other areas that could be invoked off the UI-thread? Is this one special?

There's no particular reason for this one to be any different than, say, `Text` on an `Entry` or `Rotation` on any `VisualElement`. If we go this direction, the only logically consistent path would be to wrap every native property change in `BeginInvokeOnMainThread`. Since most code executes on the main thread, thread checking/marshaling unnecessarily would be a performance drag most of the time. (Doubly so when binding is involved, since the binding system already marshals everything to the main thread.)

> Why should my ViewModels be forced to know that this random list needs to happen on the UI thread when that is a View only concern?

Well, if you're using data binding, they don't need to know. And I would argue that your view models themselves _shouldn't_ have to know, but code which _mutates you view models from another thread_ should know, and should handle the marshaling. That is, your example could be modified to look something like this:

```
public async Task GetData()
{
    var pinModels = await _DataService.RetrieveData().ConfigureAwait(false);
    Device.BeginInvokeOnMainThread(() => {
		foreach (var pinModel in pinModels)
		{
			PinList.Add(pinModel);
		}
	});
}
```

Though your example does raise a question - why are you using `ConfigureAwait(false)` if the thing you want to do when your `async` method is finished requires updating the UI? 

> That just feels like it should violate every MVVM principle I know, introducing code to the View Model that is only there because of your choice of View.

You would have this issue with any native UI control you were trying to update from a non-UI thread. It's not specific to Maps. And again, it's not that your view models should handle this; it's whatever code is updating your view models from a non-UI thread that should deal with this issue.

>  ListViews don't need that

This is true - there's some code buried in ListProxy that marshals all of the change handling code to the main thread automagically. I don't know the original reasoning behind the decision to make ListView special in this way. But ListView is the exception here.

adammeaney:
I guess what this says to me is that I should never use collection changed events as you don't think we should support them reasonably.

ObservableCollections are useless then. Is that not what you are saying? If I ever want to do some processing of my returned objects and still allow the UI to be responsive, I have to sacrifice collection changed events, or litter my VM with Device.BeginInvokeOnMainThread, which is something that absolutely should not be in my VM.

If you are saying that my app should do

```
public async Task GetData()
{
    var pinModels = await _DataService.RetrieveData().ConfigureAwait(false);

    // Some heavy data processing here, that shouldn't be on the main thread so that I don't have stuttering UIs

    Device.BeginInvokeOnMainThread(() => {
		foreach (var pinModel in pinModels)
		{
			PinList.Add(pinModel);
		}
	});
}
```

I just can't get behind that. Especially when you would let me write it as 

```
public async Task GetData()
{
    var pinModels = await _DataService.RetrieveData().ConfigureAwait(false);

    // Some heavy data processing here, that shouldn't be on the main thread so that I don't have stuttering UIs

    PinList = PinList.Concat(pinModels);
}
```

and have a perfectly functional UI. Those use cases are not different. You should not be handling them differently. That's my take.

adammeaney:
@hartez Do you handle it the same way in CollectionView? Or even the BindableLayout stuff? At least all of the controls with ItemTemplates should be consistent.

The PropertyChanged event and the CollectionChanged event are the same basic things, they are used in the Forms code to indicate that either a property on a view or a list of views changed.

They are so much the same that Microsoft has ways to use CollectionChanged as their UpdateSourceTrigger for Bindings. If that doesn't indicate that even the company thinks they should be treated the same, I don't know what would.

hartez:
> @hartez Do you handle it the same way in CollectionView? Or even the BindableLayout stuff? At least all of the controls with ItemTemplates should be consistent.

Everything other than the ListView which exposes an ItemsSource property (e.g., BindableLayout, Map, CollectionView, Picker, etc.) will throw an exception if the ItemsSource is updated (i.e., has items added/removed) off the UI thread. 

So ListView is the inconsistent one at this point. More on that later.

And for the controls which allow direct manipulation of their collection properties (e.g. Picker.Items and Map.Pins), directly modifying those collections off of the UI thread results in an exception. So we are consistent there. In those cases, we are effectively just modifying a UI property of the control; it's the same as if we were modifying the `Placeholder` property of Entry, or the `Text` property of Button. We leave the marshaling decisions for that to the application author for various performance reasons.

> The PropertyChanged event and the CollectionChanged event are the same basic things, they are used in the Forms code to indicate that either a property on a view or a list of views changed.

Well, to be clear, there's nothing about either PropertyChanged or CollectionChanged (i.e., about INotifyPropertyChanged or INotifyCollectionChanged) that has anything to do with marshaling updates to the UI thread. If you just create the standard basic implementation of those interfaces on your viewmodel and handle those events directly to modify a UI property, you have to handle the UI thread marshaling yourself. 

(Some base viewmodel implementations in other frameworks handle that for you automatically (e.g., I think Caliburn.Micro does this)).

Now, if you're using binding in Xamarin.Forms, the _binding_ will automatically handle the thread marshaling for you; that might be what you're thinking of. This is a convenience that Forms provides; other binding systems (e.g., UWP) don't automatically marshal the application of the binding to the UI thread. If you have a UI property (say, for example, a Button's `Text`) bound to a property of a viewmodel, you can change the viewmodel's property from a non-UI thread and everything will work just fine - the marshaling is done for you.

Where this gets a bit confusing is when we bind a property like ItemsSource, and then modify the contents of ItemsSource on a non-UI thread. If the bound source happens to implement INCC, then this can result in UI changes, which need to happen on the UI thread.

Since it's just an update to a bound property, it's totally reasonable to expect that it works like other bound properties (e.g., a `Text` property bound to a string in a viewmodel) - the update is marshaled to the UI thread and everything works fine. And for ListView it does; for all the other controls with an ItemsSource, it does not. 

We can't achieve consistency by removing that behavior from ListView, since it could break existing applications. But it's possible we could add that behavior to the other controls which use ItemsSource. Doing so would incur a performance penalty, but it might be worth it for consistency and fewer POLA violations. I'm going to run this by the rest of the team to see if anyone can see a reason _not_ to do this that I'm missing.

adammeaney:
I agree I was basically conflating how you guys do binding and how we implement it, its just we implement it with how you do things in mind.

Glad to hear you agree that there just hasn't been a good reason presented not to support collection changed on the UI thread in these kinds of situations at least.

I agree, its just not much different than binding in these circumstances, its basically just how list binding should be.

samhouts:
I'm unclear where we've left this PR. Should it be closed?

jcmanke:
Probably. If the idea of marshalling all ItemSource updates to the UI thread were to go through it should have its own issue and PR and not be limited to Maps.

samhouts:
Okay....thank you for the effort here! We very much appreciate your contributions. :)

@hartez, would you mind putting together the proposal? Thanks!

hartez:
> Okay....thank you for the effort here! We very much appreciate your contributions. :)
> 
> @hartez, would you mind putting together the proposal? Thanks!

Added to my to-do list.

hartez:
Related: #8392

hartez:
Sorry this took so long, but the team has discussed it and we're going to support automatic marshaling of INCC changes to the main thread here.

rmarinho:
@jcmanke do you think you can rebase this ?

rmarinho:
Oh i merged master, tests running.

jcmanke:
@hartez I'm happy to see this getting approved, but before it gets merged, I think it might be a good idea to modify the code to check `Device.IsInvokeRequired` like in #8568. Do you agree?

hartez:
@jcmanke Agreed!

Xamarin.Forms.Maps.Android/MapRenderer.cs
```diff
@@ -396,27 +396,30 @@ void MoveToRegion(MapSpan span, bool animate)
 
 		void OnPinCollectionChanged(object sender, NotifyCollectionChangedEventArgs notifyCollectionChangedEventArgs)
 		{
-			switch (notifyCollectionChangedEventArgs.Action)
+			Device.BeginInvokeOnMainThread(() =>
```
hartez: Check Device.IsInvokeRequired before pushing this onto the main thread.

Xamarin.Forms.Maps.UWP/MapRenderer.cs
```diff
@@ -123,29 +123,33 @@ protected override void Dispose(bool disposing)
 
 		void OnPinCollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
 		{
-			switch (e.Action)
+			Device.BeginInvokeOnMainThread(() =>
```
hartez: Check Device.IsInvokeRequired before pushing this onto the main thread.

Xamarin.Forms.Maps.iOS/MapRenderer.cs
```diff
@@ -410,28 +410,31 @@ void MoveToRegion(MapSpan mapSpan, bool animated = true)
 
 		void OnPinCollectionChanged(object sender, NotifyCollectionChangedEventArgs notifyCollectionChangedEventArgs)
 		{
-			switch (notifyCollectionChangedEventArgs.Action)
+			Device.BeginInvokeOnMainThread(() =>
```
hartez: Check Device.IsInvokeRequired before pushing this onto the main thread.

jcmanke:
@hartez Added Device.IsInvokeRequired checks.

samhouts:
@jcmanke Would you mind rebasing to 4.6.0 to hopefully fix the build? Thanks!

jcmanke:
@samhouts Rebased.

PR #5902, opened 2019-04-12
Add explicit references back into nuspec

PureWeen:
### Description of Change ###
References have to specified explicitly otherwise the design package gets added into VS for Mac which causes exceptions

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/846719


PureWeen:
Tested nuget package on vs mac and it looks good

PR #5915, opened 2019-04-15
[Shell] Hide Shell types from being browsable

PureWeen:
### Description of Change ###
This hides Shell<things> and just shows the Tab aliases for intellisense.

### Platforms Affected ### 
- Core/XAML (all platforms)

Set ShellItem, ShellContent, and ShellSection  to EditorBrowsable(false) so that FlyoutItem, TabContent, and Tab are all that show up but won't cause compile exceptions. 

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Shell/ShellContent.cs
+++ b/Xamarin.Forms.Core/Shell/ShellContent.cs
@@ -13,6 +13,7 @@ using Xamarin.Forms.Internals;
 
 namespace Xamarin.Forms
 {
+	[EditorBrowsable(EditorBrowsableState.Always)]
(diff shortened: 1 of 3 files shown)


mgoertz-msft:
@StephaneDelcroix Yes, XLS does look at these attributes and will hide anything with EditorBrowsableState.Never from IntelliSense. Let me know if that doesn't work for you.

PureWeen:
@StephaneDelcroix @mgoertz-msft I tested this inside XAML and the classes marked *Never* didn't show up and the ones marked *always* did

PR #5926, opened 2019-04-16
[Android] Use Fast Renderers by Default

### LINKED ISSUE #5724: [Spec] Use Android Fast Renderers by Default
# Use Android Fast Renderers by Default

Android Fast Renderers were introduced in #845 in version 2.3.5. These removed an extra `ViewGroup` wrapper around the basic controls, which made view rendering faster by reducing the number of views inflated. 

These controls have a Fast Renderer that you must currently opt in to use via experimental flag:

- `Button`
- `Image`
- `Label`
- `Frame`

Since these new renderers were not identical to the legacy renderers, they were put under an experimental flag to prevent unexpected behavior changes in existing applications. 

These renderers have now been tested heavily for almost two years, and it is time to make them the default renderers on Android!

From this point forward, all applications targeting `FormsAppCompatActivity` will use the Fast Renderers by default. We will also no longer plan to add new features to the legacy renderers.

As part of this change, we will also be un-`seal`ing the Fast Renderers and ensuring that all extension points are properly open to allow developers to create custom renderers.

# Backward Compatibility

You can override the new default by:

1. Using custom renderers targeting the legacy renderers. Any existing custom renderers will continue to function with legacy renderers.
2. Specifying a different `View.Visual`, such as `Material`. Note that these are already different renderers that are also (mostly) following the Fast Renderer pattern.
3. Enabling the `Forms.SetFlags("UseLegacyRenderers")` flag in your `MainActivity`.

samhouts:
### Description of Change ###

Use Fast Renderers by default. Introduce `UseLegacyRenderers` flag to go back. Unseal fast renderers & add the most common extension points.

### Issues Resolved ### 

- fixes #5724

### API Changes ###

 None

### Platforms Affected ### 

- Android

None

Not applicable

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
@@ -41,12 +41,10 @@ namespace Xamarin.Forms.ControlGallery.Android
 			base.OnCreate(bundle);
 
 #if TEST_EXPERIMENTAL_RENDERERS
-			// CollectionView lets us test CollectionView stuff until it's officially released
-			Forms.SetFlags("FastRenderers_Experimental"/*, "CollectionView_Experimental", "Shell_Experimental"*/); 
-#else
 			// Fake_Flag is here so we can test for flag initialization issues
-			// CollectionView lets us test CollectionView stuff until it's officially released
-			Forms.SetFlags("Fake_Flag"/*, "CollectionView_Experimental", "Shell_Experimental" */); 
+			Forms.SetFlags("Fake_Flag"/*, "CollectionView_Experimental", "Shell_Experimental"*/); 
+#else
+			Forms.SetFlags("UseLegacyRenderers"/*, "CollectionView_Experimental", "Shell_Experimental" */);
 #endif
 			Forms.Init(this, bundle);
 
--- a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
@@ -101,19 +101,19 @@ namespace Xamarin.Forms.Platform.Android
 			RegisterHandler(typeof(Picker), typeof(AppCompat.PickerRenderer), typeof(PickerRenderer));
 			RegisterHandler(typeof(CarouselPage), typeof(AppCompat.CarouselPageRenderer), typeof(CarouselPageRenderer));
 
-            if (Forms.Flags.Contains(Flags.UseLegacyRenderers))
-            {
-                RegisterHandler(typeof(Button), typeof(AppCompat.ButtonRenderer), typeof(ButtonRenderer));
-                RegisterHandler(typeof(Frame), typeof(AppCompat.FrameRenderer), typeof(FrameRenderer));
-            }
-            else
-            {
-                RegisterHandler(typeof(Button), typeof(FastRenderers.ButtonRenderer), typeof(ButtonRenderer));
-                RegisterHandler(typeof(Label), typeof(FastRenderers.LabelRenderer), typeof(LabelRenderer));
-                RegisterHandler(typeof(Image), typeof(FastRenderers.ImageRenderer), typeof(ImageRenderer));
-                RegisterHandler(typeof(Frame), typeof(FastRenderers.FrameRenderer), typeof(FrameRenderer));
-            }
-        }
+			if (Forms.Flags.Contains(Flags.UseLegacyRenderers))
+			{
+				RegisterHandler(typeof(Button), typeof(AppCompat.ButtonRenderer), typeof(ButtonRenderer));
+				RegisterHandler(typeof(Frame), typeof(AppCompat.FrameRenderer), typeof(FrameRenderer));
+			}
+			else
+			{
+				RegisterHandler(typeof(Button), typeof(FastRenderers.ButtonRenderer), typeof(ButtonRenderer));
+				RegisterHandler(typeof(Label), typeof(FastRenderers.LabelRenderer), typeof(LabelRenderer));
+				RegisterHandler(typeof(Image), typeof(FastRenderers.ImageRenderer), typeof(ImageRenderer));
+				RegisterHandler(typeof(Frame), typeof(FastRenderers.FrameRenderer), typeof(FrameRenderer));
+			}
+		}
 
 		protected void LoadApplication(Application application)
 		{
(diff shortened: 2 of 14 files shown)


samhouts:
I'm running this through automated tests first, then we'll discuss how to best flip the `TEST_EXPERIMENTAL_RENDERERS` define.

Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
```diff
@@ -41,12 +41,10 @@ protected override void OnCreate(Bundle bundle)
 			base.OnCreate(bundle);
 
 #if TEST_EXPERIMENTAL_RENDERERS
-			// CollectionView lets us test CollectionView stuff until it's officially released
-			Forms.SetFlags("FastRenderers_Experimental"/*, "CollectionView_Experimental", "Shell_Experimental"*/); 
-#else
 			// Fake_Flag is here so we can test for flag initialization issues
-			// CollectionView lets us test CollectionView stuff until it's officially released
-			Forms.SetFlags("Fake_Flag"/*, "CollectionView_Experimental", "Shell_Experimental" */); 
+			Forms.SetFlags("Fake_Flag"/*, "CollectionView_Experimental", "Shell_Experimental"*/);
```
kingces95: Why did `Fake_Flag` get moved?

Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
```diff
@@ -41,12 +41,10 @@ protected override void OnCreate(Bundle bundle)
...
 #if TEST_EXPERIMENTAL_RENDERERS
-			// CollectionView lets us test CollectionView stuff until it's officially released
-			Forms.SetFlags("FastRenderers_Experimental"/*, "CollectionView_Experimental", "Shell_Experimental"*/); 
-#else
 			// Fake_Flag is here so we can test for flag initialization issues
-			// CollectionView lets us test CollectionView stuff until it's officially released
-			Forms.SetFlags("Fake_Flag"/*, "CollectionView_Experimental", "Shell_Experimental" */); 
+			Forms.SetFlags("Fake_Flag"/*, "CollectionView_Experimental", "Shell_Experimental"*/); 
+#else
+			Forms.SetFlags("UseLegacyRenderers"/*, "CollectionView_Experimental", "Shell_Experimental" */);
```
kingces95: `UseLegacyRenderers` -> `NonFastRenderers_Legacy`? Cuz not all renderers have a legacy, right? Just those with fast versions right? And, I hate to ponder, but if we introduce new fast renderers for those yet to be converted then we might consider enumerating the renderers explicitly..

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -101,18 +101,18 @@ static void RegisterHandlers()
...
+			if (Forms.Flags.Contains(Flags.UseLegacyRenderers))
+			{
+				RegisterHandler(typeof(Button), typeof(AppCompat.ButtonRenderer), typeof(ButtonRenderer));
+				RegisterHandler(typeof(Frame), typeof(AppCompat.FrameRenderer), typeof(FrameRenderer));
+			}
+			else
 			{
 				RegisterHandler(typeof(Button), typeof(FastRenderers.ButtonRenderer), typeof(ButtonRenderer));
 				RegisterHandler(typeof(Label), typeof(FastRenderers.LabelRenderer), typeof(LabelRenderer));
 				RegisterHandler(typeof(Image), typeof(FastRenderers.ImageRenderer), typeof(ImageRenderer));
```
kingces95: Huh. Wonder why this isn't symmetric.. Why don't we have to register `AppCompat.LabelRenderer` and `AppCompat`ImageRenderer` like we did for button and frame.

Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
samhouts: The idea is that we'll always set a flag of some sort. Since I removed the flag from TEST_EXPERIMENTAL_RENDERERS and added a flag when we're not on fast renderers, I just moved the fake flag.

Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
samhouts: The idea is that we'll always set a flag of some sort. Since I removed the flag from TEST_EXPERIMENTAL_RENDERERS and added a flag when we're not on fast renderers, I just moved the fake flag.

Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
samhouts: If it's needed, we can add more fine-grained control. There are still workarounds for individual controls. For example, a developer could use an `ExportRenderer` and explicitly replace all Buttons with the old `ButtonRenderer` instead of using this flag. And yes, we may or may not be putting any new fast renderers under a new experimental flag. That's TBD. Not all renderers have a legacy at this point, but I like to think that this flag will happily encapulate them all when they do.

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
samhouts: Yeah, crazy, right? That's because there is no `AppCompat.LabelRenderer` and `AppCompat.ImageRenderer`. Those are still using the legacy legacy renderers (Pre-AppCompat), and those are registered in the forwarders via attribute.

This isn't confusing at all. 🤦‍♀️

Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
kingces95: Oh, I see.

Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
kingces95: Roger.

daniel-luberda:
```C#
internal sealed class VisualElementRenderer {}
internal static class ElevationHelper {}
```

@samhouts  Is it possible you could make these public? Currently FFImageLoading's renderer is forced to use reflection because of it.

samhouts:
@daniel-luberda Sure! Would you mind opening an issue describing what API in particular you are using reflection to get to? Thanks!

PR #5936, opened 2019-04-17
Adds UWP support to Shell

dotMorten:
### Description of Change ###

This is the first stab at adding UWP renderers for Shell

### Issues Resolved ### 
#2415

### API Changes ###
ShellRenderer (I tried to keep as much as possible internal for now).
```cs
public class ShellRenderer : NavigationView, IVisualElementRenderer, IAppearanceObserver, IFlyoutBehaviorObserver
{
   public ShellRenderer();
   protected internal Shell Element { get; private set; }
   protected virtual void OnElementPropertyChanged(object sender, PropertyChangedEventArgs e);
   protected virtual void OnElementSet(Shell shell);   
}
```

Known issues while it's a draft-PR:

 - [ ] 1. Back navigation isn't working.
 - [ ] 2. Only FileImageSource are supported for icons (limitation of NavigationViewItem and AppBarItem - considering to use a different kind)
 - [ ] 3. Try and drop back down to 16299 (requires 1809)
-  [ ] 4. Release builds fail to render the side pane (this one got me stumped, as no errors are reported - it's just completely missing from the visual tree)

It might be worth considering adding the WinUI dependency. This would mean 16299 can be continued to be used, but it does require one extra setup step to add a theme resource to App.xaml, as well as adding the dependency. In the future though, this library is where all new UWP components will be added, so sooner or later that's going to happen anyway.

### Platforms Affected ### 
- UWP

None

Before: 
image

After (Gastropods, Default Shell Template, Xappy):

Shell_3apps_gif

### Testing Procedure ###
Use @davidortinau's Gastropods sample. It seems fully functional from what I can tell.
You can also use this repo: https://github.com/dotMorten/Xamarin.Forms.UWPShell.Sample
(the sub-module points to this PR)

### PR Checklist ###

- [ ] Has automated tests 


Xamarin.Forms.Platform.Android/Properties/AssemblyInfo.cs
```diff
@@ -23,7 +23,7 @@
 
 // These renderers are now registered via the RenderWithAttribute in the Android Forwarders project.
 // Note that AppCompat and FastRenderers are also registered conditionally in FormsAppCompatActivity.LoadApplication
-#if ROOT_RENDERERS
+#if ROOT_RENDERERS || TRUE
```
bbenetskyi: This all-time be `true`. If it's done specifically, we need to remove that `#if` at all?

bbenetskyy: This all-time be `true`. If it's done specifically, we need to remove that `#if` at all?

dotMorten:
Update: I'd like to move to take a dependency on Microsoft.UI.Xaml. This allows Xamarin.Forms to continue compiling against 16299. Sooner or later I believe this will happen anyway, as this is where all new UWP controls are going to go, rather than in the OS. It also allows us to quicker address any issues with the underlying controls.

There's a couple of challenges though.
1. First of all there's a bug I encountered when moving to it. This is getting looked at already though:, and the team has been pretty responsieve  https://github.com/Microsoft/microsoft-ui-xaml/issues/595
2.  To use the library, the UWP app-head must reference the package, as they rely on .targets files to be executed, and those don't execute in dependencies. Nuget 5.0 has a way around through transitiveBuilds, but this won't work on VS2017. There's a [discussion here](https://github.com/NuGet/Home/issues/7961) regarding porting the feature back to VS2017/Nuget 4.x

One of the problems is we must reference a Microsoft.UI.Xaml resource in the app.xaml resources. This can however also be done in Form's own Resources.xaml file, so the end-developer won't have to do it, but that means the moment the resources.xaml is parsed and the WinUI package wasn't deployed right, the app would crash. I'm currently looking into whether we can load it on-demand instead, so it would only be a requirement if you want to use Shell, and thus avoid breaking existing apps.

PureWeen:
This is looking really nice

I merged with master to test the store gallery and the store gallery looks really smooth

I'm proposing this PR so that we can at least get the min requirements for WINUI setup
https://github.com/xamarin/Xamarin.Forms/pull/5981

>  and the WinUI package wasn't deployed right, the app would crash

What scenario would lead to this happening? 

If you rebase to latest master you can load in the Store Gallery through the Control Gallery

- run control gallery
- click Store shell from the top menu

one thing I noticed on android vs uwp is that the flyout header view doesn't full width on the flyout
image


But it looks awesome so far

dotMorten:
@PureWeen Thanks for the feedback. If we move to WinUI there's no need to increase from 16299. Moving that number up has the risks of inadvertently using APIs not available on 16299, so it's safer to keep max version the same as minversion.

That's why I think the WinUI approach is better, but I am finding some challenges. I've found a way to only require adding WinUI to the project head if you use Shell, so it could just be a prerequisite. Also it would only be a prereq if you use VS2017 once they start using support for transitive builds.

WRT the header not rendering the right width - I've had a hell of a time with rendering Forms UI inside the native UI. I might need a few pointers in getting it right.

PureWeen:
> Thanks for the feedback. If we move to WinUI there's no need to increase from 16299. Moving that number up has the risks of inadvertently using APIs not available on 16299, so it's safer to keep max version the same as minversion.

I haven't used Win UI yet but I was just going off of the version info from their git page

https://github.com/Microsoft/microsoft-ui-xaml#version-support

> The Microsoft.UI.Xaml 2.1 NuGet package requires your project to have TargetPlatformVersion >= 10.0.17763.0 and TargetPlatformMinVersion >= 10.0.14393.0 when building.

So I was just upping our platform projects to match those numbers and what you did here https://github.com/xamarin/Xamarin.Forms/pull/5936/files#diff-618e2bbd3673299c5f379b46b9f032afR14

> WRT the header not rendering the right width - I've had a hell of a time with rendering Forms UI inside the native UI. I might need a few pointers in getting it right.

Yea it can be quirky :-/ 
I had a heck of a time getting TitleView to render on UWP correctly or at all with a MDP

I can poke at it next week a bit but overall not super worried about the quirks. Once we have the baseline for uwp shell and winui pieces in place we can merge and then create issues for known problem areas which will let us divide and conquer a bit more as well

dotMorten:
You just have to be in that range, which 16299 is, so we're good. I have another branch (`uwp-shell-winui`) that uses the WinUI controls instead of the native one.

PureWeen:
Aw I see you added this one
> <SkipMicrosoftUIXamlCheckTargetPlatformVersion>true</SkipMicrosoftUIXamlCheckTargetPlatformVersion>

to get around the targets error

Which you've already talked about a bit over here
https://github.com/Microsoft/microsoft-ui-xaml/issues/579

It does still mean that when running WinUI components on a pre 17763 version of windows then users might hit quirks or bugs right? 

https://github.com/Microsoft/microsoft-ui-xaml/issues/16#issuecomment-436240995

dotMorten:
> It does still mean that when running WinUI components on a pre 17763 version of windows then users might hit quirks or bugs right?

Oh good point! Didn't think of that.

dotMorten:
@PureWeen Also any comments on the general approach I took to implement this? Is it in line with how this should all be done?

PureWeen:
> Oh good point! Didn't think of that.

If we want to we can multitarget UWP instead of just setting the min/max
Then just include WinUI and 17763 apis inside that target

That way we just ensure maximum compatibility. It's always worrisome telling people it may work in pre 17763 APIs because if it's there they will expect it to regardless.

So perhaps we can multi target the UWP platform to ensure that we don't start using 17763  APIs and then we only include the shell renderers with the 17763 target. 

Other down sides?

PureWeen:
I do kind of waffle about whether multi targeting is fully worth it. If Shell is based on Win UI then in theory it'll work with 16299 but the experience just might be reduced slightly 

The main area of concern I have with opening up our Target Platforms is if we accidentally use an api that's not available in 16299 so if we just have a build check for this I'm not sure if we really need to multi target. 

The point of WinUI is so you don't have to multi target right? In theory if we rewrote all of forms using WinUI then we could even lower our min target.

dotMorten:
@PureWeen I had thought about multi-target too as the platform does fully support it. But just to be clear, you'll need 1809 to use Shell then. The point of WinUI is that you'll only need 14393+, and controls gets more frequent updates/bugfixes.

>  In theory if we rewrote all of forms using WinUI then we could even lower our min target.

Yes. It would probably be a breaking change though, as you'd suddenly start returning a different type. That might be ok for a major 4.0 release, so if you want to take it that far, now is the time.

I wouldn't worry about going lower than 16299 - it's very soon the lowest supported version anyway.

The alternative is to use the platform analyzer to detect if you use a non-supported API without a platform check.

dotMorten:
I'm now sort of leaning towards my current approach + paltform specific analyzer.
The downside is it'll require 1809. but I think that's still a lot better than, and more of a limited time issue.

The question then though is: What is the behavior if you try to use it below 1809: Render blank? Exception thrown (and then what and where)? We include an analyzer to tell you not to (which might be tricky when using .NET Standard).

dotMorten:
ARGH Crap! I messed up the rebase!

dotMorten:
@PureWeen The reason the image is missing in the header is that it was never added to the UWP project head like it is with Android and iOS:
image

dotMorten:
...same goes for all the icons. Those were missing in the Univeral gallery project as well:
image

PureWeen:
> I had thought about multi-target too as the platform does fully support it. But just to be clear, you'll need 1809 to use Shell then. The point of WinUI is that you'll only need 14393+, and controls gets more frequent updates/bugfixes.

Yea multi targeting sort of defeats the purpose of WINUI :-/   The only reason I could see doing multi targeting is that we'd essentially just build everything the same for 1809 and 16299 but it'd provide us a mechanism with our own APIs to create branching logic.  So if we wanted to use an API that's in 1809 in our LabelRenderer we'd get a compile time error for the 16299 target and then be able to just if/def the versions. So it'd be less about Shell and more about everything else. 

> I'm now sort of leaning towards my current approach + platform specific analyzer.
There's just a lot of ugly things about using WinUI from a class library, that we should work with the nuget and WinUI team on addressing, but that's more of a long-term thing, and if those issue could be resolved, WinUI would be the way to go.

So are you talking about just abandoning WINUI for now?

Also UWP looks awesome right now :-) Those screen shots of the store are so cool!!!!

dotMorten:
> So are you talking about just abandoning WINUI for now?

Well to have the least impact on current users upgrading, I think that's the best way to go. We can make sure things continue to "just work", and if they choose to use Shell on UWP, they'll just get blank (or an exception telling them to target 1809 as minversion). We'd have to then target 17132 as you did in your other PR, and I think that's ok if you add the platform analyzer, because it avoids calling into unsupported APIs without doing an API check first.

One downside of not using WinUI now is that the ShellRenderer inherits from NavigationView. If we were to move to WinUI later, it would be a breaking change (the alternative would be to keep that bit internal, which should be doable)

dotMorten:
> Also UWP looks awesome right now :-) Those screen shots of the store are so cool!!!!
Thanks. About half the buttons work (however I created a new branch because I messed this one up if you want to try it, and you need the PR that adds the missing image resources).

I'm especially struggling with anything natigation-related (generally crashes). I'm not sure what I'm missing, so could really use some pointers.

jevansaks:
> > So are you talking about just abandoning WINUI for now?
> 
> Well to have the least impact on current users upgrading, I think that's the best way to go.

Could you be more explicit about what is blocking you? Is it just the buildTransitive change that you need?

dotMorten:
> Could you be more explicit about what is blocking you? Is it just the buildTransitive change that you need?

Well even with that added, VS2017 wouldn't work. And it would have an effect on users who doesn't even use Shell but just upgrade. So I'm more concerned about breaking existing apps

PureWeen:
> And it would have an effect on users who doesn't even use Shell but just upgrade.

What effect? Why would adding WINUI effect say a TextBox or a MasterDetailsPage? 

> Could you be more explicit about what is blocking you? Is it just the buildTransitive change that you need?

can we just add a build target to get around this for now on VS 2017?

dotMorten:
> What effect? Why would adding WINUI effect say a TextBox or a MasterDetailsPage?

WinUI requires you to register an app extension. The moment you load the renderers, it'll find the extension isn't' there (since the renderer relies on WinUI) and go boom. There are more requirements (like registering a resource), but I managed to get around that programmatically by loading it dynamically when needed.

> can we just add a build target to get around this for now on VS 2017?

I don't think so. It's not a simple targets file, and involves registering a bunch of resources, without knowing where to get those. We could also quickly get out of sync.

dotMorten:
I think what I need to do is avoid exposing that I'm using NavigationView. That way we can switch to WinUI later if we want without introducing breaking changes. At this point we really only get downlevel support out of it anyway, but at the cost of a lot of complexity, and backwards compatibility issues.

PureWeen:
> I think what I need to do is avoid exposing that I'm using NavigationView. That way we can switch to WinUI later if we want without introducing breaking changes. At this point we really only get downlevel support out of it anyway, but at the cost of a lot of complexity, and backwards compatibility issues.

So currently there's nothing about WINUI that makes your life easier? If that's a true statement I would say 

- just abandon it for now and we can work with the WinUI team to get our code working with Win UI as a separate effort
- Taking a dependency on 1809 and we will just use a combination of those analyzers and multi targeting (again not something I think you need to worry about)

We will just throw a runtime exception if users try to use shell on < 1809 and then at some point we will get Win ui integration working which will make it work on previous versions

dotMorten:
@PureWeen Sounds good. I'll rewrite a bit so I don't expose the NavigationView type on the renderer, so we can flip the switch later.

Will you be at build? Perhaps we can meet with the WinUI team and just work through some of these issues?

jevansaks:
> Aw I see you added this one
> 
> > SkipMicrosoftUIXamlCheckTargetPlatformVersion>true</SkipMicrosoftUIXamlCheckTargetPlatformVersion
> 
> to get around the targets error

Sorry, I missed this earlier. Can you explain what error you were getting? Was it failing the TargetPlatformVersion or the TargetPlatformMinVersion check?

PureWeen:
@jevansaks 

It was failing the TargetPlatformVersion  check. I was confused why @dotMorten 's wasn't failing and mine was then I saw this attribute

With our currently plan we won't need that setting as we'll just put the TargetPlatformVersion  at 1809 and min at 16299 

The issues @dotMorten outlines here I still need to get caught up on
https://github.com/xamarin/Xamarin.Forms/pull/5936#issuecomment-486398129

Our plan is to implement Win UI outside of @dotMorten 's PR so he can continue working on shell and then at a later point once we get Win UI integrated then we can switch shell to use Win UI

jevansaks:
> Our plan is to implement Win UI outside of @dotMorten 's PR so he can continue working on shell and then at a later point once we get Win UI integrated then we can switch shell to use Win UI

That's great! We would be excited to have Xamarin.Forms using WinUI so please let me know if there's anything you run into that's blocking you or that you need changed to make it a smooth switch.

dotMorten:
Also a reminder that this issue is a blocker for using WinUI as well: https://github.com/Microsoft/microsoft-ui-xaml/issues/595

jevansaks:
Thanks for the reminder! We’ll put that on the list too.

dotMorten:
Closing in favor of #6015

dotMorten:
> We would be excited to have Xamarin.Forms using WinUI so please let me know if there's anything you run into that's blocking you or that you need changed to make it a smooth switch.

@jevansaks I think in general WinUI has challenges when it comes any 3rd party control class library building on top of WinUI. Users of that 3rd party lib doens't just add a reference to the library - they have to also know to add a reference to WinUI in the project head and add the mux resources. It creates quite a poor developer story. Transitive builds solves some of it, but only really viable once VS2017/Nuget4.x is no longer supported, and that would be quite a while.

jevansaks:
> Users of that 3rd party lib doesn't just add a reference to the library - they have to also know to add a reference to WinUI

I admit I'm somewhat new to the nuget packaging world, but isn't this generally solved by [nuget dependencies](https://docs.microsoft.com/en-us/nuget/reference/nuspec#dependencies-element)? The Xamarin.Forms package would put WinUI as a dependency and then when apps install Xamarin.Forms they also get the dependency of WinUI.

It's very possible I could be missing something though.

dotMorten:
> isn't this generally solved by nuget dependencies? 
Only assembly references. Assets and build folders weren't.
NuGet 5.0 adds the transitive build support so you can solve it, but it's opt in, and VS2019 only.

PR #5954, opened 2019-04-18
fix infinite loops and remove public apis

PureWeen:
### Description of Change ###

- removed GoToPart from the IShellController code parts because there isn't any platform code that uses them so it's not useful at this point
- added an exception if someone tries to do a relative route to a shell section as this behavior hasn't been implemented yet so I don't want the behavior to change once we make it so relative routes push to a stack. I also added an internal overload to turn this on for unit tests so unit tests can continue unhindered with validating behavior
- I fixed  the logic when you use an absolute route to navigation via <routename> / <page>. It was causing an  infinite loop
- removed getroutekeys and a few additional classes added in pre9 that will become exposed later down the road

### API Changes ###
 Removed:
 - Routing.GetRouteKeys
- class NavigationRequest
- GotoPart from the IShellController classes as the platforms don't use these methods so they don't need to be apart of the controllers

### Platforms Affected ### 
- Core/XAML (all platforms)

### Testing Procedure ###
- there are unit tests

### PR Checklist ###

- [X] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Routing.cs
+++ b/Xamarin.Forms.Core/Routing.cs
@@ -87,8 +87,10 @@ namespace Xamarin.Forms
 
 		internal static Uri RemoveImplicit(Uri uri)
 		{
+			uri = ShellUriHandler.FormatUri(uri);
+
 			if (!uri.IsAbsoluteUri)
-				return ShellUriHandler.FormatUri(uri);
+				return uri;
 
 			string[] parts = uri.OriginalString.TrimEnd(_pathSeparator[0]).Split(_pathSeparator[0]);
 
--- a/Xamarin.Forms.Core/Routing.cs
+++ b/Xamarin.Forms.Core/Routing.cs
@@ -84,6 +84,21 @@ namespace Xamarin.Forms
 			return $"{source}/";
 		}
 
+		internal static Uri RemoveImplicit(Uri uri)
+		{
+			if (!uri.IsAbsoluteUri)
+				return uri;
+
+			string[] parts = uri.ToString().TrimEnd('/').Split('/');
+
+			List<string> toKeep = new List<string>();
+			for (int i = 0; i < parts.Length; i++)
+				if (!IsImplicit(parts[i]))
+					toKeep.Add(parts[i]);
+
+			return new Uri(string.Join("/", toKeep));
+		}
+
 		public static string FormatRoute(List<string> segments)
 		{
 			var route = FormatRoute(String.Join("/", segments));
--- a/Xamarin.Forms.Core/Routing.cs
+++ b/Xamarin.Forms.Core/Routing.cs
@@ -42,7 +42,7 @@ namespace Xamarin.Forms
 			return bindable.GetType().Name + ++s_routeCount;
 		}
 
-		public static string[] GetRouteKeys()
+		internal static string[] GetRouteKeys()
 		{
 			string[] keys = new string[s_routes.Count];
 			s_routes.Keys.CopyTo(keys, 0);
--- a/Xamarin.Forms.Core/Routing.cs
+++ b/Xamarin.Forms.Core/Routing.cs
@@ -10,6 +10,7 @@ namespace Xamarin.Forms
 		static Dictionary<string, RouteFactory> s_routes = new Dictionary<string, RouteFactory>();
 
 		internal const string ImplicitPrefix = "IMPL_";
+		const string _pathSeparator = "/";
 
 		internal static string GenerateImplicitRoute(string source)
 		{
@@ -87,21 +88,21 @@ namespace Xamarin.Forms
 		internal static Uri RemoveImplicit(Uri uri)
 		{
 			if (!uri.IsAbsoluteUri)
-				return uri;
+				return ShellUriHandler.FormatUri(uri);
 
-			string[] parts = uri.ToString().TrimEnd('/').Split('/');
+			string[] parts = uri.OriginalString.TrimEnd(_pathSeparator[0]).Split(_pathSeparator[0]);
 
 			List<string> toKeep = new List<string>();
 			for (int i = 0; i < parts.Length; i++)
 				if (!IsImplicit(parts[i]))
 					toKeep.Add(parts[i]);
 
-			return new Uri(string.Join("/", toKeep));
+			return new Uri(string.Join(_pathSeparator, toKeep));
 		}
 
 		public static string FormatRoute(List<string> segments)
 		{
-			var route = FormatRoute(String.Join("/", segments));
+			var route = FormatRoute(String.Join(_pathSeparator, segments));
 			return route;
 		}
 
@@ -143,7 +144,7 @@ namespace Xamarin.Forms
 		static void ValidateRoute(string route)
 		{
 			if (string.IsNullOrWhiteSpace(route))
-				throw new ArgumentNullException("Route cannot be an empty string");
+				throw new ArgumentNullException(nameof(route), "Route cannot be an empty string");
 
 			var uri = new Uri(route, UriKind.RelativeOrAbsolute);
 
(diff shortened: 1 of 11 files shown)


PureWeen:
[5954.zip](https://github.com/xamarin/Xamarin.Forms/files/3110102/5954.zip)

Xamarin.Forms.Core/Routing.cs
```diff
@@ -84,6 +84,21 @@ internal static string GetRoutePathIfNotImplicit(Element obj)
 			return $"{source}/";
 		}
 
+		internal static Uri RemoveImplicit(Uri uri)
+		{
+			if (!uri.IsAbsoluteUri)
+				return uri;
+
+			string[] parts = uri.ToString().TrimEnd('/').Split('/');
```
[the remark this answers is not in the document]
PureWeen: Original string is a good suggestion but I don't want to remove the empty entries here. Otherwise if you have "app://route" it will turn into "app:route"

I don't think we need to have the backslashes do we? If it's an absolute uri at this point it'll be forward slashes right?

Xamarin.Forms.Core/Shell/ShellUriHandler.cs
```diff
@@ -19,11 +19,30 @@ static Uri FormatUri(Uri path)
...
 		}
 
+		internal static Uri CreateUri(string path)
+		{
+			path = FormatUri(path);
+
+			// on iOS if the uri starts with // it'll instantiate as absolute with
+			// file: as the default scheme where as android just crashes
+			// so this checks if it starts with / and just forces relative
+			if (path.StartsWith("/", StringComparison.Ordinal))
```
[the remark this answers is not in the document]
PureWeen: Above this code I do this
```
path = FormatUri(path);
```

which formats it to have forward slashes. I got tired of having to check for both everywhere so I try to always just sanitize the inputs so it converts them to forward slashes and then all the code further down can just worry about that

Xamarin.Forms.Core/Shell/ShellUriHandler.cs
```diff
@@ -121,7 +139,8 @@ internal static List<RouteRequestBuilder> GenerateRoutePaths(Shell shell, Uri re
 			string localPath = request.LocalPath;
 
 			bool relativeMatch = false;
-			if (!originalRequest.IsAbsoluteUri && !originalRequest.OriginalString.StartsWith("/") && !originalRequest.OriginalString.StartsWith("\\"))
+			if (!originalRequest.IsAbsoluteUri &&
+				!originalRequest.OriginalString.StartsWith("/", StringComparison.Ordinal))
```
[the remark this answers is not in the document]
PureWeen: Same respond as above about converting backs to forwards

Xamarin.Forms.Core/Routing.cs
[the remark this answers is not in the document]
paymicro: Absolute path can contain a slash and backslash same time and it will be correct. Users can enter as they like. We have a 100% certainty that the input Uri will only use slashes?

Xamarin.Forms.Core/Shell/ShellUriHandler.cs
```diff
@@ -9,19 +9,36 @@ namespace Xamarin.Forms
...
 				return path;
 
 			return new Uri(FormatUri(path.OriginalString), UriKind.Relative);
 		}
 
-		static string FormatUri(string path)
+		internal static string FormatUri(string path)
 		{
-			return path.Replace("\\", "/");
+			return path.Replace("\\", _pathSeparator);
```
paymicro: here you can use constants)
```suggestion
			return path.Replace(_pathSeparators[1], _pathSeparator[0]);
```

Xamarin.Forms.Core/Shell/ShellUriHandler.cs
```diff
@@ -485,10 +530,10 @@ static IEnumerable GetItems(object node)
 			for (var i = 0; i < keys.Length; i++)
 			{
 				var key = FormatUri(keys[i]);
-				if (key.StartsWith("/") && !(node is Shell))
+				if (key.StartsWith("/", StringComparison.Ordinal) && !(node is Shell))
```
paymicro:
```suggestion
				if (key.StartsWith(_pathSeparator, StringComparison.Ordinal) && !(node is Shell))
```

Xamarin.Forms.Core/Routing.cs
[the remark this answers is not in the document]
PureWeen: alright I added code to sanitize absolute uris

https://github.com/xamarin/Xamarin.Forms/pull/5954/commits/1da11a1fc406c436e6e1073bea2867cf82afab51#diff-e3b95c85373e6c0fb300485802be7ab6R18

I just really don't want to have "/", "\\" everywhere inside internal code :-)

Xamarin.Forms.Core/Shell/ShellNavigationState.cs
```diff
@@ -7,11 +7,32 @@ namespace Xamarin.Forms
 	[DebuggerDisplay("Location = {Location}")]
 	public class ShellNavigationState
 	{
-		public Uri Location { get; set; }
+		private Uri _fullLocation;
```
samhouts:
```suggestion
		Uri _fullLocation;
```

Xamarin.Forms.Core/Shell/ShellUriHandler.cs
```diff
@@ -100,43 +116,49 @@ public static NavigationRequest GetNavigationRequest(Shell shell, Uri uri)
...
-		internal static List<RouteRequestBuilder> GenerateRoutePaths(Shell shell, Uri request, Uri originalRequest)
+		internal static List<RouteRequestBuilder> GenerateRoutePaths(Shell shell, Uri request, Uri originalRequest, bool enableRelativeShellRoutes)
 		{
-			request = FormatUri(request);
-			originalRequest = FormatUri(originalRequest);
-
 			var routeKeys = Routing.GetRouteKeys();
 			for (int i = 0; i < routeKeys.Length; i++)
 			{
+				if(routeKeys[i] == originalRequest.OriginalString)
```
samhouts:
```suggestion
				if (routeKeys[i] == originalRequest.OriginalString)
```

Xamarin.Forms.Core/Shell/ShellUriHandler.cs
```diff
@@ -100,43 +116,49 @@ public static NavigationRequest GetNavigationRequest(Shell shell, Uri uri)
...
 				{
 					var route = routeKeys[i];
-					var uri = ConvertToStandardFormat(shell, new Uri(route, UriKind.RelativeOrAbsolute));
-					// Todo is this supported?
+					var uri = ConvertToStandardFormat(shell, CreateUri(route));
 					if (uri.Equals(request))
 					{
-						var builder = new RouteRequestBuilder(route, route, null, segments);
-						return new List<RouteRequestBuilder> { builder };
+						throw new Exception($"Global routes currently cannot be the only page on the stack so absolute routing to global routes is not supported. For now just navigate to: {originalRequest.OriginalString.Replace("//","")}");
```
samhouts:
```suggestion
						throw new Exception($"Global routes currently cannot be the only page on the stack, so absolute routing to global routes is not supported. For now, just navigate to: {originalRequest.OriginalString.Replace("//","")}");
```

Xamarin.Forms.Core/Shell/ShellUriHandler.cs
```diff
@@ -160,24 +182,42 @@ internal static List<RouteRequestBuilder> GenerateRoutePaths(Shell shell, Uri re
...
-						return pureRoutesMatch;
 
+					SearchPath(currentLocation.LowestChild, null, segments, pureGlobalRoutesMatch, 0, ignoreGlobalRoutes: false);
+					ExpandOutGlobalRoutes(pureGlobalRoutesMatch, routeKeys);
+					pureGlobalRoutesMatch = GetBestMatches(pureGlobalRoutesMatch);
 					if (pureGlobalRoutesMatch.Count > 0)
+					{
+						// currently relative routes to shell routes isn't supported as we aren't creating navigation stacks
+						// So right now we will just throw an exception so that once this is implemented
+						// GotoAsync doesn't start acting inconsistely and all of a suddent starts creating routes
```
samhouts:
```suggestion
						// GotoAsync doesn't start acting inconsistently and all of a suddent starts creating routes
```

PR #5961, opened 2019-04-19
Shell hamburger

### LINKED ISSUE #3124: [Shell] 3bar.png missing for the flyout menu hamburger
### Description
The resource cannot be found. It works fine if I copy the images manually to the head projects.

### Basic Information

- Version with issue: 3.2.0.599850

### Reproduction Link

https://github.com/davidortinau/Gastropods

Parent: #2415

PureWeen:
### Description of Change ###

Remove the bundled 3bar.png files. From what I can tell these won't get auto included when bundled with a nuget. In theory you can bundle content with the nuspec but iOS won't pickup those resources and pull them into the build. We could add a target to add the resource in but seeing as how it's just 3 horizontal lines, just drawing it seemed like the best solution for now. The PNG is basically just 3 rectangles.

### Issues Resolved ### 
- fixes #3124 

### Platforms Affected ### 
- iOS

One is the png and the other is the drawing

image

### Testing Procedure ###
Run the shell and make sure the hamburger shows up and looks ok

### PR Checklist ###

- [ ] Has automated tests 


jcmanke:
I'm certain that I've made a NuGet package that included images for iOS as BundleResources before and it worked, but I like the workaround.

PureWeen:
@jcmanke I messed around with it a bit and was able to create a package that included them with the nuget but it still didn't include them at compile time.  This PR (https://github.com/NuGet/NuGet.Client/pull/225) added those types of resources to the white list but I think not until xamarin projects get converted to full on sdk projects will that type of bundling just work out of the box

UWP does work when you include assets like this but ios/android doesn't at this point (AFAICT)

Xamarin.Forms.Platform.iOS/Renderers/ShellPageRendererTracker.cs
```diff
@@ -247,14 +250,55 @@ async Task SetDrawerArrowDrawableFromFlyoutIcon()
...
+			ctx.SaveState();
+			ctx.SetStrokeColor(UIColor.Blue.CGColor);
+
+			float size = 3f;
+			float start = 4f;
+			ctx.SetLineWidth(size);
+
+			for(int i = 0; i< 3; i++)
+			{
+				ctx.MoveTo(1f, start + i * (size * 2));
```
kingces95: Nice. Just like LOGO writer.

PR #5981, opened 2019-04-23
increase TargetPlatformVersion for Windows UI compatibility

PureWeen:
### Work to be done still ###
I feel like maintaining this range of compatibility from 16299 to 17763 will be ok but I'd like to add some build processes to the CI that will do a test build of the platform at 16299 just in case we accidentally use any APIs that aren't available

### Description of Change ###
We'd like to start moving towards using https://github.com/Microsoft/microsoft-ui-xaml#version-support especially for the work being done on Shell

> The Microsoft.UI.Xaml 2.1 NuGet package requires your project to have TargetPlatformVersion >= 10.0.17763.0 and TargetPlatformMinVersion >= 10.0.14393.0 when building.

https://github.com/xamarin/Xamarin.Forms/pull/5936

### Platforms Affected ### 
- UWP

### Testing Procedure ###
- I'll run all our UI tests to see if anything additional breaks

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/EmbeddingTestBeds/Embedding.UWP/Embedding.UWP.csproj
+++ b/EmbeddingTestBeds/Embedding.UWP/Embedding.UWP.csproj
@@ -11,7 +11,7 @@
     <AssemblyName>Embedding.UWP</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion>10.0.16299.0</TargetPlatformVersion>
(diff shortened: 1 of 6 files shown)


dotMorten:
If you do this, I'd recommend also adding the PlatformSpecific Analyzer, so you don't accidentially use non-supported API without a check first: https://docs.microsoft.com/en-us/windows/communitytoolkit/platform-specific/platformspecificanalyzer

PureWeen:
related

https://github.com/xamarin/Xamarin.Forms/pull/5999

PR #5985, opened 2019-04-24
Added new named Font Sizes, iOS Named fonts now all AutoScale based on Accessibility Options

### LINKED ISSUE #5474: Accessibility: On iOS Forms should use DynamicType
On iOS, Forms should use dynamic type by default, which would provide the different sizes and weights for fonts and should respond to font size changes when instructed by the operating system.

This is necessary for users that need larger font sizes for accessibility reasons, but also would make Forms work better and consistently with the rest of the OS and applications.

Resources:
https://developer.apple.com/documentation/uikit/uifont/scaling_fonts_automatically

Anecdotally, Marco Arment reports that for his own app over 40% of his users do not use the default font size:

https://twitter.com/marcoarment/status/1099295178651783169?s=21

Clancey:
We now use the iOS Prefered Font Scaling stuff so fonts are magical

### Description of Change ###

Added new Named Fonts

fixes #5474

### API Changes ###
Added:
NamedSize Sizes
		Large = 4,
		Body = 5,
		Header = 6,
		Title = 7,
		Subtitle = 8,
		Caption = 9

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS
- Android
- UWP

iOS Font named font sizes now Scale just like Android does.

### Testing Procedure ###
Use a named font,
Set the Large Font Size  option in iOS, Accessibility options.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/FontSizeConverter.cs
+++ b/Xamarin.Forms.Core/FontSizeConverter.cs
@@ -34,6 +34,16 @@ namespace Xamarin.Forms
 					namedSize = NamedSize.Medium;
 				else if (value.Equals(nameof(NamedSize.Large), sc))
 					namedSize = NamedSize.Large;
+				else if (value.Equals(nameof(NamedSize.Body), sc))
+					namedSize = NamedSize.Body;
+				else if (value.Equals(nameof(NamedSize.Caption), sc))
+					namedSize = NamedSize.Caption;
+				else if (value.Equals(nameof(NamedSize.Header), sc))
+					namedSize = NamedSize.Header;
+				else if (value.Equals(nameof(NamedSize.Subtitle), sc))
+					namedSize = NamedSize.Subtitle;
+				else if (value.Equals(nameof(NamedSize.Title), sc))
+					namedSize = NamedSize.Title;
 				else if (Enum.TryParse(value, ignoreCase, out NamedSize ns))
 					namedSize = ns;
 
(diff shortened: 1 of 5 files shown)


Clancey:
The values for [Android](https://material.io/design/typography/the-type-system.html#type-scale) and [UWP](https://docs.microsoft.com/en-us/windows/uwp/design/style/typography) come from the Typography guides.

Clancey:
The values for [Android](https://material.io/design/typography/the-type-system.html#type-scale) and [UWP](https://docs.microsoft.com/en-us/windows/uwp/design/style/typography) come from the Typography guides.

rmarinho:
Hey @Clancey  I have been playing a little bit with this, and I think we need to revert this or think on a better way. 

1 - this introduces a breaking change, that means that now our users if they have Accessibility on the app will use the font scale and existing UI will look different , breaking existing app designs.

2 - This only works when you restart the app. Since you get the font scale when iOSPlatformServices are created and you don't listen to accessibility changes and never update 

So without your pr here's how a default app looks.

IMG_4057

When I enable your pr the UI will look different, we need the user to opt in for this imo.

IMG_4059
IMG_4058

And when I have the app running and change the Accessibility the UI font 

IMG_4060

IMG_4061

Clancey:
I know, I don't listen to the changes. The reason is the font's won't update anyways unless the user has dynamic styles enabled.

Clancey:
As for being a breaking change, This is more of an alignment. If you use Accessibility on  Android using the named fonts it scales properly just like this does now. Using named fonts never expose a specific size. This is the expected behavior and it now properly aligns with Android behavior.

PR #5990, opened 2019-04-24
[iOS] remove SetPaddingInsets api and just have it pad content and flyout by default

### LINKED ISSUE #5971: Shell layout different between iOS 10 and later versions
### Description

Layout with relation to the navigation area differs between versions.

### Steps to Reproduce

1. Run test 4684 in Control Gallery on iOS 10.
2. Run test 4684 in Control Gallery on iOS >10.

### Expected Behavior

Same layout.

### Actual Behavior

Parts of the content are not visible on iOS 10.

### Basic Information

- Version with issue: 4.0.0

### Screenshots

iOS 10:
ios10

iOS 11:
ios11

PureWeen:
### Description of Change ###
- Remove Shell.SetPaddingInsets api for now as it's not really serving it's correct purpose
- Fixed iOS so that it just pads the Flyout and page renderer the same on iOS and android which means you don't have to set SafeArea anymore
- Also fixes ios10 to pad the page area content

### Issues Resolved ### 
- fixes #5971

### API Changes ###

 Removed:
 - Shell.SetPaddingInsets 
 None

### Platforms Affected ### 
- iOS

- Flyout now offsets from notch or status bar on ios10
- when the header scrolls with the list it doesn't show up behind the clock anymore

Before / After

image

When scrolling I added a mask so it doesn't overlap the status bar

image

iOS 10
image

image

### Testing Procedure ###
- 4684 should pass on ios10

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
@@ -339,32 +339,24 @@ namespace Xamarin.Forms.Platform.iOS
 
 		void UpdateShellInsetPadding()
 		{
-			if (Element == null)
+			if (!(Element?.Parent is ShellContent))
 				return;
 
-			var setInsets = Shell.GetSetPaddingInsets(Element);
+			nfloat topPadding = 0;
+			nfloat bottomPadding = 0;
 
-			if (!setInsets && Element.Parent != null)
-				setInsets = Shell.GetSetPaddingInsets(Element.Parent);
-
-			if (setInsets)
+			if (Forms.IsiOS11OrNewer)
 			{
-				nfloat topPadding = 0;
-				nfloat bottomPadding = 0;
-
-				if (Forms.IsiOS11OrNewer)
-				{
-					topPadding = View.SafeAreaInsets.Top;
-					bottomPadding = View.SafeAreaInsets.Bottom;
-				}
-				else
-				{
-					topPadding = TopLayoutGuide.Length;
-					bottomPadding = BottomLayoutGuide.Length;
-				}
-
-				(Element as Page).Padding = new Thickness(0, topPadding, 0, bottomPadding);
+				topPadding = View.SafeAreaInsets.Top;
+				bottomPadding = View.SafeAreaInsets.Bottom;
 			}
+			else
+			{
+				topPadding = TopLayoutGuide.Length;
+				bottomPadding = BottomLayoutGuide.Length;
+			}
+
+			(Element as Page).Padding = new Thickness(0, topPadding, 0, bottomPadding);
 		}
 
 		void UpdateStatusBarPrefersHidden()
--- a/Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
@@ -301,7 +301,7 @@ namespace Xamarin.Forms.Platform.iOS
 		{
 			get
 			{
-				var animation = ((Page)Element).OnThisPlatform().PreferredStatusBarUpdateAnimation();
+				var animation = Page.OnThisPlatform().PreferredStatusBarUpdateAnimation();
 				switch (animation)
 				{
 					case (PageUIStatusBarAnimation.Fade):
@@ -356,7 +356,7 @@ namespace Xamarin.Forms.Platform.iOS
 				bottomPadding = BottomLayoutGuide.Length;
 			}
 
-			(Element as Page).Padding = new Thickness(0, topPadding, 0, bottomPadding);
+			Page.Padding = new Thickness(0, topPadding, 0, bottomPadding);
 		}
 
 		void UpdateStatusBarPrefersHidden()
@@ -364,7 +364,7 @@ namespace Xamarin.Forms.Platform.iOS
 			if (Element == null)
 				return;
 
-			var animation = ((Page)Element).OnThisPlatform().PreferredStatusBarUpdateAnimation();
+			var animation = Page.OnThisPlatform().PreferredStatusBarUpdateAnimation();
 			if (animation == PageUIStatusBarAnimation.Fade || animation == PageUIStatusBarAnimation.Slide)
 				UIView.Animate(0.25, () => SetNeedsStatusBarAppearanceUpdate());
 			else
@@ -384,7 +384,7 @@ namespace Xamarin.Forms.Platform.iOS
 
 		public override bool PrefersStatusBarHidden()
 		{
-			var mode = ((Page)Element).OnThisPlatform().PrefersStatusBarHidden();
+			var mode = Page.OnThisPlatform().PrefersStatusBarHidden();
 			switch (mode)
 			{
 				case (StatusBarHiddenMode.True):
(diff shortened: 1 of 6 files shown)


davidortinau:
In many designs, Outlook on mobile for example, the flyout has design (background or otherwise) that does extend up into the status bar "unsafe" area. How would I accomplish that?

Also notice that Outlook will hide the clock and status elements, which is possible though I do not believe we have a platform specific for this yet? In such a case I would want to opt into my content scrolling all the way and not be clipped.

PureWeen:
@davidortinau TBD :-) 

> In many designs, Outlook on mobile for example, the flyout has design (background or otherwise) that does extend up into the status bar "unsafe" area. How would I accomplish that?

> Also notice that Outlook will hide the clock and status elements, which is possible though I do not believe we have a platform specific for this yet? In such a case I would want to opt into my content scrolling all the way and not be clipped.

Before this PR the fly out on iOS just went into status bar and android didn't plus there's no way to hide the status bar so the user is just kind of stuck with weird behavior. My intention with this PR is to strip Shell down to a default so the user isn't stuck thinking "this looks really bad on iOS and I have no way to fix" 

If we are happy with the default  here then we can start adding some additional apis down the road like FlyoutLayout="FullHeight" and StatusBar="Hide" which will create a consistent behavior between iOS and Android.

Right now all the user has is *SetPaddingInsets* which really only equates to setting safe areas on ios11 and 12 and has no effect on the flyout

In a perfect world this PR would address everything you bring up and give you hooks for those. But for now I'd rather just remove a weird api and make it all work in an expected fashion

PureWeen:
Issue 4684 is now passing on ios10

The ios10 failure here (1939) isn't related to this PR

Xamarin.Forms.Platform.iOS/Renderers/ShellTableViewController.cs
```diff
@@ -91,5 +102,26 @@ void OnScrolled(object sender, UIScrollView e)
...
+
+					return (float)UIApplication.SharedApplication.StatusBarFrame.Size.Height;
+
+				}
+
+				if (UIApplication.SharedApplication.KeyWindow != null)
+					return (float)UIApplication.SharedApplication.KeyWindow.SafeAreaInsets.Top;
+
+				if (UIApplication.SharedApplication.Windows.Length > 0)
+					return (float)UIApplication.SharedApplication.Windows[0].SafeAreaInsets.Top;
```
paymicro: What does iOS11+ require? `KeyWindow` and `Windows` require iOS 2.0+

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Xamarin.Forms.Controls.Issues.Shared.projitems
```diff
@@ -1161,7 +1161,7 @@
   <ItemGroup>
     <EmbeddedResource Include="$(MSBuildThisFileDirectory)A11yTabIndex.xaml">
       <SubType>Designer</SubType>
-      <Generator>MSBuild:Compile</Generator>
+      <Generator>MSBuild:UpdateDesignTimeXaml</Generator>
```
paymicro: Is this a necessary change?

Xamarin.Forms.Platform.iOS/Renderers/ShellTableViewController.cs
```diff
@@ -91,5 +102,26 @@ void OnScrolled(object sender, UIScrollView e)
 
 			LayoutParallax();
 		}
+
+		float SafeAreaOffset
+		{
+			get
+			{
+				if (!Forms.IsiOS11OrNewer)
```
paymicro: maybe better to save the first value in a variable then give it?

Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
```diff
@@ -339,32 +339,24 @@ void UpdateUseSafeArea()
...
+				topPadding = View.SafeAreaInsets.Top;
+				bottomPadding = View.SafeAreaInsets.Bottom;
+			}
+			else
+			{
+				topPadding = TopLayoutGuide.Length;
+				bottomPadding = BottomLayoutGuide.Length;
 			}
+
+			(Element as Page).Padding = new Thickness(0, topPadding, 0, bottomPadding);
```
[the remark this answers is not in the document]
PureWeen: Yea there's actually already a page property so these should all just be converted to 
`Page.Padding`

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Xamarin.Forms.Controls.Issues.Shared.projitems
PureWeen: the IDE just makes it by default :-/ so I like to just get these kinds of these checked in otherwise it'll just keep making this change. This got defined wrong when it was initially checked in

Xamarin.Forms.Platform.iOS/Renderers/ShellTableViewController.cs
PureWeen: The problem here is timing. I found that the KeyWindow doesn't get set at the point in time where I need this value so the second check is just a fallback if the KeyWindow hasn't been set yet.  I moved this code into Platform so it can at least be centralized if there are other cases that need it which will mainly just be components that sit above everything and need to know window constraints

Xamarin.Forms.Platform.iOS/Renderers/ShellTableViewController.cs
PureWeen: the problem with saving it is that the safe area changes on device rotation so we could detect for that then clear it out but this seems safer. 

The way this all works with a rotated device + notch + rounded corners needs to be fleshed out a bit

Xamarin.Forms.Platform.iOS/Renderers/ShellTableViewController.cs
```diff
@@ -40,10 +41,20 @@ void OnStructureChanged(object sender, EventArgs e)
...
+			TableView.Frame = parent.Bounds.Inset(0, SafeAreaOffset);
 			if (_headerView != null)
-				_headerView.Frame = new CGRect(0, _headerOffset, parent.Frame.Width, _headerSize);
+			{
+				_headerView.Frame = new CGRect(0, _headerOffset + SafeAreaOffset, parent.Frame.Width, _headerSize);
+
+				if (_headerOffset < 0)
+				{
+					CAShapeLayer shapeLayer = new CAShapeLayer();
+					CGRect rect = new CGRect(0, _headerOffset * -1, parent.Frame.Width, _headerSize + _headerOffset);
```
paymicro: `_headerSize + _headerOffset` cannot be negative?

PureWeen: I don't think so... but adding a check just in case is probably worth while

paymicro: I checked, it may be negative if `FlyoutHeaderBehavior` is `CollapseOnScroll`, but no exception is thrown.

PR #5994, opened 2019-04-24
simplify App.Current

StephaneDelcroix:
simplify Application.Current getter, other minor cosmetic changes

### DIFF
--- a/Xamarin.Forms.Core/Application.cs
+++ b/Xamarin.Forms.Core/Application.cs
@@ -12,22 +12,16 @@ namespace Xamarin.Forms
 {
 	public class Application : Element, IResourcesProvider, IApplicationController, IElementConfiguration<Application>
 	{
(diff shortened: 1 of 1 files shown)


adammeaney:
```public static Application Current { get; set; }```

Did we mean to have a public setter for Application.Current? I can't imagine that is on purpose with the methods for messing with current being EditorBrowsable attributed.

File: Xamarin.Forms.Core/Application.cs [hunk 1]
```diff
@@ -62,18 +56,7 @@ public IAppLinks AppLinks
...
 
-		public static Application Current
-		{
-			get { return s_current; }
-			set
-			{
-				if (s_current == value)
-					return;
-				if (value == null)
-					s_current = null; //Allow to reset current for unittesting
```
kingces95: Lovely..

PR #5999, opened 2019-04-24
Properly guard against statusbar usage

dotMorten:
### Description of Change ###

If you add the `Microsoft.Toolkit.Uwp.PlatformSpecificAnalyzer` nuget package, you'll find the analyzer points out these locations as not correctly guarded against usage.

### Issues Resolved ### 
Potential to hit a type that isn't present.

### API Changes ###
None

 None

### Platforms Affected ### 
- UWP

None
None

Not applicable

### Testing Procedure ###
Add the `Microsoft.Toolkit.Uwp.PlatformSpecificAnalyzer` nuget package.

You'll still see the analyzer points out the MobileStatusBar property to not be guarded. There's not really away around that except to disable the warning using `#pragma warning disable UWP001 // Platform-specific`. I didn't want to add that though, unless the nuget package gets permanently added to the project, and any usage of the property will get called out by the analyzer, so it would be safe to disable here.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.UAP/Platform.cs
+++ b/Xamarin.Forms.Platform.UAP/Platform.cs
@@ -339,46 +339,52 @@ namespace Xamarin.Forms.Platform.UWP
 		{
 			_bounds = new Rectangle(0, 0, _page.ActualWidth, _page.ActualHeight);
 
-			StatusBar statusBar = MobileStatusBar;
-			if (statusBar != null)
-			{
-				bool landscape = Device.Info.CurrentOrientation.IsLandscape();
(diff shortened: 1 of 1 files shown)


PureWeen:
Is this code necessary? 

Ez already does a check for this here https://github.com/xamarin/Xamarin.Forms/blob/0aef19be27eeeb1735cce898675f823a44431495/Xamarin.Forms.Platform.UAP/Platform.cs#L26 

And then does null checks.

dotMorten:
@PureWeen If you add the analyzer (which I assume you will if you increase the SDK you compile against), you'd need either that, or `#pragma warning disable UWP001` to avoid the warning

PR #6012, opened 2019-04-25
Profiler V5

kingces95:
Debug CI


kingces95:
I introduced an optional parameter on a public method. That could cause a binary break. Changed that to an overload. So fingers crossed. If this goes through then I'll try push the whole thing including the refactor of the init code to allow for piecemeal exclusion/inclusion. The refactor should be the same as before -- works locally -- but who knows.

PR #6024, opened 2019-04-26
Ehandler2

StephaneDelcroix:
report file location (in source) in addition to xpe

### DIFF
--- a/Xamarin.Forms.Core/Internals/ResourceLoader.cs
+++ b/Xamarin.Forms.Core/Internals/ResourceLoader.cs
@@ -54,6 +54,17 @@ namespace Xamarin.Forms.Internals
 			public bool UseDesignProperties { get; set; }
 		}
 
-		internal static Action<Exception> ExceptionHandler { get; set; }
+		static Action<Exception> exceptionHandler1;
+
+		[Obsolete("2 is better than 1")]
+		internal static Action<Exception> ExceptionHandler {
+			get => exceptionHandler1;
+			set {
+				exceptionHandler1 = value;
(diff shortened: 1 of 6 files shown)


Xamarin.Forms.Core/Internals/ResourceLoader.cs
```diff
@@ -53,6 +53,17 @@ public class ResourceLoadingResponse
 			public bool UseDesignProperties { get; set; }
 		}
 
-		internal static Action<Exception> ExceptionHandler { get; set; }
+		private static Action<Exception> exceptionHandler1;
```
samhouts:
```suggestion
		static Action<Exception> exceptionHandler1;
```

samhouts:
These are internal, but should we do EditorBrowsable on the obsolete for consistency?

Xamarin.Forms.Core/Internals/ResourceLoader.cs
```diff
@@ -53,6 +53,17 @@ public class ResourceLoadingResponse
 			public bool UseDesignProperties { get; set; }
 		}
 
-		internal static Action<Exception> ExceptionHandler { get; set; }
+		static Action<Exception> exceptionHandler1;
+
+		[Obsolete("2 is better than 1")]
+		internal static Action<Exception> ExceptionHandler {
```
kingces95: An obsolete internal method? Isn't that funny? Is something stopping us from updating all the call sites to use the new and improve `ExceptionHandler`?

StephaneDelcroix: some people (you know who you are, I'm looking at you 👀) are using our internals methods with different versions of Forms. So removing this would make that (👀 that one) tool fail with newer version of forms, as we don't release at the same cycle

Issue #5656, opened 2019-03-21
[Enhancement] Setup CI to install nugets into different projects and run basic tests to verify compatibility


PureWeen:
## Summary
We've added sets of ABI comparison checks to our CI so that we can validate breaking changes between releases but it would still be a good idea to have an automated process that installs a nuget and then runs a basic set of UI tests against said app.

For example it can
- pull down smart hotel 360
- install latest nugets
- click through a few basic buttons
- This should also run all of these projects with proguard and/or the new r8/d8 settings to ensure all proguard and multidex rules are working

If we had something like this in place then our ABI break with IVisualElementRenderer would have been caught a few months ago.

Ideally we can also expand this out to test default apps from 3rd parties. We can run this same set of tests against a baseline Telerik, SyncFusion, FFImageLoading, CarouselView, etc..

The cake scripts we added recently make it fairly easy to install new nugets either through the CLI or just by doing a basic string replace on the csproj files

## Intended Use Case
Verify unexpected ABI or other breaks that will cause hardship to 3rd party libraries

jfversluis:
Old and no clue what this is about anymore. Let's close it.

PR #6032, opened 2019-04-27
additional proguard rules for shell

### LINKED ISSUE #5799: [Visual] Entry control page does not load in release
### Description
If there is an entry control on the page with Visual="Material" the page will not load in Release mode on android

### Steps to Reproduce

1.  Open TheLittleThingsPlayground
2. Change MainActivity.cs code
From
``` C#
  global::Xamarin.Forms.Forms.SetFlags( "FastRenderers_Experimental", "Shell_Experimental");
  global::Xamarin.Forms.Forms.Init(this, bundle);
  LoadApplication(new App());
```
To 
``` C#
  global::Xamarin.Forms.Forms.SetFlags( "FastRenderers_Experimental" );//, "Shell_Experimental");
  global::Xamarin.Forms.Forms.Init(this, bundle);
  Xamarin.Forms.FormsMaterial.Init( this, bundle );
  LoadApplication(new App());
```

2.  start the app in release mode
3.  Go to entries page

### Expected Behavior
Entries page loads normally

### Actual Behavior
Entries page shows only a black screen

### Basic Information

- Version with issue: XF 3.6.0.220655
- IDE: Visual Studio 2017 15.9.10
- Platform Target Frameworks:
  - Android: 9.0
- Affected Devices:
All Android

This is not exclusive to TheLittleThingsPlayground, but is a good example project to illustrate the issue

PureWeen:
### Description of Change ###
Shell needs some additional help to work with proguard.

I also added a keep for multidex which will be fixed in an upcoming VS release but I figured we might as well just include it for now to remove friction

Also added a property to disable default proguard rules if users want more fine grained control

```xml
<Target Name="IncludeProguardForAndroid" Condition="'$(XFDisableDefaultProguardConfiguration)' != 'True'">
    <ItemGroup>
      <ProguardConfiguration Include="$(MSBuildThisFileDirectory)MonoAndroid10\proguard.cfg" />
    </ItemGroup>
  </Target>
```
### Issues Resolved ### 

- fixes #5799 

### Platforms Affected ### 
- Android

### Testing Procedure ###
install the nuget for this PR into a XF project using shell.  On that project enable multidex, proguard, and set the linker to full. At this point the project should run without any errors

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/.nuspec/proguard.cfg
+++ b/.nuspec/proguard.cfg
@@ -1,2 +1,4 @@
 -keep class android.support.v7.widget.FitWindowsFrameLayout { *; }
 -dontwarn android.support.v7.widget.FitWindowsFrameLayout
+-keep class android.support.design.** { *; }
+-keep class android.support.multidex.MultiDexApplication { *; }
--- a/.nuspec/Xamarin.Forms.targets
+++ b/.nuspec/Xamarin.Forms.targets
@@ -161,7 +161,7 @@
     </PrepareForBuildDependsOn>
   </PropertyGroup>
   
-  <Target Name="IncludeProguardForAndroid">
+  <Target Name="IncludeProguardForAndroid" Condition="'$(XFDisableDefaultProguardConfiguration)' != 'True'">
     <ItemGroup>
       <ProguardConfiguration Include="$(MSBuildThisFileDirectory)MonoAndroid10\proguard.cfg" />
     </ItemGroup>
--- a/Xamarin.Forms.Sandbox.Android/proguard.cfg
+++ b/Xamarin.Forms.Sandbox.Android/proguard.cfg
@@ -1,2 +1,4 @@
 -keep class android.support.v7.widget.FitWindowsFrameLayout { *; }
 -dontwarn android.support.v7.widget.FitWindowsFrameLayout
+-keep class android.support.design.** { *; }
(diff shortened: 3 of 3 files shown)


.nuspec/proguard.cfg
```diff
@@ -1,2 +1,4 @@
 -keep class android.support.v7.widget.FitWindowsFrameLayout { *; }
--dontwarn android.support.v7.widget.FitWindowsFrameLayout
\ No newline at end of file
+-dontwarn android.support.v7.widget.FitWindowsFrameLayout
+-keep class android.support.design.** { *; }
```
jonathanpeppers: If you get CI in place to check these rules are working, it might be worth it to get more specific and not use `**`? If you could run XF's UITest suite with proguard+release, that would help this case.

proguard/r8 can probably strip out more code if you listed individual classes.

You can also do this to allow non-public members to get stripped:
```
-keep class android.support.design.** { public *; }
```

This is all trial and error though, you will have to run a release app with proguard on to see if you get any java missing-method exceptions at runtime.

PureWeen: Yea once we get this in place
https://github.com/xamarin/Xamarin.Forms/issues/5656

We'll look at refining this down

PR #6047, opened 2019-04-29
Add EventToCommandBehavior

dansiegel:
### Description of Change ###

This closes #5685 from @davidortinau for an EventToCommandBehavior. This makes two changes

1) Adds the AssociatedObject in Behavior{T} which is the major missing piece of Behavior and even shown in numerous samples and docs.
2) Adds the EventToCommandBehavior from Prism which provides support for invoking a command based on a Path Name of a property in the EventArgs and/or using a IValueConverter to convert the EventArgs/Property 

### API Changes ###

Added:
 - T Behavior<T>.AssociatedObject { get; } 
 - EventToCommandBehavior
 
### Platforms Affected ### 

- Core

None

Not applicable

### Testing Procedure ###

n/a

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Interactivity/EventToCommandBehavior.cs
+++ b/Xamarin.Forms.Core/Interactivity/EventToCommandBehavior.cs
@@ -1,0 +1,160 @@
+﻿using System;
+using System.Globalization;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using System.Windows.Input;
+
+namespace Xamarin.Forms.Core.Interactivity
+{
+	public class EventToCommandBehavior : Behavior
+	{
+		public static readonly BindableProperty EventNameProperty =
+			BindableProperty.Create(nameof(EventName), typeof(string), typeof(EventToCommandBehavior));
+
+		public static readonly BindableProperty CommandProperty =
+			BindableProperty.Create(nameof(Command), typeof(ICommand), typeof(EventToCommandBehavior));
+
+		public static readonly BindableProperty CommandParameterProperty =
+			BindableProperty.Create(nameof(CommandParameter), typeof(object), typeof(EventToCommandBehavior));
+
+		public static readonly BindableProperty EventArgsConverterProperty =
+			BindableProperty.Create(nameof(EventArgsConverter), typeof(IValueConverter), typeof(EventToCommandBehavior));
+
+		public static readonly BindableProperty EventArgsConverterParameterProperty =
+			BindableProperty.Create(nameof(EventArgsConverterParameter), typeof(object), typeof(EventToCommandBehavior));
+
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Core/Interactivity/EventToCommandBehavior.cs
```diff
@@ -0,0 +1,160 @@
...
+			if (_handler != null)
+			{
+				_eventInfo.RemoveEventHandler(bindable, _handler);
+			}
+			_handler = null;
+			_eventInfo = null;
+			base.OnDetachingFrom(bindable);
+		}
+
+		private void AddEventHandler(EventInfo eventInfo, object item, Action<object, EventArgs> action)
```
samhouts:
```suggestion
		void AddEventHandler(EventInfo eventInfo, object item, Action<object, EventArgs> action)
```

dansiegel:
@StephaneDelcroix no offense intended, I hadn't noticed your comments on the issue from @davidortinau. As I mentioned above I was just bringing this over from what Prism has been using for 2 years... the AssociatedObject on the Behavior<T> has been in every sample I've seen since the introduction of Behaviors and why so many people just created a BehaviorBase<T> that adds exactly that single property.

> Behaviors are meant to be declared in Styles, and have shared instances. So your behavior should keep track of an AssociatedObject, but an array of AssociatedObjects.

Regarding your comments on AssociatedObject(s)

- I honestly wasn't aware that Style even took Behaviors... TBH I don't recall seeing it as an option in Intellisense, nor would have occurred to me to try to use a Behavior on a Style.
- I cannot honestly think of a single case where I've ever used a Behavior in a manner where it could or would be shared between elements.
- I cannot think of any blogs (official or otherwise) or documentation that indicates that Behaviors could be used in the manner you describe. Which may not mean there isn't any, but should at least indicate that it's very likely that what I've described is probably pretty common.

As I think on what you're asking for adding an AssociatedObjects element makes ZERO sense to me. It really only make sense to me that you want access to the AssociatedObject while executing custom code say in an EventHandler or something. If you're still that adamantly against AssociatedObject, I think it makes more sense to back those changes out of this PR so we can get the EventToCommandBehavior part merged, and we can deal with the AssociatedObject issue separately.

StephaneDelcroix:
@dansiegel my wrong. I thought you replied and dismissed my review (https://github.com/xamarin/Xamarin.Forms/pull/5685#discussion_r275712600) but that wasn't you. I'll come back with a solution when I have a few free minutes. thanks for your patience.

Now that you know that you can use Behaviors in Styles (like in Silverlight or WPF), I'm sure you're thinking of 100 of ways doing that would make all of your xaml looks simpler and more beautiful :)

samhouts:
I believe we're in agreement that this is not the approach we'll take at this time, so I will close this PR. Thank you!!

PR #6062, opened 2019-05-01
[Android] Image load fixes

### LINKED ISSUE #6010: [Android]  New ImageSource needs to intelligently invalidate Elements
### Description
https://github.com/xamarin/Xamarin.Forms/pull/4915/commits/665961b0785d42edb8c1bb089a4fa7d9992306f3

The intent of  the invalidation was to handle scenarios where images are being loaded post layout but as is it's causing the layout to thrash. A basic page with a few images was calling invalidatemeasure about 900 times with these invalidation vs the normal 8 without those.

Commenting the code out will probably mean that in certain cases when loading images from the web they currently won't layout correctly or if the image is too slow to load

### LINKED ISSUE #6061: [Android] UI Test 4597 fails on API 19
### Description

The UI test for issue #4597 does not pass on API 19 devices.

### Steps to Reproduce

1. Open Control Gallery
2. Navigate to Issue 4597
3. Follow the on-screen instructions

### Expected Behavior

All of the images disappear.

### Actual Behavior

One of the images does not disappear.

### Basic Information

- Version with issue: 4.0.0

PureWeen:
### Description of Change ###
The bulk of this code is to make it so the layout and loading of images doesn't thrash the system as much
- I added a basic lrucache for images loaded from url (Added Peppers and Matt on here to validate my effort)
- I made the button invalidate the measure only after it's been laid out otherwise you get all these repetitive invalidations on initial load
- API 19 doesn't look to clear the image out when it's set to translucent so for < Lollipop I just set the drawable to null

### Issues Resolved ### 

- fixes #6061
- fixes #6010 

### Platforms Affected ### 
- Android

### Testing Procedure ###
Test all the image loading on android you can
Issue4915 does a lot of different image loading against different types

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/ButtonRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/ButtonRenderer.cs
@@ -54,7 +54,6 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 
 		public override SizeRequest GetDesiredSize(int widthConstraint, int heightConstraint)
 		{
-			_buttonLayoutManager?.Update();
 			return base.GetDesiredSize(widthConstraint, heightConstraint);
 		}
 
--- a/Xamarin.Forms.Platform.Android/AppCompat/ButtonRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/ButtonRenderer.cs
@@ -15,7 +15,7 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 {
 	public class ButtonRenderer : ViewRenderer<Button, AppCompatButton>,
 		AView.IOnAttachStateChangeListener, AView.IOnClickListener, AView.IOnTouchListener,
-		IBorderVisualElementRenderer, IButtonLayoutRenderer
+		IBorderVisualElementRenderer, IButtonLayoutRenderer, IDisposedState
 	{
 		BorderBackgroundManager _backgroundTracker;
 		TextColorSwitcher _textColorSwitcher;
@@ -202,5 +202,6 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 		}
 
 		AppCompatButton IButtonLayoutRenderer.View => Control;
+		bool IDisposedState.IsDisposed => _isDisposed;
 	}
 }
--- a/Xamarin.Forms.Platform.Android/ButtonLayoutManager.cs
+++ b/Xamarin.Forms.Platform.Android/ButtonLayoutManager.cs
@@ -276,8 +276,20 @@ namespace Xamarin.Forms.Platform.Android
 			else
 				view.CompoundDrawablePadding = (int)Context.ToPixels(layout.Spacing);
 
+			Drawable existingImage = null;
+			var images = TextViewCompat.GetCompoundDrawablesRelative(view);
+			for (int i = 0; i < images.Length; i++)
+				if(images[i] != null)
+				{
+					existingImage = images[i];
+					break;
+				}
+
 			_renderer.ApplyDrawableAsync(Button.ImageSourceProperty, Context, image =>
 			{
+				if (image == existingImage)
+					return;
+
 				switch (layout.Position)
 				{
 					case Button.ButtonContentLayout.ImagePosition.Top:
@@ -294,11 +306,8 @@ namespace Xamarin.Forms.Platform.Android
 						TextViewCompat.SetCompoundDrawablesRelativeWithIntrinsicBounds(view, image, null, null, null);
 						break;
 				}
-
-				// Invalidating here causes a crazy amount of increased measure invalidations
-				// when I tested with Issue4484 it caused about 800 calls to invalidate measure vs the 8 without this
-				// I'm pretty sure it gets into a layout / invalidation loop where these are invalidating mid layout				
-				//_element?.InvalidateMeasureNonVirtual(InvalidationTrigger.MeasureChanged);
+		
+				_element?.InvalidateMeasureNonVirtual(InvalidationTrigger.MeasureChanged);
 			});
 		}
 	}
--- a/Xamarin.Forms.Platform.Android/ButtonLayoutManager.cs
+++ b/Xamarin.Forms.Platform.Android/ButtonLayoutManager.cs
@@ -86,7 +86,7 @@ namespace Xamarin.Forms.Platform.Android
 				return;
 
 			AppCompatButton view = View;
-			if (view == null || view.Layout == null)
+			if (view == null)
 				return;
 
 			Drawable drawable = null;
--- a/Xamarin.Forms.Platform.Android/ButtonLayoutManager.cs
+++ b/Xamarin.Forms.Platform.Android/ButtonLayoutManager.cs
@@ -34,6 +34,7 @@ namespace Xamarin.Forms.Platform.Android
 		bool _preserveInitialPadding;
 		bool _borderAdjustsPadding;
 		bool _maintainLegacyMeasurements;
+		bool _hasLayoutOccurred;
 
 		public ButtonLayoutManager(IButtonLayoutRenderer renderer)
 			: this(renderer, false, false, false, true)
@@ -141,6 +142,8 @@ namespace Xamarin.Forms.Platform.Android
 					}
 				}
 			}
+
+			_hasLayoutOccurred = true;
 		}
 
 		public void OnViewAttachedToWindow(AView attachedView)
@@ -306,8 +309,12 @@ namespace Xamarin.Forms.Platform.Android
 						TextViewCompat.SetCompoundDrawablesRelativeWithIntrinsicBounds(view, image, null, null, null);
 						break;
 				}
-		
-				_element?.InvalidateMeasureNonVirtual(InvalidationTrigger.MeasureChanged);
+
+				// I'm not thrilled about the BeginInvokeOnMainThread
+				// but without it the button won't resize based on the image dimensions if the image has changed
+				// you can see this using Issue4915
+				if (_hasLayoutOccurred)
+					Device.BeginInvokeOnMainThread(() => _element?.InvalidateMeasureNonVirtual(InvalidationTrigger.MeasureChanged));
 			});
 		}
 	}
--- a/Xamarin.Forms.Platform.Android/ButtonLayoutManager.cs
+++ b/Xamarin.Forms.Platform.Android/ButtonLayoutManager.cs
@@ -310,11 +310,8 @@ namespace Xamarin.Forms.Platform.Android
 						break;
 				}
 
-				// I'm not thrilled about the BeginInvokeOnMainThread
-				// but without it the button won't resize based on the image dimensions if the image has changed
-				// you can see this using Issue4915
 				if (_hasLayoutOccurred)
-					Device.BeginInvokeOnMainThread(() => _element?.InvalidateMeasureNonVirtual(InvalidationTrigger.MeasureChanged));
+					_element?.InvalidateMeasureNonVirtual(InvalidationTrigger.MeasureChanged);
 			});
 		}
 	}
--- a/Xamarin.Forms.Platform.Android/ImageCache.cs
+++ b/Xamarin.Forms.Platform.Android/ImageCache.cs
@@ -2,6 +2,7 @@
 using System.Collections.Concurrent;
 using System.Threading;
 using System.Threading.Tasks;
+using Android.Runtime;
 
 namespace Xamarin.Forms.Platform.Android
 {
@@ -22,7 +23,7 @@ namespace Xamarin.Forms.Platform.Android
 
 		void Put(string key, TimeSpan cacheValidity, global::Android.Graphics.Bitmap cacheObject)
 		{
-			_lruCache.Put(key, new CacheEntry() { TimeToLive = DateTimeOffset.Now.Add(cacheValidity), Data = cacheObject });
+			_lruCache.Put(key, new CacheEntry() { TimeToLive = DateTimeOffset.UtcNow.Add(cacheValidity), Data = cacheObject });
 		}
 
 		public Task<Java.Lang.Object> GetAsync(string cacheKey, TimeSpan cacheValidity, Func<Task<Java.Lang.Object>> createMethod)
@@ -30,6 +31,7 @@ namespace Xamarin.Forms.Platform.Android
 			return Task.Run(async () =>
 			{
 				SemaphoreSlim semaphoreSlim = null;
+				Java.Lang.Object innerCacheObject = null;
 
 				try
 				{
@@ -38,10 +40,9 @@ namespace Xamarin.Forms.Platform.Android
 
 					var cacheEntry = _lruCache.Get(cacheKey) as CacheEntry;
 
-					if (cacheEntry?.TimeToLive < DateTimeOffset.Now)
+					if (cacheEntry?.TimeToLive < DateTimeOffset.UtcNow || cacheEntry?.IsDisposed == true)
 						cacheEntry = null;
 
-					Java.Lang.Object innerCacheObject = null;
 					if (cacheEntry == null && createMethod != null)
 					{
 						innerCacheObject = await createMethod().ConfigureAwait(false);
@@ -54,8 +55,6 @@ namespace Xamarin.Forms.Platform.Android
 					{
 						innerCacheObject = cacheEntry.Data;
 					}
-
-					return innerCacheObject;
 				}
 				catch
 				{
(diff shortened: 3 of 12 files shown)


Xamarin.Forms.Platform.Android/AppCompat/ButtonRenderer.cs
```diff
@@ -54,7 +54,6 @@ protected override void SetContentDescription()
 
 		public override SizeRequest GetDesiredSize(int widthConstraint, int heightConstraint)
 		{
-			_buttonLayoutManager?.Update();
```
PureWeen: This was causing an infinite loop of layout invalidations

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -18,6 +18,43 @@ namespace Xamarin.Forms.Platform.Android
...
+
+			return _lruCache;
+		}
+
+		class ImageCache : global::Android.Util.LruCache
+		{
+
+			static int getCacheSize()
+			{
+				// taken from lru cache docs
```
hartez: Can we get a link to the docs where you found this? Easier for future maintainers to understand what's happening.

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -18,6 +18,43 @@ namespace Xamarin.Forms.Platform.Android
...
+					_lruCache = new ImageCache();
+			}
+
+			return _lruCache;
+		}
+
+		class ImageCache : global::Android.Util.LruCache
+		{
+
+			static int getCacheSize()
```
hartez: Pascal case.

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -146,10 +184,46 @@ internal static Task ApplyDrawableAsync(this IShellContext shellContext, Bindabl
 			{
 				try
 				{
-					using (var drawable = await context.GetFormsDrawableAsync(initialSource, cancellationToken))
+					string cachKey = String.Empty;
+
+					// Todo improve for other sources
+					// volley the requests up front so that if the same request comes in it isn't requeueued
```
hartez: requeueued -> requeued  :)

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -146,10 +184,46 @@ internal static Task ApplyDrawableAsync(this IShellContext shellContext, Bindabl
 			{
 				try
 				{
-					using (var drawable = await context.GetFormsDrawableAsync(initialSource, cancellationToken))
+					string cachKey = String.Empty;
```
hartez: cachKey -> cacheKey

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -146,10 +184,46 @@ internal static Task ApplyDrawableAsync(this IShellContext shellContext, Bindabl
...
+					// volley the requests up front so that if the same request comes in it isn't requeueued
+					if(initialSource is UriImageSource uri)
 					{
-						// TODO: it might be good to make sure the renderer has not been disposed
+						cachKey = Device.PlatformServices.GetMD5Hash(uri.Uri.ToString());
+						var cache = GetCache();
+						var cacheObject = cache.Get(cachKey);
+
+						// this only really gets hit during load when there are a lot of requests for the same image
+						while(cacheObject?.ToString() == "PLEASEHOLD")
```
hartez: Make "PLEASEHOLD" a `const string`.

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -146,10 +184,46 @@ internal static Task ApplyDrawableAsync(this IShellContext shellContext, Bindabl
...
+							{
+								cache.Remove(cachKey);
+							}
+						}
+						else
+						{
+							returnValue = new BitmapDrawable(context.Resources, bitmap);
+						}
 
+						// TODO: it might be good to make sure the renderer has not been disposed
```
hartez: We should go ahead and make the check.

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -159,7 +233,31 @@ internal static Task ApplyDrawableAsync(this IShellContext shellContext, Bindabl
...
+							using(returnValue)
+								onSet(returnValue);
+						}
+					}
+					else
+					{
+
+						using (var drawable = await context.GetFormsDrawableAsync(initialSource, cancellationToken))
+						{
+							// TODO: it might be good to make sure the renderer has not been disposed
```
hartez: Add the disposed check.

Xamarin.Forms.Platform.Android/ImageCache.cs
```diff
@@ -0,0 +1,108 @@
...
+			{
+				var maxMemory = (int)(Java.Lang.Runtime.GetRuntime().MaxMemory() / 1024);
+				cacheSize = maxMemory / 8;
+			}
+			return cacheSize;
+		}
+
+		public ImageCache() : base()
+		{
+			_lruCache = new global::Android.Util.LruCache(GetCacheSize());
```
jonathanpeppers: So a note here...

`LruCache` seems like it functions on the "number of entries": https://developer.android.com/reference/android/util/LruCache#LruCache(int)

_Not_ the size in bytes... That means the pool will only hold N entries of images--regardless of those images' sizes.

When I implemented something like this myself, I actually ported `LruCache` to C# and changed it to take the actual size of the `Bitmap` into account by using [Bitmap.ByteCount](https://developer.android.com/reference/android/graphics/Bitmap#getByteCount()).

Xamarin.Forms.Platform.Android/ImageCache.cs
```diff
@@ -0,0 +1,108 @@
...
+	/// <summary>
+	/// I setup the access to all the cache elements to be async because
+	/// if I didn't then it was locking up the GC and freezing the entire app
+	/// </summary>
+	class ImageCache
+	{
+
+		const string _cachePlaceHolder = "PLEASEHOLD";
+
+		global::Android.Util.LruCache _lruCache;
```
jonathanpeppers: Just some questions here...

What happens if `LruCache` evicts a `Bitmap` and it's in use? Does `Bitmap.Dispose` get called, _should_ it get called?

Related, will `Bitmap.Dispose` get called on a `Bitmap` that is still in the `LruCache` for reuse?

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -138,17 +156,48 @@ internal static Task ApplyDrawableAsync(this IShellContext shellContext, Bindabl
...
+					string cacheKey = String.Empty;
+
+					// Todo improve for other sources
+					// volley the requests better up front so that if the same request comes in it isn't requeued
+					if (initialSource is UriImageSource uri)
 					{
-						// TODO: it might be good to make sure the renderer has not been disposed
+						cacheKey = Device.PlatformServices.GetMD5Hash(uri.Uri.ToString());
+						var cache = GetCache();
+						var cacheObject = await cache.GetAsync(cacheKey);
```
jonathanpeppers: Should the use of the new cache be behind a feature flag? `LruCache_Experimental` or something?

I don't know your team's stance on when a feature flag is needed for a certain change.

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -18,6 +18,23 @@ namespace Xamarin.Forms.Platform.Android
...
+				return _lruCache;
+
+			lock(_lruCacheHandle)
+			{
+				if (_lruCache == null)
+					_lruCache = new ImageCache();
+			}
+
+			return _lruCache;
+		}
```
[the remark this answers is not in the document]
PureWeen: For some reason I keep forgetting about Lazy :-/

Xamarin.Forms.Platform.Android/ImageCache.cs
PureWeen: > What happens if LruCache evicts a Bitmap and it's in use? Does Bitmap.Dispose get called, should it get called?

I did some research on this topic and found varying implementations. In theory we never retain drawables/bitmaps so if it's evicted and an imageview is collected the bitmap (having no references) should also be collected yes? A few implementations I saw involved people using custom drawables and then adding there own refcounters to them

https://stackoverflow.com/a/22855962/953734

> Related, will Bitmap.Dispose get called on a Bitmap that is still in the LruCache for reuse?

It shouldn't. We don't really call dispose on any of the bitmaps that get created from ResourceManager. It could be slightly breaking though if a user calls dispose so we could check the handle on the stored object as well?

Xamarin.Forms.Platform.Android/ResourceManager.cs
PureWeen: If we aren't comfortable with the LruCache going into 4.0 (next weekish) then we'd most likely just retarget the LRU part of this for master so that it can percolate inside some 4.1 pre releases first and get testing.  The main issue I was trying to combat with the LruCache is that we have a disk cache for UriSources but if multiple things request the same image from the same Url it will generate multiple bitmaps over the same stream which was causing OOM exceptions. The LruCache allows for the same underlying bitmap so it greatly streamlined cases where a user might have the same url loaded 4 or 5 times.  There's also cases where the same ImageSource might be retrigger to re-request the same image which was causing a cancel then a re-request of the same image (I realize the ImageSource here could be made more resilient) but an LruCache seemed to just be a good way to Volley all requests across different ImageSources or from the same one

Xamarin.Forms.Platform.Android/ImageCache.cs
```diff
@@ -0,0 +1,108 @@
...
+				try
+				{
+					var innerCacheObject = _lruCache.Get(cacheKey);
+
+					// this only really gets hit during load when there are a lot of requests for the same image
+					while (innerCacheObject?.ToString() == _cachePlaceHolder)
+					{
+						await Task.Delay(100).ConfigureAwait(false);
+						innerCacheObject = _lruCache.Get(cacheKey);
+					}
```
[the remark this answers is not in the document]
PureWeen: Alright I modified my implementation to use some better structures around the key. Let me know what you think

Xamarin.Forms.Platform.Android/ImageCache.cs
```diff
@@ -0,0 +1,148 @@
...
+
+		public ImageCache() : base()
+		{
+			_waiting = new ConcurrentDictionary<string, SemaphoreSlim>();
+			_lruCache = new FormsLruCache();			
+		}
+
+		async Task PutAsync(string key, TimeSpan cacheValidity, global::Android.Graphics.Bitmap cacheObject)
+		{
+			await Task.Run(() =>
```
hartez: Since PutAsync is only called from GetAsync, it's already running on a thread from the thread pool; is there any reason to use Task.Run again?

Xamarin.Forms.Platform.Android/ImageCache.cs
```diff
@@ -0,0 +1,148 @@
...
+				finally
+				{
+					semaphoreSlim?.Release();
+				}
+
+				return null;
+			});
+		}
+
+		public async void Remove(string key)
```
hartez: Is Remove a CPU-heavy operation? Is there any reason it needs to be done on a separate thread?

PureWeen: I'll just delete Remove for now as I'm not using it anymore

Xamarin.Forms.Platform.Android/ImageCache.cs
PureWeen: This was a bit of an overreaction. When I did all the cache interactions synchronously it started locking up the GC so I just moved everything to tasks to see if that would unblock that issue which it did. I've changed put to now just be synchronous and everything seems to run well.

Issue4915 is a good test for this one

Xamarin.Forms.Platform.Android/ImageCache.cs
```diff
@@ -0,0 +1,120 @@
...
+
+		public ImageCache() : base()
+		{
+			_waiting = new ConcurrentDictionary<string, SemaphoreSlim>();
+			_lruCache = new FormsLruCache();			
+		}
+
+		void Put(string key, TimeSpan cacheValidity, global::Android.Graphics.Bitmap cacheObject)
+		{
+			_lruCache.Put(key, new CacheEntry() { TimeToLive = DateTimeOffset.Now.Add(cacheValidity), Data = cacheObject });
```
jonathanpeppers: Should this use `UtcNow`?

Xamarin.Forms.Platform.Android/ImageCache.cs
```diff
@@ -0,0 +1,120 @@
...
+				SemaphoreSlim semaphoreSlim = null;
+
+				try
+				{
+					semaphoreSlim = _waiting.GetOrAdd(cacheKey, (key) => new SemaphoreSlim(1, 1));
+					await semaphoreSlim.WaitAsync().ConfigureAwait(false);
+
+					var cacheEntry = _lruCache.Get(cacheKey) as CacheEntry;
+
+					if (cacheEntry?.TimeToLive < DateTimeOffset.Now)
```
jonathanpeppers: Same here? `UtcNow`?

Xamarin.Forms.Platform.Android/ImageCache.cs
```diff
@@ -0,0 +1,120 @@
...
+		}
+
+		void Put(string key, TimeSpan cacheValidity, global::Android.Graphics.Bitmap cacheObject)
+		{
+			_lruCache.Put(key, new CacheEntry() { TimeToLive = DateTimeOffset.Now.Add(cacheValidity), Data = cacheObject });
+		}
+
+		public Task<Java.Lang.Object> GetAsync(string cacheKey, TimeSpan cacheValidity, Func<Task<Java.Lang.Object>> createMethod)
+		{
+			return Task.Run(async () =>
```
jonathanpeppers: Since you `await` the `SemaphoreSlim` can you drop this `Task.Run` now?

Xamarin.Forms.Platform.Android/ImageCache.cs
PureWeen: I don't think it matters since it's a datetimeoffset and not a datetime it'll do an accurate comparison but I'll set it to utc anyway

Xamarin.Forms.Platform.Android/ImageCache.cs
hartez: What does using UtcNow vs Now give us?

Xamarin.Forms.Platform.Android/ImageCache.cs
PureWeen: My concern with that is if the winning thread (first one that loads image and puts it into the cache) holds the semaphore then it might not change threads to a Task Pool thread. Therefore, the code afterwards (mainly the Get) will run on the calling thread. Which I'd like to avoid.

Xamarin.Forms.Platform.Android/ImageCache.cs
jonathanpeppers: `Now` is going to calculate your local time -- some timezone math is involved. Also the user's timezone can change?

Here is a good explanation: https://blogs.msdn.microsoft.com/kirillosenkov/2012/01/10/datetime-utcnow-is-generally-preferable-to-datetime-now/

I would pretty much not ever use `Now` unless you are presenting a date/time to a user, and they need to see it in their local time.

hartez: That's fair. I didn't know Now was calling UtcNow and converting; TIL.

Xamarin.Forms.Platform.Android/ImageCache.cs
PureWeen: Changed

Xamarin.Forms.Platform.Android/ImageCache.cs
PureWeen: Changed to UtcNow

Xamarin.Forms.Platform.Android/ImageCache.cs
```diff
@@ -0,0 +1,142 @@
...
+		public class FormsLruCache : global::Android.Util.LruCache
+		{
+
+			static int GetCacheSize()
+			{
+				// https://developer.android.com/topic/performance/graphics/cache-bitmap
+				int cacheSize = 4 * 1024 * 1024;
+				if (Java.Lang.Runtime.GetRuntime()?.MaxMemory() != null)
+				{
+					var maxMemory = (int)(Java.Lang.Runtime.GetRuntime().MaxMemory() / 1024);
```
jonathanpeppers: This calls `Java.Lang.Runtime.GetRuntime()?.MaxMemory()` twice, could just do:
```csharp
var max = Java.Lang.Runtime.GetRuntime()?.MaxMemory();
if (max != null)
...
```

Xamarin.Forms.Platform.Android/ButtonLayoutManager.cs
```diff
@@ -85,7 +86,7 @@ public void OnLayout(bool changed, int left, int top, int right, int bottom)
 				return;
 
 			AppCompatButton view = View;
-			if (view == null || view.Layout == null)
+			if (view == null)
```
PureWeen: @mattleibow do you by chance remember why this null check is here? I found that with it there would sometimes be cases where the layout would be null and then the padding code below would never run. So Just removed it and that fixed my scenario but I'm not sure if I broke a different one

PR #6079, opened 2019-05-02
Add separate MarkerClick and InfoWindowClick events for Pins

### LINKED ISSUE #2509: Pin.Clicked does not work
### Description
Event listener is not triggered for Pin.Clicked event,
Looks like this bug is opened again,
https://bugzilla.xamarin.com/show_bug.cgi?id=25557

### Steps to Reproduce

1. var pin = new Pin()...
2. pin.Clicked += PinOnClicked;
3. Map.Pins.Add(pin)

### Expected Behavior
PinOnClicked should be triggered,

### Actual Behavior
PinOnClicked is not called. Event is not triggered on clicking pin in the map,

### Basic Information

- IDE: Visual Studio 2017 (15.6.6)
- Platform Target Frameworks:
  - Android: 8.0
- Nuget Packages:
- Affected Devices:

### LINKED ISSUE #3131: Add Selected and Unselected events to Maps.Pin object
## Summary
I'd like to have `Selected ` and `Unselected `  events on `Xamarin.Forms.Maps.Pin` class that fires when the user tap the pin.

## API Changes
Add
- `public event EventHandler Selected;` 
- `public event EventHandler Unselected;` 

to the `Xamarin.Forms.Maps.Pin.cs` file (and implement it on platforms code).

## Intended Use Case
We already have the  pin `Clicked` event,  but it's actually fired when the user, after tapping the pin, then taps the pin label. This is a double action so that the clicked event is not really related to the pin itself, but to the label of the pin.

I do understand that changing the semantic of `Clicked` and eventually adding a `LabelClicked` event would be a breaking change on existent Xamarin.Forms code, so I'm asking for the `Selected` event that can solve the issue without breaking existing code.

Moreover, AFAIK there isn't an event that's been fired when a selected pin is unselected.

This can be useful (and much needed) at least in this scenario:

1. Suppose I want to show something related to the pin in the same page where the Map is shown, whenever the user tap a pin:

  image

As you can see the user has to tap twice to get the Clicked event fired, one for selecting the pin and get the pin label shown, the other when tapping the label to fire the clicked event.

And when the user taps on another pin, or just unselect it tapping on the map, the footer message become out of sync because is still related to the latest tapped pin label.

### LINKED ISSUE #5490: Map Pin Click only works on Android on XF and Maps 3.6
### Description
Upgraded to XF 3.6 and Maps 3.6 today all mapping working fine BUT on anything put Android if click on the map pins the click event does not fire or bubble does not come up, this previously worked

### Steps to Reproduce

1. Add a map
2. Add a pin with a click event
3. Click pin, click does not fire

### Expected Behavior
It should work on ALL platforms
### Actual Behavior
Only works on Android
### Basic Information

- Version with issue: 3.6
- Last known good version:3.1 for me
- IDE:
- Platform Target Frameworks: 
  - iOS: 12.1
  - Android: 7.1
  - UWP:  17000 
- Nuget Packages:
- Affected Devices:

jcmanke:
### Description of Change ###

Replace the `Pin.Clicked` event which has caused much confusion about when it is actually fired with two separate events, `MarkerClicked` and `InfoWindowClicked`.

### Issues Resolved ### 

- fixes #2509 
- fixes #3131
- fixes #5490

### API Changes ###

Added:
 - `public class PinClickedEventArgs`
 - Pin
   - `public event EventHandler<PinClickedEventArgs> MarkerClicked;`
   - `public event EventHandler<PinClickedEventArgs> InfoWindowClicked;`
   - `public bool SendMarkerClick()`
   - `public bool SendInfoWindowClick()`
 - Android MapRenderer
   - `protected Pin GetPinForMarker(Marker marker)`
   - `void OnMarkerClick(object sender, GoogleMap.MarkerClickEventArgs e)`
 - iOS MapRenderer
   - `protected Pin GetPinForAnnotation(IMKAnnotation annotation)`
   - `void MkMapViewOnAnnotationViewSelected(object sender, MKAnnotationViewEventArgs e)`

Changed:
 - Android MapRenderer
   - `void MapOnMarkerClick(object sender, GoogleMap.InfoWindowClickEventArgs eventArgs)` --> `void OnInfoWindowClick(object sender, GoogleMap.InfoWindowClickEventArgs e)`
 - iOS MapRenderer
   - `void OnClick(object annotationObject, UITapGestureRecognizer recognizer)` --> `void OnCalloutClicked(IMKAnnotation annotation)`
 - MacOS MapRenderer
   - `void OnClick(object annotationObject, NSClickGestureRecognizer recognizer)` --> `void OnCalloutClicked(IMKAnnotation annotation)`

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

- `Pin.Clicked` and `Pin.SendTap()` marked as obsolete.

Not applicable

### Testing Procedure ###
Go to the "Map Gallery - Legacy" page in the Control Gallery. Clicking a pin will update the Last Marker Clicked label and present the info window. Clicking the info window will update the Last Info Window Clicked label and dismiss the info window.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Maps.iOS/MapRenderer.cs
+++ b/Xamarin.Forms.Maps.iOS/MapRenderer.cs
@@ -284,7 +284,13 @@ namespace Xamarin.Forms.Maps.MacOS
 
 			if (pin != null)
 			{
-				pin.SendMarkerClick();
+				// SendMarkerClick() returns the value of PinClickedEventArgs.HideInfoWindow
+				// Hide the info window by deselecting the annotation
+				bool deselect = pin.SendMarkerClick();
+				if (deselect)
+				{
+					((MKMapView)Control).DeselectAnnotation(annotation, false);
+				}
 			}
 		}
 		
@@ -305,7 +311,14 @@ namespace Xamarin.Forms.Maps.MacOS
 #pragma warning disable CS0612
 			targetPin.SendTap();
 #pragma warning restore CS0612
-			targetPin.SendInfoWindowClicked();
+
+			// SendInfoWindowedClick() returns the value of PinClickedEventArgs.HideInfoWindow
+			// Hide the info window by deselecting the annotation
+			bool deselect = targetPin.SendInfoWindowClick();
+			if (deselect)
+			{
+				((MKMapView)Control).DeselectAnnotation(annotation, true);
+			}
 		}
 
 #if __MOBILE__
--- a/Xamarin.Forms.Maps.iOS/MapRenderer.cs
+++ b/Xamarin.Forms.Maps.iOS/MapRenderer.cs
@@ -304,9 +304,9 @@ namespace Xamarin.Forms.Maps.MacOS
 			if (_lastTouchedView is MKAnnotationView)
 				return;
 
-#pragma warning disable CS0612
+#pragma warning disable CS0618
 			targetPin.SendTap();
-#pragma warning restore CS0612
+#pragma warning restore CS0618
 
 			// SendInfoWindowClick() returns the value of PinClickedEventArgs.HideInfoWindow
 			// Hide the info window by deselecting the annotation
--- a/Xamarin.Forms.Maps.iOS/MapRenderer.cs
+++ b/Xamarin.Forms.Maps.iOS/MapRenderer.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Collections.Specialized;
 using System.ComponentModel;
+using System.Linq;
 using CoreLocation;
 using MapKit;
 using ObjCRuntime;
@@ -71,6 +72,7 @@ namespace Xamarin.Forms.Maps.MacOS
 				}
 
 				var mkMapView = (MKMapView)Control;
+				mkMapView.DidSelectAnnotationView -= MkMapViewOnAnnotationViewSelected;
 				mkMapView.RegionChanged -= MkMapViewOnRegionChanged;
 				mkMapView.GetViewForAnnotation = null;
 				if (mkMapView.Delegate != null)
@@ -145,6 +147,7 @@ namespace Xamarin.Forms.Maps.MacOS
 					SetNativeControl(mapView);
 
 					mapView.GetViewForAnnotation = GetViewForAnnotation;
+					mapView.DidSelectAnnotationView += MkMapViewOnAnnotationViewSelected;
 					mapView.RegionChanged += MkMapViewOnRegionChanged;
 #if __MOBILE__
 					mapView.AddGestureRecognizer(_mapClickedGestureRecognizer = new UITapGestureRecognizer(OnMapClicked));
@@ -256,6 +259,17 @@ namespace Xamarin.Forms.Maps.MacOS
 			mapPin.AddGestureRecognizer(recognizer);
 		}
 
+		void MkMapViewOnAnnotationViewSelected(object sender, MKAnnotationViewEventArgs e)
+		{
+			var annotation = e.View.Annotation;
+			var pin = ((Map)Element).Pins.FirstOrDefault(p => p.MarkerId == annotation);
+
+			if (pin != null)
+			{
+				pin.SendMarkerClick();
+			}
+		}
+
 #if __MOBILE__
 		void OnClick(object annotationObject, UITapGestureRecognizer recognizer)
 #else
@@ -289,6 +303,7 @@ namespace Xamarin.Forms.Maps.MacOS
 				return;
 
 			targetPin.SendTap();
+			targetPin.SendInfoWindowClicked();
 		}
 
 #if __MOBILE__
--- a/Xamarin.Forms.Maps.iOS/MapRenderer.cs
+++ b/Xamarin.Forms.Maps.iOS/MapRenderer.cs
@@ -302,7 +302,9 @@ namespace Xamarin.Forms.Maps.MacOS
 			if (_lastTouchedView is MKAnnotationView)
 				return;
 
+#pragma warning disable CS0612
 			targetPin.SendTap();
+#pragma warning restore CS0612
 			targetPin.SendInfoWindowClicked();
 		}
 
--- a/Xamarin.Forms.Maps.iOS/MapRenderer.cs
+++ b/Xamarin.Forms.Maps.iOS/MapRenderer.cs
@@ -1,15 +1,11 @@
-﻿using System;
-using System.Collections;
-using System.Collections.Generic;
+﻿using System.Collections;
 using System.Collections.ObjectModel;
 using System.Collections.Specialized;
 using System.ComponentModel;
-using System.Linq;
 using CoreLocation;
 using MapKit;
 using ObjCRuntime;
 using RectangleF = CoreGraphics.CGRect;
-using Foundation;
 
 #if __MOBILE__
 using UIKit;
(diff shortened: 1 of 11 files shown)


jcmanke:
Some questions I would appreciate input on from Xamarin people:

1. Obsoleting `Pin.Clicked` - Do I also obsolete the `SendTap()` method that the renderers use to fire the event? If so, do I put a preprocessor directive around that call to ignore the warning? What do I do about the UI tests that use the event (specifically, Bugzilla29017)?
2. For finding the Pin associated with the clicked Marker, is there a preference for looping through the `Map.Pins` list vs using the LINQ `FirstOrDefault()` method? Currently I have it written one way on Android and the other on iOS and figure I should at least be consistent.

hartez:
> Some questions I would appreciate input on from Xamarin people:
> 
> 1. Obsoleting `Pin.Clicked` - Do I also obsolete the `SendTap()` method that the renderers use to fire the event? If so, do I put a preprocessor directive around that call to ignore the warning? What do I do about the UI tests that use the event (specifically, Bugzilla29017)?

Yes, the method should be marked obsolete and if necessary, a #pragma can be added to ignore the warning.

Since we still need to make sure not to break the event while it's obsolete, any tests which still rely on the event should continue to rely on it (and use a #pragma to disable the warning). 


> 2. For finding the Pin associated with the clicked Marker, is there a preference for looping through the `Map.Pins` list vs using the LINQ `FirstOrDefault()` method? Currently I have it written one way on Android and the other on iOS and figure I should at least be consistent.

A `for` loop is the preference, since `Pins` is an `IList` and has an indexer. `FirstOrDefault` will allocate an enumerator, which is a bit slower and adds memory pressure.

> 3. Custom EventArgs for determining whether to show or hide the info window. Default behavior is to show the info window when the pin is clicked, then keep the info window presented after it is clicked. However, users may not want to show the info window when they handle the `MarkerClicked` event, or may want to dismiss the info window after handling the `InfoWindowClicked` method.

If the question is "should I create a custom subclass of EventArgs to handle these options", I think the answer is "yes".

jcmanke:
>> Custom EventArgs for determining whether to show or hide the info window. Default behavior is to show the info window when the pin is clicked, then keep the info window presented after it is clicked. However, users may not want to show the info window when they handle the MarkerClicked event, or may want to dismiss the info window after handling the InfoWindowClicked method.
> 
> If the question is "should I create a custom subclass of EventArgs to handle these options", I think the answer is "yes".

To clarify, my question is more like "Should I make one custom subclass of EventArgs or two?"

In my package that extends Xamarin.Forms.Maps, I have two classes:

    public class MarkerClickedEventArgs : EventArgs
    {
        public bool ShowInfoWindow { get;  set; }

        public MarkerClickedEventArgs()
        {
            ShowInfoWindow = true;
        }
    }

    public class InfoWindowClickedEventArgs : EventArgs
    {
        public bool DismissInfoWindow { get; set;  }

        public InfoWindowClickedEventArgs()
        {
            DismissInfoWindow = true;
        }
    }

But reviewing them as I'm porting them into the base package, I realize they're basically the same except for the name of the Boolean property so maybe they should be consolidated into one.

Xamarin.Forms.Maps/Pin.cs
```diff
@@ -65,8 +65,13 @@ public object MarkerId
 			}
 		}
 
+		[Obsolete("This event is obsolete. Please use MarkerClicked and/or InfoWindowClicked instead.")]
```
hartez: The obsolete message should include the version, e.g. "This event is obsolete as of version 4.1.0".

Xamarin.Forms.Maps/Pin.cs
```diff
@@ -101,6 +106,7 @@ public override int GetHashCode()
 		}
 
 		[EditorBrowsable(EditorBrowsableState.Never)]
+		[Obsolete]
```
hartez: Needs a message for the Obsolete attribute.

Xamarin.Forms.Maps.iOS/MapRenderer.cs
```diff
@@ -4,6 +4,7 @@
 using System.Collections.ObjectModel;
 using System.Collections.Specialized;
 using System.ComponentModel;
+using System.Linq;
```
hartez: Are you using Linq in here anywhere? If not, let's clean up the `using`.

Xamarin.Forms.Maps.iOS/MapRenderer.cs
```diff
@@ -288,7 +308,17 @@ void OnClick(object annotationObject, NSClickGestureRecognizer recognizer)
 			if (_lastTouchedView is MKAnnotationView)
 				return;
 
+#pragma warning disable CS0612
 			targetPin.SendTap();
+#pragma warning restore CS0612
+
+			// SendInfoWindowedClick() returns the value of PinClickedEventArgs.HideInfoWindow
```
hartez: SendInfoWindowedClick -> SendInfoWindowClick in comment

Xamarin.Forms.Maps/Pin.cs
jcmanke: Just didn't know when I made the PR what version this would be making it into.

jcmanke:
@hartez @PureWeen Should I change the obsolete messages to say version 4.2.0 before this gets merged, or is this going to find its way into 4.1.0?

hartez:
> @hartez @PureWeen Should I change the obsolete messages to say version 4.2.0 before this gets merged, or is this going to find its way into 4.1.0?

Yeah, looks like 4.1 stable was already tagged, so the message should read 4.2.0 now. Sorry about that.

jcmanke:
@samhouts I fixed the warnings 2 weeks ago, is there anything else holding this up?

Xamarin.Forms.Maps/Pin.cs
```diff
@@ -101,6 +106,7 @@ public override int GetHashCode()
 		}
 
 		[EditorBrowsable(EditorBrowsableState.Never)]
+		[Obsolete("This method is obsolete as of 4.2.0.")]
```
samhouts:
```suggestion
		[Obsolete("This method is obsolete as of 4.3.0.")]
```

Xamarin.Forms.Maps/Pin.cs
```diff
@@ -65,8 +65,13 @@ public object MarkerId
 			}
 		}
 
+		[Obsolete("This event is obsolete as of 4.2.0. Please use MarkerClicked and/or InfoWindowClicked instead.")]
```
samhouts:
```suggestion
		[Obsolete("This event is obsolete as of 4.3.0. Please use MarkerClicked and/or InfoWindowClicked instead.")]
```

samhouts:
@jcmanke Hmm, getting a build error:

` GalleryPages\MapGallery.xaml : error : Root element is missing. [D:\a\1\s\Xamarin.Forms.Controls\Xamarin.Forms.Controls.csproj]`

jcmanke:
@samhouts Looks like I botched a merge and ended up with MapGallery.xaml as an empty file. Rebased on master and it looks better now.

jcmanke:
@samhouts Anything I can do about the failing tests?

jsuarezruiz:
Several tests are failing, but appear to be unrelated _("Frame Size Gets Corrupted When List View Is Scrolled", "Navigating Back And Forth Does Not Crash"_, etc.).

PR #6081, opened 2019-05-03
remove route params from shell

PureWeen:
### Description of Change ###
This makes routescheme, shell.route, shell.routehost internal

Currently there is no functional purpose for exposing these APIs and there is a good chance we are just going to delete them. The purpose of this PR is to simplify down how you construct the shell and how you perform internal routing.  

### API Changes ###

 Removed:
 - Shell.Route
- Shell.RouteHost
- Shell.Scheme
 

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS
- Android
- UWP

Users will get an error if they've specified these values and they will need to update their routing code

### Testing Procedure ###
All Unit tests pass so routing should still be intact

### PR Checklist ###

- [x] Has automated tests 


Xamarin.Forms.Core.UnitTests/ShellTests.cs
```diff
@@ -270,7 +265,7 @@ public void CancelNavigation()
 			shell.Items.Add(one);
 			shell.Items.Add(two);
 
-			Assume.That(shell.CurrentState.Location.ToString(), Is.EqualTo("app:///s/one/tabone/content/"));
+			Assume.That(shell.CurrentState.Location.ToString(), Is.EqualTo("//one/tabone/content/"));
```
paymicro:
```suggestion
			Assume.That(shell.CurrentState.Location.ToString(), Is.EqualTo("//one/tabone/content"));
```
also below need to fix on the same string

dlakes-hx:
Is there a time table for when documentation will be updated to reflect this change? All the official sources, ex [the main docs](https://docs.microsoft.com/en-us/dotnet/api/xamarin.forms.shell?view=xamarin-forms) seem to reflect these fields are still public

Issue #4700, opened 2018-12-11
GTK: setting background color or border color of a frame to Transparent just makes it white


knocte:
### Description
If you add a frame to any Xamarin.Forms app, and you set its BackgroundColor property to Transparent, the frame will "blend" with the original container it's in, so that you cannot distinguish where the frame starts or ends. However, using the GTK driver, the frame will show up as white.

### Steps to Reproduce

1. Create a Xamarin.Forms app.
2. Add a frame to it, add some text to the frame.
3. Add BackgroundColor=Transparent and BorderColor=Transparent to the XAML of the frame.
4. Run it in GTK.

### Expected Behavior
Frame should be indistinguishable.

### Actual Behavior
Frame's background and border color shows as white.

### Basic Information

- Version with issue: 3.4.0.1008975
- IDE: VS4Mac
- Platform Target Frameworks:
    Android: 9.1 I believe
- Android Support Library Version: package id="Xamarin.Android.Support.v4" version="27.0.2.1" targetFramework="monoandroid81"
- Nuget Packages: 3.4.0.1008975
- Affected Devices: GTK

PR #6092, opened 2019-05-05
[GTK] Add Alpha support to Frame background

mfkl:
### Description of Change ###

- Use Xamarin.Forms.Color instead of Gdk.Color for alpha support.
- Add an extension method from Gdk.Color to XF.Color.
- Call `base.OnElementChanged` at the end of the method in `FrameRenderer.cs`
- Remove set Shadow.

### Issues Resolved ### 

- fixes https://github.com/xamarin/Xamarin.Forms/issues/4700

### API Changes ###

 None

### Platforms Affected ### 

- GTK

Having a transparent Frame becomes possible.

### Testing Procedure ###

Just a basic content page with this 
```
<Frame x:Name="frame1" BackgroundColor="Blue" BorderColor="Red">
   <Frame x:Name="frame2" BackgroundColor="Transparent" BorderColor="Transparent"/>
</Frame>
```
Should be enough to see the problem.

Local repro here: https://github.com/mfkl/Gtk-Frame-Background-Bug
Start `Gtk-Frame-Background-Bug\Gtk-Frame-Background-Bug.csproj`

Repro with fix in Xamarin.Forms repo: https://github.com/mfkl/Xamarin.Forms/tree/local-repro

Without https://github.com/xamarin/Xamarin.Forms/commit/41052bcd28f8139c2b9c61f341616e4f15b04d8b, the repro with the 2 frames positioned on top of each other (one blue, one transparent) shows a black frame. It comes from this code
```
// Draw Shadow
if (_hasShadow)
{
    var color = Color.Black;
    cr.SetSourceRGBA(color.R, color.G, color.B, color.A);
    cr.Rectangle(Allocation.Left + _shadowWidth, Allocation.Top + _shadowWidth, Allocation.Width + _shadowWidth, Allocation.Height + _shadowWidth);
    cr.Fill();
}
```
As when not run, the frames display fine. 
Also https://github.com/xamarin/Xamarin.Forms/blob/42b61a87228eb4686a9bce6c1a687a6c9ef34572/Xamarin.Forms.Platform.GTK/Controls/CustomFrame.cs#L20 this would probably need changing to allow shadow support, which I guess should be taken care of in a separate PR. Hence my commit removing the call to `Control.SetShadow();` which seemed systematic.

Let me know what you think.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.GTK/Controls/CustomFrame.cs
+++ b/Xamarin.Forms.Platform.GTK/Controls/CustomFrame.cs
@@ -6,10 +6,10 @@ namespace Xamarin.Forms.Platform.GTK.Controls
 {
 	public class CustomFrame : Gtk.Frame
 	{
-		private Gdk.Color _defaultBorderColor;
-		private Gdk.Color _defaultBackgroundColor;
-		private Gdk.Color? _borderColor;
-		private Gdk.Color? _backgroundColor;
+		private Color _defaultBorderColor;
+		private Color _defaultBackgroundColor;
+		private Color? _borderColor;
+		private Color? _backgroundColor;
 
 		private uint _borderWidth;
 		private bool _hasShadow;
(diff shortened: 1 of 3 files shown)


rmarinho:
azp run

PR #6101, opened 2019-05-06
[Xaml] better exception message in case of a TypeConversion issue

StephaneDelcroix:
### Description of Change ###

While wrapping an exception into a XPE in case of a TypeConversion issue, concat the message of the wrapped exception


paymicro:
Is it really necessary? This message is available in the inner exception

StephaneDelcroix:
@paymicro it is. some tool choose to only show the message, and in this case the information is not enough to figure out what's wrong.

I agree that when you're debugging, you can drill down to the InnerException ad see the details, but that's to the use case we try to solve here

Issue #5642, opened 2019-03-20
[Material,Android] Button Overlay Issue


thudugala:
### Description
When using Xamarin.Forms Material.

If controls are overlay on top of Button there are not visible.

### Steps to Reproduce

1. Run the attach app

### Expected Behavior

See the Left and Right Label on top of Button

### Actual Behavior

Cannot see the Left and Right Label on top of Button

### Basic Information

- Version with issue: 3.6.0.264807
- Last known good version: It works when you set Visual="Material" to Visual="Default"
- Platform Target Frameworks: 
  - Android: 
- Nuget Packages:  https://www.nuget.org/packages/Xamarin.Forms.Visual.Material/

### Screenshots

Visual="Material"
![image](https://user-images.githubusercontent.com/4112014/54718394-46ca2380-4bbf-11e9-95c3-a8091a97cb16.png)

Visual="Default"
![image](https://user-images.githubusercontent.com/4112014/54718295-108ca400-4bbf-11e9-95b4-219deff6e7f1.png)

### Reproduction Link

[ButtonOverlayIssue.zip](https://github.com/xamarin/Xamarin.Forms/files/2990085/ButtonOverlayIssue.zip)

PureWeen:
Hmm this is an interesting one. 

https://stackoverflow.com/questions/28105551/wrong-render-order-for-button-when-material-theme-is-applied

I tested setting stateListAnimator to null and that fixes the z indexes but I need to research a bit more about the best way to fix this.

neville-nazerane:
This seems important as overlaying Material buttons can be used to overlay images and create buttons with icons in them. There doesn't have a good alternative as of now. Other solutions such as using a stacklayout require relatively a lot of steps and are still not the same as a button.

PureWeen:
@neville-nazerane one fix that should work would be to use the android platform specific and set the elevation on the labels to something high. That should cause the labels to rise above

https://docs.microsoft.com/en-us/dotnet/api/xamarin.forms.platformconfiguration.androidspecific.visualelement.setelevation?view=xamarin-forms

PureWeen:
As a note this one is partially fixed by this PR

https://github.com/xamarin/Xamarin.Forms/pull/6271

the main side effect that exists is once you click on the button then the labels disappear under the button. The material renderer has a dynamic elevation so the elevation increases when it's clicked

PR #6108, opened 2019-05-06
stack the elevation of elements

### LINKED ISSUE #2989: [Android] Fast Renderers are not visually dealing with moving views around in the visual stack correctly
### Description
Adding views after a view has appeared or using Raise/Lower child doesn't work properly with Android Fast Renderers

### Steps to Reproduce

1. In the Control Gallery uncomment 
*Issue2653.InsertThenAddSetsZIndex*
*Issue2653.MoveUpAndMoveDown*

2. Enable fast renderers
3. I was testing with API 23 

### Expected Behavior
Tests should pass

### Actual Behavior
Tests fail

### Basic Information

- Version with issue: 3.0
- Last known good version:  I tested back to 2.5.1 and it was still having the same issues 
- Platform Target Frameworks:  
  - Android:  6.0

### LINKED ISSUE #3543: [Android] Can't Overlap a Button with any view
### Description
It's for now impossible to overlap a button on Android with another view.

### Steps to Reproduce

1. Add a button on a layout (AbsoluteLayout, Grid,.. "kind of layout that allows overlap")
2. Then add another view (ActivityIndicator, Image, Notification Badges, text, ..) on the layout
3. The Button always in front

### Expected Behavior
The expected hierarchy of the visual stack would be:
- Button would be on the back of the visual stack.
- The other view would be on the front of the visual stack.

### Actual Behavior
The button is stuck on the front of the visual stack.

### Basic Information
- Xamarin.Forms: 3.1.0.697729
- Android : 8.1

PureWeen:
### Description of Change ###
If you have an elevation set on a control then that is going to trump a call to BringToFront. With the fastrenderers we unwrapped Button from ViewGroup so now it has a default elevation set. This means that even if you call BringToFront on an element that overlaps button the button will still be visually higher because it has a default elevation. 

This PR attempts to correlate the visual hierarchy to the elevation values set for controls. If a user sets an elevation themselves then that elevation is honored

### Issues Resolved ### 
- fixes #2989
- fixes #3543

### Platforms Affected ### 
- Android

Controls stacked with buttons will now get their elevation set to the same values as button

### Testing Procedure ###
This PR enables UI Tests that were previously disabled

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/VisualElementPackager.cs
+++ b/Xamarin.Forms.Platform.Android/VisualElementPackager.cs
@@ -144,6 +144,7 @@ namespace Xamarin.Forms.Platform.Android
 		}
 		void EnsureChildOrder()
 		{
+			float elevationToSet = 0;
 			for (var i = 0; i < ElementController.LogicalChildren.Count; i++)
 			{
 				Element child = ElementController.LogicalChildren[i];
@@ -152,7 +153,21 @@ namespace Xamarin.Forms.Platform.Android
 				{
 					IVisualElementRenderer r = Platform.GetRenderer(element);
 					if (r != null)
+					{
 						(_renderer.View as ViewGroup)?.BringChildToFront(r.View);
+
+						if (Forms.IsLollipopOrNewer)
+						{
+							var elevation = ElevationHelper.GetElevation(r.View) ?? 0;
+							var elementElevation = ElevationHelper.GetElevation(element);
+
+							if (elevation > elevationToSet)
+								elevationToSet = elevation;
+
+							if(elementElevation == null)
+								r.View.Elevation = elevationToSet;
+						}
+					}
 				}
 			}
 		}
--- a/Xamarin.Forms.Platform.Android/VisualElementPackager.cs
+++ b/Xamarin.Forms.Platform.Android/VisualElementPackager.cs
@@ -5,7 +5,6 @@ using Android.Content;
 using Xamarin.Forms.Internals;
 using Android.Views;
 using AView = Android.Views.View;
-using System.Linq;
 
 namespace Xamarin.Forms.Platform.Android
 {
@@ -154,8 +153,6 @@ namespace Xamarin.Forms.Platform.Android
 					IVisualElementRenderer r = Platform.GetRenderer(element);
 					if (r != null)
 					{
-						(_renderer.View as ViewGroup)?.BringChildToFront(r.View);
-
 						if (Forms.IsLollipopOrNewer)
 						{
 							var elevation = ElevationHelper.GetElevation(r.View) ?? 0;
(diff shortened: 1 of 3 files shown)


samhouts:
@PureWeen Can you explain via screenshots how this might be breaking? Thanks!

PureWeen:
@samhouts I think this one is more breaking without then with. Without this change any layout that has a button now the button is going to be the dominate element along the Z axis. Basically RaiseChild and order of specification is broken with fast renderers when a button is involved.

Given the following layout
```XAML
    <AbsoluteLayout Padding="0" BackgroundColor="Transparent"
                    MinimumHeightRequest="50" VerticalOptions="Center">
        <!--<StackLayout Padding="0" Margin="0"
                     AbsoluteLayout.LayoutFlags="All"
                     AbsoluteLayout.LayoutBounds="0,0,1,1">
            <Button BackgroundColor="DarkBlue" Text="Button" />
        </StackLayout>-->

        <Button AbsoluteLayout.LayoutFlags="All"
                AbsoluteLayout.LayoutBounds="0,0,1,1"
                Text="Button"
                BackgroundColor="DarkBlue"  />

        <Label AbsoluteLayout.LayoutFlags="PositionProportional"
               AbsoluteLayout.LayoutBounds="0,0.5,AutoSize,AutoSize"
               Text="LEFT"
               TextColor="Red"
               BackgroundColor="Transparent"
               Margin="10,0" />

        <Label AbsoluteLayout.LayoutFlags="PositionProportional"
               AbsoluteLayout.LayoutBounds="1,0.5,AutoSize,AutoSize"
               Text="RIGHT"
               TextColor="Red"
               BackgroundColor="Transparent"
               Margin="10,0" />
    </AbsoluteLayout>
```

### With this change and Legacy renderers we get this
image



### without this change we get this
image


With this change it's going to cause the elevation of any elements specified after a button to inherit the same elevation as the button. I don't think this will cause any breaking for out of the box.

samhouts:
This should also fix #3543, right?

Xamarin.Forms.Platform.Android/VisualElementPackager.cs
```diff
@@ -152,7 +153,23 @@ void EnsureChildOrder()
 				{
 					IVisualElementRenderer r = Platform.GetRenderer(element);
 					if (r != null)
+					{
 						(_renderer.View as ViewGroup)?.BringChildToFront(r.View);
+
+						if (Forms.IsLollipopOrNewer)
```
samhouts: Is this only broken in Lollipop or newer, too? Or will we still have this problem in 19 & 20?

samhouts:
The test for 2653 is still failing on Android, btw.

Xamarin.Forms.Platform.Android/VisualElementPackager.cs
PureWeen: setElevation was only added in API 21 so pre 21 it's all just ignored
I ran those ui tests locally on API 19 and they all passed..

PureWeen:
This one doesn't fix
https://github.com/xamarin/Xamarin.Forms/issues/5642

the material button uses a dynamic elevation which doesn't really play well with this solution, so we'll have to get a little more creative for that one

PR #6149, opened 2019-05-09
Opt-in init to speed up startup times

kingces95:


rmarinho:
Can we have a better description on the PR @kingces95  ?

Xamarin.Forms.Core/Registrar.cs
```diff
@@ -265,21 +266,21 @@ public static void RegisterRenderers(HandlerAttribute[] attributes)
...
 		public static void RegisterStylesheets()
 		{
-			var assembly = typeof(StyleSheets.StylePropertyAttribute).GetTypeInfo().Assembly;
+			var assembly = typeof(StylePropertyAttribute).GetTypeInfo().Assembly;
 
 #if NETSTANDARD2_0
-			object[] styleAttributes = assembly.GetCustomAttributes(typeof(StyleSheets.StylePropertyAttribute), true);
+			object[] styleAttributes = assembly.GetCustomAttributes(typeof(StylePropertyAttribute), true);
 #else
 			object[] styleAttributes = assembly.GetCustomAttributes(typeof(StyleSheets.StylePropertyAttribute)).ToArray();
```
jsuarezruiz: A small detail, after adding the namespace, `StyleSheets` is no longer necessary here.

`object[] styleAttributes = assembly.GetCustomAttributes(typeof(StylePropertyAttribute)).ToArray();`

Xamarin.Forms.Core/Registrar.cs
```diff
@@ -340,13 +341,9 @@ public static void RegisterAll(Type[] attrTypes, ActivationFlags flags)
...
 
-					var length = attributes.Length;
-					for (var i = 0; i < length; i++)
-					{
-						var attribute = (HandlerAttribute)attributes[i];
-						if (attribute.ShouldRegister())
-							Registered.Register(attribute.HandlerType, attribute.TargetType, attribute.SupportedVisuals);
-					}
+					var handlerAttributes = new HandlerAttribute[attributes.Length];
+					Array.Copy(attributes, handlerAttributes, attributes.Length);
```
jsuarezruiz: :thumbsup:

kingces95:
This change will not result in dramatic performance improvements. Instead, it was used to demonstrate the feasibility of building our profiler and using that profiler to identify a performance issue and to verify the fix.

I plan to rename some of the items published in this PR (if I can get away with that) when I push the other performance fixes that, taken together, will produce a startup performance improvement worth promoting. 

Those fixes are been developed and are now going through our PR process.

PR #6151, opened 2019-05-10
Disposed objects are not handled properly in android scrollview renderer #5820

### LINKED ISSUE #5820: Disposed objects are not handled properly in android scrollview renderer.
## Summary

Disposed objects are not handled properly in android scrollview renderer.

## Intended Use Case
We have used custom listview (SfListView). We have select and scrolling the listview item asynchronously and while navigating back from the page it will leads to crash **ObjectDisposedException**.

This issue is due to you have used delay timer in the OnScrollToRequested method.  When we leave the page while scrolling the listview disposed object exception occurs. 

We have fixed the issue by handling the disposed boolean property in the OnScrollToRequested  method. Please find the code snippet below.

     private async void OnScrollToRequested(object sender, ScrollToRequestedEventArgs e)
        {
            // 99.99% of the time simply queuing to the end of the execution queue should handle this case.
            // However it is possible to end a layout cycle and STILL be layout requested. We want to
            // back off until all are done, even if they trigger layout storms over and over. So we back off
            // for 10ms tops then move on.
            var cycle = 0;
            while (IsLayoutRequested)
            {
                await Task.Delay(TimeSpan.FromMilliseconds(1));

                if (_disposed)
                    return;
                cycle++;
                if (cycle >= 10)
                    break;
            }
        }

Can you kindly merge the above fix in your source to avoid this type of exception in future.

syncsiva:
### Description of Change ###

Object disposed exception raised when navigation back from page with SfListView in random cases.

### Issues Resolved ### 
Exception raised in random cases due to the delay we are making in OnScrollRequesed method. Now after the delay code added and returned based on _disposed flag status.

fixes #5820

### API Changes ###
None

### Platforms Affected ### 
- Android

None

Not applicable

### Testing Procedure ###
NA. 

### PR Checklist ###

- [ ] Has automated tests 


rmarinho:
Hi @syncsiva can you provide us as sample so we can check ? 

thanks

syncsiva:
@rmarinho we were able to replicate crash only in customer application shared where we could not able to share here. We could not re-create the issue in simple cases. So, requesting to review based on below report.
https://github.com/xamarin/Xamarin.Forms/issues/5820

PureWeen:
@syncsiva can you rebase this to master? Some UI tests are failing and I just want to run against latest master

syncsiva:
@PureWeen  re-base done.

Xamarin.Forms.Platform.Android/Renderers/ScrollViewRenderer.cs
```diff
@@ -373,6 +373,10 @@ async void OnScrollToRequested(object sender, ScrollToRequestedEventArgs e)
 			while (IsLayoutRequested)
 			{
 				await Task.Delay(TimeSpan.FromMilliseconds(1));
+				
+				if (_disposed)
```
paymicro: I think it's better to check out native `IsDisposed` here. Since NRE can occur on call `IsLayoutRequested` method.

kvpt:
This fix is wrong.
Like dispose methods from other renderer, the problem is that events aren't unsubscribed.
So the OnScrollToRequested is called on the old instance after been disposed.
I have done a pass on AppCompat and FastRenderer in #6467 will do another pass on legacy renderer.

kvpt:
After all, the fix is right.
The while loop need to be exited when the object is disposed in all case.
But I think that the animator.Update can cause a leak, but this not the scope of this PR.

DineshBabuYadav:
Hi @samhouts ,

Most of our customer are using **Xamarin Forms** version from **3.5.0**, so, we request to merge the same patch in the lower versions too. It will be helpful for Xamarin developers who sync up with Syncfusion.

samhouts:
@DineshBabuYadav We're not currently planning to service 3.5, but we have ported this change to 3.6 in the event we are able to release a service release for 3.6. Thanks!

PR #6153, opened 2019-05-10
Allow third party libs to display animated images correctly in view recycling scenarios

daniel-luberda:
### Description of Change ###

Background: Third party libs can now be used with `Image` view (after registering `IImageSourceHandler`)

This change won't affect standard image loading. It'll allow third party libs like FFImageLoading to display animated images correctly in view recycling scenarios. Without this, after scrolling, animated images are not shown at all (layout invalidation is not initiated).

FFImageLoading also uses this workaround: https://github.com/luberda-molinet/FFImageLoading/blob/90ed8eef65ed1c9a57a4e4ce477a98c70548de5c/source/FFImageLoading.Touch/Targets/UIImageViewTarget.cs#L50-L51

### API Changes ###

None

### Platforms Affected ### 

- iOS

#### MacOS

BTW: I also tried to apply similar fix for MacOS but `ImageRenderer` doesn't use `NSImageView` at all (which supports animations). What's the reason of using `NSView` instead of using `NSImageView`?


rmarinho:
@daniel-luberda  thanks, do you think you can provide an example , maybe a test project with ffimageloading we can update with this package and test? 

Thanks

daniel-luberda:
@rmarinho It's here: https://github.com/daniel-luberda/XF-PR6153

rmarinho:
Failing ios test nto related, tested the project provided with fix and works.

daniel-luberda:
Thanks @rmarinho. Do you know what's the reason of using `NSView` instead of using `NSImageView` on MacOS? 
v3.x branch uses `NSImageView`. Why is it implemented differently in 4.0? I wanted to push a PR with the same issue on MacOS, but I don't know what to do.

Issue #5682, opened 2019-03-25
WPF: ViewCell content width does not update on ListView width change


domyd:
### Description
If the `ListView` width changes on WPF (e.g. on resizing the window), the `ViewCell`'s content is not resized to match the new width. The `ViewCell` itself seems to resize just fine (based on the hover effect), however the content does not.

### Steps to Reproduce
1. Create a new Xamarin WPF application
2. Add a `ListView` as a root of a `Page` and give it some strings to display
3. Set an `ItemTemplate` like such:
    ```xml
    <DataTemplate>
        <ViewCell>
            <Grid>
                <Grid.ColumnDefinitions>
                    <ColumnDefinition Width="Auto" />
                    <ColumnDefinition Width="*" />
                    <ColumnDefinition Width="Auto" />
                </Grid.ColumnDefinitions>
    
                <Label Grid.Column="0" Text="X" />
    
                <Label
                    Grid.Column="1"
                    HorizontalOptions="Center"
                    Text="{Binding}" />
    
                <Label Grid.Column="2" Text="X" />
            </Grid>
        </ViewCell>
    </DataTemplate>
    ```
4. Observe as the Xs (representing the boundaries of the content view) don't behave as expected.

### Expected Behavior
The `ViewCell`'s content should resize whenever the `ListView`'s width changes, such that the Xs stay near the window borders.

### Actual Behavior
The `ViewCell`'s content does not resize, resulting in the content staying the same size and either being clipped if the window shrinks, or staying centered if the window widens.

### Basic Information

- Version with issue: Latest stable (3.6.0) as well as latest prerelease (4.0.0.250467-pre7)
- IDE: Visual Studio 15.9.9
- Platform Target Frameworks: 
  - WPF:  .NET Framework 4.7.2

### Screenshots

Starting point:
![image](https://user-images.githubusercontent.com/5590413/54939425-dd347700-4f28-11e9-9080-2c21a97e12d3.png)

Widened window:
![image](https://user-images.githubusercontent.com/5590413/54939456-ef161a00-4f28-11e9-8bf4-b35b910c3dc6.png)

Shrunk window:
![image](https://user-images.githubusercontent.com/5590413/54939469-f89f8200-4f28-11e9-975e-ae91e56b67dd.png)

### Reproduction Link
[ViewCellResize.WPF.zip](https://github.com/xamarin/Xamarin.Forms/files/3004586/ViewCellResize.WPF.zip)

[...]

PR #6179, opened 2019-05-12
Fix for layouts in a cell to expand on resize (WPF) Issue #5682

jaytilly:
### Description of Change ###
The child layouts where not receiving any notification to re-layout.  This fix calls layout on the children view items.

### Issues Resolved ### 
Grid will now auto expand within a ViewCell

### PR Checklist ###
- [] Has automated tests (Visual Tested)
- [x] Rebased on top of the target branch at time of PR (Not needed)

### DIFF
--- a/Xamarin.Forms.Platform.WPF/CellControl.cs
+++ b/Xamarin.Forms.Platform.WPF/CellControl.cs
@@ -93,11 +93,11 @@ namespace Xamarin.Forms.Platform.WPF
 				{
 					foreach (var child in vc.LogicalChildren)
 					{
-						if (child is Grid childGrid)
+						if (child is Layout<View> layout)
 						{
-							if (childGrid.HorizontalOptions.Expands)
+							if (layout.HorizontalOptions.Expands)
 							{
-								childGrid.Layout(new Rectangle(childGrid.X, childGrid.Y, sizeInfo.NewSize.Width, sizeInfo.NewSize.Height));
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.WPF/CellControl.cs
```diff
@@ -85,6 +85,27 @@ void SetSource(object oldCellObj, object newCellObj)
...
+			if (Content is ViewCell vc)
+			{
+				if (vc.LogicalChildren != null && vc.LogicalChildren.Any())
+				{
+					foreach (var child in vc.LogicalChildren)
+					{
+						if (child is Layout<View> layout)
+						{
+							if (layout.HorizontalOptions.Expands)
+							{
```
rmarinho: Why only if this ? we should layout always  no ?

jaytilly: Probably - I should have seen that.  I can fix and test this evening. 

Thanks for the review

rmarinho: i still don't get why we just call layout if `layout.HorizontalOptions.Expands` ?

jaytilly: Good question.. my assumptions may not be valid. Should the child layout of a cell always expand to fit the cell?

VladislavAntonyuk: We have to always layout content. I have fixed it in my PR.

PR #6180, opened 2019-05-13
Virtualize Lifecycle Events

### LINKED ISSUE #6159: [Enhancement] Make Send methods virtual for application lifecycle
## Summary

Right now there are hidden methods in the Application class for invoking lifecycle events (i.e. Start, Resume, Sleep). Frameworks may need to tie into these events which means overriding the On{Event} method. While this works it also means that any downstream users have to explicitly know to call the base.On{Event}.

## API Changes

Current API

```cs
public void SendResume()
{
    Current = this;
    OnResume();
}

public void SendSleep()
{
    OnSleep();
    SavePropertiesAsFireAndForget();
}

public Task SendSleepAsync()
{
    OnSleep();
    return SavePropertiesAsync();
}

public void SendStart()
{
    OnStart();
}
```

Proposed

```cs
public virtual void SendResume()
{
    Current = this;
    OnResume();
}

public virtual void SendSleep()
{
    OnSleep();
    SavePropertiesAsFireAndForget();
}

public virtual Task SendSleepAsync()
{
    OnSleep();
    return SavePropertiesAsync();
}

public virtual void SendStart()
{
    OnStart();
}
```

## Intended Use Case

Allow frameworks such as Prism that provide a custom base Application class to inject their lifecycle events without requiring users to have to know to call base.On{Event}, while continuing to hide the API's from intellisense so they aren't misused.

dansiegel:
### Description of Change ###

Virtualizes lifecycle events to allow 3rd party frameworks like Prism to allow custom logic for Startup/Resume/Sleep without requiring downstream developers to know to call base.On{Event}().

### Issues Resolved ### 

- fixes #6159

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

- Application Lifecycle (Start/Resume/Sleep) Event invokers are now virtual so that 3rd parties with custom base Applications can inject additional logic without downstream users needing to know to call `base.On{Event}()`

Not applicable

### Testing Procedure ###

n/a

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Application.cs
+++ b/Xamarin.Forms.Core/Application.cs
@@ -285,28 +285,28 @@ namespace Xamarin.Forms
 		}
 
 		[EditorBrowsable(EditorBrowsableState.Never)]
-		public void SendResume()
+		public virtual void SendResume()
 		{
 			s_current = this;
 			OnResume();
 		}
 
 		[EditorBrowsable(EditorBrowsableState.Never)]
-		public void SendSleep()
+		public virtual void SendSleep()
 		{
 			OnSleep();
 			SavePropertiesAsFireAndForget();
 		}
 
 		[EditorBrowsable(EditorBrowsableState.Never)]
-		public Task SendSleepAsync()
+		public virtual Task SendSleepAsync()
 		{
 			OnSleep();
 			return SavePropertiesAsync();
 		}
 
 		[EditorBrowsable(EditorBrowsableState.Never)]
-		public void SendStart()
+		public virtual void SendStart()
 		{
 			OnStart();
 		}


rmarinho:
I would create new methods , then refactor our stuff to make sure those new methods are called.

dansiegel:
> I would create new methods , then refactor our stuff to make sure those new methods are called.

@rmarinho the idea is to give frameworks a place to sneak in a little code safely in an API 99.999% of people shouldn't even realize exists not unnecessarily duplicate one that already exists. If you have something in mind I'm certainly happy to discuss in more detail what you're thinking.

samhouts:
@dansiegel We want to get these event hooks to you. I think Rui and Stephane are correct here; this is not the right approach for our framework. However, we're already looking at adding more events as part of #2210, and we would be more than happy to add events that meet these needs, too. Thanks!!

Issue #1647, opened 2018-01-24
Support binding to attached BP


StephaneDelcroix:
The current Binding system doesn't allow to use an attached bindable property as binding source, like in 
```
<Grid>
    <Label x:Name="label0" Grid.Row="1"/>
    <Label Text="{Binding (Grid.Row), Source={x:Reference label0}}" />
</Grid>
```

Of course, it can be combined as part of a complex path 
```
<Label Text="{Binding label0.(Grid.Row)}" />
```

If the attached BP owner type is not in the default xmlns, the type has to be prefixed
```
<Label Text="{Binding label0.(local:MyLayout.Altitude)}" />
```
For this reason, this syntax is restricted to xaml only.

Challenges:
 - attach changed handlers to attached BPs
 - keep the current bindings working while refactoring
 
## Formal syntax for Binding Path
_part-element-list_:
 - _empty_
 - _part-separator_
 - _first-part_ _part-list_(opt)

_empty_:
 - ` `

_part-separator_:
 - `.`

_first-part_:
 - _part_
 - _part-separator_ _part_

_part-list_:
 - _part-separator_ _part_
 - _part-separator_ _part_ _part-list_(opt)

_part_:
 - _part-id_
 - _indexer_

_part-id_:
 - identifier
 - `(`_type_`.`_bp_`)`

_type_:
 - identifier
 - xmlnsprefix`:`identifier

_indexer_:
 - _part_id_`[`index`]`

AtosNicoS:
This issue is listed for the x:Bind project:
https://github.com/xamarin/Xamarin.Forms/projects/7

What is the current status of the x:bind implementation?

ddobrev:
When is this planned for fixing?

groege:
Bump

YZahringer:
Any news on that?

tuyen-vuduc:
I am also looking for this feature to work as well.

JunkyXL86:
Same here

acaliaro:
there are some news there?

SebastiaanPolfliet:
Any updates on this? Thanks!

marcos-marimon:
Any update? And support for TemplateBinding as well?

jfversluis:
This won't be happening for Forms anymore. I've added it to the short list for .NET MAUI!

PR #6217, opened 2019-05-17
Badge support for Shell

### LINKED ISSUE #6112: [Enhancement] Badge support for Shell
## Summary

There are requests for badge support for classic [tabs](https://github.com/xamarin/Xamarin.Forms/issues/3352) and there are also some libraries out there which kinda support that. With shell it's not possible to use this libraries without touching / replacing too much of the shell internals so it would be great if shell supports this scenario out of the box. I couldn't find anything related in [shell spec](https://github.com/xamarin/Xamarin.Forms/issues/2415), that's why I'm proposing it here.

## API Changes

I think it makes sense to add badge support to navigation elements which are shell sections (bottom tabs) and shell contents (top tabs). Don't know if this makes sense for the flyout menu too.

The most basic properties for badge as already stated in other issues are

- `BadgeText`: string to set the number of notification-like items
- `BadgeTextColor`: Foreground color of each badge text
- `BadgeColor`: background color of each badge

There might also be other things of interest like showing only a little red dot etc.

This properties should be bindable to ShellSection and ShellContent

```xml
<ShellSection BadgeText="{Binding BadgeText}" BadgeTextColor="{Binding BadgeTextColor}" BadgeColor="{Binding BadgeColor}" />
```

`BadgeColor` and `BadgeTextColor` might also be settable at shell level, so that you don't have to specify it for every shell section / shell content item.

### Android

`ShellItemRenderer` uses `BottomNavigationView` and `ShellSectionRenderer` uses `TabLayout`.

Material components will support [badges](https://github.com/material-components/material-components-android/blob/master/docs/components/BadgeDrawable.md) for `BottomNavigationView` and `TabLayout`, which when I understood it correctly can be used when [AndroidX](https://github.com/xamarin/AndroidSupportComponents/tree/AndroidX#android-support---androidx-roadmap) is supported by Xamarin.

In the meantime you could modify the `BottomNavigationItemView` by hand and add a [simple layout](https://stackoverflow.com/a/46591378/1249506) for badges. I guess the same is possible with SetCustomView for the `Tab`s of a `TabLayout`.

### iOS

`ShellItemRenderer` uses `UITabBarController` which can use the `UITabBarItem` to specify native Badge properties.

`ShellSectionRootRenderer` uses a UICollectionView, so you would have to modify `ShellSectionHeaderCell` and add a custom badge view (something like [this](http://ootips.org/yonat/the-simplest-ios-badge/)?).

## Intended Use Case

We are using badges primarily with bottom tabs to notify that there are new items available for the underlying page.

I might have some time to work on this if wanted. Thoughts?

Dresel:
### Description of Change ###

As requested draft / POC / work in progress PR of [badge support for Shell](https://github.com/xamarin/Xamarin.Forms/issues/6112).

### Issues Resolved ### 

- fixes #6112

### API Changes ###

Added:

`BaseShellItem (bindable properties)`

- `BadgeText`: string to set the number (or something else) of notification-like items
- `BadgeMoreText`: string to set the number (or something else) of the aggregated more tab
- `BadgeTextColor`: Foreground color of badge text
- `BadgeUnselectedTextColor`: Foreground color of badge text if underlying navigation item is in unselected state
- `BadgeEffectiveTextColor (readonly)`: If `BadgeUnselectedTextColor` is unset, `BadgeTextColor` will be returned. Otherwise it will return `BadgeTextColor` if the underlying navigation item is in selected state or `BadgeUnselectedTextColor` if in unselected state.
- `BadgeColor`: background color of badge
- `BadgeUnselectedColor`: background color of badge if navigation item is in unselected state
- `BadgeEffectiveColor (readonly)`: If `BadgeUnselectedColor` is unset, `BadgeColor` will be returned. Otherwise it will return `BadgeColor` if the underlying navigation item is in selected state or `BadgeUnselectedColor` if in unselected state.

### Platforms Affected ### 

- Core (all platforms)
- iOS
- Android

This PR is a draft / work in progress implementation of [badge support](https://github.com/xamarin/Xamarin.Forms/issues/6112) and should be the base for further discussions.

Please check control gallery (store shell) for demonstration.

### Status ###

|            | Android | iOS     |
|------------|:-:|:-:|
| Bottom Tab | &#9745; | &#9745; |
| More Tab Handling (basic) | &#9745; | &#9745; |
| Top Tab    | &#9745; | &#9745; |
| Flyout | &#9745; | &#9745; |
| Basic properties | &#9745; | &#9745; |
| Unselected properties | &#9745; | &#9745; |

### Screenshots ###

**Android**

**iOS**

### PR Checklist ###

- [x] Has automated tests 
- [x] Changes adhere to coding standard (I hope so)
- [x] Feature complete

### DIFF
--- a/Xamarin.Forms.Core/Shell/BaseShellItem.cs
+++ b/Xamarin.Forms.Core/Shell/BaseShellItem.cs
@@ -4,7 +4,9 @@ using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using Xamarin.Forms.Internals;
 using System.ComponentModel;
+using System.Globalization;
 using System.Linq;
+using Xamarin.Forms.Core;
 using Xamarin.Forms.StyleSheets;
 
 namespace Xamarin.Forms
@@ -96,6 +98,24 @@ namespace Xamarin.Forms
 		{
 			get { return Routing.GetRoute(this); }
 			set { Routing.SetRoute(this, value); }
+		}
+
+		public string BadgeText
+		{
+			get { return Badge.GetBadgeText(this); }
+			set { Badge.SetBadgeText(this, value); }
+		}
+
+		public Brush BadgeBackground
+		{
+			get { return Badge.GetBadgeBackground(this); }
+			set { Badge.SetBadgeBackground(this, value); }
+		}
+
+		public Color BadgeTextColor
+		{
+			get { return Badge.GetBadgeTextColor(this); }
+			set { Badge.SetBadgeTextColor(this, value); }
 		}
 
 		public string Title
@@ -432,6 +452,25 @@ namespace Xamarin.Forms
 					defaultLabelClass.Setters.Add(new Setter { Property = Label.HorizontalTextAlignmentProperty, Value = TextAlignment.Start });
 				}
 
+				var badgeLabel = new Label();
+				badgeLabel.SetBinding(Label.TextProperty, nameof(BaseShellItem.BadgeText));
+
+				Frame badgeFrame = new Frame();
+				badgeFrame.SetBinding(Frame.IsVisibleProperty, nameof(BaseShellItem.BadgeText), converter: new IsNotNullOrEmptyConverter());
+				badgeFrame.Margin = new Thickness(0, 0, 8, 0);
+				badgeFrame.SetBinding(Frame.BackgroundProperty, nameof(BaseShellItem.BadgeBackground), converter: new UseFallbackBrushIfNullOrEmptyBrushConverter(new SolidColorBrush(Color.FromRgb(255, 59, 48))));
+				badgeFrame.CornerRadius = 10;
+				badgeFrame.HasShadow = false;
+				badgeFrame.Padding = new Thickness(6, 2);
+				badgeFrame.Content = badgeLabel;
+				badgeFrame.VerticalOptions = LayoutOptions.Center;
+				badgeFrame.HorizontalOptions = LayoutOptions.End;
+
+				grid.Children.Add(badgeFrame, 2, 0);
+
+				badgeLabel.FontSize = 10;
+				badgeLabel.SetBinding(Label.TextColorProperty, nameof(BaseShellItem.BadgeTextColor), converter: new UseFallbackColorIfDefaultColorConverter(Color.White));
+
 				INameScope nameScope = new NameScope();
 				NameScope.SetNameScope(grid, nameScope);
 				nameScope.RegisterName("FlyoutItemLayout", grid);
@@ -443,6 +482,48 @@ namespace Xamarin.Forms
 				return grid;
 			});
 		}
+	}
+
+	public class IsNotNullOrEmptyConverter : IValueConverter
+	{
+		public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+		{
+			return !string.IsNullOrEmpty((string)value);
+		}
+
+		public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture) => throw new NotImplementedException();
+	}
+
+	public class UseFallbackBrushIfNullOrEmptyBrushConverter : IValueConverter
+	{
+		public Brush FallbackBrush { get; }
+
+		public UseFallbackBrushIfNullOrEmptyBrushConverter(Brush fallbackBrush)
+		{
+			FallbackBrush = fallbackBrush;
+		}
+
+		public object Convert(object value, Type targetType, object parameter, CultureInfo culture) =>
+			Brush.IsNullOrEmpty((Brush)value) ? FallbackBrush : (Brush)value;
+
+		public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+			=> throw new NotImplementedException();
+	}
+
+	public class UseFallbackColorIfDefaultColorConverter : IValueConverter
+	{
+		public Color FallbackColor { get; }
+
+		public UseFallbackColorIfDefaultColorConverter(Color fallbackColor)
+		{
+			FallbackColor = fallbackColor;
+		}
+
+		public object Convert(object value, Type targetType, object parameter, CultureInfo culture) =>
+			((Color)value).IsDefault ? FallbackColor : (Color)value;
+
+		public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+			=> throw new NotImplementedException();
 	}
 
 	public interface IQueryAttributable
--- a/Xamarin.Forms.Core/Shell/BaseShellItem.cs
+++ b/Xamarin.Forms.Core/Shell/BaseShellItem.cs
@@ -60,30 +60,6 @@ namespace Xamarin.Forms
 		public static readonly BindableProperty IsVisibleProperty =
 			BindableProperty.Create(nameof(IsVisible), typeof(bool), typeof(BaseShellItem), true);
 
-		public static readonly BindableProperty BadgeTextProperty =
-			BindableProperty.Create(nameof(BadgeText), typeof(string), typeof(BaseShellItem), null, BindingMode.OneWay);
-
-		public static readonly BindableProperty BadgeMoreTextProperty =
-			BindableProperty.Create(nameof(BadgeMoreText), typeof(string), typeof(BaseShellItem), null, BindingMode.OneWay);
-
-		public static readonly BindableProperty BadgeTextColorProperty =
-			BindableProperty.Create(nameof(BadgeTextColor), typeof(Color), typeof(BaseShellItem), Color.Default, BindingMode.OneWay);
-
-		public static readonly BindableProperty BadgeUnselectedTextColorProperty =
-			BindableProperty.Create(nameof(BadgeUnselectedTextColor), typeof(Color), typeof(BaseShellItem), Color.Default, BindingMode.OneWay);
-
-		public static readonly BindableProperty BadgeEffectiveTextColorProperty =
-			BindableProperty.CreateReadOnly(nameof(BadgeEffectiveTextColor), typeof(Color), typeof(BaseShellItem), Color.Default, BindingMode.OneWay).BindableProperty;
-
-		public static readonly BindableProperty BadgeColorProperty =
-			BindableProperty.Create(nameof(BadgeColor), typeof(Color), typeof(BaseShellItem), Color.Default, BindingMode.OneWay);
-
-		public static readonly BindableProperty BadgeUnselectedColorProperty =
-			BindableProperty.Create(nameof(BadgeUnselectedColor), typeof(Color), typeof(BaseShellItem), Color.Default, BindingMode.OneWay);
-
-		public static readonly BindableProperty BadgeEffectiveColorProperty =
-			BindableProperty.CreateReadOnly(nameof(BadgeEffectiveColor), typeof(Color), typeof(BaseShellItem), Color.Default, BindingMode.OneWay).BindableProperty;
-
 		static void OnTabIndexPropertyChanged(BindableObject bindable, object oldValue, object newValue) =>
 			((BaseShellItem)bindable).OnTabIndexPropertyChanged((int)oldValue, (int)newValue);
 
@@ -158,66 +134,6 @@ namespace Xamarin.Forms
 				return sgi.ShellElementCollection.Contains(this);
 
 			return false;
-		}
-
-		public string BadgeText
-		{
-			get { return (string)GetValue(BadgeTextProperty); }
-			set { SetValue(BadgeTextProperty, value); }
-		}
-
-		public string BadgeMoreText
-		{
-			get { return (string)GetValue(BadgeMoreTextProperty); }
-			set { SetValue(BadgeMoreTextProperty, value); }
-		}
-
-		public Color BadgeTextColor
-		{
-			get { return (Color)GetValue(BadgeTextColorProperty); }
-			set { SetValue(BadgeTextColorProperty, value); }
-		}
-
-		public Color BadgeUnselectedTextColor
-		{
-			get { return (Color)GetValue(BadgeUnselectedTextColorProperty); }
-			set { SetValue(BadgeUnselectedTextColorProperty, value); }
-		}
-
-		public Color BadgeColor
-		{
-			get { return (Color)GetValue(BadgeColorProperty); }
-			set { SetValue(BadgeColorProperty, value); }
-		}
-
-		public Color BadgeUnselectedColor
-		{
-			get { return (Color)GetValue(BadgeUnselectedColorProperty); }
-			set { SetValue(BadgeUnselectedColorProperty, value); }
-		}
-
-		public Color BadgeEffectiveTextColor => GetBadgeEffectiveTextColor(IsChecked);
-
-		public Color GetBadgeEffectiveTextColor(bool isSelected)
-		{
-			if (isSelected)
-			{
-				return BadgeTextColor;
-			}
-
-			return !BadgeUnselectedTextColor.IsDefault ? BadgeUnselectedTextColor : BadgeTextColor;
-		}
-
-		public Color BadgeEffectiveColor => GetBadgeEffectiveColor(IsChecked);
-
-		public Color GetBadgeEffectiveColor(bool isSelected)
-		{
-			if (isSelected)
-			{
-				return BadgeColor;
-			}
-
-			return !BadgeUnselectedColor.IsDefault ? BadgeUnselectedColor : BadgeColor;
 		}
 
 		internal virtual void SendAppearing()
@@ -311,17 +227,6 @@ namespace Xamarin.Forms
 		protected override void OnPropertyChanged([CallerMemberName] string propertyName = null)
 		{
 			base.OnPropertyChanged(propertyName);
-
-			if (propertyName == BadgeTextColorProperty.PropertyName || propertyName == BadgeUnselectedTextColorProperty.PropertyName || propertyName == IsCheckedProperty.PropertyName)
-			{
-				OnPropertyChanged(BadgeEffectiveTextColorProperty.PropertyName);
-			}
-
-			if (propertyName == BadgeColorProperty.PropertyName || propertyName == BadgeUnselectedColorProperty.PropertyName || propertyName == IsCheckedProperty.PropertyName)
-			{
-				OnPropertyChanged(BadgeEffectiveColorProperty.PropertyName);
-			}
-
 			if (Parent != null)
 			{
 				if (propertyName == Shell.ItemTemplateProperty.PropertyName || propertyName == nameof(Parent))
--- a/Xamarin.Forms.Core/Shell/BaseShellItem.cs
+++ b/Xamarin.Forms.Core/Shell/BaseShellItem.cs
@@ -468,7 +468,7 @@ namespace Xamarin.Forms
 
 				grid.Children.Add(badgeFrame, 2, 0);
 
-				badgeLabel.FontSize = 10;
+				badgeLabel.FontSize = Device.GetNamedSize(NamedSize.Micro, typeof(Label));
 				badgeLabel.SetBinding(Label.TextColorProperty, nameof(BaseShellItem.BadgeTextColor), converter: new UseFallbackColorIfDefaultColorConverter(Color.White));
 
 				INameScope nameScope = new NameScope();
--- a/Xamarin.Forms.Core/Shell/BaseShellItem.cs
+++ b/Xamarin.Forms.Core/Shell/BaseShellItem.cs
@@ -60,6 +60,30 @@ namespace Xamarin.Forms
 		public static readonly BindableProperty IsVisibleProperty =
 			BindableProperty.Create(nameof(IsVisible), typeof(bool), typeof(BaseShellItem), true);
 
+		public static readonly BindableProperty BadgeTextProperty =
+			BindableProperty.Create(nameof(BadgeText), typeof(string), typeof(BaseShellItem), null, BindingMode.OneWay);
+
+		public static readonly BindableProperty BadgeMoreTextProperty =
+			BindableProperty.Create(nameof(BadgeMoreText), typeof(string), typeof(BaseShellItem), null, BindingMode.OneWay);
+
+		public static readonly BindableProperty BadgeTextColorProperty =
+			BindableProperty.Create(nameof(BadgeTextColor), typeof(Color), typeof(BaseShellItem), Color.Default, BindingMode.OneWay);
+
+		public static readonly BindableProperty BadgeUnselectedTextColorProperty =
+			BindableProperty.Create(nameof(BadgeUnselectedTextColor), typeof(Color), typeof(BaseShellItem), Color.Default, BindingMode.OneWay);
+
+		public static readonly BindableProperty BadgeEffectiveTextColorProperty =
+			BindableProperty.CreateReadOnly(nameof(BadgeEffectiveTextColor), typeof(Color), typeof(BaseShellItem), Color.Default, BindingMode.OneWay).BindableProperty;
+
+		public static readonly BindableProperty BadgeColorProperty =
+			BindableProperty.Create(nameof(BadgeColor), typeof(Color), typeof(BaseShellItem), Color.Default, BindingMode.OneWay);
+
+		public static readonly BindableProperty BadgeUnselectedColorProperty =
+			BindableProperty.Create(nameof(BadgeUnselectedColor), typeof(Color), typeof(BaseShellItem), Color.Default, BindingMode.OneWay);
+
+		public static readonly BindableProperty BadgeEffectiveColorProperty =
+			BindableProperty.CreateReadOnly(nameof(BadgeEffectiveColor), typeof(Color), typeof(BaseShellItem), Color.Default, BindingMode.OneWay).BindableProperty;
+
 		static void OnTabIndexPropertyChanged(BindableObject bindable, object oldValue, object newValue) =>
 			((BaseShellItem)bindable).OnTabIndexPropertyChanged((int)oldValue, (int)newValue);
 
@@ -134,6 +158,66 @@ namespace Xamarin.Forms
 				return sgi.ShellElementCollection.Contains(this);
 
 			return false;
+		}
+
+		public string BadgeText
+		{
+			get { return (string)GetValue(BadgeTextProperty); }
+			set { SetValue(BadgeTextProperty, value); }
+		}
+
+		public string BadgeMoreText
+		{
+			get { return (string)GetValue(BadgeMoreTextProperty); }
+			set { SetValue(BadgeMoreTextProperty, value); }
+		}
+
+		public Color BadgeTextColor
+		{
+			get { return (Color)GetValue(BadgeTextColorProperty); }
+			set { SetValue(BadgeTextColorProperty, value); }
+		}
+
+		public Color BadgeUnselectedTextColor
+		{
+			get { return (Color)GetValue(BadgeUnselectedTextColorProperty); }
+			set { SetValue(BadgeUnselectedTextColorProperty, value); }
+		}
+
+		public Color BadgeColor
+		{
+			get { return (Color)GetValue(BadgeColorProperty); }
+			set { SetValue(BadgeColorProperty, value); }
+		}
+
+		public Color BadgeUnselectedColor
+		{
+			get { return (Color)GetValue(BadgeUnselectedColorProperty); }
+			set { SetValue(BadgeUnselectedColorProperty, value); }
+		}
+
+		public Color BadgeEffectiveTextColor => GetBadgeEffectiveTextColor(IsChecked);
+
+		public Color GetBadgeEffectiveTextColor(bool isSelected)
+		{
+			if (isSelected)
+			{
+				return BadgeTextColor;
+			}
+
+			return !BadgeUnselectedTextColor.IsDefault ? BadgeUnselectedTextColor : BadgeTextColor;
+		}
+
+		public Color BadgeEffectiveColor => GetBadgeEffectiveColor(IsChecked);
+
+		public Color GetBadgeEffectiveColor(bool isSelected)
+		{
+			if (isSelected)
+			{
+				return BadgeColor;
+			}
+
+			return !BadgeUnselectedColor.IsDefault ? BadgeUnselectedColor : BadgeColor;
 		}
 
 		internal virtual void SendAppearing()
@@ -227,6 +311,17 @@ namespace Xamarin.Forms
 		protected override void OnPropertyChanged([CallerMemberName] string propertyName = null)
 		{
 			base.OnPropertyChanged(propertyName);
+
+			if (propertyName == BadgeTextColorProperty.PropertyName || propertyName == BadgeUnselectedTextColorProperty.PropertyName || propertyName == IsCheckedProperty.PropertyName)
+			{
+				OnPropertyChanged(BadgeEffectiveTextColorProperty.PropertyName);
+			}
+
+			if (propertyName == BadgeColorProperty.PropertyName || propertyName == BadgeUnselectedColorProperty.PropertyName || propertyName == IsCheckedProperty.PropertyName)
+			{
+				OnPropertyChanged(BadgeEffectiveColorProperty.PropertyName);
+			}
+
 			if (Parent != null)
 			{
 				if (propertyName == Shell.ItemTemplateProperty.PropertyName || propertyName == nameof(Parent))
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/ShellBadges.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/ShellBadges.cs
@@ -2,6 +2,7 @@
 using System.ComponentModel;
 using System.Linq;
 using System.Runtime.CompilerServices;
+using Xamarin.Forms.Core;
 using Xamarin.Forms.CustomAttributes;
 using Xamarin.Forms.Internals;
 #if UITEST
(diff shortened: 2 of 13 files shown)


Dresel:
I have added a possible implementation to the `ShellSectionHeaderCell` for iOS, looks identically (I can't see a difference) to native bottom tabs (see updated Screenshot).

Ideas for "more tab" handling? I would propose a MoreTabBadgeColor, MoreTabBadgeText, etc. property?

How do you want this to be integrated / testable in the store shell?

- Some badges set by default?
- Add badges to tabs added by add top / bottom tab?
- Additional buttons to add badges with different colors and values?

alanag13:
Thanks for your hard work on this @Dresel! I have a project this would be perfect for and am excited to see it land.

Dresel:
I have updated the PR - see status - mostly property inheritance / propagation (don't know if it's done "right"), code cleanup, more tab handling (android), unselected badge color handling (android) and some buttons added for the store shell.

Dresel:
Is it possible to get the build logs so I can see why it is failing?

jfversluis:
We're working on getting the build output public, but we're not quite there yet.

Here's the errors:

> Xamarin.Forms.Platform.iOS\Renderers\ShellSectionRootHeader.cs(283,37): Error CS0117: 'BaseShellItem' does not contain a definition for 'BadgeTextColorProperty'
Xamarin.Forms.Platform.iOS\Renderers\ShellSectionRootHeader.cs(284,37): Error CS0117: 'BaseShellItem' does not contain a definition for 'BadgeColorProperty'
Xamarin.Forms.Platform.iOS\Renderers\ShellItemRenderer.cs(217,36): Error CS0117: 'ShellSection' does not contain a definition for 'BadgeTextColorProperty'
Xamarin.Forms.Platform.iOS\Renderers\ShellItemRenderer.cs(218,36): Error CS0117: 'ShellSection' does not contain a definition for 'BadgeColorProperty'
Xamarin.Forms.Platform.Android\Renderers\ShellSectionRenderer.cs(322,47): Error CS1061: 'TabLayout.Tab' does not contain a definition for 'View' and no accessible extension method 'View' accepting a first argument of type 'TabLayout.Tab' could be found (are you missing a using directive or an assembly reference?)

Dresel:
Build should be fixed. CI please.

jfversluis:
For some reason your PR is in draft. Whenever you make it final the CI should trigger

Dresel:
Why did it trigger the last time?

jfversluis:
I didn't notice it was in draft then, maybe you did it by mistake? Anyhow, it doesn't need to be a draft anymore, right? :)

Dresel:
But it's still work in progress and shouldn't be merged yet, I don't mind if it's draft or not :)

jfversluis:
It seems there are conflicts now that need to be solved first

Dresel:
Resolved.

Dresel:
Thanks, looks better now. Guess the failing test is unrelated?

iOS is now on the same level as Android, except one minor tweak needed for ShellSectionRootHeader.

I have added buttons to the StoreShell grid, so that it is easier to test certain badge features.

jfversluis:
Test is indeed unrelated. I've added the do-not-merge tag, so let us know when you're done with everything :)

samhouts:
@Dresel Do you have more work to add to this PR? Let us know! Thanks!

Dresel:
Yes, I'm currently working on the Flyout badge support.

Dresel:
I have rebased my latest changes.

I'm still eager to add dot style and css support but in a subsequent PR, since I don't have too much time at the moment anyway.

So I guess in the meantime it's time for a review (be gentle) 😆 🙈 

Some explanations:

All styling related properties should be inheritable - that means if I set e.g. the `BadgeColor` on a shell element (shell, item, section or content) all descendants should also inherit this color. For all of this properties I have added a corresponding  `GetInheritedX` function, which walks up the tree and returns the first property set.

Then there is an `EffectiveBadgeTextColor` and `EffectiveBadgeColor` property on BaseShellItem which returns the actual color an element should have, depending on if this element is selected or not and which related color properties are set (inherited).

All properties that have an effect to `EffectiveBadgeTextColor` and `EffectiveBadgeColor` do propagate changes to all descendants, that's what `PropagateBadgePropertyChanged` is for. I don't know if you can live with that or want something more fancy.

For inheritance and property changed propagation I haved added some unit tests to back this up.

For Flyout badge support on Android in the `ShellFlyoutRecyclerAdapter` I had to call [NotifyItemChanged](https://github.com/xamarin/Xamarin.Forms/blob/468f4fb3aaf7a0fc0ca173f9a5b1ab82f8b5251b/Xamarin.Forms.Platform.Android/Renderers/ShellFlyoutRecyclerAdapter.cs#L345), otherwise the UI didn't refresh. Maybe you know what's wrong here.

For `ShellSectionRootHeader` I had to add an [ShellContentToCellMapping](https://github.com/xamarin/Xamarin.Forms/blob/468f4fb3aaf7a0fc0ca173f9a5b1ab82f8b5251b/Xamarin.Forms.Platform.iOS/Renderers/ShellSectionRootHeader.cs#L66) so I can modify the cell on property changes. Don't know if this is done right, do I have to use `WeakReference` or do I have to care about reference cleanups?

Let me know if you need more information or want to discuss on a side channel.

Xamarin.Forms.Core/Internals/ShellExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Xamarin.Forms.Internals
+{
+	internal static class ShellExtensions
+	{
+		public static IEnumerable<BaseShellItem> GetShellDescendants(this BindableObject element)
+		{
+			IEnumerable<BaseShellItem> baseShellItems = new List<BaseShellItem>();
```
StephaneDelcroix: no need to allocate

Xamarin.Forms.Core/Internals/ShellExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+
+			switch (element)
+			{
+				case Shell shell:
+					baseShellItems = shell.GetShellDescendants();
```
StephaneDelcroix: you can return from here.

Xamarin.Forms.Core/Shell/BaseShellItem.cs
```diff
@@ -52,6 +60,9 @@ public class BaseShellItem : NavigableElement, IPropertyPropagationController, I
 									propertyChanged: OnTabStopPropertyChanged,
 									defaultValueCreator: TabStopDefaultValueCreator);
 
+		public static readonly BindableProperty BadgeTextProperty =
+			BindableProperty.Create(nameof(BadgeText), typeof(string), typeof(BaseShellItem), null, BindingMode.TwoWay);
```
StephaneDelcroix: why TwoWay ?

Xamarin.Forms.Core/Shell/BaseShellItem.cs
```diff
@@ -112,6 +123,51 @@ public bool IsTabStop
...
+
+		public Color BadgeTextColor => Shell.GetBadgeTextColor(this);
+
+		public Color BadgeUnselectedTextColor => Shell.GetBadgeTextColor(this);
+
+		public Color BadgeColor => Shell.GetBadgeColor(this);
+
+		public Color BadgeUnselectedColor => Shell.GetBadgeUnselectedColor(this);
+
+		public Color EffectiveBadgeTextColor => GetEffectiveBadgeTextColor(IsChecked);
```
StephaneDelcroix: those aren't bindable on the base shell item

Xamarin.Forms.Core/Shell/Shell.cs
```diff
@@ -130,6 +130,40 @@ static void OnFlyoutBehaviorChanged(BindableObject bindable, object oldValue, ob
 			BindableProperty.CreateAttached("UnselectedColor", typeof(Color), typeof(Shell), Color.Default,
 				propertyChanged: OnColorValueChanged);
 
+		public static readonly BindableProperty BadgeTextColorProperty =
```
StephaneDelcroix: why do you define them as attached on Shell ?

Xamarin.Forms.Core/Shell/BaseShellItem.cs
StephaneDelcroix: what does `EffectiveBadgeTextColor` means ?

StephaneDelcroix: I'm sure some will want to use FontImageSource for badges, shouldn't this include a BadgeFont property ?

Xamarin.Forms.Core/Shell/BaseShellItem.cs
```diff
@@ -17,7 +17,15 @@ public class BaseShellItem : NavigableElement, IPropertyPropagationController, I
 
 		#region PropertyKeys
 
-		internal static readonly BindablePropertyKey IsCheckedPropertyKey = BindableProperty.CreateReadOnly(nameof(IsChecked), typeof(bool), typeof(BaseShellItem), false);
+		internal static readonly BindablePropertyKey IsCheckedPropertyKey = BindableProperty.CreateReadOnly(nameof(IsChecked), typeof(bool), typeof(BaseShellItem), false, propertyChanged: OnCheckedValueChanged);
+
+		static void OnCheckedValueChanged(BindableObject bindable, object oldvalue, object newvalue)
```
StephaneDelcroix: this should be handled in platform code

Xamarin.Forms.Core/Shell/BaseShellItem.cs
```diff
@@ -112,6 +123,51 @@ public bool IsTabStop
...
+		public Color GetEffectiveBadgeTextColor(bool isSelected)
+		{
+			if (isSelected)
+			{
+				return Shell.Current.GetInheritedBadgeTextColor(this);
+			}
+
+			Color color = Shell.Current.GetInheritedBadgeUnselectedTextColor(this);
+
+			return !color.IsDefault ? color : Shell.Current.GetInheritedBadgeTextColor(this);
```
StephaneDelcroix: We don't do inherited BP at this time (except for a11y). I'd prefer leaving this out for now, and come up with a proper way to do BP inheritance, than cook something custom every time we think we need it.

If you want fine-grained control over the colors, use a CSS stylesheet

Xamarin.Forms.Core/Internals/ShellExtensions.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Xamarin.Forms.Internals
+{
+	internal static class ShellExtensions
+	{
+		public static IEnumerable<BaseShellItem> GetShellDescendants(this BindableObject element)
```
StephaneDelcroix: see comments about inherited BPs below. I think we can drop this all

Xamarin.Forms.Core/Shell/BaseShellItem.cs
```diff
@@ -185,6 +241,11 @@ static void OnIconChanged(BindableObject bindable, object oldValue, object newVa
 			shellItem.FlyoutIcon = (ImageSource)newValue;
 		}
 
+		internal void NotifyBadgePropertyChanged(string propertyName)
```
StephaneDelcroix: BP supports INPC on their own. you can drop this

Xamarin.Forms.Core/Shell/Shell.cs
```diff
@@ -130,6 +130,40 @@ static void OnFlyoutBehaviorChanged(BindableObject bindable, object oldValue, ob
...
+
+		public static readonly BindableProperty BadgeUnselectedColorProperty =
+			BindableProperty.CreateAttached("BadgeUnselectedColor", typeof(Color), typeof(Shell), Color.Default,
+				propertyChanged: PropagateBadgePropertyChanged(nameof(BaseShellItem.EffectiveBadgeColor)));
+
+		public static readonly BindableProperty BadgeMoreTextProperty =
+			BindableProperty.CreateAttached("BadgeMoreText", typeof(string), typeof(Shell), null,
+				propertyChanged: PropagateBadgePropertyChanged(nameof(BaseShellItem.EffectiveBadgeMoreText)));
+
+		public static BindableProperty.BindingPropertyChangedDelegate PropagateBadgePropertyChanged(string propertyName) =>
```
StephaneDelcroix: not required once the properties are moved, and we agree to drop the inheritance part.

Dresel:
@StephaneDelcroix Thanks for your input, finally some feedback.

I'm currently in the progress of moving all properties back to `BaseShellItem`. How would you apply / change the e.g. `BadgeColor` globally (at runtime)? I tried adding a global implicit style but this didn't work (see below). Should it?

```csharp

var shellContentStyle = new Style(typeof(ShellContent))
{
    Setters = {
        new Setter { Property = BaseShellItem.BadgeColorProperty, Value = Color.DarkOrange }
    }
};

Shell.Current.Resources.Add(shellContentStyle);

```

The user could iterate over all BaseShellItems and set the value, but that's cumbersome.

How would you handle `BadgeMoreText`? That's somehow a more global property, are we ok if it has to be set on every `ShellItem`?

Xamarin.Forms.Core/Internals/ShellExtensions.cs
Dresel: Dropped.

Xamarin.Forms.Core/Internals/ShellExtensions.cs
Dresel: Dropped.

Xamarin.Forms.Core/Shell/BaseShellItem.cs
Dresel: Mistake, changed.

Xamarin.Forms.Core/Shell/Shell.cs
Dresel: Moved to BaseShellItem.

Xamarin.Forms.Core/Shell/BaseShellItem.cs
Dresel: Good catch, will work on that after everything else is settled.

Xamarin.Forms.Core/Shell/BaseShellItem.cs
Dresel: Dropped.

Xamarin.Forms.Core/Shell/BaseShellItem.cs
Dresel: Dropped.

Xamarin.Forms.Core/Shell/BaseShellItem.cs
Dresel: Dropped.

Xamarin.Forms.Core/Shell/Shell.cs
Dresel: Dropped.

Xamarin.Forms.Core/Internals/ShellExtensions.cs
Dresel: Dropped.

Xamarin.Forms.Core/Shell/BaseShellItem.cs
Dresel: See answer below.

Dresel: Calculated color based on `IsChecked`, `BadgeColor` and `BadgeUnselectedColor`, see [here](https://github.com/Dresel/Xamarin.Forms/blob/badge-support/Xamarin.Forms.Core/Shell/BaseShellItem.cs#L183). Used for flyout templates [here](https://github.com/Dresel/Xamarin.Forms/blob/badge-support/Xamarin.Forms.Platform.Android/Renderers/ShellFlyoutRecyclerAdapter.cs#L250). Open for suggestions / better solutions.

Dresel:
@StephaneDelcroix

I have moved badge properties to BaseShellItem and removed property propagation as requested. I have only updated the Android version, I continue with iOS once everything is decided and settled.

It is still open if we want to support the dynamic change of default values for unset badge properties, like changing badge color to blue for all items that do not have a badge color set, without the need of binding a property to every item.

This was possible before by setting the attached property on shell and property propagation, which was now removed. Also check XamStore StorePages [here](https://github.com/Dresel/Xamarin.Forms/blob/badge-support/Xamarin.Forms.Controls/XamStore/Views/StorePages.cs#L383).

I don't know if this should be possible with dynamic styles / stylesheets?

I'm ok if we say that this is not supported, but I need at least some input.

Dresel:
Since there where a lot of changes in the meantime I will take this month to rebase and adapt to the current code base.

jfversluis:
That is awesome @Dresel, thank you so much!

TraceFerrier:
Thanks, @Dresel! We're *really* looking forward to (and desperately need) badge support for the Shell!

Dresel:
I have updated, rebased and squashed to the latest code base, for both Android and iOS.

I have also added a first batch of UI tests to ensure updates are properly reflected in the UI. I will add more UI tests in the future but this shouldn't hinder you from reviewing. Most of the functionality is integrated in the test case, so let me know if or what you want to keep in the XamStore.

To speed up and finalize this PR I have concentrated on the most important features, which are

- BadgeText
- BadgeMoreText
- BadgeTextColor
- BadgeUnselectedTextColor
- BadgeColor
- BadgeUnselectedColor

No inheritance / property propagation. Everything else like BadgeFonts, Stylesheet support, etc. is left out for future PRs. I will update the spec accordingly.

TL;DR: Ready for review 🎉

Dresel:
I'll see what I can do about those failing UI tests.

Dresel:
It seems that different API / SDK versions behave differently - is there a best practice how to handle that for UI tests? Can I mark tests to only run for specific versions? Should I avoid the hassle and just mark them as ManualReview?

It seems that newer iOS versions are using `UIImageView` for the native bottom badge that's why `SetSectionBadgeUnselectedColorTest` and `SetSectionBadgeColorTest` do not work - don't think there is a possibility to get the color here (prove me wrong).

Lower Android versions are returning `HashMap` for getColor instead of `ColorStateList` for the `GradientBackground`, that's why `SetItemBadgeUnselectedColorTest` and `SetItemBadgeColorTest` are failing here. Didn't check if there is an alternative to determine the background color yet.

For all the other failing tests - I don't know what's going on here, but I guess that's not related to the PR?

jfversluis:
I don't think we have infrastructure to run or skip tests for a certain OS version. But I think that's something you could easily incorporate yourself? For the relevant tests just query the OS version and whenever the API is not supported skip the test?

The other failing UI tests indeed do not seem to be related.

Dresel:
> 
> 
> @Dresel can you move all of the badge properties to Shell as Attached properties? When we put the `FlyoutIconProperty` on the BaseShellItem that was wrong.

Are you sure? Just curious because I was asked before to move that to the BaseShellItem. If you can double-check and confirm I will make them attached properties again.

PureWeen:
Hmm alright sorry about the back and forth @Dresel 
I've pinged Stephane internally so I can get clarity on his comments

aharwood2:
I guess this has missed the boat for 4.6?

mohamadayash:
@Dresel could you take into consideration the Right to Left support ?

pnda489:
@Dresel do you know when this will be resolved?

Dresel:
I can only guess but it looks to me that the team is concentrating on foundational improvements for MAUI so I wouldn't be surprised if this will never make it into Xamarin Forms. To be honest it's hard to tell because I rarely received any feedback.

I'm willing to help as far as I can - e.g. split this into separate PRs for flyout items, bottom tabs and tob tabs badges to reduce the initial time of review needed, but if no one talks to me there is nothing I can do. :man_shrugging:

angelru:
Any updates ??

PureWeen:
Hey @Dresel sorry for the delay just trying to work through how we want the APIs for this exposed

I should have some more concrete feedback next week just need to refine my thoughts a little bit.

The primary thinking here is that I don't want to limit this feature to being expressed on Shell. For example with TabbedPage/MDP/TabView (in development) I'd like a way to express badges in a more general way.

Here's a list of things I'm pondering over

- Currently I'm thinking we create a class `Badge` for all the APIs and they are attached properties.
- Typically we try to avoid things like selected/unselected in favor of Visual States but I'm not sure how easy that is to apply here to badges
- the Brushes PR is almost ready so we'll probably want to convert all the colors here to Brushes once that's in
- How many places can we reuse what you've done here https://github.com/xamarin/Xamarin.Forms/pull/6217/files#diff-0ea541c2f4760e7532094e9aa78ed72cR256 because the more we can base thee badges on xplat elements the more customizable they are. It'll let users do a lot more with the shape of the badge and things like RTL.  APIs like TextColor always concern me because they always evolve to just replicating all the APIs on Label so being able to just use a Label always makes me happy :-)

Dresel:
Thanks for the response, happy to get some insight.

> The primary thinking here is that I don't want to limit this feature to being expressed on Shell. For example with TabbedPage/MDP/TabView (in development) I'd like a way to express badges in a more general way.

I already noticed some badge properties appearing in the specs :wink:

My holiday has just begun so let me know when there is work todo 😄

PureWeen:
@Dresel let me know your thoughts on this

- Move these properties to a class called `Badge` and make them attached properties
- Just expose the following APIS currently: Badge.Text, Badge.TextColor, Badge.Background (formerly BadgeColor and make it a brush)
  - The unselected variations should be set through VisualStateManager or other means. This needs to be worked out a bit more but for now I'd say just remove them so we can get the basics here checked in.

I realize this removes the ability to set the badge on the more tab but the more tab is going to need something that can handle additional customization in general. We have a few issues out where people want to be able to set the text for the more tab or various other things. So, we'll probably need to add a general "MoreTabStyles" or something and then people can set the Badge there.

I really like this idea!! Can you make it do this?  ❤️ 

> One use case would be the transformation of the underlying items hidden by this more tab. For example Item 5 (Messages) has "1" new item and Item 6 (Notifications) has "2" new items, so you could show "3" in the more tab, but the user would not how how this is split between the aggregated items.

The color/background of the more badge would just be set through typical shell style propagation.

For example what we want to allow for this

```
<Shell Badge.BadgeTextColor="Green"
<FlyoutItem Badge.BadgeBackground="Purple"
```

Which will then propagate these styles down to the tabs. We don't want people to have to set the color on every tab.

There's code here
https://github.com/xamarin/Xamarin.Forms/blob/main/Xamarin.Forms.Core/Shell/Shell.cs#L1293

That walks up the tree to find the value that's set

Dresel:
> Move these properties to a class called Badge and make them attached properties

Alright. Seems logical if we want to add badge features in additional places.

> Badge.Background (formerly BadgeColor and make it a brush)

Might not work for iOS bottom tabs because the native `UITabBarItem.BadgeColor` is used, but for all other places this shouldn't be a problem.

> The unselected variations should be set through VisualStateManager or other means. This needs to be worked out a bit more but for now I'd say just remove them so we can get the basics here checked in.

Alright.

> I realize this removes the ability to set the badge on the more tab but the more tab is going to need something that can handle additional customization in general. We have a few issues out where people want to be able to set the text for the more tab or various other things. So, we'll probably need to add a general "MoreTabStyles" or something and then people can set the Badge there.

Is this something we should consider for the future or in this PR?

> I really like this idea!! Can you make it do this? ❤️

> One use case would be the transformation of the underlying items hidden by this more tab. For example Item 5 (Messages) has "1" new item and Item 6 (Notifications) has "2" new items, so you could show "3" in the more tab, but the user would not how how this is split between the aggregated items.

You are referring to this `We could think of modifying the text (e.g. for the example above "Messages (1)" and "Notifications (6)")`?

> The color/background of the more badge would just be set through typical shell style propagation.

This should work, yes. Only thing is that NotifyPropertyChanged will not be triggered for changes on parent elements, right?

Dresel:
What about

> I'm sure some will want to use FontImageSource for badges, shouldn't this include a BadgeFont property ?

PureWeen:
> I'm sure some will want to use FontImageSource for badges, shouldn't this include a BadgeFont property ?

Hmmmm my initial thoughts here are to go as simple as possible just to make sure we get this merged but just having text will probably make this a bit too thin.  I'll leave it up to you if you want to implement these additional properties or do so in the next version. You've done so much already with this PR I don't want to be too demanding :-) 

From `ITextElement`
- CharacterSpacing

From `IFontElement`
- Font
- FontFamily
- FontSize
- FontAttributes

> This should work, yes. Only thing is that NotifyPropertyChanged will not be triggered for changes on parent elements, right?


There's a whole structure for this on Shell. You'll see a thing on Shell called "SendAppearanceChanged" basically we'll just want to wire into that. I can fill in the bits here if need be.

Dresel:
Alright I'll see what I can do here. Let's start with the basic stuff.

Dresel:
I have to admit that I like brushes 😃

PureWeen:
@Dresel that's awesome!!

Can you target the 5.0 branch for this one as well?

Dresel:
I have pushed the first version based on your feedback and rebased to 5.0.0.

- UITests are disabled for now (needs rework)
- Flyout is not working at the moment, I still have to migrate the code to `Shell.GetFlyoutItemDataTemplate`
- iOS bottom tabs only support solid brush (still looking for a workaround here)
- MoreTab functionality not migrated yet

Regarding `MoreTabStyles` - should this be just another attached property container class or do you have something else in mind?

Did not have time yet to play around with selected / unselected VisualStates.

> There's a whole structure for this on Shell. You'll see a thing on Shell called "SendAppearanceChanged" basically we'll just want to wire into that. I can fill in the bits here if need be.

I still have to check how that works - but I wouldn't be sad if you take the lead here 😉

PureWeen:
> Regarding MoreTabStyles - should this be just another attached property container class or do you have something else in mind?

I wouldn't worry about this one for now

> iOS bottom tabs only support solid brush (still looking for a workaround here)

Hmmm yea that's annoying. Figuring out what will and won't work across platforms with brushes has been a fun journey.
 
I took a brief stab at figuring out how to let users supply custom UIViews for UITabs because iOS really locks this down and I found that you can do something like this inside `ShellItemRenderer`

```C#
CustomTabBar _customTabBar;
		public override void ViewDidLoad()
		{
			base.ViewDidLoad();

			_customTabBar = new CustomTabBar();
			this.SetValueForKey(_customTabBar, new NSString("tabBar"));
```

```C#
		class CustomTabBar : UITabBar
		{

		}
```

And then iOS will render your `CustomTabBar`
I don't know if this helps you at all but I was pretty excited when I found these hooks


> I still have to check how that works - but I wouldn't be sad if you take the lead here 😉

As far as ShellAppearance yea..... It has some weird quirks that will drive you crazy if you try to figure this out.  I'd like to vastly simplify the decision tree on how ShellAppearance works here but I can wire this up pretty quickly once I need.

PureWeen:
Hey @Dresel just wanted to check in and see how it's going!!

Just want to make sure in my attempts to simplify here I didn't actually make your life more complicated.

Dresel:
Hi, sorry for the delay, was on holiday 😎 

I did take a look on the flyout support. If I understood correctly, now that we moved to attached properties, we cannot use "normal" binding because of #1647, right?

I'm not aware of any workaround, so I can only think of reacting to INPC and modifying the controls directly (like ShellSectionRenderer, ShellItemRenderer). This would be [here](https://github.com/Dresel/Xamarin.Forms/blob/badge-support/Xamarin.Forms.Platform.Android/Renderers/ShellFlyoutRecyclerAdapter.cs#L295:L299) (ElementViewHolder) and [here](https://github.com/Dresel/Xamarin.Forms/blob/badge-support/Xamarin.Forms.Platform.iOS/Renderers/UIContainerCell.cs#L63:L69) (UIContainerCell), right?

PureWeen:
@Dresel hmmmmm that is annoying isn't it :-/ 

Alright, given that I would say we back peddle a little bit here just to make life a bit easier.

If we add the properties back to BaseShellItem but then still just map them to the attached properties that should do what we want yes?

```C#
		public string BadgeText
		{
			get { return Badge.GetBadgeText(this); }
			set { Badge.SetBadgeText(this, value); }
		}
```

similar to how the Route property is setup

```C#
		public string Route
		{
			get { return Routing.GetRoute(this); }
			set { Routing.SetRoute(this, value); }
		}
```

Dresel:
Yes, that makes it easier.

Flyout works now as before. We might want to add constraints as already done in the `CreateDefaultFlyoutItemCell` method and add the controls only for Android / iOS:

```csharp
if (Device.RuntimePlatform == Device.Android)
{
    // Add badge frame
}
```

I just did a quick test with the `StoreShell.xaml`, this works now (using the BaseShellItem properties directly will work too of course):

```xml
<ShellContent Title="Search" Route="search" ContentTemplate="{DataTemplate local:SearchHandlerPage}" core:Badge.BadgeText="1">
    <core:Badge.BadgeBackground>
        <LinearGradientBrush StartPoint="0,0" EndPoint="1,0">
            <GradientStop Color="Yellow" Offset="0.1"/>
            <GradientStop Color="Green" Offset="1.0"/>
        </LinearGradientBrush>
    </core:Badge.BadgeBackground>
</ShellContent>
```

PureWeen:
@Dresel  Thank you!!

Is this ready to review? Can you fix the conflicts?

Dresel:
Working on UITests, after that we are good to go.

Dresel:
After some fiddling I have managed to revive the UI tests except for ios badge background detection. Wasn't able to get the background color of the sublayer inserted by the brush extensions 🤷‍♂️ 

Also rebased to latest changes. I guess this should be ready for review.

Still open work items as discussed in this PR:

- Property propagation (ShellAppearance)
- Add font properties
- VisualStateManager support for selected / unselected states
- iOS bottom tabs support for non solid brushes (CustomTabBar)
- MoreTab badge support
- Text transformation of items behind more tab

As already said I would be happy if you grab the `property propagation` part. The other things I would like to tackle in subsequent PRs if not absolutely necessary for the initial PR.

pictos:
@Dresel can you resolve this conflict?

PureWeen:
fixed conflict

PureWeen:
@Dresel what are your thoughts about us moving this one over to the Community Toolkit? https://github.com/xamarin/XamarinCommunityToolkit

It'll be a little bit tricky at first as we open the right APIs to make all of this work but that'll help us improve the extensibility of Shell. Since we're locking down 5.0 APIs right now in preparation for .NET MAUI my plan is to start putting larger features like this one into the toolkit which will really let things like this grow over the next year opposed to stagnate as we are focusing on .NET MAUI. 

I realize we've put you through the wringer a bit on this one so someone on our side can take this code and port it over there.

Dresel:
> improve the extensibility of Shell

That's anyway one of the pain points, so I'm good with that.

So how should we proceed? Do you open another (or multiple) Issue / PR for those extension points? For the current status quo the following things need to be addressed:

Badge (flyout)

- Add the possibility to add the badge frame to the DataTemplate, returned by CreateDefaultFlyoutItemCell
- Find an alternative for the properties added to the BaseShellItem that are mapped back to the attached properties

Badge (bottom tabs)

- Add a hook to initial add the badges when the menu is created
- Add hooks for partial upgrades (OnShellItemPropertyChanged, OnShellSectionPropertyChanged)

Badge (top tabs)

Android
- Allow the modification of the created view
- Add hooks for partial upgrades (OnShellContentPropertyChanged)

iOS
- Allow the possibility to add a custom ShellSectionHeaderCell (or find an alternative for injecting the UIView / UILabel)
- Add hooks for partial upgrades (OnShellContentPropertyChanged) and some mechanism to retrieve the underlying ShellSectionHeaderCell

I'm still open for contributions, so let me know where I can help.

PureWeen:
Awesome @Dresel  !!!

Here's the other part of this I'm curious if there's synergy with.

@jsuarezruiz created a PR on XCT for a BadgeView
https://github.com/xamarin/XamarinCommunityToolkit/pull/337

Is there a way we can leverage that?

>  Do you open another (or multiple) Issue / PR for those extension points? 

Yes! We're in the process of identifying all of these places we need to open up so that XCT can work its magic.

Here's the catch all issue

https://github.com/xamarin/Xamarin.Forms/issues/12136

So if we can create PRs on the Forms side to open up what APIs you need then we can get all of this working inside XCT

Dresel:
> @jsuarezruiz created a PR on XCT for a BadgeView
> [xamarin/XamarinCommunityToolkit#337](https://github.com/xamarin/XamarinCommunityToolkit/pull/337)
> 
> Is there a way we can leverage that?

The least we could do is to use the BadgeView where we can use Xamarin Forms controls for Shell badges (= Flyout).

We might also unify what is possible (e.g. use Brush instead of Color for BadgeView background, add AutoHide for Shell badges, etc.) - since we deal with native controls directly, some things might be harder to synchronize - e.g. I don't know how to "translate" BadgeAnimation to native platform code for Shell badges yet.

For the first release we might also live with the circumstance that there are two possible badge approaches that have similar but not identical features. From an API perspective it's anyway different because on the one hand we use attached properties and on the other side wrap elements with a custom control?

> So if we can create PRs on the Forms side to open up what APIs you need then we can get all of this working inside XCT

I have some time reserved for this week, so if you don't have something else in mind I could open a draft PR for the changes I think I need?

PureWeen:
> I have some time reserved for this week, so if you don't have something else in mind I could open a draft PR for the changes I think I need and use this for further discussions?


@Dresel I think that sounds perfect thank you!!

I'm going to close this PR for now and look forward to your next ones. 

We're in the process of opening up APIs for the controls we're moving to XCT

https://github.com/xamarin/Xamarin.Forms/issues/12136

So if you have a laundry list of APIs let me know and we can start addressing those

jfversluis:
Hey @Dresel are you interested in still bringing this to the Toolkit? Or are you OK if someone else continues on your efforts? Thanks!

Dresel:
@jfversluis It's not that I'm not interested anymore but at the moment I'm really busy so I don't mind if someone wants to take this over and continue.

jfversluis:
Totally understandable! So no worries, that was why I was checking :) Thanks for letting us know! If it changes and you're still interested don't hesitate to ping me

angelru:
any progress on this?

PR #6239, opened 2019-05-21
[Shell, iOS] Fixes shell SearchBar background color

paymicro:
### Description of Change ###

Fixes shell SearchBar background color

### Issues Resolved ### 

none

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS

| Before | After |
|:------:|:-----:|
|   image  |  after |
| image | image |
| image | image |

### Testing Procedure ###

- run Control Gallery in Shell mode
- check background color in SearchBar

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/ShellSearchHandler.cs
+++ b/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/ShellSearchHandler.cs
@@ -1,0 +1,27 @@
+﻿using FormsElement = Xamarin.Forms.Shell;
+
+namespace Xamarin.Forms.PlatformConfiguration.iOSSpecific
+{
+	public static class ShellSearchHandler
+	{
+		public static readonly BindableProperty PureSearchBackgroundEnabledProperty =
+			BindableProperty.Create("PureSearchBackgroundEnabled", typeof(bool),
+			typeof(ShellSearchHandler), true);
+
+		public static bool GetSearchPureBackground(BindableObject element)
+			=> (bool)element.GetValue(PureSearchBackgroundEnabledProperty);
+
+		public static void SetSearchPureBackgroundEnabled(BindableObject element, bool value)
+			=> element.SetValue(PureSearchBackgroundEnabledProperty, value);
+
+		public static bool GetSearchPureBackgroundEnabled(this IPlatformElementConfiguration<iOS, FormsElement> config)
+			=> (bool)config.Element.GetValue(PureSearchBackgroundEnabledProperty);
+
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Platform.iOS/Extensions/ColorExtensions.cs
```diff
@@ -26,7 +26,7 @@ public static class ColorExtensions
 		public static CGColor ToCGColor(this Color color)
 		{
 #if __MOBILE__
-			return color.ToUIColor().CGColor;
+			return new CGColor(CGColorSpace.CreateGenericRgb(), new nfloat[] { (float)color.R, (float)color.G, (float)color.B, (float)color.A });
```
rmarinho: This change affects everyone, not cool :) for a shell only fix

Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/ShellSearchHandler.cs
```diff
@@ -0,0 +1,27 @@
+﻿using FormsElement = Xamarin.Forms.Shell;
```
paymicro: `IPlatformElementConfiguration` can only be applied to the type `Element`. `SearchHandler` isn't Element. Because of this I had to add to the `Shell` page. That is, all `SearchHandler` elements on the shell page will depend on this platform specific property.

PureWeen:
@paymicro is this related to this issue by chance?

https://github.com/xamarin/Xamarin.Forms/issues/6298

paymicro:
@PureWeen looks like similar, but in another place.

rmarinho:
Looking at this again, i wonder if we should just make this a property on Xamarin.Forms pages? that will be used more globally ? So if you are inside you are a nav page you could usethis too ?

@davidortinau any feedback if this is somthing users need?

samhouts:
Closing for now. Let's revisit later.

Bronson15:
@samhouts any reason this hasn't been implemented? The background color for the shell search handler defaults to a darker color of the Shell's background color and it does not look good. Should be able to make the SearchBar background color white at least. Even in the [Xaminals documentation](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/app-fundamentals/shell/search-images/searchhandler-large.png#lightbox) it doesn't look good.

samhouts:
@Bronson15 We were still unclear on the desired API for setting this property. It is still something that should be implemented. Thanks!

Bronson15:
@samhouts Anything new on this? Our search bar still looks like this and it is hard to use this way! 
image

PR #6243, opened 2019-05-21
Load resources

StephaneDelcroix:


StephaneDelcroix:
> Unit tests failing :(
 that's a previous test not resetting App.Current to null in TearDown

StephaneDelcroix:
OSX failure unrelated

PR #6272, opened 2019-05-22
Add Color.ToHex()

### LINKED ISSUE #6003: [Enhancement] Add Color.ToHexString
## Summary

Add ToHexString method on the Color to make it easier to convert a Xamarin Forms Color back to a Hex String.  

## API Changes

```cs
public string GetHexString()
    {
        var red = (int)(R * 255);
        var green = (int)(G * 255);
        var blue = (int)(B * 255);
        var alpha = (int)(A * 255);
        var hex = $"#{alpha:X2}{red:X2}{green:X2}{blue:X2}";

        return hex;
    }
```

## Intended Use Case

anywhere that it makes sense to create a native color from hex string.

DavidRoqueni:
### Description of Change ###

Implemented Color.ToHex()
Returns a string value equal to the Hex value of a given Color.

### Issues Resolved ### 

- fixes #6003 

### API Changes ###
Added:
string Color.ToHex() 

### Platforms Affected ### 
- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###
Added TestToHex() Unit Test in the color unit test package.
It ensure that given a RGB,RGBA, or HSLA color we are able to return the appropriate hex value. 

### PR Checklist ###

- [X] Has automated tests 


Xamarin.Forms.Core/Color.cs
```diff
@@ -301,6 +301,14 @@ public override string ToString()
...
 		}
 
+		public string ToHex()
+		{
+			var red = (int)(R * 255);
+			var green = (int)(G * 255);
+			var blue = (int)(B * 255);
+			var alpha = (int)(A * 255);
+			return $"#{alpha:X2}{red:X2}{green:X2}{blue:X2}";
+		}
```
StephaneDelcroix:
```suggestion
		}

```

StephaneDelcroix:
build

DavidRoqueni:
I apologize, I reran the Unit Test I provided and it appears that the HSLA to hex is not performing correctly when given an alpha below .8D the accuracy falls off by a degree of about +/- .000000001 which is enough to result in the unit test failing. Would you like me to further investigate this issue or update my unit tests?

samhouts:
@DavidRoqueni Hmm. The datatype difference between `FromHex` and `ToHex` is likely the culprit. See if you can fix it easily by using `ushort` or a datatype with a similar precision. Thanks!!

DavidRoqueni:
@samhouts Thank you for the recommendation. I was able to get to about 99.6% accuracy using the uint datatype. I was loosing a bit from the explicit conversion from double to uint. I have tried several different datatypes and strategies for tackling this and so far this is the best I have been able to achieve. I updated the unit tests as well to account for this. Thank you again for any and all help.

samhouts:
@rmarinho Can you check the failing unit tests here, please? Thanks!

StephaneDelcroix:
loosing precision while converting is expected. it's all fine

Issue #4143, opened 2018-10-19
Span GestureRecognizer inaccuracies


davidbritch:
### Description

[Previously](https://github.com/xamarin/Xamarin.Forms/issues/3525) I logged an issue for Forms 3.2 about gesture recognizer on Spans. While a `TapGestureRecognizer` worked on the `Span`, they were really finicky on iOS.

I've just tried a `TapGestureRecognizer` on a `Span` with Forms 3.3 and it's still finicky on iOS. It's also finicky on Android. Just less finicky.

### Steps to Reproduce

1. Run the attached sample.
2. Tap the "Xamarin documentation" text in the `Span`.

### Expected Behavior

The `TapGestureRecognizer` detects the tap and opens a browser on the Xamarin documentation page.

### Actual Behavior

The `TapGestureRecognizer` detects the tap and opens a browser, provided you tap on specific areas of the text (e.g. the top of the "X" in "Xamarin". Tapping on the bottom of the "m" in "Xamarin" does nothing.

### Basic Information

- Version with issue: 3.3
- Last known good version: 3.3.0.912540
- IDE: VSMac 7.6.9
- Platform Target Frameworks:
  - iOS: iOS 12
  - Android: API 27
- Android Support Library Version: 27.0.2.1

### Reproduction Link

[HyperlinkLabel.zip](https://github.com/xamarin/Xamarin.Forms/files/2495950/HyperlinkLabel.zip)

hartez:
Confirmed. It's really easy to see with an Android emulator - turn on the developer mode touch tracking, move your mouse pointer to the bottom center of the "m" in Xamarin, and click. On the upper "Xamarin" label, nothing will happen, but on the bottom one the gesture will activate (as expected).

hauteur:
I'm not positive if this is the same issue - but I'm also seeing it not register the tap if it is not the first span in the list.

If you add a FormattedString, and the span with the TapGestureRecognizer is first in the sequence, it recognizes the tap fine.  If you put it second, the tap isn't registering.

VitalyKnyazev:
For me it works on iPad in portrait mode, but does not in landscape

Blin4ik91:
Same issue here, spans are not tapable on iOS or Android.
In my example the clickable words are linebreaked. (also when not linebreaked)

adrianknight89:
[...]

PR #6299, opened 2019-05-24
Label padding

### LINKED ISSUE #4784: [Enhancement] Padding on Label
## Summary
Make padding available in Label

## API Changes
var label = new Label();
label.Padding = new Thickness(10,0,0,0);

<Label Text="Test" Padding="10,0,0,0"/>

## Intended Use Case
It can be useful for a simple header or section divider instead of creating a contentView with padding and putting a label into it.

LOOK at the picture for an example
schermata 2018-12-18 alle 09 26 20

andreinitescu:
### Description of Change ###

Added Padding property to Label.

_WARNING: It might look boring to read the text below, but it's all worth it, I promise! Thanks!_

### iOS behavior and implementation ###
All platforms beside iOS (and MacOS) support padding on their native label, so we already know that the  real issue is with iOS.

We also know this can be easily achieved on iOS by having a UILabel derived class and overriding some of the UILabel methods (which is what I did in this PR too). However, using a derived UILabel to render any Label can be a performance issue on both native iOS and Xamarin. How much it hurts the performance? It depends on how heavy a Page is, how many labels exist on it.

**The compromise for iOS:**
If a Label is created with a non-zero Padding, on iOS the label renderer will render a FormsLabel class derived from UILabel and Padding will work:
```
_myLabel = new Label() { Padding = new Thickness(40, 20) }; 
stackLayout.Add(_myLabel);

// Later on, after Page was rendered:
void OnMyBtnClicked()   
{
   _myLabel.Padding = new Thickness(5);  // This will work, padding is updated
}
```

If Padding is not set (it's zero), it will render to UILabel, like it's working today. In this case, updating the Padding **will be ignored** after the Label is created:
```
_myLabel = new Label(); 
stackLayout.Add(_myLabel);

// Later on, after Page was rendered:
void OnMyBtnClicked() 
{
    _myLabel.Padding = new Thickness(5);  // iOS ONLY: This will NOT work, padding is not set
}
```

In other words, if you need Padding on the Label, set it when you create the Label and it's added to the Page, not afterwards. I think it's a good compromise, because in practice, there are very few scenarios where Padding needs to be updated. 

**What if an app really needs to update the Padding which is initially empty?** 
If this is really needed, one option is to overload the Label c-tor with a boolean parameter to indicate the iOS renderer that Padding will be updated, just how the [`CachingStrategy`](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/listview/performance#setting-the-caching-strategy) works on ListView, and it will have a purpose only for iOS:
```
XAML: <Label IsPaddingUpdated="true" />
C#: _myLabel = new Label(isPaddingUpdated: true);
```

Of course, the **3rd option** is to just always render to the UILabel derived class FormsLabel but, like I said above, it can hurt the performance. 

**4th option**: Make the iOS Label renderer switch to the UILabel derived class (FormsLabel) when Padding is set. But I'm not sure, is this is possible at the moment? What are the implications of switching the renderer? If you think it's worth exploring please share input on this.

What do you think?

### Issues Resolved ### 

- fixes #4784

### API Changes ###

- Core/XAML (all platforms)
- iOS
- Android
- UWP
- WPF

### Testing Procedure ###

See the **Label Gallery** page (LabelCoreGalleryPage):

**(IMPORTANT: Note that because of #4143, depending on the place where you tap a Span, tapping might NOT work!!!!!)**

image

### PR Checklist ###

- [ ] Has automated tests 


andreinitescu:
I added some code examples in the description which make it very clear

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
```diff
@@ -469,5 +485,46 @@ void UpdateMaxLines()
...
+		void UpdatePadding()
+		{
+			if (Element.Padding.IsEmpty)
+				return;
+
+#if __MOBILE__
+			var formsLabel = Control as FormsLabel;
+			if (formsLabel == null)
+			{
+				Debug.WriteLine($"{nameof(LabelRenderer)}: On iOS, a Label created with no padding will ignore padding changes");
```
adrianknight89: In a distant corner of my mind, I remember @samhouts saying the team would like to avoid slowing down the debugger with logs.

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
```diff
@@ -469,5 +485,46 @@ void UpdateMaxLines()
...
+					(float)Element.Padding.Top,
+					(float)Element.Padding.Left,
+					(float)Element.Padding.Bottom,
+					(float)Element.Padding.Right);
+			UpdateLayout();
+#endif
+		}
+
+#if __MOBILE__
+		class FormsLabel : NativeLabel
```
adrianknight89: Wondering if this should be moved to its own file...

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
```diff
@@ -469,5 +485,46 @@ void UpdateMaxLines()
...
+			}
+
+			public override void DrawText(RectangleF rect) => base.DrawText(TextInsets.InsetRect(rect));
+
+			public override SizeF SizeThatFits(SizeF size) => AddInsets(base.SizeThatFits(size));
+
+			SizeF AddInsets(SizeF size) => new SizeF(
+				width: size.Width + TextInsets.Left + TextInsets.Right,
+				height: size.Height + TextInsets.Top + TextInsets.Bottom);
+		}
```
adrianknight89: Have you considered overriding `intrinsicContentSize` as well?  See https://stackoverflow.com/a/27461090/246392

adrianknight89:
@andreinitescu Thanks for the PR, man. Padding is a much needed feature for Label, imo. Let's move Forms forward :)

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
andreinitescu: @adrianknight89 this isn't part of any critical performance path (like layout update or something like that), and most importantly, it's very useful for anybody to understand what's happening.

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
andreinitescu: @adrianknight89 Nah, it's an internal small class for specific purposes. There's lots of similar classes already.

Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
andreinitescu: intrinsicContentSize is for auto layout AFAIK.

Xamarin.Forms.Core/Label.cs
```diff
@@ -210,7 +218,7 @@ public int MaxLines
 			InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
 
 		void IFontElement.OnFontChanged(Font oldValue, Font newValue) =>
-         	InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
+			 InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
```
StephaneDelcroix: meh

andreinitescu: 😱 done!

andreinitescu: Tabs vs Spaces, I spent so much time fixing those so it doesn't get committed, but I missed one

JKennedy24:
please merge this asap.

I'm hoping it will optimise some of my views where I've had to use extra Layouts to achieve this

PR #6312, opened 2019-05-25
Switch ThumbColor

### LINKED ISSUE #5415: [Enhancement] Add ThumbColor for switch
We can modify the OnColor but not the ThumbColor for a switch element.

Can you add this possibility ? Thanks.

andreinitescu:
### Description of Change ###

Adds ThumbColor property to Switch

### Issues Resolved ### 

- fixes #5415

### API Changes ###

- Core/XAML (all platforms)
- iOS
- Android
- UWP

image

image

image

### Testing Procedure ###

See Switch gallery page

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls/CoreGalleryPages/SwitchCoreGalleryPage.cs
+++ b/Xamarin.Forms.Controls/CoreGalleryPages/SwitchCoreGalleryPage.cs
@@ -14,11 +14,11 @@ namespace Xamarin.Forms.Controls
 			get { return false; }
 		}
 
-		protected override void Build (StackLayout stackLayout)
+		protected override void Build(StackLayout stackLayout)
 		{
-			base.Build (stackLayout);
+			base.Build(stackLayout);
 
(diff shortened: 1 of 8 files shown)


toomasz:
@andreinitescu Appreciate these small enhancements you are commiting, they are so needed and desired!

MaxenceSAUNIER:
Thanks you very much

andreinitescu:
Can I rebase this against 4.1.0 so it goes out asap?

StephaneDelcroix:
> Can I rebase this against 4.1.0 so it goes out asap?

@andreinitescu  no, it's an enhancement

andreinitescu:
@StephaneDelcroix OK, I got it!

PureWeen:
I think this should use SetTint instead of SetColorFilter

@andreinitescu why did you use the particular ColorFilter settings on this one that you are using?

From the image below the image on the right is using SetTint where as the one on the left is using Color Filter

image

It looks like the color filter is causing the shadow to disappear

andreinitescu:
@PureWeen Thanks for pointing out the issue. It's fixed in #6798. The code is already using color filter for the "on" color so I thought to use the same for consistency

PR #6345, opened 2019-05-29
[iOS] adds a flag to ios if user needs to disable accessibility sizing for named font sizes

PureWeen:
### Description of Change ###
https://github.com/xamarin/Xamarin.Forms/pull/5985

Changed the default behavior of named fontsizes for iOS to make it more in line with how Android and UWP work. This technically is a breaking change but we see it as an acceptable breaking change. That being said we still want to give users the ability to opt out of it if need be.

### Platforms Affected ### 
- iOS

### Testing Procedure ###
- Create a page where font sizes are being set through named values (or just make one)
- Go to accessibility settings on the iDevice, set large text on, and make the text really large
- run the app
- now set this flag
```C#

	    Forms.SetFlags("DisableAccessibilityScalingForNamedFontSizes");
            global::Xamarin.Forms.Forms.Init();
```
- Rerun the app and the font should now be smaller because it's no longer scaled

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/Application.cs
+++ b/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/Application.cs
@@ -54,27 +54,27 @@
 		}
 		#endregion
 
-		#region DisableAccessibilityScalingForNamedFontSizes
-		public static readonly BindableProperty DisableAccessibilityScalingForNamedFontSizesProperty = BindableProperty.Create("DisableAccessibilityScalingForNamedFontSizes", typeof(bool), typeof(Application), false);
+		#region EnableAccessibilityScalingForNamedFontSizes
+		public static readonly BindableProperty EnableAccessibilityScalingForNamedFontSizesProperty = BindableProperty.Create("EnableAccessibilityScalingForNamedFontSizes", typeof(bool), typeof(Application), true);
 
-		public static bool GetDisableAccessibilityScalingForNamedFontSizes(BindableObject element)
+		public static bool GetEnableAccessibilityScalingForNamedFontSizes(BindableObject element)
 		{
-			return (bool)element.GetValue(DisableAccessibilityScalingForNamedFontSizesProperty);
+			return (bool)element.GetValue(EnableAccessibilityScalingForNamedFontSizesProperty);
 		}
 
-		public static void SetDisableAccessibilityScalingForNamedFontSizes(BindableObject element, bool value)
+		public static void SetEnableAccessibilityScalingForNamedFontSizes(BindableObject element, bool value)
 		{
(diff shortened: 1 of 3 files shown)


samhouts:
@PureWeen Why did we go for a flag instead of a platform specific here?

PureWeen:
@samhouts primarily because the method this applies to doesn't get applied to a particular element it just gets asked as it relates to the type

https://github.com/xamarin/Xamarin.Forms/blob/fbb731bcdc966565c9672f917711dc1304573f31/Xamarin.Forms.Platform.iOS/Forms.cs#L201

AFAIK we'd have to add some apis to pass the actual element in that it's being applied to but we'd also need to have the global flag option if people don't have an element to apply a platform specific to.

Are you thinking of a better place I can hook into for this setting?

samhouts:
@PureWeen We tend to use flags for settings that we intend to remove at some point, which is why we sue strings instead of strong naming. If we're adding a permanent setting like this, we should probably either add a property to the `FormsApplicationDelegate` or use a platform specific. We can create platform specifics on the `Application` level (we did for Android). Thoughts?

PureWeen:
> We can create platform specifics on the Application level (we did for Android). Thoughts?

👍 

I see that now

Xamarin.Forms.Platform.iOS/Forms.cs
```diff
@@ -199,19 +200,28 @@ public double GetNamedSize(NamedSize size, Type targetElementType, bool useOldSi
 			{
 				// We make these up anyway, so new sizes didn't really change
 				// iOS docs say default button font size is 15, default label font size is 17 so we use those as the defaults.
+				var scalingFactor = _fontScalingFactor;
+
+#if __MOBILE__
+				if (Application.Current?.On<iOS>().GetDisableAccessibilityScalingForNamedFontSizes() == true)
```
StephaneDelcroix: REALLY ?

🤪

Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/Application.cs
```diff
@@ -53,5 +53,30 @@ public static bool GetHandleControlUpdatesOnMainThread(this IPlatformElementConf
 			return config;
 		}
 		#endregion
+
+		#region DisableAccessibilityScalingForNamedFontSizes
+		public static readonly BindableProperty DisableAccessibilityScalingForNamedFontSizesProperty = BindableProperty.Create("DisableAccessibilityScalingForNamedFontSizes", typeof(bool), typeof(Application), false);
```
StephaneDelcroix: I prefer positive conditions, even if the default value is true. double negation is confusing

PR #6363, opened 2019-05-31
Added option to display icons instead of text in context actions for iOS

Sebastian1989101:
### Description of Change ###

/!\ This is the same PR as https://github.com/xamarin/Xamarin.Forms/pull/6382 just implemented in a more dynamic way but in this case the More ContextAction could not changed by the app developer (because it is never an icon) and it requires more changes to Xamarin.Forms (MenuItem additional interface) - I would prefer https://github.com/xamarin/Xamarin.Forms/pull/6382 as 2nd thoughts  /!\

I've added a enum to choose if ContextActions within a ListView should be displayed as Icon or Text. This can be activated for each individual MenuItem. 

### Issues Resolved ### 

None

### API Changes ###

Added:
 - IElementConfiguration<MenuItem> to MenuItem
- Added On<iOS>().SetContextActionDisplay(ContextActionDisplay.Icon) for MenuItem

### Platforms Affected ### 

- iOS

None with default settings. When ContextActionDisplay.Icon is set for MenuItem they will appear as Icon instead of text in iOS. 

Not applicable

### Testing Procedure ###

Added "Context Action Display Test" to Control Gallery which includes one Icon ContextAction and one Text based. 

### PR Checklist ###

- [ ] Has automated tests 

### Additional noted ###

Maybe this option should be moved to the ViewCell instead. This would prevent a base class to change (in this case MenuItem which now has an additional interface) and the More-Action could also change to an Icon (maybe the UITabBarSystemItem.More?) but it will remove some freedom which should be displayed as Icon and which one should not (at least if there is any reason for this to happen?). 

/!\ This is the same PR as https://github.com/xamarin/Xamarin.Forms/pull/6382 just implemented in a more dynamic way but in this case the More ContextAction could not changed by the app developer (because it is never an icon) and it requires more changes to Xamarin.Forms (MenuItem additional interface) - I would prefer https://github.com/xamarin/Xamarin.Forms/pull/6382 as 2nd thoughts  /!\

### DIFF
--- a/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/MenuItem.cs
+++ b/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/MenuItem.cs
@@ -1,0 +1,35 @@
+﻿namespace Xamarin.Forms.PlatformConfiguration.iOSSpecific
+{
+	using FormsElement = Forms.MenuItem;
+
+	public static class MenuItem
+	{
+		public static readonly BindableProperty ContextActionDisplayProperty = BindableProperty.Create(nameof(ContextActionDisplay), typeof(ContextActionDisplay), typeof(MenuItem), ContextActionDisplay.Text);
+
+		public static ContextActionDisplay GetContextActionDisplay(BindableObject element)
+			=> (ContextActionDisplay)element.GetValue(ContextActionDisplayProperty);
+
+		public static void SetContextActionDisplay(BindableObject element, ContextActionDisplay value)
+			=> element.SetValue(ContextActionDisplayProperty, value);
+
+		public static ContextActionDisplay DefaultContextActionCellDisplay(this IPlatformElementConfiguration<iOS, FormsElement> config)
+			=> GetContextActionDisplay(config.Element);
+
+		public static ContextActionDisplay GetContextActionDisplay(this IPlatformElementConfiguration<iOS, FormsElement> config)
+		{
+			return GetContextActionDisplay(config.Element);
+		}
+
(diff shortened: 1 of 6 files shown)


Sebastian1989101:
Updated initial PR text

Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/MenuItem.cs
```diff
@@ -0,0 +1,35 @@
+﻿namespace Xamarin.Forms.PlatformConfiguration.iOSSpecific
+{
+	using FormsElement = Forms.MenuItem;
+
+	public static class MenuItem
+	{
+		public static readonly BindableProperty ContextActionDisplayProperty = BindableProperty.Create(nameof(ContextActionDisplay), typeof(ContextActionDisplay), typeof(MenuItem), ContextActionDisplay.Text);
```
hartez: `typeof(MenuItem)` - this is likely referring to this static class; this should probably be `typeof(Forms.MenuItem)` or `typeof(FormsElement)`.

geskill:
Should add the TextAndIcon or simply „Both“ value directly since adding the iOS implementation is done in a sec

PureWeen:
if it is indeed easy and @Sebastian1989101 wants to try it 👍 

But it's not something I would hold up this PR for.

rmarinho:
Hey @Sebastian1989101 do you want to get back to apply the changes requested or should we take over?

Sebastian1989101:
@rmarinho Feel free to take it if you want. I don't get the time currently to make the requested modifications.

rmarinho:
Since we have SwipeView now we will work on that feature there.

PR #6389, opened 2019-06-03
Added tap position to the Tapped event of the TapGestureRecognizer

### LINKED ISSUE #5347: [Enhancement] Add more data to TapGestureRecognizer.Tapped event
It would be far better to have the some informations about where the tap happens in the TapGestureRecognizer.Tapped event. Raising this event without any information is in some situations enough but in some you really want to know where the tap happens.

Sebastian1989101:
### Description of Change ###

Added tap position to the Tapped Event of the TapGestureRecognizer.

### Issues Resolved ### 

- fixed #5347

### API Changes ###

### DIFF
--- a/Xamarin.Forms.Controls/ControlGalleryPages/TapGestureTestPage.cs
+++ b/Xamarin.Forms.Controls/ControlGalleryPages/TapGestureTestPage.cs
@@ -15,7 +15,7 @@ namespace Xamarin.Forms.Controls
 				RowDefinitions = new RowDefinitionCollection { new RowDefinition {Height = GridLength.Auto},new RowDefinition {Height = GridLength.Auto}, new RowDefinition {Height = GridLength.Star} }
 			};
 
-			var picker = new Picker { ItemsSource = new List<int> { 1, 2 } };
+			var picker = new Picker { ItemsSource = new List<int> { 1, 2 }, Title = "Number of taps" };
 			grid.AddChild(picker, 0, 0);
 
 			outputLabel = new Label { Text = "Nothing tapped yet!", LineBreakMode = LineBreakMode.WordWrap, HorizontalTextAlignment = TextAlignment.Center };
--- a/Xamarin.Forms.Controls/ControlGalleryPages/TapGestureTestPage.cs
+++ b/Xamarin.Forms.Controls/ControlGalleryPages/TapGestureTestPage.cs
@@ -1,0 +1,56 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace Xamarin.Forms.Controls
+{
+	public class TapGestureTestPage : ContentPage
+	{
+		readonly Label outputLabel;
+
+		public TapGestureTestPage()
+		{
+			var grid = new Grid
+			{
+				ColumnDefinitions = new ColumnDefinitionCollection { new ColumnDefinition { Width = GridLength.Star }, new ColumnDefinition { Width = GridLength.Star } },
+				RowDefinitions = new RowDefinitionCollection { new RowDefinition {Height = GridLength.Auto},new RowDefinition {Height = GridLength.Auto}, new RowDefinition {Height = GridLength.Star} }
+			};
+
+			var picker = new Picker { ItemsSource = new List<int> { 1, 2 } };
+			grid.AddChild(picker, 0, 0);
+
+			outputLabel = new Label { Text = "Nothing tapped yet!", LineBreakMode = LineBreakMode.WordWrap, HorizontalTextAlignment = TextAlignment.Center };
+			grid.AddChild(outputLabel, 0, 1, 2);
+
+			var box = new BoxView { BackgroundColor = Color.Gray };
+			grid.AddChild(box, 0, 2, 2);
+
+			var button = new Button { Text = "Set", Command = new Command(() => UpdateGesture(picker, box)) };
+			grid.AddChild(button, 1, 0);
+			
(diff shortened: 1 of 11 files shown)


Xamarin.Forms.Controls/ControlGalleryPages/TapGestureTestPage.cs
```diff
@@ -0,0 +1,56 @@
...
+
+		public TapGestureTestPage()
+		{
+			var grid = new Grid
+			{
+				ColumnDefinitions = new ColumnDefinitionCollection { new ColumnDefinition { Width = GridLength.Star }, new ColumnDefinition { Width = GridLength.Star } },
+				RowDefinitions = new RowDefinitionCollection { new RowDefinition {Height = GridLength.Auto},new RowDefinition {Height = GridLength.Auto}, new RowDefinition {Height = GridLength.Star} }
+			};
+
+			var picker = new Picker { ItemsSource = new List<int> { 1, 2 } };
```
paymicro: Please add on-screen description what is that number. Ex: "Number of taps"

Xamarin.Forms.Platform.Android/TapGestureHandler.cs
```diff
@@ -39,7 +39,7 @@ public bool OnTap(int count, Point point)
 			if (children != null)
 				foreach (var recognizer in children.GetChildGesturesFor<TapGestureRecognizer>(recognizer => recognizer.NumberOfTapsRequired == count))
 				{
-					recognizer.SendTapped(view);
+					recognizer.SendTapped(view, point);
```
paymicro: Here need add `pixelTranslation` function. For example, look like made in:
https://github.com/xamarin/Xamarin.Forms/blob/d150e22238ea184a2f9aa3c4b7dcfb134db6f51d/Xamarin.Forms.Platform.Android/PanGestureHandler.cs#L27-L32

Also below.

Xamarin.Forms.Core/TappedEventArgs.cs
```diff
@@ -4,11 +4,13 @@ namespace Xamarin.Forms
 {
 	public class TappedEventArgs : EventArgs
 	{
-		public TappedEventArgs(object parameter)
+		public TappedEventArgs(object parameter, Point position)
```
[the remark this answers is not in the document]
Sebastian1989101: Is there any advantage to have two seperate constructors instead of using default value parameters? 
`public TappedEventArgs(object parameter, Point position = new Point())`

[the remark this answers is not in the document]
paymicro: Most likely the default value is sufficient.
Thanks.

Sebastian1989101:
@paymicro Is there anything still todo here for me? Because the status is still "Changes requested"?

Xamarin.Forms.Core/TapGestureRecognizer.cs
```diff
@@ -34,18 +34,16 @@ public int NumberOfTapsRequired
 			set { SetValue(NumberOfTapsRequiredProperty, value); }
 		}
 
-		public event EventHandler Tapped;
+		public event TappedEventHandler Tapped;
```
samhouts: breaking change

Xamarin.Forms.Core/TapGestureRecognizer.cs
```diff
@@ -34,18 +34,16 @@ public int NumberOfTapsRequired
...
 
 		[EditorBrowsable(EditorBrowsableState.Never)]
-		public void SendTapped(View sender)
+		public void SendTapped(View sender, Point position)
```
samhouts: breaking change

Xamarin.Forms.Core/TappedEventArgs.cs
```diff
@@ -4,11 +4,13 @@ namespace Xamarin.Forms
 {
 	public class TappedEventArgs : EventArgs
 	{
-		public TappedEventArgs(object parameter)
+		public TappedEventArgs(object parameter, Point position = new Point())
```
samhouts: breaking change

hartez:
@Sebastian1989101 The implementation looks fine, but this introduces several breaking changes. Any implementation of this feature needs to be backward-compatible; no API or ABI breaks. 

This would need to either:

1. Keep all the existing tap behavior while also providing a second event and `Send` method (e.g., `public event EventHandler Tapped` stays, and `public event TappedEventHandler LocationTapped` is added; `SendTapped` stays, and `SendLocationTapped` is added; `TappedEventArgs` stays and `LocationTappedEventArgs` is added). `TapGestureRecognizer` can then handle both versions. Or,

2. Add a whole new gesture recognizer called `LocationTapGestureRecognizer`.

(All names here are suggestions and subject to folks coming up with better ideas.)

@samhouts Any thoughts on which version you might prefer? Or other options?

Xamarin.Forms.Core/TapGestureRecognizer.cs
rmarinho: Add a new Method and call it SendTappedWithPosition

samhouts:
@Sebastian1989101 Thanks for this PR! We're unable to accept the changes as is, however, since we do not allow ABI/API breaking changes except in the rarest of circumstances. If you're able to address these concerns, feel free to reopen this PR or submit a new one.

miguelhasse:
@Sebastian1989101 don't you want to address the ABI/API breaking changes that were pointed out?

PR #6395, opened 2019-06-03
Speed up core UI tests and reduce flakiness

hartez:
### Description of Change ###

Allow the core UI tests to jump directly to the relevant test, instead of hitting the "Move Next" button over and over again.

Ensure the presence of target element before querying the property value.

Fix bugs with Entry focus tests on devices with smaller screens and software keyboard enabled.

Avoid edge case with dismissal of Date/Time picker in tap bubbling tests.

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/GestureBubblingTests.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/GestureBubblingTests.cs
@@ -73,6 +73,7 @@ namespace Xamarin.Forms.Controls.Issues
 			{
 				// These controls show a pop-up which we have to cancel/done out of before we can continue
 #if __ANDROID__
+				System.Threading.Tasks.Task.Delay(2000).Wait(); // If we hit back before the dialog is fully up, it may not dismiss
 				RunningApp.Back();
 #elif __IOS__
 				var cancelButtonText = "Done";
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/GestureBubblingTests.cs
```diff
@@ -73,6 +73,7 @@ public void VerifyTapBubbling(string menuItem, bool frameShouldRegisterTap)
 			{
 				// These controls show a pop-up which we have to cancel/done out of before we can continue
 #if __ANDROID__
+				System.Threading.Tasks.Task.Delay(2000).Wait(); // If we hit back before the dialog is fully up, it may not dismiss
```
PureWeen: Could this be *RunningApp.WaitForElement("Cancel");*

Xamarin.Forms.Controls/CoreGalleryPages/CoreGalleryPage.cs
```diff
@@ -111,80 +81,143 @@ protected virtual void Build (StackLayout stackLayout)
...
-			moveNextButton.AutomationId = "MoveNextButton";
-			moveNextButton.Clicked += delegate (object sender, EventArgs e) {
-				if (!_views.Any ())
-					return;
+			_layout = new StackLayout();
 
-				if (_currentIndex + 1 >= _views.Count) {
-					return;
-				}
+			_targetEntry = new Entry { AutomationId = "TargetViewContainer", Placeholder = "Jump To ViewContainer" };
```
PureWeen: 👏 👏 👏 👏 👏 👏

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/GestureBubblingTests.cs
hartez: Cancel doesn't work for older versions of Android - on 19, there's just 'Done'. 
And 'Done' doesn't work for newer versions, which just have 'Cancel' and 'OK'.
That's why we moved to using `Back()` in the first place, which seemed like a much simpler idea until this edge case started popping up. ¯\_(ツ)_/¯

PR #6398, opened 2019-06-04
Adds FontImage Markup Extension for FontImageSource

### LINKED ISSUE #6376: [Enhancement] Add Markup Extension for FontImageSource
## Summary
`Xamarin.Forms.FontImageSource` was completed for v3.6.0 in issue #3189. The original issues discusses creating a markup extension to easily use this new feature in any XAML property that is of type `ImageSource`.

**It appears to have never been implemented**

## API Changes
Simplify the verbose usage of `FontImagesource`

```xaml
<ImageButton>
    <ImageButton.Source>
        <FontImageSource
            FontFamily="{StaticResource MyFontFamily}"
            Glyph="{StaticResource SmileFace}"
            Color="{StaticResource PrimaryColor}" />
    </Image.Source>
</ImageButton>
```

```xaml
<Image Source="{FontImage FontFamily={StaticResource MyFontFamily}, Glyph={StaticResource SmileFace}, Color={StaticResource White}" />
```

# Intended Use Case
For a trivial `Image` this isn't that useful but the power of this markup extension is when another Element has a property of type `ImageSource`. Consider the following real-world example in the new Xamarin.Forms Shell Spec

```xaml
<TabBar>
    <Tab Title="Hello">
        <Tab.Icon>
            <FontImageSource 
                FontFamily="{StaticResource MyFontFamily}"
                Glyph="{StaticResource SmileFace}"
                Color="White" />
        </Tab.Icon>
        <ShellContent ContentTemplate="{DataTemplate views:HelloPage}" />
    </Tab>
    <Tab Title="World">
        <Tab.Icon>
            <FontImageSource 
                FontFamily="{StaticResource MyFontFamily}"
                Glyph="{StaticResource FrownFace}"
                Color="White" />
        </Tab.Icon>
        <ShellContent ContentTemplate="{DataTemplate views:BoxesPage}" />
    </Tab>
</TabBar>
```

After we implement the markup extension the verbose XAML can be simplified removing 6 lines of code with each usage.

```xaml
<TabBar>
    <Tab Title="Hello" Icon="{FontImage FontFamily={StaticResource MyFontFamily}, Glyph={StaticResource SmileFace}, Color={StaticResource White}">
        <ShellContent ContentTemplate="{DataTemplate views:HelloPage}" />
    </Tab>
    <Tab Title="World" Icon="{FontImage FontFamily={StaticResource MyFontFamily}, Glyph={StaticResource FrownFace}, Color={StaticResource White}">
        <ShellContent ContentTemplate="{DataTemplate views:BoxesPage}" />
    </Tab>
</TabBar>
```

ahoefling:
### Description of Change ###
Adds Markup Extension for `FontImageSource` to simplify usage from verbose XAML

```xaml
<ImageButton>
    <ImageButton.Source>
        <FontImageSource
            FontFamily="{StaticResource MyFontFamily}"
            Glyph="{StaticResource SmileFace}"
            Color="{StaticResource PrimaryColor}" />
    </Image.Source>
</ImageButton>
```

```xaml
<ImageButton Source="{FontImage FontFamily={StaticResource MyFontFamily}, Glyph={StaticResource SmileFace}, Color={StaticResource White}" />
```

### Issues Resolved ### 

- fixes #6376

### API Changes ###
Added MarkupExtension: `FontImageExtension` to `Xamarin.Forms.Xaml`

Added:

```c#
string FontFamily { get; set; }
string Glyph { get; set; }
Color Color { get; set; }
```

### Platforms Affected ### 

- Core/XAML (all platforms)
- Any platform that implements `FontImageSource`

None

Not applicable

### Testing Procedure ###
This markup extension has been tested in the Sandbox project and a confidential project I am working on. 

I added positive and negative unit tests in the `Xamarin.Forms.Xaml.UnitTests` project

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Xaml/MarkupExtensionParser.cs
+++ b/Xamarin.Forms.Xaml/MarkupExtensionParser.cs
@@ -24,6 +24,8 @@ namespace Xamarin.Forms.Xaml
 				markupExtension = new OnIdiomExtension();
 			else if (match == "DataTemplate")
 				markupExtension = new DataTemplateExtension();
+			else if (match == "FontImage")
+				markupExtension = new FontImageExtension();
 			else
 			{
 				if (typeResolver == null)
--- a/Xamarin.Forms.Xaml/MarkupExtensionParser.cs
+++ b/Xamarin.Forms.Xaml/MarkupExtensionParser.cs
@@ -24,8 +24,6 @@ namespace Xamarin.Forms.Xaml
 				markupExtension = new OnIdiomExtension();
 			else if (match == "DataTemplate")
 				markupExtension = new DataTemplateExtension();
-			else if (match == "FontImage")
-				markupExtension = new FontImageExtension();
 			else
 			{
 				if (typeResolver == null)
--- a/Xamarin.Forms.Xaml/MarkupExtensions/FontImageExtension.cs
+++ b/Xamarin.Forms.Xaml/MarkupExtensions/FontImageExtension.cs
@@ -3,6 +3,7 @@
 namespace Xamarin.Forms.Xaml
 {
 	[AcceptEmptyServiceProvider]
+	[ContentProperty(nameof(Glyph))]
 	public class FontImageExtension : IMarkupExtension<ImageSource>
 	{
 		public string FontFamily { get; set; }
--- a/Xamarin.Forms.Xaml/MarkupExtensions/FontImageExtension.cs
+++ b/Xamarin.Forms.Xaml/MarkupExtensions/FontImageExtension.cs
@@ -1,0 +1,27 @@
+﻿using System;
+
+namespace Xamarin.Forms.Xaml
+{
+	[AcceptEmptyServiceProvider]
+	public class FontImageExtension : IMarkupExtension<ImageSource>
+	{
+		public string FontFamily { get; set; }
+		public string Glyph { get; set; }
+		public Color Color { get; set; }
+
+		public ImageSource ProvideValue(IServiceProvider serviceProvider)
+		{
+			return new FontImageSource
+			{
+				FontFamily = FontFamily,
+				Glyph = Glyph,
+				Color = Color
+			};
+		}
+
+		object IMarkupExtension.ProvideValue(IServiceProvider serviceProvider)
+		{
+			return (this as IMarkupExtension<ImageSource>).ProvideValue(serviceProvider);
+		}
+	}
+}
--- a/Xamarin.Forms.Xaml/MarkupExtensions/FontImageExtension.cs
+++ b/Xamarin.Forms.Xaml/MarkupExtensions/FontImageExtension.cs
@@ -9,6 +9,7 @@ namespace Xamarin.Forms.Xaml
 		public string FontFamily { get; set; }
 		public string Glyph { get; set; }
 		public Color Color { get; set; }
+		public double Size { get; set; } = 30d;
 
 		public ImageSource ProvideValue(IServiceProvider serviceProvider)
 		{
@@ -16,7 +17,8 @@ namespace Xamarin.Forms.Xaml
 			{
 				FontFamily = FontFamily,
 				Glyph = Glyph,
-				Color = Color
+				Color = Color,
+				Size = Size
 			};
 		}
 
--- a/Xamarin.Forms.Xaml/MarkupExtensions/FontImageExtension.cs
+++ b/Xamarin.Forms.Xaml/MarkupExtensions/FontImageExtension.cs
@@ -9,7 +9,7 @@ namespace Xamarin.Forms.Xaml
 		public string FontFamily { get; set; }
 		public string Glyph { get; set; }
 		public Color Color { get; set; }
-		public double Size { get; set; } = 30d;
+		public double Size { get; set; } = (double)FontImageSource.SizeProperty.DefaultValue;
 
 		public ImageSource ProvideValue(IServiceProvider serviceProvider)
 		{
--- a/Xamarin.Forms.Xaml.UnitTests/FontImageExtension.xaml
+++ b/Xamarin.Forms.Xaml.UnitTests/FontImageExtension.xaml
@@ -7,5 +7,6 @@
 
     <Tab Title="Hello" Icon="{FontImage Size={x:Static local:FontImageExtension.Size}, Color={x:Static local:FontImageExtension.Color}, FontFamily={x:Static local:FontImageExtension.FontFamily}, Glyph={x:Static local:FontImageExtension.Glyph}}" />
     <Tab Title="World" Icon="{FontImage Size={x:Static local:FontImageExtension.Size}, Color={x:Static local:FontImageExtension.Color}, FontFamily={x:Static local:FontImageExtension.FontFamily}, Glyph={x:Static local:FontImageExtension.Glyph}}" />
+    <Tab Title="ContentProperty" Icon="{FontImage MyGlyph, Size={x:Static local:FontImageExtension.Size}, Color={x:Static local:FontImageExtension.Color}, FontFamily={x:Static local:FontImageExtension.FontFamily}}" />
 
 </TabBar>
--- a/Xamarin.Forms.Xaml.UnitTests/FontImageExtension.xaml
+++ b/Xamarin.Forms.Xaml.UnitTests/FontImageExtension.xaml
@@ -5,7 +5,7 @@
     xmlns:local="using:Xamarin.Forms.Xaml.UnitTests"
     x:Class="Xamarin.Forms.Xaml.UnitTests.FontImageExtension">
 
-    <Tab Title="Hello" Icon="{FontImage FontFamily={x:Static local:FontImageExtension.FontFamily}, Glyph={x:Static local:FontImageExtension.Glyph}}" />
-    <Tab Title="World" Icon="{FontImage FontFamily={x:Static local:FontImageExtension.FontFamily}, Glyph={x:Static local:FontImageExtension.Glyph}}" />
+    <Tab Title="Hello" Icon="{FontImage Size={x:Static local:FontImageExtension.Size}, Color={x:Static local:FontImageExtension.Color}, FontFamily={x:Static local:FontImageExtension.FontFamily}, Glyph={x:Static local:FontImageExtension.Glyph}}" />
+    <Tab Title="World" Icon="{FontImage Size={x:Static local:FontImageExtension.Size}, Color={x:Static local:FontImageExtension.Color}, FontFamily={x:Static local:FontImageExtension.FontFamily}, Glyph={x:Static local:FontImageExtension.Glyph}}" />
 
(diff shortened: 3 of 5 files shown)


Xamarin.Forms.Xaml/MarkupExtensionParser.cs
```diff
@@ -24,6 +24,8 @@ public object Parse(string match, ref string remaining, IServiceProvider service
 				markupExtension = new OnIdiomExtension();
 			else if (match == "DataTemplate")
 				markupExtension = new DataTemplateExtension();
+			else if (match == "FontImage")
```
StephaneDelcroix: not required. I'm not even sure we save any measurable time doing this

Xamarin.Forms.Xaml/MarkupExtensions/FontImageExtension.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+
+namespace Xamarin.Forms.Xaml
+{
+	[AcceptEmptyServiceProvider]
```
StephaneDelcroix: ❤️

StephaneDelcroix: I'd add the `Glyph` as `ContentProperty` attribute
```suggestion
	[AcceptEmptyServiceProvider]
	[ContentProperty("Glyph")]
```

ahoefling: This is not my first MarkupExtension and I noticed the other platform ones have `[ContentProperty]` but I am not sure exactly what it is for. How is it used?

Adding this change is not a problem at all

Xamarin.Forms.Xaml/MarkupExtensionParser.cs
ahoefling: I can remove this, no problem

ahoefling:
@StephaneDelcroix I have the Xamarin.Forms.Sandbox project working with this change but didn't commit it. Should I commit that change as well to make it easier for you to see this change in action?

ChaseFlorell:
ContentProperty makes it so that you don't have to define `Glyph=`.

ahoefling:
@samhouts @StephaneDelcroix 
Is there a tipping point on adding too many properties on the markup extension and how much we really gain? I am using this in one of my projects and came to a point where I really wanted to use `Size`. 

I am thinking about adding the property (`Size`) in because it is really simple but want to check with everyone before I do that.

ChaseFlorell:
> @samhouts @StephaneDelcroix
> Is there a tipping point on adding too many properties on the markup extension and how much we really gain? I am using this in one of my projects and came to a point where I really wanted to use `Size`.
> 
> I am thinking about adding the property (`Size`) in because it is really simple but want to check with everyone before I do that.

I think all options need to be there, or the extension isn't fully useful. (side note: default 30 for size is absurd).

I also put together a proposal whereby we could set some defaults in order to keep verbosity down.
https://github.com/xamarin/Xamarin.Forms/issues/6421

ahoefling:
I left it out originally because my target use case was for the Shell `TabBar` but now I am using it in another part of my app and found it really needs `Size`.

StephaneDelcroix:
@ahoefling having all properties exposed adds versatility. that's what we want (see Binding). If all the options were mandatory, that would be verbose, but that's not the case here

ChaseFlorell:
I gave this a try and got a strange result.

```xml
<!-- works good -->
<ImageButton Source="{FontImage FontFamily={StaticResource MyFontFamily}, Glyph={x:Static xml:Images.SmileFace}, Color={StaticResource White}" />
```
but
```xml
<!-- doesn't properly set Glyph -->
<ImageButton Source="{FontImage {x:Static xml:Images.SmileFace}, FontFamily={StaticResource MyFontFamily}, Color={StaticResource White}" />
```

In other words, the `ContentProperty` doesn't pick up the path the same way

ahoefling:
Could it be because we are using 

```c#
[ContentProperty[nameof(Glyph)]
```

instead of

```c#
[ContentProperty("Glyph")]
```

I don't think this would cause that, but I don't know much about how the `ContentProperty` works except what we discussed earlier

ahoefling:
I just added more unit tests to cover this and did some manually testing as well. It appears that `[ContentProperty]` is not properly picking up the markup extension of `x:Static` in the usage you demonstrated (@ChaseFlorell). When I attached the debugger to the unit test is shows the string literal of
```
{x:Static xml:Images.SmileFace
```

Yes, that is correct and it is missing the trailing `}` (curly brace). I then updated the code to pass in a hardcoded Glyph and it all appears to work. 

@ChaseFlorell I think you discovered a bug with how the `[ContentProperty]` works with nesting usage of markup extensions.

ChaseFlorell:
I'm going to be submitting a proposal today that will require that the `[ContentProperty(nameof(Glyph))]` not be included on this markup extension.

Once the proposal is in, I will link back to this for further discussion.

ahoefling:
After running several tests today I don't see a lot of value in `[ContentProperty(nameof(Glyph)]` since you can't use a markup extension with it to access your static reference. https://github.com/xamarin/Xamarin.Forms/pull/6398#issuecomment-499530535. 

I'm recommending that we remove the `[ContentProperty(nameof(Glyph))]` attribute so we can add a more useful `[ContentProperty]` in the future that won't be a breaking change.

Xamarin.Forms.Xaml/MarkupExtensions/FontImageExtension.cs
```diff
@@ -0,0 +1,30 @@
...
+	[ContentProperty(nameof(Glyph))]
+	public class FontImageExtension : IMarkupExtension<ImageSource>
+	{
+		public string FontFamily { get; set; }
+		public string Glyph { get; set; }
+		public Color Color { get; set; }
+		public double Size { get; set; } = 30d;
```
StephaneDelcroix: not sure about setting a default. does FontImageSource requires a Size ?

ChaseFlorell: Size already defaults to 30 on FontImageSource, but if you don't set the default here, it'll revert to `0`.

Perhaps you can use `FontImageSource.SizeProperty.DefaultValue`

ChaseFlorell: alternatively, I don't know if FontImageSource has a concept of a magic value `-1` to roll over to default.

ahoefling: My thought process behind this is the default size for `FontImageSource` is `30d`. If someone uses this markup extension it will default to 0 unless they explicitly provide a size. 

Perhaps we add a static default in `FontImageSource` and that way both the markup extension and `FontImageSource` can reference the default size.

**FontImageSource**
```c#
public class FontImageSource : ImageSource
{
    // ...
    public static double DefaultSize = 30d;
    // ...
}
```

**FontImageExtension**
```c#
public class FontImageExtension : IMarkupExtension<ImageSource>
{
    // ...
    public double Size { get; set; } = FontImageSource.DefaultSize;
    // ...
}
```

ChaseFlorell: You can access it directly off of the SizeProperty

```csharp
public double Size { get; set; } = (double)FontImageSource.SizeProperty.DefaultValue;
```

ahoefling: That is a great suggestion, I'll go an update that to use the `FontImageSource.SizeProperty.DefaultValue`

StephaneDelcroix:
The ContentProperty has to be a literal, not another markup extension. I don't think it's bug

ahoefling:
Is there a reason why ContentProperty doesn't support MarkupExtensions? Is this something that should be included in a proposal and feature request?

StephaneDelcroix:
> Is there a reason why ContentProperty doesn't support MarkupExtensions? Is this something that should be included in a proposal and feature request?

Here are the reasons I can think of:
1. the need never showed up
2. I have no idea if it's supported in other XAML variants,
3. It would be very-very odd to write `{Binding {StaticResource path}}`

Now, if `2.` is false and the syntax is supported in WPF or UWP (and it's supported by the XLS (/cc @mgoertz-msft)) we could consider it, but we're not going to make a precedent because of `1.`

I suggested adding a ContentProperty not to support StaticResource for the Glyph, but for the cases when you prefer to use the glyph code instead of a resource.

ahoefling:
@StephaneDelcroix thanks for the detailed explanation, I am not as familiar with the reasoning behind it and wanted to understand better.

At this point I am in agreement with adding `[ContentProperty(nameof(Glyph))]` because it requires a string literal and we can't use a markup extension. Considering the proposal that @ChaseFlorell is working on I don't think ContentProperty would create any breaking changes as we would have to use the declarative syntax of

```xaml
 <ImageButton Source="{FontImage Glyph={x:Static xml:Images.SmileFace}}" />
```

The ContentProperty is added here as a convenience utility for people that want to use a string literal

jamesmontemagno:
Question...
It looks like this doesn't work with DynamicResources in the FontFamily?

```xml
Source="{FontImage FontFamily={DynamicResource MaterialFontFamily},
                                    Glyph={StaticResource IconUpNext}, 
                                    Color={StaticResource PrimaryColor},
                                    Size=15}"
```

```xml
            <OnPlatform x:Key="MaterialFontFamily" x:TypeArguments="x:String">
                <On Platform="iOS" Value="Material Design Icons" />
                <On Platform="Android" Value="materialdesignicons-webfont.ttf#Material Design Icons" />
                <On Platform="UWP" Value="Assets/Fonts/materialdesignicons-webfont.ttf#Material Design Icons" />
            </OnPlatform>
```

Compiler gives error about FontFamily if I chagne to StaticResource it works. I guess it didn't need to be a Dynamic Resource?

StephaneDelcroix:
the extension is fully resolved at the time it's parsed, that's why DynamicResource won't work

jamesmontemagno:
Gotcha... any way to solve this to clean it up? Suggestions?

jamesmontemagno:
Something like: 

```xml
FontFamily={OnPlatform 
                        iOS={StaticResource MaterialFontFamilyiOS},
                        Android={StaticResource MaterialFontFamilyAndroid}}
``` 

?

StephaneDelcroix:
@jamesmontemagno don't use the markup extension, it doesn't provide a lot of benefits as it doesn't save a lot of precious keystrokes

jamesmontemagno:
Yeah, but it looks better! Can i define a StaticResource and update it in the App.xaml.cs?

jamesmontemagno:
NM it all works with StaticResource in place :) yay... for some reason Xamarin.Forms didn't update when I updated Xamarin.Forms.Visual.Material... weird.

PR #6405, opened 2019-06-04
[iOS] Fix Bounds for PageContainer 

### LINKED ISSUE #6334: [Bug] iOS effect no longer works after upgrading to XF 4.0 since UIView.Bounds has no size in PlatformEffect.OnAttached()
### Description
I created an effect to add a gradient background to a UI element that no longer works in iOS after upgrading to XF 4.0 (the effect continued to work on Android). The cause of the issue is that the `UIView.Bounds` property started returning a `CGRect` with no size in the `PlatformEffect.OnAttached()` method.

I am working around this issue by rolling back to XF 3.6.

### Steps to Reproduce

1. Create an iOS `PlatformEffect` and apply it to a `ContentPage`.
2. Set breakpoint in `OnAttached()` method.

### Expected Behavior
`Container.Bounds` should have the same size as the `ContentPage`.

### Actual Behavior
`Container.Bounds` has no size (i.e. height and width are both 0).

### Basic Information

- Last known good version: 3.6.0.344457
- IDE: VisualStudio

### LINKED ISSUE #6372: [Bug] [CustomRenderer] [iOS] View.Bounds all 0 OnElementChanged
### Description
Custom Renderer for a grandient background stopped working when upgrading to Xamarin Forms 4.x.x (from 3.6.x). View.Bounds width, height all 0

### Steps to Reproduce

1. Create custom iOS renderer deriving from PageRenderer class
2. Add override to OnElementChanged
3. Add code to render a grandient background

```
base.OnElementChanged(e);

            if (e.OldElement == null) // perform initial setup
            {
                var page = e.NewElement as GradientContentPage;

                var gradientLayer = new CAGradientLayer
                {
                    Frame = View.Bounds,
                    Colors = new CGColor[] { page.StartColor.ToCGColor(), page.EndColor.ToCGColor() }
                };

                View.Layer.InsertSublayer(gradientLayer, 0);
            }

```
### Expected Behavior
View.Bounds hat the actual width and height of the view. Adding a new CAGradientLayer used View.Bounds and had the current view's size.

### Actual Behavior
View.Bounds are all 0 (width, height) and thus the new Layer is not displayed (means: it's width and height are both 0)

### Basic Information

- Last known good version: 3.6.x
- IDE: Visual Studio for Mac 8.8.8 (build 2)
- Platform Target Frameworks: 
  - iOS:  12.2
- Nuget Packages: FFImageLoading
- Affected Devices: iPhone

rmarinho:
### Description of Change ###

We introduced a wrapper PageContainer and doing so we override the default LoadView of our PageRenderer ViewController. 

The default method provided a UIView with a Frame and Bounds, while our PageContainer doesn't provide one when it's created. Since our effects are attached when we set the elements for reuse on renderers when the user access the Bounds of the container there are not set ,while in previous versions there were (even if wrong ) .
Since these are related with size even previous versions should be using not the attach only but also the OnElementPropertyChanged or other methods to update there frame if the size changes. 

After some testing seems that the default size that was provided was the UIScreen even if inside a NavigationPage, this also is not very important as the Frame will be updated later on chain ViewWillLayoutSubviews.
### Issues Resolved ### 

- fixes #6334
- fixes #6372

### API Changes ###

 
 None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###

Should visit Issue 6334 and see a gradient and Success message.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6334.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6334.cs
@@ -38,7 +38,7 @@ namespace Xamarin.Forms.Controls.Issues
 		[Test]
 		public void Issue6334Test() 
 		{
-			RunningApp.WaitForElement (q => q.Marked (AutomationId));
+			RunningApp.WaitForElement (q => q.Marked ("IssuePageLabel"));
 			RunningApp.WaitForElement(q => q.Marked(Success));
 			RunningApp.Screenshot ("I see the gradient");
 		}
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6334.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6334.cs
@@ -1,0 +1,47 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.Forms.Core.UITests;
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if UITEST
+	[Category(UITestCategories.ManualReview)]
+#endif
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 6334, "iOS effect no longer works after upgrading to XF 4.0 since UIView.Bounds has no size", PlatformAffected.iOS)]
+	public class Issue6334 : TestContentPage
+	{
+		public const string EffectName = "GradientEffect";
+		public const string Success = "Success";
+		public const string Fail = "Fail";
+		
+		protected override void Init()
+		{
+			BackgroundColor = Color.Blue;
+			var effect = Effect.Resolve($"{Issues.Effects.ResolutionGroupName}.{EffectName}");
+
+			Effects.Add(effect);
+
+			Content = new Label
+			{
+				AutomationId = "IssuePageLabel",
+				Text = Fail
+			};
+		}
+
+#if UITEST && __IOS__
+		[Test]
+		public void Issue6334Test() 
+		{
+			RunningApp.WaitForElement (q => q.Marked (AutomationId));
+			RunningApp.WaitForElement(q => q.Marked(Success));
+			RunningApp.Screenshot ("I see the gradient");
+		}
+#endif
+	}
+}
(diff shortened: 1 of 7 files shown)


Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
```diff
@@ -123,8 +123,9 @@ public void SetElementSize(Size size)
 
 		public override void LoadView()
 		{
+			//by default use the MainScreen Bounds so Effects can access the Container size
 			if (_pageContainer == null)
-				_pageContainer = new PageContainer(this);
+				_pageContainer = new PageContainer(this) { Frame = UIScreen.MainScreen.Bounds};
```
rmarinho: I tested using the base.LoadView and see size created was the same as UIScreen.Bounds even inside a NavigationPage.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6334.cs
```diff
@@ -0,0 +1,47 @@
...
+			Content = new Label
+			{
+				AutomationId = "IssuePageLabel",
+				Text = Fail
+			};
+		}
+
+#if UITEST && __IOS__
+		[Test]
+		public void Issue1Test()
```
samhouts:
```suggestion
		public void Issue334Test() 
```

rmarinho:
```suggestion
		public void Issue6334Test() 
```

Xamarin.Forms.Platform.iOS/ViewRenderer.cs
```diff
@@ -40,7 +40,7 @@ public abstract class ViewRenderer<TView, TNativeView> : VisualElementRenderer<T
 		event EventHandler _controlChanging;
 		event EventHandler _controlChanged;
 
-
+		bool _isElementOrControlEmpty => (Element == null || Control == null);
```
paymicro:
```suggestion
		bool _isElementOrControlEmpty => Element == null || Control == null;
```

rmarinho: Sorry going to remove this to other PR so we can better track it

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6334.cs
```diff
@@ -0,0 +1,47 @@
...
+				AutomationId = "IssuePageLabel",
+				Text = Fail
+			};
+		}
+
+#if UITEST && __IOS__
+		[Test]
+		public void Issue6334Test() 
+		{
+			RunningApp.WaitForElement (q => q.Marked (AutomationId));
```
rmarinho:
```suggestion
			RunningApp.WaitForElement (q => q.Marked ("IssuePageLabel"));
```

PR #6422, opened 2019-06-05
Implement ItemSizingStrategy.MeasureFirstItem on Android

hartez:
### Issues Resolved ### 

- implements part of #3172

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

None

Not applicable

### Testing Procedure ###

Control Gallery, navigate to CollectionView Gallery ->Item Size Galleries ->ItemSizing Strategy

The initial selection is MeasureFirstItem; all of the items should have the same size. If you switch to MeasureAllItems, the items will have varying sizes.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/CollectionView/ItemsViewAdapter.cs
+++ b/Xamarin.Forms.Platform.Android/CollectionView/ItemsViewAdapter.cs
@@ -2,9 +2,9 @@ using System;
 using Android.Content;
 using Android.Support.V7.Widget;
 using Android.Widget;
-using AView = Android.Views.View;
 using Object = Java.Lang.Object;
 using ViewGroup = Android.Views.ViewGroup;
+using ASize = Android.Util.Size;
 
 namespace Xamarin.Forms.Platform.Android
 {
@@ -17,6 +17,7 @@ namespace Xamarin.Forms.Platform.Android
 		readonly Func<View, Context, ItemContentView> _createItemContentView;
 		internal readonly IItemsViewSource ItemsSource;
 		bool _disposed;
+		ASize _size;
 
 		internal ItemsViewAdapter(ItemsView itemsView, Func<View, Context, ItemContentView> createItemContentView = null)
 		{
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Platform.Android/CollectionView/ItemsViewAdapter.cs
```diff
@@ -65,10 +79,21 @@ public override RecyclerView.ViewHolder OnCreateViewHolder(ViewGroup parent, int
...
+		ItemContentView CreateItemContentView(Context context)
+		{
+			if(ItemsView.ItemSizingStrategy == ItemSizingStrategy.MeasureAllItems
+				|| _size == null)
+			{
+				return new ItemContentView(context);
+			}
+
+			return new ItemContentView(context);
+		}
```
paymicro: What is the meaning of this method if it always returns `new ItemContentView(context)`?

hartez: Leftover from an earlier implementation. I'll remove it.

hartez:
> Also why does the strategy MeasureAllItems ignore the first 2 elements?

The strategy doesn't ignore them, the test does. The template for it has an IValueConverter which forces the first few items to have a different width/height than the rest. That way, we can see a difference between measuring the first item and using that size, or measuring all the items individually.

hartez:
> After changing the ItemSizingStrategy to MeasureFirstItem, the images inside have a wrong vertical offset.

Huh. That is _weird_. I'll take a look at that on Monday.

paymicro:
> Huh. That is _weird_. I'll take a look at that on Monday.

I think that this is due to a change in the position of the entire CollectionView. As the text above becomes 2 lines.

hartez:
> > Huh. That is _weird_. I'll take a look at that on Monday.
> 
> I think that this is due to a change in the position of the entire CollectionView. As the text above becomes 2 lines.

I think you're right. 

Well, I don't think it's reason enough to hold this back from getting merged. We can always log that as a separate bug in the future if we need to.

PR #6437, opened 2019-06-06
[Android] Fix context actions icon so that it loads

### LINKED ISSUE #6258: [Bug] [Android] ContextActions icon not working
### Description
Icons are not displayed for ContextActions on Android

Looks like issue was introduced in https://github.com/xamarin/Xamarin.Forms/pull/6062

### Steps to Reproduce

1. Run control gallery app from Xamarin.Forms solution
2. Find "ContextActions List Gallery" page and open it
3. Hold any item to open context actionmenu
4. No icon

### Expected Behavior
Calculator icon displayed for first item

### Actual Behavior
No icon displayed

### Basic Information

- Version with issue: 4.0 release
- Last known good version: 4.0-pre10

PureWeen:
### Description of Change ###
- Images loaded from icons don't have a renderer to work from so needed to update the image loading code to work if no renderer is specified
- Fixed menuitem/toolbaritem to correctly set Automation Id using appcompat 

### Breaking changes ###
On Android the AutomationId for Toolbaritems will now set the ContentDescription which makes the UI tests fall more inline with iOS. 

### Issues Resolved ### 
- fixes #6258 

### Platforms Affected ### 
- Android

### Testing Procedure ###
Run the included UI Test

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core.UITests.Shared/Tests/ToolbarItemTests.cs
+++ b/Xamarin.Forms.Core.UITests.Shared/Tests/ToolbarItemTests.cs
@@ -106,7 +106,7 @@ namespace Xamarin.Forms.Core.UITests
 			var btn2 = App.Query(c => c.Marked(btn4Id))[0];
 			Assert.False(btn2.Enabled, "Toolbar Item  should be disable");
 #else
-			var btn1 = App.Query(c => c.Marked(btn1Id))[0];
+			var btn1 = App.WaitForElement(c => c.Marked(btn1Id))[0];
 			ShouldShowMenu();
 			//var btn2 = App.Query (c => c.Marked (btn4Id)) [0];
 			//TODO: how to check Enable for the textview
--- a/Xamarin.Forms.Core.UITests.Shared/Tests/ToolbarItemTests.cs
+++ b/Xamarin.Forms.Core.UITests.Shared/Tests/ToolbarItemTests.cs
@@ -11,11 +11,11 @@ namespace Xamarin.Forms.Core.UITests
 	[Category(UITestCategories.UwpIgnore)]
 	internal class ToolbarItemTests : BaseTestFixture
 	{
-		string btn1Id = "tb1";
-		string btn2Id = "tb2";
-		string btn4Id = "tb4";
+		string btn1Id = "toolbaritem_primary";
+		string btn2Id = "toolbaritem_primary2";
+		string btn4Id = "toolbaritem_secondary2";
 #if !__MACOS__
-		string btn3Id = "tb3";
+		string btn3Id = "toolbaritem_secondary";
 #endif 
 
 #if __ANDROID__
@@ -45,15 +45,6 @@ namespace Xamarin.Forms.Core.UITests
 		protected override void NavigateToGallery()
 		{
 			App.NavigateToGallery(GalleryQueries.ToolbarItemGallery);
-#if __IOS__
-			btn1Id = "menuIcon";
-			btn4Id = "tb4";
-			if (AppSetup.iOSVersion  >= 9)
-			{
-				btn1Id = "toolbaritem_primary";
-				btn4Id = "toolbaritem_secondary2";
-			}
-#endif
 		}
 
 		[Test]
@@ -75,9 +66,7 @@ namespace Xamarin.Forms.Core.UITests
 		public void ToolbarButtonsCommand()
 		{
 			ShouldShowMenu();
-#if __ANDROID__
-			//App.Query (c => c.Marked (btn4Id))[0];
-#else
+
 			App.WaitForElement(btn4Id);
 			App.Tap(c => c.Marked(btn4Id));
 			App.WaitForNoElement(c => c.Text("button 4 new text"));
@@ -86,13 +75,18 @@ namespace Xamarin.Forms.Core.UITests
 #else
 			App.Tap(c => c.Marked(btn3Id));
 #endif
+#if __ANDROID__
+			ShouldShowMenu();
+#endif
 			App.Tap(c => c.Marked(btn4Id));
 			App.WaitForElement(c => c.Text("button 4 new text"));
 #if __MACOS__
 			App.Tap(c => c.Button().Index(6));
 #else
-			App.Tap(c => c.Marked(btn3Id));
+#if __ANDROID__
+			ShouldShowMenu();
 #endif
+			App.Tap(c => c.Marked(btn3Id));
 #endif
 		}
 
--- a/Xamarin.Forms.Core.UITests.Shared/Tests/ToolbarItemTests.cs
+++ b/Xamarin.Forms.Core.UITests.Shared/Tests/ToolbarItemTests.cs
@@ -20,25 +20,27 @@ namespace Xamarin.Forms.Core.UITests
 #endif 
 
 #if __ANDROID__
-		static bool isSecondaryMenuOpen = false;
+		bool isSecondaryMenuOpen()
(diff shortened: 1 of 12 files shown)


Xamarin.Forms.Platform.Android/Extensions/AccessibilityExtensions.cs
```diff
@@ -102,20 +103,21 @@ public static string SetNavigationContentDescription(this global::Android.Suppor
...
 			var elemValue = ConcatenateNameAndHint(Element);
 
+			if (string.IsNullOrWhiteSpace(elemValue))
+				elemValue = Element.AutomationId;
+
 			if (!string.IsNullOrWhiteSpace(elemValue))
-				Control.SetTitle(elemValue);
+				AMenuItemCompat.SetContentDescription(Control, elemValue);
 
 			return;
```
paymicro:
```suggestion
```

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -196,10 +195,10 @@ internal static Task ApplyDrawableAsync(this IShellContext shellContext, Bindabl
 								return;
 
 							// we are back, so update the working element
-							element = bindable ?? renderer.Element;
+							element = bindable ?? renderer?.Element;
 
 							// makse sure things are good now that we are back
-							if (element == null || renderer.View == null)
+							if (element == null || (renderer != null && renderer.View == null))
```
paymicro: In line 176
```C#
if (element == null || renderer?.View == null)
```
I think it's worth doing the same in 2 places.

Xamarin.Forms.Platform.Android/Extensions/AccessibilityExtensions.cs
```diff
@@ -102,20 +103,21 @@ public static string SetNavigationContentDescription(this global::Android.Suppor
 
 		public static void SetTitleOrContentDescription(this IMenuItem Control, ToolbarItem Element)
 		{
-			if (Element == null)
-				return;
+			SetTitleOrContentDescription(Control, (MenuItem)Element);
+		}
```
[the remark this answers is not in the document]
PureWeen: The main reason I left it in was for the possibilities of an ABI break. I realize that technically I don't need it but it would change the public api so I left it in.

Xamarin.Forms.Platform.Android/ResourceManager.cs
PureWeen: The reason I didn't do it with the operator is that I don't want a null renderer to return here.

If the renderer is null then this logic doesn't apply. This logic only applies if there is a renderer being passed in

Xamarin.Forms.Platform.Android/ResourceManager.cs
```diff
@@ -174,10 +173,10 @@ internal static Task ApplyDrawableAsync(this IShellContext shellContext, Bindabl
 							return;
 
 						// we are back, so update the working element
-						element = bindable ?? renderer.Element;
+						element = bindable ?? renderer?.Element;
 
 						// makse sure things are good now that we are back
-						if (element == null || renderer.View == null)
+						if (element == null || renderer?.View == null)
```
PureWeen:
```suggestion
						if (element == null || (renderer != null && renderer.View == null))
```

Xamarin.Forms.Platform.Android/ResourceManager.cs
PureWeen: But that does mean I should apply the same thing at line 176 :-p

hartez:
@PureWeen Looks like this is consistently failing the ToolbarItemsTests on all API levels.

Xamarin.Forms.Platform.Android/CellAdapter.cs
```diff
@@ -334,5 +338,12 @@ bool OnPrepareActionModeImpl(IMenu menu)
 
 			return false;
 		}
+
+
+		protected override void Dispose(bool disposing)
```
hartez: Since this doesn't actually do anything in the Dispose override, do we need it? I realize it's checking `_isDisposed`, but doesn't the `IsDisposed()` extension cover that?

PureWeen: I wasn't sure if it did. IsDisposed() checks the Handle which I wasn't sure if that gets synchronously set to Zero when Dispose is called or not. But I checked with @jonathanpeppers and it looks like just checking if the handle is Zero is the same thing

https://github.com/xamarin/xamarin-android/blob/75e053e5eff431093c7e00d94962fcd13683be5c/src/Mono.Android/Java.Lang/Object.cs#L201-L241

PureWeen:
@hartez alright I fixed the UI Tests for toolbar items and also added an extra dismiss keyboard call since those same tests are failing elsewhere

I also added @samhouts to review the breaking changes to see if it'll cause accessibility issues

Xamarin.Forms.Platform.Android/CellAdapter.cs
hartez: Yeah, the `IsDisposed` method was just added to make the intent more explicit; otherwise we end up with tons of `Handle == IntPtr.Zero` checks sprinkled about.

Xamarin.Forms.Core.UITests.Shared/Tests/ToolbarItemTests.cs
```diff
@@ -11,11 +12,11 @@ namespace Xamarin.Forms.Core.UITests
 	[Category(UITestCategories.UwpIgnore)]
```
rmarinho: I think we might disable these tests for now, remove the asserts, take screenshots and mark as manual review

samhouts:
@PureWeen I am okay with the breaking change, since this will "fix" `Hint` on ToolbarItems. However, we need to make the behavior consistent across platforms. On iOS and UWP, it will read the `Text`, then the `Hint`. Before this change, it just read the `Text`. With this change, it just reads the `Hint`. When that is fixed, you'll want to update the test. https://github.com/xamarin/Xamarin.Forms/blob/37c22eb1cb64afc5d62b9386a50bb6090446a65a/Xamarin.Forms.Controls/ControlGalleryPages/AutomationPropertiesGallery.cs#L48

PureWeen:
@samhouts try it now

Xamarin.Forms.Platform.Android/FastRenderers/AutomationPropertiesProvider.cs
```diff
@@ -46,17 +46,23 @@ internal static void SetAutomationId(AView control, VisualElement element, strin
...
 				defaultContentDescription = control.ContentDescription;
 
-			string value = ConcatenateNameAndHelpText(element);
-			control.ContentDescription = !string.IsNullOrWhiteSpace(value) ? value : defaultContentDescription;
+			string value = ConcatenateNameAndHelpText(bindableObject);
+
+			var contentDescription = !string.IsNullOrWhiteSpace(value) ? value : defaultContentDescription;
+
+			if (String.IsNullOrWhiteSpace(contentDescription) && bindableObject is Element element)
+				contentDescription = element.AutomationId;
```
PureWeen: @samhouts is this ok? It doesn't seem like the *ContentDescription* should get overwritten here if the user has set the AutomationId to something

samhouts: works! ship it

PR #6457, opened 2019-06-07
[ios] Shell inset,  general SafeArea fixes for shell and non shell embedded pages, entry scrolling fixes

### LINKED ISSUE #5239: Top Padding not working on iOS when it is set alone.
### Description
Xamarin.Forms content page top padding is not working if it is set alone. That is if the other padding values are zero the top padding will not work. Setting one of the other values to a non zero value together with the top padding, the top padding will work.

### Steps to Reproduce

1. Create new Xamarin.Forms application
2. Delete the main page and add a new code only main page of type content page.
3. Create a label
Label label = new Label { Text = "Hello Xamarin World!" };
4. Set the page padding
Padding = new Thickness(0, 20, 0, 0);
5. set the label as the content property of the page
Content = label;

### Expected Behavior
The label should have been displayed 20 units below the iOS status bar

### Actual Behavior
The label is overlapping with the iOS status bar

### Basic Information

- Version with issue: 3.5.0.129452
- Last known good version: I don't remeber (but I guess it was working fine on the earlier version of 3.2.0.871581
- IDE: Visual Studio 2017
- Platform Target Frameworks: 
  - iOS:  
- Nuget Packages: NetStandardLibrary 2.0.3
- Affected Devices: 

### Screenshots
image

### LINKED ISSUE #6424: Shell: CollectionView or ListView content in a top tab ShellContent page is obscured behind top tabs on iOS
### Description
CollectionView or ListView content in a ShellContent's page is obscured behind the tabs on iOS. It appears that perhaps the top insets are incorrect. Video of the behavior: https://www.youtube.com/watch?v=_NuOoiVpHrw

Issue exists only in iOS, not Android.

Wrapping the CollectionView or ListView in a ScrollView appears to fix the insets problem, but then pull-to-refresh doesn't work because the ScrollView receives the gestures, not the nested ListView. (Not an issue for CollectionView yet, of course, because CollectionView doesn't yet support pull-to-refresh.)

I only included CollectionView in my repro sample.

### Steps to Reproduce

Build and run this sample repo:
https://github.com/jsauve/Xamarin.Forms.4.0.0.482894_TopTabs-CollectionView-Insets-Bug

### Expected Behavior

The top of the CollectionView or ListView is not obscured behind the top tabs.

### Actual Behavior

The top of the CollectionView or ListView is obscured behind the top tabs.

### Basic Information

My local dev environment details:
https://gist.github.com/jsauve/6eb67803db786def5d8f8c74b308f95d

### Screenshots

Video:
https://www.youtube.com/watch?v=_NuOoiVpHrw

### Reproduction Link

Build and run this sample repo:
https://github.com/jsauve/Xamarin.Forms.4.0.0.482894_TopTabs-CollectionView-Insets-Bug

### LINKED ISSUE #6499: [Bug] Shell - First page in AppShell ignores the UseSafeAreaInset flag
### Description

Shell does not seem to respect the `UseSafeAreaInset` flag and overrides the `Padding` of the first `ContentPage` in a tab or flyout.

### Steps to Reproduce

1. Run the sample project below
2. See the `ContentPage`'s light purple background above and below the lavendar `StackLayout`.
3. Click on the "Next Screen" button
4. See the the second screen does not show the `ContentPage`'s background.  You can only see the `StackLayout`'s background color.

### Expected Behavior

The `ContentPage`'s content should take up the whole screen.  In the example, `StackLayout` should take up the whole screen.

### Actual Behavior

The `ContentPage` has `Padding` set to the `SafeAreaInset` values.  This means that the content inside the page does not take up the whole screen.

### Basic Information

- Version with issue: 4.0.0.482894
- IDE: Visual Studio 2019 for Windows 16.1.2
- Platform Target Frameworks: 
  - iOS:  12.10.0.153
  - Android: N/A
  - UWP:  N/A
- Android Support Library Version: N/A
- Nuget Packages: Just what comes with the template
- Affected Devices:

### Screenshots

2019-06-11_03-30-41-PM
2019-06-11_03-30-44-PM

### Reproduction Link

https://github.com/seansparkman/ShellNavigationExample/tree/use-safe-area-ignored
https://github.com/seansparkman/ShellNavigationExample/blob/use-safe-area-ignored/ShellNavigationRegistration/ShellNavigationRegistration/Views/OnBoarding.xaml#L13

PureWeen:
### Description of Change ###

- Fix iOS so the content pads correctly for: pages with top tabs, all pages not just scroll views
- Fixed the keyboard focus auto scroll so that when you click off the entry it'll scroll back correctly
- fixed safe area so that if you turn it off content will set to 0,0 (this also applies to page with top tabs)
- fixed safe area padding to include top tabs size as well
- consolidated all the safe area setting logic to a single method so it's easy to follow the logic of why we set it when

### Issues Resolved ### 
- fixes #5239
- fixes #6424
- fixes #6499

### Platforms Affected ### 
- iOS

### Breaking ###

- prior to this PR if the user set the padding, set safearea to true, then set the safearea to false it would reset the padding to zero. After this PR it'll set the padding back to whatever the user initially set it to

### Testing Procedure ###
I added a ShellInsets test that has some automation but you can just click around on things to see if it all works to your liking as well

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/ShellInsets.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/ShellInsets.cs
@@ -25,6 +25,10 @@ namespace Xamarin.Forms.Controls.Issues
 #endif
 	public class ShellInsets : TestShell
 	{
+		const string EntryTest = "EntryTest";
+		const string EntryToClick = "EntryToClick";
+		const string EntrySuccess = "EntrySuccess";
+
 		protected override void Init()
 		{
 			SetupLandingPage();
@@ -43,7 +47,8 @@ namespace Xamarin.Forms.Controls.Issues
 					new Button()
 					{
 						Text = "Entry Inset",
-						Command = new Command(() => EntryInset())
+						Command = new Command(() => EntryInset()),
+						AutomationId = EntryTest
 					},
 					new StackLayout()
 					{
@@ -141,7 +146,7 @@ namespace Xamarin.Forms.Controls.Issues
 				{
 					Children =
 						{
-							new Label(){ VerticalOptions= LayoutOptions.FillAndExpand, Text = "Click the entry and it should scroll up and stay visible. Click off entry and this label should still be visible"},
+							new Label(){ AutomationId = EntrySuccess, VerticalOptions= LayoutOptions.FillAndExpand, Text = "Click the entry and it should scroll up and stay visible. Click off entry and this label should still be visible"},
 							new Button(){ Text = "Top Tab", Command = new Command(() => AddTopTab("top"))},
 							new Button(){ Text = "Bottom Tab", Command = new Command(() => AddBottomTab("bottom"))},
 							new Button(){ Text = "Change Navbar Visible", Command = new Command(() => Shell.SetNavBarIsVisible(view.Parent, !(Shell.GetNavBarIsVisible(view.Parent))))},
@@ -152,6 +157,9 @@ namespace Xamarin.Forms.Controls.Issues
 							},
 							new Button(){Text = "Reset", Command = new Command(() => SetupLandingPage() )},
 							new Entry()
+							{
+								AutomationId = EntryToClick
+							}
 						}
 				}
 			};
@@ -160,10 +168,16 @@ namespace Xamarin.Forms.Controls.Issues
 		}
 
 
-#if UITEST
+#if UITEST && __IOS__
 		[Test]
-		public void UpdatingSourceOfDisposedListViewDoesNotCrash()
+		public void EntryScrollTest()
 		{
+			RunningApp.Tap(EntryTest);
+			RunningApp.Tap(EntryToClick);
+			RunningApp.WaitForNoElement(EntrySuccess);
+			RunningApp.TapCoordinates(0, 0);
+			RunningApp.WaitForElement(EntrySuccess);
+
 		}
 #endif
 	}
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/ShellInsets.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/ShellInsets.cs
@@ -28,6 +28,18 @@ namespace Xamarin.Forms.Controls.Issues
 		const string EntryTest = "EntryTest";
 		const string EntryToClick = "EntryToClick";
 		const string EntrySuccess = "EntrySuccess";
+		const string Reset = "Reset";
+
+		const string ToggleSafeArea = "ToggleSafeArea";
+		const string SafeAreaTest = "SafeAreaTest";
+		const string SafeAreaTopLabel = "SafeAreaTopLabel";
+		const string SafeAreaBottomLabel = "SafeAreaBottomLabel";
+
+		const string ListViewTest = "ListViewTest";
+
+		const string PaddingTest = "PaddingTest";
+		const string PaddingEntry = "PaddingEntry";
+		const string PaddingLabel = "PaddingLabel";
 
 		protected override void Init()
 		{
@@ -37,8 +49,8 @@ namespace Xamarin.Forms.Controls.Issues
 		void SetupLandingPage()
 		{
 			var page = CreateContentPage();
-			CheckBox cbox = new CheckBox();
-			Entry entryPadding = new Entry() { WidthRequest = 150 };
+			CheckBox cbox = new CheckBox() { AutomationId = ToggleSafeArea };
+			Entry entryPadding = new Entry() { WidthRequest = 150, AutomationId = PaddingEntry };
 
 			page.Content = new StackLayout()
 			{
@@ -53,7 +65,8 @@ namespace Xamarin.Forms.Controls.Issues
 					new Button()
 					{
 						Text = "List View Scroll Test",
-						Command = new Command(() => ListViewPage())
+						Command = new Command(() => ListViewPage()),
+						AutomationId = ListViewTest
 					},
 					new StackLayout()
 					{
@@ -64,7 +77,8 @@ namespace Xamarin.Forms.Controls.Issues
 							new Button()
 							{
 								Text = "Safe Area",
-								Command = new Command(() => SafeArea(cbox.IsChecked))
+								Command = new Command(() => SafeArea(cbox.IsChecked)),
+								AutomationId = SafeAreaTest
 							}
 						},
 						HorizontalOptions = LayoutOptions.Center
@@ -78,7 +92,8 @@ namespace Xamarin.Forms.Controls.Issues
 							new Button()
 							{
 								Text = "Padding",
-								Command = new Command(() => PaddingPage(entryPadding.Text))
+								Command = new Command(() => PaddingPage(entryPadding.Text)),
+								AutomationId = PaddingTest
 							}
 						},
 						HorizontalOptions = LayoutOptions.Center
@@ -100,8 +115,10 @@ namespace Xamarin.Forms.Controls.Issues
 				ItemTemplate = new DataTemplate(() =>
 				{
 					ViewCell cell = new ViewCell();
-					var label = new Label() { Text = " I am a label" }; ;
+					var label = new Label() { Text = " I am a label" };
 					label.SetBinding(Label.TextProperty, ".");
+					label.SetBinding(Label.AutomationIdProperty, ".");
+
 					cell.View =
 					new StackLayout()
 					{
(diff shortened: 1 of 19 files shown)


Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
```diff
@@ -328,41 +348,41 @@ void IAccessibilityElementsController.ResetAccessibilityElements()
...
-		void UpdateShellInsetPadding()
-		{
-			if (!(Element?.Parent is ShellContent))
-				return;
+			bool usingSafeArea = Page.On<PlatformConfiguration.iOS>().UsingSafeArea();
+			bool isSafeAreaSet = Element.IsSet(PageSpecific.UseSafeAreaProperty);
 
-			nfloat topPadding = 0;
-			nfloat bottomPadding = 0;
+			if (IsPartOfShell && !isSafeAreaSet)
```
seansparkman: So the default behavior for `Page` is to have `UseSafeArea` default to `false`?  Why would this behave differently?  Is this going to effect more than just the root page of `ShellSection`?

PureWeen: Looking at your initial issue. The problem is that shell just flat out ignored the SafeArea setting. With this PR your code here https://github.com/seansparkman/ShellNavigationExample/blob/use-safe-area-ignored/ShellNavigationRegistration/ShellNavigationRegistration/Views/OnBoarding.xaml#L13 will cause usingSafeArea to be false

https://github.com/xamarin/Xamarin.Forms/pull/6457/files#diff-d95827e5633c20beb139e10ccd1a40d9R415 so whatever you set the padding to is going to be the padding used for the page

seansparkman: Right, but does `ContentPage` that isn't a part of `Shell` normally use the SafeAreaInsets?  Why would it behave differently for `Shell` and does this change effect all pages or just the first one?

PureWeen: Good point!! I added an additional check so if there's nothing up top then it should just take the pages full height

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/ShellInsets.cs
```diff
@@ -0,0 +1,367 @@
...
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.None, 0, "Shell Inset Test",
+		PlatformAffected.All)]
+#if UITEST
+	[NUnit.Framework.Category(UITestCategories.Shell)]
+#endif
+	public class ShellInsets : TestShell
+	{
+		const string EntryTest = "EntryTest";
```
paymicro:
```suggestion
		const string EntryTest = nameof(EntryTest);
```
How about this?

Xamarin.Forms.Platform.iOS/Renderers/KeyboardInsetTracker.cs
```diff
@@ -10,33 +10,47 @@ internal class KeyboardInsetTracker : IDisposable
...
 		{
 			if (_disposed)
 				return;
+
 			_disposed = true;
 
 			KeyboardObserver.KeyboardWillShow -= OnKeyboardShown;
 			KeyboardObserver.KeyboardWillHide -= OnKeyboardHidden;
+
+			_shellScrollViewTracker.Dispose();
```
paymicro:
```suggestion
			_shellScrollViewTracker?.Dispose();
```

Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
```diff
@@ -114,6 +118,25 @@ public void SetElement(VisualElement element)
 			}
 
 			EffectUtilities.RegisterEffectControlProvider(this, oldElement, element);
+
+			var parent = Element.Parent;
```
paymicro: I think here need to check for null of Element

Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
```diff
@@ -326,45 +356,98 @@ void IAccessibilityElementsController.ResetAccessibilityElements()
...
 		void UpdateUseSafeArea()
 		{
-			if (!Forms.IsiOS11OrNewer)
+			if (Element == null)
+				return;
+
+			if (_userOverriddenSafeArea)
+				return;
+
+			if (!IsPartOfShell && !Forms.IsiOS11OrNewer)
```
paymicro: I think there is a mistake. Because the method goes deeper only if `IsPartOfShell` and `Forms.IsiOS11OrNewer` must be `false`, but logic below expects one of them to be `true`.

Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
```diff
@@ -445,6 +528,19 @@ void UpdateHomeIndicatorAutoHidden()
 			SetNeedsUpdateOfHomeIndicatorAutoHidden();
 		}
 
+		double _tabThickness;
+		private bool _isInItems;
```
paymicro:
```suggestion
		bool _isInItems;
```

Xamarin.Forms.Platform.iOS/Renderers/ShellSectionRootRenderer.cs
```diff
@@ -139,6 +139,19 @@ protected virtual void LoadRenderers()
 				var renderer = Platform.CreateRenderer(page);
 				Platform.SetRenderer(page, renderer);
 
+				var childView = renderer.ViewController.View;
```
paymicro: The declaration of this variable is best moved inside the `if` block below.

Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
PureWeen: I don't follow. The if statement here is meant to maintain backwards compatibility because currently if you're not part of shell and you're running iOS10 then none of the safearea code will run so this is just a way to exit. 

This probably means I could remove this if statement though
https://github.com/xamarin/Xamarin.Forms/pull/6457/files#diff-d95827e5633c20beb139e10ccd1a40d9R401

but I don't think it's totally necessary to

Xamarin.Forms.Platform.iOS/Renderers/ShellSectionRootRenderer.cs
```diff
@@ -137,8 +137,7 @@ protected virtual void LoadRenderers()
 				ShellContent item = ShellSection.Items[i];
 				var page = ((IShellContentController)item).GetOrCreateContent();
 				var renderer = Platform.CreateRenderer(page);
-				Platform.SetRenderer(page, renderer);
-
+				Platform.SetRenderer(page, renderer);
```
paymicro:
```suggestion
				Platform.SetRenderer(page, renderer);
```

Xamarin.Forms.Platform.iOS/Renderers/ShellSectionRootRenderer.cs
```diff
@@ -254,16 +253,18 @@ void OnShellSectionItemsChanged(object sender, NotifyCollectionChangedEventArgs
...
-			_header.View.Frame = frame;
+			int tabThickness = 0;
+			if (_header != null)
+			{
+				tabThickness = HeaderHeight;
+				CGRect frame;
+				if (Forms.IsiOS11OrNewer)
+					frame = new CGRect(View.Bounds.X, View.SafeAreaInsets.Top, View.Bounds.Width, HeaderHeight);
+				else
+					frame = new CGRect(View.Bounds.X, TopLayoutGuide.Length, View.Bounds.Width, HeaderHeight);
```
paymicro: here you can simplify Line 260-266
```C#
var headerTop = Forms.IsiOS11OrNewer ? View.SafeAreaInsets.Top : TopLayoutGuide.Length;
_blurView.Frame = _header.View.Frame = new CGRect(View.Bounds.X, headerTop, View.Bounds.Width, HeaderHeight);
```

Xamarin.Forms.Platform.iOS/Renderers/ShellSectionRootRenderer.cs
```diff
@@ -254,16 +253,18 @@ void OnShellSectionItemsChanged(object sender, NotifyCollectionChangedEventArgs
...
-			if (Forms.IsiOS11OrNewer)
-				frame = new CGRect(View.Bounds.X, View.SafeAreaInsets.Top, View.Bounds.Width, HeaderHeight);
-			else
-				frame = new CGRect(View.Bounds.X, TopLayoutGuide.Length, View.Bounds.Width, HeaderHeight);
-			_blurView.Frame = frame;
-			_header.View.Frame = frame;
+			int tabThickness = 0;
+			if (_header != null)
+			{
+				tabThickness = HeaderHeight;
```
paymicro: `HeaderHeight` is constant. Why put its value in a local variable?

PureWeen: because it's only changing tabThickness if the _header is not null otherwise the tabThickness just stays zero

PureWeen:
One failed test on API 28 unrelated and should be fixed by
https://github.com/xamarin/Xamarin.Forms/pull/6759

Issue #5451, opened 2019-03-05
Android/UWP ListView Memory Leak


alexanderzuban:
### Description

When updating ItemsSource for ListView cell views remains in memory both on Android(not so fast) and UWP(rapidly)

### Steps to Reproduce

1.  Run sample app in Android/UWP
2.  Click a few times on any item 
3.  Check debug output or diagnostic on UWM 

### Expected Behavior
Cell Views GC

### Actual Behavior
 Cells grow in memory

Sample Project:
[ListViewMemory.zip](https://github.com/xamarin/Xamarin.Forms/files/2929297/ListViewMemory.zip)

kingces95:
Allocation here. The debugger knows the source file and line for the frame's before the ctor but for some reason consider the symbols unloaded... So could also be a tooling issue. I'm in VS 2019.

![image](https://user-images.githubusercontent.com/4120386/53996608-2e490c00-40dd-11e9-9bd6-c092a19aaac4.png)

PureWeen:
I think UWP memory leaks with the Cells is related to this
https://github.com/xamarin/Xamarin.Forms/issues/5465

alexanderzuban:
From what I can say some Cells never get Parent reset(after profiling my real app) and because they have a listener to Parent PropertyChanged (height) they keep everything(on Android) on UWP things even worse.  Maybe use weak property listener in Cell to listen for Parent.PropertyChange will help (but this maybe workaround for a situation when Parent is not reset in the first place).
`protected override void OnParentSet()
		{
			if (RealParent != null)
			{
				RealParent.PropertyChanged += OnParentPropertyChanged;
				RealParent.PropertyChanging += OnParentPropertyChanging;
			}

			base.OnParentSet();

			PropertyPropagationController.PropagatePropertyChanged(null);
		}`

verybadcat:
This issue is a problem for me as well. I would be delighted if it is fixed.

EmilAlipiev:
Is that still the case for collectionview? I use sflistview and it doesn't have this problem.

PR #6459, opened 2019-06-07
[UWP] Fix list views leaking cells due to missing event unsubscription

### LINKED ISSUE #3065: [UWP] Listview memory leak
### Description
using a large data set as ItmesSource of xamarin forms listview produce a memory leak in UWP and when I tried the same behavior in native uwp I didn't able to produce the issue
### Steps to Reproduce

1. add a listview with a complex view cell (nested grids, images, frame and labels)
2. bind an observableCollection of 3000 item to it 
3. hold the scroll by the mouse and move the mouse up and down quickly 
4. check the memory allocation in windows task manager during list view scrolling or watch the visual studio diagnostic tool memory allocation 

### Expected Behavior
ListView should work properly in uwp
### Actual Behavior
ListView produce a memory leak and performance issues in uwp
### Basic Information
- Platform Target Frameworks:
  - UWP:  16299 
- Nuget Packages:
I can produce this with forms 2.5 , 3 , 3 pre4 and the last nightly build 3.1.0.506097

### Reproduction Link
xamarin forms 
https://1drv.ms/u/s!AoIY8oQvWbhGgTFicmMFLddURVgJ
uwp without xamarin
https://1drv.ms/u/s!AoIY8oQvWbhGgTDaokspB4l3RVk1

also we appreciate if you could release a service release with all the listView issues that fixed in 3 and 3.1 for 2.5 as we have apps target 2.5 and we can't update to 3 because we want to support the earlier version of windows

### LINKED ISSUE #3339: ListView with ViewCells Still Leaks Memory on UWP
### Description

I am using a `ListView` to display a list of a fixed maximum size, where I only swap out elements (in the list). (Removing the oldest / highest index, to insert a new element at index 0.) I am using `ViewCell` via `DataTemplate`.
Simple observation: The longer it runs, the more memory the process consumes. Growth is steeper the more complex the view is.
Workaround: If I add a handler to the `cell.Disappearing` event which nulls the view (`cell.View = null`), memory doesn’t really seem to leak anymore. So it would seem that mainly my view is being leaked, and via the `ViewCell`.

### Steps to Reproduce

1. Build `ViewCell` of relatively complex layout, so instances use a noticeable chunk of memory
2. Keep adding fresh instances of this `ViewCell` to a `ListView`, and remove them again
3. Watch memory consumption grow indefinitely

### Expected Behavior

If size of elements in the `ListView` stays constant, consumed memory should also stay roughly constant.

### Actual Behavior

Memory consumption keeps growing.

### Basic Information

- Version with issue: Latest (3.1.0 Service Release 1)
- Last known good version: Not sure, UWP ListView memleaks have been around for a while
- IDE:
- Platform Target Frameworks: 
  - UWP:  16299
- Affected Devices: Windows 10 PC

dasMulli:
### Description of Change ###
This change fixes a memory leak where view cells and the bound models are leaked. The list view creates new cells but keeps a reference to cells that are no longer needed, thus creating a memory leak.
In our production case, this causes an app to crash with OOM after 3 days of intensive use.

The code that ultimately needs to run to remove the owning reference from the list to the cell is https://github.com/xamarin/Xamarin.Forms/blob/471e9f81250e777fe194a764270bfda70a1e4dfe/Xamarin.Forms.Core/Cells/Cell.cs#L183-L187 - event handlers on `RealParent` must be unsubscribed at some point.

The creation stack trace dotMemory shows for the retaining event handlers is `SetCell()` at (https://github.com/xamarin/Xamarin.Forms/blob/471e9f81250e777fe194a764270bfda70a1e4dfe/Xamarin.Forms.Platform.UAP/CellControl.cs#L334-L336) being called via `MeasureOverride()`.

The issue can be reproduced by creating a ListView using ViewCell in data templates. See linked issues for repro projects.

Currently, a workaround is possible by subclassing the cell and nulling the `.Parent` property on disappearing: https://gist.github.com/dasMulli/64d0e3627ea2e982f56b5ca8209a681f

### Issues Resolved ### 

- fixes #3065
- fixes #3339

May also help #5451 though part of this may be a duplicate of the above.

### API Changes ###
 
 None

### Platforms Affected ### 

- UWP

When a view is disappearing (after OnDisappearing event), it's Parent property will no longer reference the parent list and the cell may be garbage collected.

Not applicable

### Testing Procedure ###

Tested the change using our application and the included gallery application. Please let me know other needed tests (!).

The testing for the memory issue was performed using dotMemory by creating a ListView binding to an observable collection of pseudo elements, using a ViewCell with a Label as template. After a bit of rapid scrolling around, the number of ViewCells after a forced GC was in excess (>480) of the number of collection items (70). After the change, only around 30 view cells stayed in memory after GC (depending on the window size).

### PR Checklist ###

- Has automated tests : ⚠️ no. I do not know how to properly test this change
- [x] Changes adhere to coding standard: ⚠️ please correct me if this change does not adhere to the coding standard

⚠️ First time contributor to Xamarin.Forms, please review with care. Please advise if there is a more desirable way of fixing the underlying issue instead of the way proposed in this PR.

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla34561.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla34561.cs
@@ -16,6 +16,12 @@ namespace Xamarin.Forms.Controls.Issues
 {
 	[Preserve (AllMembers = true)]
 	[Issue (IssueTracker.Bugzilla, 34561, "[A] Navigation.PushAsync crashes when used in Context Actions (legacy)", PlatformAffected.Android)]
+
+#if UITEST
+	[NUnit.Framework.Category(UITestCategories.ListView)]
+	[NUnit.Framework.Category(UITestCategories.ContextActions)]
+	[NUnit.Framework.Category(UITestCategories.Navigation)]
+#endif
 	public class Bugzilla34561 : TestContentPage
 	{
 		protected override void Init ()
--- a/Xamarin.Forms.Platform.UAP/CellControl.cs
+++ b/Xamarin.Forms.Platform.UAP/CellControl.cs
@@ -352,6 +352,7 @@ namespace Xamarin.Forms.Platform.UWP
 			{
 				oldCell.PropertyChanged -= _propertyChangedHandler;
 				oldCell.SendDisappearing();
+				oldCell.Parent = null;
 			}
 
 			if (newCell != null)


dasMulli:
I'm open to adding an extra test, i think i can do something similar to the existing tests.
But i'd first like to get a review if the fix is the right approach to the problem.

dasMulli:
Thanks, I'll look into it next weekend. Truth be told i had some trouble getting tests to run on my machine. Most of them failed because my keyboard layout wasn't qwerty and some still failed (without my changes).

samhouts:
Unfortunately, this does not resolve the leak, and it has consequences. Our recommendation at this time is to use the CollectionView control instead of the ListView. This new control does not have the same performance issues that ListView does. Thanks for this PR! I will close it now.

dasMulli:
I'm sure there is at least one leak that this would have resolved, but there may be others. We should be fine using the workaround mentioned above for the case that turned out to be problematic for us (our UI endurance tests crash the app with OOM after a few hours when not using the workaround).

So guiding the users to a new UI component, not breaking existing users du to the behaviour implications of at least this partial fix and offering a workaround to some users seems like a reasonable way to address this.

PR #6460, opened 2019-06-07
[Android] Context actions menu is now updated when changing cells.

### LINKED ISSUE #1455: Context action are not changed when selected item changed on Android
### Description

The context actions on Android are not the action of the last selected cell in "long pressed" mode, they are the ones of the originally long pressed cell.

### Steps to Reproduce

1. Have a ListView containing at cells with different MenuItem
2. Long pressed on a cell
3. Select another cell while in "long pressed" mode

### Expected Behavior

The context actions in the ContextActionsBar are replaced by the ones of the newly selected cell.
Or it is not allowed to select another cell.

### Actual Behavior

The context actions in the ContextActionsBar are not replaced. Also, given less context actions in the last selected cell, an OutOfRangeException will result when selecting the last action(s).

### Basic Information

- Version with issue: 2.5.0.121934
- IDE: Rider 2017.2 on Mac
- Platform Target Frameworks:
  - Android: 7.1.1
- Android Support Library Version: 26.1.01
- Affected Devices: Android Simulator API 25

### LINKED ISSUE #4041: SupportActionMode Menu not updated in Android when using DataTemplateSelector
If a ListView uses a DataTemplateSelector for it's ItemTemplate, with different ContextActions, the SupportActionMode Menu is not updated when the selected item is changed.

Reproduce:
1. Create ListView with DataTemplateSelector and different ContextActions for each template
2. Run app on Android
3. Long press on Item and watch SupportActionMode Menu appear
4. Select another item with a different ItemTemplate (and of course different ContextActions)
5. SupportActionMode Menu buttons aren't updated

Possible solution:
Check if ContextActions collection has changed when ActionModeContext is set. Before setting the backing field to the new value check if collection has changed if so then set field _actionModeNeedsUpdates to true.
https://github.com/xamarin/Xamarin.Forms/blob/f79fc69b9e85ebd946bd7128c414735860bcc9f2/Xamarin.Forms.Platform.Android/CellAdapter.cs#L47

Created a sample project.
[DataTemplateSelector.zip](https://github.com/xamarin/Xamarin.Forms/files/2462031/DataTemplateSelector.zip)

GuidoNeele:
### Description of Change ###
Added _actionModeNeedsUpdates = true; to the CellAdapter.HandleContextMode method to make sure that the ActionSupprtMenu is refreshed when it's active and another item is tapped. Menu is only updated when it's on screen and the newly tapped cell has context actions.

### Issues Resolved ### 
- fixes #1455
- fixes #4041

### API Changes ###
None

### Platforms Affected ### 
- Android

When the ActionSupportMenu is visible on screen and the user taps another cell the ActionSupportMenu will be recreated to make sure the context is correctly applied and the right menu items are shown.

[Video of the problem before the fix](https://youtu.be/EbpAnD4O7nM)

### Testing Procedure ###
Problem only occurs on Android because it's the only platform with the ActionSupportMenu. Go to listview, longpress an item, wait for the context menu to popup, tap another item and now the menu is updated, previously the menu wouldn't change.

[Video of the result after the fix](https://youtu.be/wgHh1OwwT1Q)

### PR Checklist ###
- [x] Has automated tests


GuidoNeele:
The behavior right now, 4.0 and earlier, is not "closing the menu", but showing an incorrect menu, when the context actions have changed. So there are two choices when tapping another list item and context menu is active:
1. Show the correct menu (which this pull request proposes)
2. Close the menu 
I don't think the pull request is a breaking change since neither of these two options is happening at the moment.

GuidoNeele:
If you want I could create a small Xamarin Android sample and see what behavior is used there?

jfversluis:
Hi @GuidoNeele thanks for your PR!

From what I’m gathering here I see this. In the old situation, you open the context menu for cell0 but when you tap another cell (tap or long press? not sure on that. From what I can see long press to open the next context menu) the (same) context menu is still shown.

Your fix will show the right context menu for cell1 that you long press. 

Right?

If this is the case I agree with what is going on here. In the case of showing the context menu and then simply tapping another cell I agree with @samhouts and @rmarinho that the context menu should be closed.

Your last comment mentions something about a sample project? That would be awesome :) 

Also in the meantime a conflict has occurred, if you could fix that and provide us with the info we can hopefully get this done soon, thanks!

GuidoNeele:
> Hi @GuidoNeele thanks for your PR!
> 
> From what I’m gathering here I see this. In the old situation, you open the context menu for cell0 but when you tap another cell (tap or long press? not sure on that. From what I can see long press to open the next context menu) the (same) context menu is still shown.
> 
> Your fix will show the right context menu for cell1 that you long press.
> 
> Right?
> 
> If this is the case I agree with what is going on here. In the case of showing the context menu and then simply tapping another cell I agree with @samhouts and @rmarinho that the context menu should be closed.
> 
> Your last comment mentions something about a sample project? That would be awesome :)
> 
> Also in the meantime a conflict has occurred, if you could fix that and provide us with the info we can hopefully get this done soon, thanks!

Hi @jfversluis,

The steps are as followed:

Before PR (current situation)
1. Long press an item in the list
2. Context menu is shown on top
3. Tap another item in the list
4. Context menu stays open but doesn't represent the context actions of the selected item(DataTemplate).

After PR
1. Long press an item in the list
2. Context menu is shown on top
3. Tap another item in the list
4. Context menu will update if selected item (DataTemplate) has context actions defined.

The default behavior in Xamarin Android is that the context menu is kept open and you can select multiple items in the list (like WhatsApp). But I don't think you can compare native with forms that easily since the menu in native is created for the entire list whereas with the Xamarin Forms the actions are defined on each DataTemplate.

My opinion is that closing the menu is not the right way to go. Since the default action for Xamarin Forms ActionMode is to behave as a single select it should:
a. Update the menu to represent the actions defined on the DataTemplate.
or
b. Prevent changing selection. Like Instagram does in their reactions section.

jfversluis:
I can agree on that. I’m more used to the iOS ecosystem, do you maybe have any examples of (preferably stock) Android apps that show this behavior?

Between the options you are giving I would prefer the menu to update if I was a user, so then this seems like a legit solution.

Could you maybe fix the conflict so I can run the builds and tests again? Meanwhile I can see if I can get some agreement on this internally.

GuidoNeele:
@jfversluis 
Created a few sample projects. They can be found at https://github.com/GuidoNeele/XamFormsSamplePR6460. It's got a Xamarin Forms sample that shows the current behavior when using ListView with a template selector. The second sample is a Xamarin Android project which shows the action mode in stock Android.

I also fixed the conflict so the XF tests should work again.

jfversluis:
Thanks for the samples!

Although the stock Android solution doesn’t seem to do quite the same, I can see what this should do.

We have had some internal discussion about this (again) and there are some different angles here. Personally, I agree with you. In fact, I think most of us agree with you and these changes. The “problem” is that this has worked like this since forever and there might be people that have started to depend on this behavior. Wrongly, but they could have. My gut feeling says it won’t be that many, but unfortunately that is not really something we can act on.

Is there some way to achieve some kind of backwards compatibility? Introduce a (non-discoverable) property that lets us reinstate the previous (now current) implementation? I would, again personally, be all for being the right implementation (this PR) to be the default, but then we at least have to ability for people to revert back to how it was. If no one ever complains we can remove it in a future version. How does that sound @GuidoNeele?

cc: @hartez @rmarinho and @samhouts to make a final decision on this maybe?

GuidoNeele:
@jfversluis 

I've added an Android platform specific BindableProperty called "IsContextActionsLegacyModeEnabled" (let's discuss the name :-)) so it's possible to revert to the old behavior. Also updated the UI test to also test for the old behavior.

Using the old behavior will look like this:

```
<DataTemplate x:Key="Temp1">
    <ViewCell android:ViewCell.IsContextActionsLegacyModeEnabled="true">
        <ViewCell.ContextActions>
            <MenuItem Text="{Binding Item1Text}" />
            <MenuItem Text="{Binding Item2Text}" />
            <MenuItem Text="{Binding Item3Text}" />
            <MenuItem Text="{Binding Item4Text}" />
        </ViewCell.ContextActions>
        <Grid>
            <Label Text="{Binding Text}" />
        </Grid>
    </ViewCell>
</DataTemplate>
```

jsuarezruiz:
@GuidoNeele Can you review the conflict?. Thanks!.
@hartez @rmarinho and @samhouts can you take a look to the latest @jfversluis comment?.

Xamarin.Forms.Core/PlatformConfiguration/AndroidSpecific/ViewCell.cs
```diff
@@ -0,0 +1,37 @@
+﻿namespace Xamarin.Forms.PlatformConfiguration.AndroidSpecific
+{
+	using System;
+	using FormsCell = Forms.Cell;
+
+	public static class ViewCell
+		{
```
jfversluis:
```suggestion
	{
```

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue1455.xaml.cs
```diff
@@ -0,0 +1,158 @@
...
+		protected override void Init()
+		{
+			BindingContext = new MyViewModel();
+		}
+
+
+		[Preserve(AllMembers = true)]
+		public class MyViewModel : PropertyChangedBase
+		{
+			private bool _isContextActionsLegacyModeEnabled;
```
jfversluis:
```suggestion
			bool _isContextActionsLegacyModeEnabled;
```

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Xamarin.Forms.Controls.Issues.Shared.projitems
```diff
@@ -9,13 +9,17 @@
...
+    </Compile>
     <Compile Include="$(MSBuildThisFileDirectory)CollectionViewHeaderFooterString.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)CollectionViewHeaderFooterTemplate.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)CollectionViewHeaderFooterView.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)CollectionViewItemsUpdatingScrollMode.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Issue5046.xaml.cs">
-      <DependentUpon>Issue5046.xaml</DependentUpon>
-      <SubType>Code</SubType>
+      <DependentUpon>Issue5046.xaml</DependentUpon>	
+      <SubType>Code</SubType>
```
jfversluis:
```suggestion
      <SubType>Code</SubType>
```

jfversluis:
I'm fine with the name as it will be something that people shouldn't be using anyway 😬 

Just a couple of minor things to attend to, otherwise looks good to me.

Issue #4653, opened 2018-12-06
IsTabStop property not working when Grid contains ScrollView


RawoofSharief:
### Description
We are having Grid with ScrollView architecture.

**Grid->ScrollView->Label**

**IsTabStop** property set as **False** for the Grid but when I click tab button focus moved to it.

### Steps to Reproduce

1. Run the attached sample
2. Click tab button which focus on the grid.

### Expected Behavior
Focus not moved to grid
### Actual Behavior
Focus still moved to it
### Basic Information

- Version with issue: Xamarin.forms 3.4.0.1008975
- Platform Target Frameworks: 
  - UWP: 16229

### Reproduction Link

[Sample.zip](https://github.com/xamarin/Xamarin.Forms/files/2652294/Sample.zip)

PieEatingNinjas:
I don't think this is a bug, or I'm misunderstanding the issue.
When I look at the sample that you provide, I see that you're setting `IsTabStop="false"` on the Grid, not on the ScrollView.
```
<Grid Grid.Column="2" IsTabStop="False">
    <ScrollView>
        <Label Text="Label" Grid.Row="0"/>
    </ScrollView>
</Grid>
```
When you press tab, the focus is moved to the Scrollview, which is expected as you didn't define `IsTabStop="false"` on the Scrollview itself. Setting IsTabStop doesn't apply to child elements, only on the element itself. If you want to avoid focus on the Scrollview, you should set it explicitly on that element.

PR #4703, opened 2018-12-11
[UWP] Fixes IsTabStop property in Layouts

paymicro:
### Description of Change ###

Fixes IsTabStop property in Layouts

### Issues Resolved ### 

- fixes #4653

### API Changes ###

 
 None

### Platforms Affected ### 

- UWP

### Behavioral/Visual Changes ###

None

### Before/After Screenshots ### 

Not applicable

### Testing Procedure ###

- run UItest 4653
- check that the keyboard can not select any control inside grid.
- click on `Toggle Grid IsTabStop`
- tab should work on some buttons, entry
- the ScrollView does not gain focus anyway

### PR Checklist ###

- [ ] Has automated tests 
- [x] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard

PR #6468, opened 2019-06-09
[UWP] Fixes jumps at scroll view

### LINKED ISSUE #5652: Focus Inside ScrollView Changes when Non-Interactable Element is Clicked
### Description
When a scrollviewer wraps some item that's scrolled out of view, and the user taps on a portion of the contained area that isn't interactable (such as a label), the scrollviewer brings the first focusable element into view. 
This causes an unexpected jump, and makes scrollviewer unreliable.

### Steps to Reproduce

1. Create a ScrollViewer
2. At the top of the scrollviewer, place an Entry or similar control.
3. Place some other control further down. I used button with a large button.
4. In the app, click near, but not on the button, in the empty space on the page.

### Expected Behavior
The ScrollView does not move.

### Actual Behavior
The ScrollView brings the first focusable element into view, causing the page to jump unexpectedly.

### Basic Information

- Version with issue: 3.6.0.264807
- Last known good version: 3.4.0.1029999
- IDE: VS 15.7.9
- Platform Target Frameworks: 
  - UWP:  17134
- Affected Devices: Windows Desktop

### Reproduction Link
https://github.com/jkastner/XamFormsScrollviewBug

### Exact Cause
My colleague went into the Xamarin.Forms codebase, and appears to have narrowed the actual cause:

Explanation:
On UWP on a page with a scroll viewer, clicking on any non-interactive portion of the page causes the scroll viewer to "jump" to a different position

Cause
The IDontGetFocus interface on the ScrollViewRenderer triggers a call to OnGotFocus() any time the scroll viewer gets focus, which in turn moves the focus to a different element which causes the scroll viewer to jump to the position of that element.

[UWP] Fixes IsTabStop property in Layouts (#4703)
https://github.com/xamarin/Xamarin.Forms/commit/d7bfb0cb7bdbea47a09c228550e64541184fe9b9#diff-b1a5e5a01f67d1a9df69aa44c77e786c

IsTabStop property not working when Grid contains ScrollView #4653
https://github.com/xamarin/Xamarin.Forms/issues/4653

Work-around

Implement a version of ScrollViewRenderer that doesn't implement the IDontGetFocus interface

AxelUser:
### Description of Change ###

Because changes from #4703 leads to issue #5652, I've removed interface `IDontGetFocus`. However I think that problem is deeper - method [`OnGotFocus`](https://github.com/xamarin/Xamarin.Forms/blob/24bf23b2d1060be787d8a9ed4c1626f7a2165ebe/Xamarin.Forms.Platform.UAP/VisualElementRenderer.cs#L177) should respect area of `ScrollView` and shouldn't allow inner elements being focused by `FocusManager.TryMoveFocus`.

### Issues Resolved ### 

- fixes #5652

### API Changes ###

 
 None

### Platforms Affected ### 

- UWP

None

Not applicable

### Testing Procedure ###

- Open page Issue5652
- Scroll down to the button with caption "Click around"
- Click somewhere near button
- Label with placeholder "Should not jump here" will not gain focus, thus no scrollview jumping occurred.

### PR Checklist ###

- [ ] Has automated tests 


kgrosche:
Does this mean that a new version of Xamarin.forms will be available soon, in which bug 5652 is fixed? In which version will this be the case? Or do I have to do something else to avoid this bug?

regards
Klaus

AxelUser:
@kgrosche it depends whether code will be accepted or not. I'll mention you if there where some updates.

Btw, @paymicro what do you think?

charlesroddie:
Thanks @AxelUser !

@samhouts can we prioritize this and bring it forward to vCurrent or at worst vNext? This fixes a serious regression, reverting just the part of the relevant PR that caused the problem.

samhouts:
@AxelUser Can you please rebase to 3.6.0? Thanks!

AxelUser:
@samhouts sorry, I've messed up branch and PR was automatically marked as Closes.

kgrosche:
Hello
I'm in big trouble because of this issue! I need to go to version Xamarin.forms 4.x because in version 3.4 there is a bug related to TemplateBinding (UWP). Is there anyone in the world who can tell me when and in which version this bug (6468) is fixed? 
Please please, I really need help!

regards
Klaus

AxelUser:
@samhouts is there anything related with this issue I can help with?

samhouts:
@AxelUser Thank you for this contribution! We understand that this is fixing a regression. Unfortunately, entirely reverting the fix for #4703 would now constitute introducing a new regression. We need a happy medium where both functionality works as expected. Any ideas?

charlesroddie:
OK I was a little hasty in organizing this, because this change reverts what looked from the code changes to be a small part of the PR, but according to @PureWeen this breaks the motivating issue that led to the PR.

The original issue ["IsTabStop property not working when Grid contains ScrollView"](https://github.com/xamarin/Xamarin.Forms/issues/4653) is very niche, requiring both scrollviews inside grids and use of tabs.

The subsequent issue ["ScrollView Jumps when Non-Interactable Element is Clicked"](https://github.com/xamarin/Xamarin.Forms/issues/5652) is major, an issue with every scrollview.

Since https://github.com/xamarin/Xamarin.Forms/pull/4703 causes a much larger problem than it fixes, and there isn't an immediate resolution, I suggest:
- Revert https://github.com/xamarin/Xamarin.Forms/pull/4703 and re-add it it as a WIP PR, with the test from this PR attached.
- Merge it when the original issue can be resolved without causing the subsequent one.

charlesroddie:
@samhouts

samhouts:
@paymicro Would you mind looking into this to see if we can resolve both issues without having to revert? Thanks!

paymicro:
@samhouts sure. I added a commit that should fix #4653 in another way. 
@PureWeen please re-review. 
Thanks

samhouts:
@paymicro I confirmed that this re-fixes #4653! Thanks!!!!

@AxelUser If I scroll to the bottom and click anything but the Button, it doesn't jump. But If I click the Button and then click anything but the Button, it jumps back to the Entry. Is that what you expected?

charlesroddie:
@paymicro Are you able to fix the remaining jumping issue that @samhouts identified? Thanks!

charlesroddie:
@samhouts can we set a deadline for reverting https://github.com/xamarin/Xamarin.Forms/pull/4703 if the combination of issues can't be fixed? E.g. if we can can't fix and merge this PR in the coming week then revert 4703 and redo as a WIP PR.

AxelUser:
Hm, seems like after fixing regression new fixes doesn't working. I'll try to find new way

Xamarin.Forms.Platform.UAP/VisualElementRenderer.cs
```diff
@@ -155,7 +155,7 @@ public void SetElement(VisualElement element)
 
 			OnElementChanged(new ElementChangedEventArgs<TElement>(oldElement, Element));
 
-			if (_control != null && this is IDontGetFocus)
+			if (_control != null && this is ITabStopOnDescendants)
 			{
 				_control.GotFocus += OnGotFocus;
```
kvpt: I think this part need to be protected from multiple subscription.

Like this :
```cs
_control.GotFocus -= OnGotFocus;
_control.GettingFocus -= OnGettingFocus;
_control.GotFocus += OnGotFocus;
_control.GettingFocus += OnGettingFocus;
```

charlesroddie:
@samhouts can we revert 4703 **NOW** and redo as a WIP. Otherwise how can this be escalated?

Xamarin.Forms.Platform.UAP/VisualElementRenderer.cs
charlesroddie: Can you or @AxelUser check whether this fixes both issues today?

charlesroddie:
Any feedback on whether we can get a quick revert of #4703 @samhouts @davidortinau @paymicro ?

Xamarin.Forms.Platform.UAP/VisualElementRenderer.cs
AxelUser: @charlesroddie @kvpt This fixes are redundant, cause control unsubsribed from events at [Dispose](https://github.com/xamarin/Xamarin.Forms/blob/2910754f338c2620d794f2eeee340fa7afb3388d/Xamarin.Forms.Platform.UAP/VisualElementRenderer.cs#L279)

AxelUser: Besides it doesn't fix back the issue

kvpt: @AxelUser I'm not sure of that, from others renderers in the framework the SetElement method always follow this pattern :
```cs
if (oldElement != null)
{
     oldElement -= XXX;
}

if (newElement != null)
{
     newElement  += XXX;
}
```
And after that dispose is implement like this :
```cs
Element = null;
```

I think it's mean that the SetElement method can be called sometime several times.

AxelUser: Hm, that makes sense, I've added cleanup at 034d0d6. Thank you for good point!

charlesroddie: Dooes this change make any difference to the issue @samhouts [identified](https://github.com/xamarin/Xamarin.Forms/pull/6468#issuecomment-508535977) ?

kvpt:
@charlesroddie I got some time to test that issue.

The problem described by @samhouts is present.

After some tests I think I understand the issue.
The jump occur because the method ScrollViewRenderer.OnViewChanged get called with 0 as vertical offset.
And this is caused by the focus of Entry, which can be checked by setting a breaking point in VisualElementRender.OnControlGotFocus.

It's here that it's become interesting, a ScrollViewer get and lost the focus at the same time, two ScrollViewer ?
So I checked the VisualElementTree and it indeed contains two ScrollViewer.
It's because Entry element is a FormsTextBox control and you can find that the style of the control contains also a ScrollViewer.
https://github.com/xamarin/Xamarin.Forms/blob/4fd648471915f4b37f9ad102c11bc527b4d71060/Xamarin.Forms.Platform.UAP/FormsTextBoxStyle.xaml#L214

Removing the ScrollViewer from the FormsTextBoxStyle fix the issue.
So this bug is in the FormsTextBox class, it can be fixed either by removing the ScrollViewer in the style or find a way to prevent this ScrollViewer to steal the focus/scroll event.
I'm for the removal because it's never a good idea to nesting ScrollViewer, it cause issue like this and also kill performance.

charlesroddie:
@kvpt
> this is caused by the focus of Entry
> it can be fixed either by removing the ScrollViewer in the style or find a way to prevent this ScrollViewer to steal the focus/scroll event.

The focusing of Entry is the bug. Clicking on an unrelated part of a scrollview should not cause some Entry to get focus.

The presence of a StackLayout inside an Entry is presumably intentional and I would imagine that when you remove it Entry's will work differently and less well when filled with text that overfills the Entry. So I don't think the fix is to remove StackLayout.

@samhouts @davidortinau @paymicro No-one knows how to fix this at the moment. There is no evidence that a fix can be found in the next day or even in the next few days. We need to revert this pull request ASAP to get UWP working.

kvpt:
@charlesroddie I'm not mention a StackLayout, I mention a ScrollViewer, a native UWP content, see the [source file](https://github.com/xamarin/Xamarin.Forms/blob/4fd648471915f4b37f9ad102c11bc527b4d71060/Xamarin.Forms.Platform.UAP/FormsTextBoxStyle.xaml#L214).
The other platform don't have a ScrollViewer inside the Entry, it's only in UWP.

charlesroddie:
> @kvpt I'm not mention a StackLayout

Sorry for a few seconds there was a typo in my comment.

> The other platform don't have a ScrollViewer inside the Entry

In that case it could be a fix. Actually the UWP TextBox should handle any scrolling needed so ScrollViewer does seem strange.

kvpt:
Seem to be a native UWP known issue : https://github.com/microsoft/microsoft-ui-xaml/issues/597

charlesroddie:
Are you saying you know in what way that issue relates to this one?

kvpt:
The native UWP app have the same issue as Xamarin.Forms UWP app, so the bug is in the UWP side and not in the Xamarin.Forms side.
But it's not mean that this bug can't be fixed on Xamarin.Forms side.
Xamarin.Forms suscribe to the native event so we have to find when to ignore the focus and when not. I don't think that we can do that with interface, because the problem is not at compilation time but at runtime when the page which contains the entry (and it's scrollviewer on uwp) also contains another scrollviewer. I think this case can be detected and ignored.

In a perfect world, the bug will be fixed in the UWP ScrollViewer and it's done.
But in this case Xamarin.Forms will have to add custom code to prevent this bug.
We can't modify the scrollviewer focus bug, so we have to deal with it and ignore the event on some case.

I will make some tests with the FormsTextBox to see what can be done.

kvpt:
After some tests, it doesn't seem possible to identify when the focus event is triggered by the ScrollViewer when it lost focus. 
The FormsTextBox GotFocus event arg parameter **OriginalSource** is always **FormsTextBox**.
The only workaround which work is what is described in the UWP [issue](https://github.com/microsoft/microsoft-ui-xaml/issues/597).
Setting the property IsTabStop fo the ScrollViewer to True.
In this case the TextBox doesn't get the focus, but in return the ScrollView get the focus when using tab which is also problematic.

So if IsTabStop = false on the ScrollViewer we have jump if it contains TextBox,
and if IsTabStop = true we have no jump but ScrollViewer can take focus if we use tab.

I think that fixing the two issues at once is for now impossible due to UWP [issue](https://github.com/microsoft/microsoft-ui-xaml/issues/597).
For me the best solution is to make the end user to set the value by it self.

It can be easily done with an effect. Like this.

The platform Effect :
```cs
using System;
using System.Diagnostics;
using Windows.UI.Xaml.Controls;
using Xamarin.Forms;
using Xamarin.Forms.ControlGallery.WindowsUniversal;
using Xamarin.Forms.Platform.UWP;

[assembly: ExportEffect(typeof(TabStopEffect), nameof(TabStopEffect))]
namespace Xamarin.Forms.ControlGallery.WindowsUniversal
{
	public class TabStopEffect : PlatformEffect
	{
		protected override void OnAttached()
		{
			try
			{
				if (Control is ScrollViewer scrollViewer)
				{
					scrollViewer.IsTabStop = true;
				}
			}
			catch (Exception ex)
			{
				Debug.WriteLine("Cannot set property on attached control. Error: ", ex.Message);
			}
		}

		protected override void OnDetached()
		{
		}
	}
}
```

The routing Effect :
```cs
public class TabStopEffect : RoutingEffect
{
	public TabStopEffect() : base($"{Effects.ResolutionGroupName}.{nameof(TabStopEffect)}")
	{
	}
}
````

Attaching the effect :
````cs
_scrollView = new ScrollView
{
	Content = longStackLayout,
        Effects = { new TabStopEffect() }
};
````

The exemple code is based on the Issue5652 but you got the idea.

**Comment by PureWeen** [comment]
@AxelUser can you rebase and let me know your thoughts on my comment?

https://github.com/xamarin/Xamarin.Forms/pull/6468#pullrequestreview-321322173

**Comment by AxelUser** [comment]
Hi, @PureWeen , thank you for notifying! Currently I'm on a long vocation and can't test or rebase it. I'll do it when I reach computer.

**Comment by MhAllan** [comment]
Just before you try to merge again, please consider this

1- It has nothing to do with IsTabStop 
2- It has nothing to do with entries

If you create ScrollView and tap anywhere in the space between controls it will jump to top, if it has Frame and you tap that frame it will jump to top 

```csharp
<ScrollView>
	<StackLayout>
		<Button HeightRequest="100" />
		<Button HeightRequest="100" />
		<Button HeightRequest="100" />
		<Button HeightRequest="100" />
		<Button HeightRequest="100" />
		<Button HeightRequest="100" />
		<Button HeightRequest="100" />
		<Button HeightRequest="100" />
		<Button HeightRequest="100" />
		<Frame HeightRequest="100"
			 BackgroundColor="Accent" />
	</StackLayout>
</ScrollView>
```

samhouts:
I'm going to close this one for now until we're ready to return to it. Thanks, @AxelUser !

PR #6473, opened 2019-06-10
[iOS] Fix initial color of RefreshControl in ListViews

### LINKED ISSUE #5728: ListView RefreshControlColor initial
### Description
The initial color of ListView RefreshControlColor on loading first time is not the set color.

### Steps to Reproduce
i set the global color for this ActivityIndicators in App.xaml:

```xml
<Style TargetType="ListView">
    <Setter Property="RefreshControlColor" Value="{StaticResource NavigationPrimary}" />
</Style>

<Style TargetType="ActivityIndicator">
    <Setter Property="Color" Value="{StaticResource NavigationPrimary}" />
</Style>
```

### Expected Behavior
The color should be shown on first loading.

### Actual Behavior
A default color is visible at first loading.

### Basic Information

- Version with issue: 3.6.0
- Last known good version: -
- IDE: VS for Mac & Windows
- Platform Target Frameworks:
  - Tested on iOS: 11.1

### Screenshots

Inital loading:

Refresh from user (pull to refresh):

### Reproduction Link

In the forum i [posted](https://forums.xamarin.com/discussion/comment/368887#Comment_368887) a project to reproduce.

alanag13:
### Description of Change ###
On iOS, if you make a ListView begin refreshing too soon after setting `RefreshControlColor`, the first refresh that occurs after setting the color won't properly reflect. This is actually a known bug in iOS: 

https://github.com/davbeck/RefreshTintFail/
https://github.com/uruly/RefreshControlTintColorDemo

This change applies the workaround mentioned in the above repos to Xamarin.Forms.

### Issues Resolved ### 
- fixes #5728 

### Platforms Affected ### 
- iOS

Users that are setting the refresh spinner color to one other than default color should now see that color reflected immediately. Currently, the refresh indicator may stay its _previous_ color if the refresh was performed very shortly after setting the color.

### Testing Procedure ###
Run the newly created reproduction screen in the core gallery. The refresh indicator should appear and be cyan, not black.

### PR Checklist ###

- [ ] Has automated tests 


alanag13:
@samhouts anything else needed from me here? is the UITest failure related?

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -1624,9 +1624,6 @@ void RemoveRefresh()
 
 		void UpdateContentOffset(nfloat offset, Action completed = null)
 		{
-			if (!_usingLargeTitles)
```
rmarinho: Why remove this?

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
```diff
@@ -242,11 +242,11 @@ protected override void OnElementChanged(ElementChangedEventArgs<ListView> e)
 				UpdateHeader();
 				UpdateFooter();
 				UpdatePullToRefreshEnabled();
+				UpdateSpinnerColor();
```
rmarinho: Does the order matter? if so please add a comment

Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
alanag13: The fix calls this method and we don't want it to only happen when using large titles. Nothing bad happens if you call this for an app without large titles, the offset is still correctly calculated and in fact the animation looks better. I can show in gif attachment tomorrow.

PR #6482, opened 2019-06-10
Fix binding an issue with Source and delayed applying values

rotorgames:
### Description of Change ###

When I used `Source` property in `Binding` I had a problem with delaying of applying values. If I use just `Binding` without `Source` a value is applying in `SetInheritedBindingContext` immediately and I can measure a view size exactly after applying a context. But if I use `Source` property of `Binding` in my xaml layout I get a problem. A value is applying after in `BeginInvokeOnMainThread` and it happens after my code tries to measure an element. It's really critical when I work with lists because I want to know a size instantly after applying a context.

I added checking it is invoking in the main thread or not.

Xaml code where I had this problem:
```xaml
<ContentView
    x:Class="Xamarin.Forms.Sandbox.TestView"
    x:Name="This">
    <Label 
        Text="{Binding Source={x:Reference This}, Path=Text}"/>
</ContentView>
```
```csharp
namespace Xamarin.Forms.Sandbox
{
	public partial class TestView : ContentView
	{
		public static readonly BindableProperty TextProperty = BindableProperty.Create(nameof(Text), typeof(string), typeof(TestView));

		public string Text
		{
			get { return (string)GetValue(TextProperty); }
			set { SetValue(TextProperty, value); }
		}

		public TestView()
		{
			InitializeComponent();
		}
	}
}
```

### Issues Resolved ### 

None

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/BindingExpression.cs
+++ b/Xamarin.Forms.Core/BindingExpression.cs
@@ -615,7 +615,10 @@ namespace Xamarin.Forms
 					}
 				}
 
-				Device.BeginInvokeOnMainThread(() => _expression.Apply());
+				if (Device.IsInvokeRequired)
+					Device.BeginInvokeOnMainThread(() => _expression.Apply());
+				else
+					_expression.Apply();
 			}
 
 			public bool TryGetValue(object source, out object value)


samhouts:
@rotorgames Are you able to add a unit test? Thanks!

rotorgames:
@samhouts Sure. I had a day off. I'm going to do it ASAP.

rotorgames:
I nice to see that this issue was fixed some days ago in 5bc931f0892bf6e89cb8b752ae6a1c756fbc5fdb. This PR doesn't make sense. 

Closed.

PR #6511, opened 2019-06-13
Resolve fallback type before trying to make constructed generic type

TimBarham:
### Description of Change ###

The `FallbackTypeResolver` hander in the Forms previewer does not have access to the `x:TypeArguments` attribute, so if it constructs a proxy generic type from metadata it can't construct it. So the previewer has to create an unconstructed generic type, and Forms will need to construct it.

So we need to move the code that constructs the generic type after the call to `FallbackTypeResolver`.

### Issues Resolved ### 

Part of the fix for https://devdiv.visualstudio.com/DevDiv/_workitems/edit/917689

### API Changes ###
 
 None

### Platforms Affected ### 
- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###
Testing requires previewer code that isn't yet merged.

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
+++ b/Xamarin.Forms.Xaml.UnitTests/DesignTimeLoaderTests.cs
@@ -285,6 +285,25 @@ namespace Xamarin.Forms.Xaml.UnitTests
 		[Test]
 		public void UnknownGenericType()
 		{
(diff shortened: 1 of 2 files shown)


StephaneDelcroix:
please retarget that to 4.1.0

TimBarham:
Thanks @StephaneDelcroix. Retargeted.

PR #6512, opened 2019-06-13
[ios] change checkbox to use default constructor

PureWeen:
### Description of Change ###
Change the FormsCheckbox inheriting from UIBUtton to not use the UIButtonType base constructor which isn't recommended. If you use the UIButtonType base constructor it throws exceptions. This PR changes it to use the default constructor

### Platforms Affected ### 
- iOS

### Testing Procedure ###
Bring up the CheckBox in the dynamic control gallery and play with the colors and various other settings to make sure it still all works correctly

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/FormsCheckBox.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/FormsCheckBox.cs
@@ -30,7 +30,7 @@ namespace Xamarin.Forms.Platform.iOS
 			}
 		}
 
-		public FormsCheckBox() : base(UIButtonType.System)
+		public FormsCheckBox()
 		{
 			TouchUpInside += OnTouchUpInside;
 			ContentMode = UIViewContentMode.Center;
@@ -140,10 +140,10 @@ namespace Xamarin.Forms.Platform.iOS
 			}
 
 			if (_checked == null)
-				_checked = CreateCheckBox(CreateCheckMark());
+				_checked = CreateCheckBox(CreateCheckMark()).ImageWithRenderingMode(UIImageRenderingMode.AlwaysTemplate);
 
 			if (_unchecked == null)
-				_unchecked = CreateCheckBox(null);
+				_unchecked = CreateCheckBox(null).ImageWithRenderingMode(UIImageRenderingMode.AlwaysTemplate);
 
 			return IsChecked ? _checked : _unchecked;
 		}


paymicro:
It seems to me that the `checkbox` is out of the general style, because 30% less than native controls 
Screenshot_1

PureWeen:
@paymicro yea the default one could probably stand to be increased in size a bit.

Did you compare the material ones between ios and Android by chance? The sizes I picked were really just to match android.

jamesmontemagno:
Agreed on size. I think mine was a bit bigger originally?

jamesmontemagno:
This change looks fine. I don't remembering doing that though, but looks solid change.

PureWeen:
@jamesmontemagno 

> Agreed on size. I think mine was a bit bigger originally?

Yea this was a little bit of a tricky one. Originally you had the HeightRequest on ios wired up to the drawing so it made a larger checkbox. Problem is that it didn't work on any of the other platforms and there's not a super straight forward way to make it resize on Android. Maybe down the road @paymicro could work the same magic he did on slider :-) so for now it all just sizes to a made up default that was just picked to match android. Things like the Switch and those also don't resize so should checkbox? So that leaves us with just making up some height that matches a made up control on iOS. Switch seems like a valid control to size against.  Maybe just match the height of a switch and use that for the wxh

> This change looks fine. I don't remembering doing that though, but looks solid change.

Original implementation just inherited from UIView but I changed it to UIButton to be more consistent with how other platforms do it and also this allows us to easily add text

PureWeen:
alright @paymicro and @jamesmontemagno I increased the size of the default checkbox to just match the height of the switch control

I also just setup DefaultSize as a settable property so if someone creates a custom ios renderer they can easily change the height

image

paymicro:
for full compliance, it remains to add the same shadows as on the switch control

PureWeen:
@paymicro can you log an issue for the shadow?

jamesmontemagno:
Should there be a shadow? hard to say... i mean looking at iOS apps there aren't at all. This control we are sort of making up for iOS based on apps like Reminder.

paymicro:
@jamesmontemagno by default the switch circle has a shadow
image
but there are more important things to do...

jamesmontemagno:
AHHHHHHHH I see what you mean now, this wasn't for checkbox... i was confused.

PR #6542, opened 2019-06-15
[Shell] NavigationHandler hooks

### LINKED ISSUE #5166: [Enhancement] Provide Handler for Shell Navigation
## Summary

Currently there is no good way of easily integrating an App's framework into Shell. To make this easier a Handler should be provided so that a Framework can simply provide Forms a custom implementation for the core Navigation functionality.

## API Changes

These methods should all be exposed though an Interface so that it could be potentially implemented in a stand alone service or on the Application itself since this is where you would most likely have a direct reference to a DI Container.
- Create View from Shell Content... the default delegate could simply expand on ContentTemplate, where a custom delegate may utilize Dependency Injection to create the view from the route segment name.
- Handle Can Navigate From
- Handle Will Navigate To
- Handle Did Navigate From/To

## Intended Use Case

To allow frameworks like Prism to more easily integrate with Shell.

PureWeen:
### Description of Change ###

First set of proposed changes for shell NavigationHandler hooks

- fixes #5166

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/ShellNavigationTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/ShellNavigationTests.cs
@@ -1,0 +1,44 @@
+﻿using System;
+using System.Linq;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using Xamarin.Forms.Internals;
+
+namespace Xamarin.Forms.Core.UnitTests
+{
+	[TestFixture]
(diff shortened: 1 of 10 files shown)


samhouts:
@PureWeen You need to fix your commits. It's driving my script crazy. 🦀

larsduewel:
@samhouts @PureWeen Will this pull request be merged anytime soon? We want to integrate Shell into our application, but would like to use the ShellNavigationService.

Issue #1778, opened 2018-02-01
[Enhancement] Popup Control


hartez:
## Rationale 

Forms users would like to have the option of defining and displaying Popup controls. These are controls which are modal, can host arbitrary content, allow for complex user interaction, can return a response from the user, and are expressed (by default) using the appropriate controls on each platform:

- iOS – UIView presented by a UIPopoverPresentationController
- UWP – Flyout
- Android – Dialog

## Requirements

Popup display must be asynchronous.

Popups must be effectively modal; i.e., only one may be displayed at any given time.

Popups must be light-dismissable where the platform allows for it (e.g., on iPad, they should be light-dismissable; on iPhone this is not an option on all versions).

Users must be able to define their own custom Popups with arbitrary return value types.

Users must be able to host the following in a Popup:

- View 
- ContentPage
- NavigationPage

This will effectively mean that Popups must provide for hosting VisualElement; for VisualElement types which are not View, any type other than ContentPage or NavigationPage will throw a `NotSupportedException`.

Popups must allow for (optionally) specifying an "anchor" View; behavior regarding the anchor will be platform-specific:

- On iPad, the arrow of the presented PopOver will point to the anchor View. On iPhone, the anchor view will have no effect.
- On Android, the Dialog will be presented near the anchor View.
- On UWP, the anchor View will be the View to which the Flyout will be attached.

Popups must allow for a `Size` specification. If no `Size` is specified, the Popup will size to its content.

Forms should define a convenience Popup subclass for returning a 1-bit response (i.e., a yes/no, true/false, ok/cancel). In addition to providing convenience for users dealing with this common requirement, it will also serve as an example for users who need to implement other custom scenarios.

Forms should define a convenience Popup subclass for returning (effectively) a `void` value (in order to easily support users wanting to display simple content with no user response).

## API Changes

The `INavigation` interface will add the following method:

```csharp
Task<T> ShowPopupAsync<T>(Popup<T> popup);
```

The following interface will be added to Core:

```csharp
public interface IPopup<out T>
{
	VisualElement Element { get; }

	void SetDismissDelegate(Action<T> dismissDelegate);
}
```
	
The Dismiss delegate is invoked when the Popup is dismissed. 

The following classes will be added to Core:
```csharp
public abstract class Popup<T>
```

This will be the base class from which all Popups will be built and will provide the core properties and functionality.

```csharp
public struct PopupDismissed{}
public class Popup : Popup<PopupDismissed>
```

[...]

PR #6558, opened 2019-06-17
Modal Page with transparent background

andreinitescu:
### Description of Change ###

This PR changes the background color of the **modal container** of a **modal page**, from white to transparent. This way, if the **modal page** has the background color with transparency, the content behind the modal page is now visible. In other words, this PR makes the opacity of the `BackgroundColor` on a Page really work (every Xamarin.Forms developer must have tried this and discovered it doesn't work ?? ).

With this change, there's now a very easy way to finally be able to create **popup pages**, and the nice thing is it can be done using the existing simple modal navigation API.

Creating a popup page only requires a `ContentPage` with a semi-transparent `BackgroundColor`, and some content not covering the whole page (the content position can be customized easily).

```
             BackgroundColor="#A0000000"
             x:Class="Xamarin.Forms.Controls.GalleryPages.PageWithTransparentBkgnd">
    <Frame VerticalOptions="Center"
           HorizontalOptions="Center"
           WidthRequest="300"
           HeightRequest="200"
           Padding="15, 10"
           BackgroundColor="LightGreen">
        <StackLayout>
            <Label Text="Welcome to Xamarin.Forms Popups!"
                   HorizontalOptions="Center"
                   VerticalOptions="CenterAndExpand" />
            <Button Text="Close"
                    HorizontalOptions="End"
                    Clicked="ClosePageButtonClicked"
                    VerticalOptions="End" />
        </StackLayout>
    </Frame>
</ContentPage
```

Show the page as a modal page:
```
	void ShowModalBtnClicked(object sender, System.EventArgs e)
	{
		Navigation.PushModalAsync(new PageWithTransparentBkgnd());
	}
```

Closing the page is done using existing API:

		void ClosePageButtonClicked(object sender, EventArgs e)
		{
			Navigation.PopModalAsync();
		}

On Android, since the "popup" is a simple modal page, the back button just works.

Once this is implemented, more features can be added, for example popping the modal page automatically when tapping anywhere on the page except its content.

### Don't animate the overlay, only the content

If someone doesn't want to have the overlay animated but only the Page content, this can be achieved relatively easily by calling `Navigation.PushModalAsync(modalPage, animation: false)` and animating the page content.

### Animate the overlay and/or content

Similar to the previous, with `Navigation.PushModalAsync(modalPage, animation: false)` this can be done by:
- Set the Page `BackgroundColor="Transparent"`
- Add an overlay using a `BoxView` with a semi transparent `BackgroundColor=#A0000000`
- Animate the `BoxView` overlay (animate the opacity, animate scaling it, etc.)

Popups are a very common feature, so if possible it would be great to get this feature sooner.

### API Changes ###

The only public API change is on iOS, where I added a new value for the UIModalPresentationStyle platform specific enum. This is needed because the default presentation style for the modal container changed from `UIModalPresentationStyle.FullScreen` to `UIModalPresentationStyle.OverCurrentContext` in order to support a transparent background. 

AFAIK, this change should not have any backward compatibility implications, since for the page modal container, the OverCurrentContext value behaves exactly like the previous value, FullScreen.

### Issue fixed ###

This doesn't (completely) fix the #1778, but I think it opens up new possibilities with very little changes. 

### Platforms Affected ### 

- iOS
- Android

None. If you don't set semi-transparent background, there's no visual change. This is because all pages have the white color as background color by default.

### Screenshots ### 

modal

modal_android

### Testing Procedure ###

See the "Modal With Transparent Bkgnd Gallery - Legacy" page.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Page.cs
+++ b/Xamarin.Forms.Core/Page.cs
@@ -40,6 +40,8 @@ namespace Xamarin.Forms
 		[EditorBrowsable(EditorBrowsableState.Never)]
 		public static readonly BindableProperty IconProperty = IconImageSourceProperty;
 
+		public static readonly BindableProperty ModalBackgroundColorProperty = BindableProperty.Create(nameof(ModalBackgroundColor), typeof(Color), typeof(Page), Color.Default);
+
 		readonly Lazy<PlatformConfigurationRegistry<Page>> _platformConfigurationRegistry;
 
 		bool _allocatedFlag;
@@ -143,6 +145,12 @@ namespace Xamarin.Forms
 
         [EditorBrowsable(EditorBrowsableState.Never)]
         public ObservableCollection<Element> InternalChildren { get; } = new ObservableCollection<Element>();
+
+		public Color ModalBackgroundColor
+		{
+			get { return (Color)GetValue(ModalBackgroundColorProperty); }
+			set { SetValue(ModalBackgroundColorProperty, value); }
+		}
 
 		internal override IEnumerable<Element> ChildrenNotDrawnByThisElement
 		{
--- a/Xamarin.Forms.Platform.iOS/ModalWrapper.cs
+++ b/Xamarin.Forms.Platform.iOS/ModalWrapper.cs
@@ -2,6 +2,7 @@
 using System.Linq;
 using Xamarin.Forms.PlatformConfiguration.iOSSpecific;
 using UIKit;
+using System.ComponentModel;
 
 namespace Xamarin.Forms.Platform.iOS
 {
@@ -14,15 +15,16 @@ namespace Xamarin.Forms.Platform.iOS
 			_modal = modal;
 
 			var elementConfiguration = modal.Element as IElementConfiguration<Page>;
-			if (elementConfiguration?.On<PlatformConfiguration.iOS>().ModalPresentationStyle() == PlatformConfiguration.iOSSpecific.UIModalPresentationStyle.FormSheet)
-				ModalPresentationStyle = UIKit.UIModalPresentationStyle.FormSheet;
+			if (elementConfiguration?.On<PlatformConfiguration.iOS>()?.ModalPresentationStyle() is PlatformConfiguration.iOSSpecific.UIModalPresentationStyle style)
+				ModalPresentationStyle = style.ToNativeModalPresentationStyle();
 
-			View.BackgroundColor = UIColor.White;
+			UpdateBackgroundColor();
 			View.AddSubview(modal.ViewController.View);
 			TransitioningDelegate = modal.ViewController.TransitioningDelegate;
 			AddChildViewController(modal.ViewController);
 
 			modal.ViewController.DidMoveToParentViewController(this);
+			((Page)modal.Element).PropertyChanged += OnModalPagePropertyChanged;
 		}
 
 		public override void DismissViewController(bool animated, Action completionHandler)
@@ -89,14 +91,20 @@ namespace Xamarin.Forms.Platform.iOS
 
 		public override void ViewWillAppear(bool animated)
 		{
-			View.BackgroundColor = UIColor.White;
+			UpdateBackgroundColor();
 			base.ViewWillAppear(animated);
 		}
 
 		protected override void Dispose(bool disposing)
 		{
 			if (disposing)
+			{
+				if (_modal?.Element is Page modalPage)
+					modalPage.PropertyChanged -= OnModalPagePropertyChanged;
+
 				_modal = null;
+			}
+
 			base.Dispose(disposing);
 		}
 
@@ -111,6 +119,18 @@ namespace Xamarin.Forms.Platform.iOS
 		public override UIViewController ChildViewControllerForStatusBarStyle()
 		{
 			return ChildViewControllers?.LastOrDefault();
+		}
+
+		void OnModalPagePropertyChanged(object sender, PropertyChangedEventArgs e)
+		{
+			if (e.PropertyName == Page.ModalBackgroundColorProperty.PropertyName)
+				UpdateBackgroundColor();
+		}
+
+		void UpdateBackgroundColor()
+		{
+			Color modalBkgndColor = ((Page)_modal.Element).ModalBackgroundColor;
+			View.BackgroundColor = modalBkgndColor.IsDefault ? UIColor.White : modalBkgndColor.ToUIColor();
 		}
 	}
 }
--- a/Xamarin.Forms.Platform.iOS/Platform.cs
+++ b/Xamarin.Forms.Platform.iOS/Platform.cs
@@ -6,6 +6,7 @@ using CoreGraphics;
 using Foundation;
 using UIKit;
 using Xamarin.Forms.Internals;
+using Xamarin.Forms.PlatformConfiguration.iOSSpecific;
 using RectangleF = CoreGraphics.CGRect;
 
 namespace Xamarin.Forms.Platform.iOS
@@ -112,6 +113,9 @@ namespace Xamarin.Forms.Platform.iOS
 
 			modal.DisposeModalAndChildRenderers();
 
+			if (IsModalPresentedOverContext(modal))
+				GetCurentPage(Page)?.SendAppearing();
+
 			return modal;
 		}
 
@@ -143,6 +147,9 @@ namespace Xamarin.Forms.Platform.iOS
 		Task INavigation.PushModalAsync(Page modal, bool animated)
 		{
 			EndEditing();
+
+			if (_appeared && IsModalPresentedOverContext(modal))
+				GetCurentPage(Page)?.SendDisappearing();
 
 			_modals.Add(modal);
 
@@ -561,6 +568,25 @@ namespace Xamarin.Forms.Platform.iOS
 				modal.DisposeModalAndChildRenderers();
 
 			(Page.Parent as IDisposable)?.Dispose();
+		}
+
+		Page GetCurentPage(Page currentPage)
+		{
+			if (_modals.LastOrDefault() is Page modal)
+				return modal;
+			else if (currentPage is MasterDetailPage mdp)
+				return GetCurentPage(mdp.Detail);
+			else if (currentPage is IPageContainer<Page> pc)
+				return GetCurentPage(pc.CurrentPage);
+			else
+				return currentPage;
+		}
+
+		static bool IsModalPresentedOverContext(Page modal)
+		{
+			var elementConfiguration = modal as IElementConfiguration<Page>;
+			var presentationStyle = elementConfiguration?.On<PlatformConfiguration.iOS>()?.ModalPresentationStyle();
+			return presentationStyle != null && presentationStyle == PlatformConfiguration.iOSSpecific.UIModalPresentationStyle.OverFullScreen;
 		}
 	}
 }
--- a/Xamarin.Forms.Controls/CoreGallery.cs
+++ b/Xamarin.Forms.Controls/CoreGallery.cs
@@ -409,6 +409,7 @@ namespace Xamarin.Forms.Controls
 				new GalleryPageFactory(() => new ViewCellGallery(), "ViewCell Gallery - Legacy"),
 				new GalleryPageFactory(() => new WebViewGallery(), "WebView Gallery - Legacy"),
 				new GalleryPageFactory(() => new BindableLayoutGalleryPage(), "BindableLayout Gallery - Legacy"),
+				new GalleryPageFactory(() => new ShowModalWithTransparentBkgndGalleryPage(), "Modal With Transparent Bkgnd Gallery - Legacy"),
 			};
 
 		public CorePageView(Page rootPage, NavigationBehavior navigationBehavior = NavigationBehavior.PushAsync)
--- a/Xamarin.Forms.Controls/GalleryPages/PageWithTransparentBkgnd.xaml
+++ b/Xamarin.Forms.Controls/GalleryPages/PageWithTransparentBkgnd.xaml
@@ -1,0 +1,29 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             xmlns:d="http://xamarin.com/schemas/2014/forms/design"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             mc:Ignorable="d"
+             x:Class="Xamarin.Forms.Controls.GalleryPages.PageWithTransparentBkgnd"
+             xmlns:ios="clr-namespace:Xamarin.Forms.PlatformConfiguration.iOSSpecific;assembly=Xamarin.Forms.Core"
+             ios:Page.UseSafeArea="true"
+             BackgroundColor="Transparent"
+             ModalBackgroundColor="#700000FF"
+             ios:Page.ModalPresentationStyle="OverFullScreen">
+    <Frame VerticalOptions="Center"
+           HorizontalOptions="Center"
+           WidthRequest="300"
+           HeightRequest="200"
+           Padding="15, 10"
+           BackgroundColor="LightGreen">
+        <StackLayout>
+            <Label Text="Modal page with custom modal background"
+                   HorizontalOptions="Center"
+                   VerticalOptions="CenterAndExpand" />
+            <Button Text="Close"
+                    HorizontalOptions="End"
+                    Clicked="ClosePageButtonClicked"
+                    VerticalOptions="End" />
+        </StackLayout>
+    </Frame>
+</ContentPage>
--- a/Xamarin.Forms.Controls/GalleryPages/PageWithTransparentBkgnd.xaml.cs
+++ b/Xamarin.Forms.Controls/GalleryPages/PageWithTransparentBkgnd.xaml.cs
@@ -1,0 +1,19 @@
+﻿using System;
+using Xamarin.Forms.Internals;
+
+namespace Xamarin.Forms.Controls.GalleryPages
+{
+	[Preserve(AllMembers = true)]
+	public partial class PageWithTransparentBkgnd : ContentPage
+	{
+		public PageWithTransparentBkgnd()
+		{
+			InitializeComponent();
+		}
+
+		void ClosePageButtonClicked(object sender, EventArgs e)
+		{
+			Navigation.PopModalAsync();
+		}
+	}
+}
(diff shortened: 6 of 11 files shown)


rmarinho:
We will need to make some work for IOS13 i wonder if we can do this at that time.

andreinitescu:
@rmarinho Can I rebase this against 4.1.0 so we have it released asap?

andreinitescu:
Rebased against 4.1.0

rmarinho:
I don t think we can put this on 4.1 as we are cutting a stable. but we will run tests now and see

jfversluis:
@ahoefling @hartez is this something we want to have next to #1778? Effectively we would then have two ways of showing popups which might be confusing. Don't get me wrong @andreinitescu I very much appreciate the effort! Just thinking out loud.

Also, I think there are a couple of breaking changes in this PR.

[Here](https://github.com/xamarin/Xamarin.Forms/pull/6558/files#diff-d379831cb9daec7e5a1e569f9632d03eR134) you change the default value to this new way of showing popups, potentially breaking the current implementation for people.

And [here](https://github.com/xamarin/Xamarin.Forms/pull/6558/files#diff-d8840725e3c949d73630647dd7085a80R20) you change the default background color to clear instead of white. Although it's just a styling issue people might see different visuals popping up without them changing anything.

I know you mentioned these in your PR already, I think we just need to be sure on this.

ahoefling:
@andreinitescu this is a great PR and works well as far as I can tell!
@jfversluis thanks for bringing this up, as this is something I have been monitoring.

The thing that makes #1778 very difficult, is it defines the APIs that must be used natively which would encapsulate a `ContentPage`, `NavigationPage` or a `View`. For example on Android it requires use of the `Dialog` and on UWP it requires the `Flyout`. These are native elements for rendering popups in those platforms, which will give developers great extension points to customize their popup or even take advantage of specific navtive APIs in a custom renderer. I fully agree with @hartez original spec in #1778 that we should focus any popup implementation using the native controls.

This PR provides a stopgap for #1778 in my opinion as it doesn't implement every part of the spec, which is identified as a partial completion in the PR description. I am in agreement with @jfversluis and I don't think we should have 2 competing specs or even a partial spec we refine later. Once we ship a new feature we can't make any large breaking changes. If we accept this as a stopgap feature for #1778 and when that feature is fully implemented it will break many apps out there. Providing a migration path is not really a great option either. All things I considered so we could ship this code and then swap it out with the final implemenation of #1778 when it is ready.

As far as #1778 is concerned, I have been doing a lot of work on it and have a working implementation for Android and UWP, I have not had time to circle back for the iOS implementation and plan to get to it later this summer.

andreinitescu:
While I mentioned about popups and #1778, this is not meant as a fix for #1778.  I mentioned it in the original description (I edited it, you need to see the previous edited version). It clearly does not come with any new **popup** API, it's still only about the **modal**.

I don't understand why would it be confusing. It's not meant to tell developers "here's how you **should** do popups", #1778 is for that, it has a specific and richer API (ability to have buttons).
You could do "popups"/"modals" today by just overlaying ContentView in a Page and that doesn't confuse anybody I think with the current modals.

I also mentioned in the description why I think it should not break existing apps. If I miss anything, please let me know exactly how is it breaking. The modal background is not visible at all because the modal wrapper is always completely filled by the page and the page has its own default color. Again, if I miss something let me know what is it exactly and I can try fix it. Maybe there's another way.

Also, my PR and tweet is not trying to steal or put in shadow @ahoefling 's effort on #1778, it's a great effort and I'm looking forward to see that ticket finally implemented . I actually responded to him on Twitter that my PR and his PR are different things.

andreinitescu:
Like I mentioned in the original description, this gives a very raw way to create popups. And I think there's still useful even with #1778, in the case someone wants a very flexible way to show popups. (Maybe I should not have edited the description, I described this extensively. Well, it's still visible if you click on the "edited" button and see the previous version).

ahoefling:
This is some really great work you did here! 

I think the messaging was confusing as it mentioned the Popup spec but as you (@andreinitescu) have mentioned it doesn't really implement it. I see in the files changed you updated the enum to allow for a transparent background and setting the background color. Then in the platform specific code you implement the colors being configured as the background color. This effectively gives us the transparent effect you have in the screenshots.

It looks like the API does not change for displaying a modal screen, correct?

```c#
NavigationPage.ShowModalAsync(new SomePage());
```

This may still cause a breaking change for anyone that has already entered a `BackgroundColor` as this will start to change the behavior. While a minor breaking change, I can see someone upgrading their app and the modal feature starts working differently than how it did in the previous version. Again, _I see this as a minor break_

I still have some concerns about this enhancement being merged. **I can be convinced otherwise**. Merging this and #1778 would effectively create 2 ways to render the exact same popup/modal, granted the Popup Spec in #1778 has callbacks and this does not.

@andreinitescu 
My reply was not intended to discredit the work you did here, and I don't think it does to the work I am doing for the Popup Spec.

jfversluis:
> Also, my PR and tweet is not trying to steal or put in shadow @ahoefling 's effort on #1778, it's a great effort and I'm looking forward to see that ticket finally implemented . I actually responded to him on Twitter that my PR and his PR are different things.

Don't worry @andreinitescu ! Nobody is saying or thinking this in any way!
The only thing I'm not sure about is if we merge this and later (hopefully) #1778 we would have two ways of showing popups. The only thing I'm wondering is if that is desirable. Please don't think that I (or we) do not appreciate this effort, because we really appreciate all of your work :) 

I understand these are two different things, but in the end, effectively they achieve the same thing. That was the only reason I am thinking: should we end up with two ways of doing the same thing?

andreinitescu:
> That was the only reason I am thinking: should we end up with two ways of doing the same 
thing?

Thanks Gerald, I'm happy to discuss about concrete things.  Here's why I think this PR is still worth it:

1.  #1778 has an opinionated and very specific APIs to show dialogs/popups. It has API to return value, show buttons, manage navigation to multiple dialogs (it says "_only one may be displayed at any given time_"), etc. Which is good, it's practical.

2. This PR gives indirectly another way to achieve it based on modals, by controlling the Page modal background. There's obviously no new API, and it's up to developer to create his API based on the `Navigation.PushModalAsync()` API, and decide how to return a value, how to show buttons, etc. 

    It gives a similar flexibility to how it's on native. On Android, you could start off from a raw approach using android.app.Dialog (btw, I wonder if it wouldn't better if on Android, the modal page implementation should use Dialog, I'm not sure why the `ModalWrapper` custom view, but this for another discussion/ticket). On iOS, you could use the modal view-controller (i.e. `presentModalViewController()` )
  
    The more I think about it, **I'm not sure why modal pages were designed so you cannot control the modal background,** it's a limitation which, if you need popups, it requires you go find another route, either use a third-party implementation or do a lot of unnecessary work.

3. This PR should be much much faster to release than #1778 which unless I'm mistaken, it's months away (2,3,4,?) from being ready to be released. We don't know how many.

Let me know if you have more thoughts, I'm happy to discuss. It's obviously up to you (the XF team) to decide.

Xamarin.Forms.Platform.iOS/ModalWrapper.cs
```diff
@@ -14,10 +14,10 @@ internal ModalWrapper(IVisualElementRenderer modal)
 			_modal = modal;
 
 			var elementConfiguration = modal.Element as IElementConfiguration<Page>;
-			if (elementConfiguration?.On<PlatformConfiguration.iOS>().ModalPresentationStyle() == PlatformConfiguration.iOSSpecific.UIModalPresentationStyle.FormSheet)
-				ModalPresentationStyle = UIKit.UIModalPresentationStyle.FormSheet;
+			if (elementConfiguration?.On<PlatformConfiguration.iOS>()?.ModalPresentationStyle() is PlatformConfiguration.iOSSpecific.UIModalPresentationStyle style)
+				ModalPresentationStyle = style.ToNativeModalPresentationStyle();
 
-			View.BackgroundColor = UIColor.White;
+			View.BackgroundColor = UIColor.Clear;
```
PureWeen: I think this needs to stay as White if the Style requested is FullScreen otherwise it changes the look of a Full Screen Modal that has a transparent background.

andreinitescu:
I think I found a different way which I think it's cleaner and it doesn't modify the behavior of existing apps. 
Instead of modifying current behavior, I added a specific property control the modal container background color:
```
Page.ModalBackgroundColor { get; set; } \\ BP
```

This color is set to the modal wrapper of the modal page.
This is also useful in the rare case someone wants to modify the background color when using [the existing `PageSheet` modal presentation style](https://docs.microsoft.com/en-us/dotnet/api/xamarin.forms.platformconfiguration.iosspecific.uimodalpresentationstyle?view=xamarin-forms).

With this change, here's how a ContentPage can be made to have background with a specific transparent color:

```
<ContentPage ...
             x:Class="Xamarin.Forms.Controls.GalleryPages.PageWithTransparentBkgnd"
             ios:Page.UseSafeArea="true"
             BackgroundColor="Transparent"
             ModalBackgroundColor="#700000FF"
             ios:Page.ModalPresentationStyle="OverFullScreen">
```

Since the modal page fully covers the screen on both Android and iOS, it allows implementing "close modal when tap on outside the page".

There's one thing I am not sure about is on iOS, how to get the current page in order to send disappearing/appearing events when modal is presented/dismissed (because the view controller life cycle are not sent to page presenting the modal by iOS when using "OverContent" modal presentation style).
The way I did it is by computing the current page [here](https://github.com/andreinitescu/Xamarin.Forms/blob/7bb10cdb8e6be4e42c9b12646188d6a4acec88dd/Xamarin.Forms.Platform.iOS/Platform.cs#L573)  but I think there's a better way.

What do you think this approach? I think `ModalBackgroundColor` is a simple way. Maybe naming could be `ModalContainerBackgroundColor` not to be confusing with the page background color.

Xamarin.Forms.Platform.iOS/Platform.cs
```diff
@@ -144,6 +148,9 @@ Task INavigation.PushModalAsync(Page modal, bool animated)
 		{
 			EndEditing();
 
+			if (_appeared && IsModalPresentedOverContext(modal))
```
PureWeen: I think we want to invert this check so that it's not super specific to OverContext. The pseudocode I would imagine for this would go like this

```C#

bool shouldFire = true;

if( iOS13OrHigher)
{
       if(presentationStyle == FullScreen) shouldFire = false; //this is mainly for backwards compatibility
}
else
{
      if(presentationStyle == Automatic) shouldFire = false;
      else if(presentationStyle == FullScreen) shouldFire = false; //this is mainly for backwards compatibility
}

if (_appeared && IsModalPresentedOverContext(modal))
```

This way it'll just work for whatever other presentation styles come up

Xamarin.Forms.Platform.iOS/ModalWrapper.cs
```diff
@@ -112,5 +120,17 @@ public override UIViewController ChildViewControllerForStatusBarStyle()
...
+		void OnModalPagePropertyChanged(object sender, PropertyChangedEventArgs e)
+		{
+			if (e.PropertyName == Page.ModalBackgroundColorProperty.PropertyName)
+				UpdateBackgroundColor();
+		}
+
+		void UpdateBackgroundColor()
+		{
+			Color modalBkgndColor = ((Page)_modal.Element).ModalBackgroundColor;
+			View.BackgroundColor = modalBkgndColor.IsDefault ? UIColor.White : modalBkgndColor.ToUIColor();
```
PureWeen: What's the advantage to having ModalBackgroundColor? Opposed to just doing something like this
```C#
Color modalBkgndColor = ((Page)_modal.Element).BackgroundColor;
View.BackgroundColor = modalBkgndColor.IsDefault ? UIColor.White : modalBkgndColor.ToUIColor();
```

PureWeen: Actually I see why now... 
Is there ever a reason to set the ModalBackgroundColor and BackgroundColor? Or will you always set one of them to Transparent?

If we just added an Attached Property to indicate if the Modal should be transparent or not

```C#
bool Page.ModalBackgroundIsTransparent
```

Would that achieve the same thing?

PureWeen: Could we just set it to Transparent if you use anything except FullScreen for the modal behavior?

Xamarin.Forms.Core/Page.cs
```diff
@@ -40,6 +40,8 @@ public class Page : VisualElement, ILayout, IPageController, IElementConfigurati
 		[EditorBrowsable(EditorBrowsableState.Never)]
 		public static readonly BindableProperty IconProperty = IconImageSourceProperty;
 
+		public static readonly BindableProperty ModalBackgroundColorProperty = BindableProperty.Create(nameof(ModalBackgroundColor), typeof(Color), typeof(Page), Color.Default);
```
StephaneDelcroix: It's uncommon to have a single page used both in modal and non-modal scenarios, so the added property is not required, we could use the "default" one.

in the rare case the user wants to push a modal with a different bg, he knows it at the time of the push, and can change the bg.

XXcharlosXX:
Oh how I look forward to this!!

XXcharlosXX:
Ah shucks, I was hoping for this in 4.2
Thank you guys for your hard work!

XXcharlosXX:
I will continue to check in on this. I'd much rather use this than install a nuget package to my projects.
I know you guys are working hard on the recent updates to the OS's, etc. Keep up the good work!

XXcharlosXX:
Oh please add this to next release of XF!!

PureWeen:
@andreinitescu can you rebase and address changes?

Or do you want us to take this one over?

sasivishnu:
> 
> 
> @andreinitescu can you rebase and address changes?
> 
> Or do you want us to take this one over?

It will be very helpful for us to get this feature in 4.3

sasivishnu:
@PureWeen any update on when we can expect this feature to be added?

sasivishnu:
@davidortinau @PureWeen @andreinitescu @samhouts 
Hi, Sorry for repeated request but having this feature earlier will help our business app. I understand there may be too much on plate for XF team but could we atleast get some ETA? Can we expect for atleast 4.4? 

Thanks again. If any help needed to get this PR done, I am willing to help.

jfversluis:
@sasivishnu Thanks for asking! I'm afraid it's unlikely that this will make it into 4.4. We are still looking into this. I think we are ready to take over this one ourselves to get it in. Unless @PureWeen thinks otherwise.

Xamarin.Forms.Platform.iOS/Platform.cs
jfversluis: Then change `if (_appeared && IsModalPresentedOverContext(modal))` to `if (_appeared && shouldFire)`? And how about this one: https://github.com/xamarin/Xamarin.Forms/pull/6558/files/7bb10cdb8e6be4e42c9b12646188d6a4acec88dd#diff-7213603536449d008b596d1e2c2b15f4R116 ?

XXcharlosXX:
So not in 4.4, will it be 4.5?

jfversluis:
@XXcharlosXX we will do our absolute best to get it in 4.5, stay tuned

XXcharlosXX:
fingers crossed!

XXcharlosXX:
hoping y'all are able to squeeze this into 4.5!!

XXcharlosXX:
Is this live?

ahoefling:
@XXcharlosXX looks like it was moved over to #8551

dinesh4official:
@samhouts , is there any documentation, how to use this. I can't see any transparent modal pages in v4.6

jfversluis:
I don't think there is yet... @davidbritch do you know anything about that?

AlleSchonWeg:
Try this. You need min version 4.6.
```
<ContentPage
...
BackgroundColor="Transparent"
xmlns:ios="clr-namespace:Xamarin.Forms.PlatformConfiguration.iOSSpecific;assembly=Xamarin.Forms.Core"
ios:Page.ModalPresentationStyle="OverFullScreen">

<StackLayout HorizontalOptions="FillAndExpand"

</StackLayout>

</ContentPage>
```
Source: https://stackoverflow.com/a/61680286/4506984

davidbritch:
@jfversluis https://docs.microsoft.com/xamarin/xamarin-forms/platform/ios/page-presentation-style

dinesh4official:
> @samhouts , is there any documentation, how to use this. I can't see any transparent modal pages in v4.6

@samhouts  @andreinitescu  @rmarinho , Unfortunately, Transparent Modal background is not working while we push the modal page from tabbed page.

I've attached the sample to replicate the issue with latest XF version.

**Replication Procedure:**
**Step 1 :** Run the application in iOS platform and select the **TabbedPage Translucent TabBar** item. It will navigate to TabbedPage.
**Step 2 :** . Select the **Return to Platform-Specifics List** button in the view. It will push a new Modal page.
**Issue :** You can see the except the content and background color is turned to **Black** as like below pic.

**Output :**
Screenshot 2020-05-26 at 11 42 20 AM

**Sample :**
[ModalTransparencyIssue.zip](https://github.com/xamarin/Xamarin.Forms/files/4680160/ModalTransparencyIssue.zip)

AlleSchonWeg:
@dinesh4official 
Do you try?
`ios:Page.ModalPresentationStyle="OverFullScreen" `

dinesh4official:
> @dinesh4official
> Do you try?
> `ios:Page.ModalPresentationStyle="OverFullScreen" `

@AlleSchonWeg , yes i tried with that option and the issue still replicates. 

@samhouts , one more scenario to replicate the issue.

**Navigation.PushModalAsync(new NavigationPage(new TransparentPage());**

AlleSchonWeg:
We had the same problem with ios. Not sure if our ios dev has solved it. I thought he changes the ModalPresentationStyle.
Finally we move to https://github.com/rotorgames/Rg.Plugins.Popup for the popup. Because the fadein and fadeout doesn't look nice on ios with semitransparent background.

dinesh4official:
@AlleSchonWeg , thanks

dinesh4official:
@AlleSchonWeg , I've achieved the requirement through custom renderer and the content page appears with semi transparent background smoothly. Thanks :)

brepetti:
> > @dinesh4official
> > Do you try?
> > `ios:Page.ModalPresentationStyle="OverFullScreen" `
> 
> @AlleSchonWeg , yes i tried with that option and the issue still replicates.
> 
> @samhouts , one more scenario to replicate the issue.
> 
> **Navigation.PushModalAsync(new NavigationPage(new TransparentPage());**

Make sure you apply the presentation style & transparent background to the navigation page as well.  It works for me.

dinesh4official:
@brepetti , Good to hear that its working for you. But unfortunately, not for me. Anyhow Xamarin team has verified and logged as a bug report for this.

https://github.com/xamarin/Xamarin.Forms/issues/10861

And I've adapted work around for this until then which I've mentioned in the above link. Hope it may help someone who is link for alternate solution.

Thanks :)

PR #6560, opened 2019-06-17
[Android] Change elevation to use float.minvalue instead of negative infinity

PureWeen:
### Description of Change ###
API 28 on Android appears to be a fan of setting the elevation to negative infinity and expects you to use 
float.minvalue

### Platforms Affected ### 
- Android

### Testing Procedure ###
- Run shell on android api 28 and ensure 5132 is passing on all ui tests
- store shell also has some backgrounds you can interact with for the flyout if you want to test that it all still works to expectation

### PR Checklist ###

- [x] Has automated tests 


PureWeen:
I don't see any 5132 failures on the UI Tests only ToolBarItem failures which we've fixed in a different PR

rmarinho:
Just tested on my app i was hitting this issue with nightly , this works!

PR #6570, opened 2019-06-18
wait a little bit longer for go to test button to appear

PureWeen:
### Description of Change ###
When resetting gallery it will occasionally take a bit more than 10 seconds to get back to the gallery home page before it can go to the nest test


PureWeen:
All the toolbar item tests look to have ran successfully with this one

PR #6572, opened 2019-06-18
[Shell] Flyout TableView height fix and implement margins for Flyout Header

### LINKED ISSUE #6801: [Bug] Shell Flyout - Weird line at the bottom of the items
### Description

There is a strange line at the end of the FlyoutItems / MenuItems. I'm not sure if this is a bug or if it can be disabled somehow? It just looks wrong there (the line at the end of the master menu, not the iOS gesture bar in the center). Also it seems like the items are cut off there instead of scrolling endless (to avoid overlapping with the iOS bar?). 

Simulator Screen Shot - iPhone Xʀ - 2019-07-06 at 18 59 04

### Steps to Reproduce

1. Create a fresh Xamarin.Forms Shell project
2. Change the tabbed page to flyout 
3. Add the items multiple times

### Expected Behavior

It should be at least be configureable if we want the items to go all the way down. Or maybe place a custom footer there? It should respect the SafeArea tho so it needs some kind of scrolling offset. 

### Actual Behavior

Looks strange.

### Basic Information

- Version with issue: 4.1.0
- Last known good version: -
- IDE: VS2019 and VS for Mac
- Platform Target Frameworks: 
  - iOS:  Tested with iOS 12 Simulator

### Screenshots

Already added above in the description.

### Reproduction Link

Reproduction can be done with the VS template within a minute.

PureWeen:
### Description of Change ###

- fix ListView to be full height and not cut off at bottom safe area
- Respect the margins defined on the HeaderView. On iOS if the user hasn't set the margin then it just gets set to the safe area. This allows the user to zero this out if they want the view to be up against the top of the device
- Fix tableview height on iOS to be the full height of the devices 

### Issues Resolved ### 

- fixes #6801

### Platforms Affected ### 
- iOS
- Android

None

*Left is After/ Right is Before*

image

*The TableView now goes all the way to the bottom*

image

### Testing Procedure ###
- Test the store shell with various FH Scroll settings/ height/ and margin6

### PR Checklist ###

- [ ] Has automated tests 


PureWeen:
@paymicro when you review this please check to make sure I merged the changes with your background image changes correctly on Android.

PureWeen:
@paymicro yea the Tableview is supposed to extend the full height of the device. If the user doesn't want it to go above like that then the best option is to create a header view with a margin of zero that goes up against the top of the screen.

PureWeen:
@paymicro alright let me know what you think of latest commit. I realize the bottom still goes into the safe area but that'll be resolved once we add a footer view. Then it'll just work the same as the header view if someone wants to add a margin

paymicro:
@PureWeen latest commit works great. 
But if `FlyoutHeaderBehavior` is Scroll then the table and the header go beyond the edge of the safe area. Is this the expected behavior?

PureWeen:
@paymicro Try now.

PR #6588, opened 2019-06-19
Change Query to WaitForElement

PureWeen:
### Description of Change ###
- for test Issue1733 it should be using WaitForElement instead of Query so that it actually waits for it to become visible.
- renamed test so that the name is immediately useful


PureWeen:
The View UI Tests portion of this PR looks to be failing on API 28 so I'm moving that to a different PR.

I rebased and kicked off the UI Tests again just to be sure. If the UI tests contain no failures for EditorAutoResize then we can merge

PR #6601, opened 2019-06-20
Map MoveToLastRegionOnLayoutChange Property

### LINKED ISSUE #6222: Xamarin Forms Map resets back to last set region
I have one issue with the forms map. 

When control is panned and the device rotated, the map's region is set to the LastMoveToRegion which is not the current region of the map.

I observed this both on iOS and Android. The expectation is that the map center / zoom level is retained.

Any solution?

andreinitescu:
### Description of Change ###

As described in #6222 on iOS and Android, the Map renderer moves map center to last region when device is rotated. Looking at the renderers implementation, it actually moves to the last region on a layout change.

I added a new boolean property which when it's true, it doesn't move to the last region when layout changes.

### Issues Resolved ### 

- fixes #6222

### API Changes ###
 
Added:
 - bool Map.MoveToLastRegionOnLayoutChange { get; set; } //Bindable Property (default is **true**)

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android

None

Not applicable

### Testing Procedure ###

See Map Gallery Page. I added a button which toggles the MoveToLastRegionOnLayoutChange property:

IMPORTANT: **Note that you need the fix from #6603 to run it on Android**

image

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls/GalleryPages/MapGallery.xaml
+++ b/Xamarin.Forms.Controls/GalleryPages/MapGallery.xaml
@@ -1,10 +1,9 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
-<ContentPage
-    x:Class="Xamarin.Forms.Controls.MapGallery"
-    xmlns="http://xamarin.com/schemas/2014/forms"
-    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
-    xmlns:maps="clr-namespace:Xamarin.Forms.Maps;assembly=Xamarin.Forms.Maps"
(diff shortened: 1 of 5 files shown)


SnapADrag0n:
@andreinitescu @samhouts  Is this still a valid solution with latest Xamarin version?
I am testing this with Xamarin Forms V16.6.000.1062 and I could see map going map to last 'MoveToRegion' on resuming screen from a different screen. 
Android V8.1
Visual Studio 2019 V16.6.3
I didn't specifically test this on screen rotation.

PR #6603, opened 2019-06-20
[Android] Fix Map Gallery page

andreinitescu:
### Description of Change ###

The control gallery app crashes on Android when visiting Map gallery pages:
> java.lang.NoClassDefFoundError:failed resolution of :Lorg/apache/http/ProtocolVersion

This is because app targets Android 9. 
See 
https://developers.google.com/maps/documentation/android-sdk/config#specify_requirement_for_apache_http_legacy_library

I guess once Xamarin.GooglePlayServices.Maps will be updated to bind to com.google.android.gms:play-services-maps:17.0.0 and Xamarin.Forms.Android.Map updates to the latest Xamarin.GooglePlayServices.Maps. setting `uses-library` will no longer be necessary.

### Issues Resolved ### 

- fixes #

### API Changes ###

 
 None

### Platforms Affected ### 

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.ControlGallery.Android/Properties/AndroidManifest.xml
+++ b/Xamarin.Forms.ControlGallery.Android/Properties/AndroidManifest.xml
@@ -18,5 +18,6 @@
 	<uses-feature android:glEsVersion="0x00020000" android:required="true" />
 	<application android:label="AndroidControlGallery" android:supportsRtl="true">
 		<meta-data android:name="com.google.android.gms.version" android:value="@integer/google_play_services_version" />
+    <uses-library android:name="org.apache.http.legacy" android:required="false" />
 	</application>
 </manifest>


jcmanke:
Maybe this should be done with the assembly attribute so it is automatically added to the manifest to all apps consuming Xamarin.Forms.Maps? That's what I do in my custom Maps package.

`[assembly: UsesLibrary("org.apache.http.legacy", false)]`

andreinitescu:
I thought about that, not all apps might need the attribute

jcmanke:
Since it's only required for Android 9.0 and above, the attribute could be wrapped by a preprocessor directive checking the target API version.

andreinitescu:
Like I said, what if an app really doesn't want that included?
See https://developer.android.com/about/versions/pie/android-9.0-changes-28#apache-p

> As an alternative to using the runtime Apache library, apps can bundle their own version of the org.apache.http library in their APK. If you do this, you must repackage the library (with a utility like Jar Jar) to avoid class compatibility issues with the classes provided in the runtime.

PR #6623, opened 2019-06-21
Added Device.TryOpenUri

### LINKED ISSUE #6621: [Enhancement] Device.OpenUri to return result value of OS dependent call
## Summary
Currently Device.OpenUri is void. it will be nice if it returns the result value of underline call

## Intended Use Case
below is the code which tries to open FB app with the link to facebook page or group and if fails, then open link in browser . 

`  try
            {
                Device.OpenUri(fb);
            }
            catch
            {
                Device.OpenUri(u);
            }`

on android, trying to use unsupported protocol will fire an exception , but on IOS, openURL will return false if URL cant be opened and you can't distinguish was URL opened or not . in other native ios project  i use

`  if (!   [[UIApplication sharedApplication] openURL:[NSURL URLWithString:deeplink]])
     {
         [[UIApplication sharedApplication] openURL:[NSURL URLWithString:fbpage]];
     }`

it will be nice if in Xamarin.Forms.Platform.iOS/Forms.cs
public void OpenUriAction(Uri uri) would change to bool (or object) to get return value from 
UIApplication.SharedApplication.OpenUrl(url);

andreinitescu:
### Description of Change ###

The current `Device.OpenUri(Uri uri)` has no return value.
On iOS, the native call has a boolean return value which is not surfaced. This makes it impossible to know in iOS whether the call to `Device.OpenUri` call succeeded.

Following Microsoft's `TryXXXX` pattern, this PR adds an alternate method `bool Device.TryOpenUri(Uri)`. This method never throws an exception.

### Issues Resolved ### 

- fixes #6621

### API Changes ###
Added:
 - bool Device.TryOpenUri(Uri);

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

image

### Testing Procedure ###

See Device.TryOpenUri gallery page

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls/CoreGallery.cs
+++ b/Xamarin.Forms.Controls/CoreGallery.cs
@@ -409,6 +409,7 @@ namespace Xamarin.Forms.Controls
 				new GalleryPageFactory(() => new ViewCellGallery(), "ViewCell Gallery - Legacy"),
 				new GalleryPageFactory(() => new WebViewGallery(), "WebView Gallery - Legacy"),
 				new GalleryPageFactory(() => new BindableLayoutGalleryPage(), "BindableLayout Gallery - Legacy"),
+				new GalleryPageFactory(() => new DeviceTryOpenUriPage(), "Device TryOpenUri - Legacy"),
 			};
 
 		public CorePageView(Page rootPage, NavigationBehavior navigationBehavior = NavigationBehavior.PushAsync)
--- a/Xamarin.Forms.Platform.GTK/GtkPlatformServices.cs
+++ b/Xamarin.Forms.Platform.GTK/GtkPlatformServices.cs
@@ -90,6 +90,19 @@ namespace Xamarin.Forms.Platform.GTK
 			System.Diagnostics.Process.Start(uri.AbsoluteUri);
 		}
 
+		public bool TryOpenUriAction(Uri uri)
+		{
+			try
+			{
+				OpenUriAction(uri);
+				return true;
+			}
+			catch
+			{
+				return false;
+			}
+		}
+
 		public void StartTimer(TimeSpan interval, Func<bool> callback)
 		{
 			GLib.Timeout.Add((uint)interval.TotalMilliseconds, () =>
--- a/Xamarin.Forms.Controls/GalleryPages/DeviceTryOpenUri.cs
+++ b/Xamarin.Forms.Controls/GalleryPages/DeviceTryOpenUri.cs
@@ -1,0 +1,45 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Xamarin.Forms.Controls.GalleryPages
+{
+	class DeviceTryOpenUriPage : ContentPage
+	{
+		public DeviceTryOpenUriPage()
+		{
+			var instructions = new Label
+			{
+				Text = "This calls Device.TryOpenUri with the given uri." +
+				"\nDifferent than OpenUri, TryOpenUri never throws an exception if the call failed." +
+				"\nOn Android for example, calling OpenUri with an invalid uri throws an exception."
+			};
+
+			Title = "Issue 6221";
+
+			var entry = new Entry() { Text = "someProtocol:Something" };
+
+			var label = new Label();
+
+			var btn = new Button() { Text = "Call Device.TryOpenUri" };
+			btn.Clicked += (s, e) =>
+			{
+				label.Text = "Result: " + Device.TryOpenUri(new Uri(entry.Text, UriKind.Absolute)).ToString();
+			};
+
+			var layout = new StackLayout
+			{
+				Spacing = 10,
+				Children =  {
+					instructions,
+					entry,
+					label,
+					btn,
+				}
+			};
+
+			Padding = 15;
+			Content = layout;
+		}
+	}
+}
--- a/Xamarin.Forms.Core/Device.cs
+++ b/Xamarin.Forms.Core/Device.cs
@@ -229,7 +229,12 @@ namespace Xamarin.Forms
             PlatformServices.OpenUriAction(uri);
         }
 
-        public static void StartTimer(TimeSpan interval, Func<bool> callback)
+		public static bool TryOpenUri(Uri uri)
+		{
+			return PlatformServices.TryOpenUriAction(uri);
+		}
+
+		public static void StartTimer(TimeSpan interval, Func<bool> callback)
         {
             PlatformServices.StartTimer(interval, callback);
         }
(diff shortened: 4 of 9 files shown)


Xamarin.Forms.Controls/CoreGallery.cs
```diff
@@ -409,6 +409,7 @@ public override string ToString()
 				new GalleryPageFactory(() => new ViewCellGallery(), "ViewCell Gallery - Legacy"),
 				new GalleryPageFactory(() => new WebViewGallery(), "WebView Gallery - Legacy"),
 				new GalleryPageFactory(() => new BindableLayoutGalleryPage(), "BindableLayout Gallery - Legacy"),
+				new GalleryPageFactory(() => new DeviceTryOpenUriPage(), "Device TryOpenUri - Legacy"),
```
VladislavAntonyuk: I am not sure the new method is Legacy

andreinitescu: It's not about the method being a legacy.

Xamarin.Forms.Platform.GTK/GtkPlatformServices.cs
```diff
@@ -90,6 +90,32 @@ public void OpenUriAction(Uri uri)
...
+				OpenUriAction(uri);
+				return true;
+			}
+			catch
+			{
+				return false;
+			}
+		}
+
+		public bool TryOpenUriAction(Uri uri)
```
VladislavAntonyuk: looks like you paste the code twice

andreinitescu: Thanks! Fixed!

andreinitescu:
@StephaneDelcroix The existing API could still be useful if you need to know why it failed on specific platforms where the native does not throw an exception (i.e Android). From an unified perspective, it might not be ideal not always getting an exception when it failed, but that's how native (i.e. iOS) works, It could still make sense for logging purposes, I'd like to  know why when possible.

What do you think?

andreinitescu:
On a second thought, maybe have an overload of TryOpenUri which takes an Exception parameter:

```
bool TryOpenUri();
bool TryOpenUri(out Exception ex);
```

Or just have the 2nd one, today in C# it's easy to give an out parameter: 
```
Device.TryOpenUri(out Exception ex);
```

andreinitescu:
I don't like the idea of both `bool` and `Exception`, makes it confusing... How about this:

```
bool Device.TryOpenUri(out object error); 
```
_**error**: The error supplied by the native API. If the native API doesn't give an error, this parameter is null._

StephaneDelcroix:
> bool TryOpenUri(out Exception ex);

no. `bool TryFoo()` never outs an exception. the idea behind `TryFoo` is actually to prevent the exception being thrown when possible, not swallowing it, as throwing and catching exception breaks the execution flows.

> bool Device.OpenUri(out string error);

no to this too. that's the idea behind `TryFoo`, you get the results, not the reason.

I see what you're trying to do by unifying the 2 APIs, thanks for trying. it looks like it's not really possible.

The real question is wether or not there's value in the exception being thrown by `OpenUri()`. I'd say very little, as that exception is usually a native one, and handled on the common part of the project that knows nothing about it. but that's my opinion...

andreinitescu:
@StephaneDelcroix OK, what's your suggestion? So I know what to do. Thanks.

StephaneDelcroix:
@andreinitescu do nothing right now. I'm waiting for sam and ez opinion on deprecating or not

hartez:
At the very least, we'd need to either move `TryOpenUri()` to another interface or provide a default implementation, since `IPlatformServices` is public.

But since this is all just in service of providing feedback (whether the URI was successfully opened or not) on iOS/macOS, I think the right answer is to add exceptions to `OpenUriAction` on those platforms.

If `openURL(_:)` fails (returns `false`), we can just throw a custom exception with a message like "The URL could not be opened".). Then iOS/macOS folks will at least know that it failed.

Also, we should add a version check and use `open(_:options:completionHandler:)` for iOS 10+.

Redth:
FWIW we have these in Xamarin.Essentials already:

```csharp
Task<bool> Launcher.CanOpenAsync(Uri uri)
Task<bool> Launcher.CanOpenAsync(string uri)
Task Launcher.OpenAsync(Uri uri)
Task Launcher.OpenAsync(string uri)
```

StephaneDelcroix:
@hartez `IPlatformServices` is public but only to be consumed by Platforms. It is safe to assume that all plats are on the same version as the Core.

@Redth oh, great. should we deprecate ours ?

andreinitescu:
@Redth Thanks, I forgot about it.  If we don't add this to XF, it means Essentials is required. Which is OK I guess? (step by step Essentials becomes Required :) )

andreinitescu:
I am OK with closing this.

StephaneDelcroix:
@andreinitescu do a PR on Essentials for adding `Task<bool> TryOpenAsync()` while you're at it.

@samhouts I think we can go one step further and deprecate our implementation of OpenUri(). opinion ?

andreinitescu:
@StephaneDelcroix How would `Task<bool> TryOpenAsync(Uri uri)`  be different than the existing `
Task<bool> Launcher.CanOpenAsync(Uri uri)` ?

StephaneDelcroix:
@andreinitescu `CanOpen()` doesn't actually open, or so I guess by the naming. while TryOpen() does if it can, or return false

Redth:
Yeah the idea is that essentials handles all the non UI cross platform APIs you would want. Most users will want both essentials if you use forms but not necessarily the other way around. I’m in support of deprecating the API in forms.

Redth:
Added a PR in Essentials to add `TryOpenAsync`.

hartez:
> @hartez `IPlatformServices` is public but only to be consumed by Platforms. It is safe to assume that all plats are on the same version as the Core.

Nothing enforces the "only to be consumed by Platforms" rule. And it would break any 3rd-party platforms who aren't in our repo.

hartez:
> My suggestion is don't deprecate it otherwise many apps will start seeing Deprecate warnings in their build. I see no reason to deprecate `OpenUri()`

As it stands, `OpenUri()` is broken (because it doesn't behave consistently across platforms). If we're not going to fix it and instead push users to Essentials, then we should deprecate it so new users who find it are guided to use the (better) alternative. Deprecation warnings are a feature. 

My personal opinion is that we should make the iOS implementation work like the other platforms (throw an exception), and _also_ deprecate it in favor of Essentials, which has a better API.

StephaneDelcroix:
@hartez 

> Nothing enforces the "only to be consumed by Platforms" rule. 

`IPlatformServices` being 1/ in the `X.F.Internals` namespace, 2/ tagged with EditorBrowsableNever isn't part of the public API we commit not to break.

> And it would break any 3rd-party platforms who aren't in our repo.

I expect 3rd-platforms outside of our repo (if any) to depend on a single version of XF, but that might be the dreamer in me speaking.

Anyway, those points are moot, as we're closing this PR

StephaneDelcroix:
> I am OK with closing this.

done

PR #6628, opened 2019-06-21
[Shell] MenuItem font image fix

### LINKED ISSUE #6559: [Bug] FontImageSource not working with MenuItem in Shell
### Description
Image does not show up for MenuItem in Shell when using FontImageSource
### Steps to Reproduce

1. Run the code in the reproduction link

or

1. Create new app using Shell
2. Add an icon font (see James Montemagno's [article](https://montemagno.com/using-font-icons-in-xamarin-forms-goodbye-images-hello-fonts/) as a reference)
3. Add a MenuItem to the Shell in XAML (code behind not tested)
3. Add a FontImageSource using the icon font to the MenuItem.IconImageSource

### Expected Behavior
You should see the specified icon image with the MenuItem text

### Actual Behavior
Empty space where the icon should be

### Basic Information

- Version with issue: 4.0.0.497661, 4.1.0.496342-pre2
- IDE: Visual Studio 2019 (Windows) 16.0.4
- Platform Target Frameworks: 
  - iOS:  8.0
  - Android: 9.0
  - UWP:  N?A
- Android Support Library Version: 28.0.0.1, also seen with 27.0.2.1
- Nuget Packages:  
  - Xamarin.Forms , 4.1.0.496342-pre2
  - Xamarin.Essentials, 1.1.0
 - Affected Devices: Android & iOS

### Reproduction Link
[https://github.com/anotherlab/FontImageSourceMissingInMenuItem](https://github.com/anotherlab/FontImageSourceMissingInMenuItem)

felipemomm:
### Description of Change ###

- MenuItem was using FileImageSource and it doesn't work with font icons. I've changed to ImageSource, the same way it's used on ShellItem. 
- There was a missing binding on ShellItem, so I've added it as well.

### Issues Resolved ### 
- fixes #6559

### API Changes ###

 None

### Platforms Affected ### 
- Core/XAML (all platforms)

None

Before:
Screen Shot 2019-06-21 at 11 53 39

After:
Screen Shot 2019-06-21 at 11 51 52

### Testing Procedure ###
1. Run XamControl app
2. Tap "SwapRoot - Store Shell"
3. Check "Redeem" MenuItem's icon on Hamburger menu

### PR Checklist ###

- [ ] Has automated tests (it's visual changes)

### DIFF
--- a/Xamarin.Forms.Core/MenuItem.cs
+++ b/Xamarin.Forms.Core/MenuItem.cs
@@ -46,9 +46,9 @@ namespace Xamarin.Forms
 
 		[Obsolete("Icon is obsolete as of 4.0.0. Please use IconImageSource instead.")]
 		[EditorBrowsable(EditorBrowsableState.Never)]
-		public FileImageSource Icon
+		public ImageSource Icon
 		{
-			get => GetValue(IconProperty) as FileImageSource ?? default(FileImageSource);
+			get => (ImageSource)GetValue(IconProperty);
 			set => SetValue(IconProperty, value);
 		}
 
--- a/Xamarin.Forms.Core/MenuItem.cs
+++ b/Xamarin.Forms.Core/MenuItem.cs
@@ -46,9 +46,9 @@ namespace Xamarin.Forms
 
 		[Obsolete("Icon is obsolete as of 4.0.0. Please use IconImageSource instead.")]
 		[EditorBrowsable(EditorBrowsableState.Never)]
-		public ImageSource Icon
+		public FileImageSource Icon
 		{
-			get => (ImageSource)GetValue(IconProperty);
+			get => GetValue(IconProperty) as FileImageSource ?? default(FileImageSource);
 			set => SetValue(IconProperty, value);
 		}
 
--- a/Xamarin.Forms.Core/Shell/MenuShellItem.cs
+++ b/Xamarin.Forms.Core/Shell/MenuShellItem.cs
@@ -8,9 +8,9 @@ namespace Xamarin.Forms
 		{
 			MenuItem = menuItem;
 
-			SetBinding(TitleProperty, new Binding("Text", BindingMode.OneWay, source: menuItem));
-			SetBinding(IconProperty, new Binding("IconImageSource", BindingMode.OneWay, source: menuItem));
-			SetBinding(FlyoutIconProperty, new Binding("IconImageSource", BindingMode.OneWay, source: menuItem));
+			SetBinding(TitleProperty, new Binding(nameof(MenuItem.Text), BindingMode.OneWay, source: menuItem));
+			SetBinding(IconProperty, new Binding(nameof(MenuItem.IconImageSource), BindingMode.OneWay, source: menuItem));
+			SetBinding(FlyoutIconProperty, new Binding(nameof(MenuItem.IconImageSource), BindingMode.OneWay, source: menuItem));
 
 			Shell.SetMenuItemTemplate(this, Shell.GetMenuItemTemplate(MenuItem));
 			MenuItem.PropertyChanged += OnMenuItemPropertyChanged;
--- a/Xamarin.Forms.Core/Shell/MenuShellItem.cs
+++ b/Xamarin.Forms.Core/Shell/MenuShellItem.cs
@@ -9,8 +9,8 @@ namespace Xamarin.Forms
 			MenuItem = menuItem;
 
 			SetBinding(TitleProperty, new Binding("Text", BindingMode.OneWay, source: menuItem));
-			SetBinding(IconProperty, new Binding("Icon", BindingMode.OneWay, source: menuItem));
-			SetBinding(FlyoutIconProperty, new Binding("Icon", BindingMode.OneWay, source: menuItem));
+			SetBinding(IconProperty, new Binding("IconImageSource", BindingMode.OneWay, source: menuItem));
+			SetBinding(FlyoutIconProperty, new Binding("IconImageSource", BindingMode.OneWay, source: menuItem));
 
 			Shell.SetMenuItemTemplate(this, Shell.GetMenuItemTemplate(MenuItem));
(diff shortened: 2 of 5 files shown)


jcmanke:
I'm no expert on Shell but this seems like the wrong fix to me. Changing the type of `MenuItem.Icon` from `FileImageSource` to `ImageSource` would be a breaking change, that's why it was deprecated and the `IconImageSource` property to was added instead.

I think this is the real problem: https://github.com/xamarin/Xamarin.Forms/blob/62dfadabf2bb1f3211deed33b17b0a462c1ba2d9/Xamarin.Forms.Core/Shell/MenuShellItem.cs#L12

```
SetBinding(IconProperty, new Binding("Icon", BindingMode.OneWay, source: menuItem));
SetBinding(FlyoutIconProperty, new Binding("Icon", BindingMode.OneWay, source: menuItem));
```

Change these two lines to bind to "IconImageSource" rather than "Icon" and everything should be fine.

felipemomm:
I'm not an expert either, I was just studying Xamarin.Shell and decided to fix some bugs... 

As for the PR, not sure if changing from `FileImageSource` to `ImageSource` would be a breaking change, at least it worked fine when setting a file to Icon property.

However I don't know exactly the history behind Icon and IconImageSource. Was it only for naming purposes? Can anyone explain it?

BTW, your proposed change fixes the problem too.

Xamarin.Forms.Core/MenuItem.cs
```diff
@@ -46,9 +46,9 @@ public object CommandParameter
 
 		[Obsolete("Icon is obsolete as of 4.0.0. Please use IconImageSource instead.")]
 		[EditorBrowsable(EditorBrowsableState.Never)]
-		public FileImageSource Icon
+		public ImageSource Icon
```
rmarinho: But don t we have a `IconImageSource`  ?

felipemomm: It doesn't work for font icons if you use `IconImageSource`, because it uses `Icon` property and it's a FileImageSource.

Check the reproduction link:
https://github.com/anotherlab/FontImageSourceMissingInMenuItem

This XAML will not draw an icon

```
<MenuItem Text="Icon Missing">
    <MenuItem.IconImageSource>
        <FontImageSource
            FontFamily="{DynamicResource MaterialFontFamily}"
            Glyph="{x:Static icons:MaterialDesignIconFont.InformationOutline}"
            Color="Black"/>
    </MenuItem.IconImageSource>
</MenuItem>
```

TechWatching:
> I'm not an expert either, I was just studying Xamarin.Shell and decided to fix some bugs...
> 
> As for the PR, not sure if changing from `FileImageSource` to `ImageSource` would be a breaking change, at least it worked fine when setting a file to Icon property.
> 
> However I don't know exactly the history behind Icon and IconImageSource. Was it only for naming purposes? Can anyone explain it?
> 
> BTW, your proposed change fixes the problem too.

I don't know the whole history neither but from what I have understood FileImageSource was used before for properties like Icon of some components. In order to be able to use the FontImageSource, they introduce in 4.0 new properties of type ImageSource in all these components and deprecated the old properties of type FileImageSource. 

That's why as @jcmanke said, I think the correction should not be to change the type of Icon to ImageSource but instead to make the IconImageSource works. Maybe you can update your PR with what @jcmanke proposed if it works as you said.

https://devblogs.microsoft.com/xamarin/introducing-xamarin-forms-4-0-the-era-of-shell/

felipemomm:
Sounds good. I've reverted the changes and applied the fix that @jcmanke has proposed.

Xamarin.Forms.Core/Shell/MenuShellItem.cs
```diff
@@ -9,8 +9,8 @@ internal MenuShellItem(MenuItem menuItem)
 			MenuItem = menuItem;
 
 			SetBinding(TitleProperty, new Binding("Text", BindingMode.OneWay, source: menuItem));
-			SetBinding(IconProperty, new Binding("Icon", BindingMode.OneWay, source: menuItem));
-			SetBinding(FlyoutIconProperty, new Binding("Icon", BindingMode.OneWay, source: menuItem));
+			SetBinding(IconProperty, new Binding("IconImageSource", BindingMode.OneWay, source: menuItem));
```
jcmanke: If I can make a further suggestion, maybe these bindings should use `MenuItem.IconImageSourceProperty.PropertyName` instead of just magic strings. If that had been done in the first place this issue never would have happened.

Xamarin.Forms.Core/Shell/MenuShellItem.cs
```diff
@@ -8,9 +8,9 @@ internal MenuShellItem(MenuItem menuItem)
 		{
 			MenuItem = menuItem;
 
-			SetBinding(TitleProperty, new Binding("Text", BindingMode.OneWay, source: menuItem));
-			SetBinding(IconProperty, new Binding("Icon", BindingMode.OneWay, source: menuItem));
-			SetBinding(FlyoutIconProperty, new Binding("Icon", BindingMode.OneWay, source: menuItem));
+			SetBinding(TitleProperty, new Binding(nameof(MenuItem.Text), BindingMode.OneWay, source: menuItem));
+			SetBinding(IconProperty, new Binding(nameof(MenuItem.IconImageSource), BindingMode.OneWay, source: menuItem));
+			SetBinding(FlyoutIconProperty, new Binding(nameof(MenuItem.IconImageSource), BindingMode.OneWay, source: menuItem));
```
felipemomm: Agreed, it's safer.

JunkyXL86:
> **In order to be able** to use the FontImageSource, they introduce in 4.0 new properties of type ImageSource in all these components and deprecated the old properties of type FileImageSource.

I'm investigating a lot of time finding out why the Icon property simply hasn't changed to ImageSource. Since it's the base class of FileImageSource, I don't see a breaking change here. What was the reason that an additional property has been introduced?

rmarinho:
@JunkyXL86 yap.

PR #6636, opened 2019-06-23
[iOS] Clean up Platform on PlatformRenderer disposal

### LINKED ISSUE #4671: Xamarin Forms Native embedding Memory issue
### Description
I am trying to use Xamarin Forms pages in a Xamarin IOS Native app. To do that I am using the CreateViewController. The issue I am facing is that every time I push a new view controller, memory are allocated that are not released when I click on the back button. This results in that the app will crash. 
### Steps to Reproduce

Explanation about attached image

1. I am navigating back and fort to a Xamarin forms page (contains image and is created with xaml). I am doing this multiple times, and that is displayed with the stepping increase in memory consumption
2. I am waiting to se if the memory is released, but that do not happen. I am then adding native uiviewcontrollers (contains one image) to the same navigation controller the same way as Xamarin forms.
3. The memory are released for the native view controllers.

Link to testcode
 github.com/toupilsner/XamarinFormsEmbedded/blob/master/README.md

Xamairn Forms: 3.4.0

hartez:
### Description of Change ###

Native Forms gives the user a `PlatformRenderer` (as a `UIViewController`) to use for embedding a Forms `Page`. If the user wants to free the resources used by the embedded `Page`, there is currently no way to do so. Disposing of the controller itself is not sufficient, and the actual Platform object which needs to be disposed is inaccessible.

These changes arrange the cleanup so that calling `Dispose` on the returned `UIViewController` object will clean up all the necessary resources.

### Issues Resolved ### 

- fixes #4671

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.iOS/ModalWrapper.cs
+++ b/Xamarin.Forms.Platform.iOS/ModalWrapper.cs
@@ -1,0 +1,116 @@
+﻿using System;
+using System.Linq;
+using Xamarin.Forms.PlatformConfiguration.iOSSpecific;
+using UIKit;
+
+namespace Xamarin.Forms.Platform.iOS
+{
+	internal class ModalWrapper : UIViewController
+	{
+		IVisualElementRenderer _modal;
+
+		internal ModalWrapper(IVisualElementRenderer modal)
+		{
+			_modal = modal;
+
+			var elementConfiguration = modal.Element as IElementConfiguration<Page>;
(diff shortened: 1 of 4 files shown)


hartez:
iOS 10 test failure is unrelated.

PR #6650, opened 2019-06-25
DefaultDataTemplateSelector (#6649)

### LINKED ISSUE #6649: [Enhancement] BuildProperty for BindableLayout
## Summary
Expose a `Func<object, DataTemplate> Build` property in `BindableLayout` - So one does not need to use `DataTemplateSelectors`.

## API Changes

In [`BindableLayout`](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/BindableLayout.cs) add: 

```
public static readonly BindableProperty BuildProperty =
    BindableProperty.CreateAttached("Build", typeof(Func<object, DataTemplate>), typeof(Layout<View>), default(Func<object, DataTemplate>),
        propertyChanged: (b, o, n) => { GetBindableLayoutController(b).Build = (Func<object, DataTemplate>)n; });
```
Get and Set methods

```
public static void SetBuild(BindableObject b, Func<object, DataTemplate> value)
{
    b.SetValue(BuildProperty, value); 
}

public static Func<object, DataTemplate> GetBuild(BindableObject b)
{
    return (Func<object, DataTemplate>)b.GetValue(BuildProperty);
}
```
In OnControllerChanged add `newC.Build = GetBuild(b);`

---

Inside `BindableLayoutController`, add  corresponding `Func<object, DataTemplate> Build` property.

And in `CreateItemView` the following:
```
View CreateItemView(object item, Layout<View> layout)
{ 
    return CreateItemView(item, _itemTemplate ?? _itemTemplateSelector?.SelectTemplate(item, layout) ?? build?.invoke(item)); // <---
}

```

## Intended Use Case

In the code behind of my page I can have the following:

```
BindableLayout.SetBuild(stackLayout,
    (item) =>
    {
        var model = item as string;
        switch(model)
        {
            case "First":
                return special;
            default:
                return standard;
        }
    });
              
```
Where `special` and `standard` are DataTemplates defined in xaml of the same page. (Instead of needing to have a class extending [DataTemplateSelector)](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/app-fundamentals/templates/data-templates/selector)

andreinitescu:
### Description of Change ###

Currently, in order to get an instance of a `DataTemplateSelector`, you need to add your own class and implement the `DataTemplateSelector` abstract class. It should be easier, because after all, the `DataTemplateSelector` works as a mere function anyway. 

This PR makes using `DataTemplateSelector` easier by providing a default implementation for the `DataTemplateSelector`  and taking a `Func<object, BindableObject, DataTemplate>`.

### Issues Resolved ### 

- fixes #6649

### API Changes ###

Added:

	public class DefaultDataTemplateSelector : DataTemplateSelector
	{
		public DefaultDataTemplateSelector(Func<object, BindableObject, DataTemplate> func);
	}

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/DefaultDataTemplateSelector.cs
+++ b/Xamarin.Forms.Core/DefaultDataTemplateSelector.cs
@@ -1,0 +1,13 @@
+﻿using System;
+
+namespace Xamarin.Forms
+{
+	public class DefaultDataTemplateSelector : DataTemplateSelector
+	{
+		readonly Func<object, BindableObject, DataTemplate> _func;
+
+		public DefaultDataTemplateSelector(Func<object, BindableObject, DataTemplate> func) => _func = func;
+
+		protected override DataTemplate OnSelectTemplate(object item, BindableObject container) => _func(item, container);
+	}
+}


Lelelo1:
I copied the class into my code behind of a page. How can I access it in xaml? i added `xmlns:local="clr-namespace:MyProject.Views"` - but it won't show up there in xaml.

andreinitescu:
@Lelelo1 you're missing the point, if you need it in XAML, you can just use existing DataTemplateSelector

Lelelo1:
@andreinitescu Yes, but that mean I would have to subclass  `DataTemplateSelector` again.

StephaneDelcroix:
I understand the issue you're trying to solve, but I would rather not include this default DTS at this time as:
- providing a "default" implementation might force people to think that's the only way to go
- the default reflects how you use DTS, but that "default" is probably not everyone's default, and we don't want, at this time and IMHO, include the 10 most common "default" DTS in Forms

also, this solution doesn't look like it fixes the original request or 100% please the OP (I'm still unconvinced about that enhancement request too)

Would you mind if I close this ?

andreinitescu:
Yeah, I should not had wasted time on this 😒

Issue #6653, opened 2019-06-25
[UWP] Image Loading issues with URI images


PureWeen:
### Description
- Images aren't loading when using a URI
- Images and Buttons don't scale down when the image is nulled out

On a side note I think there's a bug with Appium where it can't read an AutomationID from a DataTemplate. Which is one of the reasons that the Issue4597 test is failing on UWP and I have no idea how to currently make it automated. Even if you retrieve the descendants of the ListView it only returns a single empty element so I think the Appium concept of the UI Test isn't refreshing correctly and a bug will need to be logged with them

### Steps to Reproduce

1. Run Issue4597 and go through the variations there to test UWP
2. You can also uncomment out the UI tests for windows and run them to see where they fail

### Expected Behavior
All the UI tests for 4597 should pass 

### Actual Behavior
They all currently fail

### Basic Information

- Version with issue: 4.0+
- IDE:
- Platform Target Frameworks: 
  - UWP

PR #6654, opened 2019-06-25
Split up Issue4597 into separate testable chunks

PureWeen:
### Description of Change ###
This splits out all the individual parts being tested by Issue4597 so it's more reliable and failures are more indicative of what's actually failing. Also it's way less confusing now to read. 

UWP tests are currently commented out and Issues is logged here
https://github.com/xamarin/Xamarin.Forms/issues/6653


rmarinho:
seems we need this on 4.0 to help with green run

PR #6661, opened 2019-06-26
Fix Page implicit and inherited style with ControlTemplate #6657

### LINKED ISSUE #6657: [Bug] Setting ControlTemplate on ContentPage from an implicit style results in blank content
### Description

When setting a control template on ContentPage using an implicitly applied style the ContentPresenter appears blank.

The attached reproduction sample exhibits a completely blank ContentPresenter, however in my larger solution the result I get is slightly more subtle - The content presenter seems to work but Xaml bindings elements stop working. Originally noticed the issue due to similar symptoms described in #2146.
[ControlTemplateBinding.zip](https://github.com/xamarin/Xamarin.Forms/files/3328300/ControlTemplateBinding.zip)

### Steps to Reproduce

1. Create a style that applies to your ContentPage type (without an x:Key declaration)
2. Apply a control template using the style

### Expected Behavior

The content of the page is visible in place of the <ContentPresenter> element

### Actual Behavior

Blank space

### Basic Information

- Version with issue: 3.6+ (including 4.0.0.497661)
- IDE: VS2019
- Platform Target Frameworks: 
  - iOS:  
  - Android: 9.0 
  - UWP:   
- Nuget Packages:
- Affected Devices: Samsung Galaxy Tab Active 2 (Android 8.1, Sumsung Experience 9.5)

### Reproduction Link

[ControlTemplateBinding.zip](https://github.com/xamarin/Xamarin.Forms/files/3328305/ControlTemplateBinding.zip)

andreinitescu:
### Description of Change ###

A derived Page with an implicit style with ApplyToDerivedTypes true, doesn't get content set correctly. 
This is because Page ends up with a collection already set in c-tor by the parent.

I added unit tests for both TemplatedPage and TemplatedView. TemplatedView seems to work OK.

### Issues Resolved ### 

- fixes #6657 

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 


andreinitescu:
I really hoped this would get into 4.1. It's a quick fix for a common scenario.

StephaneDelcroix:
@samhouts I'd retarget this to master. it looks simple, but might have some unknown side effects on existing apps.

andreinitescu:
@StephaneDelcroix If there's any side effect is because of wrong assumptions made by the app. 
This is a hidden issue which it's been lurking for some time. Please don't prolong releasing this fix unless there's a very good reason.

The Layout has the right behavior but apparently it was forgotten to be added to the Page too, please see 
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/Layout.cs#L73

PR #6674, opened 2019-06-26
 [Android] Unhook WebView.EvaluateJavascriptAsync on Dispose

### LINKED ISSUE #6286: [Bug] ObjectDisposedException in Android WebView.EvaluateJavascriptAsync
### Description
WebView control exists on a ContentPage. EvaluateJavascriptAsync is called in Navigated event.
works fine first time. But Navigating away and back to the ContentPage triggers  *multiple* ObjectDisposedExceptions (for the Android webview) in EvaluateJavascriptAsync, even though the Navigated event happens only once.  The number of triggered exceptions depends on the number of navigations to the ContentPage.

### Steps to Reproduce

Use Custom renderer, navigate to and from page with web view. Call EvaluateJavascriptAsync in Navigated event.

### Possible reason and fix:

Looking at the source I can see that for Android, WebViewRenderer.Dispose() does not clear the ElementController.EvaluateJavaScriptRequested event handler. I can see that that is an omission because the OnElementChanged handler does indeed clear the EvaluateJavaScriptRequested  event for the old element.

*Fix*: Using a custom renderer (for Android) , clear the ElementController.EvaluateJavaScriptRequested event by overriding the custom renderer's Dispose() method. It is necessary to use reflection to do that.

### Basic Information

- Version with issue: 3.6
- Last known good version:
- IDE: VS 2019
- Platform Target Frameworks: 
   - Android: 28

samhouts:
### Description of Change ###

Unhook WebView.EvaluateJavascriptAsync on Dispose to prevent any crashes.

### Issues Resolved ### 

- fixes #6286 (hopefully--I was unable to reproduce it)

### API Changes ###

 
 None

### Platforms Affected ### 

- Android

None

Not applicable

### Testing Procedure ###

Added test case 6286, but it does not exhibit the crash. Will need confirmation from @kwaclaw that this resolves the issue.

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6286.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6286.cs
@@ -1,0 +1,64 @@
+﻿using System;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.Forms.Core.UITests;
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if UITEST
+	[Category(UITestCategories.WebView)]
+#endif
+	[Preserve(AllMembers = true)]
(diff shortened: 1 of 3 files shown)


samhouts:
failing tests not related

PR #6677, opened 2019-06-27
Validate style target at runtime

### LINKED ISSUE #6268: [Bug] InvalidCastException encountered when Entry control is bound to a style for a Label
### Description
Binding a entry control to a style for a label that has properties that are not applicable to an Entry control causes an InvalidCastException. This should be caught at compilation time, or a minimum of a descriptive exception saying it is not correct to do this.

### Steps to Reproduce
1. Model a style for a label with a label only property:
`<Style x:Key="LeftLabel" TargetType="Label">
            <Setter Property="VerticalTextAlignment" Value="Center" />
        </Style>`
Bind an entry control to that style (or really any other type of control would probably do as well):
`<Entry Style="{StaticResource LeftLabel}"/>`
2. Run the app

### Expected Behavior
Compilation error should occur in this case.
Good runtime error should be raised in the case that this binding is performed through code instead of modeling.

### Actual Behavior
InvalidCastException:
 	0x0 in Xamarin.Forms.Label.OnVerticalTextAlignmentPropertyChanged at D:\a\1\s\Xamarin.Forms.Core\Label.cs:324,4	C#
 	0x12E in Xamarin.Forms.BindableObject.SetValueActual at D:\a\1\s\Xamarin.Forms.Core\BindableObject.cs:624,5	C#
 	0x182 in Xamarin.Forms.BindableObject.SetValueCore at D:\a\1\s\Xamarin.Forms.Core\BindableObject.cs:422,5	C#

on this line of code from Label.cs
    var label = (Label)bindable;

### Basic Information

- IDE: VS 2019
- Platform Target Frameworks: 
  - iOS:  12.2
  - Android: 8.1

- Nuget Packages:
- Affected Devices:

### LINKED ISSUE #6535: Crash XF when Font Style Applied to StackLayout/Grid[Bug]
I **mistakenly** applied a Style to a StackLayout which has font information. XF 3.x and 4.x crash in the XF Xamarin.Forms.FontElement.FontSizeDefaultValueCreator.cs line since the bindable (StackLayout) is null since it's not a IFontElement

		static object FontSizeDefaultValueCreator(BindableObject bindable)
		{
			return ((IFontElement)bindable).FontSizeDefaultValueCreator();
		}
 
### Steps to Reproduce

Create a Style in your App.XAML targeting a **Label**
```
			<Style
				x:Key="LabelNoteStyle" TargetType="Label">
				       <Setter Property="FontSize" Value="13"></Setter>
				       <Setter Property="TextColor" Value="{StaticResource BaseLightTextColor}"></Setter>
			</Style>
```
Apply this to a StackLayout
`       <StackLayout Grid.Row="0" Orientation="Horizontal"  Style="{StaticResource LabelNoteStyle}">`

### Expected Behavior
Font info Is ignored
### Actual Behavior
Crash in Anadroid and iOS in XF 3.6 and 4.x. It was hard to find as the error is just a null exception at the Main App level. I had to debug XF to find the actual bug.

andreinitescu:
### Description of Change ###

Checks if Style can be set on an element based on the type compatibility between Style TargetType and element's type. In case of type incompatibility, it throws an exception with type names:

```
Style TargetType Xamarin.Forms.Button is not compatible with element target type Xamarin.Forms.Label
```

Includes unit tests.

**This doesn't cover XAMLC.**

### Issues Resolved ### 

- fixes #6268
- fixes #6535

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/StyleTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/StyleTests.cs
@@ -759,5 +759,29 @@ namespace Xamarin.Forms.Core.UnitTests
 			Assert.That(label0.TextColor, Is.EqualTo(Color.Pink));
 			Assert.That(label1.TextColor, Is.EqualTo(Color.Lavender));
 		}
+
+		[Test]
+		public void MismatchTargetTypeThrowsError1()
+		{
+			var s = new Style(typeof(Button));
+			var t = new View();
+			Assert.Throws<ArgumentException>(() => t.Style = s);
+		}
+
+		[Test]
+		public void MismatchTargetTypeThrowsError2()
+		{
+			var s = new Style(typeof(Button));
+			var t = new Label();
+			Assert.Throws<ArgumentException>(() => t.Style = s);
+		}
+
+		[Test]
+		public void MatchTargetTypeDoesntThrowError()
+		{
+			var s = new Style(typeof(View));
+			var t = new Button();
+			Assert.DoesNotThrow(() => t.Style = s);
+		}
 	}
 }
--- a/Xamarin.Forms.Core/MergedStyle.cs
+++ b/Xamarin.Forms.Core/MergedStyle.cs
@@ -34,7 +34,14 @@ namespace Xamarin.Forms
 		public IStyle Style
 		{
 			get { return _style; }
-			set { SetStyle(ImplicitStyle, ClassStyles, value); }
+			set
+			{
+				if (_style == value)
+					return;
+				if (value != null && !value.TargetType.IsAssignableFrom(TargetType))
+					throw new ArgumentException($"Style TargetType {value.TargetType.FullName} is not compatible with element target type {TargetType}");
+				SetStyle(ImplicitStyle, ClassStyles, value);
+			}
 		}
 
 		public IList<string> StyleClass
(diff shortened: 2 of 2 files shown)


StephaneDelcroix:
> This doesn't cover XAMLC

well, it actually does. but we probably could do better and try to detect that at compilation time.

StephaneDelcroix:
@samhouts I'd merge this, and add a new task for me to add compile-time validation for this

andreinitescu:
@samhouts could you please be more specific, the ones you mentioned I think are already covered

> One where the Style target is the derived type 

		[Test]
		public void MismatchTargetTypeThrowsError1()
		{
			var s = new Style(typeof(Button));
			var t = new View();
			Assert.Throws<ArgumentException>(() => t.Style = s);
		}

>  and one where the control is a derived type

		[Test]
		public void MatchTargetTypeDoesntThrowError()
		{
			var s = new Style(typeof(View));
			var t = new Button();
			Assert.DoesNotThrow(() => t.Style = s);
		}

samhouts:
@andreinitescu Well, look at that! That's what I get for not reading closely. Thanks!!

andreinitescu:
@samhouts done

samhouts:
I'm going to move this to master (4.2) because 4.1 is now in regression-only mode. Thanks!!

PR #5044, opened 2019-01-23
[Feature] CarouselView

rmarinho:
### Description of Change ###

Implementing the CarouselView spec based on CollectionView work.

### Issues Resolved ### 

- fixes #4996

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android

### PR Checklist ###

- [x] Has automated tests 
- [x] Rebased on top of the target branch at time of PR
- [ ] Changes adhere to coding standard

PR #6690, opened 2019-06-27
[iOS] ItemsViewLayout needs strategy on the ctor

rmarinho:
### Description of Change ###

The iOS CollectionView layouts should receive the ItemSizeStrategy on the constructor, this is because some methods can be called by the UICollectionView platform and the ItemSizingStrategy was not set yet.

### Issues Resolved ### 

- helps with #5044 

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.iOS/CollectionView/GridViewLayout.cs
+++ b/Xamarin.Forms.Platform.iOS/CollectionView/GridViewLayout.cs
@@ -10,7 +10,7 @@ namespace Xamarin.Forms.Platform.iOS
 	{
 		readonly GridItemsLayout _itemsLayout;
 
-		public GridViewLayout(GridItemsLayout itemsLayout) : base(itemsLayout)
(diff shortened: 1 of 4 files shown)


rmarinho:
Failing test not related

hartez:
>  this is because some methods can be called by the UICollectionView platform and the ItemSizingStrategy was not set yet.

Which methods? What specific error/problem does this prevent?

rmarinho:
@hartez  for example this line https://github.com/xamarin/Xamarin.Forms/blob/3d53dd70bd306d8a4df1593b2bd340d8484b7150/Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewLayout.cs#L157 

this is called as soon as the Layout at that time ItemSizingStrategy is not set yet for other value.

PR #6695, opened 2019-06-27
[Shell] Apply Query string parameters even if they aren't present and correctly apply them to nested pages

### LINKED ISSUE #6543: [Bug] Shell QueryProperty Parameter Ignored when Navigating to existing subpage
### Description
When navigating to an already existing page with a new Uri the parameter in the uri is ignored. 
Tested on Android, I am unable to test on IOS currently. This issue does not occur when the main tabs are navigated to with parameters, only pages below those.

### Steps to Reproduce

1. Add a QueryProperty to a viewmodel
   `    [QueryProperty("TypeID", "id")]
    public class MonkeysPageViewModel : ViewModelBase
	{
        private string _typeId;
        public string TypeID
        {
            get
            {
                return _typeId;
            }
            set
            {
                _typeId = value;
                OnPropertyChanged();
                int mID;
                if (Int32.TryParse(TypeID, out mID))
                {
                    SelectedMonkey = Monkeys[mID];
                }
            }
        }`
2. Register the route as a "Relative Route" of another Shell tab
`Routing.RegisterRoute("monkeys/details", typeof(MonkeysPage2));`
3. Navigate with a Parameter
`await Shell.Current.GoToAsync("//monkeys/details?id=1");`
4. Using the bottom tab navigation open a different tab (Page1)
5. Navigate again with a different parameter
`await Shell.Current.GoToAsync("//monkeys/details?id=2");`

### Expected Behavior
New parameter is passed to ViewModel

### Actual Behavior
Parameter is ignored

### Basic Information

- Version with issue: 4.0.0.497661
- IDE: VS 2019 for Windows v16.1.3
- Platform Target Frameworks: 
  - Android: 28
- Affected Devices: Only tested on Android emulator and device

### Reproduction Link

[MonkeyAppPicker.zip](https://github.com/xamarin/Xamarin.Forms/files/3293415/MonkeyAppPicker.zip)

PureWeen:
### Description of Change ###
- The code that applies query string parameters wouldn't execute if the user didn't supply any query string parameters. But it should still set whatever parameters the user has to null if no Query String Parameters are specified. 
- Nested content pages are parented to ShellSection not ShellContent so the code that applied query parameters to them wasn't correct. The only reason it worked during the first navigation was a fluke because the Parent hadn't been set yet on the content page

### Issues Resolved ### 
- fixes #6543 

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS
- Android

### Testing Procedure ###
- UI tests cover the scenarios to test

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Shell/Shell.cs
+++ b/Xamarin.Forms.Core/Shell/Shell.cs
@@ -449,9 +449,6 @@ namespace Xamarin.Forms
 
 		internal static void ApplyQueryAttributes(Element element, IDictionary<string, string> query, bool isLastItem)
 		{
-			if (query.Count == 0)
-				return;
-
 			string prefix = "";
 			if (!isLastItem)
 			{
@@ -487,7 +484,7 @@ namespace Xamarin.Forms
 				filteredQuery.Add(key, q.Value);
 			}
 
-			if (baseShellItem != null)
+			if (baseShellItem != null && baseShellItem is ShellContent)
 				baseShellItem.ApplyQueryAttributes(filteredQuery);
 			else if (isLastItem)
 				element.SetValue(ShellContent.QueryAttributesProperty, query);
--- a/Xamarin.Forms.Core/Shell/Shell.cs
+++ b/Xamarin.Forms.Core/Shell/Shell.cs
@@ -484,7 +484,7 @@ namespace Xamarin.Forms
 				filteredQuery.Add(key, q.Value);
 			}
 
-			if (baseShellItem != null && baseShellItem is ShellContent)
+			if (baseShellItem is ShellContent)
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Core/Shell/Shell.cs
```diff
@@ -487,7 +484,7 @@ internal static void ApplyQueryAttributes(Element element, IDictionary<string, s
 				filteredQuery.Add(key, q.Value);
 			}
 
-			if (baseShellItem != null)
+			if (baseShellItem != null && baseShellItem is ShellContent)
```
StephaneDelcroix: isn't this redundant ?
```suggestion
			if (baseShellItem is ShellContent)
```

PR #6329, opened 2019-05-28
[Android] Fixes layout in AppCompat ButtonRenderer

paymicro:
### Description of Change ###

After changing the font and any color in the legacy Button renderer, the button had a layout corruption.

### Issues Resolved ### 

- fixes #6282 

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

### Before/After Screenshots ### 

Before

Screenshot_2

After
Screenshot_1

### Testing Procedure ###

- add flag `UseLegacyRenderers`
- run UItest 6282
or
- open Dynamic Control Gallery
- select Button control
- change Font
- change TextColor

### PR Checklist ###

- [ ] Has automated tests 
- [x] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard

PR #6704, opened 2019-06-28
[Android] RequestLayout of button if measure has changed otherwise call force layout to ensure the content will layout correctly

PureWeen:
### Description of Change ###

https://github.com/xamarin/Xamarin.Forms/pull/6329 added a ForceLayout call when calling GetDesiredSize.  This fixed the text layout issue but it now causes the button to not relayout if the WxH  has changed. So in a case where the button was going from not visible to visible and the height and width would go from zero to not zero then the force layout would just layout the control without redrawing it with the new width and height.

This PR calls ForceLayout only if the height and width haven't changed. If the height and width have changed then it calls RequestLayout if there's not a pending layout request. 

### Platforms Affected ### 
- Android

### Testing Procedure ###
- ui test 53179 on various APIs with FR's enabled
- ui test Issue6282 on various APIs
- whatever other type of button layout you feel like testing

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/ButtonLayoutManager.cs
+++ b/Xamarin.Forms.Platform.Android/ButtonLayoutManager.cs
@@ -82,8 +82,21 @@ namespace Xamarin.Forms.Platform.Android
 
 		internal SizeRequest GetDesiredSize(int widthConstraint, int heightConstraint)
 		{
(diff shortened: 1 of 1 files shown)


rmarinho:
Just tested looks good, uitests are happy too

PR #6714, opened 2019-06-29
[Core] Add ability to show modal prompts

### LINKED ISSUE #6713: [Enhancement] Display prompts
## Summary
Forms has the ability to show action sheets and alerts but no prompts. I tried using ACR User Dialogs but found out that the Android implementation is broken (soft keyboard does not appear upon initial display, incorrect keyboard type, no margins on the text input, etc.) This is not a 1-to-1 port of ACR prompts but a simple implementation that should work in most cases.

## API Changes

```
Task<string> Page.DisplayPromptAsync([arguments])
PromptArguments => will hold config data
```

## Intended Use Case
Any time when the user is asked to input value before proceeding to do what they want. User action is required in a modal view because it's important.

adrianknight89:
### Description of Change ###

See #6713 for more explanation. This only implements iOS and Android.

### Issues Resolved ### 

- fixes #6713

### API Changes ###

Added:
 - `Task<string> DisplayPromptAsync(string title, string message, string accept, string cancel, string placeholder = null, int? maxLength = null, Keyboard keyboard = default)`
- `PromptArguments(string title, string message, string accept = "OK", string cancel = "Cancel", string placeholder = null, int? maxLength = null, Keyboard keyboard = default)`
 
### Platforms Affected ### 

- iOS
- Android

thumbnail_Image-49

thumbnail_Image-50

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Page.cs
+++ b/Xamarin.Forms.Core/Page.cs
@@ -193,7 +193,7 @@ namespace Xamarin.Forms
 			return args.Result.Task;
 		}
 
-		public Task<string> DisplayPromptAsync(string title, string message, string accept = "OK", string cancel = "Cancel", string placeholder = null, int? maxLength = null, Keyboard keyboard = default(Keyboard))
+		public Task<string> DisplayPromptAsync(string title, string message, string accept = "OK", string cancel = "Cancel", string placeholder = null, int maxLength = -1, Keyboard keyboard = default(Keyboard))
 		{
 			var args = new PromptArguments(title, message, accept, cancel, placeholder, maxLength, keyboard);
 			MessagingCenter.Send(this, PromptSignalName, args);
--- a/Xamarin.Forms.Core/Page.cs
+++ b/Xamarin.Forms.Core/Page.cs
@@ -193,7 +193,7 @@ namespace Xamarin.Forms
 			return args.Result.Task;
 		}
 
-		public Task<string> DisplayPromptAsync(string title, string message, string accept = "OK", string cancel = "Cancel", string placeholder = null, int? maxLength = null, Keyboard keyboard = default)
+		public Task<string> DisplayPromptAsync(string title, string message, string accept = "OK", string cancel = "Cancel", string placeholder = null, int? maxLength = null, Keyboard keyboard = default(Keyboard))
 		{
 			var args = new PromptArguments(title, message, accept, cancel, placeholder, maxLength, keyboard);
 			MessagingCenter.Send(this, PromptSignalName, args);
--- a/Xamarin.Forms.Core/Page.cs
+++ b/Xamarin.Forms.Core/Page.cs
@@ -18,6 +18,8 @@ namespace Xamarin.Forms
 
 		public const string AlertSignalName = "Xamarin.SendAlert";
 
+		public const string PromptSignalName = "Xamarin.SendPrompt";
+
 		public const string ActionSheetSignalName = "Xamarin.ShowActionSheet";
 
 		internal static readonly BindableProperty IgnoresContainerAreaProperty = BindableProperty.Create("IgnoresContainerArea", typeof(bool), typeof(Page), false);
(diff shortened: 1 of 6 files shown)


jfversluis:
Could you resolve the conflict please? :)

adrianknight89:
Done.

adrianknight89:
Removed unsupported language features for build to pass

Xamarin.Forms.Core/Page.cs
```diff
@@ -191,6 +193,13 @@ public Task<bool> DisplayAlert(string title, string message, string accept, stri
 			return args.Result.Task;
 		}
 
+		public Task<string> DisplayPromptAsync(string title, string message, string accept = "OK", string cancel = "Cancel", string placeholder = null, int? maxLength = null, Keyboard keyboard = default(Keyboard))
```
StephaneDelcroix: don't use a nullable int, but a value of 0 or -1 to indicate that you don't want any limit

Xamarin.Forms.Core/PromptArguments.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace Xamarin.Forms.Internals
+{
+	public class PromptArguments
+	{
+		public PromptArguments(string title, string message, string accept, string cancel, string placeholder = null, int? maxLength = null, Keyboard keyboard = default(Keyboard))
```
StephaneDelcroix: needs to be decorated so it's not intelisense-visible

Xamarin.Forms.Core/PromptArguments.cs
```diff
@@ -0,0 +1,44 @@
...
+		{
+			if (string.IsNullOrWhiteSpace(accept) || string.IsNullOrWhiteSpace(cancel))
```
StephaneDelcroix: users shouldn't call this directly, so move the arg validation to the caller

Xamarin.Forms.Core/PromptArguments.cs
```diff
@@ -0,0 +1,44 @@
...
+				throw new Exception($"You must provide a value for {nameof(accept)} and {nameof(cancel)}.");
```
StephaneDelcroix: should be 2 different `ArgumentNullException`s

Xamarin.Forms.Core/PromptArguments.cs
```diff
@@ -0,0 +1,44 @@
...
+
+		public string Cancel { get; }
+
+		public string Placeholder { get; }
+
+		public int? MaxLength { get; }
+
+		public Keyboard Keyboard { get; }
+
+		public TaskCompletionSource<string> Result { get; }
```
StephaneDelcroix: the TCS shouldn't be public.

adrianknight89: The TCS is public in `AlertArguments`, and it's used in `Page.cs` when we return a task result. How do you want me to handle this?

Xamarin.Forms.Core/PromptArguments.cs
adrianknight89: If we move it to the caller, there is no need to do validation since both of these are defaulted to "OK" and "Cancel" respectively in the caller. So, I guess I'll remove the validation.

Xamarin.Forms.Core/PromptArguments.cs
StephaneDelcroix: you could expose only the returned Task, not the TCS, but as this is readonly, it's fine

jfversluis:
Failing test is unrelated

jrahma:
Will this have a mask or format? for example if I want to force number only or email format only, etc?

jfversluis:
@jrahma I don't think it will right now, no. Sorry!

jrahma:
@jfversluis appreciate if you add this as feature request..

sasivishnu:
Thanks for implementing this feature. @adrianknight89 

Will there be ability to set entry type as "Password"? That could solve the issue when we want to re-authenticate same user after session times out without moving out of the current screen.

adrianknight89:
@sasivishnu If you want this feature, please create an enhancement ticket so I or somebody else could take a look at it.

sasivishnu:
Cools thanks.. I will create one soon.

jfversluis:
@sasivishnu created an issue for the password (and others) request here #8363

sasivishnu:
> 
> 
> @sasivishnu created an issue for the password (and others) request here #8363

Thanks much. I had few other requests also. I created them separately #8365

jfversluis:
Great work @sasivishnu ! Thank you!

PR #6715, opened 2019-06-29
[Android] Fix various issues in object dispose methods - Shell edition.

### LINKED ISSUE #6640: Object reference not set to an instance with Android.ShellSectionRenderer.UnhookEvents ()
### Description
Object reference not set to an instance after clicking  log out in MasterdetailPage(This is Shell). And then press the hardware back button of Android.
### Steps to Reproduce

1. Create ContentPage ->LoginPage
2. Create MasterDetailPage -> This is the MasterPage

### Expected Behavior
When login successfully Go to MasterPage(This is Shell) using **App.Current.MainPage = MasterPage()** and click log out will go to Login Page. I use **App.Current.MainPage = Loginpage** and then press the hardware back button of Android. It will terminate the app.
### Actual Behavior
After Logout and press hardware back button of Android. Object reference not set to an instance shown.
### Basic Information

- Version with issue:Xamarin.Forms 4.0.0.497.661
  - Android

kvpt:
### Description of Change ###
Similar issues to #6467 but for the shell part.

### Issues Resolved ### 
- fixes #6640

### API Changes ###
Removed:
- ShellFragmentContainer => Removed Android Ctor (*)
- HeaderContainer => Removed Android Ctor (*)
- ShellSectionRenderer => Removed Android Ctor (*)
 
_Android Ctor (*)_
`(IntPtr javaReference, JniHandleOwnership transfer) : base(javaReference, transfer)`

### Platforms Affected ### 
- Android

None

Not applicable

### Testing Procedure ###
Run all of the solution tests cases, there must be no regression.

### PR Checklist ###
- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6640.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6640.cs
@@ -1,0 +1,95 @@
+using System;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 6640, "[Android] Crash when app go background", PlatformAffected.Android)]
+	public class Issue6640 : TestContentPage
+	{
+		protected override void Init()
+		{
+			Content = new StackLayout
+			{
+				Children =
+				{
+					new Button
+					{
+						AutomationId = "GoToShell",
+						Text = "Click the button",
+						Command = new Command(() =>
+						{
+							Application.Current.MainPage = new MasterPageShell();
+						})
+					}
+				}
+			};
+		}
+
+		[Preserve(AllMembers = true)]
+		public class MasterPageShell : TestShell
+		{
+			protected override void Init()
+			{
+				FlyoutHeader = new FlyoutHeader();
+				Items.Add(new FlyoutItem
+				{
+					Title = "Issue 6640",
+					Items =
+					{
+						new ShellSection
+						{
+							Items =
+							{
+								new ShellContent
+								{
+									Content = new ContentPage
+									{
+										Content = new StackLayout
+										{
+											Children =
+											{
+												new Button
+												{
+													AutomationId = "Logout",
+													Text = "Click the button",
+													Command = new Command(() =>
+													{
+														Application.Current.MainPage = new ContentPage
+														{
+															Content = new StackLayout
+															{
+																BackgroundColor = Color.White,
+																Children =
+																{
+																	new Label
+																	{
+																		Text = $"Press Back Arrow to send application to background and wait few seconds.{Environment.NewLine}Application must not crash." 
+																	}
+																}
+															}
+														};
+													})
+												}
+											}
+										}
+									}
+								}
+							}
+						}
+					}
+				});
+			}
+		}
+
+		[Preserve(AllMembers = true)]
+		public class FlyoutHeader : TestContentPage
+		{
+			protected override void Init()
+			{
+				Content = new Grid();
+			}
+		}
+	}
+}
--- a/Xamarin.Forms.Platform.Android/Renderers/IShellItemRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/IShellItemRenderer.cs
@@ -1,6 +1,5 @@
-﻿using Android.Support.V4.App;
+using Android.Support.V4.App;
 using System;
-using AView = Android.Views.View;
 
 namespace Xamarin.Forms.Platform.Android
 {
@@ -9,7 +8,5 @@ namespace Xamarin.Forms.Platform.Android
 		Fragment Fragment { get; }
 
 		ShellItem ShellItem { get; set; }
-
-		event EventHandler Destroyed;
 	}
 }
--- a/Xamarin.Forms.Platform.Android/Renderers/IShellItemRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/IShellItemRenderer.cs
@@ -8,5 +8,8 @@ namespace Xamarin.Forms.Platform.Android
 		Fragment Fragment { get; }
 
 		ShellItem ShellItem { get; set; }
+
+		[Obsolete("Destroyed event is obsolete as of version 4.2.0.")]
+		event EventHandler Destroyed;
 	}
 }
--- a/Xamarin.Forms.Platform.Android/Renderers/ShellBottomNavViewAppearanceTracker.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ShellBottomNavViewAppearanceTracker.cs
@@ -1,4 +1,4 @@
-﻿using Android.Content.Res;
+using Android.Content.Res;
 using Android.Graphics.Drawables;
 using Android.Support.Design.Internal;
 using Android.Support.Design.Widget;
@@ -15,6 +15,7 @@ namespace Xamarin.Forms.Platform.Android
 		ColorStateList _defaultList;
 		bool _disposed;
 		Color _lastColor = Color.Default;
+		ColorStateList _colorStateList;
 
 		public ShellBottomNavViewAppearanceTracker(IShellContext shellContext, ShellItem shellItem)
 		{
@@ -52,11 +53,9 @@ namespace Xamarin.Forms.Platform.Android
 #endif
 			}
 
-			var colorStateList = MakeColorStateList(titleColor, disabledColor, unselectedColor);
-			bottomView.ItemTextColor = colorStateList;
-			bottomView.ItemIconTintList = colorStateList;
-
-			colorStateList.Dispose();
+			_colorStateList = MakeColorStateList(titleColor, disabledColor, unselectedColor);
+			bottomView.ItemTextColor = _colorStateList;
+			bottomView.ItemIconTintList = _colorStateList;
 
 			SetBackgroundColor(bottomView, backgroundColor);
 		}
@@ -135,17 +134,17 @@ namespace Xamarin.Forms.Platform.Android
 
 		protected virtual void Dispose(bool disposing)
 		{
-			if (!_disposed)
+			if (!_disposed && disposing)
 			{
-				if (disposing)
-				{
-					_defaultList?.Dispose();
-				}
+				_disposed = true;
+
+				_defaultList?.Dispose();
+				_colorStateList?.Dispose();
 
 				_shellItem = null;
 				_shellContext = null;
 				_defaultList = null;
-				_disposed = true;
+				_colorStateList = null;
 			}
 		}
 
--- a/Xamarin.Forms.Platform.Android/Renderers/ShellBottomNavViewAppearanceTracker.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ShellBottomNavViewAppearanceTracker.cs
@@ -16,6 +16,7 @@ namespace Xamarin.Forms.Platform.Android
 		bool _disposed;
 		Color _lastColor = Color.Default;
 		ColorStateList _colorStateList;
+		ColorChangeRevealDrawable _colorChangeRevealDrawable;
 
 		public ShellBottomNavViewAppearanceTracker(IShellContext shellContext, ShellItem shellItem)
 		{
@@ -80,7 +81,9 @@ namespace Xamarin.Forms.Platform.Android
 				var child = menuView.GetChildAt(index);
 				var touchPoint = new Point(child.Left + (child.Right - child.Left) / 2, child.Top + (child.Bottom - child.Top) / 2);
 
-				bottomView.SetBackground(new ColorChangeRevealDrawable(_lastColor.ToAndroid(), color.ToAndroid(), touchPoint));
+				_colorChangeRevealDrawable = new ColorChangeRevealDrawable(_lastColor.ToAndroid(), color.ToAndroid(), touchPoint);
+
+				bottomView.SetBackground(_colorChangeRevealDrawable);
 
 				_lastColor = color;
 			}
@@ -140,11 +143,13 @@ namespace Xamarin.Forms.Platform.Android
 			{
 				_defaultList?.Dispose();
 				_colorStateList?.Dispose();
+				_colorChangeRevealDrawable?.Dispose();
 
 				_shellItem = null;
 				_shellContext = null;
 				_defaultList = null;
 				_colorStateList = null;
+				_colorChangeRevealDrawable = null;
 			}
 		}
 
--- a/Xamarin.Forms.Platform.Android/Renderers/ShellBottomNavViewAppearanceTracker.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ShellBottomNavViewAppearanceTracker.cs
@@ -43,7 +43,6 @@ namespace Xamarin.Forms.Platform.Android
 			var unselectedColor = controller.EffectiveTabBarUnselectedColor;
 			var titleColor = controller.EffectiveTabBarTitleColor;
 
-
 			if (_defaultList == null)
 			{
 #if __ANDROID_28__
@@ -65,27 +64,25 @@ namespace Xamarin.Forms.Platform.Android
 			if (_lastColor.IsDefault)
 				_lastColor = color;
 
-			using (var menuView = bottomView.GetChildAt(0) as BottomNavigationMenuView)
+			var menuView = bottomView.GetChildAt(0) as BottomNavigationMenuView;
+
+			if (menuView == null)
 			{
-				if (menuView == null)
-				{
-					bottomView.SetBackground(new ColorDrawable(color.ToAndroid()));
-				}
-				else
-				{
-					var index = _shellItem.Items.IndexOf(_shellItem.CurrentItem);
-					using (var menu = bottomView.Menu)
-						index = Math.Min(index, menu.Size() - 1);
+				bottomView.SetBackground(new ColorDrawable(color.ToAndroid()));
+			}
+			else
+			{
+				var index = _shellItem.Items.IndexOf(_shellItem.CurrentItem);
 
-					using (var child = menuView.GetChildAt(index))
-					{
-						var touchPoint = new Point(child.Left + (child.Right - child.Left) / 2, child.Top + (child.Bottom - child.Top) / 2);
+				var menu = bottomView.Menu;
+				index = Math.Min(index, menu.Size() - 1);
 
-						bottomView.Background?.Dispose();
-						bottomView.SetBackground(new ColorChangeRevealDrawable(_lastColor.ToAndroid(), color.ToAndroid(), touchPoint));
-						_lastColor = color;
-					}
-				}
+				var child = menuView.GetChildAt(index);
+				var touchPoint = new Point(child.Left + (child.Right - child.Left) / 2, child.Top + (child.Bottom - child.Top) / 2);
+
+				bottomView.SetBackground(new ColorChangeRevealDrawable(_lastColor.ToAndroid(), color.ToAndroid(), touchPoint));
+
+				_lastColor = color;
 			}
 		}
 
@@ -148,6 +145,6 @@ namespace Xamarin.Forms.Platform.Android
 			}
 		}
 
-#endregion IDisposable
+		#endregion IDisposable
 	}
 }
--- a/Xamarin.Forms.Platform.Android/Renderers/ShellBottomNavViewAppearanceTracker.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ShellBottomNavViewAppearanceTracker.cs
@@ -16,7 +16,6 @@ namespace Xamarin.Forms.Platform.Android
 		bool _disposed;
 		Color _lastColor = Color.Default;
 		ColorStateList _colorStateList;
-		ColorChangeRevealDrawable _colorChangeRevealDrawable;
 
 		public ShellBottomNavViewAppearanceTracker(IShellContext shellContext, ShellItem shellItem)
 		{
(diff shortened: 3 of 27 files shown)


kvpt:
It just the first pass.

I seen several others issues with dispose on (ShellItemRenderer, ShellSearchView, ShellToolbarTracker, ShellSectionRenderer).

I will also try to add the reproduction from issue #6640 later.

kvpt:
Added the reproduction case.
The UITest doesn't crash after the application go background, I tried to add a wait but it doesn't work.
So the test will be manual only.

kvpt:
The second pass is finished, I have removed all suspects using and dispose of ressources being used.

kvpt:
Changes can be reviewed. Thanks.

PureWeen:
@kvpt can you extract the parts out that apply to https://github.com/xamarin/Xamarin.Forms/issues/6640 ? Or does the whole thing apply to 6640?

kvpt:
What cause #6640 is that ShellRenderer dispose (so the parent) is called before ShellSectionRenderer (the child) destroy. And so when the Element is nulled in the ShellRenderer Dispose; after that it crash when ShellSectionRenderer Destroy is called. 

The true fix is the whole thing, explicitly dispose all the things in the right order.

However, if you want a rapid and simple fix, you can just remove this line.
https://github.com/xamarin/Xamarin.Forms/blob/0a157bfd603320b9dac493e85dd16633abc3d4a8/Xamarin.Forms.Platform.Android/Renderers/ShellRenderer.cs#L361

It will not correct the true problem but will fix the crash.
It's up to you if this alternative solution is acceptable or not before this pull request get merged.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6640.cs
```diff
@@ -0,0 +1,95 @@
+using System;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 6640, "[Android] Crash when app go background", PlatformAffected.Android)]
+	public class Issue6640 : TestContentPage
```
PureWeen: Can you add a UITest for this?

Xamarin.Forms.Platform.Android/Renderers/ShellBottomNavViewAppearanceTracker.cs
```diff
@@ -135,20 +131,20 @@ public void Dispose()
...
 		protected virtual void Dispose(bool disposing)
 		{
-			if (!_disposed)
+			if (!_disposed && disposing)
 			{
-				if (disposing)
-				{
-					_defaultList?.Dispose();
-				}
+				_disposed = true;
```
PureWeen: Can we move all the _disposed checks outside of this if loop like we try to do most other places

https://github.com/xamarin/Xamarin.Forms/blob/4187c67970480340d504111520462d8bb42f7a85/Xamarin.Forms.Platform.Android/AppCompat/MasterDetailPageRenderer.cs#L210

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6640.cs
kvpt: See my comment : https://github.com/xamarin/Xamarin.Forms/pull/6715#issuecomment-507080479

Xamarin.Forms.Platform.Android/Renderers/ShellBottomNavViewAppearanceTracker.cs
kvpt: It's done.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6640.cs
PureWeen: Now that unhook is part of dispose (https://github.com/xamarin/Xamarin.Forms/pull/6715/files#diff-edc32d3b021f02879f303175242f36b7R174) instead of DestroyView is there a way to emulate this behavior without the backgrounding? Perhaps after the mainpage is swapped out https://github.com/xamarin/Xamarin.Forms/pull/6715/files#diff-fbd24038d82294fdd7a696c78d6e02d4R59 the item could be removed from the collection to trigger the dispose? 

```
Application.Current.MainPage = new ContentPage
														{
															Content = new StackLayout
															{
																BackgroundColor = Color.White,
																Children =
																{
																	new Label
																	{
																		Text = $"Press Back Arrow to send application to background and wait few seconds.{Environment.NewLine}Application must not crash.",
																		AutomationId = "ReadyToGoBack"
																	}
																}
															}
														};

														Device.BeginInvokeOnMainThread(() =>
														{
															this.CreateContentPage();
															this.Items.RemoveAt(0);
														});
```

or something similar to that? We just need to trigger the destroy or dispose in order to have a UI test for this right? because that's all the backgrounding is really doing

kvpt: I think the only way to trigger the bug is on Activity Destroy.
Because it involve two methods :
- ShellRenderer.Dispose : Called when the main page change for exemple
- ShellSectionRenderer.OnDestroy : Called .... when Android want, and from this [stackoverflow answer](https://stackoverflow.com/a/17195813) it's only garanted when the fragment activity is also destroyed. This statement is coherent with the issue reproduction which is only triggered on activity destroy.
It's for that reason that I have replaced all the OnDestroy with Dispose.

kvpt:
Fixed merge conflicts and rebased on 4.2.0.

kvpt:
Rebased again on 4.2.0 to fix conflicts.

Also fixed a new issue introduced in #6572.
https://github.com/xamarin/Xamarin.Forms/blob/069f68e8b74ead5cfa5356ed6e75c7af54db34e4/Xamarin.Forms.Platform.Android/Renderers/ShellFlyoutTemplatedContentRenderer.cs#L257

kvpt:
I fixed a regression which caused an NRE during the dispose of the ShellToolbarTracker (Issue4684).
I runned locally all the tests and all the tests related to the shell are ok.

alanag13:
any idea when I can expect this to be merged? I'm running into some of the issues that this branch fixes. Thanks for your help @kvpt !

kvpt:
Fixed conflicts and rebased on master (since it's the new branch targeted).

dlakes-hx:
Is the intent that this will be part of a 4.2 service release, or part of 4.3?

rmarinho:
This one fixes #7327

rmarinho:
@kvp we want to pull this in for 4.2.0 service release, do you mind rebase it again on top of 4.2.0 ? 

thanks sorry going back and forth

Xamarin.Forms.Platform.Android/Renderers/ShellSearchView.cs
```diff
@@ -123,87 +124,98 @@ bool TextView.IOnEditorActionListener.OnEditorAction(TextView v, ImeAction actio
...
+				_linearLayout.Dispose();
 
-			_textBlock = null;
-			_clearButton = null;
-			_searchButton = null;
-			_cardView = null;
-			_clearPlaceholderButton = null;
-			_shellContext = null;
-			_searchHandlerAppearanceTracker = null;
+				_textBlock.Adapter = null;
```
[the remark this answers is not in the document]
kvpt: Good catch, fixed.

kvpt:
@rmarinho No problem, I rebased my changes onto 4.2.0 branch and applied the remarks.
But I doesn't have the right to switch the target branch of the PR so you will have to do it.

PureWeen:
Changed base and kicked off UI Tests

Xamarin.Forms.Platform.Android/Renderers/ShellFlyoutTemplatedContentRenderer.cs
```diff
@@ -191,10 +194,7 @@ protected virtual void UpdateFlyoutHeaderBehavior()
 		{
 			var context = _shellContext.AndroidContext;
 
-			Thickness margin = default(Thickness);
-
-			if (_flyoutHeader != null)
-				margin = _flyoutHeader.Margin;
+			var margin = _flyoutHeader?.Margin ?? default;
```
PureWeen: @kvpt we don't have  C# 7.3 on our 4.2 branch it's only 7.0 so can you just change this to default(Thickness) ?

PureWeen:
```suggestion
			var margin = _flyoutHeader?.Margin ?? default(Thickness);
```

PureWeen:
@samhouts I think out of that list the main ones to fix would be

> public event System.EventHandler Destroyed;

Keep this

> protected virtual void OnItemsCollectionChagned

I'm wondering if we should just obsolete this one given that it's a spelling error 👎 and then keep the new one



These constructors

> protected ShellSectionRenderer (IntPtr javaReference, Android.Runtime.JniHandleOwnership transfer);

are only used if there's a memory leak so I think it's better to get rid of them

> public override void OnDestroy ();

these are just overrides so they should be fine

kvpt:
I did it on purpose, I had to add the IDisposable interface to IShellFlyoutRenderer which is a breaking change, so I don't tried to limit the breaking changes after that because they are all compilation breaking changes.

> IShellFlyoutRenderer => Inherit IDisposable

I don't think this one can be avoided

> ShellSectionRenderer.OnItemsCollectionChagned =>ShellSectionRenderer.OnItemsCollectionChanged

I can obsolete it and add the correct one, but as there is one breaking change it's an opportunity to remove it now, faster this code is removed fewer libraries and user code are impacted

> ShellFragmentContainer - HeaderContainer - ShellSectionRenderer => Removed Android Ctor

These constructors are poisons, they add dispose problem, when Android try to restore an object with this constructor. Because when you try to acces an object disposed on the java side, instead of crashing, it silently continue, and create a memory leak.

> IShellItemRenderer - ShellItemRendererBase => Removed Destroyed event

At first I did not remove it, but after some thought, I decided to remove it, because now that we are not linked to destroy event it make no sense at all to have it. But we can obsolete them if necessary, for this one it's logic to do that if we want to be more conservative on breaking the contract.

PureWeen:
API 19 UI Test failure: FlyoutNavigationBetweenItemsWithNavigationStacks_lg_nexus_5

> I did it on purpose, I had to add the IDisposable interface to IShellFlyoutRenderer which is a breaking change, so I don't tried to limit the breaking changes after that because they are all compilation breaking changes.

Do you have to add IDisposable to IShellFlyoutRenderer? The Android implementation of IShellFlyoutRenderer already implements IDisposable so can we cast IShellFlyoutRenderer to *IDisposable* and if it's not null then call dispose.

https://github.com/xamarin/Xamarin.Forms/pull/6715/files#diff-ea33147016e2d6ca6e3029394a4cb3f7R361


> I don't think this one can be avoided
> ShellSectionRenderer.OnItemsCollectionChagned =>ShellSectionRenderer.OnItemsCollectionChanged
> I can obsolete it and add the correct one, but as there is one breaking change it's an opportunity to remove it now, faster this code is removed fewer libraries and user code are impacted

I would still just obsolete it for now and we can look at removing it when we bump up to 5.0

> ShellFragmentContainer - HeaderContainer - ShellSectionRenderer => Removed Android Ctor

> These constructors are poisons, they add dispose problem, when Android try to restore an object with this constructor. Because when you try to acces an object disposed on the java side, instead of crashing, it silently continue, and create a memory leak.

These are fine to remove


> IShellItemRenderer - ShellItemRendererBase => Removed Destroyed event
> At first I did not remove it, but after some thought, I decided to remove it, because now that we are not linked to destroy event it make no sense at all to have it. But we can obsolete them if necessary, for this one it's logic to do that if we want to be more conservative on breaking the contract.

Obsolete them. 

96.3 percent of the time that we break the ABI we have to go back and fix it anyway so it's better to just be cautious about it now

kvpt:
Ok I will try for the IDisposable change, and obsolete the others changes.

kvpt:
I reverted most breaking changes, the only remaining are the specific android contructors.

@samhouts For the 5.0 please consider to add the IDisposable to the interface IShellFlyoutRenderer, iOS version have it.

@PureWeen For the failed test, I executed the test 6738 on an api 19 emulator and it works fine for me. However I executed it manually because the test don't want to start on my machine currently.
Manually deploy the test app worked before to enable the tests to run, but not now, strange.

kvpt:
Ok from the stacktrace, it's clear it's a problem with dispose.
I think it's because ColorChangeRevealDrawable need to be disposed.

https://github.com/xamarin/Xamarin.Forms/blob/4fd2d6b45b7f287433ea2bf52581d376a06b999f/Xamarin.Forms.Platform.Android/Renderers/ColorChangeRevealDrawable.cs#L26

This part in the constructor launch an async operation, and if we dispose the BottomNavigationView before the operation finish, I think it can cause this issue. 

I will test that.

kvpt:
@PureWeen Can you relaunch UI tests please
I implemented the dispose of the ColorChangeRevealDrawable class to force the animation to end before disposing. I think it's what cause the race condition. The tests will confirm that or not.

Xamarin.Forms.Platform.Android/Renderers/ShellBottomNavViewAppearanceTracker.cs
```diff
@@ -80,7 +81,9 @@ protected virtual void SetBackgroundColor(BottomNavigationView bottomView, Color
 				var child = menuView.GetChildAt(index);
 				var touchPoint = new Point(child.Left + (child.Right - child.Left) / 2, child.Top + (child.Bottom - child.Top) / 2);
 
-				bottomView.SetBackground(new ColorChangeRevealDrawable(_lastColor.ToAndroid(), color.ToAndroid(), touchPoint));
+				_colorChangeRevealDrawable = new ColorChangeRevealDrawable(_lastColor.ToAndroid(), color.ToAndroid(), touchPoint);
+
+				bottomView.SetBackground(_colorChangeRevealDrawable);
```
PureWeen: maybe this should be set to null on dispose?

kvpt: I changed my approach.
I now only dispose the old background after it has been changed.
The final dispose is done in ShellItemRenderer.
I succeed to reproduce when disposing the AppearanceTracker before the bottom view background. But in the current order it work fine. So not sure if it will work.

kvpt:
@PureWeen Can you restart another batch of tests please.

If this time the exception occur again I will have the Xamarin Android "Revive" constructor and throw an exception with the callstack in it to see where the access on the disposed background is done.

PureWeen:
Alright @kvpt let's see if this works

The part of the UI test that's failing performs two operations
- inserts a page in the back stack
- then switches tabs

Both of those cause a SendAppearenceChanged event to fire so both of those cause the *ColorChangeRevealDrawable* to get created then destroyed. 

I'm guessing there's a race condition where the first one is being drawn after the second *SendAppearenceChanged* disposes. 

I changed the code so it doesn't try to set the background again if the color hasn't changed. It might even make sense to not call *SendAppearenceChanged* when InsertPageBefore is called as well

kvpt:
@PureWeen It make sense, this kind of issue is always a race condition.
And yes I also think it's useless to call AppearanceChanged if the page has not been rendered yet.

PureWeen:
Dang that still didn't work 👎

PureWeen:
@kvpt so on the last round I commented out the dispose completely
https://github.com/xamarin/Xamarin.Forms/pull/6715/commits/1e135afd157de4da9c6bd5c9130e6442e60da4b9#diff-0b13b4b03eb6fc03312913a8efabf296R94

And still the same exception is popping up

kvpt:
If you want to remove the dispose, the problematic one is here : 
https://github.com/kvpt/Xamarin.Forms/blob/1e135afd157de4da9c6bd5c9130e6442e60da4b9/Xamarin.Forms.Platform.Android/Renderers/ShellItemRenderer.cs#L100

And now i see this again, I think it's useless, because the dispose of the BottowView will surely dispose the background it self.

PureWeen:
@kvpt  💚 💚💚💚💚💚💚💚

Your last suggestion looks like it worked!

https://github.com/xamarin/Xamarin.Forms/pull/6715/commits/e73a0608b2b6809610f119e55cb0b5cbd2b17477#diff-4144388f1cafc65a54c1026775d46ab3R99

I'll refine and test a bit next week and then we should be able to get this merged

PureWeen:
Alright all the UI Tests relevant to this PR passed. I rebased with the latest 4.2.0 code and kicked them off for good measure but this PR should be good to review now.

Xamarin.Forms.Platform.Android/Renderers/IShellItemRenderer.cs
```diff
@@ -10,6 +9,7 @@ public interface IShellItemRenderer : IDisposable
 
 		ShellItem ShellItem { get; set; }
 
+		[Obsolete("Destroyed event is obsolete as of version 4.2.0.")]
```
rmarinho: We need to add this to release notes for sr4

Xamarin.Forms.Platform.Android/Renderers/ShellBottomNavViewAppearanceTracker.cs
```diff
@@ -135,20 +138,23 @@ public void Dispose()
...
 				_shellItem = null;
 				_shellContext = null;
 				_defaultList = null;
-				_disposed = true;
+				_colorStateList = null;
 			}
 		}
 
-#endregion IDisposable
+		#endregion IDisposable
```
rmarinho:
```suggestion
#endregion IDisposable
```

Xamarin.Forms.Platform.Android/Renderers/ShellItemRenderer.cs
```diff
@@ -253,10 +255,18 @@ protected virtual void OnMoreItemSelected(ShellSection shellSection, BottomSheet
...
 
-		protected virtual void OnMoreSheetDismissed(object sender, EventArgs e) => OnShellSectionChanged();
+		protected virtual void OnMoreSheetDismissed(object sender, EventArgs e)
+		{
+			OnShellSectionChanged();
+
+			if (_bottomSheetDialog != null)
+			{
+				_bottomSheetDialog.DismissEvent -= OnMoreSheetDismissed;
+				_bottomSheetDialog.Dispose();
```
rmarinho:
```suggestion
				_bottomSheetDialog.Dispose();
				_bottomSheetDialog = null;
```

Xamarin.Forms.Platform.Android/Renderers/ShellItemRendererBase.cs
```diff
@@ -23,14 +22,16 @@ ShellItem IShellItemRenderer.ShellItem
 			set { ShellItem = value; }
 		}
 
-		public event EventHandler Destroyed;
-
+		[Obsolete("Destroyed event is obsolete as of version 4.2.0.")]
+		public event EventHandler Destroyed;
```
rmarinho: Needs to be added to release notes

Xamarin.Forms.Platform.Android/Renderers/ShellRenderer.cs
```diff
@@ -332,15 +319,14 @@ public override void Draw(Canvas canvas)
...
-
-				paint.Color = _color;
-
-				canvas.DrawRect(new Rect(0, 0, bounds.Right, _topSize), paint);
+				using (var paint = new Paint())
+				{
+					paint.Color = _color;
 
-				canvas.DrawRect(new Rect(0, bounds.Bottom - _bottomSize, bounds.Right, bounds.Bottom), paint);
+					canvas.DrawRect(new Rect(0, 0, bounds.Right, _topSize), paint);
```
rmarinho: why two DrawRect ?

Xamarin.Forms.Platform.Android/Renderers/ShellRenderer.cs
```diff
@@ -361,19 +347,24 @@ void IDisposable.Dispose()
...
-				Element = null;
-				// TODO: free unmanaged resources (unmanaged objects) and override a finalizer below.
-				// TODO: set large fields to null.
+				((IShellController)Element).RemoveAppearanceObserver(this);
+
+				// This cast is necessary because IShellFlyoutRenderer doesn't implement IDisposable
+				(_flyoutRenderer as IDisposable)?.Dispose();
 
-				_disposed = true;
+				_currentRenderer.Dispose();
```
rmarinho:
```suggestion
				_currentRenderer.Dispose();
				_currentRenderer = null;
```

Xamarin.Forms.Platform.Android/Renderers/ShellToolbarTracker.cs
```diff
@@ -453,34 +454,39 @@ protected virtual void UpdateToolbarItems(Toolbar toolbar, Page page)
 			var menu = toolbar.Menu;
 			menu.Clear();
 
-			foreach (var item in page.ToolbarItems)
+			if (page != null)
 			{
-				using (var title = new Java.Lang.String(item.Text))
```
rmarinho: Why did we changed this from using? any special issue?

PureWeen:
@kvpt Thank you for all the work on this one!!

We've merged in the slightly slimmed down version of this here
https://github.com/xamarin/Xamarin.Forms/pull/7768

I'm going to close this for now. 

If there are still aspects of this PR you think are important to apply then can you rebase and we'll reopen?

kvpt:
@PureWeen Thank you too.
Don't have much time to follow this at the moment.
I replaced destroy with dispose call because we can't control when android call destroy, but the root cause of the problem was the dispose/destroy order which was not right.
If this part is ok now, it's ok for me.
I also see that you kept the "leak constructor" removal, that was an important thing for me.
Thank you again.

PR #6735, opened 2019-07-01
Unhook Page from embedded Application on disposal

hartez:
### Description of Change ###

Unhook Page from embedded Application on disposal so the Page can be collected. 

This is a supplementary fix to #6636; that change allowed Pages which were not hooked to the current Application object to be collected, but did not allow the most recent Page associated with the Application to be collected. This change allows the most recent page to be collected.

### Issues Resolved ### 

- additional fix for #4671

### API Changes ###
 
Application.cs

Added:
`protected internal virtual void CleanUp();`

Modified: 
`public NavigationProxy NavigationProxy { get; }` => `public NavigationProxy NavigationProxy { get; private set; }`

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Application.cs
+++ b/Xamarin.Forms.Core/Application.cs
@@ -348,7 +348,7 @@ namespace Xamarin.Forms
 		protected internal virtual void CleanUp()
 		{
 			// Unhook everything that's referencing the main page so it can be collected
-			// This is only comes up if we're disposing of an embedded Forms app, and will
+			// This only comes up if we're disposing of an embedded Forms app, and will
 			// eventually go away when we fully support multiple windows
 			if (_mainPage != null)
 			{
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Core/Application.cs
```diff
@@ -345,6 +345,21 @@ async Task SetPropertiesAsync()
 
 		}
 
+		protected internal virtual void CleanUp()
+		{
+			// Unhook everything that's referencing the main page so it can be collected
+			// This is only comes up if we're disposing of an embedded Forms app, and will
```
samhouts:
```suggestion
			// This only comes up if we're disposing of an embedded Forms app, and will
```

PR #6748, opened 2019-07-02
Updated TemplateUtilities.cs

Juansero29:
TemplateUtilities is now public, as it should be.

### Issues Resolved ### 

- Fixes #6747

### API Changes ###
TemplateUtilities is now public, so its public methods inside can be used outside the namespace Xamarin.Forms. Especially the `GetTemplateChild` method.

Changed:
 - internal static class TemplateUtilities => public static class TemplateUtilities

### Platforms Affected ### 
- Core/XAML (all platforms)

### Behavioral/Visual Changes ###
None

### Before/After Screenshots ### 
Not applicable

### PR Checklist ###

The tests had already be done by @andreinitescu in the PR #5691

- [x] Has automated tests 
- [x] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard

PR #6752, opened 2019-07-02
Add protected GetTemplateChild to ContentPage and ContentView 

### LINKED ISSUE #6747: [Enhancement] Xamarin.Forms.TemplateUtilities.GetTemplateChild(string name) should be accessible from exterior namespaces
## Summary
This method should be accessible from other namespaces than only the 'Xamarin.Forms' namespace. It was mentioned by @davidbritch on the 4.1 [pre-release blog article](https://devblogs.microsoft.com/xamarin/checkbox-xamarin-forms-4-1-0-pre-release/) as available, but it isn't. When trying to use it from a TemplatedView, I get the error `'TemplateUtilities' is inaccessible due to its protection level`. Simply put, **Xamarin.Forms.TemplateUtilities should be public, since it has public methods inside.**

## API Changes

Xamarin.Forms.TemplateUtilities and its public methods would now be accessible outside the Xamarin.Forms namespace.

e.g.

```
<controls:MyCustomControl... >
    <controls:MyCustomControl.ControlTemplate>
         <ControlTemplate>
              <Label x:Name="myLabel"/>
         </ControlTemplate>
    <controls:MyCustomControl.ControlTemplate>
</controls:MyCustomControl>
```

```
namespace MyCustomApp 
{
class MyCustomControl : TemplatedView
{
      Label _myLabel;

       protected override OnApplyTemplate()
       {  
               _myLabel = Xamarin.Forms.TemplateUtilities.GetTemplateChild("myLabel");
       }
}
}
```

## Intended Use Case
I intend to force some properties of a control (let's say a checkbox) under certain conditions inside the code of my CustomControl. This can only be achieved by having a reference to the said checkbox and changing its properties. I already use a 'self-made' GetTemplateChild method, but I would like to use the official API from Xamarin.Forms. 

@andreinitescu  see this.

andreinitescu:
### Description of Change ###

Add the protected method `GetTemplateChild(string name)` to `ContentPage` and `ContentView`. A continuation for the work on #5691 🤦‍♀ 

I also updated the unit test to really use the protected version of `GetTemplateChild` instead of the internal extension method implemented in `TemplateUtilities` 🤦‍♂ 

**`GetTemplateChild` introduced by #5691 is really useless without this PR so I hope we can get this merged in fast for the next 4.1 update.**

Instead of adding GetTemplateChild to both ContentPage and ContentView, another option is maybe to make the `TemplateUtilities` class public but this requires further thinking as some of the methods in `TemplateUtilities` are not ready to be made public and it would require some discussion and more refactoring (for example, `OnContentChanged` and `OnControlTemplateChanged` should really take an `IControlTemplated` instance and not a `BindableObject`). 
Therefore, for now, this PR should be faster to be accepted and if we decide to change later how `GetTemplateChild` works, it will be backward compatible.

### Issues Resolved ### 

- fixes #6747

### API Changes ###

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/TemplatedViewUnitTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/TemplatedViewUnitTests.cs
@@ -36,7 +36,7 @@ namespace Xamarin.Forms.Core.UnitTests
 		}
 
 		[Test]
-		public void GetTemplateChildShouldWork()
+		public void GetContentViewTemplateChildShouldWork()
 		{
 			var xaml = @"<ContentView
 					xmlns=""http://xamarin.com/schemas/2014/forms""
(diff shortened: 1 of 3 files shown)


Juansero29:
Nice. Was waiting for the PR :) Hope it gets accepted, I'll drop mine (#6748 ) if that seems correct for you too. Thanks for the fix!

andreinitescu:
@Juansero29 Thanks for taking time to report the problem and sorry again for the issue.

samhouts:
@andreinitescu Please target 4.1. Thanks!

andreinitescu:
@samhouts Done

PR #6753, opened 2019-07-02
Fixed: Button CornerRadius doesn't work in explicit style

### LINKED ISSUE #5046: Button CornerRadius doesn't work in explicit style when implicit style exists
### Description

Applies to Windows UWP only.

When you have
* an implicit style on a Button
* an explicit style on a Button containing a CornerRadius set to anything

Then, the CornerRadius gets set to a value of 5, ignoring the value that is set in the explicit style.

### Steps to Reproduce

Add some code to a page with both an implicit button style (no key), and an explicit button stlye (has a key).

Add a button that references the explicit style.

Sample code:

```
    <ContentPage.Resources>
        <Style TargetType="Button">
            <!--<Setter Property="CornerRadius" Value="0"></Setter>-->
            <Setter Property="BackgroundColor" Value="Blue"></Setter>
            <Setter Property="TextColor" Value="White"></Setter>
        </Style>
        
        <Style x:Key="ButtonStyle" TargetType="Button">
            <Setter Property="CornerRadius" Value="0"></Setter>
            <Setter Property="BackgroundColor" Value="Blue"></Setter>
            <Setter Property="TextColor" Value="Yellow"></Setter>
        </Style>
    </ContentPage.Resources>
    
    <StackLayout>
        <!-- Place new controls here -->
        <Label Text="Welcome to Xamarin.Forms!" 
           HorizontalOptions="Center" />
        
        <Button Text="Hello World" Style="{StaticResource ButtonStyle}"></Button>
    </StackLayout>
```

### Expected Behavior

The CornerRadius from the button's explicit style is used. In the example above, the CornerRadius should be 0.

### Actual Behavior

The CornerRadius from the button's explicit style is not used. In the example above, the CornerRadius is 5.

### Basic Information

- Version with issue: 3.4.0.1008975, 3.4.0.1009999, 3.5.0.62956-pre2
- Last known good version: 3.3.0.912540
- IDE: Visual Studio 2018 15.9.4
- Platform Target Frameworks: 
  - UWP:  16299
- Nuget Packages: Just Xamarin.Forms

### Screenshots

cornerradius

Mikescher:
### Description of Bug ###

See #5046 for bug description (despite what the bug says I could reproduce it also on iOS & Android)  

A little bit of digging made clear what is happening:

 - the explicit Style "ButtonStyle" is applied to our Button
     * because of that the CornerRadius value is changed from `-1` to `0`
     * for backwards-compatibility reasons the BorderRadius property is also changed to `0`
 - the Style of our button is changed from "ButtonStyle" to a merged style consisting of the explicit "ButtonStyle" and the implicit style
     * First the old style in un-applied
         * the CornerRadius value is changed back from `0` to its default value of `-1`
         * for backwards-compatibility reasons we also set the BorderRadius property to `5` (its default value)
         * the BorderRadiusPropertyChanged handler now also changes CornerRadius to `5` **(This is the bug)**
         * because of the previous step CornerRadius is now no longer marked as `SetValuePrivateFlags.FromStyle`
     * Now the new (merged) style is applied
         * The BindableObject.SetValue wants to change CornerRadius again to 0, but it doesn't because it sees that its current value was set manually and not FromStyle.

### Description of Change ###

I fixed it by adding a single `if (button.cornerOrBorderRadiusSetting) return;` in BorderRadiusPropertyChanged of the Button.  
This has the effect that when BorderRadius is changed because CornerRadius was changed (= its only changed for backwards-compatibility) then BorderRadius does not try to change CornerRadius again (CornerRadius already has the value that was intended).

Btw: I'm not sure if I should commit my "Test"-code `Issue5046.xaml` in the Xamarin.Forms.Issues project - If you don't want it I can also remove it :)

### Issues Resolved ### 

 - fixes #5046 
 
### API Changes ###

 None

### Platforms Affected ### 

- Core/XAML (all platforms)

Well, if anyone ran into this bug he will see the correct behavior now

Before:  
before

After:  
after

### Testing Procedure ###

I added an example in the *Xamarin.Forms.Issues* project - simply open it and be amazed that the button has sharp corners (or comment out the one changed line in Button.cs and look at the child-safe round corners)

### PR Checklist ###

- [x] Has automated tests 


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5046.xaml.cs
```diff
@@ -0,0 +1,27 @@
...
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using Xamarin.Forms.Xaml;
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if APP
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 5046, "CornerRadius doesn't work in explicit style when implicit style exists", 
+	       PlatformAffected.Android | PlatformAffected.iOS)]
```
jfversluis: No UWP platform? Since that was the original culprit

Xamarin.Forms.Core.UnitTests/ButtonUnitTest.cs
```diff
@@ -323,5 +324,32 @@ private void AssertButtonContentLayoutsEqual(Button.ButtonContentLayout layout1,
...
+			                     </Style>
+			                     <Style x:Key=""ButtonStyle"" TargetType=""Button"">
+			                         <Setter Property=""CornerRadius"" Value=""0""></Setter>
+			                         <Setter Property=""TextColor"" Value=""Yellow""></Setter>
+			                     </Style>
+			                 </ContentPage.Resources>
+			                     <Button x:Name=""Button0"" Style=""{StaticResource ButtonStyle}"" />
+			             </ContentPage>";
+
+			var contentPage = new MyTestContentView();
```
jfversluis: Why not a regular `ContentView`?

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5046.xaml.cs
```diff
@@ -0,0 +1,27 @@
...
+	[XamlCompilation(XamlCompilationOptions.Compile)]
+	public partial class Issue5046 : ContentPage
```
jfversluis: If you really want to go all out you could add a UI test for this. Change the base page to a `TestContentPage` and add a test that verifies this. There are a lot of examples to be found in the project. Then again, you already have a unit test so not absolutely necessary.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5046.xaml.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using Xamarin.Forms.Xaml;
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if APP
```
jfversluis: Why the `APP` directive?

jfversluis:
> Btw: I'm not sure if I should commit my "Test"-code Issue5046.xaml in the Xamarin.Forms.Issues project - If you don't want it I can also remove it :)

Please do! It helps us go through this issue and solution much faster. Else we would have to add it ourselves. As mentioned in the comments: if you add this test code you can easily add a UI test for it as well. I've added some comments which are more questions, but overall looks good!

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5046.xaml.cs
Mikescher: I could only test it on Android and iOS, but you're right the original issue was UWP and I guess this should also happen there - I added PlatformAffected.UWP

Xamarin.Forms.Core.UnitTests/ButtonUnitTest.cs
Mikescher: No special reason. I changed it now. I guess it looks cleaner to simply use a ContentView?

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5046.xaml.cs
Mikescher: I tried, but I didn't manage to assert the CornerRadius in these UI test cases. With RunningApp.Query I can only get an AppResult object with a very limited set of properties.  
All the other UITests I looked at also only verify that a component exists, or that the app doesn't crash and not a specific property of one component.
But I think the unit test should be enough (?) otherwise it would be testing the same thing twice anyway

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5046.xaml.cs
Mikescher: I copied the structure from Issue5268.xaml.cs so I thought the APP directive was in some kind neccessary for something :).  
In the other Issue classes the methods InitializeComponent and the click handlers are all wrapped in `#if APP` and if I don't do that I get some Intellisense errors (but no compiler errors).  
So I changed it now so that only InitializeComponent and Button_Clicked is wrapped in `#if APP`.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5046.xaml.cs
jfversluis: It's fine for now with the unit tests. Did you see the .Property (from the top of my head) method? I think with that you should be able to query all properties on an object :)

Xamarin.Forms.Core.UnitTests/ButtonUnitTest.cs
jfversluis: If there is no special reason to create an inheritance we might as well just use the base class indeed.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5046.xaml.cs
jfversluis: No problem, this way the UWP tests on our side will trigger as well so they will run for you there. Whenever it runs into a problem we'll let you know. Or, if you are not able to fix it since you do not have Windows, we will have a look

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5046.xaml.cs
jfversluis: Strange, I haven't seen this one before so I'm not sure what it does. But I don't think it's necessary for this.

Xamarin.Forms.Core.UnitTests/ButtonUnitTest.cs
```diff
@@ -323,5 +324,29 @@ private void AssertButtonContentLayoutsEqual(Button.ButtonContentLayout layout1,
 			Assert.AreEqual(layout1.Position, bcl.Position);
 			Assert.AreEqual(layout1.Spacing, bcl.Spacing);
 		}
+
+		[Test]
+		public void ButtonCornerRadiusSetInExplicitStyle()
```
StephaneDelcroix: should be moved to Xamarin.Forms.Xaml.UnitTests if the same behavior can't reproduced in a code-only test

StephaneDelcroix: but ideally you should reproduce this test in code only, and leave it there

Mikescher: I tried to reproduce it in code, but i didn't manage to reproduce the bug because its kinda dependent on the styles being applied and unapplied in the right order as the xaml is doing it.
But I moved it to Xamarin.Forms.Xaml.UnitTests as you wanted.

Xamarin.Forms.Xaml.UnitTests/Xamarin.Forms.Xaml.UnitTests.csproj
```diff
@@ -30,4 +30,14 @@
       <Link>BaseTestFixture.cs</Link>
     </Compile>
   </ItemGroup>
+  <ItemGroup>
```
StephaneDelcroix: drop this, we add the files by wildcard

Xamarin.Forms.Xaml.UnitTests/ButtonCornerRadius.xaml.cs
```diff
@@ -0,0 +1,43 @@
+﻿using NUnit.Framework;
+using Xamarin.Forms.Core.UnitTests;
+
+namespace Xamarin.Forms.Xaml.UnitTests
+{
+	[XamlCompilation(XamlCompilationOptions.Compile)]
```
StephaneDelcroix:
```suggestion
```

Xamarin.Forms.Xaml.UnitTests/Xamarin.Forms.Xaml.UnitTests.csproj
Mikescher: done

Issue #2758, opened 2018-05-17
The order of 'IsEnabled' and 'Command' should not matter


mattleibow:
### Description

I have two buttons:

```xml
<Button
    Text="Button1"
    Command="{Binding TheCommand}"
    IsEnabled="False" />

<Button
    Text="Button2"
    IsEnabled="False"
    Command="{Binding TheCommand}" />
```

For some reason, the `IsEnabled` property is overwritten in Button2 - this enables the button incorrectly.

### Expected Behavior

Both buttons stay disabled.

### Actual Behavior

The button with the command applied later overwrites the enabled state.

### Basic Information

- Version with issue: 2.5.x and 3.0
- Last known good version: unsure
- IDE: VS2017
- Platform Target Frameworks: all

mattleibow:
I just checked with native UWP, and the matrix is the platform does - and the order doesn't matter.

hartez:
I don't think is is a XAML issue. IsEnabledProperty and Command.CanExecute are each changing IsEnabled completely independent of one another. You could fail the matrix above with pure C#.

mattleibow:
I was looking at this and I think the real issue is the fact that the IsEnabled property is being overwritten entirely.
What we can do is make use of the coercing feature and actually use that. When IsEnabled is set, we store the actual value and then run a coerce on it. In the Button, we implement a coerce function to return the Command.CanExecute value. The resulting IsEnabled should be the `propertyValue && infusedValue`.

```csharp
class VisualElement {
    BindableProperty IsEnabledProperty = BindableProperty.Create(
        ...,
        coerceValue: OnIsEnabledCoerce);

    static object OnIsEnabledCoerce(BindableObject bindable, object value) {
        var element = (VisualElement)bindable;
        if (element == null)
            return false;
        if ((bool)value) {
            return element.IsEnabledCore;
        } else {
            return false;
        }
    }

    protected virtual bool IsEnabledCore { get { return true; } }
}

class Button {
    protected override bool IsEnabledCore {
        get { return base.IsEnabledCore && CanExecute; }
    }
    bool CanExecute {
        get { return _canExecute; }
        set {
            if (_canExecute != value) {
                _canExecute = value;
                ReCoerceValue(IsEnabledProperty);
            }
        }
    }
    void CommandCanExecuteChanged(object sender, EventArgs eventArgs) {
        CanExecute = Command?.CanExecute(CommandParameter) ?? true;
    }
}
```

[...]

PR #6762, opened 2019-07-03
[Shell] Fix back button behavior so that it only applies the property you've set opposed to replacing the entire thing

### LINKED ISSUE #5744: [Shell] BackButtonBehavior issues
### Description

`BackButtonBehavior` seems like it has some holes in it. Here's my issues:

1. Its name. It's not a `Behavior`. How about `BackButtonOptions` or `BackButtonConfiguration`?
1. Once you set a `BackButtonBehavior` on a `ContentPage`, you automatically lose the default back arrow, and have to set your own (on iOS).
1. I like the commanding model. Maybe it should have a `Clicked` event too? Are there scenarios where users will want to handle a back button intercept in their page code-behind, while still having the page `BindingContext` set to a ViewModel?
1. It seems like a typical scenario would be binding `BackButtonBehavior.Command` to an `ICommand` in a ViewModel. The idea being that you can intercept the back button press, and display an alert if there's unsaved data on the page, and cancel navigation if required. However, there's no easy way to cancel navigation in this scenario.
1. https://github.com/xamarin/Xamarin.Forms/issues/5734

PureWeen:
### Description of Change ###
This changes BackButtonBehavior so that it only overrides the properties you specify and not the entire thing. This allows users to attach a command or change the icon on the BackButton without totally breaking everything

### Issues Resolved ### 
- fixes #5744

### Platforms Affected ### 
- iOS
- Android

### Testing Procedure ###
- I added a UI test to make sure command works
- There are buttons that you can use to set all the back button behaviors. Just click on different combinations to make sure it all works how you would expect it to

### PR Checklist ###

- [x] Has automated tests 


PureWeen:
@alanag13  It'll probably be a couple weeks before I can tie a bow on this one :-) But this is basically what I was thinking for BackButtonBehavior

alanag13:
@PureWeen cool don't wait for me but I'll see if I can add anything to this before then.

PureWeen:
@paymicro 

> IsEnabled resets after sets command or command parameter

I think this is pretty standard for most things that have commands like a Button. Once we swap out the command IsEnabled just sets based on the CanExecute of the Command

paymicro:
> I think this is pretty standard for most things that have commands like a Button. Once we swap out the command IsEnabled just sets based on the CanExecute of the Command

For me, this is not obvious and not expected behavior.
it turns out that the order of setting `isEnabled` and the commands will affect the result.
If user sets 
```C# 
behavior.IsEnabled = false;
behavior.Command = new Command()
```
`IsEnabled` is true, but
```C# 
behavior.Command = new Command()
behavior.IsEnabled = false;
```
`IsEnabled` is false

PureWeen:
@paymicro yea that's been the discussion around button for some time as well

There's some discussion here about it
https://github.com/xamarin/Xamarin.Forms/issues/2758

So in theory we could change how it works here which would make it inconsistent with Button. So that the result of the BackButton being enabled is an && of *IsEnabled && Command.CanExecute(param)* but that would make its behavior inconsistent with Button.

PureWeen:
build --uitests

PureWeen:
ios10 failure unrelated and fixed here
https://github.com/xamarin/Xamarin.Forms/pull/7132

PureWeen:
All green

Xamarin.Forms.Platform.iOS/Extensions/BindableObjectExtensions.cs
```diff
@@ -0,0 +1,21 @@
+﻿#if __ANDROID__
```
samhouts: Do we have any precedent for linking between platforms like this?

PureWeen: We do it for the MaterialColors in the material libraries. I would hardly call that creating *precedent* though.

I just really didn't want to copy and paste the same code two places. I think the *precedent* has been to just duplicate code into every platform like the *IsOneOf* code

But.... it's duplicated 😢

samhouts: I think we would actually move this up to the Core and just set it to EditorBrowseable(Never), since there's nothing platform specific about it.

PureWeen: @samhouts moved!

francipvb:
I don't know if this is related to this, but How can we change the title presented to screen readers in Flyout and Back buttons?

I will post an issue on this if I can't find anything.

Thanks,

PR #6764, opened 2019-07-03
[Shell, WIP] Ongoing Shell work merged together for 3rd parties to test

PureWeen:
### Description of Change ###

Works best on Android currently. 

Link to nuget: https://dev.azure.com/xamarin/public/_build/results?buildId=14440&view=artifacts&type=publishedArtifacts

Current work with Prism: 
 https://github.com/dansiegel/Prism-Shell-Stream

### DIFF
--- a/Xamarin.Forms.Core/Shell/ShellNavigationService.cs
+++ b/Xamarin.Forms.Core/Shell/ShellNavigationService.cs
@@ -149,8 +149,18 @@ namespace Xamarin.Forms
 
 			builder.Append(queryString);
 
-			FullUriWithImplicit = ShellUriHandler.ConvertToStandardFormat(null, new Uri(builder.ToString(), UriKind.Relative));
-			FullUri = ShellUriHandler.ConvertToStandardFormat(null, new Uri(builderWithoutImplicit.ToString(), UriKind.Relative));
+			if (builder.Length > 0)
+				FullUriWithImplicit = new Uri(builder.Insert(0, @"//").ToString(), UriKind.Relative);
+
+			if (builderWithoutImplicit.Length > 0)
+				FullUri = new Uri(builderWithoutImplicit.Insert(0, @"//").ToString(), UriKind.Relative);
+
+
+			//if(builder.Length > 0)
+			//	FullUriWithImplicit = ShellUriHandler.ConvertToStandardFormat(null, new Uri(builder.Insert(0, @"//").ToString(), UriKind.Relative));
+
+			//if(builderWithoutImplicit.Length > 0)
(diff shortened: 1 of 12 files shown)


Xamarin.Forms.Core/Routing.cs
```diff
@@ -9,22 +9,32 @@ public static class Routing
...
+
+		internal static bool IsImplicit(string source, bool excludeDefault = false)
 		{
+			if (excludeDefault && source.StartsWith(ImplicitDefaultPrefix, StringComparison.Ordinal))
+				return false;
+
 			return source.StartsWith(ImplicitPrefix, StringComparison.Ordinal);
 		}
-		internal static bool IsImplicit(BindableObject source)
+		internal static bool IsImplicit(BindableObject source, bool excludeDefault = false)
```
dansiegel: as we discussed... this would be very helpful to have made public so that Custom Navigation Services can determine if a node is Implicit or not. Also 100% ok with it being hidden from intellisense.

Xamarin.Forms.Core/Shell/ShellNavigationService.cs
```diff
@@ -0,0 +1,394 @@
...
+			ShellPart = baseShellItem;
+
+			if (baseShellItem is BaseShellItem shellPart)
+				_path = shellPart.Route;
+			else
+				_path = Routing.GetRoute(baseShellItem);
+
+			NavigationParameters = navigationParameters;
+		}
+
```
dansiegel: We need to reconsider how this is setup... makes 100% sense for Forms, but for the Prism POC I was needing to override both the Path and NavigationParameters in part because we use an `IEnumerable<KeyValuePair<string, object>>` instead of a Dictionary. I would say perhaps an Interface would be a good call here...

dansiegel:
@PureWeen in addition to my code comments one thing that really has to be fix here is that if the DataTemplate / Content isn't set in the Shell Content Shell just kind of explodes. Ultimately for Xamarin.Forms it probably should do that, but where it occurs will need to be updated so if someone is using a custom Shell Navigation Service they can move the lookup there. If someone is building a Prism App with Shell for instance all we're going to ever care about is walking the nodes to build the path and then create pages for any ShellContent based on it's route, so we will never expect or use the DataTemplate

dansiegel:
We need to change ApplyParameters to return a task instead of void. This will help to support Async initialization where we actually want to pause Navigation while the View/ViewModel are initialized with their context.

dansiegel:
also @PureWeen we need to figure out why routes are coming back as like TabBar4 instead of IMPL_TabBar4 for the implicit routes.

PureWeen:
@dansiegel rebased this with master if you want to rebase the submodule
https://github.com/dansiegel/Prism-Shell-Stream

rmarinho:
We can close this no @PureWeen  ?

PureWeen:
No

PR #6771, opened 2019-07-03
[X] Bind and assign Command to Event

StephaneDelcroix:
### Description of Change ###

Allow binding, or directly setting, ICommand to any event in Xaml. If the event
is named `Clicked`, a property named `ClickedParameter` can be set, or bound, with
the ICommand parameter.

```xaml
<Button Clicked="{Binding ClickedCommand}" ClickedParameter="{Binding Id}" />
```

or

```xaml
<Button ClickedParameter="Foo">
  <Button.Clicked>
    <local:ClickedCommand />
  <Button.Clicked/>
</Button>
```

### Issues Resolved ### 

- closes #5685
- closes #6047

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [x] Has automated tests 


StephaneDelcroix:
tests are failing on CI, with MissingMethodException :(

PureWeen:
image


[Xamarin.Forms.Xaml.UnitTests.zip](https://github.com/xamarin/Xamarin.Forms/files/3356603/Xamarin.Forms.Xaml.UnitTests.zip)

StephaneDelcroix:
@PureWeen those failure are unrelated to this branch.

brianlagunas:
Please don't do this! This should really be thought out a little more. For starters, this problem has already been solved. Just add the EventToCommand to XF. 

ClickParameter? Please, please do not pollute the XF API with this.

dansiegel:
@StephaneDelcroix I really am not a fan of this implementation at all. It seems to me that it would ultimately be very limiting and not very extensible. What if you instead created a special EventToCommand Markup Extension that just provides you what you need. 

```xml
<Button Clicked="{EventToCommand FooCommand, EventArgsPath='Item'}" />
<Button Clicked="{EventToCommand BarCommand, EventArgsConverter={StaticResource BarConverter}}" />
<Button Clicked="{EventToCommand FooBarCommand, CommandParameter={Binding .}}" />
```

What you're proposing is really too magical and that's coming from me after literally introducing an IAbracadabra… Also it really doesn't fit for people who are coming from UWP or WPF.

dotMorten:
Yet another "let's be different than the other xaml and reinvent the wheel". I really don't like this approach. We already have a command and commandparameter pattern for this scenario. And UWP has a way to bind straight to a method as well. Don't try and reinvent established patterns that cause even more fragmentation in the xaml languages.

StephaneDelcroix:
@brianlagunas @dansiegel @dotMorten: responding to concerns in no particular order:

For the record, binding a command to an event hasn't been invented here (hey x:Bind). the only new part is the parameter property. we can discuss that part, but the proposed solution does not "pollute" the API at all.

We've discussed the EventToCommandBehavior in the past, (or CommandEventTrigger as it should have been named) and this approach is impractical. You're free to use your own, but we can not distribute one with XF.

An `{EventToCommand}` markup is even more magical, I'd prefer not going there

With XF, we're not trying to reinvent the wheel, we're trying to make the XAML language evolve. if you want to stick with the xaml 2006 spec, I encourage you to do so, and we will continue to support you, but that's not a reason to prevent innovation.

I get that this looks like all new (it's not), a big step forward (it's not, except in usability), and a major change (we're not breaking or planning to break any existing pattern), but I'm quite sure you'll get used to it.

love

Stéphane

andreinitescu:
```
Clicked="{Binding ClickedCommand}"
```
It could work, but the issue I see with this is that, if we add this, it's like a Pandora's box. I'm worried it could really go out of hand quickly and soon this is not going to look like XAML anymore.

Sure, you control the changes, but soon there will be requests to do other things and honestly, it won't make sense rejecting other proposals after you added this implementation. 

**Also, by doing this, you fuel the opinions of some vocal people who say they "hate" XAML because "it's too much magic".**

```
ClickedParameter="{Id}"
```
This is a bit too much in my opinion.

I agree with others that instead of transforming how the `Binding` markup extension works, it would be better to use another markup extension, otherwise it could get too confusing. There are opportunities to evolve the old XAML spec, but not like this.

### Suppor x:Bind instead
Instead of modifying `Binding` , why not support the `x:Bind`? In UWP, you can write this:
```
<Button Click="{x:Bind ViewModel.Do}" />
```
where `Do` is a public method on the view-model class. In UWP, this is called [event binding](https://docs.microsoft.com/en-us/windows/uwp/xaml-platform/x-bind-markup-extension#event-binding). The UWP event binding has limitations in how the method signature needs to be.
Xamarin Forms already support x:DataType on any element (UWP supports it only in DataTemplate). It could support calling an ICommand.

**What if someone doesn't use x:Bind?** 
Well, this is a problem :) If `x:Bind` would work perfectly, it should really become a standard using it because of the advantages brought by the strongly typed compilation. In my opinion, making sure x:Bind works perfectly should be a priority.

dotMorten:
> With XF, we're not trying to reinvent the wheel, we're trying to make the XAML language evolve. 

It's not evolving when you go your own way. I assume there's no coordination with the other  xaml owners?

michael-hawker:
To clarify `ClickedParameter` isn't a property on `Button` anymore, eh?  

It's just a magic construct specific to this implementation?

timheuer:
>For the record, binding a command to an event hasn't been invented here (hey x:Bind)
Just to help make it clear though this isn't really 'binding' but a codegen trick.  So traditional binding in the sense as XAML developers know it isn't happening here in x:Bind world.  The main goal of this was to create a more flexible open door and also provide some (better?) debuggability to binding concepts along with type-checking at compile time.  We may or may not have hit some or all of those goals depending on who you ask :-) -- but I did want to clear up that there is less magic happening in x:Bind than in traditional markup extensions like {Binding}

brianlagunas:
Beside's what everyone else is saying, you have to really consider how you are handling parameters.

Are the [EventName]Parameter names dynamically generated? Is that a compiler trick? Are they hardcoded?

If no CommandParameter is provided, are you passing the event args? I hope not, because that would be a big no-no. Do you give the dev a way to choose what object/property off of the event args they want to pass for a particular event? Can they use a converter to convert the event args into something useful in the VM?

Will the ICommand control the enabled state of the control? If so, is there a way to control if it does or doesn't?

andreinitescu:
@StephaneDelcroix 

Is passing the parameter in the example in the description wrong?
```
ClickedParameter="{Id}"
```

Because I see in unit tests you actually use the `Binding` markup extension which makes sense:
```
Happened="{Binding Happened}" HappenedParameter="{Binding Param}"
```

What if I need to pass to the command the event handler parameter?
For example, I want to call the command with the value of `SelectedItemChangedEventArgs.SelectedItem` property. The `SelectedItemChangedEventArgs` instance is parameter of the `ListView.ItemSelected` event handler.
With `EventToCommand` from Prism (but also from MVVMLight) I can do this nicely:

        <ListView.Behaviors>
            <b:EventToCommandBehavior EventName="ItemTapped" 
                                      Command="{Binding ItemTappedCommand}"
                                      EventArgsParameterPath="Item" />
        </ListView.Behaviors>

How can I do the same think once this PR gets merged?

@michael-hawker
> To clarify ClickedParameter isn't a property on Button anymore, eh?
> It's just a magic construct specific to this implementation?

Yes, it's a magical XAML-only property which doesn't actually exist on the `Button` :)

andreinitescu:
I don't think the `Button` example in the description is meaningful. Button has `Command` and `CommandParameter` already.
The most common scenario of event-to-command thing is an event like `ListView.ItemSelected` where you want to pass the selected item.

I guess the following could work?
```
<ListView x:Name="listView"
          ItemSelected="{Binding ShowDetailsCommand}" 
          ItemSelectedParameter="{Binding SelectedItem, {x:Reference listView}}">
```
But it's still a bit heavy :/

dotMorten:
> To clarify ClickedParameter isn't a property on Button anymore, eh?
> It's just a magic construct specific to this implementation?

If that's the case (and from the PR it appears to be), then I find this a really really bad new pattern to introduce. The XAML reads as if ClickedParameter is an actual property on the class, and not some odd magic. Going this route completely changes what XAML is, and it's a very steep slippery slope at that. I think this is the wrong approach. At the very least it should be namespaced like x:ClickedParameter, so it's clearly not part of the class. It also avoids potential conflicts with existing properties.

mrlacey:
I'd like to hear more about why this was/is deemed the best approach to solving the original problem, not just the potential issues with the other EventToCommand proposals.

>>To clarify ClickedParameter isn't a property on Button anymore, eh?
It's just a magic construct specific to this implementation?

>Yes, it's a magical XAML-only property which doesn't actually exist on the Button :)

I'm concerned about this (above quote).
It seems like the change introduces more than just an evolution of XAML and changes the connection between element>attribute and class>property.
It introduces a fundamental breaking of that relationship and requires a change in thinking for developers using XAML with this compiler. No longer can we think of an element's attributes as properties on underlying objects, they now might be something magical instead. In terms of setting a precedent, this could lead the way for all kinds of other magical properties in the future. Yes that might be powerful but I can see that this could easily lead to confusion and challenges in debugging.

How, can/will this work with tooling?
What needs to change to make this practical to use from a tooling perspective?
How/can the potential for this breaking existing tooling that assumes attributes map to properties be avoided? (Or is that even an issue?)
In the XAML view, will there be intellisense for the new command parameters? Will this just add the option for every event that could have such a magical property?
Will anything in intellisense (or elsewhere in the designer) indicate that these new "magical properties" are different?
 
Is there checking that the parameter is of the type expected by the command? Through looking at the code it doesn't seem like this is verified. Could it be? If it doesn't, does any incompatibility only surface at run-time?

Using a (new?) markup extension instead would avoid the need for magical properties and avoid risking changes to things that already exist. Is there a reason why that approach wasn't chosen as a way to add the desired new behavior?

Xamarin.Forms.Build.Tasks/SetPropertiesVisitor.cs
```diff
@@ -992,16 +1000,97 @@ static IEnumerable<Instruction> ConnectEvent(VariableDefinition parent, string l
...
+			if (attached)
+				return false;
+			if (parent.VariableType.GetEvent(ed => ed.Name == localName, out _) == null)
+				return false;
+			if (!context.Variables.TryGetValue(valueNode as IElementNode, out VariableDefinition varValue))
+				return false;
+			//TODO: check if parent is a BP
+			if (varValue.VariableType.InheritsFromOrImplements(module.ImportReference(("Xamarin.Forms.Core", "Xamarin.Forms", "BindingBase"))))
+				return true;
+			//if (varValue.VariableType.InheritsFromOrImplements(module.ImportReference(("netstandard", "System.Windows.Input", "ICommand"))))
```
mrlacey: Should this still be here? please delete, uncomment, or explain why commented out. I think it should be uncommented.

Xamarin.Forms.Build.Tasks/SetPropertiesVisitor.cs
```diff
@@ -992,16 +1000,97 @@ static IEnumerable<Instruction> ConnectEvent(VariableDefinition parent, string l
...
+		{
+			var module = context.Body.Method.Module;
+
+			if (attached)
+				return false;
+			if (parent.VariableType.GetEvent(ed => ed.Name == localName, out _) == null)
+				return false;
+			if (!context.Variables.TryGetValue(valueNode as IElementNode, out VariableDefinition varValue))
+				return false;
+			//TODO: check if parent is a BP
```
mrlacey: Does a follow up task/issue need to be created to ensure this is done.

Xamarin.Forms.Build.Tasks/SetPropertiesVisitor.cs
```diff
@@ -1032,14 +1121,14 @@ static bool CanSetBinding(FieldReference bpRef, INode valueNode, ILContext conte
...
-			if (!(valueNode is IElementNode elementNode))
+			if (!(valueNode is IElementNode))
 				return false;
-
 			if (!context.Variables.TryGetValue(valueNode as IElementNode, out VariableDefinition varValue))
 				return false;
 			var implicitOperator = varValue.VariableType.GetImplicitOperatorTo(module.ImportReference(("Xamarin.Forms.Core","Xamarin.Forms","BindingBase")), module);
 			if (implicitOperator != null)
 				return true;
+			//TODO: check if parent is a BP
```
mrlacey: Is a follow up issue needed here to ensure this is done?

anthcool:
Perspective of someone just starting out with XF, whom isn't a MVVM lib maintainer, employee of MSFT, etc.:

If the changes do not break anything already present, I see no problem adding in another option for those that would possibly like to use something like this at some point.  Those of you that are using existing patterns and practices for your libraries' needs...that's great.  Keep doing that.  No one will stop you.

In regards to the One Ring to rule them all magic talk, aren't attached properties (like something you'd use when working with Grid, etc.) already venturing into the realm of "your a wizard Harry!"?  Properties that don't directly exist within the object you are working with, but attached as needed?  A part of me thinks that, honestly, I've been attempting to learn and use all this for the past year, and there's already so much I've had to learn as it is to build a fleep-flippin' mobile application "the right way", that I'm thinking why not add another slight complexity into the mix?  We've got enough "right ways" for an army already as it is.  Yet, on the other hand, to code is to create, creating is inherently artistic in nature. 
 All these "right ways" or design patterns/best practices are contributions to our overall/collective ability to express ourselves through code to benefit others.  So, for instance, when working with XAML elements' attributes, mentally, I'm already required to know that they could be properties on the object I'm declaratively working with, or they could be event delegate subscriptions, etc.  This would just add another possibility (another way to add a unique brush stroke) that it could be an adaptive way to bind an event to a command declaratively as well.

On the other hand, I not knowing all the design patterns & best practices all the minds of Earth (or at times seemingly Middle Earth) have thought up _so far_, wonder if adding these changes promotes coding practices that would limit a developer's ability to use said best practices for creating adaptive code ready for changes and easy maintainable iterations.  If this change doesn't hinder that ability to provide iterative value to the end customer, then great; but if it does by causing unneeded confusion to work through, then let's think through things a bit more.

Both sides of the coin have good insights, but at the end of the day, I think a good Frosty from Wendy's is in order.  lol

Anywho, good luck y'all.  Appreciate all you guys/gals do.

StephaneDelcroix:
> ClickedParameter="{Id}"

@andreinitescu my mistake, I meant `ClickedParameter="{Binding Id}"`

ravero:
This is very useful and easy to use. People seem emotionally attached with their “specs” and ways of doing this in other platforms. May be very logical to you since you now how it works. Learning MVVM is a pain in the ass and when the user discover the limitations of Command binding because it was really not well support on the majority of components, it’s a huge turn down.

This implementation seems very logical to me. It also doesn’t invalidate another ways of doing this, like the mentioned x:Bind that we don’t really if and when will arrive to forms xaml.

This conservative and purist screams already happened before when forms team implemented the CSS thing, and when it arrived people loved to have it as an option. It’s the same thing here. TOOLS people! The most we have the better we can accomplish our jobs. This one has the bonus of facilitating the use of the main xaml patterns.

keozx:
I don't think this is anything about being "purist", there are lots of arguments here exposed and perfectly valid and some alternatives, too many open questions with no interest in being answered it seems or a will to improve, this is not how it should work, this just halts progress IMHO. 🤷🏻‍♂️

andreinitescu:
I thin it would help to summarize the discussion, it's basically about two approaches:

1. This pull-request suggests:
```
<Button Clicked="{Binding ClickedCommand}" 
        ClickedParameter="{Binding Id}" />
```
We witness here some Stephane's awesome wizardry 🧙🏻‍♂️. The binding markup extension  gets some special powers, the Clicked **event** is bound to a command. Also, the `ClickedParameter` is a "magical" property (Button doesn't have such property) which uses a convention [EventName][Parameter]. 

2. The suggested approach from the comments is to add a new markup extension called `EventToCommand`:
```
<Button Clicked="{EventToCommand ClickedCommand, Parameter={Binding Id}}"  />
```
The `EventToCommand` extension is a regular [markup extension](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/xaml/markup-extensions/creating). It subscribes to the target Clicked event and when the event is raised, it calls `Execute() ` on the given command.

andreinitescu:
@StephaneDelcroix 

Heavily inspired by your implementation and Prism's, I implemented the `EventToCommand` markup extension in a commit [here](https://github.com/andreinitescu/Xamarin.Forms/commit/91b3bfa299b1b37f5a58d5df5c72abf644f61421).  

I implemented the parameters and their behavior from Prism, here's the MainPage.xaml in the Sandbox project as an example:

```
<Button Text="Test"
        Clicked="{EventToCommand ButtonClickedCommand}" />

<!-- Call command with a bound parameter -->
<Button Text="Test with command param"
        Clicked="{EventToCommand ButtonClicked2Command, CommandParameter={Binding SomeParam}}" />

<!-- Call command with the event handler's property -->
<ListView ItemTapped="{EventToCommand ListViewItemTappedCommand, EventArgsParameterPath='Item'}"
            SelectionMode="None">
    <ListView.ItemsSource>
        <x:Array Type="{x:Type x:String}">
            <x:String>mono</x:String>
            <x:String>monodroid</x:String>
            <x:String>monotouch</x:String>
            <x:String>mononucleosis</x:String>
        </x:Array>
    </ListView.ItemsSource>
</ListView>
```

```
class ViewModel
{
	public ICommand ButtonClickedCommand { get; }
	public ICommand ButtonClicked2Command { get; }
	public ICommand ListViewItemTappedCommand { get; }
	public object SomeParam { get; } = 1;

	public ViewModel()
	{
		ButtonClickedCommand = new Command(() => { });
		ButtonClicked2Command = new Command(o => { });
		ListViewItemTappedCommand = new Command<string>(o => { });
	}
}
```
image

Some notes:
- Visiting the XAML nodes had to be modified a tiny bit
- No unit tests for now
- There's a tiny bit more of work required to take a ICommand value too.
- No XAMLC support 
- It works with any EventHandler<T> delegate type. Correct me if I am wrong, but I believe the current PR only works with an event of EventHandler type, it doesn't work with LisView.ItemTapped for example is of `EventHandler<ItemTappedEventArgs>` type

I could do a PR if you think it would be good to.

dotMorten:
I see this PR just moving forward without addressing the concerns several people brought up.

michael-hawker:
Shouldn't PR #6995 which implements this feature as a standard markup extension replace this one?

brianlagunas:
@michael-hawker I hope they don't merge this PR. This PR should be closed immediately. Unless, a discussion on how they will deal with all the concerns that have been brought up can be addressed.

samhouts:
Thanks for the lively discussion, everyone! I think it's clear this problem has no single solution that fits all needs. I like the flexibility of this solution, and I think it adds more than it subtracts from the quality of the product. However, it sounds like many of you already have a solution that works for you, so in the interest of simplicity and flexibility, I think we should close this PR. If we continue to receive signal that this is something we should continue to evaluate, we'll be happy to reconsider this or other solutions at a later time.

ravero:
Well, I think you should reconsider for sure. The fact that there are alternatives people are used to do, doesn't mean that a built in solution isn't desired.

PR #6781, opened 2019-07-04
Clear global routes on unit tests and process global routes in the right place

PureWeen:
### Description of Change ###
- Since we throw an exception now if you register the same route twice we need to clear the global routes in the teardown of the unit tests
- Global Routes were moved to only be processed if the ShellContent was specified as part of the uri but they should just be processed regardless

### Testing Procedure ###
- UI tests cover

### PR Checklist ###

- [x] Has automated tests 


PureWeen:
All green UI Tests

PR #6786, opened 2019-07-05
Use the MoreNavigationController DidShowViewController callback for shell sections

### LINKED ISSUE #6784: [Bug] [Shell] [iOS] ShellItem.CurrentItem not set when selecting shell section aggregated in more tab
### Description

There is this known [issue](https://stackoverflow.com/questions/5803352/didselectviewcontroller-not-getting-called-when-in-more-section) with UITabBarController where didSelectViewController does not trigger for items aggregated in more tab.

`ShellItemRenderer` is relying on [this](https://github.com/xamarin/Xamarin.Forms/blob/78385f9fc1fc56dc88bd98e73bf9c8f2f2d0a90a/Xamarin.Forms.Platform.iOS/Renderers/ShellItemRenderer.cs#L67) to set the `CurrentItem` of the `ShellItem`. For the above reason `ShellItem.CurrentItem` is not set when selecting shell section aggregated in more tab.

#### With debugger

1. Set breakpoint in `ShellItemRenderer.SelectedViewController`
2. Open StoreShell from ControlsGallery
3. Click on Books (aggregated in more tab)
4. Verify breakpoint not getting hit

#### Visually

1. Open StoreShell from ControlsGallery
2. Click on Books (aggregated in more tab)
3. Verify that BooksShell style is not applied (Shell.Background should change to #039BE6) because `CurrentItem` will not get set

### Expected Behavior

`ShellItem.CurrentItem` should be set when selecting shell section aggregated in more tab

### Actual Behavior

`ShellItem.CurrentItem` is not set when selecting shell section aggregated in more tab

### Basic Information

- Platform Target Frameworks:
  - iOS

Dresel:
### Description of Change ###

When `SelectedViewController` is called for `MoreNavigationController`, register for `DidShowViewController` to make it possible to set `ShellItem.CurrentItem` when selecting shell sections aggregated in more tab. Implementation is based on this [SO](https://stackoverflow.com/questions/5803352/didselectviewcontroller-not-getting-called-when-in-more-section) thread.

~~Automated test is skipped because lack of information (is there any guidance available?) of how to setup UI tests for XF.~~

### Issues Resolved ### 

- fixes #6784

### API Changes ###
 
None

### Platforms Affected ### 

- iOS

### Testing Procedure ###

Check #6784.

### PR Checklist ###

- [x] Has automated tests

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6784.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6784.cs
@@ -1,0 +1,91 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using NUnit.Framework;
+using System.Linq;
+using Xamarin.Forms.Core.UITests;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 6784, "ShellItem.CurrentItem is not set when selecting shell section aggregated in more tab", PlatformAffected.iOS)]
+#if UITEST
+	[NUnit.Framework.Category(UITestCategories.Shell)]
+#endif
+	public class Issue6784 : TestShell
+	{
+		protected override void Init()
+		{
+			var shellItem = new ShellItem()
+			{
+				Title = "ShellItem"
+			};
+
+			Items.Add(shellItem);
+
+			AddBottomTab("Tab 1");
+			AddBottomTab("Tab 2");
+			AddBottomTab("Tab 3");
+			AddBottomTab("Tab 4").AutomationId = "Tab 4 Content";
+			var contentPage5 = AddBottomTab("Tab 5");
+			AddBottomTab("Tab 6");
+
+			shellItem.PropertyChanged += (sender, e) =>
+			{
+				if (e.PropertyName == CurrentItemProperty.PropertyName)
+				{
+					if (((ShellItem)sender).CurrentItem.AutomationId == "Tab 5")
+					{
+						contentPage5.Content = new Label()
+						{
+							Text = "Success"
+						};
+					}
+				}
+			};
+		}
+
+#if UITEST && __IOS__
+		[Test]
+		public void CurrentItemIsSetWhenSelectingShellSectionAggregatedInMoreTab()
+		{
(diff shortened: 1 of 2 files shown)


samhouts:
Xamarin.Forms.Core.UnitTests.ShellTests.NavigationBetweenShellContentsPassesQueryString
System.ArgumentException : Duplicated Route: "details"

PureWeen:
@Dresel can you rebase against master? That should fix the unit test error

Dresel:
Rebase done

Dresel:
Currently busy but I'll be able to tackle this latest end of this month

woutervanoorschot:
@PureWeen I think that this is the correct behaviour. At least on iOS it is the default behaviour that a tab stays the same (also within the more page) when navigated to and from. If a page has been opened in the more tab, navigating to another tab and returning to it should return you to the opened page. The second tap returns you to the root of the tab (in the more tab: the list of pages).

This works the same as in other tabs, if you navigate within a tab it stays the same after navigating between tabs. When tapping on a tab which is selected, navigationpage within the tab goes to the root page.

Anyway, very nice that someone is looking after this, as it seems to fix the very long standing issue of the more tab in xamarin.

Dresel:
Sorry, took me longer than expected.

> 
> 
> @Dresel I ran into a slight issue while testing this
> 
> it looks like after you select the item from the more list the first time then it won't reopen the more list when navigating back to it unless you click it twice.

This can be "corrected" by calling PopToRootViewController on the `MoreNavigationController`.

For consistency we should also set `CustomizableViewControllers` to an empty array to avoid the "Edit" button on top of the `MoreNavigationController`:

> This property controls which items in the tab bar can be rearranged by the user. When the user taps the More item on the tab bar view, a custom interface appears displaying any items that did not fit on the main tab bar. This interface also contains an Edit button that allows the user to rearrange the items. Only the items whose associated view controllers are in this array can be rearranged from this interface. If the array is empty or the value of this property is nil, the tab bar does not allow any items to be rearranged.



Let me know if you want this in a separate PR or if I can combine it with this one.

I have rebased and pushed latest changes, let me know what you think.

I'm now working on a UI test case for that, it's a pity that this does not work with remote debugger on windows...

woutervanoorschot:
I think you should combine it in this PR, as it creates inconsistent behavior across platforms if the edit button is visible (as you said already). Just my 2 cents

Great work!

Dresel:
UI test added.

Dresel:
Could not reproduce the failing UI tests (SDK 12.4 build, iPhone X 11.4 simulator).

woutervanoorschot:
@PureWeen any chance this can be included in Xamarin Forms 4.5?

Issue #6492, opened 2019-06-11
[Bug] x:DataType={x:Type SomeType} breaks Compiled binding


viktorszekeress:
### Steps to Reproduce

1. Set `x:DataType` to `"{x:Type SomeType}"` instead of `"SomeType"`
2. Try to bind to some non-existing property

### Expected Behavior
Compiler should report an error that such property doesn't exist, just like it would with `"SomeType"`.

### Actual Behavior
No errors by compiler.

### Basic Information
While this is not critical bug, it annoys me, because it brings inconsistency. As we know, we can use both expressions (`"{x:Type SomeType}"` or `"SomeType"`) in Styles or DataTemplates. I prefer `"{x:Type SomeType}"` expression because text is rendered in different color so it pops up visually.

StephaneDelcroix:
x:DataType expects a string literal, not a `{x:Type}` markup extension. This should throw an exception, I wonder why it does not. We should investigate that

davidbritch:
This markup was valid in Forms until Forms 4.1.

StephaneDelcroix:
`x:DataType` with `{x:Type}` markup extension never worked, and was silently ignored (bindings weren't compiled).

This is why, with #5335 (fixing #5330), it now throws at compilation time. This is not a bug, this is the intended behavior.

tranb3r:
Wait wait wait... @StephaneDelcroix 
What about x:DataType={x:Null} now throwing an exception, while being totally valid ?
This is a really annoying regression !

StephaneDelcroix:
@tranb3r that's a good point. please open a new issue while I start working on a fix

davidbritch:
@StephaneDelcroix It's already logged at https://github.com/xamarin/Xamarin.Forms/issues/6648 (where, despite the title, it's about `x:Null`).

StephaneDelcroix:
thx @davidbritch. missed that one, as I was off last week

tranb3r:
> btw, assigning x:Null never was a feature, but a fortuitous side effect of nulling the datatype in unrecognized cases

Maybe it's not a feature, but it's the only workaround for several bugs in Xamarin Forms.
For example this one: https://github.com/xamarin/Xamarin.Forms/issues/5240

viktorszekeress:
[...]

PR #6843, opened 2019-07-10
[X] allow x:Type markup as x:DataType value

StephaneDelcroix:
### Description of Change ###

see discussion on #6492 and related issues

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Build.Tasks/SetPropertiesVisitor.cs
+++ b/Xamarin.Forms.Build.Tasks/SetPropertiesVisitor.cs
@@ -384,11 +384,23 @@ namespace Xamarin.Forms.Build.Tasks
 
 			if (   dataTypeNode is ElementNode enode
 				&& enode.XmlType.NamespaceUri == XamlParser.X2009Uri
-				&& enode.XmlType.Name == "NullExtension")
+				&& enode.XmlType.Name == nameof(Xamarin.Forms.Xaml.NullExtension))
 				yield break;
 
-			if (!((dataTypeNode as ValueNode)?.Value is string dataType))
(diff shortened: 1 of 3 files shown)


KSemenenko:
what if my type is generic?

if my type is:
``` cs 
SomeType<T> 
```
or when:
``` cs
SomeType<MyAnotherType>
```
How then can it be used?

@samhouts @StephaneDelcroix

StephaneDelcroix:
@KSemenenko this is an excellent question, but unrelated to this change. 

Generic types, in xaml, are expressed like this `local:SomeType(local:MyAnotherType)`, unless you're using it as an xml element, in that case you have to use x:TypeArguments (see some example of both syntax in https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Xaml.UnitTests/GenericsTests.xaml)

At this time, we don't have test for passing generic parameters to x:Type markup extension or x:DataType, so I can not be affirmative about the support for them.

Feel free to investigate, and open an enhancement request.

about `SomeType<T>`, we don't support open generic types

viktorszekeress:
Great to see this feature will be included in next releases. We are desperately waiting for this, as we are using Resharper which only works with `x:Type`, otherwise it will not resolve the type and will not offer properties when creating a Binding.

Two questions:
1. When we can expect this change in stable release?
2. Can you use same algorithm for `x:TypeArguments`, to have it consistent and matching with WPF and UWP ?

PR #6860, opened 2019-07-12
Fix binding to nullable types

### LINKED ISSUE #6280: [Bug] TwoWay Data Binding on Entry when using a nullable property
### Steps to Reproduce

1.  Create a new XF Project
2.  Create a VideModel that implements INotifyPropertyChanged
3.  Inside the VM, create a property that is nullable:

 
```
        private int? _stock_quantity;
        public int? stock_quantity
        {
             get => _stock_quantity;
             set
            {
                _stock_quantity = value;
                Change(nameof(stock_quantity));
            }
        }
```

4. Create an Entry whose Text value is bound to that property:

  ` <Entry Text="{Binding stock_quantity, Mode=TwoWay}" Placeholder="Units in stock"   />`

### Expected Behavior

It should update the VM but seems that the TwoWay mode is ignored.  

### Actual Behavior

It doesn't update the view model and the breakpoint set on the setter is not hit. 

### Basic Information

- Version with issue:  4.0.0.425677
- IDE: VS 2017 

andreinitescu:
### Description of Change ###

Fixes binding to nullable types. 

### Issues Resolved ### 

- fixes #6280

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Xaml.UnitTests/Issues/Issue6280.xaml
+++ b/Xamarin.Forms.Xaml.UnitTests/Issues/Issue6280.xaml
@@ -1,0 +1,13 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             xmlns:d="http://xamarin.com/schemas/2014/forms/design"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             mc:Ignorable="d"
+             xmlns:local="clr-namespace:Xamarin.Forms.Xaml.UnitTests"
+             x:Class="Xamarin.Forms.Xaml.UnitTests.Issue6280"
+             x:DataType="local:Issue6280ViewModel">
+    <Entry x:Name="_entry"
+           Text="{Binding NullableInt}"
+           VerticalOptions="Start" />
+</ContentPage>
--- a/Xamarin.Forms.Core/BindingExpression.cs
+++ b/Xamarin.Forms.Core/BindingExpression.cs
@@ -449,6 +449,8 @@ namespace Xamarin.Forms
 					return false;
 				}
 
+				convertTo = Nullable.GetUnderlyingType(convertTo) ?? convertTo;
+
 				value = Convert.ChangeType(value, convertTo, CultureInfo.InvariantCulture);
 				return true;
 			}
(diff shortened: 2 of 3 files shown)


andreinitescu:
Would it be possible to get this fixed please in the next 4.1 release? Thank you

Xamarin.Forms.Xaml.UnitTests/Xamarin.Forms.Xaml.UnitTests.csproj
```diff
@@ -30,4 +30,9 @@
       <Link>BaseTestFixture.cs</Link>
     </Compile>
   </ItemGroup>
+  <ItemGroup>
```
StephaneDelcroix: files are added by wildcards

Xamarin.Forms.Xaml.UnitTests/Issues/Issue6280.xaml
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             xmlns:d="http://xamarin.com/schemas/2014/forms/design"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             mc:Ignorable="d"
+             x:Class="Xamarin.Forms.Xaml.UnitTests.Issues.Issue6280">
```
StephaneDelcroix: add a x:DataType argument so we get a test on compiled bindings too

StephaneDelcroix:
> Since it's a fix, would it be possible to get this pushed please to the next 4.1.x release? Thank you!
> I rebased it to 4.1.0

no. 4.1 is released and we only eventually will release -sr for it. but don't worry, 4.2 will follow soon enough.

I understand that for you it looks like "just one little thing", but the more "one little thing" we release as -sr, the more we delay 4.2, and the global effect, on average, is a general slow down.

andreinitescu:
All done! 
I did run the tests before and they both passed... But I forgot about `x:DataType`, thanks for mentioning it!

andreinitescu:
> I understand that for you it looks like "just one little thing", but the more "one little thing" we release as -sr, the more we delay 4.2, and the global effect, on average, is a general slow down.

It is really a small change easy to review, but with a pretty important impact, that's why I thought it should go out asap. Imagine having to write converters for this. Which is actually I've been doing too sadly. In WPF/UWP it just works. 
But I totally understand your opinion, you guys know better and I can miss other things.

AwsomeCode:
I have nullable decimal with converter. We did too many tricks to work it properly with Entry. Hope this will reduce that complicated tricks.

PR #6897, opened 2019-07-17
[Android] Fix previewer exceptions with TabbedPage

paymicro:
### Issues Resolved ### 

- VS [bug 939527](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/939527)

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

the Previewer will show TabbedPage

Before
Screenshot_1

After
Screenshot_2

### Testing Procedure ###

- Create a Xamarin Forms project, select the `Tabbed` template 
- Open file `MainPage.xaml` (Views->MainPage.xaml) 
- Click "Design" to open previewer
- Select platform as `Android`

### PR Checklist ###

- [ ] Has automated tests 

> VS bug [#948950](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/948950)

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs
@@ -233,7 +233,8 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 			base.OnElementChanged(e);
 
 			var activity = Context.GetActivity();
-			var isDesigner = activity == null && Context.IsDesignerContext();
+			var isDesigner = Context.IsDesignerContext();
+			var themeContext = isDesigner ? Context : activity;
 
 			if (e.OldElement != null)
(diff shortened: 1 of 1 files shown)


paymicro:
@PureWeen looks alright. Thanks.
I have only thoughts on whether to unwrap the designer context if it is in `ContextWrapper` like as application activity.

PureWeen:
@paymicro like if this themeContext here should be unwrapped or not?

https://github.com/xamarin/Xamarin.Forms/pull/6897/files#diff-7feeca82b01e77527e9263a06421f443R237

paymicro:
I think that adding unwrapping is not necessary, because the context of a designer is not wrapped.
Everything looks good.

PureWeen:
@paymicro agreed. I did play with it it for a bit figuring out if it should be unwrapped or not. It only matters if you apply material. I tested this by applying material and making sure everything still worked and it did. If we get to a point of having material tabs then we might need to reconsider but for the time being I think just making it work with previewer without changing behavior is a good way to go.

PR #6923, opened 2019-07-18
[Android] Add check for multiple dispose calls on scrollview, page,  and platform

PureWeen:
### Description of Change ###

When the previewer unloads the app process it calls dispose on every single Java.Lang.Object in memory. Meaning dispose will be called on children elements multiple times. We've occasionally been seeing an exception in the dispose call on ScrollViewContainer because dispose will be called directly on ScrollViewContainer and it will be called by the ScrollViewRenderer. If the timing is just wrong it'll throw an exception from ScrollViewContainer since the dispose happens twice

### Platforms Affected ### 
- Android

### Testing Procedure ###
- Open a forms page with a scrollview, close it, open it, and then close it. Do that a few times and just makes sure it all works
- Run an android app with a page that has a scrollview. Pop the page off the stack and just make sure the dispose calls fro scrollview and scrollviewcontainer seem to work

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/Platform.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/Platform.cs
@@ -432,9 +432,11 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 
 			protected override void Dispose(bool disposing)
 			{
-				if (disposing && !_disposed)
+				if (_disposed)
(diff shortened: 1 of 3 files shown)


jfversluis:
Failing test is just an App Center timeout, actual run was all green

PR #6935, opened 2019-07-19
[Build] Share package version with cake

rmarinho:
### Description of Change ###

Writes a file with the version so cake script can use it.

### Testing Procedure ###
Run pack see if provides the metadata on package version

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -24,6 +24,7 @@ PowerShell:
 #addin "nuget:?package=Cake.Android.SdkManager&version=3.0.2"
 #addin "nuget:?package=Cake.Boots&version=1.0.0.291"
 
+#addin "nuget:?package=Cake.FileHelpers&version=3.2.0"
 //////////////////////////////////////////////////////////////////////
 // TOOLS
 //////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


build.cake
```diff
@@ -60,6 +61,10 @@ Task("NuGetPack")
 Task("_NuGetPack")
     .Does(() =>
     {
+        var nugetVersionFile = 
+            GetFiles("XamarinFormsVersionFile.txt");
+        var nugetversion = FileReadText(nugetVersionFile.First());
```
PureWeen: Is there a way it could read this from the output of MSBuild? What I'm thinking is
- I don't think the target that does WriteLinesToFile should always run. That it should itself be a target so cake can just invoke it if it wants the version. I don't think it should write to that file has a side effect of every build it should only do it if the intention is to use it.
- With that I'm wondering if there's a way to just have an MSBuild target like
```cmd
msbuild Xamarin.Forms.Core.csproj /t:nugetversion
```

That could just print out a message to the console out that's literally just the version that cake could read?
- If that's not possible though then I still think just having a specific target for outputting to the file would be the way to go.

mcumming:
Considering the repository is already using SemVer and the tags look reasonable, you could use GitVersion to populate the version for Cake, MSBuild and Nuget.

build.cake
rmarinho: I think this kinda helps on release to have a reference to the package version so we can generate imports , my idea is read this so i can know what version to update an app by pointing to artifact folder in a release where the package lives. 
But i do think we can have just a target for this .

rmarinho:
@mcumming we actually already use GitInfo to do that

Issue #6638, opened 2019-06-24
[Bug] iOS - ListView cell height


MarMarIV:
### Description
Hi, i have a problem with cell height on iOS with using DataTemplateSelector on ListView. Sometimes happend, the cell height is not calculated exactly (in screenshot)

### Steps to Reproduce

1. DataTemplateSelector for use diferent layout

### Basic Information

- Version with issue: 3.4.0.1029999
- IDE: VS2019 16.1.2 
- Platform Target Frameworks: 
  - iOS:  12.2
- Affected Devices:
  iPhone SE

### Screenshots

![8104fd3b-c4af-4cc8-bd04-963779459d9a](https://user-images.githubusercontent.com/10964247/59990836-132ae980-9645-11e9-95e3-196dd02c87f6.png)

KSemenenko:
I have the same bug too.

samhouts:
@MarMarIV @KSemenenko  Can you please attach a small project that demonstrates this issue? Thanks!

KSemenenko:
I'm already working on a project that will show this bug

KSemenenko:
@MarMarIV Maybe you can also make a test project?

samhouts:
Thanks! We'll look for it.

MarMarIV:
I try to reprodece it on empty project (+ ListView wit DateTemplateSelector), but i cannot

KSemenenko:
@MarMarIV I have not finished my test, but I think this is due to the fact that there are several DataTemplate of different heights and Entry or Editr inside.

samhouts:
Any luck with a reproduction? Thanks!

KSemenenko:
Sorry for so long, here is an example.
The problem is that when values change through binding, the cell size does not change. Also It can be related to #5115
[GroupHeaderTemplateLayoutErroriOS.zip](https://github.com/xamarin/Xamarin.Forms/files/3446940/GroupHeaderTemplateLayoutErroriOS.zip)
@samhouts

KSemenenko:
This seems to require calling ForceUpdateSize.
Or InvalidateMeasure after changing the BindingContext.

giuseppenovielli:
Yes, ListView with DataTemplateSelector and also Grouped Height the cell height is not calculated exactly. 
There are any Workaround?

PR #6936, opened 2019-07-19
[C] force update size on View updated

StephaneDelcroix:
### Description of Change ###

when a ViewCell View property is updated, the actual rendered cell is blank

### Issues Resolved ### 

fixes HotReload

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 


felipemomm:
Just wondering, does it happen only on HotReload? If so, won't it make the performance worse on release mode? Perhaps it should call this method only on DEBUG?

StephaneDelcroix:
@felipemomm it can be reproduced by switching ViewCell.View, but no one ever done that before. the perf hit only applies when actually reassigning the View, and we just proved that no one does that, so it should be fine

KSemenenko:
There is a bug #6638 that can be related, it occurs in the ListView on iOS when the ReculceElement reuse the cell, changing the BindingContext. The BindingContext is changing, but the cell size is not. @StephaneDelcroix

PR #6944, opened 2019-07-22
[iOS] Fix usage conflict of Title and AttributedTitle on ButtonRenderer

rmarinho:
### Description of Change ###

PR #5167 introduced a bug with iOS button text where it was setting a `AttributtedTitle`  and the `ButtonLayoutManager.cs ` was setting just the control `Title` when the text was changed.

### API Changes ###

- iOS

When changing the Text button it should work, you should see the new text

### Testing Procedure ###
There are several tests on ControlGallery that should pass with this fix, iOS should be green.
Can tests for example with test case  `Bugzilla46458` after clicking the button it should read `Clicked`.
 
### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls/GalleryPages/CharacterSpacingGallery.xaml
+++ b/Xamarin.Forms.Controls/GalleryPages/CharacterSpacingGallery.xaml
@@ -2,7 +2,7 @@
 
 <ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
              xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
-             x:Class="Xamarin.Forms.Controls.GalleryPages.CharacterSpacingGallery" Visual="Material">
+             x:Class="Xamarin.Forms.Controls.GalleryPages.CharacterSpacingGallery">
 
     <StackLayout>
         <Label>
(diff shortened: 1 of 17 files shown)


PureWeen:
it looks like there's an infinite ping pong happening

If you run the label gallery the bounds on the label just keeps changing back and forth

If you change the LabelRenderer calls back to
```
Control.AttributedText = newAttributedText;
```

it no longer does this

PureWeen:
@rmarinho take a look at the tests and let me know what you think. It's only the last set of tests that are failing and when I run them locally they run mostly fine

- if I run the material set before the picker ones the completed test fails with an odd NRE
- If I run each of the sets that failed in appcenter locally the whole set passes. Though I do have to try to kick them off a few times before they actually run.
- if you check appcenter the iOS device is just a black screen for the whole last part of the run which seems odd
- I saw this same behavior when running locally against 4.1.0 as well so it didn't seem special to this PR

I'll play with it a bit more tomorrow see if I can get them running better locally

PureWeen:
@rmarinho all these same tests are failing on master 

https://appcenter.ms/orgs/xtc-Xamarin-Forms/apps/XamControl/test/series/master/runs/12859e81-f366-4311-bdbd-69edb73728a8

But with this PR a lot less are failing :-) so I'd say this one is good to merge

PureWeen:
@samhouts the iOS tests that are failing are failing on other master builds as well so we're find as far as iOS is concerned so now just need to wait for Android tests to finish

rmarinho:
Just 1 test failing on iOS not related

Issue #3844, opened 2018-09-19
[Enhancement] Add a Clear button to DatePicker and TimePicker


davidortinau:
## Rationale

See https://xamarin.uservoice.com/forums/144858/suggestions/16997776

## Implementation

TBD

## API

TBD

brinawebb:
As a part of this you will also need to support a null value, DateTime?, correct?

opcodewriter:
@davidortinau I don't understand, why did you set the issue on UserVoice to COMPLETED.
It's misleading, it's not completed, why not set the issue to PLANNED?
The UserVoice on that portal, xamarin.uservoice.com is completely broken. It looks like "planned" was overwritten with "completed" status. If I select "planned" it shows "completed" features. Sigh...

davidortinau:
@opcodewriter sorry if I got your hopes up. 

I marked them completed because there is a transition happening with our use of UserVoice and the only way to grant you back your votes was to mark it completed. I tried to make the actual status clear in the comment I added. We won't be tracking those items moving forward on UserVoice.

samhouts:
related to #3728

jfversluis:
I am looking into #3728, when that is implemented; is this really added value?
What would a button look like on the different platforms? I think if we add a button, users will want to be able to style it, etc.

I would propose to just implement the nullable date (or time for that matter) and let users implement their own clearing functionality/buttons/controls.

Thoughts?

cmcclure-penlink:
Either that are create an event when the cancel button is hit. Should come that way by default.

andreinitescu:
I wanted initially to add a comment to #6947 but I want to keep discussion separated.

I think it's confusing to have a Clear button on the picker dialog. The picker dialog itself always shows a date, there's no way not have it show a date selected, like how for example selection works in ListView (SelectedItem can be null).

So I think a Clear button would make more sense on the picker text field and it makes sense to work with the new `SelectedDate` property on the DatePicker introduced by #6947:

![image](https://user-images.githubusercontent.com/743918/61663328-08fa1a80-acd9-11e9-80cb-1857dace5b26.png)

If we agree it makes sense to have a Close button this way, I can work on a separate PR to have the date picker show a close button to set the new DatePicker `SelectedDate` property to null. It would require however first getting merged #6947 and #5277 (which btw, I don't know why it was postponed again to vNext+1)

jcmanke:
[...]

PR #6947, opened 2019-07-22
Nullable DatePicker 

### LINKED ISSUE #3728: DatePicker NULL value
Hi

Please allow to set a null value to the DatePicker..

Currently it forces a value but most of the times we want to initially show it with a placeholder, e.g. [Select], Birthday, etc

Thanks,
Jassim

andreinitescu:
### Description of Change ###

Add nullable DateTime property to DatePicker.

Please see the [comment ](https://github.com/xamarin/Xamarin.Forms/pull/6947#issuecomment-514321150 )below for detailed description of the behavior

### Issues Resolved ### 

- fixes #3728 

### API Changes ###

Added:
 - DateTime? DatePicker.SelectedDate { get; set; } //Bindable Property
 - event DatePicker.SelectedDateChanged;

Obsoleted:
 - object DatePicker.Date { get; set; }
 - event DatePicker.DateSelected; 

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla60001.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Bugzilla60001.cs
@@ -23,7 +23,9 @@ namespace Xamarin.Forms.Controls.Issues
 		{
 			StackLayout layout = new StackLayout() { Orientation = StackOrientation.Vertical };
 			DatePicker picker = new DatePicker();
+#pragma warning disable 0618 // Retain until Date is removed
 			picker.Date = new DateTime(2017, 10, 7, 0, 0, 0, DateTimeKind.Utc);
+#pragma warning restore
 			Label label = new Label() { Text = "On Droid this will show as 10/7/2017, on UWP it will show as 10/06/2017.  Local TimeZone for this test was EDT.", LineBreakMode = LineBreakMode.WordWrap };
 			layout.Children.Add(picker);
 			layout.Children.Add(label);
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/DefaultColorToggleTest.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/DefaultColorToggleTest.cs
@@ -344,9 +344,10 @@ namespace Xamarin.Forms.Controls.Issues
 
 		static ContentPage DatePickerPage()
 		{
+#pragma warning disable 0618 // Retain until Date is removed
 			var pickerInit = new DatePicker { Date = new DateTime(1978, 12, 24), TextColor = Color.Red };
 			var pickerColorDefaultToggle = new DatePicker { Date = new DateTime(1978, 12, 24) };
-
+#pragma warning restore
 			var defaultText = "Should have default color text";
 			var pickerColorLabel = new Label { Text = defaultText };
 
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2987.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2987.cs
@@ -16,11 +16,16 @@ namespace Xamarin.Forms.Controls.Issues
 			var datePicker = new DatePicker { AutomationId = "datePicker" };
 			datePicker.MinimumDate = new DateTime (2015, 1, 1);
 			datePicker.MaximumDate = new DateTime (2015, 6, 1);
+#pragma warning disable 0618 // Retain until Date is removed
 			datePicker.Date = DateTime.Now;
+#pragma warning restore
 			datePicker.Format = "MMM dd, yyyy";
+
+#pragma warning disable 0618 // Retain until Date is removed
 			datePicker.DateSelected += (object sender, DateChangedEventArgs e) => {
 				Debug.WriteLine ("Date changed");
 			};
+#pragma warning restore
 
 			Padding = Device.RuntimePlatform == Device.iOS ? new Thickness(10, 20, 10, 5) : new Thickness(10, 0, 10, 5);
 
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4187.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4187.cs
@@ -102,8 +102,9 @@ namespace Xamarin.Forms.Controls.Issues
 			picker.SetBinding(Picker.ItemsSourceProperty, new Binding(nameof(Issue4187Model.PickerSource)));
 
 			var datePicker = new DatePicker();
+#pragma warning disable 0618 // Retain until Date is removed
 			datePicker.SetBinding(DatePicker.DateProperty, new Binding(nameof(Issue4187Model.Date)));
-
+#pragma warning restore
 			var entry = new Entry();
 			entry.SetBinding(Entry.TextProperty, new Binding(nameof(Issue4187Model.Text)));
 
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6947.xaml
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6947.xaml
@@ -1,0 +1,15 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             xmlns:d="http://xamarin.com/schemas/2014/forms/design"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             mc:Ignorable="d"
+             x:Class="Xamarin.Forms.Controls.Issues.Issue6947"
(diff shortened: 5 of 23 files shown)


andreinitescu:
I stumbled on an issue and I need feedback.

I can't see how adding just the SelectedDate property is enough. There's now two properties, Date (obsoleted) and SelectedDate. The renderers use a text field to display the selected date, but I can't make it to work with both properties, because when SelectedDate is null, it should not show anything.
In the same time, for backward compatibility, it should use the obsoleted Date.

Options I see for now:

1. One obvious option, but maybe the least preferred (because _It could break existing apps_™), is to just make the existing Date property nullable. 

2. Add a new separate NullableDatePicker class

Looking forward for your input on this, thanks

PureWeen:
Looks good. This is how I figured it was going to have to be done.

andreinitescu:
@PureWeen 

> This is how I figured it was going to have to be done.

Umm, not sure you saw, but I need some input because I am not sure how to proceed

PureWeen:
@andreinitescu 

oop I see what you're asking now

I think you can keep both properties without changing existing behavior. But I might be missing a nuanced detail you've come across. 

> I can't make it to work with both properties,

If we want to keep both properties synced up I would say

If the new property == null then the old property == todays date (which is the current default) 

The main scenario I see that causing trouble would be for cases where someone has a derived control with a custom renderer but I think those controls will just have to catch up with behavior. 

Those controls will still work if the users only use the Date property so it's not breaking anyone.

andreinitescu:
> If we want to keep both properties synced up I would say
> If the new property == null then the old property == todays date (which is the current default)

That part is already done. The issue is with what I described above:

> The renderers use a text field to display the selected date, but I can't make it to work with both properties, because when the `SelectedDate` is null, it should not show anything.
In the same time, for backward compatibility, it should continue show the obsoleted Date.

image

More exactly, it's this line: https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/Renderers/DatePickerRenderer.cs#L173

Make sense?

PureWeen:
Can we just set the new field to todays date by default so the behavior is the same? Then users have to manually set the new field to null in order for it to be null.

KSemenenko:
For me, the most important question is how to clear the date. We added buttons to clear the value.
In the case when there is already a selected date, and you want to clear it.
If you are interested, I have a ready implementation of these buttons for iOS and Android.
P.S. What about TimePicker?

PureWeen:
@KSemenenko clear the bindable property specifically? or just clean the contents of the field?

KSemenenko:
@PureWeen I meant the control that clears the field, well, the value of the selected date.

DatePicker:
TTP

TimePicker:
TTP

jcmanke:
In my custom nullable pickers, we have have 3 buttons: Clear (set to null), Cancel (close the dialog without changing the BindableProperty's value) and OK/Done (set the BindableProperty value to the current dialog value).

Screenshot_1563820576

ios_nullable_datepicker

KSemenenko:
@jcmanke Nice!

andreinitescu:
@PureWeen 

> Can we just set the new field to todays date by default so the behavior is the same? Then users have to manually set the new field to null in order for it to be null.

Null would have been ideal, but today's date is a good compromise I guess :) Thanks

@KSemenenko @jcmanke To make it simpler, I suggest we keep discussion for the Clear button in the dedicated ticket #3844 , please see my comment I just added. I'm interested to hear your thoughts, thanks.

adammeaney:
I think the easiest world is to obsolete the old DatePicker and to have a new NullableDatePicker class.

Trying to keep this from being a breaking change in some way just isn't going to work. 

Either it starts null and we break the functionality that it starts with today without explicitly saying it, or it starts with today, has no good way of clearing it, and clearing the value programmatically leaves us in a weird conflicting state between Date and SelectedDate.

PureWeen:
> clearing the value programmatically leaves us in a weird conflicting state between Date and SelectedDate.

so? If you use both Date and SelectedDate then you deserve to be conflicted :-) 

Date is going to be deprecated and hidden from the APIs so new users will just use SelectedDate and it'll be fine. Old users code will still work and if they want nullable behavior they will switch to SelectedDate. It's definitely not the most ideal but I feel like adding a second control is less ideal. 

As a developer if you start wanting to use CLEAR then you'll start using SelectedDate and forget Date ever existed.

adammeaney:
I agree that the programmers using it shouldn't be using both, they should convert once they upgrade.

Its things like the renderer responding to the events of both of them like Andrei said that wouldn't work.

If Date = DateTime.Now and SelectedDate = null, what displays on the screen? Whichever one happened to fire last? Do they update each other? 

How do you keep old code working with the new SelectedDate allowing it to be null? Is starting with DateTIme.Now or UtcNow or whatever the current default is even still right? Should it default to null, breaking all old code in general that didn't specify a starting date?

There are a lot of questions and concerns to backwards compatibility and best design going forward, and it might make more sense to separate the controls to avoid most of them.

PureWeen:
> If Date = DateTime.Now and SelectedDate = null, what displays on the screen? Whichever one happened to fire last? 

Nothing displays because they set SelectedDate last. Why is this a hang up? The only case where these values will be set twice is if someone using a 3rd party library and that library implements our datepickerrenderer. At that point the library author will release an update that takes advantage of the new property. Until then everyone just uses Date without it mattering. As long as it's not an ABI break then all is well.

> Do they update each other?

Yea. Why is this a point of complication? The code for updating both isn't that bad and it's only relevant for 3rd party plugins. Users that update to new versions of forms won't be effected by this at all. If they use Date they'll still use Date and everything works. If they get excited to set the thing to null then they switch to the new one and now they are using the new one. 

> How do you keep old code working with the new SelectedDate allowing it to be null?

I think I addressed this in my last comment. If not can you give me more specific scenarios? 

>  Is starting with DateTIme.Now or UtcNow or whatever the current default is even still right? 

Yes? no? maybe? Assuming we have any idea what people want the default value to be is impossible. The fact that it's todays date is just a choice that maybe applies to most cases. Most people using that field are going to be  binding to a VM and in the VM they will set it to whatever default they are wanting. Todays date? null? are just arbitrary

> Should it default to null, breaking all old code in general that didn't specify a starting date?

How is it breaking all old code? The default for Date/SelectedDate will just be todays date and everything will work as normal. 

This is all like the ImageSource changes we just made. If you are using the old field then it updates the new field. If you are using the new field then the old field will use the new field if the type is compatible.

andreinitescu:
The way I see it is like this:
The `SelectedDate` refers to the value displayed in the text field, which is the result of the selection using the dialog picker (hence its name). It's not meant to directly control the value selected in the picker dialog. More on this below.

Also, like Shane said, once `SelectedDate` is released, since `Date` will be obsolete, we should use and care only about `SelectedDate`. More thoughts:

- To keep the backward compatibility, the implementation will have `Date` and `SelectedDate` sync'ed, but this is an implementation detail (because, again, you should not care about using both of them).

- Setting `SelectedDate=null` will show the text field with no value (btw, maybe we can think of adding later a `Placeholder` property). The `Date` remains unchanged, it keeps its current value. It's just like today, you can have a binding `<DatePicker Date={Binding MyDate}>` and `MyDate` is null, and it doesn't throw error, because the DatePicker implementation has coercing and ignores the incoming null value.

- As an implementation detail, the picker dialog will continue to use the `Date` as the selected date. This is needed in order to keep the behavior unchanged for the existing apps. Since `Date` is sync'ed with `SelectedDate`, this means picker dialog works with `SelectedDate`, even if it's null. If we'll ever really remove the `Date` property, we'll just make it internal or turn into a simple field, 

- Yes, I agree it's an effort to make these changes instead of just adding a NullableDatePicker, like I also suggested above.

Today I did more work and I found a new concern:

The native date pickers on UWP and Mac OS display as a text field, but they don't have the ability to clear the selected value. (UWP actually has a `DateTime? SelectedDate` nullable property, but it requires a newer Windows 10 version than what the XF UWP backend currently targets).
    What this means is, setting `SelectedDate` to null will not clear the text field on these platforms, at least not by this PR. We can maybe look at some workaround later, since Android and iOS are the main goals.  @PureWeen This is important for the approach to take, please let me know if this is acceptable.

This also raises an important aspect. A nullable `SelectedDate` assumes the DatePicker has a way not to display a date (or show a placeholder at least). On iOS and Android this is possible because the renderers create a text field on the fly and show picker dialog when the text field is tapped. But the native pickers don't have this ability. Hmm, which makes me thing again, shouldn't be better to have a separated NullableDatePicker? The way this could work is it would wrap the current DatePicker, it would switch its content between a read-only Entry and a DatePicker. 

Phew!

andreinitescu:
This should be ready to go

jsuarezruiz:
@andreinitescu Can you review the conflict?.

After following the thread, the summary (status) is:

- To achieve the goal (Nullable DatepPicker), the `SelectedDate` property has been added.
- `Date` will become obsolete. To keep the backward compatibility both properties will be synchronized. 
- At the moment only implemented in Android and iOS.
- In a future, other PRs will add options like include a [Clear Button](https://github.com/xamarin/Xamarin.Forms/issues/3844) in the DatePicker.

groege:
How about we add a boolean property like "Clearable" 
The default value of "Clearable" is false, which is basically the current non null able date and the clear button is not shown - if the date is null the current date will be shown
If "Clearable" is set to true, the clear button is available and if the date date is null the field will be empty

groege:
Any news about this?

tbrabec:
Is this going to be merged anytime soon? I really need it.

groege:
Is there any info on why this isn't progressing at all?
Please use more resources to merge pull requests

aismaniotto:
Also waiting for this.

samhouts:
closing in favor of #10175

PR #6970, opened 2019-07-25
Remove header padding and propagate elevation from view to header layout

### LINKED ISSUE #6964: Turn off bottom shadow and set bottom padding of Flyout Header
## Summary
When defining a Flyout Header or FlyoutHeaderTemplate the rendered header includes a bottom border shadow which does not appear to be overridable. It also pads the bottom distance of the flyout header to the flyout items.

## API Changes

Create properties on the Shell:
FlyoutHeaderBottomBorder = false
FlyoutHeaderBottomPadding = 0

To help provide better control of the Flyout's Layout. As of now I cannot seem to get rid of these two issues that prevent me from implementing UX designs and prevents me from migrating to the Shell Application feature. Please note that issue #6293 would probably resolve this issue but not everyone should need to replace the entire content just to remove a drop shadow. I'm not sure why this is appearing in the first place. You could just remove the drop shadow all together.
Shell_FlyoutHeader
[App1.zip](https://github.com/xamarin/Xamarin.Forms/files/3427197/App1.zip)
 

## Intended Use Case
UX Design does not include any sort of shadow in the menu and there is strict padding between the Header logo and the menu items which I cannot replicate with the current Shell feature.

PureWeen:
### Description of Change ###

- Android is applying a 20px padding to the header view which doesn't match iOS. If users want to adding padding then they can just add a header

- Propagate the elevation platform specific from the header view to the Appbar so users can change the elevation and remove shadow if they want to

### Issues Resolved ### 
- fixes #6964

### Platforms Affected ### 

- Android

### Testing Procedure ###
- Test the header layout on the Test Control Gallery to make sure the list and header all layout correctly

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/ShellFlyoutTemplatedContentRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ShellFlyoutTemplatedContentRenderer.cs
@@ -76,7 +76,6 @@ namespace Xamarin.Forms.Platform.Android
 			appBar.AddView(_headerView);
 
 			var adapter = new ShellFlyoutRecyclerAdapter(shellContext, OnElementSelected);
-			recycler.SetPadding(0, (int)context.ToPixels(20), 0, 0);
 			recycler.SetClipToPadding(false);
 			recycler.SetLayoutManager(new LinearLayoutManager(context, (int)Orientation.Vertical, false));
 			recycler.SetAdapter(adapter);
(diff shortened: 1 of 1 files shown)


rlasker3:
My only comment on this approach is that it seems to me that anyone who has already implemented the flyout menu and accepted the layout will see a change to their layouts without having made any changes. That is why I proposed adding it as optional settings in #6964. Am I reading the code change wrong?

rlasker3:
Also, the bottom padding and shadow also appear in iOS except that it appears after the flyout and menu items. We should consider applying the change in both platforms.

PR #6972, opened 2019-07-25
[Feature] Storyboards

### LINKED ISSUE #4233: Spec: Storyboard Animations in XAML
Drafted by  @rmarinho 

## Rationale

This proposal aims to make animation more useful for VisualStateManager with the introduction of animation and storyboards with XAML support.

- Extend the `Animation` api to be used on XAML by introducing `Storyboard`, we should be able to specify at least the animation between 2 double values, colors and positions, user can specify the TargetElement (by default the attached element), TargetProperty, From, To, Duration and Easing of the animation, this should then be translated to a Animation callback to be applied on the `TargetElement`  ( by default that the `VisualState` is attached to). Storyboard should also add a `Completed` event, and a way to start and stop the animation from code behind.

- VSM should allow transitions between states

## Implementation

```cs

public class VisualTransition 
{
    public string From { get; set; }
    public string To { get; set; }
    public Storyboard Animation { get; set; }
    public double Duration { get; set; }
}

public sealed class VisualStateGroup 
{
    public string IList<VisualTransition> Transitions { get; }
}

[ContentProperty(nameof(Storyboard))]
public class VisualState
{
    ...
    public Storyboard Storyboard  { get; set; }
}


[ContentProperty(nameof(Children))]
public class Storyboard : Timeline, IList<Timeline>
{
	public static void SetTargetName(BindableObject bindable, string value)
	public static VisualElement GetTargetName(BindableObject bindable)

        public TimelineCollection Children { get; }

	public static readonly BindableProperty TargetNameProperty = BindableProperty.CreateAttached("TargetName", typeof(VisualElement), typeof(Timeline), null);

	public static readonly BindableProperty TargetPropertyProperty = BindableProperty.CreateAttached("TargetProperty", typeof(BindableProperty), typeof(Timeline), null);

	public static void SetTarget(BindableObject bindable, VisualElement value)
	{
		bindable.SetValue(TargetNameProperty, value);
	}

	public static VisualElement GetTarget(BindableObject bindable)
	{
		return (VisualElement)bindable.GetValue(TargetNameProperty);
	}

	public static void SetTargetProperty(BindableObject bindable, BindableProperty value)
	{
		bindable.SetValue(TargetPropertyProperty, value);
	}

	public static BindableProperty GetTargetProperty(BindableObject bindable)
	{
		return (BindableProperty)bindable.GetValue(TargetPropertyProperty);
	}

	public  Storyboard()  
	{  
		Children  =  new  TimelineCollection();  
	}  
  
	public  void  Add(Timeline  animation)  
	{  
		Children.Add(animation);  
	}  
  
	public  void  Remove(Timeline  animation)  
	{  
		Children.Remove(animation);  
	}  
  
	public void Begin();
	public void End();  
	public void Pause();  
	public void Resume();
	public void Loop();  
}  

public interface IAnimation
{
	void Begin();
	void End();
}
  
public interface IAnimation<T> : IAnimation
{
	T From { get; set; }
	T To { get; set; }
}
 
public interface IInterpolatable
{
	object InterpolateTo(object from, object target, double interpolation);
}

public interface IInterpolatable<T> : IInterpolatable
{
	T InterpolateTo(T from, T target, double interpolation);
}

public abstract class Timeline : BindableObject, IAnimation
{
	protected Timeline()
	{
		Duration = defaultDuration;
		Easing = Easing.SpringOut;
	}

	public uint BeginTime { get; set; }
	public uint Duration { get; set; }
	public bool Reverse { get; set; }
	public bool Loop { get; set; }
	public Easing Easing { get; set; }
	public EventHandler Completed { get; set; }
	public void Begin();
 	public void End();
}

public abstract class Animation<T> : Timeline
{
	public abstract T From { get; set; }
	public abstract T To { get; set; }
}


public abstract class InterpolatableAnimation<T> : Animation<T> where T : IInterpolatable<T>
{
}

public class PositionAnimation : Animation<Point>
{
}

public class DoubleAnimation : Animation<double>
{
	public static readonly BindableProperty FromProperty = BindableProperty.Create(nameof(From), typeof(double), typeof(DoubleAnimation), .0);

	public static readonly BindableProperty ToProperty = BindableProperty.Create(nameof(To), typeof(double), typeof(DoubleAnimation), .0);

	public override double From
	{
		get { return (double)GetValue(FromProperty); }
		set { SetValue(FromProperty, value); }
	}

	public override double To
	{
		get { return (double)GetValue(ToProperty); }
		set { SetValue(ToProperty, value); }
	}
}

public class TimelineCollection : List<Timeline>
{
}

[ContentProperty(nameof(Storyboard))]
public sealed class VisualState 
{
    ...
    public Storyboard Storyboard { get; set; }
}

public struct Color : IInterpolatable<Color>

[Xaml.TypeConversion(typeof(VisualElement))]
	public sealed class VisualElementConverter : TypeConverter, IExtendedTypeConverter
{
}

[TypeConverter(typeof(VisualElementConverter))]
public partial class VisualElement : Element, IAnimatable, IVisualElementController, IResourcesProvider, IFlowDirectionController


```

Add a ,markup extension so the developer doesn t have to worry about the type of animation and write less xaml.

```cs
public class AnimateExtension : BindableObject, IMarkupExtension<Timeline>
{

	public BindableProperty Property { set; get; }

	public string Binding { set; get; }

	public Timeline ProvideValue(IServiceProvider serviceProvider)
	{
		var anim = new PositionAnimation();

		BindingBase GetBinding()
		{
			return new Binding(Binding);
		}

		anim.SetBinding(PositionAnimation.ToProperty, GetBinding());
		Storyboard.SetTargetProperty(anim, Property);
		return anim;
	}

	object IMarkupExtension.ProvideValue(IServiceProvider serviceProvider)
	{
		return (this as IMarkupExtension<Timeline>).ProvideValue(serviceProvider);
	}
}
```

rmarinho:
### Description of Change ###

Proof of concept of implementing Storyboards and Animations from Xaml #4233 

Please comment the API on the spec and implementation here. 

### Issues Resolved ### 

- fixes #4233 

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Color.cs
+++ b/Xamarin.Forms.Core/Color.cs
@@ -8,7 +8,7 @@ namespace Xamarin.Forms
 {
 	[DebuggerDisplay("R={R}, G={G}, B={B}, A={A}, Hue={Hue}, Saturation={Saturation}, Luminosity={Luminosity}")]
 	[TypeConverter(typeof(ColorTypeConverter))]
-	public struct Color
+	public struct Color : IInterpolatable<Color>
 	{
 		readonly Mode _mode;
 
@@ -400,6 +400,24 @@ namespace Xamarin.Forms
 		{
 			return new Color(h, s, l, a, Mode.Hsl);
 		}
+
+
+		public object InterpolateTo(object from, object target, double interpolate)
+		{
+			return InterpolateTo((Color)from, (Color)target, interpolate);
+		}
+
+		public Color InterpolateTo(Color from, Color target, double interpolate)
+		{
+
+			var rR = from.R + (target.R - from.R) * interpolate;
+			var gG = from.G + (target.G - from.G) * interpolate;
+			var bB = from.B + (target.B - from.B) * interpolate;
+			var aA = from.A + (target.A - from.A) * interpolate;
+
+			return FromRgba(rR, gG, bB, aA);
+		}
+
 #if !NETSTANDARD1_0
 		public static implicit operator System.Drawing.Color(Color color)
 		{
--- a/Xamarin.Forms.Core/Interactivity/TriggerBase.cs
+++ b/Xamarin.Forms.Core/Interactivity/TriggerBase.cs
@@ -122,8 +122,8 @@ namespace Xamarin.Forms
 
 			public void Add(T item)
 			{
-				if (IsReadOnly)
-					throw new InvalidOperationException("This list is ReadOnly");
+				//if (IsReadOnly)
+				//	throw new InvalidOperationException("This list is ReadOnly");
 				_actual.Add(item);
 			}
 
--- a/Xamarin.Forms.Core/Storyboard.cs
+++ b/Xamarin.Forms.Core/Storyboard.cs
@@ -1,0 +1,175 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Xamarin.Forms
+{
+	[ContentProperty(nameof(Children))]
+	public class Storyboard : Timeline
+	{
+		Dictionary<VisualElement, List<Animation>> _animationsPerElement = new Dictionary<VisualElement, List<Animation>>();
+		List<Animation> _runningAnimations = new List<Animation>();	
+		public TimelineCollection Children { get; }
+		public static readonly BindableProperty TargetNameProperty = BindableProperty.CreateAttached("TargetName", typeof(VisualElement), typeof(Timeline), null);
+
+		public static readonly BindableProperty TargetPropertyProperty = BindableProperty.CreateAttached("TargetProperty", typeof(BindableProperty), typeof(Timeline), null);
+
+		public static void SetTarget(BindableObject bindable, VisualElement value)
+		{
+			bindable.SetValue(TargetNameProperty, value);
+		}
+
+		public static VisualElement GetTarget(BindableObject bindable)
+		{
+			return (VisualElement)bindable.GetValue(TargetNameProperty);
+		}
+
+		public static void SetTargetProperty(BindableObject bindable, BindableProperty value)
+		{
+			bindable.SetValue(TargetPropertyProperty, value);
+		}
+
+		public static BindableProperty GetTargetProperty(BindableObject bindable)
+		{
+			return (BindableProperty)bindable.GetValue(TargetPropertyProperty);
+		}
+
+		public Storyboard()
+		{
+			Children = new TimelineCollection();
+		}
+
+		public void Add(Timeline animation)
+		{
+			Children.Add(animation);
+		}
+
+		public void Remove(Timeline animation)
+		{
+			Children.Remove(animation);
+		}
+
+		public void Begin()
+		{
+			_animationsPerElement.Clear();
+			_runningAnimations.Clear();
+			var maxDuration = Duration;
+			var defaultTarget = Storyboard.GetTarget(this);
+			var defaultTargetProperty = Storyboard.GetTargetProperty(this);
+			foreach (var item in Children)
+			{
+				if (item is Animation<Color> timeline)
+				{
+					var animationTarget = Storyboard.GetTarget(timeline);
+					if (animationTarget == null)
+						animationTarget = defaultTarget;
+					
+					var animationTargetProperty = Storyboard.GetTargetProperty(timeline);
+					if (animationTargetProperty == null)
+						animationTargetProperty = defaultTargetProperty;
+					
+					var initialValue = animationTarget.GetValue(animationTargetProperty);
+
+					var finalValue = timeline.To;
+					var newAnimation = new Animation(v =>
+					{
+						object newValue = null;
+
+						if (timeline.From is IInterpolatable interpolatable)
+						{
+							newValue = interpolatable.InterpolateTo(initialValue, finalValue, v);
+						}
+
+						animationTarget.SetValue(animationTargetProperty, newValue);
+					}, 0, 1, timeline.Easing, () => timeline?.Completed?.Invoke(this, new EventArgs()));
+
+					//mainAnimation.Add(animation.BeginTime, 1, newAnimation);
+					maxDuration = Math.Max(maxDuration, timeline.Duration);
+
+					if(!_animationsPerElement.ContainsKey(animationTarget))
+					{
+						var list = new List<Animation>();
+						list.Add(newAnimation);
+						_animationsPerElement.Add(animationTarget, list);
+					}
+					else
+					{
+						_animationsPerElement[animationTarget].Add(newAnimation);
+					}
+				}
+
+			}
+
+			foreach (var anim in _animationsPerElement)
+			{
+				var elementAnimation = new Animation();
+				foreach (var animation in anim.Value)
+				{
+					elementAnimation.Add(0, 1, animation);
+			
+				}
+				_runningAnimations.Add(elementAnimation);
+				elementAnimation.Commit(anim.Key, "ChildAnimations", 16, maxDuration, finished: (v, c) => Completed?.Invoke(this, new EventArgs()));
+			}
+			//
+		}
+		public void End()
+		{
+			
+		}
+
+		//public void Pause();
+		//public void Resume();
+		//public static void SetTarget(this Storyboard s, string target);
+		//public static void SetTargetProperty(this Storyboard s, string property);
+	}
+
+	public interface IInterpolatable
+	{
+		object InterpolateTo(object from, object target, double interpolation);
+	}
+
+	public interface IInterpolatable<T> : IInterpolatable
+	{
+		T InterpolateTo(T from, T target, double interpolation);
+	}
+
+	public class Animation<T> : Timeline where T : IInterpolatable<T>
+	{
+		public T From { get; set; }
+		public T To { get; set; }
+	}
+
+	public class DoubleAnimation : Timeline
+	{
+		public double From { get; set; }
+		public double To { get; set; }
+	}
+
+	public abstract class Timeline : VisualElement
+	{
+		const uint defaultDuration = 4000;
+
+		protected Timeline()
+		{
+			Duration = defaultDuration;
+			Easing = Easing.Linear;
+		}
+
+		public uint BeginTime { get; set; }
+
+		public uint Duration { get; set; }
+
+		public bool AutoReverse { get; set; }
+
+		public Easing Easing { get; set; }
+		public EventHandler Completed { get; set; }
+
+	}
+
+	public class TimelineCollection : List<Timeline>
+	{
+
+	}
+}
--- a/Xamarin.Forms.Core/Storyboard.cs
+++ b/Xamarin.Forms.Core/Storyboard.cs
@@ -64,6 +64,42 @@ namespace Xamarin.Forms
 			Children.Remove(animation);
 		}
 
+		public void Begin(Timeline item)
+		{
+			_animationsPerElement.Clear();
+			_runningAnimations.Clear();
+			var maxDuration = Duration;
+			var defaultTarget = Storyboard.GetTarget(this);
+			var defaultTargetProperty = Storyboard.GetTargetProperty(this);
+
+
+			if (item is Animation<Color> timeline)
+			{
+				maxDuration = AddAnimationColor(maxDuration, defaultTarget, defaultTargetProperty, timeline);
+			}
+
+			if (item is DoubleAnimation doubleAnimation)
+			{
+				AddDoubleAnimation(defaultTarget, defaultTargetProperty, doubleAnimation);
+			}
+
+			if (item is PositionAnimation positionAnimation)
+			{
+				AddPositionAnimation(defaultTarget, defaultTargetProperty, positionAnimation);
+			}
+			foreach (var anim in _animationsPerElement)
+			{
+				var elementAnimation = new Animation();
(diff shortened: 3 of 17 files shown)


Xamarin.Forms.Core/Interactivity/TriggerBase.cs
```diff
@@ -122,8 +122,8 @@ public SealedList()
 
 			public void Add(T item)
 			{
-				if (IsReadOnly)
-					throw new InvalidOperationException("This list is ReadOnly");
+				//if (IsReadOnly)
```
andreinitescu: Why keep comments?

Xamarin.Forms.Core/Storyboard.cs
```diff
@@ -0,0 +1,551 @@
...
+			var animationTarget = Storyboard.GetTarget(doubleAnimation);
+			if (animationTarget == null)
+				animationTarget = defaultTarget;
+
+			var animationTargetProperty = Storyboard.GetTargetProperty(doubleAnimation);
+			if (animationTargetProperty == null)
+				animationTargetProperty = defaultTargetProperty;
+
+
+			//maxDuration = Math.Max(maxDuration, doubleAnimation.Duration);
```
andreinitescu: Comments..

Xamarin.Forms.Core/Storyboard.cs
```diff
@@ -0,0 +1,551 @@
...
+				{
+					elementAnimation.Add(0, 1, animation);
+
+				}
+				_runningAnimations.Add(elementAnimation);
+				elementAnimation.Commit(anim.Key, "ChildAnimations", 16, maxDuration, finished: (v, c) => Completed?.Invoke(this, new EventArgs()));
+			}
+		}
+
+		public new void Begin()
```
andreinitescu: Can't `new` be avoided? Shouldn't `Timeline.Begin()` be virtual? I see `Timeline.Begin` just throws `NotImplementedException` ...
This design doesn't seem right to me

Xamarin.Forms.Core/Storyboard.cs
```diff
@@ -0,0 +1,551 @@
...
+		{
+			_animationsPerElement.Clear();
+			_runningAnimations.Clear();
+			var maxDuration = Duration;
+			var defaultTarget = Storyboard.GetTarget(this);
+			var defaultTargetProperty = Storyboard.GetTargetProperty(this);
+			foreach (var item in Children)
```
andreinitescu: The whole logic here seems duplicate of Begin(Timeline item)`, or maybe I am wrong? Why not call `Begin(item)` ?

Xamarin.Forms.Core/Storyboard.cs
```diff
@@ -0,0 +1,551 @@
...
+			bindable.SetValue(TargetPropertyProperty, value);
+		}
+
+		protected override void OnBindingContextChanged()
+		{
+			base.OnBindingContextChanged();
+			foreach (var item in Children)
+			{
+				item.SetValue(BindingContextProperty, BindingContext);
+				//SetInheritedBindingContext(item, BindingContext);
```
andreinitescu: There's several code commented of lots of empty new lines...

Xamarin.Forms.Core/Storyboard.cs
```diff
@@ -0,0 +1,551 @@
...
+					if (Reverse)
+					{
+						elementAnimation.Add(0, 0.5, animation);
+						elementAnimation.Add(0.5, 1, new Animation(animation.GetCallback(), 1, 0, Easing));
+					}
+					else
+					{
+						elementAnimation.Add(0, 1, animation);
+					}
+
```
andreinitescu: When I format Storyboard.cs a difference appears just because of this line... I wish all code was formatted before committing otherwise any future commit will create a difference. And I had to work hard on fixing these differences and not to trigger changes because of formatting the code

Xamarin.Forms.Core/Storyboard.cs
```diff
@@ -0,0 +1,551 @@
...
+		void AddPositionAnimation(VisualElement defaultTarget, BindableProperty defaultTargetProperty, PositionAnimation positionAnimation)
+		{
+			var animationTarget = GetTarget(positionAnimation);
+			if (animationTarget == null)
+				animationTarget = defaultTarget;
+
+			var animationTargetProperty = GetTargetProperty(positionAnimation);
+			if (animationTargetProperty == null)
+				animationTargetProperty = defaultTargetProperty;
```
andreinitescu: This block and previous one can be written in a much more compact way. Also, would be good to try use the object type everytime it's not obvious:
```
VisualElement animationTarget = GetTarget(positionAnimation) ?? defaultTarget;
BindableProperty animationTargetProperty = GetTargetProperty(positionAnimation) ?? defaultTargetProperty;
```

Xamarin.Forms.Core/Storyboard.cs
```diff
@@ -0,0 +1,551 @@
...
+
+
+			//maxDuration = Math.Max(maxDuration, doubleAnimation.Duration);
+			AddToInternalAnimationList(positionAnimation, animationTarget, animationTargetProperty);
+		}
+
+		void AddDoubleAnimation(VisualElement defaultTarget, BindableProperty defaultTargetProperty, DoubleAnimation doubleAnimation)
```
andreinitescu: AddPositionAnimation, AddDoubleAnimation have exactly same implementation. Wouldn't be  better to have a template method for this? 
```
void AddAnimation<T>(VisualElement defaultTarget, BindableProperty defaultTargetProperty, PositionAnimation positionAnimation)
{
 ...
}
```
Their implementation is the same and just call the existing template method `AddToInternalAnimationList<T>`

Xamarin.Forms.Core/Storyboard.cs
```diff
@@ -0,0 +1,551 @@
...
+		}
+	}
+
+
+	public abstract class InterpolatableAnimation<T> : Animation<T> where T : IInterpolatable<T>
+	{
+	}
+
+
+	public class PositionAnimation : Animation<int>
```
andreinitescu: Why is it called PositionAnimation? One example shows animating the Y property:

<PositionAnimation To="{Binding Y}" Storyboard.TargetProperty="Frame.Y" />
but the Y property is a double and the PositionAnimation is implemented with int:

public class PositionAnimation : Animation<int>
I'm confused, why's it called "PositionAnimation"?

andreinitescu: Why can't I just use DoubleAnimation ?

andreinitescu:
Another thing: I see what was implemented differs from the spec in #4233
My suggestion is to either update the spec in #4233 or put the spec here. Either way would be good, otherwise it's confusing and don't know what to expect (I am left wondering, was the spec changed due to internal discussions or it's a mistake in the implementation not following the spec,, or...?)

rmarinho:
Thanks for the feedback @andreinitescu you are totally right this PR is very broken. This was just a hack to POC that it could be easy to introduce the Storyboard API.
I just open as a draft to see the changes and get more feedback.
Feel free to make changes you are suggesting on the implementation, I agree with them, and I would not have time to come back to this in the next month.
 in terms of public API please comment on the spec #4233 and we will update these the original post

Xamarin.Forms.Core/Storyboard.cs
rmarinho: I agree, well my idea here is we should translate this better to the existing extension methods that return Task for animations, and the Storyboard would just compose a Task array and await `Task.WhenAll`

Xamarin.Forms.Core/Storyboard.cs
rmarinho: I might be having something wrong on my preview machine. You are right ,  we need to add something to the repo to verify style, I think I saw something from the NETCore team that does this.

Xamarin.Forms.Core/Storyboard.cs
rmarinho: just me hacking :)

Xamarin.Forms.Core/Storyboard.cs
rmarinho: Yeah this could be refactor better, but the idea is when 1 property change you call Begin on that animation , if you Begin a Storyboard it begins all animations. But sure it could be refactored

Xamarin.Forms.Core/Storyboard.cs
rmarinho: yap when I was updating the spec I changed a couple of things that are not reflected here.

Xamarin.Forms.Core/Storyboard.cs
rmarinho: we need to implement this, I didn't had the logic to figure the maxDuration of the storyboard

Xamarin.Forms.Core/Interactivity/TriggerBase.cs
rmarinho: because this is not right, needs to be fixed

andreinitescu:
@rmarinho 
> this PR is very broken

I don't think it's very broken, it's good stuff. It just needs more work.

Xamarin.Forms.Core/Storyboard.cs
andreinitescu: This would be truly glorious to get. It might seem like a little thing, but it's so important on the long run, it saves a great amount of time for everybody.

Xamarin.Forms.Core/Color.cs
```diff
@@ -400,6 +400,24 @@ public static Color FromHsla(double h, double s, double l, double a = 1d)
...
+
+		public Color InterpolateTo(Color from, Color target, double interpolate)
+		{
+
+			var rR = from.R + (target.R - from.R) * interpolate;
+			var gG = from.G + (target.G - from.G) * interpolate;
+			var bB = from.B + (target.B - from.B) * interpolate;
+			var aA = from.A + (target.A - from.A) * interpolate;
+
+			return FromRgba(rR, gG, bB, aA);
```
StephaneDelcroix: that gives you brown, every time

color animation are better done using HSL

Xamarin.Forms.Core/Color.cs
```diff
@@ -8,7 +8,7 @@ namespace Xamarin.Forms
 {
 	[DebuggerDisplay("R={R}, G={G}, B={B}, A={A}, Hue={Hue}, Saturation={Saturation}, Luminosity={Luminosity}")]
 	[TypeConverter(typeof(ColorTypeConverter))]
-	public struct Color
+	public struct Color : IInterpolatable<Color>
```
StephaneDelcroix: a ColorAnimation class is better than this

Xamarin.Forms.Core/Storyboard.cs
```diff
@@ -0,0 +1,551 @@
...
+		//  Storyboard.SetTargetProperty(animOpacity, new PropertyPath("Opacity"));
+		//  Storyboard.SetTargetProperty(animTransform, new PropertyPath(
+
+		//public void Pause();
+		//public void Resume();
+		//public static void SetTarget(this Storyboard s, string target);
+		//public static void SetTargetProperty(this Storyboard s, string property);
+	}
+
+	public interface IInterpolatable
```
StephaneDelcroix: that kinda defeat the actual Animation API

rmarinho:
build

rmarinho:
build

rmarinho:
build

rmarinho:
build

rmarinho:
build

rmarinho:
build

rmarinho:
build

rmarinho:
build

rmarinho:
build

rmarinho:
build

Xamarin.Forms.Core/Color.cs
jsuarezruiz: I didn't see the problem (get always brown) you are commenting on here Stephane.
Seems to do the color interpolation correctly: https://www.alanzucconi.com/2016/01/06/colour-interpolation/
I did it in another place in a similar way: https://github.com/jsuarezruiz/Xamanimation/blob/master/Xamanimation/Helpers/AnimationHelper.cs#L20

Xamarin.Forms.Core/Storyboard.cs
jsuarezruiz: In this case, it makes sense to have an `IntAnimation` to animate int type properties. For example, animate the FontSize of a Label.

Xamarin.Forms.Core/Storyboard.cs
```diff
@@ -0,0 +1,551 @@
...
+			Easing = Easing.SpringOut;
+		}
+
+		public uint BeginTime { get; set; }
+
+		public uint Duration { get; set; }
+
+		public bool Reverse { get; set; }
+
+		public bool Loop { get; set; }
```
jsuarezruiz: How about something like [RepeatBehavior](https://docs.microsoft.com/es-es/dotnet/api/system.windows.media.animation.timeline.repeatbehavior?view=netframework-4.8)? With the bool type, we launch the animation 1 time, or in Loop. However, RepeatBehavior could allow to repeat the animation also a certain number of times (Example: 3 times).

xamarindevelopervietnam:
realy love this feature <3

samhouts:

Issue #6671, opened 2019-06-26
Tab error on macOS (Xamarin.forms)


vsfeedback:
VSF_TYPE_MARKDOWNOn macOS (MacBook Pro 17&quot; mid 2010) I have a problem with TabStop.

When I press the Tab key the focus jumps 2 indexes forward. I have attached screen captures of the behavior on macOS (ScreenCapture_macOS.mp4) and Windows 10 (ScreenCapture_Windows_10.mp4).

As you can see the tab stops work fine on windows 10 but on macOS it always jumps 2 indexes forward.

MainPage.xaml is attached also.

Best regards,
Eigil

_This issue has been moved from https://developercommunity.visualstudio.com/content/problem/591450/tab-error-on-macos-xamarinforms.html
VSTS ticketId: 903999_
_These are the original issue comments:_

Visual Studio Feedback System on 6/2/2019, 07:06 PM (23 days ago): <p>We have directed your feedback to the appropriate engineering team for further evaluation. The team will review the feedback and notify you about the next steps.</p>

Samantha Houts [MSFT] on 6/5/2019, 01:44 PM (20 days ago): <p>Thank you for the report! I don't see any attachments. Can you please attach a small project that demonstrates this issue? Thanks!</p>

_These are the original issue solutions:_
(no solutions)

samhouts:
Note: repro project is attached to private ticket. Thanks!

AndreiMisiukevich:
it might be fixed via my PR

jfversluis:
Taking into consideration the remaining lifetime of Xamarin.Forms, the platform this is about, the age of this issue and the traction this issue has seen I think it's unlikely that this will get fixed. If it is something that you still would like to have, please consider opening a PR. Thanks!

PR #6982, opened 2019-07-25
[macOS] Switch focus on TAB pressed

### LINKED ISSUE #6074: [Bug] macOS pressing the tab key moves to the 2nd  entry instead of first
### Description
On macOS pressing the tab key moves to the 2nd  entry instead of first

### Steps to Reproduce
1. Use this page:
```
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms" xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml" x:Class="CargoPacking.MyPage">
    <ContentPage.Content>
        <StackLayout>
            <Entry/>
            <Entry/>
            </StackLayout>
    </ContentPage.Content>
</ContentPage>
```
2. Run the app and press the tab key
3. 

### Expected Behavior
Focus should go to the 2nd entry
### Actual Behavior
Focus is just briefly on the 2nd entry and then returns to the 1st entry
### Basic Information

- Version with issue: 4.0.0.346134-pre9
- Last known good version: 2.3.5.235-pre2
- IDE: Visual Studio 2019
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Nuget Packages:
- Affected Devices:

AndreiMisiukevich:
### Description of Change ###
Tab switching hasn't been working.

The problem is that macOS switched views itself. And KeyUp is called by NEXT view. 
+ it doesn't make sense to switch focus on UP event. Because native behavior is switching on KeyDown

### Issues Resolved ### 
- fixes #6074
- related #6671

### API Changes ###
None

### Platforms Affected ### 
macOS

Switching via tab seems to be working

Not applicable

### Testing Procedure ###
Run any sample with entries / editors etc. and try to switch views with Tab and Shift+Tab

### DIFF
--- a/Xamarin.Forms.Platform.iOS/VisualElementRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/VisualElementRenderer.cs
@@ -205,7 +205,7 @@ namespace Xamarin.Forms.Platform.MacOS
 		}
 
 #if __MACOS__
-		public override void KeyUp(NSEvent theEvent)
+		public override void KeyDown(NSEvent theEvent)
 		{
 			if (theEvent.KeyCode == (ushort)NSKey.Tab)
 			{


AndreiMisiukevich:
@samhouts oh, it's [p/macOs] not iOS :)

samhouts:
@AndreiMisiukevich Yeah, it's a shared codebase, so I'll tag it with both (even if it'll only ever be accessed by macOS). Thanks!!

Issue #6884, opened 2019-07-16
[Docs] Update docs to help users run Xamarin.Forms source against their own projects


beeradmoore:
### Description
Provide updated documentation for someone wanting to add Xamarin.Forms source projects to their own project in order to debug issues.

@PureWeen has a project here that does this but it doesn't currently use the recommended way with props files
https://github.com/PureWeen/ReactShell

The other thing to keep in mind is that users will have 3rd party libraries with transitive dependencies on XF so effort has to be made so that the source projects are used and you don't get exceptions about duplicate targets being imported

https://github.com/PureWeen/SmartHotel360-Mobile/blob/submodule/Source/SmartHotel.Clients/SmartHotel.Clients/SmartHotel.Clients.csproj#L51

### Original Description
Currently having an issue within Xamarin.Forms/Xamarin.Forms.Platform.iOS/Renderers/PageContainer.cs and I am wanting to debug Xamarin.Forms to see why I am getting a null reference in there. Building XF will also be useful for making PRs in future so I'd prefer to focus on building XF now rather than focusing on that other issue.

I've tried to follow the steps from [here](https://github.com/xamarin/Xamarin.Forms/blob/master/README.md#mac)

### Steps to Reproduce
1. Download XF source.
2. Change to point to the release-4.1.0-sr1 tag.
3. Execute the build command.

Output: 

> $ ./build.sh --target vsmac
> Downloading packages.config...
> Downloading NuGet...
> Feeds used:
>   /Users/somename/.nuget/packages/
>   https://api.nuget.org/v3/index.json
> 
> Restoring NuGet package Cake.0.33.0.
> Adding package 'Cake.0.33.0' to folder '/Users/somename/git/Xamarin.Forms/caketools'
> Added package 'Cake.0.33.0' to folder '/Users/somename/git/Xamarin.Forms/caketools'
> Analyzing build script...
> Processing build script...
> Installing tools...
> Installing addins...
> The assembly 'Cake.Xamarin, Version=3.0.0.0, Culture=neutral, PublicKeyToken=null' 
> is referencing an older version of Cake.Core (0.26.1). 
> For best compatibility it should target Cake.Core version 0.33.0.
> The assembly 'Cake.Android.Adb, Version=3.0.0.0, Culture=neutral, PublicKeyToken=null' 
> is referencing an older version of Cake.Core (0.26.1). 
> For best compatibility it should target Cake.Core version 0.33.0.
[...]

PR #6993, opened 2019-07-27
remove sandbox

PureWeen:
### Description of Change ###

Can just setup project against submodule if needed
https://github.com/PureWeen/ReactShell

doesn't need to be part of main SLN


kvpt:
The approach is interesting, but not very easy to do.
Perhaps it can be simplified by using Directory.Build.props and Directory.Build.targets to isolate the maximum of things possible and make simple to bootstrap the projects.
And obviously a documentation => #6884

PureWeen:
The *Failed   ForceUpdateSizeCallsAreRateLimited* Unit test failed in release mode 🤷‍♂ 

@kvpt that's the ultimate goal. Take #6884 and document the best way to do all of this in your own projects. I just linked to the one I've been using for now as a reference.

PureWeen:
Unit tests passed on second run here
https://dev.azure.com/xamarin/public/_build/results?buildId=4743

PR #7041, opened 2019-08-01
[iOS] Added static bool and check for iOS 13

davidortinau:
### Description of Change ###

Adding this bool to work more specifically on new iOS 13 features.

No tests for the other bools, so none here.

### PR Checklist ###

- [ ] Has automated tests 


felipemomm:
Perhaps add for iOS 12 as we already have for iOS 9, 10, 11 and now 13?

I can easily see a developer trying to use it for iOS 12 and complaining why XF has this flag for all of the newer versions except 12... 🤔

PR #7057, opened 2019-08-02
[UITest] Change Query to WaitForElement on 45125 and change GetResult to await for Performance Gallery testing

PureWeen:
### Description of Change ###
- 45125 is currently using Query()[0] which will fail sometimes because of timing issues. Changed this to a WaitForElement so that it Waits for it to show up
- The Performance Gallery has been occasionally crashing from the HttpClient call throwing a TaskCancelException so I changed it to use await/async and wrapped the call in a try/catch

### PR Checklist ###

- [x] Has automated tests 


PureWeen:
All green!!

PureWeen:
@samhouts  the performance gallery run finished with Inconclusive after 7 minutes but I also checked the 4.2.0 branch UI Test runs and those finish the same way after 7 minutes

Issue #3177, opened 2018-06-28
Bindings to a DynamicObject


aloisdeniel:
### Description

Binding properties of a DynamicObject to a bindable object doesn't work.

### Steps to Reproduce

1. Create a dynamic view model class

```csharp
 public class DynamicViewModel : DynamicObject, INotifyPropertyChanged
    {
        public event PropertyChangedEventHandler PropertyChanged;

        private string title = "dynamic";

        public override bool TrySetMember(SetMemberBinder binder, object value)
        {
            if (binder.Name == nameof(title))
            {
                this.title = value as string;
            }

            return true;
        }

        public override bool TryGetMember(GetMemberBinder binder, out object result)
        {
            if (binder.Name == nameof(title))
            {
                result = this.title;
            }

            return base.TryGetMember(binder, out result);
        }

        public override IEnumerable<string> GetDynamicMemberNames()
        {
            return new[] { nameof(title) };
        }

    }
}
```
2. Bind a page to the dynamic view model

```xml
<?xml version="1.0" encoding="UTF-8"?>
<ContentPage
    xmlns="http://xamarin.com/schemas/2014/forms" 
    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml" 
    x:Class="Dynamic.Sample.HomePage">

    <Grid>
        <Label VerticalOptions="Center" Text="{Binding title}" />
    </Grid>
    
</ContentPage>
```
```csharp

  public partial class HomePage
  {
        #region Constructors

        public HomePage()
        {
            this.BindingContext = new DynamicViewModel();
            InitializeComponent();
        }

        #endregion
  }
```

### Expected Behavior

The `dynamic` value should be displayed on the screen.

### Actual Behavior

Nothing is displayed on the screen and a log appears :

```
'title' property not found on 'Dynamic.DynamicViewModel', target property: 'Xamarin.Forms.Label.Text'
```
### Basic Information

- Version with issue: 3.1.0.583944

StephaneDelcroix:
as a workaround, bind to an array

aloisdeniel:
To an array ? Did you mean `Dictionary<string,object>` with `[key]` accesses ?

StephaneDelcroix:
a dictionary is an option, any type implementing the indexer would work too

```csharp
public object this[string index]    // Indexer declaration  
{  
    get {
        //equiv to TryGetMember
    }
    set {
        //equiv to TrySetMember
    }
} 
```

[...]

PR #7082, opened 2019-08-06
[WIP] 7060 DataTemplateSelectorList

### LINKED ISSUE #7060: [Enhancement] DataTemplateSelector - Support for defining several templates in XAML code.
Creating a DataTemplateSelector manually is a little boring. it would be nice to do it right in xaml.

``` xaml
<DataTemplateSelector>
    <DataTemplate ForType="{x:Type local:Foo}">
        <Entry Text="I am a Foo" />
    </DataTemplate >
    <DataTemplate ForType="{x:Type System:String}">
        <Label Text="I am a String" />
    </DataTemplate >
</DataTemplateSelector>
```

KSemenenko:
#### Create DataTemplate for types in XAML ####
It seems to me that now to create a DataTemplateSelector we need to do some work to create DataTemplateSelector manually.
I think it would be more convenient to be able to create a DataTemplateSelector in XAML.
And I suggest adding something like _DataTemplateSelectorList_ in which I can create the right amount of DataTemplate for the types I need.

``` xaml
<ListView>
    <ListView.ItemTemplate>
        <DataTemplateSelectorList>
            <DataTemplate TemplateDataType="{x:Type local:NamedColor}">
                <Label Text="I am a local:NamedColor" />
            </DataTemplate>
            <DataTemplate TemplateDataType="{x:Type System:String}">
                <Label Text="I am a String" />
            </DataTemplate>
        </DataTemplateSelectorList>
    </ListView.ItemTemplate>
</ListView>
```
like in WPF 
https://docs.microsoft.com/en-us/dotnet/framework/wpf/data/data-templating-overview#the-datatype-property
### Issues Resolved ### 

- fixes #7060

### Platforms Affected ### 

- Core/XAML (all platforms)

### PR Checklist ###

- [ ] Has automated tests 


KSemenenko:
I would like to know your opinion about this implementation. Of course this is very rough, but I just wanted to show an idea.
 And I also want to ask you @StephaneDelcroix is it a good idea to use x:DataType there?

KSemenenko:
Also I found problem #3177 and I have the same. Although I do not have DynamicObject.

KSemenenko:
perhaps I should add a property of the DataType type to the DataTemplate (as it is in WPF) which will determine what type this template is for.

StephaneDelcroix:
@KSemenenko please provide a valid description for this PR first, then we'll start discussing :)

`x:DataType` is a bad idea, as it has a different meaning

your DataTemplateSelector implementation doesn't require any changes, and so far we tried to avoid including custom DTS in XF. if we add a DTS for target type, we'll have to add 12 of them, for the most common cases, and even the common cases might not cover everyone needs.

This is why I think this is best left out of XF core code...

KSemenenko:
@StephaneDelcroix You are right, I forgot to add a description. done.
In my projects, I create heaps of DataTemplateSelectors, but still use them in one place.
And I thought, maybe this would be appropriate in XF, a way to create sectors in XAML.

StephaneDelcroix:
I'm closing this for now, and the issue related, for the reasons stated earlier

PR #7084, opened 2019-08-07
Rename InitializationFlags -> ActivationFlags

kingces95:
### Description of Change ###

Rename `InitializationFlags` to `ActivationFlags` so I could use `ActivationFlags` during activation and `InitializationFlags` during the `Forms.Initialization` call. Duh.

This is technically a breaking change but no one should be using the flags. I introduced them to establish a mechanism whereby we can supply developers a way to disable unused feature initialization at startup. Disabling the one sample feature (css initialization) won't be noticeable by itself so that's why no one is using it. Only when all the perf tweaks are taken together is the gain noticeable. 

Introduce a new `ActivationFlags` with a flag `FixedStatusBarColor`. If the developer specifies the flag, then the developer is asking that we not burn time at startup making a system call whose only purpose is to allow the color to be changed.

Also renamed the flag from `NoCss` to `DisableCss` and `Initialize` to `Init` because that matches the method I'm overloading. 

### API Changes ###
+/- `ActivationFlags.NoCss` -> `InitializationFlags.DisableCss`
+/- `ActivationOptions` -> `InitializationOptions`
+/- `void Forms.Initialization(ActivationOptions)` -> `void Forms.Init(InitializationOptions)`

+ `InitializationFlags.FixedStatusBarColor`
+ `InitializationOptions`
+ `Forms(InitializationOptions)`

### Platforms Affected ### 

- Android

None, unless you opt-in to using the flags.

### Testing Procedure ###

UITests, XFShellStartup

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
@@ -23,6 +23,23 @@ using Xamarin.Forms.Internals;
 
 namespace Xamarin.Forms.Platform.Android
 {
+	[Flags]
+	public enum ActivationFlags : long
+	{
+		FixedStatusBarColor = 1 << 0,
+	}
+
+	public struct ActivationOptions
+	{
+		public ActivationOptions(Bundle bundle)
+		{
+			this = default(ActivationOptions);
+			this.Bundle = bundle;
+		}
+		public Bundle Bundle;
+		public ActivationFlags Flags;
+	}
+
 	public class FormsAppCompatActivity : AppCompatActivity, IDeviceInfoProvider
 	{
 		public delegate bool BackButtonPressedEventHandler(object sender, EventArgs e);
@@ -89,8 +106,8 @@ namespace Xamarin.Forms.Platform.Android
 			if (current != filter)
 				return;
 
-			Registrar.Registered.Register(target, handler);
-		}
+				Registrar.Registered.Register(target, handler);
+			}
 
 		// This is currently being used by the previewer please do not change or remove this
 		static void RegisterHandlers()
@@ -161,7 +178,19 @@ namespace Xamarin.Forms.Platform.Android
 			ActivityResultCallbackRegistry.InvokeCallback(requestCode, resultCode, data);
 		}
 
+		protected void OnCreate(ActivationOptions options)
+		{
+			OnCreate(options.Bundle, options.Flags);
+		}
+
 		protected override void OnCreate(Bundle savedInstanceState)
+		{
+			OnCreate(savedInstanceState, default(ActivationFlags));
+		}
+
+		private void OnCreate(
+			Bundle savedInstanceState, 
+			ActivationFlags flags)
 		{
 			_activityCreated = true;
 			if (!AllowFragmentRestore)
@@ -181,8 +210,10 @@ namespace Xamarin.Forms.Platform.Android
 				if (bar == null)
 					throw new InvalidOperationException("ToolbarResource must be set to a Android.Support.V7.Widget.Toolbar");
 			}
-			else
+			else 
+			{
 				bar = new AToolbar(this);
+			}
 
 			SetSupportActionBar(bar);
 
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -161,7 +178,19 @@ protected override void OnActivityResult(int requestCode, Result resultCode, Int
 			ActivityResultCallbackRegistry.InvokeCallback(requestCode, resultCode, data);
 		}
 
+		protected void OnCreate(ActivationOptions opitons)
```
djcparker: nit:  opitons => options

kingces95:
https://appcenter.ms/orgs/xtc-Xamarin-Forms/apps/AndroidControlGallery-1/test/series/pull-7084/runs/7b85267f-59aa-498c-bdc0-b108fdfef660

Issue3276, Verify Input Transparent/Image
Issue1875, Issue1898, Issue1898 2
Issue5766

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -89,8 +106,8 @@ static void RegisterHandler(Type target, Type handler, Type filter)
 			if (current != filter)
 				return;
 
-			Registrar.Registered.Register(target, handler);
-		}
+				Registrar.Registered.Register(target, handler);
```
rmarinho: is this needed ?

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -194,14 +225,16 @@ protected override void OnCreate(Bundle savedInstanceState)
 			_previousState = _currentState;
 			_currentState = AndroidApplicationLifecycleState.OnCreate;
 
-			OnStateChanged();
+				OnStateChanged();
```
rmarinho: Why the tab?

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -161,7 +178,19 @@ protected override void OnActivityResult(int requestCode, Result resultCode, Int
...
+		{
+			OnCreate(options.Bundle, options.Flags);
+		}
+
 		protected override void OnCreate(Bundle savedInstanceState)
+		{
+			OnCreate(savedInstanceState, default(ActivationFlags));
+		}
+
+		private void OnCreate(
```
rmarinho: drop private

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -23,6 +23,23 @@
 
 namespace Xamarin.Forms.Platform.Android
 {
+	[Flags]
+	public enum ActivationFlags : long
```
rmarinho: is this the right place to put this? can we use this in Non app compact ?

kingces95:
Except activation and creation are the same thing.. I'd like to give the breaking change a try. I really doubt it'll affect anyone and if so they'll be so bleeding edge that they should be happy to roll with the change.

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
kingces95: From a merge.

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
kingces95: Yeah, that's a good point. These flags are all hints. So passing the same flags to all the different derivations of Activity would mean that derivation specific (compat vs not) would be ignored by some derivations. I'm inclined to unify them because I don't think they'll be that many derivations, my guess is many flags would overlap, and the documentation is simpler. But then again, if they don't overlap.. then maybe a nested class would per derivation would be better?

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
kingces95: No, it's just a merge artifact.

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
kingces95: Also, shouldn't everyone be using AppCompat anyway? The other ones are just legacy right? So I figure if someone is worried about perf then they'd need to upgrade to the latest greatest version anyway. And so these flags will only ever be passed into the AppCompat version and, God forbid, any future VeryAppCompat versions...

PR #7092, opened 2019-08-07
[Android] Disable AndroidTicker if AnimatorDurationScale is equal or below 0

### LINKED ISSUE #5421: Animation tasks do not complete when Animator duration scale is set to off in developer options
### Steps to Reproduce

1. Set Animator duration scale to off in developer settings on an Android device.
2. Call any animation -> var animationTask = label.FadeTo(1)
3. await the resulting task

### Expected Behavior
Animation completes instantly, Task immediately completes.
### Actual Behavior
Animation does not occur, Task never completes. Tested with Xamarin Forms 3.5.0.169047

This caused our production app to completely break on several Android devices. due to fading of a loading screen never completing. Apparently many of our users turn Animator duration scale off to improve the speed of their device.

This bug is simular to this fixed issue https://github.com/xamarin/Xamarin.Forms/issues/1556

### Basic Information

- Version with issue: 3.5.0.169047
- Platform Target Frameworks: 
  - Android: Android 9.0 
- Affected Devices: most android devices

thisisthekap:
### Description of Change ###

Fixes the issue described in #5421 by disabling AndroidTicker if AnimatorDurationScale is set to 0.

### Issues Resolved ### 

- fixes #5421

### API Changes ###

None

### Platforms Affected ### 

- Android

None

Not applicable

### Testing Procedure ###

As given in the "Steps to Reproduce" in #5421:

1. Set Animator duration scale to off in developer settings on an Android device.
2. Call any animation -> var animationTask = label.FadeTo(1)
3. await the resulting task

Without the provided fix, the awaited task will never complete. If the fix is applied, no animation is shown and the task completes immediately.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/AndroidTicker.cs
+++ b/Xamarin.Forms.Platform.Android/AndroidTicker.cs
@@ -9,7 +9,8 @@ namespace Xamarin.Forms.Platform.Android
 	internal class AndroidTicker : Ticker, IDisposable
 	{
 		ValueAnimator _val;
-		bool _systemEnabled;
+		bool _energySaveModeDisabled;
+		readonly bool _animatorEnabled;
 
 		public AndroidTicker()
 		{
@@ -17,10 +18,11 @@ namespace Xamarin.Forms.Platform.Android
 			_val.SetIntValues(0, 100); // avoid crash
 			_val.RepeatCount = ValueAnimator.Infinite;
 			_val.Update += OnValOnUpdate;
+			_animatorEnabled = IsAnimatorEnabled();
 			CheckPowerSaveModeStatus();
 		}
 
-		public override bool SystemEnabled => _systemEnabled;
+		public override bool SystemEnabled => _energySaveModeDisabled && _animatorEnabled;
 
 		internal void CheckPowerSaveModeStatus()
 		{
@@ -31,7 +33,7 @@ namespace Xamarin.Forms.Platform.Android
 
 			if (!Forms.IsLollipopOrNewer)
 			{
-				_systemEnabled = true;
+				_energySaveModeDisabled = true;
 				return;
 			}
 
@@ -40,10 +42,17 @@ namespace Xamarin.Forms.Platform.Android
 			var powerSaveOn = powerManager.IsPowerSaveMode;
 
 			// If power saver is active, then animations will not run
-			_systemEnabled = !powerSaveOn;
+			_energySaveModeDisabled = !powerSaveOn;
 			
 			// Notify the ticker that this value has changed, so it can manage animations in progress
 			OnSystemEnabledChanged();
+		}
+
+		static bool IsAnimatorEnabled()
+		{
+			var resolver = global::Android.App.Application.Context.ContentResolver;
+			var scale = global::Android.Provider.Settings.Global.GetFloat(resolver, global::Android.Provider.Settings.Global.AnimatorDurationScale, 0);
+			return scale > 0;
 		}
 
 		public void Dispose()
--- a/Xamarin.Forms.Platform.Android/AndroidTicker.cs
+++ b/Xamarin.Forms.Platform.Android/AndroidTicker.cs
@@ -51,8 +51,9 @@ namespace Xamarin.Forms.Platform.Android
 		static bool IsAnimatorEnabled()
 		{
 			var resolver = global::Android.App.Application.Context?.ContentResolver;
-			if (resolver == null) {
-				return true;
+			if (resolver == null)
+			{
+				return false;
 			}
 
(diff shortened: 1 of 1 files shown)


thisisthekap:
@samhouts I have no access to the ui test reports (getting 401).

I think i need to have access to the test reports, to fix the issues accordingly :)

jfversluis:
Unfortunately we can't give you acces, but I have taken a look for you and the failing tests are unrelated to your change 🎉

Xamarin.Forms.Platform.Android/AndroidTicker.cs
```diff
@@ -40,12 +42,19 @@ internal void CheckPowerSaveModeStatus()
...
+			_energySaveModeDisabled = !powerSaveOn;
 			
 			// Notify the ticker that this value has changed, so it can manage animations in progress
 			OnSystemEnabledChanged();
 		}
 
+		static bool IsAnimatorEnabled()
+		{
+			var resolver = global::Android.App.Application.Context.ContentResolver;
+			var scale = global::Android.Provider.Settings.Global.GetFloat(resolver, global::Android.Provider.Settings.Global.AnimatorDurationScale, 0);
```
jfversluis: Could any of this be null? Do we need some error handling here?

thisisthekap: @jfversluis Should we consider to turn off animations if `Context` or `ContextResolver` is null?

By using `false` as default, we would make sure that the app stays usable.

thisisthekap:
@jfversluis Should we consider to turn off animations if Context or ContextResolver is null?

By using false as default, we would make sure that the app stays usable.

jfversluis:
What would happen with one or the other?

You say: stays usable, with the true default it will not be usable? Then yes, false should be the default! But I'm not really sure on what this default value would do, could you please elaborate?

thisisthekap:
@jfversluis If `false` is the default value, then we assume, that animations are disabled if the current state (if animations are enabled or disabled) is not detectable.

If `true` is the default value, then the issue fixed by this PR arises again if the animations are actually disabled, and the current state is not detectable.

jfversluis:
Then indeed I think we need to go with false as the default. Great thinking!

thisisthekap:
@jfversluis Is it feasible that this fix gets ported to 4.1 and 4.2 once the review is done?

jfversluis:
Not sure about 4.1 but actually this should target 4.2 :)

thisisthekap:
@jfversluis As this is my first PR in this repo: Is it my task to target 4.2, or is this done by you?

jfversluis:
I think you did it! Normally you’d have to rebase, but in this case it seems to be enough to change the target branch

thisisthekap:
@kingces95 As this issue is quite important to us to have the fix in one of the next Xamarin.Forms 4.2 releases: Can you provide any estimate on when this PR is going to be reviewed?

thisisthekap:
@samhouts Would it be feasible to have this fix also in 4.1.0, as 4.2.0 is still a prerelease?

samhouts:
@thisisthekap We try to limit service releases to regression fixes only to minimize risk. As it stands now, 4.2 is set to go stable at the same time as the next 4.1 service release. Are you willing to upgrade to 4.2 at that time?

Thanks!

thisisthekap:
@samhouts We are willing to upgrade to 4.2.0 as it is available as a stable release. So no need for porting it to 4.1.0 from our side.

Xamarin.Forms.Platform.Android/AndroidTicker.cs
```diff
@@ -40,12 +42,24 @@ internal void CheckPowerSaveModeStatus()
...
+			_energySaveModeDisabled = !powerSaveOn;
 			
 			// Notify the ticker that this value has changed, so it can manage animations in progress
 			OnSystemEnabledChanged();
 		}
 
+		static bool IsAnimatorEnabled()
+		{
+			var resolver = global::Android.App.Application.Context?.ContentResolver;
+			if (resolver == null)
```
bentmar: Should we fallback to users having animations? IMO this should be the default is nulls appear

thisisthekap: @bentmar Imho the fallback should be that animations are disabled.

Reason: If the animations are enabled if nulls appear, the app is potentially broken because of the freezing animation tasks.

thisisthekap: @bentmar Can this conversation be resolved?

bentmar: yes

PR #7101, opened 2019-08-08
Reinstated Xamarin icon for NuGets

jfversluis:
### Description of Change ###

The `.nuspec` files still pointed to a (now) non-existent URL so the NuGet icon would be broken. This adds the icon to GitHub and points it to the URL here.

🙅‍♂ 
image

❤️ 
image


clergym3n:
@rmarinho What happened to the nightly nugets? This was fixed recently, but I don't see **nightly** updates.

PR #7105, opened 2019-08-08
Reduce memory cost of struct Color

### LINKED ISSUE #6978: [Enhancement] Reduce memory cost for struct Color
## Summary
System.Runtime.InteropServices.Marshal.SizeOf(typeof(Color))
it return 32.  
which contains 7 float number + 1 enum.
the 7 float fields are: _a, _r, _g, _b, _hue, _saturation, _luminosity.
and the enum is Mode.

if the mode is Rgb, then _r, _g, _b, and _a will be used.
if the mode is Hsl, _hue, _saturation, _luminosity and _a will be used.
so there always have 3 fields is not used.
why not share 3 fileds for rgb or hsl? thus we can save 3 * 4 byte

## API Changes

No API changes

csleead:
### Description of Change ###

The _r, _g, _b, _h, _s, _l fields are replaced by _comp012 fields,
which are interepted to RGB values or HSL values according to the mode
flag. This approach reduces the memory cost of Color by 12 bytes.

### Issues Resolved ### 

- fixes #6978

### API Changes ###
None

### Platforms Affected ### 
- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###
Automated test

### PR Checklist ###
- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Color.cs
+++ b/Xamarin.Forms.Core/Color.cs
@@ -8,7 +8,7 @@ namespace Xamarin.Forms
 {
 	[DebuggerDisplay("R={R}, G={G}, B={B}, A={A}, Hue={Hue}, Saturation={Saturation}, Luminosity={Luminosity}")]
 	[TypeConverter(typeof(ColorTypeConverter))]
-	public struct Color
+	public readonly struct Color
 	{
 		readonly Mode _mode;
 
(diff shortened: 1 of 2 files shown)


Xamarin.Forms.Core/Color.cs
```diff
@@ -41,46 +41,92 @@ public double A
 			get { return _a; }
 		}
 
-		readonly float _r;
+		readonly float _rh;
```
StephaneDelcroix: user _component0, _component1, and _component2, or an array

Xamarin.Forms.Core/Color.cs
```diff
@@ -141,35 +174,89 @@ public Color AddLuminosity(double delta)
...
+
+			double s, l;
+			if (_mode == Mode.Hsl)
+			{
+				s = _gs;
+				l = _bl;
+			}
+			else
+			{
+				ConvertToRgb(_rh, _gs, _bl, out _, out float fs, out float fl);
```
StephaneDelcroix: nope. please add a test for this if it's not failing already

Xamarin.Forms.Core/Color.cs
```diff
@@ -141,35 +174,89 @@ public Color AddLuminosity(double delta)
...
+
+			double h, l;
+			if (_mode == Mode.Hsl)
+			{
+				h = _rh;
+				l = _bl;
+			}
+			else
+			{
+				ConvertToRgb(_rh, _gs, _bl, out float fh, out _, out float fl);
```
StephaneDelcroix: same

Xamarin.Forms.Core/Color.cs
```diff
@@ -141,35 +174,89 @@ public Color AddLuminosity(double delta)
...
+
+			double h, s;
+			if (_mode == Mode.Hsl)
+			{
+				h = _rh;
+				s = _gs;
+			}
+			else
+			{
+				ConvertToRgb(_rh, _gs, _bl, out float fh, out float fs, out _);
```
StephaneDelcroix: again

Xamarin.Forms.Core/Color.cs
csleead: I don't think array is a good option because it always allocates memory on heap.

 _component0, _component1, and _component2 are not as good as _rh,  _gs and _bl in my opinion because they are long and they don't explicitly state what may be inside of those field.
May I have the reason about this comment?

StephaneDelcroix: > I don't think array is a good option because it always allocates memory on heap.

starting with c#7.2, you can use `stackalloc` in an unsafe context, which is totally suitable here

StephaneDelcroix: for me, _comp0 carries more information than _rh

csleead: Not agreed, _comp0 carries no information to me. This is kind of like naming _a as _comp4.
But as you wish. :smirk:

csleead:
Code updated, please review

Happypig375:
Please mark `Color` as a `readonly struct` as well.

csleead:
> Please mark `Color` as a `readonly struct` as well.

DONE

StephaneDelcroix:
> Please mark Color as a readonly struct as well.

I'd second that as well, but:

> error CS8107: Feature 'readonly structs' is not available in C# 7.0. Please use language version 7.2 or greater

(we haven't opted-in c# 7.2 yet)

that's not a big deal, Color has always been readonly

csleead:
> > Please mark Color as a readonly struct as well.
> 
> I'd second that as well, but:
> 
> > error CS8107: Feature 'readonly structs' is not available in C# 7.0. Please use language version 7.2 or greater
> 
> (we haven't opted-in c# 7.2 yet)
> 
> that's not a big deal, Color has always been readonly

Reverted.
I guess it is better to specify lang version in csproj to avoid such mistake.

StephaneDelcroix:
> I guess it is better to specify lang version in csproj to avoid such mistake.

I'm testing a build right now to see if our build system supports it

StephaneDelcroix:
@csleead see #7131

csleead:
> we can not merge this as is, without assessing the usage of the different color components in platform code (see discussion we already had)

I added some tests. Please take a look

rmarinho:
@csleead can you rebase this on top of master? thanks

StephaneDelcroix:
in the current state, this is not mergeable, as it trades a small memory boost by a potential huge performance loss.

Fell free to reopen this as soon as the usage Color usage in all the code base and use shortcuts to avoid too many conversion.

Maybe we should rethink this, and split Color (rib), from HslColor, so the conversions are becoming explicit

PR #7108, opened 2019-08-08
[UITest] Fix 2951 and moved extension method to be more visible

PureWeen:
### Description of Change ###

- Fixed 2951 to retry querying for buttons. It looks like the UI tests are querying before the buttons have time to disappear
- Renamed the RetryUntilPresent  to QueryUntilPresent and moved it to the same namespace as Query so when people are using Query they'll see *QueryUntilPresent* and think "Hey I bet that's better to use"

### Testing Procedure ###
- Automated

### PR Checklist ###

- [x] Has automated tests 


PureWeen:
💚 🍏 📗 🥗 🇬🇱

PR #7113, opened 2019-08-08
Add priority parameter to HandlerAttribute to thwart priority by assembly order

### LINKED ISSUE #6939: [Bug] Replacing default Material Visual Renderers fails if the assemblies are scanned in the "wrong" order
### Description
Because the Material Renderers use ExportRenderer and are registered during the assembly scanning phase.  Whatever Assembly gets scanned second wins

So if there are two assemblies with

```C#
[assembly: ExportRenderer(typeof(Button), typeof(<someType>), new[] { typeof(VisualMarker.MaterialVisual) })]
```

It's a bit random who will win and since our Xamarin.Forms.Material assembly registers them like that then it's random if that one comes in second or an assembly the user has created wins

#### Possible fixes
-  Actually make the Init call on iOS do something and register the renderers by hand
- Force all known Xamarin Forms assemblies to get scanned first in the Registrar
- Somehow use RenderWith so when it's requested on demand it reaches into the Material library and uses it. Not sure how to make this work efficiently though
- create a targets file that generates a *.cs file that registers the renderers by hand. This way users don't have to call Init() on the material library and the renderers get registered before the Init call
- ????

### Steps to Reproduce

1.  git clone https://github.com/aubiyko/XamarinFormsMaterialTest
2.  Notice that the iOS renderer used to just replace Button never runs

```C#
[assembly: ExportRenderer(typeof(Button), typeof(MaterialCustomButtonRenderer), new[] { typeof(VisualMarker.MaterialVisual) })]
```

### Expected Behavior
MaterialCustomButtonRenderer should run

### Actual Behavior
MaterialCustomButtonRenderer doesn't run

### Basic Information

- Platform Target Frameworks: 
  - iOS:  Only noticed this on iOS most likely because it's randomly unlucky

### Reproduction Link

https://github.com/aubiyko/XamarinFormsMaterialTest

PureWeen:
### Description of Change ###

- Added a priority property to HandlerAttribute so you can prioritize which ExportRenderer is the most important. 

- Set all of our material renderers to have *short.MinValue* priority so if a user wants to replace all the default Material Button renderers they can just do this
```C#
[assembly: ExportRenderer(typeof(Button), typeof(DefaultMaterialRenderer), new[] { typeof(VisualMarker.MaterialVisual) })]
```

Without this change whatever Assembly is scanned last will be the winning handler used. 

### Issues Resolved ### 
- fixes #6939

### API Changes ###

Added:
 - short HandlerAttribute.Priority { get; set; } // We could make this internal for now so as to not change the external API but this property could be important for 3rd party component developers

### Platforms Affected ### 
- Core/XAML

### Testing Procedure ###
- unit tests

### PR Checklist ###

- [x] Has automated tests 


StephaneDelcroix:
> Without this change whatever Assembly is scanned last will be the winning handler used.

this is why we were scanning our assemblies first. Isn't that the case anymore ? Shouldn't we fix that instead ?

PureWeen:
@StephaneDelcroix That was one of the possible options for doing this but I was thinking this priority concept might be useful for 3rd party vendors as well who want to control how the attributes are resolved. 

https://xamarinhq.slack.com/archives/C149QN5C0/p1563590768001500?thread_ts=1563581513.000400&cid=C149QN5C0

Or we could just have a known list string array of our assemblies in the Registrar and just make sure those take lower priority than any user registered handler

Xamarin.Forms.Platform.Android/ExportRendererAttribute.cs
```diff
@@ -12,5 +12,9 @@ public ExportRendererAttribute(Type handler, Type target) : this(handler, target
 		public ExportRendererAttribute(Type handler, Type target, Type[] supportedVisuals) : base(handler, target, supportedVisuals)
 		{
 		}
+
+		public ExportRendererAttribute(Type handler, Type target, Type[] supportedVisuals, short priority) : base(handler, target, supportedVisuals, priority)
```
StephaneDelcroix: I wouldn't create an overload for an optional property. attributes supports named parameter resolved to properties, and they supported that way before we were able to use named parameters elsewhere

StephaneDelcroix:
@PureWeen feel free to dismiss my review and merge if you feel like it

Xamarin.Forms.Platform.Android/ExportRendererAttribute.cs
PureWeen: I made the changes to make the priority as internal as it can be

I also retargeted this PR to 4.2.0 since the idea now is that we aren't adding any new APIs and this is resolving a bug in the ability to replace all default material renderers

rmarinho:
Works for me i think this gives more options to have multiple renderers for the same Control, maybe you are implementing different themes for example.

jsuarezruiz:
Tested the test and it works. As Rui commented, the test that fails is not related (it is from CollectionView).

PR #7132, opened 2019-08-12
add wait before WaitForNoElement

PureWeen:
### Description of Change ###
The test only calls WaitForNoElement which isn't very definitive if it hasn't made sure that the test itself hasn't loaded.  When looking at AppCenter runs the test after this had screen shots of this test because this test was just passing before it actually finished running which would then cause the next test to fail

### Testing Procedure ###
Automated

### PR Checklist ###

- [x] Has automated tests 


PureWeen:
Failing tests not related

PR #7143, opened 2019-08-13
Replace IsDesignModeEnabled check with original Context Type Check

PureWeen:
### Description of Change ###

Revert to old implementation but just store into static variable so it only checks once


Xamarin.Forms.Platform.Android/ContextExtensions.cs
```diff
@@ -115,7 +115,23 @@ internal static Context GetThemedContext(this Context context)
 			return null;
 		}
 
-		internal static bool IsDesignerContext(this Context context) => DesignMode.IsDesignModeEnabled;
+		static bool? _isDesigerContext;
```
StephaneDelcroix: typo

PR #7148, opened 2019-08-13
Fixes issue #7109 in XF 4.0 branch

### LINKED ISSUE #7109: [Bug] [iOS] App crashes on iPhone 6 running iOS 11 after updating to XF 4.x.
### Description

Native stack trace of crash:
`DEBUG|8/5/2019 8:19:53 PM|App: Failed to init Home View|EX:Objective-C exception thrown.  Name: NSInvalidArgumentException Reason: -[Xamarin_Forms_Platform_iOS_PlatformRenderer setNeedsUpdateOfHomeIndicatorAutoHidden]: unrecognized selector sent to instance 0x104f8bea0`

This issue is due to adding functionality in XF 4.0 to support hiding the home indicator on iPhone X/Xs/Xr devices, as per:
https://github.com/xamarin/Xamarin.Forms/pull/5717

It would seem that on some devices that run iOS 11 that do not have a home indicator, i.e. any iPhone with a home button, do not have the setNeedsUpdateOfHomeIndicatorAutoHidden selector, so when XF calls:
```
if (Forms.IsiOS11OrNewer)
      SetNeedsUpdateOfHomeIndicatorAutoHidden();
```
on a phone running iOS 11 that does not support the home indicator, then the crash occurs.

An easy and seemingly harmless fix would be to check that the calling object responds to that selector, so instead of the above version check do the following selector check:
```
if (this.RespondsToSelector(new ObjCRuntime.Selector("setNeedsUpdateOfHomeIndicatorAutoHidden")))
   SetNeedsUpdateOfHomeIndicatorAutoHidden();
```

I also found anther report of the same issue where the fix was to do the resondsToSelector check: 
https://github.com/lionheart/openradar-mirror/issues/18762

These are all of the places in XF source code that call that method that I found with a search:
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/PlatformRenderer.cs#L94
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/ModalWrapper.cs#L107
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs#L193
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs#L528
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/PhoneMasterDetailRenderer.cs#L342
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/NavigationRenderer.cs#L234

### Steps to Reproduce

I cannot reproduce since I do not have an iPhone 6 running iOS 11 (or any other < iPhone X device running iOS 11). Issue does not reproduce on simulators. 

### Basic Information

- Version with issue: XF >= 4.0
- Last known good version: XF 3.6 (does not call the SetNeedsUpdateOfHomeIndicatorAutoHidden method)
- IDE:
- Platform Target Frameworks: 
  - iOS:  12.14.0.114
- Affected Devices: iPhone 6 running iOS 11 (only reported device)

jgold6:
### Description of Change ###

Changes check for iOS version >= 11 to a check to see if the `setNeedsUpdateOfHomeIndicatorAutoHidden` selector is available before calling the `SetNeedsUpdateOfHomeIndicatorAutoHidden` method.

 No automated test created as there should be no difference in behavior except to avoid a crash due to calling an Obj-C selector that may not be available on all devices running iOS >= 11. In this case, it seems that checking that the selector is available is better than checking for the iOS version as it will be a valid check regardless of iOS version, where as just the version check alone is not guaranteeing the availability of this particular selector as devices pre iPhone X do not have a Home Indicator. 

### Issues Resolved ### 

- fixes #7109 in XF 4.0.0 branch

### API Changes ###

None

### Platforms Affected ### 

- iOS

Users app won't crash when running on an iPhone 6 with iOS version 11 (and potentially other iPhones pre the iPhone X when running iOS 11)

Not applicable

### Testing Procedure ###

I was unable to test as I do not have an iPhone 6 running iOS 11. However the native stack trace provided by the customer clearly indicates that the crash is due to calling an Obj-C selector that does not exist.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Forms.cs
+++ b/Xamarin.Forms.Platform.iOS/Forms.cs
@@ -34,6 +34,7 @@ namespace Xamarin.Forms
 		static bool? s_isiOS9OrNewer;
 		static bool? s_isiOS10OrNewer;
 		static bool? s_isiOS11OrNewer;
+		static bool? s_respondsTosetNeedsUpdateOfHomeIndicatorAutoHidden;
 #endif
 
 #if __MOBILE__
@@ -65,6 +66,16 @@ namespace Xamarin.Forms
 				if (!s_isiOS11OrNewer.HasValue)
 					s_isiOS11OrNewer = UIDevice.CurrentDevice.CheckSystemVersion(11, 0);
 				return s_isiOS11OrNewer.Value;
+			}
+		}
+
+		internal static bool RespondsToSetNeedsUpdateOfHomeIndicatorAutoHidden
+		{
+			get
+			{
+				if (!s_respondsTosetNeedsUpdateOfHomeIndicatorAutoHidden.HasValue)
+					s_respondsTosetNeedsUpdateOfHomeIndicatorAutoHidden = new UIViewController().RespondsToSelector(new ObjCRuntime.Selector("setNeedsUpdateOfHomeIndicatorAutoHidden"));
+				return s_respondsTosetNeedsUpdateOfHomeIndicatorAutoHidden.Value;
 			}
 		}
 #endif
--- a/Xamarin.Forms.Platform.iOS/PlatformRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/PlatformRenderer.cs
@@ -104,7 +104,7 @@ namespace Xamarin.Forms.Platform.iOS
 		{
 			base.ViewDidLoad();
 			SetNeedsStatusBarAppearanceUpdate();
-			if (Forms.IsiOS11OrNewer)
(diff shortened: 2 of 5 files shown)


jgold6:
Oops, mistake.

PR #7158, opened 2019-08-14
[BUG/iOS] Fixed CollectionView Insets ItemSpacing

### LINKED ISSUE #7035: [Bug][iOS] CarouselView last element is clipped
### Description
This issue seems to be iOS specific.
I have created a carouselView with a horizontal orientation.
When i try to scroll to the last element, I go as far as I can and the last element is clipped.

Screen Shot 2019-07-31 at 3

### Steps to Reproduce

1. Create a CarouselView with Horizontal Orientation
2. Add a few elements so some elements are located off screen
3. Try scrolling to the last element

### Expected Behavior
Should see the element in full view.
Screen Shot 2019-07-31 at 3 11 49 PM
(IT WORKS ON ANDROID) not on iOS for some reason.
### Actual Behavior
As seen in the first image, the last element is slightly **clipped.**
### Basic Information

- Version with issue: 
Xamarin.Forms 4.1.0.618606

- Last known good version:
????

- IDE: 
VS for Mac (community) 8.2.1 (build 19)

- Platform Target Frameworks: 
  - iOS: 12.0+
  - Android: 5.0+

- Affected Devices:
iOS

### Screenshots
**ABOVE**
### Reproduction Link
**Ahh, sorry I cannot provide a link to the project due to security concerns.
below is the exact code used to create the CarouselView.**

<CarouselView
    Grid.Row="3"
    Grid.ColumnSpan= "4"
    x:Name= "CV"
    HeightRequest= "225"
    HorizontalOptions= "Center"
    VerticalOptions= "CenterAndExpand"
    Margin= "0" >
    < CarouselView.ItemsLayout >
        < GridItemsLayout
            Orientation= "Horizontal"
            HorizontalItemSpacing= "15"
            SnapPointsAlignment= "Center"
            SnapPointsType= "Mandatory" />
    </ CarouselView.ItemsLayout >
    < CarouselView.ItemTemplate >
        < DataTemplate >
            < Frame
                Padding= "0"
                Margin= "0"
                HeightRequest= "200"
                WidthRequest= "300"
                BorderColor= "LightGray"
                CornerRadius= "10"
                HasShadow= "False"
                IsClippedToBounds= "True" >
                < Grid
                    ColumnSpacing= "0"
                    RowSpacing= "0" >
                    < Grid.RowDefinitions >
                        < RowDefinition Height= "70*" />
                        < RowDefinition Height= "15*" />
                        < RowDefinition Height= "15*" />
                    </ Grid.RowDefinitions >
                    < Image
                        Grid.Row= "0"
                        Aspect= "Fill"
                        Source= "{Binding ImgUrl}" />
                    < Label
                        Grid.Row= "1"
                        Margin= "20,0,0,0"
                        FontSize= "Medium"
                        Text= "{Binding SubTitle}"
                        HorizontalOptions= "Start"
                        VerticalOptions= "End" />
                    < Label
                        Grid.Row= "2"
                        Margin= "20,0,0,0"
                        FontSize= "Large"
                        Text= "{Binding Title}"
                        HorizontalOptions= "Start"
                        VerticalOptions= "Start" />
                </ Grid >
            </ Frame >
        </ DataTemplate >
    </ CarouselView.ItemTemplate >
</ CarouselView >

jfversluis:
### Description of Change ###

Instead of returning a fixed UIEdgeInset.Zero, we now take in account the ItemSpacing so the last item will not be offscreen.

### Issues Resolved ### 

- fixes #7035

### API Changes ###

 
 None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###

Go to the Issue7035 case, if you are able to scroll to the right and the last item is fully visible, this worked!

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.ControlGallery.iOS/AppDelegate.cs
+++ b/Xamarin.Forms.ControlGallery.iOS/AppDelegate.cs
@@ -150,7 +150,7 @@ namespace Xamarin.Forms.ControlGallery.iOS
 			App.IOSVersion = int.Parse(versionPart[0]);
 
 			Xamarin.Calabash.Start();
-			Forms.SetFlags("CollectionView_Experimental");
+			// Forms.SetFlags("CollectionView_Experimental", "Shell_Experimental");
 			Forms.Init();
 			FormsMaps.Init();
 			FormsMaterial.Init();
--- a/Xamarin.Forms.ControlGallery.iOS/AppDelegate.cs
+++ b/Xamarin.Forms.ControlGallery.iOS/AppDelegate.cs
@@ -150,7 +150,7 @@ namespace Xamarin.Forms.ControlGallery.iOS
 			App.IOSVersion = int.Parse(versionPart[0]);
 
 			Xamarin.Calabash.Start();
-			// Forms.SetFlags("CollectionView_Experimental", "Shell_Experimental");
+			Forms.SetFlags("CollectionView_Experimental");
 			Forms.Init();
 			FormsMaps.Init();
 			FormsMaterial.Init();
--- a/Xamarin.Forms.Platform.iOS/CollectionView/GridViewLayout.cs
+++ b/Xamarin.Forms.Platform.iOS/CollectionView/GridViewLayout.cs
@@ -150,6 +150,9 @@ namespace Xamarin.Forms.Platform.iOS
 				return false;
 			}
 
+			if (CollectionView.NumberOfSections() == 0)
+				return false;
+
 			var itemCount = CollectionView.NumberOfItemsInSection(section);
 
 			if (itemCount < _itemsLayout.Span)
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7035.xaml
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7035.xaml
@@ -1,0 +1,73 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Xamarin.Forms.Controls.Issues.Issue7035">
+    <Grid>
+        <Grid.RowDefinitions>
+            <RowDefinition Height="300"/>
+        </Grid.RowDefinitions>
+        <BoxView BackgroundColor="Red" WidthRequest="400" HeightRequest="300" HorizontalOptions="CenterAndExpand" VerticalOptions="CenterAndExpand"/>
+    <CarouselView
+        x:Name="CV"
+        HeightRequest="225"
+        HorizontalOptions="Center" 
+        VerticalOptions="CenterAndExpand" 
+        Margin="15,10,0,20">
+        <CarouselView.ItemsLayout>
+            <GridItemsLayout
+                Orientation="Horizontal"
+                HorizontalItemSpacing="15"
+                SnapPointsAlignment="Center" 
+                SnapPointsType="Mandatory"/>
+        </CarouselView.ItemsLayout>
+        <CarouselView.ItemTemplate>
+            <DataTemplate>
+                <Frame
+                    Padding="0"
+                    Margin="0"
+                    HeightRequest="200"
+                    WidthRequest="300"
+                    BorderColor="LightGray"
+                    CornerRadius="10"
+                    HasShadow="False"
+                    IsClippedToBounds="True">
+                    <Grid
+                        ColumnSpacing="0"
+                        RowSpacing="0">
+                        <Grid.RowDefinitions>
+                            <RowDefinition Height="70*" />
+                            <RowDefinition Height="15*" />
+                            <RowDefinition Height="15*" />
+                        </Grid.RowDefinitions>
+                        <Image
+                            Grid.Row="0"
+                            Aspect="Fill"
+                            Source="xamarinlogo" />
+                        <Label
+                            Grid.Row="1"
+                            Margin="20,0,0,0"
+                            FontSize="Medium"
+                            Text="SUB"
+                            HorizontalOptions="Start"
+                            VerticalOptions="End" />
+                        <Label
+                            Grid.Row="2"
+                            Margin="20,0,0,0"
+                            FontSize="Large"
+                            Text="TITLE"
+                            HorizontalOptions="Start"
+                            VerticalOptions="Start" />
+                        <Image
+                            Margin="0,0,8,0"
+                            Grid.Row="1"
+                            Grid.RowSpan="2"
+                            HorizontalOptions="End"
+                            Aspect="AspectFit"
+                            Source="xamarinlogo" />
+                    </Grid>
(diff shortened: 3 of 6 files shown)


jfversluis:
@hartez does it make sense to do it like this? Was there a special reason to always return the zero value?

In addition I have added this check here; https://github.com/xamarin/Xamarin.Forms/pull/7158/files#diff-002de3cb5ec47d56de5914331ded80e5R147

This crashed immediately when I added the issue repro.

hartez:
@jfversluis What happens if you change the number of items in the collection to 33 and change the Span of the GridItemsLayout to 3, and then scroll all the way to the right?

jfversluis:
@hartez 

> I'm not sure why this fix is necessary

See #7035 (or repro I added in this code) only on iOS the last item runs off the screen. Coincidentally (I think not...) the amount of pixels that run off screen are the number of pixels that is the total of the HorizontalItemSpacing. 

AFAICT there is only ever one section used for this? But I guess that is indeed not taking into account the grouping.

In any case, that is why I was asking you to verify if this is a feasible solution as I am learning all details of the CollectionView.

I've tried putting in a bigger number, but as you can see with 42 items the solution still holds up.

Untitled

hartez:
> @hartez
> 
> > I'm not sure why this fix is necessary

I meant this in the sense of "the code that's in there shouldn't be causing this bug in the first place". Clearly _something_ is broken, and we'll need to address it in our code.

> 
> I've tried putting in a bigger number, but as you can see with 42 items the solution still holds up.
> 

Not just a bigger number - use a larger number of items _and_ set a value for Span. Like 42 items and a Span of 3 - what happens then?

hartez:
I updated the issue page to show what I mean - the bottom CarouselView has Span set to 3. Both CarouselViews have the HorizontalItemSpacing set to 25, and I turned off the center snapping (the snapping makes the problem hard to see):



The top one behaves as expected, but the bottom one has a huge amount of inset space on the right.

adrianknight89:
@hartez I was just testing `ItemSpacing` on Android and noticed that the spacing is added to the bottom of **each** item, which means the last item has it too. I expected `ItemSpacing` to be spacing between items, ignoring the top of the first item and the bottom of the last item. I understand that if we add more items to the data source, then spacing will continue from the last item (hence the spacing for the last item), but this does not work if all items are loaded and we don't want that extra spacing at the bottom.

Also, I'd like to be able to create equal spacing between items as well as the top of the first item and the bottom of the last item. This seems impossible with the current implementation. I added a `Margin` to `RefreshView`, but the content is cut off at the margins, which is undesirable.

That said, I think we need to be able to accomplish two things with ease:

- Create item spacing between items and ignore additional spacing at the top and bottom of the list
- Create item spacing between items and include additional spacing at the top and bottom of the list

hartez:
> @hartez I was just testing `ItemSpacing` on Android and noticed that the spacing is added to the bottom of **each** item, which means the last item has it too. I expected `ItemSpacing` to be spacing between items, ignoring the top of the first item and the bottom of the last item. 

@adrianknight89 
What you're describing is a bug. There should not be space after the last item. Please open an issue.

> Also, I'd like to be able to create equal spacing between items as well as the top of the first item and the bottom of the last item. This seems impossible with the current implementation. I added a `Margin` to `RefreshView`, but the content is cut off at the margins, which is undesirable.

So you want, e.g., a spacing of 5 _between_ items and a spacing of 10 above and below the list? You _should_ be able to do that with a ItemSpacing of 5 and a Margin of 0,10,0,10. If that's not working, please open an issue and we'll take a look.

adrianknight89:
I'll see if I can create a repro for the bug. I was testing it all with my app.

No. I want a spacing of 5 between items and a spacing of 5 above and below the list. Something like this:

UICollectionView-Intro-feature-1

Using Margin on CollectionView seems to have no effect. I could use Margin on RefreshView, but as I said this is not desirable as items do not bleed into the toolbar and navigation bar when I scroll the list up and down because RefreshView has a margin between the bars.

adrianknight89:
See this gif (easier to see if you zoom in). The spacing between items and the spacing on the sides of the CollectionView is the same. As the user scrolls, the content is being cut off by the toolbar **not** by any margin.

hartez:
Oh, you're looking for something like ContentInset. Got it. 

Yes, we need to support that scenario.

rmarinho:
Padding ?

adrianknight89:
Content inset sounds like the right thing to investigate along with (possibly) scroll indicator insets: 

https://stackoverflow.com/a/45890219
https://stackoverflow.com/questions/19844097/uicollectionview-container-padding

hartez:
> Padding ?

It's very similar to Padding, but Forms Padding would have the same problem as a Margin - the gap would _always_ be there. We need the "gap" to scroll with the content.

adrianknight89:
> It's very similar to Padding, but Forms Padding would have the same problem as a Margin - the gap would always be there. We need the "gap" to scroll with the content.

This.

Xamarin.Forms.Platform.iOS/CollectionView/GridViewLayout.cs
```diff
@@ -144,6 +144,9 @@ bool NeedsPartialColumnAdjustment(int section = 0)
 				return false;
 			}
 
+			if (CollectionView.NumberOfSections() == 0)
```
rmarinho: isnt't this from other PR ?

hartez: The other PR took it from here. This wasn't ready to be merged yet, and I wanted to get that fix in because it was crashing.

Xamarin.Forms.ControlGallery.iOS/AppDelegate.cs
```diff
@@ -150,7 +150,7 @@ public override bool FinishedLaunching(UIApplication uiApplication, NSDictionary
 			App.IOSVersion = int.Parse(versionPart[0]);
 
 			Xamarin.Calabash.Start();
-			// Forms.SetFlags("CollectionView_Experimental", "Shell_Experimental");
+			Forms.SetFlags("CollectionView_Experimental");
```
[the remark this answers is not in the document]
hartez: My bad. Fixed.

jsuarezruiz:
I've tried the sample and I can't reproduce your problem @hartez.


The only changes after your changes in the sample is a rebase from 4.3.0.
Should I get the same problem with spacing?.

hartez:
> I've tried the sample and I can't reproduce your problem @hartez.

> The only changes after your changes in the sample is a rebase from 4.3.0.
> Should I get the same problem with spacing?.

Did you turn off center snapping?

jfversluis:
Awesome work picking this up team, thanks! :)

Is there any involvement from me still needed at this time? I see a lot of discussion going on here. Should this be in a final state now? Or do I still need to look at something?

jsuarezruiz:
I have done some tests with the different options of `SnapPointsType`. It seems correct in all the cases. But the branch is outdated, `GridItemsLayout` is no longer used with the CarouselView and `ListItemsLayout` is now `LinearItemsLayout`. Could you rebase 4.3.0?

jfversluis:
@jsuarezruiz Done

rmarinho:
unit test failing is not related

PR #7172, opened 2019-08-15
Force iOS 13 to use Full Screen for Modal

PureWeen:
### Description of Change ###

If running iOS13 then force the ModalPresentationStyle to Full Screen since XF can't handle swipe dismiss. Once we are on xCode 11 and have the necessary APIs then we can more easily react to swipe dismiss

### Platforms Affected ### 
- iOS

### Testing Procedure ###
Make sure Modal pages work on iOS as expected

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.iOS/ModalWrapper.cs
+++ b/Xamarin.Forms.Platform.iOS/ModalWrapper.cs
@@ -14,8 +14,8 @@ namespace Xamarin.Forms.Platform.iOS
 			_modal = modal;
 
 			var elementConfiguration = modal.Element as IElementConfiguration<Page>;
-			if (elementConfiguration?.On<PlatformConfiguration.iOS>().ModalPresentationStyle() == PlatformConfiguration.iOSSpecific.UIModalPresentationStyle.FormSheet)
-				ModalPresentationStyle = UIKit.UIModalPresentationStyle.FormSheet;
+			var modalPresentationStyle = elementConfiguration?.On<PlatformConfiguration.iOS>()?.ModalPresentationStyle() ?? PlatformConfiguration.iOSSpecific.UIModalPresentationStyle.FullScreen;
+			ModalPresentationStyle = modalPresentationStyle.ToNativeModalPresentationStyle();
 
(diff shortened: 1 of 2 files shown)


PureWeen:
Failing ios10 test unrelated

samhouts:
@PureWeen I don't see anything specific to iOS13 here.

Xamarin.Forms.Platform.iOS/ModalWrapper.cs
```diff
@@ -14,8 +14,8 @@ internal ModalWrapper(IVisualElementRenderer modal)
 			_modal = modal;
 
 			var elementConfiguration = modal.Element as IElementConfiguration<Page>;
-			if (elementConfiguration?.On<PlatformConfiguration.iOS>().ModalPresentationStyle() == PlatformConfiguration.iOSSpecific.UIModalPresentationStyle.FormSheet)
-				ModalPresentationStyle = UIKit.UIModalPresentationStyle.FormSheet;
+			var modalPresentationStyle = elementConfiguration?.On<PlatformConfiguration.iOS>()?.ModalPresentationStyle() ?? PlatformConfiguration.iOSSpecific.UIModalPresentationStyle.FullScreen;
+			ModalPresentationStyle = modalPresentationStyle.ToNativeModalPresentationStyle();
```
PureWeen: @samhouts the change here forces *ModalPresentationStyle* to explicitly be set to FullScreen by default where as before *ModalPresentationStyle* would only get set if you set it to *FormSheet*

on iOS12 *ModalPresentationStyle* is set to *FullScreen* by default but on iOS13 it's set to *Automatic* (Which is a new property on iOS13)  

*Automatic* causes the *ModalPresentationStyle* to be FullScreen on ios12 but PageSheet on iOS13

By just setting the *ModalPresentationStyle* ourselves explicitly it will be FullScreen by default on iOS13 which makes it so users can't dismiss the *Modal* and break everything

samhouts: Perfect! That's the blurb I was hoping for. Thanks!

PR #7182, opened 2019-08-16
[Enhancement] Shell Navigation with Models

### LINKED ISSUE #6848: [Enhancement] Shell Navigation with Models
## Summary
Use a Model Class instead of using Query Property to pass data to a page.

## API Changes

In order to facilitate the new _Navigation_ api, an _object_ is added to the parameters of the _GoToAsync_ call.

    var model = new MyModel();
    await Shell.Current.GoToAsync("PAGE_URI", model); // call GoToAsync() using a Model Class

## Intended Use Case
This new feature could be useful for passing many complex data between pages, without having to pass all the data as strings.

groege:
### Description of Change ###
Allows to pass an object parameter via Shell.GoToAsync, which can be used via overriding the OnParameter method of the Page class.

First pull request for Xamarin, so I'm glad for any feedback - Thanks!

### Issues Resolved ### 

- fixes #6848

### API Changes ###

Added:
- virtual void Page.OnParameter(object parameter)
- static Task Shell.GoToParametrized(string location, object parameter, bool animate = true)

Changed:
- ShellNavigationState(string location) => ShellNavigationState(string location, object parameter = null)
- ShellNavigationState(Uri location) => ShellNavigationState(Uri location, object parameter = null)

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###

Use GoToAsync with ShellNavigationState containing an object or call GoToParametrizedAsync
Only built the Core project, having issues using it from an actual Android project, therefore couldn't test it.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core/Page.cs
+++ b/Xamarin.Forms.Core/Page.cs
@@ -261,18 +261,18 @@ namespace Xamarin.Forms
 
 			application.PopCanceled -= handler;
 			return !canceled;
-        }
+		}
 
-        internal void SendParameter(object parameter)
-        {
-            OnParameter(parameter);
-        }
+		internal void SendParameter(object parameter)
+		{
+			OnParameter(parameter);
+		}
 
-        protected virtual void OnParameter(object parameter)
-        {
-        }
+		protected virtual void OnParameter(object parameter)
+		{
+		}
 
-        protected override void OnBindingContextChanged()
+		protected override void OnBindingContextChanged()
 		{
 			base.OnBindingContextChanged();
 			foreach (ToolbarItem toolbarItem in ToolbarItems)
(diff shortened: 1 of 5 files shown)


geojorg:
Did it make it to V4.4 ?

jfversluis:
Hey @geojorg thank you for your interest. Since this is not merged yet, this didn't make it into 4.4, sorry.

PureWeen:
@groege Thank you for the PR but I'm going to close this one for now. The approach has a few breaks that aren't backwards compatible and if we introduce model passing we'd need to do it in a similar fashion as how QueryStrings are applied and stored.

PR #7183, opened 2019-08-16
[macOS] Only use default font size for Time/DatePicker to avoid clipping issues

### LINKED ISSUE #6817: [Bug] DatePicker and TimePicker not working on xamarin.forms macOS
### Description
The xamarin.forms DatePicker and TimePicker does not work on macOS. A lot of the numbers on the dates and times is not showing.

I have attached a video showing the problem. In the video I click the up- and down arrows to increase and decrease date and time. I doesn't work if I type in the numbers.

### Steps to Reproduce
The Visual Studio solution can be found on GitHub:
https://github.com/ekrogh/DateTimeTestProj.git

1. 
2. 
3. 

### Expected Behavior
on xamarin.forms macOS that all digits in dates and times are showing

### Actual Behavior
Please watch the video
[ScreenCapture_2019-7-8 02.56.06.zip](https://github.com/xamarin/Xamarin.Forms/files/3368881/ScreenCapture_2019-7-8.02.56.06.zip)

## Basic Information

- Version with issue: xamarin.forms Version 4.1.0.555618
- Last known good version: ?
- IDE: Visual Studio for mac 2019 Version 8.1.4 (build 5)
- Platform Target Frameworks: 
 - macOS

- Nuget Packages:
- Affected Devices: mac

### Screenshots
[ScreenCapture_2019-7-8 02.56.06.zip](https://github.com/xamarin/Xamarin.Forms/files/3368881/ScreenCapture_2019-7-8.02.56.06.zip)
!-- If the issue is a visual issue, please include screenshots showing the problem if possible -->

### Reproduction Link
https://github.com/ekrogh/DateTimeTestProj.git

krdmllr:
### Description of Change ###

Using a custom font size leads to clipping issue in the date and time picker since it uses a constant size for the separators. This change enforces to use the default font size for these controls to avoid the clipping issues.

### Issues Resolved ###  

- fixes #6817

### API Changes ###

None

### Platforms Affected ### 

-macOS

- FontSize does not longer apply to Date/TimePicker on macOS

Old: 
grafik
-> This should show 23.11.2019
grafik
-> This should show 12:59:59

New
grafik
grafik

### Testing Procedure ###

- enter double digit times/dates

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.MacOS/Renderers/TimePickerRenderer.cs
+++ b/Xamarin.Forms.Platform.MacOS/Renderers/TimePickerRenderer.cs
@@ -94,7 +94,7 @@ namespace Xamarin.Forms.Platform.MacOS
 		}
 
 		void UpdateFont()
-		{ 
+		{
 			if (Control == null || Element == null)
 				return;
 
@@ -105,7 +105,7 @@ namespace Xamarin.Forms.Platform.MacOS
 			// Recreate the font with the default size
 			newFont = NSFont.FromDescription(newFont.FontDescriptor, originalFontSize.FloatValue);
 
-			// Apply the font 
+			// Apply the font
 			Control.Font = newFont;
 		}
 
--- a/Xamarin.Forms.Platform.MacOS/Renderers/TimePickerRenderer.cs
+++ b/Xamarin.Forms.Platform.MacOS/Renderers/TimePickerRenderer.cs
@@ -98,7 +98,18 @@ namespace Xamarin.Forms.Platform.MacOS
 			if (Control == null || Element == null)
 				return;
 
-			Control.Font = Element.ToNSFont();
+			if (Control == null || Element == null)
+				return;
+
+			var newFont = Element.ToNSFont();
+
+			// The font needs to have the default font size to avoid clipping
+			var originalFontSize = (NSNumber)Control.Font.FontDescriptor.FontAttributes[NSFont.SizeAttribute];
+			// Recreate the font with the default size
+			newFont = NSFont.FromDescription(newFont.FontDescriptor, originalFontSize.FloatValue);
+
+			// Apply the font 
+			Control.Font = newFont;
 		}
 
 		void UpdateTime()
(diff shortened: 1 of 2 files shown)


jfversluis:
I'm not entirely up-to-date with all the Mac OS APIs etc. but is this the right way to got?

If you are enlarging your font, I can imagine people will expect this to be bigger as well. Or at least have some way of making it bigger. With this change people will just be stuck with the default font-size forever, right?

Isn't there a way to make it show up the right way when fonts get bigger?

krdmllr:
As you can see in the screenshots is that even though the view gets stretched horizontally, the value does not get more width. I cannot find documentation which describes the layout sizing behavior but I think the native view is not supposed to get resized. I tried a bit but could not find a way to modify the size manually so that you can adapt it based on the font size.

I would argue that it is a limitation of the platform you have to account for when using it.

rmarinho:
What happens to the control if you set a bigger Frame? does it show everything?

krdmllr:
@rmarinho I did some more testing and found that increasing the height of the view does in fact fix the error, perhaps because the view has enough space to scale?
grafik
This is fontsize 50, HeightRequest 400

Additionally, the native view only increases its view to fit the font size so a weird blank gap is above the view.
So I think it would make sense to ignore the HeightRequest and base the height on the FontSize?

I just don't know where I could set the size correctly. I tried to set the frame/bounds in Layout() and GetDesiredSize() which did increase the view size but the control does not changes the rendering.

Xamarin.Forms.Platform.MacOS/Renderers/TimePickerRenderer.cs
```diff
@@ -94,11 +94,19 @@ void HandleValueChanged(object sender, NSDatePickerValidatorEventArgs e)
 		}
 
 		void UpdateFont()
-		{
+		{
```
rmarinho: extra space, can you fix formating .

jfversluis:
I think a rebase might be needed to pick up on the latest build configuration changes and make the build succeed again

krdmllr:
@jfversluis done!

jfversluis:
Congratulations you are now a proud contributor!

PR #7200, opened 2019-08-17
Android X tests

PureWeen:
### Description of Change ###

- Set CG to use AndroidX migration libraries https://github.com/xamarin/Xamarin.Forms/pull/7200/files#diff-f0d93698c5f1681c6d6aa0451b8f7782R367
- Set CG and binaries up to build against Android v10.0


PureWeen:
@rmarinho when you get a chance can you set this pipeline up to run XA >= 9.4

build.cake
```diff
@@ -93,6 +95,14 @@ Task("Restore")
     });
 
 
+
+Task("Boots")
```
PureWeen: I couldn't get the dotnet CLI version to work on mac so I just used cake

I think maybe this would fix the dotnet CLI
https://github.com/dotnet/cli/issues/9114#issuecomment-494226139

build/steps/build-windows.yml
```diff
@@ -52,7 +56,7 @@ jobs:
         solution: ${{ parameters.slnPath }}
         platform: '$(BuildPlatform)'
         configuration: '$(BuildConfiguration)'
-        msbuildArguments: ${{ parameters.msbuildExtraArguments }}
+        msbuildArguments: ${{ parameters.msbuildExtraArguments }} /p:JavaSdkDirectory="$(JAVA_HOME_8_X64)"
```
PureWeen: https://developercommunity.visualstudio.com/content/problem/606218/error-xa5300-the-java-sdk-directory-could-not-be-f.html

rmarinho:
Nice , seems you got this building, but i don't think we should rely on cake for this.

PureWeen:
@rmarinho yea I just used cake because I couldn't get the dotnet cli to work on MacOS

I can try the workaround here https://github.com/dotnet/cli/issues/9114#issuecomment-494226139 but I know you are also concerned about using boots at all :-) so I'll just wait until we decide to use boots or a different method for installing/pinning XA versions as part of the YAML

Issue #2590, opened 2018-05-02
CompressedLayout.SetIsHeadless will renderer twice on iOS platform.


jessejiang0214:
### Description
I try to use CompressedLayout.SetIsHeadless on my App, I find some time it will call custom renderer twice in some situation. As it called twice OnElementChanged in some custom control and that control cannot show the right UI, like it added subview twice. And this only happens on iOS.

### Steps to Reproduce

1. Create some custom control 
2. Put it the custom control into a StackLayout set CompressedLayout.SetIsHeadless is true for StackLayout
3. Put the StackLayout into a Grid and the Grid set CompressedLayout.SetIsHeadless as false.
4. Print out something in ElementChanged.

### Expected Behavior
It should print once.

### Actual Behavior
iOS print twice, but Android print once.
### Basic Information

- Version with issue: 2.5.1.444934
- Last known good version: Unknow
- IDE: Visual Studio for Mac
- Platform Target Frameworks: 
  - iOS:  11.9.1.24
  - Android: 8.2.0.16 
  - UWP:   
- Android Support Library Version: 
- Nuget Packages:
- Affected Devices:

### Reproduction Link

[Demo project](https://www.dropbox.com/s/dqsonznv5qlaglt/LayoutCompressionIssues.zip?dl=0)

aimore:
Any news?

samhouts:
This issue doesn't seem to have had any activity in a long time. We're working on prioritizing issues and resolving them as quickly as we can. To help us get through the list, we would appreciate an update from you to let us know if this is still affecting you on the latest version of Xamarin.Forms, since it's possible that we may have resolved this as part of another related or duplicate issue. If we don't see any new activity on this issue in the next 30 days, we'll evaluate whether this issue should be closed. Thank you!

samhouts:
Since we haven't heard from you in more than 30 days, we hope this issue is no longer affecting you. If it is, please reopen this issue and provide the requested information so that we can look into it further. Thank you!

PR #7201, opened 2019-08-18
[iOS] Fix LayoutCompression Performance Issues #3475

### LINKED ISSUE #3475: [iOS] LayoutCompression Performance Issues
On iOS, using `CompressedLayout.SetIsHeadless` on multiple elements slows down the page. For example, adding 150 labels to a `Page` (each label wrapped in a `StackLayout`) takes quite a while to render:
1. The page pushed from "Without LayoutCompression" button renders in under 2 seconds
2. The page pushed from "With LayoutCompression" button renders in about 13 seconds. The only difference between the pages is that `LayoutCompression` is set on each `StackLayout`

The same issue seems to be described here: https://forums.xamarin.com/discussion/107099/layout-compression-slowing-down-pages

_Without LayoutCompression:_
without_layoutcompression_01

_With LayoutCompression:_
with_layoutcompression_01

### Expected Behavior
Layout compression should improve page rendering performance. On Android both versions (with/without `LayoutCompression` set) render in under 1 second on my phone, with slightly better performance when setting `LayoutCompression`.

### Basic Information

- Version with issue: : Xamarin.Forms 3.1.0.637273
- Last known good version: -
- IDE:
- Platform Target Frameworks:
  - iOS:  : 11.4
- Affected Devices: Tested on iPhoneX Simulator (iOS 11.4) and iPhone 5C (iOS 10.3.3)

### Reproduction Link
[LayoutCompression.zip](https://github.com/xamarin/Xamarin.Forms/files/2256873/LayoutCompression.zip)

GiampaoloGabba:
### Description of Change ###

While inspecting this issue in iOS, i noticed that the `Render` property in the `VisualElementPackager` got multiple subscriptions to `OnRendererElementChanged` when using `CompressedLayout.IsHeadless = True`.

So if we had 150 stacklayout with `CompressedLayout.IsHeadless = True`, the `Render` property, get 150 subscriptions to `OnRendererElementChanged`, slowing down a lot the page.

To solve this, i added an optional boolean property in the `VisualElementPackager` constructor, when is set to `True` (for Headless Layouts) we skip the event subscription to `OnRendererElementChanged`.
Default is `False`.

|Compression mode|Time to render|
| -------------------  | :------------------ |
|No compression|1160ms|
|With Compression before the fix|17268ms|
|**With Compression after the fix**|**745ms**|

### Issues Resolved ### 

- fixes #3475

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS

None

### Testing Procedure ###

Included specific page in control gallery, with UITest

### PR Checklist ###

- [x] Has automated tests

### DIFF
--- a/Xamarin.Forms.Platform.iOS/VisualElementPackager.cs
+++ b/Xamarin.Forms.Platform.iOS/VisualElementPackager.cs
@@ -20,13 +20,15 @@ namespace Xamarin.Forms.Platform.MacOS
 		{
 		}
 
-		VisualElementPackager(IVisualElementRenderer renderer, VisualElement element)
+		VisualElementPackager(IVisualElementRenderer renderer, VisualElement element, bool isHeadless = false)
 		{
 			if (renderer == null)
 				throw new ArgumentNullException(nameof(renderer));
 
 			Renderer = renderer;
-			renderer.ElementChanged += OnRendererElementChanged;
+			if (!isHeadless)
+				renderer.ElementChanged += OnRendererElementChanged;
+
 			SetElement(null, element ?? renderer.Element);
 		}
 
@@ -88,7 +90,7 @@ namespace Xamarin.Forms.Platform.MacOS
 			Performance.Start(out string reference);
 			if (CompressedLayout.GetIsHeadless(view))
 			{
-				var packager = new VisualElementPackager(Renderer, view);
+				var packager = new VisualElementPackager(Renderer, view, isHeadless:true);
 				view.IsPlatformEnabled = true;
 				packager.Load();
 			}
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3475.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue3475.cs
@@ -16,7 +16,7 @@ namespace Xamarin.Forms.Controls.Issues
 	[NUnit.Framework.Category(Core.UITests.UITestCategories.Layout)]
 #endif
 	[Preserve(AllMembers = true)]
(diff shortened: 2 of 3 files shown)


GiampaoloGabba:
Note:
This should fix #2590 but i havent tested that repro but reading the issue i think that is caused by the multiple OnelementChange subscription for the renderer

Xamarin.Forms.Platform.iOS/VisualElementPackager.cs
```diff
@@ -20,13 +20,15 @@ public VisualElementPackager(IVisualElementRenderer renderer) : this(renderer, n
...
 
-		VisualElementPackager(IVisualElementRenderer renderer, VisualElement element)
+		VisualElementPackager(IVisualElementRenderer renderer, VisualElement element, bool isHeadless = false)
 		{
 			if (renderer == null)
 				throw new ArgumentNullException(nameof(renderer));
 
 			Renderer = renderer;
-			renderer.ElementChanged += OnRendererElementChanged;
+			if (!isHeadless)
```
StephaneDelcroix: another option would be to keep a ref to the EventHandler, I'm quite sure you can't add it multiple times

StephaneDelcroix:
Thanks for this first contribution , and the investigation you've done to isolate this issue 🎉

GiampaoloGabba:
@StephaneDelcroix you are welcome!
It always bugged me that LayoutCompression made my ios layout slower, I like to micromanage everything to make performance as better as possible :) so i decided to bring a little contribution to the community and hunt down the problem.

if you prefer to modify my pull request with an EventHandler reference feel free to do it. In these days i'm a bit busy porting my SharedTransition plugin to Shell and i dont have much time. The changes in code are minimal, the compelling work was the "hunting" the issue :)

Issue #5593, opened 2019-03-17
Uwp Xamarin.Forms.Shell 


allessandrosj:
Uwp Xamarin.Forms.Shell  does not work correctly in uwp, the same way it works on android and ios

pauldipietro:
Shell support is only provided for Android and iOS at this time. We are continuously evaluating feedback regarding potential future UWP support and will provide more information in the future as it becomes available.

jcmontoya:
I normally use iOS and Android for my phone users and UWP for my tablet and desktop users. So having UWP support will be very important to keep windows users engaged.

TiBall:
What?? MS is really ditching their own platform. This is sad to hear!

Juansero29:
Isn't there any visibility on when UWP support will be given? Or has the decision already been made?

snow-jallen:
Shell support for UWP - please!

akuehntopf:
I'd also like to see UWP support for the Shell. 

For my company UWP is quite important because our customers use Windows Tablets in combination with some Android phones to get their work done. We won't be able to migrate to Shell if UWP support is missing.

dan-meier:
Agreed -- UWP compatibility allows me to bake in Windows tablet capability with my mobile implementation. I'd have to code this completely separately without it. Yuck! UWP compatibility is a must-have!

cbradbaer:
I don't know if it makes sense, because I have not used Shell yet.  But maybe there is a way to use all other elements of the App (besides Shell) for a UWP implementation.  After all using TabbedPage and MasterDetailPage would achieve the same thing, it seems.

JakePorter05:
I also would vote for the ability to use UWP with Shell. In Xamarin.Forms 4.0 the docs no longer say it's just for IOS and android anymore. I got really excited, but the templates don't allow you to use it with UWP?

dgerding:
UWP please... or at least some flavor of Microsoft UX API we can depend on for a few years.  This approach of "we'll let you know if we might support it later" is just mean.  

So, yes, UWP or whatever will support it but more importantly: A COMMITMENT TO WHEN WE'LL KNOW IF OR WHEN.

Can't move to forms without that certitude.

melihercan:
Not only UWP but MacOS as well please.

Mephisztoe:
[...]

PR #7214, opened 2019-08-20
[Android/iOS] RefreshView

### LINKED ISSUE #5882: [Spec] RefreshView
## RefreshView

Part of [CollectionView](https://github.com/xamarin/Xamarin.Forms/issues/3172)

Represents a container control that provides refresh functionality for its content. 

# API

```csharp
public class RefreshView : ContentView
{
	public static readonly BindableProperty CommandProperty;
	public ICommand Command { get; set; }

	public static readonly BindableProperty CommandParameterProperty;
	public object CommandParameter { get; set; }

	public static readonly BindableProperty IsEnabledProperty;
	public bool IsEnabled { get; set; }

	public bool IsRefreshing { get; set; }

	public Color BackgroundColor {get; set;}
	public Color RefreshColor {get; set;}
}
```

## UWP Platform Specific

UWP is the only platform that has this concept implemented natively so for now we will just implement a platform specific for UWP users that won't to be able to change this

```C#
public enum RefreshPullDirection  PullDirection 
```

### Properties

| API | Description |
| Command | Gets or sets the command to execute when this item is invoked. |
| CommandParameter | Optional parameter passed to the refresh command. |
| IsEnabled | Gets or sets a value indicating whether the user can interact with RefreshView. |
| IsRefreshing | This indicates the current state of the refresh control. Developers will need to set this back to false to indicate that a refresh has completed. If this is set to true it will trigger the command and the refresh visualization on the platform. This will automatically transition to true when the user has performed the required actions to trigger the platforms Refresh Control |
| BackgroundColor | iOS - Background Color of the UIView that contains the progress circle. Android- Background Color of the progress circle  | 
| RefreshColor| Color of the circle itself | 

## RefreshPullDirection

```csharp
public enum RefreshPullDirection
{
	LeftToRight,
	TopToBottom,
	RightToLeft,	
	BottomToTop
}
```

### Enum Values

| API | Description | Notes |
| ------------- | ------------- | ------------- |
| LeftToRight | The user must pull from left to right to refresh. | May be pushed to a future phase. |
| TopToBottom | The user must pull from top to bottom to refresh. |  |
| RightToLeft | The user must pull from right to left to refresh. | May be pushed to a future phase. |
| BottomToTop | The user must pull from bottom to top to refresh. | May be pushed to a future phase. |

### Examples

```XAML
<RefreshView BackgroundColor="Green" RefreshColor="Pink">
```

#### UWP
image

#### Android
image

#### iOS
image

PureWeen:
### Description of Change ###

[RefreshView](https://github.com/xamarin/Xamarin.Forms/issues/5882) represents a container control that provides **pull-to-refresh** functionality for **scrollable** content.

fixes #5882

### Screenshots

Android
refreshview-droid

iOS
refreshview-ios2

### API Changes ###

```
public class RefreshView : ContentView
{
	public static readonly BindableProperty CommandProperty;
	public ICommand Command { get; set; }

	public static readonly BindableProperty CommandParameterProperty;
	public object CommandParameter { get; set; }

	public static readonly BindableProperty IsEnabledProperty;
	public bool IsEnabled { get; set; }

	public bool IsRefreshing { get; set; }

	public Color BackgroundColor {get; set;}
	public Color RefreshColor {get; set;}
}
```
 
### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android

None

Not applicable

### Testing Procedure ###
In Core Gallery access to the RefreshView samples. Try to pull to refresh the content. A load indicator should appear and refresh the content.

### PR Checklist ###

- [ ] Has automated tests 


.nuspec/Xamarin.Forms.targets
```diff
@@ -15,6 +15,7 @@
 		<_DefaultXamlItemsEnabled>False</_DefaultXamlItemsEnabled>
 		<EnableDefaultCssItems Condition="'$(EnableDefaultCssItems)'==''">True</EnableDefaultCssItems>
 		<_DefaultCssItemsEnabled>False</_DefaultCssItemsEnabled>
+    <SkipMicrosoftUIXamlCheckTargetPlatformVersion Condition="'$(SkipMicrosoftUIXamlCheckTargetPlatformVersion)'==''">True</SkipMicrosoftUIXamlCheckTargetPlatformVersion>
```
PureWeen:
```suggestion
		<SkipMicrosoftUIXamlCheckTargetPlatformVersion Condition="'$(SkipMicrosoftUIXamlCheckTargetPlatformVersion)'==''">True</SkipMicrosoftUIXamlCheckTargetPlatformVersion>
```

Xamarin.Forms.Platform.iOS/Renderers/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,249 @@
...
+		protected override void OnElementChanged(ElementChangedEventArgs<RefreshView> e)
+		{
+			base.OnElementChanged(e);
+
+			if (e.OldElement != null || Element == null)
+				return;
+
+			_refreshControl = new UIRefreshControl();
+
+			_refreshControl.ValueChanged += OnRefresh;
```
adrianknight89: Do we need to unsubscribe this event if `e.OldElement` is not null?

Xamarin.Forms.Platform.iOS/Renderers/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,249 @@
...
+			else if (e.IsOneOf(RefreshView.RefreshColorProperty, VisualElement.BackgroundColorProperty))
+				UpdateColors();
+		}
+
+		protected override void SetBackgroundColor(Color color)
+		{
+			if (_refreshControl == null)
+				return;
+
+			if (color != Color.Default)
```
adrianknight89: To make it simpler, we can use a ternary operator here.

Xamarin.Forms.Platform.iOS/Renderers/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,249 @@
...
+				_refreshControl.BackgroundColor = color.ToUIColor();
+			else
+				_refreshControl.BackgroundColor = null;
+		}
+
+		protected override void Dispose(bool disposing)
+		{
+			base.Dispose(disposing);
+
+			if (disposing)
```
adrianknight89: reduced if nesting here. Also, introduce `_disposed` and set it true. See other renderers. We should use a similar pattern.

Xamarin.Forms.Platform.iOS/Renderers/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,249 @@
...
+					_refreshControl = null;
+				}
+
+				_refreshControlParent = null;
+			}
+		}
+
+		bool TryOffsetRefresh(UIView view, bool refreshing)
+		{
+			if (view is UITableView)
```
adrianknight89: Can we remove the UITableView stuff? ListView has a built-in refresh control already.

Xamarin.Forms.Platform.iOS/Renderers/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,249 @@
...
+
+			SetBackgroundColor(Element.BackgroundColor);
+		}
+
+		void UpdateIsRefreshing()
+		{
+			IsRefreshing = Element.IsRefreshing;
+		}
+
+		void UpdateIsEnabled()
```
adrianknight89: So many curly braces in this method. Maybe we could remove some to make it smaller.

Xamarin.Forms.Platform.Android/Renderers/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,279 @@
...
+			_renderer.View.Measure(widthConstraint, heightConstraint);
+
+			return new SizeRequest(new Size(_renderer.View.MeasuredWidth, _renderer.View.MeasuredHeight), new Size(40, 40));
+		}
+
+		public void UpdateLayout() => Tracker?.UpdateLayout();
+
+		protected override void Dispose(bool disposing)
+		{
+			base.Dispose(disposing);
```
adrianknight89: Same dispose stuff.

Xamarin.Forms.Platform.iOS/Renderers/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,243 @@
+﻿using System;
+using System.ComponentModel;
+using UIKit;
+
+namespace Xamarin.Forms.Platform.iOS
+{
+	public class RefreshViewRenderer : ViewRenderer<RefreshView, UIView>
```
adrianknight89: Android registers an effect provider. Does iOS need to do the same?

jsuarezruiz: Yes, I just included it.

adrianknight89:
If you look at the iOS Outlook app, they have a much nicer looking refresh control. I find the stock UIRefreshControl rather unappealing. Xamarin should figure out a way to support custom refresh designs in the future.

ChasakisD:
Since iOS MaterialComponents are now available to Visual Material, UIRefreshControl could also support MDActivityIndicator. The integration of the Indicator with the UIRefreshControl is pretty easy.

Xamarin.Forms.Core/RefreshView.cs
```diff
@@ -0,0 +1,88 @@
...
+        public static readonly BindableProperty IsRefreshingProperty =
+            BindableProperty.Create(nameof(IsRefreshing), typeof(bool), typeof(RefreshView), false);
+			
+        public bool IsRefreshing
+        {
+            get { return (bool)GetValue(IsRefreshingProperty); }
+            set
+            {
+                if ((bool)GetValue(IsRefreshingProperty) == value)
+                    OnPropertyChanged(nameof(IsRefreshing));
```
adrianknight89: A bit confused with this line. Shouldn't we call `OnPropertyChanged` only after the property changes? 

```
var oldValue = IsRefreshing;

SetValue(IsRefreshingProperty, value);

if(oldValue != IsRefreshing)
   OnPropertyChanged(nameof(IsRefreshing));
```

Also, doesn't a bindable property already have change notification built in? Why use `OnPropertyChanged` again?

Xamarin.Forms.Core/RefreshView.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+        
+        public static readonly BindableProperty CommandProperty =
+            BindableProperty.Create(nameof(Command), typeof(ICommand), typeof(RefreshView));
+			
+        public ICommand Command
+        {
+            get { return (ICommand)GetValue(CommandProperty); }
+            set { SetValue(CommandProperty, value); }
+        }
```
adrianknight89: You could use the lambda operator for trival get / set calls.

```
get => (ICommand)GetValue(CommandProperty);
set => SetValue(CommandProperty, value);
```

Other properties can use the same as well.

Xamarin.Forms.Core/RefreshView.cs
```diff
@@ -0,0 +1,88 @@
...
+				
+        public object CommandParameter
+        {
+            get { return GetValue(CommandParameterProperty); }
+            set { SetValue(CommandParameterProperty, value); }
+        }
+        
+        void RefreshCommandCanExecuteChanged(object sender, EventArgs eventArgs)
+        {
+            ICommand cmd = Command;
```
adrianknight89: No need for this local variable. Use `Command` instead.

Xamarin.Forms.Core/RefreshView.cs
```diff
@@ -0,0 +1,88 @@
...
+            if (cmd != null)
+                base.IsEnabled = cmd.CanExecute(CommandParameter);
```
adrianknight89: Not sure why `base` is needed?

Xamarin.Forms.Core/RefreshView.cs
```diff
@@ -0,0 +1,88 @@
...
+            set { SetValue(RefreshColorProperty, value); }
+        }
+        
+        protected override SizeRequest OnMeasure(double widthConstraint, double heightConstraint)
+        {
+            if (Content == null)
+                return new SizeRequest(new Size(100, 100));
+
+            return base.OnMeasure(widthConstraint, heightConstraint);
+		}
```
adrianknight89: Indentation is off.

Xamarin.Forms.Core/RefreshView.cs
```diff
@@ -0,0 +1,88 @@
...
+
+		public IPlatformElementConfiguration<T, RefreshView> On<T>() where T : IConfigPlatform
```
adrianknight89: Same here.

Xamarin.Forms.Platform.Android/Renderers/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,284 @@
...
+			{
+				_refreshing = value;
+
+				if (RefreshView != null && RefreshView.IsRefreshing != _refreshing)
+					RefreshView.IsRefreshing = _refreshing;
+
+				if (base.Refreshing == _refreshing)
+					return;
+
+				base.Refreshing = _refreshing;
```
adrianknight89: To make this simpler, can't we do the following?

```
if(RefreshView != null)
   RefreshView.IsRefreshing = _refreshing;

base.Refreshing = _refreshing;
```

In other words, the extra equality checks may be unnecessary since setting a property to its current value should have minimal impact, no?

Xamarin.Forms.Platform.Android/Renderers/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,284 @@
...
+				return;
+
+			if (_renderer != null)
+			{
+				_renderer.View.RemoveFromParent();
+				_renderer.Dispose();
+				_renderer = null;
+			}
+
+			if (RefreshView.Content != null)
```
adrianknight89: Line 98 checks for null. Is this if check needed? Not sure if lines 101 - 105 effect the value of `RefreshView.Content`.

Xamarin.Forms.Platform.Android/Renderers/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,284 @@
...
+			{
+				_renderer = Platform.CreateRenderer(RefreshView.Content, Context);
+
+				Platform.SetRenderer(RefreshView.Content, _renderer);
+
+				if (_renderer.View.Parent != null)
+					_renderer.View.RemoveFromParent();
+
+				using (var layout = new LayoutParams(LayoutParams.MatchParent, LayoutParams.MatchParent))
```
adrianknight89: `layoutParams` is probably a better name than `layout` as the latter implies an actual layout. Also, I'm not sure why we need a `using` here. Can't we pass the `LayoutParams` object without disposing it?

Xamarin.Forms.Platform.Android/Renderers/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,284 @@
...
+		void UpdateIsEnabled() => SwipeRefreshLayout.Enabled = RefreshView.IsEnabled;
+
+		bool CanScrollUp(AView view)
+		{
+			if (!(view is ViewGroup viewGroup))
+				return base.CanChildScrollUp();
+
+			if (Forms.SdkInt < BuildVersionCodes.JellyBean)
+			{
+				if (viewGroup.IsScrollContainer)
```
adrianknight89: These two ifs can be merged together.

Xamarin.Forms.Platform.Android/Renderers/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,284 @@
...
+			if (view is global::Android.Widget.ScrollView)
+			{
+				var scrollview = view as global::Android.Widget.ScrollView;
+				return scrollview.ScrollY < 0;
+			}
+			if (view is NestedScrollView)
+			{
+				var nestedScrollView = view as NestedScrollView;
+				return nestedScrollView.ScrollY < 0;
+			}
```
adrianknight89: There should be a newline between these four if blocks. Also, we can use the `is` type expression to get rid of explicit variable declarations. See line 139.

Xamarin.Forms.Platform.Android/Renderers/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,284 @@
...
+
+			ElementPropertyChanged?.Invoke(sender, e);
+
+		}
+
+		public SizeRequest GetDesiredSize(int widthConstraint, int heightConstraint)
+		{
+			_renderer.View.Measure(widthConstraint, heightConstraint);
+
+			return new SizeRequest(new Size(_renderer.View.MeasuredWidth, _renderer.View.MeasuredHeight), new Size(40, 40));
```
adrianknight89: I think Core requests (100, 100) at minimum.

Xamarin.Forms.Platform.Android/Renderers/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,284 @@
...
+					_renderer.View.RemoveFromParent();
+			}
+
+			_renderer?.Dispose();
+			_renderer = null;
+
+			Tracker?.Dispose();
+			Tracker = null;
+
+			_init = false;
```
adrianknight89: Shouldn't lines 250 - 256 be nested inside the `disposing` if?

Xamarin.Forms.Core/RefreshView.cs
jsuarezruiz: This is not necessary. The BindableProperty already notifies the change.

Xamarin.Forms.Platform.Android/Renderers/RefreshViewRenderer.cs
jsuarezruiz: Cut off the flow before, but nothing heavy is done afterwards, with the line 101 is enough. Removed that validation.

Xamarin.Forms.Platform.Android/Renderers/RefreshViewRenderer.cs
jsuarezruiz: Right. Changed.

PureWeen:
We need to verify if this will be an issue on VS2017 

https://github.com/xamarin/Xamarin.Forms/issues/5593#issuecomment-524555562

Before merging this in

Xamarin.Forms.Core/PlatformConfiguration/WindowsSpecific/RefreshView.cs
```diff
@@ -0,0 +1,40 @@
...
+	{
+		public enum RefreshPullDirection
+		{
+			LeftToRight,
+			TopToBottom,
+			RightToLeft,
+			BottomToTop
+		}
+
+		public static readonly BindableProperty RefreshPullDirectionProperty = BindableProperty.Create("RefreshPullDirection", typeof(RefreshPullDirection), typeof(FormsElement), RefreshPullDirection.TopToBottom);
```
jfversluis: I see only the enum was added to the "New APIs" in the description. Do we need to include the property itself as well?

Xamarin.Forms.Platform.iOS/Renderers/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,267 @@
...
+			else
+			{
+				if (_refreshControl.Superview != null)
+					_refreshControl.RemoveFromSuperview();
+			}
+		}
+
+		bool CanUseRefreshControlProperty()
+		{
+			return Forms.IsiOS10OrNewer && !_usingLargeTitles;
```
jfversluis: On [line 45](https://github.com/xamarin/Xamarin.Forms/pull/7214/files#diff-b2706bb9ae1c67c18d702987c43c1ed6R45) you're checking against iOS11 and here it's 10. Shouldn't this be the same?

PureWeen:
@hartez 

> Where's the UWP implementation?

Has will show up in  separate PR so we can workout some WinUI issues and not hold up iOS/Android

Xamarin.Forms.Core/PlatformConfiguration/WindowsSpecific/RefreshView.cs
jsuarezruiz: Yes, I added it to the "New APIs" description.

Xamarin.Forms.ControlGallery.Android/Properties/AndroidManifest.xml
```diff
@@ -18,6 +18,6 @@
 	<uses-feature android:glEsVersion="0x00020000" android:required="true" />
 	<application android:label="AndroidControlGallery" android:supportsRtl="true">
 		<meta-data android:name="com.google.android.gms.version" android:value="@integer/google_play_services_version" />
-    <uses-library android:name="org.apache.http.legacy" android:required="false" />
```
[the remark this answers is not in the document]
jsuarezruiz: Sure. Changed.

Xamarin.Forms.Platform.iOS/Renderers/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,267 @@
...
+				_refreshControlParent = null;
+			}
+
+			_isDisposed = true;
+			base.Dispose(disposing);
+		}
+
+		bool TryOffsetRefresh(UIView view, bool refreshing)
+		{
+			if (view is UITableView tableView)
```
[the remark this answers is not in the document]
jsuarezruiz: You are right. Is not necessary at all. I did some test and changed it.

Xamarin.Forms.Platform.iOS/Renderers/RefreshViewRenderer.cs
jsuarezruiz: I removed it. Is not necessary.

jsuarezruiz:
UWP implementation was included. It also had a Platform Specific to set the pull to refresh orientation. 
refreshdirection-uwp

In UWP we used the [Microsoft.UI.Xaml NuGet](https://www.nuget.org/packages/Microsoft.UI.Xaml). After several tests, we found problems to solve. To avoid delaying this functionality we have decided to split into two branches. For now, in this PR we will add the functionality in Android and iOS and as for UWP, we created another branch (https://github.com/xamarin/Xamarin.Forms/blob/swiperefresh-uwp/Xamarin.Forms.Platform.UAP/RefreshViewRenderer.cs) and are in conversation with the WinUI team to find the best solution in UWP.

PR #7271, opened 2019-08-24
Fix RTL behavior for MasterDetailsPage when running on iOS 13 iPads

### LINKED ISSUE #7224: [Bug] MasterDetail Master width issue.
### Description
Hi Team,

We are using an application that is made up of master detail page, On iOS 13 Beta we found that the master section width of the master detail gets expanded on bringing the application to background and again getting it to foreground after 10 sec. 
 
### Steps to Reproduce

1. Launch the application in landscape mode.
2. Bring the application to background and wait for 10sec.
3. Bring the application to foreground, we could notice that the master detail ,master width is expanded.

### Expected Behavior
Master detail master section width remains the same on getting the application to background and foreground.

### Actual Behavior
Master detail master section width expands on getting the application to background and foreground.

### Basic Information

- Version with issue: iOS 13 Beta 7 and Xamarin Preview 4
- Last known good version: iOS 12 and Xamarin Previous versions
- IDE: Visual Studio 8.2.2 (build 44) 
- Platform Target Frameworks: 
  - iOS:  iOS 13 Beta 7

IMG_0019
IMG_0020

- Affected Devices: iPad Air 2 iOS 13 

### Reproduction Link

https://drive.google.com/open?id=1fFpw5p3au2Mbtd3qvEFwvSa6GMwitKW_

### LINKED ISSUE #7266: [Bug] G2818 on iOS13 doesn't move the Master view from right to left and visa versa
### Description

It looks like setting the Flow Direction on the MDP on iOS 13 doesn't move the MasterView correctly from one side to the other

### Steps to Reproduce

1. Run G2818

### Expected Behavior

The masterview should move from right to left and then left to right as you click the link

### Actual Behavior

nothing happens

### Basic Information

- Platform Target Frameworks: 
  - iOS:  xcode 11 beta 6
- Affected Devices: Ipad air (generation 3) ios 13.0

PureWeen:
### Description of Change ###
- When running on iOS13 don't use the width hack when switching from LTR to RTL. It looks like iOS13 fixed whatever issue was causing that width to get set to 0.5 so it's not needed anymore
- When running on iOS 13 if the user changes the FlowDirection of the MDP remove and readd it to the view. From what I can tell the fact that this worked at all in iOS 12 was a bug and may have also been why the width became messed up. 

If you look at the appearance docs
https://developer.apple.com/documentation/uikit/uiappearance

> iOS applies appearance changes when a view enters a window, it doesn’t change the appearance of a view that’s already in a window. To change the appearance of a view that’s currently in a window, remove the view from the view hierarchy and then put it back.

So it seems like this is how it was always supposed to work. iOS 13 made a lot of changes to the UISplitViewController so this new behavior was most likely bundled up with that

### Issues Resolved ### 
- fixes #7224
- fixes #7266 

### Platforms Affected ### 
- iOS

### Testing Procedure ###
Test all of this on an iPad Air iOS 13 because it's the only iPad that demonstrated the odd width pinning issue

- Test issue Issue2818 on an iOS 13 iPad and iOS 12 iPad
- Go to the FlowDirection gallery and just test everything

I added automated tests but they are mostly useless since we don't have iPad tests running in the cloud. Run them all locally to make sure they all pass on iOS 12 and iOS 13. They are disabled for phone devices because of a different already logged issue

### PR Checklist ###

- [X] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2818.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2818.cs
@@ -1,21 +1,41 @@
 ﻿using Xamarin.Forms.CustomAttributes;
 using Xamarin.Forms.Internals;
+using System;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
 
 namespace Xamarin.Forms.Controls.Issues
 {
 	[Preserve(AllMembers = true)]
 	[Issue(IssueTracker.Github, 2818, "Right-to-Left MasterDetail in Xamarin.Forms Hamburger icon issue", PlatformAffected.Android)]
-	public class Issue2818 : MasterDetailPage
+	public class Issue2818 : TestMasterDetailPage
 	{
-		public Issue2818()
+
+		protected override void Init()
 		{
 			FlowDirection = FlowDirection.RightToLeft;
-
 			Master = new ContentPage
 			{
 				Title = "Master",
 				BackgroundColor = Color.SkyBlue,
-				IconImageSource = "menuIcon"
+				IconImageSource = "menuIcon",
+				Content = new StackLayout()
+				{
+					Children =
+					{
+						new Button()
+						{
+							Text = "If you can see me the test has passed",
+							AutomationId = "CloseMasterView",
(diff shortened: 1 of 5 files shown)


PureWeen:
@samhouts retargeted

PureWeen:
All Green!!

PR #7273, opened 2019-08-25
Fix crash with nested CollectionViews on UWP

### LINKED ISSUE #6620: [iOS] Crash when creating a CollectionView inside a CollectionView
### Description
Creating a CollectionView within a CollectionView crashes on iOS

### Steps to Reproduce

1. Create a xaml page with a CollectionView inside a CollectionView
2. Create a class for the child containing sample properties
3.  Create a class for the parent containing sample properties and a list of the child class
4.  Populate the data using Binding
5. Run the app

### Expected Behavior
Page is created and CollectionView parent is populated along side the CollectionView child

### Actual Behavior
App freezes and throws a message on the application output
*** Assertion failure in -[UICollectionView _endItemAnimationsWithInvalidationContext:tentativelyForReordering:animator:], /BuildRoot/Library/Caches/com.apple.xbs/Sources/UIKitCore_Sim/UIKit-3698.103.12/UICollectionView.m:6165

### Basic Information

- Version with issue: Xamarin.Forms 4.0.0.497661
- Last known good version:
- IDE: Visual Studio Mac 2019
- Platform Target Frameworks: 
  - iOS:  12.2

hartez:
Closes #6620

### Description of Change ###

At one point, iOS crashed if you tried to use a CollectionView in the ItemTemplate of another CollectionView. That's no longer the case (I suspect this was fixed as a side effect of changes in 4.1).

But while investigating this, it turned out that UWP also had this issue. So this PR fixes that, and adds an automated test so we can prevent this problem in the future.

### Issues Resolved ### 

- fixes #6620

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS
- UWP

None

Not applicable

### Testing Procedure ###

Control Gallery, navigate to CollectionView Galleries -> Nested CollectionViews. The app should not crash.

### PR Checklist ###

- [x] Has automated tests 
- [x] Memes generated

### DIFF
--- a/Xamarin.Forms.Platform.UAP/CollectionView/ItemsViewRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/CollectionView/ItemsViewRenderer.cs
@@ -97,6 +97,12 @@ namespace Xamarin.Forms.Platform.UWP
 
 			var itemsSource = Element.ItemsSource;
 
+			if (itemsSource == null)
+			{
+				_collectionViewSource = null;
+				return;
+			}
+
 			var itemTemplate = Element.ItemTemplate;
 			if (itemTemplate != null)
 			{
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/NestedCollectionViews.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/NestedCollectionViews.cs
@@ -1,0 +1,41 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.Forms.Core.UITests;
+using Xamarin.UITest;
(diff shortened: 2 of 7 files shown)


Xamarin.Forms.Platform.UAP/CollectionView/ItemsViewRenderer.cs
```diff
@@ -97,6 +97,12 @@ protected virtual void UpdateItemsSource()
 
 			var itemsSource = Element.ItemsSource;
 
+			if (itemsSource == null)
+			{
+				_collectionViewSource = null;
+				return;
```
jsuarezruiz: This return avoid to do the item inflate and binding process, which is correct. But, it will also prevent the ItemsSource from being set here https://github.com/xamarin/Xamarin.Forms/blob/c1eea05b9976b0db00bf155f3754ba1aac072fa6/Xamarin.Forms.Platform.UAP/CollectionView/ItemsViewRenderer.cs
I think it is correct, because EmptyView is shown with Null or empty ItemsSource, but I have doubts, It's like that, right?.

PR #7289, opened 2019-08-27
[Android] 28+ Make non-visible pickers work again

### LINKED ISSUE #5159: [Android] Calling Focus on all Pickers running an API 28 devices no longer opens Picker
### Description
When using API Level 28, calling Focus on a DatePicker with Visible set to false does not show the date picker dialog. 

Prior to this PR https://github.com/xamarin/Xamarin.Forms/pull/4344/files#diff-446294d29d78ca8d84e4c0ddc20bbc07L132 the Picker was opened as part of the *FocusRequestedEvent*. PR #4344 brought DatePicker inline with the other pickers by changing it to trigger a Click on the control which then would trigger the click listener to open the picker. 

On API 28 *FocusRequestedEvent* is still called so there is still an opportunity to react to this. The difference on API 28 is that the Focus request fails so the Click Listener never actually fires. 

Possible fix ideas
- force it to focus anyway. Maybe this is a bad idea? On iOS if the control isn't visible it's still able to be set as a First Responder so this would bring it inline with iOS
- change all the pickers to open the dialog from the *FocusRequestedEvent*. If this is done then make sure to test what #4344 fixed

### Steps to Reproduce

1. Add a date picker and a button on a Forms page. Set the visibility of the date picker to false
2. Use the button to call Focus on the date picker.
3. On API 27 and lower the dialog is shown. Not on API 28.

### Expected Behavior
DatePicker dialog is shown.

### Actual Behavior
Nothing happens.

### Basic Information

- Version with issue: 4.0.0.135214-pre4
- IDE: Visual Studio 2019 Windows
- Platform Target Frameworks: 
  - Android: API 28
- Nuget Packages:
- Affected Devices: Android

### Screenshots
n/a

### Reproduction Link
[Reproduction](https://github.com/varyamereon/FormsDatePickerFocusSample)

### LINKED ISSUE #7311: [Bug] [Android] Error back hardware button with Picker
### Description

On Android, if you use a Picker and press the BACK hardware button, it closes the picker and comes back to the page. If you press the BACK hardware button again, it opens the Picker again. And it does it on a loop.
If you unfocus the Picker, it doesn't do it anymore.

### Steps to Reproduce

1. Create a picker
2. Launch Android app
3. Click on picker
4. Click on hardware back button
5. Click on hardware back button (the picker open)

### Expected Behavior

The second BACK button pressed should not open the Picker.

### Actual Behavior

The second BACK button pressed open the Picker.

### Basic Information

- Version with issue: 4.2.0.709249
- Last known good version: 4.1.0.709244
- IDE: Visual Studio 2019 16.2.3
- Platform Target Frameworks:
  - Android: 9.0
- Affected Devices: Android (tested : 7, 8 and 9)

### Reproduction Link

[PickerAndroidBug.zip](https://github.com/xamarin/Xamarin.Forms/files/3554816/PickerAndroidBug.zip)

jfversluis:
### Description of Change ###

Checks if the TextView control is clickable, if not, takes a different route but still shows the picker dialog.

As a bonus I added the fix for #7311. The back button will not open the Picker again when it is still focused.

### Issues Resolved ### 

- fixes #5159 
- fixes #7311 

### API Changes ###

 
 None

### Platforms Affected ### 

- Android

None

Not applicable

### PR Checklist ###

- [x] Has automated tests 


jfversluis:
Moved the iOS one behind the `__ANDROID__` flag, not needed on iOS.

For the other failing test, not sure what the cause might be. I've added a little code to hide the picker dialog again after the test is done. Don't see how that should be the problem, but let's see.

It seems the app crashes for some reason after the last "picker test" and before the next test which is the failing one you pointed out. It's way too coincidental that this happens now, so my tests are probably the cause, but not sure why.

PureWeen:
> It seems the app crashes for some reason after the last "picker test" and before the next test which is the failing one you pointed out. 

Any unhandled exceptions in the device logs?

jfversluis:
> Any unhandled exceptions in the device logs?

Besides some exceptions (916 times) that do not seem to have to do anything with this, I do see this one:

> I/MonoDroid(23390): System.NotSupportedException: Unable to activate instance of type Xamarin.Forms.Platform.Android.TimePickerRenderer from native handle 0xe9500025 (key_handle 0x42b2bbd8). ---> System.MissingMethodException: No constructor found for Xamarin.Forms.Platform.Android.TimePickerRenderer::.ctor(System.IntPtr, Android.Runtime.JniHandleOwnership) ---> Java.Interop.JavaLocationException: Exception of type 'Java.Interop.JavaLocationException' was thrown.
I/MonoDroid(23390):    --- End of inner exception stack trace ---
I/MonoDroid(23390):   at Java.Interop.TypeManager.CreateProxy (System.Type type, System.IntPtr handle, Android.Runtime.JniHandleOwnership transfer) [0x00055] in <1ee35b7607e5467fa07b155162b2c8be>:0 
I/MonoDroid(23390):   at Java.Interop.TypeManager.CreateInstance (System.IntPtr handle, Android.Runtime.JniHandleOwnership transfer, System.Type targetType) [0x00116] in <1ee35b7607e5467fa07b155162b2c8be>:0 
I/MonoDroid(23390):    --- End of inner exception stack trace ---
I/MonoDroid(23390):   at Java.Interop.TypeManager.CreateInstance (System.IntPtr handle, Android.Runtime.JniHandleOwnership transfer, System.Type targetType) [0x00182] in <1ee35b7607e5467fa07b155162b2c8be>:0 
I/MonoDroid(23390):   at Java.Lang.Object.GetObject (System.IntPtr handle, Android.Runtime.JniHandleOwnership transfer, System.Type type) [0x000c1] in <1ee35b7607e5467fa07b155162b2c8be>:0 
I/MonoDroid(23390):   at Java.Lang.Object._GetObject[T] (System.IntPtr handle, Android.Runtime.JniHandleOwnership transfer) [0x00017] in <1ee35b7607e5467fa07b155162b2c8be>:0 
I/MonoDroid(23390):   at Java.Lang.Object.GetObject[T] (System.IntPtr handle, Android.Runtime.JniHandleOwnership transfer) [0x00000] in <1ee35b7607e5467fa07b155162b2c8be>:0 
I/MonoDroid(23390):   at Java.Lang.Object.GetObject[T] (System.IntPtr jnienv, System.IntPtr handle, Android.Runtime.JniHandleOwnership transfer) [0x00006] in <1ee35b7607e5467fa07b155162b2c8be>:0 
I/MonoDroid(23390):   at Android.App.TimePickerDialog+IOnTimeSetListenerInvoker.n_OnTimeSet_Landroid_widget_TimePicker_II (System.IntPtr jnienv, System.IntPtr native__this, System.IntPtr native_view, System.Int32 hourOfDay, System.Int32 minute) [0x00000] in <1ee35b7607e5467fa07b155162b2c8be>:0 
I/MonoDroid(23390):   at (wrapper dynamic-method) System.Object.85(intptr,intptr,intptr,int,int)

But not sure what that is supposed to mean. And clearly this doesn't happen all the time.

PureWeen:
@jfversluis 

> System.NotSupportedException: Unable to activate instance of type

Means that a Renderer or View has been disposed but then some action/event is trying to reuse that View or Renderer.

For example

- View Exists
- Event  wired up
- View disposed
- Event fires
- Mono sees the View has been disposed because the HandleId is zero
- Mono tries to just recreate the view itself so it can pass the event to someone
- Boom *System.NotSupportedException: Unable to activate instance of type* because we don't supply the constructors Mono need to instantiate these types itself

So it sounds like something with your changes isn't unwiring an event. This Listenerer probably needs to be removed inside dispose?
> void TimePickerDialog.IOnTimeSetListener.OnTimeSet

samhouts:
@kingces95 We do have problems with `Focus` on all platforms that _must_ be addressed. We are not following guidelines, and unexpected behavior is occurring for most controls. It'll be a larger effort, and I expect it will introduce breaking changes that we'll have to mitigate somehow. 

As it turns out, the dialogs popping up on `Focus` was previously working on all Picker types, and our customers took a dependency on that behavior. So, we need to restore that behavior to help our customers get back into a working state.

I agree with the strategy that you and I have discussed. We should fix `Focus` on all platforms so that it doesn't actually do anything for invisible views (or views that have no height/width), and instead allow the dialogs to be opened with a new API call. Again, since this is a breaking change, it will need to be communicated well and may need to wait until the next major version.

jfversluis:
@PureWeen I think the test run for the latest change is here: https://appcenter.ms/orgs/xtc-Xamarin-Forms/apps/AndroidControlGallery-1/test/series/pull-7289/runs/543e63f2-0253-4194-8c76-ad1568555b81

And seems all green?

Xamarin.Forms.Platform.Android/Renderers/TimePickerRenderer.cs
```diff
@@ -109,6 +115,25 @@ protected virtual TimePickerDialog CreateTimePickerDialog(int hours, int minutes
...
+		protected override void Dispose(bool disposing)
+		{
+			if (_isDisposed)
+				return;
+
+			_isDisposed = true;
+
+			if (disposing)
+			{
+				if (Forms.IsLollipopOrNewer && _dialog != null)
```
PureWeen:
```suggestion
				if (Forms.IsLollipopOrNewer && _dialog.IsAlive())
```

When we dispose we're going to need to start checking if the object is still alive before we call into native apis. 
This is really important for the previewer but also a good practice to start adopting in general.

PR #7294, opened 2019-08-27
Anticipate .cctors

kingces95:
### Description of Change ###

Speed up startup by making use of any extra processors on the phone to eagerly execute class constructors off the UIThread which we know the UIThread will need during startup.

_This change helps more with non-AOT debug development startup by JITting a big chunk of code off the UIThread (poor man's AOT). So, honestly, this change just lays the ground work for the more significant change (which speeds up release performance) of pre-inflating views off the UIThread by demonstrating how the `Anticipator` works (it's just a hand rolled thread pool)._ 

### Issues Resolved ### 

None.

### API Changes ###

 None

### Platforms Affected ### 

- Android

None

### Testing Procedure ###

Every UITest is a test. If the app starts, then this worked!

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/Anticipator.cs
+++ b/Xamarin.Forms.Platform.Android/Anticipator.cs
@@ -1,0 +1,80 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+namespace Xamarin.Forms.Platform.Android
+{
+	internal class Anticipator
+	{
+		const int ThreadLifeTimeSeconds = 5;
+		readonly static TimeSpan LoopTimeOut = TimeSpan.FromSeconds(ThreadLifeTimeSeconds);
+
+		readonly Thread[] _threads;
+		readonly AutoResetEvent[] _signals;
+		readonly ConcurrentQueue<Action> _actions;
+
+		internal Anticipator()
+		{
+			_actions = new ConcurrentQueue<Action>();
+
+			_threads = new Thread[Environment.ProcessorCount];
+			_signals = new AutoResetEvent[Environment.ProcessorCount];
+
+			_signals[0] = new AutoResetEvent(true);
+			_threads[0] = new Thread(Loop);
+			_threads[0].Start(_signals[0]);
+
+			Anticipate(() =>
+			{
+				for (var i = 1; i < _threads.Length; i++)
+				{
+					_signals[i] = new AutoResetEvent(true);
+					_threads[i] = new Thread(Loop);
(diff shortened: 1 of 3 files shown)


rmarinho:
Is there any measurements you could provide with this and without ?!

Xamarin.Forms.Platform.Android/Anticipator.cs
```diff
@@ -0,0 +1,80 @@
...
+				throw new ArgumentNullException(nameof(action));
+
+			_actions.Enqueue(action);
+
+			Signal();
+		}
+
+		public void AnticipateClassConstruction(Type type)
+		{
+			Anticipate(() => RuntimeHelpers.RunClassConstructor(typeof(Resource.Layout).TypeHandle));
```
StephaneDelcroix: you probably should use the `type` parameter here, instead of the hardcoded Layout reference

kingces95: Oh? I'm not familiar. I wanna run the `.cctor` of `Reosurce.Layout`. What's the alternative to that than this?

StephaneDelcroix: don't you want to run the `.cctor` on the `type` being passed as method parameter ??

because later on, you invoke this with `anticipator.AnticipateClassConstruction(typeof(Resource.Layout));`

kingces95: That'll help...

Xamarin.Forms.Platform.Android/Anticipator.cs
```diff
@@ -0,0 +1,80 @@
...
+				throw new ArgumentNullException(nameof(action));
+
+			_actions.Enqueue(action);
+
+			Signal();
+		}
+
+		public void AnticipateClassConstruction(Type type)
+		{
+			Anticipate(() => RuntimeHelpers.RunClassConstructor(type.TypeHandle));
```
StephaneDelcroix: way better :)

rmarinho:
Should this be moved to the internals namespace on Core ?

StephaneDelcroix:
@rmarinho 

> Should this be moved to the internals namespace on Core ?

no, definitely not, never, not under my watch

Xamarin.Forms.Platform.Android/Anticipator.cs
kingces95: Way.

kingces95:
My next PR replaces XAML with a public version of Anticipator. It's gonna be great.

PR #7315, opened 2019-08-29
[Enhancement] Add platform specific property to handle iOS UITabBar Translucent property

### LINKED ISSUE #7135: [Enhancement] Add platform specific property to set the iOS UITabBar Translucent property to false
### Description
When we set the BackgroundImageSource to a page that is inside a TabbedPage, the tab's colors (background color and unselected color) are shifted a bit, meaning that the actual colors are not represented correctly. This happens only for iOS. On Android, the colors are fine.

### Steps to Reproduce
1. Open the solution DemoImageOpacityIssue
2. Run the project and make a screenshot of the MainPage. 
3. Inspect the screenshot and confirm that the colors are 1:1 with the ones that are set in code (The color is in MainPage.xaml and are #f7f7f7 for BarBackgroundColor property). 
4. Uncomment the 14th line in FirstPage.xaml.cs, which is simply setting a page background. 
5. Make a screenshot again and inspect the colors. 

Notes: 
The selection color, if we set any, doesn't change.
Also, it doesn't matter if the BackgroundImageSource propery is set from the xaml, or from the code-behind.
We are using the default deselected color, but any custom set color will be changed also.

### Expected Behavior
The second screenshot should have the exact same colors as in the first screenshot (as they are set in the code)

### Actual Behavior
The colors are not the ones that we set in code.
The colors should be #f7f7f7 for bar background and #a1a1a1 for unselected icon, but the actual colors, when we have a background image, are #f5f5f5 for bar background and #a0a0a0 for unselected icon.

### Basic Information

- Version with issue: Latest stable - 4.1.0.673156, but it also happens with the latest pre-release - 4.2.0.673161-pre3
- IDE: Visual Studio Enterprise 2019 16.2.1 (windows)
- Platform Target Frameworks: 
  - iOS:  12.14.0.114
- Nuget Packages: N/A
- Affected Devices: Tested with iPhone 7 Plus and iPhone X, but it should happen on all iOS devices

### Screenshots
normal
shifted

### Reproduction Link
[DemoImageOpacityIssue.zip](https://github.com/xamarin/Xamarin.Forms/files/3495912/DemoImageOpacityIssue.zip)

AndreiMisiukevich:
# Description of Change ###

Added one more platform-specific property for iOS

### Issues Resolved ### 

- fixes #7135 

### API Changes ###

Added:
- Xamarin.Forms.PlatformConfiguration.iOSSpecific.TabbedPage (IsTranslucentTabBarProperty)

### Platforms Affected ### 
- Core (iOS specific stuff)
- iOS

Can change Translucent of TabBar

Not applicable (it's better to use Color meter)

### Testing Procedure ###
```charp
MainPage = new TabbedPage
{
    Children = {
        new ContentPage() { BackgroundColor = Color.Red },
        new ContentPage()
    }
};
Xamarin.Forms.PlatformConfiguration.iOSSpecific.TabbedPage.SetIsTranslucentTabBar(MainPage, false);
```

- Check bar colors with a color meter

### DIFF
--- a/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TabbedPage.cs
+++ b/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TabbedPage.cs
@@ -4,10 +4,8 @@
 
 	public static class TabbedPage
 	{
-		const string TranslucencyModePropertyName = "TranslucencyMode";
-
 		public static readonly BindableProperty TranslucencyModeProperty =
-			BindableProperty.Create(TranslucencyModePropertyName,
+			BindableProperty.Create("TranslucencyMode",
 				typeof(TranslucencyMode), typeof(TabbedPage), TranslucencyMode.Default);
 
 		public static TranslucencyMode GetTranslucencyMode(BindableObject element)
--- a/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TabbedPage.cs
+++ b/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TabbedPage.cs
@@ -1,0 +1,30 @@
+﻿namespace Xamarin.Forms.PlatformConfiguration.iOSSpecific
+{
+	using FormsElement = Forms.TabbedPage;
+
+	public static class TabbedPage
+	{
+		private const string TranslucencyModePropertyName = "TranslucencyMode";
+
+		public static readonly BindableProperty TranslucencyModeProperty =
+			BindableProperty.Create(TranslucencyModePropertyName,
+				typeof(TranslucencyMode), typeof(TabbedPage), TranslucencyMode.Default);
+
+		public static TranslucencyMode GetTranslucencyMode(BindableObject element)
+			=> (TranslucencyMode)element.GetValue(TranslucencyModeProperty);
+
+		public static void SetTranslucencyMode(BindableObject element, TranslucencyMode value)
+			=> element.SetValue(TranslucencyModeProperty, value);
+
+		public static TranslucencyMode GetTranslucencyMode(
+			this IPlatformElementConfiguration<iOS, FormsElement> config)
+			=> GetTranslucencyMode(config.Element);
+
+		public static IPlatformElementConfiguration<iOS, FormsElement> SetTranslucencyMode(
+			this IPlatformElementConfiguration<iOS, FormsElement> config, TranslucencyMode value)
+		{
+			SetTranslucencyMode(config.Element, value);
+			return config;
+		}
+	}
+}
--- a/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TabbedPage.cs
+++ b/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TabbedPage.cs
@@ -4,7 +4,7 @@
 
 	public static class TabbedPage
 	{
-		private const string TranslucencyModePropertyName = "TranslucencyMode";
+		const string TranslucencyModePropertyName = "TranslucencyMode";
 
 		public static readonly BindableProperty TranslucencyModeProperty =
 			BindableProperty.Create(TranslucencyModePropertyName,
--- a/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TranslucencyMode.cs
+++ b/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TranslucencyMode.cs
@@ -1,0 +1,10 @@
+﻿
+namespace Xamarin.Forms.PlatformConfiguration.iOSSpecific
+{
+	public enum TranslucencyMode
+	{
+		Default,
+		Translucent,
+		Opaque
+	}
+}
--- a/Xamarin.Forms.Platform.iOS/Renderers/TabbedRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/TabbedRenderer.cs
@@ -7,6 +7,8 @@ using UIKit;
 using Xamarin.Forms.Internals;
 using static Xamarin.Forms.PlatformConfiguration.iOSSpecific.Page;
 using PageUIStatusBarAnimation = Xamarin.Forms.PlatformConfiguration.iOSSpecific.UIStatusBarAnimation;
+using TabbedPageConfiguration = Xamarin.Forms.PlatformConfiguration.iOSSpecific.TabbedPage;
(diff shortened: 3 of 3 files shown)


AndreiMisiukevich:
image

🤣

Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TabbedPage.cs
```diff
@@ -0,0 +1,30 @@
...
+	public static class TabbedPage
+	{
+		private const string IsTranslucentTabBarPropertyName = "IsTranslucentTabBar";
+
+		public static readonly BindableProperty IsTranslucentTabBarProperty =
+			BindableProperty.Create(IsTranslucentTabBarPropertyName,
+				typeof(bool?), typeof(TabbedPage), default(bool?));
+
+		public static bool? GetIsTranslucentTabBar(BindableObject element)
+			=> element.GetValue(IsTranslucentTabBarProperty) as bool?;
```
StephaneDelcroix: use direct cast

Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TabbedPage.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+	using FormsElement = Forms.TabbedPage;
+
+	public static class TabbedPage
+	{
+		private const string IsTranslucentTabBarPropertyName = "IsTranslucentTabBar";
+
+		public static readonly BindableProperty IsTranslucentTabBarProperty =
+			BindableProperty.Create(IsTranslucentTabBarPropertyName,
+				typeof(bool?), typeof(TabbedPage), default(bool?));
```
StephaneDelcroix: an enum is probably better than a bool?

AndreiMisiukevich: > an enum is probably better than a bool?

hm. Don't think so, since this property has only two possible values (true or false)

Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TabbedPage.cs
AndreiMisiukevich: > use direct cast

sorry, my habit )

Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TabbedPage.cs
AndreiMisiukevich: @StephaneDelcroix I decided to reopen this thread.

How do you see that?

```public enum TranslucentValue { Default, Translucent, NonTranslucent  }``` ? 

I lean toward to ```bool?``` just because it is much easier to use with bindings

**Comment by AndreiMisiukevich** [comment]
@samhouts let's continue the discussion / the review

File: Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TabbedPage.cs (hunk 2 above)
Comment (StephaneDelcroix) [reply Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TabbedPage.cs]: yes, that's what I mean. We shouldn't expose tri-state bool in any public API

**Comment by AndreiMisiukevich** [comment]
@rmarinho @StephaneDelcroix updated PR in accordance with requirements (avoid optional values in public API)

File: Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TabbedPage.cs [hunk 3]
```diff
@@ -0,0 +1,30 @@
+﻿namespace Xamarin.Forms.PlatformConfiguration.iOSSpecific
+{
+	using FormsElement = Forms.TabbedPage;
+
+	public static class TabbedPage
+	{
+		private const string TranslucencyModePropertyName = "TranslucencyMode";
```
rmarinho:
```suggestion
		const string TranslucencyModePropertyName = "TranslucencyMode";
```

AndreiMisiukevich:
@rmarinho @PureWeen where do we stand with this PR guys?

Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TabbedPage.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace Xamarin.Forms.PlatformConfiguration.iOSSpecific
+{
+	using FormsElement = Forms.TabbedPage;
```
StephaneDelcroix: I sense something wrong with the identation

Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TabbedPage.cs
```diff
@@ -0,0 +1,30 @@
...
+
+	public static class TabbedPage
+	{
+		const string TranslucencyModePropertyName = "TranslucencyMode";
```
StephaneDelcroix: creating a const that is used once is overkill

AndreiMisiukevich: I will update it, but I cannot agree that it's overkill. From my perspective, it's a good practice. Anyway, constants will be put in a literal pool.

Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/TabbedPage.cs
AndreiMisiukevich: Look at this file: https://github.com/xamarin/Xamarin.Forms/blob/6df34004b219593885ef72e7fd2fe83b7c0fd915/Xamarin.Forms.Core/PlatformConfiguration/GTKSpecific/NavigationPage.cs

It uses spaces instead of tabs.

But most of the files use tabs instead of spaces (e.g https://github.com/xamarin/Xamarin.Forms/blob/6df34004b219593885ef72e7fd2fe83b7c0fd915/Xamarin.Forms.Core/AbsoluteLayout.cs)

My file uses TABS.

Is it your concern? Could you please explain what did you mean? Thx

AndreiMisiukevich: or do you hate "inside" usings? :)

AndreiMisiukevich:
Should I change anything?)

MitchBomcanhao:
is this supposed to work via XAML on 4.5.0-pre1?
I can set it up via code but not in xaml like other platform specifics...

PR #7318, opened 2019-08-29
Set MinimumWidth to the correctly calculated value for Auto Grid Columns

### LINKED ISSUE #7303: [Bug] Grid is not correctly using MinimumWidthRequest with calculating Auto Cells
### Description

The code here
https://github.com/xamarin/Xamarin.Forms/blob/409271bbfaf845683399d4faab728a1ca2dfaa10/Xamarin.Forms.Core/GridCalc.cs#L173

Is incorrectly using the actualWidth for the minimum width and it should be using the minimumWidth as calculated by the for loop above it

https://github.com/xamarin/Xamarin.Forms/blob/409271bbfaf845683399d4faab728a1ca2dfaa10/Xamarin.Forms.Core/GridCalc.cs#L168

My guess is that fixing this is going to "Break" a lot of existing layouts so I'm not sure the best approach currently to fixing this. Maybe we enable with a platform specific for now? Or we take this opportunity to obsolete these fields? 

In theory if this breaks a persons layout they would fix it by just removing the MinimumWidthRequest completely which should cause it to revert back to how it currently works

### Steps to Reproduce
Given the following XAML

```
<Grid>
        <Grid.RowDefinitions>
            <RowDefinition />
            <RowDefinition />
        </Grid.RowDefinitions>
        <Grid.ColumnDefinitions>
            <ColumnDefinition Width="Auto" />
            <ColumnDefinition Width="Auto" />
        </Grid.ColumnDefinitions>
        <BoxView HorizontalOptions="Start" Grid.Row="0" Grid.Column="0" BackgroundColor="Purple" MinimumWidthRequest="1" WidthRequest="500"></BoxView>
        <BoxView HorizontalOptions="Start" Grid.Row="0" Grid.Column="1" BackgroundColor="Green" MinimumWidthRequest="450" WidthRequest="500"></BoxView>
        <BoxView HorizontalOptions="Start" Grid.Row="1" Grid.Column="0" BackgroundColor="Purple" MinimumWidthRequest="250" WidthRequest="500"></BoxView>
        <BoxView HorizontalOptions="Start" Grid.Row="1" Grid.Column="1" BackgroundColor="Green" MinimumWidthRequest="250" WidthRequest="500"></BoxView>
    </Grid>
```

### Expected Behavior

The left BoxView should shrink as the view gets smaller

image

### Actual Behavior

Both BoxViews stay the same width (note the green looks smaller but that's only because it's off the screen). The purple boxview should be shrinking as the view is shrinking

image

### Basic Information

- Version with issue: master

PureWeen:
### Description of Change ###
When calculating the MinimumWidth for an auto column the code was incorrectly throwing that value away and just using the actual width for the minimum width.

### Issues Resolved ### 
- fixes #7303 

### Platforms Affected ### 
- Core/XAML (all platforms)

Given the following XAML

```
<Grid>
        <Grid.RowDefinitions>
            <RowDefinition />
            <RowDefinition />
        </Grid.RowDefinitions>
        <Grid.ColumnDefinitions>
            <ColumnDefinition Width="Auto" />
            <ColumnDefinition Width="Auto" />
        </Grid.ColumnDefinitions>
        <BoxView HorizontalOptions="Start" Grid.Row="0" Grid.Column="0" BackgroundColor="Purple" MinimumWidthRequest="1" WidthRequest="500"></BoxView>
        <BoxView HorizontalOptions="Start" Grid.Row="0" Grid.Column="1" BackgroundColor="Green" MinimumWidthRequest="450" WidthRequest="500"></BoxView>
        <BoxView HorizontalOptions="Start" Grid.Row="1" Grid.Column="0" BackgroundColor="Purple" MinimumWidthRequest="250" WidthRequest="500"></BoxView>
        <BoxView HorizontalOptions="Start" Grid.Row="1" Grid.Column="1" BackgroundColor="Green" MinimumWidthRequest="250" WidthRequest="500"></BoxView>
    </Grid>
```

Before the change both BoxViews stay the same width (note the green looks smaller but that's only because it's off the screen). The purple boxview should be shrinking as the view is shrinking since it has a MinimumWidthRequest set

image

After this change the left BoxView shrinks.
image

This means anyone that is using *MinimumWidthRequest* inside a Grid with the Column width set to Auto is going to experience a new layout once they update forms. Technically if they want to fix the breaking change they just have to delete the *MinimumWidthRequest* value from the elements inside the grid as that's basically how this was operating prior to this fix

### Testing Procedure ###
- Automated tests

### PR Checklist ###

- [x] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/GridTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/GridTests.cs
@@ -1541,6 +1541,115 @@ namespace Xamarin.Forms.Core.UnitTests
 		}
 
 
+		[Test]
+		public void MinimumWidthRequestInAutoCells()
+		{
+			var boxRow0Column0 = new BoxView
+			{
+				MinimumWidthRequest = 50,
+				WidthRequest = 200,
+				IsPlatformEnabled = true
+			};
+			var boxRow1Column0 = new BoxView
+			{
+				MinimumWidthRequest = 50,
+				WidthRequest = 200,
+				IsPlatformEnabled = true
+			};
+
+			var boxRow0Column1 = new BoxView
+			{
+				WidthRequest = 800,
+				IsPlatformEnabled = true
+			};
+			var boxRow1Column1 = new BoxView
+			{
+				WidthRequest = 800,
+				IsPlatformEnabled = true
+			};
+
+			var grid = new Grid
+			{
+				IsPlatformEnabled = true,
+				BackgroundColor = Color.Red
+			};
+
+			grid.ColumnDefinitions.Add(new ColumnDefinition() { Width = GridLength.Auto });
+			grid.ColumnDefinitions.Add(new ColumnDefinition() { Width = GridLength.Auto });
+
+			grid.Children.Add(boxRow0Column0, 0, 0);
+			grid.Children.Add(boxRow1Column0, 0, 1);
+			grid.Children.Add(boxRow1Column1, 1, 1);
+			grid.Children.Add(boxRow0Column1, 1, 0);
+
+			var view = new ContentView
+			{
+				IsPlatformEnabled = true,
+				Content = grid,
+			};
+			view.Layout(new Rectangle(0, 0, 800, 800));
+
+
+			Assert.AreEqual(boxRow0Column0.MinimumWidthRequest, boxRow0Column0.Width);
+			Assert.AreEqual(boxRow1Column0.MinimumWidthRequest, boxRow1Column0.Width);
+		}
+
+
+		[Test]
+		public void MinimumHeightRequestInAutoCells()
+		{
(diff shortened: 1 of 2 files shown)


jsuarezruiz:
Wow, this must have been like this for a long time. The only thing that worries me about this change is that it will impact on third parties although the fix is easy to do.

johnshardman:
Is this fixing MinimumWidthRequest (and MinimumHeightRequest) in general, or just in the context of a Grid?

There were bugs logged in Bugzilla years ago regarding these two properties, which Xamarin declined to resolve. The properties never did what people expected, and Xamarin wouldn't change them to do what people expected as it was claimed to be a breaking change (it really wasn't, as nobody used them as they were useless). If the change being done here is properly implementing them to do what people expect, this is a very good thing.

PR #7319, opened 2019-08-30
[UWP] RefreshView

PureWeen:
### Description of Change ###
RefreshView renderers for UWP

### Issues Resolved ### 
- fixes for UWP #5882

### Platforms Affected ### 
- UWP

### Screenshots ### 

image

image
*In the previous gif, in UWP, a Platform Specific is used to modify the pull direction.*

### Testing Procedure ###
- Ensure all the Refresh View gallery pages work
- Added Warnings messages if your Target Platform is less than 17763

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/.nuspec/Xamarin.Forms.nuspec
+++ b/.nuspec/Xamarin.Forms.nuspec
@@ -24,6 +24,7 @@
         <dependency id="NETStandard.Library" version="2.0.1"/>
         <dependency id="Microsoft.NETCore.UniversalWindowsPlatform" version="6.0.12" />
         <dependency id="Win2D.uwp" version="1.20.0" />
+        <dependency id="Microsoft.UI.Xaml" version="2.1.190606001" />
       </group>
       <group targetFramework="tizen40">
         <dependency id="Tizen.NET" version="4.0.0"/>
@@ -223,6 +224,7 @@
     <file src="..\Xamarin.Forms.Platform.UAP\bin\$Configuration$\StepperControl.xbf" target="lib\uap10.0\Xamarin.Forms.Platform.UAP" />
     <file src="..\Xamarin.Forms.Platform.UAP\bin\$Configuration$\CollectionView\ItemsViewStyles.xbf" target="lib\uap10.0\Xamarin.Forms.Platform.UAP\CollectionView" />
     <file src="..\Xamarin.Forms.Platform.UAP\bin\$Configuration$\FormsCheckBoxStyle.xbf" target="lib\uap10.0\Xamarin.Forms.Platform.UAP" />
+    <file src="..\Xamarin.Forms.Platform.UAP\bin\$Configuration$\Microsoft.UI.Xaml\DensityStyles\Compact.xbf" target="lib\uap10.0\Xamarin.Forms.Platform.UAP\Microsoft.UI.Xaml\DensityStyles\" />
 
     <!--Mac-->
     <file src="..\Xamarin.Forms.Core\bin\$Configuration$\netstandard2.0\Xamarin.Forms.Core.dll" target="lib\Xamarin.Mac" />
--- a/.nuspec/Xamarin.Forms.targets
+++ b/.nuspec/Xamarin.Forms.targets
@@ -15,6 +15,9 @@
 		<_DefaultXamlItemsEnabled>False</_DefaultXamlItemsEnabled>
 		<EnableDefaultCssItems Condition="'$(EnableDefaultCssItems)'==''">True</EnableDefaultCssItems>
 		<_DefaultCssItemsEnabled>False</_DefaultCssItemsEnabled>
+		<!-- The WINUI check for this only runs when there is an empty string so I just convert false to an empty string to fall in line with our other properties -->
+		<SkipMicrosoftUIXamlCheckTargetPlatformVersion Condition="'$(SkipMicrosoftUIXamlCheckTargetPlatformVersion)'==''">true</SkipMicrosoftUIXamlCheckTargetPlatformVersion>
+		<SkipMicrosoftUIXamlCheckTargetPlatformVersion Condition="'$(SkipMicrosoftUIXamlCheckTargetPlatformVersion)'=='false'"></SkipMicrosoftUIXamlCheckTargetPlatformVersion>
 	</PropertyGroup>
 
 	<Import Project="$(MSBuildThisFileDirectory)Xamarin.Forms.DefaultItems.targets" Condition="'$(MSBuildSDKsPath)'!=''" />
--- a/.nuspec/Xamarin.Forms.targets
+++ b/.nuspec/Xamarin.Forms.targets
@@ -178,4 +178,21 @@
     <Error Code="XF005"  Condition="$(TargetFrameworkVersionWithoutV) &lt; $(MinTargetFrameworkVersionForForms)"
            Text="The %24(TargetFrameworkVersion) for $(ProjectName) ($(TargetFrameworkVersion)) is less than the minimum required %24(TargetFrameworkVersion) for Xamarin.Forms ($(MinTargetFrameworkVersionForForms)). You need to increase the %24(TargetFrameworkVersion) for $(ProjectName)."   />
   </Target>
+  
+  <!-- UWP Targets-->
+  <Target Name="WinUICheckTargetPlatformVersion" BeforeTargets="PrepareForBuild"
+      Condition="'$(TargetPlatformVersion)' != '' and '$(TargetPlatformMinVersion)' != ''">
+    <PropertyGroup>
+      <MicrosoftUIXamlTargetPlatformMinCheckValue>$([System.Version]::Parse('$(TargetPlatformMinVersion)').Build)</MicrosoftUIXamlTargetPlatformMinCheckValue>
+    </PropertyGroup>
+    <Warning
+        Text="Xamarin.Forms recommends TargetPlatformMinVersion &gt;= 10.0.14393.0 (current project is $(MicrosoftUIXamlTargetPlatformMinCheckValue))"
+        Condition="$(MicrosoftUIXamlTargetPlatformMinCheckValue) &lt; 14393" />
+    <PropertyGroup>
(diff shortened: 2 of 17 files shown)


Xamarin.Forms.Platform.UAP/Platform.cs
```diff
@@ -23,8 +23,6 @@ public abstract class Platform : INavigation
 	{
 		static Task<bool> s_currentAlert;
 
-		internal static StatusBar MobileStatusBar => ApiInformation.IsTypePresent("Windows.UI.ViewManagement.StatusBar") ? StatusBar.GetForCurrentView() : null;
```
PureWeen: These changes were made  to make the https://www.nuget.org/packages/Microsoft.Toolkit.Uwp.PlatformSpecificAnalyzer happy. Though there appears to a bug with that analyzer currently that's causing it to throw an NRE which is why it's not installed

.nuspec/Xamarin.Forms.nuspec
```diff
@@ -24,6 +24,7 @@
         <dependency id="NETStandard.Library" version="2.0.1"/>
         <dependency id="Microsoft.NETCore.UniversalWindowsPlatform" version="6.0.12" />
         <dependency id="Win2D.uwp" version="1.20.0" />
+        <dependency id="Microsoft.UI.Xaml" version="2.1.190606001" />
```
dotMorten: Perhaps use just released `v2.2.190830001` as this version supports transitive build targets, so at least VS2019 users doesn't have to also add this package to their project head

.nuspec/Xamarin.Forms.nuspec
```diff
@@ -223,6 +224,7 @@
     <file src="..\Xamarin.Forms.Platform.UAP\bin\$Configuration$\StepperControl.xbf" target="lib\uap10.0\Xamarin.Forms.Platform.UAP" />
     <file src="..\Xamarin.Forms.Platform.UAP\bin\$Configuration$\CollectionView\ItemsViewStyles.xbf" target="lib\uap10.0\Xamarin.Forms.Platform.UAP\CollectionView" />
     <file src="..\Xamarin.Forms.Platform.UAP\bin\$Configuration$\FormsCheckBoxStyle.xbf" target="lib\uap10.0\Xamarin.Forms.Platform.UAP" />
+    <file src="..\Xamarin.Forms.Platform.UAP\bin\$Configuration$\Microsoft.UI.Xaml\DensityStyles\Compact.xbf" target="lib\uap10.0\Xamarin.Forms.Platform.UAP\Microsoft.UI.Xaml\DensityStyles\" />
```
dotMorten: The Microsoft.UI.Xaml package should be including this for you. I'd worry this makes it get included twice + if you decide on using a newer version of WinUI, these might not align. When I was experimenting with using WinUI to support shell I didn't have to do this.

Xamarin.Forms.ControlGallery.WindowsUniversal/Xamarin.Forms.ControlGallery.WindowsUniversal.csproj
```diff
@@ -236,6 +236,9 @@
     <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform">
       <Version>6.0.6</Version>
     </PackageReference>
+    <PackageReference Include="Microsoft.UI.Xaml">
+      <Version>2.1.190606001</Version>
+    </PackageReference>
```
dotMorten: Once you move to `v2.2.190830001`, this should only be necessary if you use VS2017 (but doesn't hurt to do, except more to maintain)

Xamarin.Forms.Platform.UAP/Resources.xaml
```diff
@@ -5,6 +5,7 @@
 	x:Class="Xamarin.Forms.Platform.UWP.Resources">
 
 	<ResourceDictionary.MergedDictionaries>
+        <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls"/>
```
dotMorten: Consider doing this programmatically. That way you can throw if WinUI wasn't referenced by the head project (which could be an issue in VS2017). It would even allow you to only throw for those controls that rely on WinUI, thus making it non-breaking and only required for RefreshView.

Xamarin.Forms.Platform.UAP/Xamarin.Forms.Platform.UAP.csproj
```diff
@@ -268,6 +270,9 @@
     <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform">
       <Version>6.0.6</Version>
     </PackageReference>
+    <PackageReference Include="Microsoft.UI.Xaml">
+      <Version>2.1.190606001</Version>
```
dotMorten: `v2.2.190830001` ?

.nuspec/Xamarin.Forms.nuspec
dotMorten: @jevansaks Should this be necessary?

jevansaks: No, especially with buildTransitive fix it shouldn't be necessary.

PureWeen: Without it I get the following exception when I try to compile
```
Severity	Code	Description	Project	File	Line	Suppression State
Error		Payload file 'C:\Users\shane\.nuget\packages\xamarin.forms\4.3.0.2061\lib\uap10.0\Xamarin.Forms.Platform.UAP\Microsoft.UI.Xaml\DensityStyles\Compact.xbf' does not exist.	App133.UWP			
```

Which is perhaps caused by the addition of this?

https://github.com/xamarin/Xamarin.Forms/pull/7319#discussion-diff-319736060R8

PureWeen: Here's a project with included nuget based on latest WINUI to demonstrate the exception above
[App133.zip](https://github.com/xamarin/Xamarin.Forms/files/3575999/App133.zip)

I also tested by removing
```
        <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls"/>
```

But same exception

.nuspec/Xamarin.Forms.targets
```diff
@@ -15,6 +15,9 @@
 		<_DefaultXamlItemsEnabled>False</_DefaultXamlItemsEnabled>
 		<EnableDefaultCssItems Condition="'$(EnableDefaultCssItems)'==''">True</EnableDefaultCssItems>
 		<_DefaultCssItemsEnabled>False</_DefaultCssItemsEnabled>
+		<!-- The WINUI check for this only runs when there is an empty string so I just convert false to an empty string to fall in line with our other properties -->
+		<SkipMicrosoftUIXamlCheckTargetPlatformVersion Condition="'$(SkipMicrosoftUIXamlCheckTargetPlatformVersion)'==''">true</SkipMicrosoftUIXamlCheckTargetPlatformVersion>
+		<SkipMicrosoftUIXamlCheckTargetPlatformVersion Condition="'$(SkipMicrosoftUIXamlCheckTargetPlatformVersion)'=='false'"></SkipMicrosoftUIXamlCheckTargetPlatformVersion>
```
jsuarezruiz: Formatting error.

.nuspec/Xamarin.Forms.nuspec
dotMorten: Hmm I see the problem. Apparently the resource gets referenced by `\lib\uap10.0\Xamarin.Forms.Platform.UAP.pri`, which is now missing.

dotMorten: @jevansaks Perhaps this should only be done with project heads, not class libraries: 
https://github.com/microsoft/microsoft-ui-xaml/blob/f2b0459553152d69168aa6929c9ff0a7188ac8ea/build/NuSpecs/MUXControls-Nuget-FrameworkPackage.props#L134-L139

dotMorten: @PureWeen Quick fix would be to explicitly remove that resource again in csproj, ie <Page Remove="**\*\DensityStyles\Compact.xaml" />.
But it's probably not a huge problem to have it there twice for now, as I'm guessing the ms-appx path used will be different.

PureWeen:
Rolled back WINUI version to 2.1 until this is resolved
https://github.com/microsoft/microsoft-ui-xaml/issues/1277

Xamarin.Forms.Platform.UAP/RefreshViewRenderer.cs
```diff
@@ -0,0 +1,176 @@
...
+		protected override void OnElementChanged(ElementChangedEventArgs<RefreshView> e)
+		{
+			if (e.NewElement != null)
+			{
+				if (Control == null)
+				{
+					var refreshControl = new RefreshContainer
+					{
+						PullDirection = RefreshPullDirection.TopToBottom,
+						Visualizer = new RefreshVisualizer()
```
PureWeen: @jsuarezruiz it looks like when you assign a Visualizer the RefreshRequested event fires twice. 

Do we need to assign a Visualizer here?

PureWeen: I see it was so you could update the colors. I added a callback to set those colors once WinUI just initializes it itself which looks to work

jsuarezruiz:
Using WinUI 2.1 everything works again as expected!.

Xamarin.Forms.Platform.UAP/Resources.xaml
jsuarezruiz: I think it's a good idea, what do you think @PureWeen?

jsuarezruiz: `MergedDictionaries.Add(new Microsoft.UI.Xaml.Controls.XamlControlsResources());`

PureWeen: https://github.com/xamarin/Xamarin.Forms/pull/7319/files#diff-fadef492a65644f17c0a99fa2e742e80R39

PureWeen:
Failing tests are all the CarouselView tests

PR #7321, opened 2019-08-30
Allow CollectionView to specify which property contains the items for a group

hartez:
This is a proposal/proof-of-concept implementation; I'm looking for feedback on the API proposed and the naming.

### Description of Change ###

At the moment, CollectionView grouping requires that the type used for a group be an `IEnumerable`. That is, to display groups of `Item` objects, given a top level model class like

```
class GroupsModel : IList<GroupModel> {}
```

`GroupModel` would have to be something like

```
class Group : IList<Item> {}
```

This is inconvenient and awkward for a number of reasons; it also poses some issues for common serialization scenarios (see the comments below "Issues Resolved" for an explanation). 

These changes allow for CollectionView users to specify an additional value `GroupItemsMemberName` - a string which tells the CollectionView which property on the "group" model contains the actual list of objects to display. So the `GroupModel` could be something like

```
class Group 
{
	public string Title { get; }
	public IList<Item> Items { get; }
}

```

Setting `GroupItemsMemberName` to "Items" would tell the CollectionView to look at the `Items` property for the objects to display. If `GroupItemsMemberName` is not specified, CollectionView assumes the default scenario shown above (where `GroupModel` is itself an `IEnumerable`).

### Issues Resolved ### 

- Concerns about serialization raised here: https://github.com/xamarin/Xamarin.Forms/issues/3172#issuecomment-493382178
- Inconvenience of requiring groups to inherit directly from `IEnumerable`

### API Changes ###

Added to GroupableItemsView:

`public static readonly BindableProperty GroupItemsMemberNameProperty =
	BindableProperty.Create(nameof(GroupItemsMemberName), typeof(string), typeof(GroupableItemsView), null);`

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android

None

Not applicable

### Testing Procedure ###

Control Gallery, CollectionView Gallery -> Grouping Gallery -> GroupItemsMemberName Property

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls/CoreGallery.cs
+++ b/Xamarin.Forms.Controls/CoreGallery.cs
@@ -254,8 +254,6 @@ namespace Xamarin.Forms.Controls
 		{
 			public GalleryPageFactory(Func<Page> create, string title)
 			{
-				Device.SetFlags(new List<string>(Device.Flags ?? new List<string>()) { "CollectionView_Experimental" });
-
 				Realize = () =>
 				{
 					var p = create();
@@ -287,7 +285,7 @@ namespace Xamarin.Forms.Controls
 				new GalleryPageFactory(() => new MemoryLeakGallery(), "Memory Leak"),
 				new GalleryPageFactory(() => new Issues.A11yTabIndex(), "Accessibility TabIndex"),
 				new GalleryPageFactory(() => new FontImageSourceGallery(), "Font ImageSource"),
-				new GalleryPageFactory(() => new CollectionViewGallery(), "_CollectionView Gallery"),
+				new GalleryPageFactory(() => new CollectionViewGallery(), "CollectionView Gallery"),
 				new GalleryPageFactory(() => new CollectionViewCoreGalleryPage(), "CollectionView Core Gallery"),
 				new GalleryPageFactory(() => new Issues.PerformanceGallery(), "Performance"),
 				new GalleryPageFactory(() => new EntryReturnTypeGalleryPage(), "Entry ReturnType "),
--- a/Xamarin.Forms.Controls/CoreGallery.cs
+++ b/Xamarin.Forms.Controls/CoreGallery.cs
@@ -254,6 +254,8 @@ namespace Xamarin.Forms.Controls
 		{
 			public GalleryPageFactory(Func<Page> create, string title)
 			{
+				Device.SetFlags(new List<string>(Device.Flags ?? new List<string>()) { "CollectionView_Experimental" });
+
 				Realize = () =>
 				{
 					var p = create();
@@ -285,7 +287,7 @@ namespace Xamarin.Forms.Controls
 				new GalleryPageFactory(() => new MemoryLeakGallery(), "Memory Leak"),
 				new GalleryPageFactory(() => new Issues.A11yTabIndex(), "Accessibility TabIndex"),
 				new GalleryPageFactory(() => new FontImageSourceGallery(), "Font ImageSource"),
-				new GalleryPageFactory(() => new CollectionViewGallery(), "CollectionView Gallery"),
+				new GalleryPageFactory(() => new CollectionViewGallery(), "_CollectionView Gallery"),
 				new GalleryPageFactory(() => new CollectionViewCoreGalleryPage(), "CollectionView Core Gallery"),
 				new GalleryPageFactory(() => new Issues.PerformanceGallery(), "Performance"),
 				new GalleryPageFactory(() => new EntryReturnTypeGalleryPage(), "Entry ReturnType "),
--- a/Xamarin.Forms.Controls/GalleryPages/CollectionViewGalleries/GroupingGalleries/GroupItemsProperty.xaml
+++ b/Xamarin.Forms.Controls/GalleryPages/CollectionViewGalleries/GroupingGalleries/GroupItemsProperty.xaml
@@ -3,37 +3,53 @@
              xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
              xmlns:d="http://xamarin.com/schemas/2014/forms/design"
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             xmlns:local="clr-namespace:Xamarin.Forms.Controls.GalleryPages.CollectionViewGalleries.GroupingGalleries"
              mc:Ignorable="d"
              x:Class="Xamarin.Forms.Controls.GalleryPages.CollectionViewGalleries.GroupingGalleries.GroupItemsProperty">
+
+    <ContentPage.Resources>
+        <local:PropertySelection x:Key="PropertySelection"></local:PropertySelection>
+    </ContentPage.Resources>
+    
     <ContentPage.Content>
-        <CollectionView x:Name="CollectionView" IsGrouped="True" Header="Heroes by City" 
-                        GroupItemsPath="Heroes">
+        <StackLayout>
 
-            <CollectionView.ItemTemplate>
-                <DataTemplate>
-                    <StackLayout>
-                        <Label Text="{Binding Name}" Margin="5,0,0,0"/>
-                    </StackLayout>
-                </DataTemplate>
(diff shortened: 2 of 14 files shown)


hartez:
@VincentH-Net Let me know what you think. Still just a proposal, but unless someone comes up with a really strong argument against, I'm cautiously optimistic.

modplug:
So you'll be able to bind an ObservableCollection<Group> with an ObservableCollection<item> to a ListView?

VincentH-Net:
I'm very happy to see this proposal!

@hartez Did you intentionally choose for a `string` in favor of a `BindingBase` type property?

I originally proposed `GroupItemsBinding` to be consistent with existing similar `ListView` properties like `GroupDisplayBinding`, which is of type `BindingBase` and serves to:

> Gets or sets the binding to use for displaying the group header

That seems similar to what we are doing here. I'm not sure what the advantages/disadvantages are of using a `string` versus a `BindingBase` in this scenario; I assumed the similar `ListView` properties were of type `BindingBase` for a reason.

One advantage I can see of using `BindingBase` is that Intellisense will present you with a list of property names to choose from when you edit `GroupItemsBinding` in XAML

adrianknight89:
> If GroupItemsPath is not specified, CollectionView assumes the default scenario shown above (where GroupModel is itself an IEnumerable).

Since CollectionView is still in preview, why don't we enforce `GroupItemsPath`? I agree that the existing implementation is weird. I'd rather have developers specify the path than end up with the current implementation. It's also somewhat more difficult to maintain the codebase with two approaches.

VincentH-Net:
@adrianknight89 Good point. If we use `GroupItemsBinding` with default `"{Binding .}"` it would be 1 code approach (because it's always a binding) but it would also allow devs to use the group itself as a collection (sometimes someone may still prefer that).

On the other hand, making the group itself a collection is definitely not a best practice so it may be better from a guidance perspective to not make that the default.

In that case there would be no default, as you propose. However I would still prefer the `BindingBase GroupItemsBinding` to a `string GroupItemsPath` because BindingBase has a known syntax to explicitly refer to the group object itself.

hartez:
> @hartez Did you intentionally choose for a `string` in favor of a `BindingBase` type property?
> 
> I originally proposed `GroupItemsBinding` to be consistent with existing similar `ListView` properties like `GroupDisplayBinding`, which is of type `BindingBase` and serves to:
> 
> > Gets or sets the binding to use for displaying the group header
> 
> That seems similar to what we are doing here. I'm not sure what the advantages/disadvantages are of using a `string` versus a `BindingBase` in this scenario; I assumed the similar `ListView` properties were of type `BindingBase` for a reason.
> 

`ListView` uses an intermediate type `TemplatedItemsList`, which is itself a `BindableObject` with a bindable `Name` property. When you set the `GroupDisplayBinding` for a `ListView`, that binding is passed through to each child `TemplatedItemsList` (one for each group) as the binding for the `Name` property. All of this happens on the Forms side (rather than the native side).

`CollectionView` doesn't use an intermediate `BindableObject` like `TemplatedItemsList`. Its renderers have native classes which do something similar, but those are not subclasses of `BindableObject`, so there's nothing to apply the binding to. And the necessary methods aren't available from native code; they're `internal` to Forms. 

> One advantage I can see of using `BindingBase` is that Intellisense will present you with a list of property names to choose from when you edit `GroupItemsBinding` in XAML; e.g. when you are editing below XAML you will get "Binding" offered after typing `{` and property names after selecting / typing `Binding`:
> 
> ```cs
> <CollectionView
>    GroupItemsBinding = "{Binding Children}"
>    ...>
> </CollectionView>
> ```

The only autocompletion options I'm seeing are stuff from the Binding markup extension (e.g. `Converter`, `ConverterParameter`, etc.). Are you using an extension that provides more Intellisense info?

Regardless, I'll have to look into whether we can give Intellisense some hints on reasonable values for autocompletion.

hartez:
Renamed `GroupItemsPath` to `GroupItemsMemberName`; it's more accurate, and doesn't imply that it can handle dotted paths.

hartez:
> Since CollectionView is still in preview, why don't we enforce `GroupItemsPath`? 

Leaving `.` as the implied default makes the transition from `ListView` easier. And everywhere else in Forms assumes that if you don't explicitly specify a binding context, then the current context should be used. I think forcing it here would be an annoying surprise to most folks.

> It's also somewhat more difficult to maintain the codebase with two approaches.

Setting the property means we have to run some code to retrieve that value from the group and pass it into the child group's constructor. Without the property set, we just pass the group itself. At this point, enforcing the property would require _more_ code than we have currently.

VincentH-Net:
@hartez 
Thanks for your explanation on `BindableBase` versus `string`; it is clear now that using `string` as you propose is appropriate. The rename is also an improvement imo.

> The only autocompletion options I'm seeing are stuff from the Binding markup extension (e.g. `Converter`, `ConverterParameter`, etc.). Are you using an extension that provides more Intellisense info?

I had seen it but did not realize it was provided by an extension. Displaying viewmodel properties in XAML intellisense is a feature of [MFractor](https://www.mfractor.com/) (see it [in action](https://cdn.shopify.com/s/files/1/1676/4055/files/binding-completion.gif?v=1515452700))

> Regardless, I'll have to look into whether we can give Intellisense some hints on reasonable values for autocompletion.

If displaying viewmodel properties in XAML Intellisense is currently not an OOB feature, I think there is no need to include that in this feature.

VincentH-Net:
@hartez 
> Renamed `GroupItemsPath` to `GroupItemsMemberName`; it's more accurate, and doesn't imply that it can handle dotted paths.

This name suggests that different member types are supported, more than property names (e.g. also fields). Is that the intention?

If only property names will be supported, maybe use `GroupItemsPropertyName` to communicate that?

rmarinho:
Why not use the way it s done on UWP by relying on IGrouping .

Add a bool `IsSourceGrouped` to CollectionView to say that is using grouping 

If it's true we will expect a `IGrouping` use `TKey` as Group, and the items are the ItemsSource. Since this  is aa common scenario for c# devs shouldn't be hard for our users to opt in.

hartez:
> Why not use the way it s done on UWP by relying on IGrouping .

`IGrouping` is part of Linq; it's used to represent the result of a `GroupBy` operation. We don't want to force users to do their grouping with Linq or to implement the `IGrouping` interface just to use this feature. 

> Add a bool `IsSourceGrouped` to CollectionView to say that is using grouping

We already have `IsGrouped`.

VincentH-Net:
Hi @hartez , looks like this PR has had ample time for feedback (3 months).

Is anything holding up merge & release?

hartez:
> Hi @hartez , looks like this PR has had ample time for feedback (3 months).
> Is anything holding up merge & release?

API-wise, nothing. PR-wise, we still need to implement this for UWP and it probably needs some automated tests. I'll be doing that sometime in the next few weeks.

As of right now, we're targeting this feature to be released in 4.5.0.

rmarinho:
@hartez  so can we move this from draft to PR ?

hartez:
After I add the missing stuff we can.

jsuarezruiz:
@hartez Old draft with new CollectionView functionality. Should we close this PR and recover it in the future in .NET MAUI?

jfversluis:
> @hartez Old draft with new CollectionView functionality. Should we close this PR and recover it in the future in .NET MAUI?

Yep, this will not happen for Forms.

VincentH-Net:
@jfversluis how can we track including this in Maui? 
Is the `potential-maui-material` label all it takes or do we need to create an issue in [Maui](https://github.com/dotnet/maui)?

jfversluis:
That's the label I use that we can filter on so we can easily spot things that we potentially want to bring over. If you're dedicated and really interested in this one, feel free to open an issue. When you do; ideally bringing over the detail and revise that with potential new insights or ideas that you might have about it now.

I guess that will help in gauging engagement and make it a bit more visible. It seems EZ was already pretty happy with it so it _might_ speed things up a little.

PR #7330, opened 2019-08-30
Gif Animation Support

### LINKED ISSUE #1704: [F100] Basic GIF animation support

PureWeen:
### Description of Change ###

Adds basic GIF image animation support on UWP/IOS/Android. **NOTE**, according to issue this implementation intends to target small animated GIF's meaning that current implementation does **NOT** include support for caching or streaming. This could of course be added in future if needed and some preparations have been done to support streaming in implemented GIF decoder.

Implementation on UWP uses OS support for animated GIF's included in 1607 (anniversary update) or later. No support added to previous versions. The minimum UWP version supported by Xamarin Forms (master branch) should be 1607.

Implementation on IOS uses IOS GIF parser and key frame animations to present correctly timed key frames to image control.

Implementation on Android includes GIF parser used to parse, decode and create bitmaps from encoded in GIF sources. Implementation uses AnimationDrawable to present correctly timed animation buffers to image control. GIFs are currently only enabled on FastRenderers so if you have the Legacy Renderers flag set then GIFs won't play

### Issues Resolved ### 
- fixes #1704

### API Changes ###

New animation properties and methods on Image:
```C#

public bool IsAnimationPlaying;

```

Starting and stopping is controlled through *IsAnimationPlaying*. Originally we were going to use Start and Stop methods but all these did was set *IsAnimationPlaying* to true or false respectively. This way it can also be started and stopped via Binding!!

### Behavioral Changes ###

Nothing, if the animation auto play property has not been set then images will be loaded normal. Trying to use animation support on none GIF image sources will fail the load and fallback trying to load as regular image (Android) or as a 1 frame keyframe animation (IOS).

### PR Checklist ###


PureWeen:
@jsuarezruiz 

Rebased and fixed merge issues should be good to go now

jsuarezruiz:
Testing it I had some ideas (not for this PR). I add them:
- SpeedRatio: Be able to control the playback speed of the Gif.
- RepeatBehavior (OneTime, Forever or an integer with the number of times): Repeat mode.

Xamarin.Forms.ControlGallery.Android/Xamarin.Forms.ControlGallery.Android.csproj
```diff
@@ -45,7 +45,7 @@
     <Debugger>Xamarin</Debugger>
     <DevInstrumentationEnabled>True</DevInstrumentationEnabled>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <AndroidSupportedAbis>armeabi-v7a</AndroidSupportedAbis>
+    <AndroidSupportedAbis>armeabi-v7a;x86</AndroidSupportedAbis>
```
kingces95: Is adding `x86` intentional?

kingces95:
Am I holding it wrong? This is on Android 9.0 running master as of this comment. Expected if I fiddled with this I could get the gif moving... got the cold shoulder tho... :sad panda:

FA6BE123-04CC-4DA0-A77F-94A6F2187730

PureWeen:
@kingces95 it only works on android with fast renderers

So need to comment out the legacy renderers on the control gallery

kingces95:
Yea! That works for start and stop.

 I expected the `IsAnimationPlaying` to toggle on when I hit start and off when I hit stop and to be able to emulate start and stop by toggling that switch. Is that not how it's supposed to work? 

I expected switching on `IsAnimationAutoplay` to begin an animation if the gif is loaded but stopped. It actually only effects gifs transitioning from unloaded to loaded. That's not a big deal, but just wanted to be sure that's intentional.

Xamarin.Forms.ControlGallery.Android/Xamarin.Forms.ControlGallery.Android.csproj
PureWeen: Yea in trying to speed up master this got removed so deploying to android device on windows needs it

PureWeen:
> Shane said at one point he dropped IsAnimationAutoplay and instead allowed users to set IsAnimationPlaying before the image loaded as an indication to play once it's loaded. And come to think of it, what happens now if the user sets IsAnimationAutoplay to false but IsAnimationPlaying to true before the image loads? Even if the behavior is defined, it's kinda icky.

I remember now why the two exist.  If we just have *IsAnimationPlaying* it will toggle on then off so the developer has to keep resetting it to true if they want it played "onload"  *IsAnimationAutoplay* is a persistent behavior indicating how you want the gif to act onload. 

>  if the user sets IsAnimationAutoplay to false but IsAnimationPlaying to true before the image loads? 

It'll just play. that doesn't seem too icky. You're just triggering it to play once the first time

> I'd vote for just having IsAnimationPlaying. Might name it IsPlaying and then have a method Play with an overload that takes an enum Once (default), and Repeat. And god help us, can these gifs embed a sound track?

Methods aren't MVVM friendly though so it makes it weird for the MVVM people right?


@kingces95 @jsuarezruiz  One thing I'm wondering about is if these APIs should be shifted to an object instead of being directly on the Image itself?

``` XAML
<Image>
<Image.AnimationSettings />

```

Thoughts?


> RepeatBehavior (OneTime, Forever or an integer with the number of times): Repeat mode.

Yea I had this on the original API 
https://github.com/xamarin/Xamarin.Forms/pull/2202#discussion_r177960608

And it got voted down :-p

kingces95:
I'm fine with the API but I'll play devils advocate for a second... 

Can we describe an app experience that would use `IsAnimationAutoplay`? Basically, that means every time the GIF appears on screen it'll animate and then stop, right? So what justifies `IsAnimationAutoplay` and not `IsAnimationLoopPlay`? Isn't `IsAnimationAutoplay` just setting `IsAnimationPlaying` to `true` in `OnAppearing`? Or.. what event triggers the "autoplay" if not `OnAppearing`? Reloading the app from being backgrounded?

PureWeen:
@kingces95 I removed IsAnimationAutoPlaying :-) 

I think interpreting what a user wants that to mean is too difficult with a single property. 

Once we get this merged I could see some additional work around defining more features of playing the gif like repeat count

But for now just having the start stop toggle is sufficient

@jsuarezruiz @kingces95 please re-review 

I also had to fix an issue on UWP where it wasn't starting the animation so had to figure out a point in time to trigger it

jsuarezruiz:
Added support to gifs from Url on Android:
gif-url-droid

jsuarezruiz:
@samhouts Should retarget this to **4.4.0**?

PureWeen:
Rebased

Kicked off build

PR #7348, opened 2019-09-02
[Android/iOS/UWP] Added BarSelectedTextColorProperty for TabbedPage

### LINKED ISSUE #3281: [Android] TabbedPage Add `BarSelectedTextColor`
Currently there is no way to set  the BarSelectedTextColor which means if you are only using Text for Tabs there's no way to set the color so that Selected and Unselected look different

The default template for tabs doesn't have images set for Android icons so if the user tries to color the text then there's no way for them to color the text differently for unselected

ahoefling:
### Description of Change ###

Added new bindable property to the `TabbedPage` which provides granular control over the color of the icons and text in the `TabbedPage`. `BarSelectedTextColor` provides custom color to apply to the text on the selected tab page so the selected tab may be uniquely identified. 

### Issues Resolved ### 

- fixes #3281

### API Changes ###
Added:
 - `Color BarSelectedTextColor { get; set; } //Bindable Property`

#### New Rules
With the addition of `BarSelectedTextColor` there is now going to be color conflicts between the following properties
* BarSelectedTextColor/BarTextColor
* SelectedTabColor/UnselectedTabColor

I have adjusted the rules to handle several scenarios where some properties are set and other are set to default

| BarSelectedTextColor | BarTextColor | SelectedTabColor | UnselectedTabColor | Selected Icon Color                                   | Selected Text Color | Unselected Icon Color                                  | Unselected Text Color|
|----------------------|--------------|------------------|--------------------|-------------------------------------------------------|---------------------|--------------------------------------------------------|----------------------|
| Default              | Default      | Green            | Pink               | Green                                                 | Green               | Pink                                                   | Pink                 |
| Orange               | White        | Default          | Default            | (android) icon color from file. (iOS) uses text Color | Orange              | (android) icon color from file. (iOS) uses text color  | White                |
| Orange               | White        | Green            | Pink               | Green                                                 | Orange              | Pink                                                   | White                |

The reasoning behind this table is to not break existing expected behavior with `SelectedTabColor` and `UnselectedTabColor`. Currently those bindable properties will update the Text and Icon across the platform (except UWP). By only setting the `BarSelectedTextColor` if it is not default we have what hopefully is a non-breaking change. If there is a breaking API please make a review in the PR so we can adjust the code or discuss trade-offs.

**UWP**
- Currently there is no support for icon colors

**Android**
- If no TabColor specified the default color of the file is used

**iOS**
- If no tabColor is specified the text color will be applied

### Platforms Affected ### 

- Core
- iOS
- Android
- UWP

#### UWP Limitations
Since the concept of Tabs in UWP doesn't really exist, the `TabbedPageRenderer` and the BarSelectedTextColor will overwrite the value for `SelectedTabColor` property if not set. This is to force a consistent look and feel with the text color across the 3 major platforms.

As far as I can tell there is no way for UWP to set the icon color of a tab. When that gets implemented we should come back to this and update the rules for TabColor vs TextColor to match what iOS and Android are doing. 
Not Applicable

selectedTextColor

### Testing Procedure ###
1. Open Control Gallery Project
2. Navigate to Root Gallery Pages
3. Select 'TAB->CONTENT'

Now you are on the correct page and you can swap between the tabs. If you want to test different scenarios open `RootTabbedContentPage` in the shared code and you can edit the following code:

```
UnselectedTabColor = Color.HotPink;
SelectedTabColor = Color.Green;
BarTextColor = Color.White;
BarSelectedTextColor = Color.Orange;
```

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs
@@ -335,6 +335,7 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 				((IPageController)tabbedPage).InternalChildren.CollectionChanged += OnChildrenCollectionChanged;
 				UpdateBarBackgroundColor();
 				UpdateBarTextColor();
+				UpdateBarSelectedTextColor();
 				UpdateItemIconColor();
 				if (!isDesigner)
 				{
@@ -366,6 +367,8 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 			}
 			else if (e.PropertyName == PlatformConfiguration.AndroidSpecific.TabbedPage.IsSwipePagingEnabledProperty.PropertyName)
 				UpdateSwipePaging();
+			else if (e.PropertyName == TabbedPage.BarSelectedTextColorProperty.PropertyName)
+				UpdateBarSelectedTextColor();
 		}
 
 		void SetNavigationRendererPadding(int paddingTop, int paddingBottom)
@@ -950,6 +953,41 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 				_bottomNavigationView.ItemTextColor = colors;
 			else
 				_tabLayout.TabTextColors = colors;
+		}
+
+		protected virtual void UpdateBarSelectedTextColor()
+		{
+			AColor selectedColor;
+			AColor unselectedColor;
+
+			if (Element != null)
+			{
+				selectedColor = Element.BarSelectedTextColor.ToAndroid();
+				unselectedColor = Element.BarTextColor.ToAndroid();
+			}
+			else
+			{
+				selectedColor = ((Color)TabbedPage.BarSelectedTextColorProperty.DefaultValue).ToAndroid();
+				unselectedColor = ((Color)TabbedPage.BarTextColorProperty.DefaultValue).ToAndroid();
+			}
+
+			for (int index = 0; index < _tabLayout.TabCount; index++)
+			{
+				var tab = _tabLayout.GetTabAt(index);
+				var icon = tab.Icon;
+
+				if (icon == null)
+				{
+					_tabLayout.TabTextColors = GetColorStateList(unselectedColor, selectedColor);
+					break;
+				}
+				else
+				{
+					var color = tab.IsSelected ? selectedColor : unselectedColor;
+					icon = ADrawableCompat.Wrap(icon);
+					icon.SetColorFilter(color, PorterDuff.Mode.SrcIn);
+				}
+			}
 		}
 
 		void SetIconColorFilter(TabLayout.Tab tab)
--- a/Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs
@@ -367,7 +367,7 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 			}
 			else if (e.PropertyName == PlatformConfiguration.AndroidSpecific.TabbedPage.IsSwipePagingEnabledProperty.PropertyName)
 				UpdateSwipePaging();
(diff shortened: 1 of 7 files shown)


ahoefling:
Is there any action I need to take on this? I understand it is low priority, but wanted to check in to see if I need to do anything

jingliancui:
So Is this any progress about this issue ?

jsuarezruiz:
@ahoefling Could you rebase master to fix the conflict?. Thanks!.

ahoefling:
@jsuarezruiz I have rebased this on master

Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs
```diff
@@ -952,6 +955,25 @@ void UpdateBarTextColor()
...
+			AColor unselectedColor;
+
+			if (Element != null)
+			{
+				selectedColor = (Element.BarSelectedTextColor.IsDefault ? Element.SelectedTabColor : Element.BarSelectedTextColor).ToAndroid();
+				unselectedColor = (Element.BarTextColor.IsDefault ? Element.UnselectedTabColor : Element.BarTextColor).ToAndroid();
+			}
+			else
+			{
+				selectedColor = ((Color)TabbedPage.BarSelectedTextColorProperty.DefaultValue).ToAndroid();
```
samhouts, line 970: Do we need to do anything if the Element is null?

Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs
```diff
@@ -952,6 +955,25 @@ void UpdateBarTextColor()
...
 		}
 
+		protected virtual void UpdateBarSelectedTextColor()
+		{
+			AColor selectedColor;
+			AColor unselectedColor;
+
+			if (Element != null)
+			{
+				selectedColor = (Element.BarSelectedTextColor.IsDefault ? Element.SelectedTabColor : Element.BarSelectedTextColor).ToAndroid();
```
samhouts, line 965: I feel like this is going to make the text color the same as the tab background color, and thus not visible, if the text color isn't set. Is this really what it was doing before?

Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs
ahoefling: I think we should set the default value if `Element` is null which will give us consistency across the platforms. Currently the platform sets the native default value which is not consistent across iOS, Android and UWP. 

Going from memory here, but I believe the text color defaults is as follows:
- iOS - blue
- Android - white
- UWP - black

Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs
ahoefling: I had to re-read this code a few times to remember what I was doing and hook up the debugger. It isn't super clear and maybe we can come up with better variable names as it is one of those cases where every part of the variable name is super important

This code will use the selected/unselected Tab Color if there is no value set for the Text Color. The Tab Color is not the same as the background color but the color of the icon to make them match. If the selected text color or unselected text color is set it will use those values.

The idea here is to sync up the foreground color of the text and icon in the tab. If the code doesn't explicitly set the color it will match the icon. This is where `SelectedTabColor` and `UnselectedTabColor` come into play.

Did I clear things up or did I just make it more confusing 😕

jingliancui:
any update here?

mos379:
That would really be great to have resolved...

jsuarezruiz:
@ahoefling This is really nice. What do you think if we move it to the Toolkit as an Effect?

ahoefling:
I think that is a great idea! Is this something that we can have someone from the Microsoft team complete or do you need the community to submit the PR?

jfversluis:
This will not happen anymore for Xamarin.Forms. If this is still important to you, make sure to check the [.NET MAUI repo](https://github.com/dotnet/maui) and see if it's already on the roadmap or open an issue with a detailed feature request. Thanks!

PR #7353, opened 2019-09-02
[Core] added error codes

paymicro:
### Description of Change ###

Added error codes.
Error codes added without reference to any resources, because they are unique and rarely match the same on [other platforms (UWP & WPF)](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/).

### Issues Resolved ### 

- fixes [AB#826363](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/826363)

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

**Before**
Screenshot_4

**After**
Screenshot_3

### Testing Procedure ###

- open any xaml in VisualStudio for Windows
- add `<Button Clicked="BtClick1234"></Button>`
- build project

### PR Checklist ###

- [ ] Has automated tests 


StephaneDelcroix:
As there's no error table, or error link, adding an actual code doesn't provide any value to the developer, it just checks a box.

StephaneDelcroix:
If we go along with adding error codes, we should add them for:
- Xaml compilation
- exceptions happening when the previewer is running
- same for HotReload

Also, the error codes should match what UWP/WPF does in the same case, or default to CS#### ones in most of the cases. If we have very custom cases in which we throw an error, the prefix should be XF#### (there is precedent for that in the .targets file), and we'd need a table and error links for those.

And maybe also take the localization into account.

@mgoertz-msft is your team responsible for generating, or handling those errors ? If so, we should discuss this further.

@maddyleger1 this is XF interacting with the tooling, can you take the lead in organizing the discussion for this ? Thanks

mgoertz-msft:
@StephaneDelcroix I'm not sure what you mean by my team being _"responsible for generating, or handling those errors_".

The language service has its own validation code to produce errors for invalid XAML and to make sure we get squiggles and error list entries. Other components, such as compiler, previewer and hot reload would need to do the same.

Perhaps we could all use a shared MEF component though that abstracts away the differences between VS for Windows and VS for Mac.
cc @drewgillies and @TimBarham

TimBarham:
If nothing else, error codes would actually be very useful to the previewer... it could help us determine what to do with certain errors (rather than relying on the error string which could easily change or be localized).

And as I believe @mgoertz-msft is suggesting, the previewer and hot reload teams would be responsible for reporting those errors through the appropriate paths so we get squiggles and error list entries.

mgoertz-msft:
On Windows we have a host error service that the language service, the designer and hot reload use to report errors, and the service is responsible for creating error list entries and squiggles.

My proposal would be to create a new MEF component for that purpose that we all could use (one for Mac and one for Windows actually). It would then replace the service on the Windows side when we port the new language service back from Mac to Windows.

PR #7363, opened 2019-09-03
Check for null Platform in OnDestroy

### LINKED ISSUE #7331: [Bug] Android Activity crashes when calling finish before Forms.Init
### Steps to Reproduce

1. On Android Activity after base.OnCreate, call Finish and return 

### Expected Behavior
It should not need to call Xamarin.Forms.Init because it does not to render Xamarin page. The reason we are doing this is sometimes we are passing extra parameter to indicate the step has finished without rendering. This used to work prior to Xamarin 4.0. After upgrading to 4.2, it does not work anymore.

### Actual Behavior
Android App crashes on dispose. 

System.NullReferenceException: Object reference not set to an instance of an object.
at Xamarin.Forms.Platform.Android.AppCompat.Platform.op_Implicit (Xamarin.Forms.Platform.Android.AppCompat.Platform canvas) [0x00000] in D:\a\1\s\Xamarin.Forms.Platform.Android\AppCompat\Platform.cs:488 
08-30 21:17:42.294 I/MonoDroid(20609):   at Xamarin.Forms.Platform.Android.FormsAppCompatActivity.OnDestroy () [0x00025] in Xamarin.Forms.Platform.Android\AppCompat\FormsAppCompatActivity.cs:214 
at Android.App.Activity.n_OnDestroy (System.IntPtr jnienv, System.IntPtr native__this) [0x00009] in <fc1f547b51c6476996598e4a1bf2fc5f>:0 
 at (wrapper dynamic-method) Android.Runtime.DynamicMethodNameCounter.9(intptr,intptr)

### Basic Information

- Version with issue:
- Last known good version:
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android: 8.1
  - UWP:   
- Nuget Packages: Xamarin.Form version 4.2.0.618605
- Affected Devices:

[Bug7331.zip](https://github.com/xamarin/Xamarin.Forms/files/3561683/Bug7331.zip)

I was not able to compile Xamarin Forms to Nuget package. Here is the link with the potential fix. 
https://github.com/acvien1234/Xamarin.Forms.git

slang25:
Missing null check is causing NRE under some circumstances.

Fixes #7331

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
@@ -278,9 +278,11 @@ namespace Xamarin.Forms.Platform.Android
 
 			PopupManager.Unsubscribe(this);
 
-			_layout.RemoveView(Platform);
-
-			Platform?.Dispose();
+			if (Platform != null)
+			{
+				_layout.RemoveView(Platform);
+				Platform.Dispose();
+			}
 
 			PreviousActivityDestroying.Set();
 


slang25:
@kvpt Is this reasonable, or is there some missing cleanup in this case that needs to be understood?

kvpt:
@slang25 Platform here is the root renderer and it was not removed from the layout, so I added this.
But I doesn't checked that if platform was null it cause an NRE, it's my fault, sorry.
Indeed if you call Destroy before the Init method is done it can be the case.
This change is the right thing to do. Thanks.

slang25:
Thank you @kvpt, do no be sorry, it was a good change 😃

slang25:
Thanks @samhouts, I'll take a look this evening and see if I can figure out what's going on.

PR #7365, opened 2019-09-03
fix back icon color for the navigation page

### LINKED ISSUE #5177: [Enhancement] Change NavigationPage back button color on Android

ravinderjangra:
### Description of Change ###
fixes: #5177

### Issues Resolved ### 
fixes #5177

### API Changes ###
 None

### Platforms Affected ### 
- Android

None

**Before**
Screenshot_1567510174

**After**
Screenshot_1567510005

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
@@ -1031,8 +1031,9 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 
 			if (_toolbar.NavigationIcon != null && !textColor.IsDefault)
 			{
-				var drawable = (_toolbar.NavigationIcon as DrawerArrowDrawable);
-				drawable.Color = textColor.ToAndroid().ToArgb();
+				var icon = _toolbar.NavigationIcon as DrawerArrowDrawable;
+				if (icon != null)
+					icon.Color = textColor.ToAndroid().ToArgb();
 			}
 
 			UpdateTitleIcon();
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs
```diff
@@ -1027,6 +1027,12 @@ void UpdateToolbar()
 
 			bar.Title = currentPage.Title ?? "";
 
+			if (_toolbar.NavigationIcon != null && !textColor.IsDefault)
+			{
+				var drawable = (_toolbar.NavigationIcon as DrawerArrowDrawable);
```
jsuarezruiz: Just include a null validation like here: https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs#L502

```
var icon = _toolbar.NavigationIcon as DrawerArrowDrawable;
if (icon == null)
     return;
```

ravinderjangra: @jsuarezruiz, why should it return from there? I understand having the null check but I don't understand the return because we still have some function calls after this `if` block like `UpdateTitleView()`.

jsuarezruiz: Right, is necessary the null check but not the return (I see that we updated the TitleView below).

samhouts:
@ravinderjangra Can you please rebase on master to fix the build? Thanks!

Issue #5815, opened 2019-04-04
Update instructions for building Xamarin.Forms


adrianknight89:
@davidortinau Now that VSTS is Azure DevOps, I believe the [instructions](https://github.com/xamarin/Xamarin.Forms/wiki/How-to-Build-Xamarin.Forms-NuGet) need to be updated. The user interface of DevOps looks significantly different and makes the existing instructions difficult to follow. Please make sure the new instructions mention building an XF fork as well as the master branch. Most of us will probably need to build our own forks.

adrianknight89:
#3430

Issue #3430, opened 2018-07-29
Add VSTS YAML Configuration


dansiegel:
### Description

Now that VSTS supports YAML configurations, a YAML build configuration should be added to the repo to better assist the community in being able to run custom builds.

#### Goals

By adding a YAML configuration this should help to do the following:

1. Remove barriers of entry for the F100 by ensuring developers can properly build/test before submitting a PR
1. Attract more adoption by making enterprises feel they can patch and run custom builds for their needs with minimal effort
1. Help developers by getting more 1st hand visibility into how Xamarin.Forms itself is built and then tested.

rmarinho:
it's on my backlog!

adrianknight89:
Please also update [this page](https://github.com/xamarin/Xamarin.Forms/wiki/How-to-Build-Xamarin.Forms-NuGet) when you're done.

samhouts:
closed by #4385

PR #7367, opened 2019-09-03
Make WkWebViewRenderer default for iOS WebView

### LINKED ISSUE #7323: [Xamarin.iOS] Deprecated API Usage UIWebView
## Before Commenting please read this comment
https://github.com/xamarin/Xamarin.Forms/issues/7323#issuecomment-542363338

---

Dear Developer,

We identified one or more issues with a recent delivery for your app, "xxxx" 1.0.11 (1.0.11). Your delivery was successful, but you may wish to correct the following issues in your next delivery:

ITMS-90809: Deprecated API Usage - Apple will stop accepting submissions of apps that use UIWebView APIs . See https://developer.apple.com/documentation/uikit/uiwebview for more information.

After you’ve corrected the issues, you can use Xcode or Application Loader to upload a new binary to App Store Connect.

Best regards,

The App Store Team

But I am sure I replaced UIWebView with WKWebView(WKWebViewRenderer).

### LINKED ISSUE #7327: [Bug] Shell Application crash when clicking hardware back button on android
### Description

Shell Application crash when clicking hardware back button on android (back to device homepage), the error shows `System.NullReferenceException`, but it occurs on latest Xamarin.Forms (4.2.0.709249) , i test with the version `4.1.0.555618` and  no crash happens , i think it's a bug. 

### Steps to Reproduce

1. Download official sample :https://docs.microsoft.com/en-us/samples/xamarin/xamarin-forms-samples/userinterface-xaminals/ .
2. Upload Xamarin.Forms to the latest .
3. Test on android,  when locates at root page , click hardware back button .

### Expected Behavior

Back to HomePage of mobile phone.

### Actual Behavior

Crash occurs.

### Basic Information

- IDE:  Visual studio 2019 16.1.3
- Xamarin.Android : 9.3.0.22 
- Nuget Packages: Xamarin.Forms (4.2.0.709249)

jfversluis:
### Description of Change ###

Deprecates the `WebViewRenderer` for iOS and uses the `WkWebViewRenderer` instead.

No (extra) automated tests were added at this stage, the ones in place should succeeded with this replacement.

To reinstate the current behavior just add the line below to the `AssemblyInfo.cs` file of you iOS project. This will register the `WebViewRenderer` again over the `WkWebViewRenderer`.

`[assembly: ExportRenderer(typeof(Xamarin.Forms.WebView), typeof(Xamarin.Forms.Platform.iOS.WebViewRenderer))]`

### Issues Resolved ### 

- fixes #7323 

### API Changes ###

 
 None

### Platforms Affected ### 

- iOS

None. Everything should function as before. Users should receive build warnings when they reference the `WebViewRenderer` directly. E.g. when there is a custom renderer in place.

Not applicable

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Stubs/Xamarin.Forms.Platform.cs
+++ b/Stubs/Xamarin.Forms.Platform.cs
@@ -72,9 +72,13 @@ namespace Xamarin.Forms.Platform
 	[RenderWith (typeof (SliderRenderer))]
 	internal class _SliderRenderer { }
 
+#if !__IOS__
 	[RenderWith (typeof (WebViewRenderer))]
 	internal class _WebViewRenderer { }
-
+#else
+	[RenderWith (typeof (WkWebViewRenderer))]
+	internal class _WebViewRenderer { }
+#endif
 	[RenderWith (typeof (SearchBarRenderer))]
 	internal class _SearchBarRenderer { }
 
--- a/Stubs/Xamarin.Forms.Platform.cs
+++ b/Stubs/Xamarin.Forms.Platform.cs
@@ -72,13 +72,14 @@ namespace Xamarin.Forms.Platform
 	[RenderWith (typeof (SliderRenderer))]
 	internal class _SliderRenderer { }
 
-#if !__IOS__
-	[RenderWith (typeof (WebViewRenderer))]
-	internal class _WebViewRenderer { }
-#else
+#if __IOS__
 	[RenderWith (typeof (WkWebViewRenderer))]
 	internal class _WebViewRenderer { }
+#else
+	[RenderWith(typeof(WebViewRenderer))]
+	internal class _WebViewRenderer { }
 #endif
+
 	[RenderWith (typeof (SearchBarRenderer))]
 	internal class _SearchBarRenderer { }
 
--- a/Xamarin.Forms.Platform.iOS/Properties/AssemblyInfo.cs
+++ b/Xamarin.Forms.Platform.iOS/Properties/AssemblyInfo.cs
@@ -58,4 +58,3 @@ using UIKit;
 [assembly: Xamarin.Forms.Dependency(typeof(Deserializer))]
 [assembly: Xamarin.Forms.Dependency(typeof(ResourcesProvider))]
 [assembly: ResolutionGroupName("Xamarin")]
-[assembly: Preserve]
--- a/Xamarin.Forms.Platform.iOS/Properties/AssemblyInfo.cs
+++ b/Xamarin.Forms.Platform.iOS/Properties/AssemblyInfo.cs
@@ -58,3 +58,4 @@ using UIKit;
 [assembly: Xamarin.Forms.Dependency(typeof(Deserializer))]
 [assembly: Xamarin.Forms.Dependency(typeof(ResourcesProvider))]
 [assembly: ResolutionGroupName("Xamarin")]
+[assembly: Preserve]
--- a/Xamarin.Forms.ControlGallery.iOS/PerformanceTrackerRenderer.cs
+++ b/Xamarin.Forms.ControlGallery.iOS/PerformanceTrackerRenderer.cs
@@ -908,7 +908,7 @@ namespace Xamarin.Forms.ControlGallery.iOS
 		}
 	}
 
-	public class PerformanceTrackingWebView : WebViewRenderer, IDrawnObservable
+	public class PerformanceTrackingWebView : WkWebViewRenderer, IDrawnObservable
 	{
 		readonly SubviewWatcher<PerformanceTrackingWebView> _watcher;
 		int _Drawn;
--- a/Xamarin.Forms.ControlGallery.iOS/local.html
+++ b/Xamarin.Forms.ControlGallery.iOS/local.html
@@ -5,6 +5,6 @@
 <body>
 <h1 id="#LocalHtmlPage">Xamarin.Forms</h1>
 <p>This is a local iOS Html page</p>
-<a href="https://www.google.com">Go to Google</a>
+<a id="#LocalHtmlPageLink" href="https://www.google.com">Go to Google</a>
 </body>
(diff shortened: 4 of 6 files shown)


jsuarezruiz:
To add a little more context to this PR. Some devs have started receiving this warning on App submissions:
 
_"ITMS-90809: Deprecated API Usage - Apple will stop accepting submissions of apps that use UIWebView APIs . See https://developer.apple.com/documentation/uikit/uiwebview for more information."_

With this PR, deprecates the `WebViewRenderer` for iOS and uses the `WkWebViewRenderer` instead.

adrianknight89:
I thought the OP said he was already using `WkWebViewRenderer` when he got the warning. If that's the case, shouldn't the old renderer be completely removed from the source code instead of being marked obsolete?

jfversluis:
I think that is something we need to investigate. I also see some reports of people implementing a renderer with `WkWebViewRenderer` and the message does disappear.

Removing the old renderer altogether is not something that I would want to do. That will immediately break peoples code whenever they have any reference whatsoever to that class.

If there is no reference to it, it should be removed by the compiler and all should be good.

PureWeen:
@jfversluis 

> If there is no reference to it, it should be removed by the compiler and all should be good.

I ran some tests with this and it is not removed.

If you run a default iOS app currently with full linker on even the *WkWebViewRenderer* sticks around if you aren't using it. The intent of the RenderWith attributes was to link anything out for elements you didn't use but I don't think that behavior is currently working correctly.

So I don't think this PR will really achieve anything currently

adrianknight89:
So, what now? Fix the linker behavior or introduce a breaking change and remove the old renderer?

PureWeen:
> Fix the linker behavior or introduce a breaking change and remove the old renderer?

My current thought would be to figure out why those renderers aren't getting linked out (which I've been working on a bit)

In that case we could switch the default type (like in this PR) and then leave the WebViewRenderer in. If people need to revert back to *WebViewRenderer* they can add the Export for it but if people aren't using it then hopefully it gets linked out.

Stubs/Xamarin.Forms.Platform.cs
```diff
@@ -72,9 +72,13 @@ internal class _CarouselViewRenderer { }
 	[RenderWith (typeof (SliderRenderer))]
 	internal class _SliderRenderer { }
 
+#if !__IOS__
```
StephaneDelcroix: avoid using !condition, esp when there's a else

PureWeen:
@jfversluis can you rebase to master? I just merged 4.3.0 up so we should be able to get this through the UI tests

In theory for this change to get rid of the warning we also have to delete this out of the AssemblyInfo
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Properties/AssemblyInfo.cs#L61

Though that currently does nothing (as we know and bug logged with iOS)

So we might want to include deleting that as part of this PR and working through whatever needs to be worked through with the linker to make sure our code still works.

PureWeen:
@jfversluis never mind about removing

https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Properties/AssemblyInfo.cs#L61

The WebViewRenderer will never get linked out no matter how much it wants to.
So let's just merge this after it's rebased and the UITests pass

Xamarin.Forms.Platform.iOS/Properties/AssemblyInfo.cs
```diff
@@ -57,5 +57,4 @@
 [assembly: InternalsVisibleTo("Xamarin.Forms.Material")]
 [assembly: Xamarin.Forms.Dependency(typeof(Deserializer))]
 [assembly: Xamarin.Forms.Dependency(typeof(ResourcesProvider))]
-[assembly: ResolutionGroupName("Xamarin")]
-[assembly: Preserve]
```
PureWeen: Just leave this

jfversluis:
Short summary: there is no way to get the (UI)WebViewRenderer linked out through the mechanism we have in place. As a first step to swap out the WebViewRenderer and WkWebViewRenderer we would merge this PR. 

People submitting apps to Apple will still receive the warning, but it's just that: a warning. Probably by the time iOS 14 starts knocking on the door Apple will start actually rejecting apps still referencing UIWebView.

Because we can't link out the (UI)WebViewRenderer, we need to delete it as a whole. This will be a breaking change for everyone still referencing it. In a follow up PR somewhere in the future we will need to do this or come up with something better. For now, this is the solution to at least warn our users that they should start removing references to this renderer.

adrianknight89:
@PureWeen @jfversluis For those of us who want to remove the old renderer completely, you could fix #5815 so that we could use Azure DevOps for custom builds. Right now, I use cake to build locally, but I'd rather integrate with DevOps.

This page needs to be updated to remove VSTS:

https://github.com/xamarin/Xamarin.Forms/wiki/How-to-Build-Xamarin.Forms-NuGet

#3430 was fixed, but this was ignored.

PureWeen:
@adrianknight89 working on that over here

https://github.com/xamarin/Xamarin.Forms/pull/7458

Once I have it to a point where there's a path that the public can use I'll let you know and you can let me know how it works out for you

adrianknight89:
Thanks Shane!

jfversluis:
Fixed the tests for iOS. It seems UITest also has some issues with UIWebView -> WKWebView. I will report that to them.

Namely:
- The `WebView` shorthand method defaults to `UIWebView`
- The `Css` method does not seem to detect all the elements anymore. There for I had to change the assert to the `a` element instead of the `h1` which is entirely not detected anymore
- `Tap` on a CSS selected element does not seem to work. The coordinates are detected but a tap is never executed

jfversluis:
Failed tests are not related (CarouselView one)

PR #7383, opened 2019-09-04
Fixed the rounding issue when the stepper value is incremented.

### LINKED ISSUE #5168: [Core] Incorrect increments in Stepper

CliffAgius:
When the increment value is small like say 0.5 per step. This is due to Double being used as the value so when it's unboxed from the binding object precision is lost.

### Description of Change ###
In the Stepper bindable property for the value this is a Double so when the value is unboxed from the Object it's not 100% precise such that when the stepper is used in the platform renderer of say Android 0.5 - 0.5 doesn't result in zero but some slight non-zero value.

### Issues Resolved ### 
- fixes #5168 

### API Changes ###
None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

To fix the issue Math.Round has been used to a precision of 6 decimal places so if a used was to use the stepper with a precision greater than this they will face an issue.  But lets be honest they are then using the wrong tool for the job...

Not applicable

### Testing Procedure ###
Use the test case g5168 on your chosen platform to confirm that the error has been fixed.

### PR Checklist ###

- [x] Has automated tests 


AndreiMisiukevich:
Hm, not sure it's a proper fix: https://github.com/xamarin/Xamarin.Forms/pull/5192

CliffAgius:
Yes having read that long back and forth I agree my fix is nothing more than a patch to hide the problem.

Ideally as mentioned it should have been a decimal type from the beginning but to change now would be a breaking change.

I like the idea of having a count tracking the number of up/down clicks and deriving the value based on the count times the interval.

If this idea has no objections I will take another bash as my first PR.

samhouts:
@CliffAgius We would be interested in another approach if you have one! Thanks!

CliffAgius:
@samhouts any idea what was wrong with this approach? I changed from the Math.Round() method to using an internal counter as suggested by @StephaneDelcroix & @PureWeen  #5192.
If neither of these methods suits then I'm not sure what the team are aiming for as the best option is to change the Double for Decimal but this would be a breaking change so not ideal.

Happy to take another swing at this but it may need a little more design if there is an underlying issue the team wish to fix at the same time.

This is my first attempt at a Code PR to Xamarin so just looking for a bit of guidance if you don't mind please. 

Thanks.

samhouts:
Ack, I missed your last commits! Sorry about that! I'll reopen for review. :)

AndreiMisiukevich:
@CliffAgius please fix merge conflicts
thx)

CliffAgius:
Sorry not sure what happen there but fixed now.

jfversluis:
Seems like unit tests are now failing because of this @CliffAgius 

> Xamarin.Forms.Core.UnitTests\StepperUnitTests.cs(181,12): Error CS1955: Non-invocable member 'Stepper.Value' cannot be used like a method.
Xamarin.Forms.Core.UnitTests\StepperUnitTests.cs(182,12): Error CS1955: Non-invocable member 'Stepper.Value' cannot be used like a method.
Xamarin.Forms.Core.UnitTests\StepperUnitTests.cs(183,12): Error CS1955: Non-invocable member 'Stepper.Value' cannot be used like a method.
Xamarin.Forms.Core.UnitTests\StepperUnitTests.cs(184,12): Error CS1955: Non-invocable member 'Stepper.Value' cannot be used like a method.
Xamarin.Forms.Core.UnitTests\StepperUnitTests.cs(185,12): Error CS1955: Non-invocable member 'Stepper.Value' cannot be used like a method.
Xamarin.Forms.Core.UnitTests\StepperUnitTests.cs(186,12): Error CS1955: Non-invocable member 'Stepper.Value' cannot be used like a method.
Xamarin.Forms.Core.UnitTests\StepperUnitTests.cs(187,12): Error CS1955: Non-invocable member 'Stepper.Value' cannot be used like a method.
Xamarin.Forms.Core.UnitTests\StepperUnitTests.cs(188,12): Error CS1955: Non-invocable member 'Stepper.Value' cannot be used like a method.

CliffAgius:
Hmmm... Ok will take a look.

Xamarin.Forms.Core.UnitTests/StepperUnitTests.cs
```diff
@@ -172,5 +172,22 @@ public void StepperValueChangedEventArgs (double initialValue, double finalValue
...
+			var stepper = new Stepper(0, 10, 0, 0.5);
+
+			stepper.Value(0.5);
+			stepper.Value(1.0);
+			stepper.Value(1.5);
+			stepper.Value(2.0);
+			stepper.Value(1.5);
+			stepper.Value(1.0);
+			stepper.Value(0.5);
+			stepper.Value(0.0);
```
AndreiMisiukevich: Value is a property of a "double" type.

Also, what does this UnitTest do?

CliffAgius:
@AndreiMisiukevich @jfversluis 
Sorry turns out that my install of VS19 was creating problems, what ever changes I made to the stepper.cs file when the tests were run they didn't get used so the tests always passed.  I re-installed VS19 and it's back to working correctly now so the tests failed (Correctly) and I have fixed the issues including a rounding error in using the (int) type conversion for the StepperPosition. However sadly I am now seeing tests fail in my other projects so not sure what VS19 has been hiding...

@AndreiMisiukevich the test you referenced was not supposed to have made it to the PR I was attempting to make a test that checked the stepper function.  But as VS19 said all tests passed I wrongly assumed all tests were OK and really should have checked back on it so sorry about that.

I think this 3rd time lucky should have it I have tested many times with lots of different values and all the tests now (Correctly) pass on my machine and I have debugged and stepped thru them all to be sure.

Thanks for sticking with this PR and me being a bit of a Noob at this OpenSource lark, but assuming this passes Ok I have 2 others I am looking at.

Cliff.

CliffAgius:
@samhouts @StephaneDelcroix just wondering if this PR is Ok or needs more work? It's been a few months now.

Thanks.

Cliff.

Xamarin.Forms.Core/Stepper.cs
```diff
@@ -43,6 +45,8 @@ public class Stepper : View, IElementConfiguration<Stepper>
 
 		public static readonly BindableProperty IncrementProperty = BindableProperty.Create("Increment", typeof(double), typeof(Stepper), 1.0);
 
+		public static readonly BindableProperty StepperPositionProperty = BindableProperty.Create("StepperPosition", typeof(int), typeof(Stepper), 0);
```
jfversluis:
```suggestion
		public static readonly BindableProperty StepperPositionProperty = BindableProperty.Create(nameof(StepperPosition), typeof(int), typeof(Stepper), 0);
```

CliffAgius:
@jfversluis Thanks for the review looks like my first code PR to Xamarin might just be about to happen... Yay!!!

jfversluis:
Oh this is your first? No way! I though you had done something before. Awesome, thank you so much! And sorry for the wait

CliffAgius:
@jfversluis This was my first and I did another after that is still awaiting review, as I had two awaiting I stopped looking at other to make sure I had the process correct, if they are both accepted then there are others I will look at as well.  The other issue if your willing to 😉 look is:

Adds Cookies to the WebView fixing #Issue3262 #8169

Xamarin.Forms.Core/Stepper.cs
```diff
@@ -41,7 +43,9 @@ public class Stepper : View, IElementConfiguration<Stepper>
 				eh(stepper, new ValueChangedEventArgs((double)oldValue, (double)newValue));
 		});
 
-		public static readonly BindableProperty IncrementProperty = BindableProperty.Create("Increment", typeof(double), typeof(Stepper), 1.0);
+		public static readonly BindableProperty IncrementProperty = BindableProperty.Create(nameof(Increment), typeof(double), typeof(Stepper), 1.0);
+
+		public static readonly BindableProperty StepperPositionProperty = BindableProperty.Create(nameof(StepperPosition), typeof(int), typeof(Stepper), 0);
```
samhouts: Does this need to be a new public bindable property?

CliffAgius: Hi @samhouts the increment is obviously needed but the StepperPositionProperty I flip/flopped on this a few times myself.  I settled on it being a Public property in the end as it was suggested to me that someone might want an Int position of the stepper rather than having to do math using Doubles to work out where the stepper is.  Foe example they know the stepper steps at 0.125 and has 50 positions rather than work out the maths for 2/5ths they can just set it to 20 or equally read 20 and know it's 2/5th of the way.

Of course if this needs to be changed then I am happy to change after all I flip/flopped quite a few times on this.

Ta

Cliff.

samhouts: I'm not opposed to new API, but I do prefer splitting API changes into separate PRs from bug fixes whenever possible!

CliffAgius: @samhouts I have made the property private as suggested for this PR if it's accepted I will open another issue/PR to revert this change as I think it may be helpful in some situations.  As you say it's an API change but not a breaking change.

lathan86:
Assembly: Xamarin.Forms.Core, Version=2.0.0.0 
nuget: xamarin.forms\4.6.0.800\lib\netstandard2.0\Xamarin.Forms.Core.dll

There is still a problem with the Stepper when the increment is smaller than 0.5.
The rounding are totaly wrong per increment.
as example:
increment: 0.06 => 0.0199999999999996, 0.0800000000000001, 0.139999999999999, 0.199999999999999, ....
increment: 0.05 => 0, 0.0500000000000007, 0.100000000000001, 0.15, ....
and so on..
can you please check.

jfversluis:
@lathan86 If you're seeing issues with this please open a new issue describing the problem and attach a reproduction. Thanks!

StephaneDelcroix:
I fear that this was wrong. see #11169, the inability to set small MinValue, the counter going off if the increment is changed, or the logic being wrong if the counter is manually set.

Xamarin.Forms.Core.UnitTests/StepperUnitTests.cs
```diff
@@ -160,17 +163,35 @@ public void StepperValueChangedEventArgs (double initialValue, double finalValue
...
-			Assert.AreEqual (finalValue, newValue);
+			Assert.AreEqual(stepper, stepperFromSender);
+			Assert.AreEqual(initialValue, oldValue);
+			Assert.AreEqual(finalValue, newValue);
+		}
+
+		[TestCase(10)]
+		public void TestReturnToZero(int steps)
+		{
+			var stepper = new Stepper(0, 10, 0, 0.5);
```
StephaneDelcroix, line 183: this test is passing without the change (.5 has an exact fp representation)

CliffAgius: Wasn't the issue reported that it was numbers less than 0.5 and with large Min/Max values, these tests are not testing that case?
I think the issue was reported with 
Min = -100000
Max = 100000
Step = Any value less than 0.5

StephaneDelcroix: don't bother, those are notes to myself, and history. I'm handling this. But I reverted your code, and this test was still passing.

lathan86: > Wasn't the issue reported that it was numbers less than 0.5 and with large Min/Max values, these tests are not testing that case?
> I think the issue was reported with
> Min = -100000
> Max = 100000
> Step = Any value less than 0.5

step value range was between -1 and 1 and larg min and max values

PR #7389, opened 2019-09-05
Update support libraries from 28.0.0.1 to 28.0.0.3

PureWeen:
### Description of Change ###
Support library version bump

### DIFF
--- a/.nuspec/Xamarin.Forms.Maps.nuspec
+++ b/.nuspec/Xamarin.Forms.Maps.nuspec
@@ -18,7 +18,7 @@
       </group>
       <group targetFramework="MonoAndroid81">
         <dependency id="Xamarin.GooglePlayServices.Maps" version="60.1142.1"/>
-        <dependency id="Xamarin.Android.Support.v7.AppCompat" version="28.0.0.1"/>
+        <dependency id="Xamarin.Android.Support.v7.AppCompat" version="28.0.0.3"/>
         <dependency id="Xamarin.Forms" version="$version$"/>
       </group>
       <group targetFramework="tizen40">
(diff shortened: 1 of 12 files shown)


PureWeen:
Relevant UI Tests all green

PureWeen:
All Green against 4.2

adrianknight89:
I'm on Windows. Do you all develop on a Mac? What does your setup look like? I'd rather use VS on Windows than VSMAC.

PureWeen:
I use both but that was the only issue I could think of you might be having.

VS for windows and our CI are all compiling 4.2.0+ without any issues.

Maybe @StephaneDelcroix has some additional insights to share about that particular exception?

> The "ValidateOnly" parameter is not supported by the "XamlCTask" task. Verify the parameter exists on the task, and it is a settable public instance property. Xamarin.Forms.Controls C:\Xamarin.Forms.nuspec\Xamarin.Forms.targets 116

PureWeen:
@adrianknight89 when doing the clean are you leaving the IDE open by chance?

You could also try clearing out the nuget cache
*nuget locals all -clear*

adrianknight89:
I can't clean when the IDE is open since some files are locked by VS. At the moment, I'm not sure what the issue is, but I ended up creating a new VM and was able to proceed with clean-installed VS.

PureWeen:
@adrianknight89 sometimes VS does like to hold onto those files in the .nuspec

Usually if I close all the VS instances/vs code/emulators it'll eventually free it up

https://docs.microsoft.com/en-us/sysinternals/downloads/process-explorer 
is also a good one for finding what's holding onto the files

Once you can run a full clean on that folder it'll most likely start working :-)

adrianknight89:
@PureWeen Another thing: Looks like this is part of 4.4.0.1153-nightly now, but I believe 28.0.0.3 was unlisted from the public. How do you expect people to update their support libraries to .3?

https://www.nuget.org/packages/Xamarin.Android.Support.v4/28.0.0.3

PureWeen:
Those will get listed once we list a stable XF release that uses those which should be soon.

For now if you just delete all support library package references from the csproj file or manually update them it should work fine

PR #7398, opened 2019-09-05
[RefreshView] Fire RefreshCommand setting IsRefreshing

### LINKED ISSUE #7388: [Bug] IsRefreshing and RefreshCommand aren't working so well together
### Description

I checked in a branch here where you can reproduce a couple issues
https://github.com/xamarin/Xamarin.Forms/commit/3c2d5ffee8d5b4f62683a721c441dfed22a9db26

### Steps to Reproduce

RefreshCommand doesn't fire if you set IsRefreshing

1. Set a break point inside the command assigned to RefreshView 
2. click *Toggle Refresh*
3. Break point is never hit

Setting IsRefreshing to False inside RefreshCommand doesn't do anything

1. Set a break point inside the command assigned to RefreshView 
2. trigger the refresh but swiping down 
3. Break point is hit
4. Notice that IsREfreshing here is set to False (it should be set to True)
5. Since it's set to false setting it to true doesn't do anything

### Reproduction Link
https://github.com/xamarin/Xamarin.Forms/compare/refreshview_fix

jsuarezruiz:
### Description of Change ###

Fixes some issues:
- RefreshCommand doesn't fire if you set IsRefreshing.
- Setting IsRefreshing to False inside RefreshCommand doesn't do anything.

### Issues Resolved ### 

- fixes #7388

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android

Setting IsRefreshing to true will execute RefreshCommand (if it exists).

#### Before
refresh-droid-before
(Same behavior on iOS).

#### After
ios-refresh-after
refresh-droid-after

### Testing Procedure ###
Open Core Gallery and navigate to the Test Cases. Search by "Refresh" and open the RefreshView Tests. Pressing the "Toggle Refresh" button or pulling to refresh must change the IsRefreshing property value.

### PR Checklist ###

- [ ] Has automated tests 

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/RefreshViewTests.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/RefreshViewTests.cs
@@ -27,7 +27,10 @@ namespace Xamarin.Forms.Controls.Issues
 		{
 			Title = "Refresh View Tests";
 			var scrollViewContent =
-				new StackLayout();
+				new StackLayout()
+				{
+					AutomationId = "LayoutContainer"
+				};
 
 			Enumerable.Range(0, 1000).Select(_ => new Label() { Text = "Pull me down to refresh me" })
 				.ForEach(x => scrollViewContent.Children.Add(x));
@@ -74,6 +77,17 @@ namespace Xamarin.Forms.Controls.Issues
 		public void IsRefreshingAndCommandTest()
 		{
 			RunningApp.Tap(q => q.Button("Toggle Refresh"));
+			RunningApp.WaitForElement(q => q.Marked("IsRefreshing: True"));
+			RunningApp.Screenshot("Refreshing");
+			RunningApp.WaitForElement(q => q.Marked("IsRefreshing: False"));
+			RunningApp.Screenshot("Refreshed");
+		}
+
+		[Test]
+		public void IsRefreshingAndCommandTest_SwipeDown()
+		{
+			RunningApp.WaitForElement(q => q.Marked("IsRefreshing: False"));
+			RunningApp.ScrollDown("LayoutContainer", ScrollStrategy.Gesture);
 			RunningApp.WaitForElement(q => q.Marked("IsRefreshing: True"));
 			RunningApp.Screenshot("Refreshing");
 			RunningApp.WaitForElement(q => q.Marked("IsRefreshing: False"));
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/RefreshViewTests.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/RefreshViewTests.cs
@@ -29,7 +29,6 @@ namespace Xamarin.Forms.Controls.Issues
 			var scrollViewContent =
 				new StackLayout()
 				{
-					AutomationId = "LayoutContainer"
 				};
 
 			Enumerable.Range(0, 1000).Select(_ => new Label() { Text = "Pull me down to refresh me" })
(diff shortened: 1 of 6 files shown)


adrianknight89:
@jsuarezruiz @PureWeen I was thinking of something recently, but I haven't had the chance to test it. What happens if we set 'IsEnabled' to false?

Doc:

> Gets or sets a value indicating whether the user can interact with RefreshView.

Since CollectionView is nested under RefreshView, if I set IsEnabled to false, I'd expect everything to be disabled so that no touch gestures are processed (including gestures on the CollectionView). If IsEnabled enables/disables the RefreshView but does not affect gestures on the CollectionView, I think Xamarin needs to make this clear in the official documentation.

Another idea is to rename this property (e.g. IsPullToRefreshEnabled) and make sure IsEnabled from the base ContentView functions the way we normally expect to see (i.e. propagate enabled/disabled state down to its child content). Since we're still in preview, it might be the time to re-think this. ;)

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/RefreshViewTests.cs
```diff
@@ -0,0 +1,62 @@
...
+		{
+			Title = "Refresh View Tests";
+
+			_refreshView = new RefreshView()
+			{
+				Content = new ScrollView(),
+				Command = new Command(async () =>
+				{
+					await Task.Delay(2000);
+					_refreshView.IsRefreshing = false;
```
PureWeen: can we also make sure this works without being after a task delay?

Maybe we could even add a check in here when the command is first executed that will throw an exception if *_refreshView.IsRefreshing* is true?

PureWeen:
Fixed the merge conflics
@jsuarezruiz can you add automated UI tests and then we should be good to go

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Xamarin.Forms.Controls.Issues.Shared.projitems
```diff
@@ -1043,6 +1043,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)Issue7053.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Issue6894.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Issue6929.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)RefreshViewTests.cs" />
```
PureWeen:
```suggestion
```

Xamarin.Forms.Core.UITests.Shared/UITestCategories.cs
```diff
@@ -56,5 +56,6 @@ internal static class UITestCategories
 		public const string Page = "Page";
 		public const string RefreshView = "RefreshView";
 		public const string TitleView = "TitleView";
+		public const string RefreshView = "RefreshView";
```
PureWeen:
```suggestion
```

jsuarezruiz:
@PureWeen Added UITest!.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/RefreshViewTests.cs
```diff
@@ -69,5 +69,16 @@ protected override void Init()
 				}
 			};
 		}
+#if UITEST
+		[Test]
+		public void IsRefreshingAndCommandTest()
```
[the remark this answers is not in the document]
PureWeen: I added an AutomationId on the iOS side on this PR
https://github.com/xamarin/Xamarin.Forms/pull/7514/files#diff-198607f4ce97efcc515a271473524b08R1673

Which should let us locate the control

PR #7458, opened 2019-09-09
Setup public and provisionator based provisioning of SDKS

PureWeen:
### Description of Change ###

This PR sets up two paths for provisioning
- Public path all public sdks
  - This uses Cake to install the sdks versions desired. I realize we had stepped away from using cake for this and just used inline scripts on the YAML file but I'd really prefer to use cake because of the reduced amount of effort it takes to make it work and for it to work cross platform. For example provisioning android APIs is a lot easier with @Redth 's cake plugins than using command line. Plus with cake it makes it easier if people want to run the provision task from the time line
- Internal path for master branches and alpha testing
  - This is the path used when we want to use provisionator and need to install things that we can only access internally. This is the path we will use for our weekly tests against the master branches of the sdks and whatever the latest xcode version is

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS
- Android
- UWP

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -62,7 +62,7 @@ jobs:
     buildPlatform: $(DefaultBuildPlatform)
     provisionatorPath : 'build/provisioning/provisioning.csx'
     DOTNET_VERSION: ${{ coalesce(variables.DOTNET_VERSION, variables.DOTNET_VERSION_DEFAULT) }}
-    nugetVersion: ${{ coalesce(variables['NUGET_VERSION'], variables['NUGET_VERSION_DEFAULT']) }}
+    NUGET_VERSION: ${{ coalesce(variables['NUGET_VERSION'], variables['NUGET_VERSION_DEFAULT']) }}
     
 - template: build/steps/build-android.yml
   parameters:
@@ -73,6 +73,8 @@ jobs:
     androidProjectArguments: '/t:"Rebuild;SignAndroidPackage" /bl:$(Build.ArtifactStagingDirectory)/android-legacy.binlog'
     buildConfiguration: $(DefaultBuildConfiguration)
     provisionatorPath : 'build/provisioning/provisioning.csx'
+    DOTNET_VERSION: ${{ coalesce(variables.DOTNET_VERSION, variables.DOTNET_VERSION_DEFAULT) }}
+    NUGET_VERSION: ${{ coalesce(variables['NUGET_VERSION'], variables['NUGET_VERSION_DEFAULT']) }}
 
 - template: build/steps/build-android.yml
   parameters:
@@ -83,6 +85,8 @@ jobs:
     androidProjectArguments: '/t:"Rebuild;SignAndroidPackage" /p:DefineConstants="TRACE DEBUG FORMS_APPLICATION_ACTIVITY APP" /bl:$(Build.ArtifactStagingDirectory)/android-preappcompact.binlog'
     buildConfiguration: $(DefaultBuildConfiguration)
     provisionatorPath : 'build/provisioning/provisioning.csx'
+    DOTNET_VERSION: ${{ coalesce(variables.DOTNET_VERSION, variables.DOTNET_VERSION_DEFAULT) }}
+    NUGET_VERSION: ${{ coalesce(variables['NUGET_VERSION'], variables['NUGET_VERSION_DEFAULT']) }}
 
 - template: build/steps/build-android.yml
   parameters:
@@ -93,6 +97,8 @@ jobs:
     androidProjectArguments: '/t:"Rebuild;SignAndroidPackage" /p:DefineConstants="TRACE DEBUG TEST_EXPERIMENTAL_RENDERERS APP" /bl:$(Build.ArtifactStagingDirectory)/android-newrenderers.binlog'
     buildConfiguration: $(DefaultBuildConfiguration)
     provisionatorPath : 'build/provisioning/provisioning.csx'
+    DOTNET_VERSION: ${{ coalesce(variables.DOTNET_VERSION, variables.DOTNET_VERSION_DEFAULT) }}
+    NUGET_VERSION: ${{ coalesce(variables['NUGET_VERSION'], variables['NUGET_VERSION_DEFAULT']) }}
 
 - job: osx
   displayName: OSX Phase
@@ -109,7 +115,8 @@ jobs:
     provisionator.signPath : 'build/provisioning/provisioning_sign.csx'
     buildConfiguration: $(DefaultBuildConfiguration)
     slnPath: $(SolutionFile)
-    nugetVersion: $(NUGET_VERSION)
+    DOTNET_VERSION: ${{ coalesce(variables.DOTNET_VERSION, variables.DOTNET_VERSION_DEFAULT) }}
+    NUGET_VERSION: ${{ coalesce(variables['NUGET_VERSION'], variables['NUGET_VERSION_DEFAULT']) }}
     iOSCertSecureFileName: 'Xamarin Forms iOS Certificate.p12'
     iOSProvisioningSecureFileName: 'Xamarin Forms iOS Provisioning.mobileprovision'
   steps:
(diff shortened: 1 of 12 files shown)


build.cake
```diff
@@ -49,13 +55,55 @@ Task("Clean")
...
+        try { AcceptLicenses (androidSdkSettings); } catch { }
+
+        AndroidSdkManagerInstall (new [] {
+                "platforms;android-29"
+            }, androidSdkSettings);
+
+        var platform = IsRunningOnWindows() ? "windows" : "macos";
+        await Boots ($"https://aka.ms/xamarin-android-commercial-d16-2-{platform}");
+
+        string monoVersion = "6.0.0.319";
```
StephaneDelcroix: can't we keep the monoVersion as a variable at the top-level ?

build.cake
```diff
@@ -49,13 +55,55 @@ Task("Clean")
...
+        if(IsRunningOnWindows())
+        {
+            string monoPath = $"{System.IO.Path.GetTempPath()}/mono.msi";
+            DownloadFile(
+                $"https://download.mono-project.com/archive/6.0.0/windows-installer/mono-{monoVersion}-x64-0.msi",
```
StephaneDelcroix: the 6.0.0 part of the path is hardcoded, make it a top-level variable so it's easier to update in the future

PureWeen: https://github.com/xamarin/Xamarin.Forms/pull/7458/files#diff-4d3634747586b964df9f7712bc3ec669R49

build.cake
PureWeen: https://github.com/xamarin/Xamarin.Forms/pull/7458/files#diff-4d3634747586b964df9f7712bc3ec669R49

build.cake
```diff
@@ -49,9 +74,81 @@ Task("Clean")
 {
     CleanDirectories("./**/obj", (fsi)=> !fsi.Path.FullPath.Contains("XFCorePostProcessor") && !fsi.Path.FullPath.StartsWith("tools"));
     CleanDirectories("./**/bin", (fsi)=> !fsi.Path.FullPath.Contains("XFCorePostProcessor") && !fsi.Path.FullPath.StartsWith("tools"));
-
 });
 
+Task("provision-macsdk")
+    .Does(async () =>
+    {
+        if(!IsRunningOnWindows() && String.IsNullOrWhiteSpace(macSDK))
```
PureWeen:
```suggestion
        if(!IsRunningOnWindows() && !String.IsNullOrWhiteSpace(macSDK))
```

PureWeen:
I broke up the provisioning tasks on cake

https://github.com/xamarin/Xamarin.Forms/pull/7458/files#diff-4d3634747586b964df9f7712bc3ec669R120

Which will let us easily do things like install mono if we need to roll back like we've been having to odo

azure-pipelines.yml
```diff
@@ -11,12 +11,18 @@ variables:
...
   value: $[counter('$(Build.SourceBranchName)_counter', 1)]
-- name: MONO_VERSION
-  value: 5_18_1
 - name: XCODE_VERSION
   value: 10.2
 - name: NUGET_VERSION
-  value: 5.0.2
+  value: 5.2.0
+- name: DOTNET_VERSION
+  value: 2.1.801
```
PureWeen:
```suggestion
  value: 3.0.100
```

build/steps/build-osx.yml
```diff
@@ -3,17 +3,28 @@ steps:
...
   - task: xamops.azdevex.provisionator-task.provisionator@1
     displayName: Provisionate Xamarin
-    condition: eq(variables['provisioningOSX'], 'true')
+    condition: eq(variables['provisioning'], 'true')
     inputs:
       provisioning_script: $(provisionator.osxPath)
       provisioning_extra_args: $(provisionator.extraArguments)
+      
+  - task: Bash@3
+    condition: eq(variables['provisioning'], 'false')
```
rmarinho: If it's false we call provision target?

PureWeen: So this probably need some naming work :-) 

The provisioning variable maps more to "use internal Provisionator" to do the provisioning

If it's false then it just installs public SDKs. 

*useProvisionator*
*enableInternalProvisioning*

?

rmarinho: maybe call it cakeProvisioning ?

azure-pipelines.yml
```diff
@@ -11,12 +11,18 @@ variables:
...
   value: $[counter('$(Build.SourceBranchName)_counter', 1)]
-- name: MONO_VERSION
-  value: 5_18_1
 - name: XCODE_VERSION
   value: 10.2
 - name: NUGET_VERSION
-  value: 5.0.2
+  value: 5.2.0
+- name: DOTNET_VERSION
+  value: 3.0.100
```
PureWeen:
```suggestion
  value: 2.1.801
```

CI currently doesn't support dotnet 3 until we have msbuild 16.3

azure-pipelines.yml
```diff
@@ -11,12 +11,18 @@ variables:
...
 - name: BuildVersion
   value: $[counter('$(Build.SourceBranchName)_counter', 1)]
-- name: MONO_VERSION
-  value: 5_18_1
 - name: XCODE_VERSION
   value: 10.2
 - name: NUGET_VERSION
-  value: 5.0.2
+  value: 5.2.0
+- name: DOTNET_VERSION
```
rmarinho: I wonder if we should pin the variables here on in AzureDev Ops

> You can choose which variables are allowed to be set at queue time and which are fixed by the pipeline author. If a variable appears in the variables block of a YAML file, it is fixed and cannot be overridden at queue time. To allow a variable to be set at queue time, make sure it doesn't appear in the variables block of a pipeline or job. You can set a default value in the editor, and that value can be overridden by the person queuing the pipeline.

PureWeen: So that's why whenever I set these they never change.

Yea I think not pinning these here would be good or fixing the code here so that this value is overridable

Like rename this to *DOTNET_VERSION_DEFAULT*

And then if there's an environment variable set for *DOTNET_VERSION* then we use that otherwise we use the default.

Is there a way to set a variable conditionally if it's not specified by the pipeline? Or is that just something the scripts and YAML tasks will need to account for?

PureWeen: alright @rmarinho until this gets resolved

https://developercommunity.visualstudio.com/content/problem/741877/variables-and-parameters-are-not-being-expanded-in.html

Let's just leave this as is

in theory it's setup to work so once the issue above is resolved then we can specify versions as parameters

build/steps/build-osx.yml
PureWeen: I removed the provisioning variable at the job level so now it's settable from the build queue variables

I think provisioning is fine for now on the variable name since the only purpose is to turn on provisionator and cake is just permanently on. Which is what we want for always installing stable bits for consistent builds and then we'll use provisionator on alpha branches for alpha testing

PureWeen:
Alright @rmarinho :-) This one is back to being ready to go

The cake and provisionator sdks are currently set to the 16.3 versions of everything
I tested this on CI by setting provisioning to true and it built with provisionator without any issues.

azure-pipelines.yml
```diff
@@ -11,13 +11,15 @@ variables:
...
 - name: BuildVersion
   value: $[counter('$(Build.SourceBranchName)_counter', 1)]
-- name: MONO_VERSION
-  value: 5_18_1
-- name: XCODE_VERSION
-  value: 10.2
 - name: NUGET_VERSION
-  value: 5.0.2
-
+  value: 5.2.0
```
PureWeen:
```suggestion
  value: 5.0.2
```

azure-pipelines.yml
```diff
@@ -11,13 +11,15 @@ variables:
...
   value: $[counter('$(Build.SourceBranchName)_counter', 1)]
-- name: MONO_VERSION
-  value: 5_18_1
-- name: XCODE_VERSION
-  value: 10.2
 - name: NUGET_VERSION
   value: 5.0.2
-
+- name: DOTNET_VERSION
+  value: 2.1.801
```
PureWeen:
```suggestion
  value: 2.2.402
```

PR #7465, opened 2019-09-10
Added IndicatorsView 

### LINKED ISSUE #6577: [Spec] IndicatorsView
# IndicatorsView

Provide a list of Indicators (shapes) to represent a number of items and a current position. This will be mainly to work with `CarouselView` #4996 

# API

## IndicatorsView
```csharp
public class IndicatorsView : View
{
	 public static readonly BindableProperty IndicatorsColorProperty = BindableProperty.Create(nameof(IndicatorsColor), typeof(Color), typeof(IndicatorsView), Color.Silver);

	public Color IndicatorsColor
	{
		get { return (Color)GetValue(IndicatorsColorProperty); }
		set { SetValue(IndicatorsColorProperty, value); }
	}

	public static readonly BindableProperty SelectedIndicatorColorProperty = BindableProperty.Create(nameof(SelectedIndicatorColor), typeof(Color), typeof(IndicatorsView), Color.Gray);

	public Color SelectedIndicatorColor
	{
		get { return (Color)GetValue(SelectedIndicatorColorProperty); }
		set { SetValue(SelectedIndicatorColorProperty, value); }
	}

	public static readonly BindableProperty IndicatorsShapeProperty = BindableProperty.Create(nameof(IndicatorsShape), typeof(IndicatorsShape), typeof(IndicatorsView), IndicatorsShape.Circle);

	public IndicatorsShape IndicatorsShape
	{
		get { return (IndicatorsShape)GetValue(IndicatorsShapeProperty); }
		set { SetValue(IndicatorsShapeProperty, value); }
	}

       public DataTemplate IndicatorTemplate
	{
		get { return (DataTemplate)GetValue(IndicatorTemplate Property); }
		set { SetValue(IndicatorTemplate Property, value); }
	}
	public static readonly BindableProperty CountProperty = BindableProperty.Create(nameof(Count), typeof(int), typeof(IndicatorsView), default(int));

	public int Count
	{
		get => (int)GetValue(CountProperty);
		set => SetValue(CountProperty, value);
	}

	 public static readonly BindableProperty MaxVisibleCountProperty = BindableProperty.Create(nameof(MaxVisibleCount), typeof(int), typeof(IndicatorsView), -1);

	public int MaxVisibleCount
	{
		get => (int)GetValue(MaxVisibleCountProperty);
		set => SetValue(MaxVisibleCountProperty, value);
	}

	public static readonly BindableProperty PositionProperty = BindableProperty.Create(nameof(Position), typeof(int), typeof(IndicatorsView), default(int), BindingMode.TwoWay, propertyChanged: OnPositionPropertyChanged);

	public int Position
	{
		get => (int)GetValue(PositionProperty);
		set => SetValue(PositionProperty, value);
	}

	public event EventHandler<PositionChangedEventArgs> PositionChanged;
}
```
### Properties

| API | Description |
| IndicatorsColor | Color of the indicators  |
| SelectedIndicatorColor | Color of the indicator that shows the selected item  |
| IndicatorShape | Shape to render on Indicator  |
| IndicatorTemplate | Template to render on Indicator  |
| Count | Number of indicators  |
| MaxVisibleCount | Number of visible indicators, defaults to -1 so Count is used  |
| Position | Current selected indicator index  |

## IndicatorShape
Specify how the Indicator should look, Circle or Square
```csharp
public enum IndicatorShape
{
    Circle,
    Square
}
```

# Scenarios
```csharp
var indicatorsView = new IndicatorsView
{
	BackgroundColor = Color.Red,
	IndicatorsTintColor = Color.Blue,
        SelectedIndicatorTintColor = Color.Yellow,
	IndicatorsShape = IndicatorsShape.Square,
	HorizontalOptions = LayoutOptions.Center,
	VerticalOptions = LayoutOptions.End,
	HeightRequest = 30,
	InputTransparent = true
};
```
```
	IndicatorsView.SetItemsSourceBy(indicatorsView, carouselView);
``` 
# Backward Compatibility
New control so no issues.

# Difficulty : [medium]

AndreiMisiukevich:
# Description of Change ###
Added new view for representing CarouselView indicators

### Issues Resolved ### 

- fixes #6577 

### API Changes ###

Added:
 - IndicatorsView

### Platforms Affected ### 
- Core

There is the ability to add indicators

with IndicatorsView
image

### Testing Procedure ###
Enable CarouselView experimental flag, go to CarouselView gallery and run an Indicators test case
(keep in mind that currently there is an issue with CarouselView and sometimes some cards are not rendered )


Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,536 @@
...
+			var count = 0;
+			var enumerator = (sender as IEnumerable)?.GetEnumerator();
+			while (enumerator?.MoveNext() ?? false)
+			{
+				count++;
+			}
+			Count = count;
+		}
+
+		// Code below is copied from StackLayout
```
adrianknight89: Can you remove these comments?

AndreiMisiukevich: @adrianknight89 yeah) we noticed it almost simultaneously )
done

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+		{
+			(bindable as IndicatorsView)?.ResetIndicatorsStyles();
+		});
+
+		internal static readonly BindableProperty ItemsSourceProperty = BindableProperty.Create(nameof(ItemsSource), typeof(IEnumerable), typeof(IndicatorsView), null, propertyChanged: (bindable, oldValue, newValue) =>
+		{
+			(bindable as IndicatorsView)?.ResetItemsSource(oldValue as IEnumerable);
+		});
+
+		static IndicatorsView()
```
adrianknight89: Why do we need a static constructor here?

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+		{
+			(bindable as IndicatorsView)?.ResetIndicatorsStyles();
+		});
+
+		public static readonly BindableProperty SelectedIndicatorColorProperty = BindableProperty.Create(nameof(SelectedIndicatorColor), typeof(Color), typeof(IndicatorsView), Color.Gray, propertyChanged: (bindable, oldValue, newValue) =>
+		{
+			(bindable as IndicatorsView)?.ResetIndicatorsStyles();
+		});
+
+		public static readonly BindableProperty IndicatorsSizeProperty = BindableProperty.Create(nameof(IndicatorsSize), typeof(double), typeof(IndicatorsView), 10.0, propertyChanged: (bindable, oldValue, newValue) =>
```
adrianknight89: There seems to be some inconsistency with naming things. Can we make `IndicatorsSize` and `IndicatorsColor` singular? Maybe drop the `s`? Also, since these bindable properties along with others are part of `IndicatorsView`, I'm wondering if we should drop the word `Indicator` from the property names.

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+			set => SetValue(PositionProperty, value);
+		}
+
+		public int Count
+		{
+			get => (int)GetValue(CountProperty);
+			set => SetValue(CountProperty, value);
+		}
+
+		public int MaxVisibleCount
```
adrianknight89: Maybe this needs to be renamed too. We have `MinimumHeightRequest`. Not sure if we should stick to the non-abbreviated name here and use `MaximumVisibleCount`.

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+			set => SetValue(SelectedIndicatorColorProperty, value);
+		}
+
+		public double IndicatorsSize
+		{
+			get => (double)GetValue(IndicatorsSizeProperty);
+			set => SetValue(IndicatorsSizeProperty, value);
+		}
+
+		[EditorBrowsable(EditorBrowsableState.Never)]
```
adrianknight89: I think this might need to be removed for intellisense to work. (not sure 100% though)

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: Don't want to load type "beforefieldsinit" (keen to lazy init)

https://stackoverflow.com/questions/610818/what-does-beforefieldinit-flag-do

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: @adrianknight89 yeah, makes sense. I decided to follow @rmarinho spec
but it is a good point indeed

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: Yep, I use the full word "Maximum" for my control in CardsView plugin, but i followed spec

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: Since spec doesn't have this property, I decided to hide it (ItemsSourceProperty is internal.) I cannot make it internal (bindings won't' work)

Xamarin.Forms.Core/IndicatorsView.cs
hartez: Unless @rmarinho has an objection, I think we should use `Maximum`.

AndreiMisiukevich:
@hartez @adrianknight89 @rmarinho 
As for SelectedIndicatorColor and IndicatorColor...
With this approach, we can manage background colors only.

In my control, I have two properties (SelectedIndicatorStyle and UnselectedIndicatorStyle).
So users can manage BackgroundColor, Size, OutlineColor etc. it's super flexible

Is it redundant stuff you? your thoughts

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+	{
+		Circle,
+		Square
+	}
+
+	public class IndicatorsView : StackLayout
+	{
+		public static readonly BindableProperty PositionProperty = BindableProperty.Create(nameof(Position), typeof(int), typeof(IndicatorsView), default(int), BindingMode.TwoWay, propertyChanged: (bindable, oldValue, newValue) =>
+		{
+			(bindable as IndicatorsView)?.ResetIndicatorsStyles();
```
StephaneDelcroix: use direct cast

StephaneDelcroix: use a lambda, or a static method, but avoid `{}` in BP definition

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+		});
+
+		public static readonly BindableProperty IndicatorShapeProperty = BindableProperty.Create(nameof(IndicatorShape), typeof(IndicatorShape), typeof(IndicatorsView), default(IndicatorShape), propertyChanged: (bindable, oldValue, newValue) =>
+		{
+			(bindable as IndicatorsView)?.ResetIndicatorsStyles();
+		});
+
+		public static readonly BindableProperty IndicatorsColorProperty = BindableProperty.Create(nameof(IndicatorsColor), typeof(Color), typeof(IndicatorsView), Color.Silver, propertyChanged: (bindable, oldValue, newValue) =>
```
StephaneDelcroix: default Color should be Color.Default, and handled at the time it's shown

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+		{
+			(bindable as IndicatorsView)?.ResetIndicatorsStyles();
+		});
+
+		public static readonly BindableProperty SelectedIndicatorColorProperty = BindableProperty.Create(nameof(SelectedIndicatorColor), typeof(Color), typeof(IndicatorsView), Color.Gray, propertyChanged: (bindable, oldValue, newValue) =>
```
StephaneDelcroix: default Color should be Color.Default, and handled in the renderer

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+
+namespace Xamarin.Forms
+{
+	public enum IndicatorShape
+	{
+		Circle,
+		Square
+	}
+
+	public class IndicatorsView : StackLayout
```
StephaneDelcroix: I'd use a singular `IndicatorView`

Xamarin.Forms.Core/IndicatorsView.cs
StephaneDelcroix: default size should probably be -1

Xamarin.Forms.Core/IndicatorsView.cs
StephaneDelcroix: please drop the static constructor

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+			(bindable as IndicatorsView)?.ResetItemsSource(oldValue as IEnumerable);
+		});
+
+		static IndicatorsView()
+		{
+		}
+
+		public IndicatorsView()
+		{
+			Orientation = StackOrientation.Horizontal;
```
StephaneDelcroix: this won't play nice with styling, but anyway, we should drop the SL inheritance

Xamarin.Forms.Core/IndicatorsView.cs
StephaneDelcroix: there's no semantic reason for this to inherit from StackLayout. it would mean one could use the StackLayout API to add animated GIFs cats to the IndicatorView.

Inherit from View, and maintain an internal StackLayout.

or way better, inherit from TemplatedView

Xamarin.Forms.Core/IndicatorsView.cs
StephaneDelcroix: don't stick to the spec this much

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public IEnumerable ItemsSource
+		{
+			get => GetValue(ItemsSourceProperty) as IEnumerable;
+			set => SetValue(ItemsSourceProperty, value);
+		}
+
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public new IList<View> Children => base.Children;
```
StephaneDelcroix: should be Items, once the SL inheritance is dropped

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+
+		public static void SetItemsSourceBy(IndicatorsView indicatorsView, CarouselView carouselView)
```
StephaneDelcroix: naming is very confusing. and although I understand the use of such an helper method, I don't think it should be part of the class. maybe move it to an extensions class instead ?

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+			=> indicatorsView.SetItemsSourceBy(carouselView);
+
+		public void SetItemsSourceBy(CarouselView carouselView)
```
StephaneDelcroix: same

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+			});
+		}
+
+		protected virtual void ApplySelectedStyle(View view, int index)
+			=> view.BackgroundColor = SelectedIndicatorColor;
+
+		protected virtual void ApplyUnselectedStyle(View view, int index)
+			=> view.BackgroundColor = IndicatorsColor;
+
+		private void AddExtraIndicatorsItems()
```
StephaneDelcroix: drop the `private`

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+
+			if (ItemsSource is INotifyCollectionChanged observableCollection)
+			{
+				observableCollection.CollectionChanged += OnCollectionChanged;
+			}
+
+			OnCollectionChanged(ItemsSource, new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
+		}
+
+		private void OnCollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
```
StephaneDelcroix: we can do better than that, this doesn't handle elements moved in a collection

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+		{
+			foreach (var item in Children.Where((v, i) => i >= Count).ToArray())
+			{
+				Children.Remove(item);
+			}
+		}
+
+		private int IndexOf(View view) => Children.IndexOf(view);
+
+		private void ApplyStyle(View view)
```
StephaneDelcroix: the Style name is unfortunate here

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+		});
+
+		public static readonly BindableProperty MaxVisibleCountProperty = BindableProperty.Create(nameof(MaxVisibleCount), typeof(int), typeof(IndicatorsView), int.MaxValue, propertyChanged: (bindable, oldValue, newValue) =>
+		{
+			(bindable as IndicatorsView)?.ResetIndicatorsStyles();
+		});
+
+		public static readonly BindableProperty IndicatorTemplateProperty = BindableProperty.Create(nameof(IndicatorTemplate), typeof(DataTemplate), typeof(IndicatorsView), default(DataTemplate), propertyChanged: (bindable, oldValue, newValue) =>
+		{
+			(bindable as IndicatorsView)?.ResetIndicatorsStyles();
```
StephaneDelcroix: if the template wasn't set, it doesn't look like the dots will be replaced

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+		});
+
+		public static readonly BindableProperty IndicatorShapeProperty = BindableProperty.Create(nameof(IndicatorShape), typeof(IndicatorShape), typeof(IndicatorsView), default(IndicatorShape), propertyChanged: (bindable, oldValue, newValue) =>
```
StephaneDelcroix: I'd drop this, stick with Circles, or template items

Xamarin.Forms.Core/IndicatorsView.cs
```diff
@@ -0,0 +1,285 @@
...
+		}
+
+		public double IndicatorsSize
+		{
+			get => (double)GetValue(IndicatorsSizeProperty);
+			set => SetValue(IndicatorsSizeProperty, value);
+		}
+
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public IEnumerable ItemsSource
```
StephaneDelcroix: what's the point of the ItemsSource, as the context is never applied ?

AndreiMisiukevich: it was added for getting the count of dots (bound to CarouselView ItemsSource) instead of binding with a converter (ItemsSource to Count via converter)

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: agreed, i'll try

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: > use direct cast

is it code style or there are any reasons?

In terms of "code readability", i think the best idea would be adding an extension method (As<T>(object o) where T: class)

and using it: bindable.As<IndicatorsView>()?.ResetIndicatorsStyles()

anyway I will amend

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: Sorry, could you please explain?

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: @StephaneDelcroix so circles are by default but everybody can change them by setting IndicatorTemplate, right?

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: hmmm, in this case, Size can be different depending on conditions
I'd like to see indicators with constant size

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: @StephaneDelcroix yeah, i thought about it. We could add protected virtual method for handling default SL orientation value.
But as you mentioned, we are refusing SL inheritance at all

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: @hartez @adrianknight89  well, let's go ahead with Maximum

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: sure
I had an idea, that uses coud extend and override these methods for customizing "style". Not back color only

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: Hmmmm, why?

if ItemsSource is INotifyCollectionChanged it means that CollectionChanged event will be fired with **Move** action, in case of any element is moved

Am i wrong?

Xamarin.Forms.Core/IndicatorsView.cs
StephaneDelcroix: should be:

```suggestion
			((IndicatorsView)bindable).ResetIndicatorsStyles();
```

like it's done everywhere else. I prefer a failing cast exception than continuing with a property being set on the wrong control

AndreiMisiukevich: makes sense

Xamarin.Forms.Core/IndicatorsView.cs
adrianknight89: Also drop the above private keywords. There's a lot more throughout the code.

AndreiMisiukevich: @adrianknight89 nice note, sure

AndreiMisiukevich:
> @AndreiMisiukevich Please rebase to 4.3.0. Thanks!!

sure, I will rebase and apply fixes per comments on the weekend

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: removed plural

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: dropped

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: done

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: made ItemsSource public

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: done

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: done

Xamarin.Forms.Core/IndicatorsView.cs
AndreiMisiukevich: done

AndreiMisiukevich:
@StephaneDelcroix @hartez @rmarinho @adrianknight89 
updated, please re-review

@samhouts rebased and retargeted

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -0,0 +1,256 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using System.Collections.Specialized;
+using System.Linq;
+using Xamarin.Forms.Platform;
+using static Xamarin.Forms.IndicatorView;
+
+namespace Xamarin.Forms
+{
+	public static class IndicatorViewExtensions
```
adrianknight89: I'd move this class to another file. It's odd to me that this is the first class I look at when I open `IndicatorView.cs`.

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -0,0 +1,256 @@
...
+			var index = Items.IndexOf(view);
+			if (index == Position)
+			{
+				ApplySelectedColor(view);
+				return;
+			}
+			ApplyRegularColor(view);
+		}
+
+		void ApplySelectedColor(View indicatorItemView)
```
adrianknight89: Do we really need `ApplySelectedColor`, `ApplyRegularColor`, and `GetColorOrDefault`? These are one-line methods and we can use their code directly in the callers instead of extracting everything into separate methods.

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -0,0 +1,256 @@
...
+			=> ((IndicatorView)bindable).ResetItemsSource((IEnumerable)oldValue));
+
+		public static readonly BindableProperty IndicatorLayoutProperty = BindableProperty.Create(nameof(IndicatorLayout), typeof(Layout<View>), typeof(IndicatorView), null, propertyChanged: TemplateUtilities.OnContentChanged);
+
+		public IndicatorView()
+			=> IndicatorLayout = new StackLayout { Orientation = StackOrientation.Horizontal };
+
+		IList<View> Items => IndicatorLayout.Children;
+
+		public Layout<View> IndicatorLayout
```
adrianknight89: What is the purpose of `IndicatorLayout` on iOS now that you're using a `UIPageControl`? Is this `StackLayout` being used for anything? How is it being used for Android?

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -0,0 +1,256 @@
...
+			{
+				ApplyColor(child as View);
+			}
+
+			IndicatorLayout.IsVisible = Count > 1 || !HidesForSingleIndicator;
+		}
+
+		void ResetItemsSource(IEnumerable oldItemsSource)
+		{
+			if (oldItemsSource is INotifyCollectionChanged oldObservableCollection)
```
adrianknight89: I'd name this `collection` instead of `oldObservableCollection` or `observableCollection` down below. Technically, there are different types of collections that implement `INotifyCollectionChanged` without being named observable.

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -0,0 +1,256 @@
...
+				observableCollection.CollectionChanged += OnCollectionChanged;
+			}
+
+			OnCollectionChanged(ItemsSource, new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
+		}
+
+		void AddExtraIndicatorItems()
+		{
+			var oldCount = Items.Count;
+			for (var i = 0; i < Count - oldCount && i < MaximumVisibleCount - oldCount; ++i)
```
adrianknight89: Convention: post increment `i` instead of pre-incrementing it.

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -0,0 +1,256 @@
...
+			{
+				var size = IndicatorSize > 0 ? IndicatorSize : 10;
```
adrianknight89: The default value for `IndicatorSize` is -1. Why don't you make it 10? Also, you can validate/coerce it for <= 0 values so that you don't have to do it here.

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -0,0 +1,256 @@
...
+				var indicator = IndicatorTemplate?.CreateContent() as View ?? new Frame
+				{
+					Padding = 0,
+					HasShadow = false,
+					BorderColor = Color.Transparent,
+					VerticalOptions = LayoutOptions.Center,
```
adrianknight89: Do we need `VerticalOptions` and `HorizontalOptions` here?

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -0,0 +1,256 @@
...
+					HorizontalOptions = LayoutOptions.Center,
+					WidthRequest = size,
+					HeightRequest = size,
+					CornerRadius = (float)size / 2
+				};
+				var itemTapGesture = new TapGestureRecognizer();
```
adrianknight89: I'd name this `tapGestureRecognizer` and rename the below arguments to `sender` and `args`.

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -0,0 +1,256 @@
...
+				itemTapGesture.Tapped += (tapSender, tapArgs) => Position = Items.IndexOf(tapSender as View);
+				indicator.GestureRecognizers.Add(itemTapGesture);
+				Items.Add(indicator);
+			}
+		}
+
+		void RemoveRedundantIndicatorItems()
+		{
+			foreach (var item in Items.Where((v, i) => i >= Count).ToArray())
```
adrianknight89: I'd avoid LINQ here because it's difficult to read in its current form and there may be slight performance overhead in using it. You could create a for loop instead and remove the redundant items. Also, does it matter if you remove the items from the beginning or end of the item source?

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -0,0 +1,256 @@
...
+			{
+				Items.Remove(item);
+			}
+		}
+
+		void ApplyColor(View view)
```
adrianknight89: Can we rename this to `SetBackgroundColor`?

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -0,0 +1,256 @@
...
+				var indicator = IndicatorTemplate?.CreateContent() as View ?? new Frame
+				{
+					Padding = 0,
+					HasShadow = false,
+					BorderColor = Color.Transparent,
+					VerticalOptions = LayoutOptions.Center,
+					HorizontalOptions = LayoutOptions.Center,
+					WidthRequest = size,
+					HeightRequest = size,
+					CornerRadius = (float)size / 2
```
adrianknight89: To get a perfect circular `Frame`, `CornerRadius` needs to be exactly half the size of frame width and height. Since `IndicatorSize` is a double and you're doing a division here while casting the result to float, I believe you might end up with rounding issues in some cases and unexpected "borders". @davidortinau Is the Drawing Spec still on? Wish we could get our own primitive shapes.

Xamarin.Forms.Platform.iOS/Renderers/IndicatorViewRenderer.cs
```diff
@@ -0,0 +1,70 @@
...
+
+		void UpdatePages()
+			=> Control.Pages = Element.Count;
+
+		void UpdateHidesForSinglePage()
+			=> Control.HidesForSinglePage = Element.HidesForSingleIndicator;
+
+		void UpdatePagesIndicatorTintColor()
+		{
+			_defaultPagesIndicatorTintColor = _defaultPagesIndicatorTintColor ?? Control.PageIndicatorTintColor;
```
adrianknight89: Setting `_defaultPagesIndicatorTintColor` and `_defaultCurrentPagesIndicatorTintColor` can be moved inside `OnElementChanged`. Here, the assignment checks will be made every time colors change, but we really need to get the default colors only once.

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: @adrianknight89 agreed. It's a good point. But I don't know what could be the best place for this file

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: @adrianknight89 you are right. They are presented because previously they were protected virtual. I decided to hide them for now. Of course, I can drop them)

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: @adrianknight89 on iOS it is not used. But I really don't know how to manage such cases.

1) iOS uses a renderer
2) All other platforms use a cross-platform solution


For Android and all other platforms, IndicatorLayout is used as "layout strategy". By default, it's horizontal stack, but you can easily change it to Vertical stack or any other layout

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: how about **oldCollection**?

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: hawkeye :))

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: @adrianknight89 we cannot do it

@StephaneDelcroix proposed to make it -1 and it makes sense because iOS uses own "default" value, which can be different with 10.

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: @adrianknight89 I can be wrong but seems yes (because of android)

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: @adrianknight89 but this name doesn't explain what is it and who is the owner.
How about **itemViewTapGestureRecognizer**?

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: 1) Well, i will handle it with while(Items.Count > Count) {  //remove }
2) No, it doesn't

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: I will drop this method, as we discussed above

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: hmmmm

Xamarin.Forms.Platform.iOS/Renderers/IndicatorViewRenderer.cs
AndreiMisiukevich: Oh, you are right. By the way, it could lead to a bug because default value is nil if I am not wrong

Xamarin.Forms.Core/IndicatorView.cs
adrianknight89: We have a lot of extension classes in the Xamarin.Forms.Core project. Search for extensions in VS. You'll see them. You could place it in Xamarin.Forms or Xamarin.Forms.Internals namespace. There doesn't seem to be consistency among extension classes in terms of namespaces. @StephaneDelcroix can guide you better. 

Also some have `[EditorBrowsable(EditorBrowsableState.Never)]` while others don't. Hmm.

Xamarin.Forms.Core/IndicatorView.cs
adrianknight89: My opinion is dropping them would be better.

Xamarin.Forms.Core/IndicatorView.cs
adrianknight89: That's what I thought was going on. To be honest, people will be confused about this or potentially other properties having no effect on iOS due to the renderer being present. For Android and others, the path forward could be creating ViewRenderers (with shapes, pagination, etc.) from scratch. This requires more work, but it will solve this mismatch issue. In my opinion, if we are leveraging Core, then we should do it for all platforms, but I know you were asked to use a renderer. :)

Xamarin.Forms.Core/IndicatorView.cs
adrianknight89: Sorry, that's what I meant. Use `oldCollection` and `collection` respectively.

Xamarin.Forms.Core/IndicatorView.cs
adrianknight89: I don't have strong feelings on this. It all comes down to preferences. Normally, what I'd do is, if there is only one gesture recognizer, then I leave the variable name bland and infer its owner based on the subsequent lines. I'll use more specific names if there are multiple gesture recognizers. Again, do what you think makes sense. :)

Xamarin.Forms.Core/IndicatorView.cs
adrianknight89: Again, this issue could be solved in native renderers to leverage native drawing APIs. Forms is slowing us down here a bit.

Xamarin.Forms.Platform.iOS/Renderers/IndicatorViewRenderer.cs
adrianknight89: The default value for the class fields is null, but you're using a null coalescing operator, so you should pick Control properties the first time.

Xamarin.Forms.Core/IndicatorView.cs
rmarinho: We are thinking in using Visual="Forms" for a cross platform solution , while No visual would use a native renderer , what do guys think of that approach?

Xamarin.Forms.Core/IndicatorView.cs
rmarinho: Yeah, avoid linq please

Xamarin.Forms.Core/IndicatorView.cs
rmarinho: I agree we can move this to internals name space and hide it. Can we also add a example of IndivatorsView binding to a VM without using carouselView so our users can understand they can use this without a CarouselVIew.

Thanks

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: done

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: done

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: done

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: this method has gone

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: sounds good

Xamarin.Forms.Core/IndicatorView.cs
AndreiMisiukevich: done

AndreiMisiukevich:
Does anybody want to help with tests?

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -0,0 +1,228 @@
...
+		public static readonly BindableProperty CountProperty = BindableProperty.Create(nameof(Count), typeof(int), typeof(IndicatorView), default(int), propertyChanged: (bindable, oldValue, newValue)
+			=> ((IndicatorView)bindable).ResetIndicatorCount((int)oldValue));
+
+		public static readonly BindableProperty MaximumVisibleCountProperty = BindableProperty.Create(nameof(MaximumVisibleCount), typeof(int), typeof(IndicatorView), int.MaxValue, propertyChanged: (bindable, oldValue, newValue)
+			=> ((IndicatorView)bindable).ResetIndicatorStyles());
+
+		public static readonly BindableProperty IndicatorTemplateProperty = BindableProperty.Create(nameof(IndicatorTemplate), typeof(DataTemplate), typeof(IndicatorView), default(DataTemplate), propertyChanged: (bindable, oldValue, newValue)
+			=> ((IndicatorView)bindable).ResetIndicatorStyles());
+
+		public static readonly BindableProperty HidesForSingleIndicatorProperty = BindableProperty.Create(nameof(HidesForSingleIndicator), typeof(bool), typeof(IndicatorView), true, propertyChanged: (bindable, oldValue, newValue)
```
StephaneDelcroix: `HidesSingle` is probably a better name

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -0,0 +1,228 @@
...
+	[RenderWith(typeof(_IndicatorViewRenderer))]
+	public class IndicatorView : TemplatedView
+	{
+		public static readonly BindableProperty PositionProperty = BindableProperty.Create(nameof(Position), typeof(int), typeof(IndicatorView), default(int), BindingMode.TwoWay, propertyChanged: (bindable, oldValue, newValue)
+			=> ((IndicatorView)bindable).ResetIndicatorStyles());
+
+		public static readonly BindableProperty CountProperty = BindableProperty.Create(nameof(Count), typeof(int), typeof(IndicatorView), default(int), propertyChanged: (bindable, oldValue, newValue)
+			=> ((IndicatorView)bindable).ResetIndicatorCount((int)oldValue));
+
+		public static readonly BindableProperty MaximumVisibleCountProperty = BindableProperty.Create(nameof(MaximumVisibleCount), typeof(int), typeof(IndicatorView), int.MaxValue, propertyChanged: (bindable, oldValue, newValue)
```
StephaneDelcroix: MaximumVisibleXXX (XXX being DOTS, or INDICATORS, but not Count), or, to make it shorter `MaximumVisible`

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -0,0 +1,228 @@
...
+			=> ((IndicatorView)bindable).ResetIndicatorStyles());
+
+		public static readonly BindableProperty IndicatorTemplateProperty = BindableProperty.Create(nameof(IndicatorTemplate), typeof(DataTemplate), typeof(IndicatorView), default(DataTemplate), propertyChanged: (bindable, oldValue, newValue)
```
StephaneDelcroix: accepting templates is allowing the user to pass a full view as the mini, and that will blow up, perf-wise.

instead of template, color, selected color (btw, it's quite hard to set the color on a templated item), I would take Image, color, selectedcolor, and tint the image according to the state. (By tinting, I mean desaturating the image, and tinting it using the color)

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -0,0 +1,228 @@
...
+		public static readonly BindableProperty SelectedIndicatorColorProperty = BindableProperty.Create(nameof(SelectedIndicatorColor), typeof(Color), typeof(IndicatorView), Color.Default, propertyChanged: (bindable, oldValue, newValue)
+			=> ((IndicatorView)bindable).ResetIndicatorStyles());
+
+		public static readonly BindableProperty IndicatorSizeProperty = BindableProperty.Create(nameof(IndicatorSize), typeof(double), typeof(IndicatorView), -1.0, propertyChanged: (bindable, oldValue, newValue)
+			=> ((IndicatorView)bindable).ResetIndicatorStyles());
+
+		public static readonly BindableProperty ItemsSourceProperty = BindableProperty.Create(nameof(ItemsSource), typeof(IEnumerable), typeof(IndicatorView), null, propertyChanged: (bindable, oldValue, newValue)
+			=> ((IndicatorView)bindable).ResetItemsSource((IEnumerable)oldValue));
+
+		public static readonly BindableProperty IndicatorLayoutProperty = BindableProperty.Create(nameof(IndicatorLayout), typeof(Layout<View>), typeof(IndicatorView), null, propertyChanged: TemplateUtilities.OnContentChanged);
```
StephaneDelcroix: there's no need to expose this. use a StackLayout (eventually expose the orientation), and if someone wants to replace it, well, this control is templatable

Xamarin.Forms.Core/Internals/IndicatorViewExtensions.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.ComponentModel;
+using static Xamarin.Forms.IndicatorView;
+
+namespace Xamarin.Forms.Internals
+{
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	public static class IndicatorViewExtensions
+	{
+		public static void SetItemsSourceBy(this IndicatorView indicatorView, CarouselView carouselView)
```
StephaneDelcroix: is this meant to be used ?

AndreiMisiukevich: in another case, we'll have to bind each property from indicators to carousel

jsuarezruiz:
I have tested **IndicatorView**. For that, I have added a new sample to Core Gallery.
Captura de pantalla 2019-11-15 a las 11 22 06

I have found the following details:
- On iOS, the initial Shape is Circle although Square is assigned.
- Changing `MaximumVisible` seems not to be working.
- Changing `IndicatorTemplate` seems not to be working.
- Changing `IndicatorsShape` seems not to be working. It works if the Shape is changed and also the Color (the Color is the one that updates the Shape).

Some gifs:
indicatorview-droid-issues-small

samhouts:
also please retarget to 4.4.0! thanks :)

AndreiMisiukevich:
Thanks, folks

PR #7468, opened 2019-09-10
[Android] Refactor the code for setting reverse layout in ItemsViewRenderer

adrianknight89:
Small refactoring attempt to simplify reverse layout setting so that the code is more readable.


adrianknight89:
@hartez Can you compare the two commits in this PR and let us know if the first one makes sense? I know there is some bitwise stuff in the extension methods.

hartez:
UI Test failures are unrelated to these changes.

PR #6596, opened 2019-06-19
[Android] Prevent race condition between activity create and dispose

kvpt:
### Description of Change ###
This change prevent race condition on MainPage change between the old activity OnDestroy and new activity OnCreated.

This case only appear when using a static application, often true when using an IoC container for resolve App object.

I fixed some cases that cause crash in #4707 :
- Missing App PropertyChanged handler unsubscribe
- Wrong dispose place
- Prevent the dispose if it the same page
- Prevent PropertyChanged to be called if the activity is in pause or to be destroyed

It drastically reduced the number of crash but didn't fix it completely.

So I have an edge case of an edge case to fix.

This [post](https://medium.com/androiddevelopers/the-android-lifecycle-cheat-sheet-part-ii-multiple-activities-a411fd139f24) here put me on the right path, it seemed possible that the destroy and create execute in parallel.

After some strategically placed breakpoints in the Framework I confirmed that it was this issue.
This case of race condition was very difficult to reproduce, an user have the bug in #5894, and when executing his reproduction it works for me but not for him because computer speed difference.

So after reproduce it, I identified the cause of the problem :
- When the page is reloaded, the app restart and call this method
https://github.com/xamarin/Xamarin.Forms/blob/a874f52adcbdba2491676b466bbfa84f2ff2cbb7/Xamarin.Forms.Platform.Android/AppCompat/Platform.cs#L353
The renderer exist so the method return
- After that two cases
  - OnLayout is called but the renderer is now null, this cause the application to crash
https://github.com/xamarin/Xamarin.Forms/blob/a874f52adcbdba2491676b466bbfa84f2ff2cbb7/Xamarin.Forms.Platform.Android/AppCompat/Platform.cs#L233
  - The app display a blank page because renderer was disposed (worse for me because unnoticed in AppCenter, so no stacktrace and impossible to identify the root cause)

After some search, that was because the `Cleanup` method in [Platform](https://github.com/xamarin/Xamarin.Forms/blob/362c683d0b8838269bc4df32ec1d10b9870fd2f0/Xamarin.Forms.Platform.Android/AppCompat/Platform.cs#L270) was asynchronous and delayed. So I make it synchronous when dispose (part of PR #6467).
After that I didn't reproduce it with the emulator, but after published the fix on my app, I have lower crash number in AppCenter but still crashs. And issue #5894 was not fixed.

So I decided to manually sync the two methods with a synchronization primitive.
I choose the simpliest and lightest way to do that, with a `ManualResetEventSlim`.
[...]

PR #7477, opened 2019-09-11
Call the application OnResume, on Android Resume instead of Restart.

### LINKED ISSUE #7400: [Bug] [Android] MainPage not changing in OnResume
### Description
On Android changing MainPage in OnResume have no effect. Looks like it was introduced in https://github.com/xamarin/Xamarin.Forms/pull/4707.

OnResume sent to application here
https://github.com/xamarin/Xamarin.Forms/blob/1d47eb0f70144211a495ce2b6e628ed5735b4e0d/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs#L442
when the  _currentState == AndroidApplicationLifecycleState.OnRestart

and MainPage changed here
https://github.com/xamarin/Xamarin.Forms/blob/1d47eb0f70144211a495ce2b6e628ed5735b4e0d/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs#L389

this check https://github.com/xamarin/Xamarin.Forms/blob/1d47eb0f70144211a495ce2b6e628ed5735b4e0d/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs#L392 prevents MainPage change handling
### Steps to Reproduce

1. Start repro
2. Move app to background
3. resume app

### Expected Behavior
Main page changed
### Actual Behavior
Main page doesn't change
### Basic Information

- Version with issue: 4.0 and later
- Last known good version: 3.6.0.709228
- IDE: vs 2019

[Xf.Sandbox.zip](https://github.com/xamarin/Xamarin.Forms/files/3580089/Xf.Sandbox.zip)

kvpt:
### Description of Change ###
Fix regression introduced by #4707.

The OnResume worked before due to a side effect.
#4707 fixed the previous activity destroy which incorrectly triggered the page change.
Now that the event state are consistant, this revealed that OnResume event on the application was not called during Android Resume but during Android Restart event.
This PR modify when the SendResume is triggered to be consistant with the Android lifecycle.

### Issues Resolved ### 
- Fixes #7400

### API Changes ###
 None

### Platforms Affected ### 
- Android

The OnResume method is called on Android Resume

None

Not applicable

### Testing Procedure ###
[issue7400.zip](https://github.com/xamarin/Xamarin.Forms/files/3600649/issue7400.zip)
The reproduction project has the framework code included. So you can test with and without the modification.

1. Start reproduction project
2. Set a breakpoint on OnResume method in the App class
3. Move app to background
4. Resume app 
5. Breakpoint must be hit

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
@@ -446,7 +446,7 @@ namespace Xamarin.Forms.Platform.Android
 
 			if (_previousState == AndroidApplicationLifecycleState.OnCreate && _currentState == AndroidApplicationLifecycleState.OnStart)
 				_application.SendStart();
-			else if (_previousState == AndroidApplicationLifecycleState.OnStop && _currentState == AndroidApplicationLifecycleState.OnRestart)
+			else if (_previousState == AndroidApplicationLifecycleState.OnRestart && _currentState == AndroidApplicationLifecycleState.OnStart)	
 				_application.SendResume();
 			else if (_previousState == AndroidApplicationLifecycleState.OnPause && _currentState == AndroidApplicationLifecycleState.OnStop)
 				_application.SendSleep();
--- a/Xamarin.Forms.Platform.Android/FormsApplicationActivity.cs
+++ b/Xamarin.Forms.Platform.Android/FormsApplicationActivity.cs
@@ -263,7 +263,7 @@ namespace Xamarin.Forms.Platform.Android
 
 			if (_previousState == AndroidApplicationLifecycleState.OnCreate && _currentState == AndroidApplicationLifecycleState.OnStart)
 				_application.SendStart();
-			else if (_previousState == AndroidApplicationLifecycleState.OnStop && _currentState == AndroidApplicationLifecycleState.OnRestart)
+			else if (_previousState == AndroidApplicationLifecycleState.OnRestart && _currentState == AndroidApplicationLifecycleState.OnStart)	
 				_application.SendResume();
 			else if (_previousState == AndroidApplicationLifecycleState.OnPause && _currentState == AndroidApplicationLifecycleState.OnStop)
 				_application.SendSleep();


jsuarezruiz:
I've tried the sample and I can't reproduce your problem @hartez.


The only changes after your changes in the sample is a rebase from 4.3.0.
Should I get the same problem with spacing?.

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -265,9 +265,16 @@ protected override void OnResume()
...
+			var isRestart = _previousState == AndroidApplicationLifecycleState.OnRestart;
+
 			_previousState = _currentState;
 			_currentState = AndroidApplicationLifecycleState.OnResume;
 
+			if (isRestart)
+			{
+				_application?.SendResume();
+			}
+
```
kingces95: Can we move this logic into `OnStateChanged`? I'm not entirely sure why we didn't sprinkle the logic in `OnStateChanged` into the overrides like you have but since we've already established the precedent let's keep it (or recast call the logic as you have here). For one thing, it'll ensure the lollipop code below runs before the callback is placed.

kvpt: @jsuarezruiz I will have some time this weekend and see what can be done.

kvpt: It's done.

kvpt:
I tried to run the reproduction on 4.0.0 branch and don't reproduce the issue (as #4707 has been merged in 4.1.0 it's logic).
So the 4.1.0 branch is the right branch to target, no need to retarget on 4.0.0.

kingces95:
Thanks for your hard work re-arranging the code per my request. Much appreciated!

I see we will call `OnResume` when the android lifecycle is resuming: `OnRestart` -> `Start`. To do that, is it necessary to not call `OnResume` during restart: `Stop` -> `Restart`? Might someone have taken a dependency on that behavior however odd? If so, should we add the new behavior and not remove the old behavior?

kvpt:
The old behavior trigger the OnResume when activity state is not started yet.
And this cause nasty issues with old and new activity race condition and dispose problem which I fixed with #4707 and #6596.

In #4707 I added this code to prevent paused or stopped activity to respond to the change.
https://github.com/xamarin/Xamarin.Forms/blob/c171efa367fd8e84f040df9882dbec79850cfd85/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs#L396
This is this change combined with the fact that the lifecycle was wrong for the restart which break that.
So if you want that the old lifecycle work we will have too revert this change, but it will reintroduce some dispose issues on main page change.

Also if we want the old lifecycle to work and the new at the same time, it will likely cause the Resume event to be called twice.
Perhaps we can use a flags or an OnPlatform (like SendAppearingEventOnResume), to allow the developer to choose the behavior of the lifecycle ?

samhouts:
@rmarinho Can you please tell me why this build is failing? Thanks!

rmarinho:
@samhouts needs rebase, yaml doesn't have dotnet 3.0 install seems a old branch

samhouts:
@kvpt Would you mind rebasing to fix the build? Thanks!

kvpt:
I was used 4.1.0 branch has the regression is introduced in this version.
But as the target branch has been changed to master, I rebased on master this time.
It's only two lines, so it will be simple to backport, regardless of the branch.

samhouts:
@kvpt I wonder if these failures on API 19 Legacy Renderers are legitimate. Would you mind investigating? Thanks!
image

kvpt:
@samhouts This code is only executed when the application lifecycle change state, start/restart/resume.
Nothing in these tests can trigger that, so I think it's not related.
I checked in API 19 emulator and doesn't reproduce the issues.

PR #7499, opened 2019-09-12
[Android/iOS] Update CollectionView when user changes out the header/footer

### LINKED ISSUE #7347: [Bug] Header/Footer do not update on Android CollectionView
### Description

Updating the value of the `Header` and `Footer` properties on Android does not update the display. 

The problem is in the ItemsViewAdapter class - the adapter detects the value change, but does not actually update the content. It only updates whether or not a header/footer is displayed.

### LINKED ISSUE #7453: [Bug] CollectionView Header not displayed without Footer
without a footer the header doesn't get displayed

- Version with issue: 4.2.0

PureWeen:
### Description of Change ###

- Fixed iOS and Android so that when you swap out or just null out the header/footer they properly show/hide
- Fixed ContentOffset adjustment on iOS to be correct when said swapping occurs
- Fixed iOS to remeasure header/footer once LayoutSubviews is called on the Adapter. The initial width of the CV when the Header/footer is realized is larger than when it actually gets to the layoutsubviews call
- on iOS dispose of the renderers when header/footer is swapped out
- On Android the Scroll Values weren't being converted back from Device pixels before being returned to Forms
- On iOS fixed the Scroll Values so that if the ContentInset is set on the CollectionView from the header that it doesn't affect the scroll values

### Issues Resolved ### 
- fixes #7453
- fixes #7347 

### Platforms Affected ### 
- iOS
- Android

### Testing Procedure ###
- PR includes modifications to the CV header/footer galleries so that you can test hiding and showing

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls/GalleryPages/CollectionViewGalleries/HeaderFooterGalleries/HeaderFooterGrid.xaml
+++ b/Xamarin.Forms.Controls/GalleryPages/CollectionViewGalleries/HeaderFooterGalleries/HeaderFooterGrid.xaml
@@ -6,33 +6,36 @@
              mc:Ignorable="d"
              x:Class="Xamarin.Forms.Controls.GalleryPages.CollectionViewGalleries.HeaderFooterGalleries.HeaderFooterGrid">
     <ContentPage.Content>
-        <CollectionView x:Name="CollectionView" >
-            <CollectionView.ItemsLayout>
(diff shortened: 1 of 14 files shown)


PureWeen:
fixed conflicts and kicked off UI Tests!!

PureWeen:
Pretty sure UI Test failures aren't related to this PR

> ScrollListViewInsideScrollView_apple_iphone_x

Isn't and there are some CarouselView failures that I'm pretty sure are the same ones we've been seeing

Issue #4565, opened 2018-11-28
[Enhancement] MultiBinding


legistek:
## Summary
Add a `MultiBinding` binding type equivalent to WPF.

## API Changes

```
// new
public class MultiBinding : BindingBase
{
     public Collection<BindingBase> Bindings {get; set;}
     public IMultiValueConverter Converter {get; set;}
}

// new
public interface IMultiValueConverter 
{
     object Convert(object[] values, Type targetType, object parameter);
     object[] ConvertBack(object value, Type[] targetTypes, object parameter);
}
```

## Intended Use Case

Example XAML usage:

```
<Button>
     <Button.IsEnabled>
          <MultiBinding Converter="{StaticResource AllMustBeTrueMultiConverter}">
               <Binding Path="ActionIsAllowed"/>
               <Binding Path="SomeOtherRequirement"/>
          </MultiBinding>
     </Button.IsEnabled>
</Button>
```

Comment - It's challenging to build highly responsive UIs in XAML without `MultiBinding`. Currently there are two workarounds, which even together do not address all cases:

(1) Define a single property in a view model that resolves to a value based on other properties, and bind in XAML to the single property.

First, this makes view models more complex and more difficult to maintain. Suppose a property `ActionIsEnabled ` is defined as:
```
public bool ActionIsEnabled => this.ActionIsAllowed && this.SomeOtherRequirement;
```
Each of the setters for the independent properties (`ActionIsAllowed` and `SomeOtherRequirement`) need to include property change notifications for `ActionIsEnabled` in addition to themselves. If the logic for `ActionIsEnabled` is changed, however, then those other property setters (and/or those of other properties) have to potentially be updated as well. 

Moreover, this assumes the dependencies are all entirely within the same view model. If the independent properties are in different classes than the dependent property, however, there may be no practical way to send a property change notification for the dependent property when the independent properties in the other classes are changed. This workaround therefore doesn't work in all cases.

(2) Wrap a control in a new layer (`Grid`, etc) for each condition.

I've seen this done a lot in Angular using nested `<div>`s for each condition. Leaving aside the inefficiency, this does work for simple cases, the most common one being a cascading property like `IsEnabled` or `IsVisible`, where all the conditions must evaluate to `true`. Anything more complex, however, and this system doesn't work either - for example, if ANY rather than ALL conditions must be true.

`MultiBinding` IMHO provides the most robust and elegant means of specifying complex dependencies between control properties and data.

[...]

PR #7506, opened 2019-09-13
[C] Fix TemplateBinding for non-logicalchildren

### LINKED ISSUE #7494: {TemplateBinding}s in Span no longer works
### Description
Label FormattedString Span Text bindings no longer work

### Sample Code
	<Style x:Key="ForecastItem.v1" TargetType="forecastItems:ForecastItem">
		<Setter Property="ControlTemplate">
			<Setter.Value>
				<ControlTemplate>
					<Grid>
						<Label
							Margin="0,2,0,-2"
							FontFamily="{TemplateBinding FontFamily}"
							FontSize="16"
							HorizontalOptions="Center"
							Text="{TemplateBinding DayText}"
							VerticalOptions="Start" />
						<Label
							Margin="0,2,0,-2"
							FontFamily="{TemplateBinding FontFamily}"
							FontSize="20"
							HorizontalOptions="Center"
							VerticalOptions="End">
							<Label.FormattedText>
								<FormattedString>
									<Span Text="{TemplateBinding HiTempText}" />
									<Span Text="{TemplateBinding DividerText}" />
									<Span Text="{TemplateBinding LoTempText}" TextColor="#AAFFFFFF" />
								</FormattedString>
							</Label.FormattedText>
						</Label>
						<weatherImages:WeatherImageSmall
							Delay="{TemplateBinding Delay}"
							HeightRequest="80"
							HorizontalOptions="Center"
							ImageIcon="{TemplateBinding ImageIcon}"
							VerticalOptions="Center"
							WidthRequest="80" />
					</Grid>
				</ControlTemplate>
			</Setter.Value>
		</Setter>
	</Style>

### Expected Behavior
text shows on page
### Actual Behavior
no text shown for label with formatted text, other label shows text
### Basic Information

- Version with issue: 4.3.0.778476
- Last known good version: 4.2.0.709249
- IDE: Android
- Platform Target Frameworks: 
  - Android: 9.0

StephaneDelcroix:
### Description of Change ###

Spans, e.g., are no LogicalChildren of FormattedString, which is not a
LogicalChildren of Label. The newly introduced way of propagating the
TemplatedParent was top-down, using LogicalChildren, was not assigning
the correct TemplatedParent on Spans.

This reverts to a bottom-up lookup, hooking on ParentSet, like it used to
be.

As a side effect, the TemplateBindings are no longer updated on reparenting,
but this was the original behavior, and I can't think of a case where this
would be needed.

The regression to `{TemplateBinding}` was introduced by #4375.

### Issues Resolved ### 
- fixes #7494

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Element.cs
+++ b/Xamarin.Forms.Core/Element.cs
@@ -12,15 +12,8 @@ namespace Xamarin.Forms
 	{
 		public static readonly BindableProperty MenuProperty = BindableProperty.CreateAttached(nameof(Menu), typeof(Menu), typeof(Element), null);
 
-		public static Menu GetMenu(BindableObject bindable)
-		{
-			return (Menu)bindable.GetValue(MenuProperty);
-		}
-
-		public static void SetMenu(BindableObject bindable, Menu menu)
-		{
-			bindable.SetValue(MenuProperty, menu);
-		}
+		public static Menu GetMenu(BindableObject bindable) => (Menu)bindable.GetValue(MenuProperty);
+		public static void SetMenu(BindableObject bindable, Menu menu) => bindable.SetValue(MenuProperty, menu);
 
 		internal static readonly ReadOnlyCollection<Element> EmptyChildren = new ReadOnlyCollection<Element>(new Element[0]);
 
@@ -59,8 +52,8 @@ namespace Xamarin.Forms
 
 		public string ClassId
 		{
-			get { return (string)GetValue(ClassIdProperty); }
-			set { SetValue(ClassIdProperty, value); }
+			get => (string)GetValue(ClassIdProperty);
+			set => SetValue(ClassIdProperty, value);
 		}
 
 		public IList<Effect> Effects
@@ -164,10 +157,7 @@ namespace Xamarin.Forms
 		[EditorBrowsable(EditorBrowsableState.Never)]
 		public Element RealParent { get; private set; }
 
-		Dictionary<BindableProperty, string> DynamicResources
-		{
-			get { return _dynamicResources ?? (_dynamicResources = new Dictionary<BindableProperty, string>()); }
-		}
+		Dictionary<BindableProperty, string> DynamicResources => _dynamicResources ?? (_dynamicResources = new Dictionary<BindableProperty, string>());
 
 		void IElement.AddResourcesChangedListener(Action<object, ResourcesChangedEventArgs> onchanged)
 		{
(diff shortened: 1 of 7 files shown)


Xamarin.Forms.Core/Element.cs
```diff
@@ -213,38 +203,6 @@ public Element Parent
...
-		void RefreshTemplatedParent()
-		{
-			var templatedParent = this.IsTemplateRoot
-				? this.Parent
-				: this.Parent?.TemplatedParent;
-			if (ReferenceEquals(templatedParent, this.TemplatedParent))
-				return;
-
-			this.TemplatedParent = templatedParent;			
-			foreach (var element in this.LogicalChildren)
```
StephaneDelcroix: so this was the top-down way

legistek:
Hey @StephaneDelcroix thanks for cc'ing me. I see the issue. Apologies for not catching it before. The problem is not the way the `TemplatedParent` is being derived (that's still bottom up) but rather the way parent change notifications are propagating (that is indeed top-down). Because we're no longer `await`ing the formation of the full parent chain before initially applying the `TemplateBinding` we are relying on the parent change notification propagating down the chain when the template root is finally assigned to the templated parent. 

However I think this could be fixed without having to go back to async void and `TemplateUtilities.FindTemplatedParentAsync` (which if you recall you were eager for me to avoid).  Rather we could use the same recursive mechanism that `AncestorType` relative source binding is using. It would then also handle reparenting correctly (though you're right, it's hard to imagine a scenario where that is relevant for `TemplateBinding`). It would also probably be more efficient than my original version because the `Binding` would only be applied once, after the chain is complete. 

Want me to take a stab at doing so? I'm not sure how quickly I could get to it so if it's high priority your fix looks fine too.

StephaneDelcroix:
@legistek as this is a regression, we'll probably move forward with this fix for 4.3, but I'd be happy if you could provide a better fix for further versions

StephaneDelcroix:
@legistek 

> I see the issue. Apologies for not catching it before

no problem at all. we can't test 100% of the cases in initial unit tests. and the regression was caught in a -pre version, so the process works

legistek:
Thanks @StephaneDelcroix ! :) I'll see about improving it with Multibinding (#4565), which will need to work properly with `TemplateBinding` and `RelativeSource` too.

PR #7511, opened 2019-09-13
Reverse CollectionView layout

### LINKED ISSUE #7277: Set scroll direction for RemainingItemsThresholdReached
## Summary
If we want to make a chat application, how can we create an infinite scroll in the opposite direction (i.e. from bottom to top)? Chat apps today load the last X (say 10) messages and scroll to the last message before the initial items source is rendered on the screen. The user, then, has to scroll up to trigger a DB call to get more messages which are added to the top of the list. The current implementation of `RemainingItemsThresholdReached` assumes the list is scrolling from top to bottom.

## API Changes

```
public enum RemainingItemsThresholdDirection
{
   Forward, // default
   Backward
}
```
`Forward` corresponds to the order the user scrolls a vertical or horizontal list naturally. By default, a horizontal list scrolls from left to right whereas a vertical list scrolls from top to bottom. `Backward` is the opposite. If the direction is set to `Backward` and `RemainingItemsThreshold` is set to 5, then `RemainingItemsThresholdReached` will be invoked when the user is looking at any of the first 5 items from the top of the list.

## Intended Use Case
Mostly chat apps

adrianknight89:
### Description of Change ###

This PR allows us to reverse CollectionView layout. On Android, I exposed a property that could be overridden in a custom renderer. On iOS, the simplest way is to turn the CollectionView and its item template upside down. Android performs badly with `ScaleY` changes and scrolling.

### Issues Resolved ### 

- fixes #7277

### Testing Procedure ###
@jfversluis This has the same UI test as #5754, but the test needs to be modified slightly to scroll up. Can you please take a look at it so it passes?

### PR Checklist ###


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Xamarin.Forms.Controls.Issues.Shared.projitems
```diff
@@ -1359,7 +1362,10 @@
     <Compile Update="$(MSBuildThisFileDirectory)Issue6254.xaml.cs">
       <DependentUpon>Issue6254.xaml</DependentUpon>
     </Compile>
-    <Compile Update="C:\Users\hartez\Documents\Xamarin\Xamarin.Forms\Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Issue7519Xaml.xaml.cs">
```
adrianknight89: @hartez Just saw this after rebase

hartez:
Adding a DNM for now - I want to get this in as a real full-fledged feature; we've had a couple of requests for it. This probably needs to be a boolean on LinearItemsLayout (or something similar). The Android implementation is probably very close to what you've already got here. The UWP implementation should be similar. I need to look into the iOS implementation more.

adrianknight89:
@hartez On iOS, I think it will be similar to what I did with the `ScaleY` changes on the CollectionView and its ItemTemplate. I need this feature soon though. If we could expose `protected virtual bool IsLayoutReversed { get; set; }` on Android for now, I could wait for a full-fledged feature.

mrkrk:
Any update?

PR #7579, opened 2019-09-18
Update Android UI test instructions

adrianknight89:
Added instructions for running Android UI tests on Windows.

@PureWeen @jfversluis @samhouts When I debug the test project on Windows, [ApkPath](https://github.com/xamarin/Xamarin.Forms/blob/65c6e5f59fa1776566f675d6e2e57e7068adb48e/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/TestPages/TestPages.cs#L19) isn't correct for me. We go 3 folders up from the current directory and append a substring to point to the Apk file, but I noticed my current directory does not point to the solution directory. It points to the Temp folder instead.

```
static IApp InitializeAndroidApp()
{
	var directory = Directory.GetCurrentDirectory(); // added this line for testing

	var app = ConfigureApp.Android.ApkFile(AppPaths.ApkPath).Debug().StartApp();

	if (bool.Parse((string)app.Invoke("IsPreAppCompat")))
	{
		IsFormsApplicationActivity = true;
	}

	return app;
}
```

`directory` returns `C:\Users\[user]\AppData\Local\Temp` for me. So, when we go 3 levels up and then append the rest of `ApkPath`, I end up with

`C:/Users/[user]/Xamarin.Forms.ControlGallery.Android/bin/Debug/AndroidControlGallery.AndroidControlGallery-Signed.apk` which is not correct because it's missing the parent folder I checked out XF solution to. Any of you saw this issue before?

Temporarily, I hard-coded the path to the Apk in local, but I'd like to know the root cause of this issue if possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ Make sure you have nuget.exe 4.0 or above and the latest dotnet core sdk (2.0.3)
 
 ### UI Tests ###
 
-#### Run Android Tests ####
+##### Run Android Tests #####
 
 Depending on your environment setup, you might need to configure a few things before being able to debug / run UI tests, especially on Windows.
 
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ Make sure you have nuget.exe 4.0 or above and the latest dotnet core sdk (2.0.3)
 
 ### UI Tests ###
 
-##### Run Android Tests #####
+##### Run Android UI Tests #####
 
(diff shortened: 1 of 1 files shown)


rmarinho:
Hum, yeah I have the same issue, but it does work fine on the Mac no ?

adrianknight89:
@rmarinho I develop on Windows, so I saw it there.

PR #7595, opened 2019-09-19
- generalize global listener

PureWeen:
### Issues Resolved ### 

- fixes #

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###


KSemenenko:
This is exactly what I need :) Would it be a good idea to add listeners in some base class for renderer?
the main problem is ListView and ScrollView, and I think it will be in CollectionView too, they have separate renderers, and listeners are usually not added. Therefore, gesture recognizers do not work in ListView.

PR #7600, opened 2019-09-20
Fix raising Navigating event for internal page content with WkWebRenderer

danardelean:
### Description of Change ###
We need to verify if the TargetFrame is the MainFrame and not raise the navigating event. If for example you have a page with an internal iframe that points to another page (for example a Video on Vimeo) you will actually raise the Navigating event even if the page doesn't change, The Navigation should be raised only if the frame is the MainFrame

### Issues Resolved ### 
When using WKWebViewRenderer on iOS we will get the Navigating event when the main page loads internal content from another link

### Platforms Affected ### 
- iOS

None

### Testing Procedure ###
Use a page with iframe that is pointing to external content. You will get two navigating events even if the page does not change

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/WkWebViewRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/WkWebViewRenderer.cs
@@ -243,6 +243,11 @@ namespace Xamarin.Forms.Platform.iOS
 			// https://stackoverflow.com/questions/37509990/migrating-from-uiwebview-to-wkwebview
 			public override void DecidePolicy(WKWebView webView, WKNavigationAction navigationAction, Action<WKNavigationActionPolicy> decisionHandler)
 			{
+				if (!navigationAction.TargetFrame.MainFrame)
+                {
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.iOS/Renderers/WkWebViewRenderer.cs
```diff
@@ -243,6 +243,11 @@ public override void DidStartProvisionalNavigation(WKWebView webView, WKNavigati
 			// https://stackoverflow.com/questions/37509990/migrating-from-uiwebview-to-wkwebview
 			public override void DecidePolicy(WKWebView webView, WKNavigationAction navigationAction, Action<WKNavigationActionPolicy> decisionHandler)
 			{
+				if (!navigationAction.TargetFrame.MainFrame)
```
rmarinho:
```suggestion
				if (!navigationAction.TargetFrame.MainFrame)
```

rmarinho:
Since we didn't got any feedback we will be closing for now. Please provide a sample of the use case at least so we can try it and take over from here.

Thanks

PR #7613, opened 2019-09-21
[Android/iOS] Handle dynamically switching between item layouts in CollectionView

### LINKED ISSUE #5354: [CollectionView] Updating the ItemsLayout type should refresh the layout
More a work task than a bug due to the ongoing development of `CollectionView` (but we'll just categorize it as a bug for now), it should update its layout when switching the `ItemsLayout` types from a `ListItemsLayout` to `GridItemsLayout`. A good example of where this would be used is in a UI where the user is allowed to toggle between those layouts, such as in a retail app. 

This is supposedly semi-functional on iOS but wasn't personally tried. Android doesn't update the layout.

### Steps to Reproduce

1. Use a CollectionView with one of the two layout types
2. Switch to the other layout type in app (such as via a button tap)

### Expected Behavior
The CollectionView updates its layout to the new type.

### Actual Behavior
The CollectionView does not update its layout until leaving and reentering the page (such as with tabs).

adrianknight89:
### Description of Change ###

If you switch between linear and grid layouts, the view should be updated now. Please note that this does not implement UWP.

### Issues Resolved ### 

- fixes #5354

### Testing Procedure ###
Run the attached test and switch between item layouts. Verify that the spacing was not lost.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/CollectionView/ItemsViewRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/CollectionView/ItemsViewRenderer.cs
@@ -591,6 +591,15 @@ namespace Xamarin.Forms.Platform.Android
 			}
 		}
 
+		protected virtual void UpdateLayoutManager()
+		{
+			ItemsLayout = GetItemsLayout();
+			SetLayoutManager(SelectLayoutManager(ItemsLayout));
+
+			UpdateFlowDirection();
+			UpdateItemSpacing();
+		}
+
 		internal void UpdateEmptyViewVisibility()
 		{
 			if (ItemsViewAdapter == null)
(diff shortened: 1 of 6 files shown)


Xamarin.Forms.Platform.Android/CollectionView/ItemsViewRenderer.cs
```diff
@@ -591,6 +591,15 @@ protected virtual void ScrollTo(ScrollToRequestEventArgs args)
 			}
 		}
 
+		protected virtual void UpdateLayoutManager()
+		{
+			ItemsLayout = GetItemsLayout();
+			SetLayoutManager(SelectLayoutManager(ItemsLayout));
+
+			UpdateFlowDirection();
+			UpdateItemSpacing();
```
adrianknight89: Not sure if we need to update anything else here. Any thoughts?

rmarinho:
Hey @adrianknight89 can you add a UITest? 

Thanks

adrianknight89:
I'll see what I can do later tonight. Thanks.

adrianknight89:
@rmarinho Added a UI test.

rmarinho:
@adrianknight89 I think the merged had some error, the build is failing, can you check again? thanks

adrianknight89:
@rmarinho I rebased. I don't see an issue when I locally build, but I ran `Xamarin.Forms.Controls.Issues.Shared.projitems` through an XML validator and got the same issue as that of build. I'll look at this later today.

In the meantime, feel free to run build again. I don't see a problem with the changes in this PR at first glance.

adrianknight89:
@rmarinho Is it building now? I don't believe those Windows errors are related.

Issue #2447, opened 2018-04-12
[Android] Force label text direction


encrypt0r:
### Description
Currently on Android `FlowDirection` has no effect on how text is aligned in `Label`. Which can be quite problematic especially when the `Label` is used inside `ListView` cells.

I am not sure if this is a bug, but I think that both iOS and Android projects should be consistent (according to @ali-h2010 `Label` on iOS respects `FlowDirection`). Aligning text based on its content is very useful in some cases, but we need to have more control over it.

### Steps to Reproduce

1. Create new project
2. Add Two labels to a page
3. Write English text in the first label and Kurdish/Arabic (`نووسینی کوردی`) text in the other
3. Set Page `FlowDirection` to `RightToLeft`

### Expected Behavior
The text of both labels should be aligned right.

### Actual Behavior
The English text is still aligned left.
### Basic Information

- Version with issue:
- Last known good version: N/A
- IDE: Visual Studio 2017.6

### Screenshots

![img](https://user-images.githubusercontent.com/16880059/38658671-c75bcb10-3e2e-11e8-9d0b-b79ce1c6b465.PNG)

![919470c57e564540a10e4d163d4e4a34](https://user-images.githubusercontent.com/16880059/38698811-fea6aa16-3e9e-11e8-808c-3fd58e0aa26a.jpg)
Screenshot provided by: @ali-h2010

### Reproduction Link
```xaml
<?xml version="1.0" encoding="utf-8" ?>
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             xmlns:local="clr-namespace:App24"
             x:Class="App24.MainPage"
             Padding="10"
             FlowDirection="RightToLeft">

    <StackLayout>
        <Label Text="English Text" FontSize="Large" />
        <Label Text="نووسینی کوردی" FontSize="Large" />
    </StackLayout>
    
</ContentPage>
```
This is the full project: [TextDirectionRepro.zip](https://github.com/xamarin/Xamarin.Forms/files/1901476/TextDirectionRepro.zip)

samhouts:
This is listed as a known limitation on #1222: `TextDirection` property needs to be added. Thanks!

huangjinshe:
In the new releaset version of nuget, the problem still there.  
My problem is more different, Looks like Android just auto handle the flwo direction by font content itselt.  

[...]

PR #7614, opened 2019-09-21
[Android] Fix FlowDirection for Labels

### LINKED ISSUE #7512: RTL CollectionView looks odd
### Description
Looks weird:
image (1)

### Steps to Reproduce

https://docs.microsoft.com/en-us/samples/xamarin/xamarin-forms-samples/userinterface-collectionviewdemos/

Nav to page as shown in description. Observe left alignment as in description.

### Expected Behavior

Expected it to be right aligned. Expected it to mirror of LTR.

### Actual Behavior

Is left aligned. 

### Basic Information

Android 9.0 on Master.

adrianknight89:
### Description of Change ###

I was looking at #7512 and noticed the issue is not a `CollectionView` problem. If you go to `FlowDirection` gallery, you'll see that RTL flow is broken. At the level of `VisualElementRenderer`, flow direction is being applied only to the layout direction. This does not seem to be working for `TextView`. Instead, we could use a different method for that.

### Issues Resolved ### 

- fixes #7512

### Testing Procedure ###
Run the attached test. Also, go to `FlowDirection` gallery and make sure the labels are aligned properly when you change the flow direction.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7512.xaml.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7512.xaml.cs
@@ -75,7 +75,6 @@ namespace Xamarin.Forms.Controls.Issues
 			{
 				Name = "Baboon",
 				Location = "Africa & Asia",
-				Details = "Baboons are African and Arabian Old World monkeys belonging to the genus Papio, part of the subfamily Cercopithecinae.",
 				ImageUrl = "http://upload.wikimedia.org/wikipedia/commons/thumb/f/fc/Papio_anubis_%28Serengeti%2C_2009%29.jpg/200px-Papio_anubis_%28Serengeti%2C_2009%29.jpg"
 			});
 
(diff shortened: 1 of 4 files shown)


rmarinho:
Screenshot_1569508479

samhouts:
Yeah, we have this as an enhancement (https://github.com/xamarin/Xamarin.Forms/issues/2447). Setting it explicitly like this should bring iOS and Android in line with each other a the moment, but I had planned to add a new core property for TextDirection so it could be overridden. Not all labels should have the text direction set explicitly like this (my interpretation of the [guidelines](https://material.io/design/usability/bidirectionality.html#) is that English _should_ be left-to-right), so I was not ready to force it. However, I'm willing to accept that my interpretation was incorrect. :)

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7512.xaml.cs
```diff
@@ -0,0 +1,204 @@
...
+			BindingContext = new ViewModel7512();
+		}
+#endif
+
+		protected override void Init()
+		{
+
+		}
+
+		private void HandleButtonClick(object sender, EventArgs e)
```
samhouts:
```suggestion
		void HandleButtonClick(object sender, EventArgs e)
```

adrianknight89:
@samhouts If the current locale is Arabic or Hebrew, then the developer can enforce a RTL flow whereas English and most other languages should follow LTR. 

> Text should always be in the correct direction for the language it’s in. 

As you said, the recommendation for English would be LTR (i.e. no mirroring). Perhaps, the right way to use `FlowDirection` is in code based on the current locale rather than XAML for all locales.

I actually haven't used `FlowDirection` in my code yet. I assumed that Android and iOS would be smart enough to handle this for us. Is that not the case?

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7512.xaml.cs
rmarinho:
```suggestion
	 void HandleButtonClick(object sender, EventArgs e)
```

PR #7615, opened 2019-09-21
[Android] Remove ripple effect on CollectionView item when SelectionMode is None

### LINKED ISSUE #6051: [Android] CollectionView items retain ripple effect even when unselectable
### Description

Only affects APIs of Lollipop and up.

Once a CollectionView item has been selected, it retains the ripple tap effect even when selection is no longer enabled.

### Steps to Reproduce

1. In Control Gallery, navigate to CollectionView Gallery -> Selection Galleries -> Selection Mode
2. Change to SelectionMode "Single"
3. Select an item
4. Change to SelectionMode "None". The item should be de-selected.
5. Tap on the previously selected item. 
6. Tap on another item that was never selected.

### Expected Behavior

Tapping on the previously selected item should have the same effect as tapping on an item which was never selected.

### Actual Behavior

The previously selected item will retain the ripple effect.

### Basic Information

- Version with issue: 4.0.0

### Screenshots

leftover_ripple_effect

adrianknight89:
### Description of Change ###

Fixes a visual error with selection mode changes on Android / CollectionView. There is also a change in `Core`, which I'd like to discuss with the team to decide whether it should be merged or dismissed.

### Issues Resolved ### 

- fixes #6051

### Testing Procedure ###
Please follow the test steps in #6051. I'll update this section with more detailed instructions if the changes in `SelectableItemsView` are approved.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Items/SelectableItemsView.cs
+++ b/Xamarin.Forms.Core/Items/SelectableItemsView.cs
@@ -138,55 +138,21 @@ namespace Xamarin.Forms
 		static void SelectionModePropertyChanged(BindableObject bindable, object oldValue, object newValue)
 		{
 			var selectableItemsView = (SelectableItemsView)bindable;
-
-			var oldMode = (SelectionMode)oldValue;
 			var newMode = (SelectionMode)newValue;
-
-			IList<object> previousSelection = new List<object>();
-			IList<object> newSelection = new List<object>();
-
-			switch (oldMode)
-			{
-				case SelectionMode.None:
-					break;
-				case SelectionMode.Single:
-					if (selectableItemsView.SelectedItem != null)
-					{
-						previousSelection.Add(selectableItemsView.SelectedItem);
-					}
-					break;
-				case SelectionMode.Multiple:
-					previousSelection = selectableItemsView.SelectedItems;
-					break;
-			}
 
 			switch (newMode)
 			{
 				case SelectionMode.None:
+					selectableItemsView.SelectedItem = null;
+					selectableItemsView.SelectedItems = null;
 					break;
 				case SelectionMode.Single:
-					if (selectableItemsView.SelectedItem != null)
-					{
-						newSelection.Add(selectableItemsView.SelectedItem);
-					}
+					selectableItemsView.SelectedItems = null;
 					break;
 				case SelectionMode.Multiple:
-					newSelection = selectableItemsView.SelectedItems;
+					selectableItemsView.SelectedItem = null;
 					break;
 			}
-
-			if (previousSelection.Count == newSelection.Count)
-			{
-				if (previousSelection.Count == 0 || (previousSelection[0] == newSelection[0]))
-				{
-					// Both selections are empty or have the same single item; no reason to signal a change
-					return;
-				}
-			}
-
-			var args = new SelectionChangedEventArgs(previousSelection, newSelection);
-
-			SelectionPropertyChanged(selectableItemsView, args);
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


adrianknight89:
Previously, `SelectionModePropertyChanged` was doing some manual work to determine old and new selected items upon selection mode changes. To me, if `SelectionMode` changes from `Single` to `None`, then I should clear out `SelectedItem` because it doesn't make sense to have a selected item when the `CollectionView` signals no item is selectable. Similarly, if I go from `Multiple` to `Single` or `None`, I should clear out `SelectedItems`.

`SelectionModePropertyChanged` was raising `SelectionPropertyChanged` without actually setting the selected item and selected items properties. This causes issues in the renderers because we set `IsSelected` property to true/false on the view holders while `Core` is out of sync.

The current changes update `SelectedItem` and `SelectedItems` and let `Core` raise `SelectionPropertyChanged`.

hartez:
SelectedItems and SelectedItem are completely independent properties. This is intentional.

SelectionMode should be thought of as a way to determine which (if any) of these properties is currently being displayed/modified by the native control:

- If SelectionMode is set to None, then the native control is not affected by/has no effect on SelectedItem or SelectedItems. 
- If SelectionMode is set to Single, then the native control only affects/is affected by SelectedItem
- If SelectionMode is set to Multiple, then the native control only affects/is affected by SelectedItems

Tying these properties together causes some subtle inconsistencies and issues. For instance, and the moment this PR is setting SelectedItem and SelectedItems to `null` when changing _to_ SelectionMode.None. This implies that it should not be possible for SelectedItem or SelectedItems to have values when SelectionMode is None. However, if I set SelectionMode to None and _then_ set a value for SelectedItem (explicitly or via a binding), then I'm in an "invalid" state (by the implied rule). 

So if we were going to keep going in the direction of "SelectedItem and SelectedItems depend on SelectionMode", to keep internall consistent we'd have to enforce some rules:

1. Setting mode to None clears all selections
2. If the mode is None, all attempts to set the selection are invalid, and the selections remain empty

The PR currently does 1; we'd have to add some validation logic to the setters to enforce 2. As soon as we do this, the order of the properties in XAML becomes important. With a full version of this PR in place, this XAML:

`<CollectionView SelectedItem="foo" SelectionMode="Single" />`

has a different result than:

`<CollectionView SelectionMode="Single" SelectedItem="foo" />`

The first example will have nothing selected, the second will have a selected item. 

And we haven't yet taken the idea that "SelectedItem/SelectedItems should make 'sense' with the SelectionMode" to its logical conclusion. Moving from Single to Multiple mode and vice versa would need to make sense. We need more rules. What happens if I move from Single to Multiple - do I clear out the selection or do I keep the single selected item as part of SelectedItems? If I keep the single item, do I still retain it in SelectedItem or do I clear that value? 

How about the other way? If I move from Multiple SelectedItems to a Single SelectedItem, do I clear the selected items or keep one of them? If I keep one, which one? Do I remove the others from SelectedItems? 

All of these are answerable and, with a little effort, things we could handle in the code. But they require additional complexity in Core for all the validation logic and add more weird XAML ordering issues. And if we make all those decisions, it makes the control less flexible for developers. Right now, a developer can decide whether or not to keep a single selected item when switching to Multiple, or which (if any) item to keep when moving from Multiple to Single. That's between them and their view model.

tl;dr; - No to those selection mode changes.

adrianknight89:
@hartez I reverted the changes in Core.

If keeping `SelectedItem` and `SelectedItems` separate was intentional, then I understand. I wasn't quite sure whether they were meant to be tied together or not. I agree that the changes proposed above add more complexity. My initial impression was `SelectionMode` should be in charge of dictating how these two properties are maintained. I suppose if I implemented `CollectionView` of my own, I'd still go that route. However, this doesn't necessarily make the control **flexible** as you mentioned. Thanks for the explanation.

PR #7617, opened 2019-09-21
[iOS] Make button click easier to see when turning on CollectionView flag

adrianknight89:
Sometimes, when I tap the top button to enable the `CollectionView` flag in `CarouselViewGallery` and `CollectionViewGallery`, I miss the tappable area of the button while assuming the button was successfully tapped and so the Gallery crashes on me. The changes here will give us a better visual clue that the tap took place.


Xamarin.Forms.Controls/GalleryPages/CollectionViewGalleries/CarouselViewGalleries/CarouselViewGallery.cs
```diff
@@ -34,5 +41,16 @@ public CarouselViewGallery()
 				}
 			};
 		}
+
+		private void Button_Clicked(object sender, System.EventArgs e)
```
rmarinho: Can we drop private and call it ButtonClicked , thanks

adrianknight89: Done.

Issue #5455, opened 2019-03-05
ItemSizingStrategy MeasureAllItems does not work for iOS


Dresel:
### Description

I was playing around with the `CollectionView` on iOS and was facing problems with uneven row height. I saw there was an `ItemSizingStrategy` introduced lately, but it's still behaving weird though.

### Steps to Reproduce

1. Open the `DataTemplateGallery` CollectionViewGallery
2. Switch to `MeasureAllItems` to get the following result

<img src="https://user-images.githubusercontent.com/1769090/53488499-c49e8180-3a8e-11e9-9bf1-52fd4e1ff4c1.PNG" width="300">

4. Swiping to right and coming back shows the following result

<img src="https://user-images.githubusercontent.com/1769090/53488498-c49e8180-3a8e-11e9-9b22-1a5967a03ee1.PNG" width="300">

### Expected Behavior

Items should have the same height / width when setting `ItemSizingStrategy` to `MeasureAllItems`

### Actual Behavior

See above

### Basic Information

- Version with issue: master branch (50dbbcf5)
- Last known good version: -
- Platform Target Frameworks: 
  - iOS:  12.1
- Affected Devices: iPhone 6 (iOS 12.1.4)

kingces95:
Reproduces as described.

Dresel:
Is there any known work around (except setting height manually)? At the moment it's really useless with variable dynamic content.

adrianknight89:
Agreed. This needs to be given priority, imo.

timahrentlov:
I can't get any dynamic sizing to work on android with Pre 8.

timahrentlov:
With regards to ItemSizingStrategy, is https://gist.github.com/hartez/7d0edd4182dbc7de65cebc6c67f72e14/revisions current? 

It is mentioned here (https://devblogs.microsoft.com/xamarin/xamarin-forms-4-0-feature-preview-an-entirely-new-point-of-collectionview/) as a way to keep track of this component.

I spent some time today trying to use CollectionView on android, and found out that ItemSizingStrategy does not work. A more prominent display of status would have saved me some grief. And time.

pfedotovsky:
> Items should have the same height/width when setting ItemSizingStrategy to MeasureAllItems

Is it correct? According to the spec, MeasureFirstItem should produce the same height/width, while MeasureAllItems should measure all items individually.

adrianknight89:
[...]

PR #7622, opened 2019-09-23
[iOS] Fix MeasureFirstItem for CollectionView

### LINKED ISSUE #7621: [iOS] MeasureFirstItem is broken for CollectionView
### Steps to Reproduce

1. Go to CollectionView Galleries -> Item Size Galleries -> ItemSizing Strategy
2. Verify that each item is given the same size as the first item when `MeasureFirstItem` is selected
3. Scroll to the right as far as 20 items
4. Slowly scroll to the left to item 0
5. Observe that items on the left side of the screen change their size

### Expected Behavior

Scrolling should not affect item sizes when `MeasureFirstItem` is selected.

### Actual Behavior

Item sizes are broken when scrolling back.

adrianknight89:
### Description of Change ###

While looking at #5455, I noticed `MeasureFirstItem` does not work either. I provided a button to switch between `MeasureFirstItem` and `MeasureAllItems`. You could use this test as well as the gallery test in #5455 when figuring out a fix for `MeasureAllItems`.

**_MeasureAllItems:_**
To me, `MeasureAllItems` seems to be working in the test provided here. I assumed that each column would be as wide as the largest item in that column, and the other item would be centered. This is the case in this test. However, **the first item** is left aligned when the page is initially shown. If you scroll right and left, you'll see that the first item will be centered eventually. Something seems odd about the 4th (yellow) item. I was expecting it to be a bit wider since there is a lot of empty space to leverage. My head hurts too much right now to think about image `AspectFit` vs CollectionView item sizing, so I wanted to document it here for now. :)

### Issues Resolved ### 

- fixes #7621

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/CollectionView/HeightConstrainedTemplatedCell.cs
+++ b/Xamarin.Forms.Platform.iOS/CollectionView/HeightConstrainedTemplatedCell.cs
@@ -12,8 +12,9 @@ namespace Xamarin.Forms.Platform.iOS
 
 		public override void ConstrainTo(CGSize constraint)
 		{
+			base.ConstrainTo(constraint);
+
 			ConstrainedDimension = constraint.Height;
-			Layout(constraint);
 		}
 
 		protected override (bool, Size) NeedsContentSizeUpdate(Size currentSize)
--- a/Xamarin.Forms.Platform.iOS/CollectionView/TemplatedCell.cs
+++ b/Xamarin.Forms.Platform.iOS/CollectionView/TemplatedCell.cs
@@ -10,6 +10,8 @@ namespace Xamarin.Forms.Platform.iOS
 	{
 		public event EventHandler<EventArgs> ContentSizeChanged;
 
+		protected CGSize ConstrainedSize;
+
 		protected nfloat ConstrainedDimension;
 
 		DataTemplate _currentTemplate;
@@ -25,9 +27,18 @@ namespace Xamarin.Forms.Platform.iOS
 
 		internal IVisualElementRenderer VisualElementRenderer { get; private set; }
 
+		public override void ConstrainTo(CGSize constraint)
+		{
+			ConstrainedSize = constraint;
+		}
+
 		public override void ConstrainTo(nfloat constant)
 		{
 			ConstrainedDimension = constant;
+
+			// Reset constrained size in case ItemSizingStrategy changes
+			// and we want to measure each item
+			ConstrainedSize = default(CGSize);
 		}
 
 		public override UICollectionViewLayoutAttributes PreferredLayoutAttributesFittingAttributes(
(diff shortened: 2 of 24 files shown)


Xamarin.Forms.Platform.iOS/CollectionView/HeightConstrainedTemplatedCell.cs
```diff
@@ -12,8 +12,9 @@ public HeightConstrainedTemplatedCell(CGRect frame) : base(frame)
 
 		public override void ConstrainTo(CGSize constraint)
 		{
+			base.ConstrainTo(constraint);
+
 			ConstrainedDimension = constraint.Height;
-			Layout(constraint);
```
adrianknight89: @hartez Previously, there was a `Layout()` call inside `ConstrainTo()`, but `PreferredLayoutAttributesFittingAttributes` does layout too. Any reason for that? Removing the call here made cell sizing better.

Xamarin.Forms.Platform.iOS/CollectionView/TemplatedCell.cs
```diff
@@ -25,18 +27,27 @@ protected TemplatedCell(CGRect frame) : base(frame)
...
 
 		public override UICollectionViewLayoutAttributes PreferredLayoutAttributesFittingAttributes(
 			UICollectionViewLayoutAttributes layoutAttributes)
 		{
 			var preferredAttributes = base.PreferredLayoutAttributesFittingAttributes(layoutAttributes);
 
-			// Measure this cell (including the Forms element)
-			var size = Measure();
+			// Measure this cell (including the Forms element) if there is no constrained size
+			var	size = ConstrainedSize == default(CGSize) ? Measure() : ConstrainedSize;
```
adrianknight89: The reason `MeasureFirstItem` was broken is we were measuring here and assigning a new size to the cell. Since `MeasureFirstItem` constrains us to the first item, I added `ConstrainedSize` which is passed down from `PrepareCellForLayout()` in `ItemsViewLayout` as `ItemSize`.

Xamarin.Forms.Platform.iOS/CollectionView/TemplatedCell.cs
```diff
@@ -25,18 +27,27 @@ protected TemplatedCell(CGRect frame) : base(frame)
...
+			ConstrainedSize = constraint;
+		}
+
 		public override void ConstrainTo(nfloat constant)
 		{
 			ConstrainedDimension = constant;
+
+			// Reset constrained size in case ItemSizingStrategy changes
+			// and we want to measure each item
+			ConstrainedSize = default(CGSize);
```
adrianknight89: You could test this functionality in CollectionView Galleries -> Item Size Galleries -> ItemSizing Strategy. 

- Changing `ItemSizingStrategy` to measure all should measure each cell in `PreferredLayoutAttributesFittingAttributes`.
- Changing `ItemSizingStrategy` to measure all and then back to measure first should **not** break measure first.

adrianknight89:
@rmarinho I'll drop the private keyword from here too once the tests are done.

rmarinho:
The previous commit passed IOS uitests @hartez

Xamarin.Forms.Platform.iOS/CollectionView/HeightConstrainedTemplatedCell.cs
hartez: TBH, I don't know why there was a Layout call there. Seems to work fine without it.

PR #7641, opened 2019-09-24
[Android] Fix ApkPath to be able to run UI tests on Windows

adrianknight89:
This is to supplement #7579. This fixes ApkPath so that we can run Android UI tests on Windows. 

@rmarinho Can you let me know if this breaks the pipelines? Perhaps local and build work differently, and we need to code for that.


rmarinho:
Seems it's workign fien on CI, @PureWeen  does this helps you?

adrianknight89:
Great.

rmarinho:
@PureWeen we might want this on 4.3.30 because with have Xamarin UItest 3.0 there too ?

PureWeen:
These changes break the UI Tests locally on OSX

I created a different PR for this over here
https://github.com/xamarin/Xamarin.Forms/pull/7657

That targets 4.3.0

PR #7431, opened 2019-09-06
Dispose of HttpResponseMessage after stream is used

PureWeen:
### Description of Change ###

It appears that a race condition with how fast the stream is returned from HttpResponse leads to an image from a stream sometimes being loaded and then sometimes not

The Dispose on HttpContent class will only dispose of the returned stream if the status on the task returned from *ReadAsStreamAsync* is *RantoCompletion*

https://github.com/dotnet/corefx/blob/d3911035f2ba3eb5c44310342cc1d654e42aa316/src/System.Net.Http/src/System/Net/Http/HttpContent.cs#L540

So I think in some cases the stream will run to completion before the dispose is called and in other cases it won't

The call to ReadAsStreamAsync returns an invoked task. It doesn't start reading once awaited

I tried moving the await after the dispose explicitly

```C#
using (var client = new HttpClient())
                using (HttpResponseMessage response = await client.GetAsync(uri, cancellationToken))
                {
                    if (!response.IsSuccessStatusCode)
                    {
                        Internals.Log.Warning("HTTP Request", $"Could not retrieve {uri}, status code {response.StatusCode}");
                        return null;
                    }
                    returnValue = response.Content.ReadAsStreamAsync();
                }

                return await returnValue;
```

to see if that would keep it around but it doesn't

We could just not dispose of *HttpResponseMessage* and in theory it would be fine because the GC will eventually dispose of it but since it might be holding on to a large byte array I'd rather it get disposed of quickly. Which is why the solution is a bit more than just not disposing of *HttpResponseMessage*

I also opted out of using the Disposed event on StreamWrapper because I don't want to create a closure 

### Issues Resolved ### 
- fixes #7248 

### Platforms Affected ### 
- Android

### Testing Procedure ###
- UI tests have a few scenarios that use URLs though it seems like it's hit or miss if those catch this one
- the original issue has a few scenarios you can test as well

### PR Checklist ###

- [x] Has automated tests 
- [ ] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard

PR #7702, opened 2019-09-26
Wait for test to load before scrolling images

### LINKED ISSUE #6773: [Bug] Issue2354 failing occasionally and the Presidents aren't loading
### Description

Issue2354 - Needs to add a WaitForElement call at the top so that it knows when the test has loaded and it needs to wait for one of the images to have actually loaded

iOS - Images are only loading when you click on the cell
Android - Images aren't loading at all

### Steps to Reproduce
1. Run Issue2354

### Expected Behavior
All the presidents should load

### Actual Behavior
iOS - they only load when you click on the cell
Android - they don't load

### Basic Information

- Version with issue: 4.0
- Last known good version: Not sure. I know this test used to work. I tested 3.5 and it's still not working there either

- Platform Target Frameworks: 
  - iOS:  
  - Android:

PureWeen:
### Description of Change ###
This was most likely fixed by this issue
https://github.com/xamarin/Xamarin.Forms/pull/7431

So I just cleaned up the UI test here a bit to be more deterministic 

### Issues Resolved ### 
- fixes #6773

### Platforms Affected ### 
- Android

### Testing Procedure ###
- Android UI tests pass
- run Issue2354 and make sure the Presidents load

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2354.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2354.cs
@@ -45,6 +45,7 @@ namespace Xamarin.Forms.Controls.Issues
 			};
 		}
 
+		[Preserve(AllMembers = true)]
 		public class President
 		{
 			public President (string name, int position, string image)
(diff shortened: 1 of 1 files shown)


rmarinho:
since the test is not just Android (no conditional) we should always run iOS and Android so we make sure it doesn't break the test, actually this was good as it found a issue with iOS loading the image on the ViewCell, and @PureWeen already logged this: https://github.com/xamarin/Xamarin.Forms/issues/6773

PR #7707, opened 2019-09-26
JavaSDK build path fix for 16.3 CI servers

PureWeen:
### Description of Change ###
Fix for running Android builds on CI Server


knocte:
@PureWeen: according to this (https://github.com/jonathanpeppers/glidex/commit/88a87fdc62407fa5675e1455f2eb63c180d6c459) it's the other way around. I mean, if you were using 16.3 already, you wouldn't need this hack. So your CI must be still using 16.2.

PureWeen:
I commented incorrectly build servers are still on 16.2 we'll be on 16.3 in the next couple of weeks

PR #7737, opened 2019-09-29
[Android/iOS] Expose CollectionView scroll listener

adrianknight89:
### Description of Change ###

There is no way currently to listen to scroll changes in a custom renderer. This PR allows us to override setting a scroll listener so that we can add more complex logic to it if needed.

### PR Checklist ###


adrianknight89:
@samhouts Can you add the iOS label as well?

PR #7753, opened 2019-09-30
.create-nuget.sh and .create-stubs.sh files

g0rdan:
### Description of Change ###

Long story short, I wanted to debug XF and see a behavior in one specific place but, unfortunately, I didn't see clear instructions how I can do that. I saw files .create-nuget.bat and .create-stubs.bat that allow to assemble nuget packages that I can use in 'raw' debugging purposes. These files are a .bat files but I don't have any Windows machine. So, I've decided to make a replicas of those files but for bash (macOS). I've tried it on "ios" and "droid" profiles and it seems working.

### Issues Resolved ### 

No issues are resolved

### API Changes ###

 None

### Platforms Affected ### 

None

None

Not applicable

### Testing Procedure ###

In order to test the result of scripts you could input `sh .create-nuget.sh ios` in Terminal and wait for Xamarin.Forms.9.9.X.nupkg file in root folder of the project. After that you may create a new XF project in vs4mac and replace exsisting XF packages in Core and iOS projects on your own.

### PR Checklist ###

### DIFF
--- a/.create-nuget.sh
+++ b/.create-nuget.sh
@@ -7,7 +7,7 @@
 
 CONSOLE_PREFIX="-- SCRIPT MESSAGE:"
 CONFIG=Debug
-NUGET_EXE=caketools/nuget.exe
+NUGET_EXE=nuget.exe
 
 # Download NuGet if it does not exist.
 if [ ! -f "$NUGET_EXE" ]; then
--- a/.create-nuget.sh
+++ b/.create-nuget.sh
@@ -1,0 +1,115 @@
+ #!/bin/bash 
+
+# This is not our official nuget build script.
+# This is used as a quick and dirty way create nuget packages used to test user issue reproductions.
+# This is updated as XF developers use it to test reproductions. As such, it may not always work.
+# This is not ideal, but it's better than nothing, and it usually works fine.
+
+CONSOLE_PREFIX="-- SCRIPT MESSAGE:"
+CONFIG=Debug
+NUGET_EXE=caketools/nuget.exe
+
+# Check if Homebrew is installed
+if [[ $(command -v brew) == "" ]]; then
+    echo "${CONSOLE_PREFIX} Installing Homebrew"
+    /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+else
+    echo "${CONSOLE_PREFIX} Updating Homebrew"
+    brew update
+fi
+
+# Check if Wine is installed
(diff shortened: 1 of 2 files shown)


.create-nuget.sh
```diff
@@ -0,0 +1,115 @@
...
+    wine $NUGET_EXE restore .xamarin.forms.nuget.sln
+    msbuild /v:m /p:platform="any cpu" /p:WarningLevel=0 .xamarin.forms.nuget.sln
+    ;;
+esac
+
+DEBUG_VERSION=0
+wine $NUGET_EXE pack .nuspec/Xamarin.Forms.nuspec -properties configuration=$CONFIG\;platform=anycpu -Version 9.9.$DEBUG_VERSION
+if [ ! -n $CREATE_MAP_NUGET ]; then
+  # Requires building x86, x64, AMD
+  wine $NUGET_EXE pack .nuspec/Xamarin.Forms.Maps.nuspec -properties configuration=$CONFIG\;platform=anycpu -Version 9.9.$DEBUG_VERSION
```
PureWeen: Does this need wine to work?  Nuget is installed as part of the mono framework and should be able to pack the nuspecs

Plus we should be able to just use 
*msbuild /t:restore" instead of *wine nuget_exe*

g0rdan: You're right. I don't know what I was thinking about.
Question. Can I use msbuild for nuget packaging? I see in docs that it is possible. For example `msbuild -t:pack <path to .csproj file> -p:NuspecFile=<path to nuspec file> -p:NuspecProperties=<> -p:NuspecBasePath=<Base path>` but I cannot pick up right parameters.

PureWeen: You might need to import this
> you must import NuGet.Build.Tasks.Pack.targets 

https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-using-a-nuspec

But might not work until those projects support sdk style or are converted to using SDK.Extras

samhouts:
@g0rdan Are you able to address the comment in https://github.com/xamarin/Xamarin.Forms/pull/7753#pullrequestreview-351303782? Thanks!

g0rdan:
@samhouts Done. Although, as @kingces95 has mentioned, the script needs to be constantly updated and maybe it's not a great idea to put it here. All the development goes on Windows anyway, I believe.

PureWeen:
@g0rdan I think this is a very useful script

What you've done gives us a baseline for getting this all working off windows (which will be really helpful for people) and then we'll just update it as we need to

rmarinho:
We are trying to use cake now for this. 

Thanks

PR #6798, opened 2019-07-06
Fix for Switch on Android losing shadow

andreinitescu:
### Description of Change ###

After #6312 Switch lost shadow on Android, see Shane's comment

fixes #5415

### API Changes ###

 
 None

### Platforms Affected ### 

- Android

### After Screenshots ### 

image

### PR Checklist ###

- [ ] Has automated tests 
- [x] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard

PR #7777, opened 2019-10-02
[Android] Fix Switch Thumb wrong color

### LINKED ISSUE #7249: [Bug] Android Switch Slider wrong color?
### Description

After updating from 4.1.0 (first service release) to 4.2.0, the switch on Android has a strange behavior. If I switch them, the circle stays white which was not the case before and is not the case when I reload the page, see: https://i.imgur.com/sWVbdWX.png - All Orange Switches where on IsToggles while the page has loaded, the white ones just got toggled after the page loads. 

This bug is only happening for the Switch itself. It does not occur with SwitchCell controls. 

### Steps to Reproduce

1. Update to 4.2.0

### Expected Behavior

The Switch should not have different color. If this was an intended behvior change it should be equal accross all switches. 

### Actual Behavior

Different if the switch was loaded with a true value for IsToggled or if it was set after the page loads.

### Basic Information

- Version with issue:
- Last known good version: 4.1.0 (first service release)
- IDE:
- Platform Target Frameworks: 
  - Android: Latest stable
- Nuget Packages:
- Affected Devices:

jsuarezruiz:
### Description of Change ###

Fix **Switch Thumb wrong color on Android**. We introduce an option to change the ThumbColor here https://github.com/xamarin/Xamarin.Forms/pull/6798
We save in a variable the color of the Thumb https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/Renderers/SwitchRenderer.cs#L87 and use it if ThumbColor is not set https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/Renderers/SwitchRenderer.cs#L141
The problem is that the default Thumb color is different if the switch is Checked or not (white or pink respectively). Applied changes to set the color of the ThumbColor only when it is not Color.Default. 

### Issues Resolved ### 

- fixes #7249

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

None

#### Before
7249-before

#### After
7249-after

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.ControlGallery.Android/Issue7249SwitchRenderer.cs
+++ b/Xamarin.Forms.ControlGallery.Android/Issue7249SwitchRenderer.cs
@@ -1,0 +1,66 @@
+﻿using Android.Content;
+using Android.Graphics;
+using Android.OS;
+using Xamarin.Forms.Platform.Android;
+using Xamarin.Forms.Controls.Issues;
+using static Android.Widget.CompoundButton;
+
+namespace Xamarin.Forms.ControlGallery.Android
+{
+	public class Issue7249SwitchRenderer : SwitchRenderer
+	{
+		Issue7249Switch _view;
+
+		public Issue7249SwitchRenderer(Context context) : base(context)
+		{
+
+		}
+
+		protected override void OnElementChanged(ElementChangedEventArgs<Switch> e)
+		{
+			base.OnElementChanged(e);
+
+			if (e.OldElement != null || e.NewElement == null)
+				return;
+
+			_view = (Issue7249Switch)Element;
+
+			if (Build.VERSION.SdkInt >= BuildVersionCodes.JellyBean)
+			{
+				if (Control != null)
+				{
+					if (Control.Checked)
+					{
+						Control.TrackDrawable.SetColorFilter(_view.SwitchOnColor.ToAndroid(), PorterDuff.Mode.SrcAtop);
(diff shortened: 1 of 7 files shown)


samhouts:
@jsuarezruiz What happens if you have a custom switch renderer that sets the color? Can you add a test case for that to make sure we won't break those? Thanks!

Xamarin.Forms.ControlGallery.Android/Issue7249SwitchRenderer.cs
```diff
@@ -0,0 +1,66 @@
...
+using Android.OS;
+using Xamarin.Forms.Platform.Android;
+using Xamarin.Forms.Controls.Issues;
+using static Android.Widget.CompoundButton;
+
+namespace Xamarin.Forms.ControlGallery.Android
+{
+	public class Issue7249SwitchRenderer : SwitchRenderer
+	{
+		private Issue7249Switch _view;
```
samhouts:
```suggestion
		Issue7249Switch _view;
```

samhouts:
@jsuarezruiz Can you rebase and fix the build? Thanks!

samhouts:
@jsuarezruiz Sorry to ask again, but can you retarget to 4.4.0? Thanks!!

jsuarezruiz:
Retarget to 4.4.0.

Issue #7758, opened 2019-10-01
[iOS] EmptyView is not rendered in screen center 


adrianknight89:
### Description

`EmptyView` is not centered in the middle of the screen when `CollectionView` has upper and lower siblings in its parent. Possibly related to #7251.

### Steps to Reproduce

1. Run the repro
2. Observe that the center of empty view content is not aligned with the red guide line
3. Comment out the black grids
4. Observe that `EmptyView` is centered properly

### Basic Information

- Version with issue: 4.4.0.1201-nightly
- Last known good version: N/A
- IDE: VS2019
- Platform Target Frameworks: 
  - iOS: 12 
- Affected Devices: Tested on XS

### Reproduction Link

[App144.zip](https://github.com/xamarin/Xamarin.Forms/files/3674630/App144.zip)

adrianknight89:
I briefly looked at this. I don't think there is anything wrong with EmptyView. The issue, at first glance, seems to be with CollectionView as it is measured to be the full screen size. Once the fix is provided, we can use the repro here as a test.

jsuarezruiz:
Tested and verified the issue.

rmarinho:
I don't see the issue in the latest versions. Please reopen if this still a issue.

![Simulator Screen Shot - iPhone 8 Plus - 2020-02-21 at 17 22 22](https://user-images.githubusercontent.com/1235097/75056359-c4cd3380-54ce-11ea-9f28-3a95a66d9432.png)
![Simulator Screen Shot - iPhone 11 - 2020-02-21 at 17 17 39](https://user-images.githubusercontent.com/1235097/75056362-c696f700-54ce-11ea-9b30-5753cb01eefc.png)

PR #7794, opened 2019-10-03
[iOS] Fixed CollectionView EmptyView incorrect bounds issue

### LINKED ISSUE #7251: [Bug] iOS - CollectionView EmptyView has incorrect bounds, height, width
### Description

I have a PopupPage (Rg.Plugins.Popup) that is acting as a multiple selection dialog, and am using a CollectionView to display the items for multiple selection. One particular use case in the app I'm working on requires selection from about 300 items. On Android, these items render instantly, however on iOS (with the update to 4.2 to get rid of the empty item on bounce) it is taking many seconds to render. I've added an empty view to display a loading indicator while waiting for those items to render, however, the empty view isn't respecting the fact that it's in a dialog with margins, and instead behaves as though it thinks the CollectionView is taking up the entire screen. The activity indicator is set up with Horizontal and VerticalOptions set to Center. It is not centered in the CollectionView, though. Instead it is offset on the X coordinate by approximately the Dialog's margin, and on the Y coordinate by the Dialog's margin plus the height of the header row. When the items are loaded, the CollectionView looks fine, only the EmptyView is having trouble here.

### Expected Behavior

CollectionView's EmptyView should be constrained to the bounds of the CollectionView.

### Actual Behavior

CollectionView's EmptyView behaves as though it is taking up the full screen regardless of the CollectionView's dimensions.

### Basic Information

- Version with issue:
- Last known good version:
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Nuget Packages:
- Affected Devices:

### Screenshots

IMG_0007 1

jsuarezruiz:
### Description of Change ###

Fixed CollectionView **EmptyView** incorrect bounds issue.
_NOTE: This fix also affects to CarouselView._

### Issues Resolved ### 

- fixes #7251 and #7758

### API Changes ###

 None

### Platforms Affected ### 

- iOS

None

#### Before
7758-before

#### After
7758-after

Captura de pantalla 2019-10-04 a las 17 28 09

### Testing Procedure ###
Launch Core Gallery and open the Issue 7758. If the red line is between Item 3 and 4, the test passes.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewController.cs
+++ b/Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewController.cs
@@ -141,9 +141,10 @@ namespace Xamarin.Forms.Platform.iOS
 			{
 				ItemsViewLayout.ConstrainTo(CollectionView.Bounds.Size);
 				_initialConstraintsSet = true;
+				UpdateEmptyView();
 			}
 		}
-
+  
 		protected virtual UICollectionViewDelegator CreateDelegator()
 		{
 			return new UICollectionViewDelegator(ItemsViewLayout, this);
@@ -370,7 +371,7 @@ namespace Xamarin.Forms.Platform.iOS
 
 					// Now that the native empty view's frame is sized to the UICollectionView, we need to handle
 					// the Forms layout for its content
-					_emptyViewFormsElement.Layout(CollectionView.Frame.ToRectangle());
+					_emptyViewFormsElement.Layout(_emptyUIView.Frame.ToRectangle());
 				}
 			}
 			else
--- a/Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewController.cs
+++ b/Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewController.cs
@@ -370,7 +370,7 @@ namespace Xamarin.Forms.Platform.iOS
 
 					// Now that the native empty view's frame is sized to the UICollectionView, we need to handle
 					// the Forms layout for its content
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewController.cs
```diff
@@ -370,7 +370,7 @@ void UpdateEmptyViewVisibility(bool isEmpty)
 
 					// Now that the native empty view's frame is sized to the UICollectionView, we need to handle
```
adrianknight89: Should this comment change? In theory, if the native empty view's frame is sized to the UICollectionView, then `_emptyUIView.Frame.ToRectangle()` should work. Looks like `_emptyUIView` still needs to be sized correctly?

jsuarezruiz: Mnn, good point. The problem is that when we initialize EmptyView, the Bounds have not yet been adjusted and occupies the entire screen size. I have applied changes and we do the initialization of EmptyView in `ViewWillLayoutSubviews`. When the UICollectionView bounds change, just adjust the size of the EmptyView subview.

jsuarezruiz:
Oops, we have two PRs now for the same.
The problem is that EmptyView does not take the correct size. In this PR it is fixed initializing EmptyView in `ViewWillLayoutSubviews` and in PR 7811 is using `ViewDidLoad` method. 
Both seems correct. ViewDidLoad will be executed before ViewWillLayoutSubviews. So, to avoid having two PRs for the same, I am going close this PR.

PR #7771, opened 2019-10-02
In Android : Reverse layout only when default layout direction is the…

Elashi:
… opposite of the user settings (#7393)
### Description of Change ###
Android: If the user adds this line `android:supportsRtl="true"` to application node in AndroidManifest.xml and add `Window.DecorView.LayoutDirection = LayoutDirection.Rtl;` into MainActivity.cs , the CollectionView fails to display in the correct direction chosen by the user when setting the property FlowDirection.

### Issues Resolved ### 
- fixes #7397 

### API Changes ###
None
 
### Platforms Affected ### 
- Android

### Behavioral/Visual Changes ###
When setting CollectionView.FlowDirection to RightToLeft and setting `Window.DecorView.LayoutDirection = LayoutDirection.Rtl` , the CollectionView will display from Right To Left

### Before/After Screenshots ### 

Before: the whole screen is display correctly from right-to-left but the CollectionView is left-to-right
image

After: After the fix, the whole screen is displayed correctly from right-to-left including the CollectionView is also right-to-left
image

### Testing Procedure ###
Android:
1 - Add `android:supportsRtl="true"` to application node in AndroidManifest.xml.
2- add `Window.DecorView.LayoutDirection = LayoutDirection.Rtl;` in MainActivity.cs
3- Set CollectionView's property `ItemsLayout="HorizontalList"`
4- Set CollectionView's direction `FlowDirection="RightToLeft"`

### PR Checklist ###

- [ ] Targets the correct branch
- [ ] Tests are passing (or failures are unrelated)

Issue #6342, opened 2019-05-29
[Bug] Shell FlowDirection follows Device FlowDirection in all cases.


aljaiban:
### Description
When setting the FlowDirection on Shell to RTL, it only take effects if the Device FlowDirection is the same i.e. RTL. When the Device FlowDirection is LTR and you set Shell FlowDirection to RTL, it does not work

### Steps to Reproduce

I tried the following scenario on Android emulator:

Shell: Explicit RightToLeft 
Page: MatchParent

When device region is Arabic/SA (RTL) : The Shell displays correctly and the page too.

When device region is English/US: The Shell displays Incorrectly i.e. LTR and the page displays correctly i.e. RTL.  See snapshot.

### Expected Behavior
Shell should use the direction specified in its FlowDirection attribute regardless of the Device FlowDirection.

### Actual Behavior
Right now it follows the Device FlowDirection.

### Basic Information

- Version with issue: Xamarin Forms 4.0
- Last known good version: None
- IDE: Visual Studio 2019 on Windows
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  
### Screenshots
![image](https://user-images.githubusercontent.com/7361168/58554672-46ce4b80-8220-11e9-995d-48c8cfb316ae.png)

samhouts:
Duplicate of #6227

aljaiban:
@samhouts  Not duplicate. In case of 6227 the Device FlowDirection is the same as the Shell Flow Direction and both are RTL.

In this case, the Shell menu appear LTR  even Shell is specified as RTL (Device is LTR.)

samhouts:
@aljaiban Okay, I'll reopen this. With any luck, we'll be able to fix them both at the same time! Thanks!

samhouts:
@aljaiban  Can you please attach a small project that demonstrates this issue? Thanks!

aljaiban:
Attached a sample project and snapshot. Note that you need to run the project on an LTR (English/US for example) device/emulator/simulator for the issue to appear.
![Snapshot](https://user-images.githubusercontent.com/7361168/58752128-333a1380-84b2-11e9-8b82-80702fe6882d.png)

https://1drv.ms/u/s!Anxli5dwfVWNpNcQNeJpj3ZhNKPryw

jfversluis:
Making sure we don't lose access to the reproduction code adding it here: 
[...]

Issue #6227, opened 2019-05-20
[Bug] Xamarin.Forms Shell Opener Issue for "RTL Language" when Tap at Hamburger


ranacodendev:
### Description
using xamarin.forms 4.0.0 -pre9 as Shell App, over all application is working fine, but right now i have one serious issue i.e.
My App is in Arabic Language , so when i Tap at Hamburger to open Shell/Menu it works differently for iPhone/iPad i.e. at iPad it open Shell/Menu Right to Left fine.
but at iPhone it open Shell/Menu Left to Right not acceptable by my Customer.

so please help me in this regards, that its behavior should be same for both Devices i.e. it should be open from Right to Left
### Steps to Reproduce

1. 
2. 
3. 

### Basic Information

- Version with issue:
- Last known good version:
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Android Support Library Version: 
- Nuget Packages:
- Affected Devices:

### Screenshots

<img width="1317" alt="WithMenu" src="https://user-images.githubusercontent.com/1707247/58007774-a5dee280-7af3-11e9-8630-a0627d3eb76f.png">
<img width="1373" alt="WithoutMenu" src="https://user-images.githubusercontent.com/1707247/58007775-a6777900-7af3-11e9-83b9-414f230650da.png">

ranacodendev:
dear @pauldipietro ,
     Is there any way that i can open this Shell/Menu in Standard way, i.e. Right to Left, may be using some Render, please help me in this regards ASAP.
Please share any source code sample if possible,
Thanks,

ranacodendev:
hello anybody there to help me PLEASE.

Hedyia:
To solve this issue the device should be in Arabic setting. or you could use this plugin.
https://github.com/saeedafshari/RedCorners.Forms

mohamadayash:
Hello I have the same problem 
![IMG_1808](https://user-images.githubusercontent.com/11196273/58631354-fcfe6780-82ea-11e9-9733-3a40552d5962.PNG)

ranacodendev:
so any update on this issue as explained above, 
is there any way to solve it, i.e. always open menu from RTL in both devices cases iPhone/iPad and also in Android.?

please share any code or any thing which helps me in this regards,
Thanks,
MN

mohamadayash:
Hello , any news about this issue ?

AmerBakeer:
[...]

PR #7809, opened 2019-10-04
In Android : Reverse layout only when default layout direction is the…

### LINKED ISSUE #7397: [Bug] Forms 4.2 : CollectionView : Android : RTL
### Description
I have a horizontal CollectionView inside a vertical CollectionView.  The horizontal CollectionView's direction was working perfectly RightToLeft in version 4.1 but when I upgraded to 4.2 , its direction now is flipped LeftToRight only in Android.

### Steps to Reproduce

1. In AppShell.xaml add property to Shell : FlowDirection="RightToLeft". I also added this property to inner CollectionView

2.  I have this block of code in MainActivity.cs as an attempt to make the default direction for the app as RTL and the language Arabic:
`            string cultureName = "ar";
            var locale = new Java.Util.Locale(cultureName);
            Java.Util.Locale.Default = locale;
            var config = new Android.Content.Res.Configuration();
            config.SetLocale(locale);
            config.SetLayoutDirection(locale);
            BaseContext.CreateConfigurationContext(config);
`
3.  In AppDelegate.cs , I have the following to force default RTL dirction:
`            var selector = new ObjCRuntime.Selector("setSemanticContentAttribute:");
            IntPtr_objc_msgSend(UIView.Appearance.Handle, selector.Handle, UISemanticContentAttribute.ForceRightToLeft);
`

### Expected Behavior
Inner CollectionView shows items displayed RightToLeft

### Actual Behavior
Inner CollectionView shows items displayed LeftToRight

### Basic Information

- Version with issue: 4.2
- Last known good version: 4.1
- IDE: Visual Studio 2019
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:  not applicable
- Nuget Packages:
- Affected Devices:

### Screenshots
Version 4.1
ver 4 1

Version 4.2
ver 4 2

If you notice, the hamburger icon also disappears.

Elashi:
… opposite of the user settings (#7397)

### Description of Change ###

Android: If CollectionView.FlowDirection=RightToLeft , the change will make sure to display the CollectionView in the correct direction (right-to-left) regardless of the application direction settings defined in AndroidManifest and MainActivity.

### Issues Resolved ### 

- fixes #7397 

### API Changes ###
 
 None

### Platforms Affected ### 
- Android

The change will force CollectionView to display direction RightToLeft even when application default direction is Rtl in AndroidManifest and MainActivity which currently does not.

Before the change, if I set the default direction of the application to right-to-left in AndroidManifest and MainActivity, everything will change direction to right to left including Shell controls but CollectionView will display left-to-right even though its FlowDirection is set to RightToLeft.
Before the change:
image
After the change
image

### Testing Procedure ###

1- in AndroidManifest.xml , add the attribute `android:supportsRtl="true"` to `application` node.
2- in MainActivity.cs , add `Window.DecorView.LayoutDirection = LayoutDirection.Rtl;` to `OnCreate` 
3- in your xaml page add the following property to your CollectionView `FlowDirection=RightToLeft`

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/CollectionView/ItemsViewRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/CollectionView/ItemsViewRenderer.cs
@@ -539,19 +539,6 @@ namespace Xamarin.Forms.Platform.Android
 			{
 				return;
 			}
-
-			var effectiveFlowDirection = ((IVisualElementController)Element).EffectiveFlowDirection;
-
-			if (effectiveFlowDirection.IsRightToLeft()
-				&& Context.GetActivity().Window.DecorView.LayoutDirection == LayoutDirection.Ltr)
-			{
-				linearLayoutManager.ReverseLayout = true;
-			}
-			else if (effectiveFlowDirection.IsLeftToRight()
-				&& Context.GetActivity().Window.DecorView.LayoutDirection == LayoutDirection.Rtl)
-			{
-				linearLayoutManager.ReverseLayout = false;
-			}
 		}
 
 		protected virtual int DetermineTargetPosition(ScrollToRequestEventArgs args)
--- a/Xamarin.Forms.Platform.Android/CollectionView/ItemsViewRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/CollectionView/ItemsViewRenderer.cs
@@ -542,11 +542,13 @@ namespace Xamarin.Forms.Platform.Android
 
 			var effectiveFlowDirection = ((IVisualElementController)Element).EffectiveFlowDirection;
 
-			if (effectiveFlowDirection.IsRightToLeft())
+			if (effectiveFlowDirection.IsRightToLeft()
+				&& Context.GetActivity().Window.DecorView.LayoutDirection == LayoutDirection.Ltr)
 			{
 				linearLayoutManager.ReverseLayout = true;
 			}
-			else if (effectiveFlowDirection.IsLeftToRight())
+			else if (effectiveFlowDirection.IsLeftToRight()
+				&& Context.GetActivity().Window.DecorView.LayoutDirection == LayoutDirection.Rtl)
 			{
 				linearLayoutManager.ReverseLayout = false;
 			}
--- a/Xamarin.Forms.Platform.Android/CollectionView/EdgeSnapHelper.cs
+++ b/Xamarin.Forms.Platform.Android/CollectionView/EdgeSnapHelper.cs
@@ -1,5 +1,6 @@
 using Android.Support.V7.Widget;
 using AView = Android.Views.View;
+using ALayoutDirection = Android.Views.LayoutDirection;
 
 namespace Xamarin.Forms.Platform.Android
 {
@@ -13,8 +14,11 @@ namespace Xamarin.Forms.Platform.Android
 		}
 
 		protected static bool IsLayoutReversed(RecyclerView.LayoutManager layoutManager)
-		{
-			if (layoutManager is LinearLayoutManager linearLayoutManager)
+        {
+            if (layoutManager.LayoutDirection == (int)(ALayoutDirection.Rtl))
+                return true;
+
+            if (layoutManager is LinearLayoutManager linearLayoutManager)
 			{
 				return linearLayoutManager.ReverseLayout;
 			}
--- a/Xamarin.Forms.Platform.Android/CollectionView/EdgeSnapHelper.cs
+++ b/Xamarin.Forms.Platform.Android/CollectionView/EdgeSnapHelper.cs
@@ -14,11 +14,11 @@ namespace Xamarin.Forms.Platform.Android
 		}
 
 		protected static bool IsLayoutReversed(RecyclerView.LayoutManager layoutManager)
-        {
-            if (layoutManager.LayoutDirection == (int)(ALayoutDirection.Rtl))
-                return true;
+		{
+			if (layoutManager.LayoutDirection == (int)(ALayoutDirection.Rtl))
+				return true;
 
-            if (layoutManager is LinearLayoutManager linearLayoutManager)
+			if (layoutManager is LinearLayoutManager linearLayoutManager)
 			{
 				return linearLayoutManager.ReverseLayout;
 			}
@@ -26,13 +26,13 @@ namespace Xamarin.Forms.Platform.Android
 			return false;
 		}
 
-		protected int[] CalculateDistanceToFinalSnap(RecyclerView.LayoutManager layoutManager, AView targetView, 
+		protected int[] CalculateDistanceToFinalSnap(RecyclerView.LayoutManager layoutManager, AView targetView,
 			int direction = 1)
 		{
 			var orientationHelper = CreateOrientationHelper(layoutManager);
 			var isHorizontal = layoutManager.CanScrollHorizontally();
 			var rtl = isHorizontal && IsLayoutReversed(layoutManager);
-			
+
 			var size = orientationHelper.GetDecoratedMeasurement(targetView);
 
 			var hiddenPortion = size - VisiblePortion(targetView, orientationHelper, rtl);
@@ -54,7 +54,7 @@ namespace Xamarin.Forms.Platform.Android
 			var size = orientationHelper.GetDecoratedMeasurement(view);
 
 			var portionInViewPort = VisiblePortion(view, orientationHelper, reversed && isHorizontal);
-			
+
 			// Is the first visible view at least halfway on screen?
 			return portionInViewPort >= size / 2;
 		}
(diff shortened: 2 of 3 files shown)


Elashi:
Not successful checks !
What shall I do now?

Xamarin.Forms.Platform.Android/CollectionView/ItemsViewRenderer.cs
```diff
@@ -542,11 +542,13 @@ protected virtual void ReconcileFlowDirectionAndLayout()
 
 			var effectiveFlowDirection = ((IVisualElementController)Element).EffectiveFlowDirection;
 
-			if (effectiveFlowDirection.IsRightToLeft())
+			if (effectiveFlowDirection.IsRightToLeft()
```
PureWeen: Can we just delete this code to fix this?

```C#
if (effectiveFlowDirection.IsRightToLeft())
			{
				linearLayoutManager.ReverseLayout = true;
			}
			else if (effectiveFlowDirection.IsLeftToRight())
			{
				linearLayoutManager.ReverseLayout = false;
			}
```

I'm not totally clear what it's accomplishing. If the flow direction on the CollectionView is set to RightToLeft then the *LayoutDirection* on the RecyclerView gets set to *LayoutDirection.RTL*

I tested with your reproduction here
https://github.com/xamarin/Xamarin.Forms/issues/7397#issuecomment-532625478

And I tested with the test procedure steps you have outlined in this PR and in both cases if we just delete the ReverseLayout code then it correctly lays out RTL


Which is all the reason why setting SupportsRTL to True was causing it to break
https://github.com/xamarin/Xamarin.Forms/issues/7397#issuecomment-532625478

Because if SupportsRTL is false than LayoutDirection doesn't do anything so when you set it to true it was then getting reversed twice. Once because LayoutDirection is set to RTL and a second time because ReverseLayout is set to True

Based on all this I feel like the fix would be to just 
- Delete the code that sets *ReverseLayout* at all
- Also modify the code inside *SingleSnapHelper.IsLayoutReversed* to test for LayoutDirection on RecyclerView

The only reason that I can think of why you'd use ReverseLayout would be if you want RTL support on the CollectionView but the user hasn't set SupportsRTL inside the Android  Manifest

But our documentation indicates that you need to set that to true in order for FlowDirection to work

https://docs.microsoft.com/en-us/xamarin/xamarin-forms/app-fundamentals/localization/right-to-left#android

If we wanted RTL to work when *SupportsRTL* is false (which I don't think we do) I think the code fix would be more like this

```C#
			if(!Context.HasRtlSupport())
			{
				if (effectiveFlowDirection.IsRightToLeft())
				{
					linearLayoutManager.ReverseLayout = true;
				}
				else if (effectiveFlowDirection.IsLeftToRight())
				{
					linearLayoutManager.ReverseLayout = false;
				}

			}
```

@hartez @samhouts  thoughts?

hartez: @PureWeen Yeah, you can probably just delete that. I either was not aware of supportsRTL when I was originally writing/testing that stuff, or I was trying to get the support to work without it because we were targeting API 15 (and supportsRTL is from 17 onward, I think).

TBH, I can't remember. But I think you're right that the code is wrong, especially now that 19 is the minimum API level.

PureWeen: Alright I pushed those changes and also modified the checks in the snap helper. Let me know if that all works as you'd expect it to. I ran a few tests with snap helpers and it seems to work fine when set to RTL.

@Elashi can you sign the CLA?

Elashi:
My humble opinion is that removing the code makes FlowDirection useless.  And leaving the code will harm no one. 
For my current need removing the code will solve my problem but I’m worried if for any reason someone wanted a CollectionView with RightToLeft direction in a LeftToRight context.  Or the opposite.
I respect you guys but I really don’t see the benefit behind removing the code.

PureWeen:
@Elashi 

> My humble opinion is that removing the code makes FlowDirection useless.  And leaving the code will harm no one.

The problem is that ReverseLayout does double duty. Our flow direction code already sets the layout to RTL so the setting ReverseLayout undoes that.  What's the advantage to using *ReverseLayout* over LayoutDirection?

> For my current need removing the code will solve my problem but I’m worried if for any reason someone wanted a CollectionView with RightToLeft direction in a LeftToRight context.  Or the opposite.

Then they just set the Page to LeftToRight and the CollectionView to RightToLeft. What am I missing?

Elashi:
> Then they just set the Page to LeftToRight and the CollectionView to RightToLeft. What am I missing?

This won't resolve the problem.  Setting the CollectionView to RightToLeft will not work as current defaults (when Manifest and MainActivity were Ltr) and Setting the CollectionView to LeftToRight will not work if Manifest and MainActivity were set to RTL.  In these cases reversing is necessary.

[Checkout our long discussion in the previous PR.](https://github.com/xamarin/Xamarin.Forms/pull/7771#discussion_r330659294)

LayoutDirection | FlowDirection | Need to Reverse
-- | -- | --
Ltr | RightToLeft | Y
Ltr | LeftToRight | N
Rtl | RightToLeft | N
Rtl | LeftToRight | Y

PureWeen:
>  if Manifest and MainActivity were set to RTL

- How do you set the Manifest to RTL?  

Setting *android:supportsRtl="true"* doesn't set the application to RTL all it does is enable the RTL APIs like layout direction. It's just a declaration to googles APIs that your application is RTL capable. For example if you go into developer options and  "Force RTL" then it will change the default LayoutDirection on all your layouts to RTL but that will only have an effect if you have supportsRTL set to true.  LayoutDirection just becomes a noop if supportsRtl is false but setting supportsRTL to true has no direct effect on an application.  There is a bug where CV doesn't respect "Force RTL" (https://github.com/xamarin/Xamarin.Forms/issues/7920). 

- Setting the decor view LayoutDirection to RTL  

I would say falls into the "not recommended" way of doing it. Forms does a lot of work to propagate flow direction so if you use forms apis (FlowDirection) then Forms can act accordingly. Once you start setting LayoutDirection yourself natively getting Forms to play with that is going to be undefined.  That being said :-)   I think the bug you are running into is in fact the one I've logged her (https://github.com/xamarin/Xamarin.Forms/issues/7920)  

If you comment out this line on the CV
https://github.com/xamarin/Xamarin.Forms/blob/823013dbc8f3095f25ee19e6257e5acc7b5e3696/Xamarin.Forms.Platform.Android/CollectionView/ItemsViewRenderer.cs#L476

And then set the DecorView to RTL then the CV correctly lays out as RTL. The CV needs to not explicitly set LayoutDirection if FlowDirection hasn't been set anywhere in the Xamarin.Forms hierarchy

Elashi:
@PureWeen I recommend that you read the previous PR and Issue.  We already discussed these points in it.  In short, Shell icons does not respond to FlowDirection=RightToLeft

Xamarin.Forms.Platform.Android/CollectionView/ItemsViewRenderer.cs
Elashi: > Because if _supportsRtl_ is false then _RecyclerView.LayoutDirection_ doesn't do anything, so when you set _supportsRtl_ to true it was then getting reversed twice. Once because _RecyclerView.LayoutDirection_ is set to RTL and a second time because _LinearLayoutManager.ReverseLayout_ is set to True

The PR will reduce the number of reverses by adding more restricting conditions.

Elashi:
> I would say falls into the "not recommended" way of doing it.

My application is in Arabic , it has to display Right To Left regardless of the settings in the device (force right to left or not)

> Forms does a lot of work to propagate flow direction so if you use forms apis (FlowDirection) then Forms can act accordingly.

Setting FlowDirection=RightToLeft for Shell does not act accordingly.

PureWeen:
> My application is in Arabic , it has to display Right To Left regardless of the settings in the device (force right to left or not)

Right! And you should be able to lean on our APIs for this. If not then we'll fix. For example setting on Shell or just on the Pages

> Setting FlowDirection=RightToLeft for Shell does not act accordingly.
> Setting Shell.FlowDirection=RightToLeft has no effect

Do these bugs outline your issues? 
 https://github.com/xamarin/Xamarin.Forms/issues/6342
https://github.com/xamarin/Xamarin.Forms/issues/6227

If those don't fully capture your shell issues can you log another one please?

> After this change, FlowDirection will only work if Manifest was changed to android:supportsRtl="true"

Correct we use native APIs to handle RTL (LayoutDirection) and those only work if you have this set to true.  

> I tested this change on my local computer and it resolved my problem

Good to hear! We'll hopefully get to the rest of the Shell RTL issues here soon

> Keeping the lines as it is (without the PR) , will show CollectionView left to right when I set Window.DecorView.LayoutDirection = LayoutDirection.Rtl

Which is caused by this issue https://github.com/xamarin/Xamarin.Forms/issues/7920. Using ReverseLayout would just be the wrong way to address the bug

Elashi:
@PureWeen Thank you for your time.  And I understand your point of view.  Let me know if I have to do anything else to get this issue resolved.

PR #7811, opened 2019-10-04
[iOS] Alternative proposal for PR 7794

### LINKED ISSUE #7251: [Bug] iOS - CollectionView EmptyView has incorrect bounds, height, width

adrianknight89:
#7794 attempts to use the `CollectionView` frame for properly laying out the empty view. However, the comment in the same block of code mentions `_emptyUIView` is measured properly, which is not the case. 

I have noticed that when `ItemsViewController.UpdateEmptyView()` is called in `ItemsViewRenderer`, `CollectionView` is sized to have the entire screen dimensions. A second call to `UpdateEmptyView()` which happens in ItemsViewController -> NumberOfSections -> CheckForEmptySource uses the correct dimensions.

I moved `UpdateEmptyView()` call into `ViewDidLoad` which fixed the issue.

fixes #7251 and #7758

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7758.xaml.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7758.xaml.cs
@@ -1,0 +1,22 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+namespace Xamarin.Forms.Controls.Issues
+{
+    [Preserve(AllMembers = true)]
+    [Issue(IssueTracker.Github, 7758, "[iOS] EmptyView is not rendered in screen center", PlatformAffected.iOS)]
+    public partial class Issue7758 : TestContentPage
+    {
+        public Issue7758()
+        {
+#if APP
+			InitializeComponent();
+#endif
+        }
+
+        protected override void Init()
+        {
+
+        }
+    }
+}
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7758.xaml.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7758.xaml.cs
@@ -10,11 +10,13 @@ namespace Xamarin.Forms.Controls.Issues
         public Issue7758()
         {
 #if APP
+			Device.SetFlags(new List<string> { CollectionView.CollectionViewExperimental });
+
 			InitializeComponent();
 #endif
-        }
+		}
 
-        protected override void Init()
+		protected override void Init()
         {
 
         }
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7758.xaml.cs
```diff
@@ -0,0 +1,25 @@
...
+        public Issue7758()
+        {
+#if APP
+			Device.SetFlags(new List<string> { CollectionView.CollectionViewExperimental });
+
+			InitializeComponent();
+#endif
+		}
+
+		protected override void Init()
```
rmarinho: Can you fix the code formatting please?

adrianknight89: Should be fixed. I think the formatting errors were introduced with copy paste from the other test.

adrianknight89:
@rmarinho I'm currently looking at device rotation. If we change the device orientation, EmptyView will not be aligned properly. I think we might want to move the fix here to `ViewWillLayoutSubviews` as @jsuarezruiz suggested or write more code to handle rotation. Give me some time.

jsuarezruiz:
`ViewWillLayoutSubviews` will be invoked when the device is rotated. If you add the UpdateEmptyView outside of the `if (!_initialConstraintsSet)` validation, will works but will do the same more times than necessary. So, my suggestion is call UpdateEmptyView inside the `if (!_initialConstraintsSet)` validation to initialize the EmptyView (if exists). Then, inside `ViewWillLayoutSubviews` validate if `_emptyUIView` is not null and if the Frame is different from the UICollectionView Frame (can use `if (IsHorizontal)` to compare Width or Height depending on the orientation) to call a new method called for example, UpdateEmptyViewLayout where the only thing to do, is `_emptyViewFormsElement.Layout(_emptyUIView.Frame.ToRectangle());`. 

Will work nice!.
Captura de pantalla 2019-10-04 a las 17 28 09

adrianknight89:
I did all those things locally. I'm researching something else now. If you set the first black grid to be auto sized and add a large text (multiple sentences), then the CollectionView and EmptyView are not sized properly even after the suggested changes.

adrianknight89:
@rmarinho @jsuarezruiz This should be good to go. The stuff I mentioned in the previous comment had to do with the way the test was setup, so I fixed the test with better instructions. Please test and let me know if it needs changes. :)

PR #7840, opened 2019-10-06
Fix double refresh issue with RefreshView

### LINKED ISSUE #7803: [Bug] RefreshView pull to refresh command firing twice on a single pull
### Description

In 4.3-pre3, it looks like RefreshView Command is firing twice on a pull gesture. This occurs on both iOS and Android. The method in my view model that's executed when the command fires is called twice.

This was working just fine when I tried it in a CollectionView (4.3-pre1), and I'm pretty certain it was fine when I tried it with a CarouselView in 4.3-pre2.

### Steps to Reproduce

1. Run the attached sample.
2. Browse to the pull to fresh page.
3. Pull to refresh.

### Expected Behavior

10 items changes to 20 items.

### Actual Behavior

10 items changes to 30 items.

### Basic Information

- Version with issue: 4.3-pre3
- Last known good version: I'm pretty certain this was working fine in 4.3-pre2
- IDE: VSMac 8.3.1
- Platform Target Frameworks: 
  - iOS: 13.0
  - Android: API 28 

### Screenshots

Prior to the refresh (10 items):

Simulator Screen Shot - iPhone X - 2019-10-03 at 16 24 00

After the refresh (30 items - should be 20):

Simulator Screen Shot - iPhone X - 2019-10-03 at 16 24 04

### Reproduction Link

[CarouselViewDemos.zip](https://github.com/xamarin/Xamarin.Forms/files/3686826/CarouselViewDemos.zip)

adrianknight89:
### Description of Change ###

If you refresh a RefreshView and add a time delay beforehand (through Task.Delay or network latency), then previously we ended up calling the refresh command twice. This PR fixes the issue. Also added a UI test.

### Issues Resolved ### 

- fixes #7803

### Testing Procedure ###
Pull to refresh and make sure that the count shows 20 instead of 30.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7803.xaml
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7803.xaml
@@ -1,0 +1,35 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<controls:TestContentPage
+    xmlns="http://xamarin.com/schemas/2014/forms"
+    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+    xmlns:controls="clr-namespace:Xamarin.Forms.Controls"
+    x:Class="Xamarin.Forms.Controls.Issues.Issue7803">
+    
+    <Grid>
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto"/>
+            <RowDefinition/>
+        </Grid.RowDefinitions>
+
+        <StackLayout Grid.Row="0" Orientation="Vertical" Spacing="5">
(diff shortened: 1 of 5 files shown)


adrianknight89:
@PureWeen Can you review again? I made some changes to the property setter since it was bloated with several lines of code.

> So in theory IsRefresh should never be set to true if in fact the Commands can execute is false but it still feels odd to let the Refresh trigger if CanExecute is false.

I'd agree with you, but the user can manually set `IsRefreshing` to true when the command is not executable. So why not allow this to be the same? I don't see any validation on `IsRefreshingProperty`  based on the command:

```
public static readonly BindableProperty IsRefreshingProperty =
BindableProperty.Create(nameof(IsRefreshing), typeof(bool), typeof(RefreshView), false, BindingMode.TwoWay);
```

PR #7847, opened 2019-10-07
Add IsVisible property to ShellItem/ShellSection/ShellContent

### LINKED ISSUE #7230: [Enhancement] Shell allow dynamic load FlyoutItem!
## Summary
Shell does not allow dynamic loading of menu options, according to a user profile, it must allow to show or hide options according to the login profile

## API changes

To facilitate the dynamic menu (FlyoutItem) loading, you can add a Visible property in a ShellSection through Binging.
And thus be able to invite from ViewModel the status of this option indicating whether the user is shown or not. (user profiles according to access)
Another send a List <Menus> indicating whether or not it will be displayed to the user.

## Intended use case
Well I am using Shell and I have this problem since the App that I am developing must load the options of the menus, according to a login where I bring the list of the options that this user has, and if you do not have permissions to some options you should not show you, in the side menu.

PureWeen:
### Description of Change ###
Add IsVisible property to BaseShellItem
- If all children items aren't visible then the parent will also not be visible. 
- Retrieving Items is now exposed through IShell<type>Controller instead of accessing the Items directly.
- If ShellContent isn't visible you can still navigate to it manually which lets you hide things from the flyout

### Issues Resolved ### 
- fixes #7230

### API Changes ###
Added:
 - EventHandler IShellContentController. IsPageVisibleChanged
 - EventHandler IShellController. ItemsCollectionChanged
 - ReadOnlyCollection<ShellItem> IShellController.GetItems()
- ReadOnlyCollection<ShellSection> IShellItemController.GetItems();
- event IShellItemController.NotifyCollectionChangedEventHandler ItemsCollectionChanged;
- ReadOnlyCollection<ShellContent> IShellSectionController.GetItems();
- event NotifyCollectionChangedEventHandler IShellSectionController.ItemsCollectionChanged;

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS
- Android
- UWP

### Testing Procedure ###
- Automated Tests included
- Interact with added test to make sure it works as you expect it to

### PR Checklist ###


PureWeen:
UI Tests are passing. iOS 13 and Android 19 failures are unrelated

jfversluis:
> ##[error]Xamarin.Forms.Platform.UAP\ListViewRenderer.cs(780,71): Error CS1061: 'Type' does not contain a definition for 'IsValueType' and no accessible extension method 'IsValueType' accepting a first argument of type 'Type' could be found (are you missing a using directive or an assembly reference?)

jamesmcmenamin:
@PureWeen @samhouts Any chance on this getting into Pre3

PureWeen:
@jamesmcmenamin Unfortunately no 😢 

The last 4.4 is closed for API changes so this one will be part of 4.5

jamesmcmenamin:
Well shoo Thanks for the update @PureWeen

Xamarin.Forms.Core/Shell/BaseShellItem.cs
```diff
@@ -33,6 +33,9 @@ public class BaseShellItem : NavigableElement, IPropertyPropagationController, I
 		public static readonly BindableProperty IsEnabledProperty =
 			BindableProperty.Create(nameof(IsEnabled), typeof(bool), typeof(BaseShellItem), true, BindingMode.OneWay);
 
+		public static readonly BindableProperty IsVisibleProperty =
+			BindableProperty.Create(nameof(IsVisible), typeof(bool), typeof(BaseShellItem), true, BindingMode.OneWay);
```
StephaneDelcroix: OneWay == Default

Xamarin.Forms.Core/Shell/IShellContentController.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace Xamarin.Forms
+﻿using System;
+
+namespace Xamarin.Forms
 {
 	public interface IShellContentController : IElementController
```
StephaneDelcroix: should be hidden from intelisense

PureWeen:
Maybe IsIncluded?

We are trying to get away from associating Shell<Elements> with visual concepts so something like that might work.

Though with the current implementation I think you could say that it does change the visibility of the item itself. Setting IsVisible is logically equivalent to completely removing the item from the Items Collection. In addition to not showing up as a tab/flyout Item you can't navigate to it. If you're on a Shell<element> and IsVisible gets set to false then you are navigated away just like if someone had removed the item. It becomes logically unreachable and nonexistent. It's basically not visible to the platform layer.

Like if you have
```xaml
<ShellItem FlyoutBehavioar=Disabled Route="Something"
```

Then that item won't show up in the menu but you can get to it via
```C#
Goto("//Something")
```
If you have
```xaml
<ShellItem FlyoutBehavioar=Disabled IsVisible=False Route="Something"
```

```C#
Goto("//Something") -> Will fail and say route not found
```
so it's more like it sends the Item into oblivion
It's meant to be a replacement for having to call Items.Clear or remove shell items at all when doing things like login flow and such

Xamarin.Forms.Core/Shell/IShellContentController.cs
rmarinho: This is not new , so why would we hide now?

Xamarin.Forms.Core/Shell/ShellItemCollection.cs
```diff
@@ -14,7 +30,103 @@ internal sealed class ShellItemCollection : IList<ShellItem>, INotifyCollectionC
...
+		internal IList<ShellItem> Inner
+		{
+			get
+			{
+				return _inner;
+			}
+			set
+			{			
+				_inner = value;
+				((INotifyCollectionChanged)_inner).CollectionChanged += InnerCollectionChanged;
```
rmarinho: Shouldn't we unsubscribe from this? On dispose or if we pass a new "inner" ?

Xamarin.Forms.Platform.UAP/Shell/ShellRenderer.cs
```diff
@@ -160,6 +160,12 @@ protected virtual void OnElementSet(Shell shell)
 			IsPaneOpen = Shell.FlyoutIsPresented;
 			((IShellController)Element).AddFlyoutBehaviorObserver(this);
 			((IShellController)shell).AddAppearanceObserver(this, shell);
+			(shell as IShellController).ItemsCollectionChanged += OnItemsCollectionChanged;
```
rmarinho: Do we need to unsubscribe this?

Xamarin.Forms.Core/Shell/Shell.cs
```diff
@@ -314,7 +317,24 @@ async Task IShellController.OnFlyoutItemSelectedAsync(Element element)
...
+				shellSection.PropertyChanged += OnShellItemPropertyChanged;
+			else
+				await GoToAsync(state).ConfigureAwait(false);
+		}
+
+		void OnShellItemPropertyChanged(object sender, PropertyChangedEventArgs e)
+		{
+			if (e.PropertyName == CurrentItemProperty.PropertyName)
+			{
+				(sender as BindableObject).PropertyChanged -= OnShellItemPropertyChanged;
```
rmarinho: Why do we unsubscribe here? Not understanding very well, can we go back and if i click again since we unsubscribed it will not navigate again

PureWeen: The only point of this code is to delay navigation until a visible stack of Shell Elements are realized so it's only relevant as the navigation is happening and the visible items are showing up.

The code gets here
https://github.com/xamarin/Xamarin.Forms/pull/7847/files/528247083f733948659463b1a2b9ea6e3089f647#diff-070ffbecea113674cafb32e2f6776e08R321

It sees that CurrentItem is null so I subscribe to an event until there's a valid ShellItem.ShellSection.ShellContent

Then it navigates there. So this is only relevant as it's navigating not to initiate navigation

Xamarin.Forms.Core/Shell/ShellItemCollection.cs
PureWeen: This list should never change. It's a readonly BP
```C#
static readonly BindablePropertyKey ItemsPropertyKey = BindableProperty.CreateReadOnly(nameof(Items), typeof(ShellItemCollection), 
```

Xamarin.Forms.Platform.UAP/Shell/ShellRenderer.cs
PureWeen: 👍

PR #7866, opened 2019-10-08
Make the Refresh View Respect Command CanExecute and add Refreshing event

### LINKED ISSUE #7803: [Bug] RefreshView pull to refresh command firing twice on a single pull

PureWeen:
### Description of Change ###
- The check for disabling the RefreshView based on a command was only firing when CommandParameter was set. This PR aligns Refresh View with how Button logic works with IsEnabled
- Add IsRefreshing event to RefreshView
- Clamp IsRefreshing to false if IsEnabled is false
- I moved the Command execution up into the Cross Platform code (much like the ListView Refresh) because it doesn't seem to make sense to have that code spread out between the renderers. This also fixed #7803 

### Issues Resolved ### 

- fixes #7803

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS
- Android
- UWP

### API Changes ###
Added: 
 - RefreshView.IsRefreshing

### Testing Procedure ###
- Tests are Automated
- Added test buttons to RefreshViewTests. Click around and make sure it all works as you'd expect it to

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7803.xaml.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7803.xaml.cs
@@ -1,0 +1,171 @@
+﻿using System.Collections.ObjectModel;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using Xamarin.Forms.Xaml;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using System.ComponentModel;
+using System;
+
+#if UITEST
+using Xamarin.UITest;
+using Xamarin.UITest.Queries;
+using NUnit.Framework;
+using Xamarin.Forms.Core.UITests;
+using System.Linq;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if UITEST
+	[NUnit.Framework.Category(UITestCategories.CollectionView)]
+#endif
+#if APP
+	[XamlCompilation(XamlCompilationOptions.Compile)]
+#endif
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 7803, "[Bug] CarouselView/RefreshView pull to refresh command firing twice on a single pull", PlatformAffected.All)]
+	public partial class Issue7803 : TestContentPage
+	{
+#if APP
+		public Issue7803()
+		{
+			Device.SetFlags(new List<string> { CollectionView.CollectionViewExperimental });
+
+			InitializeComponent();
+
+			BindingContext = new ViewModel7803();
+		}
+#endif
+
+		protected override void Init()
+		{
+
+		}
+
+#if UITEST
+		[Test]
+		public void DelayedIsRefreshingAndCommandTest_SwipeDown()
+		{
+			var collectionView = RunningApp.WaitForElement(q => q.Marked("CollectionView7803"))[0];
+
+			RunningApp.Pan(new Drag(collectionView.Rect, Drag.Direction.TopToBottom, Drag.DragLength.Medium));
+
+			RunningApp.WaitForElement(q => q.Marked("Count: 20"));
+			RunningApp.WaitForNoElement(q => q.Marked("Count: 30"));
+
+			AppResult[] lastCellResults = null;
+
+			RunningApp.QueryUntilPresent(() =>
+			{
+				RunningApp.DragCoordinates(collectionView.Rect.CenterX, collectionView.Rect.Y + collectionView.Rect.Height - 50, collectionView.Rect.CenterX, collectionView.Rect.Y + 5);
+
+				lastCellResults = RunningApp.Query("19");
+
+				return lastCellResults;
+			}, 10, 1);
+
+			Assert.IsTrue(lastCellResults?.Any() ?? false);
+		}
+#endif
+	}
+
+	[Preserve(AllMembers = true)]
+	public class ViewModel7803 : INotifyPropertyChanged
+	{
+		public ObservableCollection<Model7803> Items { get; set; } = new ObservableCollection<Model7803>();
+
+		private bool _isRefreshing;
+
+		public bool IsRefreshing
+		{
+			get
+			{
+				return _isRefreshing;
+			}
+			set
+			{
+				_isRefreshing = value;
+
+				OnPropertyChanged("IsRefreshing");
+			}
+		}
+
+		private string _text;
+
+		public string Text
+		{
+			get
+			{
+				return _text;
+			}
+			set
+			{
+				_text = value;
+
+				OnPropertyChanged("Text");
+			}
+		}
+
+		public Command RefreshCommand { get; set; }
+
+		public ViewModel7803()
+		{
+			PopulateItems();
+
+			RefreshCommand = new Command(async () =>
+			{
+				IsRefreshing = true;
+
+				await Task.Delay(2000);
+				PopulateItems();
+
+				IsRefreshing = false;
+			});
+		}
+
+		void PopulateItems()
+		{
+			var count = Items.Count;
+
+			for (var i = count; i < count + 10; i++)
+				Items.Add(new Model7803() { Position = i });
+
+			Text = "Count: " + Items.Count;
+		}
+
+		public event PropertyChangedEventHandler PropertyChanged;
+
+		protected void OnPropertyChanged(string name)
+		{
+			PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(name));
+		}
+	}
+
+	[Preserve(AllMembers = true)]
+	public class Model7803 : INotifyPropertyChanged
+	{
+		private int _position;
+
+		public int Position
+		{
+			get
+			{
+				return _position;
+			}
+			set
+			{
+				_position = value;
+
+				OnPropertyChanged("Position");
+			}
+		}
+
+		public event PropertyChangedEventHandler PropertyChanged;
+
+		protected void OnPropertyChanged(string name)
+		{
+			PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(name));
+		}
+	}
+}
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7803.xaml.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7803.xaml.cs
@@ -30,7 +30,6 @@ namespace Xamarin.Forms.Controls.Issues
 #if APP
 		public Issue7803()
 		{
-			Device.SetFlags(new List<string> { CollectionView.CollectionViewExperimental });
 
 			InitializeComponent();
 
--- a/Xamarin.Forms.Core/RefreshView.cs
+++ b/Xamarin.Forms.Core/RefreshView.cs
@@ -29,7 +29,21 @@ namespace Xamarin.Forms
 		}
 
 		public static readonly BindableProperty CommandProperty =
-			BindableProperty.Create(nameof(Command), typeof(ICommand), typeof(RefreshView));
+			BindableProperty.Create(nameof(Command), typeof(ICommand), typeof(RefreshView), propertyChanged: OnCommandChanged);
+
+		static void OnCommandChanged(BindableObject bindable, object oldValue, object newValue)
+		{
+			if (bindable is RefreshView refreshView)
+			{
+				if (oldValue is ICommand oldCommand)
+					oldCommand.CanExecuteChanged -= refreshView.RefreshCommandCanExecuteChanged;
+
+				if (newValue is ICommand newCommand)
+					newCommand.CanExecuteChanged += refreshView.RefreshCommandCanExecuteChanged;
+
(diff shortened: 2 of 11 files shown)


Xamarin.Forms.Core/RefreshView.cs
```diff
@@ -29,7 +29,21 @@ public bool IsRefreshing
 		}
 
 		public static readonly BindableProperty CommandProperty =
-			BindableProperty.Create(nameof(Command), typeof(ICommand), typeof(RefreshView));
+			BindableProperty.Create(nameof(Command), typeof(ICommand), typeof(RefreshView), propertyChanged: OnCommandChanged);
+
+		static void OnCommandChanged(BindableObject bindable, object oldValue, object newValue)
+		{
+			if (bindable is RefreshView refreshView)
```
adrianknight89: Minor suggestion: reduce nesting here

adrianknight89:
@PureWeen Correct me if I'm wrong, but here you're disabling `RefreshView` if its command can no longer execute, right? `ListView` has a `Refreshing` event. Why doesn't `RefreshView` have it too (assuming not everybody wants to use the MVVM pattern)? 

If it did have this event in the future, there won't be a way to process the event and the command independently of each other (e.g. disable the command / business logic in certain circumstances while enabling the event / UI stuff) if we set `IsEnabled` to false. So I'm not sure how I feel about this change. You said you're aligning with `Button`. I wonder how that's working today with `Clicked` & `Command`.

jsuarezruiz:
Reviewing and testing the changes, my main question here is should we allow to set IsRefreshing to true (and get the visualization) without the command execution when the CanExecute from the Command is false?.
refreshview-disabled-small

PureWeen:
@adrianknight89 

If I'm understanding this comment

> You said you're aligning with Button. I wonder how that's working today with Clicked & Command.

With our implementation of ICommand if you call Execute it doesn't check CanExecute. With the SendClicked behavior on Button it will only execute the command if the button is enabled

PureWeen:
@adrianknight89 

> If it did have this event in the future, there won't be a way to process the event and the command independently of each other (e.g. disable the command / business logic in certain circumstances while enabling the event / UI stuff) if we set IsEnabled to false.

Yea I keep waffling :-) 

> (e.g. disable the command / business logic in certain circumstances while enabling the event / UI stuff)

I'm thinking through the use cases for this

While also keeping @jsuarezruiz 's comment in mind here

https://github.com/xamarin/Xamarin.Forms/pull/7866#issuecomment-539596818

I could see it being confusing to allow the visualization but not the command.  I think if you're going to use a command use the command and if you use an event use an event. It's like the scenario with IsEnabled on Button where people set it from the View and then the Command replaces whatever you set it to. 

If you're using MVVM the MVVM is going to win out over the non MVVM APIs

> ListView has a Refreshing event. Why doesn't RefreshView have it too (assuming not everybody wants to use the MVVM pattern)?

It should have a Refreshing event

adrianknight89:
@PureWeen 

> With the SendClicked behavior on Button it will only execute the command if the button is enabled

https://github.com/xamarin/Xamarin.Forms/blob/3c3ecfd45489f119edfb922e01c0514a0bb74e2a/Xamarin.Forms.Core/ButtonElement.cs#L56-L63

Why doesn't this also check CanExecute before line 60?

adrianknight89:
> I think if you're going to use a command use the command and if you use an event use an event

You mean we shouldn't be using both at the same time? I could see how someone might end up sending a message through the command to the view (which could have been dealt with in the event handler instead). Sorry, thinking of possibilities here. I'm sure all of this could be worked around.

Xamarin.Forms.Core/RefreshView.cs
```diff
@@ -53,7 +67,9 @@ public object CommandParameter
 		void RefreshCommandCanExecuteChanged(object sender, EventArgs eventArgs)
 		{
 			if (Command != null)
-				IsEnabled = Command.CanExecute(CommandParameter);
+				SetValueCore(IsEnabledProperty, Command.CanExecute(CommandParameter));
+			else
+				SetValueCore(IsEnabledProperty, true);
```
adrianknight89: How would the below case work with this change?

```
var refreshView = new RefreshView();
refreshView.IsEnabled = false;
refreshView.Command = new Command(() => {});
```

`RefreshCommandCanExecuteChanged` gets executed and sets `IsEnabled` to true even though I explicitly set it to false.

PureWeen: Yea that's an ongoing conversation with Commanding and IsEnabled
https://github.com/xamarin/Xamarin.Forms/issues/2758

So the way it's all done here is just consistent with how it works with Button

```C#
		public static void CommandChanged(IButtonElement sender)
		{
			if (sender.Command != null)
			{
				CommandCanExecuteChanged(sender, EventArgs.Empty);
			}
			else
			{
				sender.IsEnabledCore = true;
			}
		}
```

PureWeen:
@adrianknight89 
> Why doesn't this also check CanExecute before line 60?

I can't say for sure. I think it's as unexciting as when that code was initially written that was the decision made for the rules of execution.

PureWeen:
Alright @jsuarezruiz @hartez @adrianknight89 

This one should be good to review. A few quick notes

- I added an IsRefreshing event
- @adrianknight89 we all talked internally and we don't think you should be able to set IsRefreshing to true if IsEnabled is false
- I moved all the code that fires the command from the native code and just put it into the cross platform code because the command seems like it should just fire from the cross platform code when IsRefreshing transitions to true. I'm not sure if there's an advantage or specific reason for it to be in the native code?  Doing that also fixed #7803 so I copied your UI test from that PR you did over to this one (hope that's alright)

Xamarin.Forms.Core/RefreshView.cs
```diff
@@ -20,7 +22,41 @@ public RefreshView()
...
-			BindableProperty.Create(nameof(IsRefreshing), typeof(bool), typeof(RefreshView), false, BindingMode.TwoWay);
+			BindableProperty.Create(nameof(IsRefreshing), typeof(bool), typeof(RefreshView), false, BindingMode.TwoWay, coerceValue:OnCoerceIsRefreshing, propertyChanged: OnIsRefreshingPropertyChanged);
+
+		static void OnIsRefreshingPropertyChanged(BindableObject bindable, object oldValue, object newValue)
+		{
+			bool value = (bool)newValue;
+
+			if (value)
+			{
+				var resfrehView = ((RefreshView)bindable);
```
adrianknight89: fix variable name

Xamarin.Forms.Core/RefreshView.cs
```diff
@@ -20,7 +22,41 @@ public RefreshView()
...
+				resfrehView.Refreshing?.Invoke(bindable, EventArgs.Empty);
+				if(resfrehView.Command != null)
+					resfrehView.Command.Execute(resfrehView.CommandParameter);
```
adrianknight89: so not checking if the command can execute before executing it then?

Xamarin.Forms.Core/RefreshView.cs
```diff
@@ -20,7 +22,41 @@ public RefreshView()
...
+			}
+		}
+
+		static object OnCoerceIsRefreshing(BindableObject bindable, object value)
```
adrianknight89: Should this be renamed to `OnIsRefreshingPropertyCoerced` to align with the previous PC handler?

Xamarin.Forms.Core/RefreshView.cs
```diff
@@ -20,7 +22,41 @@ public RefreshView()
...
+				return value;
+
+			if (!view.IsEnabled)
+				return false;
+
+			if (view.Command == null)
+				return value;
+
+			if (!view.Command.CanExecute(view.CommandParameter))
+				return false;
```
adrianknight89: Can we group all these if checks together? There's so many of them. Group !newValue and view.Command together. Group !view.IsEnabled and !view.Command.CanExecute() together? Not sure if it'd be more difficult to read this way. Just a thought.

adrianknight89:
@PureWeen Yeah, that's alright. I closed the other PR.

Xamarin.Forms.Core/RefreshView.cs
PureWeen: No. My thinking here is that if IsRefreshing transitions to true for any reason then I want these events to all fire.  If these fire when they shouldn't then I would say that's an upstream bug.

Xamarin.Forms.Core/RefreshView.cs
PureWeen: hehe
I'll leave it for now unless it gets another down vote.  Helps me think through it better having it split out. Especially knowing how slow future Shane is.

slang25:
I've just tried this in my app and on iOS I am getting a `UIKitThreadAccessException` from the `OnPropertyChanged` method. If I get time later I'll try to recreate it in the controls app.

Xamarin.Forms.Core/RefreshView.cs
```diff
@@ -77,5 +127,13 @@ protected override SizeRequest OnMeasure(double widthConstraint, double heightCo
 		{
 			return _platformConfigurationRegistry.Value.On<T>();
 		}
+
+		protected override void OnPropertyChanged([CallerMemberName] string propertyName = null)
+		{
+			base.OnPropertyChanged(propertyName);
+			if (IsEnabledProperty.PropertyName == propertyName)
```
adrianknight89: Isn't it easier to read if we did `propertyName == IsEnabledProperty.PropertyName`?

adrianknight89:
@slang25 Sounds like something needs to be executed on the main thread. A repro would be very helpful.

Xamarin.Forms.Core/RefreshView.cs
```diff
@@ -20,7 +22,41 @@ public RefreshView()
...
+
+			var refreshView = ((RefreshView)bindable);
+			refreshView.Refreshing?.Invoke(bindable, EventArgs.Empty);
+			if (refreshView.Command != null)
+				refreshView.Command.Execute(refreshView.CommandParameter);
+		}
+
+		static object OnIsRefreshingPropertyCoerced(BindableObject bindable, object value)
+		{
+			RefreshView view = (RefreshView)bindable;
```
adrianknight89: could use `var` here

adrianknight89:
@PureWeen Can you rebase by any chance?

adrianknight89:
@PureWeen I just tested this in my app on Android. Pull to refresh is working, but there seems to be an issue with invoking the refresh command programmatically. The state of `IsRefreshing` seems different inside the refresh command:

| Refresh type  | IsRefreshing |
| Pull | True  |
| Programmatic  | False  |

Is this the intended behavior or should we expect `IsRefreshing` to be the same regardless of whether the refresh is happening via a gesture or code?

Here's the important bit: If I set `IsRefreshing` to true inside the command, then double refresh happens if it was false to begin with.

PureWeen:
@adrianknight89 if I'm understanding you correctly are you saying if you do

```C#
RefreshView.Command.Execute(null)
```

That it doesn't set IsRefreshing to true? 

I don't think executing the command manually like that should affect the value of IsRefreshing. The command execution is a side effect of the Refresh view triggering and we don't really have a way to know someone has invoked the command from code.

adrianknight89:
@PureWeen Yep, as long as that's the intended behavior, we're good.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7803.xaml.cs
```diff
@@ -0,0 +1,171 @@
...
+	[XamlCompilation(XamlCompilationOptions.Compile)]
+#endif
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 7803, "[Bug] CarouselView/RefreshView pull to refresh command firing twice on a single pull", PlatformAffected.All)]
+	public partial class Issue7803 : TestContentPage
+	{
+#if APP
+		public Issue7803()
+		{
+			Device.SetFlags(new List<string> { CollectionView.CollectionViewExperimental });
```
hartez: You can drop this now.

PureWeen:
```suggestion
```

PR #7880, opened 2019-10-08
Cleanup material controls character spacing interactions

PureWeen:
### Description of Change ###

Character spacing PR started using the title for the placeholder on picker which we originally weren't going to do but at this point I think it's fine. It's what users will expect if they switch to material

Unfortunately (AFAICT) character spacing doesn't work with place holders on the material controls. I poked iOS a bit and couldn't get it to work. I'm hoping once we update material components on android/ios then maybe it'll work?

- Since we're now using Title I replaced the APIs needed for TitleColor
- I added an override for *UpdateAttributedPlaceHolder* on iOS so that the Material implementations can control this field. It's confusing having the non material control set this. If AttributedPlaceHolder start working for material than the Material control can be updated to handle it

### API Changes ###
Added: iOS only
 - EntryRendererBase.UpdateAttributedPlaceHolder(NSAttributedString nSAttributedString)
 - PickerRendererBase.UpdateAttributedPlaceHolder(NSAttributedString nSAttributedString)

### Platforms Affected ### 
- iOS
- Android

### Testing Procedure ###
- make sure character spacing still works and that picker obeys title color for placeholder color

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Material.iOS/MaterialEntryRenderer.cs
+++ b/Xamarin.Forms.Material.iOS/MaterialEntryRenderer.cs
@@ -1,4 +1,5 @@
 ﻿using CoreGraphics;
+using Foundation;
 using UIKit;
 using Xamarin.Forms.Platform.iOS;
 
--- a/Xamarin.Forms.Material.iOS/MaterialEntryRenderer.cs
+++ b/Xamarin.Forms.Material.iOS/MaterialEntryRenderer.cs
@@ -12,9 +12,9 @@ namespace Xamarin.Forms.Material.iOS
 		protected override void UpdateColor() => Control?.UpdateTextColor(this);
 		protected virtual void ApplyTheme() => Control?.ApplyTheme(this);
 		protected override void UpdatePlaceholder() => Control?.UpdatePlaceholder(this);
-		protected override void UpdateAttributedPlaceHolder(NSAttributedString nSAttributedString)
+		protected override void UpdateAttributedPlaceholder(NSAttributedString nsAttributedString)
 		{
-			// Attributed place holders don't currently work with Material Text Fields
+			// AttributedPlaceholder doesn't currently work with Material
 			// once/if it does start working it will be handled inside MaterialTextManager	
 		}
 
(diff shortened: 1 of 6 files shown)


Xamarin.Forms.Material.iOS/MaterialEntryRenderer.cs
```diff
@@ -11,6 +12,11 @@ public class MaterialEntryRenderer : EntryRendererBase<MaterialTextField>, IMate
 		protected override void UpdateColor() => Control?.UpdateTextColor(this);
 		protected virtual void ApplyTheme() => Control?.ApplyTheme(this);
 		protected override void UpdatePlaceholder() => Control?.UpdatePlaceholder(this);
+		protected override void UpdateAttributedPlaceHolder(NSAttributedString nSAttributedString)
```
adrianknight89: The casing in `nSAttributedString` seems a bit odd. Can we try `nsAttributedString`? here and in other places?

Xamarin.Forms.Material.iOS/MaterialEntryRenderer.cs
```diff
@@ -11,6 +12,11 @@ public class MaterialEntryRenderer : EntryRendererBase<MaterialTextField>, IMate
...
+		{
+			// Attributed place holders don't currently work with Material Text Fields
+			// once/if it does start working it will be handled inside MaterialTextManager
```
adrianknight89: once/if _they do_ start working _they_ will be handled inside MaterialTextManager

Xamarin.Forms.Material.iOS/MaterialPickerRenderer.cs
```diff
@@ -30,8 +37,8 @@ protected override void OnElementChanged(ElementChangedEventArgs<Picker> e)
 			}
 		}
 
-		string IMaterialEntryRenderer.Placeholder => string.Empty;
-		Color IMaterialEntryRenderer.PlaceholderColor => Color.Default;
+		string IMaterialEntryRenderer.Placeholder => Element?.Title;
+		Color IMaterialEntryRenderer.PlaceholderColor => Element.TitleColor;
```
adrianknight89: Do we need a `?` after `Element` similar to other examples?

PR #7891, opened 2019-10-09
Fix TemplatedItemsList incorrect grouped collection range removal

### LINKED ISSUE #7890: [Bug] TemplatedItemsList incorrect grouped collection range removal
### Description
When using grouped ListView, it handles ranged removals wrong.
It happens because of this line:
https://github.com/xamarin/Xamarin.Forms/blob/6ba52f79b51da48386df8f5395fee6b33cc3cff0/Xamarin.Forms.Core/TemplatedItemsList.cs#L813

Items are removed not sequentially, but through one.
### Steps to Reproduce
I'll submit PR several minutes later, you can check it using test

### Expected Behavior
Items are removed sequentially
### Actual Behavior
Items are removed through one
### Basic Information

- Version with issue: 
- Last known good version:
- IDE:
- Platform Target Frameworks: 
  - All
- Nuget Packages:
- Affected Devices:

melimion:
### Description of Change ###

Fixes ranged item removal for grouped ListView

### Issues Resolved ### 

- fixes #7890

### API Changes ###
 
 None

### Platforms Affected ### 

- Core(all platforms)

None

Not applicable

### PR Checklist ###


jfversluis:
Would you be able to add a UITest for this?

melimion:
@jfversluis 
Added UITest, but you should check it
Not sure it works, I have not much experience with it

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7890.cs
```diff
@@ -46,6 +59,22 @@ protected override void Init()
 			};
 		}
 
+#if UITEST
+		[Test]
+		public void RefreshingListViewCrashesWhenDisposedTest()
```
jfversluis: I think you copied a bit too much from where you got this ;)

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7890.cs
```diff
@@ -46,6 +59,22 @@ protected override void Init()
...
+		{
+			foreach (var c in Enumerable.Range(0, Count))
+			{
+				RunningApp.WaitForElement(q => q.Marked(c.ToString()));
+			}
+			RunningApp.Tap(q => q.Button("RemoveBtn"));
+			foreach (var c in Enumerable.Range(RemoveFrom, RemoveCount))
+			{
+				RunningApp.WaitForNoElement(q => q.Marked(c.ToString()));
+			}
```
jfversluis: What do you think this is doing?

I think you should just do a `WaitForElement` on the remove button, then tap that and then verify that the range that you expect to be deleted is actually gone through an Assert statement. I don't think like this the test has the desired effect.

Are you able to run the test locally?

melimion: > I think you copied a bit too much from where you got this ;)

Looks like that )

> Are you able to run the test locally?

After some research and fighting with adb, yes )

I've fixed UI test

melimion:
@rmarinho, I checked replace part, and it looks not affected.

When replacing, old item removed and then new item added within InsertGrouped method, so index is correct.

PR #7908, opened 2019-10-09
Ellipsize Picker text on Android

### LINKED ISSUE #5703: Android EditText Ellipsize not working
### Description
I have a picker with an item that has text too long for the picker control. I would like this to ellipse at the end. However when I set Control.Ellipsize = TruncateAt.End; in the renderer, the ellipse does not display.

Please find attached a test app.

### Steps to Reproduce

1. Select the first option in the picker (an item with text longer than the control)

### Expected Behavior
Ellipse is added at the end of the line in the picker control.

### Actual Behavior
The text is just cut off - no ellipse.

### Basic Information

- Version with issue: Xamarin 3.4.0.109*
- IDE: Visual Studio 2017 Pro
- Platform Target Frameworks: 
  - Android: Min Api 21, target 27

### Screenshots
image

### Test App
[App1.zip](https://github.com/xamarin/Xamarin.Forms/files/3012306/App1.zip)

jcmanke:
### Description of Change ###

Ellipsize the text in the Picker on Android if it is too long.

### Issues Resolved ### 

- fixes #5703 

### API Changes ###

Added:
 - public override void PickerEditText.SetText(ICharSequence text, BufferType type)
 - public override void MaterialPickerEditText.SetText(ICharSequence text, BufferType type)

### Platforms Affected ### 

- Android
- Android Material

Text is ellipsized.

Before:
Screenshot_1570648970

After:
Screenshot_1570649138

### Testing Procedure ###

Add a long string to the Items of a picker. Select that item.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Material.Android/MaterialPickerEditText.cs
+++ b/Xamarin.Forms.Material.Android/MaterialPickerEditText.cs
@@ -1,4 +1,8 @@
-﻿#if __ANDROID_28__
+﻿
+using Android.Support.V4.View;
+using Android.Text;
+using Java.Lang;
+#if __ANDROID_28__
 using System;
 using Android.Content;
 using Android.Graphics;
@@ -31,6 +35,24 @@ namespace Xamarin.Forms.Material.Android
 		{
 			base.OnFocusChanged(gainFocus, direction, previouslyFocusedRect);
 			PickerManager.OnFocusChanged(gainFocus, this, (IPopupTrigger)Parent.Parent);
+		}
+
+		public override void SetText(ICharSequence text, BufferType type)
+		{
+			if (ViewCompat.IsLaidOut(this) && text != null)
+			{
+				int textWidth = Width - CompoundPaddingLeft - CompoundPaddingRight;
+
+				string fullText = text.ToString();
+				string ellipsizedText = TextUtils.Ellipsize(fullText, Paint, textWidth, TextUtils.TruncateAt.End);
+
+				if (!string.IsNullOrEmpty(ellipsizedText))
+				{
+					text = new Java.Lang.String(ellipsizedText);
+				}
+			}
+
+			base.SetText(text, type);
 		}
 
 		protected override void Dispose(bool disposing)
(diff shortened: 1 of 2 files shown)


Xamarin.Forms.Material.Android/MaterialPickerEditText.cs
```diff
@@ -33,6 +37,24 @@ protected override void OnFocusChanged(bool gainFocus, [GeneratedEnum] FocusSear
 			PickerManager.OnFocusChanged(gainFocus, this, (IPopupTrigger)Parent.Parent);
 		}
 
+		public override void SetText(ICharSequence text, BufferType type)
+		{
+			if (ViewCompat.IsLaidOut(this) && text != null)
+			{
+				int textWidth = Width - CompoundPaddingLeft - CompoundPaddingRight;
+
+				string fullText = text.ToString();
```
jsuarezruiz: The same is done in this renderer and in the PickerEditText, we could create an extension method or similar.

jcmanke: Moved to an extension method in TextViewExtensions.

samhouts:
It looks like this will always ellipsize long text, whether we want it to or not. This seems to be technically a breaking change. Do we want to make this an option?

hartez:
We can't take the PR as it is, because it's a breaking change; worse, it'll make the behavior different across the platforms. 

If the PR were updated to be a Platform Specific for Android (something like `Picker.On<Android>().EllipsizeTextAtEnd()`, I'd be fine with that; we could extend it to other platforms later. 

I agree with Javier that if we want to do something like this, we should introduce an enum property (something like `TextTruncation`; `LineBreakMode` doesn't really work, since it includes wrap modes which don't apply to this control), and we should make it work on iOS/UWP.

hartez:
@jcmanke Did you consider the `Ellipsize` property of `EditText` for implementing this (rather than having to manually set the text)?

jcmanke:
@hartez

>Did you consider the Ellipsize property of EditText for implementing this (rather than having to manually set the text)?

IIRC just setting the Ellipsize property didn't work. I want to say it's something about Ellipsize being defined on the TextView class and not working for the EditText subclass.

PR #7910, opened 2019-10-09
Add IsReadOnly to Switch

### LINKED ISSUE #6953: Add property IsReadOnly to Switch control
Please add property IsReadOnly to Switch control

adammeaney:
### Description of Change ###

Add IsReadOnly property to the Switch.

This allows for focusing the control by keyboard navigation, as the Entry ReadOnly does, but not changing the value.

Both Android and UWP give a bit of visual when that happens, but never visibly toggle state.

### Issues Resolved ### 

- fixes #6953

### API Changes ###
Added:
 - bool Switch.IsReadOnly { get; set; } //Bindable Property

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

Default of IsReadOnly is false. No changes without setting the property.

When set, the Switch still appears enabled, but toggling does not occur.

Not applicable, looks the same as always

### Testing Procedure ###
Go the Switch Gallery page in the Control Gallery. Scroll to the IsReadOnly option.

Try and toggle the switch, push the Change Readonly button, try and toggle the switch.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Switch.cs
+++ b/Xamarin.Forms.Core/Switch.cs
@@ -17,6 +17,8 @@ namespace Xamarin.Forms
 
 		public static readonly BindableProperty ThumbColorProperty = BindableProperty.Create(nameof(ThumbColor), typeof(Color), typeof(Switch), Color.Default);
 
+		public static readonly BindableProperty IsReadOnlyProperty = BindableProperty.Create(nameof(IsReadOnly), typeof(bool), typeof(Switch), false);
+
 		public Color OnColor
 		{
 			get { return (Color)GetValue(OnColorProperty); }
@@ -27,6 +29,12 @@ namespace Xamarin.Forms
 		{
 			get { return (Color)GetValue(ThumbColorProperty); }
 			set { SetValue(ThumbColorProperty, value); }
+		}
+
+		public bool IsReadOnly
+		{
+			get { return (bool)GetValue(IsReadOnlyProperty); }
+			set { SetValue(IsReadOnlyProperty, value); }
 		}
 
 		readonly Lazy<PlatformConfigurationRegistry<Switch>> _platformConfigurationRegistry;
--- a/Xamarin.Forms.Controls/CoreGalleryPages/SwitchCoreGalleryPage.cs
+++ b/Xamarin.Forms.Controls/CoreGalleryPages/SwitchCoreGalleryPage.cs
@@ -43,9 +43,16 @@ namespace Xamarin.Forms.Controls
 			thumbColorContainer.ContainerLayout.Children.Add(changeThumbColorButton);
 			thumbColorContainer.ContainerLayout.Children.Add(clearThumbColorButton);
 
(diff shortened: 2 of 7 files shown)


Xamarin.Forms.Core/Switch.cs
```diff
@@ -17,6 +17,8 @@ public class Switch : View, IElementConfiguration<Switch>
 
 		public static readonly BindableProperty ThumbColorProperty = BindableProperty.Create(nameof(ThumbColor), typeof(Color), typeof(Switch), Color.Default);
 
+		public static readonly BindableProperty IsReadOnlyProperty = BindableProperty.Create(nameof(IsReadOnly), typeof(bool), typeof(Switch), false);
```
StephaneDelcroix: I'd prefer having the property named `ReadOnly`

adammeaney: > how is this different from `IsEnabled` ?

In visual, it is not toggleable but also does not have the disabled appearance.

The issue author and @PureWeen can probably describe their use cases for it better, but my reasoning has been that sometimes I want to show that a value is there, but not allow it to be toggled currently. 

Disabled look doesn't feel right in those scenarios, but an un-Switchable Switch with a different thumb color does.





> I'd prefer having the property named `ReadOnly`

I agree, but I just followed the naming conventions of IsVisible, IsEnabled, and the Entry IsReadOnly.

adammeaney:
Do you guys need anything else from me to get this going?

adammeaney:
I have an entry in my commit for that.

I am pretty sure I figured out how to work those gallery pages, let me know if I didn't get it right.

Xamarin.Forms.Core/Switch.cs
mattleibow: > does not have the disabled appearance

this might be visually confusing... "why is it not working"

mattleibow: > property named `ReadOnly`

Is this property used elsewhere? ie: for editor/entry? What is it called?

adammeaney: > > does not have the disabled appearance
> 
> this might be visually confusing... "why is it not working"

The use case I had in mind at least is a "locked" switch. It has a different thumb color, indicating is locked, but, doesn't look disabled so that it appears to still apply, even if I am not allowed to change the value.

I am not sure what other people were intending it for, but I have wanted it enough times that I wrote the PR when I saw it as an issue.


> > property named `ReadOnly`
> 
> Is this property used elsewhere? ie: for editor/entry? What is it called?

Entry/Editor appears to have gone with [IsReadOnly](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/InputView.cs#L17).

hartez:
@adammeaney Thank you for the effort, but we will not be taking this PR. 
For additional context, see the comments [here](https://github.com/xamarin/Xamarin.Forms/issues/6953#issuecomment-599791670).

PR #7979, opened 2019-10-13
[iOS] Fix iOS ScrollToAsync not completing when scrolling to current location

### LINKED ISSUE #7972: under iOS, await scrollView.ScrollToAsync(scrollView.ScrollX, scrollView.ScrollY, true) never completes
### Description
under iOS, `await scrollView.ScrollToAsync(scrollView.ScrollX, scrollView.ScrollY, true)` never completes and hangs indefinitely.
If the desired X and Y are already the same as the current scrollview X and Y, this never completes.
If the desired X or Y are different from the current scrollview X and Y, this completes normally.

Basically there is a bug if no animation is ever started because the scrollview is already at the given position.

### Steps to Reproduce

1. `await ScrollView.ScrollToAsync(scrollView.ScrollX, scrollView.ScrollY, true)`
2. the above line hangs forever under iOS.
3. 

### Expected Behavior
This is a corner case that needs to be handled gracefully (task need to complete).

### Actual Behavior
task never completes
### Basic Information

- Version with issue: 4.3.0.851321-pre3 
- IDE: VS 2019
- Platform Target Frameworks: I only tested this with iOS and VS 16.3.4

GalaxiaGuy:
### Description of Change ###

If `OnScrollToRequested` would call `SetContentOffset` with the current `ContentOffset`, call `SendScrollFinished` immediately (the same behavior as scrolling without animation).

### Issues Resolved ### 
- fixes #7972
### API Changes ###

 None

### Platforms Affected ### 
- iOS

None

Not applicable

### Testing Procedure ###

I added a label to `ScrollGallery` that says "scrolling" when scrolling starts and "completed" when it finished. Previously, when pressing the scroll button twice it would not change back to "completed".

**I also added a simple UITest testing this but haven't been able to actually run it.**

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/ScrollViewRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/ScrollViewRenderer.cs
@@ -260,8 +260,10 @@ namespace Xamarin.Forms.Platform.iOS
 				_requestedScroll = e;
 				return;
 			}
+
+			PointF newOffset = PointF.Empty;
 			if (e.Mode == ScrollToMode.Position)
-				SetContentOffset(new PointF((nfloat)e.ScrollX, (nfloat)e.ScrollY), e.ShouldAnimate);
+				newOffset = new PointF((nfloat)e.ScrollX, (nfloat)e.ScrollY);
 			else
 			{
 				var positionOnScroll = ScrollView.GetScrollPositionForElement(e.Element as VisualElement, e.Position);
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.iOS/Renderers/ScrollViewRenderer.cs
```diff
@@ -272,17 +274,20 @@ void OnScrollToRequested(object sender, ScrollToRequestedEventArgs e)
...
+						newOffset = new PointF(ContentOffset.X, (nfloat)positionOnScroll.Y);
 						break;
 					case ScrollOrientation.Both:
-						SetContentOffset(new PointF((nfloat)positionOnScroll.X, (nfloat)positionOnScroll.Y), e.ShouldAnimate);
+						newOffset = new PointF((nfloat)positionOnScroll.X, (nfloat)positionOnScroll.Y);
 						break;
 				}
 			}
-			if (!e.ShouldAnimate)
+			var sameOffset = newOffset == ContentOffset;
```
kingces95: `sameOffset` is only used once. Could inline it in the if statement.

PR #7986, opened 2019-10-14
Fix 4459 - Update UWP BoxViewRenderer to fully support corner radius

### LINKED ISSUE #4459: [UWP] BoxView CornerRadius doesn't work
### Description
When setting the corner radius, UWP doesn't render it.

### Steps to Reproduce
Run this:

https://github.com/davidortinau/TheLittleThingsPlayground/blob/master/TheLittleThingsPlayground/Views/ThreeTwoPage.xaml#L86

```
<BoxView FlexLayout.AlignSelf="Center"
                     BackgroundColor="#CC3300"
                     WidthRequest="200"
                     HeightRequest="100"
                     CornerRadius="10,0,0,10"
                     />
```

### Expected Behavior

Top left and lower right corners should be rounded.

### Actual Behavior

They are not rounded.

### Basic Information

### Screenshots

image

Android showing what is expected:

image

### Reproduction Link

https://github.com/davidortinau/TheLittleThingsPlayground/blob/master/TheLittleThingsPlayground/Views/ThreeTwoPage.xaml#L86

bmacombe:
### Fixes #4459  ###

Xamarin forms BoxView supports individual corner radius settings but the UWP BoxViewRenderer was using the Rectangle class as it's native control which does not.  This PR changes BoxViewRenderer to use the UWP Border class instead, which fully supports a unique radius for each corner.

In researching the various ways in UWP to support this, switching to border control seemed the simplest and least likely solution to cause regressions.  Other solutions I found was to us a clipping rectangle, but that seemed like it would have the same issue of only a radius setting shared on all corners.  There was also a solution that used a brush to achieve this effect, but also seemed more complicated then needed.

My only concern with switching to border is if there would be any performance impact over rectangle.  So far I've not been able to find any documentation.  Considering how simple the shapes are and given UWP is a desktop platform, I don't believe there will be a performance issue.

### Issues Resolved ### 

- fixes #4459

### API Changes ###

 
 None

### Platforms Affected ### 

- UWP

Updates the UWP BoxRenderer to use Border instead of Rectangle.  Since Rectangle has no way to individually control corner radius, switch the renderer to use Border.  Used the FrameRenderer as a reference to correctly apply the background with the corner radius.

None

Not applicable

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
+++ b/Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
@@ -8,6 +8,7 @@ using Windows.UI.Xaml.Media;
 using Windows.UI.Xaml.Shapes;
 using Xamarin.Forms.ControlGallery.WindowsUniversal;
 using Xamarin.Forms.Controls.Issues;
+using Xamarin.Forms.Platform.UAP;
 using Xamarin.Forms.Platform.UWP;
 
 [assembly: ExportRenderer(typeof(Xamarin.Forms.Controls.Issues.Bugzilla42602.TextBoxView), typeof(Xamarin.Forms.ControlGallery.WindowsUniversal.TextBoxViewRenderer))]
@@ -119,7 +120,9 @@ namespace Xamarin.Forms.ControlGallery.WindowsUniversal
 	}
 
 
+#pragma warning disable 618
 	public class TextBoxViewRenderer : BoxViewRenderer
+#pragma warning restore 618
 	{
 		Canvas m_Canvas;
 
--- a/Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
+++ b/Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
@@ -8,7 +8,6 @@ using Windows.UI.Xaml.Media;
 using Windows.UI.Xaml.Shapes;
 using Xamarin.Forms.ControlGallery.WindowsUniversal;
 using Xamarin.Forms.Controls.Issues;
-using Xamarin.Forms.Platform.UAP;
 using Xamarin.Forms.Platform.UWP;
 
 [assembly: ExportRenderer(typeof(Xamarin.Forms.Controls.Issues.Bugzilla42602.TextBoxView), typeof(Xamarin.Forms.ControlGallery.WindowsUniversal.TextBoxViewRenderer))]
--- a/Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
+++ b/Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
@@ -119,9 +119,7 @@ namespace Xamarin.Forms.ControlGallery.WindowsUniversal
 	}
 
 
-#pragma warning disable 618
-	public class TextBoxViewRenderer : BoxViewRenderer
-#pragma warning restore 618
+	public class TextBoxViewRenderer : BoxViewBorderRenderer
 	{
 		Canvas m_Canvas;
 
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml
@@ -29,7 +29,7 @@
                 <Entry x:Name="BottomLeft" Text="0" TextChanged="InputView_OnTextChanged"></Entry>
             </StackLayout>
 
-            <issues:Issue4459BoxView x:Name="BoxView" BackgroundColor="#CC3300"
+            <BoxView x:Name="BoxView" BackgroundColor="#CC3300"
                      WidthRequest="200"
                      HeightRequest="100"
                      CornerRadius="10,0,0,10"
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml
@@ -1,0 +1,38 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             xmlns:d="http://xamarin.com/schemas/2014/forms/design"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             mc:Ignorable="d"
+             x:Class="Xamarin.Forms.Controls.Issues.Issue4459">
+    <ContentPage.Content>
+        <StackLayout>
+            <StackLayout Orientation="Horizontal">
+                <Label Text="Top Left"></Label>
+                <Entry x:Name="TopLeft" Text="10" TextChanged="InputView_OnTextChanged"></Entry>
+            </StackLayout>
+
+            <StackLayout Orientation="Horizontal">
+                <Label Text="Top Right"></Label>
+                <Entry x:Name="TopRight" Text="0" TextChanged="InputView_OnTextChanged"></Entry>
+            </StackLayout>
+
+            <StackLayout Orientation="Horizontal">
+                <Label Text="Bottom Right"></Label>
+                <Entry x:Name="BottomRight" Text="10" TextChanged="InputView_OnTextChanged"></Entry>
+            </StackLayout>
+
+            <StackLayout Orientation="Horizontal">
+                <Label Text="Bottom Left"></Label>
+                <Entry x:Name="BottomLeft" Text="0" TextChanged="InputView_OnTextChanged"></Entry>
+            </StackLayout>
+
+            <BoxView x:Name="BoxView" BackgroundColor="#CC3300"
+                     WidthRequest="200"
+                     HeightRequest="100"
+                     CornerRadius="10,0,0,10"
+                     Margin="50"
+            />
+        </StackLayout>
+    </ContentPage.Content>
+</ContentPage>
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml
@@ -3,10 +3,12 @@
              xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
              xmlns:d="http://xamarin.com/schemas/2014/forms/design"
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             xmlns:issues="clr-namespace:Xamarin.Forms.Controls.Issues;assembly=Xamarin.Forms.Controls"
              mc:Ignorable="d"
              x:Class="Xamarin.Forms.Controls.Issues.Issue4459">
     <ContentPage.Content>
         <StackLayout>
+            <Label>Enter a corner radius for each corner and the box view below should update.</Label>
             <StackLayout Orientation="Horizontal">
                 <Label Text="Top Left"></Label>
                 <Entry x:Name="TopLeft" Text="10" TextChanged="InputView_OnTextChanged"></Entry>
@@ -27,7 +29,7 @@
                 <Entry x:Name="BottomLeft" Text="0" TextChanged="InputView_OnTextChanged"></Entry>
             </StackLayout>
 
-            <BoxView x:Name="BoxView" BackgroundColor="#CC3300"
+            <issues:Issue4459BoxView x:Name="BoxView" BackgroundColor="#CC3300"
                      WidthRequest="200"
                      HeightRequest="100"
                      CornerRadius="10,0,0,10"
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml.cs
@@ -20,9 +20,4 @@ namespace Xamarin.Forms.Controls.Issues
 			double.Parse(BottomLeft.Text), double.Parse(BottomRight.Text));
 		}
 	}
-
-	public class Issue4459BoxView : BoxView
-	{
-
-	}
 }
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml.cs
@@ -20,8 +20,10 @@ namespace Xamarin.Forms.Controls.Issues
 
 		void InputView_OnTextChanged(object sender, TextChangedEventArgs e)
 		{
+#if APP
 			BoxView.CornerRadius = new CornerRadius(double.Parse(TopLeft.Text), double.Parse(TopRight.Text),
 			double.Parse(BottomLeft.Text), double.Parse(BottomRight.Text));
+#endif
 		}
 	}
 }
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml.cs
@@ -6,12 +6,16 @@ namespace Xamarin.Forms.Controls.Issues
 {
 	[Preserve(AllMembers = true)]
 	[Issue(IssueTracker.Github, 4459, "[UWP] BoxView CornerRadius doesn't work", PlatformAffected.UWP)]
+#if APP
 	[XamlCompilation(XamlCompilationOptions.Compile)]
+#endif
 	public partial class Issue4459 : ContentPage
 	{
 		public Issue4459()
 		{
+#if APP
 			InitializeComponent();
+#endif
 		}
 
 		void InputView_OnTextChanged(object sender, TextChangedEventArgs e)
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml.cs
@@ -1,0 +1,23 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using Xamarin.Forms.Xaml;
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 4459, "[UWP] BoxView CornerRadius doesn't work", PlatformAffected.UWP)]
+	[XamlCompilation(XamlCompilationOptions.Compile)]
+	public partial class Issue4459 : ContentPage
+	{
+		public Issue4459()
+		{
+			InitializeComponent();
+		}
+
+		void InputView_OnTextChanged(object sender, TextChangedEventArgs e)
+		{
+			BoxView.CornerRadius = new CornerRadius(double.Parse(TopLeft.Text), double.Parse(TopRight.Text),
+			double.Parse(BottomLeft.Text), double.Parse(BottomRight.Text));
+		}
+	}
+}
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml.cs
@@ -20,4 +20,9 @@ namespace Xamarin.Forms.Controls.Issues
 			double.Parse(BottomLeft.Text), double.Parse(BottomRight.Text));
 		}
 	}
+
+	public class Issue4459BoxView : BoxView
+	{
+
+	}
 }
--- a/Xamarin.Forms.Platform.UAP/BoxViewBorderRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/BoxViewBorderRenderer.cs
@@ -3,9 +3,8 @@ using Windows.UI.Xaml;
 using Windows.UI.Xaml.Automation.Peers;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Shapes;
-using Xamarin.Forms.Platform.UWP;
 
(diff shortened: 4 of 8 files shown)


bmacombe:
I don't see any issues with using a Border vs a Rectangle to represent a BoxView but I'm not 100% sure.

StephaneDelcroix:
please provide a title and a description

Xamarin.Forms.Platform.UAP/BoxViewRenderer.cs
```diff
@@ -48,10 +49,22 @@ protected override AutomationPeer OnCreateAutomationPeer()
...
+		protected override void UpdateBackgroundColor()
+		{
+			//background color change must be handled separately
+			//because the background would protrude through the border if the corners are rounded
+			//as the background would be applied to the renderer's FrameworkElement
+			Color backgroundColor = Element.BackgroundColor;
+			if (Control != null)
+			{
+				Control.Background = backgroundColor.IsDefault ? null : backgroundColor.ToBrush();
+			}
```
felipebaltazar: Here you can use early return pattern, to validate if control is null..
eg:

```C#
			if (Control != null) return;
			
			//background color change must be handled separately
			//because the background would protrude through the border if the corners are rounded
			//as the background would be applied to the renderer's FrameworkElement
			Control.Background = backgroundColor.IsDefault ? null : Element.BackgroundColor.ToBrush();
```

bmacombe: I honestly just copied how the FrameRenderer was handling the background because I ran into the issue of the background color extending beyond the border.  Mainly a style preference, but I'm ok with changing that. 

In your example it looks like the Element.BackgroundColor property would be called twice, you want to avoid that since bindable properties are essentially list a lookup

felipebaltazar: Thats true! you have a good point about bindable properties!

Xamarin.Forms.Platform.UAP/BoxViewRenderer.cs
```diff
@@ -1,11 +1,12 @@
 ﻿using System.ComponentModel;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Automation.Peers;
+using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Shapes;
 
 namespace Xamarin.Forms.Platform.UWP
 {
-	public class BoxViewRenderer : ViewRenderer<BoxView, Windows.UI.Xaml.Shapes.Rectangle>
+	public class BoxViewRenderer : ViewRenderer<BoxView, Border>
```
samhouts: breaking change

Xamarin.Forms.Platform.UAP/BoxViewRenderer.cs
```diff
@@ -48,10 +49,22 @@ protected override AutomationPeer OnCreateAutomationPeer()
...
+			//as the background would be applied to the renderer's FrameworkElement
+			Color backgroundColor = Element.BackgroundColor;
+			if (Control != null)
+			{
+				Control.Background = backgroundColor.IsDefault ? null : backgroundColor.ToBrush();
+			}
+		}
+
+
 		private void SetCornerRadius(CornerRadius cornerRadius)
```
samhouts:
```suggestion
		void SetCornerRadius(CornerRadius cornerRadius)
```

Xamarin.Forms.Platform.UAP/BoxViewRenderer.cs
bmacombe: @samhouts  I wondered if that would be case.  I will do a little more investigation into how to maybe do it with a rectangle shape, but I"m not optimistic on that.

In the mean time, I'll create a new renderer, any suggestions on how you would name that?  BoxViewRenderer2?

bmacombe:
https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/shapes#rectangle

"There are some scenarios for UI definitions where instead of using a Rectangle, a Border might be more appropriate. If your intention is to create a rectangle shape around other content, it might be better to use Border because it can have child content and will automatically size around that content, rather than using the fixed dimensions for height and width like Rectangle does. A Border also has the option of having rounded corners if you set the CornerRadius property."

I think to enable the CornerRadius feature a border will need to be used, I'll rework this PR to be a new renderer and obsolete the current one.  If anyone has another suggestion, I'm all ears.

bmacombe:
@samhouts @jfversluis Updated to create a new renderer and obsolete the original.  Please review again at your convenience.  Thanks!

bmacombe:
Updated my master and rebased to remove conflicts

Xamarin.Forms.Platform.UAP/BoxViewRenderer2.cs
```diff
@@ -0,0 +1,69 @@
+﻿using System.ComponentModel;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Automation.Peers;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Shapes;
+using Xamarin.Forms.Platform.UWP;
+
+namespace Xamarin.Forms.Platform.UAP
+{
+	class BoxViewRenderer2 : ViewRenderer<BoxView, Border>
```
jfversluis: Needs to be public so people can write custom renderers. Also, not really a big fan of this name.

Xamarin.Forms.Platform.UAP/Properties/AssemblyInfo.cs
```diff
@@ -1,12 +1,13 @@
...
+using Xamarin.Forms.Platform.UAP;
 using Xamarin.Forms.Platform.UWP;
 
 [assembly: Dependency(typeof(WindowsSerializer))]
 
 // Views
 
 [assembly: ExportRenderer(typeof(Layout), typeof(LayoutRenderer))]
-[assembly: ExportRenderer(typeof(BoxView), typeof(BoxViewRenderer))]
+[assembly: ExportRenderer(typeof(BoxView), typeof(BoxViewRenderer2))]
```
jfversluis: I think we are not going to make this the default renderer from the start. IMO there would need to be some kind of transition period. People can use "BoxViewRenderer2" if they need it and we give people the time to transition become of the obsoleting message. Later we will switch to this new one.

Xamarin.Forms.Platform.UAP/BoxViewRenderer.cs
```diff
@@ -1,10 +1,12 @@
-﻿using System.ComponentModel;
+﻿using System;
+using System.ComponentModel;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Automation.Peers;
 using Windows.UI.Xaml.Shapes;
 
 namespace Xamarin.Forms.Platform.UWP
 {
+	[Obsolete("Replaced with BoxViewRenderer2")]
```
jfversluis: Please have a look at how other obsolete messages look and make it more like that

Xamarin.Forms.Platform.UAP/BoxViewRenderer2.cs
bmacombe: Thanks, I'll fix that.

Xamarin.Forms.Platform.UAP/Properties/AssemblyInfo.cs
bmacombe: Sounds good to me

Xamarin.Forms.Platform.UAP/BoxViewRenderer2.cs
bmacombe: Any suggestion on the name?  I'm struggling there also.  Trying to think of a meaningful name.

Xamarin.Forms.Platform.UAP/BoxViewRenderer.cs
bmacombe: "BoxViewRenderer is obsolete as of version 4.4. Please use BoxViewRenderer2 instead."  I'm taking a guess at the version this might get merged into, I might put 4.x and update it once everything is approved.

Xamarin.Forms.Platform.UAP/BoxViewRenderer2.cs
bmacombe: I'll go with BoxViewBorderRenderer for now

Xamarin.Forms.Platform.UAP/Xamarin.Forms.Platform.UAP.csproj
```diff
@@ -13,7 +13,7 @@
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
     <TargetPlatformVersion>10.0.16299.0</TargetPlatformVersion>
     <TargetPlatformMinVersion>10.0.16299.0</TargetPlatformMinVersion>
-    <SkipMicrosoftUIXamlCheckTargetPlatformVersion>true</SkipMicrosoftUIXamlCheckTargetPlatformVersion>		
+    <SkipMicrosoftUIXamlCheckTargetPlatformVersion>true</SkipMicrosoftUIXamlCheckTargetPlatformVersion>
```
bmacombe: I"m not sure where this is extra line break is coming from here, I don't see it when editing my local copy or directly on GitHub

bmacombe:
Updated to add some details to the issue page and to use the new renderer in issue page.

bmacombe:
Did some more investigation into how to handle the corner radius.  Below are two options if anyone has an opinion.

I still haven't found a way to do this without creating a new renderer.

``` xaml
         <!-- Currently how XF handles a BoxView on UWP -->
        <Rectangle Margin="25" Fill="Yellow" Width="100" Height="100" RadiusX="25" RadiusY="25"></Rectangle>

        <!-- Currently how this PR is implemented-->
        <Frame Margin="25" Width="100" Height="100" Background="blue" CornerRadius="25, 0, 25, 0"></Frame>

        <!-- Rectangle inside a grid to add corder radius-->
        <Grid Margin="25" CornerRadius="25, 0, 25, 0" Width="100" Height="100">
            <Rectangle Fill="Red"></Rectangle>
        </Grid>
```

image

bmacombe:
Just an update, I had a brief idea that maybe a control template could be updated...but rectangle has no control template...which makes sense.

I still believe the only option to enable this is to depreciate the existing Rectangle based Renderer and a new Border (which seems to be the UWP recommedation) based Renderer (how this PR is implemented).  I suggest allowing developers to optionally use the new renderer until the next major XF release at which time it would be the default, but whatever the team decides there is fine with me.

bmacombe:
rebased to remove conflicts

Xamarin.Forms.Platform.UAP/BoxViewBorderRenderer.cs
```diff
@@ -0,0 +1,69 @@
+﻿using System.ComponentModel;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Automation.Peers;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Shapes;
+using Xamarin.Forms.Platform.UWP;
+
+namespace Xamarin.Forms.Platform.UAP
```
samhouts:
```suggestion
namespace Xamarin.Forms.Platform.UWP
```

Xamarin.Forms.Platform.UAP/BoxViewBorderRenderer.cs
```diff
@@ -0,0 +1,69 @@
+﻿using System.ComponentModel;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Automation.Peers;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Shapes;
+using Xamarin.Forms.Platform.UWP;
```
samhouts:
```suggestion
```

Xamarin.Forms.Platform.UAP/Properties/AssemblyInfo.cs
```diff
@@ -1,12 +1,15 @@
 ﻿using Xamarin.Forms;
+using Xamarin.Forms.Platform.UAP;
```
samhouts:
```suggestion
```

Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
```diff
@@ -6,12 +6,14 @@
 using Windows.UI.Xaml.Shapes;
 using Xamarin.Forms.ControlGallery.WindowsUniversal;
 using Xamarin.Forms.Controls.Issues;
+using Xamarin.Forms.Platform.UAP;
```
samhouts:
```suggestion
```

bmacombe:
@samhouts Sounds good to me, I don't think it should cause any problems for anyone and since the old rendererer is still there we won't break any custom renderers.

Should the default namespace for the UWP project be changed?
image

Thanks for pointing those out, I didn't event think about checking that.

bmacombe:
@samhouts Fixed the namespace and rebased on master

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml.cs
```diff
@@ -0,0 +1,28 @@
...
+		}
+
+		void InputView_OnTextChanged(object sender, TextChangedEventArgs e)
+		{
+			BoxView.CornerRadius = new CornerRadius(double.Parse(TopLeft.Text), double.Parse(TopRight.Text),
+			double.Parse(BottomLeft.Text), double.Parse(BottomRight.Text));
+		}
+	}
+
+	public class Issue4459BoxView : BoxView
```
samhouts:
```suggestion
```

Xamarin.Forms.Platform.UAP/BoxViewRenderer.cs
```diff
@@ -1,10 +1,12 @@
-﻿using System.ComponentModel;
+﻿using System;
+using System.ComponentModel;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Automation.Peers;
 using Windows.UI.Xaml.Shapes;
 
 namespace Xamarin.Forms.Platform.UWP
 {
+	[Obsolete("BoxViewRenderer is obsolete as of version 4.X. Please use BoxViewBorderRenderer instead.", false)]
```
samhouts:
```suggestion
	[Obsolete("BoxViewRenderer is obsolete as of version 4.6.0. Please use BoxViewBorderRenderer instead.", false)]
```

Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
```diff
@@ -12,6 +12,7 @@
 [assembly: ExportRenderer(typeof(Issue1683.EntryKeyboardFlags), typeof(EntryRendererKeyboardFlags))]
 [assembly: ExportRenderer(typeof(Issue1683.EditorKeyboardFlags), typeof(EditorRendererKeyboardFlags))]
 [assembly: ExportRenderer(typeof(Issue3273.SortableListView), typeof(SortableListViewRenderer))]
+[assembly: ExportRenderer(typeof(Issue4459BoxView), typeof(Issue4459BoxViewRenderer))]
```
samhouts:
```suggestion
```

Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
```diff
@@ -163,4 +166,9 @@ protected override void OnElementChanged(ElementChangedEventArgs<BoxView> e)
 			m_Canvas.Children.Add(text);
 		}
 	}
+
+	public class Issue4459BoxViewRenderer : BoxViewBorderRenderer
```
samhouts:
```suggestion
```

Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
```diff
@@ -163,4 +166,9 @@ protected override void OnElementChanged(ElementChangedEventArgs<BoxView> e)
...
+	{
```
samhouts:
```suggestion
```

Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
```diff
@@ -163,4 +166,9 @@ protected override void OnElementChanged(ElementChangedEventArgs<BoxView> e)
 			m_Canvas.Children.Add(text);
 		}
 	}
+
```
samhouts:
```suggestion
```

Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
```diff
@@ -163,4 +166,9 @@ protected override void OnElementChanged(ElementChangedEventArgs<BoxView> e)
...
+	public class Issue4459BoxViewRenderer : BoxViewBorderRenderer
+	{
+
```
samhouts:
```suggestion
```

Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
```diff
@@ -163,4 +166,9 @@ protected override void OnElementChanged(ElementChangedEventArgs<BoxView> e)
...
+	}
```
samhouts:
```suggestion
```

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml.cs
```diff
@@ -0,0 +1,28 @@
...
+	{
```
samhouts:
```suggestion
```

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml.cs
```diff
@@ -0,0 +1,28 @@
...
+
```
samhouts:
```suggestion
```

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml.cs
```diff
@@ -0,0 +1,28 @@
...
+	}
```
samhouts:
```suggestion
```

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4459.xaml
```diff
@@ -0,0 +1,40 @@
...
+                <Label Text="Bottom Right"></Label>
+                <Entry x:Name="BottomRight" Text="10" TextChanged="InputView_OnTextChanged"></Entry>
+            </StackLayout>
+
+            <StackLayout Orientation="Horizontal">
+                <Label Text="Bottom Left"></Label>
+                <Entry x:Name="BottomLeft" Text="0" TextChanged="InputView_OnTextChanged"></Entry>
+            </StackLayout>
+
+            <issues:Issue4459BoxView x:Name="BoxView" BackgroundColor="#CC3300"
```
samhouts:
```suggestion
            <BoxView x:Name="BoxView" BackgroundColor="#CC3300"
```

Xamarin.Forms.Platform.UAP/Properties/AssemblyInfo.cs
```diff
@@ -7,7 +7,9 @@
 // Views
 
 [assembly: ExportRenderer(typeof(Layout), typeof(LayoutRenderer))]
+#pragma warning disable CS0618 // Type or member is obsolete
 [assembly: ExportRenderer(typeof(BoxView), typeof(BoxViewRenderer))]
```
bmacombe: @samhouts  Should I remove disabling that warning and change the default renderer to the new one here also?

bmacombe:
@samhouts I removed the custom renderer used for testing and make the new BoxViewBorderRenderer the default...assuming that is what you intended.  If not, just let me know.

jfversluis:
There seem to be a couple of build errors.

> ##[error]Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Issue4459.xaml.cs(9,3): Error CS0246: The type or namespace name 'XamlCompilationAttribute' could not be found (are you missing a using directive or an assembly reference?)

> ##[error]Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Issue4459.xaml.cs(9,3): Error CS0246: The type or namespace name 'XamlCompilation' could not be found (are you missing a using directive or an assembly reference?)

> ##[error]Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Issue4459.xaml.cs(9,19): Error CS0103: The name 'XamlCompilationOptions' does not exist in the current context

> ##[error]Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Issue4459.xaml.cs(19,4): Error CS0120: An object reference is required for the non-static field, method, or property 'BoxView.CornerRadius'

> ##[error]Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Issue4459.xaml.cs(19,57): Error CS0103: The name 'TopLeft' does not exist in the current context

And a couple more for all the corners :)

bmacombe:
@jfversluis I'll take a look...it built on my machine :)

jfversluis:
For the `XamlCompilation` error you probably need something like this: https://github.com/xamarin/Xamarin.Forms/blob/03266f580ad9bbcabfc7ce99dd25bc50aceadfbd/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5949_1.xaml.cs#L14

bmacombe:
@jfversluis Thanks for that tip...I was just removing that attribute, but I'll use that method instead.

bmacombe:
@jfversluis On the other errors "BoxView" is the name I gave the box view control in xaml so the code behind could reference it for the issue page.  My best guess is that unit test projects don't like that?  For now I'm just making the code that interactively updates the corner radius as only when an APP like the complied xaml attribute. Since it's not a uitest it should be ok?  Or if I'm missing something here please let me know.

Code is all udpated.

Thanks!

jfversluis:
Yep, this looks good! Triggering the build now

bmacombe:
@jfversluis Looks like it built fine...btw, it's awesome we can access the DevOp builds now!

Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
```diff
@@ -115,7 +115,9 @@ public static void SetKeyboardFlags(this FormsTextBox Control, KeyboardFlags? fl
 	}
 
 
+#pragma warning disable 618
 	public class TextBoxViewRenderer : BoxViewRenderer
```
jfversluis: We can't use the new renderer here too? If it's obsolete, I guess we shouldn't be using it anymore either 😉

bmacombe: Fine with me, updated

bmacombe:
rebased for conflicts

PR #7991, opened 2019-10-14
VisualElement.IsFocused make Bindable

### LINKED ISSUE #7944: [Enhancement] VisualElement.IsFocused make bindable
## Summary
I think that it would be nice to make VisualElement.IsFocused - Bindable. It’s very convenient from ViewModel to send focus to some text field, or to pick focus.

## API Changes

``` cs
VisualElement.IsFocused {get;set;}
```

KSemenenko:
### Description of Change ###
Solved the problem of setting focus via Binding. 
Changing the existing property "IsFocused" now  is a bindable property.
### Issues Resolved ### 

- fixes #7944

### API Changes ###
```` cs
VisualElement.IsFocused {get;set;}
````

Changed:
 - bool VisualElement.IsFocused { get; set; } 

### Platforms Affected ### 
- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###
1. open ControlGallery
2. Select "IsFocused Bindable Gallery"
4. Press "Focus" Button
5. Press "Unfocus" Button


StephaneDelcroix:
I don't think this is the right solution for the problem

StephaneDelcroix:
> fixes #7944

does not fixes it

KSemenenko:
The problem is that IsFocused gets the value from the Renderer and changing this logic creates a large number of changes.
If you have any opinion how to do this, I will be happy to implement it. @StephaneDelcroix

KSemenenko:
and I’ll try to do something with the IsFocused property as well.

KSemenenko:
@StephaneDelcroix  What do you think is the solution? I left the private property, because it is set via renderer

KSemenenko:
@StephaneDelcroix @kingces95 @samhouts  Maybe you have already watched this PR and have any comments?

KSemenenko:
@StephaneDelcroix @kingces95 I updated the branch from the last develop

KSemenenko:
@samhouts @StephaneDelcroix @kingces95 
any updates?

KSemenenko:
I fixed the tests. Now everything works as it should.
@StephaneDelcroix @kingces95 @samhouts

KSemenenko:
@samhouts @PureWeen 
any updates?

samhouts:
@KSemenenko On the list! Thanks for your patience!

StephaneDelcroix:
IsFocused has always been Bindable, but read-only. having it that way ensure that only one control is focused at the time, and the responsibility is on the platform.

you can still call Focus manually, to request Focus.

if you really need a BP, you can add an attached BP, defined in your code, that does just that.

I'm closing this for now

PR #8011, opened 2019-10-14
Fix Material Editor so TextChange delegate isn't removed

### LINKED ISSUE #5500: [iOS] Editor with material visuals value binding not working on physical device
### Description
Xamarin Forms 3.6.0.220655 & Xamarin.Forms.Visual.Material (I only tested iOS)
Using an editor with visual="Material" works on simulator but when deploying on a physical device the values entered in the editor will not be updated.
### Steps to Reproduce

1. use editor control in xaml
2. apply visual="Material" and Text={Binding ValueInModel}

### Expected Behavior
ValueInModel should change
### Actual Behavior
ValueInModel changes only in Simulator not on physical device
### Basic Information

- Version with issue:
- Platform Target Frameworks: 
  - iOS:  11.0 (12.1 has the same behavior)
- Nuget Packages: Xamarin Forms 3.6.0.220655 & Xamarin.Forms.Visual.Material
- Affected Devices: iPhone 7 and iPhone 8 Plus

PureWeen:
### Description of Change ###
After assigning events to the event handler tied to text changes store it to a local variable. If I don't do this then at some point the Delegate just gets nulled out on the material rendererers when launched to a device

### Issues Resolved ### 
- fixes #5500 

### Platforms Affected ### 
- iOS

### Testing Procedure ###
- Test included with instructions
- Also test Material Entry and that auto size based on text changes works

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5500.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5500.cs
@@ -58,38 +58,24 @@ namespace Xamarin.Forms.Controls.Issues
 		protected override void OnAppearing()
 		{
 			base.OnAppearing();
-			Device.BeginInvokeOnMainThread(async () =>
-			{
-				GarbageCollectionHelper.Collect();
-				editor.Text = "First Test";
-				await Task.Delay(1);
-				if(editor.Text != entry.Text)
-				{
-					editor.Text = "Test has failed";
-					entry.Text = "Test has failed";
-					return;
-				}
-
-				entry.Text = "Second Test";
-				await Task.Delay(1);
-				if (editor.Text != entry.Text)
-				{
-					editor.Text = "Test has failed";
-					entry.Text = "Test has failed";
-					return;
-				}
-
-				await Task.Delay(1);
-				editor.Text = "Success";
-				entry.Text = "Success";
-			});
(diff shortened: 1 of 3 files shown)


PureWeen:
UI tests for this one already passed don't need to rerun just because of the merge

Ghasan:
Awesome 🥇  Fixed it for us 👍 #5500

mastermoe01:
when will this fix go out?

jfversluis:
Shouldn't be long, can't give you an exact date, but should be as soon as next week

mastermoe01:
@jfversluis  I thought xamarin.forms 4.3.0.908675 and xamarin.forms.visual.material 4.3.0.908675 has this issue resolved? I'm still having issues in IOS where editor with visual material not binding the filled in text. it works fine in android.

PureWeen:
@mastermoe01  😢 

So it looks like this changed didn't get merged up from 4.2 before 4.3 was tagged
Can you test with the latest 4.2 and let me know if it works for you there? 

This will be fix on 4.3 with the next service release

In the mean time if you would like to stay on 4.3 you can do this to work around the issue
https://github.com/xamarin/Xamarin.Forms/issues/5500#issuecomment-542839002

mastermoe01:
@PureWeen thank you for the update, i confirmed that the going to the latest 4.2 does work properly, I will just stick with that and await the next 4.3 service release.

mastermoe01:
fixed in 4.3 service release, thank you!

PR #8048, opened 2019-10-16
[iOS] Add UpdateMode platform-specific to DatePicker and TimePicker

jcmanke:
### Description of Change ###

Add iOS-specific to UpdateMode enum from Picker to DatePicker and TimePicker.

### Issues Resolved ### 

Didn't see any GitHub issues, but some related forum threads:
- https://forums.xamarin.com/discussion/93521/what-is-date-picker-behaviour-in-xamarin-forms
- https://forums.xamarin.com/discussion/98832/add-cancel-button-to-ios-datepicker-and-timepicker-only-change-the-value-when-done-is-clicked
- https://forums.xamarin.com/discussion/94666/cant-detect-ok-button-press-on-datepicker

### API Changes ###

Added:
 - `Xamarin.Forms.PlatformConfiguration.iOSSpecific.DatePicker` class
 - `Xamarin.Forms.PlatformConfiguration.iOSSpecific.TimePicker` class

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS

When UpdateMode is set to WhenFinished, the Date/Time property will be set when the user closes the picker with the Done button rather than as they scroll.

Not applicable

### Testing Procedure ###

Use the Picker Gallery - Legacy page in the Control Gallery app. Setting the UpdateMode from the picker 4th from the bottom will affect the Picker, DatePicker, and TimePicker below it.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/DatePicker.cs
+++ b/Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/DatePicker.cs
@@ -1,6 +1,6 @@
 ﻿namespace Xamarin.Forms.PlatformConfiguration.iOSSpecific
 {
-    using FormsElement = Xamarin.Forms.DatePicker;
+	using FormsElement = Xamarin.Forms.DatePicker;
 
 	public static class DatePicker
 	{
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Core/PlatformConfiguration/iOSSpecific/DatePicker.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace Xamarin.Forms.PlatformConfiguration.iOSSpecific
+{
+    using FormsElement = Xamarin.Forms.DatePicker;
```
rmarinho: can we fix style

PR #8057, opened 2019-10-17
[Tizen] Register self-owned renderers on preload

swift-kim:
### Description of Change ###
Self-register renderers in Xamarin.Forms assemblies ahead-of-time in `Forms.Preload()`. This improves the time required for application launches up to 40 ms (including JIT compilation time) in our testing environment.

See the previous change #7440. Also related to #7209, #7441.

### Issues Resolved ### 
None

### API Changes ###
None

### Platforms Affected ### 
Tizen

None

N/A

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Tizen/Forms.cs
+++ b/Xamarin.Forms.Platform.Tizen/Forms.cs
@@ -176,6 +176,12 @@ namespace Xamarin.Forms
 			private set;
 		}
 
+		internal static bool IsPreloaded
+		{
+			get;
+			private set;
+		}
+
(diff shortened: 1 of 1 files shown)


swift-kim:
cc @rookiejava, @myroot

rookiejava:
Currently, we're focused on optimizing app startup time on tizen. And thank you(@swift-kim ) for your first contribution for Xamarin.Forms. : ) This PR is also associated with this topic. However, I think it's not a simple problem, we need to think more carefully about the various scenarios. So it would be better discuss it more and then re-submit it.

PR #8079, opened 2019-10-18
Fix issue 6878 | [Bug] ShellItem.Items.Clear() crashes when the ShellItem has bottom tabs

### LINKED ISSUE #6878: [Bug] ShellItem.Items.Clear() crashes when the ShellItem has bottom tabs
### Description
When using shell, if you want to programmatically clear and redraw the bottom tabs on a shell page, you can't use Items.Clear to get rid of all the existing ones because it crashes. Investigation suggests that this is because it's trying to set properties on a tab that is no longer there.

### Steps to Reproduce

1. Create a ShellItem in xaml that contains tabs.
2. In code, call Items.Clear() on your ShellItem.
3. A null exception will occur on line 83 of ShellBottomNavViewAppearanceTracker.SetBackgroundColor, presumably because menuView.GetChildAt(index) returns null, either because index is -1 or menuView has no children.

### Expected Behavior
The tabs are cleared and you're free to create new ones
### Actual Behavior
Exception and crash
### Basic Information

- IDE: Visual Studio 2019
- Platform Target Frameworks: 
  - iOS:  Untested
  - Android: 9.0

felipebaltazar:
### Description of Change ###

I followed @PureWeen suggestion, that's I agree, to just throw an exception that clear isn't supported for ShellItem
I tried to avoid exceptions, but Shell get lost and some navigations throw`s other exceptions

### Issues Resolved ### 

- fixes #6878 

### API Changes ###
Changed:
 - void ShellItem.ItemsCollectionChanged
	- added this validation:

```C#
if ((e.Action == NotifyCollectionChangedAction.Reset ||
	e.Action == NotifyCollectionChangedAction.Remove) && Items.Count == 0)
	throw new NotSupportedException("the ShellItem.Items must have at least one item.");
```
 
### Platforms Affected ### 
- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###
1. Create a ShellItem in that contains tabs and set background-color
2. In code, call Items.Clear() on your ShellItem.
3. A NotSupportedException will occur 

- Or just run 6878 issue at Control Galery:

<p align="center">
	<kbd>
		image
	</kbd>
</p>

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6878.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue6878.cs
@@ -1,0 +1,96 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Text;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System.Linq;
+using Xamarin.Forms.PlatformConfiguration;
+using Xamarin.Forms.PlatformConfiguration.iOSSpecific;
+using System.Threading;
+using System.ComponentModel;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+using Xamarin.Forms.Core.UITests;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 6878,
+		"ShellItem.Items.Clear() crashes when the ShellItem has bottom tabs", PlatformAffected.All)]
+
+#if UITEST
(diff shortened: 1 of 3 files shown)


PureWeen:
I agree with @StephaneDelcroix on this one

- the solution will also need to work at each level of the shell not just clearing ShellItem
- the fix should more be about a way to make this work instead of an exception
- Final thought once I work out the kinks of this PR https://github.com/xamarin/Xamarin.Forms/pull/7847 I don't think clearing is going to be necessary because with an IsVisible binding there really won't be any point to clearing the items.  Plus PR #7847 already has code that accounts for a scenario where everything is made not visible and then something is subsequently made visible

Once we have an IsVisible binding I think (HotReload) is really the only use case for clearing items. Unless there are some additional ones I haven't thought of yet :-) 

I'm going to close this one for now

PR #8080, opened 2019-10-18
Added instructions and changed the test for Issue2241

### LINKED ISSUE #2241: PreferredScreenEdgesDeferringSystemGestures can't be triggered in the PageRenderer of Xamarin.Forms
### Description

I want to override the PreferredScreenEdgesDeferringSystemGestures to delay the system behavior when touching the screen edges on iOS 11+. it works properly on Xamarin.iOS, but when I do this in a 
page renderer on Xamarin.Foms it fails. The system will not trigger this property automatically. I want to do the same thing on Forms, and how can I achieve that.

### Steps to Reproduce

1. override the PreferredScreenEdgesDeferringSystemGestures  in a page renderer:
```
public override UIRectEdge PreferredScreenEdgesDeferringSystemGestures
{
    get
    {
        return UIRectEdge.All;
    }
}
```
2. It won't be triggered when this renderer shows.

### LINKED ISSUE #2274: Issue2241 Needs instructions for tester
Currently it's just a big red square. Needs instructions for tester

adrianknight89:
### Description of Change ###

Added instructions and changed the manual test as well as the automated test. Please make sure the new automation works (I haven't had a chance to run it).

### Issues Resolved ### 

- fixes #2274

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2241.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2241.cs
@@ -3,6 +3,7 @@
 using Xamarin.Forms;
 using Xamarin.Forms.CustomAttributes;
 using Xamarin.Forms.Internals;
+using System.Linq;
 
 #if UITEST
 using NUnit.Framework;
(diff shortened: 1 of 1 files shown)


adrianknight89:
@jfversluis Sorry my bad. I changed the issue number. I believe It was 2241 on Bugzilla, but it's tracked by a different ticket on Github.

adrianknight89:
@samhouts Can you remove the inprogress label from #2241 and add it to #2274?

samhouts:
@adrianknight89 Done!

jfversluis:
The classname is 2241 and also that is linked in the attribute pointing back to Github.

So, either that needs to be changed as well, or this is implemented in the wrong class altogether :)

adrianknight89:
@jfversluis @samhouts I think the issue tracker needs to be changed to Bugzilla. If we change the class name to include 2274, then we'll lose track of the original ticket unless I add a comment to include the URL. Which option is better for you?

samhouts:
Ah, I see, that was an issue from the old GitHub repo. GitHub is technically correct. We have a bunch that overlap like this, unfortunately. You can just leave it as is. We'll figure it out. A more correct solution would be to create another tracker type for LegacyGitHub or something like that, but that would also require going through the old issues and realigning them. Not totally sure it's worth it. :)

adrianknight89:
Right. Probably not worth it for now. Thank you! :)

PR #8097, opened 2019-10-19
Fixed 5680 - [Enhancement] Added method to force value coercion

### LINKED ISSUE #5680: [Enhancement] Add method to force value coercion
## Summary
It would be nice to have a method that forces a BindableProperty to be coerced, similar to WPF's [CoerceValue method](https://docs.microsoft.com/en-us/dotnet/api/system.windows.dependencyobject.coercevalue?view=netframework-4.7.2)

## API Changes

The BindableObject class would have an additional method called CoerceValue that looks like this:

```csharp
class BindableObject
{
    void CoerceValue(BindableProperty property);
}
```

When called, this method would coerce the current value of the property if a coerceCallback is set on the property.

## Intended Use Case
This method is useful to coerce a BindableProperty even if the property itself hasn't changed. If you have some sort of TextBox control with a variable DisplayLength property for example, you could adjust the displayed text length by calling the CoerceValue method when the DisplayLength changes without changing the text itself.

techduggu:
### Description of Change ###

As requested in the issue, Added a new method `CoerceValue` to `BindableObject` which when called, it would force coerce the current value of the property if a coerceCallback is set on the property.

### Issues Resolved ### 

- fixes #5680 

### API Changes ###

Added:
Below methods are added to `BindableObject.cs` of Xamarin.Forms.Core library:
```
public void CoerceValue(BindableProperty property) => CoerceValue(property, checkAccess: true);

public void CoerceValue(BindablePropertyKey propertyKey)
{
	if (propertyKey == null)
		throw new ArgumentNullException(nameof(propertyKey));

	CoerceValue(propertyKey.BindableProperty, checkAccess: false);
}

void CoerceValue(BindableProperty property, bool checkAccess)
{
	if (property == null)
		throw new ArgumentNullException(nameof(property));

	if (checkAccess && property.IsReadOnly)
		throw new InvalidOperationException($"The BindableProperty \"{property.PropertyName}\" is readonly.");

	BindablePropertyContext bpcontext = GetContext(property);
	if (bpcontext == null)
		return;

	object currentValue = bpcontext.Value;

	if (property.ValidateValue != null && !property.ValidateValue(this, currentValue))
		throw new ArgumentException($"Value is an invalid value for {property.PropertyName}", nameof(currentValue));

	property.CoerceValue?.Invoke(this, currentValue);
}
```

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###

image

- Just run Issue5680.xaml from Control Gallery and for force coercion, try changing the value of 'Maximum angle' which will force call the Coerce method on 'Angle' (or Rotation angle) property.

Please note, As there was no sample reproduction attached, I have added the above sample testcase from Xamarin.Forms Samples (CoerceValueCallback) to reproduce it. It isn't a best testcase (which may be required to perfectly test it).

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5680.xaml
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue5680.xaml
@@ -1,0 +1,16 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<local:TestContentPage xmlns="http://xamarin.com/schemas/2014/forms" xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml" xmlns:local="clr-namespace:Xamarin.Forms.Controls" x:Class="Xamarin.Forms.Controls.Issues.Issue5680" x:Name="page1">
+    <StackLayout Padding="0,20,0,0" BindingContext="{x:Reference page1}">
+        <Label Text="Bindable Property CoerceValue Callback Demo" FontAttributes="Bold" HorizontalOptions="Center" />
+        <StackLayout Orientation="Horizontal" HorizontalOptions="Center">
+            <Label Text="Maximum angle:" />
+            <Entry Text="{Binding MaximumAngle}" WidthRequest="50" />
+        </StackLayout>
+        <StackLayout Orientation="Horizontal" HorizontalOptions="Center">
+            <Label Text="Rotation angle:" />
+            <Entry Text="{Binding Angle}" WidthRequest="50" />
+        </StackLayout>
+        <Label x:Name="CoerceResult" FontAttributes="Bold" HorizontalOptions="Center"/>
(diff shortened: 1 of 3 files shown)


techduggu:
Hi Team (@samhouts / @jfversluis ),

Thanks for approving it. I can see the above build is failing (at high level, I can see that it is unable to publish the test results). Do I need to worry about such cases or Is there any action item on me which is blocking it to merge into master branch (I guess that happens automatically as part of CI)? Please don't mind as it is my first time.
Thanks!

jfversluis:
Hey @techduggu!

Good that you're on top of this! 😉 First: don't worry about the questions, we're here to help!

I've taken a quick look at the failing tests and so far it seems nothing related. I think the test results will only be published whenever all of them are done. In this case the UI tests are maybe not even that useful, but I usually like to trigger them anyway in cases where I have doubts.

In general; whenever you see a failing build, you should be able to see what is wrong and fix it yourself, failing UI tests are not so obvious and public. We try to keep an eye on the test runs, check if the errors are related or not and then hopefully merge the PR.

If you see the test run is completed but we haven't contacted you in a while or took any other action, please feel free to ping us so we can have another look.

Thanks!

PR #8124, opened 2019-10-21
Occasionally clean up WeakReferences in Styles whose targets have been collected

hartez:
### Description of Change ###

Styles use WeakReferences to track their targets so that the target BindableObjects can be garbage collected. However, the WeakReferences themselves are never removed from the Style; this means that long-lived Styles (e.g., application-wide styles) can accumulate WeakReferences for the entire lifetime of the application, eventually running out of memory.

These changes check the number of accumulated WeakReferences when a new reference is added; if the number of references is over a threshold, the list of references are scanned and references whose targets have already been collected are removed, allowing them to be garbage collected.

### Issues Resolved ### 
- WeakReference memory leak

### API Changes ###
None

### Platforms Affected ### 
- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###
Run a Forms application which 
- uses a global Style 
- creates many BinableObjects which use that Style (e.g., scrolling forward and back in a CollectionView or ListView)

under a profiler and watch the WeakReference instance count. If the WeakReference count ever decreases, then this fix is working. 

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Interactivity/AttachedCollection.cs
+++ b/Xamarin.Forms.Core/Interactivity/AttachedCollection.cs
@@ -8,6 +8,9 @@ namespace Xamarin.Forms
 	{
 		readonly List<WeakReference> _associatedObjects = new List<WeakReference>();
 
+		const int CleanupTrigger = 128;
+		int _cleanupThreshold = CleanupTrigger;
+
 		public AttachedCollection()
 		{
 		}
@@ -64,6 +67,7 @@ namespace Xamarin.Forms
 			lock (_associatedObjects)
 			{
 				_associatedObjects.Add(new WeakReference(bindable));
+				CleanUpWeakReferences();
 			}
 			foreach (T item in this)
 				item.AttachTo(bindable);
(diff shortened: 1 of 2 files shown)


hartez:
The Android API19 failure is utterly unrelated.
The iOS 12 failures are also unrelated.

Xamarin.Forms.Core/Interactivity/AttachedCollection.cs
```diff
@@ -123,5 +127,16 @@ protected override void SetItem(int index, T item)
 				item.AttachTo(bindable);
 			}
 		}
+
+		void CleanUpWeakReferences()
```
StephaneDelcroix: AggressiveInline FTW

rlasker3:
After releasing our app with v4.4.0.991265 we are getting error's logged in AppCenter that reads:
Style+<>c.<CleanUpWeakReferences>b__47_0 (System.WeakReference`1[T] t)
System.AggregateException: Object reference not set to an instance of an object

Do you believe that this is related to this issue?

hartez:
@rlasker3 Seems likely. Please open a separate issue (and include your stack traces and any other info you can) and we'll take a look.

PR #8128, opened 2019-10-21
Change XamarinFormsVersionFile to a dot file so it's hidden

PureWeen:
### Description of Change ###

Change XamarinFormsVersionFile to a dot file so it's hidden. 

This file is only used by the CI processes and build scripts 

### Testing Procedure ###
Generate nugets from cake and make sure the *.XamarinFormsVersionFile* is generated but doesn't try to get added to SC.

Also make sure that nuget files are still generated with valid version numbers

### PR Checklist ###

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,4 +69,4 @@ caketools/
 *.binlog
 .ionide/**
 /.nuspec
-XamarinFormsVersionFile.txt
+.XamarinFormsVersionFile.txt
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -69,4 +69,4 @@ caketools/
 *.binlog
 .ionide/**
 /.nuspec
-XamarinFormsVersionFile.txt
+.XamarinFormsVersionFile.txt
```
PureWeen: @rmarinho yea

Otherwise git will try to check the files in

image

PR #8135, opened 2019-10-22
Fix Flyout behavior so it propagates from templated Content Page once it's created

### LINKED ISSUE #7985: [Bug] Shell.FlyoutBehavior="Disabled" not working android emulator
### Description
Shell.FlyoutBehavior="Disabled" does not make the flyout not appear when the user swipes from left to right. I have encountered this in android emulator (Android 9.0) and on an android phone.

### Steps to Reproduce

1.  In a xamarin.forms shell project, add flyout to AppShell.Xaml
2. `<FlyoutItem FlyoutDisplayOptions="AsMultipleItems">
        <ShellContent Title="browse" ContentTemplate="{DataTemplate local:ItemsPage}" />
        <ShellContent Title="about" ContentTemplate="{DataTemplate local:AboutPage}" />
    </FlyoutItem>`
3. In one of the pages like say AboutPage add `Shell.FlyoutBehavior="Disabled"` in the ContentPage tag

### Expected Behavior
Flyout should not appear when the user swipes from left to right

### Actual Behavior
The flyout appears.

- Version with issue: 4.2.0.848062
- Last known good version:
- IDE: Visual Studio
- Platform Target Frameworks: 
  - iOS:  
  - Android: 9.0 - API 28 
  - UWP:   
- Nuget Packages:
- Affected Devices:

PureWeen:
### Description of Change ###

Flyout Behavior was being calculated before the Content Page was realized so any attached properties on the Content Page were ignored. This change centralizes SendAppearanceChange calls as well so it's being called from a more correct place opposed to just trying to guess everywhere it should be called

### Issues Resolved ### 

- fixes #7985

### Platforms Affected ### 
- Core/XAML (all platforms)

### Testing Procedure ###
- Flyout behavior is automated but you can run UI test and make sure it works as you expect
- I also added a manual Appearance change test you can use to verify appearance changes propagate correctly

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/ShellAppearanceChange.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/ShellAppearanceChange.cs
@@ -1,0 +1,95 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System.Linq;
+
+
+#if UITEST
+using Xamarin.Forms.Core.UITests;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.None, 0, "Shell AppearanceChange",
+		PlatformAffected.All)]
+#if UITEST
+	[NUnit.Framework.Category(UITestCategories.Shell)]
+	[NUnit.Framework.Category(UITestCategories.ManualReview)]
+#endif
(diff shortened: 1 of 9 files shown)


PureWeen:
Tests all green 
Rebased

PureWeen:
@rmarinho 

I logged your second issue about the blank screen
https://github.com/xamarin/Xamarin.Forms/issues/8235

I tested that scenario without these changes and it crashes in the same way

Also with out these changes the Appearance changes just super don't work on iOS. For example when the app loads it doesn't even apply the *SetBackgroundColor* and pop doesn't correctly revert at all

PR #4539, opened 2018-11-27
[C] Expose Application.LogWarningsListener

StephaneDelcroix:
### Description of Change ###

The fix proposed in #2911 didn't worked with Release build of XF, aka
the majority of the builds used.
This fix allows the developer to set any logger he might like, but the
plan is to add a default one in the template, that uses Debug.WriteLine,
so it will Log if the user application is build with Debug, and not if
XF binaries are.

This also changes 2 remaining `Debug.WriteLine` into `Log.Warning` in the `Core`

### Issues Resolved ### 

- fixes #1517
- fixes #3873

### API Changes ###

Added:
 - `public static Action<(string Category, string Message)> Application.LogWarningsListerner {get; set;}`
 
### Platforms Affected ### 

- Core/XAML (all platforms)

### Behavioral/Visual Changes ###

None

### Before/After Screenshots ### 

Not applicable

### PR Checklist ###

- [ ] Has automated tests 
- [x] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard

PR #8142, opened 2019-10-22
Add Application.LogWarningsListener

### LINKED ISSUE #8119: [Bug] Binding errors are not reported to the developer
### Description
Is there an echo in here?

`Application.LogWarningsToApplicationOutput` was introduced in Xamarin Forms 3.2 as a `bool` flag to enable or disable logging to the developer when {Binding} expressions failed.

In XF 3.4, the release notes noted that it didn't work because of #3873: Debug calls get stripped from the XF release binaries.

A later PR was created and merged to expose a `LogWarningsListener`: #4539, and `Application.LogWarningsToApplicationOutput` was marked obsolete.

Unfortunately, it looks like some wires got crossed, and the API mentioned in the above PR was never actually introduced, and there is no `LogWarningsListener` property on `Application`.

So, as it stands, there is no "official" way for a developer to enable logging of Binding failures.

There IS the old hacky workaround way, which no one should rely on, but nevertheless works:

```C#
Xamarin.Forms.Internals.Log.Listeners.Add(new DelegateLogListener((arg1, arg2) => Debug.WriteLine(arg2)));
```

Tagging @jfversluis, because we worked on this, and you might remember some of the details here that I've forgotten.

### Steps to Reproduce

1. Try to use  `Application.LogWarningsToApplicationOutput`
2. Observe deprecation message
3. Attempt to find LogWarningsListener
4. 🤔

### Expected Behavior

Either `Application.LogWarningsToApplicationOutput` works, or `LogWarningsListener` exists.

### Actual Behavior

Neither of these!

### Basic Information

- Version with issue: Xamarin.Forms 4.2.0.848062
- Last known good version: I'm not sure this _ever_ worked quite right.
- IDE: Visual Studio 2019
- Nuget Packages: N/A
- Affected Devices: N/A

jfversluis:
### Description of Change ###

Adds the `Application.LogWarningsListener` to be able to capture output from the internal XamForms logging

### Issues Resolved ### 

- fixes #8119

### API Changes ###

Added:
 - `static Action<(string Category, string Message)> Application.LogWarningsListener`

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###

Hook into this new API by, for example, adding this line to your `App.xaml.cs`: 

`Application.LogWarningsListener = (warning) => Debug.WriteLine($"{warning.Category}: {warning.Message}");`

Now whenever a warning is raised from within Xamarin.Forms this will be outputted to the debug console. This happens when a binding could not be resolved. Add a binding to a property that does not exist, run the application and observe that you will be notified of the binding not working.

image

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Application.cs
+++ b/Xamarin.Forms.Core/Application.cs
@@ -38,6 +38,9 @@ namespace Xamarin.Forms
 				if (CurrentLogListener != null && Log.Listeners.Contains(CurrentLogListener))
 					Log.Listeners.Remove(CurrentLogListener);
 
+				if (value == null)
+					return;
+
 				CurrentLogWarningListener = value;
 				CurrentLogListener = new DelegateLogListener((category, message) => CurrentLogWarningListener.Invoke((category, message)));
 
--- a/Xamarin.Forms.Core/Application.cs
+++ b/Xamarin.Forms.Core/Application.cs
@@ -26,6 +26,25 @@ namespace Xamarin.Forms
 		[Obsolete("Assign the LogWarningsListener")]
 		public static bool LogWarningsToApplicationOutput { get; set; }
 
+		private static Action<(string Category, string Message)> CurrentLogWarningListener;
+		private static DelegateLogListener CurrentLogListener;
+
+		public static Action<(string Category, string Message)> LogWarningsListener
+		{
+			get => CurrentLogWarningListener;
+
+			set
+			{
+				if (CurrentLogListener != null && Log.Listeners.Contains(CurrentLogListener))
+					Log.Listeners.Remove(CurrentLogListener);
+
+				CurrentLogWarningListener = value;
+				CurrentLogListener = new DelegateLogListener((category, message) => CurrentLogWarningListener.Invoke((category, message)));
+
+				Log.Listeners.Add(CurrentLogListener);
+			}
+		}
+
 		public Application()
 		{
 			var f = false;


jfversluis:
This should have been introduced in #4539, but for some reason got lost. I've tried to recreate it with this PR.

I marked it regression and therefore targeted 4.3 because this functionality should've been in there, also stated by an obsolete message which is now giving false information. However, this does introduce a new API.

If a retarget is needed, please let me know!

Xamarin.Forms.Core/Application.cs
```diff
@@ -26,6 +26,25 @@ public class Application : Element, IResourcesProvider, IApplicationController,
...
+		{
+			get => CurrentLogWarningListener;
+
+			set
+			{
+				if (CurrentLogListener != null && Log.Listeners.Contains(CurrentLogListener))
+					Log.Listeners.Remove(CurrentLogListener);
+
+				CurrentLogWarningListener = value;
+				CurrentLogListener = new DelegateLogListener((category, message) => CurrentLogWarningListener.Invoke((category, message)));
```
hartez: What happens if I set `LogWarningsListener` to `null`?

jfversluis: Ha! Nice. Just tested it, nothing really did happen. Added a check for it anyway.

samhouts:
I'm going to close this one for now until we're ready to return to it.

PR #8144, opened 2019-10-22
Add VSM Setter TargetName

### LINKED ISSUE #4924: [F100] Add "Target" support to VisualStateManager

### LINKED ISSUE #5622: CollectionView SelectedItem Color

jfversluis:
### Description of Change ###

Adds the ability to target a child object by name for a Setter for visual state management or styles.

### Issues Resolved ### 

- fixes #4924 
- closes #5034
- closes #5622

### API Changes ###

Added:
 - `public string TargetName { get; set; }`

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###

I've added a core gallery page `VisualStateSetterTarget.xaml` which has a button to toggle the VSM state of the `StackLayout`. When it's pressed, the `ALabel` is also triggered because of it's reference through `TargetName`.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Setter.cs
+++ b/Xamarin.Forms.Core/Setter.cs
@@ -14,17 +14,6 @@ namespace Xamarin.Forms
 	{
 		readonly ConditionalWeakTable<BindableObject, object> _originalValues = new ConditionalWeakTable<BindableObject, object>();
 
-		/// <summary>
-		///		Set the target for the property to be changed.  Can be a <c>x:Reference</c> to a <see cref="BindableObject"/>
-		///		or the Name of an object in the attached scope.
-		/// </summary>
-		/// <example>
-		/// The following example shows the two ways of targeting a named object.
-		/// <code>
-		///		<Setter Target="{x:Reference TargetLabel1}" Property="Label.TextColor" Value="#94b0b7" />
-		///		<Setter Target="TargetLabel1" Property="Label.TextColor" Value="#94b0b7" />
-		/// </code>
-		/// </example>
 		public object Target { get; set; }
 
 		public BindableProperty Property { get; set; }
--- a/Xamarin.Forms.Core/Setter.cs
+++ b/Xamarin.Forms.Core/Setter.cs
@@ -14,7 +14,18 @@ namespace Xamarin.Forms
 	{
 		readonly ConditionalWeakTable<BindableObject, object> _originalValues = new ConditionalWeakTable<BindableObject, object>();
 
-		public string TargetName { get; set; }
+		/// <summary>
+		///		Set the target for the property to be changed.  Can be a <c>x:Reference</c> to a <see cref="BindableObject"/>
+		///		or the Name of an object in the attached scope.
+		/// </summary>
+		/// <example>
+		/// The following example shows the two ways of targeting a named object.
+		/// <code>
+		///		<Setter Target="{x:Reference TargetLabel1}" Property="Label.TextColor" Value="#94b0b7" />
+		///		<Setter Target="TargetLabel1" Property="Label.TextColor" Value="#94b0b7" />
+		/// </code>
+		/// </example>
+		public object Target { get; set; }
 
 		public BindableProperty Property { get; set; }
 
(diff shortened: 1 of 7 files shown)


Xamarin.Forms.Core/Setter.cs
```diff
@@ -47,59 +48,73 @@ object IValueProvider.ProvideValue(IServiceProvider serviceProvider)
 			return this;
 		}
 
-		internal void Apply(BindableObject target, bool fromStyle = false)
+		internal void Apply(BindableObject scope, bool fromStyle = false)
```
StephaneDelcroix: the name was correct

Xamarin.Forms.Core/Setter.cs
```diff
@@ -47,59 +48,73 @@ object IValueProvider.ProvideValue(IServiceProvider serviceProvider)
...
 		{
-			if (target == null)
-				throw new ArgumentNullException(nameof(target));
+			var targetObject = FindTargetObject(scope, TargetName);
```
StephaneDelcroix: I think the FindTargetObject doesn't bring enough value by itself
```csharp
if (target == null) throw...;
if (!string.IsNullOrEmpty(TargetName) && target is Element element)
    target == element.FindByName(TargetName) as BindableObject ?? throw ...;
```

jfversluis: I think I implemented the way you intended, if I missed anything please let me know :)

samhouts:
please target 4.4.0

jfversluis:
🤦‍♂ done!

mattleibow:
Does this support property paths? For example:

```
<Setter TargetName="ItemGrid" Property="ColumnDefinitions[2].Width" Value="0" />
```

I have an app that actually does this: https://gist.github.com/mattleibow/2f2f33c353be0b33e31bf9007095b105#file-mainpage-xaml-L35-L41

jfversluis:
> Does this support property paths? For example:
> 
> ```
> <Setter TargetName="ItemGrid" Property="ColumnDefinitions[2].Width" Value="0" />
> ```
> 
> I have an app that actually does this: https://gist.github.com/mattleibow/2f2f33c353be0b33e31bf9007095b105#file-mainpage-xaml-L35-L41

That is pretty awesome :) 

Just tried it with this PR and that doesn't work. Would it make sense to get this in first as a start and then build from there?

mattleibow:
That seems like something g that we can do still. I can always name things a bit more for now. Good work so far.

Gonna be nice to write less VSM per element :)

PR #8147, opened 2019-10-22
Fix UWP Toolbar icons

### LINKED ISSUE #7505: [Bug] Icons from Toolbaritem are not displayed on UWP if starts on second monitor
### Description

The icons from ToolbarItem elements are not displayed if application starts on second monitor.

### Steps to Reproduce

1. Create tabbed page project and choose only UWP
2. Add the line inside `<TabbedPage ...` node

```
<TabbedPage.ToolbarItems>
    <ToolbarItem Order="Primary" Text="Tab 1" Icon="{x:OnPlatform iOS='sync.png', UWP='Assets/LargeTile.scale-100.png'}" />
    <ToolbarItem Order="Primary" Text="Tab 2" Icon="{x:OnPlatform iOS='settings.png', UWP='Assets/LargeTile.scale-200.png'}"/>
  </TabbedPage.ToolbarItems>
```

3. Run application to second monitor

Attach project:

https://1drv.ms/u/s!AgtERRLVkXN4hLdai20VKSiboOSTkw?e=nL2fJC

### Expected Behavior
Expect to see icons. If you use navigation page, then icons are displayed back.

### Basic Information

- Version with issue: Xamarin.Forms 3.4.0.1008975 . It happens in 4.2.0.778463 too
- IDE: VS 2017
- Platform Target Frameworks:
  - UWP:  16299 
- Affected Devices: Windows laptop

### Screenshots

image

no_icon_toolbaritem

jfversluis:
### Description of Change ###

Change the icons on UWP ToolbarItem to map to `AppBarButton.Content` and add an `Image` in there instead of using the `AppBarButton.Icon` directly because of a bug in UWP.

### Issues Resolved ### 

- fixes #7505

### API Changes ###

 
 None

### Platforms Affected ### 

- UWP

The toolbar icons will look a bit different and less "iconized". You can now show full colored icons.

If users are not doing so already, they can create monochromatic images to mimic the old behavior themselves.

As you can see in the screenshots below there is a slight difference in the look whenever the toolbar item is primary. This change also allows full color icons to be used.

You can also see a black square, that is the full color Xamarin logo. In the old way, this was not supported, in the new way it is. You can see while using the new way for a secondary item the full colored Xamarin logo is (still) a black square. To make an icon still show up when a toolbar item is secondary, we still need to assign it directly to `AppBarButton.Icon`. I have tested if the bug with the secondary monitor also affects items that are secondary, but from what I can tell it does not.

#### Original icons used in both screenshots below

bank
calculator

#### Old

Old way of showing icons

#### New

New way of showing icons

### Testing Procedure ###

Go to Issue7505 and inspect the toolbar icons. It will take a dual monitor setup (with different resolutions) to test if the icons are not disappearing anymore. To see the old vs new appearance you would have to comment some lines in the `Platform.cs` file. Or you can compare the looks from the secondary item icons and the primary ones.

### PR Checklist ###


Xamarin.Forms.Platform.UAP/ImageSourceConverter.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using Windows.UI.Xaml.Controls;
+
+namespace Xamarin.Forms.Platform.UWP
+{
+	internal class ImageSourceConverter : Windows.UI.Xaml.Data.IValueConverter
+	{
+		public object Convert(object value, Type targetType, object parameter, string language)
+		{
+			// TODO: investigate whether we can use AsyncValue<T> instead of blocking
```
hartez: We should either do this now or open up a separate GitHub issue to track this; otherwise, this TODO will sit here forever.

jfversluis: To be completely honest I just copy/pasted it from the other value converter that was previously being used and didn't even notice it. I'll see if I can fix it or create an issue for the both of them

jfversluis: Fixed for the "new" icon. For the "old" one I don't think it's possible, so removed the TODO.

jfversluis:
I think if we really want to (in a later stage) there are some APIs available within UWP to make an image monochromatic. Which comes of course at a performance cost. I've looked into making this possible with this change immediately, but using those APIs seemed to take another good but of effort which I'm not sure is worth it at this point.

The other way around is also true: I discovered that there is a property on the icons (that are used now) which lets you make them non-monochromatic. Maybe something we could expose as a platform specific or something.

PR #8169, opened 2019-10-23
Adds Cookies to the WebView fixing #Issue3262

CliffAgius:
### Description of Change ###

Added the ability for a WebView to use Cookies, this means that you can set Cookies and they will be passed along with the request to the website.

### API Changes ###
None

Added:
 - CookieContainer Cookies { get; set; } //Bindable Property

### Platforms Affected ### 

- iOS
- Android

None

Not applicable

### Testing Procedure ###
Issue3262 has a test view and also added a Test Method to the WebViewUnitTests.cs called TestSettingOfCookie() that checks the Cookie can be set and read correctly...

### PR Checklist ###


jsuarezruiz:
Hey @CliffAgius, thanks for the PR. It seems that you have some conflicts in addition to some changes that does not correspond to the PR (are related with the Stepper). Could you check it?.

CliffAgius:
Thanks @jsuarezruiz somehow I had made the changes to the WebView leaving in my other open PR for the Stepper.  I have solved it now so you should see there are no conflicts and the Stepper items are removed.

I must mention though that the conflicts are because I changed the Android build to allow x86/x86_64 so that I can use the local emulators is there a reason the main XF project only targets to 'armeabi_v7a'

EmbeddingTestBeds/XamarinFormsVersionFile.txt
```diff
@@ -0,0 +1 @@
+4.4.205+205-sha.1b06b1b34
```
StephaneDelcroix: do not commit those files

Xamarin.Forms.Core/WebView.cs
```diff
@@ -37,6 +38,8 @@ public class WebView : View, IWebViewController, IElementConfiguration<WebView>
 
 		public static readonly BindableProperty CanGoForwardProperty = CanGoForwardPropertyKey.BindableProperty;
 
+		public static readonly BindableProperty CookiesProperty = BindableProperty.Create("Cookies", typeof(CookieContainer),typeof(WebView),default(string));
```
StephaneDelcroix: use nameof(Cookies)

CliffAgius:
If the rest of this PR is Ok will take a look at doing the same for UWP.

rmarinho:
@CliffAgius yes looks good we just need UWP

CliffAgius:
@rmarinho thanks for the reminder I totally forgot about this request will take a look, I know I looked before just never got around to finishing it.

rmarinho:
Needs rebase i think

CliffAgius:
@rmarinho yes I have had a bit of an issue with the my fork of the project so just finished putting all that right (as you can see in the commits!), seems the rebase didn't solve it as the issue was with the dll files in the bin/obj folders that the rebase wasn't touching.  Anyway have it back together now must learn that when I have a cold not to try and play with Git.  Will hopefully have a UWP version soon.

rmarinho:
is the build working for you locally?

CliffAgius:
@rmarinho sorry been a bit busy the last few weeks flying rather than coding, the last push was by mistake I hadn't finished as I was struggling.  Turns out that the UWP WebView doesn't support cookies fully and there is is a GitHub proposal that is being looked at by MS at:

[https://github.com/microsoft/microsoft-ui-xaml/issues/1298]

Also you will notice that not only are they looking at fixing the issue but making it part of the move in UWP from IE to Edge (Chromium) with what they have named WebView2.

I have this latest push working with a bit of a hack which isn't brilliant if I'm honest but gets the job done for 80% of users.  I suggest that when the new WebView2 comes in for UWP and the Xamarin team complete the binding for this they then add in the Cookies correctly.

Take a look and let me know I have tested it here and it works for the tests I have done and as we know the Droid and IOS versions already work and have been tested.

Let me know if you need anything else to have this PR accepted be nice to cross this one off as done.

Thanks

Cliff.

Xamarin.Forms.Core/WebView.cs
CliffAgius: Done.

jfversluis:
Since you asked so nicely 😉, took a quick look at this one.

It seems there is an issue with the build though :(

rmarinho:
Hum something wrong here. You are pushing the bins and more stuff. Maybe branch again, and cherry pick your commits

rmarinho:
Still not there @CliffAgius

CliffAgius:
@rmarinho @jfversluis Ok managed to resolve the issue of why the build was failing, a while back due to issues I deleted all the Bin/OBJ files from my fork.  I rebuilt the project and it works locally but when it gets built by the PipeLine tooling the folder XFCorePostProcessor.Tasks is missing and this has DLL's required for the build to work.

This means that when I was pushing changes due to the .gitignore these files where not getting pushed up to the remote but are actually needed (I Believe!) and hence me putting them back in, my mistake was that I put all the Bin/Obj folders back rather than just this one.

Do you think as this folder is important the .gitignore file needs updating to make sure this folder is tracked?

rmarinho:
I think something is wrong you don't need to push the XFCorePostProcessor changes

CliffAgius:
> I think something is wrong you don't need to push the XFCorePostProcessor changes

But the build errors only reference these files and looking at my fork those were missing hence me putting them back in.  I can build and run here perfectly so unless we ship my machine I might need a little help. 🤔

Sadly as the Build failure isn't showing in the issue anymore I can't check what they were.

Sorry for being a pain and thanks for the help.

rmarinho:
No worries, we are here to help. 

If you checkout master, can you build fine? are you on win or mac ?

CliffAgius:
I'm on windows.

Yes Can build and run on both Master and this Issue3262 branches.

rmarinho:
So and master you don't have these XFCorePostProcessor changes right? if you checkout , build there shouldn't be no new checked in files.

What i would do is checkout master, create a new branch like issue3262_new, then cherrypick your commits for cookies changes , then  rename that branch to issue3262 and force push

rmarinho:
Hum still need to remove the extra files

CliffAgius:
> Hum still need to remove the extra files

@rmarinho yes sorry hoped I could do a rebase to fix it but sadly not. I am doing a Cherry-Pick now but UWP for me will no Load I am getting a weird error in VS19

C:\Users\*******\.nuget\packages\microsoft.ui.xaml\2.1.190606001\build\Microsoft.UI.Xaml.props(134,14): error MSB4184: The expression "[System.Version]::Parse('')" cannot be evaluated. Version string portion was too short or too long.

So I have done all the other parts and tested them but can't test the UWP version, if I revert the change made to the XF.Platform.UAP.csproj 18 days ago then it works but that means my PR will have that as a change hence me taking a while to figure this out.  I want to make sure it's my machine or not before filing a Bug report as it's being a bit screwy.

I am now doing this all on a separate branch so as not to keep bothering your with PR changes. 😉

rmarinho:
to fix that uwp issue to a `git clean -fdx` it's something old getting in the way

CliffAgius:
@rmarinho thanks for the help yesterday turns out it was my laptops .nuget cache so I deleted it all and it's now solved and working.  Good news also is that I found a bug in the UWP/IOS versions that would have given a NRE if there was no cookies which I have also fixed in this latest push.

I have it all running here with no issues on all 3 OS's so hoping that the build system likes it this time.

PR #8194, opened 2019-10-24
Fixed 7649 - [Enhancement] Added accessibility support for Top tabs in TabbedPage

### LINKED ISSUE #7649: [Enhancement] AutomationId for Android TabbedPage
## Summary
It would be great to have support for AutomationId for tabs in the TabbedPage on Android. At the moment, iOS has this support working [out of the box](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/TabbedRenderer.cs#L429). It seems as if the Android support was simply never implemented.

## API Changes
No visible API changes, just the AutomationId being mapped to the tab as it does on iOS now.

## Intended Use Case
This is super helpful for doing automated UI tests as you need the AutomationId (ContentDescription on Android) or some other unique identifier to 'tap' the button. We have had to implement a custom TabbedPageRenderer to get this working for now.

techduggu:
### Description of Change ###

As requested in the ticket #7649, added support for `AutomationId` for tabs in the TabbedPage on Android.

### Issues Resolved ### 

- fixes #7649

### API Changes ###

Added:
Below code is added to the Android `TabbedPageRenderer.cs` to support the `AutomationId` for tabs
```
if (!string.IsNullOrEmpty(child.AutomationId))
	tab.SetContentDescription(child.AutomationId);
```
Note: I've added the above condition in existing UpdateTabIcons() method to avoid any performance impact as we are already looping through all children (tabs) here.

### Platforms Affected ### 

- Android

None

Not applicable

### Testing Procedure ###

- Just run `TabbedPageTests.cs` from Control Gallery. I've added `AutomationId` for the tabs which gets assigned by the TabbedPageRenderer at runtime.

- I've also added a UI Test `TabbedPageAccessibilityTest` inside `TabbedPageTests.cs` which accesses the tabs through AutomationId.

image

### PR Checklist ###


techduggu:
Hi Team,

Have a quick question:

Android's `TabbedPageRenderer` has two types of implementation for Tabbed Page: Top & BottomNavigationView. This Pull request fixes the issue with AutomationId on top tabbed pages as it was a straight forward change and doesn't impact anywhere. But for BottomNavigationView tabs, it has a dependency with other renderers (`ShellItemRenderer`) on following two methods where we need to add the AutomationId support:

```
BottomNavigationViewUtils.SetupMenu()

BottomNavigationViewUtils.CreateMoreBottomSheet()
```

This will further require changes to the existing `CreateTabList()` method inside both renderers (TabbedPage & ShellItem) to include AutomationId as part of the tuple parameter so that it can be passed to the above two methods of `BottomNavigationViewUtils`. As this requires more changes and impacts other renderers too, I wasn't sure if we need to take care with this ticket. If it's fine, I can make the above changes too. Please share your feedback.
Thanks!

techduggu:
Hi @jfversluis ,

Thanks for taking a look into this. Yes, it makes sense to rename the `UpdateTabIcons` method to something like `UpdateTabIconsAndAutomationId` or `UpdateTabIconsAndContentDescription`. I will incorporate your review comments and will let you know soon.

Thanks!

techduggu:
Hi @jfversluis 
I've incorporated above review comments. 
Thanks!

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/TabbedPageTests.cs
```diff
@@ -21,6 +21,9 @@ namespace Xamarin.Forms.Controls.Issues
 	[Issue (IssueTracker.None, 0,"TabbedPage nav tests", PlatformAffected.All)]
 	public class TabbedPageTests : TestTabbedPage
 	{
+		const string _page1AutomationId = "Page 1";
```
samhouts: Make these different from the tab titles so we're sure it's automating on the right thing.

```suggestion
		const string _page1AutomationId = "Auto 1";
```

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/TabbedPageTests.cs
```diff
@@ -21,6 +21,9 @@ namespace Xamarin.Forms.Controls.Issues
...
+		const string _page2AutomationId = "Page 2";
```
samhouts:
```suggestion
		const string _page2AutomationId = "Auto 2";
```

Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs
```diff
@@ -679,6 +679,8 @@ void UpdateTabIcons()
 				Page child = Element.Children[i];
 				TabLayout.Tab tab = tabs.GetTabAt(i);
 				SetTabIconImageSource(child, tab);
+				if (!string.IsNullOrEmpty(child.AutomationId))
+					tab.SetContentDescription(child.AutomationId);
```
samhouts: If we alter `SetContentDescription` here, we're actually changing the text that TalkBack will announce. So users with the screen reader on will now here "Auto 1" instead of "Page 1" as expected. Not good!

jfversluis:
Added DNM after Samantha's observation on the linked issue. We should probably look into that one first. Good catch!

samhouts:

PR #8090, opened 2019-10-18
[Android] fix android scrollbar visibility issue

ravinderjangra:
### Description of Change ###

fixed Android scrollview scrollbar visibility issue.

### Issues Resolved ### 

- fixes #7629
- fixes #8015

### Platforms Affected ### 
- Android

### Before/After Screenshots ### 

**Before**
before

**After**
after scrollbar

### PR Checklist ###

- [ ] Targets the correct branch
- [ ] Tests are passing (or failures are unrelated)

PR #8196, opened 2019-10-24
[Android] Make sure we use the right Context

rmarinho:
Fix a regression caused by #8090 , still checking if this would be a issue in more places.
Basically everything that will be expecting a Context that s inside the ScrollViewer will have check if it's a themewrapper, the is special worrying if people want to grab the activity. 

### Description of Change ###

Check we pass the right context not the wrapper to the underlying controls. 
I m still investigating if other controls will suffer from the same when inside the ScrollView

### Issues Resolved ### 

- fixes broken API19 tests on master

### API Changes ###

 None

### Platforms Affected ### 

- Android

None

Not applicable

### Testing Procedure ###
You should be able to visit the SearchBar gallery

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/SearchBarRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/SearchBarRenderer.cs
@@ -10,6 +10,7 @@ using Android.Text;
 using Android.Text.Method;
 using Android.Util;
 using Android.Widget;
+using Android.Views;
 using AView = Android.Views.View;
 
 namespace Xamarin.Forms.Platform.Android
@@ -60,7 +61,8 @@ namespace Xamarin.Forms.Platform.Android
 
 		protected override SearchView CreateNativeControl()
 		{
-			return new SearchView(Context);
+			var context = (Context as ContextThemeWrapper).BaseContext ?? Context;
+			return new SearchView(context);
 		}
 
 		protected override void OnFocusChangeRequested(object sender, VisualElement.FocusRequestArgs e)


PureWeen:
@rmarinho

Visual Material also uses a ContextThemeWrapper 

If you use searchview inside a Frame set to Material does it throw the same exception?

Xamarin.Forms.Platform.Android/Renderers/SearchBarRenderer.cs
```diff
@@ -60,7 +61,8 @@ public override SizeRequest GetDesiredSize(int widthConstraint, int heightConstr
 
 		protected override SearchView CreateNativeControl()
 		{
-			return new SearchView(Context);
+			var context = (Context as ContextThemeWrapper).BaseContext ?? Context;
```
PureWeen: - If this is only broken on APIi 19 then shouldn't we only do this if it's running on an API 19 device?
- I'm still curious if this also breaks if you have a Material Frame with a SearchView. If it doesn't than I wonder if the styles are setup wrong for the scrollview?
- if we end up going this route then I think we'd want to use this extension method . https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/ContextExtensions.cs#L101

rmarinho: - I think we could have a check
- It doesn't show the issue if i use material frame with a scrollview.
- It fixes the issue (crash) if i use the ThemeContext, but applies the theme to the SearchBar, for example on the gallery the SearchBar will now have a white text color i think that will be a breaking change

rmarinho:
@PureWeen going merge this, but we should sync up to talk about this when you are back from vacations.

Issue #4875, opened 2018-12-31
Height calculations for content within a layout not calculation correctly for wrapping content like flex and text


pizzahead24:
### Description
Height calculations for content within a layout not calculation correctly for wrapping content like flex and text

### Steps to Reproduce

https://forums.xamarin.com/discussion/145900/long-label-with-word-wrap-doesnt-seem-to-auto-calculate-height-correctly#latest

To product the full blown error:

1.) Create a flex layout. 
2.) Inside the layout create several other stack layouts Horizontal with title and value labels inside.  Create 5 of them and make the Text of them very long (look at the forum link above).
3.) Now divide these content within so that when you grow and shrink the content will wrap into one column or two. What I did was split it up 3 - 3 left and right and have a flex basis on the left of 300. This should make the parent flex layout wrap the right side under once the width reaches 300.
4.) When you grow and expand on UWP you can see the text gets cut off. You can also see the layout straight up wont expand all the way because the wrapping text height for some reason wont notify the parent layout to expand that much more.

### Expected Behavior

The height of wrapping text and layouts should notify its parent to update its overall height and not cut things off.

### Actual Behavior

text and content are being cut off by the height because the wrapping features are not updating the parent layouts height.

### Basic Information

- Version with issue: Latest Xamarin Forms 3.3
- Affected Devices: All, but much easier to visibly watch the flex layout with UWP.

### Screenshots

Please check out the forum link above.

pizzahead24:
#4950 This is related

hamzaoubouzid:
hi @pizzahead24
thank for all Just i have one question about flexlayout in xamarin
i create design xml but when i am binding flexlayout width my list in c#
it not work and i serch and not get eney enswer about this probleme

Issue #1538, opened 2018-01-09
Items are not clipped properly when load the item inside the stack layout


DineshBabuYadav:
Hi Team,

### Description
We have come across an issue in **Xamarin.Forms.UWP** platform while using custom layouts and rendering it with native view. 

**Layout of Custom Layouts:**
•	In the sample, we have created a new **custom ContentView** derived from **ContentView** in which **StackLayout** has added labels as child element.
•	These **custom ContentView** has been added into the **custom Layout<View>** as children and added another label in the **Layout<View>**. And, the **Custom Layout<View>** is loaded into **custom ScrollView**. Finally, custom **ScrollView** loaded into a **ContentPage**.

**Hierarchical structure of sample:**

•	**ContentPage -> Custom ScrollView -> Custom Layout<View> -> Custom ContentView -> StackLayout -> Label (Issue : it doesn’t clip properly if native UWP view is created for Custom ContentView in renderer project)** 

**Replication of issue:**
•	The child elements such as labels in the **StackLayout (loaded in custom ContentView)** doesn’t get clipped **if we create an equivalent view for custom ContentView in UWP renderer project** as like below code example.

**Code Example in PCL Custom ContentView:**

![image](https://user-images.githubusercontent.com/20282188/34720619-63d8cd48-f565-11e7-809b-a049fbf37e0d.png)

**Code Example in UWP renderer Custom ContentView:**

![image](https://user-images.githubusercontent.com/20282188/34720663-872a9ce0-f565-11e7-9b54-c0ab98d2102e.png)

### Expected Behavior

Items are clipped properly when the content is greater than the item height.

### Actual Behavior

Items are not clipped properly, it overlaps on the other item.

### Basic Information

- Platform Target Frameworks: 
  - UWP
- Nuget Packages: XAMARIN.FORMS v.2.5.0.91635

### Screenshots

**_Before writing the renderer_**

![image](https://user-images.githubusercontent.com/20282188/34719655-97d11ca8-f561-11e7-8f99-084b169c2baf.png)

**_After writing the renderer_**
![image](https://user-images.githubusercontent.com/20282188/34719679-ac51c074-f561-11e7-8957-3570f07fc32c.png)

### Reproduction Link

[ClippingBug.zip](https://github.com/xamarin/Xamarin.Forms/files/1615042/ClippingBug.zip)

samhouts:
@DineshBabuYadav  Are you still experiencing this issue? Thanks!

[...]

PR #8214, opened 2019-10-24
[UWP] Attempt to resolve entry on UWP not correctly calculating the correct height when in a scroll view

### LINKED ISSUE #2172: Height of Entry with data binding incorrect on UWP when Entry in ScrollView in Grid

### Description
On complex pages on UWP, the height of Entry views can be incorrect, clipping the bottom of the text in the Entry. This occurs when the Entry uses data binding, and is in a Layout inside a ScrollView inside a Grid.

### Steps to Reproduce

Using the following code with Xamarin.Forms 2.4.0.38779, push an instance of the BugUWPEntryBindingPageView page onto the navigation stack. Press the Button that appears at the top of the page and view the Entry that then appears. On UWP, the Entry is clipped.

Then, to see a nasty workaround, change the WorkaroundBug boolean to true and run again.

using System.Threading.Tasks;

using Xamarin.Forms;

namespace ViewsUsingXamarinForms
{
    public class BugUWPEntryBindingViewModel : BindableObject
    {
        public static readonly BindableProperty ValueOneProperty
            = BindableProperty.Create(
                nameof(ValueOne),
                typeof(string),
                typeof(BugUWPEntryBindingViewModel),
                default(string));

        public static readonly BindableProperty ValueTwoProperty
            = BindableProperty.Create(
                nameof(ValueTwo),
                typeof(string),
                typeof(BugUWPEntryBindingViewModel),
                default(string));

        public BugUWPEntryBindingViewModel()
        {
            ValueTwo = string.Empty;
            ValueOne = "123";
        }

        public string ValueOne
        {
            get => (string) GetValue(ValueOneProperty);

            set
            {
                if (ValueOne != value)
                    SetValue(ValueOneProperty, value);
            }
        }

        public string ValueTwo
        {
            get => (string) GetValue(ValueTwoProperty);

            set
            {
                if (ValueTwo != value)
                    SetValue(ValueTwoProperty, value);
            }
        }

    } // public class BugUWPEntryBindingViewModel

    public class BugUWPEntryBindingPageView : ContentPage
    {
        private static readonly BugUWPEntryBindingViewModel ViewModel = new BugUWPEntryBindingViewModel();

        private Grid _grid;

        {
            _grid = new Grid
            {
                BackgroundColor = Color.White,
                HorizontalOptions = LayoutOptions.Fill,
                VerticalOptions = LayoutOptions.FillAndExpand,
                ColumnDefinitions = new ColumnDefinitionCollection
                {
                    new ColumnDefinition { Width = GridLength.Star },
                },
                RowDefinitions = new RowDefinitionCollection
                {
                    new RowDefinition {Height = GridLength.Auto },
                    new RowDefinition {Height = GridLength.Star }
                },
            };

            _grid.Children.Add(
                new Button
                {
                    Text = "Add/replace entries",
                    BackgroundColor = Color.Blue,
                    TextColor = Color.White,
                    Command = new Command(OnAddEntries)
                }, 0, 1, 0, 1);

            BindingContext = ViewModel;
            Content = _grid;
        }

        private const bool WorkaroundBug = false;
        private void OnAddEntries()
        {
            string text;
            if (WorkaroundBug)
            {
                text = ViewModel.ValueOne;
                ViewModel.ValueOne = string.Empty;
            }

            Entry entry = new Entry
            {
                FontSize = Device.GetNamedSize(NamedSize.Medium, typeof(Entry)),
            };
            entry.SetBinding(
                Entry.TextProperty,
                new Binding(
                    nameof(BugUWPEntryBindingViewModel.ValueOne),
                    BindingMode.TwoWay,
                    null,
                    string.Empty));
            _grid.Children.Add(
                new ScrollView
                {
                    Content = new StackLayout
                    {
                        Children =
                        {
                            entry
                        }
                    }
                }, 0, 1, 1, 2);

            if (WorkaroundBug)
            {
                Task.Run(async () =>
                {
                    await Task.Delay(1);
                    Device.BeginInvokeOnMainThread(() =>
                    {
                        ViewModel.ValueOne = text;
                    });
                });
            }
        }
    }
}

### Expected Behavior
Expect to see "123"

### Actual Behavior
Expect to see "123" but with bottom of text clipped.

### Basic Information

- Version with issue: 2.4.0.38779
- IDE: VS2017
- Platform Target Frameworks: 
  - UWP:  14393
- Nuget Packages: XF 2.4.0.38779
- Affected Devices: Windows 10 x64 bit desktop

### Screenshots
entry

### LINKED ISSUE #5711: Layout issue with Entry
### Description
We are creating a custom control which has a hierarchy TemplatedView --> StackLayout --> Grid --> Entry. We are adding a row with height zero above the row (Auto height) with Entry. While doing this, the height of the custom view is not rendered properly when the Entry is added with text in load time. Hence to avoid this, we are setting string.Empty for Entry text in OnElementChanged() and updating the actual text and setting padding for native TextBox in TextBox loaded event in custom renderer. 

When this custom view is added as “FlexLayout --> ScrollView --> Grid --> CustomTemplatedView”, Entry text not displayed in UWP until we resize the window.

### Expected Behavior

Entry text should be displayed

### Screenshots

https://github.com/divi1411/TextInputLayout/raw/master/Screenshot/Screenshot.png

### Reproduction Link

https://github.com/divi1411/TextInputLayout.git

bmacombe:
### Description of Change ###
It seems that when the UWP textbox is in a visual tree that includes a scroll view it seems to mis-measures the height of the text box.  This change will set the native text box text property to string.empty while measuring which seems to result in the correct height being calculated.  Once that is calculated, it sets the text value back from the XF element.

Used the method of using a static TextBox to do the measure that was already in use in the EditorRenderer to handle autosize. I pulled that code out and modified it, adding it as a static method on FormsTextBox. I then updated the EntryRenderer and EditorRenderer to use this code for measuring their size which corrects the issue.

To help validate I recreated the previous behavior in custom renderers (in the UWP Control Gallery) to show the previous behavior and the new behavior side by side on the issue page.

### Issues Resolved ### 

- fixes #2172 
- fixes #5711

### API Changes ### 
 None

### Platforms Affected ### 

- UWP

Correctly calculates the height of a text box in a scroll view.

Top entry using the existing renderer
Buttom entry using the experimental renderer
image

### Testing Procedure ###
Open the issue page and view the entries

### PR Checklist ###


bmacombe:
This PR needs to be finalized, looking for feedback from the XF team and the community before making final changes.

bmacombe:
This solution likely won't work, after more review I see unintended events being fired by changing the text value.  The EntryRenderer updates the XF Entry when the native Textbox text changes.

bmacombe:
I wonder if this issue is related to this comment in VisualElementRenderer.

``` c#
// Disabled until reason for crashes with unhandled exceptions is discovered
// Without this some layouts may end up with improper sizes, however their children
// will position correctly
//Loaded += (sender, args) => {
if (Packager != null)
	Packager.Load();
//};
```

bmacombe:
It does appear to be related.  Calling Packager.Load() in the Loaded event in SetElement line 150 in visual state renderer resolves the issue.  But that comment appears to be from Jason Smith.  Thoughts from the XF team with more knowledge on how this process works would be appreciated.

``` c#
// Disabled until reason for crashes with unhandled exceptions is discovered
// Without this some layouts may end up with improper sizes, however their children
// will position correctly
Loaded += (sender, args) =>
{
	if (Packager != null)
		Packager.Load();
};

```

image

bmacombe:
Last observation.  I tried stepping through the layout logic and seeing if I could force the items hosted by the ScrollViewer to update their layout, but I noticed that all the UWP LayoutUpdate, Loaded, etc events were being called before the first the requested size of the entry, so they had no effect.

I then refocused on the GetDesiredSize method of VisualElementRender and after just some random experimentation I noticed that heightConstraint was double.Infinity, which makes sense inside a ScrollViewer.  But I don't think the UWP layout logic likes this.

Changing the heightConstraint to 500,000 (still effectively infinity as far as layout goes), caused the correct height of the FormsTextBox to be calculated.

```c#
if (double.IsInfinity(heightConstraint))
	heightConstraint = 500000;
return base.GetDesiredSize(widthConstraint, heightConstraint);
```

This "Test Renderer" entry is using a renderer that contains the code above in it's GetDesiredSize override.

image

@samhouts Would it be possible to get a review on method?  This seems to be a reasonable solution as I think it would have little side effects.

bmacombe:
Well, scratch that one.  It works in the simple layout in the Issue page, but not in a more complex layout in my own app.  Hopefully these notes will help spur some thought in others.

bmacombe:
Another note, I see the UWP TextBox control template also includes a ScrollViewer. Maybe an issue?

```
<ScrollViewer x:Name="ContentElement" AutomationProperties.AccessibilityView="Raw" Grid.Column="0" HorizontalScrollBarVisibility="{TemplateBinding ScrollViewer.HorizontalScrollBarVisibility}" HorizontalScrollMode="{TemplateBinding ScrollViewer.HorizontalScrollMode}" IsDeferredScrollingEnabled="{TemplateBinding ScrollViewer.IsDeferredScrollingEnabled}" IsHorizontalRailEnabled="{TemplateBinding ScrollViewer.IsHorizontalRailEnabled}" IsTabStop="False" IsVerticalRailEnabled="{TemplateBinding ScrollViewer.IsVerticalRailEnabled}" Margin="{TemplateBinding BorderThickness}" Padding="{TemplateBinding Padding}" Grid.Row="1" VerticalScrollMode="{TemplateBinding ScrollViewer.VerticalScrollMode}" VerticalScrollBarVisibility="{TemplateBinding ScrollViewer.VerticalScrollBarVisibility}" ZoomMode="Disabled"/>
```

jfversluis:
Hi @bmacombe thank you for your ongoing, extensive research and writeup on this. We really appreciate it!

I see you asking some feedback and things from time to time, but it's hard to make it concrete since you are progressing on your own 😄 

The only thing I can see is to not pay too much attention to the comments in there. They might be pretty old and not relevant anymore. Just trust your own judgement on how to solve this best and of course keeping in mind that other controls and scenarios should just remain working. Did you by any chance get the UI tests running for UWP? That might be a good indication to go on.

Is there anything concrete you want us to look at at this time?

bmacombe:
@jfversluis I think you gave me what I need, Thanks!...I think the comment just made me nervous and unfortunately predates the import into GitHub so I haven't been able to track it back to an issue or find something in Bugzilla.  

I think the right solution is to have the Packager.Load happen after the layout control has been loaded.  I really think the issue is timing of when the various native UWP elements are created and asked to measure their size.  Everything else seems way to much of hack...I might be willing to risk that in some app code...but not in framework code.

So I'll put Packager.Load back into the Loaded event and look into running the UWP UI tests.  Just makes me nervous messing with such a root level item in the UWP platform. 😅

bmacombe:
Just another observation today in my actual app.  These entries are in a SyncFusion SfTextInputLayout.  Just interesting that they size correctly when the window width changes (not height) but they get the size error again when the window goes back to it's original size

HeightOnSize

bmacombe:
Seems that entrys calculate the correct height if the scrollview is the root layout.  It's only when it's nested instead another layout that is causes an issue.

bmacombe:
Considering a new approach, I saw some comments in the UWP EditorRenderer about using a cloned FormsTextBox that is never added to the visual tree for size measurement.  Since it seems the TextBox doesn't correctly measure it's height while in a visual tree in some cases, having a static TextBox that duplicates the actual renderers TextBox, that is just used to get the desired size.

Will look into this method and how to ensure it consumes as little resources as possible.

bmacombe:
The above method appears to do the trick.  Needs to be applied to both Entry and Editor renderers.  Will cleanup and push new code soon.

Before
image

After
image

bmacombe:
@samhouts This is ready for review, I think I have a final solution.

Took the method of using a static TextBox to do the measure that was already in use in the EditorRenderer.  I pulled that code out and modified it, adding it as a static method on FormsTextBox.  I then updated the EntryRenderer and EditorRenderer to use this code for measuring their size which corrects the issue.

To help validate I recreated the previous behavior in custom renderers to show the previous behavior and the new behavior side by side on the issue page.

bmacombe:
Likely not fixed by this PR, but could be a similar issue
 #4875
 #1538

bmacombe:
I also verified this fix works in my app when I applied this fix via some custom renderers, including SyncFusion's control (though they will have to update their renderers since they don't inherit from the XF renderers)

image

Xamarin.Forms.ControlGallery.WindowsUniversal/CustomRenderers.cs
```diff
@@ -163,4 +167,107 @@ protected override void OnElementChanged(ElementChangedEventArgs<BoxView> e)
...
+			child.Measure(constraint);
+			var result = new Size(Math.Ceiling(child.DesiredSize.Width), Math.Ceiling(child.DesiredSize.Height));
+
+			return new SizeRequest(result);
+		}
+	}
+
+	public class Issue2172OldEditorRenderer : EditorRenderer
+	{
+		private static FormsTextBox _copyOfTextBox;
```
jfversluis:
```suggestion
		static FormsTextBox _copyOfTextBox;
```

bmacombe:
Thanks @jfversluis!  Thought the background would be good for reviewers...but also helpful for myself to remember what I had done as I was able to find time to work on it.  I will rebase to fix the conflict hopefully yet today.

bmacombe:
@PureWeen Would you run the UWP UITests if you feel they are needed on this one since I couldn't get them going right on my machine?

bmacombe:
@samhouts  @jfversluis Rebased to resolve conflict

bmacombe:
#8536 may be attempting to resolve a similar issue, but it appears each of these PRs resolve one aspect of the issue, but when combined do not.

See https://github.com/xamarin/Xamarin.Forms/issues/8787#issuecomment-562975014

bmacombe:
I know how to run them 😁 but they don't run right on machine...tried various versions of the driver etc.  Not all the text gets filled in the various search boxes to pull up the tests.  I had a little better luck if I changed the control gallery search bars to text boxes...but still had issues.

If anyone has some other suggestions to try I'd be glad to try them again.

PureWeen:
@jfversluis running now

PureWeen:
@jfversluis This doesn't appear to break anything with the unit tests

Is the removal of this file intentional
https://github.com/xamarin/Xamarin.Forms/pull/8214/files#diff-9d8f28a74e9d78f5c7af9201bf8f98bbL1

bmacombe:
@PureWeen I removed it because I accidentally added when I accidentally staged and checked in those resharper settings.  If there was already one there, I didn't realize it.

bmacombe:
I'll try to reverse that...new git skill to learn 😁

bmacombe:
@PureWeen @jfversluis I believe have I reversed my resharper file delete...sorry about that!

jfversluis:
It appears so, great! Thanks!

Issue #5807, opened 2019-04-04
[Visual] [iOS] Material entry text caret miss aligned


deja135:
### Description
When using an entry with HorizontalTextAlignment="Center"  the caret is slightly off to the left, and when you start typing in the field it jumps in to the midle.
### Steps to Reproduce

1. Create a entry with visual="material" and HorzontalTextAlignment="Center"

### Expected Behavior
Caret should be in the middle
### Actual Behavior
Caret is slightly off to the left

### Basic Information

- Version with issue: 3.6.0.264807
- Last known good version:
- IDE: 16.0.0

### Screenshots
![image](https://user-images.githubusercontent.com/22226357/55541305-0ab9b700-56c5-11e9-9604-ff222b1b5b2e.png)

henriqueu96:
I could not reproduce
![Screen Shot 2019-10-11 at 15 31 02](https://user-images.githubusercontent.com/14217887/66675681-32717500-ec3c-11e9-859a-4252cf2697f9.png)

PR #8236, opened 2019-10-25
Added test case for issue 5807

basdecort:
### Description of Change ###

Added test case for issue #5807.

### Issues Resolved ### 
- Added test case for issue.

### API Changes ###
None

### Platforms Affected ### 
- iOS

None

Simulator Screen Shot - iPhone 11 - 2019-10-25 at 19 48 32

### Testing Procedure ###
1. Enter text in the Entry
2. Text isn't centered (as it should be)

### PR Checklist ###


samhouts:
@basdecort Is there a bug fix that needs to go along with this test case? Thanks!

basdecort:
@jsuarezruiz great suggestion! I've added the label.

@samhouts after creating the repro, I've done some additional investigation. It seems like the issue is not in Xamarin.Forms, but in the Material component:
Simulator Screen Shot - iPhone 11 Pro Max - 2019-10-25 at 19 21 17

I've used the sample code on [their GitHub Page](https://github.com/material-components/material-components-ios/tree/develop/demos) to replicate the issue.

PR #8245, opened 2019-10-25
Anticipate calls into Android OS during startup

kingces95:
### Description of Change ### 

Use a thread pool to anticipate calls into the Android OS that the UIThread will make during startup; Use a thread pool thread to race ahead of the UIThread to place calls and cache results to remove that burden from the UIThread and thereby speedup startup.

Calling into the Android OS off the UIThread is a "gray" operation. We know to only modify the UI on the UIThread but what about getting the SDK version? The latter is likely ok and this change demonstrates how that type of call would be made. The particular anticipation in this PR is unlikely to speedup startup by an appreciable degree. That's not the point of the this PR. This PR is meant to demonstrate the pattern for future anticipations.

Doing any work off the UIThread at startup makes the startup code re-entrant. It would be a maintenance nightmare to make arbitrary startup code re-entrant. Doing so would require future developers to continually ask if the code they are working on is re-entrant and it may not be easy to discover if that's the case. To avoid this we isolate all re-entrant code in anticipator.cs and disallow scheduling arbitrary callback using the thread pool. 

### Platforms Affected ### 

- Android

None

Just for reference:
Screenshot_2019-10-25-15-08-33

### Testing Procedure ###

Every UITests stresses startup optimizations. 

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Anticipator.cs
+++ b/Xamarin.Forms.Platform.Android/Anticipator.cs
@@ -2,10 +2,59 @@
 using System.Collections.Concurrent;
 using System.Runtime.CompilerServices;
 using System.Threading;
+using ABuildVersionCodes = Android.OS.BuildVersionCodes;
+using ABuild = Android.OS.Build;
 
 namespace Xamarin.Forms.Platform.Android
 {
-	internal class Anticipator
+	/// <summary>
+	/// 
+	/// Anticipator is a hand-rolled threadpool that exists to speedup startup by activating
+	/// threads that can be used to race ahead of the UIThread in order to compute and cache results 
+	/// that the UIThread would otherwise have to compute as part of startup. This requires
+	/// making some startup code re-entrant. 
+	/// 
+	/// So developers don't need to wonder if the startup code they're writing is re-entrant we 
+	/// isolate all re-entrant code here in the static members of Anticipator.
+	/// 
+	/// Computing the results that need to be cached often require calling into the Android OS. 
+	/// Calling into the Android OS off the UIThread is a "gray" operation. E.g. we know not to 
+	/// update UI elements off the UIThread, but what about getting the SdkInt version? Likely
+	/// ok but just the same we want to track ALL Android OS APIs we call off the UIThread.
+	/// Isolating all re-entrant code in static Anticipator members simplifies accounting of
+	/// all Android OS calls potentially made off the UIThread.
+	/// 
+	/// </summary>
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.Android/Anticipator.cs
```diff
@@ -2,10 +2,59 @@
...
+			s_singleton.AnticipateGetter(() => Forms.SdkInt);
+		}
+
+		internal static ABuildVersionCodes SdkInt
+		{
+			get
+			{
+				if (!s_sdkInt.HasValue)
+					s_sdkInt = ABuild.VERSION.SdkInt;
+				return (ABuildVersionCodes)s_sdkInt;
```
kingces95: This code is re-entrant.

Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs
```diff
@@ -67,10 +68,7 @@ protected FormsAppCompatActivity()
 			_previousState = AndroidApplicationLifecycleState.Uninitialized;
 			_currentState = AndroidApplicationLifecycleState.Uninitialized;
 			PopupManager.Subscribe(this);
-
-			var anticipator = new Anticipator();
-			anticipator.AnticipateClassConstruction(typeof(Resource.Layout));
-			anticipator.AnticipateClassConstruction(typeof(Resource.Attribute));
+			RuntimeHelpers.RunClassConstructor(typeof(Anticipator).TypeHandle);
```
kingces95: This spins up the thread pool and sets its threads racing ahead of the UIThread to compute and cache results needed during startup.

Xamarin.Forms.Platform.Android/Forms.cs
```diff
@@ -72,13 +71,7 @@ public static class Forms
...
 		public static Color ColorButtonNormalOverride { get; set; }
 
-		internal static BuildVersionCodes SdkInt {
-			get {
-				if (!s_sdkInt.HasValue)
-					s_sdkInt = Build.VERSION.SdkInt;
-				return (BuildVersionCodes)s_sdkInt;
-			}
-		}
+		internal static BuildVersionCodes SdkInt => Anticipator.SdkInt;
```
kingces95: The UIThread and the thread pool threads will race to place this call. Usually the thread pool thread will win and the UIThread will be saved the effort of crossing into the android OS get get the sdk version.

PR #8259, opened 2019-10-27
Fixed 8176 - Shell BackButtonBehavior Inconsistencies & controlling FlyoutBehavior

### LINKED ISSUE #5219: [Shell,Android] BackButtonBehavior enables FlyoutBehavior
### Description

When the `FlyoutBehavior="Disabled"` and then I add a `BackButtonBehavior` on a `ContentPage`, the flyout becomes active on Android.

Android: https://www.screencast.com/t/xqjM4PeTrEH

iOS: https://www.screencast.com/t/VpYNt6aRJsj3

### Steps to Reproduce

```
<Shell xmlns="http://xamarin.com/schemas/2014/forms" 
       xmlns:local="clr-namespace:VisualChallenge"
       RouteHost="companyname.com"
       RouteScheme="app"
       Route="VisualChallenge"
       FlyoutBehavior="Disabled"
       Title="VisualChallenge"
       x:Class="VisualChallenge.AppShell"
       Visual="Material"
       >
```

```
             xmlns:sys="clr-namespace:System;assembly=mscorlib"
             x:Class="VisualChallenge.VisualChallengePage"
             Shell.NavBarIsVisible="true"
             Title="Edit Request"
             >
    
    <ContentPage.ToolbarItems>
        <ToolbarItem Text="Save"/>
    </ContentPage.ToolbarItems>
    
    <Shell.BackButtonBehavior>
        <BackButtonBehavior TextOverride="Undo" IconOverride="{OnPlatform Android=blanky}"></BackButtonBehavior>
    </Shell.BackButtonBehavior>
```

### Expected Behavior

Flyout doesn't become active.

### Basic Information

- Version with issue: 3.6-pre1
- Android 26
Parent: #2415

### LINKED ISSUE #8176: [Bug]  [Shell] Setting BackButtonBehavior IsEnabled to False inconsistent behavior
### Description
Setting BackButtonBehavior IsEnabled to False on view that is pushed onto the detail view of a MasterDetail windows behaves differently on Android and iOS.   Android displays the back button but tapping it does nothing.  iOS displays the back button and tapping it goes back.  

In Forms setting Navigation.SetHasBackButton(false) would not over-ride the hamburger / open master behavior of the left toolbar item.  I see no comparable way to do this with Shell.  I do understand that I could create my own Hamburger icon and code the command, but in the very least there should be a way to set the icon to the default hamburger icon used in the MasterDetail page.

The objective is that the Master/Detail navigation would look and behave exactly the same regardless of how many items are in the navigation stack of the detail window.

### Steps to Reproduce
1. Download the Xaminals test app or similar app using Shell
2. Add the following lines to one of the animal detail pages such as BearDetailPage.xaml
 ```
    <Shell.BackButtonBehavior >
         <BackButtonBehavior IsEnabled="False" />
     </Shell.BackButtonBehavior>
```
3. Run the app and navigate to the page, then see what is displayed in the left menu bar position.  Tap on it and note it's behavior.

### Expected Behavior
In keeping with Forms, I would expect to see the hamburger icon and tapping on it would open the Master view of the Master/detail page.
If the intention is really to disable the left toolbar item entirely, then the icon and text should not be displayed, or at least there should be an option to turn them off such as IsVisible.

### Actual Behavior
Android displays the back button but tapping it does nothing.  
iOS displays the back button and tapping it goes back.

### Basic Information
- Version with issue: 4.3.0

### PR DESCRIPTION
**Pull request by techduggu**
### Description of Change ###

As mentioned in the issue #8176 & #5219 (Possibly), I've noticed that `BackButtonBehavior` is not working as expected and it also overrides the `FlyoutBehavior` (tested with latest master branch).
As you can see in the attached 'Before' screenshot behavior, following are the issues:

-  On single Shell page (without navigation stack), when we set BackButtonBehavior `IsEnabled` to False, it disables the `FlyoutBehavior` too but hamburger icon exists (which is unexpected as BackButtonBehavior should not alter FlyoutBehavior at all).

- When a page is pushed to Navigation Stack of Shell (also described in issue #8176), when we set BackButtonBehavior `IsEnabled` to False, it displays the Back Button arrow but does nothing (which is not a good user experience as there should be a way to set the icon to the default hamburger icon used in the MasterDetail page of Shell).

As per my understanding (and more of proposal), the `BackButtonBehavior` should be applicable only on the pages pushed to Navigation Stack and should not alter the `FlyoutBehavior` at all. Also, if the `BackButtonBehavior` is disabled on the navigation pages, it should fallback to the FlyoutBehavior's current state  (i.e. it should display the default Flyout if it's enabled or hide if user has chosen it to be disabled) to maintain the user experience and keep things configurable. Let me know if it makes sense.

### Issues Resolved ### 

- fixes #8176 
- related to #5219 (Possibly as BackButtonBehavior's `IsEnabled` alters FlyoutBehavior too)

### API Changes ###

Changed:
 - Following logic has been updated inside `UpdateLeftBarButtonItem` method of `ShellToolbarTracker.cs` to handle the inconsistencies of Shell `BackButtonBehavior` (for both single and navigation pages).
```
if (command != null || CanNavigateBack)
{
	if (isEnabled)
	{
		_drawerToggle.DrawerIndicatorEnabled = false;
		toolbar.NavigationIcon = icon;
	}
	else
	{
		if (_flyoutBehavior != FlyoutBehavior.Flyout) //if Disabled or Locked
			_drawerToggle.DrawerIndicatorEnabled = false;
		else
			_drawerToggle.DrawerIndicatorEnabled = true;

		if (_flyoutBehavior != FlyoutBehavior.Disabled)
		{
			_drawerToggle.DrawerArrowDrawable = icon;
			toolbar.NavigationIcon = icon;
		}
	}
}
else if (_flyoutBehavior == FlyoutBehavior.Flyout)
{
	_drawerToggle.DrawerIndicatorEnabled = true;
	_drawerToggle.DrawerArrowDrawable = icon;
	toolbar.NavigationIcon = null;
}
else
{
	_drawerToggle.DrawerIndicatorEnabled = false;
}
```

### Platforms Affected ### 

- Android

As described above, after this change BackButtonBehavior will work as expected on both single and navigation pages.

**Before**:

8176_before

**After**:

8176_after

### Testing Procedure ###

- Just run `ShellBackButtonBehavior.cs` and `ShellFlyoutBehavior.cs` from Control Gallery as demonstrated in the above attached screen recordings.
- I've tested the following scenarios:

1. On single page of Shell, if BackButtonBehavior `IsEnabled` set to False, it will display default Flyout unless user has chosen it to be `FlyoutBehavior=Disabled`.
2. When a page is pushed to Navigation Stack, it will display Back button arrow by default (if default `BackButtonBehavior`).
3. If the BackButtonBehavior `IsEnabled` set to False on navigation page, it will display default Flyout unless user has chosen it to be `FlyoutBehavior=Disabled`.

As the above method is also responsible for overriding icons (with BackButtonBehavior), I've tested that it works as expected (overrides icons for both Flyout & Back button) but would like to propose that for Flyout, we should have separate properties to override icons and not use BackButtonBehavior to alter Flyout in any way. 

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/ShellBackButtonBehavior.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/ShellBackButtonBehavior.cs
@@ -112,7 +112,11 @@ namespace Xamarin.Forms.Controls.Issues
 					Command = new Command(ToggleIsEnabled),
 					AutomationId = ToggleIsEnabledId
 				});
-
+				layout.Children.Add(new Button()
+				{
+					Text = "Toggle Flyout Behavior",
+					Command = new Command(ToggleFlyoutBehavior),
+				});
 				layout.Children.Add(new Button()
 				{
 					Text = "Push Page",
@@ -177,6 +181,10 @@ namespace Xamarin.Forms.Controls.Issues
 			public void ToggleIsEnabled()
 			{
 				behavior.IsEnabled = !behavior.IsEnabled;
+			}
+			public void ToggleFlyoutBehavior()
+			{
+				Shell.Current.FlyoutBehavior = Shell.Current.FlyoutBehavior == FlyoutBehavior.Flyout ? FlyoutBehavior.Disabled : FlyoutBehavior.Flyout;
 			}
 		}
 
--- a/Xamarin.Forms.Platform.Android/Renderers/ShellToolbarTracker.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ShellToolbarTracker.cs
@@ -333,19 +333,30 @@ namespace Xamarin.Forms.Platform.Android
 
 			if (command != null || CanNavigateBack)
 			{
-				_drawerToggle.DrawerIndicatorEnabled = false;
-				toolbar.NavigationIcon = icon;
-			}
-			else if(_flyoutBehavior == FlyoutBehavior.Flyout)
-			{
-				_drawerToggle.DrawerIndicatorEnabled = isEnabled;
-				if(isEnabled)
+				if (isEnabled)
 				{
-					_drawerToggle.DrawerArrowDrawable = icon;
-					toolbar.NavigationIcon = null;
+					_drawerToggle.DrawerIndicatorEnabled = false;
+					toolbar.NavigationIcon = icon;
 				}
 				else
-					toolbar.NavigationIcon = icon;
+				{
+					if (_flyoutBehavior != FlyoutBehavior.Flyout) //if Disabled or Locked
+						_drawerToggle.DrawerIndicatorEnabled = false;
+					else
+						_drawerToggle.DrawerIndicatorEnabled = true;
+
+					if (_flyoutBehavior != FlyoutBehavior.Disabled)
+					{
+						_drawerToggle.DrawerArrowDrawable = icon;
+						toolbar.NavigationIcon = icon;
+					}
+				}
+			}
+			else if (_flyoutBehavior == FlyoutBehavior.Flyout)
+			{
+				_drawerToggle.DrawerIndicatorEnabled = true;
+				_drawerToggle.DrawerArrowDrawable = icon;
+				toolbar.NavigationIcon = null;
 			}
 			else
 			{

### CODE REVIEW

**Comment by rmarinho** [comment]
@techduggu you mind rebase master so we can run the build that was broken when you branched? 

Thanks

**Comment by techduggu** [comment]
@rmarinho Rebase master - Done!

**Comment by PureWeen** [comment]
> Also, if the BackButtonBehavior is disabled on the navigation pages, it should fallback to the FlyoutBehavior's current state (i.e. it should display the default Flyout if it's enabled or hide if user has chosen it to be disabled) to maintain the user experience and keep things configurable. Let me know if it makes sense.

I understand the intent but I think this is too much behavior to tie into BackButtonBehavior.IsEnabled. 

As @rmarinho said I think IsEnabled should just disable the backbutton like it currently does and if we want to reuse the FlyoutIcon across all the pages we should figure out a more direct API for that.  Right now IsEnabled is also tied to command so if the command CanExecute is false i don't want that to show the hamburger menu.

Maybe we can use the FlyoutBehavior at the page level so users could specify flyout behavior on every page? By default it'll be disabled on pushed pages but you can change that to something else per page? Then maybe this would let people add a backbutton and hamburger? Then we can also add an IsVisible property to BackButtonBehavior?

**Comment by techduggu** [comment]
@PureWeen @rmarinho Thanks for looking into this and providing the detailed feedback. I have the same opinion about keeping Flyout Behavior unaltered from Back Button Behavior as proposed earlier:

> As per my understanding (and more of proposal), the BackButtonBehavior should be applicable only on the pages pushed to Navigation Stack and should not alter the FlyoutBehavior at all. 

> but would like to propose that for Flyout, we should have separate properties to override icons and not use BackButtonBehavior to alter Flyout in any way.

Reg. this one:

> As @rmarinho said I think IsEnabled should just disable the backbutton like it currently does

I have a question that currently (in existing code before this PR) it doesn't disable the backbutton (as demonstrated in 'Before' screenshot). So we need to use this PR fix in order to disable the backbutton icon when BackButtonBehavior `IsEnabled` is false.

Also reg. this:

> Right now IsEnabled is also tied to command so if the command CanExecute is false i don't want that to show the hamburger menu. 

Isn't this expected? I meant lets say user has chosen BackButtonBehavior `IsEnabled` as false and if there is no command associated resulting `CanExecute` also false, and in that case, it will only show Flyout/hamburger Icon as a fallback if user has not chosen it to be disabled (i.e. user has enabled the Flyout in the main application page itself and not chosen it to be disabled yet). If user has chosen explicitly the Flyout to be disabled (similar to IsVisible property we are talking about), it won't display the hamburger on the pushed pages.

So in a nutshell, what I proposed and did is to give a fallback mechanism for Back Button Behavior to keep the consistent behavior what user has applied on the main Shell item itself and irrespective of how many items we push (also described in the original ticket):

> The objective is that the Master/Detail navigation would look and behave exactly the same regardless of how many items are in the navigation stack of the detail window.

In case we don't want to use fallback (which has its own benefit), I like the idea about adding a `IsVisible` property per page, the only thing is user need to handle `IsEnabled` property too in case of hiding the flyout icon completely otherwise app won't display flyout icon (as expected) but able to click on that empty area to open the drawer (unexpected). And at present, both icon visibility and behavior can be controlled by single `IsEnabled` property.

**Comment by PureWeen** [comment]
> Isn't this expected? I meant lets say user has chosen BackButtonBehavior IsEnabled as false and if there is no command associated resulting CanExecute also false, and in that case, it will only show Flyout/hamburger Icon as a fallback if user has not chosen it to be disabled

My point here is that I think the behavior would be confusing if it was

Bind Command to BackButton => Command Can Execute is False because user hasn't filled out a form "or something" => so now because IsEnabled is false the Flyout shows up

> In case we don't want to use fallback (which has its own benefit), I like the idea about adding a IsVisible property per page, the only thing is user need to handle IsEnabled property too in case of hiding the flyout icon completely otherwise app won't display flyout icon (as expected) but able to click on that empty area to open the drawer (unexpected). And at present, both icon visibility and behavior can be controlled by single IsEnabled property.

I could see doing this through an *IsVisible* property and then using the fallback in that way. Alongside that though I would want to make sure users could control the flyout at the page level

```XAML
<ContentPage Shell.FlyoutBehavior="Disabled">
<Shell.BackButtonBehavior>
<BackButtonBehavior IsVisible="false"
```

PureWeen:
@techduggu 

Can you expand this comment a bit more?

> but able to click on that empty area to open the drawer (unexpected). And at present, both icon visibility and behavior can be controlled by single IsEnabled property.

Just trying to unpack if there are some bugs here with BackButtonBehavior

techduggu:
@PureWeen Sure. I just considered a scenario where IsVisible is False and IsEnabled is True. What I meant from above point is - consider we have introduced IsVisible property and user has selected it to be False on a pushed page (but IsEnabled is still true). Now if you look at the existing code: https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/Renderers/ShellToolbarTracker.cs#L348 - we are currently using `IsEnabled` property value to set the drawer behavior as well. So in this case (if we don't change this code), user need to handle IsEnabled property too in-order to completely hide the Flyout/back button icon along with its behavior otherwise making just IsVisible as False will result in hiding the icon but `_drawerToggle.DrawerIndicatorEnabled` is still enabled to perform its functionality.

This is just a edge case scenario but can be controlled if we define under the docs once IsVisible is introduced.

samhouts:
I'm going to close this one for now until we're ready to return to it. Thanks, @techduggu !

PR #8265, opened 2019-10-27
[iOS] [macOS] Fix HTML string encoding

### LINKED ISSUE #8156: [Bug] [iOS] Label with HTML does not replace Umlaut characters properly
### Description

The HTML content is not correctly displayed because the label does not use UTF-8 (or escape the Umlaut characters like Ä, Ö, Ü, ä, ö, ü and ß) as I would expect it (because this display issue is not the case on the normal text input and neither is it on Android). 

### Steps to Reproduce

1. Place a Label with HTML content containing one or multiple of the mentioned characters

### Expected Behavior

Should display the text properly

### Actual Behavior

Error / placeholder / replacement characters are shown

### Basic Information

- Version with issue: 4.3
- Last known good version: -
- IDE: Latest stable VS for Mac Enterprise
- Platform Target Frameworks: 
  - iOS: Tested with iOS 13.1

### Screenshots

The translator section is a Label with HTML content. It was previously correctly formatted with white text, centered and no misplaced characters. I was using the [HtmlLabel](https://github.com/matteobortolazzo/HtmlLabelPlugin) plugin previously. 
Simulator Screen Shot - iPhone 11 - 2019-10-22 at 21 39 30

### Reproduction Link

Way too easy to reproduce to make a whole reproduction repo for it...

GalaxiaGuy:
### Description of Change ###

Specify explicit StringEncoding when creating attributed text from HTML on iOS and macOS.
Also updated the control gallery to demonstrate the other iOS HTML bugs.

### Issues Resolved ### 

- fixes #8156 

### API Changes ###

 None

### Platforms Affected ### 

- iOS
- macOS

Text outside the ASCII range, assumed to be UTF-8 will now display correctly.

image

### Testing Procedure ###

Control gallery label page.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
@@ -453,7 +453,8 @@ namespace Xamarin.Forms.Platform.MacOS
 #if __MOBILE__
 			var attr = new NSAttributedStringDocumentAttributes
 			{
-				DocumentType = NSDocumentType.HTML
+				DocumentType = NSDocumentType.HTML,
+				StringEncoding = NSStringEncoding.UTF8
 			};
 
 			NSError nsError = null;
@@ -463,7 +464,8 @@ namespace Xamarin.Forms.Platform.MacOS
 #else
 			var attr = new NSAttributedStringDocumentAttributes
 			{
-				DocumentType = NSDocumentType.HTML
+				DocumentType = NSDocumentType.HTML,
+				StringEncoding = NSStringEncoding.UTF8
 			};
 
 			var htmlData = new NSMutableData();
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
```diff
@@ -450,7 +450,8 @@ void UpdateTextHtml()
 #if __MOBILE__
 			var attr = new NSAttributedStringDocumentAttributes
 			{
-				DocumentType = NSDocumentType.HTML
+				DocumentType = NSDocumentType.HTML,
+				StringEncoding = NSStringEncoding.UTF8
```
rmarinho: I think a better solution would be to refactor this to a `protected virtual NSAttributedStringDocumentAttributes  GetNSAttributedStringDocumentAttributes` , that way developers can override that method and specify a different encoding if they need to.

GalaxiaGuy: Sure :)

PR #8270, opened 2019-10-28
BindableLayout allow use ViewCell

### LINKED ISSUE #8268: [Enhancement] BindableLayout allow use ViewCell
## Summary
It would be nice to have the opportunity to use ViewCell in a DataTemplate in Bindable Layout.
In this case, you can change to choose which container to use. ListView or StakLayout for example.

KSemenenko:
### Description of Change ###
It would be nice to have the opportunity to use ViewCell in a DataTemplate in Bindable Layout.
In this case, you can change to choose which container to use. ListView or StakLayout for example.

``` xaml
<DataTemplate x:Key="MyViewCellCharTemplate">
    <ViewCell>
        <Frame Padding="10" BackgroundColor="DodgerBlue">
	      <Label Text="{Binding }" />
	</Frame>
    </ViewCell>
</DataTemplate>
```

### Issues Resolved ### 

- fixes #8268

### API Changes ###
 None

### Platforms Affected ### 
- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###
1. Open Controll Gallery
2. Select BindableLayout Gallery - Legacy
3. Press Replase button
4. Check different colors for background.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls/GalleryPages/BindableLayoutGalleryPage.xaml
+++ b/Xamarin.Forms.Controls/GalleryPages/BindableLayoutGalleryPage.xaml
@@ -18,9 +18,20 @@
 				<Label Text="{Binding }" />
 			</Frame>
 		</DataTemplate>
+		<!-- Template for ViewCell items -->
+		<DataTemplate x:Key="MyViewCellCharTemplate">
+			<ViewCell>
+				<Frame Padding="10"
+					   BackgroundColor="DodgerBlue">
+					<Label Text="{Binding }" />
+				</Frame>
+			</ViewCell>
+		   
+		</DataTemplate>
 		<local:BindableLayoutItemTemplateSelector x:Key="ItemTemplateSelector"
 												  IntTemplate="{StaticResource MyIntTemplate}"
-												  CharTemplate="{StaticResource MyCharTemplate}" />
+												  CharTemplate="{StaticResource MyCharTemplate}"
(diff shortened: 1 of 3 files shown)


rmarinho:
Why would you do this not just the Frame?

KSemenenko:
@rmarinho Based on my experience, I often have to reuse the DataTemplate. And sometimes it’s more logical to use StackLayout + BindableLayout rather than ListView.
And in this case, I can not use my DataTemplate from ListView which use ViewCell.
AndI have to duplicate the code, or use my component implementations.

KSemenenko:
In general, this is for the case when we have a bunch of DataTemaplate which we can use in both ListView and StackLayout.
And we do not need to create copies.

KSemenenko:
@rmarinho Do you want to stop using ViewCell?
What good news!

KSemenenko:
Then of course, this functionality is not needed

rmarinho:
Yap, for example if you check CollectionView already doesn't use it.

hartez:
Also, ViewCell is derived from Cell, which is specifically for use in TableView/ListView; there are implications for things like context actions which we don't intend to support in BindableLayout. 

Folks who are using templates in both a ListView and a StackLayout will simply have to maintain two templates, or move their ListView to CollectionView (which does not use ViewCell), or use a DataTemplateSelector to wrap their templates in a ViewCell, or possibly use a ControlTemplate to share the code.

PR #8271, opened 2019-10-28
Add On/Off VisualStates for Switch

### LINKED ISSUE #8258: [Bug] Switch ThumbColor is broken
### Description

There are two problems, at least on Android:

1. When a switch is initialized toggled, the 'on' thumb color is always the accent color. But when a switch is initialized untoggled, the 'on' thumb color is some shade of white. So two otherwise identical switches will look different for no obvious reason.
2. A switch's thumb actually has two colors, an 'on' color and an 'off' color. However, Xamarin.Forms.Switch only offers a single property, ThumbColor, which overrides both the on and off states. Ideally, this would be broken up into two distinct properties.

### Steps to Reproduce

```
        <Switch IsToggled="True" />
        <Switch />
        <Switch ThumbColor="Blue" />
```

### Expected Behavior

1. Switch thumb colors shouldn't depend on the initial toggled state of the switch.
2. We should be able to independently control both the on and off colors for the thumb.

### Actual Behavior

As described.

### Basic Information

- Version with issue: 4.3.0.908675
- IDE: VS 2019 16.3.6
- Platform Target Frameworks: 
  - Android: 9.0
- Affected Devices: Android

### Screenshots

image

### Reproduction Link

https://github.com/mfeingol/repros/tree/master/SwitchOnColor-8258

### LINKED ISSUE #8263: [Enhancement] Add On/Off VisualStates for Switch
## Summary

Provide On/Off VisualStates for the Switch control so that users can style them in the VisualStateManager.

## API Changes

None; this would simply require adding a changed handler for the `Switch.IsToggledProperty` and calling `ChangeVisualState()`; `ChangeVisualState` would need to be overridden for the Switch to set the appropriate state.

## Intended Use Case

Users may want to change the Switch ThumbColor or other properties (see #8258) based on the state of the Switch. This will allow them to do so.

techduggu:
### Description of Change ###

As described in the ticket #8263, Added On/Off VisualStates for Switch.

### Issues Resolved ### 

- fixes #8263 
- fixes #8258 

### API Changes ###
As described, added a changed handler for the Switch `IsToggledProperty` to call `ChangeVisualState()`; `ChangeVisualState` has been overridden for the Switch to set the appropriate states (On/Off).

### Platforms Affected ### 

- Core/XAML (all platforms)

### Testing Procedure ###

- Just run `Issue8263.xaml` from Control Gallery to manually test it.

- I've also added a UI Test `SwitchOnOffVisualStatesTest` inside Issue8263.

### PR Checklist ###


techduggu:
@rmarinho I've not added 'Disable' state explicitly as it is already part of '`CommonStates`' of `VisualStateGroup` and it will call it as we are calling `base.ChangeVisualState();` if the Switch doesn't pass the first two states On/Off (or Switch is Disabled in other words). I've not tested the Disabled state yet, but let me take a look in-case of any issue.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue8263.xaml.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Linq;
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 8263, "[Enhancement] Add On/Off VisualStates for Switch")]
```
hartez: I think this needs the `ManualReview` test category applied to it.

Xamarin.Forms.Core/Switch.cs
```diff
@@ -39,7 +43,16 @@ public Switch()
 		public bool IsToggled
 		{
 			get { return (bool)GetValue(IsToggledProperty); }
-			set { SetValue(IsToggledProperty, value); }
+			set { SetValue(IsToggledProperty, value); ChangeVisualState(); }
```
StephaneDelcroix: nope. the ChangeVisualState will be triggered by the property changed handler

Xamarin.Forms.Core/Switch.cs
```diff
@@ -6,11 +6,15 @@ namespace Xamarin.Forms
...
 	{
-		public static readonly BindableProperty IsToggledProperty = BindableProperty.Create("IsToggled", typeof(bool), typeof(Switch), false, propertyChanged: (bindable, oldValue, newValue) =>
+		public const string SwitchOnVisualState = "On";
+		public const string SwitchOffVisualState = "Off";
+
+		public static readonly BindableProperty IsToggledProperty = BindableProperty.Create(nameof(IsToggled), typeof(bool), typeof(Switch), false, propertyChanged: (bindable, oldValue, newValue) =>
 		{
 			EventHandler<ToggledEventArgs> eh = ((Switch)bindable).Toggled;
 			if (eh != null)
 				eh(bindable, new ToggledEventArgs((bool)newValue));
```
StephaneDelcroix: replace with
```csharp
((Switch)bindable).Toggled?.Invoke(bindable, new ToggledEventArgs((bool)newValue));
```

while you're at it

Xamarin.Forms.Core/Switch.cs
```diff
@@ -39,7 +43,16 @@ public Switch()
...
+			set { SetValue(IsToggledProperty, value); ChangeVisualState(); }
+		}
+		protected internal override void ChangeVisualState()
+		{
+			if (IsEnabled && IsToggled)
+				VisualStateManager.GoToState(this, SwitchOnVisualState);
+			else if (IsEnabled && !IsToggled)
+				VisualStateManager.GoToState(this, SwitchOffVisualState);
+			else
+				base.ChangeVisualState();
```
StephaneDelcroix: shouldn't you always invoke the base.ChangeVisualState, then handle the specifics ?? /cc @hartez

Xamarin.Forms.Core/Switch.cs
techduggu: Hi @StephaneDelcroix 
Thanks for looking into this. I understand that we are replacing the above code (Line 14 to 16) with the single line code markup as you suggested. This is an existing code.
Quick question (just so I'm aware of before implementing): Is this just only for syntactic sugar or really has any major issue?

Xamarin.Forms.Core/Switch.cs
techduggu: I've a question, What if someone not using the MVVM approach or Binding Property (IsToggledProperty) and use this public property IsToggled to set the value (though I know they need to have their own mechanism to update the UI)? 
It is similar to what we have with Checkbox: https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/CheckBox.cs#L38

Xamarin.Forms.Core/Switch.cs
techduggu: @StephaneDelcroix Yes, you are correct. I retested it and noticed that it will never call 'Normal' state if Switch is enabled but no On/Off states defined by user in Xaml. Hence, I moved it at the top so that it always invoke the base Visual States and then our new On/Off states if available. I've tested various combinations too (added Unit Tests) and it works as expected now (except Focused state - I guess this isn't valid for Switch control and I was unable to set focus on it using `switch.Focus()` ).

techduggu:
@samhouts @StephaneDelcroix I saw a merge conflict and resolved it. Is this PR good to be merged or under consideration?

PR #8273, opened 2019-10-28
Add missing null check on ItemsSource when IsGrouped is true

### LINKED ISSUE #8269: [Bug] CollectionView exception when IsGrouped=true and null ItemSource
### Description
When using IsGrouped="true" on a CollectionView, an error is thrown when the datasource is null. Seems obvious now i write it down - but the control is more forgiving when IsGrouped="false". Also, although the exception mentions a null reference, it was tricky to track down the source when it occurred. A workaround is to set the ItemSource to an empty list rather than a null list.

### Steps to Reproduce

Using the CollectionView Demo as a base - https://github.com/xamarin/xamarin-forms-samples/tree/master/UserInterface/CollectionViewDemos 

1.  Change line 9 of GroupedAnimalsViewModel.cs to have a null value rather than a new List
2.  Run the app on Android (i haven't tried on any other platform)
3.  Click into one of the grouping example pages

### Expected Behavior
Show nothing

### Actual Behavior
Throws an exception

- Version with issue: 4.3
- IDE: VS2019

hartez:
### Description of Change ###

Adds a missing null check before trying to create a grouped data source out of a null ItemsSource.

### Issues Resolved ### 
- fixes #8269

### API Changes ###
None

### Platforms Affected ### 
- Android

None

Not applicable

### Testing Procedure ###
Automated test

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue8269.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue8269.cs
@@ -1,0 +1,49 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.Forms.Core.UITests;
+using Xamarin.UITest;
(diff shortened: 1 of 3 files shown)


samhouts:
@hartez Please rebase to 4.3.0. Thanks!

PR #8283, opened 2019-10-29
[iOS] Fix crash using an object that implements IEnumerable and INotifyCollectionChanged

### LINKED ISSUE #8200: [Bug] CollectionView on iOS assumes INotifyCollectionChanged is an IList (causing null reference exception when assumption is wrong)
### Description
We are using collection view with an object that implements `IEnumerable` and `INotifyCollectionChanged`. Starting in Xamarin.Forms 4.3.0 (the official release), this has started causing a `NullReferenceException`. This behavior changed between 4.3.0-pre3 and the official release of 4.3.0.

The issue is in `ItemsSourceFactory` -- when it is creating the source for the collection view, it checks the type of the items source. If it implements `INotifyCollectionChanged`, it attempts to cast it to `IList`.
https://github.com/xamarin/Xamarin.Forms/blob/f9114b1306f2896cce07d358725f63ce6ab8cac5/Xamarin.Forms.Platform.iOS/CollectionView/ItemsSourceFactory.cs#L20

`IList` and `INotifyCollectionChanged` are not derived from each other, so this cast can fail. When it does, `ObservableItemSource` will be instantiated with a null item source. This causes a `NullReferenceException` in the constructor when it attempts to add an event listener to `CollectionChanged`:
https://github.com/xamarin/Xamarin.Forms/blob/f9114b1306f2896cce07d358725f63ce6ab8cac5/Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs#L28

### Steps to Reproduce

1. Set `ItemsSource` on a collection view to an `IEnumerable` which also implements `INotifyCollectionChanged` (but not `IList`)

### Expected Behavior
The collection view would read in data from the source and render the view appropriately.

### Actual Behavior
The app crashes with a `NullReferenceException`

### Basic Information

- Version with issue: 4.3.0
- Last known good version: 4.3.0-pre3
- IDE: Visual Studio for Mac
- Platform Target Frameworks: 
  - iOS:  iOS 13.1
  - Android: ??
  - UWP:  N/A
- Nuget Packages:
- Affected Devices:

### Screenshots
N/A

### Reproduction Link
N/A

jsuarezruiz:
### Description of Change ###

Fix crash using an object that implements `IEnumerable` and `INotifyCollectionChanged`.

Initially, I only made changes to `ItemsSourceFactory` to use `ListSource` in this case. This would prevent the crash, however, we would ignore `INotifyCollectionChanged`. For that reason, I made changes also in `ObservableItemsSource` (on Android too) to use IEnumerable. In this way, we guarantee the use of IList (have IEnumerable as base class) and IEnumerable.

### Issues Resolved ### 

- fixes #8200

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS
- Android

None

#### Before
The app crashes with a NullReferenceException

#### After
Captura de pantalla 2019-10-29 a las 11 36 47

### Testing Procedure ###
Launch Core Gallery and navigate to the Issue 8200. If the CollectionView appear with items, the test pass.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/CollectionView/ObservableItemsSource.cs
+++ b/Xamarin.Forms.Platform.Android/CollectionView/ObservableItemsSource.cs
@@ -6,18 +6,18 @@ namespace Xamarin.Forms.Platform.Android
 {
 	internal class ObservableItemsSource : IItemsViewSource
 	{
-		readonly IList _itemsSource;
+		readonly IEnumerable _itemsSource;
 		readonly ICollectionChangedNotifier _notifier;
 		bool _disposed;
 
-		public ObservableItemsSource(IList itemSource, ICollectionChangedNotifier notifier)
+		public ObservableItemsSource(IEnumerable itemSource, ICollectionChangedNotifier notifier)
 		{
-			_itemsSource = itemSource;
+			_itemsSource = itemSource as IList ?? itemSource as IEnumerable;
 			_notifier = notifier;
 			((INotifyCollectionChanged)itemSource).CollectionChanged += CollectionChanged;
 		}
 
-		public int Count => _itemsSource.Count + (HasHeader ? 1 : 0) + (HasFooter ? 1 : 0);
+		public int Count => ItemsCount() + (HasHeader ? 1 : 0) + (HasFooter ? 1 : 0);
 
 		public bool HasHeader { get; set; }
 		public bool HasFooter { get; set; }
@@ -39,9 +39,9 @@ namespace Xamarin.Forms.Platform.Android
 
 		public int GetPosition(object item)
 		{
-			for (int n = 0; n < _itemsSource.Count; n++)
+			for (int n = 0; n < ItemsCount(); n++)
 			{
-				if (_itemsSource[n] == item)
+				if (ElementAt(n) == item)
 				{
 					return AdjustPositionForHeader(n);
 				}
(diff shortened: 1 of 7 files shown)


Xamarin.Forms.Platform.Android/CollectionView/ObservableItemsSource.cs
```diff
@@ -1,23 +1,24 @@
 using System;
-using System.Collections;
+using System.Collections.Generic;
 using System.Collections.Specialized;
+using System.Linq;
```
rmarinho: We try to avoid Linq.

jsuarezruiz: Updated. Now, if `IList` is used, exactly the same thing we were doing is done. In case of using `IEnumerable`, for example, to obtain the Count we iterate the items (there is no other option) but we avoid using Linq.

Xamarin.Forms.Platform.Android/CollectionView/ObservableItemsSource.cs
```diff
@@ -187,5 +187,48 @@ void Replace(NotifyCollectionChangedEventArgs args)
...
+			{
+				if (count == index)
+					return item;
+				count++;
+			}
+
+			return -1;
+		}
+
+		internal int IndexOf(object item)
```
rmarinho: maybe this should be internal extensions

samhouts:
@jsuarezruiz Please retarget to 4.4.0. Thanks!

jsuarezruiz:
Retarget to 4.4.0!

PR #8297, opened 2019-10-30
Fixed 7808 - [Enhancement] Allow providing custom ShellSectionRootRenderer

### LINKED ISSUE #7808: [Enhancement] Allow providing custom ShellSectionRootRenderer
## Summary
For Shell customization, the ShellRenderer itself exposes many virtual methods which are overridable to create custom IShell..Renderers such as CreateShellSectionRenderer() https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/ShellRenderer.cs. 
However more granular renderers in lower levels are not exposed. For example ShellSectionRootRenderer is private to ShellSectionRenderer https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/ShellSectionRenderer.cs line 65. 

## API Changes
Add CreateShellSectionRootRenderer() to IShellSectionRenderer instead of instantiating the renderer in LoadPages() https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/ShellSectionRenderer.cs line 179.

Add CreateShellSectionRootHeader() to IShellSectionRootRenderer instead of instantiating it internally in UpdateHeaderVisibility() https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/ShellSectionRootRenderer.cs line 208.

## Intended Use Case
To be able to modify tab styles using a custom Visual - MyVisual

```csharp
[assembly: ExportRenderer(typeof(Shell), typeof(MyShellRenderer), new[] { typeof(MyVisual) })]
namespace Sample.iOS.Visual
{
    public class MyShellRenderer : ShellRenderer
    {
        protected override IShellSectionRenderer CreateShellSectionRenderer(ShellSection shellSection)
        {
            return new MyShellSectionRenderer(this);
        }
    }
}
```

```csharp
namespace Sample.iOS.Visual
{
    public class MyShellSectionRenderer : ShellSectionRenderer
    {
        public MyShellSectionRenderer(IShellContext context) : base(context)
        {
        }

        //This is what is needed
        public override IShellSectionRootRenderer CreateShellSectionRootRenderer(...)
        {
            ... return new MyShellSectionRootHeader(...);
        }
    }
}
```

Currently, the whole ShellRenderer needs to be replaced just to customize tab styles.

techduggu:
### Description of Change ###

As mentioned in the ticket #7808 , allowed access for `ShellSectionRootRenderer` and `ShellSectionRootHeader` to allow for further customization from platform-specific renderers. 

### Issues Resolved ### 

- fixes #7808

### API Changes ###

Added:

- Added `CreateShellSectionRootRenderer()` to `IShellSectionRenderer` instead of instantiating the renderer in LoadPages() https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/ShellSectionRenderer.cs#L179

- Added `CreateShellSectionRootHeader()` to `IShellSectionRootRenderer` instead of instantiating it internally in `UpdateHeaderVisibility()` https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/ShellSectionRootRenderer.cs#L208

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###

- Just run `Issue7808.cs` from Control Gallery. I tried to create a test/use case (might not be a best usecase) and customized the `ShellSectionRootRenderer` and `ShellSectionRootHeader` (as described in the ticket). On iOS, if ShellSection View with 'Blue' and header as 'Red' background color are shown, this test has passed.

Please Note: I don't have iOS setup so I'd need help from someone from Xamarin Team to review the above test on iOS. I've implemented all the required changes (including Custom Renderer) in order to run the above test page (Issue7808.cs). Please let me know in case of any questions.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.ControlGallery.iOS/CustomRenderers.cs
+++ b/Xamarin.Forms.ControlGallery.iOS/CustomRenderers.cs
@@ -691,5 +691,35 @@ namespace Xamarin.Forms.ControlGallery.iOS
 		}
 	}
 
+	public class Issue7808ShellRenderer : ShellRenderer
+	{
+		protected override IShellSectionRenderer CreateShellSectionRenderer(ShellSection shellSection)
+		{
+			return new Issue7808ShellSectionRenderer(this);
+		}
+	}
+
+	public class Issue7808ShellSectionRenderer : ShellSectionRenderer
+	{
+		public Issue7808ShellSectionRenderer(IShellContext context) : base(context)
+		{
+		}
+
+		//This is what is needed
+		protected override IShellSectionRootRenderer CreateShellSectionRootRenderer(ShellSection shellSection)
+		{
+			var renderer = base.CreateShellSectionRootRenderer(shellSection);
+			if (renderer != null)
+			{
+				//customize your stuff here, for example:
+				(renderer as ShellSectionRootRenderer).View.BackgroundColor = UIColor.Blue;
+				var header = renderer.CreateShellSectionRootHeader();
+				if (header != null)
+					header.View.BackgroundColor = UIColor.Red;
+			}
+			return renderer;
+		}
+	}
+
 }
 
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7808.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7808.cs
@@ -1,0 +1,66 @@
+﻿using System;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System.Linq;
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+using Xamarin.Forms.Core.UITests;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 7808, "[Enhancement] Allow providing custom ShellSectionRootRenderer",
+		PlatformAffected.iOS)]
+
+	public class Issue7808 : TestShell
+	{
+		const string CreateBottomTabButton = "CreateBottomTabButton";
+		protected override void Init()
+		{
+			
+			var page = CreateContentPage();
+			page.Title = "Main";
+			page.Content = CreateEntryInsetView();
+			CurrentItem = Items.Last();
+		}
+
+		View CreateEntryInsetView()
+		{
+			var random = new Random();
+			ScrollView view = null;
+			view = new ScrollView()
+			{
+				Content = new StackLayout()
+				{
+					Children =
+						{
+							new Button()
+							{
+								Text = "Bottom Tab",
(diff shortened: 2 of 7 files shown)


techduggu:
Thanks @samhouts for the feedback and it makes sense. Let me take a look and think of different approach to expose the above granular renderers.

Sauder91:
@techduggu I was just trying to do something that would require this feature. If the methods weren't added to the interface and I had to subclass ShellSectionRenderer and ShellSectionRootRenderer so that way I could use my subclassed ShellSectionRootHeader I would be ok with that.  Its a little clunkier, but the ABI/API wouldn't change. @samhouts Would that be acceptable?

techduggu:
@samhouts I had given it a thought to this problem (dependency on existing interface) and all I can think of is using the Interface segregation principal of SOLID to solve it. If we create a new public interface (basically breaking down the interfaces into smaller chunks) and move our methods into this new interface and implement them only in corresponding classes `ShellSectionRenderer` and `ShellSectionRootRenderer`. This way we won't be forcing any client who is dependent on existing public interfaces. What do you suggest?
@Sauder91 Do you see any issue with this approach? Also, how do you plan to access ShellSectionRootHeader after sub-classing as it is currently not exposed and private to the ShellSectionRootRenderer which was the main problem described in #7808

Sauder91:
I don't see any problem adding a new public interface personally.  I wasn't planning on accessing the ShellSectionRootHeader, but instead sub-classing it,  then subclassing ShellSectionRootRenderer to use my RootHeader, and subclassing ShellSectionRenderer to use my RootRenderer. In order to do that though I need the creation of ShellSectionRootRenderer and ShellSectionRootHeader moved to virtual classes. Does that make sense?

PureWeen:
@techduggu @Sauder91 

Apologies but I had missed this PR. I'm going to close this PR for now because of a different PR I did

Can you let me know what you think of this PR?
https://github.com/xamarin/Xamarin.Forms/pull/8441

My take was to not add any interfaces to *IShellSectionRenderer*

I feel like *ShellSectionRootRenderer* is specific to our concrete implementation of  *ShellSectionRenderer*  and that someone implementing a custom *IShellSectionRenderer* won't really have any reason to create a *ShellSectionRootRenderer*

*ShellSectionRootHeader* isn't used anywhere outside of *ShellSectionRenderer*

techduggu:
@PureWeen No worries. I looked at #8441 and it exactly the same thing what I came up later and suggested in the above comment i.e. Interface segregation  (https://github.com/xamarin/Xamarin.Forms/pull/8297#issuecomment-549756202) which I can see that you have also done by creating a new public interface for ShellSectionRootHeader. 
Additionally, I like that we kept the virtual method for ShellSectionRootRenderer too rather than hardcoding the object, so this will help in-case someone wants to use it though it is internally used. Thanks for optimizing it!

PR #8298, opened 2019-10-30
Set StatusBar color and StatusBarStyle for iOS Android and UWP from Page

### LINKED ISSUE #7314: [Enhancement] Set StatusBar color for iOS and Android from any Page
## Summary
Now this is only possible in NavigationPage.
I suggest making it accessible from any page.

## API Changes

``` 
Page.SetStatusBarColor(Color color);
Page.SetStatusBarLightTheme();
Page.SetStatusBarDarkTheme();
```

or

```
enum StatusBarTheme
{
    Light,
    Dark,
}
Page.SetStatusBarColor(StatusBarTheme theme);
```

KSemenenko:
### Description of Change ###
With the advent of the black theme on iOS and generally good opportunities for adjusting colors, there was a need to be able to change the color of the StatusBar on the page.

### Issues Resolved ### 

- fixes #7314

### API Changes ###
New properties for Page
``` cs
public Color StatusBarColor { get; set; } // color for StatusBar
public StatusBarStyle StatusBarStyle { get; set; } //style for StatusBar content (Dark or Light)
```

Added:
``` cs
Page.StatusBarColor;
Page.StatusBarStyle;
```
``` cs
public enum StatusBarStyle
{
      Default = 0,
      LightContent = 1,
      DarkContent = 2
}
```
### Important
For iOS it require changes into info.plist  (maybe it can be put into Xamarin.Forms project template for VS and VS for Mac)
If this flag is not set, then the StatusBarStyle will not be changed.
``` 
<key>UIViewControllerBasedStatusBarAppearance</key>
<false/>
```
The priority settings for the StatusBar are on the container pages TabbedPage, CaruselPage, NaviationPage. it is higher than just pages. And their StatusBar settings will overwrite the settings of the page.

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS
- Android

@jsuarezruiz thanks for this animation

### Testing Procedure ###
1. Open Controll Gallery
2. Page StatusBar Gallery
3. Move R, G, and B sliders
4. Toggle Switch
5. Press buttons to check different pages

### PR Checklist ###


jsuarezruiz:
The implementation in UWP would not be complex.
You should use for Desktop:
`var titleBar = ApplicationView.GetForCurrentView().TitleBar;`
And:
`var statusBar = StatusBar.GetForCurrentView();`
For mobile.

Then, you could modify the `BackgroundColor` (you also have access to opacity, back button color, etc.).

Xamarin.Forms.Core/Page.cs
```diff
@@ -36,6 +36,8 @@ public class Page : VisualElement, ILayout, IPageController, IElementConfigurati
 
 		public static readonly BindableProperty TitleProperty = BindableProperty.Create("Title", typeof(string), typeof(Page), null);
 
+		public static readonly BindableProperty StatusBarColorProperty = BindableProperty.Create("StatusBarColor", typeof(Color), typeof(Page), Color.Default);
```
jsuarezruiz: `public static readonly BindableProperty StatusBarColorProperty = BindableProperty.Create(nameof(StatusBarColor), typeof(Color), typeof(Page), Color.Default);`

Xamarin.Forms.Platform.Android/Renderers/PageRenderer.cs
```diff
@@ -89,6 +90,8 @@ protected override void OnElementChanged(ElementChangedEventArgs<Page> e)
 		protected override void OnElementPropertyChanged(object sender, PropertyChangedEventArgs e)
 		{
 			base.OnElementPropertyChanged(sender, e);
+			if (e.PropertyName == Page.StatusBarColorProperty.PropertyName)
+				UpdateStatusBarColor();
 			if (e.PropertyName == Page.BackgroundImageSourceProperty.PropertyName)
```
jsuarezruiz: `else if`

KSemenenko:
@jsuarezruiz  thanks for help with UWP :)

Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
```diff
@@ -542,6 +544,64 @@ void IShellContentInsetObserver.OnInsetChanged(Thickness inset, double tabThickn
 			}
 		}
 
+		void UpdateStatusBarColor()
+		{
+			var statusBarColor = Page.StatusBarColor.ToUIColor();
+
+			//UIApplication.SharedApplication.SetStatusBarStyle(UIStatusBarStyle.LightContent, false);
```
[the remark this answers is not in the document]
KSemenenko: done

KSemenenko:
@jsuarezruiz I also want to add the ability to change the color of the text in the StatusBar.
but for now I have problems with this.

jsuarezruiz:
@KSemenenko Let's see if I can help you a little bit.

Let's start with **UWP**. This case is quite simple. You can access to the `TitleBar` and set custom colors for the background and text color:

```
var titleBar = ApplicationView.GetForCurrentView().TitleBar;
titleBar.ForegroundColor = Windows.UI.Colors.White;
titleBar.BackgroundColor = Windows.UI.Colors.Green;
```

It is also possible to modify back button colors, inactive status colors, etc. More info: https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar

In **Android** you have the code for the background color. After get the current Window, use the `SetStatusBarColor` method.

`currentWindow.SetStatusBarColor(Android.Graphics.Color.Red);`

Since Android M (23) it is possible to set a predefined status bar text colour theme to light or dark https://developer.android.com/reference/android/R.attr#windowLightStatusBar.

In **iOS**, since iOS 7 it is possible to set a predefined status bar text colour theme to light or dark. You have to manipulate the Info.plist.

```
<key>UIViewControllerBasedStatusBarAppearance</key>
<false/>
```

Next:

```
UIApplication.SharedApplication.SetStatusBarStyle(UIStatusBarStyle.LightContent, false);
GetCurrentViewController().SetNeedsStatusBarAppearanceUpdate();
```

I hope that helps!

KSemenenko:
@jsuarezruiz It helped a lot. thank you very much! 
I think I’ve finished, can you watch this PR when you have time?

KSemenenko:
@jsuarezruiz I will check TabbedPage  and NavigationPage.
And Shell :)
thanks!

KSemenenko:
@jsuarezruiz I checked, everything works only for Page. Container pages do not support our property. I also added a few buttons so that we can look at different types of pages
Maybe this is logical, because for example, every page per navigation can have different colors, and it is responsibility of the page itself to set the colors.
I changed the description. 

Do you think it will be enough for this task?

KSemenenko:
I have done normal behavior.
@jsuarezruiz  Can you see when you have time?

KSemenenko:
@jsuarezruiz  Now everything works as it should :)

Xamarin.Forms.ControlGallery.iOS/Info.plist
```diff
@@ -36,7 +36,7 @@
 		<string>Default-812h@3x.png</string>
 	</array>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
```
jsuarezruiz: This change is necessary to programmatically change the style of the StatusBar (light or dark). It is not necessary to change the background color. How are we going to manage this case? I think it must be prominently in the documentation.

KSemenenko: it should be indicated in the documentation, because if this option is not enabled, then the color change simply does not work.

KSemenenko:
@jsuarezruiz Thanks for taking the time to watch this PR.
The property is really necessary. Because I will fix the name to a more understandable one.

KSemenenko:
@jsuarezruiz  can you please run build again? I changed property name to
IsLoaded. The property is interesting.
Now we can use IsLoaded in the ViewModel to see if your page is shown or not.
This is an accident, but very convenient.
Also can you help me, previous build was falied because of
```
##[error]Xamarin.Forms.Platform.iOS\Renderers\ShellRenderer.cs(318,73): Error CS0117: 'UIStatusBarStyle' does not contain a definition for 'DarkContent'
```
Locally, I have everything going without problems.

jfversluis:
> Also can you help me, previous build was falied because of...

This is because we are not building against Xcode 11 yet in our pipeline

KSemenenko:
Thanks for the answer @jfversluis, does that mean that I shouldn’t change anything?

jfversluis:
That is correct, that part is totally on us. We might need you to rebase whenever we do get it into our pipeline, but until then there is nothing you can do unfortunately.

KSemenenko:
@jfversluis Okay, then I'll follow the messages :) Thanks.

Xamarin.Forms.Core/Page.cs
```diff
@@ -36,8 +36,18 @@ public class Page : VisualElement, ILayout, IPageController, IElementConfigurati
...
 
+		public static readonly BindableProperty StatusBarColorProperty = BindableProperty.Create(nameof(StatusBarColor), typeof(Color), typeof(Page), Color.Default, 
+			propertyChanged: (bo, o, n) => ((Page)bo).StatusBarColorUpdated());
+
+		public static readonly BindableProperty StatusBarStyleProperty = BindableProperty.Create(nameof(StatusBarStyle), typeof(StatusBarStyle), typeof(Page), StatusBarStyle.Default,
+			propertyChanged: (bo, o, n) => ((Page)bo).StatusBarStyleUpdated());
+
 		public static readonly BindableProperty IconImageSourceProperty = BindableProperty.Create(nameof(IconImageSource), typeof(ImageSource), typeof(Page), default(ImageSource));
 
+		static readonly BindablePropertyKey IsLoadedPropertyKey = BindableProperty.CreateReadOnly(nameof(IsLoaded), typeof(bool), typeof(Page), false);
```
PureWeen: Can you make this PR work without adding this property? We're not going to be able to add a property like this as a part of this PR.

We have a spec here https://github.com/xamarin/Xamarin.Forms/issues/8596 for this property

In the code where you are checking for IsLoaded can you instead wire up to the Appearing event or use a local variable? For example in PageRenderer

https://github.com/xamarin/Xamarin.Forms/blob/6fda5941981d4566f54b873d3ab297a76ccac21e/Xamarin.Forms.Platform.Android/Renderers/PageRenderer.cs#L57

You could just set a local property.

KSemenenko: @PureWeen Is there any news when #8596 will be ready?
Maybe we can change the name?
At first it was IsOnScreen.
Because to call the work of this property, I need to know exactly whether the page is currently on the screen or not.

KSemenenko: Would it also be a good idea to have the same functionality in 2 implementations?

PureWeen: The spec I linked to actually won't tell you what Page is onscreen

From the looks of how you've wired up *IsLoaded* it seems like you're just exposing out _hasAppeared.

I still think doing this a bit more event based around appearing and dispappering would be more resilient though. 

Just wiring up to Appearing inside 
https://github.com/xamarin/Xamarin.Forms/pull/8298/files#diff-5c15636b9e632ed12684a0c5ca1f29e8R165

that way when appearing fires it'll update the statusbarcolors and then locally inside the renderer you can just have a property to track appeared/disappeared


For example in the UAP code the UpdateStatusBar calls that you have are before the renderer sends the appearing event too the cross platform code

https://github.com/xamarin/Xamarin.Forms/pull/8298/files#diff-ebd6a6485423cd0b5a1c9af41db5bea1R68

KSemenenko: For make this functionality to work correctly, I need a property in the Page class.
So I'm ready to rename it. but we need to come up with a name.
Also, the public property will make it convenient to use it in the ViewModel
@PureWeen

PureWeen: > To make this functionality work correctly, I need a property in the Page class.

Why? What behavior in this PR can't be achieved with what I've proposed?

KSemenenko: The fact is that the color status of the bar is set for the entire application.
And a situation is possible, when:
you create Page1 and Page2.
You put the first page color red, and the second yellow.
and set MainPage = page1;
and the color of the status bar will be yellow. because as soon as you assign the color to page 2, it will be set for the entire application.
And it does not matter whether the page is displayed or we just created an instance of the class.
Another case of MultiPage, such as NaviagationPage or TabbedPage.
There is one page, but inside there are other pages and they switch.

I made a property - which will say whether the page is on the screen now or not.
And set the color only when the page is on the screen.

I think that the property of the page will be the best solution. This does not require invading all possible page renderers.
And adds convenience to the MVVM pattern :)
so, two in one

Also, for the build of this branch, we need the latest Xcode on build agent.

PureWeen: > Another case of MultiPage, such as NaviagationPage or TabbedPage.
There is one page, but inside there are other pages and they switch.

I get this and we have this concept already working multiple places
- Shell.SetNavBarIsVisible
- Shell.SetNavBarHasShadow
- NavigationPage.SetHasNavigationBar
- NavigationPage.SetHasBackButton
- ContentPage.ToolBarItems
- etc...

We already have a lot of implementations that effect global features based on the active page in different multi page scenarios so I think we should just tie into those

https://github.com/xamarin/Xamarin.Forms/blob/c579e17ee08665b5dc787b05dd530869e9792b7f/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs#L119


> IsLoaded

The annoying truth is that adding an IsLoaded property is a very large discussion and if this PR is going to hinge on that property being implemented it will be awhile before this PR can be merged. For example tying a property called IsLoaded to Appearing doesn't really make sense. If you have a TabPage with 3 tabs then technically all 3 of those pages should have IsLoaded set to true even if they aren't visible.

To create the property you want we would need to add a new property like 
*IsPresented* and wire it up differently than the Appearing event.

And even then it's arguable whether you'd want to tie the StatusBarColor change to IsPresented. For example on Shell we tie NavBarVisible to WillShowViewController (https://github.com/xamarin/Xamarin.Forms/blob/c79e187d61501f70ec8cdd40b5d3b447407980b0/Xamarin.Forms.Platform.iOS/Renderers/ShellSectionRenderer.cs#L492)

Otherwise there is a weird delayed animation with the navbar being hidden



> Shell Implementation

In a perfect world the shell implementation would be an attached property that could be applied to Shell,ShellItem, ShellSection, ShellContent, or ContentPage and utilize this mechanism

https://github.com/xamarin/Xamarin.Forms/blob/c79e187d61501f70ec8cdd40b5d3b447407980b0/Xamarin.Forms.Core/Shell/Shell.cs#L1039

Though I could see leaving that for a second PR

KSemenenko: I renamed the property now it's call IsPresented

PureWeen: Relying on the state of the application to be correct during imperative checks of IsPresented is going to be buggy

Tying *IsPresented* to the appearing event is also wrong because Appearing doesn't imply IsPresented it implies that it WILL be Presented 

We already have a lot of cases where the "active page" in multi page scenarios affect global features like the navbar or tabbars

Overall i'm not saying a property like *IsPresented* wouldn't be useful but if we're going to add a property like that it would need to be a separate PR that will probably involve a lot more discussion. 

Also, even if we added *IsPresented* my recommendation would be to add a check here 
https://github.com/xamarin/Xamarin.Forms/pull/8298/files#diff-5c15636b9e632ed12684a0c5ca1f29e8R96

that would basically be

```
else if (e.PropertyName == Page.IsPresented.PropertyName)
				UpdateStatusBarStyle();
```

Even with an *IsPresented* property we'd want the *UpdateStatusBarStyle* sets to happen from the event of IsPresented toggling

jamesmontemagno:
IMHO this should be in Xamarin.Essentials

https://github.com/xamarin/Essentials/pull/1054 i started randomly not knowing about this PR.

KSemenenko:
Hi @jamesmontemagno, maybe this will really be a good idea. I can transfer the entire implementation to your project.
what do you say?

Xamarin.Forms.Core/Page.cs
KSemenenko: @PureWeen added some code

KSemenenko:
@PureWeen What do you think about the "semi-private" property "HasAppered" in Page?
I can make it internal, but for this I need to add add assembly for visibility.

KSemenenko:
@PureWeen any chance that this pr will be ok?

PureWeen:
Yes!! I have this PR pulled down and I'll get you some feedback today. 
I have a couple more of yours at the top of my queue for this week as well 👍

KSemenenko:
Perhaps build errors are related to iOS and xcode version
https://github.com/xamarin/Xamarin.Forms/pull/8298#issuecomment-567063353

PureWeen:
@KSemenenko 

Can you tell me what you think of this?

https://github.com/xamarin/Xamarin.Forms/tree/8298_shane

PureWeen:
@KSemenenko also I only really fixed up Android at this point

just wanted to get your thoughts on the direction

KSemenenko:
@PureWeen thanks! let me check :)

jsuarezruiz:
@KSemenenko Could you rebase master to fix the conflict?. What is the status of the Shane feedback?

KSemenenko:
Hello @jsuarezruiz  I fixed the code.

KSemenenko:
But there’s something else to fix.

KSemenenko:
about ```if (Element.HasAppeared) ``` problem. While I was testing, I came to the conclusion that I can set this property before adding the page to the visual tree. but executing this command will change the color change the status bar of the current window. @jsuarezruiz @PureWeen

KSemenenko:
So, ```GetEffectiveValue``` this is a very cool idea. but unfortunately this is not enough.

KSemenenko:
I updated the branch and now the code works as it should. Pay attention ```TabbedPage Pages``` - each tab has its own color. @jsuarezruiz

KSemenenko:
If I need to fix something, please write to me, I will do everything quickly.

PureWeen:
@KSemenenko are your latest commits a merging of what I was thinking and what you have?

KSemenenko:
yes @PureWeen, that’s all I have with some of your fixes.
It seems that all this works as it should.
For example, pages in TabbedPage may have their own color.

KSemenenko:
@PureWeen  btw, thank for this 
```
if (pageContainer?.CurrentPage == null)
```

it's real cool!

KSemenenko:
Conflicts fixed.

sthewissen:
@KSemenenko I'm not sure if (or when) this could get merged into Xamarin.Forms. However, would you be willing to contribute this to the Xamarin.CommunityToolkit if it is decided that it won't make it in? I think people would love to be able to use this functionality sooner than later 👍 

https://github.com/xamarin/XamarinCommunityToolkit

KSemenenko:
Hello @sthewissen  
there are 2 ways
1. do it well, as part of the Page.
2. do it somehow.

I need to know exactly whether the page is showing now or not.
In my PR I added a property to the page. At first public (and I'm sure a lot of people are waiting for it), but it was rejected. then I added private and that's a problem too, so that's why it can not be merged.

So I have no idea how I can add this to XamarinCommunityToolkit.
Except as some kind of static extension method.

pictos:
@KSemenenko you can do it without `Page` control, on XCT you can implement just like you do when you need to access native APIs. 

```csharp
// shared layer
public static ChangeStatusBarColor(Color color) => PlatformChangeStatusBarColor(color);



//Platform Specific
static PlatformChangeStatusBarColor(color)
{
    // Do stuffs here
}

```

jsuarezruiz:
Could we move this to the Xamarin Community Toolkit?

jamesmontemagno:
I think that makes a lot of sense to move it over.

pictos:
Great, so @jamesmontemagno and @jsuarezruiz we already have an opened PR for this feature on XCT https://github.com/xamarin/XamarinCommunityToolkit/pull/812

So, can we close this one?

PR #8325, opened 2019-10-31
[CollectionView] Set default Layout ItemSizingStrategy value to MeasureFirstItem

### LINKED ISSUE #8314: [Bug]
Collection view does not add a default ItemSizingStrategy which defaults to MeasureAll.

This causes a major performance hit when using the GridLayout. With 100 items it adds a good 5 seconds to loading.  It should default to `MeasureFirstItem`

### Steps to Reproduce

1. 
2. 
3. 

### Basic Information

- Version with issue:
- Last known good version:
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Nuget Packages:
- Affected Devices:

jsuarezruiz:
### Description of Change ###

Set default Layout ItemSizingStrategy value to MeasureFirstItem. This has a positive impact on the performance and time required when rendering CollectionView (and CarouselView).

Captura de pantalla 2019-10-31 a las 14 42 05

_NOTE: Performing tests, the time difference using MeasureFirstItem or MeasureAllItems is an average of 110ms better with MeasureFirstItem._

### Issues Resolved ### 

- fixes #8314

### API Changes ###

The default Layout ItemSizingStrategy value is now MeasureFirstItem.

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###
Launch Core Gallery and navigate to the Issue8314. The time to load the elements must be below a second. The ItemSizingStrategy  property used in the Layout by default (not specified in the sample) must be MeasureFirstItem.
Change the ItemSizingStrategy to set MeasureAllItems and validate that the time required to load the items is higher. 

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue8314.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue8314.cs
@@ -1,0 +1,106 @@
+﻿using System.Collections.Generic;
+using System.Diagnostics;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.Forms.Core.UITests;
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if UITEST
+	[Category(UITestCategories.CollectionView)]
+#endif
+	[Preserve(AllMembers = true)]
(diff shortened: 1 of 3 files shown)


hartez:
The choice of "measure all" as the default was a deliberate one. It's the least surprising default for someone using the control for the first time; items are sized mostly as one would expect if they'd worked with Forms layouts in other contexts. And for most people in most situations, it performs well enough. 

If someone is looking for better performance and happens to be in a situation where all of their items can/should be the same size, they can opt for "measure first item" and get the performance benefit. 

If we are seeing a performance problem with the default strategy, step one is to investigate and hopefully improve the performance of that strategy. Let's get some better data from #8314 before we do anything hasty.

jfversluis:
Oh it was still a draft anyway. Oh well, now you're warned twice! 👻

PR #8343, opened 2019-10-31
Updating DependencyResolver to use TypedParameter

### LINKED ISSUE #3088: Update Dependency Resolver to include Dependency Type
### Description

Due to the need to include an `Android.Content.Context` for Android Renderers, the DependencyResolver uses `Resolve(Type type, params object[] args)` The problem with these args is that they are ultimately meant to be used by a DI Container to pass in as constructor arguments but we have no way of knowing that args[0] is meant to be Context without explicitly checking for the type which is not something that we can really do from shared code. It also means that this is not flexible to easily use across other platforms.

This becomes an issue as injecting a dependency at Resolve time generally requires that we know what the type is supposed to be.

```cs
// Autofac
parameters.Add(new TypedParameter(typeof(Android.Content.Context), context));
container.Resolve(type, parameters.ToArray());

// DryIoc
var resolver = container.Resolve<Func<Android.Content.Context, object>>(type);
resolver.Invoke(context);

// Unity
overrides.Add(new DependencyOverride(typeof(Android.Content.Context), context));
container.Resolve(type, overrides.ToArray());
```

### Proposal

Change object[] to another type that can be can wrap both the expected type and instance.

```cs
public class DependencyObject
{
    public Type Type { get; set; }
    public object Instance { get; set; }
}

DependencyResolver.Resolve(typeof(BoxRenderer), new DependencyObject{ Type = typeof(Context), Instance = someInstance });
```

This ensures that most DI Containers should be able to support this, and that it will be extensible from shared code moving forward regardless of what changes are made regardless of platform.

### Basic Information

- Version with issue: 3.1
- IDE: n/a

dansiegel:
### Description of Change ###

Updates the Dependency Resolver to take a new TypedParameter. This helps with providing a proper Dependency Injection tie in as containers prefer to know the type a specified instance is for. 

### Issues Resolved ### 

- fixes #3088

### API Changes ###

Public API changes are in the `DependencyResolver`

Added:
 - public static void ResolveUsing(Func<Type, TypedParameter[], object>)

Changed:
 - public static void ResolveUsing(Func<Type, object[], object) [Obsolete]
 
### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS (only via Core)
- Android (updates Platform create Renderer)
- UWP (only via Core)
- Tizen (updates GestureDetector create GestureHandler)

None

Not applicable

### Testing Procedure ###

no changes

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Registrar.cs
+++ b/Xamarin.Forms.Core/Registrar.cs
@@ -75,7 +75,7 @@ namespace Xamarin.Forms.Internals
 			return (TRegistrable)handler;
 		}
 
-		internal TRegistrable GetHandler(Type type, object source, IVisual visual, params object[] args)
+		internal TRegistrable GetHandler(Type type, object source, IVisual visual, params TypedParameter[] args)
 		{
 			if (args.Length == 0)
 			{
@@ -95,7 +95,7 @@ namespace Xamarin.Forms.Internals
 			return GetHandler(type) as TOut;
 		}
 
-		public TOut GetHandler<TOut>(Type type, params object[] args) where TOut : class, TRegistrable
+		public TOut GetHandler<TOut>(Type type, params TypedParameter[] args) where TOut : class, TRegistrable
 		{
 			return GetHandler(type, null, null, args) as TOut;
 		}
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Core/Registrar.cs
```diff
@@ -95,7 +95,7 @@ internal TRegistrable GetHandler(Type type, object source, IVisual visual, param
 			return GetHandler(type) as TOut;
 		}
 
-		public TOut GetHandler<TOut>(Type type, params object[] args) where TOut : class, TRegistrable
+		public TOut GetHandler<TOut>(Type type, params TypedParameter[] args) where TOut : class, TRegistrable
```
PureWeen: Since this is internal this is probably fine but we might still want to keep the old one.

dansiegel: I'm fine either way on this... I didn't provide an obsolete overload here because of the internal nature.

StephaneDelcroix:
We plan to deprecate our DependencyService in the near future. merging this now would send the wrong signal. So I'm closing this for now.

If our plans change and we keep our DependencyService, we'll reopen this.

PR #8372, opened 2019-11-05
GH-7015 - Implemented  handle for Placeholder Image

### LINKED ISSUE #7015: Image.ErrorSource and Image.LoadingSource
# API

## Description
- Don't require FileImageSource
- Don't modify `IImageViewHandler` the behavior should be implementable via the various Extension methods. UWP might need to be generalized a bit better for this

In a perfect world I think we'll also add these properties to ImageSource which will let users set these values if they define an ImageSource explicitly. Currently I think that'll be way too big of an effort and I think just having it on Image/ImageButton will suffice. Even if we were to add these to ImageSource I think string properties for these values on Image/ImageButton are valid for usability reasons. The majority of users will just want to set these values directly on image/imagebutton and we can see how often the use case for adding to imagesource comes up
 
## IImageElement

Add these to IImageElement
```C#
		ImageSource LoadingSource { get; }
		ImageSource ErrorSource { get; }
```

## ImageButton and Image
```C#
		ImageSource LoadingSource { get; }
		ImageSource ErrorSource { get; }
```

## Button
Button also implements IImageElement but for Button just implement these explicitly and return null

### Original Description
Hi,

Please add NoImage property ot the Image control which will have the image to be displayed when no image is loaded from database or https call.

For example, if a student page opened and the student has no photo on the server trhen it should look like this:

`NoImage="NoPhoto.png"
`

Thanks,
[Jassim](https://www.jassimrahma.net)

pictos:
### Description of Change ###

Added a Bindable Property called `PlaceholderImage`. If the `ImageSource `can't be loaded, the Image control will show the `PlaceholderImage`.

### Issues Resolved ### 

- fixes #7015

### API Changes ###

Added:
 - `ImageSource Image.LoadingPlaceholder { get; set; }` //Bindable Property
 - `ImageSource Image.ErrorPlaceholder{ get; set; }` //Bindable Property
Changed:
 
 Removed:
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

### PR Checklist ###


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7015.cs
```diff
@@ -0,0 +1,62 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using Xamarin.Forms.PlatformConfiguration;
+using Xamarin.Forms.PlatformConfiguration.AndroidSpecific;
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 6132, "Test the placeholder implementation for Image control ")]
```
jsuarezruiz: Here must use 7015 to identify this Issue. In other case we will get an exception trying to navigating to the Test Cases.

jsuarezruiz:
I know it's in Draft state (the iOS and UWP implementation is pending) but I want to leave some feedback.
I like the idea. With a `Placeholder` property, we could show the Placeholder  only in case of error or while the Source image is loading and in case of an error.
Another option is to create two Placeholder: `LoadingPlaceholder` and `ErrorPlaceholder`. This would allow to use different resources for loading and in case of error.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7015.cs
pictos: Fixed.

pictos:
@jsuarezruiz, tks for your review and suggestions. I agree with the names and already implemented those on Android.

For this feature, I'm assuming that:
* For Placeholders `ImageSource` we just allow local resources;
* This property just makes sense into `Image` control.

pictos:
@jsuarezruiz and @samhouts this PR is read to go❣

jsuarezruiz:
Hey @pictos, we have some conflicts here, could you check it out?

pictos:
For sure, I will take a look on it tomorrow, and ping you when it's done.

Xamarin.Forms.Platform.Android/Renderers/FileImageSourceHandler.cs
```diff
@@ -44,20 +44,39 @@ public Task LoadImageAsync(ImageSource imagesource, ImageView imageView, Cancell
...
+					return SetImagePlaceholderAsync(imageView, placeholder);
+				}
 			}
 
 			return Task.FromResult(true);
 		}
 
+		Task SetImagePlaceholderAsync(ImageView imageView, ImageSource placeholder, CancellationToken cancellation = default(CancellationToken))
+		{
+			string file = ((FileImageSource)placeholder).File;
```
rmarinho: What if it's uri image source?

pictos: For this feature, I'm assuming that images will be in the local resource. I think that no makes sense to get the error and loading PlaceholderImage from an external resources.

Xamarin.Forms.Platform.iOS/Renderers/ImageElementManager.cs
```diff
@@ -230,6 +242,9 @@ public static async Task SetImage(IImageVisualElementRenderer renderer, IImageEl
 					if (renderer.IsDisposed)
 						return;
 
+					if (uiimage == null)
+						uiimage = await errorPlaceholderSource.GetNativeImageAsync();
```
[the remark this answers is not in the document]
pictos: fixed

Xamarin.Forms.Platform.UAP/ImageElementManager.cs
```diff
@@ -169,15 +177,22 @@ public static async Task UpdateSource(IImageVisualElementRenderer renderer)
 				if (Control != null)
 					renderer.SetImage(imagesource);
 
-				RefreshImage(renderer);
+        RefreshImage(renderer);
```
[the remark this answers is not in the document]
pictos: Sorry... This happens when I fix the merge conflict in GitHub. Do you know how I can solve these conflicts locally?

pictos:
@rmarinho I did some updates on the code.

Answering your question... The LoadingPlaceholder will be presented when the device is loading the image (from the internet for example) and the ErrorPlaceholder will be presented when the `Image.Source` can't be displayed for some reason.

Xamarin.Forms.Platform.Android/Renderers/FileImageSourceHandler.cs
rmarinho: So i think we should check if it's a FileImageSource if it's not send a proper exception

pictos:
Hey, have been a while since I did this PR, I just want to confirm if it's everything ok or if you need me to review something, I will be very sad if this causes any regression or break something else.

MSiccDev:
On more suggestion before this goes stable - Please allow FontImageSource as well (this is also a local resource, right?)

Xamarin.Forms.Platform.Android/Renderers/FileImageSourceHandler.cs
rmarinho: @pictos this way we can't load FontImageSource right ?

pictos:
@MSiccDev can you please open an Issue for this?

MSiccDev:
@pictos done => #10876

pictos:
@MSiccDev tks,when the team approve this proposal I'll work on it.

PR #8397, opened 2019-11-06
[XamlC] Dispose assembly resolver constructed by XamlCTask

akihikodaki:
### Description of Change ###

An assembly resolver holds opened files, and leaks if it is not disposed. It is problematic if you run Xamarin.Forms.Xaml.UnitTests on Linux with the default configuration because it has a small limit for the number of opened files. It may cause stability issues also on other platforms, depending on the situation.

### Issues Resolved ### 

None

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###

Run `Xamarin.Forms.Xaml.UnitTests` with a debugging means like strace. You'll see the number of the file descriptor will not keep increasing while it does without this change.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/XamlCTask.cs
+++ b/Xamarin.Forms.Build.Tasks/XamlCTask.cs
@@ -52,212 +52,214 @@ namespace Xamarin.Forms.Build.Tasks
 				return true;
 			}
 
-			var resolver = DefaultAssemblyResolver ?? new XamlCAssemblyResolver();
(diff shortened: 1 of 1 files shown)


akihikodaki:
99e697192c29600dfc90f8e3cbe1c1aab3c633fd: rebased to 6eafaca536fd7184fbfceda33257b01880f734db.

PR #8403, opened 2019-11-06
Implement FormsStack and FormsQueue internally for VS 2017 Compatibility 

### LINKED ISSUE #8126: Investigate removing/changing all uses of Stack<T> and Queue<T>
## Summary
These types were moved 
https://github.com/mono/mono/pull/11511

If you compile anything on mono after this move it will fail to run on anything running mono before the move

Which is what led to the following issue https://github.com/xamarin/Xamarin.Forms/issues/7602. VS 2019 includes the moved types whereas VS 2017 (and VS Mac versions pre 8.3)  don't. So in order to utilize newer VS 2019 build tools but still remain VS 2017 compatible we'll need to remove all uses of Stack<T> and Queue<T>

As far as we know these are the only types that have been moved so in theory if we remove all uses of these types then we can start building our libraries against VS 2019 images which will let us build against Xcode11, Android10, and newer UWP APIs

## What to test after this change
In a perfect world we could run our Control Gallery built with VS 2017 against a nuget built with VS 2019 so that our whole suite of UI tests can validate.

In a world where time isn't infinite we'll want to dogfood the nuget built with VS 2019 on VS 2017. Make sure templates work, any dog food applications we have, smarthotel, etc...

### Copy the source code of Queue<T> and Stack<T> to our own types

The code for both of these is pretty simple
https://github.com/dotnet/corefx/blob/master/src/System.Collections/src/System/Collections/Generic/Stack.cs

https://github.com/dotnet/corefx/blob/master/src/System.Collections/src/System/Collections/Generic/Queue.cs

### Queue and stack api can be written on top of list
I expect the perf to be good enough (given that List perf is really good)

PureWeen:
### Description of Change ###
- Update everything back to using VS 2019 build components
- Copy in Stack<T> and Queue<T> implementations from corefx so that the moved types in mono don't break VS2017
   - I just linked the files into the platform projects so they can all stay internal/sealed. I assume once we totally drop VS 2017 we'll just delete all of this and roll back to using the packaged versions of these

Here are some comparison tests ran through Benchmark dotnet

|     Method |      Mean |     Error |    StdDev |
|----------- |----------:|----------:|----------:|
|  PushQueue | 27.780 ms | 0.5488 ms | 1.3148 ms |
|   PopQueue |  5.082 ms | 0.0992 ms | 0.0928 ms |
| PushFQueue | 26.596 ms | 0.5269 ms | 0.8046 ms |
|  PopFQueue |  2.397 ms | 0.0453 ms | 0.0378 ms |

|     Method |      Mean |     Error |    StdDev |    Median |
|----------- |----------:|----------:|----------:|----------:|
|  PushStack | 28.324 ms | 1.0790 ms | 3.0433 ms | 26.999 ms |
|   PopStack |  4.148 ms | 0.0827 ms | 0.0813 ms |  4.125 ms |
| PushFStack | 26.596 ms | 0.6836 ms | 1.2670 ms | 26.244 ms |
|  PopFStack |  2.086 ms | 0.0454 ms | 0.0819 ms |  2.053 ms |

Here's the test project if you would like to verify or add additional tests

[TestCollections.zip](https://github.com/xamarin/Xamarin.Forms/files/3816504/TestCollections.zip)

### Issues Resolved ### 
- fixes #8126

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

### Testing Procedure ###
- pull down the nuget and make sure you can build and deploy projects on VS 2017

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Xamarin.Forms.Platform.Android.csproj
+++ b/Xamarin.Forms.Platform.Android/Xamarin.Forms.Platform.Android.csproj
@@ -64,7 +64,6 @@
     <Compile Include="..\Xamarin.Forms.Core\corefx\FormsStack.cs">
       <Link>corefx\FormsStack.cs</Link>
     </Compile>
-    <Compile Include="Anticipator.cs" />
     <Compile Include="ActivityResultCallbackRegistry.cs" />
     <Compile Include="AndroidApplicationLifecycleState.cs" />
     <Compile Include="AndroidTitleBarVisibility.cs" />
--- a/Xamarin.Forms.Platform.Android/Xamarin.Forms.Platform.Android.csproj
+++ b/Xamarin.Forms.Platform.Android/Xamarin.Forms.Platform.Android.csproj
@@ -58,6 +58,13 @@
     <Reference Include="System.Net.Http" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="..\Xamarin.Forms.Core\corefx\FormsQueue.cs">
+      <Link>corefx\FormsQueue.cs</Link>
+    </Compile>
+    <Compile Include="..\Xamarin.Forms.Core\corefx\FormsStack.cs">
+      <Link>corefx\FormsStack.cs</Link>
+    </Compile>
+    <Compile Include="Anticipator.cs" />
     <Compile Include="ActivityResultCallbackRegistry.cs" />
     <Compile Include="AndroidApplicationLifecycleState.cs" />
     <Compile Include="AndroidTitleBarVisibility.cs" />
--- a/Xamarin.Forms.Platform.UAP/Xamarin.Forms.Platform.UAP.csproj
+++ b/Xamarin.Forms.Platform.UAP/Xamarin.Forms.Platform.UAP.csproj
@@ -11,7 +11,7 @@
     <AssemblyName>Xamarin.Forms.Platform.UAP</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion>10.0.16299.0</TargetPlatformVersion>
+    <TargetPlatformVersion>10.0.18362.0</TargetPlatformVersion>
     <TargetPlatformMinVersion>10.0.16299.0</TargetPlatformMinVersion>
     <MinimumVisualStudioVersion>14</MinimumVisualStudioVersion>
     <FileAlignment>512</FileAlignment>
@@ -275,7 +275,7 @@
     </PackageReference>
   </ItemGroup>
   <ItemGroup>
-    <SDKReference Include="WindowsMobile, Version=10.0.16299.0">
+    <SDKReference Include="WindowsMobile, Version=10.0.18362.0">
       <Name>Windows Mobile Extensions for the UWP</Name>
     </SDKReference>
   </ItemGroup>
--- a/Xamarin.Forms.Platform.UAP/Xamarin.Forms.Platform.UAP.csproj
+++ b/Xamarin.Forms.Platform.UAP/Xamarin.Forms.Platform.UAP.csproj
@@ -11,7 +11,7 @@
     <AssemblyName>Xamarin.Forms.Platform.UAP</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion>10.0.18362.0</TargetPlatformVersion>
+    <TargetPlatformVersion>10.0.16299.0</TargetPlatformVersion>
     <TargetPlatformMinVersion>10.0.16299.0</TargetPlatformMinVersion>
     <MinimumVisualStudioVersion>14</MinimumVisualStudioVersion>
     <FileAlignment>512</FileAlignment>
@@ -275,7 +275,7 @@
     </PackageReference>
   </ItemGroup>
   <ItemGroup>
-    <SDKReference Include="WindowsMobile, Version=10.0.18362.0">
+    <SDKReference Include="WindowsMobile, Version=10.0.16299.0">
       <Name>Windows Mobile Extensions for the UWP</Name>
     </SDKReference>
   </ItemGroup>
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -18,7 +18,7 @@ variables:
 - name: DOTNET_VERSION
   value: 3.0.100
 - name: winVmImage
-  value: Hosted Windows 2019 with VS2019
+  value: XamarinForms
 
 resources:
   repositories:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -15,10 +15,6 @@ variables:
   value: 5.0.2
 - name: DOTNET_SKIP_FIRST_TIME_EXPERIENCE
   value: true
-- name: DOTNET_VERSION
-  value: 3.0.100
-- name: winVmImage
-  value: Hosted Windows 2019 with VS2019
 
 resources:
   repositories:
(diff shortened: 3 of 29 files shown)


rmarinho:
Can we have 2 pr's ? one just for the Stack and Queue and other with the other changes? 

Also we should format the corefx code with our guidelines, drop private etc..

PureWeen:
@rmarinho the whole of this PR is to create a nuget that's built on VS 2019 that we can test works on VS 2017

If we split out all the changes we can't really do that

> Also we should format the corefx code with our guidelines, drop private etc..

yea I wasn't sure. This isn't really code that we are going to iterate or own we're just going to delete it sometime next year so it didn't seem worth the effort to format according to all our guidelines

samhouts:
Let's leave the code style the same as corefx, just in case we need to pull in an update.

PureWeen:
I tested the nuget with the sample given here and it now runs with these changes

https://github.com/xamarin/Xamarin.Forms/issues/7602#issuecomment-534355918

Xamarin.Forms.Platform.Android/Xamarin.Forms.Platform.Android.csproj
```diff
@@ -58,6 +58,13 @@
     <Reference Include="System.Net.Http" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="..\Xamarin.Forms.Core\corefx\FormsQueue.cs">
+      <Link>corefx\FormsQueue.cs</Link>
+    </Compile>
+    <Compile Include="..\Xamarin.Forms.Core\corefx\FormsStack.cs">
+      <Link>corefx\FormsStack.cs</Link>
+    </Compile>
+    <Compile Include="Anticipator.cs" />
```
PureWeen:
```suggestion
```

azure-pipelines.yml
```diff
@@ -15,6 +15,10 @@ variables:
   value: 5.0.2
 - name: DOTNET_SKIP_FIRST_TIME_EXPERIENCE
   value: true
+- name: DOTNET_VERSION
+  value: 3.0.100
+- name: winVmImage
+  value: Hosted Windows 2019 with VS2019
```
rmarinho: This will keeps us from updating let's say for XamarinForms (private agents) i prefer we had this as settable from the pipeline

PureWeen: Yea once we are happy with the PR I was going to delete this before merging it.

Though I still think this should be part of the yml file not the pipeline. For example if branch 4.1.0 is set to VS 2017 than when you run those builds it'll just build and work. Then when you run 4.2.0 it'll just build and work. If it's settable at the pipeline level than you are required to have knowledge external to the repository in order to make something compile.

But we can discuss that more on Thursday :-)

rmarinho: we can delete i m happy with it :P Let's merge

PureWeen: Sounds good!! I want to run some quick tests on UWP to see if incrementing the SDK build will break the icon on MDP even more or if the issue is the same.

Xamarin.Forms.Platform.UAP/Xamarin.Forms.Platform.UAP.csproj
```diff
@@ -11,7 +11,7 @@
     <AssemblyName>Xamarin.Forms.Platform.UAP</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion>10.0.16299.0</TargetPlatformVersion>
+    <TargetPlatformVersion>10.0.18362.0</TargetPlatformVersion>
```
samhouts: Why is this happening?

MagicAndre1981: maybe because VS2019 installs this last/newest SDK by default and not all older Win10 SDKs.

PureWeen: We had this before we rolled back to VS 2017 but the main reason has to do with WinUI functioning better when set to 10.0.18362.0 as the TargetPlatformVersion

And in theory it shouldn't matter increasing this since we still maintain the same mintarget

PureWeen: @samhouts regardless though I just removed these changes. Because we no longer specify the build host on the YML we have to wait until all the build hosts are updated to VS2019 otherwise once we merge this everything will break on UWP

azure-pipelines.yml
PureWeen: @rmarinho I added these back in for the time being because FormsQueue and FormsStack uses C# features that need 2019. I don't really want to modify them so they compile with C# 7.0 I'd rather just leave them untouched.

So I figure once this is merged in and the build server pipeline variables are updated then we can remove the fields?

```
Xamarin.Forms.Core\corefx\FormsQueue.cs(263,19): Error CS8107: Feature 'default literal' is not available in C# 7.0. Please use language version 7.1 or greater.
Xamarin.Forms.Core\corefx\FormsQueue.cs(278,14): Error CS8107: Feature 'default literal' is not available in C# 7.0. Please use language version 7.1 or greater.
Xamarin.Forms.Core\corefx\FormsQueue.cs(285,19): Error CS8107: Feature 'default literal' is not available in C# 7.0. Please use language version 7.1 or greater.
Xamarin.Forms.Core\corefx\FormsStack.cs(344,23): Error CS8107: Feature 'default literal' is not available in C# 7.0. Please use language version 7.1 or greater.
Xamarin.Forms.Core\corefx\FormsStack.cs(374,24): Error CS8107: Feature 'default literal' is not available in C# 7.0. Please use language version 7.1 or greater.
Xamarin.Forms.Core\corefx\FormsStack.cs(404,23): Error CS8107: Feature 'default literal' is not available in C# 7.0. Please use language version 7.1 or greater.
Xamarin.Forms.Core\corefx\FormsQueue.cs(263,19): Error CS8107: Feature 'default literal' is not available in C# 7.0. Please use language version 7.1 or greater.
Xamarin.Forms.Core\corefx\FormsQueue.cs(278,14): Error CS8107: Feature 'default literal' is not available in C# 7.0. Please use language version 7.1 or greater.
Xamarin.Forms.Core\corefx\FormsQueue.cs(285,19): Error CS8107: Feature 'default literal' is not available in C# 7.0. Please use language version 7.1 or greater.
Xamarin.Forms.Core\corefx\FormsStack.cs(344,23): Error CS8107: Feature 'default literal' is not available in C# 7.0. Please use language version 7.1 or greater.
```

build.cake
```diff
@@ -43,42 +44,48 @@ var informationalVersion = gitVersion.InformationalVersion;
 var buildVersion = gitVersion.FullBuildMetaData;
 var nugetversion = Argument<string>("packageVersion", gitVersion.NuGetVersion);
 
-var ANDROID_HOME = EnvironmentVariable ("ANDROID_HOME") ??
```
PureWeen: @rmarinho all these cake changes just make this cake script consistent with master

azure-pipelines.yml
rmarinho:
```suggestion
  value: XamarinForms
```

MagicAndre1981:
what will be the first build that includes the fix, so that I can test it with VS2017?

PureWeen:
@MagicAndre1981 
It should be up with the nightly in a couple of nights otherwise you can download the nuget from this PR here

https://dev.azure.com/xamarin/public/_build/results?buildId=10002&view=artifacts

and test it out

MagicAndre1981:
@PureWeen 

I tested the version. In my current working copy the nuget package fails first with `XF001 "Xamarin.Forms targets have been imported multiple times. Please check your project file and remove the duplicate import(s)."` Trying to set `<XFDisableTargetsValidation>true</XFDisableTargetsValidation>` causes error that I need to reference MonoAndroid90 🤔 🤷‍♂️

So I made a new checkout from my SVN source server and here the 4.2.0.910325 works with VS2017, the Android app compiles and runs fine so far.

PR #8408, opened 2019-11-07
Shell Navigation Bar Has Shadow property

PureWeen:
### Description of Change ###
Add an API to show and hide NavBar shadow on Shell. 

I also removed this check
https://github.com/xamarin/Xamarin.Forms/pull/8408/files#diff-8222d0783ad0aa5666a6d9d008a474b6L205

Because it was causing the settings to not propagate when you pushed a new page onto the stack. 

I don't really know what that check was trying to accomplish in the first place.

### API Changes ###

Added:
 - Shell.NavBarHasHadow

#### iOS
- ShellNavBarAppearanceTracker (basically a rename of SafeShellNavBarAppearanceTracker)
- ShellNavBarAppearanceTracker.SetHasShadow

### Platforms Affected ### 
- iOS
- Android
- Core

### Testing Procedure ###
- go into the store shell and play with hiding showing nav bar shadow. 
- go into the store shell and play with hiding showing nav bar 

### PR Checklist ###


jsuarezruiz:
Tested on Android and iOS. LGTM.
navbar-shadow-android
nav-shadow-ios

PR #8437, opened 2019-11-08
[Build] Update version numbers to ship from yaml

rmarinho:
Fix version numbers to ship nugets from YAML

### Description of Change ###

Add counters for old releases with counters starting with existing release numbers
### Issues Resolved ### 

Allows us to drop legacy build pipeline and just use yaml builds.

### API Changes ###

 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###

### DIFF
--- a/Version.targets
+++ b/Version.targets
@@ -13,7 +13,7 @@
           DependsOnTargets="GitVersion"
           Returns="$(Version)"
          >
-    <PropertyGroup> 
+    <PropertyGroup>
       <PreOut>$([System.Text.RegularExpressions.Regex]::Match($(GitTag), $(RegexPre)))</PreOut>
       <GitSemVerLabel Condition=" '$(GitTag)' != '' and $([System.Text.RegularExpressions.Regex]::IsMatch('$(GitTag)', $(RegexPre))) ">$(PreOut)</GitSemVerLabel>
       <GitSemVerLabel  Condition="$(CI) and '$(BUILD_REASON)' == 'Schedule'">$(NightlyTag)</GitSemVerLabel>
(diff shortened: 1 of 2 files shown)


Version.targets
```diff
@@ -1,22 +1,24 @@
...
 
   <Target Name="SetVersions"
           BeforeTargets="PrepareForBuild;GetAssemblyVersion;GetPackageVersion;Pack"
           DependsOnTargets="GitVersion"
           Returns="$(Version)"
          >
-
+    <PropertyGroup>
+      <PreOut>$([System.Text.RegularExpressions.Regex]::Match($(GitTag), $(RegexPre)))</PreOut>
+      <GitSemVerLabel Condition=" '$(GitTag)' != '' and '$(GitTag)' == '$(GitBranch)' and $([System.Text.RegularExpressions.Regex]::IsMatch('$(GitTag)', $(RegexPre))) ">$(PreOut)</GitSemVerLabel>
```
rmarinho: this checks if we are bulding a release, because GitTag uses a tag from origin (so it could have some value even if we are just building the branch), but if the GitTag is the same as the GitBranch it means we are building tag

Version.targets
```diff
@@ -34,7 +36,11 @@
       <VersionMetadataPlusLabel Condition="'$(VersionMetadataLabel)' != ''">+$(VersionMetadataLabel)</VersionMetadataPlusLabel>
       <Version>$(GitBaseVersionMajor).$(GitBaseVersionMinor).$(GitBaseVersionPatch)</Version>
       <PackageVersion>$(GitSemVerMajor).$(GitSemVerMinor).$(GitSemVerPatch)$(GitSemVerDashLabel)$(VersionMetadataPlusLabel)</PackageVersion>
+
       <PackageVersion Condition="$(CI)">$(GitSemVerMajor).$(GitSemVerMinor).$(GitBaseVersionPatch).$(BUILDVERSION)$(GitSemVerDashLabel)$(VersionMetadataPlusLabel)</PackageVersion>
+      <PackageVersion Condition="$(CI) and '$(GitBaseVersion)' == '4.2.0'">$(GitSemVerMajor).$(GitSemVerMinor).$(GitBaseVersionPatch).$(BUILDVERSION42)$(GitSemVerDashLabel)$(VersionMetadataPlusLabel)</PackageVersion>
```
rmarinho: We can remove this later, for now so we need these counters so when we ship we keep the order of new nugets , we could also branch for 4.2.1, and 4.3.1 and 4.4.1. and drop this.

But in long term we should really just use SemVer 2.0

azure-pipelines.yml
```diff
@@ -54,37 +69,36 @@ jobs:
     msbuildExtraArguments: '/nowarn:VSX1000 /p:CreateAllAndroidTargets=true /bl:$(Build.ArtifactStagingDirectory)\win.binlog'
     buildConfiguration: $(DefaultBuildConfiguration)
     buildPlatform: $(DefaultBuildPlatform)
```
PureWeen: I think most of the code below this line doesn't needs to be changed does it?

rmarinho: nop ,

PR #8446, opened 2019-11-09
WPF GTK Set flags

VladislavAntonyuk:
# Description of Change ###

Allows you to set experimental flags on WPF and GTK so that the application does not crash.

### API Changes ###

Added:
 - void SetFlags(params string[] flags)
 - IReadOnlyList<string> Flags {get;}
 

### Platforms Affected ### 

- GTK
- WPF

### Testing Procedure ###

Run ControlGallery, open CollectionView. 
Previous behavior - exception is thrown
Actual behavior - Application is running 

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.ControlGallery.GTK/Program.cs
+++ b/Xamarin.Forms.ControlGallery.GTK/Program.cs
@@ -21,6 +21,7 @@ namespace Xamarin.Forms.ControlGallery.GTK
             GtkOpenGL.Init();
             GtkThemes.Init();
             Gtk.Application.Init();
+            Forms.SetFlags("CarouselView_Experimental", "CollectionView_Experimental");
             FormsMaps.Init(string.Empty);
             Forms.Init();
             var app = new App();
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.ControlGallery.GTK/Program.cs
```diff
@@ -21,6 +21,7 @@ static void Main(string[] args)
             GtkOpenGL.Init();
             GtkThemes.Init();
             Gtk.Application.Init();
+            Forms.SetFlags("CarouselView_Experimental", "CollectionView_Experimental");
```
jsuarezruiz: The experimental label `"CollectionView_Experimental"` is no longer necessary (4.3). Currently, the CollectionView sample would not fail using the master branch.

VladislavAntonyuk: I found this flag in ExperimentalFlags.cs. Should I delete CollectionViewExperimental?

jsuarezruiz: No, it is not necessary in this case.

PR #8465, opened 2019-11-12
Color.IsDark + Color.IsLight

KSemenenko:
### Description of Change ###
With release of the black theme and the ability to change the background of the StatusBar #8298, 
it would be nice to know if the color contrast is dark or light.
Therefore, I think it will be a good addition to Color.

Calculating Color Contrast: https://24ways.org/2010/calculating-color-contrast

That mean, if our color is ``` Color.Black.IsDark ``` we need to use light text. 
So this is the opposite of the link above.

### API Changes ###

Added:
 - bool Color.IsDark {get;} 
 - bool Color.IsLight{get;} 

``` cs
Assert.IsTrue(Color.Black.IsDark); 
Assert.IsTrue(Color.Red.IsDark); 
Assert.IsTrue(Color.DarkBlue.IsDark);
//or
Assert.IsTrue(Color.White.IsLight); 
Assert.IsTrue(Color.LemonChiffon.IsLight); 
Assert.IsTrue(Color.Pink.IsLight);
```

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###

Run Unit Tests

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Color.cs
+++ b/Xamarin.Forms.Core/Color.cs
@@ -406,7 +406,7 @@ namespace Xamarin.Forms
 			var r = _r * 255;
 			var g = _g * 255;
 			var b = _b * 255;
-			return Convert.ToInt32(Math.Round((299 * r + 587 * g + 114 * b) / 1000f));
+			return Convert.ToInt32(((r * 299) + (g * 587) + (b * 114)) / 1000f);
 		}
 
 		public bool IsDark
--- a/Xamarin.Forms.Core/Color.cs
+++ b/Xamarin.Forms.Core/Color.cs
@@ -400,7 +400,27 @@ namespace Xamarin.Forms
 		{
 			return new Color(h, s, l, a, Mode.Hsl);
 		}
+
+		int CalculateYiqLuma()
+		{
+			var r = _r * 255;
+			var g = _g * 255;
+			var b = _b * 255;
+			return Convert.ToInt32(Math.Round((299 * r + 587 * g + 114 * b) / 1000f));
+		}
+
+		public bool IsDark
+		{
+			get => CalculateYiqLuma() < 128;
+		}
+
+		public bool IsWhite
+		{
+			get => CalculateYiqLuma() > 128;
+		}
+
 #if !NETSTANDARD1_0
+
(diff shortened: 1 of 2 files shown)


Xamarin.Forms.Core/Color.cs
```diff
@@ -400,7 +400,27 @@ public static Color FromHsla(double h, double s, double l, double a = 1d)
...
+		}
+
+		public bool IsDark
+		{
+			get => CalculateYiqLuma() < 128;
+		}
+
+		public bool IsWhite
+		{
+			get => CalculateYiqLuma() > 128;
```
VladislavAntonyuk: what if value == 128?

KSemenenko: it will be black

VladislavAntonyuk: but IsDark will return false.
I think it should be
```csharp
public bool IsDark 




		{ 
			get => CalculateYiqLuma() <= 128; 
		}
```

jcmanke:
Maybe `IsLight` rather than `IsWhite`?

Xamarin.Forms.Core/Color.cs
KSemenenko: Yes you are right

jcmanke: The formula in your link considers >= 128 to be light.

KSemenenko: @jcmanke I had to change this. because we show the property of our color, and not which color we need to use.
In the example, the function returns the color of the text.

KSemenenko: @jcmanke 
in the test:
``` cs
Assert.IsTrue(Color.FromHex("#FFCC00").IsLight); 
```
This means that our color is light and we need to use dark text.

KSemenenko: @jcmanke  Thank you for your attention and that you noticed an inverted condition.

StephaneDelcroix:
Contrast is relative, so a color isn't dark or light. We could add code to compute contrast, but we won't, as we already have it :)

converts your color to HSL, use the `L` channel, and do your choice of foreground color base on it

PR #8474, opened 2019-11-12
Fixes issue #7300 in XF 4.3.0 branch

### LINKED ISSUE #7300: [Bug] [Android] Content rendered above a Material button is not visible when button is pressed
### Description
When you render some content above a Material button, pressing the button on Android will give the button a higher elevation than the other content, hiding said content.
We've had a few issues around this area and they've all been closed even though the fixes didn't fix this particular scenario.

related information
PR #6271, ISSUE #6262, ISSUE #5642

@pureween commented on 5642

> this one is partially fixed by this PR
#6271
the main side effect that exists is once you click on the button then the labels disappear under the button. The material renderer has a dynamic elevation so the elevation increases when it's clicked

### Steps to Reproduce

1. Create a Layout with some input transparent content overlaying a material button.
1. Tap the button.

### Expected Behavior
The content above the button is still visible when pressing the button.
### Actual Behavior
The button is rendered above the content.

### Basic Information

- Version with issue: 3.6.0.264807 all the way to 4.2.0
- Platform Target Frameworks: 
  - Android:  

### Reproduction Link
I'm linking the repro project from issue 5642 as the issue is still there.
https://github.com/xamarin/Xamarin.Forms/files/2990085/ButtonOverlayIssue.zip

jgold6:
### Description of Change ###

- Updates the Platform.Android VisualElementPackager.EnsureChildOrder method to elevate any elements that are after a Button in the LogicalChildren of a layout by a value of 100 so they will stay above the Button when Android OS animates the button by raising the translationZ value while the button is in the Pressed state.
- Adds Issue7300.cs file in Xamarin.Forms.Controls.Issues.Shared project that tests the issue.
 
### Issues Resolved ### 

- fixes #7300

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

Before this change, if a user has a button that is covered by other elements, when the button is clicked the button will briefly appear above the the other elements while the button is in the pressed state, but will return to being covered by the other elements when the Android animation finishes. After this fix, the button will remain below the other elements and should never be seen. 

In one test scenario, the button remained over the other elements after the button was clicked, but I could only see this behavior on an Android API 28 emulator running on a Mac. The same emulator on Windows acts as described above. However the fix in EnsureChildOrder method fixes both scenarios. 

Since this issue occurs so quickly, it is difficult to get a screen shot, so here is a zip of two screen recordings, one unfixed and one fixed:

[Screen_Recordings_of_Issue_7300.zip](https://github.com/xamarin/Xamarin.Forms/files/3838607/Screen_Recordings_of_Issue_7300.zip)

### Testing Procedure ###
1. Run the ControlGallery.Android project
2. Go to Test Cases
3. Search for 7300 and open the test
4. you will see three images, one each in a Grid, AbsoluteLayout, and RelativeLayout.
5. Each layout contains a Button and an Image, with the Image fully covering the button so you don't see the button at all. 
6. Click the Image and the underlying button will be clicked. (InputTransparent)
7. If the issue is fixed, you will not see the button text "If you see me, even briefly, the test has failed."

To see the original issue, in Xamarin.Forms.Platform.Android.VisualElementPackager.EnsureChildOrder method, change line 194 from  
`r.View.Elevation = xfButtonIndex > -1 ? elevationToSet + 100 : elevationToSet;`
to 
` r.View.Elevation = elevationToSet;`

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7300.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7300.cs
@@ -25,41 +25,58 @@ namespace Xamarin.Forms.Controls.Issues
 	{
 		protected override void Init()
 		{
+			// To test FastRenderer.ButtonRenderer, change the folowing to VisualMarker.Default
+			// and comment out setting the LegacyRenderers flag in FormsAppCompatActivity in Xamarin.Forms.ControlGallery.Android
 			Visual = VisualMarker.Material;
-			// Test with Grid
-			Grid theGrid = null;
-			theGrid = new Grid();
-			SetupLayout(theGrid);
+			// Test with Grid with overlap
+			Grid theGrid = new Grid();
+			SetupLayout(theGrid, true);
 
-			// Test with AbsoluteLayout
+			// Test with AbsoluteLayout with overlap
 			AbsoluteLayout absLayout = new AbsoluteLayout();
-			SetupLayout(absLayout);
+			SetupLayout(absLayout, true);
 
-			// Test with RelativeLayout
+			// Test with RelativeLayout with overlap
 			RelativeLayout relLayout = new RelativeLayout() { HeightRequest = 50 };
-			SetupLayout(relLayout);
+			SetupLayout(relLayout, true);
 
-			StackLayout stackLayout = new StackLayout();
-			stackLayout.Children.Add(new Label { Text = "Grid" });
+			// Test with Grid without overlap
+			Grid theGrid2 = new Grid();
+			SetupLayout(theGrid2, false);
+
+			// Test with AbsoluteLayout without overlap
+			AbsoluteLayout absLayout2 = new AbsoluteLayout();
+			SetupLayout(absLayout2, false);
+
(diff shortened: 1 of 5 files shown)


jgold6:
Hold off on this PR. 

I discovered that this fix has a side effect of applying a shadow to the elements that are above the button. 

Need to rethink this a bit.

jgold6:
After discussing this with @PureWeen, there are some deeper issues here.

Basically this PR can be ignored.

In a nutshell, raising the elevation of elements adds a shadow when Material is in use, so even before the changes in this PR, the fix for issue #5642 (PR #6271) affects the shadows of the controls that overlay and cover the buttons. This may or may not be desirable by the end developer. 

What really needs to happen is that if Android animates a control by default, and that control is under other controls, then we need to handle when Android changes the translationZ and apply the same translation to the controls that cover the animating control. 

So we need to check more than just the order of the XF controls in the layout, we need to determine if they actually cover the potential animating control, like a Button, and then and only then should they follow the animating controls translationZ values. 

More info on elevation and TranslationZ in Android: https://developer.android.com/training/material/shadows-clipping.html

jgold6:
Closing this pull request so I can file a new one.

PR #8494, opened 2019-11-14
[Android] Fix content rendering above a material/fast renderer button

### LINKED ISSUE #7300: [Bug] [Android] Content rendered above a Material button is not visible when button is pressed

jgold6:
### Description of Change ###

- Adds code in FastRenderer.ButtonRenderer and MaterialButtonRenderer to check if there are any elements that overlap the button, and if so block any changes to TranslationZ in order to keep the button from appearing on top of the covering element(s). 
- Adds Issue7300.cs file in Xamarin.Forms.Controls.Issues.Shared project that tests the issue.
 
### Issues Resolved ### 

- fixes #7300

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

Before this change, if a user has a button that is covered by other elements, when the button is clicked the button will briefly appear above the the other elements while the button is in the pressed state, but will return to being covered by the other elements when the Android animation finishes. After this fix, the button will remain below the other elements and should never be seen. 

In one test scenario, the button remained over the other elements after the button was clicked, but I could only see this behavior on an Android API 28 emulator running on a Mac. The same emulator on Windows acts as described above. However this PR fixes both scenarios. 

Since this issue occurs so quickly, it is difficult to get a screen shot, so here is a zip of two screen recordings, one unfixed and one fixed:

[Screen_Recordings_of_Issue_7300.zip](https://github.com/xamarin/Xamarin.Forms/files/3838607/Screen_Recordings_of_Issue_7300.zip)

### Testing Procedure ###
1. Run the ControlGallery.Android project
2. Go to Test Cases
3. Search for 7300 and open the test
4. you will see 6 images, two each in a Grid, AbsoluteLayout, and RelativeLayout, top 3 have image covering the button, bottom three the image and the button do not overlap.
5. Click the Image in the top 3 layouts and the underlying button will be clicked. (InputTransparent)
6. If the issue is fixed, you will not see the button text "If you see me, even briefly, the test has failed" when clicking any of the top 3 images

To see the original issue, you can just comment out the override for the TranslationZ property in MaterialButtonRenderer and FastRenderer.ButtonRenderer

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7300.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7300.cs
@@ -25,41 +25,58 @@ namespace Xamarin.Forms.Controls.Issues
 	{
 		protected override void Init()
 		{
+			// To test FastRenderer.ButtonRenderer, change the folowing to VisualMarker.Default
+			// and comment out setting the LegacyRenderers flag in FormsAppCompatActivity in Xamarin.Forms.ControlGallery.Android
 			Visual = VisualMarker.Material;
-			// Test with Grid
-			Grid theGrid = null;
-			theGrid = new Grid();
-			SetupLayout(theGrid);
+			// Test with Grid with overlap
+			Grid theGrid = new Grid();
+			SetupLayout(theGrid, true);
 
-			// Test with AbsoluteLayout
+			// Test with AbsoluteLayout with overlap
 			AbsoluteLayout absLayout = new AbsoluteLayout();
-			SetupLayout(absLayout);
+			SetupLayout(absLayout, true);
 
-			// Test with RelativeLayout
+			// Test with RelativeLayout with overlap
 			RelativeLayout relLayout = new RelativeLayout() { HeightRequest = 50 };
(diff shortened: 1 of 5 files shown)


PureWeen:
@jgold6 based on our conversations I'm going to close this

PR #8513, opened 2019-11-14
[Build]Remove provisionator

rmarinho:
Remove the provisionator so everyone can build using yaml


PureWeen:
Are you sure we want to delete provisionator?

rmarinho:
We aren't using  it, the gain of anyone can use this yaml I think is better then we needing something private. 
Also the other sdks will end having more public links anyway so we could always pull those with cake . 
Also with control of our bots, we can just spin a new VM, install what we want there. And have special vms for special features, and we just set that at build time to use those new/private features. 

the build is failing because we need provisioning: true now

PureWeen:
> Also with control of our bots, we can just spin a new VM, install what we want there. And have special vms for special features, and we just set that at build time to use those new/private features.

If we have two different builds that are linked to two different Android SDKS is that going to mess up the VM? Or are we going to need to create a VM per SDK?

rmarinho:
@PureWeen we will need to test I hope option 1 works ok, we just install and it works .. no need for 2 VMS

PR #8531, opened 2019-11-16
Fix dynamic heights for iOS Flyout Item Templates

### LINKED ISSUE #8290: [Bug] [iOS] [Shell] Flyout Cells Don't Fit DataTemplate Content
### Description
The [UIContainerView](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/UIContainerCell.cs) in the [ShellTableViewSource](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Renderers/ShellTableViewSource.cs) always remains a static size, regardless of the contents height. It appears that this occurs for both MenuItems and ShellItems.

### Steps to Reproduce

1. Open the application
2. View the flyout.
3. Each of the items in the flyout should be a different height due to the HeightRequest of the content in the cell, but each remains a single static height.

### Expected Behavior
Each cell would resize to fit the height of the content inside it (Similar to Android).

### Actual Behavior
The cell is a single static size.

### Basic Information

- Version with issue: 4.2, 4.3
- IDE: Visual Studio for Mac, Visual Studio
- Platform Target Frameworks: 
  - iOS: iOS 12.2 iOS 13.1

### Screenshots
Screen Shot 2019-10-29 at 9 26 01 AM
Screen Shot 2019-10-29 at 9 26 33 AM

### Reproduction Link
[FlyoutCellHeightBug.zip](https://github.com/xamarin/Xamarin.Forms/files/3785678/FlyoutCellHeightBug.zip)

PureWeen:
### Description of Change ###
- Calculate the height of the templates used on the flyout so that they will size correctly
- Expose the creation of the TableViewController and TableViewSource for the flyout so users can override these and supply custom implementations
- This also fixes a regression on iOS where currently the flyout is unscrollable if content is too large

### Issues Resolved ### 

- fixes #8290 

### Platforms Affected ### 
- iOS

The height behavior will change on iOS if the templates that have been supplied have a HeightRequest specified or if they have a height > 44. This makes the behavior match Android and it is more expected

### Testing Procedure ###
- compare store shell with and without this change to make sure results makes sense
- verify the menu item with a templated height has specified height
- verify android and ios are consistent

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/ShellFlyoutContentRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/ShellFlyoutContentRenderer.cs
@@ -18,16 +18,23 @@ namespace Xamarin.Forms.Platform.iOS
 
 		public ShellFlyoutContentRenderer(IShellContext context)
 		{
+			_shellContext = context;
+
 			var header = ((IShellController)context.Shell).FlyoutHeader;
 			if (header != null)
 				_headerView = new UIContainerView(((IShellController)context.Shell).FlyoutHeader);
-			_tableViewController = new ShellTableViewController(context, _headerView, OnElementSelected);
+
+			_tableViewController = CreateShellTableViewController();
 
 			AddChildViewController(_tableViewController);
 
 			context.Shell.PropertyChanged += HandleShellPropertyChanged;
 
-			_shellContext = context;
+		}
+
+		protected virtual ShellTableViewController CreateShellTableViewController()
+		{
+			return new ShellTableViewController(_shellContext, _headerView, OnElementSelected);
 		}
 
 		protected virtual void HandleShellPropertyChanged(object sender, PropertyChangedEventArgs e)
--- a/Xamarin.Forms.Platform.iOS/Renderers/ShellFlyoutContentRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/ShellFlyoutContentRenderer.cs
@@ -32,7 +32,7 @@ namespace Xamarin.Forms.Platform.iOS
 
 		}
 
-		protected virtual ShellTableViewController CreateShellTableViewController()
+		protected ShellTableViewController CreateShellTableViewController()
 		{
 			return new ShellTableViewController(_shellContext, _headerView, OnElementSelected);
 		}
--- a/Xamarin.Forms.Platform.iOS/Renderers/ShellTableViewController.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/ShellTableViewController.cs
@@ -15,18 +15,25 @@ namespace Xamarin.Forms.Platform.iOS
 		double _headerOffset = 0;
 		double _headerSize;
 		bool _isDisposed;
+		Action<Element> _onElementSelected;
 
 		public ShellTableViewController(IShellContext context, UIContainerView headerView, Action<Element> onElementSelected)
 		{
 			_context = context;
+			_onElementSelected = onElementSelected;
 			_headerView = headerView;
-			_source = new ShellTableViewSource(context, onElementSelected);
+			_source = CreateShellTableViewSource();
 			_source.ScrolledEvent += OnScrolled;
 			if (_headerView != null)
 				_headerView.HeaderSizeChanged += OnHeaderSizeChanged;
 			((IShellController)_context.Shell).StructureChanged += OnStructureChanged;
 
 			_context.Shell.PropertyChanged += OnShellPropertyChanged;
+		}
+
+		protected virtual ShellTableViewSource CreateShellTableViewSource()
+		{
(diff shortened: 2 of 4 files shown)


Xamarin.Forms.Platform.iOS/Renderers/ShellTableViewController.cs
```diff
@@ -59,22 +66,16 @@ void UpdateVerticalScrollMode()
...
-					var firstCellIsVisible = TableView.Bounds.Contains(firstCellRect);
-
-					var lastRowIndex = NMath.Max(0, TableView.NumberOfRowsInSection(0) - 1);
-					var pathToLastRow = Foundation.NSIndexPath.FromRowSection(lastRowIndex, 0);
-					var cellRect = TableView.RectForRowAtIndexPath(pathToLastRow);
-					var lastCellIsVisible = TableView.Bounds.Contains(cellRect);
-
-					TableView.ScrollEnabled = !firstCellIsVisible || !lastCellIsVisible;
+					TableView.ScrollEnabled = true;
+					TableView.AlwaysBounceVertical = false;
```
PureWeen: @paymicro is this ok?   From what I can tell your changes broke scrolling when set to Auto

Xamarin.Forms.Platform.iOS/Renderers/ShellFlyoutContentRenderer.cs
```diff
@@ -18,16 +18,23 @@ public class ShellFlyoutContentRenderer : UIViewController, IShellFlyoutContentR
...
 		public ShellFlyoutContentRenderer(IShellContext context)
 		{
+			_shellContext = context;
+
 			var header = ((IShellController)context.Shell).FlyoutHeader;
 			if (header != null)
 				_headerView = new UIContainerView(((IShellController)context.Shell).FlyoutHeader);
-			_tableViewController = new ShellTableViewController(context, _headerView, OnElementSelected);
+
+			_tableViewController = CreateShellTableViewController();
```
mjfreelancing: Is there any concern with calling a virtual method from the constructor?

PureWeen: Yea best practices dictate that I should move this :-) 

I think moving this to a lazy instantiation + wiring this up inside an override of *LoadView* would be the better way to put this together

mjfreelancing: Sounds like a much better plan

Xamarin.Forms.Platform.iOS/Renderers/ShellFlyoutContentRenderer.cs
```diff
@@ -18,16 +18,23 @@ public class ShellFlyoutContentRenderer : UIViewController, IShellFlyoutContentR
...
 
 			AddChildViewController(_tableViewController);
 
 			context.Shell.PropertyChanged += HandleShellPropertyChanged;
 
-			_shellContext = context;
+		}
+
+		protected virtual ShellTableViewController CreateShellTableViewController()
```
PureWeen:
```suggestion
		protected ShellTableViewController CreateShellTableViewController()
```

Xamarin.Forms.Platform.iOS/Renderers/ShellTableViewController.cs
```diff
@@ -29,6 +31,11 @@ public ShellTableViewController(IShellContext context, UIContainerView headerVie
 			_context.Shell.PropertyChanged += OnShellPropertyChanged;
 		}
 
+		protected virtual ShellTableViewSource CreateShellTableViewSource()
```
PureWeen:
```suggestion
		protected ShellTableViewSource CreateShellTableViewSource()
```

PureWeen:
> If a sub-class calls this method what can the caller do with the returned ShellTableViewController / ShellTableViewSource in order to change behavior? Or are you exposing the method for some other reason?

If you're commenting about this commit

https://github.com/xamarin/Xamarin.Forms/pull/8531/commits/11c5bd55d4b645aafca8ff9060d3bad278f8ef28

Virtual was just removed because we've stopped accepting any API changes to 4.4 so I'll make those virtual again in 4.5

> If a sub-class calls this method what can the caller do with the returned ShellTableViewController / ShellTableViewSource in order to change behavior

A subclass wouldn't really call these methods. Once these are made virtual again then a subclass would override them to provide subclassed implementations of *ShellTableViewController / ShellTableViewSource* which are much easier to customize if you can utilize a custom type. You can't customize much with an  iOS view/viewcontroller/class if you don't control the type

mjfreelancing:
Ah ok, I didn't realize the removal of virtual was temporary - makes sense now - that's what raised my eyebrows :-)

PR #6463, opened 2019-06-08
ITextAlignmentElement VerticalTextAlignment (Entry, SearchBar, EntryCell)

andreinitescu:
- Adds VerticalTextAlignment to ITextAlignmentElement
- Adds suport for vertical text alignment on Entry, EntryCell, SearchBar

### API Changes ###

Added:
 - TextAlignment ITextAlignmentElement.VerticalTextAlignment{ get; set; } //Bindable Property
 - TextAlignment Entry.VerticalTextAlignment
 - TextAlignment EntryCell.VerticalTextAlignment
 - TextAlignment SearchBar.VerticalTextAlignment

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP
- WPF

### PR Checklist ###

- [ ] Has automated tests 
- [x] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard

PR #8536, opened 2019-11-17
Fix 8503, 8787 - text in Entry not immediately visible, or visible after IsVisible set to true

### LINKED ISSUE #8503: [Bug] Text is not visible on Entry element after animation finished
### Description
I have `Entry` element which contains a text value. That element is animated using translation effect. But the text value is not visible after animation finished and it will be visible until I click on that `Entry` element (focus) and then will be visible

### Steps to Reproduce
Click on Login_Action button and see Entry after animation finished.

### Expected Behavior
Expect to be visible. In previous 3.3.x version, it worked like a charm.

### Actual Behavior
Text value is not visible in Entry element after animation is done.

### Basic Information

- Version with issue:
- Last known good version:
- IDE: Visual Studio 2017
- Platform Target Frameworks: 
  - UWP:  17134
- Nuget Packages: Xamarin.Forms v4.3.0.947036

### Reproduction Link
See projects: 
[XamarinTests.zip](https://github.com/xamarin/Xamarin.Forms/files/3845673/XamarinTests.zip)

### LINKED ISSUE #8787: [Bug] Entry text initially invisible on UWP
### Description

On UWP, with a certain combination of views and layouts, there is a circumstance where text in an entry will initially be invisible. This can be remedied in the app by resizing the windows or clicking into the entry. Based on my investigation, there seems to be an issue where the the actual element that holds the text in UWP is not full size.

We also observed a similar but not quite the same issue when there are entries in a List View in a view cell where the text will not appear unless the cursor is brought near the entry. I don't have a repro for this yet but I bring it up because it might be related.

Finally, while I was able to make it appear by setting a height request, the behavior is not quite predictable. In our production apps, setting/not setting height requests was not able to change anything. In our production apps, we fixed it by setting a width request of 5 which for some reason resolved it in our environment.

### Steps to Reproduce

Repro project below

### Expected Behavior

Text should appear initially upon page load as it does on Android and iOS.

### Actual Behavior

The text is not initially visible in entries unless they are resized or interacted with.

### Basic Information

- Version with issue: 4.3.0.991221 (any version after the last known good one)
- Last known good version: 4.2.0.848062
- IDE: 16.4.0
- Platform Target Frameworks:
  - UWP:  1903, 18362
- Nuget Packages: Xamarin Forms 4.3.0.991221,
 Microsoft.NETCore.UniversalWIndowsPlatform 6.2.9,
Xamarin.Essentials 1.3.1

### Screenshots
Issue on UWP,
entryIssue

Issue not present on iOS as an example,
image

### Reproduction Link

https://github.com/squinn7/EntryBindingRepro

WayaFlyfeather:
### Description of Change ###

Adding a `ContentPresenter` to the `ScrollViewer` in FormsTextBoxStyle, as setting VerticalAlignment on the ScrollViewer led to unexpected results.

### Issues Resolved ### 

- fixes #8503
- fixes #8787 

### API Changes ###

 None

### Platforms Affected ### 

- UWP

Before this fix, if an Entry was changed from **IsVisible** = false to **IsVisible** true, text inside the entry would not actually become visible before it acquired focus.

[Edit January 30, 2020]: Other cases has popped up where entry text is seemingly randomly invisible at first. I have today confirmed that this fix also fixes issue #8787.

**Issue 8503**
Before 1 - before clicking button to make Entry visible
fix8503-before-1

Before 2 - after clicking button to make Entry visible. Not, the text is not visible.
fix8503-before-2

Before 3 - after clicking in the Entry, giving it focus - now showing text
fix8503-before-3

After 1 - before clicking button to make Entry visible
fix8503-after-1

After 2 - after clicking button to make entry visible. Note that the text is now visible
fix8503-after-2
 
**Issue 8787**
Before fix 1 - Page is loaded, text in entry invisible:
fix-8787-8503-before-1

Before fix 2 - after clicking in Entry to give it focus, text becomes visible;
fix-8787-8503-before-2

After fix - page is loaded, text in Entry is immediately visible:
fix-8787-8503-after

### Testing Procedure ###

**Issue 8503**
Go to test page for Issue 8503 in the ControlGallery.
Click the button _Click to show Entry for Isolated bug_.
Check that the entry appears, and it's text content is visible.
If not, check if it appears when the entry receives focus. If it does, the bug is not fixed.

A similar test is repated below this, this one mimics the original problem described in #8503, it should similarly show the text at once when the button is clicked.

**Issue 8787**
Go to test page for Issue 8787 in the ControlGallery.
Check that the text in the yellow Entry is immediately visible.
If it is, the issue is fixed. If it is not visible before the Entry gains focus or the window is resized, the issue is not fixed.

### Notes ###

This one is a bit weird.

I isolated the problem described in #8503, to occurring when an Entry started out as invisible, but was later changed to Visible. In that case it's text content would not be shown before it got focus. (The animation used in the original report changed visibility of hte item, which was the cause - it wasn't due to behaviors or animations).

I then determined that the "culprit" was this line introduced in #6463:
https://github.com/andreinitescu/Xamarin.Forms/blob/3481a842b45c463c1c0ff40c051e3c2484907c85/Xamarin.Forms.Platform.UAP/FormsTextBoxStyle.xaml#L224

If this VerticalAlignment was not present on the ScrollViewer, then everything worked fine. It also worked fine if I gave the Entry a HeightRequest, I discovered. But without a HeightRequest, the problem was there, even if changed the VerticalAlignment to "Center", i.e not binding it.

I could on Stack Overflow see that [others have had problems with strange behavior from a ScrollViewer in a TextBox with VerticalAlignment](https://stackoverflow.com/a/47909054).

Hence this proposed solution, instead og setting VerticalAlignment directly on the ScrollViewer, to have it wrap a ContentPresenter, and set VerticalContentAlignment on that instead.

It seems to be working fine, but to be perfectly honest I'm not 100% sure if I may have overlooked some potential problem with this solution.

### PR Checklist ###


WayaFlyfeather:
I have tested and confirmed that this PR also seems to fix #8787. I have updated this PR with a test page for that issue, and added screenshots.

bmacombe:
@WayaFlyfeather you might want to test the issue 2172 page against this branch also.  When I tested your changes with mine in #8214 the combo caused an issue at that time.  Since #8214 is merged already just want to make sure there is no regression there.

Looking forward to getting this merged, I've seen some of text visibility issues in UWP recently.

WayaFlyfeather:
@bmacombe Right, will do, good thinking!

bmacombe:
@WayaFlyfeather Great!, let me know if you need anything from me, glad to help.

WayaFlyfeather:
@bmacombe Yes, you are right, the two fixes do not appear to work well together, at least not out of the box. Not quite sure why, it is a bit interesting.

I know you've already done this, but for the benefit of the understanding for others (and myself), this is how the the test page for Issue 2172 looks with the current version of 4.5.0:
8214-current

Clearly, the green (your fix from #8214) looks fine, and the red (the old version of the renderers), not so fine.

Now, this is what happens when my fix in this PR is applied:
8214-with-8503

Now the green - your fix - is messed up. Intererstingly the red - the old renderers - are nearly all fine. Except for the auto-size editor, at the bottom.

And, as I understand it, your fix was basically to apply a technique already used on the EditorRenderer, where a static instance of a FormsTextBox is used to measure what the dimensions of a FormsTextBox should be, as apparently this is sometimes messed up if the FormsTextBox is within a ScrollView, and then apply this measurement to the actual FormsTextBox in the EditorRenderer/EntryRenderer?

So, as this technique was already in place for the EditorRenderer before your fix, it would appear that this is maybe the part that messes up my fix.

That lead me to an experiment - I deleted the `GetDesiredSize`-overrides from the EditorRenderer and the EntryRenderer, thus just using the inherited GetDesiredSize, and not bothering with measuring copies. I've pushed it as [branch exp-8214-problem](https://github.com/WayaFlyfeather/Xamarin.Forms/tree/exp-8214-problem) - this was the result:
8214-experiment

Which actually looks pretty good, I guess? This suggests to me, that maybe my fix inadvertently also fixes the problem behind #2172, and whatever lead to introduction of the static-copy-measurement in the EditorRenderer in the first place?

bmacombe:
@WayaFlyfeather If your fix takes care of the root issue behind the entry and editors awesome!  Trying to figure out the root issue behind the incorrect sizing was just a dead end for me, so my fix was mainly a hack that was already approved by being in the code base already.  

I remember the initial editor sizing method I borrowed and expanded upon was there to make sure an auto size editor would expand as it text wrapped.  So probably want to make sure that scenario is tested.

I'll try to pull down your new branch and play with it some also in the next couple days.

WayaFlyfeather:
@bmacombe Thanks, Brian, that would be great :)

The autosizing Editor in page 2172 does seem to work, actually - at least at first glance.

I'm not sure exactly what is happening, either - but it _does_ seem that it is somewhat more healthy to include a ContentPresenter in a ScrollViewer, instead of treating the ScrollViewer as the ContentPresenter.

WayaFlyfeather:
@bmacombe So, it seems this fix is not a miracle solution after all. I looked at the original PR for the AutoSize in the Editor, and could see that one of the problems occuring was when rapidly changing the size, such as when [holding down the return or back space keys](https://github.com/xamarin/Xamarin.Forms/blob/dfeb9f7f9122279101d3d37572629ba1b8d4192a/Xamarin.Forms.Platform.UAP/EditorRenderer.cs#L150).

The Editor with AutoSize in the 2172 test page does seem to handle holding down Return alright; unfortunately it seems when holding down back space some lines are not visually deleted, until next time the content is changed.

So I guess some solution will have to be found; either making the measuring code work with this fix, or find some other work around. Or perhaps make two styles for the FormsTextBox, one that is for multi line content and one for single lines.

Will see if I can come up with something :)

bmacombe:
@WayaFlyfeather If nothing else, maybe the measure method could be modified to remove PlaceholderTextContentPresenter from the control template for the static text box when it's created.  Saves having two control templates in case other changes are made in the future to it.

Seems a little hacky to me, but that whole measure using a static text box outside the visual tree feels like a hack.

samhouts:
Off topic comment....this collaboration is why I love this community so much! Thank you, @WayaFlyfeather and @bmacombe !!

PureWeen:
Build failure was just the OSX box timing out

bmacombe:
@PureWeen I just pulled down 4.5 after merging this. It regresses the fix in #8214 still.

image

the changes in #8214 could be undone but then the edit auto size is broken
image

You can see all of this in the #2172 issue page.

PureWeen:
@bmacombe my fault I had missed the discussion on this one 

Let me just revert this and reopen

PureWeen:
alright I've reverted this PR

https://github.com/xamarin/Xamarin.Forms/pull/9598

Once the other highlighted issues are addressed create a new PR

Thank you for the watchful eye on this one @bmacombe !!

bmacombe:
@PureWeen You're welcome!

@WayaFlyfeather Have you had a chance to dig into this more?  I might be able to find some time this weekend or next week.

WayaFlyfeather:
> @PureWeen You're welcome!
> 
> @WayaFlyfeather Have you had a chance to dig into this more? I might be able to find some time this weekend or next week.

@bmacombe Sorry, got suddenly very busy & caught up in some RL stuff. Will try to get back to this over the weekend.

@PureWeen Sorry for the trouble!

And an off-topic thank you, @samhouts! :)

PR #8610, opened 2019-11-21
[UWP] Fix MDP Button "Padding" issue

jfversluis:
### Description of Change ###

Added correct height for the `CommandBar` to make it use the full height again.

Looks like we've hit this issue on the UWP side: https://github.com/microsoft/microsoft-ui-xaml/issues/275

They decided to just leave it like this, so I implemented the suggested fix from this issue which works.

### Issues Resolved ### 

- fixes #5661

### API Changes ###
 
 None

### Platforms Affected ### 

- UWP

### Behavioral/Visual Changes ###

None

#### Before
Aantekening 2019-11-21 145933

#### After
Aantekening 2019-11-21 145620

### Testing Procedure ###

Add/remove this change and go to the Bugzilla 44453 issue in the Gallery app to verify

### PR Checklist ###

- [x] Targets the correct branch
- [x] Tests are passing (or failures are unrelated)

Issue #8602, opened 2019-11-21
[Bug] UWP NavigationPage.TitleView does not fill Header


YZahringer:
### Description
On UWP the TitleView does not take up all the available space. There is a margin at the bottom and left.

### Steps to Reproduce
````xaml
<ContentPage>
    <NavigationPage.TitleView>
        <Grid
            BackgroundColor="Red"
            HorizontalOptions="FillAndExpand"
            VerticalOptions="FillAndExpand" />
    </NavigationPage.TitleView>

    <Grid BackgroundColor="Yellow">

    </Grid>
</ContentPage> 
````

### Expected Behavior
Grid with `BackgroundColor="Red"` Fill Header without margins:
![image](https://user-images.githubusercontent.com/4254116/69325930-3e911f00-0c4b-11ea-98b1-ec861048bb3b.png)

### Actual Behavior
Grid has Left `10px` and Bottom `8px` Margin:
![image](https://user-images.githubusercontent.com/4254116/69321605-b78c7880-0c43-11ea-906d-7dfeca407807.png)
Live Visual Tree of Header:
![image](https://user-images.githubusercontent.com/4254116/69325161-ffae9980-0c49-11ea-96f5-a225fba8327c.png)
It is the `Border` in `TitleViewPresenter` that takes `width=10px` and `height=48px`. This one should have `Visibility=Collapsed`.

### Basic Information

- Version with issue: 4.3.0.991211
- IDE: 16.3.10
- Platform Target Frameworks: 
  - UWP:  18362

jfversluis:
Hm I wonder if this has some connection to what is going on in #5661

YZahringer:
@jfversluis Thanks! This also fixes the left "Padding" on TitleView?

jfversluis:
Hm that I don't think so, this focusses mostly on the bottom thing. But it should be in the same area

jfversluis:
Confirmed this happening in the attached repro project.

[Repro8602.zip](https://github.com/xamarin/Xamarin.Forms/files/3897212/Repro8602.zip)

WayaFlyfeather:
I've tried to look into this. It appears the implementation of the TitleView in UWP is such, that it co-exists with the TitleIcon and Title properties, and this is causing the problems.

This becomes more obvious if a Title is added to the repro @jfversluis provided: 
[...]

PR #8625, opened 2019-11-22
[UWP] Fix PageControl "Padding" issue

### LINKED ISSUE #5661: Xamarin Forms MasterDetailPage button padding issue in UWP
### Description

I'm currently learning Xamarin Forms and I have a small graphic issue. I have to fit the height of all the buttons to 100% (based on Navigation bar). Actually Xamarin cut some pixels in the bottom for the hamburger icon and the "three dots" icon is not set to 100% height.

### Steps to Reproduce

1. Create xamarin form project with  default template Master-Detail
2. Modify UWP target version to 17763
3. Run the app

### Expected Behavior
image

### Actual Behavior
image

### Basic Information

- Version with issue: UWP target Version 17763
- Last known good version:  UWP target Version 17134
- IDE: Visual Studio 2017
- Platform Target Frameworks: 
 - UWP:  The version of the UWP SDK 17763

### Reproduction Link

[Xamarin Forms MasterDetailPage button padding issue in UWP](https://stackoverflow.com/questions/55265495/xamarin-forms-masterdetailpage-button-padding-issue-in-uwp)

jfversluis:
### Description of Change ###

This is in addition to #8610. The padding was also there on other pages which used the `PageControlStyle.xaml` although it wasn't visible there.

This change removes the `MinHeight` from the style, which makes it use the default value and the padding then disappears.

### Issues Resolved ### 

- related to #5661

### API Changes ###
 
 None

### Platforms Affected ### 

- UWP

The `MinHeight` before was 48, now it will change to 40, maybe depending on which UWP SDK you target. I've checked the behavior with vanilla UWP in regard to how buttons and labels etc. are shown and that remains the same.

Not applicable

### Testing Procedure ###

Run the gallery app and navigate to B54977. Without the fix in place, when you press the overflow button (all the way to the right) you will notice in the animation that there is a thin line at the bottom. With this fix, that line is gone.

In addition, bump up the gallery app UWP target and run through both scenarios again. The animation has changed, the effect is the same.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/PageControlStyle.xaml
+++ b/Xamarin.Forms.Platform.UAP/PageControlStyle.xaml
@@ -18,7 +18,7 @@
 						</Grid.RowDefinitions>
 
 						<Border x:Name="TopCommandBarArea" HorizontalAlignment="Stretch" Background="{TemplateBinding ToolbarBackground}">
-							<uwp:FormsCommandBar x:Name="CommandBar" Background="{TemplateBinding ToolbarBackground}" MinHeight="{ThemeResource TitleBarHeight}" HorizontalAlignment="Stretch">
+							<uwp:FormsCommandBar x:Name="CommandBar" Background="{TemplateBinding ToolbarBackground}" HorizontalAlignment="Stretch">
 								<uwp:FormsCommandBar.Content>
 									<Border x:Name="TitleArea" Height="{ThemeResource TitleBarHeight}" Visibility="{TemplateBinding TitleVisibility}" HorizontalAlignment="Stretch">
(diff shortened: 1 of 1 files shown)


WayaFlyfeather:
Actually, I'm not sure this is the best solution, as I discovered when looking into fixing #8602. It seems if MinHeight is removed, we do avoid the padding at the bottom, but it is at the cost at clipping the title area. Instead I would propose using your resources from #8610, as they accomplish both avoiding the padding, and maintaining the size.

These screen shots should demonstrate:

A normal Title before any fixes:


A normal Title after removing MinHeight (this fix):


A normal title using the _uwp:CommandBar.Resources_ from #8610:


Besides the Title area being smaller (which may be an inconvenience for some designs, I imagine), the title is not centered anymore.

YZahringer:
any news on that?

samhouts:
I'm going to close this one for now until we're ready to return to it.

Issue #4715, opened 2018-12-12
[Android 9] System.InvalidOperationException: Sequence contains no elements


tonholis:
### Description

I upgraded an app to use the latest version of Xamarin.Forms and I caught the error described here. I couldn't reproduce it on my development device (Huawei CAN-L01 / Android 6.0). 
A user which noticed the crash explained to me that he opens the app, put it in background and open it again. So it happens when a user resumes the app.

### Steps to Reproduce

1. Open a Xamarin.forms app
2. Put it on background
3. Resume the app

### Expected Behavior

The app should resume successfully.

### Actual Behavior

It crashes when a user resumes the app on a Oneplus 6 (A6003) with Android 9

### Basic Information

- Version with issue: Xamarin.Forms 3.4.0.1008975
- IDE: Visual Studio for Mac 7.6.10 (build 27)
- Platform Target Frameworks: 
  - Android: 9
- Nuget Packages:
- Affected Devices: Oneplus 6 (A6003) with Android 9

### Screenshots

```
Manufacturer: OnePlus
Model: ONEPLUS A6003

Xamarin Exception Stack:
System.InvalidOperationException: Sequence contains no elements
  at System.Linq.Enumerable.Last[TSource] (System.Collections.Generic.IEnumerable`1[T] source) [0x00010] in <0f5fc5e6c8e94bd581de5f77b09899a9>:0
  at Xamarin.Forms.Platform.Android.AppCompat.Platform.HandleBackPressed (System.Object sender, System.EventArgs e) [0x00015] in <81626d5b2a054a23bc21d192df515fa9>:0
  at Xamarin.Forms.Platform.Android.FormsAppCompatActivity.OnBackPressed () [0x00007] in <81626d5b2a054a23bc21d192df515fa9>:0
  at Android.App.Activity.n_OnBackPressed (System.IntPtr jnienv, System.IntPtr native__this) [0x00009] in <e53ae53987e74674b24a33a90ba894a3>:0
  at (wrapper dynamic-method) System.Object.13(intptr,intptr)

Thread 2:
0   dalvik.system.VMStack.getThreadStackTrace(VMStack.java:-2)
1   java.lang.Thread.getStackTrace(Thread.java:1538)
2   java.lang.Thread.getAllStackTraces(Thread.java:1588)
3   com.microsoft.appcenter.crashes.Crashes.saveUncaughtException(Crashes.java:933)
[...]

PR #5047, opened 2019-01-23
[Android] Allow push/pop during OnStart

hartez:
### Description of Change ###

In order to keep the toolbar and appearing event handling consistent, the AppCompat NavigationPageRenderer waits until OnAttachedToWindow before pushing its existing pages. However, if pages are pushed/popped during OnStart (which happens before OnAttachedToWindow), the page list can become inconsistent. This change allows for the internal page list to remain inconsistent even when pushing/popping during OnStart, and then pushing the correct set of pages once OnAttachedToWindow has occurred.

Submitting this as an alternative to #4730.

No automated tests, as the issue requires running code during OnStart and having a different than normal page structure for Control Gallery.

### Issues Resolved ### 

- fixes #3528
- closes #4730

### API Changes ###

 None

### Platforms Affected ### 

- Android

### Behavioral/Visual Changes ###

None

### Before/After Screenshots ### 

Not applicable

### Testing Procedure ###

Build the Control Gallery project with the directive TEST_4729 and run it. If the app crashes, the test has failed. If it shows a Navigation Page with the title 4729 and a label with the text "Welcome to Xamarin.Forms!", then the test has succeeded.

### PR Checklist ###

- [ ] Has automated tests 
- [x] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard

PR #4730, opened 2018-12-13
[Android] Fix first navigation race condition with main page affectation

kvpt:
### Description of Change ###
Wait for initial page to be loaded before push new pages inside the navigation stack.

### Issues Resolved ### 
- Fix #4729
- Fix #3528
- Fix #3923

### Platforms Affected ### 
- Android

### PR Checklist ###

- [ ] Has automated tests => Only manuel test is possible
- [X] Rebased on top of the target branch at time of PR
- [X] Changes adhere to coding standard

PR #8666, opened 2019-11-26
[AND] Fix NRE in OnBackPressed

### LINKED ISSUE #6700: Enumerable.Last[TSource] (System.Collections.Generic.IEnumerable`1[T] source)
### Description
System.InvalidOperationException: Sequence contains no elements

### Platform
- Android (Versions 8, 8.1, 9)

### Reference Link

#5857, #4715 

### Issue First reported 
12 Dec 2018

Kindly fix this. This crash occurs very frequently. (CRITICAL ISSUE)

jfversluis:
### Description of Change ###

Added some defensive code to prevent an NRE.

Targets 4.3 since it is a regression since 3.4 apparently.

#### Background
This fix has been proposed a couple of times before but was never implemented. For instance because of what @hartez mentions here: https://github.com/xamarin/Xamarin.Forms/issues/4715#issuecomment-449056787

> There's no check there because a nav model with no elements in it shouldn't be handling a back button press in the first place. So somehow we're getting into an invalid navigation state. You mention that you do some navigation when the app goes to sleep - do you handle Resume at all?

While I do agree, on the other hand I also see no harm in adding it this way to prevent the NRE. But I'm basically opening this PR to get some discussion going around this.

Since no one is able to produce a reproduction, it is kind of hard to make progress on this.

### Issues Resolved ### 

- fixes #6700

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

None

Not applicable

### Testing Procedure ###

N/A

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/AppCompat/Platform.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/Platform.cs
@@ -373,8 +373,8 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 			if (NavAnimationInProgress)
 				return true;
 
-			Page root = _navModel.Roots.Last();
-			bool handled = root.SendBackButtonPressed();
+			Page root = _navModel.Roots.LastOrDefault();
+			bool handled = root?.SendBackButtonPressed() ?? false;
 
 			return handled;
 		}
--- a/Xamarin.Forms.Platform.Android/Platform.cs
+++ b/Xamarin.Forms.Platform.Android/Platform.cs
@@ -812,8 +812,8 @@ namespace Xamarin.Forms.Platform.Android
 			if (NavAnimationInProgress)
 				return true;
 
-			Page root = _navModel.Roots.Last();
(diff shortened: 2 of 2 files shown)


hartez:
Looking at this again (and with more comments from other folks with the problem), my best guess is that this is a result of a race between Forms and (possibly unintentional) multiple hardware back button presses. Normally I'd rather fix the root issue than stick a Band-Aid on it, but this might not even be fixable (especially if we still can't reproduce it after all this time). And if it really is just a weird race condition, ignoring the back button press and marking it unhandled _is_ the correct fix.

So let's do it. Best case, it's the right fix. Worst case, fewer crashes and if apps get into a weird navigation state maybe those bug reports will help us figure out what's going on.

kvpt:
@hartez A proven case of corruption of the navigation stack is the push during MainPage change.
I have made #4730 and you made #5047 to fix that.
What is the remaining work on this ?
I maintain a fork on my side just for this fix. 
Can you allocate time to complete the pull request.
Thanks.

PR #8684, opened 2019-11-26
MultiBinding! (Issue 4565)

### LINKED ISSUE #4565: [Enhancement] MultiBinding

legistek:
### Description of Change ###

Implements `MultiBinding`, per issue #4565.

fixes #4565

### API Changes ###
Added:
`public sealed class MultiBinding : BindingBase`
`public interface IMultiValueConverter`

`public static readonly object DoNothing { get; }` in `Binding`
`public static readonly object UnsetValue { get; }` in `BindableProperty`
(the last two are used in `IMultiValueConverter` implementations)

### Platforms Affected ### 
All

No changes to existing applications.

New applications can now use `MultiBinding` to bind a target property to a series of `BindingBase` expressions (including nested `MultiBinding`s) that evaluate to a single value through an `IMultiValueConverter` instance provided by the application. For example:

```
<Button Text="Issue Driver's License">
     <Button.IsEnabled>
          <MultiBinding Converter="{StaticResource AllMustBeTrueMultiConverter}">
               <Binding Path="IsOver16"/>
               <Binding Path="HasPassedExam"/>
               <Binding Path="HasCriminalRecord" Converter="{StaticResource Inverter}"/>
          </MultiBinding>
     </Button.IsEnabled>
</Button>
```

All types of `Binding`s should be valid child bindings for the `MultiBinding`, including implicit source (`BindingContext`), explicit source `Binding`, relative source `Binding`, and even additional `MultiBinding`s.

Except for allowing nested `MultiBinding`s, all efforts have been made to match the WPF spec and behavior for the same feature.

### Testing Procedure ###
Unit tests in Xamarin.Forms.Core

### PR Checklist ###


Xamarin.Forms.Core.UnitTests/MultiBindingTests.cs
```diff
@@ -0,0 +1,872 @@
...
+		{
+			base.TearDown();
+			Device.PlatformServices = null;
+		}
+
+		[Test]
+		public void TestChildOneWayOverridesMultiTwoWay()
+		{
+			// This tests a weird edge case where the MultiBinding is TwoWay but
+			// one of the child bindings is OneWay. This results a situation where
```
samhouts:
```suggestion
			// one of the child bindings is OneWay. This results in a situation where
```

Xamarin.Forms.Core.UnitTests/MultiBindingTests.cs
```diff
@@ -0,0 +1,872 @@
...
+
+			// Martial law is over
+			group.SuspendAll = false;
+			Assert.IsTrue(checkBox.IsChecked);
+
+			// But she got in an accident and now can't drive again
+			group.Person5.IsSuspended = true;
+			Assert.IsFalse(checkBox.IsChecked);
+
+			// The new PM has pardoned everyone after the end of the rebellion
```
samhouts: so nice when a unit test has a happy ending

Xamarin.Forms.Core/MultiBinding.cs
```diff
@@ -0,0 +1,293 @@
...
+			{
+				ThrowIfApplied();
+				_converterParameter = value;
+			}
+		}
+
+		public Collection<BindingBase> Bindings
+		{
+			get => _bindings ?? (_bindings = new Collection<BindingBase>());
+			set => _bindings = value;
```
StephaneDelcroix: it should probably throw if applied, and also throw if the collection is modified once the binding is applied

Xamarin.Forms.Core/MultiBinding.cs
```diff
@@ -0,0 +1,293 @@
...
+		{
+			get { return _converterParameter; }
+			set
+			{
+				ThrowIfApplied();
+				_converterParameter = value;
+			}
+		}
+
+		public Collection<BindingBase> Bindings
```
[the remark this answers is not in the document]
StephaneDelcroix: or an IList if you're not tracking INCC

Xamarin.Forms.Core/MultiBinding.cs
```diff
@@ -0,0 +1,293 @@
...
+				ConverterParameter = ConverterParameter,
+				Bindings = new Collection<BindingBase>(this.Bindings),
+				FallbackValue = FallbackValue,
+				Mode = Mode,
+				TargetNullValue = TargetNullValue,
+				StringFormat = StringFormat,
+			};
+		}
+
+		internal override void Apply(bool fromTarget)
```
StephaneDelcroix: if I understand correctly, the Converter is mandatory, so it should be nice to Log a message if the converter is missing

AmrAlSayed0:
Is MultiBinding going to be in 4.5.0?

legistek:
@AmrAlSayed0 it has not been abandoned but I have been swamped with my actual job. I'm fine with @StephaneDelcroix's changes so if he or anyone wants to just make them and run with this that's fine by me. Otherwise I'll make them when I'm able.

Xamarin.Forms.Core/MultiBinding.cs
legistek: Ok on throw if applied. To detect collection modified wouldn't it have to be an INCC though? Is it worth it?

legistek: > I know there's none in WPF, but what would be a good markup extension for this ? (we can bring that on a different PR)

I was thinking about that myself, but I don't know of any precedent for a collection being represented through a markup extension, do you?

legistek:
@StephaneDelcroix I think I'm finished here, just not sure if you actually want me to bother trying to detect binding collection changes through INCC. 

This should now be properly rebased on 4.6.0 too btw.

neobenedict:
Error		Position 56:34. Type MultiBinding not found in xmlns http://xamarin.com/schemas/2014/forms

Version 4.6.0.726+446-sha.d0dc59903-azdo.3681414

Is this not usable yet?

samhouts:
This will be part of 4.7.0. You may try it in the latest nightly build. Thanks!

ncarandini:
I've made a [Demo](https://github.com/ncarandini/XFMultiBindingDemo) and it works.
For comparison and for those who don't like using converters and multi bindings I've added an example on how to create a property in the ViewModel that does the same thing.

PR #8696, opened 2019-11-27
Prevent UWP NullReferenceException when calling SavePropertiesAsync method off the main thread

### LINKED ISSUE #8682: [Bug] [UWP] NullReferenceException when call SavePropertiesAsync method off the main thread
### Description

NullReferenceException is thrown in method Xamarin.Forms.Application.SavePropertiesAsync when called off the main thread. 

Xamarin.Forms.Application.Dispatcher property is null because method Xamarin.Forms.Platform.UWP.DispatcherProvider.GetDispatcher(object) returns null. If statement on line 22 in this method is evaluated as false.

Possible fix could be to remove return statement on line 32 in Xamarin.Forms.Platform.UWP.DispatcherProvider.GetDispatcher(object) method. Not tested.

[UwpDispatcherNullReference.zip](https://github.com/xamarin/Xamarin.Forms/files/3893354/UwpDispatcherNullReference.zip)

### Steps to Reproduce

See attachment - UwpDispatcherNullReference.MainPage.OnAppearing method.

### Basic Information

- Last known good version: Different exception in 4.3.0.947036. See bug #8167

TomOnd:
### Description of Change ###

When DispatcherProvider returns null, because context parameter does not inherit from VisualElement, then new instance of FallbackDispatcher is used.

### Issues Resolved ### 

- fixes #8682 

### API Changes ###

 None

### Platforms Affected ### 

- UWP

Issue is fixed in Xamarin.Forms.Core project, but other platforms are not affected, because for other platforms IDispatcherProvider is not implemented and thus method DispatcherExtensions.GetDispatcher already returns FallbackDispatcher instance on line 25.

None

Not applicable

### Testing Procedure ###
Automated test added; for UWP, manually run the test for Issue 8682.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue8682.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue8682.cs
@@ -70,7 +70,7 @@ namespace Xamarin.Forms.Controls.Issues
 
 #if UITEST
 		[Test]
-		public void Issue1Test() 
+		public void SavePropertiesAsyncOffMainThreadDoesNotCrash() 
 		{
 			RunningApp.WaitForElement(Run);
 			RunningApp.Tap(Run);
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue8682.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue8682.cs
@@ -1,0 +1,81 @@
+﻿using System;
+using System.Threading.Tasks;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.Forms.Core.UITests;
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if UITEST
+	[Category(UITestCategories.ManualReview)]
+#endif
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 8682, "[Bug] [UWP] NullReferenceException when call SavePropertiesAsync method off the main thread", PlatformAffected.UWP)]
+	public class Issue8682 : TestContentPage
+	{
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Core/DispatcherExtensions.cs
```diff
@@ -26,7 +26,14 @@ public static IDispatcher GetDispatcher(this BindableObject bindableObject)
 			}
 
 			// Use the DispatcherProvider to retrieve an appropriate dispatcher for this BindableObject
-			return s_current.GetDispatcher(bindableObject);
+			IDispatcher dispatcher = s_current.GetDispatcher(bindableObject);
```
StephaneDelcroix:
```suggestion
			return s_current.GetDispatcher(bindableObject) ?? new FallbackDispatcher();
```

TomOnd: What should I do now? Commit or batch? Thank you for your advice.

PureWeen: @TomOnd either way

When we merge things in we squash all the commits down to one so it doesn't really matter

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue8682.cs
```diff
@@ -0,0 +1,81 @@
...
+				catch (Exception e)
+				{
+					return $"Test failed: {e.Message}.";
+				}
+			});
+		}
+
+#if UITEST
+		[Test]
+		public void Issue1Test()
```
samhouts:
```suggestion
		public void SavePropertiesAsyncOffMainThreadDoesNotCrash() 
```

Xamarin.Forms.Core/DispatcherExtensions.cs
TomOnd: Resolved with commit: a1b0093

HemalathaMarikuma-syncfusion:
Is this rolled out ? We have checked this with the following pre release version of XF , it seems not fixed.

4.4.0.991220-pre3

4.4.0.991210-pre2

4.4.0.996621-pre1

4.3.0.991221

May I know in which release version will I get the fix ?

PR #8702, opened 2019-11-28
Add name of key in INameScope.RegisterName ArgumentException

### LINKED ISSUE #7540: [Enhancement] Include the name of the duplicate element for INameScope.RegisterName
## Summary
referenced code: https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/Internals/NameScope.cs line 21
'throw new ArgumentException("An element with the same key already exists in NameScope", nameof(name));'

It would be fantastic if the value of the 'name' variable was included in the exception. Would make it far easier to track down the duplicate name definition in nested xaml files.

jfversluis:
### Description of Change ###

Adds the actual name to the exception so it's easier to solve by the user.

### Issues Resolved ### 

- fixes #7540

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Internals/NameScope.cs
+++ b/Xamarin.Forms.Core/Internals/NameScope.cs
@@ -18,7 +18,7 @@ namespace Xamarin.Forms.Internals
 		void INameScope.RegisterName(string name, object scopedElement)
 		{
 			if (_names.ContainsKey(name))
-				throw new ArgumentException($"An element with the same key ({name}) already exists in NameScope", nameof(name));
+				throw new ArgumentException($"An element with the key '{name}' already exists in NameScope", nameof(name));
 
 			_names[name] = scopedElement;
 		}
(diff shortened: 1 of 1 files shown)


StephaneDelcroix:
this has been discussed before. the decision this far was to mimic the DuplicateKeyException message from, e.g., the Dictionary`2 implementation, that doesn't indicates the key name.

but, I'm open to change the approach, as the type of the key is known to be a string.

Xamarin.Forms.Core/Internals/NameScope.cs
```diff
@@ -18,7 +18,7 @@ object INameScope.FindByName(string name)
 		void INameScope.RegisterName(string name, object scopedElement)
 		{
 			if (_names.ContainsKey(name))
-				throw new ArgumentException("An element with the same key already exists in NameScope", nameof(name));
+				throw new ArgumentException($"An element with the same key ({name}) already exists in NameScope", nameof(name));
```
StephaneDelcroix:
```suggestion
				throw new ArgumentException($"An element with the key '{name}' already exists in NameScope", nameof(name));
```

jfversluis:
Done!

PR #8750, opened 2019-12-05
Setup UWP Platform projects to use multi targeting

PureWeen:
### Description of Change ###
Having the ControlGallery min target set to 14393 makes it really hard to test the majority of scenarios due to not being able to add nugets that have minimums higher than 14393 

I tried to just set it on the Control Gallery without all these other project changes but UWP just kept complaining about two different PDB files and it couldn't figure out which one to use.

This is setup to multitarget UWP but currently only uses the 16299/14393 packages since that's how we are currently setup. At a later date we can pull in the 18463/15063 packages and increment WinUI to 2.3/2.4

### Platforms Affected ### 
- UWP

### Testing Procedure ###
- run control gallery and make sure nugets work

### PR Checklist ###


Xamarin.Forms.Controls/Xamarin.Forms.Controls.csproj
```diff
@@ -1,8 +1,7 @@
 ﻿<Project Sdk="Msbuild.Sdk.Extras/2.0.54">
   <PropertyGroup>
     <TargetFramework Condition=" '$(OS)' != 'Windows_NT' ">netstandard2.0</TargetFramework>
-    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">netstandard2.0;uap10.0.14393</TargetFrameworks>
-    <TargetPlatformVersion Condition="$(TargetFramework) == 'uap10.0.14393'">10.0.16299.0</TargetPlatformVersion>
+    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">netstandard2.0</TargetFrameworks>
```
campersau: Looks like this can be reduced to just
```xml
<TargetFramework>netstandard2.0</TargetFramework>
```

PR #8790, opened 2019-12-07
Shell Modal Behavior

### LINKED ISSUE #7147: [Spec]  Modal Improvements
# Modal Spec

Allow users to further customize modal behavior and simplify down the interaction with modal pages. The majority of the time pages that are treated as Modal are only used modally so it makes more sense to just describe the modal behavior as a permanent attribute of the page itself. 

This is different than PopOvers (https://github.com/xamarin/Xamarin.Forms/issues/1778). You can have only one pop over and pop overs map more to alerts/dialogs associated to visible content. The difference can also be seen at the platform level with how iOS has a specific UIPopOverController vs setting the Modal Behavior on a UIViewController itself. You can push as many Modal Pages as you want onto a stack but there can be only one PopOver.

# API

#### ModalBehavior
Users will set this on a Page to define how this page will be displayed modally. With Shell this will allow users to pass in parameters on the query string to influence the modal behavior on a page. They could even bind to the IsEnabled property on ModalBehavior which will cause it to not be pushed modally

```C#
public class ModalBehavior
{
  bool IsEnabled //v1
  bool Animate //v1
  DismissOptions Dismiss //v2
  DisplayOptions Display //v2

}
```

### Properties

| API | Description |
| IsEnabled | If this is set to false than the page will just push onto the current stack intead of being pushed modally. Changing this value while the page is already displayed will have no real time effect |
| Dismiss | Indicate what ways you can dismiss the modal. Default will use a combination of platform defaults and display to determine default |
| Display | Default: Use platform default. FullScreen: Cover the entire screen. FitToDevice: Size the content based on the size of the device. Phones will be FullScreen whereas larger tablets will show content in a window  |
| Animate | Default: true. Whether to animate or not  |

```C#
public enum DisplayOptions
  Default,
  FitToDevice,
}
```

```C#
[Flags]
public enum DismissOptions
{
  Default = 0,
  Swipe  = 1 << 0
}
```

When Display and Dismiss are set to default how they act will be determined by the platform and device types. 

- iOS13, Phone: FullScreen
- iOS, IPad: FitToDevice (PageSheet)
- Android, Phone: FullScreen
- Android, Tablet: FitToDevice
- UWP: FitToDevice

### Examples

```XAML

<ContentPage x:Class="Xamarin.Forms.Sandbox.MainPage">
  <Page.ModalBehavior>
     <ModalBehavior IsEnabled={Binding IsModal} />
  </Page.ModalBehavior>
</ContentPage>

```

```C#
// This will push the page modally 
Navigation.PushAsync(new MainPage())
```

```C#
// This will push the page non modally
Shell.Current.GotoAsync("MainPage?IsModal=false")


// This will push the page modally
Shell.Current.GotoAsync("MainPage?IsModal=true")
```

### Implementation Details

- The first version can just take ModalBehavior and implement the Animated/IsEnabled properties. This will allow for better Shell navigation integration and simplify the overall navigation API

#### Android

 - Androids Modal implementation should be switched to using a DialogFragment. Currently Modal pages on Android just get added to the root ViewGroup and are animated over the top of the current view but both views are techinically still on the screen. This implementation has caused accessibility issues and generally just seems counter to the platform. When a Modal page is pushed the container can use a ViewPager or just FragmentTransations to swap out content. Both of these can be animated at the Android level.
   - This does need to be spiked a bit more to see if all current Modal behavior can be replicated if we use DialogFragments
- FitToDevice on Android will use a MinHeight/MinWidth that will match how the FormSheet sizing works on iOS

 #### iOS

-  FitToDevice on iOS will map to FormSheet 

 #### UWP
-  FitToDevice will map to using a Flyout

PureWeen:
### Description of Change ###
Adds ModalBehavior class to Shell which allows people to push modal pages via GotoAsync Navigation
### Issues Resolved ### 
- fixes (partially) #7147

### API Changes ###
```C#
[Flags]
public enum PresentationMode
{
	Animated = 1 << 1,
	Modal = 1 << 2,
// These are to help for XAML only scenarios
	ModalAnimated = PresentationMode.Animated | PresentationMode.Modal,
	ModalNotAnimated = PresentationMode.NotAnimated | PresentationMode.Modal
}
```

```C#

public class Shell
{
	PresentationMode PresentationMode {get;set;} //Attached Property
}
```

### Platforms Affected ### 
- Core/XAML (all platforms)

### Testing Procedure ###
- Testing page included and a UI test included

### Usage Examples ###
Specify the Modal property on the content Page

#### Basic Example

```XAML
<ContentPage x:Class="ModalPage" Shell.PresentationMode="ModalAnimated">
</ContentPage>
```

```C#
Routing.RegisterRoutes("ModalPage", typeof(ModalPage))
...

GotoAsync("ModalPage") // Equivalent to Navigation.PushModal
GotoAsync("//ShellRoute/ModalPage") // Navigate to Shell Route and push Modal Page
```

#### Navigation Page Example
```XAML
<NavigationPage x:Class="ModalPage" Shell.PresentationMode= "ModalAnimated">
</NavigationPage>
```

```C#
Routing.RegisterRoutes("ModalPage", typeof(ModalPage))
Routing.RegisterRoutes("LoginPage", typeof(LoginPage))
```

*This will push LoginPage onto the NavigationPage that is Modal Page*
```C#
GotoAsync("ModalPage/LoginPage") 
```

*This will push two modal pages and then LoginPage will get pushed onto the last Modal Page*
```C#
GotoAsync("ModalPage/ModalPage/LoginPage") 
```

*If you push another Modal Page than that becomes the new root. The following sample with have a ModalStack with 2 NavigationPages each with a pushed page*
```C#
GotoAsync("ModalPage/SecondPage/ModalPage/SecondPage");
```

### PR Checklist ###


rookiejava:
Here are two alternatives.

1. What about providing these bindable properties to `Shell` directly just like `NavBarIsVisibleProperty` and `NavBarHasShadowProperty`.
```cs
public static readonly BindableProperty IsModalProperty =
	BindableProperty.CreateAttached("IsModal", typeof(bool), typeof(Shell), true);

public static readonly BindableProperty IsModalAnimatedProperty =
	BindableProperty.CreateAttached("IsModalAnimated", typeof(bool), typeof(Shell), false);
``` 


2. Or if you want to use the `Behavioir` suffix because you want to be consistent with `FlyoutBehavior` and `BackButtonBehavior`, you might consider providing only one attribute of type enum like `FlyoutBehavior`.
```cs
public enum ModalBehavior
{
	Modal,
	AnimatedModal,
	Modalless
}

public static readonly BindableProperty ModalBehaviorProperty =
	BindableProperty.CreateAttached("ModalBehavior", typeof(ModalBehavior), typeof(Shell), FlyoutBehavior.Modal, propertyChanged: OnModalBehaviorChanged);
```

Xamarin.Forms.Core/Shell/ModalBehavior.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+	public class ModalBehavior : BindableObject
+	{
+		public static readonly BindableProperty ModalProperty =
+			BindableProperty.Create(nameof(Modal), typeof(bool), typeof(ModalBehavior), true);
+
+		public static readonly BindableProperty AnimatedProperty =
+			BindableProperty.Create(nameof(Animated), typeof(bool), typeof(ModalBehavior), false);
+
+		public bool Modal
```
rookiejava:
```suggestion
		public bool IsModal
```

Xamarin.Forms.Core/Shell/ModalBehavior.cs
```diff
@@ -0,0 +1,27 @@
...
+		{
+			get => (bool)GetValue(ModalProperty);
+			set => SetValue(ModalProperty, value);
+		}
+
+		public bool Animated
```
rookiejava:
```suggestion
		public bool IsAnimated
```

PureWeen:
Went with your suggestion @rookiejava

Xamarin.Forms.Core/Application.cs
```diff
@@ -76,6 +76,13 @@ public Page MainPage
 				if (_mainPage != null)
 				{
 					InternalChildren.Remove(_mainPage);
+					_mainPage.NavigationProxy.Inner = null;
+
+					// Set platform inner proxy to null
+					// otherwise the incoming page will make decisions based on the 
+					// previous application state
+					NavigationProxy.Inner = null;
```
PureWeen: This is the biggest change here I'm wondering about the approach on

The problem I noticed is that this Inner is only set once the page is realized on the platform. 

This causes an issue for the incoming page because when the incoming page checks *Navigation.ModalStack* that call gets proxied to the previous platform *INavigation* instance. So it makes decisions based on modals/pages that aren't going to exist anymore

PureWeen:
UI Tests green
https://dev.azure.com/xamarin/public/_releaseProgress?releaseId=984&_a=release-pipeline-progress

samhouts:
I'm not fully sure what PresentationMode.None means.

PureWeen:
@samhouts removed None

It was left over and I had meant to remove it

samhouts:
Ok, so now the default is "Animated" and not modal. What does that mean?

PureWeen:
@samhouts it means

GotoAsync("route", <thisParameter>)

Should it NonModalAnimated?

The idea is that the presentation of the page when navigated to is animated. It's all a bit flawed since the attribution should be on the navigating but we don't really have that

samhouts:
Nope, that sounds fine! Just having a public conversation here. ;)

IndexOutOfLimit:
Can we open Modal pages from Shell TabBar Menu?

PR #8820, opened 2019-12-10
[Tizen] Implement DisplayPromptAsync for Tizen

### LINKED ISSUE #6713: [Enhancement] Display prompts

myroot:
### Description of Change ###

 * Implement DisplayPromptAsync for Tizen
    * related PR https://github.com/xamarin/Xamarin.Forms/pull/6714
### Issues Resolved ### 

- fixes #6713

### API Changes ###

 
 None

### Platforms Affected ### 

- Tizen

None

65365380-0b80-11ea-9c85-89615f6f4758
e7bf1300-0b80-11ea-80ef-d476f7a1035a
1229c400-0bb3-11ea-86c3-a2d187a67479

### PR Checklist ###


Xamarin.Forms.Platform.Tizen/Platform.cs
```diff
@@ -562,6 +564,97 @@ void ActionSheetSignalNameHandler(Page sender, ActionSheetArguments arguments)
...
+
+			entry.Completed += (s, e) =>
+			{
+				args.SetResult(entry.Text);
+				prompt.Dismiss();
+			};
+
+			prompt.BackButtonPressed += (s, evt) =>
+			{
+				prompt.Dismiss();
```
campersau: Should the result also be set here because otherwise the tcs never completes in this case?

```suggestion
				args.SetResult(null);
				prompt.Dismiss();
```

myroot: @campersau Yes, you are right!

Xamarin.Forms.Platform.Tizen/Platform.cs
```diff
@@ -120,6 +120,7 @@ internal DefaultPlatform(EvasObject parent)
 			MessagingCenter.Subscribe<Page, bool>(this, Page.BusySetSignalName, BusySetSignalNameHandler);
 			MessagingCenter.Subscribe<Page, AlertArguments>(this, Page.AlertSignalName, AlertSignalNameHandler);
 			MessagingCenter.Subscribe<Page, ActionSheetArguments>(this, Page.ActionSheetSignalName, ActionSheetSignalNameHandler);
+			MessagingCenter.Subscribe<Page, PromptArguments>(this, Page.PromptSignalName, OnPromptRequested);
```
rookiejava: This change will conflict with #8845. I don't know which one will merge first, but I will rebase whatever it is.

PR #8865, opened 2019-12-12
[Android] Fixed Gif animation loading from files 

### LINKED ISSUE #8821: [Bug] Animations of downloaded gifs are not playing on Android
### Description
when gif is downloaded from internet and saved to filesystem and the property IsAnimationPlaying is set to true the application dumps with 
Error loading image: Android.Content.Res.Resources+NotFoundException: Resource ID #0x0
### Steps to Reproduce

1. clone git https://github.com/hulluP/AnimatedGifForms
2. run on Android device or simulator
3. app will dump once button is pressed and animation does not play. This works on iOS

### Expected Behavior
Animation plays like it does for the same gif when it is loaded from resources like the git https://github.com/jfversluis/AnimatedGifForms

### Actual Behavior
does not play animation
### Basic Information

- Version with issue: 4.4.09 pre3
- Last known good version: never worked
- IDE: Mac OS visual studio
- Platform Target Frameworks: 
  - iOS:  13.2
  - Android: 7.1

- Nuget Packages: xamarin.forms
- Affected Devices:

### Screenshots

Screenshot 2019-12-10 at 09 19 24

jsuarezruiz:
### Description of Change ###

Fixed Gif animation loading from local files (no resources).

### Issues Resolved ### 

- fixes #8821

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

None

#### Before 
issue8821-before

#### After
issue8821-after

### Testing Procedure ###
Launch Core Gallery and navigate to the issue 8821. Press the "DownloadFile" button and then press the "Animate" button. If the gif is loaded and animated, the test has passed.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/AndroidGIFImageParser.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/AndroidGIFImageParser.cs
@@ -38,6 +38,8 @@ namespace Xamarin.Forms.Platform.Android
 
 	public class FormsAnimationDrawable : AnimationDrawable, IFormsAnimationDrawable
 	{
+		const int DefaultBufferSize = 4096;
+
 		int _repeatCounter = 0;
 		int _frameCount = 0;
 		bool _finished = false;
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.Android/Renderers/AndroidGIFImageParser.cs
```diff
@@ -264,6 +257,56 @@ public static async Task<IFormsAnimationDrawable> LoadImageAnimationAsync(UriIma
...
+			}
+
+			return animation;
+		}
+
+		internal static async Task<FormsAnimationDrawable> GetFormsAnimationDrawableFromFile(string file, Context context, BitmapFactory.Options options)
+		{
+			FormsAnimationDrawable animation = null;
+
+			using (var stream = new FileStream(file, FileMode.Open, FileAccess.Read, FileShare.Read, DefaultBufferSize, true))
```
campersau: The code here looks identical to the code in the method above except that the stream is created differently.
Maybe add an other method which takes a stream instead of a string and call it from these two methods (`GetFormsAnimationDrawableFromResource`, `GetFormsAnimationDrawableFromFile`) to reduce code duplication.

Also it looks like the `GetFormsAnimationDrawableFromResource` method can directly take the `drawableIdentifier` so `ResourceManager.GetDrawableByName` does need to be called twice in this case.

jsuarezruiz: Right. I have created a new method that receives a Stream and is used in the other cases, when using a Resource or a file path.

Xamarin.Forms.Platform.Android/Renderers/AndroidGIFImageParser.cs
```diff
@@ -264,6 +257,50 @@ public static async Task<IFormsAnimationDrawable> LoadImageAnimationAsync(UriIma
...
+		{
+			FormsAnimationDrawable animation = null;
+
+			using (var stream = context.Resources.OpenRawResource(resourceId))
+				animation = await GetFormsAnimationDrawableFromStream(stream, context, options);
+
+			return animation;
+		}
+
+		internal static async Task<FormsAnimationDrawable> GetFormsAnimationDrawableFromFile(string file, Context context, BitmapFactory.Options options)
```
PureWeen: - Does this work on UWP?
- Does this work with a PNG? If I download a PNG/JPG to the file system to display it on Android does that currently work?

jsuarezruiz: I have applied changes and the sample works on Android, iOS and UWP.
issue8821-uwp

PR #8875, opened 2019-12-13
[Bug] [iOS] [Shell] "Nav Stack consistency error" when Swiping to Dismiss

### LINKED ISSUE #8461: [Bug] [iOS] [Shell] "Nav Stack consistency error"
### Description

I am getting an exception report in AppCenter, _"Nav Stack consistency error"_. Occurring in all versions of XForms (including nightly, was hoping #8135 might have helped but no effect).

Multiple versions of iOS (13, 12 & 11), affecting about ~5% of users. I have not been able to replicate and have no way to reach out to users to get retry steps.

```csharp
IShellSectionController.SendPopping (Xamarin.Forms.Page page)
ShellSectionRenderer.SendPoppedOnCompletion (System.Threading.Tasks.Task popTask)
AsyncMethodBuilderCore+<>c.<ThrowAsync>b__7_0 (System.Object state)
NSAsyncSynchronizationContextDispatcher.Apply ()
(wrapper managed-to-native) UIKit.UIApplication.UIApplicationMain(int,string[],intptr,intptr)
UIApplication.Main (System.String[] args, System.IntPtr principal, System.IntPtr delegate)
UIApplication.Main (System.String[] args, System.String principalClassName, System.String delegateClassName)
Application.Main (System.String[] args)
```

### Steps to Reproduce

I have a TabBar setup in my AppShell. Three tabs with only a single content page as the child. In the Shell's constructor I also register some modal routes, which I navigate to using `Shell.GoToAsync()`. I do not trigger navigation in any other way, nor do I pop anything off the stack manually.

### Expected Behavior

Obviously I could be at fault here, but it would be helpful if the exception contained some more information to aid in debugging. Possibly the name of the page it's trying to pop (when available) and the current state of the navigation stack.

### Actual Behavior

Generic exception, which is difficult to debug/troubleshoot.

### Basic Information

- Version with issue: All (including nightly)

### LINKED ISSUE #8944: [Bug] Shell iOS back swipe crash Xamarin.Forms 4.4.0.991265
### Description
App crash when using swipe gestures to go back (iOS, Shell, TabBar).

### Steps to Reproduce

1. Use Shell TabBar
2. Push any page to a stack
3. Start swipe gesture to navigate back, but do not finish navigation
4. Start swipe gesture again
5. The app will crash

### Expected Behavior
App will not crash.

### Actual Behavior
App crashes.

### Basic Information

- Version with issue: Xamarin.Forms 4.4.0.991265
- Last known good version: Xamarin.Forms 4.3.0.908675
- Affected Devices: iPhone 7, iOS 12.4.1

### StackTrace

```
System.NullReferenceException: Object reference not set to an instance of an object
  at Xamarin.Forms.Shell.GetNavBarIsVisible (Xamarin.Forms.BindableObject obj) [0x00000] in D:\a\1\s\Xamarin.Forms.Core\Shell\Shell.cs:75 
  at Xamarin.Forms.Platform.iOS.ShellSectionRenderer+NavDelegate.WillShowViewController (UIKit.UINavigationController navigationController, UIKit.UIViewController viewController, System.Boolean animated) [0x00028] in D:\a\1\s\Xamarin.Forms.Platform.iOS\Renderers\ShellSectionRenderer.cs:498 
  at (wrapper managed-to-native) UIKit.UIApplication.UIApplicationMain(int,string[],intptr,intptr)
  at UIKit.UIApplication.Main (System.String[] args, System.IntPtr principal, System.IntPtr delegate) [0x00005] in /Library/Frameworks/Xamarin.iOS.framework/Versions/13.6.0.12/src/Xamarin.iOS/UIKit/UIApplication.cs:86 
  at UIKit.UIApplication.Main (System.String[] args, System.String principalClassName, System.String delegateClassName) [0x0000e] in /Library/Frameworks/Xamarin.iOS.framework/Versions/13.6.0.12/src/Xamarin.iOS/UIKit/UIApplication.cs:65 
```

dmariogatto:
### Description of Change ###

Swiping to dismiss when using the Shell results in an exception "Nav Stack consistency error". This is due to `UINavigationViewController.PopViewController` getting called on the start of the back swipe gesture, even if it is cancelled. This removes the page from the Shell's navigation stack, but it still exists for the `UINavigationController`.

This change removes the override for `UINavigationViewController.PopViewController` on the `ShellSectionRenderer` and instead relies on the `UIViewControllerTransitionCoordinator` to determine whether the swipe gesture was completed or cancelled, popping the stack if required.

### Issues Resolved ### 

- fixes #8461
- fixes #8944

### API Changes ###

None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###

Use the "SwapRoot - Store Shell" in the Control Gallery sample app to test pushing and popping pages, in particular test the swipe to dismiss gesture. The navigation stacks for both the `UINavigationController`and `ShellSection` should remain consistent.

### PR Checklist ###


Xamarin.Forms.Platform.iOS/Renderers/ShellSectionRenderer.cs
```diff
@@ -174,17 +167,12 @@ protected virtual void HandlePropertyChanged(object sender, PropertyChangedEvent
 				UpdateTabBarItem();
 		}
 
-		protected virtual IShellSectionRootRenderer CreateShellSectionRootRenderer(ShellSection shellSection, IShellContext shellContext)
```
PureWeen: Can you bring this back?

PureWeen:
Yay!

Xamarin.Forms.Platform.iOS/Renderers/ShellSectionRenderer.cs
dmariogatto: Done! The only parts that are important are the removal of the `PopViewController` and the changes to the `NavDelegate`. My mistake, was copying things around... should've paid more attention to the diff!

PureWeen:
Is it possible to add an automated UI Test for this? or is it tricky to trigger?

dmariogatto:
Maybe... I'm not familiar with automated UI tests. I'd imagine the trickiest part would be to get the swipe gesture just right.

PureWeen:
@dmariogatto do you want to give it a go creating that UI Test? Or want me to take it over from here?

dmariogatto:
I wouldn't mind having a go, however, I can't get any of UI tests to run (either using "Run Test" or "Debug Test"). I've created a new issue with the error I'm having #8891, hopefully it's just something simple I'm missing, but I don't want to hold anything up.

PureWeen:
@dmariogatto 

This is probably a different issue but do you think this PR will fix this issue as well?
https://github.com/xamarin/Xamarin.Forms/pull/8875

dmariogatto:
It's a different stack trace, however, the retry steps are exactly the same.

dmariogatto:
@PureWeen - thanks for the help with getting the UI Tests to work!

I've added an automated UI Test for this issue. Tested on both the new & old code and it's producing the correct results.

dmariogatto:
@PureWeen I've fixed the test to run on iOS 11/12/13. The swipe to dismiss gesture is handled slightly differently in iOS 13. This probably means there is a risk of the test breaking in future releases.

PureWeen:
@dmariogatto 

> This probably means there is a risk of the test breaking in future releases.

Such is the iOS life :-/ Which makes having the test super important. For example, the UISplitViewController (that ipads use) completely changed with iOS 13 and required a fairly heavy rewrite.

Thank you for being so prompt with this PR!!

I've kicked off the UI tests again 🤞

jcarlosgbarett:
I've got the same exception in my app when user starts a "back" swipe from a page, but do not complete the movement.

I don't know how to solve it.

PureWeen:
@dmariogatto 

I started working on this issue
https://github.com/xamarin/Xamarin.Forms/issues/9006

And noticed that this PR isn't quite correct for iOS13. On iOS13 after you swipe to dismiss a page you can swipe to bring the page back so *NotifyWhenInteractionEndsUsingBlock* doesn't fire if it successfully popped the page.

It kind of works because then when you push the page again that causes *NotifyWhenInteractionEndsUsingBlock* to then fire which then gets the stack into a consistent state

NavigationPageRenderer has the same issue.

I don't have a workaround yet :-) still poking around my options

dmariogatto:
@PureWeen 

How are you swiping to bring back the page? Is this two separate gestures, or one continuous?

PureWeen:
@dmariogatto wait... wait....wait... 

It looks like this is actually just a quirk of the remote iOS simulator :-/ 
For some reason it's not registering when you lift your finger to complete the pop 🤷‍♂ 

But it leaves it all in a really weird state.. 

79DCBAA5-D36C-4193-A8EF-33A30B93599D

I tested on a physical device and directly on a mac and can't recreate

One other thing I noticed is that the API we used here is deprecated
https://developer.apple.com/documentation/uikit/uiviewcontrollertransitioncoordinator/1619292-notifywheninteractionendsusingbl?language=objc

dmariogatto:
@PureWeen 

That should hopefully just be a change to use [notifyWhenInteractionChanges](https://developer.apple.com/documentation/uikit/uiviewcontrollertransitioncoordinator/1829391-notifywheninteractionchanges) instead on iOS 10+. I just quickly tried it and it passed the UI test.

PureWeen:
Yea I have that change in the PR I'm working on for https://github.com/xamarin/Xamarin.Forms/issues/9006

seems to work the same

PR #8878, opened 2019-12-13
Implemented IsVisibleProperty on iOS/Android

### LINKED ISSUE #3838: [Enhancement] Add IsVisible to ToolbarItem

muak:
### Description of Change ###

Added IsVisible Property on ToolbarItem.
Implemented iOS and Android Platform Renderer.

### Issues Resolved ### 

- fixes #3838 

### API Changes ###

 
Added:
 - bool ToolbarItem.IsVisible { get; set; } // BindableProperty

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android

None

Not applicable

### Testing Procedure ###

1. Go ToolbarItems Gallery.
1. Tap tb6.
1. Toggle tb1 visibility.

### PR Checklist ###


Xamarin.Forms.Platform.iOS/Renderers/NavigationRenderer.cs
```diff
@@ -1229,6 +1238,15 @@ void UpdateToolbarItems()
 					n.UpdateToolBarVisible();
 			}
 
+			void OnToolberItemPropertyChanged(object sender, PropertyChangedEventArgs e)
```
jfversluis: Typo in the method name

muak: @jfversluis 
Thank you for your comment.
I fixed it.

muak: @jfversluis 
Can I click the Resolve conversation button?

jfversluis: Nope, because I did 😉  Thanks for resolving this so quickly!

jfversluis:
Sorry to trouble you, but would you be able to rebase this? We had an issue with the unit tests that is fixed now. If you rebase it that will go away and we can see if the build and (UI) tests still work. Do you know how to do it?

Thanks!

Xamarin.Forms.Platform.iOS/Renderers/NavigationRenderer.cs
muak: I see. Thanks!

muak:
OK, I rebased and force pushed it.
Thanks.

jfversluis:
Unfortunately, there is a build error

> ##[error]Xamarin.Forms.Controls\ControlGalleryPages\ToolbarItems.cs(61,8): Error CS0128: A local variable or function named 'tb6' is already defined in this scope

muak:
Sorry, there was a duplicated var because of rebasing.
I fixed it.

Xamarin.Forms.Platform.iOS/Renderers/NavigationRenderer.cs
```diff
@@ -1213,6 +1216,12 @@ void UpdateToolbarItems()
 				List<UIBarButtonItem> secondaries = null;
 				foreach (var item in _tracker.ToolbarItems)
 				{
+					item.PropertyChanged -= OnToolbarItemPropertyChanged;
+					item.PropertyChanged += OnToolbarItemPropertyChanged;
```
jfversluis: Why do we need to unsubscribe and subscribe to the same thing right after each other?

Also, shouldn't the `item.IsVisible` check be between here? No need to hook the event to that when it's not visible, right?

muak: @jfversluis 
Thank you for reviewing.

That is because it avoids duplicate subscriptions.

Because every time a toolbaritem's IsVisible property is changed, invoked  UpdateToolbarItems method, it should remove an old subscription and add a new subscription.

Even if the item is not visible, it needs to event hook for changing from invisible to visible.
If the item.IsVisible check is between here, it can't restore visible state.

Thanks.

muak: @jfversluis 
Did you solve your question?
I would be grateful if you could resolve conversation for my answer.

akamud:
I don't get why #3946 was closed and this is still being considered, since it is pretty much the same implementation. Is this PR really being considered as a solution to the "IsVisible" property for ToolbarItems (#3838)?

hartez:
> I don't get why #3946 was closed and this is still being considered, since it is pretty much the same implementation. Is this PR really being considered as a solution to the "IsVisible" property for ToolbarItems (#3838)?

The implementation in 3496 was adding/removing toolbar items, which has other undesirable side-effects. This implementation is setting the visibility (where possible) and keeps the items in the same order as the appear/disappear.

akamud:
> > I don't get why #3946 was closed and this is still being considered, since it is pretty much the same implementation. Is this PR really being considered as a solution to the "IsVisible" property for ToolbarItems (#3838)?
> 
> The implementation in 3496 was adding/removing toolbar items, which has other undesirable side-effects. This implementation is setting the visibility (where possible) and keeps the items in the same order as the appear/disappear.

Thanks for the explanation!

maksym-husiev:
Any updates on this improvement? Really need to have this functionality, but still forced to use workarounds :(

muak:
I made the custom renderer version.
https://github.com/muak/HidableToolbarItem

PR #8889, opened 2019-12-14
Resolve issues with Shell ToolbarItems not reacting to CanExecute Changes

PureWeen:
### Description of Change ###
Fixed Android ToolbarItems so that they update functionally and visually as CanExecuteChages toggles. This addressed both the Text and Icon not changing alpha values when disabled

The code on the NavigationPage was identical to the shell code so I extracted all of that to a reusable Extension method for both of them

### Issues Resolved ### 
- fixes #8741 
- fixes #8327

### Platforms Affected ### 
- Android
- iOS

### Testing Procedure ###
- UI test included
- run through a few tool bar item tests with navigation page to make sure it works as well

### PR Checklist ###

- [x] Targets the correct branch
- [x] Tests are passing (or failures are unrelated)

Issue #8741, opened 2019-12-04
[Bug] [Shell] [Android] ToolbarItem Enabled/Disabled behavior does not work for Shell apps


brunck:
### Description
Changing a ToolbarItem's enabled/disabled state at runtime in a Shell app works on iOS, but not on Android.
This _does_ work properly in non-Shell Android applications using the same mechanism.

### Steps to Reproduce

1. Wire up a ToolbarItem on a page, binding to its Command property. Set the text (or an icon) for it.
2. Make a button that toggles the state of the ToolbarItem between enabled and disabled, via the Command's CanExecute functionality.
3. Add a label with its Text property bound to the ToolbarItem's IsEnabled property.
4. Toggling the state back and forth changes the state of the IsEnabled property, but the UI state of the ToolbarItem itself does not change.

Using the sample repro app, just run it. The main page has the button, label and ToolbarItem. Just click on the button to change the enabled/disabled state for the ToolbarItem.

### Expected Behavior
The ToolbarItem grays out when disabled and does not respond to clicks, and is not grayed out and is clickable when enabled. This works on iOS but not on Android.

### Actual Behavior
The UI state of the ToolbarItem does not change.

### Basic Information

- Version with issue: 4.4.0.991220-pre3 and earlier
- Last known good version: unknown. This works when not using Shell
- IDE: VS 2019 16.3.8
- Platform Target Frameworks: 
  - iOS:  13.2
  - Android: 9.0
- Android Support Library Version: 28.0.0.3
- Affected Devices: all

### Reproduction Link
[TestShellToolbarEnableDisable.zip](https://github.com/xamarin/Xamarin.Forms/files/3923405/TestShellToolbarEnableDisable.zip)

brunck:
Here is an issue file which could be useful as a starting point to repro this.

[Issue8741.zip](https://github.com/xamarin/Xamarin.Forms/files/3963226/Issue8741.zip)

jamesmcmenamin:
@PureWeen Did we find out if this made the cut for next Release? If so we know the number on it yet?

PureWeen:
@jamesmcmenamin Unfortunately it probably won't make it into the next 4.4 SR
But it should be in into the next next one

jamesmcmenamin:
@PureWeen All good brotha man such a huge thank you for this one.

PR #8892, opened 2019-12-14
Fix for 8741: Assert that ToolbarItem is grayed-out (PR 8889)

brunck:
@PureWeen 
There have been some recent changes around the tint color of toolbar/menu items on Android. Such changes being the second aspect of this fix, I'd like us to assert in the issue test that the graying-out works, in order to keep the code around this stable.


brunck:
For PR #8889

eugvor:
Got a fix for #8741 problem with new version of Xamarin.Forms 4.5.0.356. But now I see another problem that with 3 ToolbarItems which I have in the Shell mode for Android, the appearance of one or another one of them gets greyd-out, looks like a bug. Press on one ToobarItem, but also another ToolbarItem changes the color to grey, but still properly clickable(CanExecute).

PureWeen:
@eugvor can you create an issue with a repro please?

eugvor:
Ok will do, thx воскресенье, 01 марта 2020г., 06:51 +01:00 от Shane Neuville  notifications@github.com :

PR #8910, opened 2019-12-16
RadioButton Implementation

### LINKED ISSUE #2404: [F100] RadioButton

jfversluis:
### Description of Change ###

Takeover from the RadioButton PR in #5349. I've copied the relevant parts here.

### Issues Resolved ### 

- fixes #2404
- closes #5349

### API Changes ###
Added:

```
public class RadioButton
{
      public event EventHandler Checked;
      public event EventHandler Unchecked;

      public bool IsChecked {get; set; }  //BP
      public string GroupName { get; set; } //BP
      public ImageSource ButtonSource { get; set; } //BP
}
```

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP
- MacOS

#### GroupName property

The "RadioButton group Gallery - Legacy" has **three** tabs which show group behavior. 
Radio-buttons with group name null are grouped by their parent.
Radio-buttons with same group name, no matter the hierarchy, are grouped at page level.

Video: https://www.youtube.com/watch?v=fwYSmvjCfbw

image

### Testing Procedure ###

See the newly added gallery pages

### PR Checklist ###

### TODO ###
- [x] Make Mac OS work with the groups, you can now check all radiobuttons, group seems to be ignored completely

image

- [x] on iOS you can uncheck a radiobutton by tapping it a second time, this should not happen


jfversluis:
Need to have a look at the review comments in the original PR

Xamarin.Forms.Platform.iOS/Xamarin.Forms.Platform.iOS.csproj
```diff
@@ -177,6 +177,8 @@
     <Compile Include="Renderers\IShellSectionRootHeader.cs" />
     <Compile Include="Renderers\PageContainer.cs" />
     <Compile Include="Renderers\CheckBoxRendererBase.cs" />
+    <Compile Include="Renderers\RadioButonCALayer.cs" />
```
samhouts: RadioButon >> RadioButton

Xamarin.Forms.Platform.iOS/Renderers/RadioButonCALayer.cs
```diff
@@ -0,0 +1,127 @@
+﻿using CoreAnimation;
+using CoreGraphics;
+using UIKit;
+
+namespace Xamarin.Forms.Platform.iOS
+{
+	public class RadioButonCALayer : CALayer
```
samhouts:
```suggestion
	public class RadioButtonCALayer : CALayer
```

Xamarin.Forms.Platform.iOS/Renderers/RadioButonCALayer.cs
```diff
@@ -0,0 +1,127 @@
...
+		const UIColor _checkBorderFillColor = null;
+		const UIColor _checkMarkStrokeColor = null;
+		const UIColor _checkMarkFillColor = null;
+
+		readonly UIButton _nativeControl;
+		readonly RadioButton _radioButton;
+		readonly CAShapeLayer _checkLayer = new CAShapeLayer();
+		readonly CAShapeLayer _containerLayer = new CAShapeLayer();
+
+		public RadioButonCALayer(RadioButton radioButton, UIButton nativeControl)
```
samhouts:
```suggestion
		public RadioButtonCALayer(RadioButton radioButton, UIButton nativeControl)
```

Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -0,0 +1,185 @@
...
+
+		public static readonly BindableProperty GroupNameProperty = BindableProperty.Create(
+			nameof(GroupName), typeof(string), typeof(RadioButton), null, propertyChanged: (b, o, n) => ((RadioButton)b).OnGroupNamePropertyChanged((string)o, (string)n));
+
+		public static readonly BindableProperty ButtonSourceProperty = BindableProperty.Create(
+			nameof(ButtonSource), typeof(ImageSource), typeof(RadioButton), null);
+
+		public event EventHandler<CheckedChangedEventArgs> CheckedChanged;
+
+		public bool IsChecked
```
samhouts: Do we want to allow tri-state?

Xamarin.Forms.Platform.Android/IButtonLayoutRenderer.cs
```diff
@@ -1,11 +1,12 @@
 ﻿using System;
 using Android.Support.V7.Widget;
+using AButton = Android.Widget.Button;
 
 namespace Xamarin.Forms.Platform.Android
 {
 	public interface IButtonLayoutRenderer
 	{
-		AppCompatButton View { get; }		
+		AButton View { get; }
```
samhouts: This is a breaking change. Why are we changing the type?

Xamarin.Forms.Platform.iOS/Renderers/RadioButonCALayer.cs
samhouts:
```suggestion
	public class RadioButtonCALayer : CALayer
```

Xamarin.Forms.Platform.iOS/Renderers/RadioButonCALayer.cs
samhouts:
```suggestion
		public RadioButtonCALayer(RadioButton radioButton, UIButton nativeControl)
```

Xamarin.Forms.Platform.iOS/Renderers/RadioButtonRenderer.cs
```diff
@@ -0,0 +1,296 @@
...
+			var elemValue = (string)Element?.GetValue(AutomationProperties.NameProperty);
+			if (string.IsNullOrWhiteSpace(elemValue) && Control?.AccessibilityLabel == Control?.Title(UIControlState.Normal))
+				return;
+
+			base.SetAccessibilityLabel();
+		}
+
+		protected virtual CALayer CreateRadioBoxLayer()
+		{
+			return new RadioButonCALayer(Element, Control);
```
samhouts:
```suggestion
			return new RadioButtonCALayer(Element, Control);
```

Xamarin.Forms.Platform.iOS/Xamarin.Forms.Platform.iOS.csproj
samhouts:
```suggestion
    <Compile Include="Renderers\RadioButtonCALayer.cs" />
```

Xamarin.Forms.Platform.Android/FastRenderers/ButtonRenderer.cs
```diff
@@ -10,8 +10,9 @@
 using Xamarin.Forms.PlatformConfiguration.AndroidSpecific;
 using AColor = Android.Graphics.Color;
 using AView = Android.Views.View;
+using AButton = Android.Widget.Button;
 
-namespace Xamarin.Forms.Platform.Android.FastRenderers
+	namespace Xamarin.Forms.Platform.Android.FastRenderers
```
samhouts:
```suggestion
namespace Xamarin.Forms.Platform.Android.FastRenderers
```

Xamarin.Forms.Platform.UAP/FormsRadioButton.cs
```diff
@@ -0,0 +1,78 @@
...
+				BackgroundColor = Background;
+
+			if (_contentPresenter != null)
+				_contentPresenter.Background = BackgroundColor;
+			Background = Color.Transparent.ToBrush();
+		}
+
+		void UpdateBorderRadius()
+		{
+
```
samhouts:
```suggestion
```

Xamarin.Forms.Platform.Android/IButtonLayoutRenderer.cs
jfversluis: Yeah @PureWeen made a comment about that too in the original PR with a proposed solution I think. So, I'll look at that

Xamarin.Forms.Core/RadioButton.cs
jfversluis: Personally not a fan, I'll throw it out there in todays sync

Xamarin.Forms.Platform.Android/IButtonLayoutRenderer.cs
jfversluis: Implemented proposed solution

Xamarin.Forms.Core/RadioButton.cs
jfversluis: As discussed; tri-state won't be a thing (for now)

Xamarin.Forms.Platform.iOS/Renderers/RadioButtonRenderer.cs
```diff
@@ -0,0 +1,298 @@
...
+				{
+					SetNativeControl(CreateNativeControl());
+					SetRadioBoxLayer(CreateRadioBoxLayer());
+					Control.HorizontalAlignment = UIControlContentHorizontalAlignment.Left;
+
+					Debug.Assert(Control != null, "Control != null");
+
+					SetControlPropertiesFromProxy();
+
+					_useLegacyColorManagement = e.NewElement.UseLegacyColorManagement();
```
rmarinho: If we have a new Control why we are checking for LegacyColorManagement?

Xamarin.Forms.Controls/CoreGalleryPages/RadioButtonCoreGalleryPage.cs
```diff
@@ -0,0 +1,140 @@
...
+				new RadioButton
+				{
+					Text = "BorderColor",
+					BackgroundColor = Color.Transparent,
+					BorderColor = Color.Red,
+					BorderWidth = 1,
+				}
+			);
+
+			var borderRadiusContainer = new ViewContainer<RadioButton>(Test.Button.BorderRadius,
```
rmarinho: Thos should be CornerRadius no ? but it also doesn't work for me if i change to something like:

```
	var borderRadiusContainer = new ViewContainer<RadioButton>(Test.Button.BorderRadius,
				new RadioButton
				{
					Text = "BorderRadius",
					BackgroundColor = Color.Transparent,
					BorderColor = Color.Red,
					BorderWidth = 1,
					CornerRadius = 5
				}
			);
```

Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -0,0 +1,185 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using Xamarin.Forms.Platform;
+
+namespace Xamarin.Forms
+{
+	[RenderWith(typeof(_RadioButtonRenderer))]
+	public class RadioButton : Button, IElementConfiguration<RadioButton>
```
rookiejava, line 9: why don't you implements `IFontElement`, `ITextElement`, `IBorderElement` instead of inheriting `Button` directly.

PR #8921, opened 2019-12-17
Improve animation behavior with regard to power saving mode

### LINKED ISSUE #7500: [Bug] [Android] Async animations can cause hang if animations are disabled
### Description

One simple way to do a looped animation is to do an async void method containing an infinite loop. Normally, this works fine because the async system allows other code to execute in the meantime.

However, if animations become disabled for some reason, such as due to power saving mode or the accessibility setting, the async calls return immediately, causing the application to hang.

### Steps to Reproduce

1. Write an async void method with an infinite loop (`while (true)`) containing simple animations (e.g. `await view.ScaleTo(0.8, 500);`)
2. Run the app to confirm the animation works and the app doesn't hang.
3. Turn off animations in Android settings, or by turning on a power savings mode.
4. Run the app again.

### Expected Behavior
The animation should run and the app shouldn't hang. Suggestion: the animation should return `Task.Delay()` of the same duration instead of immediately returning.

### Actual Behavior
The app hangs with an infinite loop.

### Basic Information

- Version with issue: 4.2.0.778463
- IDE: Visual Studio 2019 16.2.5
- Platform Target Frameworks:
  - Android: 9.0
- Nuget Packages:
- Affected Devices: Samsung Galaxy S9+

### Reproduction Link

<s>Later, if really needed.</s>

[AsyncAnimationHang.zip](https://github.com/xamarin/Xamarin.Forms/files/3644569/AsyncAnimationHang.zip)

### LINKED ISSUE #8634: [Bug] Android animation issue when device animations are turned off
### Description
Open device Android settings, open Developer Menu, turn off animations.
App animations (eg. view.TranslateTo(0,0)) causes app to freeze.
This also happens if you open Android settings and turn on "Reduce animations" on Samsung devices.

### Steps to Reproduce

1. Turn off animations in device Developer menu
2. Do any view animation such as .TranslateTo or .ScaleTo etc

### Expected Behavior
With animations turned off I would expect the animation to be skipped.

### Actual Behavior
App freezes

### Basic Information

- Version with issue: All
- IDE: VS 2019
- Platform Target Frameworks:
  - Android: 7 and 8
- Nuget Packages: Latest
- Affected Devices:
I suspect all Android devices. Definitely Samsung devices.

hartez:
### Description of Change ###

Modify animation helpers to return `true` (i.e., "canceled") when force-finishing animations because the ticker is system-disabled; this allows users of the animation methods to react accordingly when their awaited looped animations are returning immediately because animations have been disabled by the system (e.g., because of power saving mode).

These changes also add a public method to the Animation class allowing the user to determine if animations have been disabled by the system.

Finally, for newer API levels (26+), the Android power saving mode check no longer just disables the animations; rather, it makes a check with AreAnimationsEnabled; this allows the animations to continue working if the power saving mode doesn't actually disable them (e.g., on Samsung devices with "medium" power saving modes).

See also the discussion [here](https://github.com/xamarin/Xamarin.Forms/issues/7500#issuecomment-552153036)

### Issues Resolved ### 

- fixes #7500
- fixes #8634

### API Changes ###

Added:

`public static bool Animation.IsEnabled { get; }`

### Platforms Affected ### 
- Core/XAML (all platforms)
- Android

None

Not applicable

### Testing Procedure ###
Issue 1556 (manual test)

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/MotionTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/MotionTests.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Threading.Tasks;
 using NUnit.Framework;
 using Xamarin.Forms.Internals;
@@ -247,7 +248,7 @@ namespace Xamarin.Forms.Core.UnitTests
 		[Test]
 		public async Task AnimationExtensionsReturnTrueIfAnimationsDisabled() 
 		{
-			DisableTicker();
+			await DisableTicker();
 
 			var label = new Label { Text = "Foo" };
 			var result = await label.ScaleTo(2, 500);
@@ -258,7 +259,7 @@ namespace Xamarin.Forms.Core.UnitTests
 		[Test, Timeout(2000)]
 		public async Task CanExitAnimationLoopIfAnimationsDisabled() 
 		{
-			DisableTicker();
+			await DisableTicker();
 
 			var run = true;
 			var label = new Label { Text = "Foo" };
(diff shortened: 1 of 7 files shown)


hartez:
API 19 FR test failure and API 19 LR test failure are both AppCenter bugs.

PR #8943, opened 2019-12-18
fix: ios image loader should consider scale

fatfatson:
### Description of Change ###
`Xamarin.Forms.Platform.iOS.FileImageSourceHandler.LoadImageAsync` doesn't use the `scale` parameter, so fix it.

### Platforms Affected ### 

- iOS


samhouts:
@fatfatson Looks like you have more changes in here than you mean to. Can you remove the unnecessary commits? Thanks!

fatfatson:
> @fatfatson Looks like you have more changes in here than you mean to. Can you remove the unnecessary commits? Thanks!

ok, I will close and open a new one.

Issue #3494, opened 2018-08-06
[Enhancement]  TapGestureRecognizer Improvements


PureWeen:
# TapGestureRecognizer Improvements
- IsTapping
	- Specified number of touches have come down and back up but it hasn't reached the number of taps required to be recognized as Tapped.
- Updated Event Args
	- Adds an event to the Gesture Recognizer with additional information about the fingers causing the tap
- NumberOfTouches
	- Allows for multiple touch tapping

_Note: Nothing in this specification is guaranteed to be final; all features, implementations, and interfaces are subject to change._

# API

## Supporting APIs

In order to provide the user with as much available information as possible, we are going to add additional information about all currently active touches. 

#### GestureEventArgs
[Additional Touch Data Attached to Updated events](https://github.com/xamarin/Xamarin.Forms/issues/3479)

### TapGestureRecognizer
```C#
public sealed class TapGestureRecognizer : GestureRecognizer
{
  	public bool IsTapping { get; }
  
  	// Bindable Property will live on GestureRecognizer
	public ICommand CancelledCommand { get; set; }
	public object CancelledCommandParameter { get; set; }
	
	// Bindable Property will live on GestureRecognizer
	public ICommand StartedCommand { get; set; }
	public object StartedCommandParameter { get; set; }  
	
	public static readonly BindableProperty NumberOfTouchesProperty;
	public int NumberOfTouches { get; set;}
  
	public event EventHandler<TapUpdatedEventArgs> TapUpdated;
}
```
### Properties

| API | Description |
| ------------- | ------------- |
| IsTapping | Specified number of touches have come down and back up but it hasn't reached the number of taps required to be recognized as Tapped. |
| CancelledCommand | If the gesture starts then gets cancelled for some reason this command will fire.  |
| CancelledCommandParameter ||
| StartedCommand | Gesture has recognized the touch and is responding to touch changes.  |
| StartedCommandParameter ||
| NumberOfTouchesRequired | Number of touch points required for gesture command to execute (default 1). |

### Events

| API | Description |
| ------------- | ------------- |  
| TapUpdated | Occurs whenever there is a change is state for this Gesture.  |

### TapUpdatedEventArgs 

```csharp
public class TapUpdatedEventArgs : GestureEventArgs
{	
	public GestureStatus StatusType { get; }
}
```

### Properties

| API | Description |
| ------------- | ------------- |
| GestureStatus | Indicates state of gesture (Started, Running, Canceled, Completed). |

[...]

Issue #3479, opened 2018-08-03
[Enhancement] GestureEventArgs - additional touch data for Gesture Updated events


PureWeen:
# GestureEventArg

Additional telemetry data that will be added to *Updated* events.

_Note: Nothing in this specification is guaranteed to be final; all features, implementations, and interfaces are subject to change._

## Touch

```C#
public class Touch
{
	public int TouchIndex { get; set; }
	public Point ViewPosition { get; set; }
	public Point PagePosition { get; set; }
	public Point ScreenPosition { get; set; }
	public GestureStatusType StatusType { get; set; }
	View Target { get; }
}
```

### Properties

| API | Description |
| ------------- | ------------- |
| TouchIndex | First touch is index 0, second touch is index 1, etc. This indicates the touch index that updates and commands will listen to. |
| ViewPosition | The relative Position of this touch for the given view the gesture is attached to.  |
| PagePosition | The relative Position of this touch to the parent page |
| ScreenPosition | The absolute Position of the touch on the screen. |
| GestureStatusType | The state of the touch. This is mainly used when inspecting ChangedTouches to see how said touches have changed |
| Target | The view associated with this Gesture. |

## TouchEvent

```C#
public class TouchEvent
{
	public IReadOnlyList<Touch> Touches { get; set; }
	public IReadOnlyList<Touch> ChangedTouches { get; set; }
	public IReadOnlyList<Touch> TargetTouches { get; set; }
	View Target { get; }
}
```

### Properties

| API | Description |
| ------------- | ------------- |
| Touches | All active touches and their positions. This will include touches that are happening outside this view |
| ChangedTouches | All touches that changed from the last event.  |
| TargetTouches | All touches affecting this view |

### GestureEventArgs

Eventually all of the gestures eventargs will inherit from GestureEventArgs. Initially it'll just be LongPressGestureRecognizer
``` C#
public class GestureEventArgs : EventArgs
{
	public TouchEvent TouchEvent { get; private set; }
}

```

### Properties
| API | Description |
| ------------- | ------------- |
| TouchEvent | Information about the touch event that caused an update on a gesture. |

### Examples
```C#
public class LongPressContainer : ContentView
{
  BoxView touchView;
  
  public LongPressContainer ()
  {
    var longPressGesture = new LongPressGestureRecognizer ();
    longPressGesture.LongPressUpdated += OnLongPressUpdated;
    GestureRecognizers.Add (longPressGesture);
    
[...]

PR #9055, opened 2019-12-31
Continue #6389: Added tap position to the Tapped event of the TapGestureRecognizer

### LINKED ISSUE #5347: [Enhancement] Add more data to TapGestureRecognizer.Tapped event

Happypig375:
### Description of Change ###

Added tap position to the Tapped Event of the TapGestureRecognizer.
### Issues Resolved ### 

Fixes #5347
Continue #6389
### API Changes ###

 
```diff
namespace Xamarin.Forms
{
+	public class TapEventArgs : EventArgs
+	{
+		public TapEventArgs(object parameter, Point position);
+		public object Parameter { get; }
+		public Point Position { get; }
+	}
+	public delegate void TapEventHandler(object sender, TapEventArgs args);
	public sealed class TapGestureRecognizer : GestureRecognizer
	{
+		public event TapEventHandler Tap;
	}
}
```

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP
- macOS
- WPF
- GTK
- Tizen

None

Not applicable

### Testing Procedure ###

Open corresponding page in Gallery

### PR Checklist ###


Happypig375:
First PR of 2020 :rocket:

Happypig375:
@samhouts Missing macOS tag

Happypig375:
@samhouts A "high impact" PR was left sitting for months until a file conflict occurs... Really Xamarin?

Xamarin.Forms.Core/TapGestureRecognizer.cs
```diff
@@ -35,6 +35,7 @@ public int NumberOfTapsRequired
 		}
 
 		public event EventHandler Tapped;
+		public event TapEventHandler Tap;
```
jsuarezruiz, line 38: This is fine because is backward-compatible but I am concerned that it will cause confusion. Just reading the names, it appears that Tapped and Tap are exactly the same. Could we rename `Tap` to something that better indicates the difference with `Tapped`? Example: **LocationTapped**.

Happypig375: I am even more confused why Xamarin decided not to provide the tapped location in the first place. `Tap` is supposed to be `Tapped` but better. It also conveys the event order: first `Tap`, then `Tapped`. By using a shorter name, `Tap` can be more attractive than `Tapped` and will attract more usage. `Tapped`, being `Tap` but more useless, can be deprecated in the long run.

Happypig375: If you insist on the confusion argument, I can deprecate `Tapped` for you.

PureWeen: I think a lot of what you're articulating was taken into account with our gestures specs here

https://github.com/xamarin/Xamarin.Forms/issues/3494
and here
https://github.com/xamarin/Xamarin.Forms/issues/3479

In there you have `GestureStatus` to convey order and when things are canceled plus specific events and commands to hook into

The idea was to also generalize these types of args so it could be used by all the gestures..

Thoughts?

Happypig375: If those changes can be added in the next release, then I would be happy to side with you. However, seeing that the linked PR is not only stuck in limbo, but also forever in a state of WIP, I believe that a simpler PR such as this one can be merged more quickly and benefit more users.

PureWeen: The linked PR is WIP mainly because it also doesn't match the spec

can we reshape this PR slightly to match those specs?

Just taking on a new API to later be deprecated isn't really desireable

Happypig375: What are the changes you are specifically talking about?

PureWeen: Position on its own doesn't have enough context. Position to what? 

Which those specs address with
```C#
public class Touch
{
	public int TouchIndex { get; set; }
	public Point ViewPosition { get; set; }
	public Point PagePosition { get; set; }
	public Point ScreenPosition { get; set; }
	public GestureStatusType StatusType { get; set; }
	View Target { get; }
}
```

So my thinking would be to create the gesture args described here
https://github.com/xamarin/Xamarin.Forms/issues/3479

you can just implement a subset of those based on what you've already done

Or possibly just match those properties to your `TapEventArgs`

Or we can figure out why the spec we put together is problematic. Make sure we're happy with the path forward and can reach at least a minimal set of APIs that reshapes those specs or just use a subset of those APIs with this PR

Happypig375: If `VisualElement` has `ViewX`, `PageX` and `ScreenX`, then I will agree with the API. However, `VisualElement` has the one and only `X`. Therefore, context is already achieved.

samhouts:

PR #9063, opened 2020-01-02
Fix ResolveLayoutChanges crash

### LINKED ISSUE #9062: [Bug] Crash in Layout. ResolveLayoutChanges
### Description
Crash in `Layout.ResolveLayoutChanges`: `System.InvalidOperationException: Collection was modified; enumeration operation may not execute.`

### Steps to Reproduce

Steps are unknown yet, we are trying to reliably recreate it. We are seeing the crashes through AppCenter and are getting plenty of them across different Android devices.

I have a console app that replicates the underlying bug:
https://gist.github.com/slang25/f26349340f74300d67f7c552cb2b2493

### Expected Behavior
No crash.

### Actual Behavior
In a `StackLayout`, in `Layout.OnChildMeasureInvalidated`, it should queue up layout update operations in `s_resolutionList` for `Layout.ResolveLayoutChanges` to process.

`Layout.ResolveLayoutChanges` "copies" `s_resolutionList`, however it copies the reference, this means that if `GetElementDepth` is slow during `OnChildMeasureInvalidated`, it could have dereferenced the old list. By adding an item to the old `s_resolutionList` reference, it risks this exception for occurring.

### Other Information

Console app that replicates basic non-thread-safe underlying issue can be found here:
https://gist.github.com/slang25/f26349340f74300d67f7c552cb2b2493

There are a few quick fixes, including this:
https://github.com/xamarin/Xamarin.Forms/compare/xamarin:master...slang25:fix-layout-crash?expand=1

But I'd like to work on a more complete fix.

slang25:
### Description of Change ###

By un-inlining the kvp to add to `s_resolutionList`, it is dereferenced later, vastly reducing the likelihood that `s_resolutionList` is an older instance that is currently being enumerated in `ResolveLayoutChanges`.

### Issues Resolved ### 

- fixes #9062

### API Changes ### 
 None

### Platforms Affected ### 
- Android
(although other platforms should be regression tested)

None

Not applicable

### Testing Procedure ###
Unfortunately because this is a rare race condition, this is hard. We are working on getting a much more reliable recreation, but this may take some time.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Layout.cs
+++ b/Xamarin.Forms.Core/Layout.cs
@@ -355,7 +355,8 @@ namespace Xamarin.Forms
 				InvalidateMeasureInternal(InvalidationTrigger.MeasureChanged);
 			}
 
-			s_resolutionList.Add(new KeyValuePair<Layout, int>(this, GetElementDepth(this)));
+			var kvp = new KeyValuePair<Layout, int>(this, GetElementDepth(this));
+			s_resolutionList.Add(kvp);
 			if (!s_relayoutInProgress)
 			{
 				s_relayoutInProgress = true;


slang25:
I need to move this over to a release branch.

This change is a quick mitigation, I would also like to change the code around here:
https://github.com/xamarin/Xamarin.Forms/pull/9063/files#diff-39c7bc04661bbb5f63e514fd9c73e38cR383

I think `s_resolutionList` should be copied using a new `List` passing the old one into the constructor. This will add more allocations, but this can be reduced by not newing a `List` on the following line, but calling `s_resolutionList.Clear()`, the internal `Array` will be reused. I understand this is a bigger change and warrant further discussion.

slang25:
Closed for reasons explained here: https://github.com/xamarin/Xamarin.Forms/issues/9062#issuecomment-573060498

PR #9132, opened 2020-01-08
Bump WinUI Nuget version if you are using Min target >= 16299

PureWeen:
### Description of Change ###

Bump WinUI to latest released version

### Platforms Affected ### 
- UWP

### Testing Procedure ###
- make sure you can install nugets into VS2017 and VS2019 and that everything runs

### PR Checklist ###

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -226,7 +226,7 @@ Task("Build")
         MSBuild("./Xamarin.Forms.Platform.UAP/Xamarin.Forms.Platform.UAP.csproj",
                     GetMSBuildSettings()
                         .WithRestore()
-                        .WithProperty("DisableEmbeddedXbf", "true"));
+                        .WithProperty("DisableEmbeddedXbf", "false"));
     }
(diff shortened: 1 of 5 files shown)


build.cake
```diff
@@ -226,7 +226,7 @@ Task("Build")
         MSBuild("./Xamarin.Forms.Platform.UAP/Xamarin.Forms.Platform.UAP.csproj",
                     GetMSBuildSettings()
                         .WithRestore()
-                        .WithProperty("DisableEmbeddedXbf", "true"));
+                        .WithProperty("DisableEmbeddedXbf", "false"));
```
samhouts: is this supposed to be changed?

PureWeen: Yea

This is only used for cases when you're creating the nugets from the command line not the CI

Basically it's telling it to "Not Not Embed Xbf files" which is what we want opposed to just "Not Embedding them" :-)

PR #9133, opened 2020-01-08
[Enhancement] Add Asynchronous Callback to Device.StartTimer

### LINKED ISSUE #9131: [Enhancement] Add Asynchronous Callback to Device.StartTimer
## Summary

This will allow developers to start timers that use an asynchronous callback.

## API Changes

```csharp
interface IPlatformServices
{
    //Add New API
    void StartTimer(TimeSpan interval, Func<Task<bool>> callback)
}
```

## Intended Use Case

Many mobile tasks today are asynchronous, e.g. calls to web APIs, saving/retrieving items from disk, getting device sensor information like GPS coordinates.

```csharp
class App : Application
{
    protected override void OnStart()
    {
        Device.StartTime(TimeSpan.FromSeconds(5), SendUpdate);
    }

    async Task<bool> SendUpdate()
    {
        var isSuccessful = await DataServices.PostData();
        return isSuccessful;
    }
}
```

brminnick:
### Description of Change ###

Implements an asynchronous callback to `Device.StartTimer`

This will enable devs to use `StartTime` to execute asynchronous tasks, e.g. calls to web APIs, saving/retrieving items from disk, getting device sensor information like GPS coordinates.

```csharp
class App : Application
{
    protected override void OnStart()
    {
        Device.StartTime(TimeSpan.FromSeconds(5), SendUpdate);
    }

    async Task<bool> SendUpdate()
    {
        var isSuccessful = await DataServices.PostData();
        return isSuccessful;
    }
}
```

### Issues Resolved ### 

- fixes #9131 

### API Changes ###

```csharp
interface IPlatformServices
{
    //Add New API
    void StartTimer(TimeSpan interval, Func<Task<bool>> callback)
}
```
 
 None

### Platforms Affected ###

- Core/XAML (all platforms)
- iOS
- Android
- UWP
- WPF
- Tizen
- macOS
- GTK

None

Not applicable

### Testing Procedure ###

The following tests will be created for each platform:
- [ ] Call `StartTimer(TimeSpan interval, Func<Task<bool>> callback)` with a successful Task that returns `true`; confirm timer executes again
- [ ] Call `StartTimer(TimeSpan interval, Func<Task<bool>> callback)` with a successful Task that returns `false`; confirm timer does not execute again
- [ ] Call `StartTimer(TimeSpan interval, Func<Task<bool>> callback)` with a Task that throws an exception before `await` is called in that `Task`; confirm timer does not execute again at the scheduled interval
- [ ] Call `StartTimer(TimeSpan interval, Func<Task<bool>> callback)` with a Task that throws an exception after `await` is called in that `Task`; confirm timer does not execute again at the scheduled interval

### PR Checklist ###

- [ x ] Targets the correct branch

### DIFF
--- a/Xamarin.Forms.Platform.UAP/WindowsBasePlatformServices.cs
+++ b/Xamarin.Forms.Platform.UAP/WindowsBasePlatformServices.cs
@@ -152,6 +152,25 @@ namespace Xamarin.Forms.Platform.UWP
 			CompositionTarget.Rendering += renderingFrameEventHandler;
 		}
 
+		public void StartTimer(TimeSpan interval, Func<Task<bool>> callback)
+		{
+			var timerTick = 0L;
+			var stopWatch = new Stopwatch();
+			stopWatch.Start();
+			async void renderingFrameEventHandler(object sender, object args)
+			{
+				var newTimerTick = stopWatch.ElapsedMilliseconds / (long)interval.TotalMilliseconds;
+				if (newTimerTick == timerTick)
+					return;
+				timerTick = newTimerTick;
+				bool result = await callback();
+				if (result)
+					return;
+				CompositionTarget.Rendering -= renderingFrameEventHandler;
+			}
+			CompositionTarget.Rendering += renderingFrameEventHandler;
+		}
+
 		public void QuitApplication()
 		{
 			Log.Warning(nameof(WindowsBasePlatformServices), "Platform doesn't implement QuitApp");
--- a/Xamarin.Forms.Platform.UAP/WindowsBasePlatformServices.cs
+++ b/Xamarin.Forms.Platform.UAP/WindowsBasePlatformServices.cs
@@ -145,9 +145,8 @@ namespace Xamarin.Forms.Platform.UWP
 					return;
 				timerTick = newTimerTick;
 				bool result = callback();
-				if (result)
-					return;
-				CompositionTarget.Rendering -= renderingFrameEventHandler;
+				if (!result)
+					CompositionTarget.Rendering -= renderingFrameEventHandler;
 			}
 			CompositionTarget.Rendering += renderingFrameEventHandler;
 		}
@@ -164,9 +163,8 @@ namespace Xamarin.Forms.Platform.UWP
 					return;
 				timerTick = newTimerTick;
 				bool result = await callback();
-				if (result)
-					return;
-				CompositionTarget.Rendering -= renderingFrameEventHandler;
+				if (!result)
+					CompositionTarget.Rendering -= renderingFrameEventHandler;
 			}
 			CompositionTarget.Rendering += renderingFrameEventHandler;
 		}
--- a/Xamarin.Forms.Platform.WPF/WPFPlatformServices.cs
+++ b/Xamarin.Forms.Platform.WPF/WPFPlatformServices.cs
@@ -151,6 +151,18 @@ namespace Xamarin.Forms.Platform.WPF
 			};
 		}
 
+		public void StartTimer(TimeSpan interval, Func<Task<bool>> callback)
+		{
+			var timer = new DispatcherTimer(DispatcherPriority.Background, System.Windows.Application.Current.Dispatcher) { Interval = interval };
+			timer.Start();
+			timer.Tick += async (sender, args) =>
+			{
+				bool result = await callback();
+				if (!result)
+					timer.Stop();
+			};
+		}
+
 		public void QuitApplication()
 		{
 			System.Windows.Application.Current.Shutdown();
--- a/Xamarin.Forms.Platform.iOS/Forms.cs
+++ b/Xamarin.Forms.Platform.iOS/Forms.cs
@@ -334,6 +334,22 @@ namespace Xamarin.Forms
 				NSRunLoop.Main.AddTimer(timer, NSRunLoopMode.Common);
 			}
 
+			public void StartTimer(TimeSpan interval, Func<Task<bool>> callback)
+			{
+				NSTimer timer = NSTimer.CreateRepeatingTimer(interval, t =>
+				{
+					callback().ContinueWith(completedTask =>
+					{
+						if (completedTask.IsFaulted
+							|| (completedTask.IsCompleted && !completedTask.Result))
+						{
+							t.Invalidate();
+						}
+					});
+				});
+				NSRunLoop.Main.AddTimer(timer, NSRunLoopMode.Common);
+			}
+
 			HttpClient GetHttpClient()
 			{
 				var proxy = CoreFoundation.CFNetwork.GetSystemProxySettings();
--- a/Xamarin.Forms.Platform.iOS/Forms.cs
+++ b/Xamarin.Forms.Platform.iOS/Forms.cs
@@ -336,16 +336,11 @@ namespace Xamarin.Forms
 
 			public void StartTimer(TimeSpan interval, Func<Task<bool>> callback)
 			{
-				NSTimer timer = NSTimer.CreateRepeatingTimer(interval, t =>
+				NSTimer timer = NSTimer.CreateRepeatingTimer(interval, async t =>
 				{
-					callback().ContinueWith(completedTask =>
-					{
-						if (completedTask.IsFaulted
-							|| (completedTask.IsCompleted && !completedTask.Result))
-						{
-							t.Invalidate();
-						}
-					});
+					var result = await callback();
+					if (result)
+						t.Invalidate();
 				});
 				NSRunLoop.Main.AddTimer(timer, NSRunLoopMode.Common);
 			}
--- a/Xamarin.Forms.Core.UnitTests/MockPlatformServices.cs
+++ b/Xamarin.Forms.Core.UnitTests/MockPlatformServices.cs
@@ -120,14 +120,13 @@ namespace Xamarin.Forms.Core.UnitTests
 		public void StartTimer(TimeSpan interval, Func<Task<bool>> callback)
 		{
 			Timer timer = null;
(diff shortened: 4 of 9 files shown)


brminnick:
Including @ahoefling for review

Xamarin.Forms.Platform.UAP/WindowsBasePlatformServices.cs
```diff
@@ -145,9 +145,26 @@ void renderingFrameEventHandler(object sender, object args)
 					return;
 				timerTick = newTimerTick;
 				bool result = callback();
-				if (result)
+				if (!result)
```
VladislavAntonyuk: remove temp variable. if (!callback())

Xamarin.Forms.Platform.WPF/WPFPlatformServices.cs
```diff
@@ -151,6 +151,18 @@ public void StartTimer(TimeSpan interval, Func<bool> callback)
 			};
 		}
 
+		public void StartTimer(TimeSpan interval, Func<Task<bool>> callback)
+		{
+			var timer = new DispatcherTimer(DispatcherPriority.Background, System.Windows.Application.Current.Dispatcher) { Interval = interval };
+			timer.Start();
+			timer.Tick += async (sender, args) =>
```
VladislavAntonyuk: sender and args are not used

Xamarin.Forms.Platform.iOS/Forms.cs
```diff
@@ -334,6 +334,17 @@ public void StartTimer(TimeSpan interval, Func<bool> callback)
 				NSRunLoop.Main.AddTimer(timer, NSRunLoopMode.Common);
 			}
 
+			public void StartTimer(TimeSpan interval, Func<Task<bool>> callback)
+			{
+				NSTimer timer = NSTimer.CreateRepeatingTimer(interval, async t =>
+				{
+					var result = await callback();
+					if (result)
```
VladislavAntonyuk: everywhere you use if (!result), but here if (result). is it correct?

Xamarin.Forms.Platform.UAP/WindowsBasePlatformServices.cs
ahoefling: This appears to be modifying the original API `void StartTimer(TimeSpan interval, Func<bool> callback)`, where this PR is to only add a new API. 

Is this code wrong and needs to be updated? 

If I understand the event wire-up code correctly it removes the event handler when the result is true. I believe the intent of the API is to continue if it is true. Something doesn't quite add up here to me.

Given the code snippet below we should stop the timer after the first run since it returns false
```c#
Device.StartTimer(TimeSpan.FromMilliseconds(250), () =>
{
    MyMethod();
    return false;
}
```

After typing this out, I agree with the change. I am confused on how this currently works because the same code I provided above currently works in Xamarin.Forms 4.1 and I don't want to add a change that may cause problems if it reverses the behavior.

Xamarin.Forms.Platform.iOS/Forms.cs
ahoefling: This appears to be a typo to me as well. Looking at line 331 from the non `Task` API, it shows that the behavior checks for false and not true. Maybe this should be `if (!result)`

Code snippet from non-async API
https://github.com/xamarin/Xamarin.Forms/blob/75497c66b2de7941c101af2f651c82483d987ca5/Xamarin.Forms.Platform.iOS/Forms.cs#L331-L332

I'm not an iOS expert but is `t.Invalidate();` restarting the timer or re-running the function? If I understand this correctly that function stops the process.

Xamarin.Forms.Platform.WPF/WPFPlatformServices.cs
```diff
@@ -151,6 +151,18 @@ public void StartTimer(TimeSpan interval, Func<bool> callback)
 			};
 		}
 
+		public void StartTimer(TimeSpan interval, Func<Task<bool>> callback)
+		{
+			var timer = new DispatcherTimer(DispatcherPriority.Background, System.Windows.Application.Current.Dispatcher) { Interval = interval };
```
ahoefling: Is this pushes the callback to the background thread? The other implementations attempt to run the callback function on the main thread. I see that this mimics the behavior of the non-async API. Not sure if this is something that should be consistent across the platforms or if there is a technical limitation

Xamarin.Forms.Platform.iOS/Forms.cs
brminnick: Good catch! Yup - I totally forgot the `!` 🙈

brminnick:
Thanks @VladislavAntonyuk  @ahoefling!

Regarding the execution context of `StartTimer` (ie background thread vs main thread), it looks like each platform Marshalls to the main thread every time, except iOS & macOS. I agree that this is a bit strange and it would be a good idea to open a new Issue to dive deeper into it.

This PR should conform to existing functionality of the original `StartTimer` API to avoid any breaking changes.

| Platform | Behavior | Source |
| -------- | --------- | ------- |
| Tizen | Always marshalls to Main Thread | [Source](https://github.com/xamarin/Xamarin.Forms/blob/75497c66b2de7941c101af2f651c82483d987ca5/Xamarin.Forms.Platform.Tizen/TizenPlatformServices.cs#L128) |
| WPF | Always Marshalls to Main thread | [Source](https://stackoverflow.com/a/11625264/5953643) | 
| iOS | Indifferent. If `StartTimer` is called on a background thread, it will run on a background thread. If `StartTimer` is called on main thread, it will run on the main thread | [Source](https://stackoverflow.com/a/54342962/5953643) |
| macOS | Indifferent. If `StartTimer` is called on a background thread, it will run on a background thread. If `StartTimer` is called on main thread, it will run on the main thread | [Source](https://stackoverflow.com/a/54342962/5953643) |
| Android | Always marshalls to main thread | [Source](https://github.com/xamarin/Xamarin.Forms/blob/190dda803621cb61a2d93834cdbddae2bac3c777/Xamarin.Forms.Platform.Android/Forms.cs#L747)
| GTK | Always marshalls to main thread | [Source](https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html) |

Xamarin.Forms.Platform.WPF/WPFPlatformServices.cs
brminnick: `DispatcherTimer.Tick` is an EventHandler which can only subscribe to the following method signature: `void (object sender, EventArgs e)`

Xamarin.Forms.Platform.WPF/WPFPlatformServices.cs
brminnick: I'm not a WPF expert, but I it looks like this runs on the main thread.

I first thought it was running on a background thread too because of `DispatcherPriority.Background`. However the docs indicate that [`Dispatcher.Background`](https://docs.microsoft.com/dotnet/api/system.windows.threading.dispatcherpriority?view=netframework-4.8) is its priority, meaning specifically that this task will execute on the passed in thread (`Application.Current.Dispatcher`, i.e. the [Main Thread](https://stackoverflow.com/a/11625264/5953643)) at this priority:
> The enumeration value is 4. Operations are processed after all other non-idle operations are completed.

brminnick:
While testing, I found that when an Exception is thrown, `Device.Timer` continues running

The default behavior when an exception is thrown should be to return `false`.

I've opened up an Issue here for which I'll generate a PR once this PR has been merged: https://github.com/xamarin/Xamarin.Forms/issues/9146

brminnick:
@StephaneDelcroix Unit Tests & UI Tests are complete. PR is ready for review 👍

samhouts:
@brminnick The tests are not passing :(


Issue9131.Async Device Timer With Successful Shorter Task Returning False
Issue9131.Device Timer With Successful Function Returning False
Issue9131.Device Timer With Successful Funnction Returning True
Issue9131.Async Device Timer With Successful Longer Task Returning False
Issue9131.Async Device Timer With Successful Longer Task Returning True
Issue9131.Async Device Timer With Successful Shorter Task Returning True

brminnick:
No worries - I'll check out what has changed since I submitted the PR and fix it up!

brminnick:
@samhouts Should be good to go now! I fixed a couple problems due to multi-threading race conditions.

I confirmed the following tests locally:
✅ Unit Tests
✅ iOS UI Tests
✅ Android UI Tests
✅ macOS UI Tests

Skenja:
> we use device timer internally to get more control while doing animations.

Sure, but this is a public API, not an internal one - it should stand to reason that a public API should be modeled according to the needs of the public.

> in this particular use case, you're good with Task.Delay() or Timer().

`Task.Delay()` solution for this is hackish at best, especially if this needs to run periodically.
 `new Timer()` is a good approach if you want to define it in a common project, such as Xamarin.Forms project. However, `new Timer()` isn't quite the same as per-platform implementations, and it would be better to use such native API's where possible. However, if you were to use native API's in this case, it would require a bit more code, (boilerplate repetitive code might I add), which goes against the vision of Xamarin.Forms framework.

Personally, I have been involved on multiple projects where such an API would have been useful, especially in a periodical `Task` invoking capacity (multi-app multi-user solutions, highly performance-dependent distributed consensus applications). Such an API would have saved a certain amount of repetitive boilerplate code which never should have been needed in a Xamarin.Forms environment. 

Sure, there are other solutions to this problem, and this isn't strictly speaking necessary, however, the `Device.StartTimer` API feels lacking without `Func<Task<>>` support. That is especially obvious when you take into consideration the fact that most of the oft used Xamarin.Forms APIs are async oriented.

All of the above considered, combined with the complete PR, and the actual small size of the code introduced, I believe this should be at least reconsidered.

PR #9163, opened 2020-01-11
[Android][iOS] TabBar adds Reselect Event

### LINKED ISSUE #6544: [Enhancement] Shell TabBar Action on Current Tab Tapped
## Summary
When a Shell has tabs it would be great if there was an easy way to do something when the current tab is tapped. 

I can't see anything documented for this so hopefully i'm not missing something.

## API Changes
I am not too sure of the best way to go about this.
Maybe something like a command property OnCurrentTabTapped for each tab?
I am sure one of the team will have a better idea on this.

## Intended Use Case
Many popular apps have actions where the current tab is tapped e.g fitbit and youtube tapping the home tab when on the home page takes you back to the top of the page. It would also be useful when a sub page is in view to get back to the main page.

The have been some pre shell solutions for this e.g:
 https://stackoverflow.com/questions/42801222/xamarin-forms-tabbedpage-event-when-current-tab-is-tapped-to-refresh-the-page
https://stackoverflow.com/questions/44847298/how-to-detect-xamarin-forms-tabbed-page-click-ios

huybuidac:
### Description of Change ###

TabBar adds Reselect Event

### Issues Resolved ### 

- fixes #6544

### API Changes ###

Added:
 - event ShellItem.Reselected // event handler

### Platforms Affected ### 

- iOS
- Android

None

Not applicable

### PR Checklist ###


xamarindevelopervietnam:
that's nice!

Xamarin.Forms.Platform.iOS/Renderers/ShellItemRenderer.cs
```diff
@@ -85,6 +85,13 @@ public override void ViewDidLoad()
 
 			ShouldSelectViewController = (tabController, viewController) =>
 			{
+				if (viewController == SelectedViewController)
+				{
+					// let user handers it
```
samhouts:
```suggestion
					// let user handle it
```

PureWeen:
@huybuidac I'm going to close this PR for now. 

This API doesn't really work for Shell. Adding `Reselected` to only ShellItem doesn't quite work because ShellItem/ShellContent/ShellSection will all need to express this idea.

We're thinking that adding a Command Property to BaseShellitem might make the most sense and if we do add *Reselected* it would need to be an attached property that you could apply to a ContentPage

PR #9175, opened 2020-01-13
[Android] Changes in SwipeView to avoid the use of APIs that require a specific API level

### LINKED ISSUE #9030: [Bug] ClassNotFoundException when using snap points on API < 23
### Description
A `ClassNotFoundException` is thrown when using snap points in a `CollectionView` on Android versions < 6.0 (API 23).

This is due to [wiring](https://github.com/xamarin/Xamarin.Forms/blob/1de59c5e7ea5c3ee2dc7250f8ebd31eda9e650a7/Xamarin.Forms.Platform.Android/CollectionView/NongreedySnapHelper.cs#L22) into the View.ScrollChange event which is only available in API 23 and above. Refer to [Android API specification](https://developer.android.com/reference/android/view/View.OnScrollChangeListener).

Note: This event is also used in the [SwipeViewRenderer](https://github.com/xamarin/Xamarin.Forms/blob/d45bd2195063cd5064a003eedcef4caffbcea9a1/Xamarin.Forms.Platform.Android/Renderers/SwipeViewRenderer.cs#L160) likely leading to the same exception.

### Steps to Reproduce

1. Open the control samples app on an affected device.
2. Open the CollectionView gallery snap points page.

### Expected Behavior
The app should not crash.

### Actual Behavior
The app crashes with the following exception:

```
ava.Lang.ClassNotFoundException: Didn't find class "mono.android.view.View_OnScrollChangeListenerImplementor"

Android.Views.View+IOnScrollChangeListenerImplementor..ctor (System.Object sender) [0x00010] in <82e2c298095040ab99611f85aa135203>:0
View.__CreateIOnScrollChangeListenerImplementor ()
EventHelper.AddEventHandler[TInterface,TImplementor] (System.WeakReference& implementor, System.Func`1[TResult] creator, System.Action`1[T] setListener, System.Action`1[T] add)
View.add_ScrollChange (System.EventHandler`1[TEventArgs] value)
NongreedySnapHelper.AttachToRecyclerView (Android.Support.V7.Widget.RecyclerView recyclerView)
```

### Basic Information
- Version with issue: 4.4.0
- Affected Devices: Android devices < 6.0 (API 23).

### Screenshots
N/A

### Reproduction Link
N/A

jsuarezruiz:
### Description of Change ###

Changed code to avoid the usage of some APIs that require a specific API Level on Android.

### Issues Resolved ### 

- addresses parts of #9030

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

None

Not applicable

### Testing Procedure ###
Verify the Build and try any SwipeView sample on Android <6.0.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/SwipeViewRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/SwipeViewRenderer.cs
@@ -16,7 +16,6 @@ using AButton = Android.Support.V7.Widget.AppCompatButton;
 using Android.Support.V7.Widget;
 #endif
 using Android.Views;
-using Android.Widget;
 using Xamarin.Forms.Internals;
 using Xamarin.Forms.PlatformConfiguration.AndroidSpecific;
 using APointF = Android.Graphics.PointF;
@@ -32,10 +31,11 @@ namespace Xamarin.Forms.Platform.Android
 		const int SwipeThresholdMargin = 6;
 		const int SwipeItemWidth = 100;
 		const long SwipeAnimationDuration = 200;
+		const double SwipeMinimumDelta = 10;
 
 		readonly Context _context;
 		GestureDetector _detector;
-		AView _scrollParent;
+		View _scrollParent;
 		AView _contentView;
 		LinearLayoutCompat _actionView;
 		SwipeTransitionMode _swipeTransitionMode;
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.Android/Renderers/SwipeViewRenderer.cs
```diff
@@ -225,7 +230,7 @@ public override bool OnTouchEvent(MotionEvent e)
 			float x = Math.Abs((_downX - e.GetX()) / density);
 			float y = Math.Abs((_downY - e.GetY()) / density);
 
-			if (e.Action != MotionEventActions.Move | (x > 10f || y > 10f))
+			if (e.Action != MotionEventActions.Move | (x > SwipeMinimumDelta || y > SwipeMinimumDelta))
```
VladislavAntonyuk: I think we should check swipe direction
If left or right & x>delta
If top or bottom & y> delta

velocitysystems:
Thanks for this PR @jsuarezruiz, however this does not entirely fix the issue #9030 as it does not include the fix when using snap points with `CollectionView` on Android.

hartez:
> Thanks for this PR @jsuarezruiz, however this does not entirely fix the issue #9030 as it does not include the fix when using snap points with `CollectionView` on Android.

I updated the description to reflect that this is fixing _part_ of 9030.

PR #9201, opened 2020-01-14
[GTK] Fix button shrinking

### LINKED ISSUE #9092: [Bug] GTK: reenabling button doesn't work if page is a not-initial NavigationPage
### Steps to Reproduce

1. Clone https://github.com/knocte/geewallet , switch to branch `bugreport/gtkButtonDisabling`.
2. Run the GTK frontend.
3. Click on the button "FOO" to go to next page.
4. Click on the button "BAR": an ALERT modal dialog will show the text "BAZ".
5. Click "OK" to dismiss the alert.

### Expected Behavior
(What happens if you run any other frontend, such as the Android one: ) The button "BAR" was disabled and reenabled once the dialog was shown, and its text changed to "reenabled". The button can be clicked again. (If you don't use navigation pages, this is also the actual behaviour in GTK.)

### Actual Behavior
The button is still disabled and can't be clicked.

### Basic Information

- Last known good version: ?
- IDE: VS4Mac 8.3.11
- Platform Target Frameworks: 
  - Android: v9.0
- Affected Devices: GTK

### Reproduction Link

https://github.com/knocte/geewallet/tree/bugreport/gtkButtonDisabling

z3ut:
### Description of Change ###

Using buttons on page without navbar causes them to shrink to 1px width on GTK. Two things leading to this:

If a button desired size is bigger than available space, button [would shrink to width constraint](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.GTK/Extensions/WidgetExtensions.cs#L39) and would stay this way on next cycles, even if it could get more space;

Each page starts to allocate children sizes with its initial size 1px X 1px, on next cycles getting real window size. Pages with navbar would subtract navbar height from available size and get negative number, so they won't ask children to allocate space on this cycle. Pages without any navbar would calculate its children sizes on page 1px wide, stucking buttons with width=1px for all next cycles.

Overriding ButtonRendered.GetDesiredSize method to remove button size mutation fixed this behavior.

### Issues Resolved ### 

- fixes #9092 (this is the real reason behind the bug, not disabling button or changing navigation pages)

### API Changes ###

 None

### Platforms Affected ### 

- GTK

Buttons should stop to shrink

No navbar, green row on the top of the page

Before:
Xamarin Forms GTK# Backend 13-Jan-20 22_19_32

After:
Xamarin Forms GTK# Backend 13-Jan-20 22_20_13

### Testing Procedure ###

[GTK] Add button on page without navbars. Confirm that button has right width. Change button text and WidthRequest to check size changing and constraint fitting.

Tested this behavior on Windows 10 and Kubuntu 19.04

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue9092.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue9092.cs
@@ -1,0 +1,85 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.Forms.Core.UITests;
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 9092, "[Bug] GTK: reenabling button doesn't work if page is a not-initial NavigationPage", PlatformAffected.Gtk)]
+	public class Issue9092 : TestNavigationPage
+	{
+		protected override void Init()
+		{
+			var page = new ContentPage9092();
+			NavigationPage.SetHasNavigationBar(page, false);
+
+			var navigationPage = new NavigationPage(page);
+			NavigationPage.SetHasNavigationBar(navigationPage, false);
+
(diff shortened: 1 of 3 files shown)


z3ut:
My fault, didn't notice "building on" before "targeting the right branch"

knocte:
@jsuarezruiz PR has been updated, can you check again?

PR #9203, opened 2020-01-14
fix: ios image loader should consider scale

fatfatson:
### Description of Change ###
`Xamarin.Forms.Platform.iOS.FileImageSourceHandler.LoadImageAsync` doesn't use the `scale` parameter, so fix it.

### Platforms Affected ### 

- iOS


jfversluis:
No issue attached so note sure what this is supposed to fix :) Also seems like the scale can be less than 1? So this fix seems incomplete. Closing this for now.

Issue #8743, opened 2019-12-04
[Bug][UWP] SearchBar does not respect FontSize on 4.3.0 


jcmanke:
### Description

Setting a FontSize, either using a number or a NamedSize, does not have any effect on a SearchBar on UWP in version 4.3.0.

### Steps to Reproduce

1. Create a project with Xamarin.Forms 4.3.0.
2. Add a SearchBar to a page and set the FontSize.
3. Run the project.

### Expected Behavior

FontSize is applied.

### Actual Behavior

FontSize is not applied.

### Basic Information

- Version with issue: 4.3.0-pre2 and later
- Last known good version: 
  - Stable: 4.2.0.910310
  - Pre-release: 4.3.0-pre1
- IDE: Visual Studio Enterprise 2019, version 16.3.10
- Platform Target Frameworks: 
  - UWP:  16299

### Screenshots
4.2.0:
![SearchBarFontSize1](https://user-images.githubusercontent.com/6132629/70183050-abc09d80-16aa-11ea-8162-16ff61abdb5b.png)

4.3.0:
![SearchBarFontSize2](https://user-images.githubusercontent.com/6132629/70183056-af542480-16aa-11ea-9b49-8256d6057ed8.png)

### Reproduction Link

[SearchBarFontSize.zip](https://github.com/xamarin/Xamarin.Forms/files/3924030/SearchBarFontSize.zip)

WayaFlyfeather:
I can confirm this, and have located the bug to [these lines](https://github.com/xamarin/Xamarin.Forms/blob/f51099c3823cd7433d1fa95c4cff9227b2a59629/Xamarin.Forms.Platform.UAP/Forms.cs#L37-L44), introduced in #7319 - here UWP resources are merged into the ResourceDictionary, overwriting the default style for the _AutoSuggestBox_ defined in [AutoSuggestStyle.xaml](https://github.com/xamarin/Xamarin.Forms/blob/b80d3aedc716018f0e213af666307aa6d14ddcc3/Xamarin.Forms.Platform.UAP/AutoSuggestStyle.xaml).

My suggested fix to avoid conflicts with the imported UWP-styles, is to inherit _AutoSuggestBox_ to a new class, _FormsAutoSuggestBox_, and then use a default style for this instead.

AFAICT _AutoSuggestBox_ is currently used in two elements: _SearchBar_  and _ShellSection_.

[...]

PR #9211, opened 2020-01-15
Fix 9052 - restore style with no extra width for Switch / ToggleSwitch [UWP]

### LINKED ISSUE #9052: [Bug] [UWP] Switch takes up more horizontal space then before
### Description
Switch seems to take up more horizontal space in XF 4.3+.

### Steps to Reproduce

Add a Switch to a StackLayout with Horizontal orientation

### Expected Behavior
Switch only takes the horizontal space it needs

### Actual Behavior
It's taking more space then needed

### Basic Information

- Version with issue:
- Last known good version: 4.2
- IDE: VS2019 16.4.2
- Platform Target Frameworks: 
  - UWP: 16299   
- Nuget Packages:
- Affected Devices:

### Screenshots
Yellow is the background color of the switch

UWP (Win 10 1903)
image

Android
image

iOS
image

### Reproduction Link
[SwitchRepro.zip](https://github.com/xamarin/Xamarin.Forms/files/4011326/SwitchRepro.zip)

WayaFlyfeather:
### Description of Change ###

By using a new dedicated style, FormsToggleSwitchStyle, the UWP SwitchRenderer now shows a switch with no extra-margin, as before 4.3.0; previously it relied on a default style, overwritten by changes in #7319; basically the cause and fix is the same as in #8743. The fact that the previous default style is hidden is the cause of #9052.

### Issues Resolved ### 

- fixes #9052

### API Changes ###
 
 None

### Platforms Affected ### 

- UWP

Switches will have no extra width, as before 4.3. There is one possible problem for UWP developers: if they have implemented their own default ToggleSwitch style, then Xamarin Switches will no longer use that; they will have to implement their own version of FormsToggleSwitchStyle in that case.

Before:
fix-9052-before

After:
fix-9052-after

### Testing Procedure ###

Go to test case 9052; Check that the yellow background of the Switch has no extra width.

### Todo(?) ###

As the previous (pre 4.3) default ToggleSwitch is now no longer default, other uses of ToggleSwitch will probably have the extra width; this may possibly cause problems. One fix would be to have them use the new FormsToggleSwitchStyle. One such possible problem (perhaps the only one), is in DataTemplate for the SwitchCell (in Resources).

Also, in Resources, is a default style for TextBox. This is most likely also hidden now, and it would probably be prudent to delete it from Resources, as it has no effect beyond being confusing. I'm not sure whether it may be a problem that it has been hidden, as the FormsTextBox already uses it's own style.

I have not at this point done any of these as part of this PR, as they had nothing to do with fixing #9052, as such.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue9052.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue9052.cs
@@ -27,12 +27,13 @@ namespace Xamarin.Forms.Controls.Issues
 
 			Label instructions = new Label
 			{
-				Text = "Check that the yellow background of the Switch below is only around the Switch and text - no extra space around."
+				Text = "Check that the yellow background of the Switch below is only around the Switch and On/Off-text - no extra space to the right."
 			};
 
 			Switch theSwitch = new Switch
 			{
-				BackgroundColor = Color.Yellow
+				BackgroundColor = Color.Yellow,
+				HorizontalOptions = LayoutOptions.Start
 			};
 
 			layout.Children.Add(instructions);
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue9052.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue9052.cs
@@ -1,0 +1,44 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Text;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System.Linq;
+using System.Threading.Tasks;
+
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+using Xamarin.Forms.Core.UITests;
+#endif
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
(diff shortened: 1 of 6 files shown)


PureWeen:
Here's a Xamarin.Forms app against XF 4.2 with the TPV set to 18632

image

Here's the same app with TPV set to 16299

image

PureWeen:
I'm thinking to fix this we'll need to just break out the entire control template and hard code

ToggleSwitchThemeMinWidth

to Zero

WayaFlyfeather:
Right - nice catch, @PureWeen.

So, to sum up: this fix brought the Switch on UWP back to the state it was in before #7319. However, that state isn't really good enough, as it is susceptible to changes in the default ControlTemplate.

So we also need to provide our own ControlTemplate, for us to be sure what it looks like, and doesn't add spacing as per the Theme.

Luckily, that's is pretty easy to do, now we have a FormsToggleSwitchStyle. In my recent commit I have added a copy of the default ControlTemplate from 16299 to the style - it doesn't use ToggleSwitchMinWidth, so is not affected by changes to that.

With TPV 16299:
fix-9052-after-16299

With TPV 18632:
fix-9052-after-18632

WayaFlyfeather:
Note, this of course also means that the notes in the PR for Behavioral/Visual changes are not complete. There will also be a change where the extra width on are removed from Switches for clients with Target Platform Version > 16299.

samhouts:
@WayaFlyfeather Are you able to address @PureWeen's comments? Thanks!

samhouts:
I'm going to close this one for now until we're ready to return to it. Thanks, @WayaFlyfeather !

PR #9218, opened 2020-01-15
[Enhancement] Shapes

### LINKED ISSUE #2452: Xamarin.Forms Drawing Spec
# Xamarin.Forms Drawing Spec

# Theming

To make MaterialShell work it not only needs a shell which appears to be material design, but all of its contents also should be material design. Today that means Custom Renderers. This is heavy and undesired from a user work perspective. Instead a better approach is required.

## Simple drawing API sample
```xaml
<Button x:Class="Local.MyButton">
  <Button.Template>
    <DrawingTemplate>
      <Drawing>
        <Grid>
          <RoundedRectangle Background="{x:Bind BackgroundColor}" />
          <Ellipse x:Name="TouchFeedback" Opacity="0" />
          <Text Content="{x:Bind Text}" />
        </Grid>
      </Drawing>
    </DrawingTemplate>
  </Button>
</Button>
```
View templates, at least for now, must be a `DrawingTemplate`. The first child of a `DrawingTemplate` is always a `Drawing`. Inside of a `Drawing` you may use Layouts and drawing primitives.

Drawing primitives may be x:named and looked up by name and manipulated in code behind as any other `View`. In effect a drawing primitive is simply a view who has no renderer but rather must be inside a `Drawing` in order to work. `Drawing` is limited to children it can support as all are collapsed into simple drawing commands.

Drawing primative objects may use a Brush type rather than a Color type for providing colors/backgrounds/etc.

When using a MaterialShell, all relevant controls will receive a default `Template` which themes them according to the Material design guidelines and unifies their look and feel across platform. This theme can be overridden at any layer of the hierarchy simply by blocking its propagation in resource dictionaries.

Once realized a Control may not have its Template changed as DrawingTemplated controls may at times require different renderers.

### DrawingTemplate
This is mostly a marker type. In the future we will eliminate the requirement that Templates be DrawingTemplates.
```csharp
public class DrawingTemplate : ControlTemplate {}
```
### Drawing
A View which has no renderer and is instead consumed by its parent view as a native drawing. Drawing is a very basic layout which can be measured but always sizes all of its children to the same size as itself when laid out.
```csharp
public class Drawing : Layout<View> {}
```
### New API for Drawing in View
```csharp
public class View : VisualElement // new API only
{
  public ControlTemplate Template { get; set; }

  protected BindableObject GetTemplateChild(string childName);

  protected virtual void OnApplyTemplate ();
}
```
### Brush
```csharp
public class Brush : BindableObject
{
  public static readonly BindableProperty OpacityProperty;
  public double Opacity { get; set; }
}
```
### SolidColorBrush
```csharp
public class SolidColorBrush : Brush
{
}
```
### GradientBrush
```csharp
public class LinearGradientBrush : Brush
{
  public static readonly BindableProperty GradientStopsProperty;
  [Content]
  public GradientStopCollection GradientStops { get; set; }
}
```
### GradientStopCollection
```csharp
public sealed class GradientStopCollection : IEnumerable<GradientStop>, IList<GradientStop>
```
### GradientStop
```csharp
public sealed class GradientStop : BindableObject
{

  public static readonly BindableProperty OffsetProperty;
  public double Offset { get; set; }

  public static readonly BindableProperty StartPointProperty;
  public Point StartPoint { get; set; }

  public static readonly BindableProperty EndPointProperty;
  public Point EndPoint { get; set; }
}
```
### Drawing Primitives
There will be a need for a large number of drawing primitives with associated properties. This document does not attempt to define all of them, just note some of the obvious ones that will need to exist. The drawing primitive API is not intended to be a wholesale replacement for SkiaSharp. It is however intended to be agnostic to the usage of Skia or native drawing backends for the paltform.

A good path forward for Skia would be to add a SkiaDrawing element that would instruct the Drawing to be rendered via Skia. Skia could then render all the stock primitives as well as provide a Skia drawing element which allows for coded drawing.

### Shape
```csharp
namespace Xamarin.Forms.Shapes 
{
  public class Shape : View
  {
    public static readonly BindableProperty FillProperty;
    public Brush Fill { get; set; }

    public static readonly BindableProperty StrokeProperty;
    public Brush Stroke { get; set; }

    public static readonly BindableProperty StrokeThicknessProperty;
    public double StrokeThickness { get; set; }
  }
}
```

### Line
```csharp
namespace Xamarin.Forms.Shapes 
{
  public sealed class Line : Shape
  {
    public Point Start { get; set; }
    public Point End { get; set; }
  }
}
```

### Ellipse
```csharp
namespace Xamarin.Forms.Shapes 
{
  public sealed class Ellipse : Shape
  {
  }
}
```

### Text
```csharp
public sealed class Text : Shape 
{
  // All the same properties as Label more or less
}

```

### Rectangle
```csharp
namespace Xamarin.Forms.Shapes 
{
  public sealed class Rectangle : Shape
  {
    public CornerRadius CornerRadius { get; set; }
  }
}
```

### BezierLine
This differs quite a bit from UWP, however this is capable of drawing an Bezier curved path/shape, as well as regular polygons and polylines.
```csharp
namespace Xamarin.Forms.Shapes 
{
  public sealed class BezierLine : Shape
  {
    public IList<BezierPoint> Points { get; }
    public bool ShouldClose { get; set; }
  }
}

```

### BezierPoint
```csharp
namespace Xamarin.Forms.Shapes 
{
  public sealed class BezierPoint : Point
  {
    public Size LeftControlOffset { get; set; }
    public Size RightControlOffset { get; set; }
  }
}

```

# TODO
- [x] Add API for drawing
- [x] Fill out brush API

# Issues

## Shapes

There is not currents a great way to draw just an arc. The mechanism in UWP for doing this is a bit... uhg just not fun.

There is also something to be said for the fact that Shapes are currently Views. This is to make sure they can be consumed by standard Layout's which is nice because the user does not need to learn new layouting mechanisms. The downside is drawing primitives only work in the context of a Drawing, but the compiler will not stop you from using them external to one. The counter case is to make Drawing specific layouts which currently seems to be the greater evil.

Ideally Layout would allow any child which implements some ILayoutable interface, of which View and and Drawing Primitives would implement. This would however be a breaking change, so at the moment it seems View is the only viable option.

## Brush

ImageBrush is probably going to be needed. Research needs to be done to ensure that every platform can support ImageBrush everywhere brushes are used.

## DrawingTemplate

There currently exists no mechanism in core to switch the renderer based on a property as this will require. That will need to be added. Ideally this should be more generic than a switch based on the Template.

### LINKED ISSUE #9178: [Spec]  Shapes
# Shapes

A **Shape** is a type of View that enables you to draw a shape to the screen.

_But, hey, why Shapes?_

Currently there are Views such as BoxView or Frame that allow us to create elements such as rectangles or ellipses but, there are many more  complex shapes. Having a set of Shapes (lines, polygons, etc.) allows a greater number of possibilities when creating attractive and rich user interfaces.

shapes

Notice that in the design there are:
- Dash Line
- Circle (Ellipse)
- Bubble (Path)
- Nested Circles (Path)

# API

Next, the Shapes API definition.

**NOTE:** This API definition is based on WPF Shapes API: https://docs.microsoft.com/en-us/dotnet/framework/wpf/graphics-multimedia/shapes-and-basic-drawing-in-wpf-overview

## Shape 

To render a **Shape**, must set the Fill property of the Shape to the Color you want. A Shape can also have a Stroke, which is a line that is drawn around the shape's perimeter. A Stroke also requires a Color that defines its appearance, and should have a non-zero value for StrokeThickness. StrokeThickness is a property that defines the perimeter's thickness around the shape edge. 

```csharp
public class Shape : View
{
    public static readonly BindableProperty FillProperty =
        BindableProperty.Create(nameof(Fill), typeof(Color), typeof(Shape), null);

    public static readonly BindableProperty StrokeProperty =
        BindableProperty.Create(nameof(Stroke), typeof(Color), typeof(Shape), null);

    public static readonly BindableProperty StrokeThicknessProperty =
        BindableProperty.Create(nameof(StrokeThickness), typeof(double), typeof(Shape), 1.0);

    public static readonly BindableProperty StrokeDashArrayProperty =
        BindableProperty.Create(nameof(StrokeDashArray), typeof(DoubleCollection), typeof(Shape), null,
            defaultValueCreator: bindable => new DoubleCollection());

    public static readonly BindableProperty StrokeDashOffsetProperty =
        BindableProperty.Create(nameof(StrokeDashOffset), typeof(double), typeof(Shape), 0.0);

    public static readonly BindableProperty AspectProperty =
        BindableProperty.Create(nameof(Aspect), typeof(Stretch), typeof(Shape), Stretch.None);

    public Color Fill
    {
        set { SetValue(FillProperty, value); }
        get { return (Color)GetValue(FillProperty); }
    }

    public Color Stroke
    {
        set { SetValue(StrokeProperty, value); }
        get { return (Color)GetValue(StrokeProperty); }
    }

    public double StrokeThickness
    {
        set { SetValue(StrokeThicknessProperty, value); }
        get { return (double)GetValue(StrokeThicknessProperty); }
    }

    public DoubleCollection StrokeDashArray
    {
        set { SetValue(StrokeDashArrayProperty, value); }
        get { return (DoubleCollection)GetValue(StrokeDashArrayProperty); }
    }

    public double StrokeDashOffset
    {
        set { SetValue(StrokeDashOffsetProperty, value); }
        get { return (double)GetValue(StrokeDashOffsetProperty); }
    }

    public Stretch Aspect
    {
        set { SetValue(AspectProperty, value); }
        get { return (Stretch)GetValue(AspectProperty); }
    }
}
```

### Ellipse

An Ellipse is a shape with a curved perimeter. To create a basic Ellipse, specify a WidthRequest, HeightRequest, and a Color for the Fill.

```csharp
public sealed class Ellipse : Shape
{

}
```
Example:

```xaml
<Ellipse Fill="SteelBlue" HeightRequest="200" WidthRequest="200" />
```

Ellipse

### Rectangle

A Rectangle is a four-sided shape with its opposite sides being equal. To create a basic Rectangle, specify a WidthRequest, a HeightRequest, and a Fill.
You can round the corners of a Rectangle. To create rounded corners, specify a value for the RadiusX and RadiusY properties. These properties specify the x-axis and y-axis of an ellipse that defines the curve of the corners. 

```csharp
public sealed class Rectangle : Shape
{
    public static readonly BindableProperty RadiusXProperty =
        BindableProperty.Create(nameof(RadiusX), typeof(double), typeof(Rectangle), 0.0);

    public static readonly BindableProperty RadiusYProperty =
        BindableProperty.Create(nameof(RadiusY), typeof(double), typeof(Rectangle), 0.0);

    public double RadiusX
    {
        set { SetValue(RadiusXProperty, value); }
        get { return (double)GetValue(RadiusXProperty); }
    }

    public double RadiusY
    {
        set { SetValue(RadiusYProperty, value); }
        get { return (double)GetValue(RadiusYProperty); }
    }
}
```

Example:

```xaml
<Rectangle Fill="Blue" Width="200" Height="100" Stroke="Black" StrokeThickness="3" RadiusX="50" RadiusY="10" />
```

Rectangle

### Polygon

A Polygon is a shape with a boundary defined by an arbitrary number of points. The boundary is created by connecting a line from one point to the next, with the last point connected to the first point. The Points property defines the collection of points that make up the boundary. 

```csharp
public sealed class Polygon : Shape
{
    public static readonly BindableProperty PointsProperty =
        BindableProperty.Create(nameof(Points), typeof(PointCollection), typeof(Polygon), null, defaultValueCreator: bindable => new PointCollection());

    public static readonly BindableProperty FillRuleProperty =
        BindableProperty.Create(nameof(FillRule), typeof(FillRule), typeof(Polygon), FillRule.EvenOdd);

    public PointCollection Points
    {
        set { SetValue(PointsProperty, value); }
        get { return (PointCollection)GetValue(PointsProperty); }
    }

    public FillRule FillRule
    {
        set { SetValue(FillRuleProperty, value); }
        get { return (FillRule)GetValue(FillRuleProperty); }
    }
}
```

The FillRule property specifies a "rule" which the composite shape uses to determine whether a given point is part of the geometry. There are two possible values for FillRule: EvenOdd and Nonzero. 

Example:

```xaml
<Polygon Fill="LightBlue" Points="10,200,60,140,130,140,180,200" />
```

Polygon

### Line

A Line is simply a line drawn between two points in coordinate space. For a Line, make sure to specify values for the Stroke and StrokeThickness properties, because otherwise the Line won't render.

```csharp
public sealed class Line : Shape
{
    public static readonly BindableProperty X1Property =
        BindableProperty.Create(nameof(X1), typeof(double), typeof(Line), 0.0);

    public static readonly BindableProperty Y1Property =
        BindableProperty.Create(nameof(Y1), typeof(double), typeof(Line), 0.0);

    public static readonly BindableProperty X2Property =
        BindableProperty.Create(nameof(X2), typeof(double), typeof(Line), 0.0);

    public static readonly BindableProperty Y2Property =
        BindableProperty.Create(nameof(Y2), typeof(double), typeof(Line), 0.0);

    public double X1
    {
        set { SetValue(X1Property, value); }
        get { return (double)GetValue(X1Property); }
    }

    public double Y1
    {
        set { SetValue(Y1Property, value); }
        get { return (double)GetValue(Y1Property); }
    }

    public double X2
    {
        set { SetValue(X2Property, value); }
        get { return (double)GetValue(X2Property); }
    }

    public double Y2
    {
        set { SetValue(Y2Property, value); }
        get { return (double)GetValue(Y2Property); }
    }
}
```
Example:

```xaml
<Line Stroke="Red" X2="400"/>
```

### Polyline

A Polyline is similar to a Polygon in that the boundary of the shape is defined by a set of points, except the last point in a Polyline is not connected to the first point.

```csharp
public sealed class Polyline : Shape
{
    public static readonly BindableProperty PointsProperty =
        BindableProperty.Create(nameof(Points), typeof(PointCollection), typeof(Polyline), null, defaultValueCreator: bindable => new PointCollection());

    public static readonly BindableProperty FillRuleProperty =
        BindableProperty.Create(nameof(FillRule), typeof(FillRule), typeof(Polyline), FillRule.EvenOdd);

    public PointCollection Points
    {
        set { SetValue(PointsProperty, value); }
        get { return (PointCollection)GetValue(PointsProperty); }
    }

    public FillRule FillRule
    {
        set { SetValue(FillRuleProperty, value); }
        get { return (FillRule)GetValue(FillRuleProperty); }
    }
}
```
Example:

```xaml
<Polyline Stroke="Black" StrokeThickness="4" Points="10,200,60,140,130,140,180,200" />
```

polyline

### Path

A Path is the most versatile Shape because you can use it to define an arbitrary geometry. But with this versatility comes complexity. Define the geometry of a path with the Data property. 

```csharp
public sealed class Path : Shape
{
    public static readonly BindableProperty DataProperty =
            BindableProperty.Create(nameof(Data), typeof(Geometry), typeof(Path), null);

    public Geometry Data
    {
        set { SetValue(DataProperty, value); }
        get { return (Geometry)GetValue(DataProperty); }
    }
}

```

#### Geometry

Where Geometry provides a base class for objects that define geometric shapes.

```csharp
public class Geometry : BindableObject
{

}
```

There will be:
- LineGeometry
- EllipseGeometry
- RectangleGeometry
- PathGeometry

**LineGeometry**

Represents the geometry of a line.

```
public class LineGeometry : Geometry
{
    public static readonly BindableProperty StartPointProperty =
        BindableProperty.Create(nameof(StartPoint), typeof(Point), typeof(LineGeometry), new Point());

    public static readonly BindableProperty EndPointProperty =
        BindableProperty.Create(nameof(EndPoint), typeof(Point), typeof(LineGeometry), new Point());

    public Point StartPoint
    {
        set { SetValue(StartPointProperty, value); }
        get { return (Point)GetValue(StartPointProperty); }
    }

    public Point EndPoint
    {
        set { SetValue(StartPointProperty, value); }
        get { return (Point)GetValue(StartPointProperty); }
    }
}
```

**EllipseGeometry**

Represents the geometry of a circle or ellipse.

```
public class EllipseGeometry : Geometry
{
    public static readonly BindableProperty CenterProperty =
        BindableProperty.Create(nameof(Center), typeof(Point), typeof(EllipseGeometry), new Point());

    public static readonly BindableProperty RadiusXProperty =
        BindableProperty.Create(nameof(RadiusX), typeof(double), typeof(EllipseGeometry), 0.0);

    public static readonly BindableProperty RadiusYProperty =
        BindableProperty.Create(nameof(RadiusY), typeof(double), typeof(EllipseGeometry), 0.0);

    public Point Center
    {
        set { SetValue(CenterProperty, value); }
        get { return (Point)GetValue(CenterProperty); }
    }

    public double RadiusX
    {
        set { SetValue(RadiusXProperty, value); }
        get { return (double)GetValue(RadiusXProperty); }
    }

    public double RadiusY
    {
        set { SetValue(RadiusYProperty, value); }
        get { return (double)GetValue(RadiusYProperty); }
    }
}
```

**RectangleGeometry**

Describes a two-dimensional rectangle.

```
public class RectangleGeometry : Geometry
{
    public static readonly BindableProperty RectangleProperty =
        BindableProperty.Create(nameof(Rectangle), typeof(Rectangle), typeof(RectangleGeometry), new Rectangle());

    public Rectangle Rectangle
    {
        set { SetValue(RectangleProperty, value); }
        get { return (Rectangle)GetValue(RectangleProperty); }
    }
}
```

**PathGeometry**

Represents a complex shape that may be composed of arcs, curves, ellipses, lines, and rectangles.

```
public sealed class PathGeometry : Geometry
{
    public PathGeometry()
    {
        Figures = new PathFigureCollection();
    }

    public static readonly BindableProperty FiguresProperty =
        BindableProperty.Create(nameof(Figures), typeof(PathFigureCollection), typeof(PathGeometry), null);

    public static readonly BindableProperty FillRuleProperty =
        BindableProperty.Create(nameof(FillRule), typeof(FillRule), typeof(PathGeometry), FillRule.EvenOdd);

    [TypeConverter(typeof(PathFigureCollectionConverter))]
    public PathFigureCollection Figures
    {
        set { SetValue(FiguresProperty, value); }
        get { return (PathFigureCollection)GetValue(FiguresProperty); }
    }

    public FillRule FillRule
    {
        set { SetValue(FillRuleProperty, value); }
        get { return (FillRule)GetValue(FillRuleProperty); }
    }
}
```

Example:

```xaml
<Path Stroke="DarkGoldenRod" StrokeThickness="3" Data="M 100,200 C 100,25 400,350 400,175 H 280" />
```

path

## Clip Views

All views (inherited from VisualElement) have the **Clip** property of type `Geometry`.

clip-image

(In the previous sample, the header background image is clipped to achieve the diagonal effect).

The following example shows an Image without a defined clip region. 

```xaml
<Image 
     Source="waterlilies.jpg" 
     WidthRequest="200"
     HeightRequest="150" />
```
image

In the next example, an identical Image is created, except that it has a defined clip region.

```xaml
<Image 
     Source="waterlilies.jpg" 
     WidthRequest="200" 
     HeightRequest="150">
     <Image.Clip>
     <EllipseGeometry
          RadiusX="100"
          RadiusY="75"
          Center="100, 75"/>
     </Image.Clip>
</Image>

```
clip-image

## Applying Styles

We can customize the appearance of Shapes using styles. We can create the styles using **XAML** or **CSS**.

Using XAML:

```
<Style x:Key="MyRectangle" Target="Rectangle">
     <Setter Property="Fill" Value="Red"  />     
     <Setter Property="Stroke" Value="Black"  />
     <Setter Property="StrokeThickness" Value="2"  />
     <Setter Property="StrokeDashArray" Value="2"  />
</Style>
```

Using CSS:
```
#MyRectangle {
  fill: red;
  stroke: black;
  stroke-width: 2px;
  stroke-dasharray: 2;
}
```
# Difficulty : Medium

# More info

This Spec is related to https://github.com/xamarin/Xamarin.Forms/issues/2452

jsuarezruiz:
### Description of Change ###

Shapes support is added:
- Ellipse
- Rectangle
- Polygon
- Line
- Polyline

### Issues Resolved ### 

- fixes #2452 (partially)
- fixes #9178

### API Changes ###

Added:

```
public class Shape 
{
    public Color Fill { get; set; }

    public Color Stroke { get; set; }

    public double StrokeThickness { get; set; }

    public DoubleCollection StrokeDashArray { get; set; }

    public double StrokeDashOffset { get; set; }

    public Stretch Aspect { get; set; }
}
```

Ellipse

```
public sealed class Ellipse : Shape
{

}
```

Rectangle
```
public sealed class Rectangle : Shape
{
    public double RadiusX { get; set; }

    public double RadiusY { get; set; }
}
```

Polygon
```
public sealed class Polygon : Shape
{
    public PointCollection Points { get; set; }

    public FillRule FillRule { get; set; }
}
```

Line
```
public sealed class Line : Shape
{
    public double X1 { get; set; }

    public double Y1 { get; set; }

    public double X2 { get; set; }

    public double Y2 { get; set; }
}
```
Polyline
```
public sealed class Polyline : Shape
{
    public PointCollection Points { get; set; }

    public FillRule FillRule { get; set; }
}
```

Also added:

- PenLineJoin
- PenLineCap

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- macOS
- UWP
- WPF

None

### Screenshots ### 
ellipse-ios
line-ios
polygon-ios
polyline-ios
rect-ios
linejoin
linecap
shapes-wpf
shapes-macOS

### Testing Procedure ###
Launch Core Gallery and navigate to the new Shapes Gallery.

### Notes

- A **Path** is the most versatile Shape because can use it to define an arbitrary geometry. But with this versatility comes complexity. Path include a big amount of auxiliary classes (Geometry, etc), etc. To prevent this PR from being very very large (and difficult to review), I am going to divide the Shapes support into two PRs. On the one hand this PR where Ellipse, Rectangle, Line, Polyline and Polygon are added and then **another PR** where Path will be added.
- We already had a class called **Rectangle**. I have called the class that define the Rectangle Shape, Rect. What is the best approach in this case?, rename the original Rectangle class to Rect?, etc.

### PR Checklist ###


programmation:
I note that the implementation creates a pile of new renderers. How about instead creating a `ShapeRenderer<T>` where `T` can provide its own drawing code? Otherwise if I want a `Rectangle` and an `Ellipse` I'm going to need 2 `View`s + their renderers, which is only going to compound the performance problems we have (especially on Android) where we're already using too many `View`s. I realise that `Path` may alleviate some of these issues, but that says to me that the other shapes should be implemented in terms of `Path` rather than leaving `Path` to last.

bcaceiro:
@jsuarezruiz Will it be possible to animate the drawing of such shapes? for example, animating drawing a circle or whatsoever. that would be huge in my opinion.

mattleibow:
Will this support gradients and images for background and borders?

jsuarezruiz:
@bcaceiro Yes, it will be possible to apply animations or transformations to Shapes.
@mattleibow The support for Brushes is work in progress here https://github.com/xamarin/Xamarin.Forms/pull/9220. After merge that PR (before this one), the Fill and Stroke properties would become a Brush.
About supporting images, do you mean something like ImageBrush: https://www.c-sharpcorner.com/uploadfile/mahesh/using-imagebrush-in-wpf/?. I have done some tests, and it would be possible to have it.

jsuarezruiz:
Added "Shapes_Experimental" flag.

charlesroddie:
I think this PR is going down the wrong path with separate renderers for each platform. It's the typical Xamarin.Forms route which maximizes bugs and discrepancies between platforms. The rendering should be done in one place, in a platform-independent way. Taking a SkiaSharp dependency would be a good way to do this.

roubachof:
@charlesroddie I strongly disagree with this point. SkiaSharp is a gigantic lib (12Mb on android), and I don't want to add this dependency weight in my apk (or ipa).

jsuarezruiz:
As many of you have seen in this PR we expose Shapes using the native APIs on each platform. This has advantages such as:
- Great possibilities.
- Good performance.
- We do not require any dependency (libraries, etc.).

To add to the Xamarin.Forms Core, although it has cons with respect some of the ideas raised here (with SkiaSharp we define the renderer once instead of one per platform), it is the option that we want to propose at the moment.

Bret Johnson has a fantastic [shape library using SkiaSharp](https://github.com/BretJohnson/XGraphics).

charlesroddie:
> SkiaSharp is a gigantic lib (12Mb on android)

It's [5MB](https://github.com/mono/SkiaSharp/issues/1081#issuecomment-569226920). Maybe you are not using aab.

roubachof:
I'm just looking at the dll size in the nuget package.

image

Xamarin.Forms.Core/FillRule.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Xamarin.Forms
+{
+    public enum FillRule
```
rmarinho, line 3: What's this?

jsuarezruiz: A value that determines how the intersecting areas are combined. 
https://docs.microsoft.com/en-us/dotnet/api/system.windows.media.pathgeometry.fillrule?view=netcore-3.1

Xamarin.Forms.Core/Shapes/Ellipse.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Xamarin.Forms.Platform;
+
+namespace Xamarin.Forms.Shapes
+{
+    [RenderWith(typeof(_EllipseRenderer))]
+    public sealed class Ellipse : Shape
+    {
+		public Ellipse()
+		{
+			Aspect = Stretch.Fill;
```
StephaneDelcroix, line 10: you're not setting a default property here, but a strongly set one, preventing styling. Use defaultValueCreator for that

Xamarin.Forms.Core/Shapes/Line.cs
```diff
@@ -0,0 +1,44 @@
...
+        public static readonly BindableProperty X1Property =
+            BindableProperty.Create(nameof(X1), typeof(double), typeof(Line), 0.0d);
+
+        public static readonly BindableProperty Y1Property =
+            BindableProperty.Create(nameof(Y1), typeof(double), typeof(Line), 0.0d);
+
+        public static readonly BindableProperty X2Property =
+            BindableProperty.Create(nameof(X2), typeof(double), typeof(Line), 0.0d);
+
+        public static readonly BindableProperty Y2Property =
```
StephaneDelcroix, line 17: it's quite inconsistent to have Line defined with 2 points, and other shapes with width and height

Xamarin.Forms.Core/Shapes/PointCollectionConverter.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Globalization;
+
+namespace Xamarin.Forms.Shapes
+{
+    public class PointCollectionConverter : TypeConverter
+    {
+        public override object ConvertFromInvariantString(string value)
+        {
+            string[] points = value.Split(new char[] { ' ', ',' });
```
StephaneDelcroix, line 10: we so far only use ',' as list separator in xaml

Xamarin.Forms.Core/Shapes/Rect.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Xamarin.Forms.Platform;
+
+namespace Xamarin.Forms.Shapes
+{
+    [RenderWith(typeof(_RectangleRenderer))]
+    public sealed class Rectangle : Shape
```
StephaneDelcroix, line 6: rename the file

Xamarin.Forms.Core/Shapes/Rect.cs
```diff
@@ -0,0 +1,31 @@
...
+			Aspect = Stretch.Fill; 
+		}
+
+        public static readonly BindableProperty RadiusXProperty =
+            BindableProperty.Create(nameof(RadiusX), typeof(double), typeof(Rectangle), 0.0d);
+
+        public static readonly BindableProperty RadiusYProperty =
+            BindableProperty.Create(nameof(RadiusY), typeof(double), typeof(Rectangle), 0.0d);
+
+        public double RadiusX
```
StephaneDelcroix, line 19: why don't have CornerRadius instead of RX, RY ?

PR #9220, opened 2020-01-15
[Enhancement] Brushes

### LINKED ISSUE #1856: [Enhancement] Add brush support for background
### Rationale

Many times there's a need to have a gradient brush, especially on a `Page`. Using images is not an option. 

### Summary

Add `Background` property to `VisualElement`:

`public Brush Background { get; set; }`

where `Brush` is a `public abstract class Brush`.

Type of brushes:

`public class SolidColorBrush: Brush`
`public class GradientBrush: Brush`
`public class LinearGradientBrush: GradientBrush`
`public class RadialGradientBrush: GradientBrush`

Nothing new, these are all taken from Windows XAML (_XAML Standard where art thou?_)

### LINKED ISSUE #2452: Xamarin.Forms Drawing Spec

### LINKED ISSUE #7293: [Spec] Brushes
# Gradients Spec

A **gradient** is the gradual blending from one color to another. 

_But, hey, why gradients?_

Mobile design trends have changed rapidly in recent years, with some things disappearing for a while and then making a gradual comeback. That’s the case with gradients. Gradients are making a comeback. 

artboard_4x-2

## GradientBrush

A gradient brush paints an area with multiple colors that blend into each other along an axis. 

```
public abstract class Brush : BindableObject
{

}
```
And the GradientBrush definition:

```
[ContentProperty(nameof(GradientStops))]
public class GradientBrush : Brush
{
    public static readonly BindableProperty GradientStopsProperty = BindableProperty.Create(
        nameof(GradientStops), typeof(IList<GradientStop>), typeof(GradientBrush), null);

    public IList<GradientStop> GradientStops
    {
        get => (IList<GradientStop>)GetValue(GradientStopsProperty);
        set => SetValue(GradientStopsProperty, value);
    }
}
```

The GradientStop is the basic building block of a gradient brush. A gradient stop specifies a Color at an Offset along the gradient axis.

```
public class GradientStop : BindableObject
{
    public static readonly BindableProperty ColorProperty = BindableProperty.Create(
        nameof(Color), typeof(Color), typeof(GradientStop), Color.White);

    public Color Color
    {
        get => (Color)GetValue(ColorProperty);
        set => SetValue(ColorProperty, value);
    }

    public static readonly BindableProperty OffsetProperty = BindableProperty.Create(
        nameof(Offset), typeof(float), typeof(GradientStop), 0f);

    public float Offset
    {
        get => (float)GetValue(OffsetProperty);
        set => SetValue(OffsetProperty, value);
    }
}

```

We will add a new property to `VisualElement` to define the background using Brushes.

```
public static readonly BindableProperty BackgroundProperty = BindableProperty.Create("Background", typeof(GradientBrush), typeof(VisualElement), default(GradientBrush));

```

This would allow gradients to be used as background in any control, although the BackgroundColor property would still exist (deprecated).

### LinearGradientBrush

A LinearGradientBrush paints an area with a gradient that's defined along a line. This line is called the gradient axis. 
You specify the gradient's colors and their locations along the gradient axis using GradientStop objects. By default, the gradient axis runs from the upper left corner to the lower right corner of the area that the brush paints, resulting in a diagonal shading.

LinearGradientBrush

```
public class LinearGradientBrush : GradientBrush
{
    public static readonly BindableProperty StartPointProperty = BindableProperty.Create(
        nameof(StartPoint), typeof(Point), typeof(LinearGradientBrush), default(Point));

    public Point StartPoint
    {
        get => (Point)GetValue(StartPointProperty);
        set => SetValue(StartPointProperty, value);
    }

    public static readonly BindableProperty EndPointProperty = BindableProperty.Create(
        nameof(EndPoint), typeof(Point), typeof(LinearGradientBrush), default(Point));

    public Point EndPoint
    {
        get => (Point)GetValue(EndPointProperty);
        set => SetValue(EndPointProperty, value);
    }
}
```

### RadialGradientBrush

A radial gradient brush paints an area with a radial gradient that has a circle, along with a focal point, to define the gradient behavior. The focal point defines the center of the gradient and has default value 0.0. 

RadialGradientBrush

```
public class RadialGradientBrush : GradientBrush
{
    public static readonly BindableProperty CenterProperty = BindableProperty.Create(
        nameof(Center), typeof(Point), typeof(RadialGradientBrush), default(Point));

    public Point Center
    {
        get => (Point)GetValue(CenterProperty);
        set => SetValue(CenterProperty, value);
    }

    public static readonly BindableProperty GradientOriginProperty = BindableProperty.Create(
        nameof(GradientOrigin), typeof(Point), typeof(RadialGradientBrush), default(Point));

    public Point GradientOrigin
    {
        get => (Point)GetValue(GradientOriginProperty);
        set => SetValue(GradientOriginProperty, value);
    }

    public static readonly BindableProperty RadiusXProperty = BindableProperty.Create(
        nameof(RadiusX), typeof(double), typeof(RadialGradientBrush), default(Point));

    public double RadiusX
    {
        get => (double)GetValue(RadiusXProperty);
        set => SetValue(RadiusXProperty, value);
    }

    public static readonly BindableProperty RadiusYProperty = BindableProperty.Create(
        nameof(RadiusY), typeof(double), typeof(RadialGradientBrush), default(Point));

    public double RadiusY
    {
        get => (double)GetValue(RadiusYProperty);
        set => SetValue(RadiusYProperty, value);
    }
}
```

To implement the Spec we can use a PaintDrawable setting a Shader (LinearGradient, etc.) in the case of Android. In iOS, can use a CALayer (DrawLinearGradient, etc.). And with UWP the GradientBrushes.

_**Note:** Nothing in this specification is guaranteed to be final; all features, implementations, and interfaces are subject to change._

# Scenarios

Let's take a look at a simple sample:

## XAML
```
<Grid>
    <Grid.Background>
        <LinearGradientBrush StartPoint="0.5,0" EndPoint="0.5,1">
            <GradientStop Color="Yellow" Offset="0.0" />
            <GradientStop Color="Red" Offset="0.25" />
            <GradientStop Color="Blue" Offset="0.75" />
            <GradientStop Color="LimeGreen" Offset="1.0" />
        </LinearGradientBrush>
    </Grid.Background>
<Grid>
``` 

## CSS

(see #7374)
```
#RootContainer{
    background: linear-gradient(45deg, rgba(218, 64, 244, 0.26) 0%, rgba(218, 64, 244, 0.26) 3%,rgba(184, 81, 207, 0.26) 3%, rgba(184, 81, 207, 0.26) 26%,rgba(149, 97, 169, 0.26) 26%, rgba(149, 97, 169, 0.26) 27%,rgba(115, 114, 132, 0.26) 27%, rgba(115, 114, 132, 0.26) 46%,rgba(80, 130, 94, 0.26) 46%, rgba(80, 130, 94, 0.26) 87%,rgba(46, 147, 57, 0.26) 87%, rgba(46, 147, 57, 0.26) 100%),linear-gradient(0deg, rgba(247, 80, 105, 0.26) 0%, rgba(247, 80, 105, 0.26) 1%,rgba(223, 84, 119, 0.26) 1%, rgba(223, 84, 119, 0.26) 11%,rgba(199, 88, 133, 0.26) 11%, rgba(199, 88, 133, 0.26) 46%,rgba(174, 91, 147, 0.26) 46%, rgba(174, 91, 147, 0.26) 54%,rgba(150, 95, 161, 0.26) 54%, rgba(150, 95, 161, 0.26) 73%,rgba(126, 99, 175, 0.26) 73%, rgba(126, 99, 175, 0.26) 100%),linear-gradient(90deg, rgb(74, 13, 231) 0%, rgb(74, 13, 231) 18%,rgb(96, 13, 230) 18%, rgb(96, 13, 230) 21%,rgb(119, 13, 229) 21%, rgb(119, 13, 229) 26%,rgb(141, 13, 228) 26%, rgb(141, 13, 228) 32%,rgb(163, 12, 226) 32%, rgb(163, 12, 226) 44%,rgb(185, 12, 225) 44%, rgb(185, 12, 225) 56%,rgb(208, 12, 224) 56%, rgb(208, 12, 224) 64%,rgb(230, 12, 223) 64%, rgb(230, 12, 223) 100%)
}
```

# Backward Compatibility

`BackgroundColor` becomes deprecated (it can still be used to avoid breaking third parties) although its replacement will be to use `Background` and a `SolidColorBrush`.

Before
`<Grid BackgroundColor="Red" />`

After
`<Grid Background="Red" />`

# Scope

The new `Background` property as well as `BackgroundColor` will be at the same level, in all the views where it is currently possible to use BackgroundColor it will be possible to use Background.
Other options such as BorderColors, etc. are not included in the Spec, etc. (depending on the interest it could come later).

## Difficulty : Medium

jsuarezruiz:
### Description of Change ###

Added Brushes support. A Brush "paints" an area with its output. Different brushes have different types of output. Some brushes paint an area with a solid color and others with a gradient.

gradients2

Examples:

Using the new **Background** property you can set the background of any Page, Layout or View using a Brush.

```
<Grid>
    <Grid.Background>
        <LinearGradientBrush StartPoint="0.5,0" EndPoint="0.5,1">
            <GradientStop Color="Yellow" Offset="0.0" />
            <GradientStop Color="Red" Offset="0.25" />
            <GradientStop Color="Blue" Offset="0.75" />
            <GradientStop Color="LimeGreen" Offset="1.0" />
        </LinearGradientBrush>
    </Grid.Background>
<Grid>
```

### Issues Resolved ### 

- fixes  (partially) #2452
- fixes #1856
- fixes #7293

### API Changes ###

Added:

```
public abstract class Brush : BindableObject
{

}
```

GradientBrush
```
public class GradientBrush : Brush
{
    public IList<GradientStop> GradientStops { get; set; }
}
```

GradientStop
```
public class GradientStop : BindableObject
{

    public float Offset { get; set; }
}
```

SolidColorBrush
```
public class SolidColorBrush : GradientBrush
{

}
```

LinearGradientBrush
```
public class LinearGradientBrush : GradientBrush
{
    public Point StartPoint { get; set; }

    public Point EndPoint { get; set; }
}
```

RadialGradientBrush
```
public class RadialGradientBrush : GradientBrush
{
    public Point Center { get; set; }

    public double Radius { get; set; }
}
```

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- macOS
- UWP
- WPF

None

### Testing Procedure ###

Launch Core Gallery and navigate to the new Gradients Gallery.

### Screenshots
solidcolorbrush
lineargradientbrush-h
lineargradientbrush-v

macos-brushes

css-gradient2

### Notes
- **BackgroundColor** property will be deprecated.

### PR Checklist ###


jcmanke:
> What should happen with the BackgroundColor property?. In the current state of the PR can use BackgroundColor (Color) and Background (Brush) properties at the same time. If both properties are used, BackgroundColor is ignored, but should we deprecate BackgroundColor?

Basically, I don't want to ever create a SolidColorBrush object in either XAML or code, I just want to specify the color. There should be an implicit converter so that `Background = Color.Blue` works. (If this is already included, great. I haven't looked through the files yet.)

jsuarezruiz:
@jcmanke  This PR is still a Draft and we continue working on it so, thanks for your feedback. A TypeConverter was added to allow `Background="Red"`.

MichaelStgt:
I had a closer look at the source and the sample inside the Gallery.
I assume that x = 0 and y = 0 is the upper left corner of the Element.

Your example contains two for a LinearGradientBrush, where one is
```
**HorizontalLinearGradient**
StartPoint="0, 0"
EndPoint="1, 0"
```
and the other one is
```
**VerticalLinearGradient**
StartPoint="0, 0"
EndPoint="0, 1"
```
**But should it not be like this:**
```
**FromTopLeftToBottomRight (diagonal)**
StartPoint="0, 0"
EndPoint="1, 1"
```

```
**HorizontalLinearGradient**
StartPoint="0.5, 0"
EndPoint="0.5, 1"
```

```
**VerticalLinearGradient**
StartPoint="0, 0.5"
EndPoint="1, 0.5"
```
```
**FromTopRightToBottomLeft (diagonal)**
StartPoint="1, 0"
EndPoint="0, 1"
```

jsuarezruiz:
@MichaelStgt Good question. I have created more samples to test and cover all these cases. Let me add a screenshot where you see each most common case using LinearGradientBrush.

lineargradientpoints

Xamarin.Forms.Platform.UAP/RadialGradientBrush.cs
```diff
@@ -0,0 +1,373 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
michael-hawker, line 1: Would need to call out somewhere in a third party notice that you're using this code from the .NET Foundation/Windows Community Toolkit.

Though, you should probably just use the new WinUI 2.4 version as it also supports animations.

samhouts: Do we need to call it out, since we're also part of the .NET Foundation?

Xamarin.Forms.Core/VisualElement.cs
```diff
@@ -152,6 +152,46 @@ static void OnTransformChanged(BindableObject bindable, object oldValue, object
...
+				if (oldvalue != null)
+					(bindable as VisualElement)?.StopNotifyingBackgroundChanges();
+			},
+			propertyChanged: (bindable, oldvalue, newvalue) =>
+			{
+				if (newvalue != null)
+					(bindable as VisualElement)?.NotifyBackgroundChanges();
+			});
+
+		void NotifyBackgroundChanges()
```
PureWeen, line 167: Technically we need this everywhere we use brushes right? @StephaneDelcroix do you have any thoughts or any existing BO magic that we can use here so we don't have to do this every time there's a brush to use? 

Or should this be pushed down to the platforms to watch for brush changes?

Xamarin.Forms.Platform.Android/Extensions/BrushExtensions.cs
```diff
@@ -0,0 +1,236 @@
...
+
+			view.SetPaintGradient(brush);
+		}
+
+		public static void UpdateBackground(this Paint paint, Brush brush, int height, int width)
+		{
+			if (paint == null || brush == null || brush.IsEmpty)
+				return;
+
+			if (brush is SolidColorBrush solidColorBrush)
```
PureWeen, line 50: We don't have to do this now but is there a way we can generalize this out? Instead of using if statements? Maybe we need to create a brush handler like we have with ImageSource handler?

Xamarin.Forms.Platform.Android/GradientStrokeDrawable.cs
```diff
@@ -0,0 +1,179 @@
...
+		{
+			base.OnDraw(shape, canvas, paint);
+
+			if (_backgroundColor != null)
+				paint.Color = _backgroundColor;
+
+			shape.Draw(canvas, _strokePaint);
+		}
+
+		public abstract class IGradientShader
```
PureWeen, line 83: Is this supposed to be `interface` ? Or should it just be called a `GradientShared`?

Xamarin.Forms.Platform.MacOS/Renderers/FrameRenderer.cs
```diff
@@ -19,12 +19,49 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
 			base.OnElementPropertyChanged(sender, e);
 
 			if (e.PropertyName == VisualElement.BackgroundColorProperty.PropertyName ||
-			    e.PropertyName == Xamarin.Forms.Frame.BorderColorProperty.PropertyName ||
+				e.PropertyName == VisualElement.BackgroundProperty.PropertyName ||
```
PureWeen, line 22: Replace this with the IsAny extension?

Xamarin.Forms.Platform.UAP/RadialGradientBrush.cs
jsuarezruiz: I think the best way to have RadialGradientBrush is update WinUI to 2.4 https://github.com/xamarin/Xamarin.Forms/pull/10976
Thanks for the feedback @michael-hawker

Xamarin.Forms.Core/Brush.cs
```diff
@@ -0,0 +1,184 @@
...
+				{
+					var radialGradientBrush = (RadialGradientBrush)this;
+					return radialGradientBrush == null || radialGradientBrush.GradientStops.Count == 0;
+				}
+
+				return true;
+			}
+		}
+
+		public static readonly SolidColorBrush AliceBlue = new SolidColorBrush(Color.AliceBlue);
```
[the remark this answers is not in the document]
jsuarezruiz: Only if we want to allow creating a SolidColorBrush like this:
`SolidColorBrush brush = Brush.Red;`

Instead of:
`SolidColorBrush brush = new SolidColorBrush(Color.Red);`

Xamarin.Forms.Core/Brush.cs
```diff
@@ -0,0 +1,184 @@
...
+{
+	public abstract class Brush : BindableObject
+	{
+		public static Brush Default
+		{
+			get { return new SolidColorBrush(Color.Default); }
+		}
+
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public bool IsEmpty
```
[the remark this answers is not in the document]
jsuarezruiz: I have changed it. Now it is abstract and we implemented it in every brush. It is no longer hidden.

Xamarin.Forms.Platform.Android/GradientStrokeDrawable.cs
jsuarezruiz: Changed to `GradientShader`

Xamarin.Forms.Platform.Android/Extensions/BrushExtensions.cs
jsuarezruiz: We could create handlers in the same way as we have with the images.

Xamarin.Forms.Platform.MacOS/Renderers/FrameRenderer.cs
jsuarezruiz: Ok

Xamarin.Forms.Core/GradientStopCollection.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Collections.ObjectModel;
+
+namespace Xamarin.Forms
+{
+    public sealed class GradientStopCollection : ObservableCollection<GradientStop>
```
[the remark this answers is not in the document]
jsuarezruiz: Mnn, yes, to detect changes adding or removing `GradientStops` and update the Brush.

jsuarezruiz:
Thanks for all the feedback @PureWeen. Some fixes:
- Brush in NavBar on iOS 10.
- Fix RadialGradientBrush using right position with CSS styles.
- Small fix with bindable Brush.

fix-ios-brush

Xamarin.Forms.Core/Brush.cs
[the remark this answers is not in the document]
PureWeen: 👍

Xamarin.Forms.Core/GradientStopCollection.cs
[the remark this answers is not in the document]
PureWeen: Do we do that anywhere? I looked around for this is being utilized and couldn't find it.

Xamarin.Forms.Platform.Android/Extensions/BrushExtensions.cs
PureWeen: Yea, maybe create a new issue/spec to cover this down the road?

Xamarin.Forms.Platform.UAP/RadialGradientBrush.cs
jsuarezruiz: Using WinUI 2.4. 
winui2 4-radial

jsuarezruiz:
Added brushes support to Shapes:

 shapes-brushes
shapes-brushes-wpf
shapes-brushes-uwp

jsuarezruiz:
Added brushes support to Visual Renderers:
visual-brushes

Xamarin.Forms.Platform.Android/Renderers/FrameRenderer.cs
```diff
@@ -183,10 +183,18 @@ void DrawBackground(ACanvas canvas, int width, int height, float cornerRadius, b
 					float ry = _convertToPixels(cornerRadius);
 					path.AddRoundRect(rect, rx, ry, direction);
 
-					global::Android.Graphics.Color color = _frame.BackgroundColor.ToAndroid();
-
 					paint.SetStyle(style);
-					paint.Color = color;
+
+					if (_frame.Background != null && !_frame.Background.IsEmpty)
```
PureWeen, line 188: Can this be moved to an extension method? I see this check a few places

Xamarin.Forms.Platform.UAP/BackgroundTracker.cs
```diff
@@ -21,7 +21,9 @@ public BackgroundTracker(DependencyProperty backgroundProperty)
 
 		protected override void OnPropertyChanged(object sender, PropertyChangedEventArgs e)
 		{
-			if (e.PropertyName == VisualElement.BackgroundColorProperty.PropertyName || e.PropertyName == Page.BackgroundImageSourceProperty.PropertyName)
+			if (e.PropertyName == VisualElement.BackgroundColorProperty.PropertyName
```
PureWeen, line 24: IsOneOf

Xamarin.Forms.Platform.UAP/BoxViewRenderer.cs
```diff
@@ -23,7 +23,8 @@ protected override void OnElementChanged(ElementChangedEventArgs<BoxView> e)
 					};
 
 					rect.SetBinding(WShape.FillProperty, new Windows.UI.Xaml.Data.Binding { Converter = new ColorConverter(), Path = new PropertyPath("Color") });
-	
+					rect.SetBinding(WShape.FillProperty, new Windows.UI.Xaml.Data.Binding { Converter = new BrushConverter(), ConverterParameter = e.NewElement.Color, Path = new PropertyPath("Background") });
```
PureWeen, line 26: Is this code correct? It seems like this will overwrite the previous binding?

Xamarin.Forms.Platform.UAP/DatePickerRenderer.cs
```diff
@@ -104,6 +106,8 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
 				UpdateMaximumDate();
 			else if (e.PropertyName == DatePicker.MinimumDateProperty.PropertyName)
 				UpdateMinimumDate();
+			else if (e.PropertyName == VisualElement.BackgroundProperty.PropertyName)
```
PureWeen, line 109: Isn't this call already taken care of by the base class?

Xamarin.Forms.Platform.UAP/LayoutRenderer.cs
```diff
@@ -39,10 +40,18 @@ protected override void UpdateBackgroundColor()
...
 			{
 				// Forces the layout to take up actual space if it's otherwise empty
-				Background = new SolidColorBrush(Colors.Transparent);
+				Background = new WSolidColorBrush(Colors.Transparent);
 			}
 		}
 
+		protected override void UpdateBackground()
+		{
+			base.UpdateBackgroundColor();
```
PureWeen, line 49: I'm guessing this is the wrong base call?

Xamarin.Forms.Platform.UAP/MasterDetailControlStyle.xaml
```diff
@@ -56,7 +56,7 @@
                                         </uwp:FormsCommandBar.Resources>
                                         <uwp:FormsCommandBar.Content>
 											<Border x:Name="TitleArea" Height="{ThemeResource TitleBarHeight}" Visibility="{TemplateBinding DetailTitleVisibility}" HorizontalAlignment="Stretch">
-												<Grid x:Name="TitleViewPresenter" VerticalAlignment="Center" Background="{TemplateBinding ToolbarBackground}" HorizontalAlignment="Stretch">
+												<Grid x:Name="TitleViewPresenter" VerticalAlignment="Center" HorizontalAlignment="Stretch">
```
PureWeen, line 59: Did you mean to remove this binding?

Xamarin.Forms.Platform.iOS/Renderers/FrameRenderer.cs
```diff
@@ -87,6 +89,20 @@ public virtual void SetupLayer()
 				_actualView.Layer.BackgroundColor = Element.BackgroundColor.ToCGColor();
 			}
 
+			_actualView.Layer.RemoveBackgroundLayer();
```
PureWeen, line 92: Is this safe if the user doesn't set the Background at all? Is it basically a noop?

Xamarin.Forms.Platform.iOS/Renderers/NavigationRenderer.cs
```diff
@@ -676,6 +677,10 @@ void UpdateBarBackgroundColor()
 					navigationBarAppearance.BackgroundColor = barBackgroundColor.ToUIColor();
 				}
 
+				var barBackgroundBrush = NavPage.BarBackground;
```
PureWeen, line 680: Does this code need to be changed so it only executes if the background brush is actually set?

Xamarin.Forms.Platform.iOS/VisualElementRenderer.cs
```diff
@@ -316,13 +320,20 @@ public override SizeF SizeThatFits(SizeF size)
...
 			base.LayoutSubviews();
+
 			if (_blur != null && Superview != null)
 			{
 				_blur.Frame = Bounds;
 				if (_blur.Superview == null)
 					Superview.Add(_blur);
 			}
+
+			bool hasBackground = Element.Background != null && !Element.Background.IsEmpty;
```
PureWeen, line 331: Extension method for hasBackground check

Xamarin.Forms.Material.Android/MaterialButtonRenderer.cs
```diff
@@ -215,7 +215,7 @@ protected virtual void OnElementPropertyChanged(object sender, PropertyChangedEv
 				UpdateBorder();
 			else if (e.PropertyName == Button.FontProperty.PropertyName)
 				UpdateFont();
-			else if (e.PropertyName == Button.TextColorProperty.PropertyName || e.PropertyName == VisualElement.BackgroundColorProperty.PropertyName)
+			else if (e.PropertyName == Button.TextColorProperty.PropertyName || e.PropertyName == VisualElement.BackgroundColorProperty.PropertyName || e.PropertyName == VisualElement.BackgroundProperty.PropertyName)
```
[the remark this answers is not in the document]
jsuarezruiz: Changed.

Xamarin.Forms.Platform.UAP/MasterDetailControlStyle.xaml
jsuarezruiz: Seems like a merge mistake.

Xamarin.Forms.Platform.iOS/Renderers/FrameRenderer.cs
jsuarezruiz: The `RemoveBackgroundLayer` method internally check if there is a specific Layer, otherwise it will do nothing.

Xamarin.Forms.Platform.UAP/BackgroundTracker.cs
jsuarezruiz: Changed.

Xamarin.Forms.Platform.UAP/DatePickerRenderer.cs
jsuarezruiz: Right. This is unnecessary. I reviewed and removed some places updating the brush.

Xamarin.Forms.Platform.UAP/BoxViewRenderer.cs
jsuarezruiz: Not at all, with the second binding the first one is no longer neccesary.

jsuarezruiz:
@PureWeen Reviewed Shell FlyoutBackground:
flyoutbackground

PR #8858, opened 2019-12-11
[android] remove Anticipator.cs for now

jonathanpeppers:
### Description of Change ###

Context: https://github.com/jonathanpeppers/anticipator
Context: https://github.com/xamarin/Xamarin.Forms/pull/8245

I was profiling the Blank Xamarin.Forms app template and saw this:

    Method call summary
    Total(ms) Self(ms)      Calls Method name
       40095        0         11 (wrapper runtime-invoke) <Module>:runtime_invoke_void__this___object (object,intptr,intptr,intptr)
       40094        0          8 System.Threading.ThreadHelper:ThreadStart (object)
       40094        0          8 System.Threading.ExecutionContext:Run (System.Threading.ExecutionContext,System.Threading.ContextCallback,object)
       40094        0          8 System.Threading.ExecutionContext:Run (System.Threading.ExecutionContext,System.Threading.ContextCallback,object,bool)
       40094        0          8 System.Threading.ExecutionContext:RunInternal (System.Threading.ExecutionContext,System.Threading.ContextCallback,object,bool)
       40092        0          8 System.Threading.ThreadHelper:ThreadStart_Context (object)
       40092        0          8 Xamarin.Forms.Platform.Android.Anticipator:Loop (object)
       40035        0         16 System.Threading.WaitHandle:WaitOne (System.TimeSpan)
       40035        0         16 System.Threading.WaitHandle:WaitOne (System.TimeSpan,bool)
       40035        0         16 System.Threading.WaitHandle:WaitOne (long,bool)
       40035        0         16 System.Threading.WaitHandle:InternalWaitOne (System.Runtime.InteropServices.SafeHandle,long,bool,bool)
       40034        0         16 System.Threading.WaitHandle:WaitOneNative (System.Runtime.InteropServices.SafeHandle,uint,bool,bool)
       40034    40034         16 (wrapper managed-to-native) System.Threading.WaitHandle:Wait_internal (intptr*,int,bool,int)

The `Anticipator` class stuck out, so I profiled it separately in a
blank app (not using Xamarin.Forms at all). This project isolates the
`Anticipator`, in timing specifically `Build.VERSION.SdkInt` and the
`Resource.designer.cs` class. This launches an app 10 times and
averages the result.

The results were that it hurt performance in every scenario:

    HAXM Emulator / Debug
        Anticipator: 141.273
        No Anticipator: 108.182
    HAXM Emulator / Release
        Anticipator: 86.182
        No Anticipator: 63.545
[...]

PR #8867, opened 2019-12-12
[android] added a script to profile Android startup

jonathanpeppers:
### Description of Change ###

To make a useful way to profile startup of Xamarin.Forms, we need two
things:

1) A "default" empty Xamarin.Forms app.

2) A script that works cross-platform to launch the app several times
   and display an average.

I created a "HelloForms" project using the "Blank Forms App" template
in Visual Studio 2019 16.4. The only other changes I made were to
wire-up Xamarin.Forms in-tree and removed the `<PackageReference/>`. I
may have removed some extraneous junk or comments, too.

I created a script at `tools\scripts\profile-android.ps1`, which will
display rich help text via:

    .\tools\scripts\profile-android.ps1 -?

By default it will launch `HelloForms.Android.csproj` 10 times. It
will parse `adb logcat` output and average the result.

For example:

    PS> profile-android.ps1 -iterations 3
    Launching: com.xamarin.forms.helloforms
    Launching: com.xamarin.forms.helloforms
    Launching: com.xamarin.forms.helloforms
    12-12 09:13:15.593  1876  1898 I ActivityManager: Displayed com.xamarin.forms.helloforms/crc6450e568c951913723.MainActivity: +1s501ms
    12-12 09:13:19.696  1876  1898 I ActivityManager: Displayed com.xamarin.forms.helloforms/crc6450e568c951913723.MainActivity: +1s475ms
    12-12 09:13:23.863  1876  1898 I ActivityManager: Displayed com.xamarin.forms.helloforms/crc6450e568c951913723.MainActivity: +1s564ms
    Average(ms): 1513.33333333333

The script can be used against other projects as well, by using the
`-project` parameter.

Going forward, this script will be useful for verifying any
performance improvements. I hope to send at least a few!

### Issues Resolved ### 

None

### API Changes ###
 
None

### Platforms Affected ### 

- Android (really just adding a script, though)

### Behavioral/Visual Changes ###

None

### Before/After Screenshots ### 

Not applicable

### Testing Procedure ###

Try out `tools/scripts/profile-android.ps1`

### PR Checklist ###

- [ ] Targets the correct branch
- [ ] Tests are passing (or failures are unrelated)

PR #9226, opened 2020-01-16
[android] remove the Anticipator for 4.5.0 and master

jonathanpeppers:
### Description of Change ###

Context: https://github.com/xamarin/Xamarin.Forms/pull/8858

After merging from 4.4.0, etc. this class came back.

@kingces95 did not want to remove the `Anticipator.cs` file. Since
`Xamarin.Forms.Platform.Android.csproj` is now a short-form MSBuild
project, you have to add an explicit:

`<Compile Remove="Anticipator.cs"/>` to remove it.

### Issues Resolved ### 

None

### API Changes ###

None

### Platforms Affected ### 

- Android

None

Not applicable

### Testing Procedure ###

CI should be OK

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Forms.cs
+++ b/Xamarin.Forms.Platform.Android/Forms.cs
@@ -53,6 +53,7 @@ namespace Xamarin.Forms
 
 		const int TabletCrossover = 600;
 
+		static BuildVersionCodes? s_sdkInt;
 		static bool? s_isLollipopOrNewer;
 		static bool? s_is29OrNewer;
 		static bool? s_isMarshmallowOrNewer;
@@ -72,7 +73,13 @@ namespace Xamarin.Forms
 		static Color _ColorButtonNormal = Color.Default;
 		public static Color ColorButtonNormalOverride { get; set; }
 
-		internal static BuildVersionCodes SdkInt => Anticipator.SdkInt;
+		internal static BuildVersionCodes SdkInt {
+			get {
+				if (!s_sdkInt.HasValue)
+					s_sdkInt = Build.VERSION.SdkInt;
+				return (BuildVersionCodes)s_sdkInt;
+			}
+		}
 
 		internal static bool Is29OrNewer
 		{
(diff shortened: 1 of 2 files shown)


jonathanpeppers:
I'm also OK to delete `Anticipator.cs` here as another option.

samhouts:
I'm not sure why it's important to remove the class. Nothing is calling it, right?

Xamarin.Forms.Platform.Android/Forms.cs
```diff
@@ -72,7 +73,13 @@ public static class Forms
 		static Color _ColorButtonNormal = Color.Default;
 		public static Color ColorButtonNormalOverride { get; set; }
 
-		internal static BuildVersionCodes SdkInt => Anticipator.SdkInt;
```
jonathanpeppers: @samhouts this line is calling it.

The reason to remove `Anticipator` is just that it hurts performance, when the goal of the `Anticipator` was to improve it, I think?

I think the concept is somewhat flawed, as starting new threads to run [Android.Build.VERSION.SdkInt](https://github.com/xamarin/Xamarin.Forms/blob/df5dc1c232d2447e48a26a1c98902a26c151cd6b/Xamarin.Forms.Platform.Android/Anticipator.cs#L39) is not going to speed anything up. There is also the problem where Android puts all but one core asleep when on battery power. Your code will have to wait for Android to "wake up" the other cores. I wouldn't recommend running any threads on startup, unless there is a network request.

This isn't as much of an issue on iOS starting threads, Android is just janky.

samhouts: Gotcha, I missed the SdkInt part. 

This was just a POC for the Anticipator to make sure it didn't break anything in the wild. It's not expected to improve performance until the rest of the implementation is complete.

It's fine to remove all references until then.

Thanks!

jonathanpeppers: @samhouts if you have a PR that brings `Anticipator` back, let me know so I can review and time it, thanks!

I also sent a PR for the script I've been using, as I thought it would help the team be able to measure perf improvements on Android: https://github.com/xamarin/Xamarin.Forms/pull/8867

PR #9237, opened 2020-01-16
Fix RTL on Shell to correctly propagate FlowDirection through the UIView Stack

### LINKED ISSUE #6227: [Bug] Xamarin.Forms Shell Opener Issue for "RTL Language" when Tap at Hamburger

### LINKED ISSUE #6342: [Bug] Shell FlowDirection follows Device FlowDirection in all cases.

PureWeen:
### Description of Change ###

All these changes were applied to both iPhone and iPad based implementations

- Calculate the flyout positioning based on RTL settings
- propagate RTL to search handler
- propagate RTL to ContentPage
- propagate RTL to Navigation Bar
- Modify EffectiveFLowDirection so that it's set during FlowDirectionChanging not FlowDirectionChanged
- Don't influence the RTL setting set by the OS if the user hasn't specified an RTL value

### Issues Resolved ### 

- fixes #6227 
- fixes #6342

### Breaking Changes ###
Previously Xamarin.Forms always forced LTR regardless of the OS setting but with this PR if the user has set the device language to something that's RTL then it will use RTL as the default.

### Platforms Affected ### 
- iOS

#### RTL Store Shell

image

image

image
image

### Testing Procedure ###
- Go into Store shell and play with the Flow Direction button that has been added

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/VisualElement.cs
+++ b/Xamarin.Forms.Core/VisualElement.cs
@@ -187,7 +187,7 @@ namespace Xamarin.Forms
 
 		public static readonly BindableProperty IsFocusedProperty = IsFocusedPropertyKey.BindableProperty;
 
-		public static readonly BindableProperty FlowDirectionProperty = BindableProperty.Create(nameof(FlowDirection), typeof(FlowDirection), typeof(VisualElement), FlowDirection.MatchParent, propertyChanged: FlowDirectionChanged);
+		public static readonly BindableProperty FlowDirectionProperty = BindableProperty.Create(nameof(FlowDirection), typeof(FlowDirection), typeof(VisualElement), FlowDirection.MatchParent, propertyChanging:FlowDirectionChanging, propertyChanged: FlowDirectionChanged);
 
 		public static readonly BindableProperty TabIndexProperty =
 			BindableProperty.Create(nameof(TabIndex),
@@ -218,7 +218,6 @@ namespace Xamarin.Forms
 			((VisualElement)bindable).TabStopDefaultValueCreator();
 
 		IFlowDirectionController FlowController => this;
-		IPropertyPropagationController PropertyPropagationController => this;
 
 		public FlowDirection FlowDirection
 		{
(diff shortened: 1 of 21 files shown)


Xamarin.Forms.Core/VisualElement.cs
```diff
@@ -929,18 +936,24 @@ static void OnVisualChanged(BindableObject bindable, object oldValue, object new
 			(self as IPropertyPropagationController)?.PropagatePropertyChanged(VisualElement.VisualProperty.PropertyName);
 		}
 
-		static void FlowDirectionChanged(BindableObject bindable, object oldValue, object newValue)
+		static void FlowDirectionChanging(BindableObject bindable, object oldValue, object newValue)
```
PureWeen: @samhouts I moved the EffectiveFlowDirection calculations here so that once the FlowDirection propertyChange fires the EffectiveFlowDirection has already been calculated

jsuarezruiz:
I have tested on iPhone and iPad and it works great. There are some UITests failing but seems unrelated (Example: EntryScrollTest).

PR #9276, opened 2020-01-22
Fix 7257 [UWP] Make Translate in combination with Scale work same as other platforms

### LINKED ISSUE #7257: [Bug] Scale/Translate works in different order between UWP and Android
### Description

When using the properties ScaleX/ScaleY and TranslationX/TranslationY on a view (in this case AbsoluteLayout) the position of the object is different between UWP and Android.

### Steps to Reproduce

With an AbsoluteLayout as the current object being drawn on, run the following code :

Xamarin.Forms.AbsoluteLayout AL = new Xamarin.Forms.AbsoluteLayout () ;
AL.BackgroundColor = Xamarin.Forms.Color.Green ;
Xamarin.Forms.AbsoluteLayout.SetLayoutBounds (AL, new Xamarin.Forms.Rectangle (0, 0, 100, 100));
Xamarin.Forms.AbsoluteLayout.SetLayoutFlags (AL, Xamarin.Forms.AbsoluteLayoutFlags.None);
AL.AnchorX = 0 ;
AL.AnchorY = 0 ;
AL.ScaleX = 0.5 ;
AL.ScaleY = 0.5 ;
AL.TranslationX = 100 ;
AL.TranslationY = 100 ;
Children.Add ( AL ) ;

### Expected Behavior

The expected behaviour is that the position of the green square should be the same on Android and on UWP (iOS not tried).
From the documentation it is not possible to tell which is expected, but they should both give the same result.

### Actual Behavior

On UWP: the green square of side 50 is at (x,y) = (200,200).
On Android: the green square of side 50 is at (x,y) = (100, 100).

### Basic Information

On Android the translate is done with non scaled coordinates, on UWP you have to multiply the translation by the scale to get the same result.

- Version with issue:
Xamarin.Forms v4.2.0.709249

- Last known good version:
- IDE: Visual Studio 2019
- Platform Target Frameworks: 
  - Android:  8.1
  - UWP:   16299
- Android Support Library Version:  v28.0.0.1
- Nuget Packages: standard Xamarin
- Affected Devices:

WayaFlyfeather:
### Description of Change ###

Changes the way translate-transforms are applied on FrameworkElements in UWP, when the element is also scaled.

Without this fix, the translate-transform is adjusted so as to neutralize a scaling of the transform that doesn't happen. (Probably because in 3D-projection that apparently _does_ happen, and an [adjustment is necessary](https://github.com/xamarin/Xamarin.Forms/blob/e962467bdd6ce2b612d3e5c97e0992e94bd3dd6c/Xamarin.Forms.Platform.UAP/VisualElementTracker.cs#L553-L555).)

This causes 2D scaling in combination with translating to behave differently on UWP compared to Android/iOS.

With this fix, the translation is not adjusted for scaling in 2D (as there is no need), and therefore behaves the same as on the other platforms.

### Issues Resolved ### 

- fixes #7257

### API Changes ###
 
 None

### Platforms Affected ### 

- UWP

2D translate in combination with scaling will behave like it does on Andorid/iOS. Without this fix, if something is scaled to half size, the translation will be doubled.

On Android, for comparison:
Android Before 7257

On UWP, before fix:
UWP 7257 Before

On UWP, after fix:
UWP 7257 After

### Testing Procedure ###

In UWP to Test Case 7257 in the Control Gallery. Check that the green and blue squares are vertically aligned with the red bar to the left, and that they are the same size. (The green square is the one with a 2D transform. The blue square is rotated an imperceptible amount in 3D, and therefore uses FrameworkElement.Projection instead of FrameworkElement.RenderTransform.)

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7257.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7257.cs
@@ -1,0 +1,82 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Text;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System.Linq;
+using System.Threading.Tasks;
+
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+using Xamarin.Forms.Core.UITests;
+#endif
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 7257, "[Bug] Scale/Translate works in different order between UWP and Android", PlatformAffected.UWP)]
+#if UITEST
+	[NUnit.Framework.Category(UITestCategories.SearchBar)]
+#endif
+	public class Issue7257 : TestContentPage
+	{
+		protected override void Init()
+		{
+			Title = "Issue 7257";
+
+			StackLayout contentSL = new StackLayout();
+
+			Label instructions = new Label
+			{
+				Text = "Look at the layout below. The green and the blue squares should both be at the same vertical position as the red bar to the left, and should be same size. "
+					 + "If the green square is below the blue square, it means there is a problem.",
+				Margin = 10
+			};
+
+			contentSL.Children.Add(instructions);
+
+			AbsoluteLayout testAL = new AbsoluteLayout();
+
+			// Reference red rectangle on left edge 100 from top, 50 high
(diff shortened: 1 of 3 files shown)


nickrandolph:
The changes proposed look good to me.

@WayaFlyfeather are you able to fix the build error - It looks like a missing file perhaps? 
@samhouts what else needs to happen with the PR for it to be good to merge?

pictos:
I think the #11933 fix this issue. @mattleibow can you confirm that?

mattleibow:
@pictos that is correct.

pictos:
cc: @samhouts

rmarinho:
Closing since #11933  fixes the other versions.

PR #9294, opened 2020-01-23
[HR] complement the sourceinfo with assemblyname

StephaneDelcroix:
### Description of Change ###

make sure we can trace the relative uri back to the originator

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###


GitHubUser30:
I would recommend adding a helper method to centralize URI formatting rather than repeating it in 6 places. Also why not follow WPF format, at least it relative form, e.g.
ABSOLUTE: pack://application:,,,/MyApp;component/Controls/MyGrid.xaml
RELATIVE:   MyApp;component/Controls/MyGrid.xaml

StephaneDelcroix:
thanks for the feedback, @GitHubUser30. as we're still finalizing that API, we prefer keeping the repetition. as for pack url, we pondered using them at other places in XF, and agree, with our community, not to.

PR #9332, opened 2020-01-27
[Core] Allow CarouselView to specify IndicatorView 

### LINKED ISSUE #8992: [Bug] IndicatorView - ItemsSourceBy not showing up in Intellisense
### Description
For IndicatorView control, the **ItemsSourceBy** property is not shown up by Intellisense

image

Also, XAML Hot Reload does not work with IndicatorView. When you change something in XAML, the IndicatorView is disappearing because probably XAML Hot Reload thinks that ItemsSourceBy property does not exist.

### Steps to Reproduce

1. Try to create IndicatorView.
2. Try to add ItemsSourceBy property to the IndicatorView
3. As you can see the Intellisense is not working for this property.

### Expected Behavior
Intellisense should show ItemsSourceBy property for IndicatorView. And XAML Hot Reload should also work for this property.

### Actual Behavior
Intellisense does not show ItemsSourceBy property for IndicatorView and XAML Hot Reload is not working with this property.

### Basic Information

- Version with issue: 4.4.0.991265
- IDE: Visual Studio 2019 16.4.2
- Platform Target Frameworks: 
  - Android: 9.0 (Pie)

rmarinho:
### Description of Change ###

CarouselView should provide a way to link to a IndicatorView. 
 
fixes #8992 

### API Changes ###

Added:
 - `public IndicatorView IndicatorView`

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###

Try create a custom IndicatorView and provide an override to link to other control.

### DIFF
--- a/Xamarin.Forms.Core/IndicatorView.cs
+++ b/Xamarin.Forms.Core/IndicatorView.cs
@@ -107,23 +107,6 @@ namespace Xamarin.Forms
 		}
 
 
-		[TypeConverter(typeof(ReferenceTypeConverter))]
-		public ItemsView ItemsSourceBy
-		{
-			get => (ItemsView)GetValue(ItemsSourceByProperty);
-			set => SetValue(ItemsSourceByProperty, value);
-		}
-
-		public static readonly BindableProperty ItemsSourceByProperty = BindableProperty.Create(nameof(ItemsSourceBy), typeof(ItemsView), typeof(IndicatorView), default(VisualElement), propertyChanged: (bindable, oldValue, newValue)
-		 => ((IndicatorView)bindable).LinkToItemsView(newValue as ItemsView));
-
-
-		public virtual void LinkToItemsView(ItemsView itemsView)
-		{
-			if (itemsView is CarouselView carouselView)
-				LinkToCarouselView(this, carouselView);
-		}
-
 		protected override SizeRequest OnMeasure(double widthConstraint, double heightConstraint)
 		{
 			var baseRequest = base.OnMeasure(widthConstraint, heightConstraint);
@@ -148,24 +131,6 @@ namespace Xamarin.Forms
 				(indicatorView.IndicatorLayout as IndicatorStackLayout).Remove();
 				indicatorView.IndicatorLayout = null;
 			}
-		}
-
-		static void LinkToCarouselView(IndicatorView indicatorView, CarouselView carouselView)
-		{
-			if (carouselView == null || indicatorView == null)
-				return;
-
-			indicatorView.SetBinding(PositionProperty, new Binding
-			{
-				Path = nameof(CarouselView.Position),
-				Source = carouselView
-			});
-
-			indicatorView.SetBinding(ItemsSourceProperty, new Binding
-			{
-				Path = nameof(ItemsView.ItemsSource),
-				Source = carouselView
-			});
 		}
 
 		void ResetItemsSource(IEnumerable oldItemsSource)
--- a/Xamarin.Forms.Core/IndicatorView.cs
+++ b/Xamarin.Forms.Core/IndicatorView.cs
@@ -108,14 +108,14 @@ namespace Xamarin.Forms
 
 
 		[TypeConverter(typeof(ReferenceTypeConverter))]
-		public VisualElement ItemsSourceBy
+		public ItemsView ItemsSourceBy
 		{
-			get => (VisualElement)GetValue(ItemsSourceByProperty);
+			get => (ItemsView)GetValue(ItemsSourceByProperty);
 			set => SetValue(ItemsSourceByProperty, value);
 		}
 
-		public static readonly BindableProperty ItemsSourceByProperty = BindableProperty.Create(nameof(ItemsSourceBy), typeof(VisualElement), typeof(IndicatorView), default(VisualElement), propertyChanged: (bindable, oldValue, newValue)
-		 => (bindable as IndicatorView)?.LinkToVisualElement(newValue as VisualElement));
+		public static readonly BindableProperty ItemsSourceByProperty = BindableProperty.Create(nameof(ItemsSourceBy), typeof(ItemsView), typeof(IndicatorView), default(VisualElement), propertyChanged: (bindable, oldValue, newValue)
+		 => ((IndicatorView)bindable).LinkToVisualElement(newValue as ItemsView));
 
 
 		public virtual void LinkToVisualElement(VisualElement visualElement)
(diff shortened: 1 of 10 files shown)


Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -106,18 +106,22 @@ public IEnumerable ItemsSource
 			set => SetValue(ItemsSourceProperty, value);
 		}
 
-		public static readonly BindableProperty ItemsSourceByProperty = BindableProperty.Create("ItemsSourceBy", typeof(VisualElement), typeof(IndicatorView), default(VisualElement), propertyChanged: (bindable, oldValue, newValue)
-		 => LinkToCarouselView(bindable as IndicatorView, newValue as CarouselView));
 
 		[TypeConverter(typeof(ReferenceTypeConverter))]
-		public static VisualElement GetItemsSourceBy(BindableObject bindable)
+		public VisualElement ItemsSourceBy
```
StephaneDelcroix: if we go with this, we need a better name

StephaneDelcroix: the property accepts any `VisualElement` but the app will do nothing unless it's a CarouselView ?

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -106,18 +106,22 @@ public IEnumerable ItemsSource
...
 		{
-			return (VisualElement)bindable.GetValue(ItemsSourceByProperty);
+			get => (VisualElement)GetValue(ItemsSourceByProperty);
+			set => SetValue(ItemsSourceByProperty, value);
 		}
 
-		public static void SetItemsSourceBy(BindableObject bindable, VisualElement value)
+		public static readonly BindableProperty ItemsSourceByProperty = BindableProperty.Create(nameof(ItemsSourceBy), typeof(VisualElement), typeof(IndicatorView), default(VisualElement), propertyChanged: (bindable, oldValue, newValue)
+		 => (bindable as IndicatorView)?.LinkToVisualElement(newValue as VisualElement));
```
StephaneDelcroix: direct cast here

Xamarin.Forms.Core/IndicatorView.cs
```diff
@@ -106,18 +106,22 @@ public IEnumerable ItemsSource
...
+
+
+		public virtual void LinkToVisualElement(VisualElement visualElement)
 		{
-			bindable.SetValue(ItemsSourceByProperty, value);
+			if (visualElement is CarouselView carouselView)
```
samhouts: Can we link to an ItemsView by default, maybe?

rmarinho: Sure

Xamarin.Forms.Core/IndicatorView.cs
rmarinho: Fixed

Xamarin.Forms.Core/IndicatorView.cs
rmarinho: Only takes ItemsView now, just works if it's a carousel but user can override the virtual method for other controls

Xamarin.Forms.Core/Items/CarouselView.cs
```diff
@@ -168,6 +168,38 @@ public LinearItemsLayout ItemsLayout
 			set => SetValue(ItemsLayoutProperty, value);
 		}
 
+
+		[TypeConverter(typeof(ReferenceTypeConverter))]
+		public IndicatorView IndicatorView
```
StephaneDelcroix: does this need to be a BP? because it'll keep a ref to the indicatorview.

Xamarin.Forms.Core/Items/CarouselView.cs
```diff
@@ -168,6 +168,38 @@ public LinearItemsLayout ItemsLayout
...
+		{
+			get => (IndicatorView)GetValue(IndicatorViewProperty);
+			set => SetValue(IndicatorViewProperty, value);
+		}
+
+		public static readonly BindableProperty IndicatorViewProperty = BindableProperty.Create(nameof(IndicatorView), typeof(IndicatorView), typeof(CarouselView), default(IndicatorView), propertyChanged: (bindable, oldValue, newValue)
+		 => LinkToIndicatorView(newValue as IndicatorView, bindable as CarouselView));
```
StephaneDelcroix: use direct cast

Xamarin.Forms.Core/Items/CarouselView.cs
```diff
@@ -168,6 +168,38 @@ public LinearItemsLayout ItemsLayout
...
+
+
+		static void LinkToIndicatorView(IndicatorView indicatorView, CarouselView carouselView)
```
StephaneDelcroix: swap the order of the params (the first one being the carousel view)

Xamarin.Forms.Core/Items/CarouselView.cs
```diff
@@ -168,6 +168,38 @@ public LinearItemsLayout ItemsLayout
...
+		{
+			if (carouselView == null || indicatorView == null)
```
StephaneDelcroix: carouselview is never null here

Xamarin.Forms.Core/Items/CarouselView.cs
```diff
@@ -168,6 +168,38 @@ public LinearItemsLayout ItemsLayout
...
+			{
+				Path = nameof(CarouselView.Position),
+				Source = carouselView
+			});
+
+			indicatorView.SetBinding(IndicatorView.ItemsSourceProperty, new Binding
+			{
+				Path = nameof(ItemsView.ItemsSource),
+				Source = carouselView
+			});
```
StephaneDelcroix: setting bindings is one thing, unsetting the previous ones is another...

Xamarin.Forms.Core/Items/CarouselView.cs
rmarinho: Fixed

samhouts:
@StephaneDelcroix Indicators originally belonged to CarouselView. It was part of the CarouselView code. Then we decided to break the dependency on CarouselView because IndicatorView could be used for any control that the user wanted to create. 

This flexibility came at a cost for people who wanted an easy way to show indicators for a CarouselView, so we invented this way to easily bind the two together. We don't need to easily bind the IndicatorView to a custom control because at that point, the user is already opting for a more advanced scenario.

If we put a strong link on the IndicatorView, it suggests that the IndicatorView is only for the CarouselView, which is false. If we try to make it generic, now the IndicatorView is trying to unbox something to see if it has a known path for binding to it, which is inefficient and is the wrong place for that responsibility. And to have users override that to provide their own link is more code than just writing the bindings, as you pointed out.

I like this convenience property for the CarouselView, but I'd rather remove it than put it back on IndicatorView, personally.

Thoughts?

rmarinho:
Since this is under the flag we gonna see if users give us some feedback. Also we don't think the scenario of multiple indicators used by a carousel to be common, and it that case we will recommend bind to the position on the view model.

PR #9389, opened 2020-02-01
Replacement ActivityIndicator for WPF

### LINKED ISSUE #9372: [Enhancement] [WPF] ActivityIndicatorRenderer to use circular animation
## Summary
Currently the WPF implementation of ActivityIndicator uses an indeterminate ProgressBar which doesn't really make sense in most scenarios. Windows 8 and above have used a circular spinning dots animation for this purpose although it doesn't exist as a Win32 or WPF control. I have implemented a custom renderer (https://github.com/inthehand/InTheHand.Forms/blob/master/InTheHand.Forms/InTheHand.Forms.Platform.WPF/ActivityIndicatorRenderer.cs) to achieve this consistent look using a very simple WPF UserControl which is modelled on the WinRT/UWP control. I'm proposing a Pull Request to replace the current renderer with this one which provides a better UI.

## API Changes

No API changes are required to the control. The renderer fully supports the IsRunning and Color properties of the ActivityIndicator control. This change will affect the appearance on the WPF platform only.

## Intended Use Case
The control is intended for showing indeterminate activity and on other platforms uses some variant of a circular spinning device. If you have designed a Xamarin Forms UI with the control a stretched ProgressBar will look out of place (and there is already a separate control for that).

peterfoot:
### Description of Change ###

Replaces the default ProgressBar with a proper Activity spinner to show indeterminate activity. Style is consistent with Windows features and is the same as used throughout Windows 8 and above.

### Issues Resolved ### 
- fixes #9372 

### API Changes ###
 None

### Platforms Affected ### 
- WPF

Before the change an ActivityIndicator is rendered as a progress bar with indeterminate style which doesn't fit with the look across multiple platforms or the control used in various Windows components. After the change the ActivityIndicator is rendered as a circular animated control with multiple dots chasing around. This is similar to the control used on UWP but works on any WPF platform.

Before:-
activityindicator-before

After:-
activityindicator-after

### Testing Procedure ###
The new behaviour can be tested on the existing control gallery project.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.WPF/Controls/FormsProgressRing.xaml.cs
+++ b/Xamarin.Forms.Platform.WPF/Controls/FormsProgressRing.xaml.cs
@@ -31,12 +31,12 @@ namespace Xamarin.Forms.Platform.WPF.Controls
 			}
 		}
 
-		private static void IsActiveChanged(DependencyObject sender, DependencyPropertyChangedEventArgs e)
+		static void IsActiveChanged(DependencyObject sender, DependencyPropertyChangedEventArgs e)
 		{
 			((FormsProgressRing)sender).OnIsActiveChanged(Convert.ToBoolean(e.NewValue));
 		}
 
-		private void OnIsActiveChanged(bool newValue)
+		void OnIsActiveChanged(bool newValue)
 		{
 			if (newValue)
 			{
--- a/Xamarin.Forms.Platform.WPF/Controls/FormsProgressRing.xaml.cs
+++ b/Xamarin.Forms.Platform.WPF/Controls/FormsProgressRing.xaml.cs
@@ -1,0 +1,59 @@
+﻿using System;
+using System.Windows;
+using System.Windows.Controls;
+using System.Windows.Media.Animation;
+
+namespace Xamarin.Forms.Platform.WPF.Controls
+{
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Platform.WPF/Controls/FormsProgressRing.xaml.cs
```diff
@@ -0,0 +1,59 @@
...
+				return (bool)GetValue(IsActiveProperty);
+			}
+
+			set
+			{
+				SetValue(IsActiveProperty, value);
+			}
+		}
+
+		private static void IsActiveChanged(DependencyObject sender, DependencyPropertyChangedEventArgs e)
```
rmarinho: let's drop the `private` on these methods

peterfoot: Fixed this now

rmarinho: Thanks, building now, I m going to test this.

peterfoot:
Thanks @rmarinho - If Color.Default is used (as is the default) it was leaving the foreground transparent. I've changed this now to pick up Color.Accent from the system in this case.

PR #9390, opened 2020-02-01
fixes #6679 [Bug] WebView UWP not displaying

### LINKED ISSUE #6679: [Bug]  WebView UWP not displaying (BugZilla: 57451)
In production the WebView that I'm using to display the terms and condition when my app starts randomly doesn't show up. I opened a [StackOverflow question ](https://stackoverflow.com/questions/56770456/xamarin-webview-sometimes-doesnt-show-in-uwp?noredirect=1#comment100125670_56770456)for the issue. 

Later I found an existing bug report with comes with a reproduce-able sample on your old BugZilla: https://xamarin.github.io/bugzilla-archives/57/57451/bug.html

I'm a bit surprised to see a bug of a basic function like the WebView with local files being open for 2 years. Given that when I currently search for 57451 on this Github I found no issue, I opened this bug report to draw more attention.

artemious7:
**If you plan to make service releases for previous versions of Xamarin.Forms, please, apply this fix to all of them, cause this bug is really-really old and exists since version 2.4.0.**

### Description of Change ###
The problem with the WebView appeared because of `var internalWebView` being garbage-collected. This variable is created inside a method, and once the method is done, the variable may be garbage-collected, and may not be (you never know when it happens, but it happens more often on devices with less memory, I guess). 
Putting this `internalWebView` into a private field fixes this bug.

### Issues Resolved ### 

- fixes #6679 [Bug] WebView UWP not displaying

### API Changes ###
 
 None

### Platforms Affected ### 

- UWP

None

Not applicable

### Testing Procedure ###
If you put `GC.Collect()` right after call to `LoadHtml()` (that is [at this line](https://github.com/xamarin/Xamarin.Forms/blob/bd31e1e9fc8b2f9ad94cc99e0c7ab058174821f3/Xamarin.Forms.Core/HtmlWebViewSource.cs#L29)), you will see that the WebView will stay empty, when an `HtmlWebViewSource` is set as its `Source`.

After applying this fix, `GC.Collect()` doesn't affect the WebView, it displays just fine.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/WebViewRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/WebViewRenderer.cs
@@ -14,7 +14,7 @@ namespace Xamarin.Forms.Platform.UWP
 	{
 		WebNavigationEvent _eventState;
 		bool _updating;
-		private Windows.UI.Xaml.Controls.WebView _internalWebView;
+		Windows.UI.Xaml.Controls.WebView _internalWebView;
 		const string LocalScheme = "ms-appx-web:///";
 
 		// Script to insert a <base> tag into an HTML document
--- a/Xamarin.Forms.Platform.UAP/WebViewRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/WebViewRenderer.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.ComponentModel;
 using Windows.UI.Core;
 using Windows.UI.Xaml.Controls;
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.UAP/WebViewRenderer.cs
```diff
@@ -14,6 +14,7 @@ public class WebViewRenderer : ViewRenderer<WebView, Windows.UI.Xaml.Controls.We
 	{
 		WebNavigationEvent _eventState;
 		bool _updating;
+		private Windows.UI.Xaml.Controls.WebView _internalWebView;
```
samhouts:
```suggestion
		Windows.UI.Xaml.Controls.WebView _internalWebView;
```

artemious7:
Ohh too bad this fix is not in the service release 4

PR #9422, opened 2020-02-04
[Enhancement] AdaptiveTriggers

### LINKED ISSUE #4232: Spec: Visual State Manager Improvements
Drafted by @rmarinho 

## Rationale

Xamarin.Forms 3.0 introduced [VisualStateManager](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/visual-state-manager). This proposal is to add support for states, triggers, and provide a foundation for animations.

- VSM should make easy to activate states from XAML, for this we could introduce the concept of a `StateTrigger`, a state trigger activates the attached state depending of a number conditions, we could have builtin `AdaptiveTrigger` where we activate a state depending on features like screen size or orientation, or a `EventStateTrigger` that could activate a state when a event is fired and deactivate when other event is fired. We already have this kind of concept and a existing `EventTigger` and `Triggerbase` .  
That way the user doesn't have to call the GoToState on code behind. A StateTrigger should only apply when all conditions are met, if any of the conditions isn't met all the modifications to the properties made by the corresponding VisualState are automatically removed and the values provided initally take effect.

- VSM should support more out of the box states for each of the Views like Button, Entry, Editor,  ListView 
- VisualState should provide a way to execute animations instead of just setting properties on the elements. We should consider extend `Animation` or to introduce the a new api like `Storyboard`(similar to Animation but for XAML usage)

It's important this works with OnPlatform also. 

## Implementation

```cs

public class StateTrigger : TriggerBase
{
    public void SetActive(bool active);
    public bool IsActive { get; }
}

public class AdaptiveTrigger : StateTrigger
{
    public double MinHeight { get; set; }
    public double MinWidth { get; set; }
    public Orientation CurrentOrientation { get; set; }
}

public class EventStateTrigger : StateTrigger
{
    public string ActivateEvent { get; set; }
    public string DeactivateEvent { get; set; }
}


public class VisualState
{
    public IList<StateTrigger> StateTriggers { get; }
}

public class VisualStateManager
{
    public static bool GoToState(VisualElement visualElement, string name, bool useTransition = false)
}

internal class CommonStates
{
    ...
    public const string Pressed = "Pressed";
    public const string Released = "Released";
    public const string Hovered = "Hovered";
    public const string Selected = "Selected";
    public const string UnSelected = "UnSelected";
    public const string Toggled = "Toggled";
    public const string Running = "Running";
}

```
XAML Sample

```
<ContentPage.Content>
		<StackLayout Orientation="Vertical"
				<VisualStateGroup>
					<VisualState x:Name="Orientation">
						<VisualState.StateTriggers>
							<AdaptiveTrigger MinWidth="640" />
						</VisualState.StateTriggers>

						<VisualState.Setters>
							<Setter Property="Orientation" Value="Horizontal" />
						</VisualState.Setters>
					</VisualState>
				</VisualStateGroup>

			<Label Text="Welcome to Xamarin.Forms!">
					<VisualStateGroup>
						<VisualState x:Name="Focused">
							<VisualState.StateTriggers>
								<EventTrigger Event="Focused" />
							</VisualState.StateTriggers>

							<VisualState.Setters>
								<Setter Property="TextColor" Value="Red" />
							</VisualState.Setters>
						</VisualState>
					</VisualStateGroup>
			</Label>
			<Label Text="Welcome to Xamarin.Forms!" />
		</StackLayout>
	</ContentPage.Content>
```

- Need to track Toogled state on Switch 
- Need to track Running state on ActivityIndicator 
- Need to track state Selected / UnSelected on ListView

#### Android

- We already track device orientation changes, we should track screen size too

#### iOS

- We already track device orientation changes, we should track screen size too

#### UWP

- We already track device orientation changes, we should track screen size too
- We need to track the hover state on a button

#### MacOS

- We should track screen size 
- We need to track the hover state on a button

#### GTK

- We should track screen size too
- We need to track the hover state on a button

### Implications for CSS

We could try to map media queries to `AdaptiveTriggers`but I think this is out of the scope for this spec.

### Backward Compatibility

There should be no changes to the existing usage of VSM without triggers, transitions or animations.
Using none of the new features should just work as before.

### Difficulty : Medium

### LINKED ISSUE #9189: i wish some day you add “AdaptiveTrigger” of UWP Visual states to xamarin forms too
## Summary
Please provide a brief summary of your proposal. Two to three sentences is best here.

## API Changes

Include a list of all API changes, additions, subtractions as would be required by your proposal. These APIs should be considered placeholders, so the naming is not as important as getting the concepts correct. If possible you should include some "example" code of usage of your new API.

e.g.

In order to facilitate the new Shiny Button api, a bool is added to the Button class. This is done as a bool because it is simpler to data bind and other reasons...

    var button = new Button ();
    button.MakeShiny = true; // new API

The MakeShiny API works even if the button is already visible.

## Intended Use Case
Provide a detailed example of where your proposal would be used and for what purpose.

jsuarezruiz:
### Description of Change ###

An **AdaptiveTrigger** provides an easy way to set the threshold (also called 'breakpoint') where a state is applied. A VisualState defines property values that are applied to an element when it’s in a particular state. You group visual states in a VisualStateManager that applies the appropriate VisualState when the specified conditions are met.

Example:

```
<VisualStateGroupList>
        <VisualStateGroup>
            <VisualState x:Name="Narrow">
            <VisualState.StateTriggers>
                <AdaptiveTrigger MinWindowWidth="0" />
            </VisualState.StateTriggers>
            <VisualState.Setters>
                <Setter Property="BackgroundColor" Value="Blue" />
            </VisualState.Setters>
        </VisualState>
        <VisualState x:Name="Wide">
            <VisualState.StateTriggers>
                <AdaptiveTrigger MinWindowWidth="1000" />
            </VisualState.StateTriggers>
            <VisualState.Setters>
                <Setter Property="BackgroundColor" Value="Green" />
            </VisualState.Setters>
        </VisualState>
    </VisualStateGroup>
</VisualStateGroupList>
```

### Issues Resolved ### 

- fixes #4232
- fixes #9189

### API Changes ###

Added:

```
public class StateTriggerBase
{
    public void SetActive(bool active);
    public bool IsActive { get; }
}

public class AdaptiveTrigger : StateTriggerBase
{
    public double MinHeight { get; set; }
    public double MinWidth { get; set; }
}

public class OrientationStateTrigger : StateTriggerBase
{
    public DeviceOrientation Orientation { get; set; }
}

public class DeviceStateTrigger : StateTriggerBase
{
    public string Device { get; set; }
}

public class CompareStateTrigger : StateTriggerBase
{
    public object Property { get; set; }
    public object Value { get; set; }
}
```

In Xamarin.Forms.DualScreen:

```
public class DualScreenSpanModeStateTrigger : StateTriggerBase
{
    public TwoPaneViewMode SpanMode { get; set; }
}
```

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

**AdaptiveTrigger**
adaptivetrigger2

**DeviceStateTrigger**

```
<VisualStateGroupList>
    <VisualStateGroup>
        <VisualState
            x:Name="Android">
            <VisualState.StateTriggers>
                <DeviceStateTrigger Device="Android" />
            </VisualState.StateTriggers>
            <VisualState.Setters>
                <Setter Property="BackgroundColor" Value="Blue" />
            </VisualState.Setters>
        </VisualState>
        <VisualState
            x:Name="iOS">
            <VisualState.StateTriggers>
                <DeviceStateTrigger Device="iOS" />
            </VisualState.StateTriggers>
            <VisualState.Setters>
                <Setter Property="BackgroundColor" Value="Red" />
            </VisualState.Setters>
        </VisualState>
    </VisualStateGroup>
</VisualStateGroupList>
```

Running on Android
devicestate-droid

Running on iOS
devicestate-ios

**OrientationStateTrigger**

```
<VisualStateGroupList>
        <VisualStateGroup>
        <VisualState
            x:Name="Landscape">
            <VisualState.StateTriggers>
                <OrientationStateTrigger Orientation="Landscape" />
            </VisualState.StateTriggers>
            <VisualState.Setters>
                <Setter Property="BackgroundColor" Value="Blue" />
            </VisualState.Setters>
        </VisualState>
        <VisualState
            x:Name="Portrait">
            <VisualState.StateTriggers>
                <OrientationStateTrigger Orientation="Portrait" />
            </VisualState.StateTriggers>
            <VisualState.Setters>
                <Setter Property="BackgroundColor" Value="Red" />
            </VisualState.Setters>
        </VisualState>
    </VisualStateGroup>
</VisualStateGroupList>
```

orientationtrigger

**CompareStateTrigger**

```
<VisualStateGroupList>
    <VisualStateGroup>
        <VisualState x:Name="Wide">
            <VisualState.StateTriggers>
                <CompareStateTrigger Property="{Binding IsChecked, Source={x:Reference CheckBox}}" Value="True" />
            </VisualState.StateTriggers>
            <VisualState.Setters>
                <Setter Property="BackgroundColor" Value="Green" />
            </VisualState.Setters>
        </VisualState>
        <VisualState x:Name="Narrow">
            <VisualState.StateTriggers>
                <CompareStateTrigger Property="{Binding IsChecked, Source={x:Reference CheckBox}}" Value="False" />
            </VisualState.StateTriggers>
            <VisualState.Setters>
                <Setter Property="BackgroundColor" Value="Red" />
            </VisualState.Setters>
        </VisualState>
    </VisualStateGroup>
</VisualStateGroupList>
```

comparestatetrigger

**DualScreenStateTrigger**

```
<Grid>
        <VisualStateGroup>
            <VisualState x:Name="Normal">
                <VisualState.StateTriggers>
                    <dualScreen:DualScreenSpanModeStateTrigger SpanMode="SinglePane"/>
                </VisualState.StateTriggers>
                <VisualState.Setters>
                    <Setter Property="BackgroundColor" Value="Red" />
                </VisualState.Setters>
            </VisualState>
            <VisualState x:Name="Spanned">
                <VisualState.StateTriggers>
                    <dualScreen:DualScreenSpanModeStateTrigger SpanMode="Wide" />
                </VisualState.StateTriggers>
                <VisualState.Setters>
                    <Setter Property="BackgroundColor" Value="Green" />
                </VisualState.Setters>
            </VisualState>
        </VisualStateGroup>
</Grid>
```

one-screen
two-screens

### Testing Procedure ###
Launch Core Gallery and navigate to the VisualStateManager samples. Here, can find new samples related with Adaptive Triggers.

### PR Checklist ###


Xamarin.Forms.Core/WeakEventListener.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.ComponentModel;
+
+namespace Xamarin.Forms
+{
+	[EditorBrowsable(EditorBrowsableState.Never)]
+    public sealed class WeakEventListener<TInstance, TSource, TEventArgs>
```
PureWeen: Can we use WeakEventManager
https://github.com/xamarin/Xamarin.Forms/blob/c8cf3ba9edbc50bd2933515af39812949380d034/Xamarin.Forms.Core/WeakEventManager.cs#L10

Xamarin.Forms.Controls/GalleryPages/VisualStateManagerGalleries/StateTriggersDirectlyOnElements.xaml
```diff
@@ -0,0 +1,38 @@
...
+    x:Class="Xamarin.Forms.Controls.GalleryPages.VisualStateManagerGalleries.StateTriggersDirectlyOnElements"
+    Title="StateTriggers directly on Elements">
+    <ContentPage.Content>
+        <Grid>
+            <VisualStateManager.VisualStateGroups>
+                <VisualStateGroup>
+                    <VisualState
+                        x:Name="Landscape">
+                        <VisualState.StateTriggers>
+                            <StateTrigger IsActive="{Binding ToggleStateInverse}" />
```
PureWeen: I updated this sample to use StateTrigger with a binding and it doesn't seem like the bindingcontext is propagating.

The BindingContext of the Grid should propagate to the state triggers.

Here's an example from UWP
https://github.com/microsoft/Windows-universal-samples/blob/master/Samples/XamlStateTriggers/cs/ItemTemplateControl.xaml#L37

Using a similar concept

Xamarin.Forms.Core/AdaptiveTrigger.cs
```diff
@@ -0,0 +1,97 @@
...
+		public AdaptiveTrigger()
+		{
+			UpdateState();
+
+			if (!DesignMode.IsDesignModeEnabled)
+			{
+				var weakEvent = new WeakEventListener<AdaptiveTrigger, object, EventArgs>(this)
+				{
+					OnEventAction = (instance, source, eventArgs) => OnSizeChanged(source, eventArgs),
+					OnDetachAction = (listener) => { Application.Current.MainPage.SizeChanged -= listener.OnEvent; }
```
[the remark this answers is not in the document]
jsuarezruiz: Applied changes to use `Device.info.ScaledScreenSize`

Xamarin.Forms.Controls/GalleryPages/VisualStateManagerGalleries/StateTriggersDirectlyOnElements.xaml
jsuarezruiz: Good one. Fixed and included more samples using Bindings with StateTriggers.

jsuarezruiz: statetrigger-binding

jsuarezruiz:
@PureWeen I have added some unit tests.

Xamarin.Forms.Core/View.cs
```diff
@@ -155,9 +155,25 @@ public LayoutOptions VerticalOptions
 		protected override void OnBindingContextChanged()
 		{
 			this.PropagateBindingContext(GestureRecognizers);
+			PropagateBindingContextToStateTriggers();
```
PureWeen: Should this be at the VisualElement level?

*VisualStateManager.VisualStateGroupsProperty*

Can be applied to anything that's of type VisualElement

```C#
public static readonly BindableProperty VisualStateGroupsProperty =
			BindableProperty.CreateAttached("VisualStateGroups", typeof(VisualStateGroupList), typeof(VisualElement),
				defaultValue: null, propertyChanged: VisualStateGroupsPropertyChanged,
				defaultValueCreator: bindable => new VisualStateGroupList { VisualElement = (VisualElement)bindable });
```

jsuarezruiz: Yeah, better. I moved it.

dotMorten:
Thank you for making IsActive a public getter instead of protected. I'd like to propose you also add an event for when it changes. The reason for that is to be able to make composite state triggers. I state my case for that in this design issue:
https://github.com/microsoft/microsoft-ui-xaml/issues/1460

jsuarezruiz:
Thanks for the feedback @dotMorten. I think one of the most interesting things related to StateTriggers is the possibility of creating new ones and your proposal is interesting because it is simple but enables new options.

Xamarin.Forms.Core/WeakEventListener.cs
PureWeen: Can we make this internal?

Xamarin.Forms.Core/OrientationStateTrigger.cs
```diff
@@ -0,0 +1,70 @@
...
+	{
+		public OrientationStateTrigger()
+		{
+			UpdateState();
+
+			if (!DesignMode.IsDesignModeEnabled)
+			{
+				var weakEvent = new WeakEventListener<OrientationStateTrigger, object, EventArgs>(this)
+				{
+					OnEventAction = (instance, source, eventArgs) => OnSizeChanged(source, eventArgs),
```
PureWeen: DeviceInfo implements INPC so you can just subscribe to that and watch for Orientation changes

Xamarin.Forms.Core/StateTriggerBase.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.ComponentModel;
+
+namespace Xamarin.Forms
+{
+	public class StateTriggerBase : BindableObject
```
PureWeen: abstract?

Xamarin.Forms.Core/AdaptiveTrigger.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
+
+namespace Xamarin.Forms
+{
+	public class AdaptiveTrigger : StateTriggerBase
```
PureWeen: Let's make all the triggers we are adding sealed for now

Xamarin.Forms.Core/StateTriggerBase.cs
dotMorten: Oddly enough the class this design was taken from isn't abstract (albeit I agree it makes sense, apart from that would diverge on a common design): https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.StateTriggerBase

Xamarin.Forms.Core/AdaptiveTrigger.cs
dotMorten: Also same comment: Makes sense as there's nothing to override, but the UWP design doesn't have it sealed

Xamarin.Forms.Core/View.cs
```diff
@@ -155,6 +155,7 @@ public LayoutOptions VerticalOptions
 		protected override void OnBindingContextChanged()
 		{
 			this.PropagateBindingContext(GestureRecognizers);
+
```
PureWeen: remove
```suggestion
```

Xamarin.Forms.Core/VisualElement.cs
```diff
@@ -890,6 +896,20 @@ internal void UnmockBounds()
...
+		void PropagateBindingContextToStateTriggers()
+		{
+			var stateTriggers = new List<StateTriggerBase>();
+
+			var groups = (IList<VisualStateGroup>)GetValue(VisualStateManager.VisualStateGroupsProperty);
+
+			foreach (var group in groups)
+				foreach (var state in group.States)
+					foreach (var stateTrigger in state.StateTriggers)
+						stateTriggers.Add(stateTrigger);
```
PureWeen: Can this just use
```C#
SetInheritedBindingContext(stateTrigger, BindingContext);
```

Instead of having to create the list?

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -274,6 +291,9 @@ public VisualStateGroup()
 		public IList<VisualState> States { get; }
 		public VisualState CurrentState { get; internal set; }
 
+		[EditorBrowsable(EditorBrowsableState.Never)]
```
PureWeen:
```suggestion
```

Xamarin.Forms.Core/VisualStateManager.cs
```diff
@@ -287,61 +307,197 @@ internal VisualState GetState(string name)
...
+			StateTriggers = new WatchAddList<StateTriggerBase>(OnStateTriggersChanged);
 		}
 
 		public string Name { get; set; }
-		public IList<Setter> Setters { get;}
+		public IList<Setter> Setters { get; }
+		public IList<StateTriggerBase> StateTriggers { get; }
 		public Type TargetType { get; set; }
 
+		[EditorBrowsable(EditorBrowsableState.Never)]
```
PureWeen:
```suggestion
```

Xamarin.Forms.Core/StateTriggerBase.cs
```diff
@@ -0,0 +1,24 @@
...
+	public abstract class StateTriggerBase : BindableObject
+	{
+		public StateTriggerBase()
+		{
+			ExperimentalFlags.VerifyFlagEnabled(nameof(IndicatorView), ExperimentalFlags.AdaptiveTriggersExperimental);
+		}
+
+		internal bool IsTriggerActive { get; set; }
+
+		[EditorBrowsable(EditorBrowsableState.Never)]
```
PureWeen:
```suggestion
```

Xamarin.Forms.Core/WeakEventListener.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.ComponentModel;
+
+namespace Xamarin.Forms
+{
+	[EditorBrowsable(EditorBrowsableState.Never)]
```
PureWeen:
```suggestion
```

Xamarin.Forms.Core/ExperimentalFlags.cs
```diff
@@ -9,6 +9,7 @@ namespace Xamarin.Forms
 {
 	internal static class ExperimentalFlags
 	{
+		internal const string AdaptiveTriggersExperimental = "AdaptiveTriggers_Experimental";
```
PureWeen: I think this should be StateTriggers

PureWeen:
@dotMorten 

IsActive only exists on StateTrigger 
https://github.com/xamarin/Xamarin.Forms/pull/9422/files#diff-076f6683a045207a713a5d05d28028b9R5

On StateTriggerBase the property is IsTriggerActive (https://github.com/xamarin/Xamarin.Forms/pull/9422/files#diff-31f161e53637b04031f6ae93a28b9a58R15) 

So it seems like matching the event to the name at that level makes sense

Xamarin.Forms.Core/StateTrigger.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Xamarin.Forms
+{
+	public class StateTrigger : StateTriggerBase
```
PureWeen:
```suggestion
	public sealed class StateTrigger : StateTriggerBase
```

Xamarin.Forms.DualScreen/SpanModeStateTrigger.shared.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System.ComponentModel;
+
+namespace Xamarin.Forms.DualScreen
+{
+	public class SpanModeStateTrigger : StateTriggerBase
```
PureWeen:
```suggestion
	public sealed class SpanModeStateTrigger : StateTriggerBase
```

dotMorten:
@PureWeen Ah missed that! Isn't is weird to have the same property both on the base class and the subclass? Why not just move IsActive down to the base-class then, and not also have IsTriggerActive? I realize that goes against something else I'd like to see: Consistency between the UWP and Forms APIs, but it feels like a cleaner design.

PR #9426, opened 2020-02-04
Expose Resolve and Get methods on DependencyService

### LINKED ISSUE #9424: [Enhancement] Expose Resolve(Type) and Get(Type) methods on DependencyService to be used in a non generic way
## Summary
Expose a Resolve(Type) and Get(Type) methods on [DependencyService ](https://github.com/xamarin/Xamarin.Forms/blob/f0087cbf5af2b11cca6b5bc6977c1ea75da0dd40/Xamarin.Forms.Core/DependencyService.cs#L19)for used in a non generic way.

## API Changes

A consumer can now use the api like,

```csharp
Type someType ;//some type from a external source, like a MarkupExtension 
var instance = DependencyService.Resolve(someType); //new API method
```
same for Get<T> method

The methods will look like:
```csharp
public static object Resolve(Type serviceType, DependencyFetchTarget fallbackFetchTarget = DependencyFetchTarget.GlobalInstance);

public static object Get(Type serviceType, DependencyFetchTarget fallbackFetchTarget = DependencyFetchTarget.GlobalInstance);
```
We can convert the main methods in this and expose the generic version as extensions methods or included in the same class. 

This change will not affect any existing app.
## Intended Use Case
This is useful in cases where the generic parameter is not possible, like a MarkupExtension  it will allow to external libraries easy relay on the default service resolution mechanism provided by the framework, eg. 

An MarkupExtensions that provide a service to a given type can pass only the Type of the service and non the generic parameter, since the DependencyService.Resolve use the DependencyResolver and a user can change the default resolution strategy to use its how resolution strategy.

I know that this can be archive using reflection but, since this change non break any code can be implemented with safety, and avoid the payload of use reflection. 

## Related issues
Support for DependencyTemplateExtension https://github.com/xamarin/Xamarin.Forms/issues/9425

### LINKED ISSUE #9425: [Spec]  DependencyTemplateExtension
# DependencyTemplateExtension 
Now that shell become common for all the community and has a easy way to use lazy load of component using the TemplateExtension we can add a new markup extension that can resolve the instance of the related DataTemplate using the DependencyService.Resolve method, this way we can integrate any IoC container since we can change the default factory method used by the DependencyResolver.

# API
## DependencyTemplateExtension 
A markup extensions that receive the service "Type" to be resolved and use DependencyService.Resolve to create the service.

```csharp

[ContentProperty("TypeName")]
public class DpendencyTemplateExtension : IMarkupExtension<DataTemplate>, IMarkupExtension
{
     public string TypeName { get; set; }
   
     public DependencyFetchTarget FallbackTarget { get; set; }

     public DataTemplate ProvideValue(IServiceProvider serviceProvider);
     
     object IMarkupExtension.ProvideValue(IServiceProvider serviceProvider)
}
``` 

### Properties

| API | Description |
| TypeName | Store the type mame to be resolved by the IXamlTypeResolver |

# Scenarios
Use in a Shell 
## XAML Example
```xaml
<ShellContent ContentTemplate="{DenpendencyTemplate local:AboutPage}" />
```

# Backward Compatibility

Minimum API levels: 4
Breaking changes: None
Unsupported platforms: none

# Difficulty : [low]

danielmeza:
### Expose Resolve and Get methods on DependencyService ###

Add support for #9425 and #9424 proposals
### Issues Resolved ### 
- fixes #9424
- fixes #9425 

## DependencyService
Added:
 - public static object Resolve(Type serviceType, DependencyFetchTarget fallbackFetchTarget = DependencyFetchTarget.GlobalInstance)
 - public static object Get(Type targetType, DependencyFetchTarget fetchTarget = DependencyFetchTarget.GlobalInstance)

## DependencyTemplateExtension 
Total new! 

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

#### DependencyService
Since the logic here has no change and the old API relay on the new the same test apply to it should be enough.

#### DependencyTemplateExtension 
Since its relay on DependencyService the test are minimal

>We need to decide witch is  the default behavior if the service wasn't find in the DependencyService, now I [throw ](https://github.com/xamarin/Xamarin.Forms/pull/9426/files#diff-345412ac341da79219d496ceebb0ef38R36) and exception and wrap any exception into a XamlParseException to provide contextual info about where the exception happen in the xaml. It's import to catch exceptions [here ](https://github.com/xamarin/Xamarin.Forms/pull/9426/files#diff-345412ac341da79219d496ceebb0ef38R40) because the exception can be produced by a triparty IoC container.  

### PR Checklist ###


Xamarin.Forms.Xaml/MarkupExtensions/DependencyTemplateExtension.cs
```diff
@@ -0,0 +1,63 @@
...
+
+		public DataTemplate ProvideValue(IServiceProvider serviceProvider)
+		{
+			if (serviceProvider == null)
+				throw new ArgumentNullException(nameof(serviceProvider));
+			if (!(serviceProvider.GetService(typeof(IXamlTypeResolver)) is IXamlTypeResolver typeResolver))
+				throw new ArgumentException("No IXamlTypeResolver in IServiceProvider");
+			if (string.IsNullOrEmpty(TypeName))
+			{
+				var li = (serviceProvider.GetService(typeof(IXmlLineInfoProvider)) is IXmlLineInfoProvider lip) ? lip.XmlLineInfo : new XmlLineInfo();
```
campersau:
```suggestion
				var li = GetLineInfo(serviceProvider);
```

danielmeza:
Sure, letme rollback the last changes

samhouts:
@danielmeza Did you mean to  check in additional changes? I don't see any new commits.

Xamarin.Forms.Xaml/MarkupExtensionParser.cs
```diff
@@ -24,6 +25,8 @@ public object Parse(string match, ref string remaining, IServiceProvider service
 				markupExtension = new OnIdiomExtension();
 			else if (match == "DataTemplate")
 				markupExtension = new DataTemplateExtension();
+			else if (match == "DependencyTemplate")
```
danielmeza, line 28: Will remove this in order to revert the DependencyTemplateExtension implementation

danielmeza:
Hi @samhouts I did the changes, sorry for the delay I was busy at work.

danielmeza:
@StephaneDelcroix I made the change

jfversluis:
This will not happen anymore for Xamarin.Forms. If this is still important to you, make sure to check the [.NET MAUI repo](https://github.com/dotnet/maui) and see if it's already on the roadmap or open an issue with a detailed feature request. Thanks!

PR #9427, opened 2020-02-04
Fixes TPV so it works anywhere on the screen

PureWeen:
### Description of Change ###

Fix TPV to always take into account the hinge regardless of size and position
- If any layout is spanned over the hinge it'll correctly size pane1/pane2
- If the hinge is only overlapping the tpv one one of the sides but the tpv doesn't span across the hinge then this will add padding to the tpv so content doesn't flow underneath the hinge
- Added constructor for DualScreenInfo to take a layout so users can supply whatever layout they want to to DualScreenInfo and it will return the calculations for that layout relative to the Hinge

### API Changes ###
Added:
 - public DualScreenInfo(Layout layout)

### Platforms Affected ### 
- Core/XAML (all platforms)

### Testing Procedure ###
- unit tests included

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.DualScreen/DualScreenInfo.shared.cs
+++ b/Xamarin.Forms.DualScreen/DualScreenInfo.shared.cs
@@ -15,9 +15,23 @@ namespace Xamarin.Forms.DualScreen
         Rectangle _hingeBounds;
         bool _isLandscape;
         TwoPaneViewMode _spanMode;
+		TwoPaneViewLayoutGuide _twoPaneViewLayoutGuide;
+		public static DualScreenInfo Current => _dualScreenInfo.Value;
 
-        public static DualScreenInfo Current => _dualScreenInfo.Value;
-        public Rectangle[] SpanningBounds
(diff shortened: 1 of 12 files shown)


Xamarin.Forms.DualScreen/DualScreenInfo.shared.cs
```diff
@@ -15,9 +15,23 @@ public class DualScreenInfo : INotifyPropertyChanged
...
+		public DualScreenInfo(Layout layout)
+		{
+			if(layout == null)
+			{
+				_twoPaneViewLayoutGuide = TwoPaneViewLayoutGuide.Instance;
+			}
+			else
+			{
+				_twoPaneViewLayoutGuide = new TwoPaneViewLayoutGuide(layout);
+				_twoPaneViewLayoutGuide.PropertyChanged += OnTwoPaneViewLayoutGuideChanged;
```
jsuarezruiz: Could we unsubscribe this?

PR #9452, opened 2020-02-06
Fix UWP volume, mediasource exception, android position

PureWeen:
### Description of Change ###

- fix uwp to set volume after media has loaded otherwise it always starts at 1
- check to see if they've used an empty string with ms-appx:///
- add some null checks on android

### Platforms Affected ### 
- Android

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
@@ -106,10 +106,13 @@ namespace Xamarin.Forms.Platform.Android
 				UpdateBackgroundColor();
 			}
 
-			MediaElement.PropertyChanged += OnElementPropertyChanged;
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Platform.Android/FastRenderers/MediaElementRenderer.cs
```diff
@@ -99,6 +99,7 @@ void IVisualElementRenderer.SetElement(VisualElement element)
 			MediaElement.PropertyChanged += OnElementPropertyChanged;
 			MediaElement.SeekRequested += SeekRequested;
 			MediaElement.StateRequested += StateRequested;
+			MediaElement.PositionRequested += OnPositionRequested;
```
jsuarezruiz: We can use here oldElement to unsubscribe events.

```
if (oldElement != null)
{
     oldElement.PropertyChanged -= OnElementPropertyChanged;
}
```

PR #9462, opened 2020-02-07
[iOS] Fix NRE when Entry Completed fires

### LINKED ISSUE #9460: [Bug] Entry - changing app main page in Entry.OnCompleted causes crash in iOS
### Description

For better user experience we use the entry's OnCompleted event to fire the login command, which sets the app main page to a different one. This causes crash in iOS.
### Steps to Reproduce

1. App main page is set to LoginPage.xaml. Login Page consists of Entry and a Button
2. When Entry's OnCompleted event occurs, Button event handler is fired, which sets the app main page to different one. iOS app crashes with
System.NullReferenceException: Object reference not set to an instance of an object
at Xamarin.Forms.Platform.iOS.EntryRendererBase`1[TControl].OnShouldReturn (UIKit.UITextField view) 

### Expected Behavior

iOS - Application should **not crash**

### Actual Behavior
iOS - Application **crashes**

### Basic Information

- Version with issue: 4.5.0.266-pre3, 4.4.0.991640. I Downgraded to 4.2.0.910310 - still crashes
- Last known good version: 4.0.0.709238
- IDE: Visual Studio 16.4.3
- iOS:  13.2

### Reproduction Link

[EntryIosSetMainPageCrash.zip](https://github.com/xamarin/Xamarin.Forms/files/4169899/EntryIosSetMainPageCrash.zip)

Full StackTrace:
[EntryShouldReturnStackTrace.txt](https://github.com/xamarin/Xamarin.Forms/files/4169922/EntryShouldReturnStackTrace.txt)

### Workaround

Applying 
**`await Task.Delay(TimeSpan.FromMilliseconds(400));`**
before setting the new main page solves the problem.

rmarinho:
### Description of Change ###

When we fire the completed event on the entry on iOS we could also be disposed , so the Element will be set to null.  Since in that event we were checking the ReturnType it could throw a NRE

### Issues Resolved ### 
- fixes #9460 

### API Changes ###
 None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###
on App.cs use this code :

```
	var page = new ContentPage();
        var entry = new Entry();
	entry.Completed += (s,e) => SetMainPage(CreateDefaultMainPage());
        page.Content = entry;
	MainPage = page;
```

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
@@ -246,7 +246,7 @@ namespace Xamarin.Forms.Platform.iOS
 			Control.ResignFirstResponder();
 			((IEntryController)Element).SendCompleted();
 
-			if (Element.ReturnType == ReturnType.Next)
+			if (Element != null && Element.ReturnType == ReturnType.Next)
 			{
 				FocusSearch(true);
 			}


Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
```diff
@@ -246,7 +246,7 @@ protected virtual bool OnShouldReturn(UITextField view)
 			Control.ResignFirstResponder();
 			((IEntryController)Element).SendCompleted();
```
jcmanke: If the goal is to fix a NRE when firing the completed event, shouldn't you be null-checking on/before the line where the Completed event is actually fired?

PR #9479, opened 2020-02-07
Add Android Setup Script for SDKs

jamesmontemagno:
Use @redth's amazing new tool to help developers get up and running to contribute to XF.

### DIFF
--- a/android-setup.ps1
+++ b/android-setup.ps1
@@ -1,0 +1,41 @@
+$androidSdkHome = $ENV:ANDROID_HOME
+
+if ([string]::IsNullOrWhiteSpace($androidSdkHome))
+{
+    if ($ENV:OS -eq "Windows_NT")
+    {
+        $androidSdkHome = "C:\Program Files (x86)\Android\android-sdk\"
+    }
+    else
+    {
+        $androidSdkHome = Resolve-Path "~/Library/Developer/Xamarin/android-sdk-macosx"
+    }
+
+    Write-Host "Couldn't locate ANDROID_HOME, using: $androidSdkHome"
+}
+
+Write-Host "Using Android SDK Home: $androidSdkHome"
+
+Write-Host "Installing Global dotnet android tool..."
+dotnet tool install --global --add-source https://www.myget.org/F/xam-dotnet-tools/api/v3/index.json AndroidSdk.Tool
+
+Write-Host "Installing and/or Updating Android SDK..."
+Write-Host "This may take awhile..."
+android --home="$androidSdkHome" sdk download
+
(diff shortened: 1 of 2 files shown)


android-setup.ps1
```diff
@@ -0,0 +1,41 @@
...
+{
+    if ($ENV:OS -eq "Windows_NT")
+    {
+        $androidSdkHome = "C:\Program Files (x86)\Android\android-sdk\"
+    }
+    else
+    {
+        $androidSdkHome = Resolve-Path "~/Library/Developer/Xamarin/android-sdk-macosx"
+    }
+
```
VladislavAntonyuk: Should we check the $androidSdkHome path exists?
do we also support Linux as a build machine? Is it the same path as on macOS?

PureWeen:
@jamesmontemagno @Redth can we just use cake?

We already have hooks for this in cake
https://github.com/xamarin/Xamarin.Forms/blob/master/build.cake#L69

Looking at your script I think we can just create a task inside cake that does everything there plus it'll work all the places our current cake script works (sh, cmd, ps1)

jamesmontemagno:
We could, this is more of a "someone wants to contribute to the repo and needs to get their environment setup" type of script is our thinking.

PureWeen:
yea and we could just make this the default for the cake script. So you just run

./build.ps1
build.cmd
./build.sh

And it'll provision and do whatever needs to be done

PureWeen:
my fear with adding more scripts is that we'll have too many scripts
The cake scripts are documented as part of the cake script, users can use them to create nugets, and they are cross platform 

https://github.com/xamarin/Xamarin.Forms/blob/master/build.cake#L168

And you can run it with -showdescription to see all the things available to you

image

We could formalize this all a bit more so if you ran the script without targets it shows you the options that would be interesting to you.

Redth:
Yeah so for now the cake addin you are using is based on some older code. I’m planning on updating the cake addin to also use the same nuget dependency this dotnet global tool is using. If you can make this happen more easily for new users I’d agree another script is just noise.

jamesmontemagno:
I agree on the extra script as I didn't know that the cake thing was here. I guess one thing I want to get out of this is to have a script where there isn't additional dependencies to get someone up and running. Just run this file and boom your machine is ready to contribute and build the solution.

PureWeen:
Cake doesn't have any additional dependencies 

It'll just download anything that it needs. We use it to provision our CI servers so they have everything they need to build the project

jamesmontemagno:
Closing per discussion and will chat with the team about some simple setup scripts :)

PR #9481, opened 2020-02-08
C# 8

mattleibow:
Coming in here with a huge hammer. Try C# 8.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -11,7 +11,7 @@
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
     <Company>Microsoft</Company>
     <Product>Xamarin.Forms</Product>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>8.0</LangVersion>
     <ProduceReferenceAssembly Condition="'$(UsingMicrosoftNETSdk)' == 'True' AND '$(Configuration)' == 'Debug'">True</ProduceReferenceAssembly>
   </PropertyGroup>
   <ItemGroup>


slang25:
It's so weird to see my blog while idly browsing around XF issues 😄 Thank you for the mention.

That blog post was aimed at "unsupported targets", however targeting mono 6.4 and above is also a supported platform, so all of the features will work, including the runtime Default Interface Members feature.

slang25:
Interesting (I think), one of the motivations (of many) for C# adopting Default Interface Members is that Java has it, so the Xamarin Android bindings can take advantage of this by having a more direct translation of APIs.

[See here](https://docs.microsoft.com/en-gb/xamarin/android/release-notes/10/10.1#c-80-default-interface-members-for-bindings-projects). It can be enabled in Xamarin Android projects by adding `<_EnableInterfaceMembers>true</_EnableInterfaceMembers>`.

PureWeen:
@mattleibow can we set the ControlGalleries to still use <LangVersion>7.3</LangVersion> ?
And just the SDKs will compile with 8.0? 

We need to maintain verification that people can build forms projects with VS 2017

mattleibow:
@PureWeen that is something we can do. Is there a .props file that is just pulled in for the SDK, or some variable? How would I distinguish between these as this props is at the root - or do I add this to each project?

PureWeen:
@mattleibow I think you'll need to add it to each CG project

I added this line to most of the CG projects
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.ControlGallery.Android/Xamarin.Forms.ControlGallery.Android.csproj#L3


So that the props files would be imported earlier since old style csproj files import directory.build.props way down at the targets import instead of at the beginning like sdk style projects.

So you could change that line to import something like "Cg.build.props" which also imports the existing *Environment.Build.props*

```
<Import Condition="'$(EnvironmentBuildPropsImported)' != 'True'" Project="..\Environment.Build.props" />
```

Then with the 8.0 setting it'll only set it if it wasn't previously set

making this all work with vsmac + how old project styles only kind of work with sdk features has been a bit challenging

charlesroddie:
This huge hammer is absolutely needed and fast. Especially enabling NRTs across the repo.

https://github.com/xamarin/Xamarin.Forms/issues?q=nre+is%3Aopen

> We need to maintain verification that people can build forms projects with VS 2017

Just make sure you don't unpublish older versions of XF for people who don't want to update software. Problem solved.

PR #9485, opened 2020-02-08
[Android] Properly unsubscribe ToolbarItem on removal

### LINKED ISSUE #9419: ObjectDisposedException setting property value on property bound to disposed page
### Description

Apologies for the post-and-run on this. I'm traveling and seeing this in my travel app while using it.

When a bound property changes on a page that's already been disposed, it appears to throw an unhandled ObjectDisposedException. It seems unreasonable for the user to have to catch this kind of exception, or to find ways to completely avoid interactions with pages that might be disposed. But advice appreciated regardless.

Here's the stack trace.

```
JniPeerMembers.AssertSelf (Java.Interop.IJavaPeerable self)
JniPeerMembers+JniInstanceMethods.InvokeNonvirtualObjectMethod (System.String encodedMember, Java.Interop.IJavaPeerable self, Java.Interop.JniArgumentValue* parameters)
View.get_Context ()
NavigationPageRenderer.OnToolbarItemPropertyChanged (System.Object sender, System.ComponentModel.PropertyChangedEventArgs e)
BindableObject.OnPropertyChanged (System.String propertyName)
Element.OnPropertyChanged (System.String propertyName)
BindableObject.SetValueActual (Xamarin.Forms.BindableProperty property, Xamarin.Forms.BindableObject+BindablePropertyContext context, System.Object value, System.Boolean currentlyApplying, Xamarin.Forms.Internals.SetValueFlags attributes, System.Boolean silent)
BindableObject.SetValueCore (Xamarin.Forms.BindableProperty property, System.Object value, Xamarin.Forms.Internals.SetValueFlags attributes, Xamarin.Forms.BindableObject+SetValuePrivateFlags privateAttributes)
BindingExpression.ApplyCore (System.Object sourceObject, Xamarin.Forms.BindableObject target, Xamarin.Forms.BindableProperty property, System.Boolean fromTarget)
BindingExpression.Apply (System.Boolean fromTarget)
BindingExpression+BindingExpressionPart.<PropertyChanged>b__49_0 ()
BindingExpression+BindingExpressionPart.PropertyChanged (System.Object sender, System.ComponentModel.PropertyChangedEventArgs args)
BindingExpression+WeakPropertyChangedProxy.OnPropertyChanged (System.Object sender, System.ComponentModel.PropertyChangedEventArgs e)
(wrapper delegate-invoke) <Module>.invoke_void_object_PropertyChangedEventArgs(object,System.ComponentModel.PropertyChangedEventArgs)
Observable.OnPropertyChanged (System.String propertyName)
Observable.Set[T] (T& storage, T value, System.String propertyName)
TrackingLogic.set_HasTrackPoints (System.Boolean value)
TrackingLogic.<OnTrackCleared>b__84_0 ()
Thread+RunnableImplementor.Run ()
IRunnableInvoker.n_Run (System.IntPtr jnienv, System.IntPtr native__this)
(wrapper dynamic-method) Android.Runtime.DynamicMethodNameCounter.42(intptr,intptr)
```

### Steps to Reproduce

The issue seems to happen when a property is changed (perhaps due to a callback) on a page's viewmodel after the page has been disposed.

### Expected Behavior

The case should be succeed silently and not result in an ObjectDisposedException, since it no longer really matters.

### Actual Behavior

An ObjectDisposedException is thrown back to the user.

### Basic Information

- Version with issue: 4.4.0.991537
- Last known good version: Unknown, but this seems fairly new
- IDE: VS 2019
- Platform Target Frameworks: 
  - Android: 9
- Nuget Packages:
- Affected Devices:

kvpt:
### Description of Change ###
The NavigationPageRenderer doesn't track the removal of an ToolbarItem.
If the user decide to remove an item, the NavigationPageRenderer will rebuild the menu but not unsubscribe PropertyChange EventHandler subscribed in the ToolbarExtensions class.
When the renderer is disposed it try to dispose items by unsubscribe it, but in this case, they were deleted, so they aren't disposed correctly.
To fix that we have to track when item are removed, but we can't do that with the current event CollectionChange on the ToolbarTracker. So I introduced a new collection to store the current toolbar items.

### Issues Resolved ### 
- fixes #9419 

### API Changes ###
 None

### Platforms Affected ### 
- Android

None

Not applicable

### Testing Procedure ###
Run UiTest included in this PR

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Extensions/ToolbarExtensions.cs
+++ b/Xamarin.Forms.Platform.Android/Extensions/ToolbarExtensions.cs
@@ -69,14 +69,13 @@ namespace Xamarin.Forms.Platform.Android
 			item.PropertyChanged -= toolbarItemChanged;
 			item.PropertyChanged += toolbarItemChanged;
 
-			toolbarItemsCreated?.Add(item);
-
 			IMenuItem menuitem;
 
 			if (menuItemIndex == null)
 			{
 				menuitem = menu.Add(new Java.Lang.String(item.Text));
 				menuItemsCreated?.Add(menuitem);
+				toolbarItemsCreated?.Add(item);
 			}
 			else
 			{
--- a/Xamarin.Forms.Platform.Android/Extensions/ToolbarExtensions.cs
+++ b/Xamarin.Forms.Platform.Android/Extensions/ToolbarExtensions.cs
@@ -1,12 +1,11 @@
-﻿using System.ComponentModel;
+using System.ComponentModel;
 using Android.Views;
 using AToolbar = Android.Support.V7.Widget.Toolbar;
-using ATextView = global::Android.Widget.TextView;
+using ATextView = Android.Widget.TextView;
 using Android.Content;
 using Android.Graphics;
 using System.Collections.Generic;
 using System;
-using System.Linq;
 
 namespace Xamarin.Forms.Platform.Android
 {
@@ -28,8 +27,10 @@ namespace Xamarin.Forms.Platform.Android
 			Color? tintColor,
 			PropertyChangedEventHandler toolbarItemChanged,
 			List<IMenuItem> menuItemsCreated,
-			Action<Context, IMenuItem, ToolbarItem> updateMenuItemIcon = null
-			)
+			List<ToolbarItem> toolbarItemsCreated,
+			Action<Context, 
+			IMenuItem, 
+			ToolbarItem> updateMenuItemIcon = null)
 		{
 			if (sortedToolbarItems == null || menuItemsCreated == null)
 				return;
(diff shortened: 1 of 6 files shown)


rmarinho:
Why can't we just change the Event to have the NotifyCollectionChangedEventArgs ? Since this a internal class

kvpt:
It is just too be safe and minimize regression.
This event handler (CollectionChanged) is called when the collection change, but also at others places where the collection not change, PagePropertyChanged and PageChange. It's why NotifyCollectionChangedEventArgs it's not used.
So I preferred to introduce a new event handler just called when the collection really change.
But if the Team decide that it's safe and that we can merge the two, no problem for me.

rmarinho:
yeah i just thing it would be confusing since they are similar and emitted at the same time.
I m not sure we have people using this tracker, but what do you think @samhouts ?

PureWeen:
I think ideally we'd move this logic to ToolbarExtensions

https://github.com/xamarin/Xamarin.Forms/blob/30228a2f3587e1b8c10674076a4a6a64908a1c19/Xamarin.Forms.Platform.Android/Extensions/ToolbarExtensions.cs#L17

Instead of being apart of the tracker because this same issue is going to exist on Shell

As far as the best approach with the tracker :-/ Maybe we just obsolete the old one and use the new one for now

And we'll plan to remove it in the future.

Another option might be to keep track of the last set of ToolBarItems that were passed into UpdateMenuItems

We started keeping track of the IMenuItems so they don't get disposed of
https://github.com/xamarin/Xamarin.Forms/blob/30228a2f3587e1b8c10674076a4a6a64908a1c19/Xamarin.Forms.Platform.Android/Extensions/ToolbarExtensions.cs#L34

we could just retain a local list as well of OldToolBarItems

And then pass it into
https://github.com/xamarin/Xamarin.Forms/blob/30228a2f3587e1b8c10674076a4a6a64908a1c19/Xamarin.Forms.Platform.Android/Extensions/ToolbarExtensions.cs#L57

That might also be better because it keeps everything as part of the same loop..

Any time you modify that list it calls *UpdateMenuItems* to reprocess the list

kvpt:
I think that the use of a static class here is not very suitable because there is a bunch of variables to track or pass between methods, in this case a class similar to the managers class seem more suitable for me.

@PureWeen I choose this solution to not overload the UpdateMenuItem method too much, but you right, it's better to have all the logic at the same place and avoid duplication. 
I will try that.

PureWeen:
The main reason for static is so we don't accrue more managed classes

ToolBarItems are probably simple enough where we could just instance the tracker but statics aren't that tricky to work with

It's a little extra with parameters but nothing unmanageable

kvpt:
@PureWeen It's done.

Xamarin.Forms.Platform.Android/Extensions/ToolbarExtensions.cs
```diff
@@ -40,37 +41,42 @@ public static void DisposeMenuItems(this AToolbar toolbar, IEnumerable<ToolbarIt
...
+			Action<Context,
+			IMenuItem, 
+			ToolbarItem> updateMenuItemIcon = null)
 		{
 			IMenu menu = toolbar.Menu;
 			item.PropertyChanged -= toolbarItemChanged;
 			item.PropertyChanged += toolbarItemChanged;
 
-			IMenuItem menuitem = null;
+			toolbarItemsCreated?.Add(item);
```
PureWeen: I think this change needs to be moved down to

https://github.com/xamarin/Xamarin.Forms/pull/9485/files#diff-b461663043f6182b652f5aeb4b221297R79

If you test with Issue9360 and click to change the icon you'll see toolbarItemsCreated keeps getting items added that it already contains

kvpt: I had a doubt on this one, I did not understand well the menuItemIndex == null.
But now it's logic, it is if the item is in the collection or not.

rmarinho:
@kvpt can you rebase so i can run tests? thanks

kvpt:
@rmarinho It's done.

rmarinho:
I m sorry @kvpt need a new rebase :( there was a error in the script

kvpt:
@rmarinho No problem, done again.

PR #9487, opened 2020-02-08
TPV Additional Samples and event notification changes

PureWeen:
### Description of Change ###
- Added all of the TPV samples into CG so we can easily test changes
- fixed a timing issue on CV https://github.com/xamarin/Xamarin.Forms/compare/4.5.0...tpv?expand=1#diff-a1be28687ab96ea024e59996701f8ff8R26 if the source get swapped out but a measureinvalidated still fires
- Changed code that detects whether renderer is attached and should watch for changes
- Disabled the Hinge Sensor code because we aren't currently exposing any hinge data

### Platforms Affected ### 
- Core/XAML (all platforms)

### Testing Procedure ###
- test through added CG pages

### PR Checklist ###


Xamarin.Forms.DualScreen/DualScreenService.android.cs
```diff
@@ -80,26 +80,47 @@ public static void Init(Activity activity)
 
 				if (_HingeService._hingeSensor != null)
 				{
-					_HingeService._hingeSensor.OnSensorChanged -= _HingeService.OnSensorChanged;
-					_HingeService._hingeSensor.StopListening();
+					//_HingeService._hingeSensor.OnSensorChanged -= _HingeService.OnSensorChanged;
```
jsuarezruiz: Is it causing some kind of problems?, can the refresh rate be reduced perhaps? (SensorDelay).

PureWeen: Yea @Redth is going to look into that

I wonder if this is something we should expose cross platform as well.. Turning listening off and on so it's just not firing off by default.

For now I'm just going to add a getter for the android dll so you can retrieve the sensor

PR #9489, opened 2020-02-09
[Enhancement] Added Resumed and Asleep events to Application

AndreiMisiukevich:
### Description of Change ###
Added **Resumed** and **Asleep** events to Application class.
These properties would help developers of 3rd party libraries to determine the app state. Sometimes it is important to know, that the app is going to background, but currently, it's not possible to get the state in cross-platform code, since there are only methods **OnResume** and **OnSleep** and there are no events/signals there.

### API Changes ###
Added:
 - **event** Resumed; // in Application
 - **event** Asleep; // in Application

### Platforms Affected ### 
- Core

None

Not applicable

### DIFF
--- a/Xamarin.Forms.Core/Application.cs
+++ b/Xamarin.Forms.Core/Application.cs
@@ -167,6 +167,10 @@ namespace Xamarin.Forms
 
 		public event EventHandler<Page> PageDisappearing;
 
+		public event EventHandler Asleep;
+
+		public event EventHandler Resumed;
+
 		async void SaveProperties()
 		{
 			try
@@ -272,12 +276,14 @@ namespace Xamarin.Forms
 		{
 			Current = this;
 			OnResume();
+			Resumed?.Invoke(this, EventArgs.Empty);
 		}
 
 		[EditorBrowsable(EditorBrowsableState.Never)]
 		public void SendSleep()
 		{
 			OnSleep();
+			Asleep?.Invoke(this, EventArgs.Empty);
 			SavePropertiesAsFireAndForget();
 		}
 
@@ -285,6 +291,7 @@ namespace Xamarin.Forms
 		public Task SendSleepAsync()
 		{
 			OnSleep();
+			Asleep?.Invoke(this, EventArgs.Empty);
 			return SavePropertiesAsync();
 		}
 


jamesmontemagno:
How does this differ from OnResume and OnSleep events?

Android has: OnStart, OnResume, OnPause, OnStop, OnDestroy. It is extremely complicated to have so may events IMHO.

AndreiMisiukevich:
Hey @jamesmontemagno thanks for your comment.
It’s good point and I agree with that, but look at my case. I as a 3rd party library developer need to get current status of the application? How can get it?
It isn’t good practice to force users to call methods themselves onResume and onSleep.
Also this approach is more convenient IMHO

mattleibow:
Just want to note that this will probably cause people to leak things. Events usually do, so docs will be required to explain that all things NEED to be unsubscribed.

SittenSpynne:
I'd be happy with this.

I communicate with devices over BLE (and USB on Android). It's important to know when the app has suspended/resumed, so I've had to add a bit of boilerplate to *make* events like this in my App class. 

The main reason for this as a use case for me is I have different behaviors for some pages, so one global suspend/resume method isn't appropriate. For example, what the "acquire data" page needs to do on suspend/resume is dramatically different from the "firmware update" page's needs. One can merely re-initialize the connection and keep going, the other has to sternly inform the user they've done a bad thing and ask them to restart the devices.

AndreiMisiukevich:
@mattleibow hm) I think every developer should understand this :) but anyway you are right.


I am OK with markings these events EditorBrowsable.Never...

haavamoa:
I wouldn't recommend people to have to subscribe / unsubscribe to events. I would much rather go through the current page / viewmodel or every page / viewmodel in my container, look for a interface and Invoke a method that they commonly implement from `OnResume` or `OnSleep`. Having to subscribe / unsubscribe tend to get forgotten, and that is costly.

jamesmontemagno:
Oh, I see these are events and not additional overrides. I would have a name that is the same, however I agree with @mattleibow that this will cause potentially a lot of leaks in apps. 

If I was creating a library that needed the hooks I would have developers add a bit of code to notify my library. This is what we will be doing mostly in Essentials for some things that we can't get to. Essentials also exposes Android lifecycle events which is nifty.

AndreiMisiukevich:
@jamesmontemagno well, then marking those events by `[EditorBrowsable(EditorBrowsableState. Advanced)]` shouldn't be a mistake, should it?
If I consider myself as *advanced user*, I understand that I have to unsubscribe from those events for preventing memory leaks.

PR #9577, opened 2020-02-13
Fix Landscape Calculations and use DualScreenServices in platform code

PureWeen:
### Description of Change ###

- IsLandscape wasn't working correctly on Android when not on a dual screen device
- Pulled interfaces into UWP and Android so the platform code could check for an implementation of IDualScreenService and then use that for reporting back landscape/portrait since those values get reported back incorrectly when spanned
- removed IDisposable from DualScreenService because it's a singleton that never goes away
- Fixed issue where relative contentarea used for region calculation opposed to screen relative position
- Changed DualScreenInfo to take Visualelement so it can work for anything

#### Changed
- SpanModeStateTrigger changed to apply to VisualElement it's attached to
- Added WindowSpanModeStateTrigger which watches to see if the entire window has changed to being spanned
- DualScreenInfo now takes a VisualElement as a constructor argument

### Platforms Affected ### 
- Android
- UWP

### Testing Procedure ###
- test included gallery on single screen device and duo device

### PR Checklist ###


jsuarezruiz:
The new `WindowSpanModeStateTrigger` looks good.

PR #9615, opened 2020-02-17
Propagate cleared shell elements correctly

PureWeen:
### Description of Change ###
When clearing out shell collections the removal of those elements wasn't being propagated out correctly

### Platforms Affected ### 
- Core/XAML (all platforms)

### Testing Procedure ###
- unit tests included

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Shell/ShellSectionCollection.cs
+++ b/Xamarin.Forms.Core/Shell/ShellSectionCollection.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Collections.Specialized;
+using System.Linq;
 
 namespace Xamarin.Forms
 {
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Core/Shell/ShellSectionCollection.cs
```diff
@@ -119,7 +123,13 @@ void CheckVisibility(ShellSection section)
 
 		public void Add(ShellSection item) => Inner.Add(item);
 
-		public void Clear() => Inner.Clear();
+		public void Clear()
```
felipebaltazar: This logic is duplicated in `ShellItemCollection.cs`,  maybe we can create a base class with this code (and remove logic)
What do you think? 😊

PureWeen: It's probably about that time :-)
But I'll leave that for the next PR for now

There are type specifics that will probably need a few generics and probably some generic interfaces as well.

All good stuff but something better left for a later branch for now

PureWeen:
ios12 test failure unrelated

Xamarin.Forms.Core/Shell/ShellSectionCollection.cs
felipebaltazar: Maybe I can help with this, I've created a suggestion for this change.
What do you think?

https://github.com/xamarin/Xamarin.Forms/pull/9626

PR #9633, opened 2020-02-19
added InvokeOnMainThreadIfRequired

AndreiMisiukevich:
### API Changes ###
Added:
 - void InvokeOnMainThreadIfRequired (Action action) in **Device**

### Platforms Affected ### 
- Core

None

Not applicable

### DIFF
--- a/Xamarin.Forms.Core/Device.cs
+++ b/Xamarin.Forms.Core/Device.cs
@@ -104,6 +104,16 @@ namespace Xamarin.Forms
 			PlatformServices.BeginInvokeOnMainThread(action);
 		}
 
+		public static void InvokeOnMainThreadIfRequired(Action action)
+		{
+			if (IsInvokeRequired)
+			{
+				BeginInvokeOnMainThread(action);
+				return;
+			}
+			action.Invoke();
+		}
+
 		public static Task<T> InvokeOnMainThreadAsync<T>(Func<T> func)
 		{
 			var tcs = new TaskCompletionSource<T>();


jamesmontemagno:
I think that the method should just do the right thing and if on main thread then execute, else schedule. This is what Essentials does. If for some reason there is back compat issues then I would pass a book into the current method. No need for a new method IMHO

hartez:
> I think that the method should just do the right thing and if on main thread then execute, else schedule. This is what Essentials does. If for some reason there is back compat issues then I would pass a book into the current method. No need for a new method IMHO

Are you suggesting changing the behavior of `BeginInvokeOnMainThread`? That would _definitely_ cause some backward compatibility issues. 

Also, until we implement a separate layout scheduler, that change would grind complex layouts to a halt.

brminnick:
I agree with @jamesmontemagno that this should be implemented into `Device.BeginInvokeOnMainThread` and `Device.InvokeOnMainThreadAsync`.

As for breaking changes, these are the ones I foresee (albeit insignificant) for adding this functionality into the existing APIs:
- `await Device.InvokeOnMainThreadAsync(...)` would not change threads when already on the Main Thread
    - I.e. If a non-Main Thread, e.g. Thread 4, calls `InvokeOnMainThreadAsync`, it will yield to the Main Thread, e.g. Thread 1, until the Main Thread has finished executing
- `Device.BeginInvokeOnMainThread(...)` would run in-line rather when already on the main thread, rather than be scheduled
    - I.e. All code written on the lines preceding  `Device.BeginInvokeOnMainThread(...)` would run immediately after its `Action`, instead of its `Action` being scheduled back onto the main thread

 But since the behavior we want from `InvokeOnMainThreadAsync` and `BeginInvokeOnMainThread` is to run its `Action` on the Main Thread, it still accomplishes the original goal _and_ we get a performance gain by avoiding a context switch.

Does Xamarin.Forms has an equivalent to `Xamarin.Essentials.MainThread.IsMainThread`? I like [`Xamarin.Essentials`' implementation](https://github.com/xamarin/Essentials/blob/d73fb42418566a45e4bd3fc0e425bb2b382c38c0/Xamarin.Essentials/MainThread/MainThread.shared.cs#L12-L129):

```csharp
//Xamarin.Essentials Implementation
public static void BeginInvokeOnMainThread(Action action)
{
    if (IsMainThread)
    {
        action();
    }
    else
    {
        PlatformBeginInvokeOnMainThread(action);
    }
}
```

brminnick:
On a side, note - I had no idea that Xamarin.Essentials had this [baked-in until reading James' comment](https://github.com/xamarin/Xamarin.Forms/pull/9633#issuecomment-588346611) then [digging into its implementation](https://github.com/xamarin/Essentials/blob/d73fb42418566a45e4bd3fc0e425bb2b382c38c0/Xamarin.Essentials/MainThread/MainThread.shared.cs#L12-L129) - love it! 

I had been mixing `Xamarin.Forms.Device.InvokeOnMainThreadAsync` and `Xamarin.Essentials.MainThread.IsMainThread` to get this performance boost, myself: 
https://github.com/brminnick/GitTrends/blob/995a1c271533f814e5d7e837b6a911e2e44dbc4c/GitTrends/Services/BaseMobileApiService.cs#L28-L54

I'll be switching my apps over to `Xamarin.Essentials.MainThread.InvokeOnMainThread` to get this performance boost (if/until this gets folded into Xamarin.Forms 😊).

AndreiMisiukevich:
@brminnick @jamesmontemagno I decided to expose another method due to one reason: Current implementation always makes a delay before running the action. So probably some systems expect such behavior.

Please take a look at this example (I agree, it doesn't make sense, but just for example)
```csharp
SomeObject obj = new SomeObject();

void MakeAction() {
	BeginInvokeOnMainThread(() => {
		obj = null;
	})
	obj.MakeTheLastActionBeforeLoosingReference();
}
```

jamesmontemagno:
So, I would recommend 2 options:

1. BeginInvokeOnMainThread should check if on main thread, if so execute, else schedule. 

This is a change in functionality, however I don't know how it could possibly cause issues. So, the other option:

2.) Add a boolean to BeginInvokeOnMainThread for "checkMainThreadFirst" or something like that that defaults to false so no change in functionality at all.

In essentials we decided to always check the main thread. I would recommend not adding yet another method that does something slightly different, there are already a lot.

AndreiMisiukevich:
@jamesmontemagno I'd recommend the 3rd option

3. Mark `BeginInvokeOnMainThread` **Obsolete** and expose new method `InvokeOnMainThread` or `BeginInvokeOnUiThread`.

It will allow avoiding redundant parameters like checkMainThreadFirst and also won't break backward compatibility.

AndreiMisiukevich:
> 2.) Add a boolean to BeginInvokeOnMainThread for "checkMainThreadFirst" or something like that that defaults to false so no change in functionality at all.

Btw it would be breaking change. If somebody uses Device.BeginInvokeOnMainThread method group as parameter ```DoSomething(Device.BeginInvokeOnMainThread);```
It this case better solution would create a new overload with a parameter, but I am still don't like this approach so much.

hartez:
> Does Xamarin.Forms has an equivalent to Xamarin.Essentials.MainThread.IsMainThread? I like

Forms has `IsInvokeRequired`. Which is the inverse of Essentials's `IsMainThread`.

hartez:
Closing this for now. As has been noted, Essentials already provides a helper that does this; if Forms ends up taking a dependency on Essentials then this would become another conflict. And if users want this method without using Essentials, it's possible to do in an extension method. 

As we revise how we handle dispatching in future versions, we may resurrect this PR.

Issue #5322, opened 2019-02-20
The `rate` parameter in IAnimatable.Animate() method doesn't take effect


yanshouwang:
### Description

The `rate` parameter in IAnimatable.Animate(string name, Animation animation, uint rate = 16, uint length = 250, Easing easing = null, Action<double, bool> finished = null, Func<bool> repeat = null) method doesn't take effect.
I have read Xamarin.Forms' source code at [AnimationExtensions.cs](https://github.com/xamarin/Xamarin.Forms/blob/bd31e1e9fc8b2f9ad94cc99e0c7ab058174821f3/Xamarin.Forms.Core/AnimationExtensions.cs) and only the Info Class has used the `rate` parameter but do nothing
it seems that Xamarin.Forms used the system ticker to implement animation on each platform( ValueAnimator on Android and CADisplayLink on iOS) so the `rate` parameter has no effect to the animation...
I don't know if I am right about this, but this `rate` parameter really confused me for a long time because the docs explains it as followed:

> rate UInt32 
The time, in milliseconds, between frames.

### Steps to Reproduce

1. Add an ProgressBar and use the Animate() method with `rate` set to 20.
2. Run the app and then change the `rate` to 2000.
3. Run the app again and there is no difference with the animation's rate.

### Expected Behavior

 The animation's rate changes when the `rate` parameter has changed.

### Actual Behavior

 The animation's rate doesn't change when the `rate` parameter has changed.

### Basic Information

- Version with issue: Test on 3.5.0.129452
- Last known good version: NULL
- IDE: VS 2017
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Android Support Library Version: 
- Nuget Packages:
- Affected Devices:

samhouts:
@yanshouwang Can you please attach your sample project? Thanks!

yanshouwang:
@samhouts OK, I‘ve done here: [Confusions.Forms](https://github.com/yanshouwang/Confusions)

gitcoinbot:
Issue Status: 1. Open 2. Started 3. Submitted 4. **Done** 

<hr>

__This issue now has a funding of 480.0 DAI (480.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/xamarin/Xamarin.Forms/5322/4081).
[...]

PR #9649, opened 2020-02-20
Add support for scale and rotation of images on GTK

### LINKED ISSUE #8732: [Bug] Image animation not working in GTK
### Description
It seems like image animation is not working in GTK. Similar code worked on the UWP backend.

### Steps to Reproduce
https://github.com/ysangkok/XamarinGTKAnimationIssue

[src.zip](https://github.com/xamarin/Xamarin.Forms/files/3935547/src.zip)

### Expected Behavior
Image should be animated.

### Actual Behavior
Image is not animated.

### Basic Information

- Version with issue: 4.4.0.991210-pre2
- Last known good version: don't know
- Platform Target Frameworks: Ubuntu 18.04 with gtk-sharp 2.12.45-0xamarin17+ubuntu1804
- Nuget Packages: See linked repository

Bobface:
### Description of Change ###

This PR adds support for rotation and scaling of the Image Control  on the GTK Platform.

### Issues Resolved ### 

fixes #8732 

### Platforms Affected ### 
- GTK

Below is a GIF of a rotating and scaling image as per the issue above. 

_The GIF is a bit wonky because of low capture framerate. In reality the animation is smooth._

Gif

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.GTK/Renderers/ImageRenderer.cs
+++ b/Xamarin.Forms.Platform.GTK/Renderers/ImageRenderer.cs
@@ -1,4 +1,4 @@
-﻿using Gdk;
+using Gdk;
 using System;
 using System.ComponentModel;
 using System.Drawing;
@@ -49,6 +49,9 @@ namespace Xamarin.Forms.Platform.GTK.Renderers
 				SetImage(e.OldElement);
 				SetAspect();
 				SetOpacity();
+				SetScaleX();
+				SetScaleY();
+				SetRotation();
 			}
 
 			base.OnElementChanged(e);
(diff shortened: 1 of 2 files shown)


knocte:
@ysangkok can you test this please?

Xamarin.Forms.Platform.GTK/Renderers/ImageRenderer.cs
```diff
@@ -64,6 +67,14 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
 				SetOpacity();
 			else if (e.PropertyName == Image.AspectProperty.PropertyName)
 				SetAspect();
+			else if (e.PropertyName == Image.ScaleProperty.PropertyName)
+				SetScale();
```
jsuarezruiz: Thanks for the contribution. Mnn, could we move this implementation to `VisualElementTracker` https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.GTK/VisualElementTracker.cs#L279?. In this way, the transformation can be applied to any visual element.

Bobface: The implementation is specifically designed for rotation and scaling of images and uses Cairo etc. to do so. I don't think it's easily portable to support any visual element.

knocte: @jsuarezruiz ping? I think Bobface raises a good point here. BTW, issue #5322 has been fixed in master branch (4.7.x?), which was related to this.

knocte:
@hartez can you advise on what's the next step here?

Xamarin.Forms.Platform.GTK/Renderers/ImageRenderer.cs
jsuarezruiz: Ideally this should be managed by `VisualElementTracker` for all the Views. I understand that the implementation is not easy for everything, if we allow this it would be an exception (manage transformations of a specific view like an Image in the view renderer). Otherwise, it works correctly.

PR #9657, opened 2020-02-20
Remove unreliable scroll from TestItemsPosition

hartez:
These changes add more information to the TestItemsPosition method when a test fails. They also remove the scrolling from the test to make it more reliable.


hartez:
The KeepItemsInView test failure is unrelated, and being worked on elsewhere.

Issue #8848, opened 2019-12-11
a11y: Voiceover position not resetted when invoked from tab selection


jantanis:
### Description
When I switch tabs I want the voiceover to start at the top of the page. I can produce this "reset" function from a button, but it won't work from a tab somehow.

### Steps to Reproduce

1. Open A11y.sln
2. Start iOS project on device (not simulator)
3. Enable voiceover
4. Select button
5. Double tap button -> Voiceover focus gets resetted to top item
6. Switch tab to page 2 and back to page 1 -> Voiceover does not get resetted

### Expected Behavior
When invoking UIAccessibility.PostNotification(UIAccessibilityPostNotification.ScreenChanged, null) voiceover focus gets resetted

### Actual Behavior
Does not work when invoked from tab

### Basic Information

- Version with issue: 4.3.0.991211
- Last known good version: n/a
- IDE: VS for Mac 8.3.10.2
- Platform Target Frameworks: 
  - iOS:  11.2
- Android Support Library Version: 
- Nuget Packages:
- Affected Devices: iPhone with iOS 11

### Screenshots
![ezgif com-video-to-gif](https://user-images.githubusercontent.com/11005949/70615201-43267300-1c0c-11ea-9f68-cd7dc7bcaaf1.gif)

### Reproduction Link
[A11Y.zip](https://github.com/xamarin/Xamarin.Forms/files/3949924/A11Y.zip)

PureWeen:
I'm going to close this one for now

@rachelkang and I played with this one for a bit and the current behavior is all expected from native iOS

Xamarin.Forms itself isn't getting in the way here at all.

I've attached a sample Xamarin.iOS native app where you can see the same behavior
[AcessTest.zip](https://github.com/xamarin/Xamarin.Forms/files/5426920/AcessTest.zip)

If you can elaborate on the type of accessibility requirement you are trying to meet that would help us understand the limitation here and possibly add some additional APIs to help out. If you're trying to select the content after selecting a new tab then you're going to need to work against the default a11y settings of iOS to accomplish this

Issue #7568, opened 2019-09-18
[Bug] Xamarin.Forms TabIndex not working on Android and iOS


davantigroup:
### Sample

https://github.com/xamarin/xamarin-forms-samples/tree/master/UserInterface/Accessibility

### Description

It seems the TabIndex is not working anymore, even in the sample provided by Microsoft regarding to Keyboard Accessibility it only focuses the first button, it works properly for entry or stepper tough.

### Steps to Reproduce

1. Open the sample project
2. Go to Tab Order
3. Press Tab multiple times
4. When the focus is on the first button press Tab again

### Expected Behavior

The focus should move to the next button

### Actual Behavior

The focus is stuck on the first button

### Information
  - Platform with the issue: 
  - iOS:  12.4
  - Android:  8.1, 9.0
![image](https://user-images.githubusercontent.com/54096887/65041058-88fdd300-d95e-11e9-8f8f-6178a0864db9.png)

rnic99:
@samhouts Id like to add some extra to this bug, in the same accessibility app we have two different pages, "Automation Properties" and "Tab Order".  The bug was originally reported for TabIndex not working on the sample app. Testing the accessibility app in XF 4.3 I was not having any issues navigating on the Tab Order sample but I have encountered an interesting bug in the accessibility demo portion of the app, and its very similar to the one described here. When trying to navigate using a screen reader I have found that I cannot navigate past the "Enter your name:" label when its wrapped in a stack layout. However when I removed the stack layout i could freely navigate past this control. Ive found that wrapping other controls in stack layouts either perform identical or similar issues in regards to how the screen reader behaves with with the ui elements. 
However I have found that wrapping multiple elements in stack layouts at once also produces different results.

Not entirely sure if this warrants another bug or similar enough to this one. I can easily create another bug if need be.

samhouts:
@rnic99 Yeah, I would open a new bug for that. Thanks!!!

rnic99:
@samhouts [https://github.com/xamarin/Xamarin.Forms/issues/8613](url)

samhouts:
@davantigroup It looks like TabIndex behaves the same on iOS no matter what version I'm using. Are you really seeing this issue on iOS, too? Thanks!

davantigroup:
@samhouts  We had the same issue on iOS too, we can test it again if needed

samhouts:
Please do! If you can post a gif or video, too, that would be immensely helpful! Thanks!!

[...]

PR #9702, opened 2020-02-25
[iOS|Android] Fix screen reader order

### LINKED ISSUE #4220: iOS VoiceOver/Android TalkBack reads elements on Master menu page when scrolling over elements on Detail page
### Description

With a MasterDetail page whose master page is a menu in a ListView, and whose Detail page uses a StackLayout with default options (as in the MasterDetailPage template project) , when scrolling over items in the Detail page's StackLayout, the iOS VoiceOver reader will read out menu items from the Master menu page even if it is not showing. This occurs on iPhone X but not on iPad Air 2. 

### Steps to Reproduce

1. Open Visual Studio for Mac and create a MasterDetailPage template solution
2. Open ItemsPage.xaml and add a Label above the ListView, e.g.:

```
      <StackLayout>
          <Label Text="Header 1" TextColor="Red" />
          <ListView  ...>
          </ListView>
      </StackLayout>
```

3. Launch app to iPhone device (easier to use VoiceOver accessibility function on device)
4. Open Settings app and go to **General > Accessibility > VoiceOver** and turn on the VoiceOver feature. You will now have to tap an item once to select it and then double tap the item to execute it. 
5. Go back to the Forms MasterDetail app and swipe over the elements on  the ItemsPage

### Expected Behavior

I should hear the names of the elements I scroll over and I should not hear the names of any non-visible elements.

### Actual Behavior

As I scroll between "Header 1" Label and the top of the ListView, I hear "About" and see a box around and empty element between "Header 1" Label and the ListView. 

## Notes

I confirmed that what is being read is the "About" menu item in the Master page because removing the "About" menu item resolves the issue. 

It seemed to me that there must be a space between the label and the ListView such that iOS voiceover is seeing the hidden "About" menu item, and indeed I see the default spacing for a Stack Layout is 6. Setting the StackLayout.Spacing to 0 works around the issue. 

    <StackLayout Spacing="0">

## Workaround

Set the Spacing for the StackLayout to 0

### Basic Information

- Version with issue: 3.1.0.697729 and 3.3.0.912540
- IDE:  Visual Studio for Mac
- Platform Target Frameworks: iOS
  - iOS:  12
  - Android: N/A 
  - UWP:  N/A 

- Affected Devices: iPhone (not iPad)

### LINKED ISSUE #4270: [Android] AutomationProperties.IsInAccessibleTree is ignored
### Description

 TalkBack is the accessibility mode for Android. When using talkback, users can swipe to navigate similar to what tab would do in a webpage, that is, it causes focus to move from one item to the next. AutomationProperties.IsInAccessibleTree is set to false, we expect that swiping gesture would cause focus to move to the next item in the accessibility tree. However, focus is moved to controls (switch, button, etc.) where AutomationProperties.IsInAccessibleTree is set to false.

### Steps to Reproduce

1. Enable TalkBack. Ensure the swiping gesture is allowing you to move between items in accessibility tree.
2. Add this to any layout in XF, where this layout is included and you will see repro 100% of time: (I created a new project and added this to the starting page)
``` xaml
<Switch AutomationProperties.IsInAccessibleTree="False"/>
<Button AutomationProperties.IsInAccessibleTree="False"/>
```
3. Run the app on Android, visit the page where the controls were added. Begin swipe navigation to move focus from one control to the next. 

Also, `Label` is crazy. It is generally always skipped unless you explicitly set `IsInAccessibleTree` to true; However, if the `Label` is the first element on the page, it'll be read even if you explicitly set `IsInAccessibleTree` to false.

### Expected Behavior

focus is not moved to controls where  IsInAccessibleTree is set to false.

### Actual Behavior

focus does move to controls where IsInAccessibleTree is set to false.

### Basic Information

- Version with issue:
- Last known good version:
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Nuget Packages:
- Affected Devices:

### Reproduction Link

[707192](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/707192)

### LINKED ISSUE #5879: [Bug]  Tab index not resetting on iPAD
### Description
https://github.com/xamarin/Xamarin.Forms/pull/3989#issuecomment-477360686

### Steps to Reproduce

1.  run control gallery
2. Accessibility Tab Index
3. on an iPad if you swipe to the end it stops on "14" and won't go any further. @samhouts tested this on a device and on the device it works fine

### Expected Behavior
Expectations are that iOS/Android will incorrectly read the master page contents first (#4220), then BONG that you're at the end of the elements; wait a second, and then it will allow you to start from the beginning again.

### Actual Behavior
iPad is not allowing you to start from the beginning

### Basic Information

- Version with issue: 4.0 pre9
- Platform Target Frameworks: 
  - iOS:  12
- Affected Devices: IPAD

### LINKED ISSUE #7606: [Bug] When a view appears it is not accessible via VoiceOver
### Description

If a previously hidden view is toggled to visible, it can not be accessed via VoiceOver

This appears to be linked to https://github.com/xamarin/Xamarin.Forms/issues/6929

### Steps to Reproduce
- Open the app on an iOS device
- Turn voice over on
- Toggle a previously hidden view to visible

### Expected Behavior

- When the view appears, it is reachable by swiping left/right on voice over

### Actual Behavior

- When the view appears, it is not reachable by swiping left/right on voice over

### Basic Information

- Version with issue: 4.2.0.815419
- Last known good version:
- IDE: Visual studio professional (both windows and mac)
- Platform Target Frameworks: 
  - iOS:  12.4 
- Affected Devices:
  - Observed on iPhone 6S

### Reproduction Link

The following sample project contains a demonstration of the issue:
https://github.com/boyd-m/VisibilityA11y

### LINKED ISSUE #8613: [Bug] Accessibility, screenreader ignores or skips items in nested stacklayout
### Sample 
https://github.com/xamarin/xamarin-forms-samples/tree/master/UserInterface/Accessibility

### Description
Encountered an issue on an app I was working on where our custom entry fields were being skipped when swiping over them when using a screen reader. Its almost as if these items were put at the bottom of the accessibility tree, because if you continue to navigate down the page in question eventually you will be navigated to the controls that were skipped. Given It was a custom control I thought it was something on our end, but I was able to reproduce the issue on a Xamarin forms sample app. This issue occurs on both Android and iOS however their behavior is different from each other.

### Steps to Reproduce
1. Load the Accessibility sample app/Turn on Screen reader.  
2. Navigate to the app and click on the automation properties button .
3. While on the accessibility demo page swipe from left to right to navigate down the page. 
4. On Android the ("Enter your name label, and Enter name here entry field)" are skipped)/  iOS your are stopped from navigating further into the page at the ( Enter name here entry field). 

### Expected Behavior
Screen readers should be able to navigate from left to right and then down unless a tab index has been set or the item in question has been removed from the accessibility tree. 

### Actual Behavior
Screen readers are skipping or stopping at certain elements within nested stack layouts.
### Basic Information

- Version with issue: 4.0 and up 
- Last known good version: 3.6 
- IDE: VS for mac
- iOS:  13.1.2, 12.1.3
- Android: 10, 9
- Affected Devices: Pixel 2 XL, OnePlus 3T, Ipad (6th gen), iPhone 6.

### Screenshots
Android
One swipe on Android skips the ui elements

Screenshot_20191121-100056

Screenshot_20191121-100105
 
iOS

Adding AutomationProperties.IsInAccessibleTree="true" fixes the issue somewhat for iOS, this allows you to navigate past the entry field however it groups it together. If you continue to swipe down the page it will eventually move back to the fields individually. 

IMG_0011

This is how the screen reader handles the app without any changes I cannot navigate past the entry field.

IMG_0012

### Additional Information 
Removing the stack layout surrounding the label and entry will resolve this issue on both Android and iOS.

### LINKED ISSUE #8691: A11y: TabIndex is ignored for first element on page for VoiceOver
### Description
When setting all TabIndexes on the page. The first element on the page receives VoiceOver focus. Not the one with the lowest TabIndex. (Will start on NavigationBar if it is present.)

### Steps to Reproduce
1. Open attached solution
2. Run
3. Enable VoiceOver

### Expected Behavior
Voiceover starts on the element with the lowest TabIndex.

### Actual Behavior
The element on top of the page receives VoiceOver focus.

### Basic Information
- IDE: VS MAC 8.3.10.2
- Platform Target Frameworks: 
  - iOS:  11.2
- Affected Devices: iOS

### Reproduction Link
[A11Y.zip](https://github.com/xamarin/Xamarin.Forms/files/3897211/A11Y.zip)

### LINKED ISSUE #9137: A11y: Image in a11y tree stops voiceover from hopping to the next element
### Description
While working on A11y, I've found this issue: having an image in a11y tree with TabIndexes set (for A11y). Voiceover does not get past this image if I put it somewhere in between. If I put it at the end it works fine.

### Steps to Reproduce
1. Open and run A11y.sln on iOS. 
2. Start voiceover
3. Scroll through elements, all is well
4. Change TabIndex of Image that has TabIndex set to 80 to 20
5. Restart app
6. Scroll through elements, you can't get past this image

### Expected Behavior
Voiceover able to scroll to next element (with TabIndex 30)

### Actual Behavior
Voiceover stops on Image with TabIndex 20.

### Basic Information
- IDE: VS Mac 8.3.11.1
- Platform Target Frameworks: 
  - iOS:  11.2
- Affected Devices: iOS

### Reproduction Link
[A11Y.zip](https://github.com/xamarin/Xamarin.Forms/files/4039228/A11Y.zip)

### LINKED ISSUE #9191: [Bug] If users don't specify any TabIndexes or TabStops then just let the native OS do it's thing
### Description

By default iOS/Android to a fairly good job articulating TabIndex orders based on view orders and the changes in 4.0 broke this implicit behavior. 

We need to look at just letting the operating system do its thing if the user hasn't opted into any TabIndex or TabStop behaviors

### Steps to Reproduce

1.  Compare default TabIndex behavior on a 3.6 app to a 4.0 app

### Expected Behavior
both apps should act the same

### Actual Behavior
4.0 is too heavy handed and replaces default behavior

### Basic Information

- Version with issue: 4.0
- Last known good version: 3.6
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   

### Workaround

- On iOS the workaround has been to override ViewLoad and return a different view then PageContainer

### LINKED ISSUE #9653: [Bug] Accessibility Show Focus Traps or Out-Of-Order Reading
### Description
For both Android and iOS, we are experiencing multiple types of accessibility issues. Issues include out-of-order reading in which a view will be skipped and read after view later in the hierarchy. In iOS in particular, we are seeing focus traps in which voice over navigation cannot advance passed or before a collection of views. We have provided a sample project to demonstrate this as well as videos showing the issues we are seeing. 

### Steps to Reproduce

1. Create a new forms project
2. Create a static layout normally containing two orthogonal stack layouts (although any view group will work)
3. Navigating through the view hierarchy will trigger either a focus trap or view skipping

### Expected Behavior
Voiceover should be able to navigate views in a sane linear method. For us this is top to bottom, left to right.

### Actual Behavior
Views are either skipped or navigation gets stuck in a sub view group and voiceover is incapable of leaving the view without resorting to untypical navigation methods.

### Basic Information

- Version with issue: 4.5.0.142-pre as well as 4.4
- IDE: Rider and Visual Studio Mac
- Platform Target Frameworks: Android, iOS
  - iOS:  11.3
  - Android: 9 and 10
- Affected Devices:

### Screenshots

A test video is attached in to the sample project

### Workaround

None known.

[AccessibilityTestForms.zip](https://github.com/xamarin/Xamarin.Forms/files/4232086/AccessibilityTestForms.zip)

samhouts:
### Description of Change ###

* iOS MasterDetailPage will no longer read MasterPage contents when the master page is hidden and will no longer read DetailPage contents when the master page is visible.
* iOS can now tab stop on a Frame and can read its contents properly.
* iOS/Android screen readers should no longer skip over layouts and should read all children in declaration order and/or TabIndex order.
* TabIndex should no longer return elements that are `Enabled = false` or `IsAccessibilityElement = false`.

### Issues Resolved ### 

- fixes #9653
- fixes #4220 
- fixes #9191 
- fixes #8613
- fixes #9137
- fixes #8691 
- fixes #7606
- fixes #4270
- fixes #5879

### API Changes ###
 
These are breaking API changes, but they drive me crazy. I will revert if someone has an issue with this.

`TabIndexExtensions`:
```diff
-public static SortedDictionary<int, List<ITabStopElement>> GetSortedTabIndexesOnParentPage(this VisualElement element, out int countChildrensWithTabStopWithoutThis)
+public static SortedDictionary<int, List<ITabStopElement>> GetSortedTabIndexesOnParentPage(this VisualElement element)


-public static IDictionary<int, List<ITabStopElement>> GetTabIndexesOnParentPage(this ITabStopElement element, out int countChildrensWithTabStopWithoutThis, bool checkContainsElement = true)
+public static IDictionary<int, List<ITabStopElement>> GetTabIndexesOnParentPage(this ITabStopElement element, out int countChildrenWithTabStopWithoutThis)
```

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### Testing Procedure ###

1. Use TalkBack/VoiceOver to test the Control Gallery and make sure everything is read in an expected order.
2. Use keyboard to test that TabIndex is still functioning as expected.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
@@ -173,6 +173,10 @@ namespace Xamarin.Forms.Platform.Android
 			if (tabIndexes == null)
 				return base.FocusSearch(focused, direction);
 
+			// use OS default--there's no need for us to keep going if there's one or fewer tab indexes!
+			if (tabIndexes.Count <= 1)
+				return base.FocusSearch(focused, direction);
+
 			int tabIndex = element.TabIndex;
 			AView control = null;
 			int attempt = 0;
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/A11yTabIndex.xaml
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/A11yTabIndex.xaml
@@ -42,14 +42,14 @@
 
         <Button Text="Skip" AutomationProperties.IsInAccessibleTree="False" TabIndex="3" Grid.Row="3"/>
 
-        <Frame BackgroundColor="Cornsilk" Margin="20" CornerRadius="2" BorderColor="Gray" Grid.Row="4">
+        <Frame BackgroundColor="Cornsilk" Margin="20" CornerRadius="2" BorderColor="Gray" Grid.Row="4"  AutomationProperties.IsInAccessibleTree="False">
             <StackLayout>
                 <Button Text="Twelve" TabIndex="3"/>
                 <Button Text="Ten" TabIndex="2"/>
             </StackLayout>
         </Frame>
 
-        <Frame BackgroundColor="Cornsilk" Margin="20" CornerRadius="2" BorderColor="Gray" Grid.Row="5">
+        <Frame BackgroundColor="Cornsilk" Margin="20" CornerRadius="2" BorderColor="Gray" Grid.Row="5"  AutomationProperties.IsInAccessibleTree="False">
             <StackLayout>
                 <StackLayout>
                     <Button Text="Six" />
--- a/Xamarin.Forms.Controls/CoreGallery.cs
+++ b/Xamarin.Forms.Controls/CoreGallery.cs
@@ -286,6 +286,7 @@ namespace Xamarin.Forms.Controls
 		}
 
 		List<GalleryPageFactory> _pages = new List<GalleryPageFactory> {
+				new GalleryPageFactory(() => new TabIndexTest.TabIndex(), "Accessibility TabIndex (2)"),
 				new GalleryPageFactory(() => new PlatformTestsConsole(), "Platform Automated Tests"),
 				new GalleryPageFactory(() => new MemoryLeakGallery(), "Memory Leak"),
 				new GalleryPageFactory(() => new Issues.A11yTabIndex(), "Accessibility TabIndex"),
--- a/Xamarin.Forms.Controls/TabIndexTest/DayView.xaml
+++ b/Xamarin.Forms.Controls/TabIndexTest/DayView.xaml
@@ -60,37 +60,37 @@
                 <Frame.Triggers>
                     <DataTrigger TargetType="Frame" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Sunday}">
                         <Setter Property="AutomationProperties.Name" Value="Sunday"/>
-                        <Setter Property="TabIndex" Value="1"/>
+                        <!--<Setter Property="TabIndex" Value="1"/>-->
                     </DataTrigger>
 
                     <DataTrigger TargetType="Frame" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Monday}">
                         <Setter Property="AutomationProperties.Name" Value="Monday"/>
-                        <Setter Property="TabIndex" Value="2"/>
+                        <!--<Setter Property="TabIndex" Value="2"/>-->
                     </DataTrigger>
 
                     <DataTrigger TargetType="Frame" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Tuesday}">
                         <Setter Property="AutomationProperties.Name" Value="Tuesday"/>
-                        <Setter Property="TabIndex" Value="3"/>
+                        <!--<Setter Property="TabIndex" Value="3"/>-->
                     </DataTrigger>
 
                     <DataTrigger TargetType="Frame" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Wednesday}">
                         <Setter Property="AutomationProperties.Name" Value="Wednesday"/>
-                        <Setter Property="TabIndex" Value="4"/>
+                        <!--<Setter Property="TabIndex" Value="4"/>-->
                     </DataTrigger>
 
                     <DataTrigger TargetType="Frame" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Thursday}">
                         <Setter Property="AutomationProperties.Name" Value="Thursday"/>
-                        <Setter Property="TabIndex" Value="5"/>
+                        <!--<Setter Property="TabIndex" Value="5"/>-->
                     </DataTrigger>
 
                     <DataTrigger TargetType="Frame" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Friday}">
                         <Setter Property="AutomationProperties.Name" Value="Friday"/>
-                        <Setter Property="TabIndex" Value="6"/>
+                        <!--<Setter Property="TabIndex" Value="6"/>-->
                     </DataTrigger>
 
                     <DataTrigger TargetType="Frame" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Saturday}">
                         <Setter Property="AutomationProperties.Name" Value="Saturday"/>
-                        <Setter Property="TabIndex" Value="7"/>
+                        <!--<Setter Property="TabIndex" Value="7"/>-->
                     </DataTrigger>
 
                     <DataTrigger TargetType="Frame" Binding="{Binding IsSelected}" Value="True">
--- a/Xamarin.Forms.Controls/TabIndexTest/DayView.xaml
+++ b/Xamarin.Forms.Controls/TabIndexTest/DayView.xaml
@@ -10,7 +10,7 @@
   <ContentView.Content>
         <StackLayout BindingContext="{x:Reference this}"
                      HorizontalOptions="FillAndExpand"
-                     AutomationProperties.IsInAccessibleTree="False" IsTabStop="False">
+                     AutomationProperties.IsInAccessibleTree="False">
             <Frame Margin="0"
                    Padding="0"
                    HasShadow="False"
@@ -25,7 +25,7 @@
                        HorizontalOptions="FillAndExpand"
                        VerticalOptions="FillAndExpand"
                        VerticalTextAlignment="Center"
-                       HorizontalTextAlignment="Center" IsTabStop="False">
+                       HorizontalTextAlignment="Center">
                     <Label.Triggers>
                         <DataTrigger TargetType="Label" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Sunday}">
                             <Setter Property="Text" Value="S"/>
@@ -60,30 +60,37 @@
                 <Frame.Triggers>
                     <DataTrigger TargetType="Frame" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Sunday}">
                         <Setter Property="AutomationProperties.Name" Value="Sunday"/>
+                        <Setter Property="TabIndex" Value="1"/>
                     </DataTrigger>
 
                     <DataTrigger TargetType="Frame" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Monday}">
                         <Setter Property="AutomationProperties.Name" Value="Monday"/>
+                        <Setter Property="TabIndex" Value="2"/>
                     </DataTrigger>
 
                     <DataTrigger TargetType="Frame" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Tuesday}">
                         <Setter Property="AutomationProperties.Name" Value="Tuesday"/>
+                        <Setter Property="TabIndex" Value="3"/>
                     </DataTrigger>
 
                     <DataTrigger TargetType="Frame" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Wednesday}">
                         <Setter Property="AutomationProperties.Name" Value="Wednesday"/>
+                        <Setter Property="TabIndex" Value="4"/>
                     </DataTrigger>
 
                     <DataTrigger TargetType="Frame" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Thursday}">
                         <Setter Property="AutomationProperties.Name" Value="Thursday"/>
+                        <Setter Property="TabIndex" Value="5"/>
                     </DataTrigger>
 
                     <DataTrigger TargetType="Frame" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Friday}">
                         <Setter Property="AutomationProperties.Name" Value="Friday"/>
+                        <Setter Property="TabIndex" Value="6"/>
                     </DataTrigger>
 
                     <DataTrigger TargetType="Frame" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Saturday}">
                         <Setter Property="AutomationProperties.Name" Value="Saturday"/>
+                        <Setter Property="TabIndex" Value="7"/>
                     </DataTrigger>
 
                     <DataTrigger TargetType="Frame" Binding="{Binding IsSelected}" Value="True">
--- a/Xamarin.Forms.Controls/TabIndexTest/DayView.xaml
+++ b/Xamarin.Forms.Controls/TabIndexTest/DayView.xaml
@@ -1,0 +1,103 @@
+﻿<?xml version="1.0" encoding="UTF-8"?>
+<ContentView xmlns="http://xamarin.com/schemas/2014/forms" 
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             xmlns:d="http://xamarin.com/schemas/2014/forms/design"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             mc:Ignorable="d"
+             xmlns:enums="clr-namespace:Xamarin.Forms.Controls.TabIndexTest"
+             x:Name="this"
+             x:Class="Xamarin.Forms.Controls.TabIndexTest.DayView">
+  <ContentView.Content>
+        <StackLayout BindingContext="{x:Reference this}"
+                     HorizontalOptions="FillAndExpand"
+                     AutomationProperties.IsInAccessibleTree="False" IsTabStop="False">
+            <Frame Margin="0"
+                   Padding="0"
+                   HasShadow="False"
+                   WidthRequest="32" 
+                   HeightRequest="32" 
+                   CornerRadius="16"
+                   HorizontalOptions="Center"
+                   BackgroundColor="{Binding ButtonBackgroundColor}"
+                   AutomationProperties.IsInAccessibleTree="True">
+                <Label TextColor="{Binding TextColor}"
+                       AutomationProperties.IsInAccessibleTree="False"
+                       HorizontalOptions="FillAndExpand"
+                       VerticalOptions="FillAndExpand"
+                       VerticalTextAlignment="Center"
+                       HorizontalTextAlignment="Center" IsTabStop="False">
+                    <Label.Triggers>
+                        <DataTrigger TargetType="Label" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Sunday}">
+                            <Setter Property="Text" Value="S"/>
+                        </DataTrigger>
+
+                        <DataTrigger TargetType="Label" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Monday}">
+                            <Setter Property="Text" Value="M"/>
+                        </DataTrigger>
+
+                        <DataTrigger TargetType="Label" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Tuesday}">
+                            <Setter Property="Text" Value="T"/>
+                        </DataTrigger>
+
+                        <DataTrigger TargetType="Label" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Wednesday}">
+                            <Setter Property="Text" Value="W"/>
+                        </DataTrigger>
+
+                        <DataTrigger TargetType="Label" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Thursday}">
+                            <Setter Property="Text" Value="T"/>
+                        </DataTrigger>
+
+                        <DataTrigger TargetType="Label" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Friday}">
+                            <Setter Property="Text" Value="F"/>
+                        </DataTrigger>
+
+                        <DataTrigger TargetType="Label" Binding="{Binding Day}" Value="{x:Static enums:DaysOfWeek.Saturday}">
+                            <Setter Property="Text" Value="S"/>
+                        </DataTrigger>
+                    </Label.Triggers>
+                </Label>
+                <!-- The view's accessibility label triggers -->
+                <Frame.Triggers>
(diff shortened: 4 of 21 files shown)


samhouts:
This currently makes #8848 worse because that custom renderer is expecting the PageRenderer to always return a non-null AccessibilityElements. I should probably stick all the subviews in there....

bjarteskogoy:
How far away is a release with this merged in?

samhouts:
@bjarteskogoy I have to make sure that any changes made here do not make any other reported issues worse. Once I completed that validation, this should be merged for the next service release, hopefully within the next 2-3 weeks. Thanks!

arevellfraedom:
@samhouts Can I just weigh in with another consideration.
One of the issues we are experiencing was caused by the 4.X changes to TabSto/TabIndex and is the same as this one https://github.com/xamarin/Xamarin.Forms/issues/7568 in regards to the way external keyboards navigate through the app. Although the effect is different, the cause is very similar to the above - ie every element having `TabIndex = 0` and `TabStop = true` when none are set means the default (good) OS behaviour is overridden.

The cause appears to be in https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/VisualElementRenderer.cs in how `FocusSearch` interprets the result of `GetTabIndexesOnParentPage`. It seems the hope is that method returns null when no `TabIndex` have been overridden. This is unfortunately not true and results a search occurring to locate the next element to tab to. This usually doesn't do a better job than OS.

Would you consider applying the same fix in this PR to this situation too?

In particular I am thinking `if (tabIndexes.Count <= 1) return;` in `Xamarin.Forms.Platform.Android/Renderers/PageRenderer.cs`

I think a very similar change in `FocusSearch` would resolve our issue.

This issue is serious enough for us that we need to consider forking Xamarin.Forms in order to pass a stringent a11y audit.

You can tell if this change is successfully with the reproduction in https://github.com/xamarin/Xamarin.Forms/issues/7568:

- Steps to Reproduce
- Open the sample project
- Go to Tab Order
- Press Tab multiple times
- When the focus is on the first button press Tab again

When fixed, you should find you can happily tab through all items on the page. This was true in Xamarin Forms 3.X

Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
```diff
@@ -173,6 +173,10 @@ public override AView FocusSearch(AView focused, [GeneratedEnum] FocusSearchDire
 			if (tabIndexes == null)
 				return base.FocusSearch(focused, direction);
 
+			// use OS default--there's no need for us to keep going if there's one or fewer tab indexes!
+			if (tabIndexes.Count <= 1)
```
arevellfraedom: Thanks for the quick response @samhouts !

Now I can see this change in place, will the `tabIndexes` indeed be empty when none are set?
Looking at `TabIndexExtensions`, `GetTabIndexesOnParentPage` only seems to care if `IsTabStop` is set, which is true by default for every control.

So will leaving `TabIndex` unset mean `GetTabIndexesOnParentPage` will return an empty list?

It would appear for this solution to work with us, we would have to leave in our "set IsTabStop to false" reflection hack which obviously we would prefer not to have.

samhouts: So, this is actually looking for TabIndex groups. If you haven't set them, the TabIndexes default to 0. In that case, this returns a single group of all elements with TabIndex == 0. If there's a single group, then there's no need to use this logic.

joshuangfraedom: Nice, does that mean that apps will work as they did before Xamarin 4.0 if they are not using tab stops @samhouts ?

arevellfraedom: Ah.... makes sense. Could I get an out of band nuget for these changes so I could try them out? I think this entire branch would good to validate against our cases.

We are currently on 4.3, but I think I can test 4.4 without much problem

samhouts:
@arevellfraedom Have you noticed this as an issue with iOS, too? I've tested the sample with 3.6, 4.0, and this PR, and it behaves the same in all cases.

arevellfraedom:
> @arevellfraedom Have you noticed this as an issue with iOS, too? I've tested the sample with 3.6, 4.0, and this PR, and it behaves the same in all cases.

iOS seems fine with keyboard. We did have to do the workaround mentioned in https://github.com/xamarin/Xamarin.Forms/issues/9191 for Voice Over, but not for keyboard (as iOS uses the NextResponder chain which is quite different)

Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
samhouts: @joshuangfraedom I hope so!

@arevellfraedom As soon as this build finishes, you should be able to grab the nuget from the artifacts. Let me know if you have any trouble accessing it, and I'll post it on the PR instead. https://dev.azure.com/xamarin/public/_build/results?buildId=14812&view=results

samhouts:
NuGet for testing: 
[Xamarin.Forms.4.4.0.993411.nupkg.zip](https://github.com/xamarin/Xamarin.Forms/files/4278851/Xamarin.Forms.4.4.0.993411.nupkg.zip)

Thanks!!!

joshuangfraedom:
Thanks @samhouts, I started testing that package in  https://github.com/xamarin/xamarin-forms-samples/tree/master/UserInterface/Accessibility (Pulled as a zip from microsoft docs) and while the first page behaves as expected (with no tab indexes set). The tab-indexes page will allow the user to tab to the first button (Descending TabIndex), after which it gets stuck there.

arevellfraedom:
> Thanks @samhouts, I started testing that package in https://github.com/xamarin/xamarin-forms-samples/tree/master/UserInterface/Accessibility (Pulled as a zip from microsoft docs) and while the first page behaves as expected (with no tab indexes set). The tab-indexes page will allow the user to tab to the first button (Descending TabIndex), after which it gets stuck there.

For us, this is not an issue (as we don't intend to use `TabIndex`) but thought you'd prefer to know that.

Xamarin.Forms.Platform.Android/VisualElementRenderer.cs
```diff
@@ -173,6 +173,10 @@ public override AView FocusSearch(AView focused, [GeneratedEnum] FocusSearchDire
...
+				return base.FocusSearch(focused, direction);
+
 			int tabIndex = element.TabIndex;
 			AView control = null;
 			int attempt = 0;
```
joshuangfraedom: @samhouts looking at the below code, it looks like the discovery of the next tab stop in the do while loop will only happen once, as maxAttempts = 0, and attempt starts at 0 but will be 1 at this stage, am I missing something?

joshuangfraedom: Ah sorry missed the out in the extension

arevellfraedom:
@samhouts hate to bring up the old "when will this be out", but.... when do you expect this to be out? We need to rely on this change for our upcoming release, which realistically means it being part of a 4.5 patch with 1-2 weeks. If this is not realistic, can you give an estimate of when it may be part of an official release so we may plan?

samhouts:
@arevellfraedom So, first, whenever we merge something to a release branch like 4.4.0, it gets merged up through all the active releases and then to master. Thus it will be in 4.4.0, 4.5.0, and 4.6.0, and in the nightlies from master. 

Second, I can't make any promises about release dates, but yes, this has a high probability of being released within two weeks. 🤞

arevellfraedom:
That would be amazing and I highly appreciate the effort. This change is very positive and will improve the accessibility of probably 90% of Xamarin forms apps, which I’m sure those users affected will very much appreciate.

I will put some effort in shortly to look beyond the cases we don’t specifically need, but others might, to see if I can provide any more feedback to ensure this PR is as good as possible.

arevellfraedom:
Hi @samhouts - I have done additional validation of the keyboard tabbing on Android and have a couple of observations.

These are all using https://github.com/xamarin/xamarin-forms-samples/tree/master/UserInterface/Accessibility:

4.4.0.993041 

1. Ascending tabindex 
- **Tabbing from nav bar: gets stuck on "Ascending tabindex" button**
- _If manually tap in "Enter data", then can tab through fields_ until **"Select a Monkey" after which point it is impossible to proceed**
 - _If skip to "Enter search term" then can tab through the last three elements_, **but then  which can't be tabbed through the same as "Select a Monkey"**
2. Descending tabindex
- **Same pass/fail as ascending (can't exit nav bar etc), just in reverse**
3. All tabindex = 0 
- _Can tab from nav bar and through buttons_
- **Can't tab through Date or "Select a Monkey"**
- **On wrap around, skips buttons**
4. Toggle IsTabStop
- _No fields can be tabbed through_
- **Ignored if all tabindex = 0**
5. Alternating IsTabStop
- **Tabbing from nav bar: gets stuck on "Ascending tabindex" button**
- **Ignored if all tabindex = 0**
- _If "Enter data" tapped, then "Enter name" correctly skipped on tab_

4.4.0.991757
- **Tabbing from nav bar: gets stuck on some outer element and does not get into buttons**
- _If manually tap in "Enter data", then can tab through fields_ until **"Select a Monkey" after which point it is impossible to proceed**
 - _If skip to "Enter search term" then can tab through the last three elements_, **but then  which can't be tabbed through the same as "Select a Monkey"**
2. Descending tabindex
- **Same pass/fail as ascending (can't exit nav bar etc), just in reverse**
3. All tabindex = 0 
- **Tabbing from nav bar: gets stuck on some outer element and does not get into buttons**
- **Can't tab through Date or "Select a Monkey"**
- **On wrap around, skips buttons**
4. Toggle IsTabStop
- _No fields can be tabbed through_
- _Still works if tabindex = 0_
5. Alternating IsTabStop
- **Tabbing from nav bar: gets stuck on some outer element and does not get into buttons**
- _Still works if tabindex = 0_
- _If "Enter data" tapped, then "Enter name" correctly skipped on tab_

3.6.0

1. Ascending tabindex 
- _Can tab from nav bar and through buttons_
- _Tabbing through "Select a Monkey" tabs to next field after popup value selected_
- _On wrap, buttons are tabbed through then fields_
2. Descending tabindex
- _Same passes as Ascending_
3. All tabindex = 0 
- _Same passes as Ascending_
4. Toggle IsTabStop
- _Fields cannot be tabbed through_
5. Alternating IsTabStop
- _Same passes as Ascending_

**Summary**
3.6.0 was basically perfect
4.4.0.991757 had particular problems getting from the navbar and when tabindex = 0
4.4.0.993041 corrects behaviour when tabindex = 0, but still struggles otherwise, in particular gets stuck on the first button, doesn't allow fields such as "Select a Monkey" to be tabbed through and also has a regression when tabindex = 0 but tabstop = false

4.4.0.993041 is still good for us, but does restore full 3.6.0 functionality

samhouts:
@arevellfraedom Thank you so much for the detailed test and review!! Given this does improve behavior, we'll merge it, and then we'll continue working on the remaining issues you mention.

The basic problem is that once we override the accessibility tree in iOS, we're stuck with that, even if we don't need to alter it. I'm having trouble finding a way to accurately reproduce it, since iOS doesn't seem to expose it. If ANYONE has has success with that, I'd love to hear what you did!

Thanks to EVERYONE who has been pushing this area forward, either by your reports, testing, review, etc!!

arevellfraedom:
> @arevellfraedom Thank you so much for the detailed test and review!! Given this does improve behavior, we'll merge it, and then we'll continue working on the remaining issues you mention.
> 
> The basic problem is that once we override the accessibility tree in iOS, we're stuck with that, even if we don't need to alter it. I'm having trouble finding a way to accurately reproduce it, since iOS doesn't seem to expose it. If ANYONE has has success with that, I'd love to hear what you did!
> 
> Thanks to EVERYONE who has been pushing this area forward, either by your reports, testing, review, etc!!

@samhouts , if I understand the issue you are facing with iOS correctly, I may be able to help.
You are correct that once you implement the `UIAccessibilityContainer` protocol (through exporting `accessibilityElementCount` etc) you are stuck with overriding the children of that view for accessibility. Once you have implemented the protocol, you can't get back to asking the OS to do it instead, at least not without obj-c method swizzling.

I see two options:
1. Have the `PageContainer` (the one implementing the protocol) be the View of the VC conditional on whether you want to override the accessibility tree. Not sure how practical this is but it would be the most flexible option (I think)

2. Use the less known fact you can call the `setAccessibilityElements` method of `UIAccessibilityContainer`. 
So the strange thing about this protocol is that you can either act by responding as a getter (whenever the OS feels it needs to know the accessibility children) or simply set them whenever you choose. In many way, acting as a getter is easier and more reliable (as it is called at the right times and can respond with elements based on the situation), but it has the major downside, as you have seen, that it is impossible to reply with "do the default". That is what `setAccessibilityElements` can be useful for.

So to make this useful, the first thing to determine is when to call it - the good news is that the answer is "any time before voiceover scans the page". So ViewDidLoad for example is just fine. So simply go through the logic of TabStop and TabIndex, and when you are ready, call `setAccessibilityElements` with the children or simply don't call it if you don't want to override. I _think_ sending null might even remove an override, but haven't checked.

The second thing is to know how to call it - this is not so easy as this method is so strange it does not appear on the `IUIAccessibilityContainer` interface. What you need to do instead is:

1. Create a selector for it
```SetAccessibilityElementsSelector = new Selector("setAccessibilityElements:");```
2. Prepare a `NSMutableArray` of the child elements you wish to set
3. Prepare a way of sending a message to the container you want to set the children on
```[DllImport("/usr/lib/libobjc.dylib", EntryPoint = "objc_msgSend")]
        private static extern void void_objc_msgSend_IntPtr(IntPtr receiver, IntPtr selector, IntPtr arg1);```
4. Send the message
```void_objc_msgSend_IntPtr(containingView.Handle, SetAccessibilityElementsSelector.Handle, array?.Handle ?? IntPtr.Zero);
```
Hopefully I have understood the question and that gives you some new options.

bjarteskogoy:
Will this show up in Xamarin.Forms 4.5 soon?

PureWeen:
This should be available in the nuget that went out yesterday. 

@samhouts correct me if I'm wrong

samhouts:
Yep, that's correct. It's now available in the latest 4.4.0, 4.5.0, and 4.6.0 releases. Thanks!

rnic99:
@samhouts Im not seeing any major difference in accessibility behavior in the lastest stable of 4.5, however I am seeing a significant improvement on the 4.6 pre-release

samhouts:
@rnic99 You tried 4.5.0.495 and there was no difference?

rnic99:
@samhouts Yep pulled down 4.5.0.495 this morning and the behavior was identical.

arevellfraedom:
It would appear 4.5 SR2 was taken shortly _before_ this PR was merged, although I cannot be sure

samhouts:
4.5.0.495 was release-4.5.0-sr2.2, and this commit is in it. That's strange.

rnic99:
@samhouts I just re-downloaded the stable release and it seems to be working fine on Android, and iOS. Or atleast its showing the same behavior as it was in 4.6 pre. My apologies for the confusion. I have a feeling there was a problem where even though it was showing 4.5 under my XF packages it was still deploying as 4.4.

samhouts:
@rnic99 Phew! I was hoping that was the case. Thanks for checking again!!

rnic99:
@samhouts of course, sorry for causing distress

PR #9187, opened 2020-01-13
[Android] Fix BottomNavigationItemView issue with MasterDetailPage

kvpt:
### Description of Change ###
On certains conditions MasterDetailPage can be in conflict with BottomNavigationView.
Like this : 
Screenshot_1578940463

The issue disappear if the BottomNavigationItem id are correctly set.
The BottomNavigationItem id was 0, 1, 2 ... which seem to cause issue with the ViewPager because these id are already used is others views in Android.
ViewId

I tried to make an UI test case based on the reproduction in #9143, and you know what, impossible to reproduce in the UI test app. The problem seem to only occur during the app start.

### Issues Resolved ### 
- fixes #9143 

### API Changes ###
None

### Platforms Affected ### 
- Android

### Behavioral/Visual Changes ###
None

### Before/After Screenshots ### 
Not applicable

### Testing Procedure ###
Use reproduction in #9143

### PR Checklist ###
- [x] Targets the correct branch
- [x] Tests are passing (or failures are unrelated)

Issue #9143, opened 2020-01-09
Android: Master Navigation page displayed as third bottom tabbar item


alekseevabah:
### Description
We have the following pages structure:
MainPage = MasterDetailPage:
 - Master: a Navigation page with a single Content page
 - Detail: a Tabbed page with 5 tabs:
 --- each tab is a Navigation page with a single Content page in it

Toolbar placement is set to Bottom for Android.

### Steps to Reproduce
Run the attached test project on android device.

### Expected Behavior
Third bottom tabbar item is displayed similar to other tabbar items.
Master page is accessible via the hamburger icon.

### Actual Behavior
Master page is displayed instead of third bottom tabbar item.
Master page is not displayed when tapping on the hamburger icon.

### Basic Information

- Version with issue: 4.3.0.908675
- Last known good version: 4.2.0.910310
- IDE: Visual Studio 2019 (16.4.0) for Windows
- Platform Target Frameworks:
  - Android: 9.0
- Nuget Packages: NETStandard.Library (2.0.3), Xamarin.Essentials (1.3.1)
- Affected Devices: HUAWEI VTR-L09 and others (we experienced this on multiple of our test devices)

### Screenshots
![Screenshot_20200120-173631](https://user-images.githubusercontent.com/24394092/72086198-cff54700-3306-11ea-8f03-419cfeb63a36.png)
![Screenshot_20200120-173639](https://user-images.githubusercontent.com/24394092/72086648-89ecb300-3307-11ea-955d-7fdeecc6011e.png)

### Reproduction Link
[AppWithMenuInsteadOfTabbedItem.zip](https://github.com/xamarin/Xamarin.Forms/files/4041450/AppWithMenuInsteadOfTabbedItem.zip)

### Notes
Looks like the issue was introduced with this pull request: https://github.com/xamarin/Xamarin.Forms/pull/6471 (we reverted this file and the bottom tabbar behaved correctly).
Using random numbers as Ids for MasterDetailContainer also solves the issue (looks like Master navigation page gets the same ID assigned as third bottom tabbar item?).

PureWeen:
@kvpt

kvpt:
I use this code in my apps for more than 6 months, but I don't use bottom bar.
I will look at it, thanks for the reproduction.
I will add an UI test at the same time as the fix.

kvpt:
[...]

PR #9709, opened 2020-02-25
revert [Android] Fix BottomNavigationItemView issue with MasterDetailPage

PureWeen:
### Description of Change ###

#9187 broke the following UI tests on master

TabBottomIconsAndTitlesChange_google_pixel_2
AddAndRemovePages_google_pixel_2-8_1_0
BottomTabbedPageWithModalIssueTestsPopFromSecondTab_google_pixel_2-8_1_0

### Testing Procedure ###
- All Green on Master

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/BottomNavigationViewUtils.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/BottomNavigationViewUtils.cs
@@ -67,7 +67,7 @@ namespace Xamarin.Forms.Platform.Android
 				var item = items[i];
 				using (var title = new Java.Lang.String(item.title))
 				{
-					var menuItem = menu.Add(0, Platform.GenerateViewId(), 0, title);
+					var menuItem = menu.Add(0, i, 0, title);
 					menuItems.Add(menuItem);
 					loadTasks.Add(SetMenuItemIcon(menuItem, item.icon, context));
 					UpdateEnabled(item.tabEnabled, menuItem);


PureWeen:
FYI @kvpt

kvpt:
Oh, sorry for that.

In the code you referenced
```cs
Element.Children.Count > item.ItemId
```
With autogenerated id this condition will never be true.
But what is the purpose of this condition ?

PureWeen:
item.ItemId typically mapped to the index in the array

The following line uses the ItemId to retrieve the Page
https://github.com/xamarin/Xamarin.Forms/blob/7a52542d50797ccc69ae1d8dd84259190d96bdb4/Xamarin.Forms.Platform.Android/AppCompat/TabbedPageRenderer.cs#L875

so that check is just guarded programming but it's debatable if it makes sense to guard there.

Like if the index is wrong something is wrong and it should probably throw an exception

kvpt:
We can't use the item id anymore but we can retrieve the item index in the menu.

```cs
				var itemIndex = -1;

				for (int i = 0; i < _bottomNavigationView.Menu.Size(); i++)
				{
					if (_bottomNavigationView.Menu.GetItem(i) == item)
					{
						itemIndex = i;
						break;
					}
				}

				if (itemIndex != -1 && _bottomNavigationView.SelectedItemId != item.ItemId && Element.Children.Count > itemIndex)
					Element.CurrentPage = Element.Children[itemIndex];
```
I tested with BottomTabbedPageTests and it seem to work.

@PureWeen What do you think about this approach, did I miss something ?

PureWeen:
@kvpt that makes sense

What are your thoughts on this?
> I'm wondering if your change should be applied everywhere that we add menus so that all MenuItems always have a unique id?

There are a couple places where we use menu items (ToolBarItems)

It feels like we should apply this same fix to those to future proof against future possible weirdness.

Just give everyone their own ID

kvpt:
Yes, on the pure Android side, it shouldn't break anything and protect us of some weird cases like the one in #9143. 
We just have to be sure that we don't rely on them as index, like here.

Not sure, but I think this place also need to be fixed :
https://github.com/xamarin/Xamarin.Forms/blob/efb4779bc2dd119a87e375ee4a9f7efe94abba4f/Xamarin.Forms.Platform.Android/Renderers/BottomNavigationViewUtils.cs#L77

PR #9733, opened 2020-02-26
Don't Init DualScreenService on non duo device

PureWeen:
### Description of Change ###

If a device isn't a Duo then don't initialize any of the duo services.
Fixed an issue where Pane1 was staying non visible after Pane2 was set as priority

### Platforms Affected ### 
- Android

### Testing Procedure ###
- Use the CG App on an API 29 device (non duo) and make sure you can change the orientation on the device without crashing
- On a duo device set Pane2 as the priority, tweak the minwide/tall mode settings so only pane2 shows up, span the app, unspan the app, and verify both panes are displaying

### PR Checklist ###


PureWeen:
UI test failure not related on API 19

PureWeen:
`CarouselViewHorizontal("CarouselView (Code, Horizontal)")_google_pixel_2-8_1_0` failing test isn't related

PR #9743, opened 2020-02-27
Fix null SwipeItem Parent in SwipeView

### LINKED ISSUE #9738: [Enhancement] SwipeViewItem parent is null, shouldn't be
## Summary
SwipeViewItem in SwipeView in the context of templated use in CollectionView has null parent. This means that RelativeSource binding fails to locate the ancestral BindingContext.... which would generally be the ViewModel.

This means that, in XAML implementation, the Command property for a SwipeViewItem is very difficult to set. 

## API Changes

Would be great if the following would work:
`<SwipeItemView Command="{Binding Source={RelativeSource AncestorType={x:Type myns:MyViewModel}}, Path=SwipeItemTapCommand}" CommandParameter="{Binding .}">`

## Intended Use Case
To bind ViewModel Command properties to the UI in the same way we would for other templated collection views

jsuarezruiz:
### Description of Change ###

This PR has changes to set the parent of a `SwipeItemView`. In this way, the use of RelatveSource when using commands for example, will work.

```
<SwipeItemView
     Command="{Binding Source={RelativeSource AncestorType={x:Type vm:ViewModel}}, Path=SwipeItemTapCommand}"
     CommandParameter="{Binding .}">
```

### Issues Resolved ### 

- fixes #9738 

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

#### Before
issue9738-before

#### After
issue9738-after

### Testing Procedure ###
Launch Core Gallery and navigate to the SwipeView samples. Select the "SwipeView BindingContext Gallery" and try to swipe and tap on a SwipeItem. In the top side, the last tapped element must be updated.

### PR Checklist ###


Xamarin.Forms.Core/SwipeView.cs
```diff
@@ -54,6 +61,12 @@ public SwipeItems BottomItems
 			set { SetValue(BottomItemsProperty, value); }
 		}
 
+		static void OnSwipeItemsChanged(BindableObject bindable, object oldValue, object newValue)
+		{
+			if (bindable is SwipeView swipeView)
```
StephaneDelcroix: you can use direct cast here

jsuarezruiz: You are right. Changed!

PR #9346, opened 2020-01-28
[iOS] Fix navigation bar background color in iOS13

rmarinho:
### Description of Change ###
(copy from #8031)

Incorporated changes in iOS13 pertaining to navigation bar background color.

### Issues Resolved ### 

- fixes #7701
- fixes #7350 
- fixes #7679 (possibly)

### PR Checklist ###

- [x] Targets the correct branch
- [ ] Tests are passing (or failures are unrelated)

PR #9760, opened 2020-02-27
Ios 2017 compatible Xamarin.Forms.Platform.iOS

PureWeen:
### Description of Change ###
Build different ios platform dlls for xcode10 and xcode11 compatibility

The problem with our previous approach for Xcode10 compatibility is that at runtime or link time VS 2017 will encounter types that it doesn't understand. 

So given the scenario of a Xamarin.Forms.iOS application built on VS2017 running on iOS13 you will hit exceptions if you encounter types that VS2017 doesn't understand

This PR doesn't build our DLL on VS 2017 it only if defs out any apis that are iOS13 specific which should produce a dll that will compile on VS2017 and then run/link successfully

I've also included an extra iOS build lane that installs all the VS 2017 sdks and compiles our control gallery to ensure we aren't exposing any APIS that VS 2017 doesn't understand

### Platforms Affected ### 
- iOS

### Testing Procedure ###
- use this nuget with full linking on vs2017 and vs2019. Make sure everything runs

### PR Checklist ###

### DIFF
--- a/.nuspec/Xamarin.Forms.nuspec
+++ b/.nuspec/Xamarin.Forms.nuspec
@@ -298,10 +298,9 @@
     <!--Mac-->
     <file src="..\Xamarin.Forms.Core\bin\$Configuration$\netstandard2.0\Xamarin.Forms.Core.dll" target="lib\Xamarin.Mac" />
     <file src="..\Xamarin.Forms.Xaml\bin\$Configuration$\netstandard2.0\Xamarin.Forms.Xaml.dll" target="lib\Xamarin.Mac" />
+    <file src="..\Xamarin.Forms.Platform.MacOS\bin\$Configuration$\Xamarin.Forms.Platform.dll" target="lib\Xamarin.Mac" />
     <file src="..\Xamarin.Forms.Platform.MacOS\bin\$Configuration$\Xamarin.Forms.Platform.macOS.dll" target="build\XCODE11" />
-    <file src="..\Xamarin.Forms.Platform.MacOS\bin\$Configuration$\Xamarin.Forms.Platform.dll" target="build\XCODE11" />
     <file src="..\Xamarin.Forms.Platform.MacOS\bin\$Configuration$\2017\Xamarin.Forms.Platform.macOS.dll" target="build\XCODE10" />
-    <file src="..\Xamarin.Forms.Platform.MacOS\bin\$Configuration$\2017\Xamarin.Forms.Platform.dll" target="build\XCODE10" />
 
     <!-- iOS Localized String Resource Assemblies -->
     <file src="..\Xamarin.Forms.Platform.iOS\bin\$Configuration$\ar\Xamarin.Forms.Platform.iOS.resources.dll" target="lib\Xamarin.iOS10\ar" />
--- a/.nuspec/Xamarin.Forms.nuspec
+++ b/.nuspec/Xamarin.Forms.nuspec
@@ -207,9 +207,14 @@
     <file src="..\Stubs\Xamarin.Forms.Platform.Android\bin\$Configuration$\MonoAndroid10.0\Xamarin.Forms.Platform.dll" target="lib\MonoAndroid10.0" />
 
     <!--iPhone Unified-->
-    <file src="..\Xamarin.Forms.Platform.iOS\bin\$Configuration$\Xamarin.Forms.Platform.iOS.dll" target="lib\Xamarin.iOS10" />
-    <file src="..\Xamarin.Forms.Platform.iOS\bin\$Configuration$\Xamarin.Forms.Platform.iOS.*pdb" target="lib\Xamarin.iOS10" />
-    <file src="..\Xamarin.Forms.Platform.iOS\bin\$Configuration$\Xamarin.Forms.Platform.iOS.*mdb" target="lib\Xamarin.iOS10" />
+    <file src="_._" target="lib\lib\Xamarin.iOS10\_._" />
+    <file src="..\Xamarin.Forms.Platform.iOS\bin\$Configuration$\Xamarin.Forms.Platform.iOS.dll" target="build\XCODE10" />
+    <file src="..\Xamarin.Forms.Platform.iOS\bin\$Configuration$\Xamarin.Forms.Platform.iOS.*pdb" target="build\XCODE10" />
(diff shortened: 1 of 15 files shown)


.nuspec/Xamarin.Forms.targets
```diff
@@ -183,7 +183,39 @@
...
+  </PropertyGroup>
+
   
+  <Target Name="IncludeCorrectXIOSReference" Condition="'$(TargetFrameworkIdentifier)' == 'Xamarin.iOS'">
+	<PropertyGroup>
+		<Use2017 Condition="'$(Use2017)' == '' AND '$(MSBuildRuntimeType)' == 'Mono'">$(FrameworkSDKRoot.Contains('/Versions/5'))</Use2017>
+		<Use2017 Condition="'$(Use2017)' == '' AND '$(MSBuildAssemblyVersion)' &lt; '16.0'">true</Use2017>
+		<Use2017 Condition="'$(Use2017)' == ''">false</Use2017>
+	</PropertyGroup>
+    <ItemGroup Condition="'$(Use2017)' != 'true'">
```
VladislavAntonyuk: Use positive condition (== 'false' is easier to read)

Xamarin.Forms.Platform.iOS/Xamarin.Forms.Platform.iOS.csproj
```diff
@@ -17,7 +18,7 @@
     <DebugSymbols>true</DebugSymbols>
     <Optimize>false</Optimize>
     <OutputPath>bin\Debug</OutputPath>
-    <DefineConstants>DEBUG;</DefineConstants>
+    <DefineConstants>$(DefineConstants);UWP_14393</DefineConstants>
```
VladislavAntonyuk: Do we need UWP in iOS project?

PureWeen:
```suggestion
    <DefineConstants>$(DefineConstants);DEBUG;</DefineConstants>
```

.nuspec/Xamarin.Forms.targets
PureWeen: But that's not the equivalent check. The equivalent check would be

```XAML
<ItemGroup Condition="'$(Use2017)' == 'false' OR "'$(Use2017)' == ''>
```

it's good to use != anywhere that you want to articulate the default path. That way at some point if Use2017 goes away this path will still run because it represents the default

Line 205 is probably kind of the excessive one but I don't really care. It's easier to have it set to something so when I inspect the binlogs I can validate that as well

Xamarin.Forms.Platform.iOS/Renderers/SwipeViewRenderer.cs
```diff
@@ -137,9 +137,11 @@ protected override void SetBackgroundColor(Color color)
 		{
 			UIColor backgroundColor;
 
+#if __XCODE11__
 			if (Forms.IsiOS13OrNewer)
 				backgroundColor = UIColor.SystemBackgroundColor;
 			else
+#endif
 				backgroundColor = UIColor.White;
```
PureWeen: @jsuarezruiz is this correct?

build/steps/build-windows.yml
```diff
@@ -70,23 +65,9 @@ jobs:
       inputs:
         restoreSolution: ${{ parameters.slnPath }}
 
-    - task: MSBuild@1
```
PureWeen: @rmarinho So I moved all these builds into cake, because of that whole issue where branches get merged into the head branch which causes the yaml to be inconsistent with the build.

I'd like to just start moving whatever we can on the YML into cake so that the build can just more easily travel with the code. 

Plus this let's me run the same build tasks for generating the nugets locally as we will run on the servers.

Xamarin.Forms.Platform.iOS/Renderers/NavigationRenderer.cs
```diff
@@ -619,14 +619,8 @@ void UpdateBarBackgroundColor()
...
-			if (!Forms.IsiOS13OrNewer)
-			{
-				// Set navigation bar background color
-				NavigationBar.BarTintColor = barBackgroundColor == Color.Default
-				? UINavigationBar.Appearance.BarTintColor
-				: barBackgroundColor.ToUIColor();
-			}
-			else
+#if __XCODE11__
+			if (Forms.IsiOS13OrNewer)
```
PureWeen: @rmarinho can you verify all the changes to this file please? 

Basically if you are compiling on VS 2017 I'm removing all the code you checked in for this PR

https://github.com/xamarin/Xamarin.Forms/pull/9346

.nuspec/Xamarin.Forms.nuspec
```diff
@@ -207,9 +206,13 @@
     <file src="..\Stubs\Xamarin.Forms.Platform.Android\bin\$Configuration$\MonoAndroid10.0\Xamarin.Forms.Platform.dll" target="lib\MonoAndroid10.0" />
 
     <!--iPhone Unified-->
-    <file src="..\Xamarin.Forms.Platform.iOS\bin\$Configuration$\Xamarin.Forms.Platform.iOS.dll" target="lib\Xamarin.iOS10" />
```
mattleibow: If you remove this line, then does it install into ios? Or is it picking the netstandard dll?

PureWeen: This dll isn't referenced anywhere else in the nuspec so it there isn't a NS version that comes in. Since the dll doesn't even exist in the lib folder anymore I had to remove this for the nuget to even package

build.cake
```diff
@@ -37,37 +37,63 @@ PowerShell:
...
+bool buildForVS2017 = Convert.ToBoolean(Argument("buildForVS2017", "false"));
 
+string artifactStagingDirectory = Argument("Build_ArtifactStagingDirectory", (string)null) ?? EnvironmentVariable("Build.ArtifactStagingDirectory") ?? EnvironmentVariable("Build_ArtifactStagingDirectory") ?? ".";
 var ANDROID_HOME = EnvironmentVariable ("ANDROID_HOME") ??
     (IsRunningOnWindows () ? "C:\\Program Files (x86)\\Android\\android-sdk\\" : "");
 
-string monoMajorVersion = "5.18.1";
-string monoPatchVersion = "28";
-string monoVersion = $"{monoMajorVersion}.{monoPatchVersion}";
+string[] androidSdkManagerInstalls = new string[0];//new [] { "platforms;android-24", "platforms;android-28"};
```
jfversluis: Is this correct?

PR #9821, opened 2020-03-03
Fix Issue7242 iOS FontSize is calculated incorrectly

### LINKED ISSUE #7242: [Bug] iOS FontSize is calculated incorrectly
### Description
On iOS named fontsizes are calculated incorrectly usually resulting in 1 point less than expected.
### Steps to Reproduce
1.  Executing 
` var result = new FontSizeConverter().ConvertFromInvariantString(nameof(NamedSize.Micro));`
or add a label with fontsize micro

### Expected Behavior
NamedSize.Micro should return 12 with default settings
### Actual Behavior
Returns 11
### Basic Information
I think the issue lies in [https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.iOS/Forms.cs](url)

The font scaling factor is calculated on line 188 like this.
`_fontScalingFactor = (double)UIFont.PreferredBody.PointSize / 18f;`
I think it should be :
`_fontScalingFactor = (double)UIFont.PreferredBody.PointSize / 17f;`

Why?
The default UIFont.PreferredBody.PointSize is 17 on my emulator iPhone 7 iOS 12.2.
Also on the Apple guidelines [https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/](url)
It states that the default body Size in points is 17.

And on line 224 in the source it is also stated that
`// iOS docs say default button font size is 15, default label font size is 17 so we use those as the defaults.`

- Version with issue: Xamarin.Forms 4.1.0.618606
- Last known good version: Xamarin.Forms 4.0.0.425677
- IDE: Visual Studio 2019
- Platform Target Frameworks: 
  - iOS:  8.0
- Affected Devices: only tested on iPhone 7 iOS 12.2

Omniq-APL-Rollving:
### Description of Change ###
Changed the _fontScalingFactor so that the font size values corresponds with apples human interface guidelines (https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/). 

The change was made in file C:\…\Xamarin.Forms\Xamarin.Forms.Platform.iOS\Forms.cs, line 199 was changed from:
“_fontScalingFactor = (double)UIFont.PreferredBody.PointSize / 18f;”
to:
“_fontScalingFactor = (double)UIFont.PreferredBody.PointSize / 17f;”.
The issue was first introduced when scaling was added 
https://github.com/xamarin/Xamarin.Forms/pull/5985.

### Issues Resolved ### 
- fixes #7242

### API Changes ###
 None

### Platforms Affected ### 
- iOS

Text sizes will increase slightly because of the slight bump in font size for example micro that changes from 11 to 12 and medium/default goes from 16 to 17 which is apples recommended body text size.

Before: iPhone 11 iOS 13 3 2020-03-03 14_34_53

After: iPhone 11 iOS 13 3 2020-03-03 14_29_11

### Testing Procedure ###
I was unable to test the changes in any other way than testing manually checking values from the issue case and comparing results to apples guidelines.

### PR Checklist ###


AndreiMisiukevich:
Hey, I assume you selected the wrong target branch :)

Omniq-APL-Rollving:
Oh, I'm sorry I might have misunderstood the instructions what branch should I have targeted?
@AndreiMisiukevich

AndreiMisiukevich:
> Oh, I'm sorry I might have misunderstood the instructions what branch should I have targeted?
> @AndreiMisiukevich

what branch did you select as base for your PR?

Omniq-APL-Rollving:
Thanks for the heads up, I missed that I pulled version 4.6 and not  4.5

Xamarin.Forms.Platform.iOS/Forms.cs
```diff
@@ -196,7 +196,7 @@ public IOSPlatformServices()
 #if __MOBILE__
 				//The standard accisibility size for a font is 18, we can get a
```
VladislavAntonyuk: Comment also need to be updated

Xamarin.Forms.Platform.iOS/Forms.cs
```diff
@@ -194,9 +194,9 @@ class IOSPlatformServices : IPlatformServices
 			public IOSPlatformServices()
 			{
 #if __MOBILE__
-				//The standard accisibility size for a font is 18, we can get a
+				//The standard accisibility size for a font is 17, we can get a
```
StephaneDelcroix: fix the typo while you're at it...

is there a reference doc somewhere for this ? if so, please add a comment

jfversluis:
@Omniq-APL-Rollving if you can address the last comment to fix the typo and add a relevant doc then I think we're all good to go here

PR #9856, opened 2020-03-05
Add support for animation rate parameter

### LINKED ISSUE #5322: The `rate` parameter in IAnimatable.Animate() method doesn't take effect

Bobface:
### Description of Change ###

Adds support for the `rate` parameter for `Animation`. This parameter existed but was unused.

### Issues Resolved ### 
fixes #5322

### API Changes ###

Added:
- `public Tweener(uint length, uint rate)`
 - `uint Tweener.Rate { get; }`

The existing constructor of `Tweener` is also kept and defaults `rate` to `1`. This way existing code which uses the `Tweener` does not have to be updated.

### Platforms Affected ### 
- Core/XAML (all platforms)

None

Below are three GIFs for an example animation with 50ms, 200ms and 500ms rate. 

#### Rate=50 -> 20 frames per second
50ms
#### Rate=200 -> 5 frames per second
200ms
#### Rate=500 -> 2 frames per second
500ms

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/AnimationExtensions.cs
+++ b/Xamarin.Forms.Core/AnimationExtensions.cs
@@ -176,7 +176,7 @@ namespace Xamarin.Forms
 
 			var info = new Info { Rate = rate, Length = length, Easing = easing ?? Easing.Linear };
 
-			var tweener = new Tweener(info.Length);
+			var tweener = new Tweener(info.Length, info.Rate);
 			tweener.Handle = key;
 			tweener.ValueUpdated += HandleTweenerUpdated;
 			tweener.Finished += HandleTweenerFinished;
--- a/Xamarin.Forms.Core/Tweener.cs
+++ b/Xamarin.Forms.Core/Tweener.cs
@@ -94,7 +94,7 @@ namespace Xamarin.Forms
 					_lastMilliseconds = ms;
 				}
 
-				long wantedFrames = _lastMilliseconds / Rate;
+				long wantedFrames = (_lastMilliseconds / Rate) + 1;
 				if(wantedFrames > _frames || Value >= 1.0f)
 				{
 					ValueUpdated?.Invoke(this, EventArgs.Empty);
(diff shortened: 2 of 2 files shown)


ysangkok:
I have tested this, and it works.

knocte:
@StephaneDelcroix can you review?

knocte:
or @rmarinho maybe?

knocte:
@samhouts Samantha, by any chance can you review here? or find a reviewer? many thanks

Bobface:
@samhouts I saw you added the *API-change* label. Just to let you know, the PR does not break backwards-compatibility, it just extends the API.

Issue #9804, opened 2020-03-02
[Spec] OS Themes (Dark Mode support)


jfversluis:
# OS Themes (Dark Mode)

In order to be able to use the different themes that are now supported in iOS, Android and UWP, we should surface these APIs to Xamarin.Forms for our end-users.

This spec describes all the changes needed to leverage the different functionalities that are available in the underlaying platforms today. This way, people will have the ability to implement OS Theme mode support in their apps easier.

### Different App Themes
We determine the supported user interface styles through an enum, provided by Essentials. Repeated here for clarity.

```csharp
public enum AppTheme
{
    Unspecified,
    Light,
    Dark
}
```

#### DeviceInfo
In the `Device.DeviceInfo` we add a property to be able to see what appearance is currently active

| API | Description |
| ------------- | ------------- |
| `AppTheme CurrentAppTheme` | What is the actual theme currently for this device |

Already [available through Essentials](https://github.com/xamarin/Essentials/pull/927)

We wrap the Essentials call for this in an API of our own. This will make the API for this more discoverable.

##### AppThemeChanged Event
To be able to respond to changes we expose this event on all visual elements.

| API | Description |
| ------------- | ------------- |
| `void AppThemeChanged(AppThemeChangedEventArgs args)` | Event that is fired whenever the app theme changed |

```csharp
public class AppThemeChangedEventArgs : EventArgs
{
    public AppTheme RequestedTheme { get; }
}
```

##### ~Fix a `VisualElement` on a Certain `AppTheme`~ Not possible on all major platforms
We want to be able to set a fixed app theme per `VisualElement`. If this property is not set it will just use the current setting from the OS which will be reflected by `AppTheme.Unspecified`.

| API | Description |
| ------------- | ------------- |
| `AppTheme VisualElement.AppTheme` | Property to set the specific page to a specific theme. Default value: `AppTheme.Unspecified` |

### Images & Colors
To be able to differentiate images and colors between themes we will introduce the `OnAppTheme` which will act alike `OnIdiom` but now for the different app themes. We will also make this available as a markup extension for XAML.

#### Example 1: Different Colors per AppTheme
```xml
<Label.TextColor>
     <OnAppTheme x:TypeArguments="Color" Default="Black" Light="Red" Dark="Green" />
</Label.TextColor>
```

#### Example 2: Different Colors per AppTheme Markup Extension
```
<Label ... TextColor="{OnAppTheme Black, Light=Red, Dark=Green}" />
```
[...]

PR #9869, opened 2020-03-06
[AND] Default style to DayNight (Dark Theme support)

jfversluis:
### Description of Change ###

Pull the default style into the Android platform project and update it to use DayNight to support Dark Theme.

### Issues Resolved ### 

- related #9804

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

From what I can think of these would be the scenarios:
- When the user has created customizations in the theme we supply in the templates by default, this should still override whatever we do here.

- If the user has created their own theme altogether, they are not hit by this change

- If the user is still using the default theme from the templates, nothing changes. It will act as an override to the theme we specify here

Basically the only way the user will see this change is if they throw out the current theme in their project and whenever we decide to remove the theme from the template and use this as the default. Whenever that happens, the user will see that the app responds to changes in light/dark theme as expected

Not applicable

### Testing Procedure ###
Go through the scenarios under "Behavioral/Visual Changes" as much as possible. If you have any (sample) apps laying around, pull down the NuGets, install them on that project and see if it behaves as you would expect.

There is not much to see from the gallery app as it has its own custom theme. You can fiddle around with that to see what happens as well.

All of this across different API levels to ensure backwards compatibility.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Resources/values/styles.xml
+++ b/Xamarin.Forms.Platform.Android/Resources/values/styles.xml
@@ -7,4 +7,28 @@
   <style name="NestedScrollBarStyle" android:id="@+id/nestedScrollViewStyle">
     <item name="android:scrollbars">vertical|horizontal</item>
   </style>
+    <style name="MainTheme" parent="MainTheme.Base">
+    </style>
+    <!-- Base theme applied no matter what API -->
+    <style name="MainTheme.Base" parent="Theme.AppCompat.DayNight.DarkActionBar">
+        <!--If you are using revision 22.1 please use just windowNoTitle. Without android:-->
+        <item name="windowNoTitle">true</item>
+        <!--We will be using the toolbar so no need to show ActionBar-->
+        <item name="windowActionBar">false</item>
+        <!-- Set theme colors from https://aka.ms/material-colors -->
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.Android/Resources/values/styles.xml
```diff
@@ -7,4 +7,28 @@
   <style name="NestedScrollBarStyle" android:id="@+id/nestedScrollViewStyle">
     <item name="android:scrollbars">vertical|horizontal</item>
   </style>
-</resources>
+    <style name="MainTheme" parent="MainTheme.Base">
+    </style>
+    <!-- Base theme applied no matter what API -->
+    <style name="MainTheme.Base" parent="Theme.AppCompat.DayNight.DarkActionBar">
```
PureWeen: Will this theme work for projects targeting MonoAndroid9.0?

jfversluis: Yes, it's built on top of the regular AppCompat themes so it will just fallback to the ones available there. Just took the NuGets and verified that it in fact does

image

jfversluis:
Fails not related. Just our good old friend

image

PR #9922, opened 2020-03-10
[XamlC] use XmlReader for detecting XAML and x:Class

jonathanpeppers:
### Description of Change ###

Context: https://www.jetbrains.com/profiler/

I was profiling Xamarin.Android project builds with `dotTrace`, and
noticed an `IsXaml` method from XamlC taking time:

    0.07%   IsXaml  •  40/0 ms  •  Xamarin.Forms.Build.Tasks.CecilExtensions.IsXaml(EmbeddedResource, ModuleDefinition, out String)
      0.07%   Execute  •  Xamarin.Forms.Build.Tasks.XamlCTask.Execute(out IList)
        0.07%   Execute  •  Xamarin.Forms.Build.Tasks.XamlTask.Execute

In this build, I had changed only XAML and the netstandard project was
using `<ProduceReferenceAssembly>True</ProduceReferenceAssembly>`.

The project is here:

https://github.com/xamarin/xamarin-android/tree/master/tests/Xamarin.Forms-Performance-Integration

Reviewing `IsXaml`, it uses `XmlDocument` to parse every
`EmbeddedResource` with a `.xaml` extension. The problem with this, is
that `XmlDocument` loads the entire XML file into memory and parses
it. This is a step before the actual work of XamlC, so it is currently
parsing every file twice.

If we switch this code to use `XmlReader` instead, we can read to the
first XML element and stop there to get the information needed. We
don't need to parse the entire file at this step.

The results of building the Xamarin.Forms.Controls project on
Windows:

    Before:
    1973 ms  XamlCTask                                  1 calls
    After:
    1917 ms  XamlCTask                                  1 calls

The same project running on Mono / macOS:

    Before:
    6785 ms  XamlCTask                                  1 calls
    After:
    6224 ms  XamlCTask                                  1 calls

My Mac is a slightly slower machine than my PC, but the big difference
is Mono vs .NET Framework.

### Issues Resolved ### 

n/a

### API Changes ###

 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###

I added some unit tests here. I think the existing CI will be sufficient.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/XamlTask.cs
+++ b/Xamarin.Forms.Build.Tasks/XamlTask.cs
@@ -78,22 +78,19 @@ namespace Xamarin.Forms.Build.Tasks
 			if (!resource.Name.EndsWith(".xaml", StringComparison.InvariantCulture))
 				return false;
 
-			using (var resourceStream = resource.GetResourceStream()) {
-				var xmlDoc = new XmlDocument();
-				xmlDoc.Load(resourceStream);
+			using (var resourceStream = resource.GetResourceStream())
+			using (var reader = XmlReader.Create(resourceStream))
+			{
+				// Read to the first Element
+				while (reader.Read() && reader.NodeType != XmlNodeType.Element) ;
 
-				var nsmgr = new XmlNamespaceManager(xmlDoc.NameTable);
-
-				var root = xmlDoc.SelectSingleNode("/*", nsmgr);
-				if (root == null)
+				if (reader.NodeType != XmlNodeType.Element)
 					return false;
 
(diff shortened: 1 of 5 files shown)


StephaneDelcroix:
I'm surprised that XamlG doesn't show up in your performances then, as it also uses XmlDocument and then does some XPath queries too...

jonathanpeppers:
> I'm surprised that XamlG doesn't show up in your performances

Yeah, I'm using this "hot path" feature:

image

Not sure why XamlG isn't showing up, this is how long it actually took:

image

37ms is pretty quick, but I'll take a look and see if it has something similar.

PR #9931, opened 2020-03-11
Unhook Header/Footer measure handlers on Dispose

### LINKED ISSUE #8303: Swiping from left to right This lets the Label dissapear immediately in iOS.
### Description
I have a Label in StackLayout inside a ContentPage, which is in a NavigationPage. VerticalTextAlignment is set to End. Swiping from left to right This lets the Label dissapear immediately in iOS.

 I have two page which is in the name of HomePage & MainPage.
1. HomePage having only one button for navigates to Main page.
2. MainPage having Label with VerticalTextAlignment.

I have set NavigationPage.HasNavigationBar="False" to Home Page.

### Steps to Reproduce

1. Run the attached sample.
2. Swipe from left to right.

### Expected Behavior

The label should be visible while swiping.

### Actual Behavior

The label is not visible while swiping.

### Basic Information

Version with issue: 4.3.0.947036
IDE: Visual Studio 2019 in Mac.
Platform Target Frameworks:
Simulator Version - iPhoneXR iOS - 12.4
### Screenshots

Video: [https://www.syncfusion.com/downloads/support/directtrac/general/ze/NavigationIssue-1488618650](url)

### Reproduction Link
Sample : [https://www.syncfusion.com/downloads/support/directtrac/general/ze/SfTabView-1036229027 ](url)

### LINKED ISSUE #8308: [Bug] [iOS] Cannot access a disposed object. Object name: 'GroupableItemsViewController`1
### Description
A collection view with header causes a crash after navigating via the shell flyout.
The bug is iOS only. On Android it works fine.

The app wont crash if you use tabs.

### Steps to Reproduce

1. Open the App
2. Use the shell flyout to navigate to Page 2
3. Use the shell flyout to navigate back to Page 1

--> The App will crash

### Expected Behavior
Not crashing.
### Actual Behavior
Throws:

> System.ObjectDisposedException: 'Cannot access a disposed object.
> Object name: 'GroupableItemsViewController`1'.'
and crashes.

### Basic Information

- Version with issue: 4.3.0.947036, 4.3.0.908675, ..., 4.4.0.991265
- Last known good version: 4.2.0.910310
- IDE: Visual Studio 2019
- Platform Target Frameworks: 
  - iOS:  13.1, 13.2 (Probbably other versions too)

### Reproduction Link

[BugExample.zip](https://github.com/xamarin/Xamarin.Forms/files/3789323/BugExample.zip)

hartez:
### Description of Change ###

The handlers set up to remeasure the Header/Footer when their Forms element sizes changed were not being unhooked on Dispose; this led to the events firing and calling the handlers with a reference to an already disposed UICollectionView.

These changes unhook the handlers on Dispose.

### Issues Resolved ### 

- fixes #8308 

### API Changes ###
 
None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###

Automated test

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.ControlGallery.iOS/Tests/NavigationTests.cs
+++ b/Xamarin.Forms.ControlGallery.iOS/Tests/NavigationTests.cs
@@ -14,12 +14,15 @@ namespace Xamarin.Forms.ControlGallery.iOS.Tests
 				Title = "root",
 				Content = new Label { Text = "Hello" }
 			};
-			var navPage = new NavigationPage(root);
-			var renderer = GetRenderer(navPage);
 
-			// Calling Dispose more than once should be fine
-			renderer.Dispose();
-			renderer.Dispose();
+			Device.InvokeOnMainThreadAsync(() => { 
+				var navPage = new NavigationPage(root);
+				var renderer = GetRenderer(navPage);
+
+				// Calling Dispose more than once should be fine
+				renderer.Dispose();
+				renderer.Dispose();
+			});
 		}
 	}
 }
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue8308.xaml
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue8308.xaml
@@ -1,0 +1,46 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<local:TestShell xmlns="http://xamarin.com/schemas/2014/forms"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             xmlns:d="http://xamarin.com/schemas/2014/forms/design"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             mc:Ignorable="d"
+             Title="Test 8308" xmlns:local="using:Xamarin.Forms.Controls"
+             x:Class="Xamarin.Forms.Controls.Issues.Issue8308">
+
+    <FlyoutItem Title="Page 1" AutomationId="Page1">
+        <Tab Title="Page 1">
+            <ShellContent>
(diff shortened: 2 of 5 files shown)


Xamarin.Forms.ControlGallery.iOS/Tests/NavigationTests.cs
```diff
@@ -14,12 +14,15 @@ public void NavigationRendererDoubleDisposal()
...
 
-			// Calling Dispose more than once should be fine
-			renderer.Dispose();
-			renderer.Dispose();
+			Device.InvokeOnMainThreadAsync(() => { 
+				var navPage = new NavigationPage(root);
+				var renderer = GetRenderer(navPage);
+
+				// Calling Dispose more than once should be fine
+				renderer.Dispose();
```
VladislavAntonyuk: Could you please leave a comment why do we call Dispose twice

rmarinho:
Something wrong here, trying to understand why Test fails on Android but it's green on VSTS, 

On VSTS on Androdi reports 47 tests on ISSUES, it should be around 500

Xamarin.Forms.ControlGallery.iOS/Tests/NavigationTests.cs
kvpt: I think it's because dispose method should handle multiple calls and not cause a crash if called after the object has already been disposed.
So a dispose test should also test for this case.

hartez: We don't really do it on purpose, but `Dispose(true)` can occasionally be called more than once in some race conditions when we're doing things like popping/replacing pages, when we've got weak event references, and possibly when we've just written bad code. 
And `Dispose(false)` might happen under some conditions when objects are being finalized. 

So it's just good practice to ensure your `Dispose` implementations can gracefully handle being called more than once.

PR #9934, opened 2020-03-11
Layered font image source [WIP]

GalaxiaGuy:
### Description of Change ###

The original FontIconSource spec (https://github.com/xamarin/Xamarin.Forms/issues/3189) included a layered option. This is an implementation for iOS. My plan is to add the other platforms, but I wanted confirmation this was still likely to be accepted before continuing.

### API Changes ###
```csharp
[ContentProperty("Layers")]
public class LayeredFontImageSource : ImageSource
{
	public IList<FontImageSource> Layers { get; set; }
	public string FontFamily { get; set; }
	public double Size { get; set; }
}
```
 -->
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS

Simulator Screen Shot - iPad Air (3rd generation) - 2020-03-11 at 13 08 34

### Testing Procedure ###

Some layered icons have been added to the Font ImageSource gallery

### Open questions

I've repurposed `FontImageSourceHandler` to handle layered and non-layered to share code easier. This does mean that the non-layered case allocates a List. Would it be better to keep them separate.

I initially thought `FontFamily` and `Size` were highly likely to be shared between layers, so adding them made sense. `Color` seemed less likely, but possible. `Glyph` seems unlikely, but allowing it to be shared wasn't any harder.

Should `LayeredFontImageSource` inherit from `FontImageSource`? It doesn't really make sense conceptually and would imply you could have a layer of layers. On the other hand it would be handy fallback if platforms don't get a proper implementation.

I believe I can implement Android and macOS. I could possibly do UWP.

On the subject of UWP, I'm not sure what `LoadIconSourceAsync` and `LoadIconElementAsync` should do.

### PR Checklist ###


GalaxiaGuy:
I've added an implementation for macOS. There is some odd clipping happening (that was also happening before) that I haven't been able to fix.

I've also changed the macOS implementation to support screen scaling.

GalaxiaGuy:
I've now added Android.
Screenshot_1585486172

GalaxiaGuy:
I've had issues getting Forms to compile on Windows, so I'm going to have to leave it at Android, iOS and macOS.

The UWP implementation should be straightforward.

PR #9958, opened 2020-03-13
Cross-Platform OS Theme APIs

jfversluis:
### Description of Change ###

Adds the cross-platform APIs to support different OS Themes

#### Background
See #9804 for the relevant APIs and how to use it. Because of timing I have removed the dependency on Essentials from this (see #9926) so we can get this merged in time for 4.6. After that, we can easily swap out this implementation for the Essentials one. By keeping naming and enum values, etc. the same this should be an easy task.

### Issues Resolved ### 

- related to #9804

### API Changes ###

Added:
 - public enum AppTheme { Unspecified, Light, Dark }
 - public class AppThemeColor
 - public class OnAppTheme<T>
 - AppTheme Application.RequestedTheme { get; }
 - public event EventHandler<AppThemeChangedEventArgs> Application.RequestedThemeChanged;
 - public void Application.OnRequestedThemeChanged(AppThemeChangedEventArgs args)
 - public class AppThemeChangedEventArgs
 - protected virtual void OnRequestedThemeChanged(AppTheme newValue)

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP
- MacOS
- WPF
- Tizen

None

Not applicable

### Testing Procedure ###

I have added a gallery page which shows (nearly) all ways to use this new functionality. Check if that works as expected. Whenever the values are used as a `DynamicResource` they should update whenever the theme is changed while the app is running. Check out the gallery page or, to take it a step further, download the NuGets and take it for a spin in your own (sample) app.

### PR Checklist ###


jsuarezruiz:
I have done some tests. If you change the system theme, when initializing a page we initialize `OnAppTheme` and load the correct theme color. However, if we move the App to the background, change the theme and return to the App, the change will not be detected until navigating, etc. 

xamarinforms-themes

Still missing to implement an event to detect change in appearance mode, right?. Is the PR in progress or are we going to divide the Spec into different PRs?.

Xamarin.Forms.Core/AppThemeColor.cs
```diff
@@ -0,0 +1,52 @@
...
+	public class AppThemeColor
+	{
+		Color _light;
+		Color _dark;
+		Color _default;
+		bool _isLightSet;
+		bool _isDarkSet;
+		bool _isDefaultSet;
+
+		public Color Light
```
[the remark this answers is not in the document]
jfversluis: Done! :)

Xamarin.Forms.Controls/GalleryPages/OSThemesGalleryXaml.xaml
```diff
@@ -0,0 +1,55 @@
...
+                </OnAppTheme>
+            </StyleSheetExtension.Style>
+        </StyleSheet>
+    </ContentPage.Resources>
+    <ScrollView>
+        <StackLayout HorizontalOptions="Center" VerticalOptions="Center">
+            <Label TextColor="{OnAppTheme Light=Green, Dark=Red}">This text is green or red depending on Light (or default) or Dark</Label>
+            <Label>
+                <Label.TextColor>
+                    <OnAppTheme x:DataType="Color" Light="Green" Dark="Red" />
```
jsuarezruiz: Should the OnAppTheme extension detect theme changes update dynamically?

Xamarin.Forms.Platform.UAP/WindowsBasePlatformServices.cs
```diff
@@ -28,10 +29,12 @@ internal abstract class WindowsBasePlatformServices : IPlatformServices
 	{
 		const string WrongThreadError = "RPC_E_WRONG_THREAD";
 		readonly CoreDispatcher _dispatcher;
+		readonly UISettings _uiSettings = new UISettings();
 
 		protected WindowsBasePlatformServices(CoreDispatcher dispatcher)
 		{
 			_dispatcher = dispatcher ?? throw new ArgumentNullException(nameof(dispatcher));
+			_uiSettings.ColorValuesChanged += UISettingsColorValuesChanged;
```
jsuarezruiz: Could we unsubscribe this event?

Xamarin.Forms.Controls/GalleryPages/OSThemesGalleryXaml.xaml
```diff
@@ -0,0 +1,55 @@
...
+                </Label.TextColor>
+            </Label>
+            <Label Style="{DynamicResource Key=OSThemeStyle}">This text is green or red depending on Light (or default) or Dark (through style)</Label>
+            <Label TextColor="{DynamicResource MyColor}">This text is HotPink or Yellow depending on Light (or default) or Dark (through AppThemeColor)</Label>
+            <Label TextColor="{StaticResource MyColor}">This text is HotPink or Yellow depending on Light (or default) or Dark (through AppThemeColor)</Label>
+            <Label x:Name="cssStyledLabel">This text is Purple or Orange depending on Light (or default) or Dark (through CSS)</Label>
+            <Label>The image below is a Xamarin logo or fruits depending on Light (or default) or Dark</Label>
+            <Image Source="{OnAppTheme Light=xamarinlogo, Dark=Fruits.jpg}" />
```
jsuarezruiz: It is not important, but the xamarinlogo image will not work in UWP (path).

jsuarezruiz:
The failed tests seem unrelated.

Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
```diff
@@ -18,7 +18,7 @@ namespace Xamarin.Forms.ControlGallery.Android
 
 	[Activity(Label = "Control Gallery", Icon = "@drawable/icon", Theme = "@style/MyTheme",
 		MainLauncher = true, HardwareAccelerated = true, 
-		ConfigurationChanges = ConfigChanges.ScreenSize | ConfigChanges.Orientation | ConfigChanges.ScreenLayout | ConfigChanges.SmallestScreenSize)]
+		ConfigurationChanges = ConfigChanges.ScreenSize | ConfigChanges.Orientation | ConfigChanges.ScreenLayout | ConfigChanges.SmallestScreenSize | ConfigChanges.UiMode)]
```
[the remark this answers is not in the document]
jfversluis: Yep! Opened a PR for that

Xamarin.Forms.Controls/GalleryPages/OSThemesGalleryXaml.xaml
jfversluis: Ideally yes, just not sure how. Could be something for a next iteration

Xamarin.Forms.Platform.UAP/WindowsBasePlatformServices.cs
jfversluis: Nothing in this class is disposed or unsubscribed. I think this lives for the entire lifecycle of the application, right? So it only goes away when the app stops and then is cleaned up with everything else?

Xamarin.Forms.Controls/GalleryPages/OSThemesGalleryXaml.xaml
jfversluis: Fixed!

Xamarin.Forms.Controls/GalleryPages/OSThemesGalleryXaml.xaml
jfversluis: Implemented some code that should make the runtime updating experience better

Xamarin.Forms.Core/AppThemeColor.cs
```diff
@@ -0,0 +1,84 @@
...
+		public static readonly BindableProperty DefaultProperty = BindableProperty.Create(nameof(Default), typeof(Color), typeof(AppThemeColor), default(Color), propertyChanged: (bo, __, ___) => UpdateActualValue(bo));
+
+		public Color Default
+		{
+			get => (Color)GetValue(DefaultProperty);
+			set => SetValue(DefaultProperty, value);
+		}
+
+		private Color _actualValue;
+		public Color ActualValue
```
PureWeen, line 37: Why `ActualValue` and not just `Value`?

Xamarin.Forms.Core/AppThemeColor.cs
```diff
@@ -0,0 +1,84 @@
...
+				case AppTheme.Light:
+					return appThemeColor.IsSet(LightProperty) ? appThemeColor.Light : (appThemeColor.IsSet(DefaultProperty) ? appThemeColor.Default : default(Color));
+				case AppTheme.Dark:
+					return appThemeColor.IsSet(DarkProperty) ? appThemeColor.Dark : (appThemeColor.IsSet(DefaultProperty) ? appThemeColor.Default : default(Color));
+			}
+		}
+
+		public void Dispose()
+		{
+			Application.Current.RequestedThemeChanged -= RequestedThemeChanged;
```
PureWeen, line 61: Same comment here about dispose

Xamarin.Forms.Platform.Android/Renderers/PageRenderer.cs
```diff
@@ -102,6 +103,13 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
 				UpdateHeight();
 		}
 
+		protected override void OnConfigurationChanged(Configuration newConfig)
+		{
+			base.OnConfigurationChanged(newConfig);
+
+			Element.Resources.Reload();
```
PureWeen, line 110: Does it makes sense for all these reload calls to be tied to something else at the explat level? Like a themechanged event or something there on the Page itself?

Or instead of calling reload can it call into some other method on the Page itself?

Xamarin.Forms.Platform.UAP/WindowsBasePlatformServices.cs
```diff
@@ -158,6 +161,11 @@ public SizeRequest GetNativeSize(VisualElement view, double widthConstraint, dou
 			return Platform.GetNativeSize(view, widthConstraint, heightConstraint);
 		}
 
+		void UISettingsColorValuesChanged(UISettings sender, object args)
+		{
+			Device.BeginInvokeOnMainThread(() => Application.Current?.OnRequestedThemeChanged(new AppThemeChangedEventArgs(Application.Current.RequestedTheme)));
```
PureWeen, line 166: Can we change this to just use the native dispatcher?

```C#
this.Dispatcher
```

Xamarin.Forms.Platform.iOS/Forms.cs
```diff
@@ -411,6 +412,95 @@ public SizeRequest GetNativeSize(VisualElement view, double widthConstraint, dou
...
+
+				if (throwIfNull && viewController == null)
+					throw new InvalidOperationException("Could not find current view controller.");
+
+				return viewController;
+			}
+
+			static UIWindow GetCurrentWindow(bool throwIfNull = true)
+			{
+				var window = UIApplication.SharedApplication.KeyWindow;
```
PureWeen, line 484: should this code use?
https://github.com/xamarin/Xamarin.Forms/blob/30228a2f3587e1b8c10674076a4a6a64908a1c19/Xamarin.Forms.Platform.iOS/Extensions/UIApplicationExtensions.cs#L8

Xamarin.Forms.Platform.iOS/Forms.cs
```diff
@@ -411,6 +412,95 @@ public SizeRequest GetNativeSize(VisualElement view, double widthConstraint, dou
...
+#if __IOS__ || __TVOS__
+
+			static UIViewController GetCurrentUIViewController() =>
+				GetCurrentViewController(false);
+
+			static UIViewController GetCurrentViewController(bool throwIfNull = true)
+			{
+				UIViewController viewController = null;
+
+				var window = UIApplication.SharedApplication.KeyWindow;
```
PureWeen, line 455: should this code use
https://github.com/xamarin/Xamarin.Forms/blob/30228a2f3587e1b8c10674076a4a6a64908a1c19/Xamarin.Forms.Platform.iOS/Extensions/UIApplicationExtensions.cs#L8

Xamarin.Forms.Xaml/MarkupExtensions/OnAppThemeExtension.cs
```diff
@@ -0,0 +1,142 @@
...
+
+			if (!(value is Binding))
+				return new Binding(nameof(ActualValue), source: this);
+			else
+				return ret;
+		}
+
+		public void Dispose()
+		{
+			Application.Current.RequestedThemeChanged -= RequestedThemeChanged;
```
PureWeen, line 122: Same Dispose comment

Xamarin.Forms.Core/AppThemeColor.cs
```diff
@@ -0,0 +1,84 @@
+﻿using System;
+
+namespace Xamarin.Forms
+{
+	public class AppThemeColor : BindableObject, IDisposable
```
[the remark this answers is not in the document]
jfversluis: That makes much more sense, done!

Xamarin.Forms.Core/AppThemeColor.cs
jfversluis: At the time I thought it made sense. It can be either the `Light`, `Dark` or `Default` one and this is the _actual_ value that is currently provided by this. But I guess a value is always an actual value :P 

Changed

Xamarin.Forms.Core/AppThemeColor.cs
jfversluis: Removed

Xamarin.Forms.Platform.Android/Renderers/PageRenderer.cs
jfversluis: I didn't have the event before, but now that I do it makes more sense to do that more centrally, yes

Xamarin.Forms.Platform.iOS/Forms.cs
jfversluis: Ooh good call. I honestly just copied this from Essentials because this code will be swapped out for that. In fact, when looking again, `GetCurrentWindow` is not even used so I removed it.

Xamarin.Forms.Platform.iOS/Forms.cs
jfversluis: Same as above; Essentials, but changed it for now to be sure.

Xamarin.Forms.Xaml/MarkupExtensions/OnAppThemeExtension.cs
jfversluis: Removed

Xamarin.Forms.Platform.UAP/WindowsBasePlatformServices.cs
jfversluis: Yes we can!

Xamarin.Forms.Core/AppTheme.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Xamarin.Forms
+{
+	public enum AppTheme
+	{
+		Unspecified,
+		Light,
+		Dark
```
StephaneDelcroix, line 7: are we sure we won't need to extend this list ? if so, we might want to use strings...

Xamarin.Forms.Core/AppThemeColor.cs
StephaneDelcroix, line 5: this definitely doesn't look like a BindableObject. It's never parented, so it probably never gets a BindingContext set...

Xamarin.Forms.Core/OnAppTheme.cs
```diff
@@ -0,0 +1,84 @@
+﻿using System;
+
+namespace Xamarin.Forms
+{
+	public class OnAppTheme<T> : BindableObject, IDisposable
```
StephaneDelcroix, line 5: not a BO. look at OnPlatform, OnDevice

Xamarin.Forms.Core/ResourceDictionary.cs
```diff
@@ -327,6 +327,13 @@ void OnValuesChanged(params KeyValuePair<string, object>[] values)
 			ValuesChanged?.Invoke(this, new ResourcesChangedEventArgs(values));
 		}
 
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public void Reload()
+		{
+			foreach (var mr in MergedResources)
+				OnValuesChanged(mr);
```
StephaneDelcroix, line 334: I'd like to avoid doing that

Xamarin.Forms.Core/OnAppTheme.cs
StephaneDelcroix: also, you might want a xaml markup language extension

Xamarin.Forms.Core/AppTheme.cs
MagicAndre1981: @StephaneDelcroix [Essentials defined it first in this way](https://docs.microsoft.com/en-us/xamarin/essentials/app-theme?context=xamarin%2Fxamarin-forms&tabs=android#obtaining-theme-information), so I think this was done to keep this in sync

Xamarin.Forms.Core/ResourceDictionary.cs
jfversluis: In the latest changes it has at least changed to internal. I know it's not pretty and open to alternatives. But the problem is that the actual value doesn't really change. Depending on the theme the implicit operator will return a different color and that is why it's not picked up normally.

The only time this happens is when the theme changes on OS level and then a number of expensive operations already happen to redraw the whole screen already.

Xamarin.Forms.Core/AppThemeColor.cs
jfversluis: Because I'm using bindable properties? Should I be using something different?

Xamarin.Forms.Core/OnAppTheme.cs
jfversluis: Because I'm using bindable properties? Should I be using something different? And there is an extension as well; https://github.com/xamarin/Xamarin.Forms/pull/9958/files#diff-c65a7e8e0eed508c58cd0802b61b2cb1

Xamarin.Forms.Core/AppTheme.cs
jfversluis: Indeed follow the Essentials way for now. I'd like to think this is like the OS level theming and on top of that people can have their own custom theming, probably as an extension of this (naming will be a hell...). But if you implement your own theme as a user, you still might want to also follow the light and dark mode by the OS and show different colors based on that.

PR #9967, opened 2020-03-16
Remove RadioButton.ButtonSource API

### LINKED ISSUE #9948: [Bug] RadioButton.ButtonSource is only implemented on Android
### Description

The new `RadioButton` control has a `ButtonSource` property that's only implemented on Android. If that's going to remain the case, shouldn't it be a platform-specific?

### Basic Information

- Version with issue: 4.6-pre1

jfversluis:
### Description of Change ###

Removes the `RadioButton.ButtonSource` API for now. 

This functionality was only implemented on Android and needs implementation on iOS, UWP and MacOS.

### Issues Resolved ### 

- fixes #9948

### API Changes ###

 Removed:
 - `public static readonly BindableProperty ButtonSourceProperty`
 - `public ImageSource ButtonSource { get; set; }`

### Platforms Affected ### 

- Core/XAML (all platforms)
- Android

Android users on the 4.6-pre1 version will lose the ability to use a custom image for the radiobutton. When already in use, there will be a compilation error since the API has been removed.

Not applicable

### Testing Procedure ###
N/A

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/RadioButton.cs
+++ b/Xamarin.Forms.Core/RadioButton.cs
@@ -20,8 +20,9 @@ namespace Xamarin.Forms
 		public static readonly BindableProperty GroupNameProperty = BindableProperty.Create(
 			nameof(GroupName), typeof(string), typeof(RadioButton), null, propertyChanged: (b, o, n) => ((RadioButton)b).OnGroupNamePropertyChanged((string)o, (string)n));
 
-		public static readonly BindableProperty ButtonSourceProperty = BindableProperty.Create(
-			nameof(ButtonSource), typeof(ImageSource), typeof(RadioButton), null);
+		// TODO Needs implementations beyond Android
+		//public static readonly BindableProperty ButtonSourceProperty = BindableProperty.Create(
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Core/RadioButton.cs
```diff
@@ -37,11 +38,12 @@ public string GroupName
 			set { SetValue(GroupNameProperty, value); }
 		}
 
-		public ImageSource ButtonSource
-		{
-			get { return (ImageSource)GetValue(ButtonSourceProperty); }
-			set { SetValue(ButtonSourceProperty, value); }
-		}
+		// TODO Needs implementations beyond Android
+		//public ImageSource ButtonSource
```
PureWeen: what happens right now if people use ImmageSource?

jfversluis: It’s kind of a funny API anyway. You’re able to set a custom image for the checkbox, but you need it together with the VSM to dynamically set the checked state. See the example in the gallery I commented out. 

If people use it on platforms where it’s not implemented, nothing happens right now.

PR #10015, opened 2020-03-18
[iOS] Fix issue refreshing and disabling the RefreshView

### LINKED ISSUE #9326: [Bug] RefreshView should delay effects of IsEnabled while a refresh is happening and iOS is handling IsEnabled false incorrectly
### Description

In a lot of cases Commands will get marked as CanExecute false while they are executing. This confuses the refreshview as the refreshview disables or, in the iOS case, removes the view from the superview. 

IsEnabled should be ignored until after IsRefreshing toggles back. At which point IsEnabled should be processed after the Refresh Control is no longer visible. Ideally we can tap into something on the RefreshControl to know when it's no longer visible.

On iOS the code to disable doesn't seem correct. It purely removes the control from the superview which seems to causes issues with the UIScrollView scenario where it's just set ono the control. Setting it back on the control doesn't appear to add it back to the superview

### Steps to Reproduce

1.  https://github.com/RLittlesII/Reactive.Sandbox/tree/master/xamarin/forms/ReactiveUI.2299
2.  change to a XAML binding https://github.com/reactiveui/ReactiveUI/issues/2299#issuecomment-575558274

### Expected Behavior
Ii should be able to keep refreshing after the first refresh and the refresh control shouldn't be removed from the superview while refreshing

### Actual Behavior

RefreshControl is removed from the superview during refresh and even the IsEnabled toggles back the RefreshControl is never added back

### Basic Information

- Version with issue: 4.3

jsuarezruiz:
### Description of Change ###

Avoid remove the `UIRefreshControl` disabling the RefreshView while refreshing. If the UIRefreshControl is refreshing we just disable the control but not remove it from the parent view.

### Issues Resolved ### 

- fixes #9326 

### API Changes ###

 None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###
Launch Core Gallery and navigate to the issue 9326. Try to pull to refresh more than once. If the RefreshView continues refreshing after the first time, the test has passed.

### PR Checklist ###


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue9326.xaml.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+		protected override void Init()
+		{
+
+		}
+    }
+
+	/*
+    public class Issue9326ViewModel : ReactiveObject
```
rmarinho: Should this be comment out?

jsuarezruiz: Good question. The issue is using ReactiveUI refreshing and disabling the RefreshView at the same time. So, we need a reference to ReactiveUI package to test it. I include this comment: https://github.com/xamarin/Xamarin.Forms/blob/f309b8460dffef5b283b2a3370329ff05bd5e7a4/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue9326.xaml.cs#L19
Should we include the reference an uncomment this lines?

PureWeen: I don't think the UI test needs reactiveUI.. The behavior is just toggling CanExecute to false directly after executing so if you do the following order of commands


Execute => inside execute fire canexecutechanged which will set canexecute to false so the command gets disabled

At this point the RefreshView should stay visible

PR #10016, opened 2020-03-18
Platformtests IVT

hartez:
### Description of Change ###

Moving the platform tests to their own assemblies; adding InternalsVisibleTo so we can test platform internals.


hartez:
Test failures were all from the "Chrome Terms of Service" pop-up bug.

hartez:
All testing code, nothing that ships.

PR #10028, opened 2020-03-19
Remove all uses of md5

PureWeen:
### Description of Change ###

Replace all uses of md5 for creating a hash. The only place we really use this is for hashing a uri so we have a consistent length to use for a hashkey

### API Changes ###

Added:
 - string IPlatformServices.GetHash

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS
- Android
- UWP

### Testing Procedure ###
- UI tests pass
- Images load and are cached

### PR Checklist ###


transis2:
Hey @pureween, I'm curious about The reason to deprecate md5.
O use md5 a lot in my projects

PR #10047, opened 2020-03-21
NotchMode on xamarin forms

### LINKED ISSUE #10090: [Enhancement] Notch mode on Xamarin Forms!
## Summary
Since Android and IOS has a notch space and works with some "safe area", Xamarin forms can provide it for use it at the page level

## API Changes

NochMode Page.NotchMode { get; set; } //Bindable Property

The NotchMode API will change native notch space behavior

## Intended Use Case
Unify the notch configuration for android and IOS

felipebaltazar:
### Description of Change ###
I'm creating this Pr to suggest a way to manipulate Notch mode from xamarin forms, since android and IOS has an implementation of this feature.

**If this feature considered valid for Xamarin forms, I will include the Tests and images**

### Issues Resolved ### 
fixes #10090 

### API Changes ###

Added:
 - NochMode Page.NotchMode { get; set; } //Bindable Property
 - CutoutMode AndroidSpecific.Page.CutoutMode { get; set; }

 
### Platforms Affected ### 
- Core/XAML (all platforms)
- Android

When Page.NotchMode is set, the notch safe area will be manipulated.

- **NotchMode.PlatformDefault** will keep default platform behaviour ex: Ios UseSafeArea= false and Android LAYOUT_IN_DISPLAY_CUTOUT_MODE_DEFAULT

- **NotchMode.UseFullScreen** will use notch area ex: Ios UseSafeArea= false and Android LAYOUT_IN_DISPLAY_CUTOUT_MODE_SHORT_EDGES

- **NotchMode.SaveNotchArea** will use notch area ex: Ios UseSafeArea= true and Android LAYOUT_IN_DISPLAY_CUTOUT_MODE_NEVER

### Testing Procedure ###
- Set Page.NotchMode = NotchMode.UseFullScreen and your page will use the notch area
- Set Page.NotchMode = NotchMode.SaveNotchArea and your page will save the notch area
- Set Page.NotchMode = NotchMode.DefaultPlatform and your page will keep the defaut platform behavior for the notch area

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/NotchMode.cs
+++ b/Xamarin.Forms.Core/NotchMode.cs
@@ -1,0 +1,9 @@
+﻿namespace Xamarin.Forms
+{
+	public enum NotchMode
+	{
+		PlatformDefault = 0,
+		UseFullScreen = 1,
+		SaveNotchArea = 2
+	}
+}
--- a/Xamarin.Forms.Core/Page.cs
+++ b/Xamarin.Forms.Core/Page.cs
@@ -8,6 +8,9 @@ using System.Threading;
 using System.Threading.Tasks;
 using Xamarin.Forms.Internals;
 using Xamarin.Forms.Platform;
+using Xamarin.Forms.PlatformConfiguration;
+using Xamarin.Forms.PlatformConfiguration.AndroidSpecific;
+using Xamarin.Forms.PlatformConfiguration.iOSSpecific;
 
 namespace Xamarin.Forms
 {
@@ -41,6 +44,9 @@ namespace Xamarin.Forms
 		[Obsolete("IconProperty is obsolete as of 4.0.0. Please use IconImageSourceProperty instead.")]
 		[EditorBrowsable(EditorBrowsableState.Never)]
 		public static readonly BindableProperty IconProperty = IconImageSourceProperty;
+
+		public static readonly BindableProperty NotchModeProperty = BindableProperty.Create(nameof(NotchMode), typeof(NotchMode), typeof(Page), NotchMode.PlatformDefault, propertyChanged:(bo, o, n)=>((Page)bo).OnNotchModeChanged((NotchMode)n));
+
 
 		readonly Lazy<PlatformConfigurationRegistry<Page>> _platformConfigurationRegistry;
 
@@ -103,6 +109,12 @@ namespace Xamarin.Forms
 		{
 			get { return (bool)GetValue(IsBusyProperty); }
 			set { SetValue(IsBusyProperty, value); }
+		}
+
+		public NotchMode NotchMode
+		{
+			get => (NotchMode)GetValue(NotchModeProperty);
+			set => SetValue(NotchModeProperty, value);
 		}
 
 		public Thickness Padding
@@ -489,6 +501,12 @@ namespace Xamarin.Forms
 				return;
 			foreach (IElement item in args.NewItems)
 				item.Parent = this;
+		}
+
+		void OnNotchModeChanged(NotchMode newValue)
+		{
+			On<iOS>().SetUseSafeArea(newValue == NotchMode.SaveNotchArea);
+			On<Android>().SetCutoutMode((CutoutMode)(int)newValue);
 		}
 
 		bool ShouldLayoutChildren()
--- a/Xamarin.Forms.Core/PlatformConfiguration/AndroidSpecific/CutoutMode.cs
+++ b/Xamarin.Forms.Core/PlatformConfiguration/AndroidSpecific/CutoutMode.cs
@@ -1,0 +1,9 @@
+﻿namespace Xamarin.Forms.PlatformConfiguration.AndroidSpecific
+{
+	public enum CutoutMode
+	{
+		Default = 0,
+		ShortEdges = 1,
+		Never = 2
+	}
+}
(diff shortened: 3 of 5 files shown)


PureWeen:
@felipebaltazar I'm not super familiar with the android notch mode settings. Do these basically just pad the layout? 

In theory could we make safe area cross platform
https://docs.microsoft.com/en-us/xamarin/xamarin-forms/platform/ios/page-safe-area-layout

And then just use that?

I'm not saying we should because I'm not a big fan of the SafeArea setting :-) but just curious

felipebaltazar:
> @felipebaltazar I'm not super familiar with the android notch mode settings. Do these basically just pad the layout?
> 
> In theory could we make safe area cross platform
> https://docs.microsoft.com/en-us/xamarin/xamarin-forms/platform/ios/page-safe-area-layout
> 
> And then just use that?

@PureWeen 

Maybe we can use the "SafeArea", but we will need to check the `WindowManager.LayoutParams` flags. 
I think that an enum can be more flexible than a boolean, in this case.
The Android API provides 3 options for NotchMode and in the future, we can grow this feature (Cross) with the new API features.

https://developer.android.com/reference/android/view/WindowManager.LayoutParams#layoutInDisplayCutoutMode

We can get the Notch insets from android API too:
```C#
 WindowInsets.GetDisplayCutout();
```

For IOS I keep using the [feature that you have mentioned ](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/platform/ios/page-safe-area-layout)



> I'm not saying we should because I'm not a big fan of the SafeArea setting :-) but just curious

Yeah! I agree! I'm not saying that we need to implement in this way, too :D
I'm only opening a discussion about centralizing the Notch mode at cross-platform and

StephaneDelcroix:
I would prefer exposing a SafeArea (or the negative UnsafeArea) property, giving the full power on what to do. I would probably bind that info (OneTime binding) to the Margin of the content...

Tommigun1980:
Please don't introduce another Android-specific API for something that already exists (albeit in iOS form only), but make the existing one work with Android also.

A use case that must be supported is to be able to just read the padding values, as for example having a background image in an element must fill the entire page, so the user would then assign some padding values to whatever elements he deems fit. So an umbrella 'add automatic padding to the entire page' wouldn't be enough, and the SafeAreaInsets() method already exists so it'd be convenient if Android would start to support it as well.

Thanks.

Tommigun1980:
How about just...
a) Keeping the safe area API exactly as it is
b) Leave it in the iOS namespace for backwards compatibility
c) Add support for Android
d) Introduce it in a different namespace

That's how other Xamarin things work - they abstract the iOS and Android implementations to one common API, and I think it'd be a mistake to introduce a completely separate API.

felipebaltazar:
> How about just...
> a) Keep the safe area API exactly as it is
> b) Leave it in the iOS namespace for backwards compatibility
> c) Add support for Android
> d) Introduce it in a different namespace, now working on both platforms
> e) Developers would get Android support by changing the namespace, or keep it iOS only by doing nothing
> 
> ( f) Other possible extensions/refinements to the safe area API should come in a different pull request, work on both platforms, and be tracked separately from Android support imho )
> 
> That's how other Xamarin things work - they abstract the iOS and Android implementations to one common API, and I think it'd be a mistake to introduce a completely separate API for Android.

@Tommigun1980 Yeah good points here, thanks!

My first question here: "Is this a valid feature?"
I've opened this PR as a suggestion, I don't know if it will be useful or only more code to maintain.

About your points, I think that, since it will be merged on MAUI we can think in another better API.
For android, we need to set 3 different flags and for IOS we have only 2 states (true and false) so, it does not will provide the fully platform-specific. I Agree with the "**E**" and "**F**" points.

Thank you again for your feedback! 💪

Tommigun1980:
> > How about just...
> > a) Keep the safe area API exactly as it is
> > b) Leave it in the iOS namespace for backwards compatibility
> > c) Add support for Android
> > d) Introduce it in a different namespace, now working on both platforms
> > e) Developers would get Android support by changing the namespace, or keep it iOS only by doing nothing
> > ( f) Other possible extensions/refinements to the safe area API should come in a different pull request, work on both platforms, and be tracked separately from Android support imho )
> > That's how other Xamarin things work - they abstract the iOS and Android implementations to one common API, and I think it'd be a mistake to introduce a completely separate API for Android.
> 
> @Tommigun1980 Yeah good points here, thanks!
> 
> My first question here: "Is this a valid feature?"
> I've opened this PR as a suggestion, I don't know if it will be useful or only more code to maintain.
> 
> About your points, I think that, since it will be merged on MAUI we can think in another better API.
> For android, we need to set 3 different flags and for IOS we have only 2 states (true and false) so, it does not will provide the fully platform-specific. I Agree with the "**E**" and "**F**" points.
> 
> Thank you again for your feedback! 💪

Oh I think it’s absolutely a valid, nay, mandatory feature, but best served as a unified API. Thanks.

jfversluis:
This will not happen anymore for Xamarin.Forms. If this is still important to you, make sure to check the [.NET MAUI repo](https://github.com/dotnet/maui) and see if it's already on the roadmap or open an issue with a detailed feature request.

We appreciate the work that has gone into this PR although it won't be merged. Thank you so much for your precious time that you have donated to us. Hopefully it was still a fun and a great learning experience for you :)

PR #10072, opened 2020-03-24
Inverse Parent/Child fields for VisualDiagnostics.SendVisualTreeChanged

drasticactions:
### Description of Change ###

Inverse Parent/Child fields parameters for VisualDiagnostics.SendVisualTreeChanged.

### Issues Resolved ### 
In the places that currently use `VisualDiagnostics.SendVisualTreeChanged`, the parameters being passed in are flipped so the relationship to the objects they represent are incorrect.

For example, when the `Element` sends that it's visual tree has changed, you could end up with a Button being a parent of a StackPanel, when really, the Button is _inside_ the StackPanel.

### API Changes ###

The API Itself remains the same, but parameters being passed in are flipped. If anyone was depending on those parameters being in that order, they would need to flip their logic to the correct one.

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###

- Attach to `VisualDiagnostics.VisualTreeChanged`, make changes to the Visual Tree.

### PR Checklist ###


Xamarin.Forms.Core/Element.cs
```diff
@@ -201,7 +201,7 @@ public Element Parent
 					SetInheritedBindingContext(this, null);
 				}
 
-				VisualDiagnostics.SendVisualTreeChanged(this, value);
+				VisualDiagnostics.SendVisualTreeChanged(value, this);
```
StephaneDelcroix: this change 100% right, value IS the parent 👍

Xamarin.Forms.Xaml/XamlLoader.cs
```diff
@@ -93,7 +93,7 @@ public static void Load(object view, string xaml, Assembly rootAssembly, bool us
 					var rootnode = new RuntimeRootNode(new XmlType(reader.NamespaceURI, reader.Name, null), view, (IXmlNamespaceResolver)reader) { LineNumber = ((IXmlLineInfo)reader).LineNumber, LinePosition = ((IXmlLineInfo)reader).LinePosition };
 					if (XamlFilePathAttribute.GetFilePathForObject(view) is string path) {
 						VisualDiagnostics.RegisterSourceInfo(view, new Uri($"{path};assembly={view.GetType().GetTypeInfo().Assembly.GetName().Name}", UriKind.Relative), ((IXmlLineInfo)rootnode).LineNumber, ((IXmlLineInfo)rootnode).LinePosition);
-						VisualDiagnostics.SendVisualTreeChanged(null, view);
+						VisualDiagnostics.SendVisualTreeChanged(view, null);
```
StephaneDelcroix: I don't think this change is required

Xamarin.Forms.Xaml/XamlLoader.cs
```diff
@@ -144,7 +144,7 @@ public static object Create(string xaml, bool doNotThrow, bool useDesignProperti
 					if (XamlFilePathAttribute.GetFilePathForObject(inflatedView) is string path)
 					{
 						VisualDiagnostics.RegisterSourceInfo(inflatedView, new Uri($"{path};assembly={inflatedView.GetType().GetTypeInfo().Assembly.GetName().Name}", UriKind.Relative), ((IXmlLineInfo)rootnode).LineNumber, ((IXmlLineInfo)rootnode).LinePosition);
-						VisualDiagnostics.SendVisualTreeChanged(null, inflatedView);
+						VisualDiagnostics.SendVisualTreeChanged(inflatedView, null);
```
StephaneDelcroix: same here

Xamarin.Forms.Xaml/XamlLoader.cs
drasticactions: Reverted: We're not using this event and talking offline, I can see your point. It makes sense to leave it for now.

Xamarin.Forms.Xaml/XamlLoader.cs
drasticactions: same as above

drasticactions:
Re-rereverted the last commit after fully verifying if we act on XamlLoader.

Turns out, we totally do! It's used when we're editing references on Pages. So that can stay. 👍

drasticactions:
@StephaneDelcroix Is this going to be merged? Need a decision soon if it is or not...

PR #10079, opened 2020-03-24
[Android] Let user swap out Flyout Header on Shell

### LINKED ISSUE #6161: [Bug] Changing the Shell Flyout Header after it's already rendered doesn't work
### Description
If you try to change the Shell .FlyoutHeader after the app is already running it won't work. On iOS it just becomes blank

### Steps to Reproduce

1. start shell app
2. use a button or similar so you can trigger a change to flyout header
3. click the button

### Expected Behavior
flyoutheader should re-render

### Actual Behavior
it does not.

PureWeen:
### Description of Change ###

Watch for flyout header changes and swap it out

### Issues Resolved ### 

- partially fixes #6161

### Platforms Affected ### 
- Android

### Testing Procedure ###
- automated UI Tests

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android.UnitTests/ShellTests.cs
+++ b/Xamarin.Forms.Platform.Android.UnitTests/ShellTests.cs
@@ -11,7 +11,9 @@ using Android.Support.Design.Widget;
 using Android.Views;
 using Android.Widget;
 using NUnit.Framework;
+using Xamarin.Forms;
 using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Platform.Android.UnitTests;
 
 [assembly: ExportRenderer(typeof(TestShell), typeof(TestShellRenderer))]
 namespace Xamarin.Forms.Platform.Android.UnitTests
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Platform.Android.UnitTests/ShellTests.cs
```diff
@@ -0,0 +1,100 @@
...
+		{
+			var shell = CreateShell();
+			var initialHeader = new Label() { Text = "Hello" };
+			var newHeader = new Label() { Text = "Hello part 2" };
+			shell.FlyoutHeader = initialHeader;
+			var addedView = await Device.InvokeOnMainThreadAsync(() =>
+			{
+				var r = GetRenderer(shell);
+				var window = (Context.GetActivity()).Window;
+				(window.DecorView as global::Android.Views.ViewGroup).AddView(r.View);
```
PureWeen: @hartez the only way I could get shell to work was to add the view to the window. If I don't do this then Fragment inflation doesn't work. Maybe you know a better way to do this? Instead of just adding it to the global window?

hartez: I don't know a better way off the top of my head.

PureWeen:
Test failures are all the ExpanderView which has been fixed

PR #10089, opened 2020-03-25
[Enhancement] Add WebView UserAgent

### LINKED ISSUE #8432: [Enhancement] WebView - possibility to set custom user agent
## Summary
It would be nice if there was a possibility to set custom user agent forWebView in Xamarin.Forms. Thanks to that we could easily differentiate between WebView and mobile browser on the webpage.

On this website it is described how to set custom user agent on Android

https://mobiforge.com/research-analysis/webviews-and-user-agent-strings

image

## API Changes

In order to set custom user agent for a webview with x:Name ="webView" we could write something like this:
```
webView.setUserAgentString("My funky UA string")
```
This should work for Android and iOS

## Intended Use Case
Thanks to that feature we would be able to easily differentiate between webview traffic and browser traffic.

CliffAgius:
### Description of Change ###

Added the ability for the UserAgent used by the Webview to be set/changed for IOS/Android and UWP.

There is now a new Bindable property called setUserAgentString that when set will change the user agent of the HTTPRequest when using the WebView.

### Issues Resolved ### 

- fixes #8432 

### API Changes ###

Added:
 - string setUserAgentString { get; set; } //Bindable Property

### Platforms Affected ### 

- iOS
- Android
- UWP

None

Not applicable

### Testing Procedure ###
In the ControlGallery opening the Issue 8432 will show a WebView on the device that will point to the website whatismybrowser.com and in here you will see the UserAgent that has been set.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/WebView.cs
+++ b/Xamarin.Forms.Core/WebView.cs
@@ -40,7 +40,7 @@ namespace Xamarin.Forms
 
 		public static readonly BindableProperty CookiesProperty = BindableProperty.Create(nameof(Cookies), typeof(CookieContainer), typeof(WebView), null);
 
-		public static readonly BindableProperty UserAgentProperty = BindableProperty.Create(nameof(setUserAgentString), typeof(string), typeof(WebView), default(string));
+		public static readonly BindableProperty UserAgentProperty = BindableProperty.Create(nameof(UserAgentString), typeof(string), typeof(WebView), default(string));
 
 		readonly Lazy<PlatformConfigurationRegistry<WebView>> _platformConfigurationRegistry;
 
@@ -79,7 +79,7 @@ namespace Xamarin.Forms
 			set { SetValue(CookiesProperty, value); }
 		}
 
-		public string setUserAgentString
+		public string UserAgentString
 		{
 			get { return (string)GetValue(UserAgentProperty); }
 			set { SetValue(UserAgentProperty, value); }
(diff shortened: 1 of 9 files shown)


Xamarin.Forms.Core/WebView.cs
```diff
@@ -77,6 +79,12 @@ public CookieContainer Cookies
 			set { SetValue(CookiesProperty, value); }
 		}
 
+		public string setUserAgentString
```
VladislavAntonyuk: UserAgentString, property name should be a noun, when method should be a verb

CliffAgius: @VladislavAntonyuk I am using what was suggested in the Issue and the subsequent chat as the name, but feel free to suggest something better.
I must admit that I have only implemented the Setting of the UserAgentString hence the name making sense in that regard so possibly should remove the Getter.

Open to suggestions.

CliffAgius:
Any progress on this be nice to cross it off the Todo list as done and pick another Issue to look at.

CliffAgius:
Just wondering if this can be either accepted or rejected please.

@samhouts @VladislavAntonyuk

VladislavAntonyuk:
Please resolve conflicts. Also is it possible to implement it for other platforms (macOS, WPF, Tizen, GTK)?

CliffAgius:
@VladislavAntonyuk I have fixed the Conflicts and tested it here as working.

As for the other platforms they are not platforms I have experience with or ability to test here so if they are required they will need to be done by others I'm afraid sorry.

jfversluis:
Was on the fence to still add this one, but since the rebase was hurtful I think I'll have to leave this behind in order not to break other things. Thanks so much @CliffAgius for your time and effort on this! I'm looking forward to seeing this in .NET MAUI :)

PR #10120, opened 2020-03-28
[MacOS] Fix resetting TextColor and Font on Label

### LINKED ISSUE #10101: [Bug] Changing text on a Label on macOS resets font color to Black
### Description
The Issue started appearing in XF 4.5, 4.4 is OK

### Steps to Reproduce

1. Define a Label on a page with some text in it
2. Define Label style in app.xaml
```
        <!-- FONTS -->
        <OnPlatform x:TypeArguments="Font" x:Key="FontNormal">
            <On Platform="macOS" Value="IBM Plex Mono, 15" />
            <On Platform="iOS" Value="IBM Plex Mono, 15" />
            <On Platform="WPF" Value="./Fonts/IBMPlexMono-Regular.ttf#IBM Plex Mono, 15" />
            <On Platform="Android, iOS" Value="IBMPlexMono-Regular.ttf#IBMPlexMono-Regular, 15" />
        </OnPlatform>
        <!-- LABELS -->
        <Style TargetType="Label" ApplyToDerivedTypes="True">
            <Setter Property="Font" Value="{StaticResource FontNormal}" />
            <Setter Property="TextColor" Value="Red" />
        </Style>
```
3. Run the app, see the color of text is applied
4. Change the text of the Label in code behind (e.g. on button click)

### Expected Behavior

The text color stays as it is defined in Style

### Actual Behavior

The color is set to black

jfversluis:
### Description of Change ###

Reorder setting the initial properties for `TextColor` and `Font` on MacOS & iOS to overcome losing values in subsequent calls to set other properties.

#### Background 

Whenever you set `StringValue` it will automatically fill `AttributedStringValue` as well. On iOS (while property names are slightly different) it seems to be no problem if you manipulate the attributed string directly, for instance when you set the character spacing and then set the text color. It will update it for you accordingly.

However, on MacOS that doesn't seem to be the case. From what I can see, whenever you change something in the `AttributedStringValue` it will lose sync and then trying to set the font or text color will have no effect. Unfortunately, unlike for iOS, the behavior in this area is not documented for macOS.

I also removed code that was initially dropped in to overcome sort of the same problem, but effectively was duplicating functionality in a different place.

### Issues Resolved ### 

- fixes #10101

### API Changes ###

 None

### Platforms Affected ### 

- iOS
- MacOS

Should be none, other than that everything starts working as it should 🎉 

Not applicable

### Testing Procedure ###
Navigate to issue 10101 in the control gallery and observe how properties are being preserved when we set a different text.

In addition, make sure that all styling on a label still works as it should on both iOS and MacOS.

### PR Checklist ###


jfversluis:
Thanks for looking into this @ffgiraldez! Are you testing this on your own project? Would you happen to have some repro project that I can try this on?

ffgiraldez:
yes I will try

ffgiraldez:
@jfversluis here is the results:
change TextColor + Text
Using Xamarin.Froms 4.5.0.495
label-color-text-current
Using Xamarin.Froms 4.5.0.1642 
lable-color-text

change only TextColor
Using Xamarin.Froms 4.5.0.1642 
label-color

here is the sample solution
[test-label.zip](https://github.com/xamarin/Xamarin.Forms/files/4413132/test-label.zip)

jfversluis:
@ffgiraldez sorry it took a while. I have added your repro code to my sample here and it seems to work just fine. Are you sure you were looking at the right NuGets? And did a full rebuild, deleted bin/obj folders to make sure there is no caching involved?

ffgiraldez:
Not worries, could be some caching issue, I could not assert now, but if with my repro code works now, should be OK, the key part is change only color and not both text and color at the same time, if that works LGTM 👍

jfversluis:
Hm in that case I might misinterpret what your repro is supposed to do. Could you maybe explain that?

What is the expected behavior? I was thinking that when I click it both the text should be updated _and_ have the right color, but that is not the case?

ffgiraldez:
@jfversluis this is the key
ezgif com-video-to-gif(3)

`Welcome Xamarin Forms!` never change is text, only text color, and using previous Xamarin.Forms 4.5.0.1642 does not work, but using X.F 4.4 it does, at least when I tested it.

jfversluis:
I'm afraid I don't understand. From the repro you added earlier, you simply change the text when the `OnTapped` fires

```
void OnTapped(object sender, EventArgs e)
		{
			var color = RandomColor();
			labelColor.TextColor = color.Color;
			labelText.Text = $"TextColor should be {color.Name}";
			labelColor.Text = $"TextColor should be {color.Name}";
		}
```

Why should it not change text?

ffgiraldez:
sorry, in my first two gifs
the last line
`labelColor.Text = $"TextColor should be {color.Name}";` as you could se top text change and keep sync with bottom text
lable-color-text
but in second example (It was my fault to not put in another page in the repro example) I comment the last line of the `OnTapped` method in order to only change text color of `labelColor` aka top text and you could see the bug
Using Xamarin.Forms 4.5.0.1642 
label-color
Using Xamarin.Forms 4.4
ezgif com-video-to-gif(3)

jfversluis:
Screen Recording 2020-06-03 at 09 10 20

This is what it looks like right now

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue10101.xaml.cs
```diff
@@ -0,0 +1,76 @@
...
+			var colors = GetSystemColorList();
+			return colors[_random.Next(0, colors.Count - 1)];
+		}
+
+		void OnTapped(object sender, EventArgs e)
+		{
+			var color = RandomColor();
+			labelColor.TextColor = color.Color;
+			labelText.Text = $"TextColor should be {color.Name}";
+			labelColor.Text = $"TextColor should be {color.Name}";
```
ffgiraldez, line 66:
```suggestion
			//labelColor.Text = $"TextColor should be {color.Name}";
```
with that you will have this scenario
ezgif com-video-to-gif(3)

jfversluis: Ugh, yeah. That doesn't work :(

jfversluis:
So what I _think_ is happening is that the AttributedString value is now being filled while it wasn't before. The `AddCharacterSpacing` then does not seem to take into account custom font or color.

I'm thinking of maybe reverting the addition of character spacing support on MacOS to unblock people for this and then re-add that later

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue10101.xaml.cs
ffgiraldez: sorry for the not so clear repro scenario 😞

jfversluis: No worries! IT seems it's a very specific bug. It seems that color doesn't update when the text doesn't update. So the repro is definitely valid. I can only appreciate your patience trying to make this clear to us :)

nor0x:
any updates on this issue #10101 ? it's still critical. do you need more repro?

jfversluis:
@nor0x I'm sorry about that. It's still on my radar and I want to have another look. I understand that it's critical to you, at this time macOS is community supported though. That will change during the shift to .NET MAUI. If you have any input on what I can do to improve this PR and make it work in all scenarios I am very open to that.

DennisWelu:
Very important issue to get fixed. Thanks @jfversluis for the efforts.

nor0x:
was this PR ever merged? in my case the issue #10101 seems fixed in version 4.7.0.1239

jfversluis:
That is very interesting @nor0x because no, this was not merged :D 

@ffgiraldez would you be able to confirm that your issue is fixed with this?

ffgiraldez:
Still there @jfversluis , checked using  4.7.0.1239 and previous repro project
> @jfversluis here is the results:
> change TextColor + Text
> Using Xamarin.Forms 4.5.0.495
> label-color-text-current
> Using Xamarin.Forms 4.5.0.1642
> lable-color-text
> 
> change only TextColor
> Using Xamarin.Forms 4.5.0.1642
> label-color
> Using Xamarin.Forms 4.4
> ezgif com-video-to-gif(3)
> 
> here is the sample solution
> [test-label.zip](https://github.com/xamarin/Xamarin.Forms/files/4413132/test-label.zip)

DennisWelu:
@jfversluis @nor0x I am running 1239 and unfortunately am still seeing the issue unless I use a renderer to workaround it.

nor0x:
my bad! just noticed that i was using an old repro with a workaround renderer not disabled. Sorry everyone for the confusion 🙈

PR #10138, opened 2020-03-30
Add equivalent androidx proguard rules

PureWeen:
### Description of Change ###

When a project isn't running androidx migration (so not referencing any support libraries) than the migration doesn't migrate the proguard rules. So we need to have the equivalent androidx proguard rules in place

### Platforms Affected ### 
- Android

### Testing Procedure ###
- download nugets
- create new forms projects
- update essentials/forms to latest androidx dependent packages
- ensure you're targeting API 29
- enable r8/d8 and linker
- make sure project runs

### PR Checklist ###


PureWeen:
I ran the nuget against each of our samples and they all ran successfully

PR #10150, opened 2020-03-31
[UWP] Fix TextBox styling to adopt Dark Theme

jfversluis:
### Description of Change ###

Fix minor styling in the UWP `TextBox` to support the adoption of Dark Theme. The only control that stood out was the `TextBox` which would get a white background and black text when focused. 

This fix uses the background determined by the system as well as the font.

From what I can tell the `TextBox` also doesn't play nicely on pure native UWP. Maybe it's intended if I read this; https://github.com/microsoft/microsoft-ui-xaml/issues/2068 but I disagree 😬 

### Issues Resolved ### 

- related to #9804

### API Changes ###
 
 None

### Platforms Affected ### 

- UWP

In itself; none. When combined with the template fix that is coming _or_ when users remove the fixed `RequestedTheme="Light"` from the App.xaml (in UWP project, like in the changes of this PR) the Dark theme will be applied and now also to the `Entry`

For light theme no change

#### Before
image

#### After

image

### Testing Procedure ###
Open gallery app and find every `Entry` control you can, see if it looks funny

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.ControlGallery.WindowsUniversal/App.xaml
+++ b/Xamarin.Forms.ControlGallery.WindowsUniversal/App.xaml
@@ -2,8 +2,7 @@
     x:Class="Xamarin.Forms.ControlGallery.WindowsUniversal.App"
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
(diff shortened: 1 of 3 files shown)


MitchBomcanhao:
@jfversluis is the "after" screenshot representative of the final code? if so, what happened to the clear and search icons that were visible in the "before" screenshot?

jfversluis:
Good eye @MitchBomcanhao! I have opened #10328 to fix that

PR #10152, opened 2020-03-31
Fix 9939 [iOS][UWP] Dispose method not called when adding control inside the NavigationPage.TitleView

### LINKED ISSUE #9939: [Bug] Dispose method not called when adding control inside the NavigationPage.TitleView
### Description
I have a custom control that unmanaged resources(unsubscribe the events, null the reference) are cleared by the overring Dispose() method of each platform renderer. Control objects are cleared from memory when doing page navigation or removed it from parents (because unmanaged resources are cleared by calling the Dispose method on each platform renderer). But when adding control in NavigationPage.TitleView and do the page navigation it won't clear from memory, because of Dispose method doesn't fire, hence unmanaged resource holds the control reference. 

### Steps to Reproduce

1. Run the attached sample 
2. Do the page navigation back and forth
3. Application gets the crash because the dispose method is not called in order to close the stream.  
4. Add the same control as the content of ContentPage instead of TitleView, it will be working fine. 

[Note: just created sample for reproduce issue in UWP, this is not actually scenarios]

### Expected Behavior
Dispose method is should be called so that unmanaged control resources should be cleared. 
### Actual Behavior
Dispose method is not to be called. 

### Basic Information
- IDE: VS2019 Enterprise 
- Nuget Packages: XF 4.3.0.908675

### Reproduction Link

https://github.com/sheiksyedm/TitleViewDemo

yurkinh:
### Description of Change ###
Dispose() method is not called for a custom view added to  TitleView on iOS and UWP

### Issues Resolved ### 

- fixes #9939 

### Platforms Affected ### 

- iOS
- UWP

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/NavigationPageRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/NavigationPageRenderer.cs
@@ -453,7 +453,10 @@ namespace Xamarin.Forms.Platform.UWP
 			if (_currentPage != null)
 			{
 				if (isPopping)
+				{
 					_currentPage.Cleanup();
+					_container.TitleView?.Cleanup();
+				}
 
 				_container.Content = null;
 				_currentPage.PropertyChanged -= OnCurrentPagePropertyChanged;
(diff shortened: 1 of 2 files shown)


jsuarezruiz:
On Android we are disposing the TitleView here: https://github.com/xamarin/Xamarin.Forms/blob/17a0497eba0dd879aaa7eb1fecb61808bf3480fe/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs#L215

PR #10156, opened 2020-03-31
Fix route removal to keep one default route if it's needed

PureWeen:
### Description of Change ###
As part of this https://github.com/xamarin/Xamarin.Forms/pull/10064 I removed all the default routes so that hot reload would be able to reroute to where the user was working.

This introduced a bug where if you've pushed a global route then it'll strip off all routes that are part of shell if everything is default. 

If the user has navigated to a global route than we have to retain the last default route to give an accurate indication of where the user is 

### Platforms Affected ### 
- Core/XAML (all platforms)

### Testing Procedure ###
- unit tests included

### PR Checklist ###


PureWeen:
Failures not related

PR #10276, opened 2020-04-09
PopModalAsync hides navigation bar on macOS #9536

### LINKED ISSUE #9536: PopModalAsync hides navigation bar on macOS
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/content/problem/906251/popmodalasync-hides-navigation-bar-on-macos.html)._

---
When I push a page modally in my Forms app, I wrap the modal page in a NavigationPage so the modal page has a place for a title and toolbar buttons. This works quite well on iOS, UWP and Android. However on macOS, the navigation bar on the main page disappears after calling Navigation.PopModalAsync to remove the modal page.

To reproduce this behavior, create a two page app. On the main page add a button to initiate the modal push. Its action is: Navigation.PushModalAsync(new NavigationPage(new ModalPage())). This displays the modal page with its own navigation bar and toolbar buttons. However when issuing Navigation.PopModalAsync from the modal page, the navigation bar on the main page will disappear.

If the action instead is Navigation.PushModalAsync(new ModalPage(), true), the main page navigation bar no longer disappears after Navigation.PopModalAsync call.

Note also that the size of the window changes when the navigation bar is present or not. That is unexpected behavior.

See the attached project for an example demonstrating the inconsistency.

---
#### Visual Studio Feedback System on 2/4/2020, 07:05 PM: 

<p>We have directed your feedback to the appropriate engineering team for further evaluation. The team will review the feedback and notify you about the next steps.</p>

---
### Original Solutions
(no solutions)

hot33331:
### Description of Change ###
- added a page in gallery to allow for reproduction of bug
- main change to fix the bug is in Xamarin.Forms.Platform.iOS/DisposeHelpers.cs

This might be problematic due to the fact that part of the dispose code on mac is not run. Someone with a deeper understanding than me should thoroughly take a look. Probably there's a better solution than this that I just couldn't come up with.

### Issues Resolved ### 
- fixes #9536 

### API Changes ###
None

### Platforms Affected ### 
- macOS

None

Not applicable

### Testing Procedure ###
1. Both on iOS (not affected, just test that still everything works) and macOS go to test cases in the gallery app and search for 9536.
2. Select the entry for #9536 - "PopModalAsync...".
3. Tap the button to open modal.
4. Tap the button to close modal.
5. Navigation on top back to search should be there.

### PR Checklist ###


Xamarin.Forms.Platform.iOS/DisposeHelpers.cs
```diff
@@ -25,20 +25,25 @@ internal static void DisposeModalAndChildRenderers(this Element view)
...
+#if __MOBILE__
+			  )
+#else
+			    && !(view is NavigationPage))
+#endif
 			{
 				renderer = Platform.GetRenderer(visualElement);
 				if (renderer != null)
 				{
+#if __MOBILE__
```
hot33331, line 38: I figured this could already be placed here since the if statement does not have content if __MOBILE__ is false.

Xamarin.Forms.Platform.iOS/DisposeHelpers.cs
```diff
@@ -25,20 +25,25 @@ internal static void DisposeModalAndChildRenderers(this Element view)
 				}
 			}
 
-			if (view is VisualElement visualElement)
+			if (view is VisualElement visualElement
+#if __MOBILE__
+			  )
+#else
+			    && !(view is NavigationPage))
+#endif
```
hot33331, line 33: This is the main change and more symptomatic than actually a fix. I'd be keen to know if there's a better way to fix in and if so how.

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Github9536.xaml.cs
```diff
@@ -0,0 +1,23 @@
...
+    [Issue(IssueTracker.Github, 9536, "PopModalAsync hides navigation bar on macOS", PlatformAffected.macOS)]
+    public partial class Github9536:ContentPage
+    {
+        public Github9536()
+        {
+            InitializeComponent();
+        }
+        
+        void Button_Clicked(object sender, System.EventArgs e) {
+            Navigation.PushModalAsync(new NavigationPage(new Issue2964.ModalPage()));
```
hot33331, line 20: I used an existing modal page here. I hope this is okay. Not a fan of code repetition, but could understand if necessary.

hot33331, line 20: I used an existing modal page here. I hope this is okay. Not a fan of code repetition, but could understand if necessary.

rmarinho:
Found a better fix #10558

PR #10304, opened 2020-04-11
Fixed navigation events for custom WebView on Android.

### LINKED ISSUE #8954: [Bug] Navigated event fired at startup for custom WebView on Android
### Description
If you create a custom renderer for Android WebView, the **Navigated** event is called incorrectly at when the app is starting. This results in behavior that **Navigated** is called twice when launching the app.
- First call when is when the app is starting - INCORRECT
- The second call is when page is loaded - CORRECT

### Steps to Reproduce

1. Download and open this project:
[App2.zip](https://github.com/xamarin/Xamarin.Forms/files/3979670/App2.zip)

2. Put a breakpoint in this line in **MainPage.xaml.cs**

```
Debug.WriteLine("Navigated");
```

3. Launch the app on Android.
4. As you can see **Navigated** event is called twice. The first call is when the app is starting (during splash screen). **Navigated** event definitely should not be called at this moment.

### Expected Behavior
**Navigated** event should be called only when the WebView page is loaded.

### Actual Behavior
On Android **Navigated** event is called during app startup.

### Basic Information

- Version with issue: 4.4.0.991265
- IDE: Visual Studio 2019 16.4.1
- Platform Target Frameworks: 
  - Android: 9.0 (Pie)

Mikilll94:
### Description of Change ###
This PR fixes a bug on Android when using custom renderer for WebView. When you have a custom renderer for WebView,  the **Navigating** and **Navigated** events are called twice at the start. The worst thing with this bug is the fact that the first call of the **Navigated** event happens before even the WebView is visible on the page.

### Issues Resolved ### 
- fixes #8954

### API Changes ###
 None

### Platforms Affected ### 
Only Android

### Testing Procedure ###
1. Go to the gallery page with this issue.
2. Look at the "Application output". First, the "Navigating" event should be logged to the console. Then when the WebView page has loaded completely, the "Navigated" event should be logged.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.ControlGallery.Android/Xamarin.Forms.ControlGallery.Android.csproj
+++ b/Xamarin.Forms.ControlGallery.Android/Xamarin.Forms.ControlGallery.Android.csproj
@@ -136,6 +136,7 @@
     </Compile>
     <Compile Include="ApiLabelRenderer.cs" />
     <Compile Include="Issue7249SwitchRenderer.cs" />
+    <Compile Include="_8954CustomRenderer.cs" />
     <Compile Include="_9087CustomRenderer.cs" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


jfversluis:
No response to request for changes after 2 years. I think this won't be happening :D

PR #10355, opened 2020-04-17
[Enhancement] Switch: Added OffColor property

### LINKED ISSUE #10341: [Enhancement] OffColor to Switch
## Summary
Make it possible to set OffColor for a Switch

## API Changes
**Properties to Add:**

- OffColor:Color

**Fields to Add:**

- OffColorProperty:BindableProperty

## Intended Use Case
To set the color of Switch when the Switch not is toggled. The default color could be hard to see if the background color is similar.

AndreiMisiukevich:
### Description of Change ###
Added possibility to set "Off" color of Switch control.

**Note: iOS versions less than 13**
Off-color is a TintColor + Background colors, that's why the code looks like it looks.
If BackgroundColor is set for a switch on, we ignore OffColor for backward compatibility.
But since iOS 13 Off-color works independently from BackgroundColor.

Some info I found: https://medium.com/@myshkinasasha/ios-13-uiswitch-modal-presentation-8e91c9add7f1

### Issues Resolved ### 
- fixes #10341 

### API Changes ###
Added:
 - Color Switch.OffColor { get; set; } //Bindable Property

### Platforms Affected ### 
- Core
- iOS
- Android

Apr-17-2020 04-01-07

### Testing Procedure ###
The playground
```csharp
MainPage = new ContentPage
{
	Content = new StackLayout
	{
		Padding = new Thickness(100),
		Children =
		{
			new Label { Text = "Regular" },
			new Switch(),
			new Label { Text = "With On Color only" },
			new Switch { OnColor = Color.Gold },
			new Label { Text = "With Off Color only" },
			new Switch { OffColor = Color.Aqua },
			new Label { Text = "With On/Off" },
			new Switch { OnColor = Color.Gold, OffColor = Color.Aqua }
		}
	}
};
```

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/SwitchRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/SwitchRenderer.cs
@@ -1,6 +1,7 @@
 using System;
 using System.ComponentModel;
 using System.Drawing;
+using System.Linq;
 using UIKit;
 
 namespace Xamarin.Forms.Platform.iOS
@@ -39,12 +40,14 @@ namespace Xamarin.Forms.Platform.iOS
 				}
 
 				_defaultOnColor = UISwitch.Appearance.OnTintColor;
-				_defaultOffColor = UISwitch.Appearance.TintColor;
+				_defaultOffColor = Forms.IsiOS13OrNewer
+					? Control.Subviews?.FirstOrDefault().Subviews?.FirstOrDefault()?.BackgroundColor
+					: UISwitch.Appearance.TintColor;
 				_defaultThumbColor = UISwitch.Appearance.ThumbTintColor;
 				Control.On = Element.IsToggled;
 				e.NewElement.Toggled += OnElementToggled;
 				UpdateOnColor();
-				UpdateOffColor(Element.BackgroundColor);
+				UpdateOffColor();
 				UpdateThumbColor();
 			}
 
@@ -53,33 +56,38 @@ namespace Xamarin.Forms.Platform.iOS
 
 		void UpdateOnColor()
 		{
-			if (Element == null)
+			if (IsElementOrControlEmpty)
 				return;
+
 			Control.OnTintColor = Element.OnColor != Color.Default
 				? Element.OnColor.ToUIColor()
 				: _defaultOnColor;
 		}
 
-		void UpdateOffColor(Color backgroundColor)
+		void UpdateOffColor()
 		{
-			if (Element == null)
+			if (IsElementOrControlEmpty)
 				return;
-			if (Element.OffColor == Color.Default || backgroundColor != Color.Default)
+
+			if (!Forms.IsiOS13OrNewer)
 			{
-				Control.TintColor = _defaultOffColor;
-				Control.Layer.CornerRadius = 0;
-				base.SetBackgroundColor(backgroundColor);
+				HandleOffColorForOlderSystemVersions(Element.BackgroundColor);
 				return;
 			}
-			var color = Element.OffColor.ToUIColor();
-			Control.TintColor = color;
-			Control.BackgroundColor = color;
-			Control.Layer.CornerRadius = Control.Bounds.Height / 2;
+
+			var offColorView = Control.Subviews.FirstOrDefault()?.Subviews.FirstOrDefault();
+			if (offColorView == null)
+				return;
+			
+			offColorView.BackgroundColor = Element.OffColor != Color.Default
+				? Element.OffColor.ToUIColor()
+				: _defaultOffColor;
+			
 		}
 
 		void UpdateThumbColor()
 		{
-			if (Element == null)
+			if (IsElementOrControlEmpty)
 				return;
 
 			Color thumbColor = Element.ThumbColor;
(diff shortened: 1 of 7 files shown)


Xamarin.Forms.Platform.iOS/Renderers/SwitchRenderer.cs
```diff
@@ -78,13 +104,44 @@ void OnElementToggled(object sender, EventArgs e)
...
+		}
+
+		void HandleOffColorForOlderSystemVersions(Color backgroundColor)
+		{
+			if (Element.OffColor == Color.Default || backgroundColor != Color.Default)
+			{
+				Control.TintColor = _defaultOffColor;
+				Control.Layer.CornerRadius = 0;
+				base.SetBackgroundColor(backgroundColor);
+				return;
```
PureWeen, line 128: I don't think I quite follow here what you're saying about maintaining backwards compatibility. 

If you remove this return than the offcolor colors nicely on pre ios13 devices

PureWeen:
Also we're about to merge Brushes so 

https://github.com/xamarin/Xamarin.Forms/pull/9220

so maybe once that's been merged we just create a property that uses Brush

Xamarin.Forms.Platform.iOS/Renderers/SwitchRenderer.cs
AndreiMisiukevich: Yes, because ```HandleOffColorForOlderSystemVersions``` works only for pre iOS13 devices.
That's impossible to set "off" color and background color for pre IOS13 devices.

Just try to remove IF's body, set background color to switch, and run **WITH** OffColor set and **WITHOUT** OffColor set. You will see the difference (colored rectangle over the Switch).
I am not sure that anybody uses background color for Switch, but I decided to preserve initial behavior.

AndreiMisiukevich:
> What are your thoughts on just deprecating OnColor and Replacing with Color and then setting up VSM states for this instead of having On and Off Colors?
> 
> @StephaneDelcroix @AndreiMisiukevich ?
> 
> Also can you do the UWP implementation as well? if not let me know and I should be able to finish that pretty easily

1. I cannot realize how to achieve that, because OnColor uses certain API method (On iOS exactly), but OffColor uses totally different one. Single "Color" property will call different API methods depending on the state? 
2. It would be great if you could assist me, cuz I don't have a windows machine. Thanks a lot!

Xamarin.Forms.Platform.iOS/Renderers/SwitchRenderer.cs
PureWeen: Alright. what about this?

```C#
Control.Subviews[0].Subviews[0].Subviews[0].BackgroundColor = offColor;
```

it seems like on pre ios13 if you go down one more subview level you can color the off color?

image

AndreiMisiukevich: > Alright. what about this?
> 
> ```cs
> Control.Subviews[0].Subviews[0].Subviews[0].BackgroundColor = offColor;
> ```
> 
> it seems like on pre ios13 if you go down one more subview level you can color the off color?
> 
> image

Interesting.
Does it work on iOS 11/10 as well?
When I was working on this PR I didn't find any solution for older versions.

PureWeen:
1) I'm hoping we can make this work contextually 🤞 but it might be really tricky getting the timing correct. So when it's toggled or when the color changes we can set the appropriate part of the control.  This only really comes up on iOS right? Android looks like it's the same thing and AFAICT UWP is just the same rect


2) once we move to Brushes I'll add the UWP implementation

AndreiMisiukevich:
> 1. I'm hoping we can make this work contextually 🤞 but it might be really tricky getting the timing correct. So when it's toggled or when the color changes we can set the appropriate part of the control.  This only really comes up on iOS right? Android looks like it's the same thing and AFAICT UWP is just the same rect
> 2. once we move to Brushes I'll add the UWP implementation

Then I am not sure if it makes sense to keep this PR opened since your approach is fundamentally different 👍

Xamarin.Forms.Platform.iOS/Renderers/SwitchRenderer.cs
PureWeen: seems like it

I tested on iphone 7 with ios 10

AndreiMisiukevich: LoL So then those "tricky" code is my failure. Please don't hate me so much xD

PureWeen: I mean... I happened upon the solution like 2 pages deep on an SO post so it was more just luck than anything.

PureWeen:
I think the PR is still valid it's in a good state to continue from and see if we can condense this into a single brush

StephaneDelcroix:
my 0.02€ here. some switch implementation might want to show both colors at the same time (while switching). this is something we can't achieve with VSM.

I don't think VSM is the right hammer to hit this nail, and On/Off colors are probably a better semantic (to me)

StephaneDelcroix:
can you add

```csharp
[assembly: StyleProperty("-xf-off-color", typeof(Switch), nameof(Switch.OffColorProperty))]
```

to the AssemblyInfo.cs, so the off color become style-able in  CSS ?

AndreiMisiukevich:
> can you add
> 
> ```cs
> [assembly: StyleProperty("-xf-off-color", typeof(Switch), nameof(Switch.OffColorProperty))]
> ```
> 
> to the AssemblyInfo.cs, so the off color become style-able in CSS ?

Good point, thanks.
Updated.

PureWeen:
@AndreiMisiukevich or @jsuarezruiz can we update this PR to use brushes?

AndreiMisiukevich:
@jsuarezruiz can you please take care of this issue?

Codelisk:
Whats the status on this PR?

AndreiMisiukevich:
@PureWeen did https://github.com/xamarin/Xamarin.Forms/pull/12196 make the work of this PR?)

mhrastegary77:
This PR is really needed especially in switching app themes to Dark/light :( when this will be merged?

AndreiMisiukevich:
> This PR is really needed especially in switching app themes to Dark/light :( when this will be merged?

I think this won't be merged. Please wait for MAUI, hope this feature will be there

PR #10384, opened 2020-04-20
Fix previewer flakiness on Android

PureWeen:
### Description of Change ###

Don't use the global layout listener if you're in designer mode. The designer just goes nuts and disposes of everything so global listeners will sometimes cause disposal exceptions

### Platforms Affected ### 
- Android

### Testing Procedure ###
- tests pass
- pull down the nuget and test the previewer on vsmac and vs windows

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/VisualElementTracker.cs
+++ b/Xamarin.Forms.Platform.Android/VisualElementTracker.cs
@@ -37,7 +37,14 @@ namespace Xamarin.Forms.Platform.Android
 
 			renderer.View.SetCameraDistance(3600);
 
-			renderer.View.AddOnAttachStateChangeListener(AttachTracker.Instance);
+			if(_context.IsDesignerContext())
+			{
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.Android/VisualElementTracker.cs
```diff
@@ -60,7 +68,13 @@ protected virtual void Dispose(bool disposing)
 				if (_renderer != null)
 				{
 					_renderer.ElementChanged -= RendererOnElementChanged;
-					_renderer.View.RemoveOnAttachStateChangeListener(AttachTracker.Instance);
+
+					if (_renderer.View.IsAlive() && _attachTracker.IsAlive())
```
jfversluis, line 72: We don't need the `_context.IsDesignerContext()` check here? That is covered by `.IsAlive()`, right? Just checking

PureWeen: Right I was hoping to leave the designer check out of the dispose method :-) just have as few of them peppered around as possible

PR #10401, opened 2020-04-21
Indicator shows # of indicators = MaximumVisible even when there are fewer items

### LINKED ISSUE #9826: [Bug] Indicator shows # of indicators = MaximumVisible even when there are fewer items
# Description

Create a CarouselView with a bound IndicatorView. Add two items to CarouselView. Set MaximumVisible to 10. See 10 indicators instead of 2.

### Steps to Reproduce

1. Create a CarouselView with a bound IndicatorView.
2.  Add two items to CarouselView.
3. Set MaximumVisible to 10.

### Expected Behavior

Two indicators

### Actual Behavior

10 indicators

### Basic Information

- Version with issue: 4.6.0-pre1
- Last known good version:
- IDE: VS 16.4
- Platform Target Frameworks: 
  - iOS:  
  - Android: API 28
  - UWP:  
- Nuget Packages:
- Affected Devices:

### Reproduction Link

[PeakyBlunders.zip](https://github.com/xamarin/Xamarin.Forms/files/4283444/PeakyBlunders.zip)

### Workaround

n/a

jfversluis:
### Description of Change ###

Adds functionality to coerce the `MaximumVisible` and `ItemsSource.Count` value. If `ItemsSource.Count` is lower than `MaximumVisible` it will show `MaximumVisible` and vice versa.

In addition rearranged some pages in the gallery app to make it more logical and added this specific case.

### Issues Resolved ### 

- fixes #9826

### API Changes ###
 
None

### Platforms Affected ### 

- Core/XAML (all platforms)

Users already using this should see the right amount of maximum indicators after updating. This _can_ be less than they would see before, because before it would _always_ show the number set in `MaximumVisible` even thought there weren't even that many items.

Not applicable

### Testing Procedure ###
Go to the `IndicatorView` gallery and choose the "Indicator MaxVisible Gallery" option. There is a button to decrease the `MaximumVisible` value and a button to set the `MaximumValue` to `ItemsSource.Count`+1. There is also a button to add/remove items from the bound collection.

When ItemsSource.Count > MaximumVisible -> MaximumVisible should be shown
When ItemsSource.Count == MaximumVisible -> Doesn't matter, whichever one should be shown
When ItemsSource.Count < MaximumVisible -> ItemsSource.Count should be shown

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls/CoreGallery.cs
+++ b/Xamarin.Forms.Controls/CoreGallery.cs
@@ -300,7 +300,7 @@ namespace Xamarin.Forms.Controls
 				new GalleryPageFactory(() => new RadioButtonCoreGalleryPage(), "RadioButton Core Gallery"),
 				new GalleryPageFactory(() => new FontImageSourceGallery(), "Font ImageSource"),
 				new GalleryPageFactory(() => new ExpanderGalleries(), "Expander Gallery"),
-				new GalleryPageFactory(() => new IndicatorsSample(), "Indicator Gallery"),
+				new GalleryPageFactory(() => new IndicatorGalleries(), "IndicatorView Gallery"),				
 				new GalleryPageFactory(() => new CarouselViewGallery(), "CarouselView Gallery"),
 				new GalleryPageFactory(() => new CarouselViewCoreGalleryPage(), "CarouselView Core Gallery"),
 				new GalleryPageFactory(() => new CollectionViewGallery(), "CollectionView Gallery"),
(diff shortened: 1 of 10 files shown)


rmarinho:
RIght @PureWeen we can fix it on the renderers also.

jfversluis:
Added an `ActualMaximumVisible` to the control so the logic to determine that is shared and using that value in the renderers. Is that better?

@rmarinho @PureWeen ?

rmarinho:
Hum i wonder if it's confusing,. maybe we should hide it with from the api at least

jfversluis:
Moved it to the renderers. I just want to make you happy... :P

PR #10414, opened 2020-04-22
[Android,iOS] Fix setting position on Ctor and OnAppearing on CarouselView

### LINKED ISSUE #10170: [Bug] Issues setting the initial position on a CarouselView to a non-zero value
### Description
I have a CarouselView which I need to initially show one of the items and I want the CarouselView to go directly to this item and not show any other item or do any animation to the desired item.  Everything works fine when the first item is initially shown and this was working in v4.4.0.  With v4.5.0-sr2, I get the following results.

When setting the desired position in the page's constructor on Android, the first item will be shown and then scrolls with animation to the desired item even with IsScrollAnimated set to false.  On iOS, the first item will be shown and it will not go to the desired item even though the position is set correctly.

I also tried setting the position in the OnAppearing method although this is less desirable since I only want to do this the first time the screen is shown.  On Android, the correct item does show but changing the position before any swiping does not work (the item in the CarouselView does not change even though the Position property changes and the change event occurs).  After swiping to the next or previous item, changing the position works correctly.  On iOS, the first item shows briefly before switching to the correct item (no animation with IsScrollAnimated set to false). 

### Steps to Reproduce

1. Run the attached project: [CarouselTest.zip](https://github.com/xamarin/Xamarin.Forms/files/4415834/CarouselTest.zip)
2. The first button will go to a page with a CarouselView and sets the position to the 4th item using the constructor.  The second button will do the same except using the OnAppearing method.  
3. At the bottom is the current position of the CarouselView and buttons to increment and decrement the position.

### Expected Behavior
The CarouselView should be able to initially go directly to an item (with no animation) and changing the position should always work. 

### Actual Behavior
See the description.

### Basic Information

- Version with issue: 4.5.0.495
- Last known good version: 4.4.0.991265
- IDE: Visual Studio 15.9
- Platform Target Frameworks: 
  - iOS:  12.4
  - Android: 9.1.7.0

### LINKED ISSUE #10291: [Bug] CarouselView needs to check IsScrollAnimated/AnimateCurrentItemChanges/AnimatePositionChanges on UpdateItemSpacing also
https://github.com/xamarin/Xamarin.Forms/blob/8a458aebef77d6a7b2614cfdb1b3aaadd3396015/Xamarin.Forms.Platform.Android/CollectionView/CarouselViewRenderer.cs#L134

The CarouselView method `UpdateItemSpacing` on Android uses the `ScrollTo` method, but it uses in contrary to the other `ScrollTo` methods not the `AnimateCurrentItemChanges/AnimatePositionChanges` value:
**`animate: Carousel.AnimateCurrentItemChanges`**

I think it needs to use that value here in this methods also.

If you have `IsScrollAnimated `is False and set the initial position on the CarouselView, the CarouselView animates the scroll to this position. If I override the `UpdateItemSpacing `method in a CustomRenderer and do just a single return. The CarouselView inits on the correct position without any animation.

rmarinho:
### Description of Change ###

Fix issues when setting the Position on CarouselView on the ctor, on a different order or on appearing when the itemsource is set later.

Fix a small issue with `IndicatorView` when setting collection after the position would not update to the current position.

### Issues Resolved ### 

- fixes #10170 
- fixes #10291 

### API Changes ###

 None

### Platforms Affected ### 

- iOS
- Android

Should scroll to the item position on page load by setting on ctor or onappearing

Not applicable

### Testing Procedure ###

Go to gallery CarouselItemsGallery

pick some of the options 
"CarouselView SetPosition Appearing",
"CarouselView SetPosition Ctor" 
"CarouselView SetPosition Ctor No Animation"

Should always go to the Item4 (3rd index)

### PR Checklist ###


jfversluis:
Some tests are failing, although it seems more like because the gallery page is not in a scrollview that it can't reach certain cases anymore?

rmarinho:
thanks @jfversluis  fixed the tests

Xamarin.Forms.Platform.iOS/CollectionView/CarouselViewController.cs
```diff
@@ -200,18 +191,19 @@ void SetCurrentItem(int carouselPosition)
...
 
-		void ScrollToPosition(int goToPosition, int carouselPosition)
+		void ScrollToPosition(int goToPosition, int carouselPosition, bool animate, bool forceScroll = false)
 		{
-			if (_gotoPosition == -1 && goToPosition != carouselPosition)
+			if (_gotoPosition == -1 && (goToPosition != carouselPosition || forceScroll))
 			{
 				_gotoPosition = goToPosition;
-				Carousel.ScrollTo(goToPosition, position: Xamarin.Forms.ScrollToPosition.Center, animate: Carousel.AnimateCurrentItemChanges);
+				System.Diagnostics.Debug.WriteLine($"Scrolling to {_gotoPosition}");
```
jfversluis, line 205: Do we need to keep this one?

rmarinho: oh no good catch

Xamarin.Forms.Controls/GalleryPages/CollectionViewGalleries/CarouselViewGalleries/CarouselViewGallery.cs
```diff
@@ -34,19 +34,25 @@ public CarouselViewGallery()
 						GalleryBuilder.NavButton("CarouselView (XAML, Horizontal)", () =>
 							new CarouselXamlGallery(), Navigation),
 						GalleryBuilder.NavButton("CarouselView (Indicators Forms)", () =>
-							new CarouselItemsGallery(false,false,false), Navigation),
+							new CarouselItemsGallery(false,false,false,false,false,true), Navigation),
```
hartez, line 37: Can we make all these constructor parameters optional with a default value of `false`? It's pretty hard to tell what's being tested here.

rmarinho: fixed @hartez

Juansero29:
In which XF release is this PR being included?

jfversluis:
@Juansero29 it should be in the next service release of 4.6 that should be here soon

PR #10440, opened 2020-04-24
[UWP] Fixed issue disposing Pages

### LINKED ISSUE #10423: [Bug] [UWP] - !!SERIOUS MEMORYLEAK!!! PageRenderer.Dispose does not free resources
### Steps to Reproduce

1. Start Xamarin.Forms UWP app and navigate to ANY page
2. Navigate back and run GC.Collect() (Or use dotMemory for profiling)
3. You will see, that the PageRenderer and the Page are still in memory

### Expected Behavior

Resources should become garbage collected

### Actual Behavior

Does not free memory

### Basic Information

- Version with issue: 4.4.0
- Last known good version: (no idea)
- IDE: Visual Studio 2019
- Platform Target Frameworks: 
  - UWP: irrelevant  
- Nuget Packages: irrelevant
- Affected Devices: all UWP devices...

### Workaround

This is the code of `Xamarin.Forms.Platform.UWP.PageRenderer.Dispose()`

```csharp

		{
			if (!disposing || _disposed)
				return;

			_disposed = true;

			if (Element != null)
			{
				ReadOnlyCollection<Element> children = ((IElementController)Element).LogicalChildren;
				for (var i = 0; i < children.Count; i++)
				{
					var visualChild = children[i] as VisualElement;
					visualChild?.Cleanup();
				}
				Element?.SendDisappearing();
			}

			base.Dispose(); // the main culprit!
		}
```

because `base.Dispose()` is called, the code of `Xamarin.Forms.Platform.UWP.PageRenderer.VisualElementRenderer`
it - in turn - will invoke the overload that takes a bool:
```csharp
public void Dispose()
		{
                        /// !!! BECAUSE this is a protected virutal void Dispose(bool d) method
                        /// which is overwritten in PageRenderer,
                        /// the Dispose(bool b) method of VisualElementRenderer is NEVER executed!!!!!  
			Dispose(true); 
			GC.SuppressFinalize(this);
		}
```
What the flow should be:
1. `PageRenderer.Dispose(true)` is called
2. `VisualElementRenderer.Dispose(true)` is called

What actually happens
1. `PageRenderer.Dispose(true)` is called
2. `PageRenderer` forwards to `base.Dispose()` (`VisualElementRenderer.Dispose()`))
3. `VisualElementRenderer.Dispose()` calls `VisualElementRenderer.Dispose(true)`, but because that is overwritten by `PageRenderer`, instead, `PageRenderer.Dispose(true)` is called again.

`VisualElementRenderer.Dispose(true)` is NEVER CALLED!!!

This is the obvious fix:

```csharp

		{
			if (!disposing || _disposed)
				return;

			_disposed = true;

			if (Element != null)
			{
				ReadOnlyCollection<Element> children = ((IElementController)Element).LogicalChildren;
				for (var i = 0; i < children.Count; i++)
				{
					var visualChild = children[i] as VisualElement;
					visualChild?.Cleanup();
				}
				Element?.SendDisappearing();
                                // additional memoryleaks: Eventhandlers are not unsubscribed from!!
				Unloaded -= OnUnloaded;
				Loaded -= OnLoaded;
			}

                        // Forward to the correct method! 🚀
			base.Dispose(disposing);
		}

´´´

You should really really fix that. This is a game stopper unfortunately

### PR DESCRIPTION
**Pull request by jsuarezruiz**
### Description of Change ###

With this change we call `VisualElementRenderer.Dispose(true)` disposing Pages.

### Issues Resolved ### 

- fixes #10423 

### API Changes ###
 
 None

### Platforms Affected ### 

- UWP

None

Not applicable

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/PageRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/PageRenderer.cs
@@ -34,7 +34,7 @@ namespace Xamarin.Forms.Platform.UWP
 				Element?.SendDisappearing();
 			}
 
-			base.Dispose();
+			base.Dispose(true);
 		}
 
 		protected override void OnElementChanged(ElementChangedEventArgs<Page> e)
--- a/Xamarin.Forms.Platform.UAP/PageRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/PageRenderer.cs
@@ -18,23 +18,28 @@ namespace Xamarin.Forms.Platform.UWP
 
 		protected override void Dispose(bool disposing)
 		{
-			if (!disposing || _disposed)
+			if (_disposed)
+			{
 				return;
+			}
 
 			_disposed = true;
 
-			if (Element != null)
+			if (disposing)
 			{
-				ReadOnlyCollection<Element> children = ((IElementController)Element).LogicalChildren;
-				for (var i = 0; i < children.Count; i++)
+				if (Element != null)
 				{
-					var visualChild = children[i] as VisualElement;
-					visualChild?.Cleanup();
+					ReadOnlyCollection<Element> children = ((IElementController)Element).LogicalChildren;
+					for (var i = 0; i < children.Count; i++)
+					{
+						var visualChild = children[i] as VisualElement;
+						visualChild?.Cleanup();
+					}
+					Element?.SendDisappearing();
 				}
-				Element?.SendDisappearing();
 			}
 
-			base.Dispose(true);
+			base.Dispose(disposing);
 		}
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Xamarin.Forms.Platform.UAP/PageRenderer.cs [hunk 1]
```diff
@@ -34,7 +34,7 @@ protected override void Dispose(bool disposing)
 				Element?.SendDisappearing();
 			}
 
-			base.Dispose();
+			base.Dispose(true);
```
hartez: This should be `base.Dispose(disposing);`

PR #9911, opened 2020-03-09
Handle "groups but no actual cells" cases for reloading UICollectionView

hartez:
### Description of Change ###

On iOS, a CollectionView with all empty groups is crashing when adding the first item; UICollectionView internal accounting gets confused.

These changes check for that circumstance and tell the UICollectionView to do a data reload instead, which allows it to get the internal accounting correct (and prevent the crash).

This also adds a check to prevent UICollectionView full reloads while the UICollectionView is in the process of animating other changes (which results in a garbled cell arrangement).

### Issues Resolved ### 

- Fixes #9580
- Fixes #9686

### API Changes ###
 
None

### Platforms Affected ### 

- iOS

### Behavioral/Visual Changes ###

None

### Before/After Screenshots ### 

Not applicable

### Testing Procedure ###

Automated test (Issue 9580)
Automated test (Issue 9686)

### PR Checklist ###

- [ ] Targets the correct branch
- [ ] Tests are passing (or failures are unrelated)

PR #10450, opened 2020-04-24
[iOS, Android] Fix when we call ReloadData on CollectionView, Fix issue adding item to start of Collection on ANdroid

### LINKED ISSUE #10300: [Bug] [iOS] ObservableCollection.RemoveAt(index) with a valid index raises ArgementOutOfRangeException.
In latest XF version **(4.5.0.617)** using `ObservableCollection.RemoveAt(index)` with a valid index raises `ArgementOutOfRangeException`.
It is an **iOS** exception only. On **Android** that exact same code it is working fine.

The ObservableCollection is bound to the `ItemsSource `of a `CarouselView `and the `ObservableCollection.RemoveAt(index)` is executed in an event raised bij the `Messenger-Service` (not on the main thread). If the code is changed to:
```
var list = ObservableCollection.ToList();
list.RemoveAt(index)
ObservableCollection = new ObservableCollection(list)
```
The codes executes fine on iOS and without raising any Exception.
The index is valid. For example 5 in a collection count of 20.

rmarinho:
### Description of Change ###

I found that #9911 introduced a new concept of Reload the CollectionView if it wasn't visible and an update to the items was in progress.
Seems this approach doesn't work well when we push a modal on top of the CollectionView  and then on dispose we add/remove items from the CollectionView. This was actually preventing the item from being removed. So my first try was to remove the check for `_collectionViewController.View.Window == null` , this worked but was going around the real problem.

It seems that also the `_batchUpdating.WaitAsync()` on the Reload method isn't working as expected, it was getting stuck because there were no available threads on the `_batchUpdating` semaphore. The problem was after a BatchUpdates was executed (for example add a item) the semaphore wasn't releasing and making the threads available. So the next time for example if it wanted to Reload it was getting stuck.

Fix issue on Android that Scrolling to the current position when adding an item would scroll the last item. 

This is actual a regression from previous versions. 

### Issues Resolved ### 

- fixes #10300

### API Changes ###
 None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###

Added UITest 10300
Click Add
Click Delete
Click Close,
You should see the item `2`

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs
+++ b/Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs
@@ -284,7 +284,7 @@ namespace Xamarin.Forms.Platform.iOS
 		{
 			// If the UICollectionView hasn't actually been loaded, then calling InsertItems or DeleteItems is 
 			// going to crash or get in an unusable state; instead, ReloadData should be used
-			return !_collectionViewController.IsViewLoaded || _collectionViewController.View.Window == null;
+			return !_collectionViewController.IsViewLoaded;
 		}
 
 		bool ReloadRequired()
--- a/Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs
+++ b/Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs
@@ -284,7 +284,7 @@ namespace Xamarin.Forms.Platform.iOS
 		{
 			// If the UICollectionView hasn't actually been loaded, then calling InsertItems or DeleteItems is 
 			// going to crash or get in an unusable state; instead, ReloadData should be used
-			return !_collectionViewController.IsViewLoaded;
+			return !_collectionViewController.IsViewLoaded || _collectionViewController.View.Window == null;
 		}
 
 		bool ReloadRequired()
@@ -320,10 +320,7 @@ namespace Xamarin.Forms.Platform.iOS
 			},
 					(_) =>
 					{
-						if (_batchUpdating.CurrentCount > 0)
-						{
-							_batchUpdating.Release();
-						}
+						_batchUpdating.Release();
 					});
 		}
 	}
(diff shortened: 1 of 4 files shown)


rmarinho:
Not sure if we should still would remove this check  https://github.com/xamarin/Xamarin.Forms/blob/4.6.0/Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs#L287

`_collectionViewController.View.Window == null`

In the issue i created Window is null because we were showing a Modal and going back to the page with collectionview

Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs
```diff
@@ -320,10 +320,7 @@ void BatchUpdate(Action update)
 			},
 					(_) =>
 					{
-						if (_batchUpdating.CurrentCount > 0)
-						{
-							_batchUpdating.Release();
-						}
+						_batchUpdating.Release();
```
PureWeen, line 323: @hartez `CurrentCount` is 

> The number of remaining threads that can enter the semaphore.

so checking here if CurrentCount > 0 probably isn't correct.

Also `_batchUpdating.Wait();` is going to block the thread so I don't feel like that is right either

Would this work?
```C#
async void BatchUpdate(Action update)
		{
			await _batchUpdating.WaitAsync();
			_collectionView.PerformBatchUpdates(() =>
			{
				update();
			},
			(_) =>
			{
				_batchUpdating.Release();
			});
		}
```

hartez: Yeah, what's in here right now is wrong. I fixed this for ObservableGroupedSource, but forgot to apply the same fix here: https://github.com/xamarin/Xamarin.Forms/pull/9911/commits/220715c11b9543b6b2b7d98ac9acefddc54e1db8

hartez: > Also _batchUpdating.Wait(); is going to block the thread so I don't feel like that is right either

It won't block the thread because the semaphore is available. Since `CurrentCount` is greater than 0, `Wait()` decrements the count and returns immediately. 

```
{
	_batchUpdating.Release();
});
```

We need the check for `CurrentCount == 0` here. If multiple calls to `PerformBatchUpdates` are batched together, blindly calling `Release()` could result in an `SemaphoreFullException` (https://docs.microsoft.com/en-us/dotnet/api/system.threading.semaphoreslim.release?view=netcore-3.1#System_Threading_SemaphoreSlim_Release)

PR #10456, opened 2020-04-26
Use FlyoutBehavior on Shell if user sets it

PureWeen:
### Description of Change ###

When using a TabBar or FlyoutItem the implicit behavior of the Flyout is different on one vs the other. This was implemented by just setting the attached property on the Item. The problem with doing it that way is that it doesn't let the user set the FlyoutItem at the Shell level

### Platforms Affected ### 

- Core/XAML (all platforms)

### Testing Procedure ###
- unit tests included
- Play with setting in Control Gallery

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Shell/Shell.cs
+++ b/Xamarin.Forms.Core/Shell/Shell.cs
@@ -811,6 +811,9 @@ namespace Xamarin.Forms
 			set => SetValue(CurrentItemProperty, value);
 		}
 
+		internal ShellContent CurrentContent => CurrentItem?.CurrentItem?.CurrentItem;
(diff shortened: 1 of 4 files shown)


Xamarin.Forms.Core/Shell/Shell.cs
```diff
@@ -811,6 +811,9 @@ public ShellItem CurrentItem
 			set => SetValue(CurrentItemProperty, value);
 		}
 
+		internal ShellContent CurrentContent => CurrentItem?.CurrentItem?.CurrentItem;
```
samhouts, line 814: 🐢🐢🐢

PR #10460, opened 2020-04-26
Removed useless Expander.Spacing property (Bug fix)

### LINKED ISSUE #10449: [Bug] Expander - Animation cut off when Spacing is used
### Description
Collapse and Expand animation are being cut off when "Spacing" property has a value.

I tested it in CollectionView, ListView, BindableLayout and a simple grid and the animation is skipped in all of them. All of them are also in the repro sample for easier testing.

It seems that the Spacing just moves down the content by the value in the Spacing property and just leaves an empty space between the Content and Header. I do not see any use case for this but maybe I am missing something.

Spacing

### Steps to Reproduce

1. Run the repro sample
2. Expand and Collapse any item

### Expected Behavior
Animation should not be skipped when spacing is set.

### Actual Behavior
Expand and Collapase animation skip the spacing.

### Basic Information

- Version with issue: 4.7.0.688-nightly
- Last known good version:
- IDE: VS 16.5.4
- Platform Target Frameworks: 
  - Android: Android 10 emulator

### Screenshots

Example from CollectionView

ExpanderSpacingAnimationCutOff

### Reproduction Link

[ExpanderTest.zip](https://github.com/xamarin/Xamarin.Forms/files/4531370/ExpanderTest.zip)

AndreiMisiukevich:
### Description of Change ###

Please firstly look through #10449 

Collapse and Expand animation are being cut off when "Spacing" property has a value.t seems that the Spacing just moves down the content by the value in the Spacing property and just leaves an empty space between the Content and Header. I do not see any use case for this. (Spacing can easily be managed by Margin/Padding and it won't cause UI glitches).

I didn't mark it as Obsolete because stable 4.6.0 is not released yet.

### Issues Resolved ### 
- fixes #10449 

### API Changes ###
Removed:
 - double Expander.Spacing
 
### Platforms Affected ### 
- Core

N/A

Not applicable

### Testing Procedure ###
The sample is attached to the bug ticket.

### DIFF
--- a/Xamarin.Forms.Core/Expander.cs
+++ b/Xamarin.Forms.Core/Expander.cs
@@ -13,9 +13,6 @@ namespace Xamarin.Forms
 
 		public event EventHandler Tapped;
 
-		public static readonly BindableProperty SpacingProperty = BindableProperty.Create(nameof(Spacing), typeof(double), typeof(Expander), 0d, propertyChanged: (bindable, oldvalue, newvalue)
-			=> ((Expander)bindable).ExpanderLayout.Spacing = (double)newvalue);
-
 		public static readonly BindableProperty HeaderProperty = BindableProperty.Create(nameof(Header), typeof(View), typeof(Expander), default(View), propertyChanged: (bindable, oldValue, newValue)
 			=> ((Expander)bindable).SetHeader((View)oldValue));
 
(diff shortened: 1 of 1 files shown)


AndreiMisiukevich:
@samhouts

PR #7518, opened 2019-09-14
iOS: Frame with HasShadow set to true and BackgroundColor alpha < 1 casts shadow on all child views

PieEatingNinjas:
### Description of Change ###
On iOS a CALayer casts shadows on all child elements when the background color of the layer has an alpha < 1. This is not what you'd expect when setting HasShadow to true on a Frame.
To fix this I'm removing the shadow properties on the layer itself and I'm adding an additional View (ShadowView) to the SuperView underneath the View itself. On this ShadowView I'm setting the Shadow properties. By doing so the shadows are not casted on the child elements of the 'original' View.

### Issues Resolved ### 

- fixes #5108 

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS

### Behavioral/Visual Changes ###
When using a Frame with HasShadow set to true and the background has an Alpha value of '1', there shouldn't be any visual differences.
When using a Frame with HasShadow set to true and the background has an Alpha less then '1', instead of seeing shadows on all child views, only a shadow around the border of the view should be visible.
There should be no visual differences when the HasShadow property is set to false.

### Before/After Screenshots ### 
Before (with background alpha < 1):
PR5108-trans-before

After(with background alpha < 1):
PR5108-trans-after

Before (with background alpha == 1):
PR5108-hotpink-before

After (with background alpha == 1):
PR5108-hotpink-after

### Testing Procedure ###
Wasn't able to write any UI tests for this... However, the Page I created in the Controls.Issues projects has some functionality that I used the check the behavior. True-hero-style, here's a video showing it to you:

### PR Checklist ###

- [ ] Targets the correct branch
- [ ] Tests are passing (or failures are unrelated)

Issue #5108, opened 2019-02-01
Frame with a shadow on iOS adds a shadow to the contents


stevehurcombe:
### Description
Frames on iOS have drop shadows on the contents of the frame.

Given a simple frame like this:

```
<Frame class="ColouredFrame" CornerRadius="10" >
  <StackLayout Spacing="0">
    <StackLayout Orientation="Horizontal" Margin="0, 5, 0, 20">
      <Label class="RadioText" HorizontalOptions="StartAndExpand" VerticalTextAlignment="Start" FontAttributes="Bold">Please review and confirm that you agree to our terms and conditions before continuing.</Label>
      <Button Text="View" Clicked="Button_Clicked" class="Primary" HorizontalOptions="End" VerticalOptions="Start" BorderRadius="8" WidthRequest="80"></Button>
    </StackLayout>
    <StackLayout Orientation="Horizontal" Margin="0, 5, 0, 5">
      <Label class="RadioText" FontAttributes="Bold" VerticalTextAlignment="Center">I agree</Label>
      <Switch x:Name="rbTerms" HorizontalOptions="StartAndExpand" />
      <Button x:Name="btnContinue" StyleClass="Primary" Text="Continue" HorizontalOptions="Center" Clicked="ContinueButton_Clicked" BorderRadius="8" IsEnabled="False" WidthRequest="150"></Button>
    </StackLayout>
  </StackLayout>
</Frame>
```
and some CSS:

```
.ColouredFrame {
    border-color: #57BABE;
    background-color: #2257BABE;
    padding: 10 10 10 10;
}

    .ColouredFrame label {
        color: #57BABE;
    }

```
On iOS the shadow applies to everything within the Frame on Android, it only applies to the frame.

### Expected Behavior
I would not expect the shadow to apply to the contents of the frame. At the very least iOS and Android should be consistent.

- Version with issue: Forms 4.0.0.94569-pre3
- Last known good version: N/A

### Screenshots
Android:
![screenshot_20190201-android](https://user-images.githubusercontent.com/2033186/52125082-2c45e600-2623-11e9-8be1-7e18410c2d0e.jpg)

iOS:
![screenshot_20190201-ios](https://user-images.githubusercontent.com/2033186/52125083-2c45e600-2623-11e9-9063-c3fefae56811.jpg)

### Workaround
Adding HasShadow='False' to the frame solves the issue.

kingces95:
[...]

PR #10469, opened 2020-04-27
[iOS] Fix Frame issues

### LINKED ISSUE #9555: [Bug][iOS] Applying an Effect to Frame adds a shadow
### Description

When applying an Effect to a `<Frame>` a shadow is applied on iOS.

### Steps to Reproduce

1. Add an effect to a Frame

```
<Frame.Effects>
     <controls:ElevationEffect/>
</Frame.Effects>
```

2. Launch on iOS
3. See the shadow applied (not expected)

### Expected Behavior

The effect is applied and no extra shadow is added

### Actual Behavior

The effect is applied and a shadow is added like if HasShadow = True

### Basic Information

- Version with issue:
- Last known good version: (Forms) 4.4.0.991640
- IDE: Visual Studio Mac 8.4.5 (19)
- Platform Target Frameworks: 4.5.0.142-pre1 +
  - iOS:  4.5.0.142-pre1 
  - Android: Doesn't not apply
  - UWP:  Not tested
- Nuget Packages:
- Affected Devices: iPhone 7 (iOS 13) (real device) also apply on simulator

### Screenshots

**Ok (4.4.0.991640)**

My custom effect is applied (a light shadow)

OK

**Ko : (4.5.0.142-pre1 +)**

My custom effect is applied and a shadow is added

ko

### Workaround

Apply your effect as expected then force HasShadow value per platform

```
<Frame.HasShadow>
      <OnPlatform x:TypeArguments="x:Boolean">
             <On Platform="Android" Value="True" />
             <On Platform="iOS" Value="False" />
      </OnPlatform>
</Frame.HasShadow>
```

### LINKED ISSUE #9774: [Bug] Xamarin.Forms 4.5 breaks custom frame renderer shadow functionality on iOS
### Description

When making a custom frame renderer for iOS and changing the shadow properties, they do not get applied on Xamarin 4.5. Moreover, the custom renderer does not extend, but rather overwrite the Xamarin Forms implementation. 

### Steps to Reproduce

1. Make a new Xamarin Forms project
2. Create a custom frame renderer
3. Modify the shadow of the custom frame renderer

### Expected Behavior
The custom frame renderer extends the properties of the current frame, modifying only the shadow.
### Actual Behavior
The custom frame renderer does not affect the shadow, sets the default shadow.
### Basic Information

- Version with issue: 4.5.0.356
- Last known good version: 4.4.0.991757
- IDE: Visual Studio for Mac
- Nuget Packages: Xamarin.Forms, Xamarin.Essentials
- Affected Devices: iOS

### Screenshots

How it should look:
Screenshot 2020-02-28 at 16 24 45

How it looks in Xamarin.Forms 4.5
Screenshot 2020-02-28 at 16 37 52

### Reproduction Link

https://github.com/doriangh/Xamarin-Forms-4.5-issue

### LINKED ISSUE #9990: Frame's shadow doesn't translate with the Frame by TranslatTo method (iOS)
### Description
When translating `Frame` with `TranslateTo` method, its shadow doesn't translate with it.
### Steps to Reproduce

1. Set `HasShadow` of a `Frame` to `true`
2. Move the `Frame` away from its position using `TranslateTo` method.

```
<StackLayout>
    <Button Text="Move" Clicked="BtnMove" HorizontalOptions="Center"/>
    <StackLayout WidthRequest="500">
        <Frame HorizontalOptions="Center" 
               CornerRadius="30"
               BackgroundColor="Red" HasShadow="True" x:Name="frm"
               WidthRequest="15"
               HeightRequest="15"/>
    </StackLayout>
</StackLayout>
```

```
int x = 60;
private void BtnMove(object sender, EventArgs e)
{
    frm.TranslateTo(x, 0);
    x = -x;
}
```

### Expected Behavior
The whole frame is translated to the new position (the Frame and its shadow)
### Actual Behavior
The Frame moves to the new position, but the shadow is still in the same original position.

### Basic Information

- Version with issue: multiple versions (didn't notice a version without the issue), but this tested against last XF version 4.5.0.396
- Last known good version:
- IDE: VS for Windows (16.5.0)
## Reproduction link

https://drive.google.com/file/d/133UWKhmZRHSSqej__JTg7jq8ljHhmZvr/view?usp=sharing

### Workaround
Set `HasShadow` to `false`

### LINKED ISSUE #10024: Frame animation issue/inconsistency in latest version of Forms/Xamarin iOS
I'm having an issue where the latest version (not sure how back it was introduced but was working on Xamarin.iOS 12.14) has introduced an issue with animating a frame to screen when using a combination of TranslateTo and RotateYTo.

I've narrowed the issue down to when HasShadow="True".  If the frame is not showing a shadow, the animation works correctly.  If the frame is showing a shadow, the animation goes all skewif with the part of the view that looks to be the shadow.

You can see the behaviour specifically in this github repo and there are two videos showing the behaviour so at least you don't need to pull down the repo to see ...

https://github.com/brad302/Xamarin/tree/master/FrameIssue

Keen to get the Xamarin teams views on this.  It's not a showstopper but it is annoying, the shadow formed part of the effect, especially given it's meant to look like it's floating on top of the window and this was working previously.

### LINKED ISSUE #10166: [Bug] [iOS] IsClippedToBounds Property is ignored by Frames
### Description
IsClippedToBounds Property is ignored by Frames. 

### Steps to Reproduce

1. Use Xamarin.Forms 4.5.0.495
2. Create ContentPage and use sample code

### Expected Behavior
If CornerRadius is > 0, no automatic clipping should take place.
If IsClippedToBounds is true, clipping should be done

### Actual Behavior
If CornerRadius>0, the frame is always be clipped
If CornerRadius=0, the frame will never be clipped, even if you set IsClippedToBounds to true.

### Basic Information

- Version with issue: 4.5.0.495
- Last known good version: 4.4.0.991640
- IDE: Rider
- Platform Target Frameworks:
  - iOS:  10
  - Android: working on android
- Affected Devices: iPhone 11 pro

### Sample Code
```
        <Grid>
                <Grid.ColumnDefinitions>
                    <ColumnDefinition Width="200" />
                    <ColumnDefinition Width="*" />
                </Grid.ColumnDefinitions>
                
                <Label Grid.Column="0" Grid.Row="0" Text="Frame, NoClipping, NoCornerRadius"/>
                <Label Grid.Column="0" Grid.Row="1" Text="Frame, Clipping, NoCornerRadius"/>
                <Label Grid.Column="0" Grid.Row="2" Text="Frame, NoClipping, CornerRadius"/>
                
                <Label Grid.Column="0" Grid.Row="3" Text="ContentView, NoClipping"/>
                <Label Grid.Column="0" Grid.Row="4" Text="ContentView, Clipping"/>

                <Frame Grid.Column="1" Grid.Row="0" IsClippedToBounds="False" HeightRequest="20" BackgroundColor="Red"
                       HasShadow="False" Padding="0"
                       CornerRadius="0" HorizontalOptions="Start" WidthRequest="100">
                    <Label Text="Wwwwwwwwwwwwwwwwwwww" TranslationY="10" TranslationX="50" LineBreakMode="WordWrap"
                           MaxLines="1" />
                </Frame>
                <Frame Grid.Column="1" Grid.Row="1" IsClippedToBounds="True" HeightRequest="20" BackgroundColor="Yellow"
                       HasShadow="False" Padding="0"
                       CornerRadius="0" HorizontalOptions="Start" WidthRequest="100">
                    <Label Text="Wwwwwwwwwwwwwwwwwwww" TranslationY="10" TranslationX="50" LineBreakMode="WordWrap"
                           MaxLines="1" />
                </Frame>
                <Frame Grid.Column="1" Grid.Row="2" IsClippedToBounds="False" HeightRequest="20" BackgroundColor="Aqua"
                       HasShadow="False" Padding="0"
                       CornerRadius="1" HorizontalOptions="Start" WidthRequest="100">
                    <Label Text="Wwwwwwwwwwwwwwwwwwww" TranslationY="10" TranslationX="50" LineBreakMode="WordWrap"
                           MaxLines="1" />
                </Frame>
                <ContentView Grid.Column="1" Grid.Row="3" IsClippedToBounds="False" HeightRequest="20"
                             BackgroundColor="Red" HorizontalOptions="Start"
                             WidthRequest="100">
                    <Label Text="Wwwwwwwwwwwwwwwwwwww" TranslationY="10" TranslationX="50" LineBreakMode="WordWrap"
                           MaxLines="1" />
                </ContentView>
                <ContentView Grid.Column="1" Grid.Row="4" IsClippedToBounds="True" HeightRequest="20"
                             BackgroundColor="Yellow" HorizontalOptions="Start"
                             WidthRequest="100">
                    <Label Text="Wwwwwwwwwwwwwwwwwwww" TranslationY="10" TranslationX="50" LineBreakMode="WordWrap"
                           MaxLines="1" />
                </ContentView>
            </Grid>
```

### Screenshots

image

### LINKED ISSUE #10326: An exception occurs from FrameRenderer when create a native renderer for forms view in Xamarin.Forms.iOS
### Description

An exception occurs from FrameRenderer when creating a native renderer for forms view in Xamarin.Forms.iOS. it is working fine 4.4.0.991757. if I change the frame to Stacklayout it will work. 

In my sample, I have two buttons . In the first button click, I have passed the Frame to create the native renderer. In a second button click, I have passed the StackLayout to create the native renderer. 

First but click thrown an exception. but the second one not.

### Steps to Reproduce

1.  Run the attached sample
2. Click the Click me button
3.  See the exception occurs

### Expected Behavior
Need to create the native view renderer

### Actual Behavior
Exception occurs.

### Basic Information

- Version with issue:
- Last known good version:
- IDE:
- Platform Target Frameworks: 

  - iOS:  13.4
- Nuget Packages:  4.5.0.617

### Reproduction Link
[Test (2).zip](https://github.com/xamarin/Xamarin.Forms/files/4475136/Test.2.zip)

### LINKED ISSUE #10333: [Bug] [iOS] Opacity on Frame behavior change in 4.5
### Description

In Xamarin.Forms 4.4, with some controls in an AbsoluteLayout, and a Frame in the AbsoluteLayout set to cover all the controls (i.e. not include the controls in the Frame's content, but rather just position the Frame to cover the controls), if the background color is set to black and the opacity set to, say, 0.25, then the covered controls, as well as the background color of the AbsoluteLayout will dim. 

In Xamarin.Forms >= 4.5, doing the same causes a Label to not show through at all, and also the background color of the absolute layout does not show through, as if there was no Opacity set, or Opacity = 1... but only where the Frame covers a Label or directly covers the AbsoluteLayout. All of the other controls in the layout get dimmed as expected.

### Notes

It seems to me that using a Frame to do this dimming might not be the best approach, as a Frame is intended to contain controls, not cover them, so that a shadow and/or border can be added to a View or Layout. A BoxView would seem to be a better fit to cover and dim other views, and a BoxView works as expected. 

So this may not be a bug, but reporting it as such as the behavior is different between XF 4.4 and 4.5 and also between iOS and Android (did not test UWP).

### Steps to Reproduce

1.  Open the attached test project and deploy to an iOS simulator (tested on iPhone 11 iOS 13.4)

### Expected Behavior
The first visible control should be a Label with text "Can you see this Label? Am I dimmed?", and the background color, green, of the AbsoluteLayout should show through a bit. 

### Actual Behavior
The first visible control is a Button with text "Can you see this button? Am I dimmed?", and the green background color of the AbsoluteLayout does not show through at all. The Label and the background color of the AbsoluteLayout are covered by solid black with no transparency. 

### Basic Information

- Version with issue: tested 4.5.0.617, 4.6.0.616-pre4, and 4.7.0.620-nightly
- Last known good version: 4.4.0.991864
- IDE: Visual Studio 2019 16.5.3
- Platform Target Frameworks: 
  - iOS:  13.4
  
- Nuget Packages: Xamarin.Forms, Xamarin.Essentials

### Screenshots
Expected:
FrameOpacityExpected

Actual:
FrameOpacityActual

### Reproduction Link

[TestOpacityXF.zip](https://github.com/xamarin/Xamarin.Forms/files/4477619/TestOpacityXF.zip)

### Workaround

Use a BoxView rather than a Frame.

### LINKED ISSUE #10348: [Bug] Cannot change shadow appearance on Frame from version 4.5.x
### Description
In my custom render I have this settings for change the appearance of the shadow in iOS
```
protected override void OnElementChanged(ElementChangedEventArgs<Frame> e)
{
    base.OnElementChanged(e);
    Layer.ShadowRadius = 10f;
    Layer.ShadowColor = UIColor.Black.CGColor;
    Layer.ShadowOffset = new CGSize(0,10);
    Layer.ShadowOpacity = 0.16f;
}
```

but from Xamarin Forms 4.5.x not works anymore.

What I see is that in the [FrameRenderer](https://github.com/xamarin/Xamarin.Forms/blob/4.5.0/Xamarin.Forms.Platform.iOS/Renderers/FrameRenderer.cs) was added a new ShadowView that in not accessible from the custom renderer.

### Workaround

In The [FrameRender](https://github.com/xamarin/Xamarin.Forms/blob/4.5.0/Xamarin.Forms.Platform.iOS/Renderers/FrameRenderer.cs) add a public function like this for access the ShadowView::
```
public ShadowView InnerShadowView
{
   get => _shadowView;
}
```
So in the custom renderer I can do:
```
protected override void OnElementChanged(ElementChangedEventArgs<Frame> e)
{
    base.OnElementChanged(e);
    InnerShadowView.Layer.ShadowRadius = 10f;
    InnerShadowView.Layer.ShadowColor = UIColor.Black.CGColor;
    InnerShadowView.Layer.ShadowOffset = new CGSize(0,10);
    InnerShadowView.Layer.ShadowOpacity = 0.16f;
}
```

### LINKED ISSUE #10422: [Bug] [iOS] Frame render issue using TranslationY
### Description
I have a Sliding Panel control which is a Frame subclass and it stopped working after update Xamarin.Forms version.

Frame renders duplicate when use of TranslationY. One frame is displayed in the original TranslationY while the other is translated.
Issue occurs only on iOS.

### Steps to Reproduce

1. Create a blank XF App project.
2. Create a Frame subclass with a PanGestureRecognizer that updates TranslationY.
3. Update Xamarin.Forms to 4.5.0.356 or above.

### Expected Behavior
Frame moves up and down according to Pan Gesture.

### Actual Behavior
Frame moves up and down according to Pan Gesture but a second Frame remains at original TranslationY position.

### Basic Information

- Version with issue: 4.5.0.356 and above
- Last known good version: 4.4.0.991864
- IDE: Visual Studio 2019 16.5.4
- Platform Target Frameworks:
  - iOS:  13.16.0.13
- Affected Devices: Tested on simulators and iPhone 11 Pro.

### Screenshots

1 is the Frame moved using pan gesture. 2 is the duplicated Frame remaining in original position.
Simulator Screen Shot - iPhone 11 Pro - 2020-04-23 at 13 09 37

### Reproduction Link

https://github.com/HeikkiDev/XF4.5BugFrameTranslation

### Workaround

None found

### LINKED ISSUE #10906: [Bug] Control placed inside frame when animated to edge of frame getting disappeared
### Description
I am trying to built a [Material Entry Control](https://material.io/components/text-fields). In order to make an entry control with outline border feature I have used a frame. When user tap entry the placeholder which is mimicked via a label control is animated (translated) to the top of frame's outline edge. This was working fine till Xamarin.Forms version 4.3, after upgrading the XF version above 4.3 it's not working. The label (placeholder) when animated is getting disappear inside frame instead of sticking to the edge of frame.
### Steps to Reproduce

1. A custom control contains a entry and a label control and both are placed inside a grid to overlap
2. This custom control is placed inside a frame with outline border.
3. Animate the label (placeholder) to the top edge of frame.
4. Observe the behaviour once in Xamarin.Forms version 4.3 and then observe the behaviour after upgrading the Xamarin.Forms version above 4.3 or latest stable version.

### Source Code
https://github.com/himanshudwd765/XFMaterialEntry

### Expected Behavior:- Works with XF 4.3
The label should animate and stick to top of the frame's edge. This works fine for android but the when label is animated is not overlapping on top of frame's edge in case of iOS.
xf4
### Actual Behavior:- Doesn't work with XF above 4.3
The label is animating but not sticking to top of Frame's outline edge, after upgrading to Xamarin.Forms version above 4.3
xf45

### Basic Information

- Version with issue: Above Xamarin.Forms 4.3 including latest version
- Last known good version: XF 4.3.0.90867
- IDE: Visual studio

jfversluis:
### Description of Change ###

Overhauls the iOS `Frame` implementation to overcome the bugs introduced by #7518.

With this change we achieve the same thing but without the bugs and makes it more sustainable to add potential other features. Basically what happens is I add a second layer which holds the actual content so that can be clipped, etc. and the original `UIView` that has always been there is just responsible for the shadow.

For the implementation I have gotten some inspiration from PancakeView, so it's been tested for a while too.

### Issues Resolved ### 

- fixes #10422  ✅ 
- fixes #10348 ✅ 
- fixes #10333 ✅ 
- fixes #10166 ✅ 
- fixes #10024 ✅ 
- fixes #9990 ✅ 
- fixes #9774 ✅ 
- fixes #9555 ✅
- fixes #10326 ✅
- fixes #10906 ✅

### API Changes ###
 
None

### Platforms Affected ### 

- iOS

All the bugs introduced by #7518 should be gone and the `Frame` should work even better than before.

However, people that based a renderer off of the previous FrameRenderer might in some (edge) cases find that something breaks because I moved the content of the `Frame` to a separate layer. 
I think we should make the extra layer available so users can use that as well, but I think that will be something for a separate PR in vNext to not introduce new APIs on previous versions.

Not applicable

### Testing Procedure ###
Go through as much `Frame` test cases as you can and verify things work as you'd expect. The bugs introduced mostly had to do with transforms. `RotationX`, `TranslateY` that kind of things. The previous solution introduced a separate `UIView` that did not take into account any of those things.

Also, pay special attention to gesture recognizers.

I've added reproductions in the gallery for almost all issues mentioned above.

### PR Checklist ###


jfversluis:
Added DNM for now. I want to verify that all the linked issues are resolved by the NuGets generated from this.

Xamarin.Forms.Controls/CoreGalleryPages/FrameCoreGalleryPage.cs
```diff
@@ -17,23 +18,27 @@ protected override void InitializeElement (Frame element)
 			element.BorderColor = Color.Olive;
 		}
 
+		readonly Random _randomValue = new Random();
 		protected override void Build (StackLayout stackLayout)
 		{
 			base.Build (stackLayout);
 
 			var hasShadowContainer = new StateViewContainer<Frame> (Test.Frame.HasShadow, new Frame { HasShadow = true });
-			var outlineColorContainer = new StateViewContainer<Frame> (Test.Frame.OutlineColor, new Frame { BorderColor = Color.Teal, });
```
jfversluis, line 25: API is obsolete, so removed this test case

jfversluis:
Implemented a slightly better solution which allows to keep the same base class for this renderer and only adds a separate frame that will hold the actual content. Taking off the API change label.

Still want to take these NuGets and go through the "fixed issues" list to see if this works in all of these scenarios.

jfversluis:
Removed DNM. I verified that this fixes all of the issues mentioned in the OP. Unfortunately, it seems like there is some CarouselView test in the way of making that all green. But that doesn't seem related at all.

jfversluis:
Note to person merging; this does reopen #5108 (needs to happen manually) unfortunately, which is the initial bug that was attempted to fix that introduced all these regressions

samhouts:
Hmm, new failures:

Carousel View UITests.Carousel View Horizontal("CarouselView (Code, Horizontal)")
Carousel View UITests.Carousel View Remove And Update Current Item("CarouselView (XAML, Horizontal)")
Carousel View UITests.Carousel View Remove First Current Item("CarouselView (XAML, Horizontal)")
Carousel View UITests.Carousel View Remove Last Current Item("CarouselView (XAML, Horizontal)")
Carousel View UITests.Carousel View Vertical("CarouselView (Code, Vertical)")

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue10024.xaml
```diff
@@ -0,0 +1,31 @@
...
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             mc:Ignorable="d"
+             xmlns:controls="clr-namespace:Xamarin.Forms.Controls"  
+             x:Class="Xamarin.Forms.Controls.Issues.Issue10024">
+    <AbsoluteLayout VerticalOptions="FillAndExpand" HorizontalOptions="FillAndExpand">
+        <AbsoluteLayout.GestureRecognizers>
+            <TapGestureRecognizer Tapped="TapGestureRecognizer_Tapped" />
+        </AbsoluteLayout.GestureRecognizers>
+
+        <Label Text="Tap anywhere on the screen to show a animated frame and tap outside the new frame to hide it. It show show no visual atrifacts while animating" />
```
samhouts, line 14:
```suggestion
        <Label Text="Tap anywhere on the screen to show a animated frame and tap outside the new frame to hide it. It show show no visual artifacts while animating" />
```

rmarinho:
Failing test on iOS
` VerifyTapBubbling("Frame",True)_apple_iphone_11_13_1_3`

PR #10498, opened 2020-04-28
[Enhancement] Expander: Direction support

### LINKED ISSUE #10914: [Enhancement] [Expander] Set expander direction
## Summary

It should be possible to set the expansion direction of the Expander control. Currently this is limited only to being expanded downwards but it should be possible to expand upwards or sideways too.

## API Changes

This would be with a property such as 'Direction', for example:

`<Expander Direction="Left"/>`

## Intended Use Case

My use case is an Expander which currently expands from the top of the page on mobile devices but I would like to expand sideways from the right of the page on tablet devices. I can also envisage situations where you might want to expand from the bottom of the page such as for an order summary.

AndreiMisiukevich:
### Description of Change ###
Exposed a new property **Orientation** for determining expander direction.

### Fixes

- fixes #10914

### API Changes ###
Added:
 - ExpanderOrientation Expander.Orientation { get; set; } // Bindable Property

```csharp
public enum ExpandDirection
{
    Bottom,
    Top,
    Left,
    Right
}
```

### Platforms Affected ### 
- Core

### Sample ### 
expander

### Testing Procedure ###
- Start control's gallery sample project;
- Navigate to ExpanderDirectionGallery;
- Smoke test it.


Xamarin.Forms.Core/ExpanderOrientationExtensions.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Xamarin.Forms
+{
+	public static class ExpanderOrientationExtensions
```
AndreiMisiukevich, line 3: I could make it internal or even remove, but I think it would be useful for users

Xamarin.Forms.Core/ExpanderOrientation.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Xamarin.Forms
+{
+	public enum ExpanderOrientation
+	{
+		TopToBottom,
+		BottomToTop,
+		LeadingToTrailing,
+		TrailingToLeading
```
AndreiMisiukevich, line 8: Used Leading/Trailing because of "RTL"

rmarinho: StartToEnd ? Like we have LayoutOptions

AndreiMisiukevich: > StartToEnd ? Like we have LayoutOptions

I thought about it. But Start & End are used both for Vertical and Horizontal LayoutOptions.

Then it would make sense to rename all options in this enum
VerticalStartToEnd
VerticalEndToStart
HorizontalStartToEnd
HorizontalEndToStart

AndreiMisiukevich:
@rmarinho 

I guess I should rename the property to "Direction",
But I am not sure that "Right" and "Left" names are good

How about

Direction { VerticalEnd, VerticalStart, HorizontalEnd, HorizontalStart }

image

divyesh008:
> @rmarinho
> 
> I guess I should rename the property to "Direction",
> But I am not sure that "Right" and "Left" names are good
> 
> How about
> 
> Orientation { Vertical, Horizontal }
> Direction {StartToEnd, EndTostart }
> 
> ?
> 
> image

It can be ExpandInLeft, ExpandInRight

Xamarin.Forms.Core/ExpanderOrientationExtensions.cs
AndreiMisiukevich: Made internal for now

AndreiMisiukevich:
@rmarinho @samhouts 

Updated this PR: made consistent with UWP Expander's API

AndreiMisiukevich:
@rmarinho are we OK with this enhancement?
This PR introduces a lot of internal changes and it's quite difficult to work on other things (Don't want to resolve difficult merge conflicts then :) )

haavamoa:
As `SwipeView` now has a elegant way of opening it programtically, I feel a bit confused to what the differences is between that and this. I get that one is mainly used to swipe while this one is clickable, but they can both be used to do the same thing and has the same directions. hHve you guys considered merging the two components to simplify a bit more @jsuarezruiz and @AndreiMisiukevich?

AndreiMisiukevich:
seems this one is stuck :)

If you need some code clean-up/update from my side, then just ping me and I will do it 👍

AndreiMisiukevich:
Moving Expander to XCT

PR #10500, opened 2020-04-29
iOS 10 fix for Top Tabs on Shell

PureWeen:
### Description of Change ###
- On iOS 10 when reloading after structure changes we need to invalidate the CollectionViewLayout otherwise we get an inconsistency exception

### Platforms Affected ### 
- iOS

### Testing Procedure ###
- play around with included ui test
- ui test should pass on ios 10 

### PR Checklist ###


rmarinho:
Test pass on iOS10 too! wee

PR #10515, opened 2020-04-30
Set Current Item correctly after items cleared

PureWeen:
### Description of Change ###
- re-adding items wasn't correctly resetting CurrentItem up the hierarchy
- don't propagate structure changes from items that aren't visible
- null check renderer on iOS in case it's already been disposed of

Changes aren't perfect. The Collections should more conservatively propagate themselves to the renderers while they are in flux, but that's a different PR for the next official release.

This should fix some additional rough edges on 4.6 for now

### Platforms Affected ### 
- iOS

### Testing Procedure ###
- click around on the shell is visible tests

### PR Checklist ###


rmarinho:
Seems tests are failing

PR #10523, opened 2020-04-30
[X] Xaml BindingDiagnostics

StephaneDelcroix:
### Description of Change ###

first pass on issuing binding diagnositcs messages on failure, with xaml source info

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###


maxbrister:
@StephaneDelcroix @spadapet reviving feedback lost in teams.
1. We need some way of caching binding diagnostics until the tap is cached. Wpf keeps the first 2k events and flushes them when the first listener attaches. See [here](https://github.com/dotnet/wpf/blob/a8dcc312d825d543511d6a031ccecb2e8955cff6/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingDiagnostics.cs).
2. The binding object itself doesn't have any source info. In Wpf Peter leverages BindingExpressionBase.Target to get the source information from the target. Since Xamarin doesn't have a similar field, we'll need to be explicitly passed the target object.

spadapet:
> 2. The binding object itself doesn't have any source info. In Wpf Peter leverages BindingExpressionBase.Target to get the source information from the target. Since Xamarin doesn't have a similar field, we'll need to be explicitly passed the target object.

Better yet would be the source info for the actual {Binding}, rather than the target element. With WPF, the target element was as close as I could get to the property with the {Binding}. But if Xamarin actually knows the source location of the {Binding}, providing that would allow for a more accurate "go to code" action.

StephaneDelcroix:
@maxbrister we're not going to start piling up info that might eventually be consumed later on, on mobile where resources are scarce, until the time we can detect with 99% certainty that we are running under a HotReload scenario

Xamarin.Forms.Core/BindingExpression.cs
```diff
@@ -149,6 +151,7 @@ void ApplyCore(object sourceObject, BindableObject target, BindableProperty prop
 				if (!TryConvert(ref value, property, property.ReturnType, true))
 				{
 					Log.Warning("Binding", "'{0}' can not be converted to type '{1}'.", value, property.ReturnType);
+					BindingDiagnostics.SendBindingFailure(Binding, null, "{0} can not be converted to type '{1}'", new[] { value, property.ReturnType });
```
spadapet, line 154: Add "current, target, property" to this and the next SendBindingFailure.

spadapet:
> @maxbrister we're not going to start piling up info that might eventually be consumed later on, on mobile where resources are scarce, until the time we can detect with 99% certainty that we are running under a HotReload scenario

WPF does it with an environment variable that VS sets on the debugged process. Without caching the BindingFailed events for Xamarin, the initial set of binding failures on app start won't have source information. The failures won't get lost since they will still be found in the debug output.

PR #10524, opened 2020-04-30
On requested theme changed android

fredyadriano90:
### Description of Change ###

The event `Xamarin.Forms.Application.RequestedThemeChanged` is fired when the device rotates. 

### Issues Resolved ### 

Fire the event `Xamarin.Forms.Application.RequestedThemeChanged` only when the theme change

- fixes #

### API Changes ###

none

### Platforms Affected ### 

- Android

None

Not applicable

### Testing Procedure ###

- On the App.cs subscribe to the event `Xamarin.Forms.Application.RequestedThemeChanged`
- Rotate the device
- The event will be fired

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Application.cs
+++ b/Xamarin.Forms.Core/Application.cs
@@ -39,6 +39,8 @@ namespace Xamarin.Forms
 			SystemResources = DependencyService.Get<ISystemResourcesProvider>().GetSystemResources();
 			SystemResources.ValuesChanged += OnParentResourcesChanged;
 			_platformConfigurationRegistry = new Lazy<PlatformConfigurationRegistry<Application>>(() => new PlatformConfigurationRegistry<Application>(this));
+			// Initialize this value, when the app loads
+			LastAppTheme = RequestedTheme;
 		}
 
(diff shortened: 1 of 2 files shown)


fredyadriano90:
I would like to keep LastAppTheme public property to check that value from Android and do not fire the RequestTheme event without need.

fredyadriano90:
@jfversluis the idea of `LastAppTheme` public is to not dispatch the event

jfversluis:
I see why you are doing it! But as a side-effect we're exposing a new API to our users that will then have to maintain. And yes, this one is not very exciting but also not very useful to the end-user IMO.

My preference would be to keep it private.

fredyadriano90:
@jfversluis  I got your point, and I think that you are right. The only thing that I would like to add is the initialization to lastAppTheme on the ctor

fredyadriano90:
@jfversluis Hi, I was testing XF vs 4.6.0.800 and if you don't initialize the _lastAppTheme on the ctor the RequestedThemeChanged event is invoked as soon as you rotate the device

PR #10429, opened 2020-04-23
Make managed cookies opt-in

steverichey:
### Description ###

Added a new bound property that controls whether or not Xamarin Forms manages web view cookies on iOS, Android, and UAP. I wasn't sure if this should go into master or the 4.6.0 branch; let me know if I need to target a different branch.

### Issues Resolved ### 

- fixes #10318 

### API Changes ###

Added:
 - bool WebView.ShouldManageCookies { get; set; } // Bindable Property

### Platforms Affected ### 

- Core

### Behavioral/Visual Changes ###

This will impact users on the pre-release 4.6.0 version of Forms; the cookie management behavior introduced in #8169 will no longer be automatic; it is now opt-in.

### Testing Procedure ###

I was unable to build Xamarin Forms on my MacBook, so I wasn't able to build a test case; a simple project with a webview that requires cookies would suffice.

### PR Checklist ###

- [ ] Targets the correct branch
- [ ] Tests are passing (or failures are unrelated)

PR #10526, opened 2020-04-30
Use ShouldManageCookies in WkWebViewRenderer

### LINKED ISSUE #10318: [Bug] [Android] Pre-release version of Xamarin Forms forgets cookies
### Description

We have an application that presents content in web views within the application. We authenticate requests for content using a cookie. As of the latest pre-release for Xamarin Forms (which we are using for various other bug fixes) we have noticed that this authentication cookie will vanish when loading content. This appears to be the result of changes in [this](https://github.com/xamarin/Xamarin.Forms/pull/8169/files) pull request.

### Steps to Reproduce

1. Send a JWT to a server endpoint to retrieve an authentication cookie
2. Attempt to load content requiring authentication using this cookie

### Expected Behavior

Content loads as normal, as the user has a valid session cookie.

### Actual Behavior

The user is redirected to the login screen, as no valid session cookie exists.

### Basic Information

- Last known good version: I believe we were using 4.5 with no issue
- IDE: Visual Studio
- Platform Target Frameworks: 
  - Android: 9.0
- Android Support Library Version: 28

### Reproduction Link

Providing a reproduction project would be challenging, as you would need a server, authentication, a webview, and so on.

### Workaround

No known workaround as yet.

steverichey:
### Description of Change ###

This is to resolve an oversight on my part in an earlier pull request, #10429. I forgot that iOS has two web view renderers. This enables the opt-in cookie behavior I added to the other web view renderers for `WkWebViewRenderer`.

fixes #10318

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###

I was unable to test this change as I am not able to build Forms in Visual Studio for Mac.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/WkWebViewRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/WkWebViewRenderer.cs
@@ -103,7 +103,7 @@ namespace Xamarin.Forms.Platform.iOS
 			var safeRelativeUri = new Uri($"{uri.PathAndQuery}{uri.Fragment}", UriKind.Relative);
 			NSUrlRequest request = new NSUrlRequest(new Uri(safeHostUri, safeRelativeUri));
 
-			if (WebView.Cookies != null)
+			if (WebView.Cookies != null && WebView.ShouldManageCookies)
 			{
 				if (Forms.IsiOS11OrNewer)
 				{


PureWeen:
@steverichey 

Instead of adding a new property `ShouldManageCookies` 

What if we just don't clear the cookies?

If the Forms Cookie Collection is empty just don't do anything.

Just make the cookies additive for now to fix the regression and then for 4.7 or a 4.6 SR we can create a better syncing behavior.

I think the way this would work is that the WebView.Cookies collection would just fill up with the cookies returned and then users could add/remove them

steverichey:
@PureWeen I'm open to alternative implementations, but I'm not sure it would be clear to the user if the API behaves differently when the cookies are empty. In any case, this pull request is just to get this renderer to behave the same as the others; I think a change to the API would require another pull request after this is merged.

PureWeen:
>  I'm not sure it would be clear to the user if the API behaves differently when the cookies are empty

can you expand on this? 

What I'm proposing would make it work a bit more like asp.net response/request cookies are managed. 

So the Cookie Collection is a communication between the xplat and the native. Cookies from the webview sync to the cookie collection and then users can choose to clear those or not

steverichey:
> can you expand on this?

The behavior when the user has set one cookie is quite different from when they set zero cookies, if we maintain the current behavior: with no cookies the native cookies would remain, but with one cookie the native cookies are removed.

Again, I'm fine with changes, but I think those would be a separate PR. This is just to get iOS renderers to behave consistently.

PureWeen:
>  but with one cookie the native cookies are removed.

I'm recommending we should change the current behavior to be additive only to at least fix the regression for 4.6 SR1.  Then we set our sights on an implementation that is more synchronization based for 4.6 SR? or 4.7

It's not ideal but I think it's more correct for what the future implementation will look like.

I feel like this second API is just going to get obsoleted

steverichey:
Okay, since the scope of this change has changed significantly and this addition is no longer needed, I'm going to open a new pull request to remove `ShouldManageCookies` entirely. I'll link that merge to this one when it's ready.

PureWeen:
@steverichey Thank you! But don't worry about it

I restructured how the Cookie parts here
https://github.com/xamarin/Xamarin.Forms/pull/10571

The fundamental issue of the breaking change was mainly that Android cleared the cookies no matter what but in my PR I'm just converting the cookie container over to being more of a syncing mechanisms so you can add/remove

If you don't set the Cookies to anything that will resolve the initial issue you logged here
https://github.com/xamarin/Xamarin.Forms/issues/10318

I'm also curious for you to test your project by setting the cookie container.

Setting it should no longer be destructive

PR #10539, opened 2020-05-01
Adding the parameter less method to Shell.GoBackAsync()

ederbond:
### Description of Change ###

Added a handy parameter less method to Shell to allow navigating backward without having to pass ".."

### API Changes ###

- Core/XAML (all platforms)
- iOS
- Android
- UWP

- None

- Not applicable

### Testing Procedure ###
Just perform multiple forward navigation using Shell and then call 
Shell.GoBackAsync() 
and see you app navigating backward 1 time;
There's also a unit test on this PR called GoBackWithoutDotDotTests

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/ShellModalTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/ShellModalTests.cs
@@ -317,6 +317,24 @@ namespace Xamarin.Forms.Core.UnitTests
 		}
 
 		[Test]
+		public async Task GoBackWithoutDotDotTests()
+		{
+			var shell = new Shell();
+
+			var item = CreateShellItem(shellSectionRoute: "section2");
+			shell.Items.Add(item);
+
+			await shell.GoToAsync(new ShellNavigationState($"ModalTestPage"));
+			var modalPage = (shell.CurrentItem.CurrentItem as IShellSectionController).PresentedPage as ModalTestPageBase;
+			Assert.NotNull(modalPage);
+
+			await shell.GoBackAsync();
+
+			var previousPage = (shell.CurrentItem.CurrentItem as IShellSectionController).PresentedPage as ContentPage;
+			Assert.NotNull(previousPage);
+		}
+
+		[Test]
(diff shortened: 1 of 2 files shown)


PureWeen:
I'm on the fence about adding something that is specifically `GoBackAsync`. 

Mainly because I would ask the question "GoBack" to what? 

GotoAsync("..") has meaning because it's saying take the current route and strip the last route from the current `uri`

GoBack I feel is going to open up a lot of scenarios where people will be confused about why it didn't go back to the last tab, or go back to the last flyout item clicked.

The discoverability of ".." isn't great so maybe there's another way to go about this. 

Maybe an Enum with common concepts? though I can't think of too many other enums to use 

`GotoAsync(ShellPath.RemoveLastRoute)`

Maybe some sort of PathBuilder or extension methods for NavigationState that can perform operations against a Uri

NavigationState.AddRouteWithPArams
NavigationState.PopRoutes(count)

Etc..

Thoughts?

pictos:
@PureWeen I think the `".."` is ok to use (for me better than the Enum approach). Also, if it's possible, we can extend this behavior, and use `"..\.."` to Go Back two times, just like the terminal syntax.

PureWeen:
@pictos 

"..\.."

Should work.

All the code does is use the Path apis to apply ".." to whatever the current route is

https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/Shell/ShellUriHandler.cs#L20

And then it calls gotoasync with whatever that route is

so if you have

//route/page1/page2

and call gotoasync(..\..)

That should remove page1/page2

pictos:
@PureWeen, I see, I tried with just `\` and it throws an exception, I'm ok to use `/` or `\\`. Tks

PureWeen:
@pictos Aw!! I'll need to fix that

The intent is to make so you can use whatever slash direction you desire :-)

pictos:
@PureWeen, it's my fault I didn't use the `@` before, so the C# was using an invalid character.

samhouts:

PR #10571, opened 2020-05-05
Make cookie handling consistent and non destructive

### LINKED ISSUE #10318: [Bug] [Android] Pre-release version of Xamarin Forms forgets cookies

PureWeen:
### Description of Change ###

- UWP wasn't deleting cookies at all or updating cookies on refresh
- Android was just deleting all the cookies no matter what
- iOS was only deleting when cookies > 0

Determining how to do handle Cookies on each platform was different. This PR brings all the platforms in line. If the user never sets the Cookies property than nothing ever happens.

This PR also removes the API `ShouldManageCookies` which isn't public yet. Because if the CookieContainer is null that will already act as if `ShouldManageCookies` is set false. Android was the only platform that was totally destructive even when Cookies was null

This PR keeps the Cookies property in sync with the webview cookies. After the first load of a url the Cookies container (if set) will fill with cookies. At this point if you add/remove cookies it'll try to add/remove those from the webview

I also added calls to sync cookies after the navigating call this way people can modify the cookies during navigating if they need to

### Issues Resolved ### 

- fixes #10318

### Platforms Affected ### 
- iOS
- Android
- UWP

### Testing Procedure ###
- ui tests included

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/WkWebViewRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/WkWebViewRenderer.cs
@@ -108,7 +108,7 @@ namespace Xamarin.Forms.Platform.iOS
 				var safeRelativeUri = new Uri($"{uri.PathAndQuery}{uri.Fragment}", UriKind.Relative);
 				NSUrlRequest request = new NSUrlRequest(new Uri(safeHostUri, safeRelativeUri));
 
-				await SyncCookies(url);
+				await SyncNativeCookies(url);
 				LoadRequest(request);
 			}
 			catch (Exception exc)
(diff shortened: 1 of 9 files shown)


PureWeen:
This PR is fine to review. The API 19 LR test failed because one of the buttons is semi off the screen so the app runner couldn't click on it. Pushing up a button re-arrangement which should fix

Xamarin.Forms.Platform.iOS/Renderers/WkWebViewRenderer.cs
```diff
@@ -228,23 +438,25 @@ void UpdateCanGoBackForward()
...
-		{ 
-			var jCookies = WebView.Cookies.GetCookies(url);
-
+		string GetCookieString(List<Cookie> existingCookies)
+		{
 			StringBuilder cookieBuilder = new StringBuilder();
-			foreach (System.Net.Cookie jCookie in jCookies)
+			foreach (System.Net.Cookie jCookie in existingCookies)
 			{
+				// cookiesToKeep.Add(jCookie.Name);
```
samhouts, line 446: delete comment?

Xamarin.Forms.Platform.iOS/Renderers/WkWebViewRenderer.cs
```diff
@@ -167,6 +153,220 @@ protected virtual void OnElementChanged(VisualElementChangedEventArgs e)
...
+				return;
+
+			var retrieveCurrentWebCookies = await GetCookiesFromNativeStore(url);
+
+			List<Cookie> cookiesToSet = new List<Cookie>();
+			foreach (Cookie cookie in cookies)
+			{
+				bool changeCookie = true;
+
+				// on iOS 10 we have to rewrite all the cookies
```
samhouts, line 283: this comment might need more explanation

Xamarin.Forms.Platform.iOS/Renderers/WkWebViewRenderer.cs
PureWeen, line 446:
```suggestion
```

PR #10442, opened 2020-04-24
Runtime updating of AppTheme values

jfversluis:
### Description of Change ###

This implements the runtime updating of AppTheme values.

### Issues Resolved ### 

- fixes #10391

### API Changes ###

Added:
 - void BindableObject.SetOnAppTheme<T>(T Light, T Dark, T Default = default);
 - void BindableObject.SetAppThemeColor(Color Light, Color Dark, Color Default = default);

### Platforms Affected ### 

- Core/XAML (all platforms)

### Behavioral/Visual Changes ###

The `INotifyPropertyChanged` was removed from the API. It doesn't make much sense to have that on there (anymore). People that were using that might see breaks.

#### iOS
ezgif com-video-to-gif

#### Android
ezgif com-video-to-gif (1)

### Testing Procedure ###
Go to the right gallery page(s) and switch appearance

### PR Checklist ###

- [ ] Targets the correct branch
- [ ] Tests are passing (or failures are unrelated)

PR #10592, opened 2020-05-06
[C/X] Clean API and impl of OnAppTheme

### LINKED ISSUE #10395: [Bug] Unable to support AppTheme and user preference at same time
### Description

Consider this scenario:

image

In this case as a user I can choose to let my app opt-in to OS changes as they occur, or to control dark and light myself.

Note these choices in this UI are mutually exclusive. As soon as I choose to toggle the dark mode, I've opted out of OS changes (aka device settings).

image

Similarly if I opt-in to "use device settings" then I can no longer manually set the dark mode on or off.

Currently I see these problems:

* cannot opt out of OS theme changes
* cannot manually set user preferred theme

I propose adding:

```csharp
public enum AppTheme
	{
		Unspecified,
		Light,
		Dark
	}
```

Then `OSAppTheme` is that type, but it can be shared by a new type below.

```csharp
public bool Application.UseOSAppTheme { get; set; } = true;
```

When `OSAppTheme` changes are requested by the OS, this bool should be checked before triggering any changes.

```csharp
public AppTheme UserAppTheme {get; set; }
```

Some variable to store the user chosen preference. Perhaps OSTheme is a new common type. Then when the user chooses to toggle their app theme between dark and light modes, they can use an API like:

```csharp
App.Current.UserAppTheme = AppTheme.Dark;
```

This I think also makes more sense with the other naming of `AppThemeColor` and `OnAppTheme` which can be used both for OS theme changes and user theme changes.

StephaneDelcroix:
### Description of Change ###

With the issues being fixed in #10442 (thx @jfversluis), I re-reviewed the exposed API and implementation of App Theme Binding.

This ensure the exposed API is consistent. See review comments for the various reasons and explanations.

It also enables a manual override for app theme (#10395)

### Issues Resolved ### 

- fixes #10395

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###


Xamarin.Forms.Core/AppThemeColor.cs
```diff
@@ -1,4 +0,0 @@
-﻿namespace Xamarin.Forms
-{
-	public class AppThemeColor : OnAppTheme<Color> { }
```
StephaneDelcroix, line 3: I was going to turn this helper internal, but if it's internal, it's not needed at all

Xamarin.Forms.Core/Application.cs
```diff
@@ -167,7 +167,7 @@ public ResourceDictionary Resources
 		OSAppTheme _lastAppTheme;
 
 		[EditorBrowsable(EditorBrowsableState.Never)]
-		public void OnRequestedThemeChanged(AppThemeChangedEventArgs args)
+		public void TriggerThemeChanged(AppThemeChangedEventArgs args)
```
StephaneDelcroix, line 170: we use OnFooxxxed for event handlers, not triggers. changed the name

Xamarin.Forms.Core/BindableObjectExtensions.cs
```diff
@@ -67,21 +67,5 @@ public static T GetPropertyIfSet<T>(this BindableObject bindableObject, Bindable
 
 			return returnIfNotSet;
 		}
-
-		public static void SetOnAppTheme<T>(this BindableObject self, BindableProperty targetProperty, T light, T dark, T defaultValue = default)
```
StephaneDelcroix, line 71: moved

Xamarin.Forms.Core/OnAppTheme.cs
```diff
@@ -2,29 +2,14 @@
 
 namespace Xamarin.Forms
 {
-	public class OnAppTheme<T> : BindingBase
+	class OnAppTheme<T> : BindingBase
```
StephaneDelcroix, line 5: I don't think we want to expose this as-is, as it has little value unless used as a Binding. Keeping this would be confusing, and would led to customers no getting the intended behavior.

Xamarin.Forms.Core/OnAppTheme.cs
```diff
@@ -2,29 +2,14 @@
...
 
-		void ThemeChanged(object sender, AppThemeChangedEventArgs e)
-		{
-			Device.BeginInvokeOnMainThread(() => ApplyCore());
-		}
-		public new BindingMode Mode
-		{
-			get => base.Mode;
-			private set { }
-		}
```
StephaneDelcroix, line 23: as it's not public anymore, no need to override Mode

Xamarin.Forms.Core/OnAppTheme.cs
```diff
@@ -39,6 +24,14 @@ internal override void Apply(object context, BindableObject bindObj, BindablePro
 			base.Apply(context, bindObj, targetProperty, fromBindingContextChanged);
 			ApplyCore();
 		}
+
+		internal override void Unapply(bool fromBindingContextChanged = false)
+		{
```
StephaneDelcroix, line 29: not really required, but I like to have it here

Xamarin.Forms.Core/Page.cs
```diff
@@ -250,13 +250,6 @@ public bool SendBackButtonPressed()
 			return OnBackButtonPressed();
 		}
 
-		protected override void OnRequestedThemeChanged(OSAppTheme newValue)
```
StephaneDelcroix, line 253: this seemed a workaround. also, why only reload resources on pages ?

Xamarin.Forms.Core/VisualElement.cs
```diff
@@ -275,13 +275,6 @@ void SetEffectiveFlowDirection(EffectiveFlowDirection value, bool fireFlowDirect
 
 		internal VisualElement()
 		{
-			if (Application.Current != null)
-				Application.Current.RequestedThemeChanged += (s, a) => OnRequestedThemeChanged(a.RequestedTheme);
-		}
-
-		protected virtual void OnRequestedThemeChanged(OSAppTheme newValue)
```
StephaneDelcroix, line 282: dropped, as I dropped the Page override

Xamarin.Forms.Core/VisualElementExtensions.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Xamarin.Forms
+{
+	public static class VisualElementExtensions
+	{
+		public static void SetOnAppTheme<T>(this VisualElement self, BindableProperty targetProperty, T light, T dark)
```
StephaneDelcroix, line 8: move from BOExtensions. make more sense to keep that on VE

Xamarin.Forms.Xaml/MarkupExtensions/AppThemeBindingExtension.cs
```diff
@@ -0,0 +1,100 @@
+﻿using System;
+using System.Globalization;
+using System.Reflection;
+
+namespace Xamarin.Forms.Xaml
+{
+	[ContentProperty(nameof(Default))]
+	public class AppThemeBindingExtension : IMarkupExtension<BindingBase>
```
StephaneDelcroix, line 8: changed the name so it contains Binding (like TemplateBinding, etc...)

Xamarin.Forms.Xaml/MarkupExtensions/AppThemeBindingExtension.cs
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		public object Default { get; set; }
+		public object Light { get; set; }
+		public object Dark { get; set; }
+		public object Value	{ get; private set;	}
+
+		public object ProvideValue(IServiceProvider serviceProvider) => (this as IMarkupExtension<BindingBase>).ProvideValue(serviceProvider);
+
+		BindingBase IMarkupExtension<BindingBase>.ProvideValue(IServiceProvider serviceProvider)
```
StephaneDelcroix, line 22: drop the converter as it was only working if the object passed was a Xaml primitive type

Xamarin.Forms.Xaml.UnitTests/OnAppThemeTests.cs
```diff
@@ -0,0 +1,159 @@
+﻿using NUnit.Framework;
+
+using Xamarin.Forms.Core.UnitTests;
+
+namespace Xamarin.Forms.Xaml.UnitTests
+{
+	[TestFixture]
+	public class OnAppThemeTests : BaseTestFixture
```
jsuarezruiz, line 8: :heart_eyes:

PR #10632, opened 2020-05-10
PresentsWithGesture incorrectly toggled back to true when Flyout Disabled

PureWeen:
### Description of Change ###

On an iPad the PresentsWithGesture is being incorrectly re-enabled when the flyout behavior was set to disabled

### Platforms Affected ### 
- iOS

### Testing Procedure ###
- On an iPad run the store shell or go to the "shell Flyout Behavior" issue and play with disabling the flyout. Once you've disable it try to swipe it open with a gesture

### PR Checklist ###


EmilAlipiev:
is that fix only for Shell? how about non-shell?

Issue #9029, opened 2019-12-27
[Bug] WPF: ViewCell content width does incorrect update


Xamarin2019:
### Description
If the ListView width changes on WPF (e.g. on resizing the window), the part of ViewCell's  is resized but the others ViewCell's is not resized. The behavior depends on the content, if StackLayout, then when the application starts,  the first few Viewcell's are incorrect width. After repeatedly changing the window width and scrolling, Viewcell's have different widths in the same ListView.

### Steps to Reproduce
Try to make Viewcell's in the ListView with elements bound to the left and right side.
For example: the text of the left edge and the checkbox on the right

### Expected Behavior
The ViewCell's content should resize whenever the ListView's width changes, such that the checkboxes stay near the right window border.

### Actual Behavior
Checkboxes disappear beyond the window border or may have a different distance from the window border in the same ListView.

### Basic Information

- Version with issue: Latest stable 4.4.0.991265
- IDE: Visual Studio 15.9.18
- Platform Target Frameworks: WPF: .NET Framework 4.6.1

### Screenshots
In the first case of ListView, I applied StackLayout in the second Grid, and in the third I tried to fix this behavior by changing the width of Viewcell's on the width change event, but it still doesn't always work.
On Android all three Listviews work as expected.

**At start**
![image](https://user-images.githubusercontent.com/59281307/71521030-1642b300-28d0-11ea-8309-130ec8654469.png)
**After scrolling**
![image](https://user-images.githubusercontent.com/59281307/71521075-383c3580-28d0-11ea-84be-a21e5500b9f7.png)
**full screen mode**
![image](https://user-images.githubusercontent.com/59281307/71521254-fc55a000-28d0-11ea-848b-1cb5e5fcc56d.png)

**Return from full screen mode**
![image](https://user-images.githubusercontent.com/59281307/71521300-2eff9880-28d1-11ea-8297-bf1423fe4569.png)
**Changing the window width and scrolling**
![image](https://user-images.githubusercontent.com/59281307/71521344-6bcb8f80-28d1-11ea-8d1a-b8902c9b90d4.png)

**On Android 4.4 is OK**
![image](https://user-images.githubusercontent.com/59281307/71520982-e1366080-28cf-11ea-9f97-a10346ad924f.png)

### Reproduction Link

[AppListViewBug.zip](https://github.com/xamarin/Xamarin.Forms/files/4005597/AppListViewBug.zip)
without **packages** folder

[...]

Issue #8039, opened 2019-10-16
[Bug] WPF ListView ArgumentOutOfRangeException when tapped


breyed:
In WPF, if you tap the first item in a ListView with grouping enabled, the following unhandled exception is thrown:
> System.ArgumentOutOfRangeException: 'Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')'

Clicking with the mouse works. The exception only occurs when using a touchscreen.

Call stack:
```
System.Private.CoreLib.dll!System.ThrowHelper.ThrowArgumentOutOfRange_IndexException()
System.Private.CoreLib.dll!System.Collections.Generic.List<System.__Canon>.System.Collections.IList.get_Item(int index)
Xamarin.Forms.Core.dll!Xamarin.Forms.ListProxy.TryGetValue.AnonymousMethod__0()
Xamarin.Forms.Core.dll!Xamarin.Forms.ListProxy.TryGetValue(int index, out object value)
Xamarin.Forms.Core.dll!Xamarin.Forms.ListProxy.this[int].get(int index)
Xamarin.Forms.Core.dll!Xamarin.Forms.ListProxy.System.Collections.IList.get_Item(int index)
Xamarin.Forms.Core.dll!Xamarin.Forms.Internals.TemplatedItemsList<Xamarin.Forms.ItemsView<Xamarin.Forms.Cell>, Xamarin.Forms.Cell>.this[int].get(int index)
Xamarin.Forms.Core.dll!Xamarin.Forms.ListView.NotifyRowTapped(int groupIndex, int inGroupIndex, Xamarin.Forms.Cell cell)
Xamarin.Forms.Core.dll!Xamarin.Forms.ListView.NotifyRowTapped(int index, Xamarin.Forms.Cell cell)
Xamarin.Forms.Platform.WPF.dll!Xamarin.Forms.Platform.WPF.ListViewRenderer.OnNativeStylusUp(object sender, System.Windows.Input.StylusEventArgs e)
PresentationCore.dll!System.Windows.Input.StylusEventArgs.InvokeEventHandler(System.Delegate genericHandler, object genericTarget)
PresentationCore.dll!System.Windows.RoutedEventArgs.InvokeHandler(System.Delegate handler, object target)
PresentationCore.dll!System.Windows.RoutedEventHandlerInfo.InvokeHandler(object target, System.Windows.RoutedEventArgs routedEventArgs)
[...]

Issue #6842, opened 2019-07-10
[Bug] [WPF] ViewCell in scrollable list view does not update width when restoring from maximized window


domyd:
### Description

When restoring a WPF Window from maximized state, `ViewCell`s are not resized if they're in a scrollable (in the non-maximized state) `ListView`. The `Xamarin.Forms.Platform.WPF.CellControl.OnRenderSizeChanged` method, which gets called if I simply resize the window horizontally, does _not_ get called in this case.

If the `ListView` doesn't have a scrollbar in the non-maximized state, the `ViewCell` width gets updated as expected.

Related: #5682

### Steps to Reproduce

1. Build and run the attached repro application.
2. Make sure the displayed list view has a scroll bar.
3. Maximize the window
4. Restore the window back to its normal state and size. At this point you should see that the `ViewCell`s are still as wide as they were in the maximized state.
5. Scroll down, which should update the `ViewCell` to their correct size. (Horizontally resizing the window doesn't update the `ViewCell`, however ...)

### Expected Behavior
The `ViewCell`s should resize when the window size is restored from the maximized state, even if they're in a scrollable `ListView`.

### Actual Behavior
They aren't.

### Basic Information

- Version with issue: Stable (4.1.0.581479) as well as Nightly (4.3.0.936-nightly)
- IDE: Visual Studio 2019 v16.1.6
- Nuget Packages: None except Xamarin.Forms
- Affected Devices: Windows 10 v1903 (Build 18362.116)

### Screenshots

State after opening the app. The X's represent the left/right edges of the ViewCell, the test# text is centered.
![step1](https://user-images.githubusercontent.com/5590413/60968338-1b5d7700-a31d-11e9-9b45-31a24f77e31c.png)

Maximize:
![step2](https://user-images.githubusercontent.com/5590413/60968339-1bf60d80-a31d-11e9-9165-15871112d577.png)

Restore:
![step3](https://user-images.githubusercontent.com/5590413/60968341-1bf60d80-a31d-11e9-90d0-f307b2f01c03.png)

Scroll down:
![step4](https://user-images.githubusercontent.com/5590413/60968342-1bf60d80-a31d-11e9-895a-a9764b54bf9c.png)

### Reproduction Link

[ViewCellResize.WPF.zip](https://github.com/xamarin/Xamarin.Forms/files/3377358/ViewCellResize.WPF.zip)

jsuarezruiz:
[...]

Issue #6436, opened 2019-06-06
[Bug] [WPF] Grid + ListView -> first row height not correct


ZeProgFactory:
### Description
If you put a `ListView` in a `Grid` the first `RowHeight` is to small unless you specify a correct default value.

### Steps to Reproduce
```
   <Grid>
      <ListView ItemsSource="{Binding Items}" Margin="50" >
         <ListView.ItemTemplate>
            <DataTemplate>
               <ViewCell>
                  <Grid>
                     <Label HorizontalOptions="Center" Text="{Binding}" />
                  </Grid>
               </ViewCell>
            </DataTemplate>
         </ListView.ItemTemplate>
      </ListView>
   </Grid>
```

### Expected Behavior
![2019-06-06 (2)](https://user-images.githubusercontent.com/17034030/59047525-54b94780-8884-11e9-9fee-b611b6d625da.png)

### Actual Behavior
![2019-06-06 (1)](https://user-images.githubusercontent.com/17034030/59047468-3c492d00-8884-11e9-92fd-156a43be6487.png)

### Basic Information

- Version with issue:
- Last known good version: ?
- IDE: Microsoft Visual Studio Professional 2019 Version 16.1.2
- Nuget Packages: XF 4.0.0.482894

### Reproduction Link
[ViewCellSize.WPF.zip](https://github.com/xamarin/Xamarin.Forms/files/3262448/ViewCellSize.WPF.zip)

jdcarrillo:
> ### Description
> If you put a `ListView` in a `Grid` the first `RowHeight` is to small unless you specify a correct default value.
> 
> ### Steps to Reproduce
> ```
>    <Grid>
>       <ListView ItemsSource="{Binding Items}" Margin="50" >
>          <ListView.ItemTemplate>
>             <DataTemplate>
>                <ViewCell>
>                   <Grid>
>                      <Label HorizontalOptions="Center" Text="{Binding}" />
>                   </Grid>
>                </ViewCell>
>             </DataTemplate>
>          </ListView.ItemTemplate>
>       </ListView>
>    </Grid>
> ```
> 
> ### Expected Behavior
> ![2019-06-06 (2)](https://user-images.githubusercontent.com/17034030/59047525-54b94780-8884-11e9-9fee-b611b6d625da.png)
> 
> ### Actual Behavior
[...]

PR #10653, opened 2020-05-11
[WPF] Fix FormsPanel MeasureOverride negative requested size

tessarolli:
### Description of Change ###

On WPF Platform, Element.Measure().Request(); sometimes return negative values (-1) on Width or Height. If that happened, instantiating a new Windows.Size with negative values would throw an exception that prevented software execution.
This fix is to prevent that from happening, as it checks and replaces -1 values with zero.

### Issues Resolved ### 

Fix Negative Width or Height on new Windows.Size();

### API Changes ###
 None

### Platforms Affected ### 

WPF

None

Not applicable

### PR Checklist ###


Xamarin.Forms.Platform.WPF/FormsPanel.cs
```diff
@@ -82,6 +82,13 @@ protected override System.Windows.Size MeasureOverride(System.Windows.Size avail
 			if (double.IsInfinity(elementDesiredWidth) || double.IsPositiveInfinity(elementDesiredHeight))
 			{
 				Size request = Element.Measure(elementDesiredWidth, elementDesiredHeight, MeasureFlags.IncludeMargins).Request;
+
+				if (request.Width == -1)
```
VladislavAntonyuk, line 86: I think it would be better to check if request.Width < 0. it will cover all invalid values

VladislavAntonyuk:
Thank you for the PR. please make sure you didn't break fixes for: 
#6436
#8039
#9029
#5682
#6842

Xamarin.Forms.Platform.WPF/FormsPanel.cs
tessarolli: OK, did it

tessarolli:
Hi Vladislav,

I've update to check for all negative values.

Personally, I dont believe it will break any of those fixes you mentioned, because in case you do receive a negative request size, it would throw an exception trying to instantiate a new Size(), that would prevent software execution.

I did some testings on my end and all seems fine.

samhouts (approved):

PR #10682, opened 2020-05-13
[Android] Fix NRE accessing a disposed SwipeView

### LINKED ISSUE #10679: [Bug] SwipeView crash with NullReferenceException
### Description
When using swipeview to swipe  navigate to another page, if I remove or pop a page from the stack, the app always crashes with the same error.

### Steps to Reproduce
1. Run the provided demo.
2. Click the button on the homepage to go to Page 1.
3. Swipe Left or Right to go to Page 2.
4. The error is thrown here

### Expected Behavior
Page 2 is displayed and page 1 was popped out of the navigation stack with no errors.

### Actual Behavior
Throws error: "System.NullReferenceException: Object reference not set to an instance of an object"

### AppCenter Trace
``` cs
SwipeViewRenderer.UpdateIsOpen (System.Boolean isOpen)
SwipeViewRenderer.DisposeSwipeItems ()
SwipeViewRenderer.<ResetSwipe>b__71_0 ()
Runnable.Run ()
IRunnableInvoker.n_Run (System.IntPtr jnienv, System.IntPtr native__this)
(wrapper dynamic-method) Android.Runtime.DynamicMethodNameCounter.47(intptr,intptr)
```
### Basic Information
Visual Studio 2019
Xamarin.Forms v4.6

### Reproduction Link
[SwipeViewDemo.zip](https://github.com/xamarin/Xamarin.Forms/files/4622221/SwipeViewDemo.zip)

jsuarezruiz:
### Description of Change ###

Avoid NRE accessing a disposed SwipeView.

### Issues Resolved ### 

- fixes #10679 

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS
- Android

None

fix10679

### Testing Procedure ###
Launch Core Gallery and navigate to the issue 10679. Navigate to the second page and the swipe (left or right). If swiping can navigate to another Page without problems, the test has passed.

### PR Checklist ###


Xamarin.Forms.Platform.Android/Renderers/SwipeViewRenderer.cs
```diff
@@ -1318,7 +1318,8 @@ void ProgrammaticallyOpenSwipeItem(OpenSwipeItem openSwipeItem)
 
 		void UpdateIsOpen(bool isOpen)
 		{
-			((ISwipeViewController)Element).IsOpen = isOpen;
+			if (Element is ISwipeViewController swipeViewController)
+				swipeViewController.IsOpen = isOpen;
```
PureWeen, line 1322: Is this going to lead to wrong behavior upstream?

When testing this I noticed this is called from `DisposeSwipeItems`

So the ISwipeViewController that's associated with this renderers keeps the "IsOpen" value set to true.

Is that ok? Does the dispose need to happen in a different order so that IsOpen correctly propagates to SwipeView?


image

samhouts:
@jsuarezruiz It looks like this was already resolved? https://github.com/xamarin/Xamarin.Forms/issues/10679

Xamarin.Forms.Platform.Android/Renderers/SwipeViewRenderer.cs
jsuarezruiz: As @samhouts mentioned, the issue was fixed by other PR. The current changes only add a new tests in Core Gallery.

PR #10730, opened 2020-05-16
Add FlyoutBackdrop Color for Shell

PureWeen:
### Description of Change ###

Add Flyout Backdrop Color to Shell Appearance structure so users can change the color of the overlay

The attached property only takes hold when applied to Shell and ShellItem

This also makes the behavior of closing the flyout when clicking off the flyout on iOS more reliable

### API Changes ###
Added:
 - Color ShellAppearance.FlyoutBackdropColor
 - Shell.FlyoutBackdropColor // Attached Property

 
 None

### Platforms Affected ### 
- iOS
- Android
- UWP 

#### Android
6F1D65D4-DD8E-4C3B-A7FD-116E42C857C9

#### iOS

On iOS there is a subtle change for the Default Behavior. The opacity is now animated by default whereas before the opacity was just set immediately when the progress > 0

97155BF9-B74D-4B95-9C1C-321B4F829FE1

#### UWP

image

### Testing Procedure ###

Play with the added BackdropColor button on the Store Shell

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Shell/Shell.cs
+++ b/Xamarin.Forms.Core/Shell/Shell.cs
@@ -268,7 +268,10 @@ namespace Xamarin.Forms
 		void IShellController.AddAppearanceObserver(IAppearanceObserver observer, Element pivot)
 		{
 			_appearanceObservers.Add((observer, pivot));
-			observer.OnAppearanceChanged(GetAppearanceForPivot(pivot));
+			var appearance = GetAppearanceForPivot(pivot);
+
+			if(appearance != null)
+				observer.OnAppearanceChanged(appearance);
 		}
 
 		void IShellController.AddFlyoutBehaviorObserver(IFlyoutBehaviorObserver observer)
--- a/Xamarin.Forms.Core/Shell/Shell.cs
+++ b/Xamarin.Forms.Core/Shell/Shell.cs
@@ -146,6 +146,10 @@ namespace Xamarin.Forms
 			BindableProperty.CreateAttached("UnselectedColor", typeof(Color), typeof(Shell), Color.Default,
 				propertyChanged: OnColorValueChanged);
 
+		public static readonly BindableProperty FlyoutBackdropColorProperty =
+			BindableProperty.CreateAttached("FlyoutBackdropColor", typeof(Color), typeof(Shell), Color.Default,
+				propertyChanged: OnColorValueChanged);
+
 		public static Color GetBackgroundColor(BindableObject obj) => (Color)obj.GetValue(BackgroundColorProperty);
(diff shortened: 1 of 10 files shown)


Xamarin.Forms.Core/Shell/ShellAppearance.cs
```diff
@@ -65,32 +68,25 @@ internal ShellAppearance()
...
-				   EqualityComparer<Color>.Default.Equals(TabBarBackgroundColor, appearance.TabBarBackgroundColor) &&
-				   EqualityComparer<Color>.Default.Equals(TabBarDisabledColor, appearance.TabBarDisabledColor) &&
-				   EqualityComparer<Color>.Default.Equals(TabBarForegroundColor, appearance.TabBarForegroundColor) &&
-				   EqualityComparer<Color>.Default.Equals(TabBarTitleColor, appearance.TabBarTitleColor) &&
-				   EqualityComparer<Color>.Default.Equals(TabBarUnselectedColor, appearance.TabBarUnselectedColor) &&
-				   EqualityComparer<Color>.Default.Equals(TitleColor, appearance.TitleColor) &&
-				   EqualityComparer<Color>.Default.Equals(UnselectedColor, appearance.UnselectedColor);
+
+			if (appearance == null)
+				return false;
```
pictos, line 73: Any reason to don't use this approach?

```csharp
if (!(obj is ShellAppearance))
	return false;
```

Xamarin.Forms.Core/Shell/Shell.cs
```diff
@@ -313,7 +323,10 @@ void IShellController.AppearanceChanged(Element source, bool appearanceSet)
 				{
 					if (leaf == target)
 					{
-						observer.OnAppearanceChanged(GetAppearanceForPivot(pivot));
+						var appearance = GetAppearanceForPivot(pivot);
+						if(appearance != null)
```
VladislavAntonyuk, line 327: Check for null in OnAppearanceChanged

PureWeen: where?

Xamarin.Forms.Core/Shell/ShellAppearance.cs
PureWeen: 👍

Xamarin.Forms.Core/Shell/Shell.cs
VladislavAntonyuk: ShellFlyoutRenderer.cs line 24

PureWeen: 👍

PR #10756, opened 2020-05-18
Added AutoClose property to SwipeView

### LINKED ISSUE #10709: [Enhancement] When using a SwipeView within a ListView,  an option to allow a swiped item to auto close when another item is being swiped would be useful.
## Summary
When using a SwipeView within a ListView,  an option to allow a swiped item to auto close when another item is being swiped would be useful. Therefore only 1 visible swipe menu per ListView.

## Intended Use Case
Only allow one row in a ListView to have an SwipeView menu open at any one time, so the behavior is similar to  "ViewCell.ContextActions"

jsuarezruiz:
### Description of Change ###

Added `AutoClose` property in SwipeView. If AutoClose is true, the SwipeView will automatically close when another one is opened.

### Issues Resolved ### 

- fixes #10709 

### API Changes ###
 
 Added **AutoClose** Property to SwipeView.

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android

None

swipeview-autoclose

### Testing Procedure ###
Launch Core Gallery and navigate to the SwipeView samples.

### PR Checklist ###


Xamarin.Forms.Core/SwipeView.cs
```diff
@@ -64,6 +69,12 @@ public SwipeItems BottomItems
 			set { SetValue(BottomItemsProperty, value); }
 		}
 
+		public bool AutoClose
```
jsuarezruiz, line 72: I have thought different ideas. This option is the simplest one, my question is, is it easy to understand?.

marioantoniou:
@jsuarezruiz ,  Thank you for quick response, much appreciated. This is exactly what we need and works perfectly on Android.   There are 2 minor issues on iOS that also happen with my own temporarily solution  using the latest forms version 4.6.0.726.

Issue 1: if you swipe  right using 2 fingers so you are swiping more than row at a time, a list item is left open without the swipe menu visible.  see row 6  in image IMG_0474

Issue 2:  With a  swipe menu open, say row 2, and you then scroll up, so that row 2 is out of the frame, then row 14  coming into the frame will have it's swipe menu open.   Assume this is to do with the Recycle Element caching strategy of the ListView.   

IMG_0474
)

kaminipatel03:
In which Xamarian.forms version this property available?

I have installed 4.7.0.813-pre2 version but this property do not come in swipview.

marioantoniou:
@kaminipatel03, I think it's still in development. @jsuarezruiz added the option in this pull request. I tested by Launching Core Gallery and navigated to the SwipeView samples.

jsuarezruiz:
@marioantoniou What you describe is due to reuse of cells. I will test it. What is your feedback about this new `AutoClose` property? Does it solve the case that you describe?
@kaminipatel03 We still have to validate everything, it is not available in public versions. As @marioantoniou mentioned, you can try it using this branch, etc.

marioantoniou:
@jsuarezruiz  The new AutoClose property is great, it solves the scenarios encountered.

MhAllan:
I think it is very confusing and not understood. AutoClose for me sounds like it will close after some time or when it loses focus. And the sentence "closes when another opens" doesn't define what is "another"? is it in same page? same layout? At the end, Is it the job for a framework like XF simplify simple things (because this can be solved in 2 lines of code)

marioantoniou:
This feature relates to a SwipeView when used within a  ListView ViewCell.   Currently, the swipe behaviour of each row is independent of other rows in the list.  This is unlike the expected behaviour of a native ListView c.f.  ViewCell.ContextActions  where only one row’s swipe menu is open at any given time.

MhAllan:
"Currently, the swipe behaviour of each row is independent of other rows in the list" And that is how it should be. An object shouldn't have logic that facilitates it's client responsibilities. An object should encapsulate it's own behaviour knowing nothing about the world. If there should be a property like AutoClose, It should reside in the container that contains items because then that is it's responsibility. but to put AutoClose in an object for the sake of desired client behaviour is wrong OOP.

Can I use SwipeView outside listing scenario? Yes, hence AutoClose is violating OOP if you put in this class.

marioantoniou:
OK, fair enough, but that’s a matter of implementation.  That does not negate the need for this functionality.

MhAllan:
Attached property may do. 

Thanks for your patience and replies.

Xamarin.Forms.Core/SwipeView.cs
```diff
@@ -9,6 +9,8 @@ namespace Xamarin.Forms
 	[RenderWith(typeof(_SwipeViewRenderer))]
 	public class SwipeView : ContentView, IElementConfiguration<SwipeView>, ISwipeViewController
 	{
+		public const string AutoCloseSwipeViewSignalName = "Xamarin.CloseSwipeView";
```
StephaneDelcroix, line 12: this should be hidden

jsuarezruiz: Changed!

jsuarezruiz:
The behavior we get with `AutoClose="True"` is the behavior of [SwipeContol](https://docs.microsoft.com/en-us/uwp/api/microsoft.ui.xaml.controls.swipecontrol?view=winui-2.4) in WinUI. Should we apply this behavior by default, without adding a new property?

marioantoniou:
I don't have a strong opinion on WinUI.

jfversluis:
This one is cool! 😱 sorry this won't make it into Forms anymore. Let's make sure we do it for .NET MAUI though!

PR #10662, opened 2020-05-12
[X] l10n of build errors

StephaneDelcroix:
### Description of Change ###

Provide error code, error links, and allow l10n of XamlC exceptions.

Localization of string resources using xliff (https://github.com/dotnet/xliff-tasks). Update translation sources using `msbuild /t:UpdateXlf`.

This is a reduced version of #8505 only focusing on compilation errors.

### Issues Resolved ### 

- fixes [AB#826363](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/826363)
- fixes [AB#1030395](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1030395)
- closes #7353
- closes #8505

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

### Behavioral/Visual Changes ###

None

### Before/After Screenshots ### 

Not applicable

### PR Checklist ###

- [ ] Targets the correct branch
- [ ] Tests are passing (or failures are unrelated)

PR #10762, opened 2020-05-18
[C] Catch typeconverter exceptions on bindings

### LINKED ISSUE #10405: [Bug] string.Empty causes InvalidOperationException($"Cannot convert \"{value}\" into {typeof(Color)}")
### Description
ColorTypeConverter throws InvalidOperationException on an empty string

### Steps to Reproduce

1. Add a Boxview to a page and set the color to `{Binding MyColor}`
2. Add a MyColor property to your viewmodel `public string MyColor { get;set; } = "";`
3. Run

### Expected Behavior
App should not crash

### Actual Behavior
App crashes with InvalidOperationException
https://github.com/xamarin/Xamarin.Forms/blob/6b827ac010d52203a0653a00b8b736c9a3bbcf1e/Xamarin.Forms.Core/ColorTypeConverter.cs#L281 

### Basic Information

- IDE: Rider
- Platform Target Frameworks: 
  - iOS:  
  - Android: 10.0
  - UWP:   
- Nuget Packages:
- Affected Devices:
Google Pixel 4XL
Samsung Galaxy S8

### Reproduction Link

https://github.com/ChaseFlorell/ColorConverterCrash

### Workaround
You need to attach a Converter to the binding and handle the empty string.

StephaneDelcroix:
### Description of Change ###

[C] Catch typeconverter exceptions on bindings

this doesn't apply to TypedBindings as you wouldn't be able to create a typed binding in this situation

### Issues Resolved ### 

- fixes #10405

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/BindingExpression.cs
+++ b/Xamarin.Forms.Core/BindingExpression.cs
@@ -148,7 +148,7 @@ namespace Xamarin.Forms
 
 				if (!TryConvert(ref value, property, property.ReturnType, true))
 				{
-					Log.Warning("Binding", "{0} can not be converted to type '{1}'", value, property.ReturnType);
+					Log.Warning("Binding", "'{0}' can not be converted to type '{1}'.", value, property.ReturnType);
 					return;
 				}
 
@@ -160,7 +160,7 @@ namespace Xamarin.Forms
 
 				if (!TryConvert(ref value, property, part.SetterType, false))
 				{
-					Log.Warning("Binding", "{0} can not be converted to type '{1}'", value, part.SetterType);
+					Log.Warning("Binding", "'{0}' can not be converted to type '{1}'.", value, part.SetterType);
 					return;
 				}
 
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Core/BindingExpression.cs
```diff
@@ -148,7 +148,7 @@ void ApplyCore(object sourceObject, BindableObject target, BindableProperty prop
 
 				if (!TryConvert(ref value, property, property.ReturnType, true))
 				{
-					Log.Warning("Binding", "{0} can not be converted to type '{1}'", value, property.ReturnType);
+					Log.Warning("Binding", "'{0}' can not be converted to type '{1}'.", value, property.ReturnType);
```
VladislavAntonyuk, line 151: Please use string interpolation

StephaneDelcroix: thanks for the suggestion, but we're in the process of moving the errors and warnings in a place where string.Format is better than string interpolation

ChaseFlorell: @VladislavAntonyuk @StephaneDelcroix personal opinion here.
If you leave the text at the source of the error, then when we as a user search for errors on the Google, we are taken to the source of the error. If you move the error elsewhere (say a Resource), then we have to do a second search to find the usage of the resource key.

just my $0.02

samhouts: Hmmm that's a good point. We have to move them because of localization requirements; however, it might be useful to leave the text in a comment here. It might get out of sync, though.

ChaseFlorell: @samhouts it will very likely get out of sync. Perhaps (and this would be a much larger task) Forms can start using unique error codes that are specified at the source of the error. That way we don't care as much about the text as we do the code.

StephaneDelcroix: we're in the process of doing just that https://github.com/xamarin/Xamarin.Forms/pull/10662 for build errors

helplinks is what comes next, hopefully

PR #10776, opened 2020-05-19
[UWP] Fix RadioButton disposing

yurkinh:
### Description of Change ###

Added Dispose method with events unsubscription 

### Issues Resolved ### 
Proper disposing of RadioButton control

### Platforms Affected ### 

- UWP

### Testing Procedure ###
Controls gallery

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/RadioButtonRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/RadioButtonRenderer.cs
@@ -9,6 +9,10 @@ namespace Xamarin.Forms.Platform.UWP
 	public class RadioButtonRenderer : ViewRenderer<RadioButton, FormsRadioButton>
 	{
 		bool _fontApplied;
+		FormsRadioButton button;
+		PointerEventHandler _pointerPressedHandler;
+
+		public bool IsDisposed { get; private set; }
 
 		protected override void OnElementChanged(ElementChangedEventArgs<RadioButton> e)
 		{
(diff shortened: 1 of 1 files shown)


hartez:
@yurkinh Do you have a test case which shows this renderer leaking memory because it doesn't unhook the event handlers? Can you describe a situation in which that would happen?

hartez:
Adding a DNM tag until we have some indication that this is necessary.

Xamarin.Forms.Platform.UAP/RadioButtonRenderer.cs
```diff
@@ -209,5 +213,25 @@ void UpdateCheck()
 
 			Control.IsChecked = Element.IsChecked;
 		}
+
+		protected override void Dispose(bool disposing)
+		{
+			if (IsDisposed)
+				return;
+			if (button != null)
```
hartez, line 221: This is missing a check for `disposing == true`.

Xamarin.Forms.Platform.UAP/RadioButtonRenderer.cs
```diff
@@ -9,6 +9,10 @@ namespace Xamarin.Forms.Platform.UWP
 	public class RadioButtonRenderer : ViewRenderer<RadioButton, FormsRadioButton>
 	{
 		bool _fontApplied;
+		FormsRadioButton button;
+		PointerEventHandler _pointerPressedHandler;
+
+		public bool IsDisposed { get; private set; }
```
hartez, line 15: No need for this to be public; a private `_disposed` member would be sufficient.

Xamarin.Forms.Platform.UAP/RadioButtonRenderer.cs
```diff
@@ -9,6 +9,10 @@ namespace Xamarin.Forms.Platform.UWP
 	public class RadioButtonRenderer : ViewRenderer<RadioButton, FormsRadioButton>
 	{
 		bool _fontApplied;
+		FormsRadioButton button;
```
hartez, line 12: No need to add another member; `Control` will be a reference to the Button if it exists.

Xamarin.Forms.Platform.UAP/RadioButtonRenderer.cs
hartez, line 221: `button` -> `Control`

samhouts:
Since we have an alternate RadioButton implementation in progress, I'll close this one for now. Thank you so much for the work you put into this! :)

PR #10793, opened 2020-05-20
[UWP] Fix Button disposing

yurkinh:
### Description of Change ###

Added Dispose method with events unsubscription 

### Issues Resolved ### 
Proper disposing of Button control

### Platforms Affected ### 

- UWP

### Testing Procedure ###
Controls gallery

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/ButtonRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/ButtonRenderer.cs
@@ -16,6 +16,9 @@ namespace Xamarin.Forms.Platform.UWP
 	{
 		bool _fontApplied;
 
+		FormsButton _button;
+		PointerEventHandler _pointerPressedHandler;		
+
 		protected override void OnElementChanged(ElementChangedEventArgs<Button> e)
 		{
 			base.OnElementChanged(e);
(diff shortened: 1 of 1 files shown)


hartez:
@yurkinh Why is this necessary? Do you have an example of a situation where not unhooking these events causes a memory leak or a performance issue?

yurkinh:
@hartez  I have not checked it with profiler yet only ControlsGallery whether it is disposed or not.
 Also I have noticed that other UWP renderers that have click events like CheckBox or RadioButton implement Dispose method with similar unsubscription logic.

Xamarin.Forms.Platform.UAP/ButtonRenderer.cs
```diff
@@ -290,5 +293,24 @@ void UpdatePadding()
 				Element.Padding.Bottom
 			);
 		}
+
+		bool _isDisposed;
+		protected override void Dispose(bool disposing)
+		{
+			if (_isDisposed)
+				return;
+			if (_button != null)
```
hartez, line 302: This is missing the check for `disposing == true`; if this is called by the finalizer, it could potentially crash in weird ways.

PR #10941, opened 2020-06-03
[Enhancement] Public StepperPositionProperty 

### LINKED ISSUE #10148: [Enhancement] Make StepperPositionProperty field public on Stepper
## Summary

We're currently in the process of adding support of Xamarin.Forms 4.5 to an open-source MVU library, [Fabulous.XamarinForms](https://github.com/fsprojects/Fabulous).
This library relies on the `XXXProperty` fields for various reasons (from automated bindings generation - much like native libs bindings - for supporting Xamarin.Forms, to clearing previously used properties)

I noticed that the `StepperPositionProperty` field is marked `private` whereas every other BindableProperty fields are usually marked `public`.

Any reason why?

The linked property has public getter/setter so it should be ok for the backing bindable field to be public.

This prevents us from correctly supporting this property.

## API Changes

Change the following line to be `public`.

https://github.com/xamarin/Xamarin.Forms/blob/2877eb35f673821bb1e8062a88720e88804aac9e/Xamarin.Forms.Core/Stepper.cs#L48

yurkinh:
### Description of Change ###

Made StepperPositionProperty field public for Stepper control

### Issues Resolved ### 
- fixes #10148

### API Changes ###

Changed:
private static readonly BindableProperty StepperPositionProperty => public static readonly BindableProperty StepperPositionProperty 

### Platforms Affected ### 

- Core/XAML (all platforms)

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Stepper.cs
+++ b/Xamarin.Forms.Core/Stepper.cs
@@ -45,7 +45,7 @@ namespace Xamarin.Forms
 
 		public static readonly BindableProperty IncrementProperty = BindableProperty.Create(nameof(Increment), typeof(double), typeof(Stepper), 1.0);
 
-		private static readonly BindableProperty StepperPositionProperty = BindableProperty.Create(nameof(StepperPosition), typeof(int), typeof(Stepper), 0);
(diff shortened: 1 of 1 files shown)


StephaneDelcroix:
this was wrong

StephaneDelcroix:
StepperPosition, and the BP, shouldn't have been made public. If one changes it, the computation goes south

PR #10979, opened 2020-06-07
[GH-10729] Added Current Page in Shell

### LINKED ISSUE #10729: [Enhancement] Provide an easy way to get the Current Page at Shell
## Summary

As a developer, I want to get the Current Page presented at the Shell.

## API Changes
Today to have access to Current Page, we need to do a boring code, like this:
```csharp
void OnShellNavigated(object sender, ShellNavigatedEventArgs e)
{
     var page = Shell.CurrentItem.CurrentItem as ShellSection;
     CurrentPage = ((IShellSectionController)page).PresentedPage;
}
```
For this, we just need to expose a Property. Something like:
```csharp
// Shell Side

public Page CurrentPage {get;}  // Read-only property


// using scenario
void OnShellNavigated(object sender, ShellNavigatedEventArgs e)
{
     CurrentPage = Shell.Current.CurrentPage;
}
```

## Intended Use Case
Whit this addition we can easily get the presented page and do some operations on it, for example when we have our NavigationServices implementations, like [this](https://github.com/pictos/ShellPresentation/blob/master/ShellPresentation/ShellPresentation/Services/NavigationService.cs)

pictos:
### Description of Change ###

This PR provides an easy way to access the PresentedPage in Shell applications.

### Issues Resolved ### 

- fixes #10729 

### API Changes ###

Added:
 - `Page Shell.CurrentPage { get; }`
 - `Shell : IPageContainer<Page>`

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###

Create an AppShell, and access the `Shell.Current.CurrentPage`, its value should be the presented page. Also, I added some unit tests.

### PR Checklist ###


StephaneDelcroix:
@PureWeen opinion on this ?

gresolio:
Definitely a very useful enhancement, it would be nice to have access to the PresentedPage and stop using workarounds like this: (Shell.Current?.CurrentItem?.CurrentItem as IShellSectionController)?.PresentedPage

Xamarin.Forms.Core/Shell/Shell.cs
```diff
@@ -1078,8 +1080,11 @@ internal void ProcessNavigated(ShellNavigatedEventArgs args)
 				{
 					baseShellItem.OnAppearing(() =>
 					{
-						OnNavigated(args);
-						Navigated?.Invoke(this, args);
+						Device.BeginInvokeOnMainThread(() =>
```
PureWeen, line 1083: @pictos What's the use case here?

pictos: @PureWeen, for some reason, when the application initializes this event is not trigged (the breakpoint hits the `Navigated?.Invoke(this,args)` the listeners didn't fire), with that the CurrentPage will be null at the first time. After that, the event is trigged normally. So to fix this I force those calls to run on the UI thread.
If my explanation is confusing, please let me know.

PureWeen: Boo! Alright, this is a great find but we'll want to figure out how to accomplish this same thing without using `Device.BeginInvokeOnMainThread`

Anything that requires `await Task.Delay(10)` or `Device.BeginInvokeOnMainThread` needs to be fixed so those aren't needed

pictos: @PureWeen I agree, well I will dig in into the code and see if I can find something.

pictos:
@PureWeen, something else fixed that issue, the code works without the `Device.BeginInvokeOnMainThread`, also all the core Shell tests are green.

Xamarin.Forms.Core.UnitTests/ShellModalTests.cs
```diff
@@ -345,6 +345,22 @@ public async Task NavigatingAndNavigatedFiresForShellModal()
...
+			shell.Items.Add(CreateShellItem(shellItemRoute: "NewRoute", shellSectionRoute: "Section", shellContentRoute: "Content"));
+
+			shell.Navigated += (_, __) =>
+			{
+				var page = shell.CurrentPage;
+				Assert.IsNotNull(page);
+				Assert.AreEqual(page.GetType(), typeof(ModalTestPage));
+			};
+
+			await shell.GoToAsync("ModalTestPage");
```
PureWeen, line 361: Can you move the asserts below the GoToAsync? If Navigated stops firing (for whatever reason) this test will still pass.  You can probably just store CurrentPage to a local variable and then move the asserts below gotoasync

Xamarin.Forms.Core.UnitTests/ShellTests.cs
```diff
@@ -1545,6 +1545,69 @@ public void ClearingShellSectionAndReAddingSetsCurrentItem()
...
+			Routing.RegisterRoute("cat", typeof(ContentPage));
+
+			shell.Navigated += (_, __) =>
+			{
+				var page = shell.CurrentPage;
+				Assert.IsNotNull(page);
+				Assert.AreEqual(page.GetType(), typeof(ContentPage));
+			};
+
+			await shell.GoToAsync("cat");
```
PureWeen, line 1564: Same here

Xamarin.Forms.Core.UnitTests/ShellTests.cs
```diff
@@ -1545,6 +1545,69 @@ public void ClearingShellSectionAndReAddingSetsCurrentItem()
...
+			shell.Items.Add(one);
+			shell.Items.Add(two);
+
+			shell.Navigated += (_, __) =>
+			{
+				var page = shell.CurrentPage;
+				Assert.IsNotNull(page);
+				Assert.AreEqual(page.GetType(), typeof(ShellTestPage));
+			};
+			shell.GoToAsync(new ShellNavigationState("//two/tabfour/"));
```
PureWeen, line 1594: same here

Xamarin.Forms.Core.UnitTests/ShellTests.cs
```diff
@@ -1545,6 +1545,69 @@ public void ClearingShellSectionAndReAddingSetsCurrentItem()
...
+		}
+
+		[Test]
+		public void GetCurrentPageOnInit()
+		{
+			var shell = new Shell();
+			shell.Navigated += (_, __) =>
+			{
+				var page = shell.CurrentPage;
+				Assert.IsNotNull(page);
```
PureWeen, line 1604: same here

Xamarin.Forms.Core/Shell/Shell.cs
```diff
@@ -13,8 +13,10 @@
 namespace Xamarin.Forms
 {
 	[ContentProperty(nameof(Items))]
-	public class Shell : Page, IShellController, IPropertyPropagationController
+	public class Shell : Page, IShellController, IPropertyPropagationController, IPageContainer<Page>
 	{
+		public Page CurrentPage => (CurrentSection as IShellSectionController).PresentedPage;
```
PureWeen, line 18:
```suggestion
		public Page CurrentPage => (CurrentSection as IShellSectionController)?.PresentedPage;
```

PureWeen:
Test run is here

https://dev.azure.com/xamarin/public/_releaseProgress?_a=release-pipeline-progress&releaseId=2176

PR #10980, opened 2020-06-07
BaseShellItem IsVisible and FlyoutItem.IsVisible

### LINKED ISSUE #5232: [Enhancement] Add FlyoutItem.IsVisible, Tab.IsVisible, BaseShellItem.IsVisible
## Summary

Provide attached properties so users can more directly define whether something is visible in the Flyout or in the TabBar.

## Future API Suggestions

We should add options so when users use `FlyoutDisplayOptions="AsMultipleItems"`so they can choose to display the parent as a Header. With the expansion control we could also add that as a setting as well.

# API

```C#
FlyoutItem.IsVisible //BP Attached Property
Tab.IsVisible //BP Attached Property
BaseShellItem.IsVisible // BP (not attached property)
```

## Examples BaseShellItem.IsVisible 

If you set BaseShellIitem IsVisible to false it won't show up in the Flyout or on any of the tabs

## Examples Tab.IsVisible

```xaml
<TabBar>
    <Tab Title="Tab 1">
          <ShellContent Title="Top Tab 1" />
          <ShellContent Title="Top Tab 2"  Tab.IsVisible="false"/>
          <ShellContent Title="Top Tab 3" />
    </Tab>
    <Tab Title="Tab 2" Tab.IsVisible="false">
    </Tab>
</TabBar>
```

## Examples FlyoutItem.IsVisible

Users wants a Flyout that doesn't reflect any of the tabs on the screen. 

```XAML
<MenuItem title="item 1 "/>
<MenuItem title="item 2" />
<MenuItem title="item 3" />
<TabBar Shell.FlyoutBehavior="Flyout" FlyoutItem.IsVisible="False">
    <Tab>
    </Tab>
</TabBar>

```
Renders 3 Flyout Items 
- Item 1
- Item 2
- Item 3
---

```XAML
<FlyoutItem title="item 1 " />
<FlyoutItem title="item 2" />
<ShellContent title="Content" FlyoutItem.IsVisible="False" />
```

Renders 2 Flyout Items even if you are on the ShellContent element
- Item 1
- Item 2

---

## Intended Use Case
The use case I have in mind is to have an item in the Hamburger menu for the user to upgrade the app to Premium. Once the user has upgraded the item should then no longer be visible.
Parent: #2415

### LINKED ISSUE #7230: [Enhancement] Shell allow dynamic load FlyoutItem!

PureWeen:
### Description of Change ###
- Add IsVisible property to BaseShellItem which will hide that particular shell element from all UI structures and the element will be treated as if it doesn't exist

- Add FlyoutItem.IsVisible which will just hide the item from the flyout menu

### Issues Resolved ### 
- partially fixes #5232

### API Changes ###
Added:

```C#
public class FlyoutItem
{
      public new bool IsVisible //BP
}
```

```C#
public class BaseShellItem
{
      public bool IsVisible //BP
}
```

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS
- Android
- UWP

### Testing Procedure ###
- Automated Tests included

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core.UnitTests/ShellTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/ShellTests.cs
@@ -1043,7 +1043,7 @@ namespace Xamarin.Forms.Core.UnitTests
 		}
 
 		[Test]
-		public async Task ShellItemNotVisible()
+		public async Task ShellItemNotVisibleWhenContentPageNotVisible()
 		{
 			var shell = new Shell();
 			var item1 = CreateShellItem(new ContentPage() { IsVisible = false });
@@ -1054,6 +1054,74 @@ namespace Xamarin.Forms.Core.UnitTests
 
 			Assert.IsFalse(GetItems(shell).Contains(item1));
 			Assert.IsTrue(GetItems(shell).Contains(item2));
+		}
+
+		[Test]
+		public async Task BaseShellItemNotVisible()
+		{
+			var shell = new Shell();
+			var item1 = CreateShellItem();
+			var item2 = CreateShellItem();
+
+			shell.Items.Add(item1);
+			shell.Items.Add(item2);
+
+			item1.IsVisible = false;
+			Assert.IsFalse(GetItems(shell).Contains(item1));
+			Assert.IsTrue(GetItems(shell).Contains(item2));
+
+			item1.IsVisible = true;
+			Assert.IsTrue(GetItems(shell).Contains(item1));
+
+			item1.Items[0].IsVisible = false;
+			Assert.IsFalse(GetItems(shell).Contains(item1));
+			item1.Items[0].IsVisible = true;
+			Assert.IsTrue(GetItems(shell).Contains(item1));
+
+			item1.Items[0].Items[0].IsVisible = false;
+			Assert.IsFalse(GetItems(shell).Contains(item1));
+			item1.Items[0].Items[0].IsVisible = true;
+			Assert.IsTrue(GetItems(shell).Contains(item1));
+		}
+
+		[Test]
+		public async Task CanStillNavigateToNotVisibleShellItem()
+		{
+			var shell = new Shell();
+			var item1 = CreateShellItem(shellItemRoute:"NotVisible");
+			var item2 = CreateShellItem();
+
+			shell.Items.Add(item1);
+			shell.Items.Add(item2);
+
+			item1.IsVisible = false;
+
+			shell.GoToAsync("//NotVisible");
+
+			Assert.AreEqual("//NotVisible", shell.CurrentState.Location.ToString());
+		}
+
+
+		[Test]
+		public async Task FlyoutItemVisible()
+		{
+			var shell = new Shell();
+			var item1 = CreateShellItem<FlyoutItem>(shellItemRoute: "NotVisible");
+			var item2 = CreateShellItem();
+
+			shell.Items.Add(item1);
+			shell.Items.Add(item2);
+
+			FlyoutItem.SetIsVisible(item1, false);
+			Assert.IsTrue(GetItems(shell).Contains(item1));
+
+			bool hasFlyoutItem =
+				(shell as IShellController)
+					.GenerateFlyoutGrouping()
+					.SelectMany(i => i)
+					.Contains(item1);
+
+			Assert.IsFalse(hasFlyoutItem);
 		}
 
 		[Test]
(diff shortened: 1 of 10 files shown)


Xamarin.Forms.Core.UnitTests/ShellTests.cs
```diff
@@ -1056,6 +1056,74 @@ public async Task ShellItemNotVisible()
...
+			var shell = new Shell();
+			var item1 = CreateShellItem(shellItemRoute:"NotVisible");
+			var item2 = CreateShellItem();
+
+			shell.Items.Add(item1);
+			shell.Items.Add(item2);
+
+			item1.IsVisible = false;
+
+			shell.GoToAsync("//NotVisible");
```
pictos, line 1099: we are missing the `await` keyword here (;
```suggestion
			await shell.GoToAsync("//NotVisible");
```

PureWeen: 👍

PureWeen:
test failures are fixed by this PR
https://github.com/xamarin/Xamarin.Forms/pull/11001

Xamarin.Forms.Core/Shell/ShellItem.cs
```diff
@@ -21,6 +21,20 @@ public FlyoutItem()
 		{
 
 		}
+
+		public static readonly new BindableProperty IsVisibleProperty =
```
StephaneDelcroix, line 25: in Shell code, we often reuse the same BP for attached and non-attached scenarios. I think this applies here too

StephaneDelcroix: overriding the BP will likely confuse users, and eventually tooling

PureWeen: The main reason for this is the ability to articulate different types of visibility. Maybe I need a new property?

For example someone could do


```XAML
<FlyoutItem />
<FlyoutItem FlyoutDisplayOptions="AsMultipleItems" >
      <TabBar FlyoutItem.IsVisible="false".  <!-- this means just don't show up in the flyout -->
      <TabBar IsVisible="false"  <!-- this means just don't show up anywhere -->
</FlyoutItem>
<FlyoutItem FlyoutDisplayOptions="AsMultipleItems" IsVisible="false">  <!-- this means just don't show up anywhere -->
</FlyoutItem>

```

So the idea was to use the different attached properties to articulate what the `IsVisible` is visually being applied to

Does it need to be an enum or a different BP?

PureWeen: @StephaneDelcroix are you aa bigger fan of having something like

`Shell.FlyoutitemIsVisible`

StephaneDelcroix: no, that's good. make sure the XAML is parsed as intended

wagenheimer:
I'm testing it and already have a bug to report!

If I make some change when the first view is loading on the OnAppearing Method, the Flyouyt Menu invisible items are only updated after I access some other view.

PureWeen:
@wagenheimer can you create a new issue with a repro please?

IgorKravchenko10:
> I'm testing it and already have a bug to report!
> 
> If I make some change when the first view is loading on the OnAppearing Method, the Flyouyt Menu invisible items are only updated after I access some other view.

> @wagenheimer can you create a new issue with a repro please?

I have same issue. I have created small repro. I change IsVisible property of tab "Browse2" when clicking Logout. 
Changing IsVisible doesn't work when:
1. Staying in About1.
2. Click Logout.
3. Nothing changed.

Changing IsVisible does work when:
1. Staing in About2.
2. Click Logout.
3. Browse2 will show/hide.

So, changing IsVisible working only when navigating to same FlyoutItem.
[ShellTest.zip](https://github.com/xamarin/Xamarin.Forms/files/5681426/ShellTest.zip)

PR #10992, opened 2020-06-09
[C] MultiBinding overhaul

### LINKED ISSUE #10750: [Enhancement] Support StringFormat on MultiBinding
## Summary
Add `StringFormat` support in `MultiBinding`.

## API Changes
No API changes, `StringFormat` is already declared in `BindingBase`.

## Intended Use Case
Example XAML usage:
````xaml
<Label>
  <Label.Text>
    <MultiBinding StringFormat="{}{0} -- Now only {1:C}!">
      <Binding Path="Description"/>
      <Binding Path="Price"/>
    </MultiBinding>
  </Label.Text>
</Label>
````
Documentation: [BindingBase.StringFormat](https://docs.microsoft.com/en-us/dotnet/api/system.windows.data.bindingbase.stringformat)

### LINKED ISSUE #10822: [Bug] Setting MultiBinding StringFormat results in InvalidOperationException
### Description

There's an [enhancement request](https://github.com/xamarin/Xamarin.Forms/issues/10750) to add `StringFormat` support to `MultiBinding`. However, I'm also logging a bug because attempting to use `StringFormat` right now results in an exception being thrown.

The following `StringFormat` should be valid, but results in an `InvalidOperationException`:

```
        <Label Grid.Row="1">
            <Label.Text>
                <MultiBinding StringFormat="{}{0} {1} {2}">
                    <Binding Path="Employee1.Forename" />
                    <Binding Path="Employee1.MiddleName" />
                    <Binding Path="Employee1.Surname" />
                </MultiBinding>
            </Label.Text>
        </Label>
```

While I don't think that `StringFormat` support is required for the 4.7 stable release, I do think that setting it should be handled more gracefully. We shouldn't require users to read the doc to say that `StringFormat` is currently unsupported.

### Steps to Reproduce

1. Run the attached sample.

### Expected Behavior

Lack of `StringFormat` support is handled gracefully.

### Actual Behavior

`InvalidOperationException` thrown with message "MultiBinding requires Converter be set".

### Basic Information

- Version with issue: 4.7-pre2
- Platform Target Frameworks: Xamarin.Forms.Core issue.

### Reproduction Link

[DataBindingDemos.zip](https://github.com/xamarin/Xamarin.Forms/files/4666924/DataBindingDemos.zip)

### PR DESCRIPTION
**Pull request by StephaneDelcroix**
### Description of Change ###

Various fixes, changes, and simplification on the MultiBinding code.
Also avoid throwing FormatException, but instead log a warning.

### Issues Resolved ### 

- fixes #10750
- fixes #10822

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

### CODE REVIEW

File: Xamarin.Forms.Core.UnitTests/MultiBindingTests.cs [hunk 1]
```diff
@@ -33,16 +33,8 @@ public override void TearDown()
...
-		public void TestChildOneWayOverridesMultiTwoWay()
+		public void TestChildOneWayOnMultiTwoWay()
 		{
-			// This tests a weird edge case where the MultiBinding is TwoWay but
-			// one of the child bindings is OneWay. This results in a situation where
-			// you actually can't change the target value to be inconsistent with the
-			// the source value for the OneWay because ConvertBack causes Convert
-			// to get re-evaluated (and this effectively negates the change to the
-			// target). This ensures that the target and source never go out of sync,
-			// as the overall binding is still TwoWay.
```
StephaneDelcroix, line 44: changed Multibinding to use target BP default mode in case of none specified. sub Bindings defaults to the same mode.

it's totally fine to have an intermediate model out of sync, it's better than forcing 1W bindings to behave 2W

Xamarin.Forms.Core.UnitTests/MultiBindingTests.cs
```diff
@@ -71,8 +63,7 @@ public void TestChildOneWayOverridesMultiTwoWay()
 			Assert.AreEqual(oldName, label.Text);
 			Assert.AreEqual(oldName, group.Person1.FullName);
 
-			label.Text = $"{oldFirstName.ToUpper()} {oldMiddleName} {oldLastName.ToUpper()}";
```
StephaneDelcroix, line 74: setting the label.Text directly is supposed to remove the MultiBinding. the fact that some changes were still happening after that indicates some references weren't cleared

Xamarin.Forms.Core.UnitTests/MultiBindingTests.cs
```diff
@@ -162,7 +149,8 @@ public void TestNestedMultiBindings()
 							$"{nameof(Element.BindingContext)}.{nameof(GroupViewModel.PardonAllSuspensions)}",
 							source: new RelativeBindingSource(RelativeBindingSourceMode.FindAncestor, typeof(StackLayout))),
 					},
-					Converter = new AnyTrueMultiConverter()
+					Converter = new AnyTrueMultiConverter(),
+					FallbackValue = false,
```
StephaneDelcroix, line 153: the converter sends UnsetValue for the purpose of getting the fallback, but no fallback was defined

Xamarin.Forms.Core.UnitTests/MultiBindingTests.cs
```diff
@@ -235,22 +223,22 @@ public void TestConverterReturnValues()
 
 			var label2 = GenerateNameLabel(nameof(group.Person2), BindingMode.TwoWay);
 			stack.Children.Add(label2);
-			label2.Text = $"DoNothing {oldMiddleName} {oldLastName.ToUpper()}";
+			label2.SetValueCore(Label.TextProperty, $"DoNothing {oldMiddleName} {oldLastName.ToUpper()}", Internals.SetValueFlags.None);
 			Assert.AreEqual($"{oldFirstName} {oldMiddleName} {oldLastName.ToUpper()}", group.Person2.FullName);
-			Assert.AreEqual(group.Person2.FullName, label2.Text);
+			Assert.AreEqual($"DoNothing {oldMiddleName} {oldLastName.ToUpper()}", label2.Text);
```
StephaneDelcroix, line 228: in case when some conversion use DoNothing or UnsetValue, do not expect your explicitly set Label.Text to be changed. that'd be weird.

changed (here and at some other places) tests to make sure the set value is the one on display, even if the converter prevents binding propagation

Xamarin.Forms.Core.UnitTests/MultiBindingTests.cs
```diff
@@ -259,67 +247,65 @@ public void TestConverterReturnValues()
...
-			Assert.AreEqual(3, converter.Converts);
-			Assert.AreEqual(0, converter.ConvertBacks);
-
-			// Target change results in 1 ConvertBack, one additional Convert
-			label.Text = oldName;
-			Assert.AreEqual(oldName, group.Person1.FullName);
-			Assert.AreEqual(4, converter.Converts);
-			Assert.AreEqual(1, converter.ConvertBacks);
-		}
+		//[Test]
```
StephaneDelcroix, line 262: disabling this for now, the new code do a few extra conversion

Xamarin.Forms.Core.UnitTests/MultiBindingTests.cs
```diff
@@ -369,6 +352,24 @@ public void TestBindingModes()
 			Assert.AreEqual(group.Person1.FullName, label1T.Text);
 		}
 
+		[Test]
+		public void TestStringFormat()
+		{
+			var property = BindableProperty.Create("foo", typeof(string), typeof(MockBindable), null);
```
StephaneDelcroix, line 358: StringFormat for all your simple use cases :)

Xamarin.Forms.Core.UnitTests/MultiBindingTests.cs
```diff
@@ -511,13 +512,13 @@ public object[] ConvertBack(object value, Type[] targetTypes, object parameter,
 				if (s == "null" || string.IsNullOrEmpty(s))
 					return null;
 
-				string concatenator = parameter as string ?? " ";
+				string separator = parameter as string ?? " ";
 
-				if (!targetTypes.All(t => t == typeof(string)))
+				if (!targetTypes.All(t=>t==typeof(object)) && !targetTypes.All(t => t == typeof(string)))
```
StephaneDelcroix, line 517: at first, the object types might be unknown (all objects)

Xamarin.Forms.Core.UnitTests/MultiBindingTests.cs
```diff
@@ -511,13 +512,13 @@ public object[] ConvertBack(object value, Type[] targetTypes, object parameter,
...
 					// Normally we'd return null but throw exception just for unit test to catch
 					throw new Exception("Invalid targetTypes");
```
StephaneDelcroix, line 519: never ever throw in a real world converter. those exception are not, and will never be caught and it'll crash your app. you've been warned

Xamarin.Forms.Core/BindingBase.cs
```diff
@@ -129,17 +123,39 @@ internal virtual object GetSourceValue(object value, Type targetPropertyType)
 			if (value == null && TargetNullValue != null)
 				return TargetNullValue;
 
-			if (StringFormat != null)
-				return string.Format(StringFormat, value);
+			if (StringFormat != null && TryFormat(StringFormat, value, out var formatted))
```
StephaneDelcroix, line 126: we don't want binding code to throw in this case. we want logging, and ideally BindingDiagnostic

Xamarin.Forms.Core/Internals/MultiBindingProxy.cs
```diff
@@ -1,72 +0,0 @@
-﻿namespace Xamarin.Forms.Internals
-{
-	// This class serves as a "proxy" for the target and the sources,
-	// with the MultiBinding serving as the bridge between them. 
-	// All the BindingExpression's actually bind between the proxies and the
-	// target/sources. This avoids the need for what would otherwise be some very 
-	// ugly subclassing or modifications to BindingExpression. And it
-	// makes it very easy to supported nested MultiBinding's which is 
-	// not possible with WPF.
-	internal class MultiBindingProxy : BindableObject
```
StephaneDelcroix, line 10: this is gone, without loosing any capability, replaced by a lightweight object

Xamarin.Forms.Core/MultiBinding.cs
```diff
@@ -70,238 +66,147 @@ internal override BindingBase Clone()
 
 		internal override void Apply(bool fromTarget)
 		{
-			VerifyConverterBeforeApply();
```
StephaneDelcroix, line 73: converter is no longer required now that we have StringFormat

Xamarin.Forms.Core/MultiBinding.cs
```diff
@@ -13,12 +10,12 @@ public sealed class MultiBinding : BindingBase
 	{
```
StephaneDelcroix, line 10: review this file using split mode. the diff doesn't make sense as most of the code have been changed

Xamarin.Forms.Core/MultiBinding.cs
```diff
@@ -70,238 +66,147 @@ internal override BindingBase Clone()
...
-
-			base.Unapply(fromBindingContextChanged: fromBindingContextChanged);
+			if (_bindings == null)
+				throw new InvalidOperationException("Bindings is null");
+
+			base.Apply(context, targetObject, targetProperty, fromBindingContextChanged);
+
+			if (!ReferenceEquals(_targetObject, targetObject)) {
+				_targetObject = targetObject;
+				_proxyObject = new ProxyElement() { Parent = targetObject as Element };
```
StephaneDelcroix, line 123: ProxyElement is our lightweight proxy. it's an element so it has a parent (for RelativeSource Bindings)

Xamarin.Forms.Core/MultiBinding.cs
```diff
@@ -13,12 +10,12 @@ public sealed class MultiBinding : BindingBase
 	{
 		IMultiValueConverter _converter;
 		object _converterParameter;
-		BindingExpression _expression;
 		IList<BindingBase> _bindings;
 		BindableProperty _targetProperty;
-		bool _isApplying;
-		bool _isCreating;
-		bool _hasSuccessfullyConverted;
+		BindableObject _targetObject;
```
StephaneDelcroix, line 15: this could be a WeakRef in case you hold a ref on the Binding, and not on the target

Xamarin.Forms.Core.UnitTests/MultiBindingTests.cs
legistek: Ah good catch. The proxy is still hanging around listening for the change on the target property and updating the source even after the target binding is cleared.

Xamarin.Forms.Core/MultiBinding.cs
```diff
@@ -13,12 +10,12 @@ public sealed class MultiBinding : BindingBase
 	{
 		IMultiValueConverter _converter;
 		object _converterParameter;
-		BindingExpression _expression;
 		IList<BindingBase> _bindings;
 		BindableProperty _targetProperty;
-		bool _isApplying;
-		bool _isCreating;
-		bool _hasSuccessfullyConverted;
```
legistek, line 21: FYI these were mostly to minimize redundant convert's and convertback's which I saw you removed the unit test for.

legistek:
Thanks for flagging this for me. I didn't look too too closely, but if it functions the same it's all good by me. My only concern would be if there are a lot of redundant convert's or convertback's now. The converters *should* be agile but you never know.

Xamarin.Forms.Core.UnitTests/MultiBindingTests.cs
legistek: Making the default mode match that of the BP sounds good but as for the second part of that, I was trying to make sure that if a multibinding is two-way but a child is one-way, the target's value can't become inconsistent with any of the sources of the one-way bindings. That feels like the right behavior and is consistent with WPF; not sure if your changes actually impacted that or not.

Xamarin.Forms.Core/MultiBinding.cs
StephaneDelcroix: yes, I plan to look at minimising the number of converts, maybe a bit later

PR #11035, opened 2020-06-12
Revert 8372 - Placeholder and Error Images

PureWeen:
### Description of Change ###

- ABI break https://github.com/xamarin/Xamarin.Forms/pull/8372/files#diff-e47ca366fb15a83fa4c70559cb9f0e54R12
- Placeholder and Error Image shouldn't be limited to FileImageSource. If users want to uses a slow loading URI image for these that's on them. Also limiting to FileImageSource removes the ability to use FontImageSource
- Personally I'm not convinced this should be on Image. If we give ImageSource these capabilities then all of this behavior gets adopted to all those controls. The code that loads an ImageSource is fairly contained. There are probably some nuances here I'm missing though that would make this tricky everywhere :-)  If we decide to just keep this on Image then we should also add to ImageButton possibly via the IImageElement interface

### API Changes ###

 Removed:
 - Image.ErrorPlaceholder and Image.LoadingPlaceholder
 
 
 None

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS
- Android
- UWP

### Testing Procedure ###
- make sure image loading code still works on all the platforms

### PR Checklist ###


pictos:
@PureWeen  I'm ok to reimplement this feature in a different way. But I have some points here.

> Placeholder and Error Image shouldn't be limited to FileImageSource. If users want to uses a slow loading URI image for these that's on them. Also limiting to FileImageSource removes the ability to use FontImageSource

If we allow the use of  `UriImageSource` and an error occurs due to a bad Internet connection, the placeholders will not appear. So I implemented this to receive only one `FileImageSource`. Even if this in mind do you think that we should accept  `UriImageSource`?

> Personally I'm not convinced this should be on Image. If we give ImageSource these capabilities then all of this behavior gets adopted to all those controls. The code that loads an ImageSource is fairly contained. There are probably some nuances here I'm missing though that would make this tricky everywhere :-) If we decide to just keep this on Image then we should also add to ImageButton possibly via the IImageElement interface

If the team still wants this feature, I can try to implement this inside the `ImageSource` class.

Also when this PR is merged we should reopen the #7015.

samhouts:
@PureWeen Please open a new spec for the Placeholder/Error functionality on ImageSource. Thanks!

@pictos Thank you so much for your work on this! This let us visualize and test out the functionality, and thus create a better spec for what this API should be. Once that new spec is up, we'd be happy to have you work on it again! AMAZING STUFF, @pictos !!!! You are very much appreciated.

samhouts:
Oh, and of course, we want your feedback on the new spec, as well. 👍

PR #11140, opened 2020-06-19
Fix 8503, 8787 - text in Entry not immediately visible, or visible after IsVisible set to true

### LINKED ISSUE #8503: [Bug] Text is not visible on Entry element after animation finished

### LINKED ISSUE #8787: [Bug] Entry text initially invisible on UWP

### LINKED ISSUE #10137: Cases where in UWP an Entry does not display existing data until the field is tapped. Works in Droid[Bug]
### Description
If an Entry is contained within a ScrollView, in UWP, the second and subsequent ones to not display their content until they get focus. It works in Droid and iOS.
### Steps to Reproduce

1. create a basic Xamarin Forms cross platform Solution (include Droid, iOS, UWP
2. use this xaml as the Content:
`    <StackLayout Orientation="Vertical" BackgroundColor="White">
        <ScrollView>
            <StackLayout  Orientation="Vertical" BackgroundColor="Black">

                        <Label FontSize="20"   TextColor="Black" Text="Description" 
                            HorizontalOptions="Start"   HeightRequest="28"  VerticalOptions="Center"/>

                        <Entry  BackgroundColor="White" 
                            FontSize="18" 
                            VerticalOptions="Center"
                            Text="Sam"/>
                </StackLayout>

                    <Label FontSize="20"   TextColor="Black" Text="Description" 
                        HorizontalOptions="Start"   HeightRequest="28"  VerticalOptions="Center"/>

                    <Entry  BackgroundColor="White" 
                        FontSize="18" 
                        VerticalOptions="Center"
                        Text="Fred"/>
                </StackLayout>

            </StackLayout>
        </ScrollView>
    </StackLayout>`

3. run in Droid and in UWP

### Expected Behavior
Both Sam and Fred appear immediately in all platforms.

### Actual Behavior
Sam and Fred appear immediately in Droid and iOS. 
In UWP, Sam (the content of the first Entry) appears but Fred only appears when it received focus.
However, if the ScrollView is removed, then everything works in all platforms
### Basic Information

- Version with issue: XF 4.5.0.495
- IDE:VS 2019 16.4.5
- Platform Target Frameworks: 
  - iOS:      n/a
  - Android:  n/a
  - UWP:      18362
- Nuget Packages:
Microsoft.NETCore.UniversalWindowsPlatform 6.2.9
NETStandard.Library 2.0.3
Xamarin.Android.Support.Design 28.0.0.3
Xamarin.Android.Support.v4  28.0.0.3
Xamarin.Android.Support.v7.AppCompat 28.0.0.3
Xamarin.Android.Support.v7.CardView 28.0.0.3
Xamarin.Android.Support.v7.MediaRouter 28.0.0.3
Xamarin.Forms 4.5.0.495

- Affected Devices:

### Workaround
Do not include Entry within a ScrollView or remove the ScrollView.

bmacombe:
### Description of Change ###

This PR is a re-submission of #8536 by @WayaFlyfeather with a minor change to fix the conflict it was reverted for.

Adding a ContentPresenter to the ScrollViewer in FormsTextBoxStyle, as setting VerticalAlignment on the ScrollViewer led to unexpected results.  This caused problems with merged PR #8214 which fixed the incorrect TextBox hight calculation #2172.

The solution here is to copy the FormsTextBoxStyle and make a FormsTextBoxStyleMeasure, which is used only when measuring the text box size and excludes the newly added ContentPresenter.

I was not able to find a way to dynamically remove the new ContentPresenter at runtime while using the existing FormsTextBoxSyle, so I had to create that second template.

### Issues Resolved ### 

- fixes #8503
- fixes #8787
- likely fixes #10137 (not tested)
- resolves conflict with #8214

### API Changes ###
None

### Platforms Affected ### 
- UWP

### Behavioral/Visual Changes
Before this fix, if an Entry was changed from **IsVisible** = false to **IsVisible** true, text inside the entry would not actually become visible before it acquired focus.

[Edit January 30, 2020]: Other cases has popped up where entry text is seemingly randomly invisible at first. I have today confirmed that this fix also fixes issue #8787.

### Before/After Screenshots
**Issue 8503**
Before 1 - before clicking button to make Entry visible
fix8503-before-1

Before 2 - after clicking button to make Entry visible. Not, the text is not visible.
fix8503-before-2
 
Before 3 - after clicking in the Entry, giving it focus - now showing text
fix8503-before-3

After 1 - before clicking button to make Entry visible
fix8503-after-1
 
After 2 - after clicking button to make entry visible. Note that the text is now visible
fix8503-after-2
 
**Issue 8787**
Before fix 1 - Page is loaded, text in entry invisible:
fix-8787-8503-before-1
 
Before fix 2 - after clicking in Entry to give it focus, text becomes visible;
fix-8787-8503-before-2
 
After fix - page is loaded, text in Entry is immediately visible:
fix-8787-8503-after
 
### Testing Procedure

**Issue 2172**
Open the issue page and view the entries

**Issue 8503**
Go to test page for Issue 8503 in the ControlGallery.
Click the button _Click to show Entry for Isolated bug_.
Check that the entry appears, and it's text content is visible.
If not, check if it appears when the entry receives focus. If it does, the bug is not fixed.

A similar test is repated below this, this one mimics the original problem described in #8503, it should similarly show the text at once when the button is clicked.

**Issue 8787**
Go to test page for Issue 8787 in the ControlGallery.
Check that the text in the yellow Entry is immediately visible.
If it is, the issue is fixed. If it is not visible before the Entry gains focus or the window is resized, the issue is not fixed.

### Notes
This one is a bit weird.

I isolated the problem described in #8503, to occurring when an Entry started out as invisible, but was later changed to Visible. In that case it's text content would not be shown before it got focus. (The animation used in the original report changed visibility of hte item, which was the cause - it wasn't due to behaviors or animations).

I then determined that the "culprit" was this line introduced in #6463:
https://github.com/andreinitescu/Xamarin.Forms/blob/3481a842b45c463c1c0ff40c051e3c2484907c85/Xamarin.Forms.Platform.UAP/FormsTextBoxStyle.xaml#L224

If this VerticalAlignment was not present on the ScrollViewer, then everything worked fine. It also worked fine if I gave the Entry a HeightRequest, I discovered. But without a HeightRequest, the problem was there, even if changed the VerticalAlignment to "Center", i.e not binding it.

I could on Stack Overflow see that [others have had problems with strange behavior from a ScrollViewer in a TextBox with VerticalAlignment](https://stackoverflow.com/a/47909054).

Hence this proposed solution, instead og setting VerticalAlignment directly on the ScrollViewer, to have it wrap a ContentPresenter, and set VerticalContentAlignment on that instead.

It seems to be working fine, but to be perfectly honest I'm not 100% sure if I may have overlooked some potential problem with this solution.

### PR Checklist ###


bmacombe:
@hartez and @WayaFlyfeather a review would be great if you can

bmacombe:
This two template method is my favorite and I'd love to still track down the root issue.  But this appears to at least bandaid all the problems.

I somewhat wonder if everything in the UWP backend being wrapped in a panel has something to do with it, causing some timing issue when controls are added to the visual tree...I'm just wildely guessing here.  I'd hope that with the new slim renderer in MAUI that maybe all those panels will go away and it might help.

bmacombe:
@samhouts Any chance this could be targeted for a 4.6 service pack?

bmacombe:
Let me know if I need to clean up this branch, I just noticed these 3 commits below.  Wonder if I messed something up rebasing on master when I pulled in the original PR branch.  Feels like they shouldn't be there.

image

bmacombe:
I decided it was messed up, so I created a new branch off master, applied the actual commits that should be in this PR, retested and pushed.  Hopefully, this will be easier to review now!

hartez:
@bmacombe So which were the relevant line changes between FormsTextBoxStyle and FormsTextBoxMeasureStyle?

Xamarin.Forms.Platform.UAP/FormsTextBoxStyle.xaml
```diff
@@ -211,17 +211,17 @@
...
 						              HorizontalScrollBarVisibility="{TemplateBinding ScrollViewer.HorizontalScrollBarVisibility}"
 						              IsTabStop="False" IsHorizontalRailEnabled="{TemplateBinding ScrollViewer.IsHorizontalRailEnabled}"
 						              IsVerticalRailEnabled="{TemplateBinding ScrollViewer.IsVerticalRailEnabled}"
 						              IsDeferredScrollingEnabled="{TemplateBinding ScrollViewer.IsDeferredScrollingEnabled}"
 						              Margin="{TemplateBinding BorderThickness}" Padding="{TemplateBinding Padding}" Grid.Row="1"
 						              VerticalScrollBarVisibility="{TemplateBinding ScrollViewer.VerticalScrollBarVisibility}"
-						              VerticalScrollMode="{TemplateBinding ScrollViewer.VerticalScrollMode}" ZoomMode="Disabled"
-                                      VerticalAlignment="{TemplateBinding VerticalContentAlignment}" />
+						              VerticalScrollMode="{TemplateBinding ScrollViewer.VerticalScrollMode}" ZoomMode="Disabled">
+                            <ContentPresenter x:Name="ContentElement" VerticalContentAlignment="{TemplateBinding VerticalContentAlignment}" />
```
bmacombe, line 223: @hartez This content presenter is the only difference between the two.  That is what was added to fix the text not showing right away but caused the height measure to then work incorrectly.

bmacombe: that line and which element is labeled "ContentElement" is the only difference.

I was hoping to be able to pull out the ContentPresentor in OnTemplateApply in the FormsTextBox to avoid the extra template, but I don't think the template is created by the time it's needed to measure.

Also a note on the ScrollViewer as the Content Element, I made a blank UWP project and pulled up the control template, the ScrollViewer is in there as the Content Element in a stock template.

hartez: Okay, I see it. I'll try to test and review tomorrow. 

So with this change, does anything still use the other control template (FormsTextBoxStyle)?

bmacombe: yes, that is used by all the displayed text boxes...the "Measure" one is only used by the off visual tree textbox used for measuring

Xamarin.Forms.Platform.UAP/FormsTextBox.cs
```diff
@@ -461,7 +461,7 @@ public static Size GetCopyOfSize(FormsTextBox control, Windows.Foundation.Size c
 			{
 				_copyOfTextBox = new FormsTextBox
 				{
-					Style = Windows.UI.Xaml.Application.Current.Resources["FormsTextBoxStyle"] as Windows.UI.Xaml.Style
+					Style = Windows.UI.Xaml.Application.Current.Resources["FormsTextBoxStyleMeasure"] as Windows.UI.Xaml.Style
```
bmacombe, line 464: @hartez Here is the only place the Measure template is used, in that single instance used calc the size off the visual tree.  Which was the fix for #2172

hartez:
>  I'll try to test and review tomorrow.

Well, that comment didn't age well :(

If not tomorrow, then this weekend.

hartez:
This definitely fixes the problem.

> I was hoping to be able to pull out the ContentPresentor in OnTemplateApply in the FormsTextBox to avoid the extra template, but I don't think the template is created by the time it's needed to measure.

I'm still wondering if there's a way to get to the fix via that path, and simplify this. But let's get this fix in place for now and maybe we can revisit that later.

bmacombe:
Yes I would like that also...I'm worried if future changes are made to the template they might not get applied to both.

Two issues I see with the single template that I'm not sure how to fix

1 I had all the code added to try grab a reference in OnApplyTemplate to the presenter so I could remove it from measure textbox at runtime.  But it was always null, so I put breakpoints in and OnApplyTemplate wasn't being called before the get size method.  I tired a couple things to force it to apply without luck.

2 Assuming we get 1 figured out, I'm not sure if you can add an element name at runtime to a template.  Since removing the presenter would remove the ContentElment.  That name would need to be applied to the ScrollViewer I think.

I'm not the most experienced with UWP control templates, so maybe if anyone knows someone who could chime in here that would be great.

But I do agree with merging for now, this has been a headache for a couple months for UWP users.

Thanks for the review @hartez!

bmacombe:
@hartez I just want to make sure you saw this Win UI issue and that comment in particular and some of the discussion below it at least seems to add some context to the issues we are trying to address.

https://github.com/microsoft/microsoft-ui-xaml/issues/1226#issuecomment-526033765

It would be awesome with MAUI if the UWP/WinUI team could help with the backend!  Seems that folks with so much knowledge of the platform could really help improve the implementation.  As much as this and #8214 "fix" the problem, they feel like hacks to me and I really don't like it...but I really don't like all the UWP TextBox issues either, so I'm good with them for now!

ConnorJBrown:
It looks like this fix causes an issue where you cannot see what you are typing if it is longer than the entry. It used to scroll to the end as you are typing. See the following gif:

bmacombe:
I wonder if that's from the content presenter being added, maybe it needs to be set to expand or fill.

bmacombe:
So I've been playing more. So far here is where I'm at.

I can get the height, text visibility and text alignment to work, but not the keep scrolling as you type
or
I can get the height, text visibility and keep scrolling as you type, but not text alignment

and from what I find searching around, text alignment on a UWP textbox appears to be a common issue.
https://www.google.com/search?q=uwp+textbox+text+alignment&oq=uwp+textbox+text

The change introduced in this PR fixed the alignment but is what broke the text visibility #6463

I'm going to see if there is another way to address text alignment without the template binding maybe

bmacombe:
I've tried the following so far

Wrapping the ScrollViewer in another control, Panel, ContentPresenter, etc...still get the text visibility issue

Set the VerticalContentAlignment on the ScrollViewer instead of the VerticalAlignment, visibility fixed, text alignment broken

Just in case anyone else has an idea to try.

JayMansel:
Trying ourselves here too, will let you know if we figure it out

bmacombe:
Everything I'm finding on updating the vertical alignment of text in UWP says to override the template and set what you want the alignment to be.  I'm not finding any examples of how to allow it to be set dynamically as we would need here.

@jaymansel Thanks for helping!

bmacombe:
Little bit of progress.

Pulled the scrollviewer out of the template in FormsTextBox OnApplyTemplate
image

Then I hooked up to the Loaded event and then set the vertical alignment on the scroll viewer
image


#8503 and #2172 are good and so is the entry vertical alignment gallary pages.

but #8787 doesn't pass

bmacombe:
Manually calling focus after changing the vertical alignment seems to fix all cases...but I don't like it...anyone have any thoughts on any problems it might cause?

image

bmacombe:
@JayMansel Take a look at this PR again if you have time.

I've removed the "measure" style and gone back to a single style for FormsTextBox.  I removed the ContentPresenter that was causing the text entry issue you discovered and the VerticalAlignment bound to the template on the ScrollViewer.

Instead, after the text box has been loaded and added to the visual tree I then set the VerticalAlignment on the scrollviewer in FormsTextBox.cs

image

I had to add that Focus call to ensure the text was displayed.

But I noticed checking the page of issue #8222 that the edtior broke with this code, so I added a property to control if the OnLoaded method is used so the editor renderer can disable that behavior.

So now #2172, #8503, #8787, #8222 and the Entry control gallery appears to all work as expected.  I didn't notice any behavior change by calling Focus in the OnLoaded event handler.

The only remaining issue I'll address is a bit, is I'm not sure right now if someone changed the VerticalAlignment of the Entry, it would be updated in the FormsTextBox.  So I'll look into that and add support if needed later today.

This again seems hacky, but it's all I got right now!  Feedback or suggestions appreciated!

bmacombe:
Added support for updating the vertical text alignment after the control is loaded.  Updated issue #2172 page to include a test for switching between all alignments.

Xamarin.Forms.Platform.UAP/FormsTextBox.cs
```diff
@@ -167,6 +172,30 @@ protected override void OnApplyTemplate()
 				if (_DeleteButtonVisibleStateGroups != null)
 					_DeleteButtonVisibleState = _DeleteButtonVisibleStateGroups.States.SingleOrDefault(s => s.Name == "ButtonVisible");
 			}
+
+			_scrollViewer= (Windows.UI.Xaml.Controls.ScrollViewer)GetTemplateChild("ContentElement");
```
StephaneDelcroix, line 176: us `as` for casting, to prevent exception if the template is modified

bmacombe: I'll submit a PR to fix that and the other casts in OnApplyTemplate.  I also noticed that we no longer need to grab a reference to the border element, 1 because it's no longer reference except to set it and phone is no longer supported on UWP

bmacombe: #11285 Submitted

bmacombe:
@rmarinho Thx for fixing that control gallery error, I just noticed it failed and was going to go in and fix it and was pleasantly surprised to find it fixed!  I always forget you have to put that pre compiler statement in there!

PR #11162, opened 2020-06-22
[WPF] Pass through AutomationProperties from Xamarin Forms

### LINKED ISSUE #11161: [Bug] WPF does not set AutomationProperties so the experience is poor when using Narrator
### Steps to Reproduce

1. Create or run an existing Xamarin Forms application which uses AutomationProperties to set name, help text etc on controls
2. Notice that none of this accessibility info is used and you get only the bare minimum Narrator support
3. Repeat with UWP to hear the difference with the same Xamarin Forms code/XAML

### Expected Behavior
Native Controls should use the Name/HelpText set in the Xamarin Forms app. Behaviour should be as near as possible to consistent across platforms.

### Actual Behavior
On WPF because there is no implementation the Narrator experience is barely useable.

### Basic Information

- Version with issue: All
- IDE: n/a
- Platform Target Frameworks:
  - WPF:  All
- Nuget Packages: All
- Affected Devices: All

### Screenshots

n/a

### Reproduction Link

https://github.com/xamarin/xamarin-forms-samples/tree/master/UserInterface/Accessibility
(Needs a WPF head project adding)

### Workaround

n/a

peterfoot:
### Description of Change ### 

Currently the WPF implementation of Xamarin Forms has no support for screen readers - properties set using AutomationProperties are not set on the native controls so the experience is very poor. This pull requests implements the core AutomationProperties in the base ViewRenderer to provide a minimal implementation. It should be noted that this is not a complete and perfect implementation as there will be other issues specific to certain types of controls. It is noted however that UWP also has some of these other issues still to be addressed such as screen reader behaviour of list view items. The aim should be for both UWP and WPF implementations to fully support the Narrator functionality built into Windows.

### Issues Resolved ### 

- fixes #11161 

### API Changes ###
 
 None

### Platforms Affected ### 

- WPF

When using Narrator the app will now use the AutomationProperties set in your code/XAML. No visual changes.

Not applicable

### Testing Procedure ###

The sample (https://github.com/xamarin/xamarin-forms-samples/tree/master/UserInterface/Accessibility) can be used and the behavior compared with the UWP head project.

### PR Checklist ###


rmarinho:
The basic seems to work, thanks, this is a good start.

PR #11172, opened 2020-06-23
[iOS] Call page appearing when restoring app from background

### LINKED ISSUE #4986: [iOS] OnAppearing not firing when returning to app
### Description

Manual test for 52318 does not match the test instructions. Specifically, the test says "Only the last page should alert" when the app is reactivated after hitting the Home button. But currently _no alert_ displays when reactivating the app.

We need to determine if this is a regression or simply incorrect test instructions.

Original bug: https://bugzilla.xamarin.com/show_bug.cgi?id=52318

### Steps to Reproduce

1. Open the Control Gallery.
2. Navigate to issue 52318
3. Follow the test instructions.

### Expected Behavior

Instructions match the behavior of the app.

### Actual Behavior

The app displays different behavior than the instructions.

### Basic Information

- Version with issue: 3.5.0
- Last known good version: 2.3.5, assuming the instructions are correct.

rmarinho:
### Description of Change ###

Handle calling appearing and disappearing of pages when the Application is going to background and restored.

### Issues Resolved ### 

- fixes #4986 

### API Changes ###

```
internal class PageLifecycleManager : IDisposable
{
        public PageLifecycleManager(IPageController pageController)
	protected virtual void Dispose(bool disposing)

	public void Dispose()

	public void HandlePageAppearing()

	public void HandlePageDisappearing()
}
```
### Platforms Affected ### 

- iOS

When the app is restored the current page should have the Appearing event being called, like on Android.

Not applicable

### Testing Procedure ###

use `new NavigationPage(new ContentPage52318())`in the MainPage

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/PageLifecycleManager.cs
+++ b/Xamarin.Forms.Platform.iOS/PageLifecycleManager.cs
@@ -1,8 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using CoreTelephony;
 using Foundation;
 using UIKit;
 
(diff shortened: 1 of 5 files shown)


daltonks:
I typically use this logic to not have appearing/disappearing events on Android:
```
On<Android>()
    .SendDisappearingEventOnPause(false)
    .SendAppearingEventOnResume(false);
```

Is there an option to opt out of this on iOS too?

Tommigun1980:
What is the use of these events as there's no action points on a per page level when the app is suspended and resumed? It just takes a lot of work to get rid of these unnecessary calls. OnAppearing and OnDisappearing are the closest thing to lifecycle events and I can't see them being used for anything else than as a pseudo constructor and destructor. The app being suspended and resumed has no bearing on individual pages, and for generic sleep/resume code there's OnSleep and OnResume anyway.

Tommigun1980:
Also, isn't this a breaking change as it completely changes the event system (and for the worse imho)?

Tommigun1980:
I get the feel this change was made to conform to incorrect documentation, instead of serving a real purpose.

daltonks:
Well put. This will break a lot of apps, especially since it's not marked as a breaking change in the release notes.

Tommigun1980:
Edit: I now see that this was done in order to make iOS work the same way as Android, which of course is a good thing. May I suggest that instead of enabling this behavior on iOS you’d remove it from Android? By doing so both platforms would work the same way, and because OnAppearing/OnDisappearing being called on sleep/resume are highly problematic, that’d solve an issue instead of creating one.

Thanks.

yarram-manoj:
I have updated to 4.8 version and it is causing an issue due to this fix-4986, I have reverted to 4.7 for now. Can we expect something in the future releases which includes something similar to Android where we can turn that off? 
Like this.
On<Android>()
    .SendDisappearingEventOnPause(false)
    .SendAppearingEventOnResume(false);

Cliffy786:
Has this issue been resolved as yet?

Serena247Pro:
Updated to 4.8.0.1687, still didn't hit the break point in OnAppearing(). I click the home button on IOS simulator and then return to the app after a while. Is it coming in the next release?

PR #11187, opened 2020-06-24
Add Rect struct

jsuarezruiz:
### Description of Change ###

Allow to create a Rect struct (X, Y, Height, Width) from XAML. 

### API Changes ###

Added:

```
struct Rect
{
     public double X { get; set; }

     public double Y { get; set; }

     public double Width { get; set; }

     public double Height { get; set; }
}
```

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###
Launch Core Gallery, navigate to the issues and search the "Rect Test" sample. If the 

### PR Checklist ###


Xamarin.Forms.Core/Rect.cs
```diff
@@ -0,0 +1,239 @@
+﻿using System;
+using System.Diagnostics;
+using System.Globalization;
+
+namespace Xamarin.Forms
+{
+	[DebuggerDisplay("X={X}, Y={Y}, Width={Width}, Height={Height}")]
+	[TypeConverter(typeof(RectTypeConverter))]
+	public struct Rect
```
jsuarezruiz, line 9: Should we mark the Rectangle struct as deprecated?

IoTFier: Hi @jsuarezruiz  - marking Rectangle struct as deprecated while releasing this new struct is as good as renaming existing
Rectangle struct.

Introduction of new set of APIs such as Shapes - should rather come in its own namespace or it should have distinct type names that won't create any ambiguity with existing ones such as this Rectangle struct.

I sincerely request Xamarin Forms team to release any new or (especially or  ambiguous) breaking API changes only in the pre1 of the new XF version. That will give community, OSS vendors and solution providers couple of cycles to find such issues before that XF version is released for GA.

This time, Graph API was introduced in XF 4.7 pre4 and within couple days XF 4.7 GA was released. 

Thank you for your help and support.

IoTFier:
> I would retarget this change, potentially breaking, to 4.8.0

@StephaneDelcroix - does this mean, I should hold off from upgrading my app to XF 4.7 ?
Again - the name "Rect" is too odd and ambiguous. Any thoughts on having a separate name space for Shapes that we can always refer in the XAML ?

Xamarin.Forms.Core/Rect.cs
jsuarezruiz: Agree, should be something to do in 4.8.0.

IoTFier: What is there are other breaking changes in my App due to XF 4.7 ?
Shall I discover those in 4.8? Disagree with this strategy.

Please consider fixing this in XF 4.7 service release - if needed refactor for good in 4.8.
One more time - Let Shapes API be in its own namespace. That can resolve many other (if at all) issues related to types defined in Shapes that are conflicting with types from Xamarin.Form.Core .

jsuarezruiz:
Fixed the build.

PR #11216, opened 2020-06-26
Enable macOS slider rotation

JunielKatarn:
### Description of Change ###

Allow rotating a slider control in macOS to match other platforms' behavior.

### API Changes ###
 
 None

### Platforms Affected ### 

- macOS

Setting the `Rotation` property on a `Slider` element effectively rotates the element visually on macOS.

(Setting Rotation to 270)

Before:
image

After:
image

### Testing Procedure ###

Manual testing through the ControlGallery app.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.MacOS/Renderers/SliderRenderer.cs
+++ b/Xamarin.Forms.Platform.MacOS/Renderers/SliderRenderer.cs
@@ -113,6 +113,7 @@ namespace Xamarin.Forms.Platform.MacOS
 				UpdateMinimumTrackColor();
 				UpdateMaximumTrackColor();
 				UpdateThumbColor();
+				UpdateRotation();
 			}
 
 			base.OnElementChanged(e);
@@ -136,6 +137,8 @@ namespace Xamarin.Forms.Platform.MacOS
 			{
 				UpdateThumbColor();
 			}
+			else if (e.PropertyName == VisualElement.RotationProperty.PropertyName)
+				UpdateRotation();
 		}
 
 		void UpdateMaximumTrackColor()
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.MacOS/Renderers/SliderRenderer.cs
```diff
@@ -212,5 +215,10 @@ void UpdateValue()
 			if (Math.Abs(Element.Value - Control.DoubleValue) > 0)
 				Control.DoubleValue = (float)Element.Value;
 		}
+
+		void UpdateRotation()
+		{
+			Control.BoundsRotation = (float)Element.Rotation;
```
jsuarezruiz, line 221: Could we review/move this transformation in **VisualElementTracker**?. `VisualElementTracker` manage all the view transformations (rotation, scale, etc.).

JunielKatarn: Please note, this is my first PR.
I'll look into this.

jsuarezruiz: @JunielKatarn Of course, and we really appreciate your effort. For that reason, let me know if we can help you with something.

JunielKatarn: @jsuarezruiz I tried using `VisualElementTracker` for this purpose, but ran into issues.
image

The problem is, the `BoundsRotation` property seems to be needed especially for `NSSlider`.
If I set that property to any NSView (within VisualElementTracker), rendering becomes incorrect as shown in the attached screenshot.

To get the right behavior, we want to set `BoundsRotation` only on the `NSSlider` control.
Thus, it would seem like the appropriate place for this is still the `SliderRenderer` class.

Note, I have not been able to achieve any rotation by setting the transform in VisualElementTracker for the macOS Slider. Again, `BoundsRotation` seems mandatory.

jfversluis:
Unfortunately with the latest code the slider disappears completely :( Let me know if you still want to look into this and add it, else make sure to look at it for .NET MAUI if you're still interested.

Thanks for your time and effort on this nonetheless! <3

PR #11219, opened 2020-06-26
#10842 - ios - collectionview - invalid sizes when using a datatemplateselector

### LINKED ISSUE #10842: [Bug] [iOS] CollectionView - Invalid Height Measure when using different datatemplates
### Description
When we define two datatemplates with a different height in a datatemplateSelector and use that selector for the collectionview, then some heights of the cells are invalid.
- Check the gif atttached.
<br/>
More related: #8583 #9520
Possibly Related:
#5455
#7788
#10625

### Steps to Reproduce

1. use different datatemplates, where both templates differ in the heightrequest!
2. write a datatemplateselector
3. add different templates to the collection

### Expected Behavior
- the height of the already added items should not change
- the new added item shouldn use the height of the cell located above

### Actual Behavior
- It seems like that the new item added gets the height of the item located above it.
- All items located below the new added item get the same invalid height like the new added one
- this visual layout issue is only visible for a second

### Basic Information

- Version with issue: 4.4 - master
- Last known good version: -
- IDE: VS 2019
- Platform Target Frameworks: 
  - iOS:  -
- Nuget Packages: none
- Affected Devices: general ios issue

### Screenshots
ios-collectionView1

### Reproduction Link
Just create a new ContentPage and copy and paste the content into the view and the c# code into the code behind.

View:
```xml
             xmlns:d="http://xamarin.com/schemas/2014/forms/design"
             xmlns:collectionViewGalleries="clr-namespace:Xamarin.Forms.Controls.GalleryPages.CollectionViewGalleries;assembly=Xamarin.Forms.Controls"
             mc:Ignorable="d"
             x:Class="Xamarin.Forms.Controls.GalleryPages.CollectionViewGalleries.VaryingDataTempleteSelectorGallery">
    <ContentPage.Resources>
        <ResourceDictionary>
            <DataTemplate x:Key="CarTemplate">
                <Frame BorderColor="Red" BackgroundColor="Yellow">
                    <StackLayout HeightRequest="100">
                        <Label Text="{Binding Name}" />
                    </StackLayout>
                </Frame>
            </DataTemplate>

            <DataTemplate x:Key="PlaneTemplate">
                <Frame BorderColor="Red" BackgroundColor="Aqua">
                    <StackLayout HeightRequest="50">
                        <Label Text="{Binding Name}" />
                    </StackLayout>
                </Frame>
            </DataTemplate>

            <collectionViewGalleries:VehicleTemplateSelector x:Key="VehicleTemplateSelector"
                                                             CareTemplate="{StaticResource CarTemplate}"
                                                             PlaneTemplate="{StaticResource PlaneTemplate}"/>
        </ResourceDictionary>
    </ContentPage.Resources>

    <ContentPage.Content>
        <Grid>
            <Grid.RowDefinitions>
                <RowDefinition Height="*"/>
                <RowDefinition Height="Auto"/>
            </Grid.RowDefinitions>

            <CollectionView ItemsSource="{Binding Items}" 
                            ItemTemplate="{StaticResource VehicleTemplateSelector}"
                            VerticalOptions="FillAndExpand">

            </CollectionView>

            <StackLayout Grid.Row="1">
                    <Label Text="Index"/>
                    <Entry Text="{Binding Index}"/>
                </StackLayout>
                <Button Text="InsertCar" Clicked="InsertCar_OnClicked"/>
                <Button Text="InsertPlane" Clicked="InsertPlane_OnClicked"/>
            </StackLayout>
        </Grid>
    </ContentPage.Content>
</ContentPage>
```

the code behind:

```cs
public partial class VaryingDataTempleteSelectorGallery : ContentPage, INotifyPropertyChanged
	{
		string _index;

		public VaryingDataTempleteSelectorGallery()
		{
			InitializeComponent();
			BindingContext = this;

			foreach (var vehicle in CreateDefaultVehicles())
			{
				Items.Add(vehicle);
			}

			IEnumerable<VehicleBase> CreateDefaultVehicles()
			{
				yield return new Plane("Plane1");
				yield return new Plane("Plane2");
				yield return new Car("Car1");
				yield return new Plane("Plane3");
				yield return new Car("Car2");
				yield return new Plane("Plane4");
			}
		}

		public ObservableCollection<VehicleBase> Items { get; set; } =
			new ObservableCollection<VehicleBase>();

		public string Index
		{
			get => _index;
			set => SetValue(ref _index, value);
		}

		void InsertCar_OnClicked(object sender, EventArgs e)
		{
			if (string.IsNullOrWhiteSpace(Index)) return;
			if (!int.TryParse(Index, out var index)) return;
			if (index > Items.Count || index < 0) return;

			Items.Insert(index, new Car("rCar " + new Random().Next(0,1000).ToString()));
		}

		void InsertPlane_OnClicked(object sender, EventArgs e)
		{
			if (string.IsNullOrWhiteSpace(Index)) return;
			if (!int.TryParse(Index, out var index)) return;
			if (index > Items.Count || index < 0) return;

			Items.Insert(index, new Plane("rPlane " + new Random().Next(0,1000).ToString()));
		}

		void SetValue<T>(ref T backingField, in T value, [CallerMemberName] string callerName = null)
		{
			if (Equals(backingField, value)) return;
			OnPropertyChanging(callerName);
			backingField = value;
			OnPropertyChanged(callerName);
		}
	}

	class VehicleTemplateSelector : DataTemplateSelector
	{
		public DataTemplate CareTemplate { get; set; }
		public DataTemplate PlaneTemplate { get; set; }
		protected override DataTemplate OnSelectTemplate(object item, BindableObject container)
		{
			if (CareTemplate == null || PlaneTemplate == null) throw new ArgumentNullException();

			if (item is Car car) return CareTemplate;
			if (item is Plane plane) return PlaneTemplate;

			throw new ArgumentOutOfRangeException();
		}
	}

	public abstract class VehicleBase
	{
		protected VehicleBase(string name) => Name = name;

	}

	class Car : VehicleBase
	{
		public Car(string name) : base(name) { }
	}

	class Plane : VehicleBase
	{
		public Plane(string name) : base(name) { }
	}
```

### Workaround
none

BrayanKhosravian:
Description of Change
Fix sizing issue when we use a `DatatemplateSelector` where the templates have a different size

Issues Resolved
fixes #10842 
API Changes
None

Platforms Affected
iOS

Behavioral/Visual Changes
Items in a collectionview dont resize everytime when adding/inserting new items

Before
82812253-de499700-9e92-11ea-8da1-c80ca12c69aa

After
85296063-1bf50c00-b4a1-11ea-9efb-c9c8fba8525c

Testing Procedure
Launch Core Gallery and search for `CollectionView Gallery`
Then select `DataTemplate Galery` and finally select `VaryingFataTemplateSelector`

PR Checklist
 Targets the correct branch
 Tests are passing (or failures are unrelated)


Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewDelegator.cs
```diff
@@ -10,15 +11,62 @@ public class ItemsViewDelegator<TItemsView, TViewController> : UICollectionViewD
...
+
+		IDictionary<Type, CGSize> _sizeChache = new Dictionary<Type,CGSize>();
 		public ItemsViewLayout ItemsViewLayout { get; }
 		public TViewController ViewController { get; }
+		public IItemsViewSource ItemsSource { get; }
 
 		protected float PreviousHorizontalOffset, PreviousVerticalOffset;
 
-		public ItemsViewDelegator(ItemsViewLayout itemsViewLayout, TViewController itemsViewController)
+		public ItemsViewDelegator(ItemsViewLayout itemsViewLayout, TViewController itemsViewController, IItemsViewSource itemsViewSource)
```
BrayanKhosravian, line 22: I didnt find a better way to pass the `IITemsVIewSource` to the `ItemsVIewDelegator`
Maybe there are better solutions which can be applied after a review

BrayanKhosravian: i Know that there is the `UICollectionViewDataSource` which can be get/set by calling `collectionView.Source`
on a `UICollectionViewInstance` but this could change the whole xf architecture.

i tried to implement this fix/feature as defensive as possible because i do not want break anything.

Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewDelegator.cs
```diff
@@ -10,15 +11,62 @@ public class ItemsViewDelegator<TItemsView, TViewController> : UICollectionViewD
...
+		public override CGSize GetSizeForItem(UICollectionView collectionView, UICollectionViewLayout layout, NSIndexPath indexPath)
+		{
+			if (ItemsViewLayout.ItemSizingStrategy == ItemSizingStrategy.MeasureFirstItem)
+				return ItemsViewLayout.EstimatedItemSize;
+
+			var vm = ItemsSource[indexPath];
+
+			// try to get a cached size
+			if (_sizeChache.ContainsKey(vm.GetType()))
+				return _sizeChache[vm.GetType()];
```
BrayanKhosravian, line 46: This can still be improved by using a `IDictionary<IndexPath, CGSize>` as cache.

And whenever the item changes in size we would have to update the value at the given index path.

however, there is still a catch: the method which check for a layout invalidation is not located in the delegate.
it is located in either the `ItemsViewController` or in the `ItemsViewLayout` (i cant rememebr right now and im rly in hurry. will maybe update this later or work on it)

This means that the _cache has to be accessible in that class as well.
But then this will be a perfectly working solution.

Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewDelegator.cs
BrayanKhosravian: if we would assign a UICollectionViewDataSource to the ItemsViewCell then i would not have to pass the `IItemsViewSource` to the ctor of this class.
i could just call `collectionView.Source` directly in the method `ItemsViewDelegator.GetSizeForItem`

BrayanKhosravian:
this does only work for templates which have a fixed size!
could we tag this with `do not merge`?

BrayanKhosravian:
closing this PR as the issue was resolved by @hartez

PR #11228, opened 2020-06-27
[GH-11218] - Fixed crash when a user removes a page in Shell

### LINKED ISSUE #11218: [Bug] The Shell RemovePage function causes crash on iOS
### Description
Removing pages from the navigation stack using RemovePage crashes on iOS

### Steps to Reproduce

1. GoToAsync() to a page a few times
2. The code below removes all the pages in the navigation stack. Except homepage because that one is not in this list. 

``` 
            var existingPages = Shell.Current.Navigation.NavigationStack.ToList();
            foreach (var page in existingPages)
            {
                if (page != null)
                {
                    Shell.Current.Navigation.RemovePage(page);
                }
            }
```

3. iOS TargetInvocationException when navigating or null reference exception when trying to go back.

### Expected Behavior
Navigation stack cleaned and you can go one page back to the home page. Any pages in between are removed.
### Actual Behavior
iOS crashes

### Basic Information

- Version with issue: 4,7 
  - iOS: 13

### Reproduction files
[RemovePageReproduction.zip](https://github.com/xamarin/Xamarin.Forms/files/4837051/RemovePageReproduction.zip)

pictos:
### Description of Change ###

I removed the `OnPopRequested(e);` call inside the `Xamarin.Forms.Platform.iOS/Renderers/ShellSectionRenderer.OnRemoveRequested(NavigationRequestedEventArgs e)` method.

Also, I added a null check, for the `element`, inside the `WillShowViewController` method

### Issues Resolved ### 

- fixes #11218 

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS

Now the app doesn't crash when the user tries to Remove pages.

https://youtu.be/z95Y5XhknyE

### PR Checklist ###


pictos:
cc @PureWeen

pictos:
Which version of NUnity do I need to install to run the UITests? I forgot 😅

ymebrugts:
Great thanks!

rmarinho:
> Which version of NUnity do I need to install to run the UITests? I forgot 😅

NUNit3

ymebrugts:
Can this be merged into the next pre-release? I need this particular functionality for a high profile application.

PureWeen:
I need to play with this a bit more

The way I'd like to fix this is to better orchestrate navigation events at the xplat level

The sample application is modifying the Navigation stack inside the constructor so the flow is currently

Push Page  (navigation is starting) => New page is constructed and inside the constructor the navigation stack is modified => current page pushes  (navigation finishes)

This PR works but you can see the first page flash really quickly as it's navigating to the second page. What I'd prefer here is to detect that we are in the middle of a navigation which will then delay the remove page calls from propagating to the platform layer until the push has completed

This will miss the next SR but should be able to get this one into the one after (4.8.0-sr3)

pictos:
@PureWeen when I worked on this, I remember that the navigation behavior of Shell is different from the `NavigationPage` behavior, the `RemovePage` is trigged before the `Push`. Even if we call the `RemovePage` on the OnAppearing this code will fail. Maybe we can do something to prevent the _Navigation Methods_ (push, pop, remove) to run in parallel.

If you try to reproduce the sample in the issue using NavigationPage, you can see the difference.

> What I'd prefer here is to detect that we are in the middle of a navigation which will then delay the remove page calls from propagating to the platform layer until the push has completed

I will try to see if I can archive this behavior

PureWeen:
> Even if we call the RemovePage on the OnAppearing this code will fail.
Yea I found that out as well :-/ 

> Maybe we can do something to prevent the Navigation Methods (push, pop, remove) to run in parallel.

That's what I'm thinking as well. Need to play with this a bit more to see the best way to achieve this. In the linked issue the RemovePage is called in the middle of the GotoAsync call executing so I'd like to queue the removepage call until after the GotoAsync finishes. 

Not sure offhand how involved this type of queue will take to implement but that's my current thinking

programmation:
@PureWeen I would wholeheartedly support taking a dependency on Dataflow (System.Threading.Tasks.Dataflow) for this and many other useful `Task`-related work.

Xamarin.Forms.Platform.iOS/Renderers/ShellSectionRenderer.cs
```diff
@@ -415,13 +415,8 @@ protected virtual void OnRemoveRequested(NavigationRequestedEventArgs e)
 
 			if (viewController != null)
 			{
-				if (viewController == TopViewController)
-				{
-					e.Animated = false;
-					OnPopRequested(e);
-				}
-
```
PureWeen, line 423: can we just add an if statement here?

```C#
if(ViewControllers.Contains(viewController))
```

marksarcade:
How close is this one to being finished off?

PureWeen:
https://github.com/xamarin/Xamarin.Forms/issues/11218#issuecomment-723288810

Xamarin.Forms.Platform.iOS/Renderers/ShellSectionRenderer.cs
```diff
@@ -81,22 +81,22 @@ public ShellSectionRenderer(IShellContext context)
 			SendPop();
 
 		internal bool SendPop()
-		{ 
+		{
```
pictos, line 84: @PureWeen If I'm not wrong the changes are just _code style_ fixes from VS.

jfversluis:
Original issue is closed and this seemd to need some work, so I think it's safe to assume that this won't happen anymore. Thanks for all the time and effort here everyone!

PR #11231, opened 2020-06-28
Add shell renderer to page instead of window

PureWeen:
### Description of Change ###

This adds the ability to create a new activity and host content there for cases where the renderer needs an activity to be hosted inside

### Testing Procedure ###
- the tests don't schmear 

### PR Checklist ###


hartez:
@PureWeen still smearing.
Screenshot_1593447945

PR #11230, opened 2020-06-27
Replacing master with main

nickrandolph:
### Description of Change ###
Fixing a bunch of broken links/references due to renaming master branch to main

### Issues Resolved ### 
Link to Contributing docs was broken (ie https://github.com/xamarin/Xamarin.Forms/blob/master/.github/CONTRIBUTING.md doesn't resolve).

### API Changes ###
None

### Platforms Affected ### 
N/A

### Behavioral/Visual Changes ###
None

### Before/After Screenshots ### 
Not applicable

### Testing Procedure ###
Look for anything that might still be broken due to renaming master to main

### PR Checklist ###

- [x] Targets the correct branch
- [ ] Tests are passing (or failures are unrelated)

PR #11240, opened 2020-06-29
Fix some links with main branch rename

StephaneDelcroix:
### Description of Change ###

Was #11230, with added fixes

### Issues Resolved ### 

- fixes #

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###


Xamarin.Forms.Core/Crc64.cs
```diff
@@ -6,7 +6,7 @@
 using System.Text;
 
 /// <summary>
-/// https://github.com/xamarin/java.interop/blob/master/src/Java.Interop.Tools.JavaCallableWrappers/Java.Interop.Tools.JavaCallableWrappers/Crc64.cs
+/// https://github.com/xamarin/java.interop/blob/main/src/Java.Interop.Tools.JavaCallableWrappers/Java.Interop.Tools.JavaCallableWrappers/Crc64.cs
```
StephaneDelcroix, line 9:
```suggestion
/// https://github.com/xamarin/java.interop/blob/master/src/Java.Interop.Tools.JavaCallableWrappers/Java.Interop.Tools.JavaCallableWrappers/Crc64.cs
```

StephaneDelcroix:
instead of targeting 4.6, we should probably re-add the `master` branch for the tests to pass.

we should probably also review all of our links pointing to `main` or `master` and make them point to a release or commit, to avoid breaking tests when/if we remove the resource

PR #11243, opened 2020-06-29
Hide tabbar if only one tab is present

PureWeen:
### Description of Change ###

When the logic for determining tab visibility was moved into the xplat code it wasn't correctly returning false if there was only one tab present

### Platforms Affected ### 
- iOS

### Testing Procedure ###
- ui test Issue11107 passes

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/ShellItemRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ShellItemRenderer.cs
@@ -419,14 +419,7 @@ namespace Xamarin.Forms.Platform.Android
 			if (DisplayedPage == null)
 				return;
 
-			bool visible = ShellItemController.ShowTabs;
-			using (var menu = _bottomView.Menu)
-			{
-				if (menu.Size() == 1)
-					visible = false;
-			}
-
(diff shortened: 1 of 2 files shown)


Xamarin.Forms.Platform.Android/Renderers/ShellItemRenderer.cs
```diff
@@ -419,14 +419,7 @@ void UpdateTabBarVisibility()
 			if (DisplayedPage == null)
 				return;
 
-			bool visible = ShellItemController.ShowTabs;
```
samhouts, line 422: You say you're fixing a bug for iOS, so what is this code change doing here?

PureWeen: The point of `ShowTabs` was to move all the tab display logic up into the xplat code so we don't have to duplicate it across the renderers. I noticed this code on the android side when verifying that it matches up with iOS

It's currently a noop since the logic is taken care of by `showtabs` but since it's duplicated logic this change avoids 5 months from now when I modify the behavior of `showtabs` but android for some reason works differently :-)

rmarinho:
Failing tests not related, and UITest for this feature now passes on iOS

PR #11254, opened 2020-07-01
Revert [Core] Added TextTransform property #3690

samhouts:
### Description of Change ###

I suspect this is causing the span gestures tests to fail, but as I can't get them to fail locally, I will try it on CI.

### PR Checklist ###


rmarinho:
Nop they still fail @samhouts

Issue #10520, opened 2020-04-30
[Bug] TapGestureRecognizer not working with Spans on iOS when text in the span is multiline


abdelbena:
### Description
Having a Label with a span that has multi-line text with TapGestureRecognizers not working on iOS.
### Steps to Reproduce

1. Add the next Label to MainPage.xaml:
```
<Label WidthRequest="100"
    VerticalTextAlignment="Center"
    HorizontalTextAlignment="Center">
    <Label.FormattedText>
        <FormattedString>
            <Span Text="Long text about how we respect your privacy but you should actually check  Long text about how we respect your privacy but you should actually check " TextColor="Blue">
                <Span.GestureRecognizers>
                    <TapGestureRecognizer Tapped="Tap_EventHandler" />
                </Span.GestureRecognizers>
            </Span>
        </FormattedString>
    </Label.FormattedText>
</Label>
```
2. Add EventHandler Tap_EventHandler

### Expected Behavior
When the user tap on the span the action described in the EventHandler should happen.
### Actual Behavior
When the user tap on the span the action described in the EventHandler never happen.
### Basic Information

- Version with issue: 4.6.0.726

hartez:
Also not working on Android.

hartez:
[_10520 Repro.zip](https://github.com/xamarin/Xamarin.Forms/files/4560470/_10520.Repro.zip)

PR #11321, opened 2020-07-06
GestureRecognizers don't work if the Span is on the last line of a Label in iOS

adamped:
### Description of Change ###

The height of the Label seems to be off by 1 pixel, most likely due to a rounding error further down in the XF framework. This seems to affect iOS only, Android does not have the problem. Untested in UWP.

As such, when `layoutManager.BoundingRectForGlyphRange` is called for anything on the last line, it will return 0, because the glyph is outside the bounds of the textContainer. By adding the 1 extra pixel of height, it brings the glyph inside the bounds and correctly returns the CharacterBounds (CGRect). Once it has the character bounds, the region is correctly determined and GestureRecognizers will work.

If it fails to determine the CharacterBounds, a region is 0 and the GestureRecognizer will not work on the Span because there is no region for the gesture to work in.

### Issues Resolved ### 

No known issues currently raised for this.

### API Changes ###

 None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###

Create a multi line Label where a span is on the last line of the label. Add a TapGestureRecognizer to that last span.

It will not currently fire, because the Region is 0.

With the proposed patch, the region is now populated, and the GestureRecognizer will work.

### PR Checklist ###


StephaneDelcroix:
look who's back !

adamped:
Just passing through. Experienced this issue myself, thought it would help the community. :) 

Also, just scanned the open issues list, this one looks similar to this issue, but I didn't experience it on Android. Not sure if they are the same: https://github.com/xamarin/Xamarin.Forms/issues/10520

adamped:
Found more issues locally based on scaling, and updated the patch to expand the layout container (rather than Control.Frame) by the screen density, rather than just 1, as we found an issue where it didn't work with adding 1 pixel more, but worked when I added 2, with the screen density being 2.

adamped:
The other option here with the height, is just doing the height as nfloat.MaxValue since we wouldn't care about the total height when just measuring the text.

jsuarezruiz (approved):

adamped:
@jsuarezruiz - actually can I put a hold on this PR. During further testing, I found another case where it didn't pick up the last line again. My answer was to change the RoundingCorrection value to just nfloat.MaxValue because we honestly don't care about the height of the label when measuring, only the width. It's most likely I'm misunderstanding the true root cause of why the height is slightly off depending upon device, but considering the option of just ignoring any height constraint, I probably don't need to.

adamped:
@jsuarezruiz - I just made the change, if you don't see a problem with it, using the nfloat.MaxValue works in all scenarios, screen sizes etc.

adamped:
While this fix works I believe I found the root cause.

Doing this `layoutManager.UsesFontLeading = false;` manages to fix a lot of alignment issues. But I haven't run any tests to see if this is font specific or not and that was also with my existing fix still implemented.

I ran a test with an overlay and it finally matched everything up pixel perfect. I'll just close this PR since it's more of a band-aid rather than a fix, and something with the UsesFontLeading should be implemented instead.

PR #11375, opened 2020-07-09
[iOS] Fix controls added to Frame multiple times when container is recycled on ListView

### LINKED ISSUE #11266: Overlapping labels
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/content/problem/1090550/overlapping-labels.html)._

---
<p>I have updated to the latest version of xamarin form (4.7.0.968) and all the labels are overlapping on iOS.</p> <p><a href="https://developercommunityapi.westus.cloudapp.azure.com/storage/attachments/143201-simulator-screen-shot-iphone-11-pro-max-2020-06-24.png"> Image:143201-simulator-screen-shot-iphone-11-pro-max-2020-06-24.png</a></p><p></p>

---
#### Visual Studio Feedback System on 6/27/2020, 10:51 PM: 

<p>We have directed your feedback to the appropriate engineering team for further evaluation. The team will review the feedback and notify you about the next steps.</p>

---
### Original Solutions
(no solutions)

### Reproduction
[Archive.zip](https://github.com/xamarin/Xamarin.Forms/files/4893291/Archive.zip)

PureWeen:
### Description of Change ###

These changes let us keep the same actualview without instantiating it everytime

- When calling AddSubView on the framerenderer just forward those calls to ActualView
- When calling removefromparentview on ActualView just forward those to the views added to ActualView

Instead of forwarding the Gesture Recognizers just set UserInteractionEnabled to false on the inserted middle view. 
I ran tests with this to make sure content inside the frame still worked and APIs like InputTranparent still function correctly. Issue10422.xaml.cs has a PanGestureRecognizer you can test with

### Issues Resolved ### 

- fixes #11266

### Platforms Affected ### 
- iOS

### Testing Procedure ###
- Test all the issues here https://github.com/xamarin/Xamarin.Forms/pull/10469 make sure nothing regressed. One of the tests in there doesn't clip quite right but that's fixed by a PR @jsuarezruiz did on 4.7 

- test nuget against following sample https://github.com/xamarin/Xamarin.Forms/files/4893291/Archive.zip and make sure the frames update correctly with updated bindings

- Verify tapping things inside a frame and using gesture recognizers with a frame still all works

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/FrameRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/FrameRenderer.cs
@@ -128,7 +128,7 @@ namespace Xamarin.Forms.Platform.iOS
 
 		public override void Draw(CGRect rect)
 		{
-			if (_actualView == null)
+			if (_actualView != null)
 				_actualView.Frame = Bounds;
 
 			base.Draw(rect);
--- a/Xamarin.Forms.Platform.iOS/Renderers/FrameRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/FrameRenderer.cs
@@ -7,7 +7,7 @@ namespace Xamarin.Forms.Platform.iOS
 {
 	public class FrameRenderer : VisualElementRenderer<Frame>, ITabStop
 	{
-		UIView _actualView = new UIView();
+		UIView _actualView;
 		CGSize _previousSize;
 
 		UIView ITabStop.TabStop => this;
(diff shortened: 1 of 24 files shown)


Xamarin.Forms.Platform.iOS/Renderers/FrameRenderer.cs
```diff
@@ -23,24 +29,29 @@ protected override void OnElementChanged(ElementChangedEventArgs<Frame> e)
...
-						_actualView.AddGestureRecognizer(gesture);
-				}
-				else if (_actualView.Subviews.Length == 0)
-				{
-					_actualView.UserInteractionEnabled = false;
+					_actualView = new UIView();
+					// Add the subviews to the actual view.
+					foreach (var item in NativeView.Subviews)
+					{
+						_actualView.AddSubview(item);
```
PureWeen, line 38: @jfversluis 

When this is first hit nothing happens because the subviews are added by the VisualElementPackager

https://github.com/xamarin/Xamarin.Forms/blob/719fc7a604ff0cce8922d717c99bfb0fa17e35e0/Xamarin.Forms.Platform.iOS/VisualElementPackager.cs#L104

I don't think adding _actualView has really ever done anything TBH because subviews will always be empty on the first pass and then the VisualElementPackager will add it directly to the renderer later on. So, it appears that it does something but the subview is never actually added _actualView

jfversluis: So the thing I _tried_ to do is basically have a bottom view which does the shadow and the top view that holds the actual content and is clipped etc. This so that the shadow isn't clipped when stuff like that happens.

The tricky thing is that some properties have to be only applied to the top view, like the clipping, but some to both, like translation.

I wasn't really aware of the packager thing. Like you say; it appeared to work. I got inspiration from PancakeView which also does it largely like this: https://github.com/sthewissen/Xamarin.Forms.PancakeView/blob/master/src/Xamarin.Forms.PancakeView.Multi/Platforms/iOS/PancakeViewRenderer.cs

I'll pull it down and do some quick tests on this

PureWeen:
@jfversluis test again please!!

PureWeen:
Adding DNM for now

I feel like there's a more efficient way to handle the renderer recycling. 

Right now if you use this with a listview/collectionview it does a lot of remove/adding of views and news up a uiview with each element changed

ndastur:
@PureWeen I suspect there is a more efficient way to handle this, however, this fix is related to a serious bug on iOS that causes labels inside a frame to overlap on scrolling. So maybe a different feature to improve efficiency and fix the UI bug?

jsuarezruiz:
There are two tests failing but in UWP and they don't seem related.

PR #11351, opened 2020-07-08
[UWP] Fixes 8787 Entry text not visible until focus or window resize

bmacombe:
### Description of Change ###

Updates VisualElementRenderer to delay calling Packager.Load until the Loaded event fires.

PR #11350 should be merged first

I'm not sure I completely understand the layout pipeline of the UWP backend, so I'm not 100% sure of all the possible effects of this change could have.  It should be noted that the code to call Packager.Load in the Loaded event was previously commented out for causing an unhandled exception.  I have not experienced any exceptions during testing, but I can find no other reference to the issue other than the comment left in the code.  That comment pre-dates the import into GitHub, so it's rather old at this point.

My working theory is that the text not showing initially sometimes in the FormsTextBox is a timing issue of when the element is added to the visual tree when there are many layouts nested.  What I believe happens with this change is the elements are added later in the process when Packager.Load is executed in the Loaded event vs just called in SetElement method of VisualElementRenderer.

Fair warning, my theory could be totally wrong too!

Other similar issues where the FormsTextBox is not as deeply nested in layouts seem to work file with the changes in #11140 (minus the Focus() call) #8503 as an example

### Issues Resolved ### 

- fixes #8787 
- fixes #11259

### API Changes ###
 
 None

### Platforms Affected ### 
- UWP

### Behavioral/Visual Changes ###

Ensures Entry text is visible immediately 

### Before/After Screenshots ### 

Not applicable

### Testing Procedure ###
Issue page 8787 and full UI tests

### PR Checklist ###

- [ ] Targets the correct branch
- [ ] Tests are passing (or failures are unrelated)

PR #11394, opened 2020-07-10
[Bug] [UWP] Modify ItemsViewRenderer to handle ItemsSource being set during Layout

### LINKED ISSUE #8814: [Bug] UWP Shell cannot host CollectionView/CarouselView
### Steps to Reproduce

1. Create new Shell template
2. Add UWP project
3. Add `global::Xamarin.Forms.Forms.SetFlags("Shell_UWP_Experimental");`
3. Convert ListView to CollectionView
4. Run project on UWP

### Expected Behavior

CollectionView is displayed

### Actual Behavior

CollectionView is missing

### Basic Information

- Version with issue: 4.3.0
- IDE: VS 15.5 P1
- Platform Target Frameworks: 
  - UWP:  16299+

### Reproduction Link

[Repro.zip](https://github.com/xamarin/Xamarin.Forms/files/3942510/TestyMcTestikins.zip)

### LINKED ISSUE #10977: [Bug] UWP: Page with CollectionView is not displayed until the window is resized.
###  Description
In some situations, only blank page is rendered when the application starts. After the window is resized by the user, all the content suddenly appears. I have prepared a simplified version of my project:

[UwpBug-2020-06-06.zip](https://github.com/xamarin/Xamarin.Forms/files/4740020/UwpBug-2020-06-06.zip)

There is a page with a SearchBar, CollectionView and a Button. When the app is started, only blank page is displayed. When I resize the window, all the content (SearchBar, CollectionView and a Button) becomes visible.

This is even more complicated as this bug is present about 90% of times, but sometimes when I run the app, the content is visible immediately (usually when I delete bin and obj folders manually). When I rerun the app, the bug is there.

Another complication is that depending on Xamarin.Forms version, the behavior seems to be sometimes different depending on if you use Debug or Release versions! I have found the bug when just before uploading my app to Microsoft Store. I have used only Debug version during development. Before uploading the final version I wanted to try also the Release version just to be sure. And I have found out that it is not working. In my case, version 4.6.0.726 worked correctly in Debug but not in the Release mode. Upgrading to 4.6.0.847 made things even worse and the bug was there in both Debug and Release modes. However, I am not able to reproduce this difference in by repro solution.

Just creating the repro solution took me a few hours because I started with my app and removed one thing by another and tested if the bug is still there (both in debug and release; and UWP release build is ridiculously slow). I needed to make sure that the bug is not caused by some other Nuget.

This bug might be connected to the Issue #10862. However, there is no need to click and navigate in this case (compared to the situation mentioned in #10862). So it can be a different problem.

### Steps to Reproduce

1. Download the attached solution.
2. Run the UWP project. The page is blank (if it is not, close the app and run it again).
3. Resize the window. Now the content is visible.
4. I have included also Android project for comparison to make sure there is no problem in the code. If you try the Android project, it is displayed correctly. The bug is only in UWP.

### Expected Behavior
The content (SearchBar, CollectionView and a Button) is visible right after the start. Also the behavior should be deterministic and it should be the same for Debug and Release versions.

### Actual Behavior
The content is hidden (most of the times) until the window is resized.

### Basic Information

- Version with issue: Xamarin.Forms 4.6.0.847
- IDE: VS 2019 (16.6.1)
- Platform Target Frameworks:
  - UWP:  1903
- Nuget Packages: Only basic packages that are included when new Xamarin.Forms project is created in VS (Xamarin.Forms, Essentials, etc.).

### Screenshots
When the app is started, nothing is there:
UWPBug-release-after start

When I resize the window, all content become visible:
UWPBug-release-after resize

### LINKED ISSUE #12009: [Bug] In a UWP AppShell app, using a CollectionView within a StackLayout along with a label gives a blank content page.
I
### Description
In a UWP AppShell app, using a CollectionView within a StackLayout along with a label gives a blank screen.
### Steps to Reproduce

1.  I took the the xamarin xaminals sample app
2. I edited the CatsPage.xaml adding a couple of lines like so:
    <StackLayout>
        

        <Label Text="Test" HorizontalOptions="Center" />
        <CollectionView Margin="20"
                    ItemsSource="{x:Static data:CatData.Cats}"
                    ItemTemplate="{StaticResource AnimalTemplate}"
                    SelectionMode="Single"
                    SelectionChanged="OnCollectionViewSelectionChanged" />
    </StackLayout>

3.  Run the modified sample, in the xaminals sample the 1st screen displayed is CatsPage.xaml. The content is black. If you uncomment the Button control, run again, the page is displayed as expected.

### Basic Information
This was tested on a Windows PC using Visual Studio 2019 Version 16.7.2, Xamarin Forms v4.8.0.1364.

### Screenshots
Without Button
With Button

bmacombe:
### Description of Change ###

This PR prevents the ItemsViewRenderer from setting the ItemSource on ListViewBase before it has been through the layout process.  Waiting for the ListViewBase SizeChanged event to be called and for it to have a width is used as the check for layout completed.  UpdateItemSource was split into two methods with setting the ListViewBase.ItemSource and Empty view only occurring after SizeChanged has been called.

**Original Description**
Forces a Layout of the ItemList in the ItemsViewRenderer (CollectionView and CarouselView) when the ItemsSource is updated.

As shown in #10977 if the ItemsSource of the CollectionView is changed during a page OnAppearing event, it causes an issue with the UWP layout cycle.  It's some type of timing issues as adding a delay of even 1ms resolves the issue in testing, but relying on a simple delay could be inconsistent depending on the machine, etc.

Ideally, I would want to detect if we were in OnAppearing event and only force the UploadLayout in that case.  But I've been unable to find a way to detect that.  This PR could be improved if there is a way to do that.

Since UWP is a desktop platform, it's likely that forcing a Layout cycle of just the ListViewBase and it's sub tree won't be very noticeable.

I did test the same example in 10977 with a ListView and it worked fine to set ItemSource in OnAppearing, so it's likely that existing apps my use this method to set ItemsSource and should be expected to work.

### Issues Resolved ### 

- fixes #10977
- fixes #8814 
- likely fixes #12009 - Don't have a repro to test

### API Changes ###
 
 None

### Platforms Affected ### 

- UWP

Forces the collection view to layout and be displayed when the itemsSource is updated during page appearance.

10977
Before

Before

After

After

8814
Before

Before

After

After

### Testing Procedure ###
Open Issue 10977 page and ensure all controls are visible and click both test buttons

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/CollectionView/ItemsViewRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/CollectionView/ItemsViewRenderer.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Specialized;
 using System.ComponentModel;
 using System.Threading.Tasks;
@@ -25,6 +25,7 @@ namespace Xamarin.Forms.Platform.UWP
 		View _formsEmptyView; 
 		bool _emptyViewDisplayed;
 		ScrollViewer _scrollViewer;
+		bool _listViewBaseHasSize;
 		double _previousHorizontalOffset;
 		double _previousVerticalOffset;
 
@@ -99,6 +100,14 @@ namespace Xamarin.Forms.Platform.UWP
 				incc.CollectionChanged += ItemsChanged;
 			}
 
+			SetItemSourceAndEmptyView();
+		}
+
+		void SetItemSourceAndEmptyView()
+		{
+			if (!_listViewBaseHasSize) 
+				return;
+			
 			ListViewBase.ItemsSource = GetCollectionView(CollectionViewSource);
 			
 			UpdateEmptyViewVisibility();
@@ -188,6 +197,7 @@ namespace Xamarin.Forms.Platform.UWP
 			{
 				ListViewBase = SelectListViewBase();
 				ListViewBase.IsSynchronizedWithCurrentItem = false;
+				ListViewBase.SizeChanged += ListViewBaseOnSizeChanged;
 
 				FindScrollViewer(ListViewBase);
 
@@ -204,6 +214,21 @@ namespace Xamarin.Forms.Platform.UWP
 
 			// Listen for ScrollTo requests
 			newElement.ScrollToRequested += ScrollToRequested;
+		}
+
+		void ListViewBaseOnSizeChanged(object sender, SizeChangedEventArgs e)
+		{
+			if (e.NewSize.Width < 0)
+			{
+				return;
+			}
+
+			// The ListViewBase must have a size before attempting to set the ItemSource
+			// otherwise it will cause layout issues, controls not visible, etc
+			_listViewBaseHasSize = true;
+			SetItemSourceAndEmptyView();
+
+			ListViewBase.SizeChanged -= ListViewBaseOnSizeChanged;
 		}
 
 		protected virtual void TearDownOldElement(ItemsView oldElement)
@@ -407,12 +432,14 @@ namespace Xamarin.Forms.Platform.UWP
 
 			if (ListViewBase != null)
 			{
+				_listViewBaseHasSize = false;
 				ListViewBase.ItemsSource = null;
 				ListViewBase = null;
 			}
 
 			ListViewBase = SelectListViewBase();
 			ListViewBase.IsSynchronizedWithCurrentItem = false;
+			ListViewBase.SizeChanged += ListViewBaseOnSizeChanged;
 
 			FindScrollViewer(ListViewBase);
 
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue10977.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue10977.cs
@@ -1,0 +1,82 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Xaml;
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if APP
+	[XamlCompilation(XamlCompilationOptions.Compile)]
+#endif
+	[Issue(IssueTracker.Github, 10977, "[Bug] UWP: Page with CollectionView is not displayed until the window is resized.", PlatformAffected.UWP)]
+	public partial class Issue10977 : ContentPage
+	{
+		private MainViewModel viewModel;
+
+		public Issue10977()
+		{
+			viewModel = new MainViewModel();
+			BindingContext = viewModel;
+#if APP
+			InitializeComponent();
+#endif
+		}
+
+		protected override void OnAppearing()
+		{
+			base.OnAppearing();
+
+			viewModel.Items = new List<MainModel>
+			{
+				new MainModel { Text = "a"},
+				new MainModel { Text = "b"},
+				new MainModel { Text = "c"}
+			};
+		}
+
+		public class MainViewModel : INotifyPropertyChanged
+		{
+			private IEnumerable<MainModel> items;
+			public IEnumerable<MainModel> Items
+			{
+				get => items;
+				set
+				{
+					items = value;
+					OnPropertyChanged(nameof(Items));
+				}
+			}
+
+
+			public event PropertyChangedEventHandler PropertyChanged;
+			private void OnPropertyChanged(string propertyName)
+			{
+				PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
+			}
+		}
+
+		public class MainModel
+		{
+			public string Text { get; set; }
+		}
+
+#if APP
+		void Button_OnClicked_NewSource(object sender, EventArgs e)
+		{
+			viewModel.Items = new List<MainModel>
+			{
+				new MainModel { Text = $"{DateTime.Now.Second} d"},
+				new MainModel { Text = $"{DateTime.Now.Second} e"},
+				new MainModel { Text = $"{DateTime.Now.Second} f"}
+			};
+		}
+
+		void Button_OnClicked_NewLayout(object sender, EventArgs e)
+		{
+			var gridLayout = new GridItemsLayout(ItemsLayoutOrientation.Horizontal);
+			CV.ItemsLayout = gridLayout;
+		}
+#endif
+	}
+}
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue10977.xaml
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue10977.xaml
@@ -1,0 +1,29 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             xmlns:issues="clr-namespace:Xamarin.Forms.Controls.Issues"
+             x:Class=" Xamarin.Forms.Controls.Issues.Issue10977"
+             x:DataType="issues:MainViewModel"
+             Title="All element should be visible on load">
+
+    <ContentPage.Resources>
+        <ResourceDictionary>
+            <DataTemplate x:Key="ItemsDataTemplate">
+                <Label Text="{Binding Text}" />
+            </DataTemplate>
+        </ResourceDictionary>
+    </ContentPage.Resources>
+
+    <StackLayout>
+        <SearchBar Placeholder="Search text..." />
+        <Button Text="Click to test loading new items source after initial load" Clicked="Button_OnClicked_NewSource"/>
+        <Button Text="Click to test creating a new items layout" Clicked="Button_OnClicked_NewLayout"/>
+
+        <CollectionView x:Name="CV"
+            ItemsSource="{Binding Items}"
+            ItemTemplate="{StaticResource ItemsDataTemplate}">
+        </CollectionView>
+
+        <Button Text="OK" />
+    </StackLayout>
+</ContentPage>
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Xamarin.Forms.Controls.Issues.Shared.projitems
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Xamarin.Forms.Controls.Issues.Shared.projitems
@@ -37,6 +37,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)Issue9137.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Issue8691.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Issue7606.cs" />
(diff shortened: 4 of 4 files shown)


Xamarin.Forms.Platform.UAP/CollectionView/ItemsViewRenderer.cs
```diff
@@ -101,6 +101,10 @@ protected virtual void UpdateItemsSource()
 			ListViewBase.ItemsSource = CollectionViewSource.View;
 
 			UpdateEmptyViewVisibility();
+
+			// This is needed to ensure the layout is correct if the items source is updated while
+			// the page containing the ItemsView is still in layout.
+			ListViewBase.UpdateLayout();
```
bmacombe, line 107: @hartez What do you think of the performance implications of this?

bmacombe:
See this comment and down in 10977 for some history of the investigation into this
https://github.com/xamarin/Xamarin.Forms/issues/10977#issuecomment-656477786

bmacombe:
Updating the ItemTemplate during OnAppearing causes the same problem, but it's because it calls UpdateItemsSource so this PR fixes that scenario also

MitchBomcanhao:
> Since UWP is a desktop platform, it's likely that forcing a Layout cycle of just the ListViewBase and it's sub tree won't be very noticeable.
I know what you mean but an assumption that there is enough power and memory for things to not be noticeable is what gets us slow apps with high memory usage.
I'm working an a cross platform app and, as an example, the listview works noticeably better on an iPhone 7 than it does on my work desktop. Every time that people assume it is ok because there is enough power on the pc, it adds to this problem.

That said, I'm still happy whenever people contribute to the UWP project :)

bmacombe:
Yes I don't consider it ideal at all.  The other option I know would work would be to add a few ms delay instead.  I saw that the android platform does that in a few cases.

I'm open to suggestions on the best fix which is why this is marked draft.

@MitchBomcanhao What do you think of the delay option?  I need to find a non-asyc way to add it calling Task.Delay.Wait seems risky for me.  In my own code I've found call in wait in a non-asyc method sometimes hangs up.

bmacombe:
So I was thinking about the option of just putting a delay in today to fix this.  Here is my concern there and maybe it's unfounded.  In my testing, a Delay of even 1 ms was enough to get everything to render correctly in this specific scenario.  But that's also on an i9-9900K desktop system...is a 1 ms delay enough on say a Surface Go with a Pentium Gold CPU?

So then maybe you do a 10 ms delay to be safe as I found in an Android renderer that seems to have been added for a similar type of issue.  I'll admit I've done something like that in my app also from time to time.

Layout cycle adds time, eats cpu cycles, maybe a little ram and power...but likely will work in all cases
Delay adds time, but shouldn't cost anything else...but maybe won't always work

So which is best?

A) Always force a layout and incur a Layout Cycle (which I think should be a minimum amount, but I need to look at the code more) (Current method in this PR)
B) Add a say 10 ms delay
C) Only force a layout in the specific case it's needed (which I've failed to find a way to do)
D) Another method that someone more experienced with the layout process or UWP comes up with

To me, C is best
A and B are tied with the universal effectiveness of a delay being the tiebreaker in favor of B
And D would be totally awesome...hint hint! 😁

bmacombe:
I found the code in the Android Platform that appears to use delay for similar issues

https://github.com/xamarin/Xamarin.Forms/blob/a603efd7b338713f38ba6789ad31634208a8e71d/Xamarin.Forms.Platform.Android/Renderers/ScrollViewRenderer.cs#L380-L396

https://github.com/xamarin/Xamarin.Forms/blob/a603efd7b338713f38ba6789ad31634208a8e71d/Xamarin.Forms.Platform.Android/AppCompat/NavigationPageRenderer.cs#L745-L749

bmacombe:
I decided to switch to the delay approach using the device timer as I found in the Android platform.  I tried 1ms with no luck on my machine, 5ms seemed to work but decided to set to 10 to be safe.

https://github.com/xamarin/Xamarin.Forms/blob/4914ac2a763f406ab2c3646def066ad970082702/Xamarin.Forms.Platform.UAP/CollectionView/ItemsViewRenderer.cs#L101-L108

velocitysystems:
Thanks for the PR @bmacombe, we are also affected by us but I would really love to try and get to the root cause.
While the delay may work in certain cases, hard-coded delays are never guaranteed to work under all scenarios and their effectiveness varies with CPU workload, system performance etc.

I'd be happy to take a look at this with you and see if we can figure out a solution using (D).

bmacombe:
Yeah that is a concern I have also.  What I've been finding with UWP is that the timing of when things happens really matters.

Which seems to be the case here again just like the text box issues.

Would be glad for the help...this pr is meant more of an investigation anyway.

bmacombe:
I have a new proposed solution that somewhat fits option C above.

Based on what @PureWeen sad about UWP items sometimes needing a size set before you load them on #11351 I decided to use the SizeChanged event to determine if the UWP ListView or GridView was actually loaded.

https://stackoverflow.com/questions/34340134/how-to-know-when-a-frameworkelement-has-been-totally-rendered

Per that StackOverflow link, SizeChanged is only called once the layout has happened.

So I use a timer to delay up to 10 ms or until SizeChanged has been called.  This ensures the list or grid has been laid out and then continues to set the source or empty view

https://github.com/xamarin/Xamarin.Forms/blob/522827102fa47e6400e7a4bc1a4452c5dd512aeb/Xamarin.Forms.Platform.UAP/CollectionView/ItemsViewRenderer.cs#L101-L114

@velocitysystems What do you think?
@PureWeen Does this somewhat match what you were thinking with the FormsTextBox SizeChanged in #11351?

velocitysystems:
@bmacombe I think this is definitely a step in the right direction. What do you think about...
- Rather than using a looped timer to see if the `SizeChanged` event has fired, and then bailing if it hasn't....
- Can we simply perform the initialization _once_ the `SizeChanged` event has fired and width/height is > 0?

We'd just need to ensure once initialization is complete, the event handler self-unsubscribes i.e.

```
void OnSizeChanged(object sender, SizeChangedEventArgs e)
{
    if (e.Width <= 0)
    {
        return;
    }

    // Width and height are valid. Perform initialization.
    ListViewBase.ItemsSource = CollectionViewSource.View;
    UpdateEmptyViewVisibility();

    // Unsubscribe event handler.
    ListViewBase.SizeChanged -= OnSizeChanged;
}
```

bmacombe:
@velocitysystems 

Yeah, I like that vs the timer, but I still think we need a flag to have UpdateItemsSource bail if the size hasn't changed yet.  I'll give that combo a try today.  Thanks!

bmacombe:
So now the OnSized handler looks like this

https://github.com/xamarin/Xamarin.Forms/blob/de1015d85040518532957dc6c3fbea22ca5fd54b/Xamarin.Forms.Platform.UAP/CollectionView/ItemsViewRenderer.cs#L214-L227

That handler is hooked up when the ListViewBase is created or updated and unsubscribes itself.  When the size is set, the handler will just call UpdateItemsSource so no logic is duplicated.

In UpdateItemsSource it will now exit if the ListViewBase is null or there is no size.

https://github.com/xamarin/Xamarin.Forms/blob/9b3a096ea4eb0ba7837fe12a82ee97bcf6c3a66b/Xamarin.Forms.Platform.UAP/CollectionView/ItemsViewRenderer.cs#L79-L86

When using the 10977 issue page the UpdateItemsSource body is only executed once and the page display properly.  Ran the UWP UI tests via the script and all passed.  Doing a couple more runs also.

Manually reviewing random items in the collection view gallery look good also.

bmacombe:
@samhouts Would you mind starting a CI build to confirm the UI test and run the regression suite?  Thanks!

Xamarin.Forms.Platform.UAP/CollectionView/ItemsViewRenderer.cs
```diff
@@ -97,10 +97,16 @@ protected virtual void UpdateItemsSource()
 				incc.CollectionChanged += ItemsChanged;
 			}
 
-			ListViewBase.ItemsSource = CollectionViewSource.View;
+			SetItemSourceAndEmptyView();
+		}
+
+		void SetItemSourceAndEmptyView()
```
bmacombe, line 103: Not calling the rest of the UpdateItemsSource body before the ListViewBase has a size caused some issues with carousel view.  So I separated setting the source and empty view to another method.

bmacombe: Now the rest will be called as soon as the CollectionViewSource is set, but will only be passed to the ListViewBase once it's displayed

velocitysystems: @bmacombe Looks good and great to hear this solves the issue in the repro page!

Just one minor thing: `ListViewBaseHasSize` should probably match the naming format for private fields used in the XF project i.e. _bool _listViewBaseHasSize;_ and declared at the top of the ItemsViewRenderer along with all the other private fields used.

bmacombe: oh, yeah it should.  Good catch.

bmacombe:
I ran the same carousel view pages on iOS to see how they responded and I don't think this PR actually causes any problems with carousel view on UWP, but it might be good for someone more familiar with carousel view's current state in UWP to check.

bmacombe:
I think this bug report is what I was seeing with the carousel view #11506

bmacombe:
Add #8814 as another issue this PR will resolve

EmilAlipiev:
This await task.delay(100) workaround only works for me for loading data in OnAppearing method. But when I have CollectionView is visible=False and clicking a Button which set visible=true and at the same time raises ObservableCollection, it doesnt work. CollectionView is not visible until i resize the page in uwp. So delay is not a solution for all cases.

bmacombe:
@EmilAlipiev this pr doesn't use a delay.  It waits for the collection view to be sized by the layout system.  It should work in your case.

EmilAlipiev:
Beside that another problem, if you want to display emptyview, on uwp it will never be displayed even you use Task.Delay before setting the itemsource as null or empty. It doesnt even actually work within the constructor.

EmilAlipiev:
I know i have checked your pr. Just saying that workaround doesn't work in some cases. So we need this pr ASAP to be merged if none suggest an alternative. Have you also tested the case with empty view when you bind empty or null to the itemsource if empty view is displayed with your PR?

bmacombe:
I haven't tested an empty view.  But like the items source it's only set after the cv has a size.  I'll try a empty view soon if I get some time.

bmacombe:
@samhouts Do I need to update this PR to target a new branch?  I just noticed the retarget required label.

samhouts:
@bmacombe Yeah, you probably want this in 5.0.0 :)

bmacombe:
Done, phew no conflicts!

bmacombe:
Rebased for conflict

bmacombe:
Likely will partially resolve #11732, the non data template parts

hartez:
@bmacombe you were right, this was a timing/race condition issue. Looks like this issue was fixed by https://github.com/xamarin/Xamarin.Forms/pull/12017.  So updating to 5.0.0 pre2 or higher will take care of it.

hartez:
I'm going to go ahead and close this; if you think you have one of the issues addressed here and updating to 5.0.0 pre-2 or higher doesn't address it, ping me here and I'll reopen.

bmacombe:
@hartez  Great!, glad the other layout changes addressed this problem.  I was wondering if a lower level issue might have been involved.

PR #11410, opened 2020-07-12
Light up UWP Shell Navigation

### LINKED ISSUE #8498: [Bug] Shell Navigation.PushAsync ignored on UWP platform
**Description**
Using the solution created by VS2019 for shell and then adding a UWP project to it does not work correctly, the call to Navigation.PushAsync to bring up the detail form is ignored but the same project works correctly with Android and iOS

**Steps to Reproduce**
Create a new solution with VS2019, select shell project for iOS and Android.
Add in the UWP project, update the UWP code for Xamarin as documented.
https://stackoverflow.com/questions/41192203/how-can-i-add-a-uwp-target-to-an-existing-xamarin-forms-project

Upgrade the Nuget for the project to use the 4.3 release of Xamarin.Forms

Add the line global::Xamarin.Forms.Forms.SetFlags("Shell_UWP_Experimental"); prior to calling Xamarin.Forms.Init

Compile and run in UWP mode

When the app runs select the "First item", this should then display a new page with the details of the first item.  On UWP it does not, it stays on the same page. it seems that the line

await Navigation.PushAsync(new ItemDetailPage(new ItemDetailViewModel(item)));

is being ignored in UWP

**Expected Behavior**

I expected that when I selected the "First item" it would bring up the details page for this first item.

PureWeen:
### Description of Change ###

Build the basics for push/pop/etc.. with Shell

### Issues Resolved ### 
- fixes #8498

### Platforms Affected ### 
- UWP

### Testing Procedure ###
- click around on store shell

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/Shell/ShellItemRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/Shell/ShellItemRenderer.cs
@@ -115,11 +115,14 @@ namespace Xamarin.Forms.Platform.UWP
 			double inset = 10;
 			if (ShellContext.IsPaneToggleButtonVisible)
 				inset += 45;
-			if (Windows.Foundation.Metadata.ApiInformation.IsPropertyPresent("Controls.NavigationView", "IsBackButtonVisible"))
+
+			if (Windows.Foundation.Metadata.ApiInformation.IsPropertyPresent("Microsoft.UI.Xaml.Controls.NavigationView", "IsBackButtonVisible"))
 			{
-				if (ShellContext.IsBackButtonVisible != Microsoft.UI.Xaml.Controls.NavigationViewBackButtonVisible.Collapsed)
+				if (ShellContext.IsBackButtonVisible != Microsoft.UI.Xaml.Controls.NavigationViewBackButtonVisible.Collapsed &&
+					ShellContext.IsBackEnabled)
 					inset += 45;
 			}
+
 			_HeaderArea.Padding = new UwpThickness(inset, 0, 0, 0);
 		}
 
@@ -370,10 +373,6 @@ namespace Xamarin.Forms.Platform.UWP
 		void UpdatePageTitle()
 		{
 			_Title.Text = DisplayedPage?.Title ?? ShellSection?.Title ?? "";
-			if(ShellItem.Navigation.NavigationStack.Count > 1)
-				_Title.Padding = new UwpThickness(45, 0, 0, 0);
-			else
-				_Title.Padding = new UwpThickness(0, 0, 0, 0);
 		}
 
 		void UpdateBottomBarVisibility()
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Platform.UAP/Shell/ShellItemRenderer.cs
```diff
@@ -370,6 +370,10 @@ void UpdateNavBarVisibility()
 		void UpdatePageTitle()
 		{
 			_Title.Text = DisplayedPage?.Title ?? ShellSection?.Title ?? "";
+			if(ShellItem.Navigation.NavigationStack.Count > 1)
```
VladislavAntonyuk, line 373: _Title.Padding = new UwpThickness(ShellItem.Navigation.NavigationStack.Count > 1 ? 45 : 0, 0, 0, 0);
or 
_Title.Padding = ShellItem.Navigation.NavigationStack.Count > 1 ? new UwpThickness( 45, 0, 0, 0) : new UwpThickness(0, 0, 0, 0);

mattleibow: What about:

```csharp
var leftPad = ShellItem.Navigation.NavigationStack.Count > 1 ? 45 : 0;
_Title.Padding = new UwpThickness(leftPad, 0, 0, 0);
```

And that 45? Is that a magic number? Will it ever change? Ever?

PureWeen: It was scientifically created by changing the value until it lined up :-)

I didn't worry about it too much because these values (I find) are very often hardcoded into UWP

For example that back button has a hardcoded size of 40
https://github.com/microsoft/microsoft-ui-xaml/blob/8f1a6824d3bda36b2fdaaf25ca848d62f5c19a38/dev/NavigationView/NavigationView.cpp#L80

I still haven't quite tracked down why this is needed in the first place :-/ but I don't want that detail to block this PR

I'm not sure why when there's a back button the title grid doesn't have left padding whereas when it's a hamburger it does. Without this padding the title overlaps the back button when it's present

PureWeen: Oh there it is...

Here's the better code for handling this
https://github.com/xamarin/Xamarin.Forms/pull/11410/commits/e968d041433d28fb05b4b037bd412b6970af1508

Still using the "magic" numbers that @dotMorten originally setup.

dotMorten: giphy-downsized-large

g4mb10r:
Shell.Current.GoToAsync doesn't work. @PureWeen @samhouts @dotMorten

samhouts:
@g4mb10r Can you please open a new issue for that? Thanks!

PR #11420, opened 2020-07-13
[iOS] Null check Mask.Name

### LINKED ISSUE #11132: [Bug] [iOS] UpdateClip throws NullReferenceException when the Name of the Mask of the Layer is null
### Description
I have a custom control. Inside its renderer I am using the Layer for drawing. When I set a Mask of my own a NullReferenceException is thrown. I need to set name for the Mask in order to resolve it. I have noticed a similar issue (#11118) logged about the same method. The fix (#11131) applied for it won't work for my scenario. I believe [more checks](https://github.com/xamarin/Xamarin.Forms/blob/5dc0e2c210370eef56e456d181cd4ae2e707bde6/Xamarin.Forms.Platform.iOS/VisualElementTracker.cs#L455) are needed for the Name of the Mask.

### Steps to Reproduce

1. Run the attached project on iOS
2. Exception is thrown

### Expected Behavior
The sample should run as expected.
### Actual Behavior
Exception is thrown by the UpdateClip method
### Basic Information

- Version with issue: 4.7
- Last known good version: 4.6
- IDE: Visual Studio for Mac
- Platform Target Frameworks: 
  - iOS:  13.5

### Reproduction Link

[ClipNullReferenceIssue.zip](https://github.com/xamarin/Xamarin.Forms/files/4804164/ClipNullReferenceIssue.zip)

### Workaround
Set Name for the Mask applied for the Layer.

samhouts:
### Description of Change ###

Not sure why the gallery doesn't crash when the mask name is null, but this should be the last of the required null checks 🤞 

### Issues Resolved ### 

- fixes #11132

### API Changes ###

 
 None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###
Test with the project included in #11132 

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.ControlGallery.iOS/_11132CustomRenderer.cs
+++ b/Xamarin.Forms.ControlGallery.iOS/_11132CustomRenderer.cs
@@ -38,7 +38,7 @@ namespace Xamarin.Forms.ControlGallery.iOS
 					Path = clipPath
 				};
 				layer.Mask = clipShapeLayer;
-                layer.Mask.Name = "test";
+                layer.Mask.Name = null;
 
                 Debug.WriteLine($"_11132CustomRenderer Layer Name { layer.Mask.Name}");
             }
--- a/Xamarin.Forms.Platform.iOS/VisualElementTracker.cs
+++ b/Xamarin.Forms.Platform.iOS/VisualElementTracker.cs
@@ -452,19 +452,19 @@ namespace Xamarin.Forms.Platform.MacOS
 				hasClipShapeLayer =
 					uiview.Layer != null &&
 					uiview.Layer.Mask != null &&
-					uiview.Layer.Mask.Name == ClipShapeLayer;
+					uiview.Layer.Mask?.Name == ClipShapeLayer;
 			else
 			{
 				hasClipShapeLayer =
 					uiview.MaskView != null &&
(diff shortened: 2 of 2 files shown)


radioactiveman:
@samhouts: I don't understand why the added null checks are necessary.

The variable `Mask` can never be null in the last comparison with `ClipShapeLayer`, since in all three cases there is a null check done before. If it is null, the last comparison is not executed.

What still could cause a `NullReferenceExecption` though is in the `__MOBILE__` else-branch.
There `uiview.MaskView.Layer` could still be potentially `null`.

https://github.com/xamarin/Xamarin.Forms/blob/3fb89278a8d7f5b4235c20a80b597d4ceae3e4a7/Xamarin.Forms.Platform.iOS/VisualElementTracker.cs#L460

samhouts:
Yeah, that's weird. When I made the changes, it was putting the null check before a call to `.Equals`. I think some merge craziness must have happened. Will investigate.

samhouts:
Oh, this one beat me to it. I must not have had latest pulled. So weird. I would have expected that to conflict when I tried to merge it. Oh well! https://github.com/xamarin/Xamarin.Forms/commit/9d5aa980169feb397027a51800d4f1924dee9962#diff-def685157a4f2ba9e41464c6a6b802ab

PR #11437, opened 2020-07-15
[C] allow appthemebinding on BO

### LINKED ISSUE #11395: [Bug] Span cannot be styled for the Dark Mode (it does not support SetAppThemeColor method)
### Description
I am converting my project from Xamarin.Forms 4.6 (which used AppThemeColor) to the latest Xamarin.Forms 4.7 (where AppThemeColor has been replaced with AppThemeBinding). The AppThemeBinding is very hard to work with in general. I have found one scenario which I am not able to convert so that it worked in Xamarin.Forms 4.7.

I have created a HyperlinkSpan custom control. However I am not able to set different color for the light mode and for the dark mode. Here is a simplified version of my code:

```csharp
    public class HyperlinkSpan : Span
    {
        public HyperlinkSpan()
        {
            this.SetAppThemeColor(Span.TextColorProperty, Color.Blue, Color.Aqua);
            // ... setting other things such as underline and a TapGestureRecognizer ...
        }
    }
```

This code does not compile because Span does not contain definition of SetAppThemeColor. But if a the control would inherit from Label instead of Span, the method could be used (this is however not usable for my case because I need links in a text). After some searching I have found out that SetAppThemeColor is an extension method defined for VisualElement. Label inherits from VisualElement but Span does not.

### Expected Behavior
Everything that was possible with AppThemeColor (that was removed from Xamarin.Forms) should be also possible with new AppThemeBinding.

### Actual Behavior
Some things that were possible in Xamarin.Forms 4.6 are not possible in 4.7. This can prevent developers from using the current version.

### Basic Information

- Version with issue: Xamarin.Forms 4.7

StephaneDelcroix:
### Description of Change ###

what's possible in xaml should be possible in code

### Issues Resolved ### 

- fixes #11395

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###


Xamarin.Forms.Core/VisualElementExtensions.cs
```diff
@@ -5,12 +5,12 @@ namespace Xamarin.Forms
 {
 	public static class VisualElementExtensions
```
VladislavAntonyuk, line 6: please rename to BindableObjectExtensions

StephaneDelcroix: good catch

PR #11469, opened 2020-07-17
Fixes a layout issue with margins and Rtl

### LINKED ISSUE #3066: RTL does't switch the margin and the padding
### Description
switch from LTR to RTL doesn't mirror the margin and the padding 
### Steps to Reproduce

1. add the below XAML to a content page and check the result
` <StackLayout>

        <Label Text="LeftToRight"
               Margin="10" />
        <Grid FlowDirection="LeftToRight">
            <BoxView Margin="50,0,0,0"
                     WidthRequest="100"
                     HeightRequest="100"
                     Color="Red"
                     HorizontalOptions="Start"
                     VerticalOptions="Start"
                     FlowDirection="LeftToRight" />
        </Grid>
        <Label Text="RightToLeft"
               Margin="10" />
        <Grid FlowDirection="RightToLeft">
            <BoxView Margin="50,0,0,0"
                     WidthRequest="100"
                     HeightRequest="100"
                     Color="Red"
                     HorizontalOptions="Start"
                     VerticalOptions="Start"
                     FlowDirection="RightToLeft" />
        </Grid>

    </StackLayout>`

### Expected Behavior
switching from LTR to RTL Should mirror the margin and the padding 
### Actual Behavior
switching from LTR to RTL doesn't mirror the margin and the padding 
### Basic Information
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   

- Nuget Packages:
xamarin forms 3

### Screenshots

current state
screenshot_1529295623

desired state
screenshot_1529295404

### Reproduction Link

[RTLMarginIssue.zip](https://github.com/xamarin/Xamarin.Forms/files/2109881/RTLMarginIssue.zip)

AmrAlSayed0:
### Description of Change ###

Fixed an issue were margins were not correctly accounted for when the view's flow direction is Right-To-Left

### Issues Resolved ### 

- fixes #3066

### API Changes ###

 None

### Platforms Affected ### 

- Core/XAML (all platforms)

When different margins are applied to the left and right sides of a view and it's flow direction is right to left, the views position is going to change.

Before:
Issue3066_before
After:
Issue3066_after

### Testing Procedure ###

Issue was added to the issue gallery with its number (3066) and a unit test was created.

### PR Checklist ###


AmrAlSayed0:
I just realized that there are other places where margins were used like the other overload of `LayoutChildIntoBoundingRegion`. I will be fixing that too on the week end. Also I will be moving the test to `LayoutChildIntoBoundingRegionTests` since that seems to be a more appropriate place for it but I would also like your input on what to do.

AmrAlSayed0:
While ```LayoutChildIntoBoundingRegionTests``` seems like a more appropriate place for this test it would require that I rewrite all the test cases to take multiple margins as arguments (currently they only take one argument which caused this issue in the first place since all the margins are equal and it didn't matter what's left/start or right/end in RTL) or duplicate all of the RTL tests and modify them to take multiple parameters for uneven Margins. I think that the simple test in ```MarginTests``` is enough for this issue to not regress.

jfversluis:
This is unfortunately way old by now. Sorry this didn't get the attention it deserved. Thank you so much for your time and effort here! At this time in the Xamarin.Forms lifecycle we won't be merging this anymore. Closing as such.

PR #11474, opened 2020-07-17
Fix datetimepickerlocalization

### LINKED ISSUE #11471: # TimePicker Inconsistencies
# TimePicker Inconsistencies 

### Android 

On Android, it operates normally when the format string is "HH:mm". Both the picker and display on the TimePicker change to a 24-hour or 12-hour clock. The *Format* property also overrides the device settings if set. However, if the time format string has "HH" but isn't "HH:mm," then the display and picker have different formats. More specifically, the display is 24-hour while the picker is 12-hour.

### iOS 

On iOS, the *Format* property only changes the display of the TimePicker while the picker widget always maintains the device’s default culture. Thus, if the *Format* property is specified to be different than the operating system time format, the display and picker aren’t consistent.  

iOS also has two separate settings for region and language. However, if the region and language settings are set separately, then the TimePicker display and picker have different formats. For example, if the language is English and the region is France, then the picker is 24-hour while the display is 12-hour. 

### UWP 

Finally, on UWP, the TimePicker’s current *Format* property does not have any effect on the TimePicker on UWP. 

This bugfix will allow developers to get consistent default behaviour of the display and picker, as well as specify a *Format* property that will change both the display and picker widget accurately on all platforms. 

### Android Format

_Phone region: US, Format = "HH:m"_
 

### iOS Format 

_Phone region: US, Format = "HH:mm"_ 
 

### UWP 

_Format = "HH:mm"_
 

**_NOTE: [Here is a demo repo](https://github.com/memu8/DateTimeFormatTester) that illustrates what the format string changes on the TimePicker. Running this repo will show you what is happening in real time._**

## Implementation: 

We will parse the *Format* property, determine the clock type (12-hour or 24-hour), and then update the both the display and picker on the TimePicker (TimePicker Widget on Android, UITimePicker for iOS, and TimePicker ClockIdentifier on UWP) to use that clock. 

## Expected Result: 

If the region and language are set differently, then the TimePicker’s default format should follow the region’s cultural format. Both the display and the picker should stay consistent with the region’s default. 

If the *Format* property is set and the hour string in the TimePicker *Format* property is “HH,” then regardless of the device’s culture settings, both the TimePicker display and picker will both be in 24-hour. Similarly, if the hour string is “hh,” then both the TimePicker display and picker will both be in 12-hour. 

_If the Format property starts with “HH” (ex. “HH:mm” or “HH:mm tt”), then the expected result is a 24-hour clock display and picker regardless of device culture._

_If the Format property starts with “hh” (ex. “hh:mm” or “hh tt”), then the expected result is a 12-hour clock display and picker regardless of device culture._

## Example Usage: 

If the region and language are set differently, then the TimePicker’s default format should follow the region’s cultural format. Both the display and the picker should stay consistent with the region’s default. However, if the *Format* property is set, then it should override the region’s default. 

### Android and iOS Format: 

We are changing the effects of the *Format* property on the picker. There could be a potential issue if a developer wants the format of the display and the picker to be different. For example, if they deliberately wanted the display to be in 24-hour while the picker to be in 12-hour. Otherwise, there should be no significant changes that need to be made. 

### iOS Region vs. Language: 

We are changing the effects of the region and language on the TimePicker display format. There could be a potential issue if a developer wants the format of the display and the picker to be different. For example, if they deliberately wanted the display to follow language formatting while the picker to follows region formatting. Otherwise, there should be no significant changes that need to be made. 

### UWP: 

Since we are adding a feature and making the *Format* property accurately affect UWP TimePickers, there should be no compatibility problems. The only issue is if developers wanted the format string on TimePickers to not affect UWP applications.

_Originally posted by @memu8 in https://github.com/xamarin/Xamarin.Forms/discussions/11331_

### LINKED ISSUE #11472: # DatePicker Inconsistencies
# DatePicker Inconsistencies 

### Android 

On Android, the DatePicker operates normally. Since the picker is a calendar graphic, the display on the DatePicker changes to follow the month, day, year ordering depending on the string in the *Format* property. The *Format* property also overrides the device default if set.  

### iOS 

On iOS, there are two separate settings for region and language. However, if the region and language settings are set separately, then the DatePicker display and picker have different formats. For example, if the language is English and the region is France, then the picker shows day/month/year while the display shows month/day/year.  

### UWP: 

Finally, UWP’s DatePicker feature currently is only able to support the ordering month, day, and then year. However, it still can support the numeric (ex. November as 11) and abbreviated (ex. November as Nov.) of months. It can also support an abbreviated year (ex. 2019 as 19). In Xamarin.Forms, the DatePicker’s current *Format* property does not have any effect on UWP. For example, when the format is set as “MM/dd/yyyy” the month still appears as November rather than 11. 

This bugfix will allow developers to get consistent default behaviour of the display and picker, as well as specify a *Format* property that will change both the display and picker widget accurately on all platforms. 

### UWP: 

_Format = “MM/dd/yyyy”_ 

**_NOTE: [Here is a demo repo](https://github.com/memu8/DateTimeFormatTester) that illustrates what the format string changes on the DatePicker. Running this repo will show you what is happening in real time._** 

## Implementation: 

We will use the native platform’s date format given the region and language settings, and then update the display to stay consistent with that format. If the _Format_ property is set, then we will parse the _Format_ property, determine the type of month and year (numeric/abbreviated/default), and then update the DatePicker to use that format. 

### iOS: 

If the region and language are set differently, then the DatePicker’s default format should follow the Apple’s date/time display in settings. Both the display and the picker should stay consistent with the the iOS default while any strings (ex. month names) should respect the language setting. 

_If Region = US, Language = French, then the expected result is a day/month/year display with the month names in French._ 

 Screenshot 2020-07-02 at 11 59 57 AM

_If Region = France, Language = English, then the expected result is a day/month/year display with the month names in English._ 

 Screenshot 2020-07-02 at 12 01 51 PM

### UWP: 

If the month string in the DatePicker *Format* property is “MM,” then regardless of the device’s culture settings, both the DatePicker display and picker will both have month in a numeric format. Similarly, if the month string is “MMM,” then both the DatePicker display and picker will have abbreviated months. Finally, if the year string is “yy,” then the DatePicker display and picker will both have abbreviated years.  

Since the ordering of the DatePicker on UWP is inflexible, we will throw an exception if the format string has any other format besides month, day, year. 

_If the format string is in any other order than month/day/year, then the expected result is throwing an exception._ 

_If the month is “MM,” then the expected result is having a numerical month like 11 instead of November._ 

_If the month is “MMM,” then the expected result is having an abbreviated month like Nov._ 

_If the year is “yy,” then the expected result is 19 instead of 2019._

## Example Usage: 

It would still be used in the same way that the *Format* property is currently being used. 

### iOS: 

We are changing the effects of the region and language on the DatePicker display format. There could be a potential issue if a developer wants the format of the display and the picker to be different. For example, if they deliberately wanted the display to follow the language formatting while the picker followed region. Otherwise, there should be no significant changes that need to be made. 

### UWP: 

Since we are adding a feature and making the *Format* property accurately affect UWP DatePickers, there should be no compatibility problems. The only issue is if developers wanted the format string on DatePickers to not affect UWP applications.

_Originally posted by @memu8 in https://github.com/xamarin/Xamarin.Forms/discussions/11332_

### LINKED ISSUE #12556: How to Time Picker Set to Binding Value
Why can't I set the xaml Control TimePicker to the database field Date as follows, the result is 00:00:00 or 12:00:00 if Format is Format="HH:mm"

<TimePicker x:Name="_timePicker2" Format  ="hh:mm" Time="{Binding Date}" ></TimePicker>

Otherwise TimePicker works as far as Setting the time manually.

---
#### Document details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 9d81a385-0a5c-f7e0-da81-da565e5100c9
* Version Independent ID: 07691376-7c3c-8e9d-416e-9a92082ffab1
* Content: [TimePicker.Time Property (Xamarin.Forms)](https://docs.microsoft.com/en-au/dotnet/api/Xamarin.Forms.TimePicker.Time?view=xamarin-forms#feedback)
* Content Source: [docs/Xamarin.Forms/TimePicker.xml](https://github.com/xamarin/Xamarin.Forms-api-docs/blob/master/docs/Xamarin.Forms/TimePicker.xml)
* Product: **xamarin**
* Technology: **xamarin-forms**
* GitHub Login: @dotnet-bot
* Microsoft Alias: **dotnetcontent**

memu8:
### Description of Change ###

The Format property of the DatePicker and TimePicker creates consistent changes of the date and time displays. Also, the DatePicker format is consistent with the iOS default for specific region and language settings.

### Issues Resolved ### 

- fixes #11471 
- fixes #11472 
- fixes #12556

### API Changes ###

 None

### Platforms Affected ### 

- iOS
- Android
- UWP

The DatePicker and TimePicker displays and pickers might change depending on the format property as well as the regional settings.

Before:
pic1timepicker

After:
pic4timepicker

### Testing Procedure ###

Create DatePicker and TimePickers on Android, iOS, and UWP and play around with different format properties as well as different regional settings to see if the date and time picker displays and picker stay consistent and follow the region's culture default format for date and time respectively

### PR Checklist ###


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/xamarin/Xamarin.Forms?pullRequest=11474) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/xamarin/Xamarin.Forms?pullRequest=11474) before we can accept your contribution.<br/><br/>:x: memu8 [sign now](https://cla.dotnetfoundation.org/xamarin/Xamarin.Forms?pullRequest=11474)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/xamarin/Xamarin.Forms?pullRequest=11474) it.</sub>

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/DateTimePickerLocalizationTests.cs
```diff
@@ -0,0 +1,393 @@
...
+            RunningApp.EnterText("settingTime", "23, 59");
+            RunningApp.PressEnter();
+            var text = RunningApp.WaitForElement("timeClockOptions")[0].Text;
+            Assert.AreEqual("abceijklnopqruvwx", text);
+        }
+
+        // DatePicker Format String: d, M, y
+        // Separator: "/"
+        [Test]
+        public void DatePickerOneDigit()
```
hartez, line 235: It looks like several of these tests follow the same general format - would it be possible to use the [TestCase attribute](https://docs.nunit.org/articles/nunit/writing-tests/attributes/testcase.html) and create fewer test methods?

Xamarin.Forms.Platform.iOS/Renderers/DatePickerRenderer.cs
```diff
@@ -159,8 +161,28 @@ void UpdateDateFromModel(bool animate)
 		{
 			if (_picker.Date.ToDateTime().Date != Element.Date.Date)
 				_picker.SetDate(Element.Date.ToNSDate(), animate);
-
-			Control.Text = Element.Date.ToString(Element.Format);
+			if (Element.Format.Equals("") || Element.Format.Equals("d"))
+			{
+				NSDateFormatter dateFormatter = new NSDateFormatter();
```
hartez, line 166: For the format strings "d" and "D", doesn't `Control.Text = Element.Date.ToString(Element.Format);` produce the same results as using the NSDateFormatter?

Xamarin.Forms.Platform.UAP/DatePickerRenderer.cs
```diff
@@ -142,6 +142,50 @@ void UpdateDate(DateTime date)
...
+					Control.MonthFormat = "month.full";
+				if (year == 0)
+					Control.YearVisible = false;
+				else if (year <= 2)
+					Control.YearFormat = "year.abbreviated";
+				else
+					Control.YearFormat = "year.full";
+				if (day == 0)
+					Control.DayVisible = false;
+				else if (day == 3)
```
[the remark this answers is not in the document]
memu8: I switched it for abbreviated day of the week, but if the format requires a full day of the week, adding the integer automatically switches it to abbreviated. So, I left the full day of the week as just "dayofweek.full"

Xamarin.Forms.Platform.iOS/Renderers/DatePickerRenderer.cs
memu8: No, because when region and language are set differently the Element.Format display doesn't show the proper string. So, the only way I found to get the iOS default string for a given region and language is using an NSDateFormatter.

Xamarin.Forms.Platform.Android/Renderers/DatePickerRenderer.cs
```diff
@@ -175,7 +176,18 @@ void OnCancelButtonClicked(object sender, EventArgs e)
 
 		void SetDate(DateTime date)
 		{
-			EditText.Text = date.ToString(Element.Format);
+			if (Element.Format == "")
+			{
+				EditText.Text = date.ToShortDateString();
+			}
+			else if (Element.Format.Contains('/'))
```
samhouts, line 183: Can `Element.Format` ever be null? If so, this will cause a `NullReferenceException`. You can prevent that by using a [null-conditional operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-) here, and you probably want to review other places where you call a method from `Element.Format` to make sure you're accounting for a null value.

Xamarin.Forms.Platform.UAP/Xamarin.Forms.Platform.UAP.csproj
```diff
@@ -123,4 +123,7 @@
     <PackageReference Include="Win2D.uwp" Condition="'$(OS)' == 'Windows_NT'" Version="1.20.0">
     </PackageReference>
   </ItemGroup>
+  <ItemGroup>
+    <None Remove="obj\Debug\netstandard2.0\GitInfo.cache" />
```
samhouts, line 127: Was this intentional?

Xamarin.Forms.Platform.iOS/Renderers/DatePickerRenderer.cs
```diff
@@ -160,7 +162,31 @@ void UpdateDateFromModel(bool animate)
 			if (_picker.Date.ToDateTime().Date != Element.Date.Date)
 				_picker.SetDate(Element.Date.ToNSDate(), animate);
 
-			Control.Text = Element.Date.ToString(Element.Format);
+			//can't use Element.Format because it won't display the correct format if the region and language are set differently
+			if (Element.Format.Equals("") || Element.Format.Equals("d") || Element.Format.Equals("D"))
```
samhouts, line 166: Do you think we should use something like `string.IsNullOrWhitespace` instead of `.Equals("")`?

Xamarin.Forms.Platform.Android/Renderers/DatePickerRenderer.cs
memu8: When I switched Element.Format to null, it didn't throw a NullReferenceException. However, I did switch the first if to IsNullOrWhitespace.

memu8: I switched the first check to IsNullOrWhitespace, and when I tested setting the Element.Format to null, it didn't cause a NullReferenceException.

BurkusCat:
I think the issue #11520 still exists after this PR. On UWP I tried using "HH:mm" but it does not use a leading 0 in the picker or in the display of the selected value:
image

I think the format the UWP picker is displaying is "H:mm".

Xamarin.Forms.Platform.iOS/Renderers/TimePickerRenderer.cs
```diff
@@ -195,7 +197,35 @@ void UpdateCharacterSpacing()
 		void UpdateTime()
 		{
 			_picker.Date = new DateTime(1, 1, 1).Add(Element.Time).ToNSDate();
-			Control.Text = DateTime.Today.Add(Element.Time).ToString(Element.Format);
+			if (Element.Format.Equals(""))
+			{
+				string iOSLocale = NSLocale.CurrentLocale.CountryCode;
+				var cultureInfos = CultureInfo.GetCultures(CultureTypes.AllCultures)
+							  .Where(c => c.Name.EndsWith("-" + iOSLocale));
```
PureWeen, line 204: can you just change this code to 
```C#
var cultureInfo = CultureInfo.GetCultures(CultureTypes.AllCultures)
							  .Where(c => c.Name.EndsWith("-" + iOSLocale)).First();
```

It doesn't look like you are using the list of `cultureInfos` but always just the first element after this.
It also looks like you could move this search above the if statement because it's doing the same search inside the else as it is here.

As a side note about `Linq` (i.e. `.Where`)

`Linq` operators don't perform the given operation against the provided list until you access the data so every time this code calls `cultureInfos.First()` it's iterating through the data again and again.

It's best to just save off the data you need as early as you can and then operate against that.

You probably don't even need `linq` you could just foreach over `cultureInfos` and store that result into a local variable and then use that.

Xamarin.Forms.Platform.Android/Renderers/TimePickerRenderer.cs
```diff
@@ -164,8 +165,17 @@ void OnCancelButtonClicked(object sender, EventArgs e)
 
 		void SetTime(TimeSpan time)
 		{
-			var timeFormat = Is24HourView ? "HH:mm" : Element.Format;
-			EditText.Text = DateTime.Today.Add(time).ToString(timeFormat);
+			if (Element.Format.Equals(""))
```
PureWeen, line 168: Can this use? `String.IsNullOrEmpty` 

If not you can also do
`if (Element.Format.Equals(String.Empty))`

Xamarin.Forms.Platform.UAP/DatePickerRenderer.cs
```diff
@@ -142,6 +142,83 @@ void UpdateDate(DateTime date)
...
+
+			UpdateDay();
+			UpdateMonth();
+			UpdateYear();
+		}
+
+		void UpdateMonth()
+		{
+			Control.MonthVisible = true;
+			if (String.IsNullOrWhiteSpace(Element.Format) || Element.Format.Equals("d"))
```
PureWeen, line 154: Move this check to a helper or a method? It looks like this same check is performed a few times.

Xamarin.Forms.Controls/GalleryPages/DateTimePickerGalleries/DatePage.xaml.cs
```diff
@@ -0,0 +1,40 @@
...
+	public partial class DatePage : ContentPage
+	{
+		public DatePage()
+		{
+			InitializeComponent();
+			BindingContext = this;
+			var dep = DependencyService.Get<ILocalize>();
+			if (dep != null)
+			{
+				datesphoneculture.Text = "Device Culture: " + dep.GetCurrentCultureInfo();
```
PureWeen, line 17: Not a necessary change but another way to do this would be
`datesphoneculture.Text = $"Device Culture: {dep.GetCurrentCultureInfo()}";`

Xamarin.Forms.Controls/GalleryPages/DateTimePickerGalleries/TimePage.xaml
```diff
@@ -0,0 +1,53 @@
+﻿<?xml version="1.0" encoding="UTF-8"?>
+<ContentPage xmlns="http://xamarin.com/schemas/2014/forms" xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="Xamarin.Forms.Controls.GalleryPages.DateTimePickerGalleries.TimePage"
+             Title="TimePicker">
+    <ContentPage.Content>
+        <StackLayout VerticalOptions="Center" Margin="5">
```
PureWeen, line 6: Can you put this all inside a ScrollView for each of the gallery pages?

Xamarin.Forms.Platform.UAP/DatePickerRenderer.cs
```diff
@@ -100,7 +100,7 @@ protected override void OnElementPropertyChanged(object sender, PropertyChangedE
 		{
 			base.OnElementPropertyChanged(sender, e);
 
-			if (e.PropertyName == DatePicker.DateProperty.PropertyName)
+			if (e.PropertyName == DatePicker.DateProperty.PropertyName || e.PropertyName == DatePicker.FormatProperty.PropertyName)
```
PureWeen, line 103:
```suggestion
			if (e.IsOneOf(DatePicker.DateProperty, DatePicker.FormatProperty))
```

Xamarin.Forms.ControlGallery.iOS/GalleryPages/RegionalLocale.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using Foundation;
+using Xamarin.Forms;
+using Xamarin.Forms.Controls.GalleryPages.DateTimePickerGalleries;
+
+[assembly: Dependency(typeof(Xamarin.Forms.ControlGallery.iOS.GalleryPages.RegionalLocale))]
+namespace Xamarin.Forms.ControlGallery.iOS.GalleryPages
+{
+	public class RegionalLocale : ILocalize
```
PureWeen, line 9: Out of curiosity is this just providing an extension point for people and this is just your test for that?

Is this class here useful outside of the gallery in any way?

memu8: It's not useful outside the gallery. I think we just wanted the users to be able to tell the device locale and language from the control gallery page that I'm using. However, on iOS getting the device culture makes the language match the region. So, even if the region is US and language is French, the device culture would give fr-FR rather than en-FR which is why we had to create this class.

Xamarin.Forms.Platform.iOS/Renderers/TimePickerRenderer.cs
```diff
@@ -195,7 +197,32 @@ void UpdateCharacterSpacing()
...
+				string timeformat = cultureInfos.DateTimeFormat.ShortTimePattern;
+				NSLocale locale = new NSLocale(cultureInfos.TwoLetterISOLanguageName);
+				Control.Text = DateTime.Today.Add(Element.Time).ToString(timeformat, cultureInfos);
+				_picker.Locale = locale;
+			}
+			else
+			{
+				Control.Text = DateTime.Today.Add(Element.Time).ToString(Element.Format, cultureInfos);
+			}
+			if (Element.Format.Contains('H'))
```
PureWeen, line 214: Should this be

```C#
Element.Format?.Contains('H') == true
```

I notice a few checks around on if Format is null or empty so might be good to protect against Format being set to null?

I tested this a little bit and iOS seems to be the only platform that crashes when you set Format to null

But probably good to just sweep through those renderers see if there are any places that would be good to check for null

rmarinho (approved):

PR #11486, opened 2020-07-20
Fix [iOS] Button, ImageButton and RadioButton have minimum height of 44px

### LINKED ISSUE #11130: [Bug] [iOS] ImageButton has a minimum height of 44px
### Description
`ImageButton` has a minimum height of 44px on iOS.
This value is hard-coded into the renderer and forces a minimum height.

_Note: No minimum height is forced on other platforms including Android or UWP._

### Steps to Reproduce

1. Create an `ImageButton`
2. Use an image with a height less than 44px or set a `HeightRequest`.

### Expected Behavior
The button should renderer the same across all platforms.

### Actual Behavior
The button image is not rendered properly. Why? See code excerpt below:

https://github.com/xamarin/Xamarin.Forms/blob/db45ed23a7771a35d20abc515e866cce1d3a6f21/Xamarin.Forms.Platform.iOS/Renderers/ImageButtonRenderer.cs#L13-L43

### Basic Information
- Version with issue: All XF versions

### Workaround
Create a custom `ImageButtonRenderer` and override the `SizeThatFits` method:

```
/// <inheritdoc/>
public override CGSize SizeThatFits(CGSize size)
{
    return Control.SizeThatFits(size);
}
```

velocitysystems:
Remove minimum height of 44px enforced on `Button`, `ImageButton` and `RadioButton`.
Apple guidelines recommend a minimum touch target of 44px or greater, but this value should not be forced upon the user nor is is mandated by Apple.

_"Provide ample touch targets for interactive elements. Try to maintain a minimum tappable area of 44pt x 44pt for all controls."_ ([Source](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/adaptivity-and-layout/))

### Issues Resolved ### 
- Fixes #11130

### API Changes ###
None

### Platforms Affected ### 
- iOS

Minimum height of 44px no longer enforced on these controls.
Brings behavior on iOS to be consistent with other platforms (i.e. Android, UWP).

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/ImageButtonRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/ImageButtonRenderer.cs
@@ -14,32 +14,12 @@ namespace Xamarin.Forms.Platform.iOS
 	{
 		bool _isDisposed;
 
-		// This looks like it should be a const under iOS Classic,
-		// but that doesn't work under iOS 
-		// ReSharper disable once BuiltInTypeReferenceStyle
-		// Under iOS Classic Resharper wants to suggest this use the built-in type ref
-		// but under iOS that suggestion won't work
-		readonly nfloat _minimumButtonHeight = 44; // Apple docs
-
-
 		[Preserve(Conditional = true)]
 		public ImageButtonRenderer() : base()
 		{
 			ButtonElementManager.Init(this);
 			BorderElementManager.Init(this);
 			ImageElementManager.Init(this);
-		}
-
-		public override SizeF SizeThatFits(SizeF size)
-		{
-			var result = base.SizeThatFits(size);
-
-			if (result.Height < _minimumButtonHeight)
-			{
-				result.Height = _minimumButtonHeight;
-			}
-
-			return result;
 		}
 
 		protected override void Dispose(bool disposing)


velocitysystems:
> I think this would be a breaking change for users that are expecting to have a minimum default size.
> 
> I think we might need to make a virtual method or add a platform specific for this.
> 
> maybe we can then remove that on 5.0

Thanks @rmarinho. I agree this is _partially_ a breaking change, but it corrects erroneous behaviour which only exists on iOS.
On Android and UWP, it is possible to create an `ImageButton` smaller than 44px. However on iOS, a default min size of 44px will be enforced.

Q: What would be the **impact** of this change? For users who have an asset smaller than 44px in an `ImageButton` on iOS, it would now display at 1:1 size. However, simply updating the iOS image asset would correct any display issues and this could be called out under breaking changes. I agree possibly it may be more suited to target this in 4.8 or 5.0, but it feels like a little bit of a step backward to put this underneath a platform-specific when effectively it is only introducing the correct behaviour to match the other platforms.

What are your thoughts @rmarinho @samhouts?

rmarinho:
At least i think we should do this for 5.0

velocitysystems:
Thanks @rmarinho.
Should I retarget to master?

rmarinho:
@velocitysystems yes

velocitysystems:
@rmarinho Done, I have targeted `main`.

velocitysystems:
> This enforcement exists a few places
> 
> https://github.com/xamarin/Xamarin.Forms/search?q=_minimumButtonHeight&unscoped_q=_minimumButtonHeight
> 
> I don't think we can straight up delete this.
> 
> I'm thinking we'll need to keep this value in place unless the user has specifically indicated a heightrequest/minheightrequest < 44
> 
> Otherwise people are going to just see a lot of layouts just randomly adjust when they update to 5.0

I think this is a good compromise @PureWeen. So to clarify, I'll modify the PR to include `Button`, `ImageButton` and `RadioButton` with the following changes:
- If `HeightRequest`/`MinHeightRequest` **is specified**, then use this value _even if it is less_ than 44px.
- If `HeightRequest`/`MinHeightRequest` **is not specified**, then utilize the current behavior for backward compatibility.

What do you think about that?

PureWeen:
@velocitysystems this sounds good but just ignore `RadioButton` for now. That's going through a heavy rewrite

velocitysystems:
@PureWeen After reviewing the pro's and con's of this, I'm abandoning this PR for now.
Perhaps may I simply recommend this workaround is removed in the future for Maui.

PR #11601, opened 2020-07-30
Easing type converter

AndreiMisiukevich:
### Description of Change ###
Added EasingTypeConverter for convenient use Easings with xaml.

### API Changes ###
Added:
 - EasingTypeConverter

### Platforms Affected ### 
- Core/XAML (all platforms)

**BEFORE**
```xaml
<Expander ExpandAnimationEasing="{x:Static Easing.CubicIn}"
          CollapseAnimationEasing="{x:Static Easing.CubicOut}">
<!-- Content -->
</Expander>
```

**AFTER**
```xaml
<Expander ExpandAnimationEasing="CubicIn"
          CollapseAnimationEasing="Easing.CubicOut">
<!-- Content -->
</Expander>
```

Not applicable

### Testing Procedure ###
Run unit test **TestEasingTypeConverter** test case.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/CompiledConverters/EasingTypeConverter.cs
+++ b/Xamarin.Forms.Build.Tasks/CompiledConverters/EasingTypeConverter.cs
@@ -17,10 +17,15 @@ namespace Xamarin.Forms.Core.XamlC
 				yield break;
 			}
 
+			value = value?.Trim() ?? "";
+			var parts = value.Split('.');
+			if (parts.Length == 2 && parts[0] == nameof(Easing))
+				value = parts[parts.Length - 1];
+
 			var assemblyTypeInfo = ("Xamarin.Forms.Core", "Xamarin.Forms", nameof(Easing));
-			var easing = value?.Trim();
+			
 			var module = context.Body.Method.Module;
-			var fieldReference = module.ImportFieldReference(assemblyTypeInfo, easing, isStatic: true, caseSensitive: false);
+			var fieldReference = module.ImportFieldReference(assemblyTypeInfo, value, isStatic: true, caseSensitive: false);
 
 			if (fieldReference != null)
 			{
--- a/Xamarin.Forms.Build.Tasks/CompiledConverters/EasingTypeConverter.cs
+++ b/Xamarin.Forms.Build.Tasks/CompiledConverters/EasingTypeConverter.cs
@@ -1,0 +1,34 @@
+﻿using System.Collections.Generic;
+using Mono.Cecil.Cil;
+using Xamarin.Forms.Xaml;
+using Xamarin.Forms.Build.Tasks;
+using static Mono.Cecil.Cil.Instruction;
+using static Mono.Cecil.Cil.OpCodes;
+
+namespace Xamarin.Forms.Core.XamlC
+{
+	class EasingTypeConverter : ICompiledTypeConverter
+	{
+		public IEnumerable<Instruction> ConvertFromString(string value, ILContext context, BaseNode node)
+		{
+			if (string.IsNullOrWhiteSpace(value))
+			{
+				yield return Create(Ldnull);
+				yield break;
+			}
+
+			var assemblyTypeInfo = ("Xamarin.Forms.Core", "Xamarin.Forms", nameof(Easing));
+			var easing = value?.Trim();
+			var module = context.Body.Method.Module;
+			var fieldReference = module.ImportFieldReference(assemblyTypeInfo, easing, isStatic: true, caseSensitive: false);
+
+			if (fieldReference != null)
+			{
+				yield return Create(Ldsfld, fieldReference);
+				yield break;
+			}
+
+			throw new BuildException(BuildExceptionCode.Conversion, node, null, value, typeof(Easing));
+		}
+	}
+}
--- a/Xamarin.Forms.Core.UnitTests/EasingTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/EasingTests.cs
@@ -28,5 +28,49 @@ namespace Xamarin.Forms.Core.UnitTests
 			Assert.True (Math.Abs (val - Easing.SpringIn.Ease (val)) < epsilon);
 			Assert.True (Math.Abs (val - Easing.SpringOut.Ease (val)) < epsilon);
 		}
+
+		[Test]
+		public void TestEasingTypeConverter()
+		{
+			var converter = new EasingTypeConverter();
+			Assert.True(converter.CanConvertFrom(typeof(string)));
+			Assert.Null(converter.ConvertFromInvariantString(null));
+			Assert.Null(converter.ConvertFromInvariantString(string.Empty));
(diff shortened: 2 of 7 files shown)


jfversluis:
If it helps anything in deciding whether to add this or not; we are using this in the XamarinCommunityToolkit. We could use this there to make the XAML syntax of it a bit prettier :)

Xamarin.Forms.Core/EasingTypeConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+				return null;
+
+			value = value?.Trim() ?? "";
+			var parts = value.Split('.');
+			if (parts.Length == 2 && parts[0] == nameof(Easing))
+				value = parts[parts.Length - 1];
+
+			switch (value)
+			{
+				case string easing when AreEqual(easing, nameof(Linear)): return Linear;
```
StephaneDelcroix, line 22: there's no need to try casting the value to string each time. at this time, we're already sure it's a non-empty string

Xamarin.Forms.Core/EasingTypeConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+				case string easing when AreEqual(easing, nameof(CubicInOut)): return CubicInOut;
+				case string easing when AreEqual(easing, nameof(BounceIn)): return BounceIn;
+				case string easing when AreEqual(easing, nameof(BounceOut)): return BounceOut;
+				case string easing when AreEqual(easing, nameof(SpringIn)): return SpringIn;
+				case string easing when AreEqual(easing, nameof(SpringOut)): return SpringOut;
+				default: throw new InvalidOperationException($"Cannot convert \"{value}\" into {typeof(Easing)}");
+			}
+		}
+
+		bool AreEqual(string first, string second)
```
StephaneDelcroix, line 37: I'm not sure this helper helps a lot, I'd remove it so the code is easier to read

Xamarin.Forms.Core.UnitTests/EasingTests.cs
```diff
@@ -28,5 +28,49 @@ public void AllRunFromZeroToOne ([Values (0.0, 1.0)] double val)
 			Assert.True (Math.Abs (val - Easing.SpringIn.Ease (val)) < epsilon);
 			Assert.True (Math.Abs (val - Easing.SpringOut.Ease (val)) < epsilon);
 		}
+
+		[Test]
+		public void TestEasingTypeConverter()
```
StephaneDelcroix, line 33: I would prefer a XAML unit test that will also test the compiled converter

Xamarin.Forms.Build.Tasks/CompiledConverters/EasingTypeConverter.cs
```diff
@@ -0,0 +1,39 @@
...
+using Xamarin.Forms.Xaml;
+using Xamarin.Forms.Build.Tasks;
+using static Mono.Cecil.Cil.Instruction;
+using static Mono.Cecil.Cil.OpCodes;
+
+namespace Xamarin.Forms.Core.XamlC
+{
+	class EasingTypeConverter : ICompiledTypeConverter
+	{
+		public IEnumerable<Instruction> ConvertFromString(string value, ILContext context, BaseNode node)
```
StephaneDelcroix, line 12: this isn't tested, see comment in the test file

Xamarin.Forms.Core/EasingTypeConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+				case string easing when AreEqual(easing, nameof(SinOut)): return SinOut;
+				case string easing when AreEqual(easing, nameof(SinInOut)): return SinInOut;
+				case string easing when AreEqual(easing, nameof(CubicIn)): return CubicIn;
+				case string easing when AreEqual(easing, nameof(CubicOut)): return CubicOut;
+				case string easing when AreEqual(easing, nameof(CubicInOut)): return CubicInOut;
+				case string easing when AreEqual(easing, nameof(BounceIn)): return BounceIn;
+				case string easing when AreEqual(easing, nameof(BounceOut)): return BounceOut;
+				case string easing when AreEqual(easing, nameof(SpringIn)): return SpringIn;
+				case string easing when AreEqual(easing, nameof(SpringOut)): return SpringOut;
+				default: throw new InvalidOperationException($"Cannot convert \"{value}\" into {typeof(Easing)}");
```
StephaneDelcroix, line 33: I like that this doesn't use reflection, but it's not future proof, and could differ in behaviour with the compiled converter if/when someone add an easing and forget to add a case here.

what we do in `Color` e.g. is this kind of switch, plus a fallback using reflection as a safety net

Xamarin.Forms.Core/EasingTypeConverter.cs
AndreiMisiukevich: I decided to not allocate new strings by calling "ToLower" or "ToUpper" method, that's why I added "AreEqual" method which compares two strings case-independently.

Xamarin.Forms.Core/EasingTypeConverter.cs
AndreiMisiukevich: I did it for case-insensitive strings comparison.
I don't know how to simplify this syntax( Appreciate any hints

Xamarin.Forms.Core/EasingTypeConverter.cs
AndreiMisiukevich: Hmm, how about writing the next test case.

With reflection, I will retrieve all public static read-only properties which have type "Easing".
Then we will pass their names into the original test-case.
And if an exception occurred, the test will be failed and it will be noticed (And no need for Reflection in Easing class).

I hope you got my idea

StephaneDelcroix: you can't unfortunately test the fallback with an automatic unit test

Xamarin.Forms.Core/EasingTypeConverter.cs
StephaneDelcroix: you can still use string comparison without the helper

Xamarin.Forms.Core/EasingTypeConverter.cs
StephaneDelcroix: the most efficient way to write this (if you care about perf) is as follow.

```csharp
if (value.Equals(name(SinIn))
   return SinIn;
if (value.Equals(name(SinOut))
   return SinOut;
if (value.Equals(name(SinInOut))
   return SinInOut;
...
```

- going through the helper method reduce the number of method invocation
- there's no locals created for the method (from the pattern matching)

AndreiMisiukevich:
@StephaneDelcroix all comments are addressed.
The code is updated. Please re-review

StephaneDelcroix:
> @StephaneDelcroix all comments are addressed.

well, not the one about the reflection fallback, but 👍

StephaneDelcroix:
@AndreiMisiukevich please consider retargeting this to 5.0.0

AndreiMisiukevich:
@StephaneDelcroix thanks for retargeting 👍

PR #11602, opened 2020-07-30
Don't update safe area offset on non visible page

### LINKED ISSUE #10608: [Bug] [Shell] [iOS] Locked flyout causes application to freezes when quickly switching between tabs
### Description

Having the flyout locked and the tabs displayed as multiple items in the flyout allows the user to switch between different tabs in different flyout items rapidly. If the tabs have more than one ShellContent as children, and the user switches between tabs in different flyout items quickly enough, the app becomes unresponsive without an exception. 

Note that this will only happen when each Tab has multiple ShellContent children. Having a single ShellContent in this scenario does not result in the unresponsive behavior. 

Additionally, if the switch between tabs is slow enough, the app works as expected. It is only when the user moves quickly between the two.

### Steps to Reproduce

1. Create a new shell application.
2. Add two FlyoutItems to the shell. 
3. Add FlyoutDisplayOption="AsMultipleItems" to each FlyoutItem. 
4. Under each FlyoutItem, create a Tab.
5. Under each Tab create 2 or more ShellContent items (Note that it must be more than 1 to recreate).
6. Add Shell.FlyoutBehavior="Locked" to each of the ShellContents.
7. Run the application and quickly tap back and forth between the two Tabs displayed the locked flyout multiple time.

### Expected Behavior

The application navigates between the tabs and displays the corresponding page of the first ShellContent.

### Actual Behavior

The application becomes unresponsive without an exception.

### Basic Information

- Version with issue: 4.6.0.726
- IDE: Visual Studio for Mac
- Platform Target Frameworks: 
  - iOS:  13.4
- Affected Devices: Recreated on iPhone 11 Simulator, iPad Pro (4th gen) Simulator, and Physical iPad (7th gen). Looks to affect all devices.

### Screenshots

App freezes on quick switch of tabs

### Reproduction Link

[FlyoutItemRapidSwitchFreeze.zip](https://github.com/xamarin/Xamarin.Forms/files/4595038/FlyoutItemRapidSwitchFreeze.zip)

PureWeen:
### Description of Change ###

- the ios page renderer when disappearing was still propagating its inset information to shell which was causing a ping pong effect
- Automation properties weren't being set on top tabs

### Issues Resolved ### 
- fixes #10608

### Platforms Affected ### 
- iOS
- Android

### Testing Procedure ###
- ui tests included

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue10608.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue10608.cs
@@ -1,0 +1,97 @@
+﻿using System.Threading.Tasks;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.Forms.Core.UITests;
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if UITEST
+	[Category(UITestCategories.Shell)]
+#endif
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 10608, "[Bug] [Shell] [iOS] Locked flyout causes application to freezes when quickly switching between tabs", PlatformAffected.iOS)]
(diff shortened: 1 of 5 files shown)


rmarinho:
Failing tests don't seem related

Issue #5482, opened 2019-03-06
[Enhancement, Visual]  Changing Visual at runtime won't rerender the control


PureWeen:
## Summary
Changing the visual after a control has renderered currently has no effect

@PureWeen  almost has this code working it just needs to be tweaked for Cells and possibly CollectionView

jfversluis:
Let's check for .NET MAUI, this won't be happening for Forms anymore.

Issue #5005, opened 2019-01-17
Customizing a specific Visual 


PureWeen:
# Spec for adding properties to customize the behavior of a visual 

- **Button Styles**  
  There are 4 button styles in material (https://material.io/design/components/buttons.html#specs): **Contained** _(current style)_, **Border**, **Text**, **Toggle / Icon**. We could create an attached property that allows us to set this style: `Material.ButtonStyle="Contained"`. One thing to note with this is that Android will be hard, if not impossible, to dynamically change. iOS just has a theme that you apply (not so sure if it supports undoing a style and may not actually work). Android has a style resource that probably needs to be applied in the constructor of the renderer. To support dynamic changes like this, we will probably not be able to use the "fast renderer" style and would need to use the `ViewRenderer<T, T>` base. iOS is already using this, so we can always recreate the control, but for Android we may lose the optimizations gained.
- **Entry / Editor Styles**
- **Frame / Card Styles**

### Attached Properties

```XAML
<Button material:Button.Style="Outline"></Button>
````
#### Pros
- Attached properties can easily be split up according to the Element it's going to be applied to.
#### Cons
- having to import custom namespaces is cumbersome for developers. 

### Specify the renderer with dotnet MAUI

Because dotnet MAUI inverts the dependencies and now a Button knows about its renderer we could use that detail to customize visual

https://github.com/dotnet/maui/issues/28
```XAML
<Button>
   <Button.Renderer>
     <MaterialButtonRenderer Style="Outline" >
         <MaterialButtonTemplate Style="Outline"  />
     </MaterialButtonRenderer>
   </Button.Renderer>
</Button>
````

Once we have xplat controls we could also use this same technique

```XAML
<Button>
   <Button.Renderer>
     <ControlTemplateRenderer Style="Outline" >
         <ControlTemplate>
               <Grid></Grid>
         </ControlTemplate>
     </ControlTemplateRenderer>
   </Button.Renderer>
</Button>
````

### Templating

If the user specifies a MaterialButtonTemplate than they don't have to specify the Material Visual. This doesn't obsolete Visual though because Visual is still used to hierarchically style an application. When you apply Visual at the parent level it will indicate what control template to use down the hierarchy

```XAML
<Button>
     <VisualTemplate>
         <MaterialButtonTemplate Style="Outline"  />
     </VisualTemplate>
</Button>
````

[...]

PR #11628, opened 2020-07-31
Templated RadioButtons and RadioButtonGroups

hartez:
#### Control Templates

[Update: we've moved back to the ToString() behavior instead of exceptions; also, UWP now natively supports arbitrary Forms Views for Content.]

These changes make RadioButton a TemplatedView. This allows for the use of ControlTemplate (to make RadioButton look and behave the same on all platforms), or for the use of a native Renderer (to utilize the native RadioButton control available on platforms which support it).

This also allows for the RadioButton to fall back to the default ControlTemplate on iOS, which does not have a RadioButton control.

The RadioButton no longer has a `Text` property. Instead, it has a `Content` property which can be set to string or to arbitrary content. If the `Content` property is set to string, that string is used as the displayed text on any native renderers. Any platform using a ControlTemplate will display the string content as a Label. 

If the `Content` property is set to View, the native renderers which do not support arbitrary content will simply display the result of the `ToString()` method. ControlTemplates will display the View in the `ContentPresenter` section of the template. Platforms which support arbitrary content natively (i.e., UWP) will convert the View to a native renderer and display that.

So this works on all platforms/renderers:
`<RadioButton Content="Option A" />`

If we set a ControlTemplate explicitly, we can use arbitrary Views as Content on all platforms:
```
<RadioButton ControlTemplate="{x:Static RadioButton.DefaultTemplate}">
    <RadioButton.Content>
        <Image Source="coffee.png"/>
    </RadioButton.Content>
</RadioButton>
```

Setting arbitrary Views as Content without using a ControlTemplate will simply display the string representation. So on Android, the following markup:
```
<RadioButton>
    <RadioButton.Content>
        <Image Source="coffee.png"/>
    </RadioButton.Content>
</RadioButton>
```

Will simply display the string "[Xamarin.Forms.Image]". On UWP, the Forms Image will be converted to a UWP Image and displayed. Because iOS is _always_ using a template for RadioButton - there is no native renderer on that platform - an Image will be displayed.

#### RadioButtonGroups

These changes also introduce RadioButtonGroup attached properties, which allow any Layout to be turned into a radio button group:

```
<StackLayout RadioButtonGroup.GroupName="foo">
    <RadioButton Content="Option A" />
    <RadioButton Content="Option B" />
    <RadioButton Content="Option C" />
</StackLayout>
``` 

All of the RadioButtons in the StackLayout above will be given the GroupName of "foo" automatically, and will be mutually exclusive. 

```
<StackLayout RadioButtonGroup.GroupName="{Binding GroupName}" 
                         RadioButtonGroup.SelectedValue="{Binding Selection}">
    <RadioButton Content="Option A" Value="yes" />
    <RadioButton Content="Option B" Value="no" />
    <RadioButton Content="Option C" Value="maybe" />
</StackLayout>
``` 

All of the RadioButtons in the StackLayout above will be give a GroupName based on the bound value of `GroupName` in the BindingContext. The `Value` property of whichever RadioButton is selected will be bound to the `Selection` property in the BindingContext.

### Issues Resolved ### 

None

### API Changes ###

RadioButton is no longer a subclass of Button; rather, it inherits from TemplatedView

- Removed the `Text` property
- Added the `Content` property
- Added the `Value` property

Added class:

public static class RadioButtonGroup
	public static readonly BindableProperty GroupNameProperty
	public static readonly BindableProperty SelectedValueProperty

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP
- WPF
- Tizen
- macOS

None

Not applicable

### Testing Procedure ###

Control Gallery -> RadioButton Galleries

Also a bunch of unit tests

### PR Checklist ###


andreinitescu:
> If the Content property is set to View, the native renderers will throw an exception

It's important not to throw an exception in my opinion, the native renderers can just use `Content?.ToString()` instead.

andreinitescu:
Quite interesting how you decided such a breaking change, I remember piles of discussion of "oh, no, we can't do this, it would be a breaking change". Happy to see you're moving forward.

hartez:
A case we haven't considered here is what happens if a ControlTemplate is set after the native renderer has already been selected (and is possibly already on screen)? The easy option would be to simply ignore it - is that reasonable? 
@PureWeen Any thoughts on this? What happens when a Visual changes at runtime?

hartez:
> It's important not to throw an exception in my opinion, the native renderers can just use `Content?.ToString()` instead.

That's what the first draft did. The reason for the change was to make it as obvious as possible to the user that what they are trying to do won't work. And to give them an explicit message explaining why (instead of just seeing weird behavior like a RadioButton with a label text of "[Xamarin.Forms.Image]").

The "property order" problem isn't as big here as it would be if we were doing it in Core - we don't have to worry about XAML order or anything like that, because these exceptions don't happen until we've reached the native renderers. By definition, we couldn't be in this stage if ControlTemplate were anything but `null`. 

Can we come up with a concrete example of a situation where this behavior would cause confusion or catastrophe? (That might depend on the outcome of my [previous comment](https://github.com/xamarin/Xamarin.Forms/pull/11628#issuecomment-667558142)),

bmacombe:
> If the `Content` property is set to View, the native renderers will throw an exception. ControlTemplates will display the View in the `ContentPresenter` section of the template.

When I first read that I didn't catch the "View" part.  So would it only throw if you put a view as the content and not other random object or binding?

If it's only if you assign a View subclass, then I think the exception is good.  But let's say I bind the content to some random object, I would expect it to just ToString it.

andreinitescu:
Like I said on Twitter, in my opinion having a specific mandatory order for setting properties otherwise it throws an exception is not a good design. I don’t remember any API in .NET which works like this. 
I don’t have an example right now when it won’t actually work. Even if there is none, the fact I have to set properties in a specific order doesn’t sound good to me.
I realized you want to throw an exception so the developer can see it, but in this particular scenario I don’t think it’s a good decision.
At most you could write a warning message to the device log.

andreinitescu:
Beside a warning in device log, having a clear concise comment on the Content property is effective too

PureWeen:
Overall works great. 

Should this still render text into the ContentPresenter?

```XAML
        <RadioButton>
            <RadioButton.ControlTemplate>
                <ControlTemplate>
                        <ContentPresenter></ContentPresenter>
                    </StackLayout>
                </ControlTemplate>
            </RadioButton.ControlTemplate>
            <RadioButton.Content>
                Rabbits
            </RadioButton.Content>
        </RadioButton>
```


>  Any thoughts on this? What happens when a Visual changes at runtime?

Nothing happens when you change visual at runtime. Visual is such a drastic change with how a field is displayed that it's not a scenario people have ran into. 

Here's the Issue about it
https://github.com/xamarin/Xamarin.Forms/issues/5482
No thumbs up on it yet :-) 

I would default the direction of just doing nothing for now and seeing how people use it.  It's weird to switch from the native renderer to the xplat renderer because the radiobutton itself renderers so different.  If specifying the ControlTemplate created a completely symmetrical RadioButton then I think it'd make a bit more sense but I feel like you're either going to do one or the other. 

Here was the PR I had started for visual which would cause elements to recreate
https://github.com/xamarin/Xamarin.Forms/commit/b3d53cbadb34cfcac73f98afe8be8afe632b4c43

It was just a really quick spike but I feel like not worrying about this case for now makes the most sense

> some thoughts on exceptions

If we keep the current behavior I'm a fan of the exceptions. 

I type the following sentence a lot "did you look at your application output for warnings about bindings?". If I specify an `Image` or a `Label` for my content and nothing displays on android but it does display on iOS my first thought will be that there's a bug on Android

Going the exception route
- Should Android throw an exception when you specify complex content so that the user is aware of why the content won't display? 
- If we keep the exceptions should they refer to the platform and not the renderer? 
`$"WPF only supports string values for the {nameof(RadioButton)} {RadioButton.ContentProperty.PropertyName} property."`
vs
`$"{nameof(RadioButtonRenderer)} only supports string values for the {nameof(RadioButton)} {RadioButton.ContentProperty.PropertyName} property."`
- `Can we come up with a concrete example of a situation where this behavior would cause confusion or catastrophe?` I can't off hand but this might just be a failure of imagination because, like you said, these only throw once they hit the rendering stage of life. The main case I can think of here is with CollectionView if you have something like a DataTemplateSelector where some are native and some aren't, but again I feel like that's not going to come up.

> The just make it work option opposed to exceptions

For the indicators @rmarinho made is so if you specify a template it just uses the xplat indicators and if you don't then it uses the native ones. Would it make sense to follow a similar strategy here?

So if I specify
```XAML
<RadioButton>
            <RadioButton.Content>
                <Image Source="coffee.png"/>
            </RadioButton.Content>
        </RadioButton>
```

It should just opt in to use the `DefaultTemplate`? I refer back to this statement `If I specify an Image for my content and nothing displays on android but it does display on iOS then my first thought will be that there's a bug on Android.`


> UncheckedButton and CheckedIndicator

Do those names come from something? 

I don't feel like `UncheckedButton` is accurate because for me that implies it won't be visible when you've selected the radiobutton but the `UncheckedButton` is visible.

Maybe "Button" and ("CheckedIndicator" or "Checkmark" or "RadioButton")

PureWeen:
I know we had chatted a bit where visual fits into all of this which opens a few other options. @jsuarezruiz, @rmarinho , and I have talked about this a bit with other implementations. For example when @rmarinho first did the indicators he had added a "FORMS" Visual which indicated to the control that it should use xplat instead of platform. We eventually got rid of that for indicators because it didn't really offer a different experience setting the visual to "forms" vs "default"

With RadioButton that's a bit different, especially once material and fluent enter the mix. There were a few ideas we've been floating around visual 
https://github.com/xamarin/Xamarin.Forms/issues/5005

So for RadioButton it would look something like

```XAML
<RadioButton>
     <RadioButton.VisualTemplate>
         <Forms Content="Text" UncheckedButton="" ControlTemplate=""   />
     </RadioButton.VisualTemplate>
</RadioButton>
```

```XAML
<RadioButton>
     <RadioButton.VisualTemplate>
         <Native Text="Text"  />
     </RadioButton.VisualTemplate>
</RadioButton>
```

We would still let people specify "Content" on the RadioButton itself but then we could add discoverability of customization this way.

mattleibow:
Hey folks, I am having a look at the basic `TemplatedView` and came across an "issue" with the app not automatically importing styles. I created this issue (https://github.com/xamarin/Xamarin.Forms/issues/11798) which has a bit more info. I think something like this should be done by the framework.

hartez:
> > If the `Content` property is set to View, the native renderers will throw an exception. ControlTemplates will display the View in the `ContentPresenter` section of the template.
> 
> When I first read that I didn't catch the "View" part. So would it only throw if you put a view as the content and not other random object or binding?

Right now it'll throw if you set the Content to anything _but_ a `string`. So some other random object would also throw. Which I suppose would be a shock to someone who has bound the Content property to, say, an integer.

hartez:
Also, I apparently forgot to add the `string` check on Android 🤦 .

bmacombe:
> Right now it'll throw if you set the Content to anything _but_ a `string`. So some other random object would also throw. Which I suppose would be a shock to someone who has bound the Content property to, say, an integer.

My two cents fwiw, is anything that isn't a View should just ToString(), it's the behavior I would naturally expect.  But I like the exception if it's a View and you are obviously doing something that isn't intended.  

I sometimes bind a business object that has a ToString override, probably not the best practice, but I do that so I can see what the object is when debugging easier.  I sometimes then bind an object to a Content Property when I'm just doing the initial layout of a UI just to confirm things are working before refining, etc.

hartez:
Okay, at the very least I need to update the Content checks to _only_ throw an exception if the type is a View (or subclass of View). 

But I'm starting to wonder if the original policy of just displaying `Content.ToString()` was the better option. Then it's consistent for any type you give it, even a View. 

> I type the following sentence a lot "did you look at your application output for warnings about bindings?". If I specify an Image or a Label for my content and nothing displays on android but it does display on iOS my first thought will be that there's a bug on Android

We definitely have a problem with surfacing runtime warnings and errors to users, but perhaps we just need to improve their visibility and our docs.

> If we keep the exceptions should they refer to the platform and not the renderer? 

I think that's a good idea. The renderers have a concept of current platform (the Platform Specifics use it), so that should be easy enough to do.


> I don't feel like UncheckedButton is accurate because for me that implies it won't be visible when you've selected the radiobutton but the UncheckedButton is visible.

> Maybe "Button" and ("CheckedIndicator" or "Checkmark" or "RadioButton")

Agreed.

hartez:
> For the indicators @rmarinho made is so if you specify a template it just uses the xplat indicators and if you don't then it uses the native ones. Would it make sense to follow a similar strategy here?

Are you suggesting that if the Content type is View and the developer has not specified a ControlTemplate, we automatically choose the default template? 

I'm not keen on that, and here's why: it has occurred to me since I first implemented this stuff that there's no reason we can't support Content as a View on UWP and WPF. On both platforms the native RadioButton supports arbitrary content. So for those platforms, we can just create a renderer for the xplat Content and use that in the native control. 

But we can't do that if we implement the suggested rule.

bmacombe:
> But I'm starting to wonder if the original policy of just displaying `Content.ToString()` was the better option. Then it's consistent for any type you give it, even a View.

I would be ok with that unless I'm going crazy or losing my memory (which is possible!) that is how WPF, Silverlight, and most other Forms Content properties behave?

hartez:
> > But I'm starting to wonder if the original policy of just displaying `Content.ToString()` was the better option. Then it's consistent for any type you give it, even a View.
> 
> I would be ok with that unless I'm going crazy or losing my memory (which is possible!) that is how WPF, Silverlight, and most other Forms Content properties behave?

You're not crazy. 

@davidortinau and @pureween - I'm leaning very heavily in the direction of going back to `Content.ToString()` for this and simply logging a warning for the unsupported platforms. It's more consistent, easier to implement, and easier to maintain. I know we have visibility issues with those warnings, but I think that's a separate issue. 

Change my mind or I'm updating this PR later this week. :)

hartez:
Okay, updates have been made. ToString() is called where arbitrary Content is not supported, a warning message is logged where appropriate, and UWP will now display Views as Content in the native RadioButton as well as the ControlTemplate version.

PureWeen:
> I'm leaning very heavily in the direction of going back to Content.ToString() for this and simply logging a warning for the unsupported platforms.

Sounds good! 

For .net maui i think we'll need to reconcile this and visual a bit
https://github.com/xamarin/Xamarin.Forms/pull/11628#issuecomment-674422715

Because it feels a bit like overlapping APIs but for now and XF 5.0 we probably don't need to worry about it much

davidortinau:
Been working with this PR and have a few issues that are either my lack of understanding or perhaps things not yet implemented.

1. I cannot assign a control template to a type in a style:

```xaml
<Style TargetType="RadioButton">
    <Setter Property="ControlTemplate" Value="{StaticResource CalendarRadioTemplate}"/>
</Style>
```

2. I cannot seem to tie into the visual states to change the UI of my template state. How should this happen? Is the state Normal and Checked? IsChecked? Is the state being triggered from RadioButton to the control template, or how do I make that work?

```xaml
<ControlTemplate x:Key="CalendarRadioTemplate">
            <Frame BorderColor="#F3F2F1" BackgroundColor="#F3F2F1" HasShadow="False" HeightRequest="100" WidthRequest="100" HorizontalOptions="Start" VerticalOptions="Start" Padding="0">
                    <VisualStateGroupList>
                        <VisualStateGroup x:Name="CommonStates">
                            <VisualState x:Name="Normal">
                                <VisualState.Setters>
                                    <Setter Property="BackgroundColor" Value="#f3f2f1"/>
                                    <Setter Property="BorderColor" Value="#f3f2f1"/>
                                </VisualState.Setters>
                            </VisualState>
                            <VisualState x:Name="Checked">
                                <VisualState.Setters>
                                    <Setter Property="BorderColor" Value="#FF3300"/>
                                </VisualState.Setters>
                            </VisualState>
                        </VisualStateGroup>
                    </VisualStateGroupList>
                <Grid Margin="4" WidthRequest="100">

                    <Grid WidthRequest="18" HeightRequest="18" HorizontalOptions="End" VerticalOptions="Start">
                        <Ellipse Stroke="Blue" WidthRequest="16" HeightRequest="16" StrokeThickness="0.5" VerticalOptions="Center" HorizontalOptions="Center" Fill="White" />
                        <Ellipse WidthRequest="8" HeightRequest="8" Fill="Blue" VerticalOptions="Center" HorizontalOptions="Center" />
                    </Grid>
                    <ContentPresenter></ContentPresenter>
                </Grid>
            </Frame>
        </ControlTemplate>
```

rmarinho:
Uitests were failing all in the Viewtests,  heads up @hartez i rebased and going to try again .

hartez:
@davidortinau Okay, these latest changes should address your issues with the control templates. They should be settable from Styles now. The Control Gallery has an example which basically uses the same template you posted upthread.

This is the modified version used in CG:

```
<ControlTemplate x:Key="CalendarRadioTemplate">
	<Frame BorderColor="#F3F2F1" BackgroundColor="#F3F2F1" HasShadow="False" 
		   HeightRequest="100" WidthRequest="100" HorizontalOptions="Start" VerticalOptions="Start" Padding="0">

			<VisualStateGroupList>
				<VisualStateGroup x:Name="CheckedStates">

					<VisualState x:Name="Checked">
						<VisualState.Setters>
							<Setter Property="BorderColor" Value="#FF3300"/>
							<Setter TargetName="Check" Property="Opacity" Value="1"/>
						</VisualState.Setters>
					</VisualState>
					
					<VisualState x:Name="Unchecked">
						<VisualState.Setters>
							<Setter Property="BackgroundColor" Value="#f3f2f1"/>
							<Setter Property="BorderColor" Value="#f3f2f1"/>
							<Setter TargetName="Check" Property="Opacity" Value="0"/>
						</VisualState.Setters>
					</VisualState>
					
				</VisualStateGroup>
			</VisualStateGroupList>

		<Grid Margin="4" WidthRequest="100">
			<Grid WidthRequest="18" HeightRequest="18" HorizontalOptions="End" VerticalOptions="Start">
				<Ellipse Stroke="Blue" WidthRequest="16" HeightRequest="16" StrokeThickness="0.5" VerticalOptions="Center" HorizontalOptions="Center" Fill="White" />
				<Ellipse x:Name="Check" WidthRequest="8" HeightRequest="8" Fill="Blue" VerticalOptions="Center" HorizontalOptions="Center" />
			</Grid>
			<ContentPresenter></ContentPresenter>
		</Grid>
	</Frame>
</ControlTemplate>
```

hartez:
An example of further customization - we can also modify properties like rotation:
```
<VisualStateGroup x:Name="CheckedStates">
	<VisualState x:Name="Checked">
		<VisualState.Setters>
			<Setter Property="BorderColor" Value="#FF3300"/>
			<Setter TargetName="Check" Property="Opacity" Value="1"/>
			<Setter TargetName="Check" Property="Rotation" Value="45"/>
		</VisualState.Setters>
	</VisualState>
	
	<VisualState x:Name="Unchecked">
		<VisualState.Setters>
			<Setter Property="BackgroundColor" Value="#f3f2f1"/>
			<Setter Property="BorderColor" Value="#f3f2f1"/>
			<Setter TargetName="Check" Property="Opacity" Value="0.3"/>
			<Setter TargetName="Check" Property="Rotation" Value="0"/>
		</VisualState.Setters>
	</VisualState>
</VisualStateGroup>
```

mattleibow:
This is looking good!!!

rmarinho:
Failing tests were not related.

rmarinho (approved):

PR #11682, opened 2020-08-07
[WIP] GH-7015 - Implemented handle for Placeholder Image

### LINKED ISSUE #7015: Image.ErrorSource and Image.LoadingSource

pictos:
### Description of Change ###

Added a Bindable Property called `PlaceholderImage`. If the `ImageSource `can't be loaded, the Image control will show the `PlaceholderImage`.

### Issues Resolved ### 

- fix #7015

### API Changes ###

Added:
 - `ImageSource Image.LoadingPlaceholder { get; set; }` //Bindable Property
 - `ImageSource Image.ErrorPlaceholder{ get; set; }` //Bindable Property
Changed:
 
 Removed:
 
 None

### Platforms Affected ### 

- [ ] Core/XAML (all platforms) 
- [ ]  iOS 
- [ ]  Android
- [ ]  UWP

### PR Checklist ###


pictos:
@PureWeen, this PR is ready for review 🎉.

Xamarin.Forms.Core.UnitTests/ImageTests.cs
```diff
@@ -141,6 +141,62 @@ public void TestSource ()
...
+			Assert.IsNull(image.LoadingSource);
+			bool signaled = false;
+
+			image.PropertyChanged += (_, e) =>
+			{
+				if (e.PropertyName == Image.LoadingSourceProperty.PropertyName)
+					signaled = true;
+			};
+
+			var errorSource = ImageSource.FromFile("error.png");
```
PureWeen, line 191: Change this name to loadingSource

Also can you have this same type of test on the other elements? (Button, ImageButton, etc...)

I fixed a copy/paste BindableProperty error on Button that I think would have been picked up by a test like this on Button

Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
```diff
@@ -65,8 +67,10 @@ internal static class ImageViewExtensions
 							using (var drawable = await imageView.Context.GetFormsDrawableAsync(newImageSource))
 							{
 								// only set the image if we are still on the same one
-								if (!imageView.IsDisposed() && SourceIsNotChanged(newView, newImageSource))
+								if (!imageView.IsDisposed() && SourceIsNotChanged(newView, newImageSource) && drawable != null)
 									imageView.SetImageDrawable(drawable);
+								else
```
PureWeen, line 72: should this be

```C#
else if (!imageView.IsDisposed() && SourceIsNotChanged(newView, newImageSource) && drawable == null)
									await imageView.SetImagePlaceholder(newView.ErrorSource);
```

- If the imageView is disposed we don't want to set anything
- If the source has changed then we don't want to do anything and there's an update pending
- we only want to update if those two conditions are still true and drawable == null right?

so maybe this?
```C#
if (!imageView.IsDisposed() && SourceIsNotChanged(newView, newImageSource))
								{
									if(drawable == null)
										await imageView.SetImagePlaceholder(newView.ErrorSource);
									else
										imageView.SetImageDrawable(drawable);
								}
```

Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
```diff
@@ -103,6 +107,16 @@ bool SourceIsNotChanged(IImageElement imageElement, ImageSource imageSource)
 			}
 		}
 
+		public static async Task SetImagePlaceholder(this AImageView imageView, ImageSource placeholder)
+		{
+			using (var drawable = await imageView.Context.GetFormsDrawableAsync(placeholder))
+			{
+				// only set the image if we are still on the same one
+				if (!imageView.IsDisposed())
```
PureWeen, line 115: This if statement will also need this check `SourceIsNotChanged(newView, newImageSource)`

Maybe move this check `!imageView.IsDisposed() && SourceIsNotChanged(newView, newImageSource)`

into an extension method? `ShouldStillSetImage`

And then reuse that both places?

Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
```diff
@@ -53,7 +55,7 @@ internal static class ImageViewExtensions
 							animation = await animationHandler.LoadImageAnimationAsync(newImageSource, imageView.Context);
 					}
 
-					if(animation == null)
+					if (animation == null)
```
PureWeen, line 58: We should still set the `ErrorSource` for an animation if the animation fails to load yes?

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7015.cs
```diff
@@ -0,0 +1,75 @@
...
+			var button = new Button
+			{
+				Text = "Change image",
+				VerticalOptions = LayoutOptions.CenterAndExpand,
+				BackgroundColor = Color.Fuchsia
+			};
+
+			button.Clicked += async (_, __) =>
+			{
+				var s = image.Source.ToString();
```
PureWeen, line 49: Can you add descriptions here as I click the button so I know if what I'm seeing is the correct thing?

For example
"Image is being set to an invalid url so you should see a bank which is the error source"

Xamarin.Forms.Platform.UAP/ImageElementManager.cs
```diff
@@ -145,20 +145,27 @@ public static async Task UpdateSource(IImageVisualElementRenderer renderer)
 			}
 
 			var imageController = Element as IImageController;
+			
+			if (imageElement.LoadingSource != null)
+			{
+				var imageSource = await imageElement.LoadingSource.ToWindowsImageSourceAsync();
+				renderer.SetImage(imageSource);
```
PureWeen, line 152: It looks like UWP is missing the checks that iOS and Android have to make sure the source hasn't changed :-/ 

Can you add those checks in ?

Xamarin.Forms.Platform.Android/FastRenderers/ImageElementManager.cs
```diff
@@ -146,6 +146,7 @@ async static Task TryUpdateBitmap(IImageRendererController rendererController, I
 			catch (Exception ex)
 			{
 				Log.Warning(nameof(ImageElementManager), "Error loading image: {0}", ex);
+				await Control.SetImagePlaceholder(newImage.ErrorSource).ConfigureAwait(false);
```
PureWeen, line 149: I think there's a version of this that needs to be adding to UWP and iOS as well

UWP and iOS both have similar places where this could be set

PureWeen (changes requested):

Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
pictos: I can't reproduce this... Do you know how can I check if the animation fails?

Xamarin.Forms.Platform.Android/FastRenderers/ImageElementManager.cs
pictos: @PureWeen Maybe it's better to remove it. Because the way that `ImageElementManager` works on Android is different from iOS and UWP. On Android, we set the Image using an [external method ](https://github.com/xamarin/Xamarin.Forms/blob/4a2b823045a48b704d7cea561f3ed59bdf9f5c0f/Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs#L24) and we have an isolated [method to set the placeholder](https://github.com/xamarin/Xamarin.Forms/blob/4a2b823045a48b704d7cea561f3ed59bdf9f5c0f/Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs#L110), this doesn't occur on UWP and iOS.

For UWP, if the `ImageSource` doesn't exist or something went wrong the value will be `null` if that is the case, we will set the `ErrorSource`, as you can[ see here](https://github.com/xamarin/Xamarin.Forms/pull/11682/files#diff-8d87756798298b7c41f0e38383d8b106R160)
For iOS is similar also, as you can see [here](https://github.com/xamarin/Xamarin.Forms/pull/11682/files#diff-e7e02ce38379deb0c027bd4e04e17265R239) 

If you have any thoughts about it, please let me know.

Xamarin.Forms.Core/Button.cs
```diff
@@ -138,6 +142,23 @@ public ImageSource ImageSource
 			set { SetValue(ImageSourceProperty, value); }
 		}
 
+
+
+		[TypeConverter(typeof(ImageSourceConverter))]
+		public ImageSource LoadingSource
```
PureWeen, line 148: For now can we not add these to Button.cs? Just use an explicit interface on these for the time being and return null

PureWeen (changes requested):

Xamarin.Forms.Core/Button.cs
pictos: Done

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue7015.cs
```diff
@@ -0,0 +1,93 @@
...
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 7015, "Image placeholder", PlatformAffected.All)]
+	public class Issue7015 : TestContentPage
+	{
+		Image image;
+		CheckBox checkBox;
+		ImageButton imageButton;
+
+		string url = "https://cdn.wallpaperhub.app/cloudcache/0/1/7/7/1/4/0177141d23e1d77489fa02a20f66302fe01033fd.jpg";
```
PureWeen, line 18: Can you change these two images to images in our gallery?

Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
```diff
@@ -33,12 +32,15 @@ internal static class ImageViewExtensions
 			if (newImageSource != null && Equals(previousImageSource, newImageSource))
 				return;
 
-			imageController?.SetIsLoading(true);
```
PureWeen, line 36: Did you mean to move this?

Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
PureWeen: Placeholder here still needs to check if the source has changed and view hasn't been disposed

Xamarin.Forms.Platform.Android/FastRenderers/ImageElementManager.cs
PureWeen: Can we move this to here?

https://github.com/xamarin/Xamarin.Forms/blob/4a2b823045a48b704d7cea561f3ed59bdf9f5c0f/Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs#L98

And if the source isn't null and nothing got set then set it to the error source?

PureWeen (changes requested):

Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
pictos: Yes, I moved this to line 42. In my head, I think that we should `SetIsLoading(true)` after we set the LoadingPlaceholder. If you don't like this idea, I can move this call to the top

Xamarin.Forms.Platform.Android/FastRenderers/ImageElementManager.cs
pictos: > And if the source isn't null and nothing got set then set it to the error source?

@PureWeen Are you saying at the end of the method? Something like this...

```csharp

//line 96 Xamarin.Forms.Platform.Android.ImageViewExtensions method
      finally
			{
				// only mark as finished if we are still working on the same image
				if (SourceIsNotChanged(newView, newImageSource))
				{
					imageController?.SetIsLoading(false);
					imageController?.NativeSizeChanged();
				}
                 // Add this at the end to see if we have an image in loaded
				if (imageView.Drawable != null)
				{
					await imageView.SetImagePlaceholder(newView?.ErrorSource);
				}
			}
```

Xamarin.Forms.Platform.Android/Extensions/ImageViewExtensions.cs
PureWeen: I think it makes sense to still be here. 

Placeholder => Source => Error 

I think are all part of the same Loading operation

samhouts:
Fails to build for macOS:
`D:\agent\1\s\Xamarin.Forms.Platform.iOS\Renderers\ImageElementManager.cs(260,68): error CS0246: The type or namespace name 'UIImageView' could not be found (are you missing a using directive or an assembly reference?)`

pictos:
@samhouts can you run the build again?

PureWeen (changes requested):
Extensions\ImageViewExtensions.cs(51,36): error CS0103: The name 'animation' does not exist in the current context [D:\agent\1\s\Xamarin.Forms.Platform.Android\Xamarin.Forms.Platform.Android.csproj]

pictos:
@PureWeen can you trigger the build again, please? I tested on UWP as well, and is good to go

jsuarezruiz:
@pictos Could you rebase to fix the current conflict?. Thanks!.

jsuarezruiz:
I did rebase myself, it is not necessary!

pictos:
Thanks for that @jsuarezruiz

rmarinho:
Hey @pictos  some uitests are failing , can you check them ? 

Screenshot 2020-12-14 at 23 10 47

pictos:
Hey @rmarinho for sure, I'll do that by tomorrow.

pictos:
@rmarinho I just saw now that the errors are on iOS... I can't that right now, because of mac access. I'm sorry about that :( . I'm not sure, but I guess that I'll be able to run next week. Please let me know if that works for you, and I'm sorry again.

jfversluis:
Thanks for all your efforts here @pictos as you know by now, this won't go into Forms anymore. I would love to have this for .NET MAUI though! I've added a label so we can revisit later. Also feel free to open a spec/proposal for this on the .NET MAUI repo if there isn't one already.

Thanks!

PR #11741, opened 2020-08-11
Fix for issue #10307, embedded fonts not working on UWP

### LINKED ISSUE #10307: [Bug] ExportFont not working on UWP
### Description

Custom fonts are not showing up in UWP when the new ExportFontAttribute is used.

### Steps to Reproduce

1. Add a TTF or OTF file from Google Fonts, FontAwesome, icomoon, etc. as an embedded resource in your shared project
2. Add the ExportFont attribute to your assembly. e.g. `[assembly: ExportFont("FA5Regular.otf", Alias = "FontAwesome")]`
3. Reference the font family in XAML markup
4. Compile and run the UWP app

### Expected Behavior

Custom font should be used.

### Actual Behavior

Custom font is used in iOS and Android, but not UWP.

### Basic Information

- Version with issue: 4.5.0.617
- Platform Target Frameworks: 
  - iOS:  13.4
  - Android: 9.0
  - UWP:  18362

### Reproduction Link

https://github.com/jfversluis/EmbeddedFontsSample

activa:
Embedded fonts on UWP don't work when the font name does not match the file name of the font, which is the case with many Google fonts (and FontAwesome icons)

To fix this, the font family name has to be extracted from the font file and appended to the font name when the font is created from the font file.

### Issues Resolved ### 

- fixes #10307 

### API Changes ###

None

### Platforms Affected ### 

- UWP

None (fixes bug #10307)

Before:
ApplicationFrameHost_2020-08-10_19-42-17

After:
ApplicationFrameHost_2020-08-10_19-38-44

Not applicable

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/FontExtensions.cs
+++ b/Xamarin.Forms.Platform.UAP/FontExtensions.cs
@@ -111,10 +111,10 @@ namespace Xamarin.Forms.Platform.UWP
 			var (hasFontAlias, fontPostScriptName) = FontRegistrar.HasFont(fontFamily);
 			if (hasFontAlias)
 			{
+				var fontName = FontRegistrar.GetFontName(fontFamily);
 				var file = FontFile.FromString(IOPath.GetFileName(fontPostScriptName));
-				var formated = $"{fontPostScriptName}#{file.GetPostScriptNameWithSpaces()}";
-				yield return formated;
-				yield return fontFamily;
+				var formatted = $"{fontPostScriptName}#{fontName ?? file.GetPostScriptNameWithSpaces()}";
+				yield return formatted;
 				yield break;
 			}
 
@@ -133,8 +133,8 @@ namespace Xamarin.Forms.Platform.UWP
 				var (hasFont, filePath) = FontRegistrar.HasFont(fontFile.FileNameWithExtension());
 				if (hasFont)
 				{
-					var formated = $"{filePath}#{fontFile.GetPostScriptNameWithSpaces()}";
-					yield return formated;
+					var formatted = $"{filePath}#{fontFile.GetPostScriptNameWithSpaces()}";
+					yield return formatted;
 					yield break;
 				}
 				else
--- a/Xamarin.Forms.Platform.UAP/FontExtensions.cs
+++ b/Xamarin.Forms.Platform.UAP/FontExtensions.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Graphics.Canvas.Text;
 using Windows.UI.Text;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Media;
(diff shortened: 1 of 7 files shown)


activa:
This would require a documentation change on embedded fonts. The `ExportFont `attribute needs the property `FontName` when the name of the font is different than the file name.

Xamarin.Forms.Platform.UAP/FontExtensions.cs
```diff
@@ -111,10 +111,10 @@ static IEnumerable<string> GetAllFontPossibilities(string fontFamily)
 			var (hasFontAlias, fontPostScriptName) = FontRegistrar.HasFont(fontFamily);
 			if (hasFontAlias)
 			{
+				var fontName = FontRegistrar.GetFontName(fontFamily);
 				var file = FontFile.FromString(IOPath.GetFileName(fontPostScriptName));
-				var formated = $"{fontPostScriptName}#{file.GetPostScriptNameWithSpaces()}";
-				yield return formated;
-				yield return fontFamily;
```
activa, line 117: Note: this line was not needed when the font was found in the FontRegistrar. In fact, it caused problems when setting FontFamily in a CanvasTextFormat object (see FontImageSourceHandler.cs)

activa:
@samhouts Is anything else needed before this can be reviewed?

samhouts:
@activa This looks great! Since this is an API change, can we move it to 5.0.0? Thanks!

mattleibow:
Hey wait there... Before we start moving this "bug" 😄  

@activa I was thinking that I probably wouldn't want to scratch around in files to find things and add names and all that... seems like too much work 😆 

So I had a look at maybe trying to just ask the font file what the correct name was. Seems this is totally doable without adding any new dependencies or special logic:

I just made use of the already used Win2D:

```csharp
var name = "fa-solid-900.ttf";
var uriString = $"ms-appdata:///local/fonts/{name}";
var uri = new Uri(uriString);
var cfs = new CanvasFontSet(uri);
var familyName = cfs.GetPropertyValues(CanvasFontPropertyIdentifier.FamilyName).FirstOrDefault().Value;

var fullFontName = $"{uriString}#{familyName}";

Content = new TextBlock
{
	Text = "hello \uf0c8",
	VerticalAlignment = VerticalAlignment.Center,
	FontFamily = new FontFamily(fullFontName)
};
```

Maybe you could tweak your PR to do this instead? And, then this will also clean up all the other logic bits when loading fonts.

mattleibow:
Using Win2D will also work on the WPF platform as well. Might be able to use things there too.

mattleibow:
My gosh, why do I have the power to close issues? I can't be trusted! My mouse slipped!

activa:
@mattleibow I'll update the PR with your suggested changes.

activa:
PR was updated. No API changes were introduced and the font family is now extracted from the font file on UWP.

mattleibow (approved):
Yeah! Now that is what I call development! 😎

Xamarin.Forms.Platform.UAP/FontExtensions.cs
```diff
@@ -105,16 +107,27 @@ public static FontFamily ToFontFamily(this string fontFamily)
...
 		}
 
+		static string FindFontFamilyName(string fontFile)
+		{
+			using (var fontSet = new CanvasFontSet(new Uri(fontFile)))
+			{
+				if (fontSet.Fonts.Count == 0)
+					return null;
+
+				return fontSet.GetPropertyValues(CanvasFontPropertyIdentifier.FamilyName).FirstOrDefault().Value;
```
PureWeen, line 117: Will `FirstOrDefault` here always return something?

Does this need a null check?

mattleibow: The result is a struct: https://microsoft.github.io/Win2D/html/T_Microsoft_Graphics_Canvas_Text_CanvasFontProperty.htm

so the `.Value` should be `null`.

activa: What @mattleibow said. Also, I was thinking about caching the return value but it would be pointless because the resulting FontFamily objects are already being cached here.

mattleibow:
Confirmation that it just runs on UWP without any changes to my existing app:
image

@activa you are a beautiful dev! We need more like you in the world! Thanks for getting this PR in! Now we just got to get the Team to merge this for the next SR!

mattleibow:
So much yay!!! My app will now look beautiful again. Thanks!

activa:
@samhouts Is there an issue with this PR? The "UWP-BUGZILLA" check has been in progress for 5 days.

PR #11752, opened 2020-08-11
[UWP] Fixes for font when using MultiWindow

### LINKED ISSUE #11705: [Bug] [UWP] Setting the FontFamily crashes a multi-window app
### Description

Since #2432, Xamarin.Forms supports creating apps with multiple windows in UWP.
However, if we use the same custom font on these windows, then we get an exception when trying to create the secondary windows.

### Steps to Reproduce

[The Control Gallery already has a multi-window sample](https://github.com/xamarin/Xamarin.Forms/blob/29b92aa1fc2b9d834bb3b95cc224e07d0cc3e9b4/Xamarin.Forms.Controls/CoreGallery.cs#L603), we just need to update it to set a custom FontFamily:

1. Update the content page that is created for these secondary windows:
https://github.com/xamarin/Xamarin.Forms/blob/29b92aa1fc2b9d834bb3b95cc224e07d0cc3e9b4/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2482.cs#L88
After this line, add:

```C#
if (Device.RuntimePlatform == Device.UWP)
    layout.Children.Add(new Label { Text = "\xE76E", FontFamily = "Segoe MDL2 Assets", FontSize = 32 });
```

2. Run the Control Gallery app and click on the "Open Secondary Window" button twice.

### Expected Behavior

The app should create as many windows as we want.

### Actual Behavior

The app crashes when we try to create its third window.

### Basic Information

- Version with issue: 4.8.0.1269
- IDE: Visual Studio 2019 16.7
- Platform Target Frameworks:
  - UWP:  18362

### Solution

This issue is caused by the fact that fonts are cached; so they end up being shared by all windows (which is not allowed):

https://github.com/xamarin/Xamarin.Forms/blob/719fc7a604ff0cce8922d717c99bfb0fa17e35e0/Xamarin.Forms.Platform.UAP/FontExtensions.cs#L14

The solution should be to change that cache to be thread-static:

```C#
[ThreadStatic]
static Dictionary<string, FontFamily> FontFamilies;
```

Then, instantiate it before this line:

https://github.com/xamarin/Xamarin.Forms/blob/719fc7a604ff0cce8922d717c99bfb0fa17e35e0/Xamarin.Forms.Platform.UAP/FontExtensions.cs#L96

```C#
if (FontFamilies == null)
    FontFamilies = new Dictionary<string, FontFamily>();
```

That should be enough.

rmarinho:
### Description of Change ###

When using multiple windows we need to take in account that each CoreApplicationView has it's own thread so caching items doesn't work if they are shared. Like FontFamilies for example

We also need to relayout our contents when the window is resized.

### Issues Resolved ### 

- fixes #11705

### API Changes ###

 None

### Platforms Affected ### 

- UWP

None

Not applicable

### Testing Procedure ###

Try open 2 our 3 windows from the control gallery 

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.ControlGallery.WindowsUniversal/SecondaryWindowService.cs
+++ b/Xamarin.Forms.ControlGallery.WindowsUniversal/SecondaryWindowService.cs
@@ -26,10 +26,6 @@ namespace Xamarin.Forms.ControlGallery.WindowsUniversal
 				ContentPage instance = (ContentPage)Activator.CreateInstance(pageType);
 				frame.Navigate(instance);
 				Window.Current.Content = frame;
-				Window.Current.SizeChanged += (sender, args) =>
-				{
-					instance.Layout(new Rectangle(0, 0, args.Size.Width, args.Size.Height));
-				};
 				Window.Current.Activate();
 
 				newViewId = ApplicationView.GetForCurrentView().Id;
--- a/Xamarin.Forms.ControlGallery.WindowsUniversal/SecondaryWindowService.cs
+++ b/Xamarin.Forms.ControlGallery.WindowsUniversal/SecondaryWindowService.cs
@@ -26,6 +26,10 @@ namespace Xamarin.Forms.ControlGallery.WindowsUniversal
 				ContentPage instance = (ContentPage)Activator.CreateInstance(pageType);
 				frame.Navigate(instance);
 				Window.Current.Content = frame;
+				Window.Current.SizeChanged += (sender, args) =>
+				{
+					instance.Layout(new Rectangle(0, 0, args.Size.Width, args.Size.Height));
+				};
 				Window.Current.Activate();
 
 				newViewId = ApplicationView.GetForCurrentView().Id;
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2482.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2482.cs
@@ -87,6 +87,9 @@ namespace Xamarin.Forms.Controls.Issues
 			layout.Children.Add(_result);
 			layout.Children.Add(button);
 
+			if (Device.RuntimePlatform == Device.UWP)
+				layout.Children.Add(new Label { Text = "\xE76E", FontFamily = "Segoe MDL2 Assets", FontSize = 32 });
+
 			Content = layout;
 		}
 
--- a/Xamarin.Forms.Platform.UAP/FontExtensions.cs
+++ b/Xamarin.Forms.Platform.UAP/FontExtensions.cs
@@ -11,6 +11,7 @@ namespace Xamarin.Forms.Platform.UWP
 {
 	public static class FontExtensions
 	{
+		[ThreadStatic]
 		static Dictionary<string, FontFamily> FontFamilies = new Dictionary<string, FontFamily>();
 		static double DefaultFontSize = double.NegativeInfinity;
 
@@ -92,6 +93,8 @@ namespace Xamarin.Forms.Platform.UWP
 			if (string.IsNullOrWhiteSpace(fontFamily))
 				return (FontFamily)WApplication.Current.Resources["ContentControlThemeFontFamily"];
 
+			if (FontFamilies == null)
+				FontFamilies = new Dictionary<string, FontFamily>();
 			//Return from Cache!
 			if (FontFamilies.TryGetValue(fontFamily, out var f))
 			{


KPixel:
Regarding the fix for resizing secondary windows, my suggestion was a workaround.

Can you please investigate why that is not an issue for the main window? I would think that they should be resized in the same way.

KPixel:
@rmarinho @hartez Please, only commit the fix for `FontExtensions`.

The issue with resizing windows should be investigated first...

Xamarin.Forms.ControlGallery.WindowsUniversal/SecondaryWindowService.cs
```diff
@@ -26,6 +26,10 @@ public async Task OpenSecondaryWindow(Type pageType)
 				ContentPage instance = (ContentPage)Activator.CreateInstance(pageType);
 				frame.Navigate(instance);
 				Window.Current.Content = frame;
+				Window.Current.SizeChanged += (sender, args) =>
```
hartez, line 29: @KPixel makes a good point - why does this work? We should pull this out and create a separate issue.

rmarinho: Ok i removed that code.. Also another code that works is something like this bellow. But we don't want to create a new Window right? 
```
public class NewSecondaryWindowsPage : WindowsPage
{
	protected override void OnNavigatedTo(Windows.UI.Xaml.Navigation.NavigationEventArgs e)
	{
		var pageType = e.Parameter as Type;
		ContentPage instance = (ContentPage)Activator.CreateInstance(pageType);
		RegisterWindow(instance);
		base.OnNavigatedTo(e);
	}
}

public async Task OpenSecondaryWindow(Type pageType)
{
	CoreApplicationView newView = CoreApplication.CreateNewView();
	int newViewId = 0;
	await newView.Dispatcher.RunAsync(CoreDispatcherPriority.Normal, () =>
	{
		var frame = new Windows.UI.Xaml.Controls.Frame();
		Window.Current.Content = frame;
		frame.Navigate(typeof(NewSecondaryWindowsPage), pageType);
		Window.Current.Activate();
		newViewId = ApplicationView.GetForCurrentView().Id;
	});
	bool viewShown = await ApplicationViewSwitcher.TryShowAsStandaloneAsync(newViewId);
}
```

KPixel: Interesting. I think this points to the root of the issue: [Application.MainPage](https://github.com/xamarin/Xamarin.Forms/blob/29b92aa1fc2b9d834bb3b95cc224e07d0cc3e9b4/Xamarin.Forms.Core/Application.cs#L66)...
I would like to create a new issue to discuss a solution there. Ok?

KPixel:
Let's ship this fix! :)

Xamarin.Forms.ControlGallery.WindowsUniversal/SecondaryWindowService.cs
rmarinho: Yeah the issue from what i understand is here :

https://github.com/xamarin/Xamarin.Forms/blob/719fc7a604ff0cce8922d717c99bfb0fa17e35e0/Xamarin.Forms.Platform.UAP/PageExtensions.cs#L58

We create a page just to fake it as "root" and so the code that listens to size changes here:

https://github.com/xamarin/Xamarin.Forms/blob/3ca1250330e8d22d4544aaa225301d42a8b5ec98/Xamarin.Forms.Platform.UAP/Platform.cs#L91

But the _page is not added to the VisualTree

KPixel: There is also the following code which expects all windows to have a specific structure:

https://github.com/xamarin/Xamarin.Forms/blob/3ca1250330e8d22d4544aaa225301d42a8b5ec98/Xamarin.Forms.Platform.UAP/Platform.cs#L51-L59

So, your code above seems like the expected way to initialize a secondary window.

I've updated my app to follow this approach, and it work.

KPixel: This means that the "fix" for the resize issue is to update `OpenSecondaryWindow()` in the Control Gallery app.

Also, the [PageExtensions](https://github.com/xamarin/Xamarin.Forms/blob/aa0b9c66f37b05f6c290a68aba39a0fc0e51aa3d/Xamarin.Forms.Platform.UAP/PageExtensions.cs#L44) class was a trap. From what I can see, in this entire repo, it is used in two scenarios:

1. For `FormsEmbeddedPageWrapper` pages (and there is no sample/test for that).
2. For the [UWP version of DualScreen](https://github.com/xamarin/Xamarin.Forms/blob/aa0b9c66f37b05f6c290a68aba39a0fc0e51aa3d/Xamarin.Forms.DualScreen/DualScreenHelper.uwp.cs#L43). And I think that code should be updated to follow the `WindowsPage` approach, otherwise, it will suffer from the same resize bug.

KPixel:
Hello,

Can we please merge this simple fix before 5.0?

It is ready to go!

jsuarezruiz (approved):

PR #11805, opened 2020-08-16
[iOS] Fix Keyboard Inset when Editor is one of the last components in a collection view

### LINKED ISSUE #9691: [Bug][OS Keyboard hides CollectionView elements on iOS
### Steps to Reproduce

1. 
2. 
3. 

### Basic Information

- Version with issue:
- Last known good version:
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Nuget Packages:
- Affected Devices:

mjmostachetti:
### Description of Change ###

I removed this code that disabled the native collectionview from adjusting correctly.
`CollectionView.ContentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentBehavior.Never;`

There was a comment that was above the line I removed that described the reason for having that line of code to begin with. I could not see any visible change when flipping my simulator into landscape mode w/ and w/o that line of code above - except for the fact that removing the code allowed me to see my editor's text. If someone could help explain, or include some repro steps in why it was there in the first place, that would be great as there may be side affects I'm unaware of from removing that code.

### Issues Resolved ### 
- fixes #9691

### API Changes ###
 
 None

### Platforms Affected ### 
- iOS

The user can now see what they've typed in an editor, other keyboard components.

Not applicable

### Testing Procedure ###
Go to the Issue9691 Test Case and ensure that when you click into the last Editor in the collection view, the view scrolls up and you can view the text you are entering into the editor.

### PR Checklist ###


samhouts (approved):

mjmostachetti:
@samhouts @hartez Any thoughts on this one?

hartez:
That `CollectionView.ContentInsetAdjustmentBehavior` does have a purpose; without it, horizontal grid layouts in CollectionViews break pretty badly. So your fix isn't quite right.

I'll take a look at this on Monday - it looks like we might have a conflict between the CollectionView's handling of safe area and the overall handling in Forms, for which `UIScrollViewContentInsetAdjustmentBehavior.Never` is just a band-aid. The automatic behavior (what you get when you remove that line of code) *is* what we want for handling keyboard shifting. So the fix probably lies elsewhere. 

But in the worst case (where we can't reconcile the two), we can add a keyboard listener to the CollectionView. The listener can tell us when the keyboard is popping up and allow us to adjust the insets accordingly.

hartez:
Adding a do-not-merge label until I can get back to this next week.

mjmostachetti:
@hartez Thanks for taking a look. I appreciate it.

jfversluis:
Unfortunately grew stale and not wise to add this to Forms anymore at this point in time. Thanks for all the time and effort here!

PR #11949, opened 2020-08-28
[iOS] Block the scroll if is swiping a SwipeView 

### LINKED ISSUE #11651: [Bug] [Regression] [iOS] Swipeview still swipes while scrolling
### Description

On iOS with version 4.7.1239 when swiping an item in a scrollable list the item can both be swiped and scrolled at the same time.  This means that swiping down and slightly one way causes the execute action to be triggered.  This is a significant usability problem if Swipe Mode is Execute.

This does not happen on Android and does not happen on Forms 4.4.0.991220-pre3 for either.  Therefore it is a regression on iOS.

### Steps to Reproduce

1. run https://github.com/jsuarezruiz/FavFighters as is (Forms 4.4.0.991220-pre3).  Both Android and iOS are fine (one cannot swipe and scroll at the same time).
2. upgrade that project to 4.7.1239 and add SetFlags("SwipeView_Experimental") to AppDelegate.cs and MainActivity.cs.
3. now notice that Android remains ok, but on iOS one can both swipe and scroll at the same time.

### Expected Behavior

As noted above, it should not be possible to swipe and scroll at the same time.

### Actual Behavior

On iOS, one can both swipe and scroll at the same time.

Also note that there are drawing issues on 4.7 which might be described in other issue records.

### Basic Information

- Version with issue: 4.7.1239
- Last known good version: 4.4.0.991220-pre3 (may be more)
- IDE: VS 2019
- Platform Target Frameworks: 
  - iOS: 13.6
  - Android: 9.0

### LINKED ISSUE #11932: [Bug] [iOS] When in CollectionView's ItemTemplate, the SwipeView emits swipe events together with scroll events
### Description
On iOS, SwipeView, when in CollectionView's ItemTemplate, doesn't take all of the gestures during swipe and as soon as the user starts swiping, the CollectionView is very easily scrollable, which is annoying and is not the intended behavior. All apps with swipes done right don't emit multiple gestures during swiping. (e.g. the Mail app on iOS)

This is happening only on iOS. The SwipeView behaves correctly on Android.
I've attached both a sample project & a gif illustration of the issue.

### Steps to Reproduce
1. Create a CollectionView with a SwipeView in its ItemTemplate
2. Add many dummy rows so that the CollectionView can be scrolled
3. Try to swipe a row and at the same to scroll the CollectionView (swipe diagonally)

### Expected Behavior
The swipe gestures shouldn't fight for which event to be executed. As soon as the swipe starts, there shouldn't be any scrolling at all.

### Actual Behavior
The swipe gestures & the scroll gestures are happening simultaneously, which is very confusing and occasionally it's often not the indented behaviour that the user has been expecting.

### Basic Information
- Version with issue: 4.8.0.1269
- IDE: Rider for macOS
- Nuget Packages: None
- Affected Devices: iOS devices

### Screenshots
Aug-26-2020 19-51-56

### Reproduction Link
[DemoSwipeIssue.zip](https://github.com/xamarin/Xamarin.Forms/files/5131213/DemoSwipeIssue.zip)

### Workaround
No workaround.

jsuarezruiz:
### Description of Change ###

Block the scroll if is swiping a SwipeView on iOS (same behavior as in Android).

### Issues Resolved ### 

- fixes #11651 
- fixes #11932

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS

None

fix11949

### Testing Procedure ###
Launch Core Gallery and navigate to the SwipeView Galleries. Select the CollectionView (or any other gallery using a parent with scroll like a ListView or ScrollView), and scroll. Verify that you can scroll without problems, and also open a SwipeView. Try to scroll swiping the SwipeView and must not be possible.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/SwipeViewRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/SwipeViewRenderer.cs
@@ -13,7 +13,7 @@ namespace Xamarin.Forms.Platform.iOS
 {
 	public class SwipeViewRenderer : ViewRenderer<SwipeView, UIView>
 	{
-		const float MinimumOpenSwipeThresholdPercentage = 0.2f; // 20%
+		const float MinimumOpenSwipeThresholdPercentage = 0.15f; // 15%
 		const float OpenSwipeThresholdPercentage = 0.6f; // 60%
 		const double SwipeThreshold = 250;
 		const double SwipeItemWidth = 100;
--- a/Xamarin.Forms.Platform.iOS/Renderers/SwipeViewRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/SwipeViewRenderer.cs
@@ -13,6 +13,7 @@ namespace Xamarin.Forms.Platform.iOS
 {
 	public class SwipeViewRenderer : ViewRenderer<SwipeView, UIView>
 	{
+		const float MinimumOpenSwipeThresholdPercentage = 0.2f; // 20%
 		const float OpenSwipeThresholdPercentage = 0.6f; // 60%
 		const double SwipeThreshold = 250;
 		const double SwipeItemWidth = 100;
@@ -38,6 +39,7 @@ namespace Xamarin.Forms.Platform.iOS
 		double _previousScrollY;
 		int _previousFirstVisibleIndex;
 		bool _isSwipeEnabled;
+		bool _isScrollEnabled;
 		bool _isResettingSwipe;
 		bool _isOpen;
 		bool _isDisposed;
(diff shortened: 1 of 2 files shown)


Xamarin.Forms.Platform.iOS/Renderers/SwipeViewRenderer.cs
```diff
@@ -13,6 +13,7 @@ namespace Xamarin.Forms.Platform.iOS
 {
 	public class SwipeViewRenderer : ViewRenderer<SwipeView, UIView>
 	{
+		const float MinimumOpenSwipeThresholdPercentage = 0.15f; // 15%
```
jsuarezruiz, line 16: Scrolling quickly, we can slightly open the SwipeView. For that reason, we do not block the scroll unless the SwipeView is opened 15%.

rmarinho (approved):

PR #12043, opened 2020-09-07
[X] Add a TrySetPropertyValue overload for HR

StephaneDelcroix:
### Description of Change ###

Add a `TrySetPropertyValue()` that doesn't require the parsing and hydration context. For most case, it will be enough if the serviceProvider contains an `IValueConverterProvider` and a default `IValueConverterProvider`. (see https://github.com/xamarin/Xamarin.Forms/blob/main/Xamarin.Forms.Xaml/XamlServiceProvider.cs). More complex cases might require extra services.

### Issues Resolved ### 

- fixes [AB#1192221](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1192221)

### API Changes ###
 
 None

### Platforms Affected ### 

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Xaml.UnitTests/Issues/Bz47950.xaml.cs
+++ b/Xamarin.Forms.Xaml.UnitTests/Issues/Bz47950.xaml.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using NUnit.Framework;
 using Xamarin.Forms;
+using Xamarin.Forms.Core.UnitTests;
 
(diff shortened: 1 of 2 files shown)


samhouts (approved):

StephaneDelcroix:
we could consider retargeting to 4.8, but that would require a manual check of all the callers of the modified API (changing HydrationContext to object doesn't trigger a compiler error)

PR #12099, opened 2020-09-11
Set Backdrop Color after Pane Opened

### LINKED ISSUE #12018: [Bug] FlyoutBackdrop does not work correctly on UWP
### Description
When Shell menu is opened for the first time, FlyoutBackdrop is ignored and the default color is used instead. Only on the second and the subsequent openings, the FlyoutBackdrop color is applied. This issue is only on UWP. FlyoutBackdrop works well on Android.

### Steps to Reproduce

1. Create a new empty project in VS, add shell, update all Nugets, and set FlyoutBackdrop to red color (or use the included repro project).
2. Click the menu button to open the flyout menu.
3. The backdrop should be red but it is silver instead.
4. Close the menu and open it again.
5. Now the backdrop is red.

### Expected Behavior
When the menu is opened for the first time, the FlyoutBackdrop color (red in the example) should be used.

### Actual Behavior
When the menu is opened for the first time , the default color (silver) is used and the value of FlyoutBackdrop is ignored.

### Basic Information

- Version with issue: Xamarin.Forms 4.8.0.1364
- Last known good version:
- IDE: VS 2019
- Platform Target Frameworks:
  - UWP: 18362
- Nuget Packages:  Xamarin.Forms 4.8.0.1364, Xamarin.Essentials1.5.3.2, NETStandardLibrary 2.0.3, Microsoft.NETCore.UniversalWindowsPlatform 6.2.10

### Screenshots
Menu is opened for the first time:
UwpShellBackdropBug1

Menu is opened for the second time:
UwpShellBackdropBug2

### Reproduction Link

You can use this repro project. It contains both UWP and Android version so you can see that it works on Android but not on UWP.
[ShellTest.zip](https://github.com/xamarin/Xamarin.Forms/files/5171028/ShellTest.zip)

### Workaround

No workaround.

PureWeen:
### Description of Change ###

On the first initial opening of the side pane UWP resets the light layer back to the default color. This sets the color after the pane has opened just to ensure the back drop color has propagated to the light layer

### Issues Resolved ### 

- fixes #12018 

### Platforms Affected ### 
- UWP

### Testing Procedure ###
- test setting back drop layer on uwp

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/Shell/ShellRenderer.cs
+++ b/Xamarin.Forms.Platform.UAP/Shell/ShellRenderer.cs
@@ -42,6 +42,10 @@ namespace Xamarin.Forms.Platform.UWP
 				PaneClosing += (s, e) => OnPaneClosed();
 			if (ApiInformation.IsEventPresent("Windows.UI.Xaml.Controls.NavigationView", "PaneOpening"))
 				PaneOpening += (s, e) => OnPaneOpening();
+			if (ApiInformation.IsEventPresent("Windows.UI.Xaml.Controls.NavigationView", "PaneOpened"))
+				PaneOpened += (s, e) => OnPaneOpened();
+
+
 			ItemInvoked += OnMenuItemInvoked;
 			BackRequested += OnBackRequested;
 			Style = Windows.UI.Xaml.Application.Current.Resources["ShellNavigationView"] as Windows.UI.Xaml.Style;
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.UAP/Shell/ShellRenderer.cs
```diff
@@ -42,6 +42,10 @@ public ShellRenderer()
 				PaneClosing += (s, e) => OnPaneClosed();
 			if (ApiInformation.IsEventPresent("Windows.UI.Xaml.Controls.NavigationView", "PaneOpening"))
 				PaneOpening += (s, e) => OnPaneOpening();
+			if (ApiInformation.IsEventPresent("Windows.UI.Xaml.Controls.NavigationView", "PaneOpened"))
+				PaneOpened += (s, e) => OnPaneOpened();
```
jsuarezruiz, line 46: The changes looks nice but, could we add the unsubscription of these events? Maybe using `OnElementSet`, if `_shell` exists?

jsuarezruiz (changes requested):

Xamarin.Forms.Platform.UAP/Shell/ShellRenderer.cs
PureWeen: 👍

samhouts (approved):

PR #11933, opened 2020-08-26
[Bug] Correct the transformation calculation of views

mattleibow:
### Description of Change ###

When the translation offset for a view is calculated, it incorrectly uses the scale to multiply/divide the actual offset. Side bug, seems WPF does not scale at all.

This PR will correct the math, I hope.

### Issues Resolved ### 

- Fixes #11931

### API Changes ###

 
 None

### Platforms Affected ### 

- iOS: the rotation now happens before the scale, and thus the scale matrix is no longer being rotated
- UWP: the scale no longer affectes the translate
- WPF: added support for scaling along a single axis, and rotate before the scale

### Behavioral/Visual Changes ###

None

### Before/After Screenshots ### 

| Android| iOS | UWP | WPF |
| :-: | :-: | :-: | :-: |
| | | | |
| **BEFORE** | **BEFORE** | **BEFORE** | **BEFORE** |
| android | ios-before | uwp-before | wpf-before |
| **AFTER** | **AFTER** | **AFTER** | **AFTER** |
|  | ios-after | uwp-after | wpf-after |

Not applicable

### PR Checklist ###

- [ ] Targets the correct branch
- [ ] Tests are passing (or failures are unrelated)

PR #12105, opened 2020-09-14
[Tizen] Fix the order of the scale and rotation

rookiejava:
### Description of Change ###

This is addendum of #11933 for Tizen. 
With this PR, I've confirmed that the transformation of views works well as follows.

**BEFORE** | **AFTER**
--- | --- 
 |


mattleibow (approved):

mattleibow:
Thanks!

PR #11815, opened 2020-08-17
Fix input for decimal/float/double and nullable

jonkas2211:
### Description of Change ###

The input for floating numbers had some errors in different cultures.
When using german for example, the decimal seperator is a ',' and not a dot.

You can also use Nullable Types now. Before the Decimal Seperator would be cut off for Nullables.

### Issues Resolved ### 

- fixes #7996 

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

### Behavioral/Visual Changes ###
The user of an app can now use the correct decimal sperator for his language.

### Before/After Screenshots ### 

Not applicable

### Testing Procedure ###
I have tested this on android.
I added a UI Test to the Gallery (G7996).
Unfortunately i cant test this on other platforms right now.
I also added a new Unit Test to prove the behaviour.

### PR Checklist ###

- [ ] Targets the correct branch
- [ ] Tests are passing (or failures are unrelated)

PR #12123, opened 2020-09-15
Respect the UI Culture when converting data bound values

fekberg:
### Description of Change ###

Respect the UI culture when converting data bound values.

If a decimal is bound to an Entry, you cannot use any other decimal separator than `.` (dot) because it uses InvariantCulture in the conversion.

### Issues Resolved ### 

- fixes #

### API Changes ###

 None

### Platforms Affected ### 

- Core/XAML (all platforms)

Not applicable

### Testing Procedure ###
This can easily be replicated by binding an entry to a field like this:

`<Entry Text="{Binding MyNumber}" />`

```
public class MainPageViewModel
{
	public decimal MyNumber { get; set; }
}

{
	public MainPage()
	{
		InitializeComponent();

		CultureInfo.CurrentUICulture = CultureInfo.GetCultureInfo("sv-SE");

		BindingContext = new MainPageViewModel();
	}
}
```

If the value `10,5` is entered it was previously converted into `105` while when the UI Culture is respected this is converted into `10.5`.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/BindingExpression.cs
+++ b/Xamarin.Forms.Core/BindingExpression.cs
@@ -456,7 +456,7 @@ namespace Xamarin.Forms
 
 				convertTo = Nullable.GetUnderlyingType(convertTo) ?? convertTo;
 
-				value = Convert.ChangeType(value, convertTo, CultureInfo.InvariantCulture);
+				value = Convert.ChangeType(value, convertTo, CultureInfo.CurrentUICulture);
 				return true;
 			}
 			catch (Exception ex) when (ex is InvalidCastException || ex is FormatException || ex is InvalidOperationException || ex is OverflowException) {


jonkas2211:
See at PR #11815

fekberg:
@jonkas2211 Excellent, shall we close this PR then?

EmilAlipiev:
wow, I cannot believe Xamarin team used InvariantCulture instead of CurrentCulture and this bug was reported 3 years ago, still open. Dozens of forum and SO posts exists about this issue and Xamarin team is not even taking serious.  Other PR is still open since 27 August how annoying and sad is that please.

jfversluis:
Hey @fekberg sorry that we kept this laying around so long. Thanks so much for your time and effort on this. I have merged the PR (mentioned) by @jonkas2211. If this is still relevant to you, could you please check if this already fixes the problems you were facing and see if this is then still needed on top of that?

PR #12135, opened 2020-09-16
Implement Nullable Reference Types on Handler projects

jsuarezruiz:
### Description of Change ###

Implement Nullable Reference Types on Handler projects.
More information: https://github.com/dotnet/maui/issues/232

### PR Checklist ###

### DIFF
--- a/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/IViewHandler.Android.cs
+++ b/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/IViewHandler.Android.cs
@@ -7,6 +7,6 @@ namespace Xamarin.Platform
 	{
 		void SetContext(Context context);
 
-		AView View { get; }
+		AView? View { get; }
 	}
 }
--- a/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/View/AbstractViewHandler.cs
+++ b/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/View/AbstractViewHandler.cs
@@ -1,6 +1,4 @@
-﻿using Xamarin.Forms;
-
-#if __IOS__
+﻿#if __IOS__
 using NativeView = UIKit.UIView;
 #elif __MACOS__
 using NativeView = AppKit.NSView;
@@ -20,25 +18,25 @@ namespace Xamarin.Platform.Handlers
 		where TNativeView : NativeView
 #endif
 	{
-		protected readonly PropertyMapper defaultMapper;
-		protected PropertyMapper mapper;
+		protected readonly PropertyMapper _defaultMapper;
+		protected PropertyMapper? _mapper;
 		bool _hasContainer;
 		static bool HasSetDefaults;
 
 		protected AbstractViewHandler(PropertyMapper mapper)
 		{
-			defaultMapper = mapper;
+			_defaultMapper = mapper;
 		}
 
 		protected abstract TNativeView CreateView();
 
-		public NativeView View => TypedNativeView;
+		protected TNativeView TypedNativeView { get; private set; } = default!;
 
-		public TNativeView TypedNativeView { get; private set; }
+		protected TVirtualView? VirtualView { get; private set; } = default!;
 
-		protected TVirtualView VirtualView { get; private set; }
+		public NativeView? View => TypedNativeView;
 
-		public object NativeView => TypedNativeView;
+		public object? NativeView => TypedNativeView;
 
 		public virtual void SetView(IView view)
 		{
@@ -51,7 +49,7 @@ namespace Xamarin.Platform.Handlers
 				HasSetDefaults = true;
 			}
 
-			mapper = defaultMapper;
+			_mapper = _defaultMapper;
 
 			if (VirtualView is IPropertyMapperView imv)
 			{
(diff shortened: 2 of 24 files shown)


src/Platform.Handlers/src/Xamarin.Platform.Handlers/Xamarin.Platform.Handlers.csproj
```diff
@@ -1,8 +1,10 @@
 ﻿<Project Sdk="MSBuild.Sdk.Extras">
   <PropertyGroup>
-    <TargetFrameworks>$(TargetFrameworks);netstandard2.0;Xamarin.iOS10;MonoAndroid10.0;Xamarin.Mac20</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1;Xamarin.iOS10;MonoAndroid10.0;Xamarin.Mac20</TargetFrameworks>
     <RootNamespace>Xamarin.Platform</RootNamespace>
     <AssemblyName>Xamarin.Platform</AssemblyName>
+    <LangVersion>8.0</LangVersion>
+    <Nullable>enable</Nullable>
```
jsuarezruiz, line 7: This is the important change!

StephaneDelcroix (approved):

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Xamarin.Platform.Handlers.csproj
StephaneDelcroix: I'd go one step further, enable it in Directory.Build.props, and disable it on projects (almost all) that aren't ready to support it yet. Same effect, different message

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Button/ButtonHandler.iOS.cs
```diff
@@ -9,7 +8,8 @@ public partial class ButtonHandler : AbstractViewHandler<IButton, UIButton>
 
 		public static void MapText(IViewHandler handler, IButton view)
 		{
-			((UIButton)handler.NativeView).SetText(view.Text);
+			if (handler.NativeView is UIButton uIButton)
+				uIButton.SetText(view.Text);
```
StephaneDelcroix, line 12: checking that NativeView isn't null isn't enough, we probably should check wether it's disposed. Or are setting NativeView to null when we dispose it ?

also, why do we need casting ? don't we have a generic type parameter of type UIButton ?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/IViewHandler.Android.cs
```diff
@@ -7,6 +7,6 @@ public interface IAndroidViewHandler : IViewHandler
 	{
 		void SetContext(Context context);
 
-		AView View { get; }
+		AView? View { get; }
```
StephaneDelcroix, line 10: everything would be so simple, or at least simpler, if the View was never null. could we consider this ? (View created as part of the ctor, e.g.)

StephaneDelcroix (changes requested):
making views nullable adds more check than we had, so maybe the view doesn't need to be nullable after all ?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Xamarin.Platform.Handlers.csproj
jsuarezruiz: Nice suggestion but currently can be only enable in the handlers project and 8.0 is already all setup everywhere.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/IViewHandler.Android.cs
PureWeen: Agreed

What in here that's marked nullable can we guarantee won't be null.

Like does a ViewHandler with a null View make any sense?

PureWeen (changes requested):

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/IViewHandler.Android.cs
PureWeen: The views are created after the ctror and can be swapped out

https://github.com/xamarin/Xamarin.Forms/blob/7aedbc7d3e592a531879d2eb05a6daa673e16715/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/View/AbstractViewHandler.cs#L43

I'm assuming we'll need this to be nullable

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/View/AbstractViewHandler.cs
```diff
@@ -20,25 +18,25 @@ public abstract partial class AbstractViewHandler<TVirtualView, TNativeView> : I
...
 		protected AbstractViewHandler(PropertyMapper mapper)
 		{
-			defaultMapper = mapper;
+			_defaultMapper = mapper;
 		}
 
 		protected abstract TNativeView CreateView();
 
-		public NativeView View => TypedNativeView;
+		protected TNativeView TypedNativeView { get; private set; } = default!;
```
PureWeen, line 33: Why did you change the access modifiers on these?

PureWeen (approved):

PR #12163, opened 2020-09-18
Fix WkWebViewRenderer to properly dispose and clear renderers

### LINKED ISSUE #11962: [Bug] iOS Only- 'Cannot access a disposed object. Object name: 'WkWebViewRenderer'.'
### Description
When a Page with WebView is revisited with fresh data is crashing. This does not happen when using Tabbed page.

### Steps to Reproduce
1.  Create a Search Page with a List of items 
2.  On click/tapping the item, navigate to details with a WebView.
3. Navigate back to the Search page
4. Click on the second item to get to the details page.
---> App crashes with  'Cannot access a disposed object. Object name: 'WkWebViewRenderer'.'

### Expected Behavior

App should not crash and display the new Html content in the WebView.

### Actual Behavior
System.ObjectDisposedException: 'Cannot access a disposed object.
Object name: 'WkWebViewRenderer'.'

### Basic Information

- Version with issue:  4.8.0.1269
- Last known good version:
- IDE: 2019
- Platform Target Frameworks: 
  - iOS:  13.6 (Probbably other versions too)

### LINKED ISSUE #12134: WkWebView does not handle cookies consistently
### Description
Setting the cookie to the webview does not seem to work 100% times on iOS (WkWebView).
In our case we have one webview in one page and clicking on the link inside this webview opens a new page with another webview with the same session cookie. In more than 50% of cases the session cookie is not there in the second webview.
Looks like it could be related to the native iOS cookies issues: https://developer.apple.com/forums/thread/99674

### Steps to Reproduce
(Sample project attached)
1. Open the app: it will load "google.com" in one webview and after navigation is done show an alert with cookies
2. Click on "Load another webview" -> this will load one more webview with the same cookie and same domain. After loading is done the alert with cookie will be shown
3. Press on "Reload the page" button to reload the page with webviews.

### Expected Behavior
Loading the webviews should always result in cookie name/value being displayed in alert dialog.

### Actual Behavior
Sometimes no cookie is shown in the alert dialog after the webview was loaded -> so webview has no access to the cookie which was set before loading it.

### Basic Information

- Version with issue: 4.6.0.1180 (latest has it as well)
- Last known good version: -
- IDE: Visual Studio for Windows 16.5.4
- Platform Target Frameworks: 
  - iOS:  8.0
  - Android: -
  - UWP:  -
- Android Support Library Version: -
- Nuget Packages: -
- Affected Devices: physical iPhones (iOS 13.*) as well as iPhoneSimulator

### Reproduction Link

[WkWebViewAndCookies.zip](https://github.com/xamarin/Xamarin.Forms/files/5232089/WkWebViewAndCookies.zip)

PureWeen:
### Description of Change ###

- fixed WkWebViewRenderer to check for double dispose
- fixed WkWebViewRenderer to properly remove and dispose of associated renderers
- After I started cleaning up the WkWebViewRenderer the UI tests around cookie management started failing on iOS11 so I had to add less then ideal hacks to make all that work. WkWebView and Cookie management on iOS 11 is really really quirky. I'm guessing that it was working fine before the dispose because the leaked WkWebView probably was keeping around a reference to the process pool which caused it to all just work. 

### Issues Resolved ### 
- fixes #11962 
- fixes #12134

### Platforms Affected ### 
- iOS

### Testing Procedure ###
- ui tests pass

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue11962.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue11962.cs
@@ -1,0 +1,91 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Text;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+using Xamarin.Forms.Core.UITests;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 11962, "[iOS] Cannot access a disposed object. Object name: 'WkWebViewRenderer",
+		PlatformAffected.iOS)]
+#if UITEST
+	[NUnit.Framework.Category(Core.UITests.UITestCategories.Github10000)]
+	[NUnit.Framework.Category(UITestCategories.Shell)]
+#endif
+	public class Issue11962 : TestShell
+	{
+		protected override void Init()
+		{
+			ContentPage contentPage = new ContentPage()
+			{
+				Content = new Button()
+				{
+					Text = "Go to the next page and back twice. If app doesn't crash test has passed.",
+					Command = new Command(async () =>
+					{
+						await GoToAsync("//user");
+					}),
+					AutomationId = "NextButton"
+				}
+			};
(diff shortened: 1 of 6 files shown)


rmarinho (changes requested):
Build is failing, , think is with issue reproduction file.

rmarinho (approved):

PureWeen:
Here are the UI tests for this PR

https://dev.azure.com/xamarin/public/_releaseProgress?_a=release-pipeline-progress&releaseId=2338

samhouts (approved):

Issue #11988, opened 2020-09-01
WebView Navigated Status Bug In Android


john-larson:
On Android, Navigated event agrs contains wrong status of "Failed" when an asset request receives an error.  Instead, OnReceivedError should check if the request url of the OnReceivedError method is equal to the navigated url.

PR #12169, opened 2020-09-18
Fix WebView Navigated Status Bug In Android

john-larson:
On Android, Navigated event args contain wrong status of "Failed" when an asset request receives an error even though the page itself is loaded successfully. Instead, OnReceivedError should check if the request url of the OnReceivedError method is equal to the navigated url. (issue #11988)

This PR fixes the issue by adding the necessary url check.


hartez (approved):

john-larson:
@samhouts @hartez  Is there any action required on my part?

rmarinho (approved):
Failing tests not related

PR #12186, opened 2020-09-21
[Shell] Fix query parsing and allow more types for parameters

### LINKED ISSUE #10899: [Bug] [Shell] Navigation URL parameter not correctly encoded/decoded
### Description

Using Xamarin.Forms Shell navigation a URL is not correctly passed from one viewmodel to another viewmodel. Either the source viewmodel is not correctly encoding the URL, or the destination viewmodel is not correctly decoding it.

The :// part of the URL ends up with :/

For example, when you pass **http://www.google.com** to the destination viewmodel, the resulting URL becomes **http:/www.google.com** (with a single slash)

### Steps to Reproduce

Using the following code in the source viewmodel:

```
string url = "http://www.google.com";
await Shell.Current.GoToAsync($"general/showUrl?myUrl={url}");
```

In the destination viewmodel:

```
[QueryProperty("MyUrl", "myUrl")]
public class ShowImageViewModel : BaseViewModel
{
    private string _myUrl;

    public string MyUrl
    {
        get => _myUrl;
        set
        {
            _myUrl = Uri.UnescapeDataString(value);
            OnPropertyChanged("MyUrl");
        }
    }
}
```

Here, _myUrl is http:/www.google.com.

I also tried to escape the URL before passing it, but that doesn't work:

`await Shell.Current.GoToAsync($"general/showUrl?myUrl={Uri.EscapeUriString(url)}");`

### Expected Behavior

Source URL == Destination URL

### Actual Behavior

2 slashes end up as 1 slash when passed.

### Basic Information

- Version with issue: 4.6.0.800
- IDE: Visual Studio MAC
- Platform Target Frameworks: 
  - Android: 9.0 

### Workaround

In destination ViewModel:

`_myUrl = Uri.UnescapeDataString(value).Replace(":/", "://");`

PureWeen:
### Description of Change ###

- if the destination type isn't a string try to convert it to property type
- if the parameter is a string url decode it before passing it into the destination
- Fix an issue where url's passed as parameters was removing too many "/"

### Issues Resolved ### 
- fixes #10899 

### Platforms Affected ### 
- Core/XAML (all platforms)

- This code will url decode the parameter before passed to the VM so the behavior is slightly different. But if the user was already decoding then their decode should be a noop

### Testing Procedure ###
- unit tests included

### PR Checklist ###


rmarinho:
can we make ApplyQueryAttributes public? :)

PureWeen:
@rmarinho  I think ApplyQueryAttributes is going to go through some changes so probably not now

pictos (approved):
Tested here on my side, and everything works like a charm

rmarinho (approved):

syedhoque89:
How do I pass an object (lets a instance of a class or a custom property "Foo") via query params? Some examples would be nice please :)

PR #12187, opened 2020-09-21
Move the intention of PageLifecycleManager up to the App level opposed to being at the Page level

### LINKED ISSUE #12055: [Bug] [iOS] [Shell] Multiple PageAppearing/OnAppearing Issues
## Description

After the iOS OnAppearing lifecycle [update](https://github.com/xamarin/Xamarin.Forms/pull/11172) in 4.8, the OnAppearing of pages in the Shell are called when they should not be. In the examples below the shell is build the following way:
                                                                                                                                                                                                                                            
Shell 
-> FlyoutItem with FlyoutDisplayOptions="AsMultipleItems" 
-----> Multiple Tabs
---------> Multiple ShellContents with Shell.TabBarIsVisible="False"

Each ShellContent has the same type of page that displays where OnAppearing is called from. Note that each page is loaded via ContentTemplate.

### Steps to Reproduce

1. Open the app.
2. Navigate to another 'Tab 1, Page 2'.
3. Navigate to 'Tab 2'.
4. Put the app into the background.
5. Bring it back to the foreground.

### Expected Behavior

- After Step 1, the app has called OnAppearing for 'Tab 1, Page 1' (and no other pages).
- After Step 2, the app calls OnAppearing for 'Tab 1, Page 2'.
- After Step 5, the app only called OnAppearing for the visible page ('Tab 2, Page 1'). 

### Actual Behavior

- After Step 1, the app has called OnAppearing for all pages under 'Tab 1'.
- After Step 2, the app doesn't call OnAppearing for 'Tab 1, Page 2'.
- After Step 5, the app calls OnAppearing for all pages under all previously loaded tabs (both Tab 1 and Tab 2). 

### Basic Information

- Version with issue: 4.8.0.1364
- Last known good version: 4.7
- IDE: VS for Mac, Visual Studio
- Platform Target Frameworks:
  - iOS:  13.7
- Affected Devices: Tested on iPad (7th Generation)

### Screenshots

**Scenario 1:** The App is Started

![OnAppearingCalledEarlyAndNotOnNavigation](https://user-images.githubusercontent.com/6966412/92500060-01464f80-f1d3-11ea-8c3b-42d76f662fbb.gif)

Note that all OnAppearing are called for the Children of Tab 1 when app starts. Additionally, navigating to one of those tabs loaded on app start doesn't call OnAppearing (until the second time navigating to the page).

**Scenario 2:** App Returned From Background

![OnAppearingCalledForAllPagesAfterBackground](https://user-images.githubusercontent.com/6966412/92500093-0a372100-f1d3-11ea-83cd-7192b24bef35.gif)

Note that all of the pages from all previously loaded tabs have OnAppearing called. 

### Reproduction Link

[ShellBackgroundingBug.zip](https://github.com/xamarin/Xamarin.Forms/files/5189613/ShellBackgroundingBug.zip)

PureWeen:
### Description of Change ###

PageLifecycleManager added in 4.8 needed some extra logic to detect the valid page to call OnAppearing for on iOS

### Issues Resolved ### 
- fixes #12055 

### Platforms Affected ### 
- iOS

### Testing Procedure ###
- UI Test included

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Page.cs
+++ b/Xamarin.Forms.Core/Page.cs
@@ -424,6 +424,9 @@ namespace Xamarin.Forms
 			if (RealParent is ShellContent sc && !sc.IsVisibleContent)
 				return;
 
+			if (RealParent is IPageContainer<Page> multipage && multipage.CurrentPage != this)
+				return;
+
 			_hasAppeared = true;
 
 			if (IsBusy)
@@ -439,6 +442,12 @@ namespace Xamarin.Forms
 
 			var pageContainer = this as IPageContainer<Page>;
 			pageContainer?.CurrentPage?.SendAppearing();
+
+			if(this is MasterDetailPage mdp)
+			{
+				mdp.Master?.SendAppearing();
+				mdp.Detail?.SendAppearing();
+			}
 
 			FindApplication(this)?.OnPageAppearing(this);
 		}
@@ -456,6 +465,12 @@ namespace Xamarin.Forms
 
 			var pageContainer = this as IPageContainer<Page>;
 			pageContainer?.CurrentPage?.SendDisappearing();
+
+			if (this is MasterDetailPage mdp)
+			{
+				mdp.Master?.SendDisappearing();
+				mdp.Detail?.SendDisappearing();
+			}
 
 			OnDisappearing();
 			Disappearing?.Invoke(this, EventArgs.Empty);
--- a/Xamarin.Forms.Core/Page.cs
+++ b/Xamarin.Forms.Core/Page.cs
@@ -285,7 +285,7 @@ namespace Xamarin.Forms
 
 		protected virtual bool OnBackButtonPressed()
 		{
-			if (RealParent is BaseShellItem || RealParent is Shell)
+			if (IsShell)
 				return false;
 
 			var application = RealParent as Application;
@@ -421,6 +421,9 @@ namespace Xamarin.Forms
 			if (_hasAppeared)
 				return;
 
+			if (RealParent is ShellContent sc && !sc.IsVisibleContent)
+				return;
+
 			_hasAppeared = true;
 
 			if (IsBusy)
@@ -542,6 +545,8 @@ namespace Xamarin.Forms
 			}
 			return !any;
 		}
+
+		bool IsShell => RealParent is BaseShellItem || RealParent is Shell;
 
 		public IPlatformElementConfiguration<T, Page> On<T>() where T : IConfigPlatform
 		{
--- a/Xamarin.Forms.Platform.Android/AppCompat/MasterDetailPageRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/AppCompat/MasterDetailPageRenderer.cs
@@ -128,8 +128,6 @@ namespace Xamarin.Forms.Platform.Android.AppCompat
 				((IMasterDetailPageController)oldElement).BackButtonPressed -= OnBackButtonPressed;
 
 				oldElement.PropertyChanged -= HandlePropertyChanged;
-				oldElement.Appearing -= MasterDetailPageAppearing;
-				oldElement.Disappearing -= MasterDetailPageDisappearing;
 
 				RemoveDrawerListener(this);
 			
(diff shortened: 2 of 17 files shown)


Xamarin.Forms.Platform.iOS/PageLifecycleManager.cs
```diff
@@ -69,25 +68,17 @@ public void Dispose()
 
 		public void HandlePageAppearing()
 		{
-			if (_appeared)
```
PureWeen, line 72: Page.cs already has these checks

https://github.com/xamarin/Xamarin.Forms/blob/5.0.0/Xamarin.Forms.Core/Page.cs#L421

Having the logic inside Page determine whether it's in a valid state to appear or not also keeps the logic in the xplat.

For example with Shell the Page needs to figure out if it's the visible ShellContent. Putting that logic inside this manager I think would be excessive.

I also feel like this class should be alerting our App.cs that an appearance has happened and then the xplat figures out where to route that appearing action. The xplat code knows what page is visible. Having one of these per Page doesn't seem necessary because the PageController here isn't playing any roll with the observer, since the observer is a global event.

Xamarin.Forms.Core/Page.cs
```diff
@@ -437,6 +443,12 @@ public void SendAppearing()
 			var pageContainer = this as IPageContainer<Page>;
 			pageContainer?.CurrentPage?.SendAppearing();
 
+			if(this is MasterDetailPage mdp)
+			{
+				mdp.Master?.SendAppearing();
```
PureWeen, line 448: This is the same logic used inside the native android MDP it should just be moved xplat for consistency

Xamarin.Forms.Platform.Android/AppCompat/MasterDetailPageRenderer.cs
```diff
@@ -128,8 +128,6 @@ void IVisualElementRenderer.SetElement(VisualElement element)
 				((IMasterDetailPageController)oldElement).BackButtonPressed -= OnBackButtonPressed;
 
 				oldElement.PropertyChanged -= HandlePropertyChanged;
-				oldElement.Appearing -= MasterDetailPageAppearing;
```
PureWeen, line 131: Moved this logic into page appearing

Xamarin.Forms.Platform.iOS/PageLifecycleManager.cs
```diff
@@ -4,99 +4,31 @@
...
+			if (_pageController == null)
 				return;
 
-			_appeared = false;
 			_pageController.SendDisappearing();
 		}
-
-		public bool Appeared => _appeared;
-
-		bool CheckIfWeAreTheCurrentPage()
```
PureWeen, line 91: The problem with the a page trying to assess whether it should send appearing is that it has to go pretty far up the chain to know if it should send appearing.  Appearing needs to be top down not bottom up. For example if you have a TabbedPage with nested NavigationPages this causes appearing to fire on non visible TabbedPages because the immediate parent is the NavPage which indicates that it is the current page.

Xamarin.Forms.Platform.iOS/Renderers/ShellRenderer.cs
```diff
@@ -195,6 +195,7 @@ protected override void Dispose(bool disposing)
 			{
 				_disposed = true;
 				FlyoutRenderer?.Dispose();
+				SetCurrentShellItemController(null);
```
PureWeen, line 198: When the main page was getting swapped out the previous shell instance is sticking around because it's not getting cleaned up

daltonks:
The [revert PR](https://github.com/xamarin/Xamarin.Forms/pull/12263) is good, but with this new metrhod, I still need a way to opt-out of page appearing/disappearing events when the app is put into the background. Android has a way to do this.

Tommigun1980:
> The [revert PR](https://github.com/xamarin/Xamarin.Forms/pull/12263) is good, but with this new method, I still need a way to opt-out of page appearing/disappearing events when the app is put into the background. Android has a way to do this.
> 
> Please look at the comments on the initial PR: #11172.

This. I didn't look at the PR but I absolutely don't want to end up in a situation where OnAppearing/OnDisappearing are sent when the app goes to background or foreground, so if this would enable that behaviour please reconsider. There isn't anything that can be done on a page level in app lifecycle calls (suspend/resume) and the only thing that can be done is to try filtering out any calls from app lifecycle calls. The **page** itself doesn't appear or disappear, the **app* does.

If I misunderstood this I apologize.

Thank you.

jfversluis:
This is too big of a change to still add to Xamarin.Forms at this point in time. Thanks all for the time and effort here!

PR #12192, opened 2020-09-21
[Core] Slider Handlers

jsuarezruiz:
### Description of Change ###

Slider Handlers.

| API | Android | iOS | 
| ----|:--------:|:---:|
| DragCompleted | ✅  | ✅ | 
| DragCompletedCommand | ✅  | ✅  |
| DragStarted | ✅  | ✅  |
| DragStartedCommand | ✅  | ✅  |
| Maximum | ✅  | ✅  | 
| MaximumTrackColor | ✅  | ✅  | 
| Minimum | ✅  | ✅  | 
| MinimumTrackColor | ✅  | ✅  | 
| ThumbColor | ✅  | ✅  | 
| ThumbImageSource | ⚠️  | ⚠️  | 
| Value | ✅  | ✅  | 
| ValueChanged | ✅  | ✅  | 

### Platforms Affected ### 

- Core
- iOS
- Android

### PR Checklist ###


src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Slider/SliderHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+		public static void MapValue(SliderHandler handler, ISlider slider)
+		{
+			ViewHandler.CheckParameters(handler, slider);
+			handler.TypedNativeView?.UpdateValue(slider);
+		}
+
+		public static void MapMinimumTrackColor(SliderHandler handler, ISlider slider)
+		{
+			ViewHandler.CheckParameters(handler, slider);
+			handler.TypedNativeView?.UpdateMinimumTrackColor(handler, slider);
```
jsuarezruiz, line 36: At this point we also pass the handler as a parameter. But... why?. Well, we need to use default colors in the extension methods. The colors have been defined in the handler and set when initializing the native view. There are more ways to achieve the same result, feedback?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Slider/SliderHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+			[nameof(ISlider.Value)] = MapValue,
+			[nameof(ISlider.MinimumTrackColor)] = MapMinimumTrackColor,
+			[nameof(ISlider.MaximumTrackColor)] = MapMaximumTrackColor,
+			[nameof(ISlider.ThumbColor)] = MapThumbColor
+		};
+
+		public static void MapMinimum(SliderHandler handler, ISlider slider)
+		{
+			ViewHandler.CheckParameters(handler, slider);
+			handler.TypedNativeView?.UpdateMinimum(slider);
```
jsuarezruiz, line 18: The name of the extension methods should be:
- Update {Property}
- Set {Property}
?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Slider/SliderHandler.iOS.cs
```diff
@@ -0,0 +1,55 @@
...
+
+			return slider;
+		}
+
+		void UpdateDefaultColors(UISlider uISlider)
+		{
+			if (uISlider == null)
+				return;
+
+			_defaultMinTrackColor = uISlider.MinimumTrackTintColor;
```
PureWeen, line 31: Let's move these to static properties that just get set once the first time the default property is pulled

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/SliderExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+		public static void UpdateValue(this SeekBar seekBar, ISlider slider)
+		{
+			var min = slider.Minimum;
+			var max = slider.Maximum;
+			var value = slider.Value;
+
+			seekBar.Progress = (int)((value - min) / (max - min) * 1000.0);
+		}
+
+		public static void UpdateMinimumTrackColor(this SeekBar seekBar, SliderHandler sliderHandler, ISlider slider)
```
PureWeen, line 30: This should become

```C#
public static void UpdateMinimumTrackColor(this SeekBar seekBar, ISlider slider, Color DefaultProgressTintList, TintMode DefaultProgressTintMod)
```

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/ISlider.cs
```diff
@@ -0,0 +1,23 @@
...
+		double Value { get; set; }
+
+		Color MinimumTrackColor { get; }
+		Color MaximumTrackColor { get; }
+		Color ThumbColor { get; }
+
+		ICommand DragStartedCommand { get; }
+		ICommand DragCompletedCommand { get; }
+
+		void ValueChanged();
```
PureWeen, line 19: Remove ValueChanged

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/ISlider.cs
```diff
@@ -0,0 +1,23 @@
...
+	{
+		double Minimum { get; }
+		double Maximum { get; }
+		double Value { get; set; }
+
+		Color MinimumTrackColor { get; }
+		Color MaximumTrackColor { get; }
+		Color ThumbColor { get; }
+
+		ICommand DragStartedCommand { get; }
```
PureWeen, line 16: Remove ICommands

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Slider/SliderHandler.cs
PureWeen: Update

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/DrawableExtensions.Android.cs
```diff
@@ -1,23 +1,22 @@
 ﻿using System;
 using Android.Graphics;
+using Xamarin.Forms;
 using AColor = Android.Graphics.Color;
 using AColorFilter = Android.Graphics.ColorFilter;
 using ADrawable = Android.Graphics.Drawables.Drawable;
+
+#if __ANDROID_29__
```
PureWeen, line 8: You can get rid of all these if defs

this is only going to be API 29+ so they aren't needed

Xamarin.Forms.Platform.Android/Renderers/ShellToolbarTracker.cs
```diff
@@ -272,7 +273,7 @@ protected virtual void OnPagePropertyChanged(object sender, PropertyChangedEvent
 
 		void OnBackButtonBehaviorChanged(object sender, PropertyChangedEventArgs e)
 		{
-			if (!e.Is(BackButtonBehavior.CommandParameterProperty))
+			if(!e.Is(BackButtonBehavior.CommandParameterProperty))
```
hartez, line 276: Can we drop all these extra whitespace changes to make this easier to review?

hartez (changes requested):

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/DrawableExtensions.Android.cs
jsuarezruiz: But this DrawableExtensions class is also used by the Xamarin.Forms renderers.

Xamarin.Forms.Platform.Android/Renderers/ShellToolbarTracker.cs
jsuarezruiz: Sure, I have reverted these changes.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/ISlider.cs
jsuarezruiz: Removed.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Slider/SliderHandler.iOS.cs
jsuarezruiz: Should be properties or just static variables?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/DrawableExtensions.Android.cs
PureWeen: That's fine we've removed API 28 from main on Xamarin.Forms 

We should remove all `#if __ANDROID_29__` from the main branch as well

PureWeen (changes requested):
- Converted the slider handlers over to using generic versions so we don't need typing
- Cleaned up a few parts of the code

Can you sweep through this one and remove any changes that aren't related? For example ShellToolbarTracker?

src/Platform.Handlers/samples/Sample/Controls/Slider.cs
```diff
@@ -0,0 +1,84 @@
...
+
+				_value = value;
+			}
+		}
+
+		public Color MinimumTrackColor { get; set; }
+		public Color MaximumTrackColor { get; set; }
+		public Color ThumbColor { get; set; }
+
+		public Action<double> ValueChanged { get; set; }
```
PureWeen, line 59: Is this still needed?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/View/AbstractViewHandler.cs
```diff
@@ -33,7 +33,7 @@ protected AbstractViewHandler(PropertyMapper mapper)
 			_mapper = _defaultMapper;
 		}
 
-		protected abstract TNativeView CreateView();
+		protected abstract TNativeView CreateNativeView();
```
PureWeen, line 36: Changed name to CreateNativeView to distinguish this from VirtualView

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/View/AbstractViewHandler.cs
```diff
@@ -43,16 +43,28 @@ protected AbstractViewHandler(PropertyMapper mapper)
...
 
-		public virtual void SetView(IView view)
+
+		public void SetView(IView view)
 		{
 			_ = view ?? throw new ArgumentNullException(nameof(view));
 
+			if (VirtualView != null)
+			{
+				TearDownOldVirtualView(VirtualView);
```
PureWeen, line 53: Added `TearDownOldVirtualView` for when the view is replaced with a new virtual view so there is a hook point to do what you want to do

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/View/AbstractViewHandler.cs
```diff
@@ -43,16 +43,28 @@ protected AbstractViewHandler(PropertyMapper mapper)
 
 		public object? NativeView => TypedNativeView;
 
-		public virtual void SetView(IView view)
+
+		public void SetView(IView view)
```
PureWeen, line 47: Removed the virtual on `SetView`  as I wasn't sure if we wanted to allow the user to not call this code at all. I added other life cycle points below that users can use to setup/teardown.

I also liked using specific methods like `TearDownOldVirtualView` because we can make it so they are only called after a "not null" check which keeps out a bunch of `if null` checks inside the handlers

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/View/AbstractViewHandler.cs
```diff
@@ -43,16 +43,28 @@ protected AbstractViewHandler(PropertyMapper mapper)
...
 		{
 			_ = view ?? throw new ArgumentNullException(nameof(view));
 
+			if (VirtualView != null)
+			{
+				TearDownOldVirtualView(VirtualView);
+				_mapper.UpdateProperty(this, VirtualView, nameof(TearDownOldVirtualView));
```
PureWeen, line 54: Is this the best way to hook this? I wanted a way for people to be able to wire into these life cycle points without having to override the handler itself

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/View/AbstractViewHandler.cs
```diff
@@ -43,16 +43,28 @@ protected AbstractViewHandler(PropertyMapper mapper)
...
 			VirtualView = view as TVirtualView;
-			TypedNativeView ??= CreateView();
+			TypedNativeView ??= CreateNativeView();
 
 			if (!HasSetDefaults)
 			{
-				SetupDefaults();
+				if (TypedNativeView != null)
+				{
+					SetupDefaults(TypedNativeView);
```
PureWeen, line 64: Changed `SetupDefaults` to take a TypedNativeView so the override doesn't have to null check and it can just operate on the passed in view

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/View/AbstractViewHandler.cs
```diff
@@ -72,18 +84,43 @@ public virtual void SetView(IView view)
 				}
 			}
 
+			if (VirtualView != null)
+			{
+				SetUpNewVirtualView(VirtualView);
```
PureWeen, line 89: Does it make sense that this is called before `UpdateProperties` ?  I think so

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Slider/SliderHandler.Android.cs
```diff
@@ -0,0 +1,124 @@
...
+		public static void MapThumbColor(SliderHandler handler, ISlider slider)
+		{
+			ViewHandler.CheckParameters(handler, slider);
+
+			handler.TypedNativeView?.UpdateThumbColor(slider, DefaultThumbColorFilter);
+		}
+
+		internal class SeekBarChangeListener : Java.Lang.Object, SeekBar.IOnSeekBarChangeListener
+		{
+			public ISlider? VirtualView { get; set; }
```
PureWeen, line 99: I changed this to be settable because the handler might get its virtual view swapped out. We can't just instantiate this with a VirtualView. The other option here would be to instantiate this with a Func<ISlider> but I think this is fine. This way we don't have to resolve a "Func" each time we want the view

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/View/AbstractViewHandler.cs
jsuarezruiz: I hadn't thought about possible confusion in this case, but this way it's even more explicit. Looks good to me.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/View/AbstractViewHandler.cs
jsuarezruiz: :+1:

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/View/AbstractViewHandler.cs
jsuarezruiz: I think so.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/View/AbstractViewHandler.cs
jsuarezruiz: We have a `TeardownNativeView`, should be just `TearDownVirtualView`?. Would it make sense to only have a single Teardown method with VirtualView and NativeView as parameters? I am concerned that it could cause some confusion  and not knowing the difference and when to use.

PureWeen: The reason for the separation is that the handler is reusable across different VirtualViews. Think CollectionView reusable cells. So, the handler will get different VirtualViews during its life time but it will only ever have the one NativeView. Once the NativeView is torn down then the handler should be considered done with.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Slider/SliderHandler.Android.cs
Clancey: Comet re-uses Handlers. So that makes sense.

PureWeen (approved):

PR #12196, opened 2020-09-21
[Core] Switch Handlers

jsuarezruiz:
### Description of Change ###

Switch Handlers.

| MAUI API | Android | iOS | 
| ----|:--------:|:---:|
| IsToggled |  ✅  | ✅  | 
| TrackColor (OnColor) |  ✅  | ✅  | 
| ThumbColor| ✅  | ✅  |  

### Platforms Affected ### 

- Core
- iOS
- Android

### PR Checklist ###

- [x] Include Handler tests


src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Switch/SwitchHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+	public partial class SwitchHandler
+	{
+		public static PropertyMapper<ISwitch, SwitchHandler> SwitchMapper = new PropertyMapper<ISwitch, SwitchHandler>(ViewHandler.ViewMapper)
+		{
+			[nameof(ISwitch.IsToggled)] = MapIsToggled,
+			[nameof(ISwitch.OnColor)] = MapOnColor,
+			[nameof(ISwitch.ThumbColor)] = MapThumbColor
+		};
+
+		public static void MapIsToggled(SwitchHandler handler, ISwitch view) => handler.TypedNativeView?.UpdateIsToggled(view);
```
jsuarezruiz, line 12: We can have all the logic of the Handler in a single class. However, this will require quite a few compilation directives:
- Namespaces.
- CreateView.
- Teardown.
- GetDesiredSize.
- Specific native control events, etc.

This makes reading the code more complicated.
For that reason, I think a simple way to maintain everything is:
- Map all the properties in the common handler class.
- Use extension methods to do the necessary logic.
- Have the specific platform code in .Android, .iOS, etc. classes.

Have sense?

jsuarezruiz: Reviewed and approved in a team meeting.

Xamarin.Forms.Platform.Android/AppCompat/SwitchRenderer.cs
```diff
@@ -0,0 +1,172 @@
+using System;
```
rmarinho, line 1: why are you adding this file?

src/Platform.Renderers/src/Xamarin.Forms.Platform.Android/Extensions/DrawableExtensions.cs
```diff
@@ -72,36 +75,52 @@ public static void SetColorFilter(this ADrawable drawable, AColorFilter colorFil
...
+			drawable.SetColorFilter(color.ToNative(), mode);
 		}
 
+#pragma warning disable CS0612 // Type or member is obsolete
+#pragma warning disable CS0618 // Type or member is obsolete
 		public static void SetColorFilter(this ADrawable drawable, AColor color, FilterMode mode)
 		{
-			if (Forms.Is29OrNewer)
-				drawable.SetColorFilter(new BlendModeColorFilter(color, GetFilterMode(mode)));
+#if __ANDROID_29__
```
rmarinho, line 101: this is not the same thing as `Forms.Is29OrNewer` is it?

src/Platform.Renderers/src/Xamarin.Forms.Platform.Android/Extensions/DrawableExtensions.cs
```diff
@@ -1,23 +1,22 @@
...
 using ADrawable = Android.Graphics.Drawables.Drawable;
+
+#if __ANDROID_29__
 using ADrawableCompat = AndroidX.Core.Graphics.Drawable.DrawableCompat;
+#else
+using ADrawableCompat = Android.Support.V4.Graphics.Drawable.DrawableCompat;
+#endif
 
-namespace Xamarin.Forms.Platform.Android
+namespace Xamarin.Platform
```
rmarinho, line 14: If this is Xamarin.Platform shouldn't this move to Platform.Handlers folder =?

src/Platform.Renderers/src/Xamarin.Forms.Platform.Android/AppCompat/SwitchRenderer.cs
```diff
@@ -156,8 +157,11 @@ void UpdateThumbColor()
 					_changedThumbColor = false;
 				}
 			}
+
+			Control.ThumbDrawable.SetColorFilter(Element.ThumbColor, FilterMode.SrcAtop);
```
rmarinho, line 161: why we are adding this ?

src/Platform.Renderers/src/Xamarin.Forms.Platform.Android/AppCompat/SwitchRenderer.cs
```diff
@@ -103,49 +104,49 @@ protected override void OnElementChanged(ElementChangedEventArgs<Switch> e)
 
 		protected override void OnElementPropertyChanged(object sender, PropertyChangedEventArgs e)
 		{
-			if (this.IsDisposed())
```
rmarinho, line 106: How are we sure we don't need this?

rmarinho (commented):
Seems we have some extra files, and i have some questions about changes

Xamarin.Forms.Platform.Android/AppCompat/SwitchRenderer.cs
jsuarezruiz: Merge error. Fixed. This file no longer exists.

src/Platform.Renderers/src/Xamarin.Forms.Platform.Android/Extensions/DrawableExtensions.cs
jsuarezruiz: `DrawableExtensions.cs` have been moved to Platform.Handlers.

src/Platform.Renderers/src/Xamarin.Forms.Platform.Android/AppCompat/SwitchRenderer.cs
jsuarezruiz: This line remains.

src/Platform.Renderers/src/Xamarin.Forms.Platform.Android/AppCompat/SwitchRenderer.cs
jsuarezruiz: Removed.

jsuarezruiz:
I have rebased and moved the macOS implementation to: https://github.com/xamarin/Xamarin.Forms/tree/switchhandler-mac in order to facilitate the revision and merge of this PR.

@rmarinho There should be no changes to any not needed files now.

jsuarezruiz:

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/ISwitch.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Xamarin.Forms;
+
+namespace Xamarin.Platform
+{
+	public interface ISwitch : IView
+	{
+		bool IsToggled { get; set; }
+		Color OnColor { get; }
+		Color ThumbColor { get; }
```
mattleibow, line 9: Does there need to be a default/off color or is this just background color?

mattleibow (approved):

PureWeen (changes requested):
I'm fixing the code that tests for colors

mattleibow (changes requested):
@PureWeen just pointed out that this should be `SwitchCompat`

PureWeen:
Should be fixed now @mattleibow

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/SwitchExtensions.cs
```diff
@@ -0,0 +1,64 @@
+﻿using Android.Graphics.Drawables;
+using Xamarin.Forms;
+using ASwitch = AndroidX.AppCompat.Widget.SwitchCompat;
+
+namespace Xamarin.Platform
+{
+	public static class SwitchExtensions
+	{
+		static bool ChangedThumbColor;
```
PureWeen, line 9: I don't think this can be here. On the renderer this was a per instance value so this can't be a static shared across all the handlers

I wonder for the ThumbDrawable if we can use `SetTintList` on the drawable and define the different colors for those states

For API 23+ I think we can use these APIs

https://developer.android.com/reference/android/widget/Switch#setThumbTintList(android.content.res.ColorStateList)

https://developer.android.com/reference/android/widget/Switch#setTrackTintList(android.content.res.ColorStateList)

PureWeen (changes requested):

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/SwitchExtensions.cs
```diff
@@ -0,0 +1,254 @@
...
+				var newList = csl.CreateForState(
+					new int[] { AAttribute.StateChecked },
+					nativeOnColor);
+
+				if (newList != aSwitch.TrackTintList)
+					aSwitch.TrackTintList = newList;
+			}
+		}
+
+		public static void UpdateOnColorSimple(this ASwitch aSwitch, ISwitch view)
```
PureWeen, line 114: @mattleibow @jsuarezruiz so this is the really simple one that works and just toggles based on state. So this is called every single time the checked gets changeed

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/SwitchExtensions.cs
```diff
@@ -0,0 +1,254 @@
...
+					};
+
+
+		public static void UpdateIsToggled(this ASwitch aSwitch, ISwitch view)
+		{
+			aSwitch.Checked = view.IsToggled;
+		}
+
+		//static Drawable? _defaultTrackDrawable;
+		public static void UpdateOnColor(this ASwitch aSwitch, ISwitch view)
```
PureWeen, line 83: @mattleibow @jsuarezruiz With this route we can just set the ColorStateList and then let it just do its thing.

We no longer have to call into UpdateOnColor every time the check changes

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/ISwitch.cs
PureWeen: Yea there's a conversation here about the OffColor

https://github.com/xamarin/Xamarin.Forms/pull/10355

I'm still a fan of just having a property called "TrackColor" and then we just get rid of OnColor

There's some conversation in the issue above around that.

We can still leave oncolor and offColor inside Forms but on the Maui side use TrackColor

mattleibow: I see. Because of the wait the native things work - having colors fade and transition under the hood - it might make sense to keep those in the handler. There is talk of switching to brushes, but is it supported on any of the platforms? And also, is the VSM at the handlers level or just Controls? Not sure if Comet has a concept of VSM?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/SwitchExtensions.cs
mattleibow: How does this work when the switch is halfway? Is it different to using the state lists?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/ISwitch.cs
PureWeen: As far as I remember Comet has something similar to VSM where you can define the state/color relationship

>  is the VSM at the handlers level or just Controls

Currently just at the Controls level. We had talked about this a little bit a couple months ago and the approach I've taken here is kind of the approach we were curious about. Could we just set the colors based on state and the handlers would then just "learn". So, the way I built this if the user is toggling on VSM then the ColorStateList will eventually match.  This does take one iteration to work but maybe we could eventually project the VSM over the handler on init? If we have a Controls.SwitchHandler that knows about VSM then that's a possible scenario.

I realize the comments on the Switch were saying to keep both because of native transitions but this same idea kind of extrapolates out across every single control right? Pretty much all the coloring APIs on iOS/UWP/Android do it via some sort of state based coloring. This isn't unique to switch

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/SwitchExtensions.cs
PureWeen: Not sure, I haven't stared deeply into the colors yet :-)  In a perfect world we could project the VSM over the native platform. 

Though, I think with either implementation we'll need some aspect of what I've done in this PR with extracting the colors out via the attributes and copying the android source.  I'm not really sure how else we can rewind the drawable tints to default. The only other approach I found to work was to make a deep copy of the drawable but that seems like a bad idea from a performance stand point

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/SwitchExtensions.cs
```diff
@@ -0,0 +1,254 @@
...
+using Android.Content;
+using Android.Util;
+
+namespace Xamarin.Platform
+{
+	public static class SwitchExtensions
+	{
+		static ColorTrackingColorStateList? mSwitchTrackStateList;
+		// taken from android sourcec ccode
+		static ColorTrackingColorStateList getSwitchTrackColorStateList(this Context context)
```
PureWeen, line 20: @mattleibow @jsuarezruiz I think we'll need some version of this code regardless if we use ColorStateLists or not

Unless you all can find a better approach for setting the SwitchTrack Drawable back to `Color.Default`

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/ISwitch.cs
PureWeen: @mattleibow even if we don't want to be smart about matching ColorStateLists to current states it's just nicer in general to use ColorStateLists for setting the color for all states. If I want red I just set a CSL once for all the state attributes and then it just works.

When using the color filter you have to call "SetColorFilter" every time the color changes
https://github.com/xamarin/Xamarin.Forms/pull/12196/files#diff-5f11a6818b5ece8c3e67df6b8e71a734267ac65f49c428525efb61ec8122a08bR74

To "re-apply" the color the user want and override the underlying CSL

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/iOS/SwitchExtensions.cs
```diff
@@ -0,0 +1,66 @@
...
+			if (view.TrackColor == Forms.Color.Default)
+				uiSwitch.OnTintColor = defaultOnTrackColor;
+			else
+				uiSwitch.OnTintColor = view.TrackColor.ToNative();
+
+			UIView uIView;
+			if (NativeVersion.IsAtLeast(13))
+				uIView = uiSwitch.Subviews[0].Subviews[0];
+			else
+				uIView = uiSwitch.Subviews[0].Subviews[0].Subviews[0];
```
jsuarezruiz, line 30: Could we create a` UIView FindChildren <T>` or similar extension method? It would be useful in other cases, and does this type of case where we need to access a specific child easier.

src/Platform.Handlers/tests/Xamarin.Platform.Handlers.DeviceTests/AssertionExtensions.Android.cs
```diff
@@ -45,12 +47,20 @@ public static AColor ColorAtPoint(this Bitmap bitmap, int x, int y, bool include
 
 		public static Bitmap ToBitmap(this AView view)
 		{
+			var layout = new FrameLayout(view.Context);
+			layout.LayoutParameters = new FrameLayout.LayoutParams(500, 500);
```
jsuarezruiz, line 51: Why 500 as height and width?

jsuarezruiz (commented):
Looks very nice, the handler test passes.

src/Platform.Handlers/tests/Xamarin.Platform.Handlers.DeviceTests/AssertionExtensions.Android.cs
PureWeen: just randomly picked something 
No science to this really. 

It's just the container frame and then the added control will just calculates its own height/width

If we get to a point where we're screen shotting largers views we need to take pictures of we can feed these properties in

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/iOS/SwitchExtensions.cs
PureWeen: This feels fairly edge case since I'm basically just accessing a specific child and I'm not "Searching" for a specific type

PureWeen (approved):

PR #12197, opened 2020-09-21
[Core] Stepper Handlers

jsuarezruiz:
### Description of Change ###

Stepper Handlers.

| API | Android | iOS | 
| ----|:--------:|:---:|
| Increment | ✅  | ✅  |  
| Maximum  | ✅  | ✅  |  
| Minimum  | ✅  | ✅  |  
| Value  | ✅  | ✅  |  
| ValueChanged  | ✅  | ✅  |  

### Platforms Affected ### 

- Core
- iOS
- Android

### PR Checklist ###

- [x]  Include Handler tests

### DIFF
--- a/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/IRange.cs
+++ b/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/IRange.cs
@@ -1,0 +1,9 @@
+﻿namespace Xamarin.Platform
+{
+	public interface IRange : IView
+	{
+		double Minimum { get; }
+		double Maximum { get; }
+		double Value { get; set; }
+	}
+}
--- a/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/IStepper.cs
+++ b/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/IStepper.cs
@@ -1,0 +1,7 @@
+﻿namespace Xamarin.Platform
+{
+	public interface IStepper : IRange
+	{
+		double Increment { get; }
+	}
+}
--- a/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Stepper/StepperHandler.Standard.cs
+++ b/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Stepper/StepperHandler.Standard.cs
@@ -1,0 +1,9 @@
+﻿using System;
+
+namespace Xamarin.Platform.Handlers
+{
+	public partial class StepperHandler : AbstractViewHandler<IStepper, object>
+	{
+		protected override object CreateNativeView() => throw new NotImplementedException();
+	}
+}
--- a/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Stepper/StepperHandler.cs
+++ b/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Stepper/StepperHandler.cs
@@ -1,0 +1,57 @@
+﻿namespace Xamarin.Platform.Handlers
+{
+	public partial class StepperHandler
+	{
+		public static PropertyMapper<IStepper, StepperHandler> StepperMapper = new PropertyMapper<IStepper, StepperHandler>(ViewHandler.ViewMapper)
+		{
+			[nameof(IStepper.Minimum)] = MapMinimum,
+			[nameof(IStepper.Maximum)] = MapMaximum,
+			[nameof(IStepper.Increment)] = MapIncrement,
+			[nameof(IStepper.Value)] = MapValue
+#if __ANDROID__ || NETCOREAPP
+			,[nameof(IStepper.IsEnabled)] = MapIsEnabled
+#endif
+		};
+
+		public static void MapMinimum(StepperHandler handler, IStepper stepper)
+		{
+			ViewHandler.CheckParameters(handler, stepper);
+			handler.TypedNativeView?.UpdateMinimum(stepper);
+		}
+
+		public static void MapMaximum(StepperHandler handler, IStepper stepper)
+		{
+			ViewHandler.CheckParameters(handler, stepper);
+			handler.TypedNativeView?.UpdateMaximum(stepper);
+		}
+
+		public static void MapIncrement(StepperHandler handler, IStepper stepper)
+		{
+			ViewHandler.CheckParameters(handler, stepper);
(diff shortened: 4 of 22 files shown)


src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/IRange.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Xamarin.Platform
+{
+	public interface IRange : IView
+	{
+		double Minimum { get; }
```
StephaneDelcroix, line 5: why no setter ?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/IStepper.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Xamarin.Platform
+{
+	public interface IStepper : IRange
+	{
+		double Increment { get; }
```
StephaneDelcroix, line 5: why no setter ?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Stepper/StepperHandler.Standard.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+
+namespace Xamarin.Platform.Handlers
+{
+	public partial class StepperHandler : AbstractViewHandler<IStepper, object>
+	{
+		protected override object CreateView() => throw new NotImplementedException();
```
StephaneDelcroix, line 7: shouldn't that throw in the default (base) implementation ?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Stepper/StepperHandler.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace Xamarin.Platform.Handlers
+{
+	public partial class StepperHandler
+	{
+		public static PropertyMapper<IStepper> StepperMapper = new PropertyMapper<IStepper>(ViewHandler.ViewMapper)
```
StephaneDelcroix, line 5: wouldn't it make more sense to define the mapper in StepperHandler.*.cs, that is closer to the implementation. This way we could also _not_ map properties that do not make sense in some platform, and thus have a capability (by code) to figure out if a property is mapped on a platform or not ?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Stepper/StepperHandler.iOS.cs
```diff
@@ -0,0 +1,57 @@
...
+			return uIStepper;
+		}
+
+		protected override void DisposeView(UIStepper uIStepper)
+		{
+			uIStepper.ValueChanged -= OnValueChanged;
+			base.DisposeView(uIStepper);
+		}
+
+		public static void MapMinimum(IViewHandler handler, IStepper slider) => (handler as StepperHandler)?.UpdateMinimum();
```
StephaneDelcroix, line 22: we should use direct casting, and no null check. if this happen, we'd better fail (also, @PureWeen has another PR to gentrify this)

StephaneDelcroix: also, you can actually set the value directly, without defining an UpdateIncrement method

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Stepper/StepperHandler.cs
jsuarezruiz: But we repeat the same mapping N times, many times being the same, right?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Stepper/StepperHandler.iOS.cs
jsuarezruiz: This has changed a bit since the comment.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Stepper/StepperHandler.Standard.cs
jsuarezruiz: But it is an abstract method.

jsuarezruiz:
I have rebased and moved the macOS implementation to: https://github.com/xamarin/Xamarin.Forms/tree/stepperhandler-mac in order to facilitate the revision and merge of this PR.

jsuarezruiz:

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/View/AbstractViewHandler.Android.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace Xamarin.Platform.Handlers
 {
-	public partial class AbstractViewHandler<TVirtualView, TNativeView> : IAndroidViewHandler
+	public partial class AbstractViewHandler<TVirtualView, TNativeView> : Java.Lang.Object, IAndroidViewHandler
```
PureWeen, line 7: Why do you need to do this? I don't think the AbstractViewHandler should inherit from Java.Lang.Object

PureWeen (changes requested):

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/View/AbstractViewHandler.Android.cs
jsuarezruiz: To get the handler here (using the View Tag): https://github.com/xamarin/Xamarin.Forms/blob/1008a65edcf3ca8bd4ade01b5c7bf537db3076d1/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/StepperHandlerManager.cs#L41

Probably we can achieve the same in another way.  Will take a look.

jsuarezruiz: Changed to do in another way.

jsuarezruiz:
Moved to https://github.com/dotnet/maui/pull/517

PR #12231, opened 2020-09-24
[Core] Handle Timer with TimeSpan of 0

### LINKED ISSUE #12220: [Bug] Divide by zero exception when calling Device.StartTimer() with zero seconds
### Description
I would like to start off a timer with zero delay between calls.
Thus, I pass FromSeconds(0).

However, this creates a divide by zero exception.

### Steps to Reproduce

 Device.StartTimer(TimeSpan.FromSeconds(0), YourCallBack);

Xamarin.Forms version v5.0.1.1.467-nightly.

rmarinho:
### Description of Change ###

Fix issue when we want a timer with a TimeSpan with 0 seconds, this case we want it to fire immediatly
### Issues Resolved ### 

- fixes #12220

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### Testing Procedure ###

Added unit test

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Device.cs
+++ b/Xamarin.Forms.Core/Device.cs
@@ -232,6 +232,12 @@ namespace Xamarin.Forms
 
 		public static void StartTimer(TimeSpan interval, Func<bool> callback)
 		{
+			if (interval.TotalMilliseconds == 0)
+			{
+				while (callback())
+					;
+				return;
+			}
 			PlatformServices.StartTimer(interval, callback);
 		}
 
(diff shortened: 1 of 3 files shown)


Xamarin.Forms.Core/Device.cs
```diff
@@ -232,6 +232,12 @@ public static void OpenUri(Uri uri)
 
 		public static void StartTimer(TimeSpan interval, Func<bool> callback)
 		{
+			if (interval.TotalMilliseconds == 0)
+			{
+				while (callback())
+					;
+				return;
```
StephaneDelcroix, line 239: nothing else will ever get a chance to get executed. is this what we want ? Shouldn't we have a minimal (internal) value of 15ms ?

rmarinho: what if the user wants 0ms, that's actual the use case

jfversluis:
I think opinions were divided on this one. Let's not merge this anymore as it didn't get a lot of traction anyway. Thanks!

Issue #12136, opened 2020-09-16
[Spec]  APIS That need to be made public for contributors and Xamarin Community Tookit


PureWeen:
# As we move controls over to the XCT we need to expose some properties to the outside world

Here's the list so far. As we create issues to fix these please link them here

Internal
- Binding.SelfPath constant
- GestureRecognizer default ctor
- NullEffect class
Private
- FontElement class
- PaddingElement class

Xamarin.Forms.Core.UnitTests:
Internal:
- RelativeLayoutTests.UnitExpressionSearch class

In addition, the Markup unit tests use reflection to access these private API's in Xamarin.Forms.Core in the BindableObject class: GetContext method, BindableContextAttributes enum , BindablePropertyContext nested class , SetValueArgs nested class

- [Add protected Accessor to VisualElement Constructor](https://github.com/xamarin/Xamarin.Forms/pull/12125)

VincentH-Net:
Thanks for creating this issue @PureWeen; 

Most of the listed API's are used by C# Markup. Once these API's are available in the XTC we can move C# Markup there.

bmacombe:
VisualElement constructor to protected.  There is a PR #12125 for it already, but just to have a complete list.

ahoefling:
As I work through migrating the Popup Control (Xamarin/XamarinCommunityToolkit#292) I will document some of the APIs required that are currently internal. Right, now I have created the referenced PR #12125 to allow protected constructor for `VisualElement`. I know UWP is already giving me some issues and I'll document those here.

pictos:
@PureWeen would be good to have the PlatformSpecific bits public as well. Even if they are decorated with `EditorBrowsable.Never` attribute.
please see:
https://github.com/jfversluis/XamarinCommunityToolkit-1/commit/faf782d8d28193c3f3764505022fda1ce9ffde3c#commitcomment-41827281

ahoefling:
I just submitted 2 PRs that open up several APIs for iOS/UWP that are required to complete the Popup Control port from Xamarin.Forms to the Xamarin Community Toolkit. Once those are merged I will be very close to finishing up my work on the Popup Control

Redth:
@StephaneDelcroix can you weigh in here?

VincentH-Net:
@PureWeen could you add the internal `IPaddingElement` interface to the issue description? 
C# Markup also needs to access that one; I missed it earlier (my bad)
Thx!

PR #9616, opened 2020-02-17
[Core, iOS, UWP, Android] Popup Control Implementation

ahoefling:
### Description of Change ###
This is an incomplete PR as the iOS implementation is not complete. I am submitting it as I have been talking with Xamarin Team members and Community members for almost an entire year about this Popup Control. The goal is to submit it now and start getting feedback and help on the iOS implementation

Implementation of Popup Control (#1778), this change adds native popup control for the major platforms Android, UWP, iOS (incomplete) as defined in the initial work item
- Android - `Dialog`
- UWP - `Flyout`
- iOS - `Popover`

I have been dog fooding the android implementation as the official Popup Control for our internal projects at FileOnQ. The API has worked very well for us over the past 8+ months. 

### Issues Resolved ### 
- fixes #1778

### API Changes (INavigation) ###
Added ability to show Popups from the `INavigation` interface

Added:
 - `Task<T> ShowPopup<T>(Popup<T> popup);`

### API Changes (BasePopup) ###
Added `BasePopup` control designed for rendering a popup that does not care about the return result

Added:
- `View View { get; set; }`
- `Color Color { get; set; }`
- `Color BorderColor { get; set; } // UWP ONLY - wasn't originally in spec`
- `View Anchor { get; set; }`
- `Size Size { get; set; }`
- `bool IsLightDismissEnabled { get; set; }`
- `event EventHandler<PopupDismissedEventArgs> Dismissed;`

 ### API Changes (Popup<T>) ###
Added `Popup<T>` control designed for rendering a popup and returning a value

Added:
- `View View { get; set; }`
- `Color Color { get; set; }`
- `Color BorderColor { get; set; } // UWP ONLY - wasn't originally in spec`
- `View Anchor { get; set; }`
- `Size Size { get; set; }`
- `bool IsLightDismissEnabled { get; set; }`
- `event EventHandler<PopupDismissedEventArgs> Dismissed;`
- `void Reset()`
- `void Dismiss(T result)`
- `Task<T> Result`

### Library Addition ###
In one of the new Control Gallery pages I needed device window information and it was easiest to pull in [Xamarin.Essentials](https://github.com/xamarin/Essentials) for that test project. 

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

### Usage ###
The popup control can be used just like a `ContentPage` or `ContentView`, the power in our implementation is being able to create a simple XAML page that defines how your popup is rendered.

**SimplePopup.xaml**
[...]

PR #12280, opened 2020-09-27
Change Accessors for iOS Specific APIs to public [Xamarin Community Toolkit]

ahoefling:
### Description of Change ###
Changes several internal APIs to public for iOS. This change is required for the Popup Control implementation that is being ported from a Xamarin.Forms PR to the Xamarin Community Toolkit.

related to Xamarin/XamarinCommunityToolkit#292

### Issues Resolved ### 
Related to #12136 

### API Changes ###
**Changed:**
Platform.cs
- internal static readonly BindableProperty RendererProperty => public static readonly BindableProperty RendererProperty

Forms.cs
- internal static void SendViewInitialized => public static void SendViewInitialized

### Platforms Affected ###
- iOS

None

Not applicable

### Testing Procedure ###
Not applicable

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Forms.cs
+++ b/Xamarin.Forms.Platform.iOS/Forms.cs
@@ -209,7 +209,7 @@ namespace Xamarin.Forms
 
 		public static event EventHandler<ViewInitializedEventArgs> ViewInitialized;
 
-		internal static void SendViewInitialized(this VisualElement self, TNativeView nativeView)
+		public static void SendViewInitialized(this VisualElement self, TNativeView nativeView)
 		{
 			ViewInitialized?.Invoke(self, new ViewInitializedEventArgs { View = self, NativeView = nativeView });
 		}
--- a/Xamarin.Forms.Platform.iOS/Platform.cs
+++ b/Xamarin.Forms.Platform.iOS/Platform.cs
@@ -17,7 +17,7 @@ namespace Xamarin.Forms.Platform.iOS
 		, IPlatform
 #pragma warning restore
 	{
-		internal static readonly BindableProperty RendererProperty = BindableProperty.CreateAttached("Renderer", typeof(IVisualElementRenderer), typeof(Platform), default(IVisualElementRenderer),
+		public static readonly BindableProperty RendererProperty = BindableProperty.CreateAttached("Renderer", typeof(IVisualElementRenderer), typeof(Platform), default(IVisualElementRenderer),
 			propertyChanged: (bindable, oldvalue, newvalue) =>
 			{
 #if DEBUG
--- a/Xamarin.Forms.Platform.iOS/Platform.cs
+++ b/Xamarin.Forms.Platform.iOS/Platform.cs
@@ -17,7 +17,7 @@ namespace Xamarin.Forms.Platform.iOS
 		, IPlatform
 #pragma warning restore
 	{
-		public static readonly BindableProperty RendererProperty = BindableProperty.CreateAttached("Renderer", typeof(IVisualElementRenderer), typeof(Platform), default(IVisualElementRenderer),
+		internal static readonly BindableProperty RendererProperty = BindableProperty.CreateAttached("Renderer", typeof(IVisualElementRenderer), typeof(Platform), default(IVisualElementRenderer),
 			propertyChanged: (bindable, oldvalue, newvalue) =>
 			{
 #if DEBUG


Xamarin.Forms.Platform.iOS/Forms.cs
```diff
@@ -209,7 +209,7 @@ public static void Init()
 
 		public static event EventHandler<ViewInitializedEventArgs> ViewInitialized;
 
-		internal static void SendViewInitialized(this VisualElement self, TNativeView nativeView)
+		public static void SendViewInitialized(this VisualElement self, TNativeView nativeView)
```
rmarinho, line 212: Do you need this for something , this should be need for anything right now

Xamarin.Forms.Platform.iOS/Platform.cs
```diff
@@ -17,7 +17,7 @@ public class Platform : BindableObject, INavigation, IDisposable
 		, IPlatform
 #pragma warning restore
 	{
-		internal static readonly BindableProperty RendererProperty = BindableProperty.CreateAttached("Renderer", typeof(IVisualElementRenderer), typeof(Platform), default(IVisualElementRenderer),
+		public static readonly BindableProperty RendererProperty = BindableProperty.CreateAttached("Renderer", typeof(IVisualElementRenderer), typeof(Platform), default(IVisualElementRenderer),
```
PureWeen, line 20: I definitely like the spirit of this change but what's your use case here?

ahoefling: I'm not 100% sure if it is needed. When I did the original PR for the Popup Control I followed some standard rules I found in other built-in renderers. Here is the usage from the original fork 
https://github.com/ahoefling/Xamarin.Forms/blob/489c149260ce116ac8c68705c45d511e29285c97/Xamarin.Forms.Platform.iOS/Renderers/PopupRenderer.cs#L212-L213

```c#
if (iOS.Platform.GetRenderer(Element) == this)
    Element.ClearValue(iOS.Platform.RendererProperty);
```

As far as I understand the code, it is useful during the `Dispose` invocation to clear the renderer from memory. I can see this being valuable for app performance and keeping the memory allocation small.

If you or other reviewers don't think this is necessary I can omit this code in the Popup Control I am working on for the Xamarin Community Toolkit

Xamarin.Forms.Platform.iOS/Forms.cs
ahoefling: I am not sure if this is needed for the Xamarin Community Toolkit. In my original Popup Control that I PRed to Xamarin.Forms I followed similar techniques from other built-in renderers which used this.  Is this only to be used internal? If so I am fine leaving it internal.

Here are the links to the original PR fork that used this API
https://github.com/ahoefling/Xamarin.Forms/blob/489c149260ce116ac8c68705c45d511e29285c97/Xamarin.Forms.Platform.iOS/Renderers/PopupRenderer.cs#L64

```c#
Element?.SendViewInitialized(NativeView);
```

ahoefling:
@PureWeen @rmarinho Thank you for responding to this Pull Request, I have commented to each of your concerns. I am not 100% sure if we need this PR but I wanted to open this to have a discussion on things that were used in the Popup Control that I am trying to port over to the toolkit.

I linked to the code snippets in my fork, but if you need to review the original PR it is #9616

Xamarin.Forms.Platform.iOS/Platform.cs
PureWeen: It should really matter to clear the renderer. That's all connected weakly and it should dispose together.

That being said... I think you could also call this code

Platform.SetRenderer(Element, null);

And it'll have the same effect

ahoefling: I took a look at the inner workings of Platform.SetRenderer and it makes sense to me to use your recommendation
```
Platform.SetRenderer(Element, null);
```

I'll update the PR to roll this change out

Xamarin.Forms.Platform.iOS/Forms.cs
rmarinho: Yo don't need to use this. it's not need anymore on the renderers. This was used to set th AutmoationID for UITests.

rmarinho (changes requested):

rmarinho:
This PR is not needed, you can ignore this on Xamarin Forms Toolkit

Xamarin.Forms.Platform.iOS/Forms.cs
ahoefling: That makes sense, thanks for the clarification

PR #12317, opened 2020-09-30
Leave IMasterDetailPageController intact

### LINKED ISSUE #12224: [Bug] Xamarin.Forms 5 MasterDetailPage renaming broke plugins
Hi.

I just updated to Xamarin.Forms 5 pre-1, and the NuGets in my project that reference MasterDetailPage broke. I'm not even using MDP in my projects, but some plugins refer to it.

I don't know in how many months (if ever) they will be updated, so how can I use Xamarin.Forms 5 in the meantime?

I think this is an irresponsible change as it may kill off projects that are dependent on legacy plugins, unless there is some kind of fallback for this. So what is the fallback, or should I just accept that due to this I may have to remake my entire project, if updates for the plugins don't materialise soon (hint: they never do)?

This change is an order of magnitude larger than just changing the branch name in GitHub. I hope you understand how much trouble this will cause if you decide to go with it, or at least provide some fallback for the plugin issue.

PureWeen:
### Description of Change ###

Separate out IMasterDetailPageController  from IFlyoutPageController so it's not an ABI break

### Issues Resolved ### 
- fixes #12224

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS
- Android
- UWP

### Testing Procedure ###
- I'll reach out to user to verify

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/IFlyoutPageController.cs
+++ b/Xamarin.Forms.Core/IFlyoutPageController.cs
@@ -17,8 +17,18 @@ namespace Xamarin.Forms
 		event EventHandler<BackButtonPressedEventArgs> BackButtonPressed;
 	}
 
-	public interface IMasterDetailPageController : IFlyoutPageController
+	public interface IMasterDetailPageController
 	{
+		bool CanChangeIsPresented { get; set; }
+
+		Rectangle DetailBounds { get; set; }
+
 		Rectangle MasterBounds { get; set; }
+
+		bool ShouldShowSplitMode { get; }
+
+		void UpdateMasterBehavior();
+
+		event EventHandler<BackButtonPressedEventArgs> BackButtonPressed;
 	}
 }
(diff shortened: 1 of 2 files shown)


Xamarin.Forms.Core/IFlyoutPageController.cs
```diff
@@ -17,8 +17,18 @@ public interface IFlyoutPageController
 		event EventHandler<BackButtonPressedEventArgs> BackButtonPressed;
 	}
 
-	public interface IMasterDetailPageController : IFlyoutPageController
+	public interface IMasterDetailPageController
```
samhouts, line 20: Should we put an obsolete on it?

PureWeen: 👍

PureWeen:
UI Tests are here

https://dev.azure.com/xamarin/public/_releaseProgress?_a=release-pipeline-progress&releaseId=2370

samhouts (approved):

PR #12336, opened 2020-10-02
Add ability to set WebView ExecutionMode property (addresses #4720)

### LINKED ISSUE #4720: UWP: Webview: Memory Leak
### Description
Using Xamarin.Forms (3.4.0.1008975), for UWP (windows 10) the webview control will take more and more memory space until it crashes the application
### Steps to Reproduce

1. Create a simple Xamarin.Forms application where page A navigates to page B that encapsulates a webview
2. Run the application and note the memory in Task Manager that the application is consuming
3. Navigate to the page that contains the webview.  Notice that Task Manager is increasing the memory footprint of the application.
4.  Navigate back to the root (page A).  Notice that the memory footprint in task manager is not fully recycled. 
5. Repeat steps 3 & 4 until application crashes

### Expected Behavior
The webview control should closely mimic the resource allocation of Edge browser

### Actual Behavior
After three or four navigation cycles (depends on webpage) the application crashes.

### Basic Information

- Last known good version:
- IDE: Visual Studio 2017
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:  17134
- Nuget Packages:
- Affected Devices:
UWP (Windows 10)

### Screenshots
http://dynamiconnections.com/downloads/screenshot.png

### Reproduction Link

http://dynamiconnections.com/downloads/navigationsample.zip

t-johnson:
### Description of Change ###

Add option for custom renderers to control the execution mode of the WebView control.
This is 'opt-in' as by default this will not change behavior of existing applications.  To opt-in, people would need to set the ExecutionMode property in the constructor of a custom WebViewRenderer, like:

```
  public class MyWebViewRenderer : WebViewRenderer
  {
    public MyWebViewRenderer()
    {
      ExecutionMode = Windows.UI.Xaml.Controls.WebViewExecutionMode.SeparateProcess;
    }
  }
```

When set as 'SeparateProcess', the memory allocated by the WebView itself is all handled in a sub-process, which ensures the main process never crashes from running out of memory here from doing things like opening and closing youtube every 5 seconds.  This behavior will likely crash when the WebView is in-process due to the huge amounts of memory that is required by that website.

This change is related to issue #4720, however I do not believe that issue is actually a bug.  This change does nothing to address a "memory leak" directly, and due to the addition of the new protected member, I wasn't sure if this is classed as an "API change" but thought it would be better to be on the safe side...

### Issues Resolved ### 
- fixes #4720 

### API Changes ###

Added:

Xamarin.Forms.Platform.UWP.WebViewRenderer
 - protected WebViewExecutionMode ExecutionMode
 
Changed:
 None

Removed:
 None 

### Platforms Affected ### 
- UWP

None

Not applicable

### Testing Procedure ###
I've added a new UI test here which reproduces the reported issue if the custom renderer (_4720CustomRenderer) is removed.

Simply opening and closing a view that contains a WebView, pointed at something like youtube.com will cause huge memory usage which is not immediately reclaimed.  When the WebView is running in SeparateProcess mode, there is a child process that is created that handles the memory itself (see [ExecutionMode property](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.webview.executionmode?view=winrt-19041)

### PR Checklist ###


t-johnson:
ugh I'm going to re-do this PR to fix the mess i made by forgetting to create a branch!

PR #12381, opened 2020-10-06
[UWP] Allow to open ListView ContextActions using Ctrl+O 

### LINKED ISSUE #12355: [Bug] UWP ListView ContextActions menu not accessible using keyboard
### Description

It appears that the XF ListView does not support keyboard access to the ContextActions menu on UWP, making it inaccessible to many users who rely on the keyboard or accessibility tools that simulate the keyboard.

### Steps to Reproduce

Run any existing ContextActions sample and try to pop up the context menu for a ListView item using just the keyboard on UWP.

### Expected Behavior

The correct key combination should result in the context menu appearing and the focus being put on the first menu item in that context menu.

### Actual Behavior

No key combination tried results in the context menu appearing.

### Basic Information

- Version with issue: XF 4.8

### Workaround

No workaround found. Looking at the source code (ok, I checked against an early snapshot of the Maui repository, but I doubt things have changed in the corresponding code in XF recently), pretty much everything in CellControl (such as OpenContextMenu) is neither public nor protected, so painful for a custom renderer to get to (it would currently require reflection to get to internal methods etc).

jsuarezruiz:
### Description of Change ###

Allow to open ListView ContextActions using Ctrl+O.

Perhaps working in .NET MAUI in the parity of Desktop APIs, we can cover in a better way these cases also allowing to define the Key and KeyModifiers. If we consider that this PR does not make sense without allowing to set a custom Key + KeyModifier, it can be closed.

### Issues Resolved ### 

- fixes #12355 

### API Changes ###
 
 None

### Platforms Affected ### 

- UWP

None

Not applicable

### Testing Procedure ###
Open Core Gallery and navigate to the ContextActions sample. Use the Tab to se the focus on the ListView. Then can use the arrow keys to move to a specific cell. Press Tab and then, Ctrl+O to open the contextual menu. If the Flyout menu is opened, the test has passed.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/CellControl.cs
+++ b/Xamarin.Forms.Platform.UAP/CellControl.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.ComponentModel;
+using Windows.System;
 using Windows.UI.Input;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Automation.Peers;
@@ -110,6 +111,14 @@ namespace Xamarin.Forms.Platform.UWP
 			SetDefaultSwitchColor();
 
 			return result;
+		}
+
+		protected override void OnProcessKeyboardAccelerators(ProcessKeyboardAcceleratorEventArgs args)
+		{
+			base.OnProcessKeyboardAccelerators(args);
+
+			if (Cell.HasContextActions && args.Modifiers == VirtualKeyModifiers.Control && args.Key == VirtualKey.)
+				OpenContextMenu();
 		}
 
 		ListView GetListView()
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.UAP/CellControl.cs
```diff
@@ -112,6 +113,14 @@ protected override Windows.Foundation.Size MeasureOverride(Windows.Foundation.Si
 			return result;
 		}
 
+		protected override void OnProcessKeyboardAccelerators(ProcessKeyboardAcceleratorEventArgs args)
+		{
+			base.OnProcessKeyboardAccelerators(args);
+
+			if (Cell.HasContextActions && args.Modifiers == VirtualKeyModifiers.Control && args.Key == VirtualKey.)
```
campersau, line 120:
```suggestion
			if (Cell.HasContextActions && args.Modifiers == VirtualKeyModifiers.Control && args.Key == VirtualKey.O)
```

rmarinho:
Let's rebase this on main take a look on it for maui/desktop support

jfversluis:
Tried this to maybe still add it, but seems to do nothing when I press CTRL+O (also not from a virtual keyboard). So let's skip this for Forms and revisit for .NET MAUI

PR #12390, opened 2020-10-06
VerticalStackLayout and HorizontalStackLayout

hartez:
### Description of Change ###

Okay, this is some really, really basic layout stuff so that we can start getting feedback/bugs and get a stake in the ground, layout-wise.

This includes VerticalStackLayout and HorizontalStackLayout; they work pretty much like you'd expect. Each one also includes a Spacing property to provide equal inter-item spacing. They are deliberately separated to make the implementations as straightforward as possible. There are probably lots of minor performance tweaks to be had, but we'll worry about those once we've fully established the behavior and properties.  

Eventually the Forms implementation may (probably will) include a StackLayout implementation with a bindable Orientation property which will internally use these classes. 

IFrameworkElement adds Width and Height - these are replacements for WidthRequest and HeightRequest, and function as much stronger suggestions about the eventual size of the control. This does not include Minimum/Maximum Width/Height yet.


src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/LayoutViewGroup.Android.cs
```diff
@@ -0,0 +1,78 @@
...
+			{
+				return;
+			}
+
+			var deviceIndependentLeft = Context.FromPixels(l);
+			var deviceIndependentTop = Context.FromPixels(t);
+			var deviceIndependentRight = Context.FromPixels(r);
+			var deviceIndependentBottom = Context.FromPixels(b);
+
+			var destination = Rectangle.FromLTRB(deviceIndependentLeft, deviceIndependentTop,
```
AmrAlSayed0, line 69: Account for RTL? #3066 🤔

hartez: When we implement margins, we will need to account for it.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/LayoutViewGroup.Android.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Xamarin.Platform.Handlers
+{
+	public class LayoutViewGroup : ViewGroup
+	{
+		public LayoutViewGroup(Context context) : base(context)
+		{
+		}
+
+		public LayoutViewGroup(IntPtr javaReference, JniHandleOwnership transfer) : base(javaReference, transfer)
```
PureWeen, line 17: Are these here intentionally for cases where the layout will be inflated from an XML file?

PureWeen (approved):

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/IStackLayout.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Xamarin.Platform
+{
+	public interface IStackLayout : ILayout
```
rmarinho, line 3: Grid also has spacing should be this ISpacingLayout

KPixel:
What do you think of the idea of splitting the `Layout` class into two parts: `ItemsView` which hosts the children elements, and `Layout` classes which contain the "layout algorithms"?

The immediate benefit is the ability to change layouts without having to swap the control. This can be very powerful when writing responsive apps. For example, an app can use a stack layout when running on a small screen/window, and switch to a wrap layout when the screen/window is larger. It also makes it easier to plug in a different layout by just writing the (cross platform) algorithm for it.

For backward compatibility reasons, the actual implementation would be to keep the current Layout with both parts, and introduce these new concepts behind the scene. Then, old apps will continue to work as expected, and new apps can directly tap into these new concepts.

Clancey (approved):

PR #12395, opened 2020-10-07
[iOS] Fix Background issue in Button CustomRenderer

### LINKED ISSUE #12372: [Bug] XF 4.8 breaks custom renderer (Button) background color on iOS
### Description
After upgrading to XF 4.8 our custom button renderer does not have a background / color until the user presses the button and then the background is shown.  This custom renderer has worked until XF 4.8.  Downgrading to XF 4.7 and the button color works again.

### Steps to Reproduce

1. Download repo.
2. Run on iOS and notice the "Testing" button has no background.
3. Press the "Testing" button and see the blue background appear (and work from there on out).
4. Downgrade to XF 4.7 and notice the "Testing" button has a blue background on startup (no press needed).

### Expected Behavior
Control.BackgroundColor should work after it is set.

### Actual Behavior
Control.BackgroundColor is not set on the UI until the button is interacted with.

### Basic Information

- Version with issue: XF 4.8
- Last known good version: XF 4.7
- IDE: Visual Studio 2019
- Platform Target Frameworks:
  - iOS:  Fails on iOS 12 and iOS 14
  - Android: Works on Android
  - UWP:  N/A

### Screenshots
image

### Reproduction Link
https://github.com/jimiatnymbl/CustomButtonIOS

jsuarezruiz:
### Description of Change ###

Fix Background issue in Button CustomRenderer in iOS.

### Issues Resolved ### 

- fixes #12372 

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS

None

fix12372

### Testing Procedure ###
Launch Core Gallery and navigate to the issue 12372. If you can see the background color of the custom Button below, the test has passed.

### PR Checklist ###


Xamarin.Forms.Platform.iOS/Renderers/ButtonRenderer.cs
```diff
@@ -170,9 +166,12 @@ protected override void SetBackground(Brush brush)
 					var backgroundImage = this.GetBackgroundImage(brush);
 					backgroundColor = backgroundImage != null ? UIColor.FromPatternImage(backgroundImage) : UIColor.Clear;
 				}
+
+				_useBackgroundBrush = true;
 			}
 
-			Control.BackgroundColor = backgroundColor;
+			if (_useBackgroundBrush)
```
PureWeen, line 173: Should this be using a locally stored _defaultBackgroundColor instead of a bool check?

Basically how SetBackgroundColor is setup?

```C#
		protected override void SetBackgroundColor(Color color)
		{
			if (IsElementOrControlEmpty)
				return;
#if __MOBILE__
			if (color == Color.Default)
				Control.BackgroundColor = _defaultColor;
			else
				Control.BackgroundColor = color.ToUIColor();
#else
			Control.Layer.BackgroundColor = color == Color.Default ? _defaultColor : color.ToCGColor();
#endif
		}
```

PureWeen, line 173: Should this be using a locally stored _defaultBackgroundColor instead of a bool check?

Basically how SetBackgroundColor is setup?

```C#
		protected override void SetBackgroundColor(Color color)
		{
			if (IsElementOrControlEmpty)
				return;
#if __MOBILE__
			if (color == Color.Default)
				Control.BackgroundColor = _defaultColor;
			else
				Control.BackgroundColor = color.ToUIColor();
#else
			Control.Layer.BackgroundColor = color == Color.Default ? _defaultColor : color.ToCGColor();
#endif
		}
```

PureWeen (changes requested):

rachelkang (approved):

Xamarin.Forms.Platform.iOS/Renderers/ButtonRenderer.cs
jsuarezruiz: Yes, we already have a variable with the default color, the new bool variable only indicates if a brush is being used or not. We have two properties `Background` and `BackgroundColor` and this variable just indicate us when we can update the background using a color.

PureWeen: This still feels a little bit off to me. `_useBackgroundBrush` can only ever be toggled one direction so once it's true it's always true

So, after you set the Background to something the `BackgroundColor` effectively becomes useless

```C#
        private void Button_Clicked(object sender, EventArgs e)
        {
            btnMe.Background = SolidColorBrush.Default;
            btnMe.BackgroundColor = Color.Green;
        }

        private void Button_Clicked2(object sender, EventArgs e)
        {
           // this code stops working after you've set the Background to a brush. 
            btnMe.Background = SolidColorBrush.Purple;
        }
```

Is this expected behavior?

jsuarezruiz: Shane, I totally see your feedback and I understand it. I have applied changes with a different approach with more sense.
Now, the custom renderer works and also your case works without using any internal variable etc.

PureWeen (approved):

PR #12431, opened 2020-10-09
[X] add ConvertTo to typeconverters

StephaneDelcroix:
to help integration with tooling

Note that converting back to string does not succeed everytime, can result in loosing some precision, might not return to original value, can be more verbose than the original, etc.

e.g. in case of Color we return #aarrggbb, in every case. we could cache the predefined value when it's constructed, or avoid `aa` when it's `FF`, but we do not right now... 

/cc @maxbrister

### PR Checklist ###


maxbrister (approved):

StephaneDelcroix:
@maxbrister and @etvorun, make sure you catch `NotSupportedException()`

gitKRM:

Issue #8662, opened 2019-11-26
[Bug] ListView Group Header Height Doesn't Autosize to Content


foehammer88:
### Description

ListView Group Header height doesn't autosize to content. Unlike ListView cells and Headers, that auto size their heights to content, ListView Group Headers do not auto size, thus causing manual calculations of label height and sizes to set the height, which introduces bugs on various platforms and device sizes.

### Steps to Reproduce

1. Create a Sample project with a ListView and data source
2. Make ListView data source support Item Groups
3. Populate Groups with Items, where at least one Group title is very long
4. Bind the data in ListView XAML

### Expected Behavior

ListView Group Headers should auto size to content and not have any adverse visual artifacts.

### Actual Behavior

ListView Group Header height stays constant and label/content spills out beneath the Group Header

### Basic Information

- Version with issue: 4.3.0.991211
- Last known good version:
- IDE:
- Platform Target Frameworks: 
  - iOS:  iOS 13.2.2
  - Android: 10.0
- Nuget Packages:
  - Xamarin.Forms 4.3.0.991211
  - Xamarin.Essentials 1.3.1
- Affected Devices:
  - iPhone
  - Android

### Screenshots

![image](https://user-images.githubusercontent.com/1407512/69594400-11c56a80-0fb1-11ea-916a-e3e442a09658.png)

### Reproduction Link

[ListViewGroupHeaderHeightIssueProject.zip](https://github.com/xamarin/Xamarin.Forms/files/3889774/ListViewGroupHeaderHeightIssueProject.zip)

nbevans:
Duplicate of https://github.com/xamarin/Xamarin.Forms/issues/3769#issuecomment-533532673

Fix proposed in here but not dealt with my XF team yet.

PureWeen:
@foehammer88 @nbevans have you tried this with CollectionView? 

I tried the attached sample and it looks like the group header sizes correctly

nielscup:
Same issue here, groupHeader does not auto size on iOS, Android works fine. Also when visibility of the view inside the groupHeader is set to false, the groupHeader still takes up the original space on iOS.

nbevans:
> @foehammer88 @nbevans have you tried this with CollectionView?
> 
> I tried the attached sample and it looks like the group header sizes correctly

No I haven't tried CollectionView - but a lot of codebases are still using ListView.

Jeanjean:
The following workaround worked for me:

```
[...]

PR #12447, opened 2020-10-11
Fixes memory leak from grouped ListView with HasUnevenRows set

### LINKED ISSUE #6742: [Bug] Memory leak in grouping enabled ListView
### Description
First ListView(IsGroupingEnabled="TRUE", HasUnevenRows="TRUE") item is never collected by GC.
If HasUnevenRows is set to "FALSE",  first ListView item is collected by GC.

### Steps to Reproduce

1. Create ListView with IsGroupingEnabled="TRUE" and HasUnevenRows="TRUE".
2. Clear ListView group items and their items.
3. First ListView item was not collected.

### Expected Behavior
First ListView item is collected.
### Actual Behavior
First ListView item was not collected.
### Basic Information

- Version with issue: 4.1.0.555618
- Last known good version: I don't know
- IDE: Visual studio 2019 Professional
- Platform Target Frameworks: 
  - iOS: 12.3.1 
- Nuget Packages:
- Affected Devices:
iPod touch 6
### Reproduction Link

[MemoryTest2.zip](https://github.com/xamarin/Xamarin.Forms/files/3348673/MemoryTest2.zip)

### LINKED ISSUE #10820: [Bug]  ListView with HasUnevenRows="True" is leaking memory (items)
### Description

ListView with HasUnevenRows="True" somehow retains items bound to the list view in memory even after full data reload. Basically every reload dumps old entries and creates new items, but old items don’t go away. I was troubleshooting it, trying to find, what exactly retains the entries and found one particular line which caused that: HasUnevenRows="True". Having that information I recreated a sample solution and was able to reproduce the issue by adding that line to my simple ListView with a simple data binding to an ObservableCollection. The number of retained items in memory depends on the Grouping flag, if enabled then all items are retained, if disabled, only 1 per each refresh cycle is retained in memory.

### Additional tests

1. CollectionView in a similar scenario does not leak memory (`ItemSizingStrategy= MeasureAllItems`)
2. If a page which holds the listview is closed, then the list view and all items deallocated as well (even leaked)
3. Grouping is not the primary cause of the leak but with that enabled all the items are leaking
4. This memory leak affects iOS only, and no memory leak is happening on Android.

### Steps to Reproduce

1. Launch the sample app with Xamarin Profiler attached (allocations mode)
2. The page will be automatically dumping all bound data and reloading new (10 cycles)
3. In Xamarin Profiler filter to `MyTestItem` class, mark only live objects and observe that no all of them are garbage collected

### Expected Behavior

Only items currently bound to the listview stay in memory

### Actual Behavior

Some of the items are retained in memory. If you enabled grouping then ALL items will be retained.

### Basic Information

- Version with issue: 4.6.0.800
- IDE: VS4Mac 8.6.4526
- Platform Target Frameworks: iOS (only)
  - iOS:  13.18.1.31
- Nuget Packages: None
- Affected Devices: All

### Screenshots

xf listview memoryleak 1
xf listview memoryleak

### Reproduction Link

The sample project is here (master branch):
https://github.com/alexeystrakh/xamarinforms-listview-memoryleak

There are a few additional tests with grouping enabled in a separate branch:
https://github.com/alexeystrakh/xamarinforms-listview-memoryleak/tree/master-leak-with-grouping

And navigation which frees the ListView:
https://github.com/alexeystrakh/xamarinforms-listview-memoryleak/tree/master-leak-with-navigation

### Workaround

If the page which holds the listview is a root page, there is no workaround, but if the page is a details page, closing that page will let the listview go, which in turn lets all the retained items go.

t-johnson:
### Description of Change ###

The native cell that is created in the GetEstimatedRowHeight was never removed, and was remaining as a child of the native list forever.  This fix ensures that the cell is removed.

With this fix, we do not see the number of items increasing, and GC.GetTotalMemory is stable when resetting the collection the list is bound to.

### Issues Resolved ### 

- fixes #6742
- fixes #10820

### API Changes ###
 None

### Platforms Affected ### 
- iOS

None

Not applicable

### Testing Procedure ###

Use the gallery to access test G6742.  This test binds an ObservableCollection of ObservableCollections to a grouped listview which has HasUnevenRows = true.
Clicking the button in this test clears and rebuilds the collections.  There are WeakReferences to the items in the collections, and we count the number of items kept after garbage collection.  this number should remain stable, and not increase, as it was doing before the fix.

### PR Checklist ###


Codelisk:
Will this make it possible to set the GroupHeader to Height zero in iOS now? That would be amazing

t-johnson:
@Codelisk is there an issue logged about 0 height group headers?  i don't think my change here would have any affect on the current behavior in this area, as I was just addressing memory leaks.

Codelisk:
[https://github.com/xamarin/Xamarin.Forms/issues/8662](url)

@t-johnson this link details the issue. Basically at the moment it is not possible to make the group header invisible in ios it always takes space. 
Yeah probably your pr has nothing to do with it

t-johnson:
@Codelisk  I checked the sample on #8662, and this change has no affect on it. sorry.

Codelisk:
@t-johnson okay, thank you very much!

gitKRM:

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Xamarin.Forms.Controls.Issues.Shared.projitems
```diff
@@ -2491,6 +2494,7 @@
     </EmbeddedResource>
   </ItemGroup>
   <ItemGroup>
+    <EmbeddedResource Include="$(MSBuildThisFileDirectory)Issue6742.xaml">
```
jsuarezruiz, line 2497: There is a small error here:

`Error MSB4024: The 'EmbeddedResource' start tag on line 2497 position 6 does not match the end tag of 'ItemGroup'. Line 2506, position 5. (MSB4024) (Xamarin.Forms.ControlGallery.iOS)`

The build is failing for this reason, could you review it? Let me know if you need help with something.

jsuarezruiz (changes requested):

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Xamarin.Forms.Controls.Issues.Shared.projitems
t-johnson: So sorry - fixed.

nickrandolph:
@samhouts / @jsuarezruiz is there anything blocking this PR being merged? 
Also, is there a reason why we're not seeing the PR based nuget package being built - it looks like the CI process isn't picking up this PR for some reason.

t-johnson:
There is also PR #12535 that is in a similar area, that has been waiting for almost the same amount of time...

jsuarezruiz:
Something weird happened and the Build (also pass unit tests, ui tests etc) for this PR was not launched. I have launched the Build and will review it.

jsuarezruiz (approved):

AlanYost:
Guys I tested this on build 5.0.0.6998 and it DOES NOT fix the memory leak issue we were seeing when using HasUnevenRows="True"

t-johnson:
@AlanYost How are you testing it?  Given it has been several months since i last looked at this fix, i'll have to take some time to get my head back in this space...

AlanYost:
> @AlanYost How are you testing it? Given it has been several months since i last looked at this fix, i'll have to take some time to get my head back in this space...

Hi Tim - I have a test harness that I wrote that tracks the instance counts and where they are no longer considered active by the GC.  I can provide this in a GITHUB repo if you are interested?

t-johnson:
@AlanYost how does your test harness compare to the test case added in this PR?  If you can share it with me I can take a look.

AlanYost:
@t-johnson - I did not run the testcase in the PR - I downloaded the repo from the pipeline and upgraded our test harness.  I'm just having macOS Big Sur upgrade issues at the moment that will require me to reboot etc.  I will create a repo and send the link - hopefully before my machine decides it wants to die (what happened to Apple QA ... :-()

AlanYost:
@t-johnson - Please use the following [repo](https://github.com/AlanYost/xf-mvx-memory-leak-test.git) to test.  You may need to refer to your own XF nuget package as mine is sourced from a local version from the pipeline

t-johnson:
> @t-johnson - I did not run the testcase in the PR - I downloaded the repo from the pipeline and upgraded our test harness. I'm just having macOS Big Sur upgrade issues at the moment that will require me to reboot etc. I will create a repo and send the link - hopefully before my machine decides it wants to die (what happened to Apple QA ... :-()

I've re-tested with the test case that I built here and still definitely still see an improvement.  Its highly likely there are other memory leaks that are not triggered in my test.

AlanYost:
@t-johnson - Thanks for looking into it.  I'm sure there are bugs too - Not sure how memory leaks are tested on XF - but it fails using our approach that we use in many other areas.... Did you try it?

t-johnson:
> @t-johnson - Thanks for looking into it. I'm sure there are bugs too - Not sure how memory leaks are tested on XF - but it fails using our approach that we use in many other areas.... Did you try it?

I've got your app running, using the latest stable release of Xamarin Forms.  With that app, I'm seeing leaks for all the different views (ListView, ListViewGrouping, CollectionView) when the number of items is larger than what can be shown on the screen.  This seems to be a very different case than what I've fixed in this PR.

That said, I modified the test case i built for this PR to have more items than will fit on the screen, and I am still not seeing leaks in my test case...

AlanYost:
@t-johnson - You need to scroll thru the list and we also add/delete in the collection helps.  Are you able to automate this in your tests?

t-johnson:
@AlanYost  in your sample you are copying the items from the view model into an ObservableCollection in the View.  its this collection that is used in the binding the ListView.  If I change your sample so that the ListView binds directly to the ObservableCollection in the view model (along with appropriate changes to re-build that collection when the list is tapped) then the leak problem seems to be much reduced. regardless of number of items or scrolling, the sample only ever shows 1 instance of the View and ViewModel left behind.  This one left over seems to be removed after navigating to a different view and then back, so it seems like the app is keeping the last-used View alive on purpose.

AlanYost:
@t-johnson - We have in the past tried different ways to bind to the collection and had the same results.  Would it be possible for you to check in your code against my repo and I'll have a closer look? I really appreciate you looking into this and would love to know that it can be fixed an alternative way.

rmarinho:
Can you rebase please, to fix the cert failure?

t-johnson:
> Can you rebase please, to fix the cert failure?

is it okay now?

rmarinho:
yes @t-johnson  running iOS tests now.

rmarinho (approved):

PR #12460, opened 2020-10-12
[Handlers] Register handlers and use Microsoft.Extensions for Hosting and Dependency Injection

rmarinho:
### Description of Change ###

Adding the `Host` concept similar to  ASPNET , this will allow to use the containers and providers for our handlers  as well for services during the application, will open possiblity to use as `Configuration` with appsettings file, or other features of `IHost` like a built-in lifecycle and services, logging extensions etc.

Adding the `App` abstract class that allows the basic for working with the AppBuilder. Also we would have `MauiApp`that allows to work with Handlers and Windows. 

IWindow
Windows must be provided by the user by implementing `IWindow GetWindowFor(IActivationState state)` on there own `MauiApp` class. 

IAppHostBuilder
Users can register services and configure the default `AppHostBuilder` with there own services or configurations, this can be done by overiding `virtual IAppHostBuilder CreateBuilder()´

User must register Windows as  `IWindow` , or override `Windows` on App class.

Example:
 ```
public class MyApp : MauiApp
{
	//override the default builder so we can configure and register our own services
	public override IAppHostBuilder CreateBuilder()
	{
		var builder = base.CreateBuilder()
			   .ConfigureLogging(logging =>
			   {
			   	logging.ClearProviders();
			   	logging.AddConsole();
			   })
			   .ConfigureAppConfiguration((hostingContext, config) =>
			   {
				   config.AddInMemoryCollection(new Dictionary<string, string>
									{
									   {"MyKey", "Dictionary MyKey Value"},
									   {":Title", "Dictionary_Title"},
									   {"Position:Name", "Dictionary_Name" },
									   {"Logging:LogLevel:Default", "Warning"}
									});
			   })
			   .ConfigureServices(ConfigureServices);
		return builder;
	}
	
	// we need to implement this so we can return new windows when they are requested by the OS 
	// IActivationState will be related to
	public override IWindow GetWindowFor(IActivationState state)
	{
		return Services.GetService<IWindow>();
	}
        //user can configure extra services on the container
	void ConfigureServices(HostBuilderContext ctx, IServiceCollection services)
	{
		services.AddLogging();
		services.AddSingleton<ITextService, TextService>();
		services.AddTransient<MainPageViewModel>();
		services.AddTransient<MainPage>();
		services.AddTransient<IWindow, MainWindow>();
	}
}
```

The AppBuilder might  be hidden from the user , but we will allow the user to override it, ant the user can extend by registering native services, configure logging , register new handlers.

Example : Register a custom handler to override the default one, configure native services

```
var builder = App.CreateDefaultBuilder()
					.RegisterHandler<IButton, CustomHandlers.CustomPinkTextButtonHandler>()
					.ConfigureServices(ConfigureExtraServices)
					
```

Using the Microsoft Dependency Injection to register handlers in a our own `ServiceCollection`  implementation based on a dictionary. We assume that we only want 1 handler for a specific type then the use of dictionary. We still allow users to override it by just registering there own factory container of  `IServiceProviderFactory<ServiceCollection>` type.

Example to use the default ServiceCollection instead of the Xamarin.Forms one
```
public override IAppHostBuilder CreateBuilder()
{
	var builder = base.CreateBuilder()
					 .UseServiceProviderFactory<ServiceCollection>(new DIExtensionsServiceProviderFactory();
						 
	return builder;
}
   
public class DIExtensionsServiceProviderFactory : IServiceProviderFactory<ServiceCollection>
{
	public ServiceCollection CreateBuilder(IServiceCollection services)
		=> new ServiceCollection { services };

	public IServiceProvider CreateServiceProvider(ServiceCollection containerBuilder)
		=> containerBuilder.BuildServiceProvider();
}
```

This PR also adds a couple of things

- Maui.Controls.Sample that is just the Handlers + Maui.Controls

- Add benchmark project to better get a sense of improvements on MAUI non specific platform code.

- Add `profile-android.ps1` to help measure startup time on the Android application. 
usage: 
```
//old registrar
cd  eng\scripts folder
./profile-android.ps1 -project ..\..\src\Platform.Handlers\samples\Sample.Droid\Sample.Droid.csproj  -configuration Release -package com.microsoft.sample_droid

//new builder
cd  eng\scripts folder
./profile-android.ps1 -project ..\..\src\Forms\samples\Maui.Controls.Sample.Droid\Maui.Controls.Sample.Droid.csproj  -configuration Release -package com.microsoft.maui
```

Current baseline to Register and Get Handlers via Dependency Injection system and a basic Dictionary 

|                        Method |        Job | UnrollFactor |      Mean |     Error |    StdDev |     Gen 0 | Gen 1 | Gen 2 |  Allocated |
|------------------------------ |----------- |------------- |----------:|----------:|----------:|----------:|------:|------:|-----------:|
| RegisterHandlerUsingRegistrar | DefaultJob |           16 |  2.279 ms | 0.0497 ms | 0.1409 ms |         - |     - |     - |          - |
|             GetHandlerUsingDI | DefaultJob |           16 | 36.325 ms | 0.7165 ms | 1.5114 ms | 7071.4286 |     - |     - | 29606000 B |
|      GetHandlerUsingRegistrar | DefaultJob |           16 | 35.231 ms | 0.7023 ms | 1.0725 ms | 7071.4286 |     - |     - | 29606014 B |
|        RegisterHandlerUsingDI | Job-LQJADY |            1 |  2.946 ms | 0.3277 ms | 0.9662 ms |         - |     - |     - |  2097504 B |

Current StartupTime using Release/AOT using Google Pixel 2 

|                        Method |         Mean |     Error |    StdDev |  
|------------------------------ |----------- |------------- |----------:|
| Empty App |  307 ms | 5.22 ms | 16.51 ms |  
| Registrar | 367.9 ms | 6.53 ms | 20.67 ms | 
| AppBuilder+AppHost | 428.231 ms | 5.02 ms | 15.88 ms | 

### Issues Resolved ### 

- None

### API Changes ###

Trying to map IApp/MauiApp to Application on Android and AppDelegate. on iOS while Windows will map to Acitivites and SceneDelegates.

```
public class MauiApplication<TApplication> : global::Android.App.Application where TApplication : MauiApp

public class MauiAppCompatActivity : AppCompatActivity

public class MauiUIApplicationDelegate<TApplication> : UIApplicationDelegate, IUIApplicationDelegate where TApplication : MauiApp
```

The user can register it's own IHandlerServiceProvider and override our handler system. This provider shouldn't validate that implementation implements the required service, since we Register IButton,ButtonHandler, but ButtonHandler doesn't implement IButton, this is just for mapping one type to the other.
```
public interface IMauiServiceProvider : IServiceProvider
```

Special ServiceCollection based on Dictionary 
```
public interface IMauiServiceCollection : IServiceCollection, IDictionary<Type, Func<IServiceProvider, object?>?>
```

This  might exist for all IFrameworkElements and can be platform specific, will consolidate all the info to needed to generate a renderer from a IView, we might also be adding the Dispatcher here too
```
public interface IMauiContext
{
	IServiceProvider Provider { get; }

	IMauiServiceProvider Handlers { get; }
#if __ANDROID__
	global::Android.Content.Context Context { get; }
#elif __IOS__

#endif
}
```

A point where we can get handlers our other services, HotReload service could also be retrived via `App.Services`
```
public interface IApp
{
	IServiceProvider Services { get; }
}
```

```
public abstract class App : IApp
{
	public static App? Current { get; private set; }

	public IServiceProvider? Services ;

	public virtual IAppHostBuilder CreateBuilder() => CreateDefaultBuilder();

	internal void SetServiceProvider(IServiceProvider provider)
	
	public static IAppHostBuilder CreateDefaultBuilder()
	{
		var builder = new AppBuilder();

		builder.UseMauiHandlers();

		return builder;
	}
}
```

The IMauiContext will move to IFrameworkElement
```
public abstract class MauiApp : App
{
	public IMauiContext? Context;
	public abstract IWindow GetWindowFor(IActivationState state);
	internal void SetHandlerContext(IHandlersContext? context)
}
```

```
public static class ServiceProviderExtensions
{
	internal static IServiceProvider BuildServiceProvider(this IMauiServiceCollection serviceCollection)
			=> new MauiServiceProvider(serviceCollection);


	public static IViewHandler? GetHandler(this IServiceProvider services, Type type)
			=> services.GetService(type) as IViewHandler;

	public static IViewHandler? GetHandler<T>(this IServiceProvider services) where T : IView
			=> GetHandler(services, typeof(T));
}
```

```
public interface IAppHostBuilder : IHostBuilder
{
	IHostBuilder ConfigureHandlers(Action<HostBuilderContext, IServiceCollection> configureDelegate);
	new IAppHostBuilder ConfigureAppConfiguration(Action<HostBuilderContext, IConfigurationBuilder> configureDelegate);
	new IAppHostBuilder ConfigureContainer<TContainerBuilder>(Action<HostBuilderContext, TContainerBuilder> configureDelegate);
	new IAppHostBuilder ConfigureHostConfiguration(Action<IConfigurationBuilder> configureDelegate);
	new IAppHostBuilder ConfigureServices(Action<HostBuilderContext, IServiceCollection> configureDelegate);
	new IAppHostBuilder UseServiceProviderFactory<TContainerBuilder>(IServiceProviderFactory<TContainerBuilder> factory);
	new IAppHostBuilder UseServiceProviderFactory<TContainerBuilder>(Func<HostBuilderContext, IServiceProviderFactory<TContainerBuilder>> factory);
	IHost Build(IApp app);
}
```

The IAppHostBuilder implementation, when we call Build(app) we get an IAppHost
```
public class AppHostBuilder : IAppHostBuilder
{	
}
```

```
public static class AppHostBuilderExtensions
{
	public static IAppHostBuilder RegisterHandler<TType, TTypeRender>(this IAppHostBuilder builder)
			where TType : IFrameworkElement
			where TTypeRender : IViewHandler
	public static IAppHostBuilder RegisterHandlers(this IAppHostBuilder builder, Dictionary<Type, Type> handlers)
	public static IAppHostBuilder UseMauiHandlers(this IAppHostBuilder builder)
}
```

```
public interface IWindow
{
		public IPage Page { get; set; }
		public IHandlersContext HandlersContext { get; set; }
```

```
public interface IPage : IView
{
		public IView View { get; set; }
}
```

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android

None

Not applicable

### Testing Procedure ###

Unit tests present

### PR Checklist ###

### CODE REVIEW

File: src/Platform.Handlers/samples/Sample/ViewModel/ViewModelBase.cs [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+
+		#region INotifyPropertyChanged
+		public event PropertyChangedEventHandler PropertyChanged;
+		protected void OnPropertyChanged([CallerMemberName] string propertyName = "")
+		{
+			var changed = PropertyChanged;
+			if (changed == null)
+				return;
+
+			changed.Invoke(this, new PropertyChangedEventArgs(propertyName));
```
StephaneDelcroix, line 32:
```suggestion
			PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
```

src/Platform.Handlers/samples/Sample.iOS/Sample.iOS.csproj
```diff
@@ -29,6 +29,8 @@
     <MtouchArch>x86_64</MtouchArch>
     <MtouchLink>None</MtouchLink>
     <MtouchDebug>true</MtouchDebug>
+    <CodesignKey>Apple Development: Rui Marinho (G6DXESA369)</CodesignKey>
```
rmarinho, line 32: need to remove this

src/Platform.Handlers/tests/Xamarin.Platform.Handlers.UnitTests/TestClasses/MockViewHadler.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace Xamarin.Platform.Handlers.Tests
+{
+	public partial class HostBuilderTests
+	{
+		class MockViewHadler : AbstractViewHandler<IMockView, NativeMockView>
```
StephaneDelcroix, line 5: typo

src/Platform.Handlers/src/Xamarin.Platform.Handlers/App.cs
```diff
@@ -0,0 +1,94 @@
+﻿using System;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Xamarin.Platform.Hosting;
+
+namespace Xamarin.Platform
+{
+	public abstract class App : IApp
```
PureWeen, line 8: Can this class have platform specific versions that all inherit from the application concept of that given platform?

For example 

```C#
	public abstract class App :

#if __ANDROID__
		global::Android.App.Application,
#elif __IOS__
		global::UIKit.UIApplicationDelegate,
#endif
		IApp
```

And then the user will just inherit from this App for MyApp and then that's the app that's also used to load up the application? 

I think in a perfect world there's only once concept of an "App"

src/Platform.Handlers/samples/Sample.Droid/MainActivity.cs
```diff
@@ -27,8 +30,68 @@ protected override void OnCreate(Bundle savedInstanceState)
...
 			_page = FindViewById<ViewGroup>(Resource.Id.pageLayout);
 
-			var app = new MyApp();
-			Add(app.CreateView());
+
+#if __REGISTRAR__
+			UseRegistrar();
+			Add(GetContentView());
+#else
+			var (host,app) = App.CreateDefaultBuilder()
```
PureWeen, line 38: I left a comment above about the App but I don't think we'll want to perform all of this inside the Activity because in theory we'll want multi-window/multi-activity support so the App should exist at a level above the activity. 

But maybe this works for now until we define that?

src/Platform.Handlers/tests/Xamarin.Platform.Handlers.Benchmarks/HandlersBenchmarker.cs
```diff
@@ -0,0 +1,81 @@
...
+		[Benchmark]
+		public void GetHandlerUsingDI()
+		{
+			for (int i = 0; i < _numberOfItems; i++)
+			{
+				var defaultHandler = _app.Handlers.GetHandler<IButton>();
+			}
+		}
+
+		[Benchmark]
```
pictos, line 67: This will add one more column into the benchmark results, telling how much other methods are faster or slower than this one. We can just have one `Baseline = true`, so I'm guessing this is the one.
```suggestion
		[Benchmark(Baseline = true)]
```

src/Platform.Handlers/src/Xamarin.Platform.Handlers/App.cs
rmarinho: Yeah i was thinking something similar next, or this or native extension methods on the AppBuilder .. 

For example:
`.Init<TApplication>(global::Android.App.Application application)`
` .Init<TApplication>(global::UIKit.UIApplicationDelegate)`

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Xamarin.Platform.Handlers.csproj
```diff
@@ -2,12 +2,20 @@
...
+    <AssemblyName>Xamarin.Platform</AssemblyName>
     <Nullable>enable</Nullable>
   </PropertyGroup>
   <PropertyGroup>
     <PackageId>Xamarin.Platform</PackageId>
     <Product>Xamarin</Product>
   </PropertyGroup>
   <Import Project="..\..\..\..\.nuspec\Xamarin.Forms.MultiTargeting.targets" />
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Hosting" Version="5.0.0-rc.1.20451.14" />
```
eerhardt, line 14: The full Hosting package is pretty heavy weight and has a lot of dependencies.

Do you need to reference `Microsoft.Extensions.Hosting` or would it be better if you referenced `Microsoft.Extensions.Hosting.Abstractions` and `Microsoft.Extensions.DependencyInjection.Abstractions`?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Hosting/Internal/AppHostLifetime.cs
```diff
@@ -0,0 +1,99 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Options;
+
+namespace Xamarin.Platform.Hosting.Internal
+{
+	public class AppHostLifetime : IHostLifetime, IDisposable
```
eerhardt, line 11: Does this class need to be `public`? It is in an `Internal` namespace, which seems odd.

I know that's how Microsoft.Extensions.Hosting is, but I'm not sure if that was intentional, or a mistake.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Hosting/AppBuilder.cs
```diff
@@ -0,0 +1,271 @@
...
+		{
+			_configureHandlersActions.Add(configureDelegate ?? throw new ArgumentNullException(nameof(configureDelegate)));
+			return this;
+		}
+
+#pragma warning disable CS8714 // The type cannot be used as type parameter in the generic type or method. Nullability of type argument doesn't match 'notnull' constraint.
+		public IHostBuilder UseServiceProviderFactory<TContainerBuilder>(IServiceProviderFactory<TContainerBuilder> factory)
+		{
+			if (factory != null)
+				_serviceProviderFactory = new ServiceFactoryAdapter<TContainerBuilder>(factory ?? throw new ArgumentNullException(nameof(factory)));
```
eerhardt, line 130: Why the double check for `factory` being null? The line above checks that it is not null, so the `?? throw new ArgumentNullException` is dead code. Same applies in the method below.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Hosting/IStartup.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Xamarin.Platform.Hosting
+{
+	public interface IStartup
```
eerhardt, line 3: I'm confused at the intention of this empty interface. What is this used for?

We usually don't create empty interfaces.

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/interface
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1040

In @PureWeen's prototype, there was a `Page` property:

https://github.com/PureWeen/app-contacts/commit/37b2a28939f7bf6ff9e0540d077e679dffbfb049#diff-a15127609366cfa5a2e48fa5414bfcd78d8fb16e46f53b4ba334eb8d82218594R8

```C#
    public interface IStartupPage
    {
        public Page Page { get; }
    }
```

This may be more preferable over the casting that is currently happening.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Hosting/AppBuilder.cs
```diff
@@ -0,0 +1,271 @@
...
+		public IHostBuilder RegisterHandler<TType, TTypeRender>()
+			where TType : IFrameworkElement
+			where TTypeRender : IViewHandler
+		{
+			ConfigureHandlers((context, handlersCollection) => handlersCollection.AddTransient(typeof(TType), typeof(TTypeRender)));
+		
+			return this;
+		}
+
+		public (IHost Host, TApplication App) Init<TApplication>() where TApplication : class, IApp
```
eerhardt, line 51: I wonder if there needs to be a separation between a `Host` and an `Application`. Why both - what's the difference?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Hosting/AppBuilder.cs
```diff
@@ -0,0 +1,271 @@
...
+		{
+			//User services so its almost empty and fast
+			ServiceCollection servicesCollection = new ServiceCollection();
+
+			var host = Build();
+
+			if (_handlersServiceProvider != null)
+				servicesCollection.AddSingleton<IHandlerServiceProvider>(_handlersServiceProvider);
+
+			var app = (TApplication)Activator.CreateInstance(typeof(TApplication));
```
eerhardt, line 61: Why isn't the Application created through `DI`? That way it can participate in DI and get services itself.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Hosting/AppBuilder.cs
```diff
@@ -0,0 +1,271 @@
...
+
+			if (_hostBuilderContext != null)
+				AppLoader.ConfigureAppServices<TApplication>(_hostBuilderContext, servicesCollection, app);
+
+			var services = servicesCollection.BuildServiceProvider();
```
eerhardt, line 66: Is it intentional to have 2 service containers? This one and the one created in `CreateServiceProvider()`?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Hosting/Internal/AppHost.cs
```diff
@@ -0,0 +1,147 @@
...
+			if (_hostedServices != null)
+			{
+				foreach (IHostedService hostedService in _hostedServices)
+				{
+					// Fire IHostedService.Start
+					await hostedService.StartAsync(combinedCancellationToken).ConfigureAwait(false);
+
+					if (hostedService is BackgroundService backgroundService)
+					{
+						//_ = HandleBackgroundException(backgroundService);
```
eerhardt, line 58: The reason we do this in Hosting is because exceptions that occur after an `await` in a BackgroundService aren't handled at all. So in Hosting, we handle them here and write it to the log.

In a mobile application, maybe it should crash the app?

eerhardt (commented):
I took a casual look through some of the major pieces. Just some initial thoughts.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Hosting/AppBuilder.cs
rmarinho: yes our service container for handlers is registering using a different way, the service doesn't need to implement the interface

src/Platform.Handlers/src/Xamarin.Platform.Handlers/App.cs
rmarinho: Fixed @PureWeen

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Xamarin.Platform.Handlers.csproj
```diff
@@ -2,12 +2,34 @@
...
+    <AssemblyName>Xamarin.Platform</AssemblyName>
     <Nullable>enable</Nullable>
   </PropertyGroup>
   <PropertyGroup>
     <PackageId>Xamarin.Platform</PackageId>
     <Product>Xamarin</Product>
   </PropertyGroup>
   <Import Project="..\..\..\..\.nuspec\Xamarin.Forms.MultiTargeting.targets" />
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Http" Version="5.0.0" />
```
rmarinho, line 14: I need to clean this

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Xamarin.Platform.Handlers.csproj
```diff
@@ -2,12 +2,34 @@
...
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="5.0.0" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Update="Hosting\Internal\AppHostEnvironment.cs">
```
rmarinho, line 18: and this

src/Platform.Handlers/tests/Xamarin.Platform.Handlers.UnitTests/TestClasses/MockButton.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using Xamarin.Forms;
+using Xamarin.Platform.Hosting;
+
+namespace Xamarin.Platform.Handlers.Tests
+{
+	class MockButton : View, IButton, IStartup
```
rmarinho, line 7: IStartup was removed..

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Hosting/HandlerServiceProvider.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Xamarin.Platform.Hosting
+{
+	class HandlerServiceProvider : IHandlerServiceProvider
```
PureWeen, line 6: This code seems to be based on an older version of our Registrar that we had inside the handlers. 

Should we update this code? or just have this code call into the Registrar? 

The factory code added as part of the renderer shim code needs to be accounted for yes?

src/Platform.Handlers/tests/Xamarin.Platform.Handlers.Benchmarks/HandlersBenchmarker.cs
```diff
@@ -0,0 +1,80 @@
...
+		public void GetHandlerUsingRegistrar()
+		{
+			for (int i = 0; i < _numberOfItems; i++)
+			{
+				var defaultHandler = Registrar.Handlers.GetHandler<IButton>();
+			}
+		}
+	}
+
+	class MockApp : App
```
PureWeen, line 76: We should probably keep with calling these stubs yes?

PureWeen (approved):

src/Platform.Handlers/src/Xamarin.Platform.Handlers/MauiApp.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Xamarin.Platform
+{
+	public abstract class MauiApp : App
```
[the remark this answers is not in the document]
rmarinho: Yes, or for example if you want to do a "NativeApp" using the App. for services but don't want the helpers.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/IWindow.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Xamarin.Platform
+{
+	public interface IWindow
+	{
+		public IMauiContext HandlersContext { get; set; }
+		public IPage Page { get; set; }
```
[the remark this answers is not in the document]
rmarinho: That's a good question, for now i think following what we have in forms if there's multiple pages you should have 1 NavigationPage with multiple children .. so there's only 1 root page per window. But we can discuss this further.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Hosting/MauiServiceProvider.cs
```diff
@@ -0,0 +1,49 @@
...
+		IDictionary<Type, Func<IServiceProvider, object?>?> _implementationsFactories;
+
+		public MauiServiceProvider(IMauiServiceCollection collection)
+		{
+			if (collection == null)
+				throw new ArgumentNullException(nameof(collection));
+			_implementationsFactories = collection;
+		}
+
+		public object? GetService(Type serviceType)
```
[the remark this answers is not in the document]
rmarinho: we implement ourselfs because the one on DI package is slower (30 ms)

src/Platform.Handlers/src/Xamarin.Platform.Handlers/MauiApp.cs
```diff
@@ -0,0 +1,15 @@
...
+	{
+		public abstract IWindow GetWindowFor(IActivationState state);
```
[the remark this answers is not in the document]
rmarinho: All Windows, state will be platform dependent and can have things like intent,  url link, state to restore etc.. everything that will allow the user to return the window that is needed, a new one or a existing one after restoring the state.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Hosting/MauiServiceProvider.cs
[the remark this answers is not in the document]
mattleibow: Oooh, very nice!

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Hosting/Internal/AppHost.cs
```diff
@@ -0,0 +1,54 @@
...
+			_logger?.Stopping();
+
+			_ = await Task.FromResult(new object());
+
+			_logger?.Stopped();
+		}
+
+		public void Dispose() => DisposeAsync().AsTask().GetAwaiter().GetResult();
+
+		public async ValueTask DisposeAsync()
```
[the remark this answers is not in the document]
rmarinho: this was copied from aspnet implementation.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Hosting/Internal/AppHost.cs
```diff
@@ -0,0 +1,54 @@
...
+			_logger = logger;
+		}
+
+		public IServiceProvider Services { get; }
+
+		public async Task StartAsync(CancellationToken cancellationToken = default)
+		{
+			_logger?.Starting();
+
+			await Task.Run(() => { });
```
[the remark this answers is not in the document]
rmarinho: we will eventually want to add stuff inside. :)

src/Platform.Handlers/src/Xamarin.Platform.Handlers/App.cs
```diff
@@ -0,0 +1,48 @@
...
+	{
+		IServiceProvider? _serviceProvider;
+		IMauiContext? _context;
+
+		protected App()
+		{
+			Current = this;
+		}
+
+		public static App? Current { get; private set; }
```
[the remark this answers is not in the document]
rmarinho: yes good catch, i will do a check

src/Platform.Handlers/src/Xamarin.Platform.Handlers/MauiApp.cs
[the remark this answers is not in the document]
PureWeen: Maybe the confusion here is that it's called `GetWindowFor` and it should be called `CreateNewWindowFor`

src/Platform.Handlers/src/Xamarin.Platform.Handlers/App.cs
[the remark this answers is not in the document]
PureWeen: Does App need to be nullable? Can't it be readonly and not null since it's initialized in the constructor?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/IWindow.cs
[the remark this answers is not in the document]
PureWeen: Pretty sure we'll just have the one page at the root but that Page could have features that will return back to you all the pages it contains.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Hosting/Internal/ServiceFactoryAdapter.cs
```diff
@@ -0,0 +1,51 @@
...
+		{
+			_contextResolver = contextResolver ?? throw new ArgumentNullException(nameof(contextResolver));
+			_factoryResolver = factoryResolver ?? throw new ArgumentNullException(nameof(factoryResolver));
+		}
+
+		public object CreateBuilder(IServiceCollection services)
+		{
+			if (_serviceProviderFactory == null)
+			{
+				_serviceProviderFactory = _factoryResolver!(_contextResolver!());
```
mattleibow, line 29: Just a note on what I did here to remove the warnings. I added a `!` or null-forgiving operator to pinky swear we know what we are doing.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/MauiApp.cs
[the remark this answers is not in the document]
mattleibow: But if it gets the windows that is already visible...

src/Forms/samples/Maui.Controls.Sample.Droid/MainApplication.cs
```diff
@@ -0,0 +1,15 @@
...
+using Android.App;
+using Android.Runtime;
+using Xamarin.Platform;
+
+namespace Maui.Controls.Sample.Droid
+{
+	[Application]
+	public class MainApplication : MauiApplication<MyApp>
+	{
+		public MainApplication(IntPtr handle, JniHandleOwnership ownerShip) : base(handle, ownerShip)
```
jonathanpeppers, line 11: @dellis1972 it looks like they are planning on using this for default Maui apps

Is there still some issue with Fast Deployment and a custom application class?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/IWindow.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Xamarin.Platform
+{
+	public interface IWindow
+	{
+		public IMauiContext HandlersContext { get; set; }
```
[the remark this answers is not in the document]
rmarinho: Fixed

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/IMauiContext.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+
+namespace Xamarin.Platform
+{
+	public interface IMauiContext
+	{
+		IServiceProvider Provider { get; }
+
+		IMauiServiceProvider Handlers { get; }
```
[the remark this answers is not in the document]
rmarinho: Fixed

src/Platform.Handlers/src/Xamarin.Platform.Handlers/MauiApp.cs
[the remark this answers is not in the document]
rmarinho: Right Create seems it's always new, could be a Window already exists ?!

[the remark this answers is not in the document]
Redth: It doesn't get already visible, it's supposed to be when the platform asks you to create a new window for the given state/args ... this is to deal with the fact that on android and iOS you aren't always in control of when a window instantiation is requested.  On iOS, as a user, you can use the task manager to request that an app creates a new window for you.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/IMauiContext.cs
```diff
@@ -0,0 +1,17 @@
...
+
+#if __ANDROID__
+		global::Android.Content.Context Context { get; }
```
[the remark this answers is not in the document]
rmarinho: will leaving at is

src/Platform.Handlers/src/Xamarin.Platform.Handlers/App.cs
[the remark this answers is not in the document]
rmarinho: add a check and a test

mattleibow (approved):

src/Forms/samples/Maui.Controls.Sample.Droid/MainApplication.cs
dellis1972: There is but only for Enhanced Fast Deployment. That is where we fast deploy the dex files. It just won't work since we need the app class  (java) to be in the classes.dex in the apk. Which it won't be. There is a possible way around it but it would involve allot of work.

PR #12509, opened 2020-10-16
Add ability to set WebView ExecutionMode property (addresses #4720)

### LINKED ISSUE #4720: UWP: Webview: Memory Leak

t-johnson:
### Description of Change ###

Add platform specific to set the ExecutionMode on a WebView on UWP

When set as 'SeparateProcess', the memory allocated by the WebView itself is all handled in a sub-process, which ensures the main process never crashes from running out of memory here from doing things like opening and closing youtube every 5 seconds.  This behavior will likely crash when the WebView is in-process due to the huge amounts of memory that is required by that website.

This change is related to issue #4720, however I do not believe that issue is actually a bug.  This change does nothing to address a "memory leak" directly, and due to the addition of the new protected member, I wasn't sure if this is classed as an "API change" but thought it would be better to be on the safe side...

### Issues Resolved ### 
- partially fixes #4720 

### API Changes ###

Added:
```C#
        public static readonly BindableProperty ExecutionModeProperty = BindableProperty.Create("ExecutionMode", typeof(WebViewExecutionMode), typeof(WebView), WebViewExecutionMode.SameThread);
	public enum WebViewExecutionMode
	{
		SameThread = 0,
		SeparateThread = 1,
		SeparateProcess = 2
	}
```

Xamarin.Forms.Platform.UWP.WebViewRenderer
 - protected WebViewExecutionMode ExecutionMode
 
Changed:
 None

Removed:
 None 

### Platforms Affected ### 
- UWP

None

Not applicable

### Testing Procedure ###
Simply opening and closing a view that contains a WebView, pointed at something like youtube.com will cause huge memory usage which is not immediately reclaimed.  When the WebView is running in SeparateProcess mode, there is a child process that is created that handles the memory itself (see [ExecutionMode property](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.webview.executionmode?view=winrt-19041)

### PR Checklist ###


PureWeen (changes requested):
It looks like this PR is having some issues with `Issue12134.CookiesCorrectlyLoadWithMultipleWebViews()`

```
  at Xamarin.Forms.Controls.Issues.Issue12134.CookiesCorrectlyLoadWithMultipleWebViews() in D:\agent\1\s\Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Issue12134.cs:line 120
```

t-johnson:
@PureWeen  I'll take a look and see if I can figure out the test failure.

t-johnson:
So this may be a bit of a known problem with the SeparateProcess mode and cookies:

https://github.com/microsoft/microsoft-ui-xaml/issues/1297

t-johnson:
@PureWeen Is there a way to see the test failures reported in UWP-ISSUES and xamarin-forms-uitests ?  With the [latest change](https://github.com/xamarin/Xamarin.Forms/pull/12509/commits/a6506e06946df630cb23a72d5fea3d8fa64e62ad) I was expecting that there would not be any unexpected side-effects.  The test for this issue passes for me, so is it possibly the tests behave differently somehow in the pipeline?

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/GitHub4720.cs
```diff
@@ -49,6 +49,11 @@ public async Task GitHub4720Test()
 
 #endif
 	}
+	
+	public class _4720WebView : WebView
```
PureWeen, line 53: Can you move these two classes into `class GitHub4720`? It helps keeps the `Xamarin.Forms.Controls.Issues` namespace a bit cleaner

PureWeen (changes requested):

PureWeen:
@t-johnson 
- I rebased this PR (hopefully that doesn't mess you up) 
- I added some instructions (let me know if those seem correct)

Xamarin.Forms.Platform.UAP/WebViewRenderer.cs
```diff
@@ -401,5 +417,17 @@ void UpdateCanGoBackForward()
 			((IWebViewController)Element).CanGoBack = Control.CanGoBack;
 			((IWebViewController)Element).CanGoForward = Control.CanGoForward;
 		}
+
+		void OnSeparateProcessLost(Windows.UI.Xaml.Controls.WebView sender, WebViewSeparateProcessLostEventArgs e)
```
PureWeen, line 421: What are the conditions that cause this to get called?
I tried a few different things but it never seems to fire. 

Also does this need to unwire the previous webview when this is called? Is there a previous webview when this is called?

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue4720.cs
```diff
@@ -0,0 +1,133 @@
...
+				Debug.WriteLine($"++++++++ Issue4720WebPage : Constructor, count is {s_count}");
+
+				var label = new Label { Text = "Test case for GitHub issue #4720." };
+
+				var button = new Button { Text = "Close Page" };
+				button.Clicked += Button_Clicked;
+
+				var wv = new _4720WebView()
+				{
+					Source = new UrlWebViewSource { Url = "https://www.youtube.com/" },
```
PureWeen, line 72: Can we change this to https://dotnet.microsoft.com/ ?

t-johnson: It can be changed to any webpage, however to actually see the behavior reported by the issue creator, it seems that the webpage has to have a lot of media on it.  i don't think dotnet.microsoft.com would be "busy" enough for the test to fail without this enhancement.

Xamarin.Forms.Platform.UAP/WebViewRenderer.cs
t-johnson: I'm not certain what conditions would cause this to happen.  the [documentation](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.webview?view=winrt-19041#execution-modes) don't elaborate other than suggesting that it is a possibility. I'm _guessing_ if the process is killed or crashed this would happen.
The sample given in the documentation does not unwire the event handlers - i based the implementation off this.

PureWeen (approved):
I couldn't quite get it to crash even when loading YouTube a bunch of times. I did notice that after awhile it would start to lag. For now I changed the UI Test to just use Microsoft and reload the page ever 200 Ms

Waiting 2 seconds and reloading it 10 times just makes this test a bit too slow. I also just converted this property over to a platform specific so users can opt into it if they want to from the xplat code opposed to a renderer

hartez (approved):

t-johnson (commented):
thanks, much cleaner to have the platform specific

PR #12529, opened 2020-10-19
[Core] BoxView Handlers

jsuarezruiz:
### Description of Change ###

BoxView Handlers.

| API | Android | iOS | 
| ----|:-------:|:---:|
| Color  | ✅  | ✅  | 
| CornerRadius | ✅  | ✅  | 

### Platforms Affected ### 

- Core
- iOS
- Android

### PR Checklist ###

- [x]  Include Handler tests


jsuarezruiz:

src/Platform.Handlers/tests/Xamarin.Platform.Handlers.DeviceTests/Handlers/BoxView/BoxViewHandlerTests.cs
```diff
@@ -0,0 +1,32 @@
...
+using Xamarin.Forms;
+using Xamarin.Platform.Handlers.DeviceTests.Stubs;
+using Xunit;
+
+namespace Xamarin.Platform.Handlers.DeviceTests
+{
+	public partial class BoxViewHandlerTests : HandlerTestBase<BoxViewHandler>
+	{
+		[Fact]
+		public async Task CoolorRadiusInitializesCorrectly()
```
rachelkang, line 11: typo: ColorInitializesCorrectly
```suggestion
		public async Task ColorInitializesCorrectly()
```

jsuarezruiz: Oops, updated!.

mattleibow:
Seems CornerRadius was missing?

mattleibow (approved):

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Primitives/CornerRadius.cs
```diff
@@ -3,7 +3,6 @@
 namespace Xamarin.Forms
 {
 	[DebuggerDisplay("TopLeft={TopLeft}, TopRight={TopRight}, BottomLeft={BottomLeft}, BottomRight={BottomRight}")]
-	[TypeConverter(typeof(CornerRadiusTypeConverter))]
 	public struct CornerRadius
```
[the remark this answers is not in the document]
mattleibow: It did break Forms... We probably need to wrap this with `#if` for now?

mattleibow:
Note, that the BoxView is technically not working. See https://github.com/xamarin/Xamarin.Forms/pull/12913

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Primitives/CornerRadius.cs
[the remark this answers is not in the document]
mattleibow: I could attach this attribute to all properties that I find, but it still will break anyone who creates a new control and uses this type. This really needs to be on the struct type so it can be consumed. Or, Forms needs a way to register things like this.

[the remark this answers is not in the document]
mattleibow: I might be able to get it to work by adding this to the `TypeConverterAttribute.KnownConverters` list...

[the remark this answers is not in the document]
mattleibow: This should be fixed in this: https://github.com/xamarin/Xamarin.Forms/pull/12529/commits/39caa92837da445debeb258cf07f8bbff6bbf986

CI will confim.

mattleibow (approved):
Things are looking good so far.

One comment I will make is the Color property... What is the point of that since the color comes from the BackgroundColor property in the base view. Also, what is the point of BackgroundColor when we are going to have Background with all the extras?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/BoxViewExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+				color = boxView.Color;
+			else
+				color = boxView.BackgroundColor;
+
+			nativeView.Color = color;
+		}
+
+		public static void UpdateCornerRadius(this NativeBoxView nativeView, IBox boxView)
+		{
+			nativeView.CornerRadius = boxView.CornerRadius;
```
AmrAlSayed0, line 21: X.F dpi to px conversion needed?

src/Platform.Handlers/samples/Sample/Controls/BoxView.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Xamarin.Forms;
+using Xamarin.Platform;
+
+namespace Sample
+{
+	public class BoxView : Xamarin.Forms.View, IBox
+	{
+		public Color Color { get; set; }
```
PureWeen, line 8: @mattleibow @jsuarezruiz  Do we even need this property?

How does this property differ from setting `BackgroundColor` on the IView?

https://github.com/xamarin/Xamarin.Forms/blob/a01f00590694d9d20cb4ee40d783e81f2bb4c881/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/IFrameworkElement.cs#L9

Like on a current BoxView in Forms what's different when I set Color vs BackgroundColor for a BoxView?

PureWeen (changes requested):

src/Platform.Handlers/samples/Sample/Controls/BoxView.cs
mattleibow: I agree this property is not useful. Might even be detrimental for UX. Which one wins? If we are going to do gradients or images or something, then now we got this property in the way.

I vote to remove it. The forms layer can add it back for compact if need be. But on a base handler level, this is not useful.

PureWeen (changes requested):
Does this handler make sense to have? This is basically just a native view with some coloring properties and a single property for CornerRadius

It feels like this would just be a handler that's specifics to Maui.Controls but in Maui.Core we wouldn't really have a boxview handler?

mattleibow:
@PureWeen I am feeling that too. If the base view has background and corner radius support, then this view is really useless.

How are we going to do round corners in the handlers? Is it some clipping feature, or the corner radius type? Maybe both?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/BoxViewExtensions.cs
jsuarezruiz: Is done in the `NativeBoxView` class.

PureWeen:
Closing PR

BoxViewHandler doesn't really make sense as a specific handler to have in the Maui.Core code

PR #12536, opened 2020-10-20
[Handlers] Fixed Slider inconsistent behavior between Android and iOS

jsuarezruiz:
### Description of Change ###

Fixed Slider inconsistent behavior between Android and iOS

### Issues Resolved ### 

- fixes https://github.com/dotnet/maui/issues/260

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

None

Not applicable

### Testing Procedure ###
Launch Xamarin Platform sample and test the Slider in Android and iOS. If the behavior is the same, the test has passed.

### PR Checklist ###


src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/SliderExtensions.cs
```diff
@@ -7,24 +7,18 @@ namespace Xamarin.Platform
...
 	public static class SliderExtensions
 	{
-		public static void UpdateMinimum(this SeekBar seekBar, ISlider slider)
-		{
-			if (NativeVersion.Supports(NativeApis.SeekBarSetMin))
-			{
-				seekBar.Min = (int)slider.Minimum;
-			}
-		}
+		static double Min;
```
PureWeen, line 10: I don't think these can be in here because if you have different sliders then these will just keep changing. Max and Min are part of ISlider so you don't need to store these here for UpdateValue and/or you can just have a method that takes in a max/min parameter

PureWeen (changes requested):

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/SliderExtensions.cs
jsuarezruiz: Right, updated.

mattleibow (approved):

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Slider/SliderHandler.Android.cs
```diff
@@ -81,10 +88,15 @@ public static void MapThumbColor(SliderHandler handler, ISlider slider)
...
 		{
-			if (VirtualView == null)
+			if (VirtualView == null || !fromUser)
 				return;
 
-			VirtualView.Value = progress;
+			var min = VirtualView.Minimum;
+			var max = VirtualView.Maximum;
+
+			var value = min + (max - min) * (progress / 1000.0);
```
mattleibow, line 97: Can this 1000.0 be a constant? It will get lost and also a nice name will explain.

mattleibow: I see the progress PR uses a 10K number. What is a good number for here? What if this slider needs to be a bigger range. For example, I am making an app that wants to show weight in KG for filling a plane. I can't get any higher precision, even if I set my max in the app to 100,000. Maybe this should be higher? Maybe it should be based on the input range? So if I say 1000-50,000, then this is calculate to mean that the slider needs to support a range from 0-500,000, so I could get better precision.

With the current formula, the range 1K-50K basically means that each step is 49. So 1049, 1098, 1147.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/SliderExtensions.cs
```diff
@@ -7,24 +7,19 @@ namespace Xamarin.Platform
...
 	public static class SliderExtensions
 	{
-		public static void UpdateMinimum(this SeekBar seekBar, ISlider slider)
-		{
-			if (NativeVersion.Supports(NativeApis.SeekBarSetMin))
-			{
-				seekBar.Min = (int)slider.Minimum;
-			}
-		}
+		public const double NativeMaxValue = int.MaxValue;
```
mattleibow, line 10: Use a very large value so we get the smoothest and most precise values.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/SliderExtensions.cs
```diff
@@ -7,24 +7,19 @@ namespace Xamarin.Platform
...
+
+		public static void UpdateMinimum(this SeekBar seekBar, ISlider slider) => UpdateValue(seekBar, slider);
 
-		public static void UpdateMaximum(this SeekBar seekBar, ISlider slider) =>
-			seekBar.Max = (int)slider.Maximum;
+		public static void UpdateMaximum(this SeekBar seekBar, ISlider slider) => UpdateValue(seekBar, slider);
```
mattleibow, line 14: Forward the min/max values to the value as they all work together.

PureWeen (approved):

PR #12546, opened 2020-10-20
Renderer to Handler and Handler to Renderer Shims

PureWeen:
### Description of Change ###

- Wire up our Registrar so that everything registered via assembly scanning can be auto auto shimmed or use a handler
- Create `HandlerToRendererShim` for cases where people are still in the Context of a Forms app and will call Platform.CreateRenderer
- Create `RendererToHandlerShim` for cases where the user has registered a renderer that only implements IVisualElementRenderer

### API Changes ###

```C#
public class RendererToHandlerShim : Xamarin.Platform.Handlers.AbstractViewHandler<Xamarin.Platform.IView, Android.Views.View>
```

```C#
public class HandlerToRendererShim : IVisualElementRenderer
```

```C#
FormsHandler.Init()
```

This is an xplat way to register the handlers as we add handlers you can register the new handler here and it'll be used for CG.

https://github.com/xamarin/Xamarin.Forms/pull/12546/files#diff-5a7d3f99f1c9eb4b20a63099a3a317f3a39ba74e719355aa92a8803836209fe0R7

### Platforms Affected ### 
- iOS
- Android

#### Legacy Apps
Given the following code 
```C#
			SetMainPage(new ContentPage()
			{
				Content = new StackLayout()
				{
					Children = {
						new Button()
					}
				}
			});
```

Button can be registered to either a Handler or a Renderer and the registrar will react accordingly. The ContentPage calls CreateRenderer on its content which will return an IVisualElementRenderer. If the Content is of type IViewHandler then it gets wrapped with a HandlerToRendererShim so that ContentPage can still render the control successfully

#### Future Apps

If you just register a Shim against the handler then the Shim will reach out to the old registrar for the type

```C#
			Xamarin.Forms.Internals.Registrar.Registered.Register(typeof(Xamarin.Forms.Button), typeof(Xamarin.Forms.Platform.Android.FastRenderers.ButtonRenderer));
			RegistrarHandlers.Handlers.Register<Xamarin.Forms.Button>((type) =>
			{
				RendererToHandlerShim rendererToHandlerShim = new RendererToHandlerShim();
				return rendererToHandlerShim;
			});
```

### PR Checklist ###


src/Forms/src/Xamarin.Forms.Core/View.cs
```diff
@@ -173,88 +173,5 @@ void ValidateGesture(IGestureRecognizer gesture)
 			if (gesture is PinchGestureRecognizer && _gestureRecognizers.GetGesturesFor<PinchGestureRecognizer>().Count() > 1)
 				throw new InvalidOperationException($"Only one {nameof(PinchGestureRecognizer)} per view is allowed");
 		}
-
-		#region IView
```
PureWeen, line 177: @hartez is it alright that I moved these to VisualElement?

gitKRM:

jsuarezruiz (changes requested):
@PureWeen Could we add some tests to validate the use of `RendererToHandlerShim` and `RendererToHandlerShim`?

rmarinho:
@PureWeen can e rebase and add some tests to make this works fine?

PureWeen:
@rmarinho @jsuarezruiz 

So in theory our Control Gallery tests are the UI Tests for the shims.
You Shim a Handler into the Control.Gallery and then all the Control Gallery tests pass. 

Also, we can't really do any tests for RendererToHandlerShim until we have the Handler.Testing project setup or the Appium project

At that point we can write some tests to verify that the renderers shim correctly into handlers.

PR #12573, opened 2020-10-22
[docs] Rethink the documentation workflow

jonpryor:
Context: https://github.com/xamarin/Xamarin.Forms-api-docs

Currently, Xamarin.Forms documentation is maintained in a separate git
repo: xamarin/Xamarin.Forms-api-docs.  This requires that whenever
new types or members are added to e.g. `Xamarin.Forms.Core.dll`,
the following steps must be performed:

 1. Make the desired changes within Xamarin.Forms.Core.
 2. Run `make docs` or `make Xamarin.Forms.Core.docs` to generate new
    [**mdoc**(5)][0] documentation stubs for `Xamarin.Forms.Core.dll`.
 3. Write the documentation within the files generated in (2).
 4. Manually merge those changes with the corresponding files from
    [Xamarin.Forms-api-docs/docs][1].

This workflow is complicated, and scary -- particularly step (4).

There is an alternative workflow: Instead of treating
xamarin/Xamarin.Forms-api-docs as the "source of truth" for API
documentation, we can instead use "conventional"
[C# documentation comments][2] to be the single source of truth.
C# documentation comments can be "inline" -- as is already done for
several files within this repo *already* -- or they can be stored in
separate XML files, via the [`<include/>` tag][3].
See [xamarin/Java.Interop@9d6e160c][4] for an example.

When C# documentation comments are used, the C# compiler will emit an
e.g. `Xamarin.Forms.Core.xml` file as part of the build process.

`Xamarin.Forms.Core.xml` can be fed to `mdoc update --import PATH`,
which will cause `mdoc update` to import the documentation present
within `Xamarin.Forms.Core.xml` and merge the contents into the
generated **mdoc**(5) XML files.

Note: `mdoc update` *always* prefers imported documentation over
whatever documentation already exists.

For demonstration purposes, I've:

 1. Updated `Xamarin.Forms.Core.csproj` to generate the
    `Xamarin.Forms.Core.xml` documentation file.

 2. Updated `Xamarin.Forms.Core.csproj` to *ignore* the
    CS1573 and CS1591 warnings, so that it still builds;
    it will take some time to import all existing docs into the
    xamarin/Xamarin.Forms repo, if this is desirable.

 3. Merged the API docs for
    `Xamarin.Forms.Internals.ActionSheetArguments` from
    xamarin/Xamarin.Forms-api-docs into `ActionSheetArguments.cs`

 4. Updated the `make Xamarin.Forms.Core.docs` target to use
    `mdoc update --import Xamarin.Forms.Core.xml`.

This allows the generated `docs/Xamarin.Forms.Core` directory to
contain the original C# documentation comments-based documentation.

[0]: http://docs.go-mono.com/?link=man%3amdoc(5)
[1]: https://github.com/xamarin/Xamarin.Forms-api-docs/tree/c403091dcf4f16e441f49f09b0c73315390148eb/docs
[2]: https://docs.microsoft.com/dotnet/csharp/language-reference/language-specification/documentation-comments
[3]: https://docs.microsoft.com/dotnet/csharp/language-reference/language-specification/documentation-comments#include
[4]: https://github.com/xamarin/Java.Interop/commit/9d6e160cb76ee76902314c3c9f35c140de1bd2d4

### Issues Resolved ### 

- fixes #

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ endif
 DOTNETPCL=$(MONOHOME)/xbuild-frameworks/.NETPortable/v4.5/Profile/Profile259/
 CORE=Xamarin.Forms.Core
 COREASSEMBLY=$(CORE)/bin/Release/netstandard2.0/$(CORE).dll
+COREXML=$(CORE)/bin/Release/netstandard2.0/$(CORE).xml
 XAML=Xamarin.Forms.Xaml
 XAMLASSEMBLY=$(XAML)/bin/Release/netstandard2.0/$(XAML).dll
 MAPS=Xamarin.Forms.Maps
@@ -20,7 +21,7 @@ MDOC=mdoc
 docs: $(CORE).docs $(MAPS).docs $(XAML).docs $(PAGES).docs
 
 $(CORE).docs: $(COREASSEMBLY)
-	$(MDOC) update --delete -o docs/$(CORE) $(COREASSEMBLY) -L $(DOTNETPCL)
+	$(MDOC) update --delete -import "$(COREXML)" -o docs/$(CORE) $(COREASSEMBLY) -L $(DOTNETPCL)
 
 $(XAML).docs: $(XAMLASSEMBLY)
 	$(MDOC) update --delete -o docs/$(XAML) $(XAMLASSEMBLY) -L $(DOTNETPCL)
--- a/Xamarin.Forms.Core/ActionSheetArguments.cs
+++ b/Xamarin.Forms.Core/ActionSheetArguments.cs
@@ -5,9 +5,15 @@ using System.Threading.Tasks;
 
 namespace Xamarin.Forms.Internals
 {
+	///  <summary>For internal use by the Xamarin.Forms platform.</summary>
 	[EditorBrowsable(EditorBrowsableState.Never)]
 	public class ActionSheetArguments
 	{
+		/// <param name="title">For internal use by the Xamarin.Forms platform.</param>
+		/// <param name="cancel">For internal use by the Xamarin.Forms platform.</param>
+		/// <param name="destruction">For internal use by the Xamarin.Forms platform.</param>
+		/// <param name="buttons">For internal use by the Xamarin.Forms platform.</param>
+		/// <summary>For internal use by the Xamarin.Forms platform.</summary>
 		public ActionSheetArguments(string title, string cancel, string destruction, IEnumerable<string> buttons)
 		{
 			Title = title;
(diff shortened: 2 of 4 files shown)


hartez:
We were actually talking about this the other day; the roadblock we ran into was the public docs editing process. 

Right now the source of truth for the online docs is the api-docs repo. Like most MS docs now, anyone can post issues and PR edits directly from the website; the issues and PRs are directed at the api-docs repo.

So, we would either need some sort of synchronization process to pull edits to the api-docs repo back into the /// comments in the source, OR we would need a way for the public edit PRs to be directed appropriately to the Forms repo.

Alternatively, we would need to disable the public editing process.

samhouts:
We just turned off the feedback control on the website to make this a bit simpler. But it might be a good next step to make the repo readonly if we go this route.

hartez:
> We just turned off the feedback control on the website to make this a bit simpler. But it might be a good next step to make the repo readonly if we go this route.

Which feedback control? I can still see a Feedback section on the docs pages.

Redth:
I think that change may not be merged to live yet. Will be looking at this Monday. 

We should get some docs people to weigh in on making the c# source the single source of truth. I do agree that the current workflow is really cumbersome and I actually think we might see more community help with improving docs especially around including them for new feature PR’s if the process was easier. Essentials has this same issue.

jfversluis:
> We should get some docs people to weigh in on making the c# source the single source of truth. 

@davidbritch ? Also because we were talking about this/setting this up for the Toolkit?

mattleibow:
We will need to check with @joelmartinez if this works nicely with the frameworks system.

We will also want to use that so we can track differences in the API on the docs site. I know I was looking at some of the issues where people aren't sure what goes with which versions.

davidbritch:
A couple of things spring to mind from a docs perspective.

**Localization**

There are 14 repos for the Xamarin.Forms API docs - the original en-US source, and then 13 (private) localized repos. This is primarily to support localized intellisense. From the description I've read above, it doesn't sound like this will impact on the localization process (which is fired by check-ins to https://github.com/xamarin/Xamarin.Forms-api-docs). I'm mentioning this just in case anyone *can* think of a reason why it could impact on loc.

**Air gapped content**

There's a requirement for all content on docs.microsoft.com to work in air-gapped environments. The Xamarin conceptual docs have already been made air-gap compliant, and the API docs will be made air gap compliant soon(ish). My point here is that we want to avoid a process where Xamarin.Forms.Core.xml has some non-air gapped content, which keeps over writing an air-gapped version in the Xamarin.Forms-api-docs repo. I can think of a few ways of dealing with this, which shouldn't cause you any work.

**Make the API docs repo read-only**

The API docs repo must accept PRs to ensure the content is air-gapped and remains air-gapped. The solution here could be to make the repo private. I'm not sure if there's any knock-on impact if the bulk of the API docs content is made closed source in the short term (I get that the ideal is that the existing content is converted to /// and added to the source code, but I'm guessing it could be a while before that happens).

hartez:
> I'm mentioning this just in case anyone _can_ think of a reason why it could impact on loc.

It would require a couple of minor changes in the script CI uses to add the localized docs to the NuGet packages (for Intellisense).

joelmartinez:
A few notes:

- Switching to an "import" workflow is definitely feasible, where the source of truth is ///
  - though has @jonpryor mentioned, that means /// will always be the source of truth, and documentation contributions from the community must come through edits to the ///.
- If you make the api docs repo readonly, and switch the workflow so that updates to the docs are simply: 
  - 1. run mdoc to update stubs 2. include roslyn generated xml docs for import 
  - That can still leave the loc workflow untouched. the xamarin.forms-api-docs repo then simply becomes the build artifact from which docs are published and loc changes are sourced.
- To @mattleibow's point, switching to a moniker/framework-based system wouldn't necessarily work with the workflow as-is, but it could certainly work with an import/`///` model
  - to move to that would require that all tracked binaries (ie. previous versions) be maintained _somewhere_. 
  - Traditionally, we've done this by having a "binaries" repository where we publish new versions to ... but that's only an implementation detail as long as previous versions are available.

All in all, there's nothing really stopping you from optimizing your docs process ... you just have to make some strong decisions about "source of truth", and whether you want to start tracking previous versions or not; the rest is just CI updates ;)

davidbritch:
I spoke to the guy who ensures that docs content is air-gap compatible. He said you'll have to ensure that /// comments are air-gap compatible yourselves. It's no biggy though. I can discuss this further in Teams.

davidbritch:
The .NET Core team is doing something similar: https://github.com/dotnet/runtime/issues/44969

In particular, they mention a porting tool they've written for porting API docs back to ///.

gewarren:
@carlossanlop FYI.

PR #12588, opened 2020-10-23
Add margins to layouts

hartez:
### Description of Change ###

Adds Margin to IFrameworkElement and updates layout code to account for margins.

Also adds unit tests for layout code.

### DIFF
--- a/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Layouts/LayoutExtensions.cs
+++ b/src/Platform.Handlers/src/Xamarin.Platform.Handlers/Layouts/LayoutExtensions.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using Xamarin.Forms;
+
+namespace Xamarin.Platform.Layouts
+{
+	public static class LayoutExtensions
+	{
+		public static Size ComputeDesiredSize(this IFrameworkElement frameworkElement, double widthConstraint, double heightConstraint) 
+		{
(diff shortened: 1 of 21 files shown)


src/Platform.Handlers/src/Xamarin.Platform.Handlers/Layouts/LayoutExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+
+		public static Rectangle ComputeFrame(this IFrameworkElement frameworkElement, Rectangle bounds) 
+		{
+			var margin = frameworkElement.Margin;
+
+			// If the margins are too big for the bounds, then simply collapse them to zero
+			var frameWidth = Math.Max(0, bounds.Width - margin.HorizontalThickness);
+			var frameHeight = Math.Max(0, bounds.Height - margin.VerticalThickness);
+
+			return new Rectangle(bounds.X + margin.Left, bounds.Y + margin.Top,
```
AmrAlSayed0, line 43: I'm sorry I'm being picky about this 😅 but accounting for RTL #3066 is really easy, it is a one line fix #11469 . No tests will fail if you do this because all of the current margin tests (even RTL ones) use equal margins on both sides so the issue doesn't appear.

hartez: Yes, this will be in a future PR.

Issue #9905, opened 2020-03-09
[Bug] BindableLayout EmptyView doesn't work with a DataTemplateSelector


davidbritch:
### Description

Attempting to use a `DataTemplateSelector` to return a `DataTemplate` to the `BindableLayout.EmptyViewTemplate` property results in a `TargetInvocationException` on iOS and Android.

It might be an unlikely scenario, but it'd be good if it was at least handled gracefully.

### Steps to Reproduce

1. Run the attached sample.
2. Browse to the "EmptyViewTemplate demo - Data template selector" page.

### Expected Behavior

DataTemplateSelector selects a DataTemplate, which defines the displayed empty view.

### Actual Behavior

`TargetInvocationException`.

### Basic Information

- Version with issue: 4.5-pre1
- IDE: VSMac 8.4.8
- Platform Target Frameworks:
  - iOS:  13.3
  - Android: API 28

### Reproduction Link

[BindableLayouts.zip](https://github.com/xamarin/Xamarin.Forms/files/4307283/BindableLayouts.zip)

PR #11428, opened 2020-07-14
CSharpForMarkup Part 2

VincentH-Net:
>## Note: C# Markup has been moved to the Xamarin Community Toolkit
>The features in this PR will not be included in Xamarin Forms. Instead they have been included in the [Xamarin Community Toolkit C# Markup](https://github.com/MicrosoftDocs/xamarin-communitytoolkit/blob/68fe0de991a99b364ed61e19bc280e3da94ff02e/docs/markup.md) release. Go check it out!

Adds support to C# Markup for:

- `MultiBinding`
- `MultiConverter`
- `BindableLayout`
- `RelativeLayout`
- `DynamicResource`

Improvements:
- `Bind` helpers are now more strongly typed, and `converterParameter` was removed from `Bind` overloads that did not use it
- `Font` helper now has `family` as first parameter (reduces the need to use named parameters across all `FontElement` extensions)
- `Padding` and `Paddings` helpers now work on all `IPaddingElement` view types
- New `Stars(double)` overload for defining rows and columns in Grids

### Issues Resolved ### 
Fix for issue encountered by new C# Markup unit tests:
`BindableLayout.SetEmptyViewTemplate` sets `EmptyView` instead of `EmptyViewTemplate`
Possibly this fixes #9905

### Multi-Binding Helpers
New overloads of the `Bind` helper offer support for  [multi-binding](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/app-fundamentals/data-binding/multibinding).

There are overloads that support 2, 3 or 4 bindings with a type-safe inline convertor. E.g:

```CSharp
new Label { }
    .Bind (Label.TextProperty,
        new Binding (nameof(vm.Name)),
        new Binding (nameof(vm.Score)),
        ((string name, bool score) v) => $"{v.name} Score: { v.score }"
    )
```
The value for all bindings are passed in as a `ValueTuple` with type-safe members.

You can also pass in a type-safe converter parameter:
```CSharp
new Label { }
    .Bind (Label.TextProperty,
        new Binding (nameof(vm.Name)),
        new Binding (nameof(vm.Score)),
        ((string name, int Score) v, bool winner) => $"{v.name} Score: { v.Score } Winner: { winner }",
        converterParameter: true
    )
```
Here `bool winner` gets the value from the `converterParameter`.

You can specify two-way conversion inline:
```CSharp
new Entry { }
    .Bind(Entry.TextProperty,
        new Binding (nameof(vm.Emoticon)),
[...]

PR #12623, opened 2020-10-27
APIS That need to be made public for contributors and Xamarin Community Toolkit

### LINKED ISSUE #12136: [Spec]  APIS That need to be made public for contributors and Xamarin Community Tookit

jsuarezruiz:
### Description of Change ###

Changes in the accessibility level of some APIs that need to be public for contributors and Xamarin Community Tookit.

### Issues Resolved ### 

- fixes #12136 
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/FontElement.cs
+++ b/Xamarin.Forms.Core/FontElement.cs
@@ -1,4 +1,3 @@
-using System.ComponentModel;
 using Xamarin.Forms.Internals;
 
 namespace Xamarin.Forms
--- a/Xamarin.Forms.Core/FontElement.cs
+++ b/Xamarin.Forms.Core/FontElement.cs
@@ -1,8 +1,10 @@
+using System.ComponentModel;
 using Xamarin.Forms.Internals;
 
 namespace Xamarin.Forms
 {
-	static class FontElement
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	public static class FontElement
 	{
 		public static readonly BindableProperty FontProperty =
 			BindableProperty.Create("Font", typeof(Font), typeof(IFontElement), default(Font),
--- a/Xamarin.Forms.Core/FontElement.cs
+++ b/Xamarin.Forms.Core/FontElement.cs
@@ -3,8 +3,7 @@ using Xamarin.Forms.Internals;
 
 namespace Xamarin.Forms
 {
-	[EditorBrowsable(EditorBrowsableState.Never)]
-	public static class FontElement
+	static class FontElement
 	{
 		public static readonly BindableProperty FontProperty =
 			BindableProperty.Create("Font", typeof(Font), typeof(IFontElement), default(Font),
--- a/Xamarin.Forms.Core/Binding.cs
+++ b/Xamarin.Forms.Core/Binding.cs
@@ -13,7 +13,7 @@ namespace Xamarin.Forms
 {
 	public sealed class Binding : BindingBase
 	{
-		internal const string SelfPath = ".";
+		public const string SelfPath = ".";
 		IValueConverter _converter;
 		object _converterParameter;
 
--- a/Xamarin.Forms.Core/GestureRecognizer.cs
+++ b/Xamarin.Forms.Core/GestureRecognizer.cs
@@ -2,7 +2,7 @@ namespace Xamarin.Forms
 {
 	public class GestureRecognizer : Element, IGestureRecognizer
 	{
-		internal GestureRecognizer()
+		public GestureRecognizer()
 		{
 		}
 	}
--- a/Xamarin.Forms.Core/IPaddingElement.cs
+++ b/Xamarin.Forms.Core/IPaddingElement.cs
@@ -1,6 +1,9 @@
-﻿namespace Xamarin.Forms
+﻿using System.ComponentModel;
+
+namespace Xamarin.Forms
 {
-	interface IPaddingElement
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	public interface IPaddingElement
 	{
 		//note to implementor: implement this property publicly
 		Thickness Padding { get; }
--- a/Xamarin.Forms.Core/NullEffect.cs
+++ b/Xamarin.Forms.Core/NullEffect.cs
@@ -1,6 +1,6 @@
 namespace Xamarin.Forms
 {
-	internal class NullEffect : Effect
+	public class NullEffect : Effect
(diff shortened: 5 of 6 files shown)


jsuarezruiz:
@VincentDondain `RelativeLayoutTests.UnitExpressionSearch` class I suppose that it is not something critical and it can be accessed via reflection.

Xamarin.Forms.Core/FontElement.cs
```diff
@@ -1,8 +1,10 @@
+using System.ComponentModel;
 using Xamarin.Forms.Internals;
 
 namespace Xamarin.Forms
 {
-	static class FontElement
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	public static class FontElement
```
jsuarezruiz, line 7: @VincentDondain Like anything that inherits from `IFontElement` uses those bindable properties https://github.com/xamarin/Xamarin.Forms/blob/5.0.0/Xamarin.Forms.Core/Label.cs#L38, technically you could just change all the places that it's using FontElement.<Property> to Label.<Property> (for example).

jsuarezruiz:
@StephaneDelcroix Could you take a look at these changes?, any feedback?

Xamarin.Forms.Core/FontElement.cs
VincentH-Net: @jsuarezruiz I think you mean me, @VincentH-Net ?

jsuarezruiz: Yes, updated the comment!.

VincentH-Net: Thx @jsuarezruiz I built against the packages from this PR; results:
IPaddingElement` also needs to be made public - I missed that one when documenting, my bad.

Better keep those changes; reverting them would require duplication of all extension methods targeting `IPaddingElement` and `IFontElement` for all view that implement them. Even worse, it would make those methods unavailable for any (current/future) views in XCT or other libraries that have padding and font properties

VincentH-Net:
> @VincentH-Net `RelativeLayoutTests.UnitExpressionSearch` class I suppose that it is not something critical and it can be accessed via reflection.

@jsuarezruiz I copied the class - only 25 lines

Xamarin.Forms.Core/FontElement.cs
VincentH-Net: @jsuarezruiz could you make `IFontElement` public? Thx

StephaneDelcroix: I would wait until we can have default interface members in XF, and get rid of the static *Element types

VincentH-Net: @StephaneDelcroix , @jsuarezruiz The decision whether to wait/release should take @davidortinau 's goal into account,  not to delay XF5 stable longer than needed:
> What's the status on moving all the planned things from xamarin/xamarin.forms to xamarin/xamarincommunitytoolkit?
@VincentH.NET @jfversluis @ShaneNeuville @samhouts @redth (tagging anyone that might know so we don't delay XF 5 stable longer than needed)
(Quote from [David's recent discus message](https://discord.com/channels/732297728826277939/738065609740517467/770032960429621268))

Please let me know when / what you decide; I am mostly done with the move and could complete quickly once this PR incl public `IPaddingElement` is merged

PureWeen: @VincentH-Net for now can we just replace your FontElement.BP and PaddingElement.BP with 

Label.BP

I know it's kind of ugly and it's not ideal but it's functionally the same and later down the road we can do what @StephaneDelcroix suggests

VincentH-Net:
@jsuarezruiz The [C# Markup Part 2 PR](https://github.com/xamarin/Xamarin.Forms/pull/11428) incudes a fix for a Forms issue that is still present in this PR branch:
> Fix for issue encountered by new C# Markup unit tests:
`BindableLayout.SetEmptyViewTemplate` sets `EmptyView` instead of `EmptyViewTemplate`
Possibly this fixes #9905

Here is [the fix](https://github.com/xamarin/Xamarin.Forms/pull/11428/files?file-filters%5B%5D=.cs#r514964441); it is a single statement change:
`b.SetValue(EmptyViewProperty, value);`  -> `b.SetValue(EmptyViewTemplateProperty, value);`

Can we include this fix in this PR? 
If not, I can create a new issue.

jsuarezruiz:
@VincentH-Net To keep track of that issue easily, do you mind creating another separate PR (or let me know and I'll create it)?

Xamarin.Forms.Core/FontElement.cs
jsuarezruiz: Yes, the @PureWeen suggestion is what I tried to agree with my first question. We can maintain the static Element types internal and open all the necessary interfaces. From the Markup extensions, replace `FontElement.{Property}` by `Label.{Property}`. @VincentH-Net What do you think?

VincentH-Net:
> @VincentH-Net To keep track of that issue easily, do you mind creating another separate PR (or let me know and I'll create it)?

Sure; I created issue https://github.com/xamarin/Xamarin.Forms/issues/12683 and will create PR for that

Xamarin.Forms.Core/FontElement.cs
VincentH-Net: Yes that is fine by me. I don't mind some internal ugliness, as long as it does not hurt the XCT users.

So OK to keep the static *Element types internal and the interfaces public.
(apologies that I am a bit late in responding; I was up to my eyebrows in the PR)

StephaneDelcroix (approved):

PR #12632, opened 2020-10-28
iOS: CollectionView: ItemsViewLayout: Fix null reference on invalidate layout

hemantbeast:
### Description of Change ###

On changing the layout of collection at runtime, the iOS app gets crashed showing an null reference exception on invalidate layout which clearly mentioned that CollectionView is getting null here. Whereas in Android it works just fine without any issue/exception.

### API Changes ###

 None

### Platforms Affected ### 

- iOS

### DIFF
--- a/Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewLayout.cs
+++ b/Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewLayout.cs
@@ -532,7 +532,9 @@ namespace Xamarin.Forms.Platform.iOS
 
 		public override void InvalidateLayout()
 		{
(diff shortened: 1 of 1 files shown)


rmarinho (approved):

hartez (approved):

roeldewit:
I don't think this is the correct solution for this problem. I'm using a CollectionView where all items have an icon, wich you can tap to reveal extra information. On Android this looks like this:
Android

But on iOS the null reference exception occured. So I downloaded the most recent preview version (5.0.0.1709-pre4) and tried it out. The exception doesn't occur anymore, but it's also not expanding correctly:

iOS

PR #12634, opened 2020-10-28
[C] get the name of a scoped object

### LINKED ISSUE #12608: [Enhancement] Add VisualElement.Name property
## Summary
Add support for getting element name matching x:Name in XAML, e.g. <Label x:Name='MyLabel'/>. So that Hot Reload can display it in Live Visual Tree (and thus users can search for named elements there). 

## API Changes
Add VisualElement.Name property as a first class citizen.

StephaneDelcroix:
- fixes #12608

### DIFF
--- a/Xamarin.Forms.Core/Internals/NameScope.cs
+++ b/Xamarin.Forms.Core/Internals/NameScope.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Linq;
 
 namespace Xamarin.Forms.Internals
 {
@@ -11,6 +12,7 @@ namespace Xamarin.Forms.Internals
 			BindableProperty.CreateAttached("NameScope", typeof(INameScope), typeof(NameScope), default(INameScope));
 
 		readonly Dictionary<string, object> _names = new Dictionary<string, object>();
+		readonly Dictionary<object, string> _values = new Dictionary<object, string>();
 
 		object INameScope.FindByName(string name)
 			=> _names.TryGetValue(name, out var element) ? element : null;
@@ -21,7 +23,13 @@ namespace Xamarin.Forms.Internals
 				throw new ArgumentException($"An element with the key '{name}' already exists in NameScope", nameof(name));
 
 			_names[name] = scopedElement;
+			_values[scopedElement] = name;
 		}
+
+		//used by VS Live Visual Tree
+		internal string NameOf(object scopedObject)
+			=> _values.TryGetValue(scopedObject, out var name) ? name : null;
+		
 
 		public static INameScope GetNameScope(BindableObject bindable) => (INameScope)bindable.GetValue(NameScopeProperty);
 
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Core/Internals/NameScope.cs
```diff
@@ -23,6 +24,15 @@ void INameScope.RegisterName(string name, object scopedElement)
 			_names[name] = scopedElement;
 		}
 
+		//used by VS Live Visual Tree
+		internal string NameOf(object scopedObject)
+		{
+			var index = _names.Values.IndexOf(scopedObject);
+			if (index < 0)
+				return null;
+			return _names.Keys.Skip(index - 1).FirstOrDefault();
```
campersau, line 33: Does this work correctly if the Value is found at index `0` or `1`?

```c#
"abc".Skip(0 - 1).FirstOrDefault(); // => "a"
"abc".Skip(1 - 1).FirstOrDefault(); // => "a"
```

StephaneDelcroix: I didn't meant the `-1`. thanks for spotting

StephaneDelcroix, line 33:
```suggestion
			return _names.Keys.Skip(index).FirstOrDefault();
```

rmarinho:
Can we add a unit test?

Xamarin.Forms.Core/Internals/NameScope.cs
```diff
@@ -23,6 +24,15 @@ void INameScope.RegisterName(string name, object scopedElement)
 			_names[name] = scopedElement;
 		}
 
+		//used by VS Live Visual Tree
+		internal string NameOf(object scopedObject)
+		{
+			var index = _names.Values.IndexOf(scopedObject);
```
etvorun, line 30: Linear search is not ideal. Hot Reload will call it for each element in visual tree.

StephaneDelcroix: linear search on HR scenario is better than doing paying the perf of caching keys and values for every other scenarios (running apps, when HR is not plugged in)

etvorun: At a minimum you should do one linear search rather than two (first iterate thru values, then thru keys). A better way would be to enum _names, i.e.
`foreach(var entry in _names) { if (entry.Value == element) return entry.Key; } return null;`
I my little I see that double iteration does take twice the time.

I think realistically names elements will make up 10%-30% of all elements, i.e. adding another dictionary is not expensive memory usage wise. However perf gain will be significant. I my little app I creates 2,000 elements and named only 100 of them. Then looked up names for each of elements and aggregated total time. Note that because 1,900 elements did not have names we had to do full linear search for those elements. The numbers are:
  Total double linear search (current PR): 3.5ms
  Total single  linear search ("small fix" above): 1.5ms
  Total reverse dictionary lookup: 0.2ms
This is in a tiny stand alone app. In real app this might vary. I also tried different scenarios with high vs low percentage of names elements. On average linear search is 10 times slower.

StephaneDelcroix: > I think realistically names elements will make up 10%-30% of all elements

that's a good point

etvorun (approved):

rmarinho (approved):

PR #12736, opened 2020-11-04
[Android] SwipeView used in a CollectionView fires the SelectionChangedCommand unintentionally

### LINKED ISSUE #11982: [Bug] [Android]
### Description
On Android (as of XF 4.7 and still in XF 4.8 sr1), using a SwipeView as a root container for a CollectionView DataTemplate, if the CollectionView uses the SelectionChangedCommand and the content of the SwipeView (ItemSource bound to a list of viewmodels) has some element (eg Button) with a command responding to the user tapping, the tap on that element improperly fires the SelectionChangedCommand on the CV (either completely ignoring the element tapped or firing both the element command and the CV SelectionChangedCommand).
See attached sample project.

### Steps to Reproduce

1. Have a CollectionView using a SelectionChangedCommand, also bind its ItemSource to a viewmodel that has a command to be later bound to an interactive element such as a button, ecc...
2. As its DataTemplate root element use a SwipeView.
3. Inside the SwipeView add an input element such as a Button and bind that to the command in the viewmodel.

### Expected Behavior
When the element inside the SwipeView is tapped you would expect to just fire the command of the viewmodel just like when you do not have a SwipeView, but most importantly you would not expect the SelectionChangedCommand of the CV to be fired at all.

### Actual Behavior
The SelectionChangedCommand is always fired. Also, the button Command is not always fired (it seems to be fired if you tap on the rightmost side).

### Basic Information

- Version with issue: XF 4.7, still happens in XF 4.8 sr1
- IDE: Visual Studio Pro for Mac  8.7.4
- Platform Target Frameworks:
  - Android: tested on Android 9 and 10.

### Reproduction Link
Attached sample project (in MainPage.xaml comment/uncomment with/without SwipeView to see the two behaviours)
[SwipeViewBug1.zip](https://github.com/xamarin/Xamarin.Forms/files/5154676/SwipeViewBug1.zip)

### Workaround
Using the SwipeView not as a DataTemplate root element but as a child element that is contained inside a Grid/ecc... seems to solve the problem, but it is not ideal.

jsuarezruiz:
### Description of Change ###

Avoid fire the CollectionView SelectionChangedCommand event if the user tap a Button for example. With the change from this PR, the behavior is the same the behavior as without using SwipeView.

### Issues Resolved ### 

- fixes #11982

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

None

#### Before
issue11982

#### After
fix11982

### Testing Procedure ###
Launch Core Gallery and navigate to the issue 11982. In the CollectionView using the SwipeView, tap the Button. If not trigger the SelectionChanged, the test has passed.

### PR Checklist ###


jsuarezruiz:
Added changes to fix the build error.

rachelkang:
Perhaps not within the scope of this PR, but what's the expected behavior on iOS? I'm seeing that it's different on iOS than on Android

Also, what's the expected behavior when clicking on the "Command" button itself? I sometimes get the dialog and I sometimes see the "-PRESSED" added on

rmarinho:
@jsuarezruiz  build is failing

rmarinho:
@jsuarezruiz  can we UITest this?

jsuarezruiz:
Sure

jsuarezruiz:
@rachelkang Could you try again? The behavior should be the same on Android and iOS.

Xamarin.Forms.Platform.Android/Renderers/SwipeViewRenderer.cs
```diff
@@ -283,6 +286,14 @@ bool ShouldInterceptScrollChildrenTouch(SwipeDirection swipeDirection)
...
 		}
 
+		bool IsViewInBounds(AView view, int x, int y)
+		{
+			ARect outRect = new ARect();
+			view.GetHitRect(outRect);
+
+			return x > outRect.Left && x < outRect.Right && y > outRect.Top && y < outRect.Bottom;
+		}
+
```
rachelkang, line 296: Why did you move this code up?

rachelkang (commented):
No longer seeing the different behaviors described in last comment

rmarinho (changes requested):
Build is failing 

```
D:\agent\1\s\Xamarin.Forms.Controls\Xamarin.Forms.Controls.csproj" (GetTargetFrameworks target) (9:21) ->
  D:\agent\1\s\Xamarin.Forms.Controls\Xamarin.Forms.Controls.csproj(160,3): error MSB4024: The imported project file "D:\agent\1\s\Xamarin.Forms.Controls.Issues\Xamarin.Forms.Controls.Issues.Shared\Xamarin.Forms.Controls.Issues.Shared.projitems" could not be loaded. The 'EmbeddedResource' start tag on line 2026 position 6 does not match the end tag of 'ItemGroup'. Line 2078, position 5.
```

jfversluis:
Not sure if this fix is still correct and seeing where we are in the Xamarin lifecycle, we probably won't be merging this anymore

PR #12039, opened 2020-09-06
Shell navigating deferral token

PureWeen:
### Description of Change ###

Provide a deferral token during navigating that users can use to invoke async operations 

- Implementation uses the existing "cancel" behavior to "cancel" user initiated navigation. Once they complete the deferral then it uses GotoAsync to replay the navigation
- `GotoAsync` and all of the `INavigation` methods will not resolve until the user has completed all deferrals
- If user tries to navigate while there is a pending deferral then GoToAsync will throw an exception
- reworked all the `INavigation` methods to route through `GoToAsync` so that all behavior is symmetrical

### Issues Resolved ### 
- fixes #10638

### API Changes ###

```C#
class ShellNavigatingDeferral
{
    public void Complete();
}
```

```C#
class ShellNavigatingEventArgs
{
     public ShellNavigatingDeferral GetDeferral()
}

```

### Platforms Affected ### 
- Core/XAML (all platforms)
- iOS
- Android
- UWP

### Testing Procedure ###
- Unit tests included

### Example usage

Given the following override of OnNavigating if the user tries to do any navigation they will be given a display sheet confirming if the want to navigate. If they click "no" then navigation will be cancelled

If the user calls `GoToAsync` the GoToAsync call will only resolve after the deferral token has been completed

```C#
public class MyShell : Shell
{
		protected override async void OnNavigating(ShellNavigatingEventArgs args)
		{
			base.OnNavigating(args);

			var token = args.GetDeferral();

			var result  = await DisplayActionSheet(
					"Are you sure?",
					"cancel",
					"destruction",
					"Yes", "No");

			if (result != "Yes")
				args.Cancel();

			token.Complete();
		}
```

#### DEMO

F6633D2C-EBDD-4E75-950B-5B7E1998AA33

### PR Checklist ###

- [ ] Targets the correct branch
- [ ] Tests are passing (or failures are unrelated)

PR #12756, opened 2020-11-06
WebView Navigation Can Now Be Cancelled Using A Task #12720

### LINKED ISSUE #12720: [Bug] WebView Navigating Race Condition
### Description
_This issue report is both a bug AND an enhancement. I'm essentially saying that the current functionality should be changed because it is very inflexible. The issue is caused by the way the `WebViewRenderer` and `WebView` are interacting with the base platforms, The functionality I am describing in this report is easy to implement in a native app. I would call this a design bug as opposed to a code bug._

The `Xamarin.Forms.WebView` has the event `Navigating`. This event passes `WebNavigatingEventArgs` which contain the property `Cancel`. When the `Navigating` event is fired, you can override the navigation event and prevent it from occuring by setting `e.Cancel = false`.

This is useful functionality if you want to prevent the user from browsing to certain sites. The issue arises when you want to evaluate whether the user can navigate to a url using a background thread. If you subscribe to the `WebView.Navigating` method your code may look like:

```csharp
private void OnNavigating(object sender, WebNavigatingEventArgs e)
{
    var allowed = CanUserNavigateToSite(e.Url)

    e.Cancel = !allowed;
}
```

This code executes no problem, but If I say want to something more complicated like talk to an API to see if the url's domain name is allowed, the code will not work.

```csharp
private async void OnNavigating(object sender, WebNavigatingEventArgs e)
{
    var allowed = await CanUserNavigateToSiteAsync(e.Url)

    e.Cancel = !allowed;
}
```

This is caused because at the renderer level, the `Navigating` event is sent to the forms object to raise, then the value of the args is used to determine whether navigation should be cancelled. If your code can complete fast enough then you can prevent the navigation, if you do some lengthy task (say using `async await`) then the native code will have already executed before your task completed allowing navigation to the site, even if you didn't want it.

### Steps to Reproduce

I've provided a fully coded example

1. Go to my example reproduction repository: [https://github.com/Axemasta/WebViewNavigatingIssue](https://github.com/Axemasta/WebViewNavigatingIssue)
2. Read the [README](https://github.com/Axemasta/WebViewNavigatingIssue/blob/main/README.md), I have explained the issue, my code and proposed a solution (open to feedback).
3. Pull the code and run the app, on each tab try using google to search for & browse to Facebook. When you are executing the code on the same thread the site access is blocked. When you are executing the code on a background thread the browser navigates before evaluating the site access request thus demonstrating the race condition that exists in the `Xamarin.Forms` code.

### Expected Behavior

`Xamarin.Forms.WebView` can be prevented from navigating using asynchronous code (`WebView.CanNavigate = await _myService.EvaluateCanNavigate("https://google.co.uk")`).

### Actual Behavior

`Xamarin.Forms.WebView` navigates before evaluating asynchronous code, which returns after the navigation has occurred.

### Basic Information

- Version with issue: 4.8.0.1560 (but all issues appear to be affected)
- Last known good version: ?
- IDE: Visual Studio For Mac 2019 / Visual Studio Professional 2019
- Platform Target Frameworks: 
  - iOS:  14.1
  - Android: 9.0
- Android Support Library Version: AndroidX
- Nuget Packages:  Out of the box only
- Affected Devices: All

### Screenshots

**Sample app browsing using both sync & async**
_Both facebook and twitter should be banned from navigation_
bug_example_browser

**Console output when using sync method**

sync_output

**Console output when using async method**

async_output

### Reproduction Link

[WebViewNavigatingIssue](https://github.com/Axemasta/WebViewNavigatingIssue)

### Workaround

Requires code changes at both the `Xamarin.Forms.WebView` level and at the native renderer level.

Axemasta:
### Issues Resolved ### 

- fixes #12720 

### API Changes ###

Added:
 - `public Func<Task<bool>> CancelTask { get; set; }` to `WebNavigatingEventArgs`

### Platforms Affected ### 

- iOS
- Android

By default there are no behavioural changes, developers would have to opt into this functionality.

I've added the ability for the `WebView` to cancel navigation asynchronously. In the current implementation the `WebView` raised `WebNavigatingEventArgs` which contain `public bool Cancel { get; set; }`. The value is mutable and will be read by the control to determine whether or not to cancel navigation.

From [WkWebViewRenderer, Line 777-781](https://github.com/xamarin/Xamarin.Forms/blob/496204e262bc3d661cc0d693d966d02e9f588f74/Xamarin.Forms.Platform.iOS/Renderers/WkWebViewRenderer.cs#L777):
```csharp
var args = new WebNavigatingEventArgs(navEvent, new UrlWebViewSource { Url = lastUrl }, lastUrl);

WebView.SendNavigating(args);
_renderer.UpdateCanGoBackForward();
decisionHandler(args.Cancel ? WKNavigationActionPolicy.Cancel : WKNavigationActionPolicy.Allow);
```

As you can see if `args.Cancel` is mutated (in this case by the developer writing their forms app) the navigation will be cancelled. However if it takes too long for the developer to update the `args.Cancel` value then the native control will have already navigated. This becomes restrictive when trying to use `Xamarin.Forms` to build a `WebView` based application as conditional access is an important feature.

I have implemented this functionality on iOS and Android, I had raised an issue (#12720) but theres been no discussion on the issue so I've implemented what I think should be updated and want to see what peoples thoughts are.

Not applicable - No visual changes have been made & the behaviour is difficult to show in a gif.

### Testing Procedure ###
iOS and Android are currently the only platforms affected.

I used the controls gallery app in the `Xamarin.Forms` solution to test my changes. I created a page with a standard `WebView` and verified that the `Navigating` event was being fired. I then wrote a method that would return whether or not I could navigate to a site (synchronously) and tested it was behaving as expected. I then wrapped that method in. a task (and ran a `await Task.Delay(2000)`) and ran that method, this allowed me to navigate to a site eventhough when my task completed it reported that navigation to site should be blocked. I implemented my code changes ensuring that when i had updated the code the original behaviour was the same as before I did the changes. I didn't check in any changes to the control app etc, I know the [Xamarin Community Toolkit](https://github.com/xamarin/XamarinCommunityToolkit) has a sample app where you can put your new feature / change. 

Once the code was complete I increased the delay on my async task and tested using the WebView. I ran the following test where the `facebook.com` domain was banned:
- Load `https://google.co.uk`
- Seach for `"facebook"`
- Click any link to '*.facebook.com*`

Google would load and work normally, when clicking through to facebook nothing would happen, I would see through debug logs that the site was being blocked. Whilst the task is running the UI thread is not blocked so the `WebView` can be interacted with until the task completes.

### PR Checklist ###

### DIFF
--- a/Environment.Build.props
+++ b/Environment.Build.props
@@ -13,7 +13,6 @@
     <!--
       Set one of these if you want to test a particular set of renderers
       <ANDROID_RENDERERS>FAST</ANDROID_RENDERERS>
-      <ANDROID_RENDERERS>PREAPPCOMPAT</ANDROID_RENDERERS>
       <ANDROID_RENDERERS>LEGACY</ANDROID_RENDERERS>
     -->
   </PropertyGroup>
--- a/Xamarin.Forms.ControlGallery.Android/CustomRenderers.cs
+++ b/Xamarin.Forms.ControlGallery.Android/CustomRenderers.cs
@@ -51,12 +51,8 @@ using NestedScrollView = global::AndroidX.Core.Widget.NestedScrollView;
 [assembly: ExportRenderer(typeof(Issue8801.PopupStackLayout), typeof(Issue8801StackLayoutRenderer))]
 [assembly: ExportRenderer(typeof(Xamarin.Forms.Controls.Tests.TestClasses.CustomButton), typeof(CustomButtonRenderer))]
 [assembly: ExportRenderer(typeof(ScrolView11185), typeof(ScrollViewFadeRenderer))]
-
-#if PRE_APPLICATION_CLASS
-#elif FORMS_APPLICATION_ACTIVITY
-#else
 [assembly: ExportRenderer(typeof(FlyoutPage), typeof(NativeDroidFlyoutPage))]
-#endif
+
 namespace Xamarin.Forms.ControlGallery.Android
 {
 	public sealed class ScrollViewFadeRenderer : ScrollViewRenderer
@@ -909,11 +905,7 @@ namespace Xamarin.Forms.ControlGallery.Android
 
 #pragma warning disable CS0618 // Leaving in old constructor so we can verify it works
 	public class NoFlashTestNavigationPage
-#if FORMS_APPLICATION_ACTIVITY
-		: Xamarin.Forms.Platform.Android.NavigationRenderer
-#else
 		: Xamarin.Forms.Platform.Android.AppCompat.NavigationPageRenderer
-#endif
 	{
 #if !FORMS_APPLICATION_ACTIVITY
 		protected override void SetupPageTransition(FragmentTransaction transaction, bool isPush)
@@ -926,11 +918,7 @@ namespace Xamarin.Forms.ControlGallery.Android
 
 #pragma warning disable CS0618 // Leaving in old constructor so we can verify it works
 	public class QuickCollectNavigationPageRenderer
-#if FORMS_APPLICATION_ACTIVITY
-		: Xamarin.Forms.Platform.Android.NavigationRenderer
-#else
 		: Xamarin.Forms.Platform.Android.AppCompat.NavigationPageRenderer
-#endif
 	{
 		bool _disposed;
 		NavigationPage _page;
--- a/Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
+++ b/Xamarin.Forms.ControlGallery.Android/FormsAppCompatActivity.cs
@@ -1,6 +1,4 @@
-﻿#if !FORMS_APPLICATION_ACTIVITY && !PRE_APPLICATION_CLASS
-
-using Android.App;
+﻿using Android.App;
 using Android.Content;
 using Android.Content.PM;
 using Android.OS;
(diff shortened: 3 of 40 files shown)


rmarinho:
This needs rebase on top of Main.

Axemasta:
@rmarinho  I've rebased on main, I also marked the PR as draft because I want to gauge opinion on what I've proposed here 😁

rmarinho:
@PureWeen  is this similar to what you did for navigation in shell? should we use the same/similar approach ?

PureWeen:
@rmarinho yes!

I think we should go the same approach as shell with this
https://github.com/xamarin/Xamarin.Forms/pull/12039

- It's closer to what UWP does
- The advantage of a deferral token is that multiple subscribers can influence the deferral. If you're relying on a task then coordination between multiple subscribers can be awkward

Axemasta:
@PureWeen  I'll have a look at this PR and update mine accordingly 😄

Nice to know there is a pattern I can follow, I wasn't happy with how mine needed an extra property on the event args, when I was originally spitballing this code I was adding methods to set on the webview to invoke the function etc, quite nasty imho.

Axemasta:
I'm closing this PR as I've opened up a new one with the new implementation (deferral token) discussed here #14137

Issue #12762, opened 2020-11-06
[Bug] Java.Lang.IndexOutOfBoundsException: setSpan


transis2:
### Description
IndexOutOfBoundsException observed in Xamarin.Android/Xamarin.Forms code in the wild.

Stack trace:

```
JniEnvironment+InstanceMethods.CallNonvirtualVoidMethod (Java.Interop.JniObjectReference instance, Java.Interop.JniObjectReference type, Java.Interop.JniMethodInfo method, Java.Interop.JniArgumentValue* args)
JniPeerMembers+JniInstanceMethods.InvokeVirtualVoidMethod (System.String encodedMember, Java.Interop.IJavaPeerable self, Java.Interop.JniArgumentValue* parameters)
EditText.SetSelection (System.Int32 index)
EntryRendererBase`1[TControl].UpdateText ()
EntryRendererBase`1[TControl].OnElementPropertyChanged (System.Object sender, System.ComponentModel.PropertyChangedEventArgs e)
(wrapper delegate-invoke) <Module>.invoke_void_object_PropertyChangedEventArgs(object,System.ComponentModel.PropertyChangedEventArgs)
BindableObject.OnPropertyChanged (System.String propertyName)
Element.OnPropertyChanged (System.String propertyName)
BindableObject.SetValueActual (Xamarin.Forms.BindableProperty property, Xamarin.Forms.BindableObject+BindablePropertyContext context, System.Object value, System.Boolean currentlyApplying, Xamarin.Forms.Internals.SetValueFlags attributes, System.Boolean silent)
BindableObject.SetValueCore (Xamarin.Forms.BindableProperty property, System.Object value, Xamarin.Forms.Internals.SetValueFlags attributes, Xamarin.Forms.BindableObject+SetValuePrivateFlags privateAttributes)
BindableObject.SetValueCore (Xamarin.Forms.BindableProperty property, System.Object value, Xamarin.Forms.Internals.SetValueFlags attributes)
Element.SetValueFromRenderer (Xamarin.Forms.BindableProperty property, System.Object value)
TextTransformUtilites.SetPlainText (Xamarin.Forms.InputView inputView, System.String nativeText)
ITextWatcher.OnTextChanged (Java.Lang.ICharSequence s, System.Int32 start, System.Int32 before, System.Int32 count)
[...]

PR #12764, opened 2020-11-06
[Android] Fix to Issue Java.Lang.IndexOutOfBoundsException: setSpan

transis2:
Fix to Issue #12762
Java.Lang.IndexOutOfBoundsException: setSpan (nn ... nn) ends beyond length n-1

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/EntryRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/EntryRenderer.cs
@@ -538,7 +538,14 @@ namespace Xamarin.Forms.Platform.Android
 			EditText.Text = text;
 			if (EditText.IsFocused)
 			{
-				EditText.SetSelection(text.Length);
+                try
+                {
+					EditText.SetSelection(text.Length);
+                }
+				catch (System.Exception ex)
+				{
+					Internals.Log.Warning("Entry", $"Failed to set Control.Selection from CursorPosition/SelectionLength: {ex}");
+				}
 				EditText.ShowKeyboard();
 			}
 		}
--- a/Xamarin.Forms.Platform.Android/Renderers/EntryRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/EntryRenderer.cs
@@ -541,7 +541,7 @@ namespace Xamarin.Forms.Platform.Android
 			EditText.Text = text;
 			if (EditText.IsFocused)
 			{
-				EditText.SetSelection(text.Length);
+				EditText.SetSelection(EditText.Text.Length);
 				EditText.ShowKeyboard();
 			}
 		}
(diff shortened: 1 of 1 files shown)


rmarinho:
I m not sure this is the right fix.. Maybe we need something like this:

```
void UpdateText()
{
	if (EditText == null || Element == null)
		return;

	var text = Element.UpdateFormsText(Element.Text, Element.TextTransform);

	if (EditText.Text == text)
		return;

	EditText.Text = text;
	if (EditText.IsFocused)
	{
		EditText.SetSelection(EditText.Text.Length);
		EditText.ShowKeyboard();
	}
}
```

transis2:
@rmarinho Yes, your solution is better.
Just to learn: You set the text to entry and the control will limit the size of string. Só, it will never há e an selection bigger than the size of the field, wright?

Do you want to create an PR or I can change mine with your suggestion?

rmarinho:
@transis2  yes that's my idea, or with the check above if EditText is null (was disposed) we just return to also make it safer. 

I tried to reproduce this issue with a number ways but wasn't able, but that's the only idea i have right now to try to fix it.  
You can update your PR.

transis2:
@rmarinho Update the fix with your suggestion.

Thanks!

transis2:
Hi @rmarinho 

Our PR is showing thar some checks were not successful.
May I create a new one from the latest XF version ?

rmarinho:
@transis2  i think i fixed the issue, running a new build and tests

felipemomm:
@rmarinho When is it going to be merged? I am facing the same issue in my project, so I had to implement my own UpdateText method on OnElementPropertyChanged to avoid it (XF 4.8).

nrmiller:
@rmarinho Is this PR still being considered? It looks like it has been around for awhile, and I'm running into this error as late as Xamarin.Forms 5.0.0.2021 on a Google Pixel 3 XL

My scenario to trigger this bug with 100% reliability:
1. Set Entry.MaxLength to 5 via XAML
2. Enter "3910" with numeric keyboard.
3. Without dismissing the keyboard, tap a radio button* to trigger setting the text to "1539.37" via a view-model property.

*Specifically, I have two radio buttons (1 for centimeters and 1 for inches). The goal is to convert what the user enters upon changing the select radio button. Because converting 3910cm yields 1539.37in and because "1539.37" > 5 characters, the crash occurs.
I tried to trigger this issue by creating a simple button and setting the Text to "000000" in the bound view-model command. This did not work, so I'm assuming something tricky is happening specifically with triggering via `RadioButtonGroup.SelectedValue`. Perhaps this is related to how MessagingCenter is used under the hood?

FWIW I've attached the stacktrace from the Xamarin Forms side (I'm stilling getting the IndexOutOfBounds on the Android side)
[entry text stacktrace.txt](https://github.com/xamarin/Xamarin.Forms/files/6781453/entry.text.stacktrace.txt)

PR #12790, opened 2020-11-09
Check for Internet Connection before running tests that Require an Internet Connection

PureWeen:
### Description of Change ###

- Categorize tests that require internet access that are currently failing on App Center
- Before these tests run verify if the device has a valid internet connection. If it doesn't then just mark the test as Inconclusive
- For the most part tests inside of AppCenter run without issue but the internet connection on the device seems to fail enough that we need to detect for this and not fail the whole group but just mark a given set inconclusive

### Platforms Affected ### 
- iOS
- Android
- UWP

### Testing Procedure ###
- Make sure UI tests all pass

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2354.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2354.cs
@@ -17,6 +17,7 @@ namespace Xamarin.Forms.Controls.Issues
 	[NUnit.Framework.Category(Core.UITests.UITestCategories.Github5000)]
 	[NUnit.Framework.Category(Core.UITests.UITestCategories.Image)]
 	[NUnit.Framework.Category(Core.UITests.UITestCategories.ListView)]
+	[NUnit.Framework.Category(UITestCategories.RequiresInternetConnection)]
 #endif
 	[Preserve(AllMembers = true)]
 	[Issue(IssueTracker.Github, 2354, "ListView, ImageCell and disabled source cache and same image url", PlatformAffected.iOS | PlatformAffected.Android)]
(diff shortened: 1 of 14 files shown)


Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2354.cs
```diff
@@ -8,12 +8,15 @@
...
 using NUnit.Framework;
 using Xamarin.UITest;
+using Xamarin.Forms.Core.UITests;
 #endif
 
 namespace Xamarin.Forms.Controls.Issues
 {
 #if UITEST
 	[NUnit.Framework.Category(Core.UITests.UITestCategories.Github5000)]
+	[NUnit.Framework.Category(Core.UITests.UITestCategories.Image)]
```
hartez, line 18: Doesn't this one need the RequiresInternetConnection attribute?

hartez (changes requested):

Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2354.cs
PureWeen: It's on the test itself

https://github.com/xamarin/Xamarin.Forms/blob/e4a86cef2fc24fcfce5db1f885e710c137e0bf65/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue2354.cs#L127

I couldn't find a straightforward way to read the categories off the class other than scanning the assemblies but might as well just scan the assembly :-)

https://github.com/xamarin/Xamarin.Forms/pull/12790/files#diff-ad2af0a0a8e3e439c59c593a96c27eddcea548a938c7a47e02bd1ecc87bafa35R476

rmarinho (approved):

hartez (approved):

PR #12871, opened 2020-11-16
[iOS] Fix pushing same page as a modal on iOS13 in split mode

### LINKED ISSUE #8988: [Bug] App freezes on iPadOS 13.3 when in split view mode (multi-tasking) and change between pages
### Description
I have come across a very strange bug -- not sure if it in Xamarin.Forms or iOS itself.

My app switches between several pages derived from ContentPage. When I have my app and another app, for example the iOS "Pages" app on my iPad side-by-side (e.g., half screen), changing pages more than once can cause my app to freeze up (no interaction) and also background color goes to white.  But, if I move the slider to make it a 2/3 split (or 1/3, or full screen), it immediately works.

When my app is to the left in 2/3 screen mode, the result is a bit comical --  first half of my app is OK (the leftmost third of the screen), but other half (middle third of screen) refuses to allow interaction (and has white background). Weirdest thing -- if the buttons straddles the break, the left part works fine, the right does not. LOL.

I've tried this on an iPad with 12.4.1 and 10.3.2 and I do not see this behavior.

I was able to create a simple app that demonstrates this, which I am attaching. It is "shared code", which I have stripped down to only iOS, and made minimal changes to.  (I checked and see bug with .NET standard too.)   Also attached is a screen shot.

MY IMPRESSION:  This oddness was introduced with either iPadOS 13 or 13.2

### Steps to Reproduce

1. Build attached project, SplitViewFreezeSample.  Start App on an iPad with 13.2 or above.
2. Start a secod app, placing it to the right of this app.
3. On App, click "Next Page" button at bottom, then "Go Back to Main Page", then "Next Page" again.

### Expected Behavior
Switching between pages should work indefinitely, and does when not in split screen.

### Actual Behavior
All or portion of app does not display properly, and buttons do not work.

### Basic Information
- Version with issue:  XF Forms 4.4, 4.3
- iPadOS 13.3 running on iPad Air 2

- Last known good version:
- IDE:  VS PC 16.4.1
- Platform Target Frameworks: 
  - iOS:  8 & 13.2
- Nuget Packages: none
- Affected Devices: iPad

### Screenshots
SplitViewFreezeSmall

### Reproduction Link
[SplitViewFreezeSample.zip](https://github.com/xamarin/Xamarin.Forms/files/3989203/SplitViewFreezeSample.zip)

rmarinho:
### Description of Change ###

There's a issue on iOS13  that pushing a modal using your app in split mode, (side by side with other app),  it causes issues rendering the second time. 
The particular issue is also related with retaining the pushed page , it keeps it's bounds, and we need to force the layout.
Workaround could be to just reset is bounds when is popped.

### Issues Resolved ### 

- fixes #8988

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS

None

### Testing Procedure ###

Go to Issue8988 , click Next, check the page loaded, click Go Back, then click again Next and make sure the page loads fine.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/DisposeHelpers.cs
+++ b/Xamarin.Forms.Platform.iOS/DisposeHelpers.cs
@@ -33,14 +33,15 @@ namespace Xamarin.Forms.Platform.MacOS
 #if __MOBILE__
 					if (renderer.ViewController != null)
 					{
-
-						var modalWrapper = renderer.ViewController.ParentViewController as ModalWrapper;
-						if (modalWrapper != null)
+						if (renderer.ViewController.ParentViewController is ModalWrapper modalWrapper)
 							modalWrapper.Dispose();
 					}
 #endif
 					renderer.NativeView.RemoveFromSuperview();
 					renderer.Dispose();
+					//reset layout size so we can re-layout the layer if we are reusing the same page.
+					//fixes a issue on split apps on iPad and iOS13
+					visualElement.Layout(new Rectangle(0, 0, -1, -1));
 				}
 				view.ClearValue(Platform.RendererProperty);
 			}
--- a/Xamarin.Forms.Platform.iOS/DisposeHelpers.cs
+++ b/Xamarin.Forms.Platform.iOS/DisposeHelpers.cs
@@ -39,11 +39,12 @@ namespace Xamarin.Forms.Platform.MacOS
 #endif
 					renderer.NativeView.RemoveFromSuperview();
 					renderer.Dispose();
-					//reset layout size so we can re-layout the layer if we are reusing the same page.
-					//fixes a issue on split apps on iPad and iOS13
-					visualElement.Layout(new Rectangle(0, 0, -1, -1));
 				}
 				view.ClearValue(Platform.RendererProperty);
+				//reset layout size so we can re-layout the layer if we are reusing the same page.
+				//fixes a issue on split apps on iPad and iOS13
+				if (Forms.IsiOS13OrNewer && visualElement is Page)
+					visualElement.Layout(new Rectangle(0, 0, -1, -1));
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Platform.iOS/ModalWrapper.cs
```diff
@@ -49,9 +49,6 @@ internal ModalWrapper(IVisualElementRenderer modal)
 				PresentationController.Delegate = this;
 
 			((Page)modal.Element).PropertyChanged += OnModalPagePropertyChanged;
-
-			if (Forms.IsiOS13OrNewer)
```
rmarinho, line 53: this was duplicated

Xamarin.Forms.Platform.iOS/DisposeHelpers.cs
```diff
@@ -33,14 +33,15 @@ internal static void DisposeModalAndChildRenderers(this Element view)
...
-						if (modalWrapper != null)
+						if (renderer.ViewController.ParentViewController is ModalWrapper modalWrapper)
 							modalWrapper.Dispose();
 					}
 #endif
 					renderer.NativeView.RemoveFromSuperview();
 					renderer.Dispose();
+					//reset layout size so we can re-layout the layer if we are reusing the same page.
+					//fixes a issue on split apps on iPad and iOS13
+					visualElement.Layout(new Rectangle(0, 0, -1, -1));
```
rmarinho, line 44: The fix is here.

jsuarezruiz (approved):

PureWeen (changes requested):
@rmarinho curious your thoughts on this approach.

https://github.com/xamarin/Xamarin.Forms/compare/fix-8988-shane

rmarinho:
@PureWeen  just tested, and yeah looks a better fix

PureWeen (approved):

jsuarezruiz (approved):

PR #12872, opened 2020-11-16
[Core] Propagate BindingContext to Path Transforms

jsuarezruiz:
### Description of Change ###

Correctly propagate the BC from the Path to the Path RenderTransforms.

### Issues Resolved ### 
There really is no issue registered, David has notified me detecting the issue creating https://github.com/davidortinau/easings

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

fix-pbc-transform

### Testing Procedure ###
Launch Core Gallery and navigate to the Shapes galleries. Select the new "Path Transform BindingContext Gallery" and modify the Sliders. If changing the Sliders transform the Path, the test has passed.

### PR Checklist ###


StephaneDelcroix (changes requested):
and then we'll realise that, oh, DynamicResources do not work. Why not hitting 2 birds with a single stone ? Do _not_ set Inherited context, but instead, properly parent the Elements (promote from BO if necessary)

jsuarezruiz:
@StephaneDelcroix Mnn, correct. Although we would have to change `Transform` from BindableObject to Element.

PR #12895, opened 2020-11-18
[iOS] Fixed RequestedThemeChanged raised twice issue

### LINKED ISSUE #12512: [Bug] Application.RequestedThemeChanged raised twice on iOS when switching to home screen and back
### Steps to Reproduce

1. Create a Xamarin.Forms app that listens to the `Application.RequestedThemeChanged` event and set a breakpoint on the event handler (or log it somewhere)
2. Launch app on iOS Simulator
3. App will launch, life is good
4. Switch away from app back to the device home screen and then switch back to your app

### Expected Behavior

No ThemeChanged event should fire at all (because it didn't change!).

### Actual Behavior

The ThemeChanged event fires twice: once to the theme that it _isn't_ and then once back to the actual current theme.

### Basic Information

- Version with issue: `<PackageReference Include="Xamarin.Forms" Version="4.8.0.1451" />`
- Affected Devices: iOS

### Screenshots

Watch video here:
[XamarinDoubleThemeChangeVideo.zip](https://github.com/xamarin/Xamarin.Forms/files/5394542/XamarinDoubleThemeChangeVideo.zip)

### Reproduction Link

Clone repo: https://github.com/Eilon/XamarinThemeChange

### Workaround

No workaround needed per se because you ultimately get the right theme. But my app does extra work because of the double theme change.

jsuarezruiz:
### Description of Change ###

The issue is in fact an iOS feature. Is doing multiple snapshots of the app's UI that will be presented in the App Switcher. And since the user could change to Dark Mode while your app is in the background, iOS does snapshots in both interface styles to always show the correct one.

Happens something like: The system takes a screenshot of your Light UI, switches to Dark (first trigger), takes a screenshot (off-screen), and switches back to Light (second trigger). And we have raised twice `RequestedThemeChanged`.

This PR include changes to only raise `RequestedThemeChanged` if the App is not in Background. Why?. To avoid raise the event  while app snapshots are being taken when going to background. In that way, the App raise the event just 1 time when pass from background to foreground.

### Issues Resolved ### 

- fixes #12512.

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###
Launch Core Gallery and navigate to the issue 12512. Then, follow the instructions.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
@@ -364,7 +364,9 @@ namespace Xamarin.Forms.Platform.iOS
 		{
 			base.TraitCollectionDidChange(previousTraitCollection);
 
-			if (Forms.IsiOS13OrNewer && previousTraitCollection.UserInterfaceStyle != TraitCollection.UserInterfaceStyle)
+			if (Forms.IsiOS13OrNewer &&
+				previousTraitCollection.UserInterfaceStyle != TraitCollection.UserInterfaceStyle &&
+				UIApplication.SharedApplication.ApplicationState != UIApplicationState.Background)
 				Application.Current?.TriggerThemeChanged(new AppThemeChangedEventArgs(Application.Current.RequestedTheme));
 		}
 
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue12512.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue12512.cs
@@ -1,0 +1,70 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System.Diagnostics;
+using System;
+
+#if UITEST
+using Xamarin.Forms.Core.UITests;
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if UITEST
+	[NUnit.Framework.Category(UITestCategories.ManualReview)]
+#endif
+	[Preserve(AllMembers = true)]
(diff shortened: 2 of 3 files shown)


jfversluis (approved):
Awesome! Yeah, this should bring down the theme switching even more. Did you test what happens when you put the app in the background, then switch theme and then go back to the app? Does it then also pick up the change?

From code this looks like it makes sense!

rmarinho (approved):

jsuarezruiz:
@jfversluis Yes, tested that case and is working.

darrencBAS:
There is an issue with this that is stopping me from using anything newer than -pre4, when double clicking home button to see the application manager on ios, if you have changed the theme the app still in background doesn't show the theme change as it hasn't yet been to foreground to re-render. Prior to pre5 it worked perfectly and as expected, all other apps work like this too in that the app manager shows the app with the correct theme applied.

Xamarin.Forms.Platform.iOS/Renderers/PageRenderer.cs
```diff
@@ -364,7 +364,9 @@ public override void TraitCollectionDidChange(UITraitCollection previousTraitCol
 		{
 			base.TraitCollectionDidChange(previousTraitCollection);
 
-			if (Forms.IsiOS13OrNewer && previousTraitCollection.UserInterfaceStyle != TraitCollection.UserInterfaceStyle)
+			if (Forms.IsiOS13OrNewer &&
+				previousTraitCollection.UserInterfaceStyle != TraitCollection.UserInterfaceStyle &&
+				UIApplication.SharedApplication.ApplicationState != UIApplicationState.Background)
```
darrencBAS, line 369: This is stopping ios appliaction manager from showing the theme changes until the app is brought back to foreground by the user. It used to show the theme changes because when it went to background ios calls the theme change twice on the app to take a static screenshot of both light and dark theme ready for use in the application manager. Suggest removing UIApplication.SharedApplication.ApplicationState != UIApplicationState.Background from the if statement, this should rectify the problem.

PR #12983, opened 2020-11-27
[iOS] Update Font Header size

### LINKED ISSUE #11194: NamedSize.Header seems to be wrong size on Android
On Android the NamedSize.Header is way larger than any other named size. Header is 96 and Title is only 24. I think that's is a bug:

https://github.com/xamarin/Xamarin.Forms/blob/5062f533715e629eb572e5ed8a03388ed05a1d47/Xamarin.Forms.Platform.Android/Forms.cs#L705
https://github.com/xamarin/Xamarin.Forms/blob/5062f533715e629eb572e5ed8a03388ed05a1d47/Xamarin.Forms.Platform.Android/Forms.cs#L737

jsuarezruiz:
### Description of Change ###

Update Font Header size in iOS.

### Issues Resolved ### 

- fixes #11194

### API Changes ###

 None

### Platforms Affected ### 

- iOS

None

Not applicable

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Forms.cs
+++ b/Xamarin.Forms.Platform.Android/Forms.cs
@@ -734,7 +734,7 @@ namespace Xamarin.Forms
 						case NamedSize.Caption:
 							return 12;
 						case NamedSize.Header:
-							return 96;
+							return 14;
 						case NamedSize.Subtitle:
 							return 16;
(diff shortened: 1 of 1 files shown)


jfversluis:
Technically a breaking change but since it's unlikely anyone is using this value as it is today, I'll allow it.

PR #13003, opened 2020-11-30
[iOS] Fix NRE using custom font and Italic attribute

### LINKED ISSUE #11720: [Bug] Crash: Value cannot be null. Parameter name: descriptor <-- suddenly started crashing without editing code
I have been developing my app for months, and today out of the blue the app started crashing on the iOS simulator. I had not changed anything, it just started to happen by itself. The error is:

System.ArgumentNullException:  Value cannot be null. Parameter name: descriptor

Stack trace:

```
  at UIKit.UIFont._FromDescriptor (UIKit.UIFontDescriptor descriptor, System.nfloat pointSize) [0x00003] in /Library/Frameworks/Xamarin.iOS.framework/Versions/13.20.2.2/src/Xamarin.iOS/UIKit/UIFont.g.cs:115 
  at UIKit.UIFont.FromDescriptor (UIKit.UIFontDescriptor descriptor, System.nfloat pointSize) [0x00000] in /Library/Frameworks/Xamarin.iOS.framework/Versions/13.20.2.2/src/Xamarin.iOS/UIKit/UIFont.cs:183 
  at Xamarin.Forms.Platform.iOS.FontExtensions._ToNativeFont (System.String family, System.Single size, Xamarin.Forms.FontAttributes attributes) [0x00062] in D:\a\1\s\Xamarin.Forms.Platform.iOS\Extensions\FontExtensions.cs:41
``` 

Happens only on iOS. I had to backtrack almost a month back to find a revision that didn't crash -- even though I have developed this on the same machine. Safe to say is it fails to load some font but that has worked for months.

By binary division of the revisions I was able to find that removing the FontAttributes="Italic" from here fixed it:
```xaml
<Label FontAttributes="Italic" Text="This has started to crash, but has been working until it suddenly stopped" />
```

(The label didn't originally say that, but I simplified it and it still crashes as long as it is using italics).

*But this has been working* - I have seen it running on the screen in the simulator, I have tweaked the layouts, I have also run this on a real device. But out of the sudden I had to remove the italic from a label that I added months back because of the sudden error.

To be clear - All revisions from the last few weeks suddenly started to crash, until I backtracked to that exact line, removed it, and now everything works again.

What could possible cause the simulator to suddenly have trouble with this? I am very very worried right now and lost a lot of time on this -- does anyone have any kind of idea of what could be going on?

Thanks a lot!

jsuarezruiz:
### Description of Change ###

The problem is when using _UIFont.FromDescriptor_ with a null UIFontDescriptor. An internal exception occurs in Xamarin.iOS (we are passing a null UIFontDescriptor parameter). This PR include changes to avoid the Xamarin.iOS internal exception. 
However, the behavior is exactly the same. If cannot set a specific font, use the default font.

### Issues Resolved ### 

- fixes #11720 

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###
Use the NuGet package generated by this PR. Use the code from the Issue11720. Without any exception, the test has passed.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Extensions/FontExtensions.cs
+++ b/Xamarin.Forms.Platform.iOS/Extensions/FontExtensions.cs
@@ -38,7 +38,12 @@ namespace Xamarin.Forms.Platform.iOS
 								traits = traits | UIFontDescriptorSymbolicTraits.Italic;
 
 							descriptor = descriptor.CreateWithTraits(traits);
+
+							if (descriptor != null)
+								throw new Exception();
+
 							result = UIFont.FromDescriptor(descriptor, size);
+
 							if (result != null)
 								return result;
 						}
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue11720.xaml
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue11720.xaml
@@ -1,0 +1,13 @@
+﻿<?xml version="1.0" encoding="UTF-8"?>
+<ContentPage
+    xmlns="http://xamarin.com/schemas/2014/forms"
+    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+    x:Class="Xamarin.Forms.Controls.Issues.Issue11720"
+    Title="Issue 11720">
+    <StackLayout>
+        <Label
+            FontFamily="Font Awesome 5 Free"
+            FontAttributes="Italic"
+            Text="Without crash, the test has passed." />
+    </StackLayout>
+</ContentPage>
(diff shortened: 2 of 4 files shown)


Xamarin.Forms.Platform.iOS/Extensions/FontExtensions.cs
```diff
@@ -38,7 +38,12 @@ static UIFont _ToNativeFont(string family, float size, FontAttributes attributes
 								traits = traits | UIFontDescriptorSymbolicTraits.Italic;
 
 							descriptor = descriptor.CreateWithTraits(traits);
+
+							if (descriptor != null)
+								throw new Exception();
```
jsuarezruiz, line 43: This PR, in case of null `UIFontDescriptor`, throws an exception handled by https://github.com/xamarin/Xamarin.Forms/blob/869868fa7b9483079f57b734e5be0a5a50fe2a36/Xamarin.Forms.Platform.iOS/Extensions/FontExtensions.cs#L72
After that, we use a default font setting the font size and the attributes.

jfversluis:
See comment on issue, I don't think this is an actual runtime crash

PR #13101, opened 2020-12-10
[iOS/macOS] Same offsets behavior in GradientBrush between platforms

### LINKED ISSUE #13092: [Bug] Frame.Background not displaying GradientBrush on iOS
### Description
Set GradientBrush as Frame's background doesn't work on iOS, but working fine on Android.

### Expected Behavior
Frame's background can be rendered as expected on Android.
微信截图_20201210102831

### Actual Behavior
Frame's background can't be rendered correctly on iOS.
微信截图_20201210102923

### Basic Information

- Version with issue: 5.0-pre5
- Platform Target Frameworks: 
  - iOS:  14.2
  - Android: 10

jsuarezruiz:
### Description of Change ###

On Android, using a `GradientDrawable` renders the gradient even without setting the offsets. On iOS (or macOS), the gradient does not render without setting the offsets.

This PR adds changes to set the default values in case of not having any offset set in the Brush. In this way, we have the same behavior between different platforms.

### Issues Resolved ### 

- fixes #13092 

### API Changes ###
 
 None

### Platforms Affected ### 

- iOS
- macOS

None

Captura de pantalla 2020-12-10 a las 14 00 01

### Testing Procedure ###

Launch Core Gallery and navigate to the Brushes samples. Select the new "GradientBrush Offsets Gallery" to verify that the gradient is rendered in the same way on iOS (and macOS) as on Android.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls/GalleryPages/GradientGalleries/GradientBrushDefaultOffsetGallery.xaml
+++ b/Xamarin.Forms.Controls/GalleryPages/GradientGalleries/GradientBrushDefaultOffsetGallery.xaml
@@ -1,0 +1,56 @@
+﻿<?xml version="1.0" encoding="UTF-8"?>
+<ContentPage
+    xmlns="http://xamarin.com/schemas/2014/forms"
+    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+    x:Class="Xamarin.Forms.Controls.GalleryPages.GradientGalleries.GradientBrushDefaultOffsetGallery"
+    Title="GradientBrush default Offset Gallery">
+    <ContentPage.Content>
+        <ScrollView>
+            <StackLayout
+                Padding="12">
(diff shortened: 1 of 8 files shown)


rmarinho (approved):

mattleibow (commented):
Looking at the code, they are the same. Is there no way to move this into a CAUtils or something?

Also, the adjacent `GetRadialGradientBrushEndPoint` methods differ already. macOS has additional checks that might also be needed in the iOS side. Or at least have some sort of extension method for these types that can be shared if possible.

jsuarezruiz:
@mattleibow Thanks for the feedback!.  I have created a shared `BrushExtensions` class between iOS and macOS. Now `GetRadialGradientBrushEndPoint` is the same method.

mattleibow (approved):
Looking good!

PR #13152, opened 2020-12-15
Fix poor choice for FlyoutItem BP Name

PureWeen:
### Description of Change ###

The idea here was to have an attached property called `FlyoutItem.IsVisible` that users could use to indicate when they wanted to hide just the FlyoutItem. This works great in code but in XAML when binding to "IsVisible" on a FlyoutItem that will now gets routed to the Attached property that's shadowing the non attached property. 

### API Changes ###
Added:
 - Shell.FlyoutItemIsVisible //Attached BP
 - BaseShellItem.FlyoutItemIsVisible //Attached BP that allows for easy access to this property on each BaseShellItem

Removed:
 - FlyoutItem.IsVisible //Attached BP

 
### Platforms Affected ### 
- Core/XAML (all platforms)

- Removed the attached property FlyoutItem.IsVisible. Users wanting to just hide something in the Flyout will need to use Shell.FlyoutItemIsVisible.  

None

### Testing Procedure ###
- unit tests modified
- UI Test Included

### PR Checklist ###


StephaneDelcroix (approved):

rmarinho (approved):
Related tests were passing

PR #13186, opened 2020-12-17
[xamlg] improve performance by reusing Mono.Cecil objects

jonathanpeppers:
### Description of Change ###

I found that for each iteration of a `XamlFile`, `<XamlGTask/>` would:

* Iterate over each item in `@(ReferencePath)`.
* Load the assembly with Mono.Cecil.
* Find all the `XmlnsDefinitionAttribute` and create dictionaries.

This would happen for every `.xaml.g.cs` file generated. In
`Xamarin.Forms.Controls.csproj` this resulted in 122 references X 379
`.xaml.g.cs` files.

I moved this logic into a new `XmlnsCache` class. This work can be
done once, and `_xmlnsDefinitions` and `_xmlnsModules` can be reused
for each `.xaml.g.cs` file.

This changes the "lifetime" of `XamlGenerator` a bit, now unit tests
should dispose of `XamlGenerator`, while the `<XamlGTask/>` should
only dispose the `XmlnsCache` after all work is complete.

I also wrote the `XmlnsCache`, so the assemblies are loaded lazily if
the `XmlnsDefinitions` or `XmlnsModules` properties are not accessed.

Other small improvements:

* Don't create the `$(ReferencedAssemblies)` property. This is
  equivalent to doing a `string.Join()` with many full paths.
* We can use `@(ReferencePath)` as-is, to pass into a `string[]`
  property to the `<XamlGTask/>`. This way we don't have to split on
  `;` ourselves.
* The `AssemblyIsSystem()` method should use
  `StringComparison.OrdinalIgnoreCase` as `CurrentCultureIgnoreCase`
  is slower and can produce incorrect results in some cultures.

### Results ###

On Windows, when building `Xamarin.Forms.Controls`:

    Before:
    2643 ms  XamlG                                      1 calls
    2626 ms  XamlGTask                                  1 calls
    After:
     388 ms  XamlG                                      1 calls
     374 ms  XamlGTask                                  1 calls

On macOS, running under Mono:

    Before:
    5003 ms  XamlG                                      1 calls
    4926 ms  XamlGTask                                  1 calls
    After:
     772 ms  XamlG                                      1 calls
     688 ms  XamlGTask                                  1 calls

`Xamarin.Forms.Controls` might resemble a larger customer project?
These changes would help the most for projects with lots of
`.xaml.g.cs` files.

### Issues Resolved ### 

None

### API Changes ###
 
None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###

We may want to do some manual testing in IDEs, I can help with this.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/XmlnsCache.cs
+++ b/Xamarin.Forms.Build.Tasks/XmlnsCache.cs
@@ -1,0 +1,86 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Mono.Cecil;
+using IOPath = System.IO.Path;
+
+namespace Xamarin.Forms.Build.Tasks
+{
+	class XmlnsCache : IDisposable
+	{
+		readonly string[] _references;
+		readonly List<XmlnsDefinitionAttribute> _xmlnsDefinitions = new List<XmlnsDefinitionAttribute>();
+		readonly Dictionary<string, ModuleDefinition> _xmlnsModules = new Dictionary<string, ModuleDefinition>();
+		bool _initialized;
+
+		public List<XmlnsDefinitionAttribute> XmlnsDefinitions
+		{
+			get { Initialize(); return _xmlnsDefinitions; }
+		}
+
+		public Dictionary<string, ModuleDefinition> XmlnsModules
+		{
+			get { Initialize(); return _xmlnsModules; }
+		}
+
+		public XmlnsCache(string[] references)
+		{
+			_references = references;
+		}
+
+		void Initialize()
+		{
+			if (_initialized)
+				return;
+			_initialized = true;
+
+			var paths = _references?.ToList() ?? new List<string>();
+			//Load xmlnsdef from Core and Xaml
+			paths.Add(typeof(Label).Assembly.Location);
+			paths.Add(typeof(Xamarin.Forms.Xaml.Extensions).Assembly.Location);
+
+			foreach (var path in paths.Distinct())
+			{
+				string asmName = IOPath.GetFileName(path);
+				if (AssemblyIsSystem(asmName))
+					// Skip the myriad "System." assemblies and others
+					continue;
+
+				using (var asmDef = AssemblyDefinition.ReadAssembly(path))
+				{
+					foreach (var ca in asmDef.CustomAttributes)
+					{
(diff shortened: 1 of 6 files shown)


jonathanpeppers:
Let me know I can retarget `main` if I need to, thanks.

StephaneDelcroix:
Thanks for investigating this, and proposing this patch. I love it.

It's ok to keep the references alive in a cache, as they are shared across all xaml files of a single assembly. Does this make sure the cache is invalidated when XamlG runs on xaml files in multiples assemblies ?

jonathanpeppers:
> Does this make sure the cache is invalidated when XamlG runs on xaml files in multiples assemblies ?

Every time the `<XamlGTask/>` runs it creates a new `XmlnsCache` that is disposed at the end of `Execute()`. So there should be a fresh instance of `XmlnsCache` for each project/assembly.

jonathanpeppers:
Did this one miss the boat on 5.0? Should it go to another branch?

transis2:
This kind of improvement should be done with
All Net 5+ projects

jsuarezruiz (approved):

Xamarin.Forms.Build.Tasks/XmlnsCache.cs
```diff
@@ -0,0 +1,86 @@
...
+							_xmlnsModules[asmDef.FullName] = asmDef.MainModule;
+						}
+					}
+				}
+			}
+		}
+
+		bool AssemblyIsSystem(string name)
+		{
+			if (name.StartsWith("System.Maui", StringComparison.OrdinalIgnoreCase))
```
jsuarezruiz, line 65: We need to update this to be "Microsoft.Maui"

jsuarezruiz (changes requested):

Xamarin.Forms.Build.Tasks/XmlnsCache.cs
jonathanpeppers: @jsuarezruiz this PR can't apply to MAUI because XamlG is a Roslyn source generator now.

Since Xamarin.Forms 5.0 is basically service mode now, I'm inclined to just close this. What do you think?

PR #9728, opened 2020-02-26
[Android] TalkBack now reads name and helptext on buttons.

ryl:
### Description of Change ###

Modified `AutomationPropertiesProvider` such that `AutomationProperties.Name` and `AutomationProperties.HelpText` will be read by TalkBack.

Re-created from https://github.com/xamarin/Xamarin.Forms/pull/5237

### Issues Resolved ### 

- fixes #5150
- fixes #1529

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

### Behavioral/Visual Changes ###

TalkBack will read `AutomationProperties.Name` and `AutomationProperties.HelpText` on a button if provided.

### Before/After Screenshots ### 

Not applicable

### Testing Procedure ###

In the Control Gallery, visit G5150. Labels containing the ContentDescription for each button should appear with appropriate text.

### PR Checklist ###

- [x] Has automated tests 
- [x] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard

PR #13244, opened 2020-12-28
[Android] TalkBack now reads name and helptext on buttons. 

### LINKED ISSUE #5150: AutomationProperties.Name, AutomationProperties.HelpText on Button not read by Android TalkBack
### Description

`AutomationProperties.Name` and `AutomationProperties.HelpText` are not read by TalkBack on Android when placed on a `Button` element.

### Steps to Reproduce

1. Set`Text` property on a button.
2. Add `AutomationProperties.Name` and `AutomationProperties.HelpText` values to button.
3. Enable TalkBack through Android Settings.
4. Navigate to button, notice neither `Name`, nor `HelpText` are read.

### Expected Behavior

I expect an experience more similar to iOS.

Text is not null, Name is Null, HelpText is not null
"Text. HelpText" is read by TalkBack

Text is not null, Name is Not Null, HelpText is not null
"Name. HelpText" is read by TalkBack 

### Actual Behavior

Name and HelpText are not read when Text is set.

### Basic Information

- Version with issue: 3.4, 3.5
- IDE: Visual Studio 2017 Professional Mac
- Platform Target Frameworks: 
  - Android: Android SDK 9.0
- Nuget Packages: none
- Affected Devices: Any android device it seems.

PureWeen:
### Description of Change ###

Original PR here
https://github.com/xamarin/Xamarin.Forms/pull/9728

- Removed fix on original PR for fixing AutomationId and Content Description. That fix needs more review
- This PR applies a different fix to resolve the wrong name/help text being read. The main problem is that the Renderers for Switch/Button override the SetBasicContentDescription from this PR https://github.com/xamarin/Xamarin.Forms/pull/4094. This was done so that the Hint doesn't get set on the Switch/Button but it needs to use the actual control and not the ViewGroup container
- Fixed an issue where a reused control would get set to the previous elements AutomationId. The defaultContentDescription would only get set when ContentDescription wasn't null. This leads to the defaultContentDescription getting set to the AutomationId because you can't just use a null check to determine if a default should be set for a string. This would make it so when a control gets recycled to a new element it would set the ContentDescription to the previous AutomationId thinking that was the `defaultContentDescription`. The defaultContentDescription/defaultHint are now set only once explicitly the first time the AutomationId/ContentDescription are set. 

### Issues Resolved ### 
- fixes #5150

### Platforms Affected ### 
- Android

### Testing Procedure ###
Switch to Legacy Renderers as the change is only relevant for the Legacy Button Renderer. Switch Renderer doesn't have a fast renderer so it's relevant either way

- Verify UI test 5150
- Verify UI Test 4053 

### PR Checklist ###


rmarinho (changes requested):
Hey @PureWeen  seems it broke the test Bugzilla38112_SwitchIsStillOnScreen_google_pixel_2-8_1_0

rmarinho:
@PureWeen I think it's normal with legacy renderes to talk the container first then the button with content description right? But what is strange is that on 4053 test It didn't talkback the container for the button but on 5150 it does.

PureWeen:
@rmarinho The automation has just applied to the actual control and not the container for some time now. The Container always gets set to {AutomationID}_Container.  If you set the helptext/name to the container then things like Button won't ready correctly.

> But what is strange is that on 4053 test It didn't talkback the container for the button but on 5150 it does.

I don't follow this one. I tested both 4053 and 5150 and they seem to work the same

rmarinho (approved):
Good now if it passes UITEsts

PureWeen:
@vs-mobiletools-engineering-service2 backport 4.8.0

PR #13259, opened 2020-12-31
Add Exception Message when trying to use FlyoutPage with NonAppCompat

PureWeen:
### Description of Change ###

As part of the shift to FlyoutPage from MasterDetailPage we only moved over the AppCompat renderers so if someone is still on non appcompat and tries to use FlyoutPage they will just get a blank screen. This PR will give them a more useful message. 

### Platforms Affected ### 
- Android

### Testing Procedure ###
- Make sure Flyout Page still works with AppCompat (and other platforms)
- Verify Exception Message when switching to non appcompat

### PR Checklist ###

### DIFF
--- a/Stubs/Xamarin.Forms.Platform.cs
+++ b/Stubs/Xamarin.Forms.Platform.cs
@@ -163,7 +163,11 @@ namespace Xamarin.Forms.Platform
 
 
 
-#if !__IOS__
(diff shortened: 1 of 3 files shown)


PureWeen:
Test Failures Unrelated

rmarinho (approved):

PR #13260, opened 2020-12-31
Invoke AnimationFinished from fragment if no animation has been added

PureWeen:
### Description of Change ###

ShellItemRenederer on Android let's users override `SetupAnimation` to define custom animations.  The navigation code on android subscribes to "AnimationFinished" to know when the animation has finished and complete the navigation Task. The problem here is that if the SetupAnimation doesn't actually setup an animation then "AnimationFinished" never gets called from ShellContentFragment.  The real fix here will be to add a better API for the SetupAnimation override. Currently it's just

```C#
SetupAnimation(ShellNavigationSource navSource, AndroidX.Fragment.App.FragmentTransaction t, Page page)
```

But it should really be something like
```C#
SetupAnimation(ShellNavigationSource navSource, AndroidX.Fragment.App.FragmentTransaction t, Page page, Action callMeWhenAnimationHasFinished)
```

Or something similar to indicate to the caller that they need to tell the calling code when the animation has completed.

### Platforms Affected ### 
- Android

### Testing Procedure ###
- UI Test Included

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/ShellContentFragment.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ShellContentFragment.cs
@@ -71,7 +71,7 @@ namespace Xamarin.Forms.Platform.Android
 		public event EventHandler AnimationFinished;
 
 		public Fragment Fragment => this;
-
+				
 		public override AndroidAnimation OnCreateAnimation(int transit, bool enter, int nextAnim)
 		{
 			var result = base.OnCreateAnimation(transit, enter, nextAnim);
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Platform.Android/Renderers/ShellContentFragment.cs
```diff
@@ -82,7 +82,10 @@ public override AndroidAnimation OnCreateAnimation(int transit, bool enter, int
 			}
 
 			if (result == null)
+			{
+				AnimationFinished?.Invoke(this, EventArgs.Empty);
```
PureWeen, line 86: Basically if the user creates a custom renderer that doesn't actually add an animation then the following code

https://github.com/xamarin/Xamarin.Forms/blob/5.0.0/Xamarin.Forms.Platform.Android/Renderers/ShellItemRendererBase.cs#L254

will never execute and the "async push" operation will never complete

This invokes the AnimationFinished event so the calling code can continue on and complete the animation Task

rmarinho (approved):

rmarinho:
Failing tests not related

PR #13320, opened 2021-01-07
[C] Invalidate Grid when row/col def is replaced

### LINKED ISSUE #13154: [Bug] Updating RowDefinitions via C# or Hot Reload doesn't work
### Description

When RowDefinitions is set to a populated RowDefinitionCollection, that change doesn't cause the Grid to be updated.
Updating an existing RowDefinitions value triggers an update, but setting a completely new value doesn't.

### Steps to Reproduce

1. Create XAML like this:

```
             x:Class="FormsScratchPad.MainPage">

    <StackLayout>
        <Button Clicked="Button_Clicked" Text="Update RowDefinitions"/>
        <Grid x:Name="MyGrid" RowDefinitions="20,20,20">
            <Label Grid.Row="1" Text='Hey'/>
            <Label Grid.Row="2" Text='Hey'/>
            <Label Grid.Row="3" Text='Hey'/>
        </Grid>
    </StackLayout>

</ContentPage>
```

with code behind like this:

```
    {
        public MainPage()
        {
            InitializeComponent();
        }

        private void Button_Clicked(object sender, EventArgs e)
        {
            var row1 = new RowDefinition();
            row1.Height = 400;

            var row2 = new RowDefinition();
            row2.Height = 100;

            var row3 = new RowDefinition();
            row3.Height = 200;

            var rows = new RowDefinitionCollection();
            rows.Add(row1);
            rows.Add(row2);
            rows.Add(row3);

            MyGrid.RowDefinitions = rows;
        }
    }
```

2. Run the app and hit the Update RowDefinitions button

### Expected Behavior

The Grid should update, with the new row heights.

### Actual Behavior

Nothing changes visually, at least not at first.
Note that if the code is changed to set an empty collection first and then update the collection, like below, it works. But it doesn't work when the property is set directly with the fully populated collection.
```

            var rows = new RowDefinitionCollection();
            MyGrid.RowDefinitions = rows;
            rows.Add(row1);
            rows.Add(row2);
            rows.Add(row3);

```

This is a problem for Hot Reload, when updating RowDefinitions using the string syntax (`RowDefinitions="400,100,200"`). In that case Hot Reload invokes the type converter, getting a fully populated RowDefinitionCollection, then sets the RowDefinitions property & the grid doesn't update.

### Basic Information

- Version with issue: 5.0 pre5
- Last known good version: none known
- Platform Target Frameworks: 
  - Android:  VS 16.9
- Android Support Library / AndroidX Version:

StephaneDelcroix:
### Description of Change ###

We invalidate when the items of the row/col def are modified. With this,
we also invalidate when the full collection is swapped.

### Issues Resolved ### 

- fixes #13154

### API Changes ###

 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Grid.cs
+++ b/Xamarin.Forms.Core/Grid.cs
@@ -30,6 +30,7 @@ namespace Xamarin.Forms
 					((ColumnDefinitionCollection)oldvalue).ItemSizeChanged -= ((Grid)bindable).OnDefinitionChanged;
 				if (newvalue != null)
 					((ColumnDefinitionCollection)newvalue).ItemSizeChanged += ((Grid)bindable).OnDefinitionChanged;
+				((Grid)bindable).OnDefinitionChanged(bindable, EventArgs.Empty);
 			}, defaultValueCreator: bindable =>
 			{
 				var colDef = new ColumnDefinitionCollection();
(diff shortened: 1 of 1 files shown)


rmarinho:
Can we add a unit test @StephaneDelcroix ?

jsuarezruiz (approved):

rmarinho (approved):

PR #13367, opened 2021-01-11
Use internal tag android

Redth:
### Description of Change ###

1. Added an Android resource identifier `maui_internal_tag` for use with `GetTag(int, Object)` and `SetTag(int, Object)` where `int` is required to be an android resource identifier.

2. Updated all usages of `.Tag` on Android to instead use `SetTag` and `GetTag` with the `maui_internal_tag` resource id so that `Tag` is free for other uses.

### Platforms Affected ### 
- Android

None

Not applicable

### PR Checklist ###


jfversluis:
@Redth closing this one, I see some .NET MAUI references, I assume we don't need this here anymore? Let me know if you think otherwise

PR #13370, opened 2021-01-12
Invalidate CollectionView Layout on transition to IsVisible = true

### LINKED ISSUE #13203: [Bug] [iOS] CollectionView does not bind to items if `IsVisible=False`
### Description

Bind CollectionView to an item source while not visible (i.e. `IsVisible="False"`), then change visibility to true, no items in the list.

### Steps to Reproduce

```csharp
using Xamarin.Forms;

namespace CollectionViewBrokenAgain
{
    {
        public MainPage()
        {
            InitializeComponent();
            BindingContext = this;
        }

        public List<int> Items { get; } = new List<int>() { 1, 2, 3, 4, 5, 6, 7, 8, 9 };

        {
            base.OnAppearing();
           // CollectionView is bound to Items, but will be empty if visibility was initially set to false
            ItemsView.IsVisible = true;
        }
    }
}
```

```xml
<ContentPage
    x:Class="CollectionViewBrokenAgain.MainPage"
    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml">

    <CollectionView
        x:Name="ItemsView"
        IsVisible="False"
        ItemsSource="{Binding Items}">
        <CollectionView.ItemTemplate>
            <DataTemplate>
                <Frame>
                    <Label Text="{Binding}" />
                </Frame>
            </DataTemplate>
    </CollectionView>

</ContentPage>
```

### Expected Behavior

Should should items after visibilty is changed to `true`. 

### Actual Behavior

No items displayed.

### Basic Information

- Version with issue: 4.8.0.1821
- Last known good version: 4.8.0.1687
- Platform Target Frameworks: 
  - iOS:  14.2

### Workaround

Can wrap CollectionView in a ContentView, and toggle the visibility on that instead.

### Repro link
Repro: [Issue13203.zip](https://github.com/xamarin/Xamarin.Forms/files/5726025/Issue13203.zip)

hartez:
### Description of Change ###

If the CollectionView is the root Content element for a page on iOS and IsVisible transitions from false to true, the CollectionView's layout is not invalidating. This change invalidates the layout so that the CollectionView is refreshed.

Note that this isn't an issue if the CollectionView is inside some other view (e.g., a StackLayout); it's only an issue if it's the root of the page Content.

### Issues Resolved ### 

- fixes #13203

### API Changes ###

None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###

Automated UI test

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue13203.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue13203.cs
@@ -1,0 +1,76 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Text;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+
+#if UITEST
+using Xamarin.UITest;
+using NUnit.Framework;
+using Xamarin.Forms.Core.UITests;
+#endif
+
+
+namespace Xamarin.Forms.Controls.Issues
+{
+	[Issue(IssueTracker.Github, 13203, "[Bug] [iOS] CollectionView does not bind to items if `IsVisible=False`", PlatformAffected.iOS)]
+#if UITEST
+	[NUnit.Framework.Category(UITestCategories.CollectionView)]
+#endif
+	public class Issue13203 : TestContentPage
+	{
+		const string RunTest = "RunTest";
+		const string Success = "Success";
+
+		protected override void Init()
+		{
+			var cv = new CollectionView
+			{
+				IsVisible = false,
+
+				ItemTemplate = new DataTemplate(() =>
+				{
+					var label = new Label();
+					label.SetBinding(Label.TextProperty, new Binding(nameof(Item.Text)));
+					return label;
+				})
+			};
+
+			var source = new List<Item> { new Item { Text = Success } };
+			cv.ItemsSource = source;
+
+			var button = new Button { Text = "Set IsVisible to True", AutomationId = RunTest };
+
+			var layout = new StackLayout
+			{
+				Children = { button, cv }
+			};
+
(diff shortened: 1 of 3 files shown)


rmarinho:
Are you missing a button with "RunTest" @hartez  ?

hartez:
> Are you missing a button with "RunTest" @hartez ?

Yep, I forgot I changed the test to use Appearing. :(

AlleSchonWeg:
> Note that this isn't an issue if the CollectionView is inside some other view (e.g., a StackLayout); it's only an issue if it's the root of the page Content.

I have the same issue, but my CollectionView is inside some other view.

rmarinho (approved):

hartez:
> > Note that this isn't an issue if the CollectionView is inside some other view (e.g., a StackLayout); it's only an issue if it's the root of the page Content.
> 
> I have the same issue, but my CollectionView is inside some other view.

@AlleSchonWeg I should clarify - it's not an issue if the container view is one that will be invalidated if one of its children changes its size or visibility. StackLayout, Grid, FlexLayout, etc. all recompute their layouts if a child control's visibility changes. There may be other containers which do not. I just included that note for folks who need an immediate workaround; wrapping the CollectionView in a StackLayout or Grid will fix the issue temporarily while waiting for the next release.

It's all moot after this change is merged, because the CollectionView itself will invalidate its own layout after a visibility change. So it shouldn't matter what type the container is.

hartez:
> Does this need to be unregistered? Wouldn't want a new memory leak.

It'd be unusual, but I suppose a situation where the user is holding on to ItemsView and reusing it in another renderer could leak the controller. Added the unsubscribe during Dispose.

rachelkang (commented):
Are these failing tests relevant?

image

PureWeen:
@rachelkang those failures are not relevant

rachelkang (approved):

PR #13383, opened 2021-01-13
Fix parsing of azure devops unit test filters

PureWeen:
### Description of Change ###

When running UI Tests against iOS and Android via cake the nunit where isn't parsing correctly. This fixes the issue

### Testing Procedure ###
- make sure iOS 14 tests and UWP tests run and are filtered correctly 

### PR Checklist ###

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -978,6 +978,25 @@ Task("_cg-ios-run-tests")
                 }
             };
 
+        if(isCIBuild)
+        {
+            Information("defaults write com.apple.CrashReporter DialogType none");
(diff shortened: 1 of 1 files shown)


build.cake
```diff
@@ -1074,6 +1047,45 @@ Task("Default")
 
 RunTarget(target);
 
+
+
+void SetNunitEnvironmentVariables()
```
rmarinho, line 1052: Do we need this?

PureWeen: No... I thought I deleted that :-/ 

pushed a fix

PR #13384, opened 2021-01-13
[Android] Changes in CarouselView initial position 

jsuarezruiz:
### Description of Change ###

Recently, fixing a TabView issue https://github.com/xamarin/XamarinCommunityToolkit/issues/630 (which internally uses CarouselView) in the Xamarin Community Toolkit, I noticed some unexpected behavior with the initial position of the CarouselView.

The problem is that the casting:

`var items = Carousel.ItemsSource as IList;`

It is not correct in all cases. This PR adds changes to use an enumerator.

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

None

Not applicable

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/CollectionView/CarouselViewRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/CollectionView/CarouselViewRenderer.cs
@@ -335,18 +335,18 @@ namespace Xamarin.Forms.Platform.Android
 		void UpdateInitialPosition()
 		{
 			int position = 0;
-			var items = Carousel.ItemsSource as IList;
-			var itemCount = items?.Count ?? 0;
+			int itemCount = 0;
 
-			if (Carousel.CurrentItem != null || items == null)
+			if (Carousel.CurrentItem != null)
 			{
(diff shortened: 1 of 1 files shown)


Xamarin.Forms.Platform.Android/CollectionView/CarouselViewRenderer.cs
```diff
@@ -335,18 +335,18 @@ void UpdateItemDecoration()
...
+
+				while (carouselEnumerator.MoveNext())
 				{
-					if (items[n] == Carousel.CurrentItem)
-					{
-						position = n;
-						break;
-					}
+					if(carouselEnumerator.Current == Carousel.CurrentItem)
+						position = itemCount;
```
rmarinho, line 347: Shouldn't this break so it doesn't loop the full collection ?

jsuarezruiz: We need `itemCount` right after this iteration, so we need it to complete the loop.

rmarinho (approved):

PR #13423, opened 2021-01-16
Don't set IsAccessibilityElement for everything

### LINKED ISSUE #12252: [Bug][Accessibility] iOS CollectionView with VoiceOver not usable in Xamarin.Forms 4.8
### Description
In Xamarin.Forms 4.8, when using VoiceOver, the contents of a CollectionView do not get accessible and thus, items cannot be selected

### Steps to Reproduce
see repro project
1. add a CollectionView with a few items and SelectionMode="Single"
2. deploy to an iOS device
3. turn VoiceOver on
4. try navigating through the CollectionView

### Expected Behavior
The CollectionView is accessible by VoiceOver and items can be selected

### Actual Behavior
The CollectionView cannot be accessed by voiceOver. The user is unable to navigate through the items or select them

### Basic Information

- Version with issue: 4.8.0.1451
- Last known good version: 4.7.0.1080
- Platform Target Frameworks: 
  - iOS:  13.7
- Affected Devices: iOS devices

### Reproduction Link
[collectionview.zip](https://github.com/xamarin/Xamarin.Forms/files/5282037/collectionview.zip)

PureWeen:
### Description of Change ###

The original comment says this
```
		// UIKit UIViews created via storyboard default IsAccessibilityElement to true, BUT	
			// UIViews created programmatically default IsAccessibilityElement to false.	
			// We need to default to true to allow all elements to be accessible by default and	
			// allow users to override this later via AutomationProperties.IsInAccessibleTree
```

I don't feel like that's completely accurate. If you inspect IsAccessibilityElement directly after instantiating a control then it is indeed false but by the time the control is rendering it is set to the correct value for the given control. If you breakpoint SetAccessibilityElement inside the Automation Extensions you will see that IsAccessibilityElement is true when it should be true and false when it should be false

There are certain controls (UIScrollView, ListView, really anything that's a container of views) that should always remain false. If they get set to true then it causes all the content inside the view to become unreachable. 

Just let the OS do what the OS is going to do

### Issues Resolved ### 
- fixes #12252

### Platforms Affected ### 
- iOS

### Testing Procedure ###
- Test any of the collection view samples using voice over and make sure they still work
- Test all samples that are voice over related and make sure they still work

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Extensions/AccessibilityExtensions.cs
+++ b/Xamarin.Forms.Platform.iOS/Extensions/AccessibilityExtensions.cs
@@ -88,9 +88,24 @@ namespace Xamarin.Forms.Platform.MacOS
 			if (Element == null || Control == null)
 				return _defaultIsAccessibilityElement;
 
+			// If the user hasn't set IsInAccessibleTree then just don't do anything
+			if (!Element.IsSet(AutomationProperties.IsInAccessibleTreeProperty))
+				return null;
+
 #if __MOBILE__
 			if (!_defaultIsAccessibilityElement.HasValue)
-				_defaultIsAccessibilityElement = Control.IsAccessibilityElement;
+			{
+				// iOS sets the default value for IsAccessibilityElement late in the layout cycle
+				// But if we set it to false ourselves then that causes it to act like it's false
+
+				// from the docs:
+				// https://developer.apple.com/documentation/objectivec/nsobject/1615141-isaccessibilityelement
+				// The default value for this property is false unless the receiver is a standard UIKit control,
+				// in which case the value is true.
+				//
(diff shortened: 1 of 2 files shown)


jsuarezruiz (approved):

rachelkang (commented):
Some previously accessible things are no longer accessible

i.e. Open iOS ControlGallery >> CollectionView Gallery
The various buttons linking to different CV sample galleries are no longer accessible. Where they were previously focusable with VO, they no longer are. Will these issues be resolved elsewhere? If not, can we move this bit of code to several other areas where they may still be applicable?

rachelkang (approved):

PureWeen:
Failing tests unrelated

PR #13462, opened 2021-01-19
Setup Shell to only use the LogicalChildren for reporting its LogicalChildren

### LINKED ISSUE #13171: [Bug] Shell Flyout Header/Footer/more aren't included in visual tree - missing mutations
### Description

With shell visual tree mutations aren't sent for header/footer and seem to be incorrect (unparented) for FlyoutItem, causing Hot Reload to not work and the Live Visual Tree to be incomplete.

### Steps to Reproduce

1. Do File / New Project / Mobile App / Flyout
2. In `AppShell.xaml` add the following

```
    <Shell.FlyoutHeader>
        <Label Text="Hello there"/>
    </Shell.FlyoutHeader>
```
3. Debug the app, say on Android (though likely all platforms have the issue)
4. Tap on the top left hamburger menu to popup the Flyout menu
5. Update the "Hello there" text to something else.

### Expected Behavior

Text updates on device. Live Visual Tree includes all controls shown in the UI.

### Actual Behavior

There is no update.

The LVT as below. Note that the header isn't shown.
Here we never get mutations for header / footer, so they aren't included in the visual tree.
image

Also note that the Label and other visual elements that make up a FlyoutItem aren't included in the LVT. ShellContent is there, but nothing for the menu item.

In that case, the issue looks to be a little different - we are getting mutation events for those controls, but they have no parent.
See the "Xamarin.Forms: XAML Hot Reload - core team weekly" Teams chat for more details on that - it's private but Stephane is a member and can see the screenshot from Tim showing this, etc.

### Basic Information

- Version with issue: 5.0 pre5
- Affected Devices: all

PureWeen:
### Description of Change ###
Shift Shell over so it returns everything from the Logical Children structure which plays into Hot Reload and the Live Visual Tree better. Initially I didn't use LogicalChildren because Page (the base class of Shell) makes a bunch of assumptions against its LogicalChildren. This PR intercepts those assumptions and just lets Shell manage its logical children 

### Issues Resolved ### 

- fixes #13171 

### Testing Procedure ###
- Validate that Live Visual Tree displays flyout items/flyoutheader/flyoutfooter/flyoutcontent

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Core/Page.cs
+++ b/Xamarin.Forms.Core/Page.cs
@@ -379,7 +379,7 @@ namespace Xamarin.Forms
 				var element = LogicalChildren[i];
 				if (element is VisualElement c)
 				{
-					if (c.Bounds != startingLayout[i])
+					if (startingLayout.Count <= i || c.Bounds != startingLayout[i])
 					{
 						LayoutChanged?.Invoke(this, EventArgs.Empty);
 						return;
--- a/Xamarin.Forms.Core.UnitTests/ShellTests.cs
+++ b/Xamarin.Forms.Core.UnitTests/ShellTests.cs
@@ -768,10 +768,10 @@ namespace Xamarin.Forms.Core.UnitTests
 			shell.FlyoutHeader = null;
 			shell.FlyoutHeader = layout;
 
-			Assert.True(shell.ChildrenNotDrawnByThisElement.Contains(layout));
+			Assert.True(shell.LogicalChildren.Contains(layout));
 			shell.FlyoutHeader = null;
 
-			Assert.False(shell.ChildrenNotDrawnByThisElement.Contains(layout));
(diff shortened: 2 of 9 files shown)


Xamarin.Forms.Core/Page.cs
```diff
@@ -379,7 +379,7 @@ protected void UpdateChildrenLayout()
 				var element = LogicalChildren[i];
 				if (element is VisualElement c)
 				{
-					if (c.Bounds != startingLayout[i])
+					if (startingLayout.Count <= i || c.Bounds != startingLayout[i])
```
PureWeen, line 382: Now that generated flyout items are added to the LogicalChildren the `startingLayout.Count` no longer matches up with `LogicalChildren.Count` after LayoutChildren is called

BretJohnson (approved):

drasticactions (approved):

PR #13484, opened 2021-01-21
[X] Accept RD.Source with assembly

### LINKED ISSUE #2660: Cross Assembly CSS Portability
### Description

There are a number of times, particularly with a modular app design where a CSS file may not necessarily reside within the same assembly as a Page/View consuming that CSS resource. There needs to be a decent way from XAML where we can load this resource. 

cc: @jassmith @kzu 

### Expected Behavior

There are a few ways that this could probably be done:

#### Location Handler

Having something like a CSS Resource Location Handler that could be exported at the assembly level could be good to provide a global way of dealing with where to look for css files being referenced within a particular assembly or perhaps at an application level.

#### Make Source an Object

Changing source from a Uri to an object and determine if it's a Uri, StyleSheet, or a TextReader would offer a lot of flexibility as you could write custom Markup Extensions that could simply do the work of locating the resource and then either creating the StyleSheet or providing the TextReader to create the StyleSheet.

```cs
switch(Source)
{
    case Uri uri:
        return DoCurrent(uri);
    case StyleSheet stylesheet:
        return stylesheet;
    case TextReader reader:
        return StyleSheet.FromReader(reader);
    default:
        return throw new NotSupportedException();
}
```

#### Assembly Property

Adding an Assembly Property to the StyleSheetExtension would provide a nice clean and more verbose way of providing context for the StyleSheetExtension to locate the resource.

```xml
<StyleSheet Source="theme/style.css" Assembly="MyAwesomeLibrary" />
```

### Actual Behavior

There is nothing that works from XAML OOB AFAIK

### Basic Information

- Version with issue: 3.X
- IDE: n/a

### LINKED ISSUE #13187: [Bug] ResourceDictionary.Source does not include assembly name
### Steps to Reproduce

1.  Open attached solution in VS 2019
[XamarinApp9.zip](https://github.com/xamarin/Xamarin.Forms/files/5712375/XamarinApp9.zip).
2. Set Debug x86.
3. Build and Run.
4. Click on Click Me button

### Actual Behavior
Appears text "ResourceDictionary.Source is 'CustomRD.xaml'"

### Expected Behavior
ResourceDictionary.Source is 'fx://XamarinApp9/CustomRD.xaml' or similar. With assembly name.

StephaneDelcroix:
### Description of Change ###

[X] Accept RD.Source with assembly
this also always add the assembly part in the Source url

### Issues Resolved ### 

- fixes #13187
- fixes #2660

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Build.Tasks/CompiledConverters/RDSourceTypeConverter.cs
+++ b/Xamarin.Forms.Build.Tasks/CompiledConverters/RDSourceTypeConverter.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
 using Xamarin.Forms.Build.Tasks;
@@ -13,7 +14,7 @@ namespace Xamarin.Forms.Core.XamlC
 	{
 		public IEnumerable<Instruction> ConvertFromString(string value, ILContext context, BaseNode node)
 		{
-			var module = context.Body.Method.Module;
+			var currentModule = context.Body.Method.Module;
 			var body = context.Body;
 
 			INode rootNode = node;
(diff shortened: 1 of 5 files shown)


Xamarin.Forms.Build.Tasks/CompiledConverters/RDSourceTypeConverter.cs
```diff
@@ -22,7 +23,23 @@ public IEnumerable<Instruction> ConvertFromString(string value, ILContext contex
...
+			var module = currentModule;
+			string asmName = null;
+			if (value.Contains(";assembly="))
+			{
+				var parts = value.Split(new[] { ";assembly=" }, StringSplitOptions.RemoveEmptyEntries);
+				value = parts[0];
+				asmName = parts[1];
+				if (currentModule.Assembly.Name.Name != asmName)
+				{
+					var ar = currentModule.AssemblyReferences.FirstOrDefault(ar => ar.Name == asmName);
```
rmarinho, line 37: Should we check if the assembly exists ? And throw a more meaningful  error if it doesn't  , let's say i set something like:

```
SharedResourceDictionary.xaml;assembly=FooWithANNamingError",
````

StephaneDelcroix: you get ResourceMissing. We could have a specialised message, but not in a -sr (as it involves translation)

rmarinho (approved):

PureWeen (approved):

Xamarin.Forms.Build.Tasks/CompiledConverters/RDSourceTypeConverter.cs
Tommigun1980: I tried to add a comment here but must have missed to submit it.

Wouldn't it be safer to split the string on '=' and parse the left and right parts independently? The code right now assumes that 'assembly=' is the last parameter. It will also crash if 'assembly=' is used, which to the best of my knowledge is valid and should just be ignored.

PR #13528, opened 2021-01-25
#13527 - OnAppearing() is now called on every back navigation again b…

### LINKED ISSUE #13527: [Bug] When navigating back to a previous page (modal stack), OnAppearing() is not called
### Description
When you have multiple pages in a modal stack and you navigate back, the `OnAppearing()` callback of the previous page that you return to will only be called for the most inner level.
E.g.: You have 3 modal pages: 1, 2 and 3. You start with page 1, then you navigate to 2 and from 2 to 3 (modally) => your stack will be 1/2/3;
Now, navigating back from 3 to 2 will call `2.OnAppearing()`, but navigating back from 2 to 1 **will not call** `1.OnAppearing()` wich breaks a lot of things!

This is because when you navigate to the next modal leve, UAP `Platform.cs` line 348 overwrites any `_modalBackgroundPage` even if there is already one!

### Steps to Reproduce

1. Create a sample that has a navigation page, pushes a nother navigation page as modal and then a 2nd one
2. Navigating back one step to the 1st modal navigation page still triggers its "OnAppearing" callback
3. However, navigating back to the root does not trigger "OnAppearing" anymore

### Expected Behavior
Whenever you navigate back to the previous page, its "OnAppearing" callback should be called

### Actual Behavior
It works if you only have one level of nesting. However, as soon as you have 2 modal pages, the outer ones "OnAppearing" is not called anymore

### Basic Information

- Version with issue: 4.8.x and 5.0.0
- Last known good version: Anyhing before commit d103974c which broke everything (issue that was resoled: #8551)
- Platform Target Frameworks: 
  - UWP:  any version. This is a bug in UAP `Platform.cs`
- Android Support Library / AndroidX Version: 

- Affected Devices: all UWP devices

@jfversluis  this issue occurs because of your fix for tranparent background of modals

gentledepp:
…y not storing the previous page in a variable, but managing all previous pages in a Stack<Page>

### Issues Resolved ### 

- fixes #13527: When you have multiple pages in a modal stack and you navigate back, the `OnAppearing()` callback of the previous page that you return to will only be called for the most inner level.
E.g.: You have 3 modal pages: 1/2/3
navigating back from 3 to 2 will call 2.OnAppearing(), but navigating back from 2 to 1 will **not** call 1.OnAppearing() wich breaks a lot of things!

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/Platform.cs
+++ b/Xamarin.Forms.Platform.UAP/Platform.cs
@@ -275,7 +275,7 @@ namespace Xamarin.Forms.Platform.UWP
 		readonly Windows.UI.Xaml.Controls.Page _page;
 		Windows.UI.Xaml.Controls.ProgressBar _busyIndicator;
 		Page _currentPage;
-		Page _modalBackgroundPage;
+		readonly Stack<Page> _modalBackgroundPages = new Stack<Page>();
 		readonly NavigationModel _navModel = new NavigationModel();
 		readonly ToolbarTracker _toolbarTracker = new ToolbarTracker();
 		readonly ImageConverter _imageConverter = new ImageConverter();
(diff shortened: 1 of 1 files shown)


gentledepp:
@jfversluis  I just created the fix for the issue. Please comment

Xamarin.Forms.Platform.UAP/Platform.cs
```diff
@@ -345,19 +345,23 @@ async void SetCurrent(Page newPage, bool popping = false, bool modal = false, Ac
 					Page previousPage = _currentPage;
 
 					if (modal && !popping && !newPage.BackgroundColor.IsDefault)
-						_modalBackgroundPage = previousPage;
```
gentledepp, line 348: Because of this line, when you navigate to the second modal level, `_modalBackgroundPage` already has a value and will be overwritten.
As of that, when you navigate back the second time, `_modalBackgroundPage` will already be `null` (this is done in the first pop of the navigation stack at line 360 below!). So you lost the outer modal page and cannot call its `OnAppeare()` callback anymore!

jfversluis:
Alright, this seems good to me. Sorry we left this around for so long! Congrats on becoming a Xamarin.Forms contributor @gentledepp and many thanks for your time and effort on this!

bitshftr:
Any chance this "fix" has now broken navigation when using Shell.Current.GoToAsync(...) ?  Whenever I call Shell.Current.GoToAsync(...) now, I get the new window for a second, then navigation returns to the primary window without further interaction.

PR #13567, opened 2021-01-28
Move core tests to xUnit

hartez:
Move core tests from NUnit -> xUnit for consistency.

Also added a couple of custom xUnit attributes to make the test output nicer and make categories easier:

FactAttribute - defaults the display name of the test to the method name, rather than the full namespace + method.
TheoryAttribute - same thing, but for Theory
CategoryAttribute - Convenience attribute for setting a Trait with the name "Category" on test classes and methods. Handles multiple categories at once.


mattleibow (approved):
Looking good!

PureWeen (approved):

PR #13666, opened 2021-02-05
Fix material colors to match previous version better

### LINKED ISSUE #13416: Material-UI bug on several ui controls
### Description
We found some ui-issues using Xamarin.Forms 5.0.0.1874 and Xamarin.Forms.Visual.Material 5.0.0.1874.
After updating the nuget packages the ui was broken because buttons looked like they were disabled even they were not.

The backgroundcolor of DatePickers, TimePickers, Entries, Editors and Pickers is changing to white (or black if you use the dark theme) if they are disabled and you use material design.
You're not able to change the backgroundcolor even if you set it by AppThemeBinding light and dark.
If you enable these controls all looks like they should be.

Buttons that use material design and an image inside they are looking like disabled buttons when you open a new page. You're still able to push them.
If you remove these images everything looks fine.

light1
light2
dark1
dark2

I created a new project to reproduce it. And found some more controls that have these issues.

[https://github.com/KonstantinKellermann/MaterialTestProject](url)

The last good version was 4.8.0.1821.

### Steps to Reproduce

1. Take a TimePicker, DatePicker, Entry, Editor or a Picker
2. set Visual="Material"
3. choose a backgroundcolor (not white or black so you can see the effect)
4. set IsEnabled="False"
5. compare the settings IsEnabled="False" and IsEnabled="True" you will see the difference

or

1. Take a Button
2. set Visual="Material"
3. choose a backgroundcolor (not white or black so you can see the effect)
4. set the ImageSource
5. compare it to a button with the same settings only with no image in it

### Expected Behavior
The controls should use the given backgroundcolor or it should be possible to change the DisabledBackgroundColor, that would be nice to have, too.

### Basic Information

- Version with issue: 
  - Xamarin.Forms 5.0.0.1874
  - Xamarin.Forms.Visual.Material 5.0.0.1874
- Last known good version: 
  - Xamarin.Forms 4.8.0.1821
  - Xamarin.Forms.Visual.Material 4.8.0.1821
- Affected Devices:
  - Pixel 3a XL - Andorid 10
  - Pixel 3a XL - Andorid 11
  - Samsung Galaxy S9 - Android 10

PureWeen:
### Description of Change ###

- fixed android so the disabled colors are correct
- fixed iOS so if you set the placeholder color and the control is disabled it'll set the color correctly
- fixed the underline on android so when not focused it styles correctly relative to the place holder
- fixed the underline on android to shade correctly when not focused
- Fixed it so if you have a button with an image that gets pushed onto the stack the background color will render correctly

### Issues Resolved ### 

- fixes #13416 

### Platforms Affected ### 
- Android

- on iOS when the control is disabled the PH color now sets correctly

### Testing Procedure ###
- Visual Gallery Included tests so you can compare Colors

### PR Checklist ###


jsuarezruiz:
@PureWeen Should we mark this PR with the `blocker` flag?

jsuarezruiz (approved):
fix13666

jsuarezruiz (changes requested):
When the Entry is disabled in iOS, the bottom line is dashed, is an expected behavior?

PureWeen:
@jsuarezruiz looking at 4.8 that hasn't changed so we can fix that with a different PR if we need to

image

Redth (approved):
Other than the underline differences @jsuarezruiz noticed, lgtm!

PR #13673, opened 2021-02-05
[iOS, Android] Change the SwipeView Content BackgroundColor Behavior

### LINKED ISSUE #13670: [Bug] [XF 5] SwipeView not rendering properly on iOS and Android when swiping
### Description

Starting from XF 5.0 SwipeView controls begun not to render properly during the swiping gesture: e.g. the whole SwipeView.LeftItems are shown since the beginning of the gesture in transparency. See attached sample project.

(Possible duplicate of https://github.com/xamarin/Xamarin.Forms/issues/13496 or https://github.com/xamarin/Xamarin.Forms/issues/13495 ?)

### Expected Behavior

The SwipeView.LeftItems should be revealed by the SwipeView content while dragging, gradually. Like in XF 4.8:
XF4_8

### Actual Behavior
The content of the SwipeView.LeftItems is revealed at once at the beginning of the dragging gesture.
xf5

### Basic Information

- Version with issue: 5.0.0.1931
- Last known good version: 4.8
- Platform Target Frameworks:
  - iOS
  - Android

### Sample Project
[SwipeViewBug2.zip](https://github.com/xamarin/Xamarin.Forms/files/5931831/SwipeViewBug2.zip)

jsuarezruiz:
### Description of Change ###

Change the SwipeView Content BackgroundColor Behavior.

_What should be the default behavior of the background color of the SwipeView content?_

Previously (Xamarin.Forms 4.8), if a background color was not assigned to the content of the SwipeView, the background color of the SwipeView was set to the content.

Pros
- Most devs assign the background color to the SwipeView, not the content.

Cons
- If you want to use the default color (transparent), you must assign it explicitly.

The current behavior assigns the background color of the SwipeView in the renderer, but does nothing with the Content. Content must be explicitly assigned.

Pros
- Allows content with transparent color.

Cons
- Without assigning the background color, when swiping the effect is weird because the content is transparent. 

This PR apply changes to use the behavior from 4.8. If we agree that must be the default behavior we can merge this PR, in other case close it and maybe update the docs.

What do you think about it? cc @PureWeen @davidbritch 

### Issues Resolved ### 

- fixes #13670 

### API Changes ###

 None

### Platforms Affected ### 

- iOS
- Android

None

fix13670

### Testing Procedure ###
Launch Core Gallery and navigate to the issue 13670. Without setting the BackgroundColor of the SwipeView Content, the BackgroundColor of the SwipeView is used.

### PR Checklist ###


davidbritch:
I think the behaviour from 4.8 is correct. This seems like it should be the default behaviour, which users can opt out of if they want to. Whereas in 5.0, by default, you have weird behaviour if you don't set the BackgroundColor. Sure, the docs can point that out, but ideally (IMHO) controls should be easy enough to use that it's not necessary to read the doc.

jsuarezruiz:
Thanks for the feedback @davidbritch!

rachelkang (commented):
@jsuarezruiz On Android, when swiping left to collapse the extra content, the swipe animation is not as smooth as on iOS and is a bit glitchy. Is this behavior related or a potentially separate issue?

Other than this, all tests are passing, fix works great against the original issue, and LGTM on both iOS and Android!

PR #13681, opened 2021-02-06
Fix FlyoutItem recycling

PureWeen:
### Description of Change ###

Currently on iOS and Android when you scroll the Flyout the Flyout Items will clear and recreate renderers. They will also clear and recreate renderers if you remove items or rearrange. This causes a lot of churn. This PR disables cell recycling on iOS and Android since we're not dealing with massive lists. If users want full virtualization for the FlyoutItems they should use a custom FlyoutContent with a CollectionView. These changes also keep track of FlyoutItems so when the list is modified that cells/renderers are reused instead of recreating everything. 

These changes are optimized around a small to medium sized list

### Platforms Affected ### 
- iOS
- Android

### Testing Procedure ###
- make sure the store shell flyout scrolls nicely
- Test a few of the UI tests that involve flyout visibility to make sure items show and vanish

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/ShellFlyoutRecyclerAdapter.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/ShellFlyoutRecyclerAdapter.cs
@@ -18,10 +18,10 @@ namespace Xamarin.Forms.Platform.Android
 		List<List<Element>> _flyoutGroupings;
 		Action<Element> _selectedCallback;
 		bool _disposed;
-		ElementViewHolder _elementViewHolder;
 
 		public ShellFlyoutRecyclerAdapter(IShellContext shellContext, Action<Element> selectedCallback)
 		{
+			HasStableIds = true;
 			_shellContext = shellContext;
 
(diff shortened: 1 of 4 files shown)


PureWeen:
Failing Tests not related

rmarinho (approved):

rmarinho:
@PureWeen  works great, the first scroll is slow (it was also before), but then after that it butter smooth..

Xamarin.Forms.Platform.Android/Renderers/ShellFlyoutRecyclerAdapter.cs
```diff
@@ -43,7 +42,22 @@ public ShellFlyoutRecyclerAdapter(IShellContext shellContext, Action<Element> se
 
 		public override int GetItemViewType(int position)
 		{
-			var item = _listItems[position];
+			return _listItems[position].Index;
```
Redth, line 45: This is probably fine, but I'm curious why this instead of just returning the position passed in which will also always be unique?

Redth (approved):

PR #13762, opened 2021-02-15
Start Running Device Tests

mattleibow:
### Description of Change ###

This PR runs the device tests as it stands today. Just a simple run with a clean Debug build and then run on device. In future we need to take the build artifacts from the main build and test those. This will give us the bits to test XamlC and not and anything else.

But for now, this is a good smoke test and we have already uncovered 3 failing handlers tests and a case where we are using API 23 things and not falling back nicely on API 21

### Issues Resolved ### 

- fixes #

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###


eng/pipelines/steps/device-tests.yml
```diff
@@ -0,0 +1,74 @@
+parameters:
+  platform: '' # [ android, ios ]
```
rmarinho, line 2: Can we move this to common/ we are deleting the steps folder

Redth (approved):

rmarinho (approved):

PR #13764, opened 2021-02-15
[Core] LabelHandler Preview

jsuarezruiz:
### Description of Change ###

Label Handlers. This PR will allow adding a basic Label support, in the first .NET MAUI Preview. 

| API | Android | iOS |
| ----|:--------:|:---:|
| Text  | ✅  | ✅  | 
| TextColor  | ✅  | ✅  | 

### Platforms Affected ### 

- Core
- iOS
- Android

### PR Checklist ###

- [x]  Include Handler tests


src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/ILabel.cs
```diff
@@ -1,10 +1,17 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
+﻿using Xamarin.Forms;
 
 namespace Xamarin.Platform
 {
-	public interface ILabel : IText
+	public interface ILabel : IText, IPadding
```
PureWeen, line 5: Can we remove all the properties we're not implementing?

src/Forms/src/Xamarin.Forms.Core/HandlerImpl/Button.Impl.cs
```diff
@@ -1,32 +1,26 @@
-﻿using System;
-using Xamarin.Platform;
+﻿using Xamarin.Platform;
 
 namespace Xamarin.Forms
 {
 	public partial class Button : IButton
```
PureWeen, line 5: This all probably shouldn't be in here

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/ILabel.cs
jsuarezruiz: Sure. Added all the properties to include all the methods in the Mapper and easily know what is pending, do we delete or leave these properties and empty methods?

jsuarezruiz: Removed!

src/Forms/src/Xamarin.Forms.Core/HandlerImpl/Button.Impl.cs
jsuarezruiz: Changed.

hartez (approved):

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Label/LabelHandler.Android.cs
```diff
@@ -0,0 +1,38 @@
+﻿using Android.Widget;
+using Xamarin.Forms;
+
+namespace Xamarin.Platform.Handlers
+{
+	public partial class LabelHandler : AbstractViewHandler<ILabel, TextView>
+	{
+		static Color DefaultTextColor { get; set; }
```
mattleibow, line 8: Just adding a note for the future me. This is fine for a few reasons:
 - this should always be the same value
 - it only ever runs once because the base has a check and runs it once per handler

mattleibow (approved):
Lookin' good.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/LabelExtensions.cs
```diff
@@ -4,7 +4,28 @@ namespace Xamarin.Platform
...
+		{
 			textView.Text = label.Text;
+		}
+
+		public static void UpdateTextColor(this TextView textView, ILabel label)
+		{
+			textView.UpdateTextColor(label, null);
+		}
+
+		public static void UpdateTextColor(this TextView textView, ILabel label, Forms.Color? lastUpdateColor)
```
PureWeen, line 17: should this be called defaultColor?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/LabelExtensions.cs
```diff
@@ -4,7 +4,28 @@ namespace Xamarin.Platform
...
 	{
-		public static void UpdateText(this TextView textView, ILabel label) =>
+		public static void UpdateText(this TextView textView, ILabel label)
+		{
 			textView.Text = label.Text;
+		}
+
+		public static void UpdateTextColor(this TextView textView, ILabel label)
+		{
+			textView.UpdateTextColor(label, null);
```
PureWeen, line 14: I'm curious about these places here where we pass in null because technically null is an actual setting.

It feels like when we do calls like this it should pass in whatever value we typically use for the default

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Label/LabelHandler.iOS.cs
```diff
@@ -0,0 +1,23 @@
...
+			ViewHandler.CheckParameters(handler, label);
+
+			handler.TypedNativeView?.UpdateText(label);
+		}
+
+		public static void MapTextColor(LabelHandler handler, ILabel label)
+		{
+			ViewHandler.CheckParameters(handler, label);
+
+			handler.TypedNativeView?.UpdateTextColor(label);
```
PureWeen, line 20: What about the iOS defaults?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/LabelExtensions.cs
mattleibow: We got that DefaultTextColor property in LabelHandler...

PureWeen: For example on Switch I do this

https://github.com/xamarin/Xamarin.Forms/pull/12196/files#diff-abebb275026c8510de8beeb9b1b1cb717f886959f66de018ec1f3d5a8c772099R39

Because if no default is passed in and the color used is Color.Default

It needs to do something useful...  Setting the color to null is wrong

It should either just set it back to the default or if it's Color.Default it should maybe do nothing?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Label/LabelHandler.iOS.cs
mattleibow: I think it is added in the LabelExtensions if color == default, then ColorExtensions.LabelColor

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/LabelExtensions.cs
mattleibow: Why does this have an option to pass in a default if we have the default on the handler?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/LabelExtensions.cs
mattleibow: Can we remove and be like ios and just use the default inside?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/IText.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Xamarin.Platform
 {
-	public interface IText : IFont, ITextAlignment
+	public interface IText : IFont
```
mattleibow, line 5: Might be worth removing the `TextTransform` as it is limited and might be not sufficient as it is. @hartez mentions a possible case for a `Func<string, string>` being better.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/LabelExtensions.cs
hartez: Oh, yeah, it should. That's left over from something else that was here.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Label/LabelHandler.iOS.cs
hartez: Yeah, the iOS default is just black. It's already dealt with in the UpdateTextColor method.

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/LabelExtensions.cs
hartez: `public static void UpdateTextColor(this TextView textView, ILabel label)` is never called, gonna just drop it and remove the nullability in the other method.

PureWeen: Yea that probably works for now

I think the idea here was that these extension methods are useful outside the concept of the Handlers so it's useful to think of these extension methods as something users will just use themselves from their own code

So having extensions that don't take default could be useful in those scenarios?

But we can probably just remove them for now

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Core/ILabel.cs
```diff
@@ -1,10 +1,10 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
+﻿using Xamarin.Forms;
 
 namespace Xamarin.Platform
 {
-	public interface ILabel : IText
+	public interface ILabel : IView
```
mattleibow, line 5: Should this be IText again?

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Handlers/Label/LabelHandler.iOS.cs
PureWeen: but now that's kind of inconsistent right? 

Like we could do this for literally every default. Why even store the defaults at all on the handler?

mattleibow (approved):

mattleibow (approved):

PR #13766, opened 2021-02-15
[Android] EntryRenderer.cs - SelectionChanged, does not properly handle selection anchored from right

omnilogix:
Modify  SelectionChanged method to use absolute value of new selection range.   Currently, an exception is thrown when selecting from right to left, and the SelectionLength value is not updated because the value would be negative.  This method currently only works properly when selecting from left to right.

### Issues Resolved ### 

- fixes #

### API Changes ###

 
 None

### Platforms Affected ### 

- Android
- Maybe others

None

Not applicable

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/Renderers/EntryRenderer.cs
+++ b/Xamarin.Forms.Platform.Android/Renderers/EntryRenderer.cs
@@ -434,7 +434,8 @@ namespace Xamarin.Forms.Platform.Android
 			{
 				int elementSelectionLength = System.Math.Min(EditText.Text.Length - cursorPosition, Element.SelectionLength);
 
-				var controlSelectionLength = EditText.SelectionEnd - selectionStart;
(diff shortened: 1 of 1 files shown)


jsuarezruiz:
Is there any associated issue? Above all to include tests that validate the issue.

omnillogix:
The associated issue is that the SelectedLength property is not set and not valid after making a selection from right to left.  I have a Behavior that depends on the SelectionLength being correct when a selection is made.   To validate the behavior, place an Entry control on the Xamarin form, enter some text, then select a range of characters from right to left.  The debugger output will show the exception being thrown immediately.    I am not sure if this can be unit tested because the class itself does not allow setting of the start/end points of the selection from code.

jfversluis:
We won't be doing this for Xamarin.Forms anymore in this point in time, sorry. Thanks for your time and effort on this!

PR #13784, opened 2021-02-17
Add Resizetizer.NT

mattleibow:
### Description of Change ###

Add Resizetizer.NT

### Issues Resolved ### 

- fixes #

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###


Redth (changes requested):
I am thinking we should probably not use Resizetizer as the name in the folder at least...  Less concerned about the namespace since it's just build tasks.  Maybe we need to nest Resizetizer under a SingleProject folder so we can have `/src/SingleProject/Resizetizer/src` which lets us section off things a bit more in the future too.

Redth (approved):

PR #13789, opened 2021-02-17
[Core] WebViewHandler Preview

jsuarezruiz:
### Description of Change ###

Entry Handlers. This PR will allow adding a basic Entry support, in the first .NET MAUI Preview. 

| API | Android | iOS |
| ----|:--------:|:---:|
| Source  | ✅  | ✅  | 

### Platforms Affected ### 

- Core
- iOS
- Android

### PR Checklist ###

- [x]  Include Handler tests


src/Forms/src/Xamarin.Forms.Core/HandlerImpl/WebView.Impl.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Xamarin.Platform;
+
+namespace Xamarin.Forms
+{
+	public partial class WebView : IWebView
+	{
+		WebViewSource2 IWebView.Source { get; set; }
```
jsuarezruiz, line 7: `WebViewSource2 ` must be temporary. Where we use WebViewSource in Xamarin.Forms require BindableProperties.
How are we going to manage these cases?

PR #13792, opened 2021-02-17
Move the new Layout classes to Forms; make the new layouts work in CG

hartez:
### Description of Change ###

Move the layout classes from the Sample app to Forms.

Adds a small gallery to demo using the new layouts in a Forms Page with shimmed Labels.

### DIFF
--- a/src/ControlGallery/src/Xamarin.Forms.ControlGallery.Android/Resources/layout/Tabbar.axml
+++ b/src/ControlGallery/src/Xamarin.Forms.ControlGallery.Android/Resources/layout/Tabbar.axml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<android.support.design.widget.TabLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<com.google.android.material.tabs.TabLayout xmlns:android="http://schemas.android.com/apk/res/android"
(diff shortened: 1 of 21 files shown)


jsuarezruiz (commented):
There is a compilation error in the Sample with .NET 6.
`D:\a\1\s\src\Platform.Handlers\samples\Sample.iOS\AppDelegate.cs(54,30): error CS0246: The type or namespace name 'VerticalStackLayout' could not be found (are you missing a using directive or an assembly reference?)`

PureWeen (changes requested):
So for this one I'm thinking what we should do is

- Move the existing samples (https://github.com/xamarin/Xamarin.Forms/tree/main-handler/src/Platform.Handlers/samples) to here https://github.com/xamarin/Xamarin.Forms/tree/main-handler/src/Forms
- Make those samples just use Xamarin.Forms types and not any of the inheritance
- And then just do a little cleanup so what is being sampled is a bit more clear and specific to MAUI

@rmarinho thoughts?

rmarinho:
I m think for now is fine to just add this here. But we need to have another PR that we create Maui.Controls.Samples where we start putting down a kinda of Kitchen Sink . with Maui.Controls. 

This is failing to build on .NET6 .. @hartez  you need to pull the Maui.sln and check

PureWeen (approved):

rmarinho (approved):

PR #13822, opened 2021-02-21
If the measure is invalid don't perform an arrange

PureWeen:
### Description of Change ###

Fix it so Arrange isn't called if the measure is invalid. 

As is the arrange was getting called and then the measure but then not another arrange so nothing would layout

### Testing Procedure ###
- Run the new .net maui galleries on CG. Without this change the pages are blank
- Automated unit tests
 
### PR Checklist ###


PureWeen:
I can't approve my own PR :-p 

I approve the changes @hartez  has made to this PR ✔️

hartez (approved):

PR #13846, opened 2021-02-24
Fix infinite slider width issue

hartez:
### Description of Change ###

Fix issue with UISlider (and possibly other controls) misbehaving when calling SizeThatFits with infinite width.

If the control tries to claim infinite width or height, call SizeToFit and return the frame instead. (This is basically the same fix Forms has for SliderRenderer).

### Issues Resolved ### 

- (hopefully) fixes broken 7361 test in CG

### API Changes ###
 
None

### Platforms Affected ### 

- iOS

### Testing Procedure ###

Automated device test

### PR Checklist ###

### DIFF
--- a/src/Compatibility/Core/src/Android/HandlerToRendererShim.cs
+++ b/src/Compatibility/Core/src/Android/HandlerToRendererShim.cs
@@ -37,8 +37,7 @@ namespace Microsoft.Maui.Controls.Compatibility
 
 		public SizeRequest GetDesiredSize(int widthConstraint, int heightConstraint)
 		{
-			View.Measure(widthConstraint, heightConstraint);
-			return new Size(View.MeasuredWidth, View.MeasuredHeight);
+			return ViewHandler.GetDesiredSize(widthConstraint, heightConstraint);
 		}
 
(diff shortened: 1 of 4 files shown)


PureWeen (changes requested):
Should we also change this to call GetDesiredSize on the Android side?

https://github.com/xamarin/Xamarin.Forms/blob/5c6e540b0573258b1bb6f2eae546ffdf1d5d5536/src/Platform.Renderers/src/Xamarin.Forms.Platform.Android/HandlerToRendererShim.cs#L41

hartez:
> Should we also change this to call GetDesiredSize on the Android side?
> 
> https://github.com/xamarin/Xamarin.Forms/blob/5c6e540b0573258b1bb6f2eae546ffdf1d5d5536/src/Platform.Renderers/src/Xamarin.Forms.Platform.Android/HandlerToRendererShim.cs#L41

Good call. Done.

mattleibow (approved):
Can't comment on how everything fits together as I am still grasping all the internals, but the code looks sound.

jsuarezruiz (approved):

PR #13849, opened 2021-02-24
[Handlers] Map CharacterSpacing property in LabelHandlers

jsuarezruiz:
### Description of Change ###

Map **CharacterSpacing** property in LabelHandlers.

### Platforms Affected ### 

- iOS
- Android

### PR Checklist ###

- [x] Include handlers tests

### DIFF
--- a/src/Core/src/Platform/iOS/CharacterSpacingExtensions.cs
+++ b/src/Core/src/Platform/iOS/CharacterSpacingExtensions.cs
@@ -1,0 +1,58 @@
+﻿using Foundation;
+using UIKit;
+
+namespace Microsoft.Maui
+{
+	public static class CharacterSpacingExtensions
+	{
+		public static NSMutableAttributedString? AddCharacterSpacing(this NSAttributedString attributedString, string text, double characterSpacing)
+		{
+			if (attributedString == null && characterSpacing == 0)
+				return null;
+
+			NSMutableAttributedString? mutableAttributedString = attributedString as NSMutableAttributedString;
+
+			if (attributedString == null || attributedString.Length == 0)
+			{
+				mutableAttributedString = text == null ? new NSMutableAttributedString() : new NSMutableAttributedString(text);
+			}
+			else
+			{
+				mutableAttributedString = new NSMutableAttributedString(attributedString);
+			}
+
+			AddKerningAdjustment(mutableAttributedString, text, characterSpacing);
+
+			return mutableAttributedString;
+		}
+
+		internal static bool HasCharacterAdjustment(this NSMutableAttributedString mutableAttributedString)
+		{
+			if (mutableAttributedString == null)
+				return false;
+
+			var attributes = mutableAttributedString.GetAttributes(0, out NSRange removalRange);
+
+			for (uint i = 0; i < attributes.Count; i++)
+				if (attributes.Keys[i] is NSString nSString && nSString == UIStringAttributeKey.KerningAdjustment)
+					return true;
+
+			return false;
+		}
+
+		internal static void AddKerningAdjustment(NSMutableAttributedString mutableAttributedString, string? text, double characterSpacing)
+		{
+			if (!string.IsNullOrEmpty(text))
+			{
+				if (characterSpacing == 0 && !mutableAttributedString.HasCharacterAdjustment())
+					return;
+
+				mutableAttributedString.AddAttribute
+				(
+					UIStringAttributeKey.KerningAdjustment,
+					NSObject.FromObject(characterSpacing), new NSRange(0, text != null ? text.Length - 1 : 0)
+				);
+			}
+		}
+	}
+}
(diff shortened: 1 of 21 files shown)


src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/UnitExtensions.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Xamarin.Platform
+{
+	public static class UnitExtensions
+	{
+		public static float EmCoefficient = 0.0624f;
+
+		public static float ToEm(this double pt)
+		{
+			return (float)pt * EmCoefficient; //Coefficient for converting Pt to Em
```
hartez, line 9: This ... can't be right. The value for em is dependent on the font size. And even if it weren't, this assumes that we're using something like a 16.025pt font.

hartez (changes requested):

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/UnitExtensions.cs
hartez: Looks like this came straight out of our existing code; I'll open up an issue for it, and we'll deal with it separately.

hartez: https://github.com/xamarin/Xamarin.Forms/issues/13859

src/Platform.Handlers/src/Xamarin.Platform.Handlers/Platform/Android/LabelExtensions.cs
```diff
@@ -20,7 +20,15 @@ public static void UpdateTextColor(this TextView textView, ILabel label, Forms.C
 			else
 			{
 				textView.SetTextColor(textColor.ToNative());
-			}				
+			}
+		}
+
+		public static void UpdateCharacterSpacing(this TextView textView, ILabel label)
+		{
+			if (NativeVersion.IsAtLeast(21))
```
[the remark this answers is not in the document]
jsuarezruiz: Right!. Removed the unnecessary API Level check.

hartez (approved):

src/Core/src/Platform/iOS/CharacterSpacingExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+	public static class CharacterSpacingExtensions
+	{
+		public static NSMutableAttributedString? AddCharacterSpacing(this NSAttributedString attributedString, string text, double characterSpacing)
+		{
+			if (attributedString == null && characterSpacing == 0)
+				return null;
+
+			NSMutableAttributedString? mutableAttributedString = attributedString as NSMutableAttributedString;
+
+			if (attributedString == null || attributedString.Length == 0)
```
mattleibow, line 15: Should this be checking `mutableAttributedString ` instead? If not, then what is the logic above for? We could just have a variable definition. But this check also does not seem to fit.

src/Core/src/Platform/iOS/LabelExtensions.cs
```diff
@@ -23,5 +23,16 @@ public static void UpdateTextColor(this UILabel nativeLabel, ILabel label)
 				nativeLabel.TextColor = textColor.ToNative(textColor);
 			}
 		}
+
+		public static void UpdateCharacterSpacing(this UILabel nativeLabel, ILabel label)
+		{
+			if (string.IsNullOrEmpty(label.Text))
+				return;
+
+			var textAttr = nativeLabel.AttributedText?.AddCharacterSpacing(label.Text, label.CharacterSpacing);
```
mattleibow, line 32: This logic with the attributed text is very different to the decorations PR. That one check for things, this just uses text. Will this override decorations?

src/Core/src/Platform/iOS/CharacterSpacingExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+		{
+			if (!string.IsNullOrEmpty(text))
+			{
+				if (characterSpacing == 0 && !mutableAttributedString.HasCharacterAdjustment())
+					return;
+
+				mutableAttributedString.AddAttribute
+				(
+					UIStringAttributeKey.KerningAdjustment,
+					NSObject.FromObject(characterSpacing), new NSRange(0, text != null ? text.Length - 1 : 0)
```
mattleibow, line 53: If this is based on text length, what happens when the text changes?

jsuarezruiz:
Moved to https://github.com/dotnet/maui/pull/463
Let's continue the review and discussions there!

src/Core/src/Platform/iOS/CharacterSpacingExtensions.cs
jsuarezruiz: Like in the Renderer, in the Handler need to invoke again the CharacterSpacing logic.

PR #13850, opened 2021-02-24
[Handlers] Map MaxLines and LineBreakMode properties in LabelHandlers

jsuarezruiz:
### Description of Change ###

Map `MaxLines` and `LineBreakMode` property in LabelHandlers.

NOTE: Include directly two properties because are very related ones.

### Platforms Affected ### 

- iOS
- Android

### PR Checklist ###

- [x] Include handlers tests


hartez (approved):

src/Core/src/Platform/Android/LabelExtensions.cs
```diff
@@ -22,5 +23,63 @@ public static void UpdateTextColor(this TextView textView, ILabel label,  Maui.C
...
+        public static void UpdateLineBreakMode(this TextView textView, ILabel label)
+        {
+            textView.SetLineBreakMode(label);
+        }
+
+        public static void UpdateMaxLines(this TextView textView, ILabel label)
+        {
+            var maxLines = label.MaxLines;
+
+            if (maxLines == 0)
```
rachelkang, line 36: is the default value 0?
In original code: `(int)Label.MaxLinesProperty.DefaultValue`

rachelkang (changes requested):
also need to rebase main-handler again!

src/Core/src/Platform/Android/LabelExtensions.cs
```diff
@@ -22,5 +23,63 @@ public static void UpdateTextColor(this TextView textView, ILabel label,  Maui.C
 				textView.SetTextColor(textColor.ToNative());
 			}				
 		}
-	}
+
+        public static void UpdateLineBreakMode(this TextView textView, ILabel label)
+        {
+            textView.SetLineBreakMode(label);
```
mattleibow, line 29: What happens if I set max lines, then change this. For example, I set MaxLines to 3 and Break to Character. Then when I swap to Word, it is going to reset the max lines to int.MaxValue.

src/Core/src/Platform/Android/LabelExtensions.cs
mattleibow: If anything, could this be a named constant so we can at least know what this does? `int InfiniteLines = 0;`

src/Core/src/Platform/Android/LabelExtensions.cs
```diff
@@ -22,5 +23,63 @@ public static void UpdateTextColor(this TextView textView, ILabel label,  Maui.C
...
+            }
+
+            textView.SetMaxLines(maxLines);
+        }
+
+        internal static void SetLineBreakMode(this TextView textView, ILabel label)
+        {
+            var lineBreakMode = label.LineBreakMode;
+
+            int maxLines = int.MaxValue;
```
mattleibow, line 50: As noted in a comment above, this will always override this previously set max lines. Changing the break mode resets the maxLines.

mattleibow (changes requested):

mattleibow:
I updated the PR with some tests. Let me get to fixin' 'em!

src/Core/tests/DeviceTests/Handlers/Label/LabelHandlerTests.Android.cs
```diff
@@ -23,5 +25,36 @@ Task ValidateNativeBackgroundColor(ILabel label, Color color)
...
-	}
+
+        int GetNativeMaxLines(LabelHandler labelHandler) =>
+            GetNativeLabel(labelHandler).MaxLines;
+
+        TextUtils.TruncateAt GetNativeEllipsize(LabelHandler labelHandler) =>
+           GetNativeLabel(labelHandler).Ellipsize;
+
+        [Fact(DisplayName = "[LabelHandler] LineBreakMode Initializes Correctly")]
+        public async Task LineBreakModeInitializesCorrectly()
```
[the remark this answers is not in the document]
mattleibow: Moved to the shared tests

src/Controls/samples/Controls.Sample/Pages/MainPage.cs
```diff
@@ -8,6 +8,18 @@ namespace Maui.Controls.Sample.Pages
...
 	public class MainPage : ContentPage, IPage
 	{
+		const string longLoremIpsum =
+			"Lorem ipsum dolor sit amet, consectetur adipiscing elit. " +
+			"Quisque ut dolor metus. Duis vel iaculis mauris, sit amet finibus mi. " +
+			"Etiam congue ornare risus, in facilisis libero tempor eget. " +
+			"Phasellus mattis mollis libero ut semper. In sit amet sapien odio. " +
+			"Sed interdum ullamcorper dui eu rutrum. Vestibulum non sagittis justo. " +
+			"Cras rutrum scelerisque elit, et porta est lobortis ac. " +
+			"Pellentesque eu ornare tortor. Sed bibendum a nisl at laoreet.";
```
mattleibow, line 18: @hartez I cant actually use this long text in the vertical stack. As soon as the text lakes the label wider than the screen, it crashes on Android. Something about an index out of bounds in the measure.

mattleibow:
OK... So I did stuff, hopefully it is correct logic.

I linked the two properties so that they correctly set the native view's values depending on the break mode selected. When setting either of the properties, both native values are changed in the pattern:

wrapping -> use max lines
no wrap -> use 1

max <=0 -> all lines, respecting wrap
max > 0 -> x lines, respecting wrap

With regards to the max lines, this is also a weird case because the docs say an impossibility: https://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/text/label#display-a-specific-number-of-lines

> When MaxLines is 0, the Label isn't displayed

The platforms don't reflect this. In fact, the whole system looks weird as they don't respect the break mode really. 

0 lines mean all the lines on iOS but only 1 on Android. And, when the max lines are set, the no wrap is ignored.

Look at this with Forms today


Screenshot 2021-02-27 at 04 01 23

mattleibow:
Actually, if you set max lines > 1 and then set a truncation, the platforms are even weirder. Android forces a single line, iOS on the other hand has multiple lines, but applies the truncation on the last line.

jsuarezruiz:
`MaxLines`property implementation moved to: https://github.com/dotnet/maui/pull/457

jsuarezruiz:
`LineBreakMode` property implementation moved to https://github.com/dotnet/maui/pull/458

jsuarezruiz:
I close this PR and continue the discussions and review in .NET MAUI repository. cc @mattleibow

PR #13853, opened 2021-02-24
[Handlers] Map TextDecorations property in LabelHandlers

jsuarezruiz:
### Description of Change ###

Map `TextDecorations` property in LabelHandlers.

### Platforms Affected ### 

- Android
- UWP

### PR Checklist ###

- [x]  Include handlers tests

### DIFF
--- a/src/Forms/samples/Maui.Controls.Sample/Pages/MainPage.cs
+++ b/src/Forms/samples/Maui.Controls.Sample/Pages/MainPage.cs
@@ -9,10 +9,12 @@ namespace Maui.Controls.Sample.Pages
 	public class MainPage : ContentPage, IPage
 	{
 		MainPageViewModel _viewModel;
+
 		public MainPage() : this(App.Current.Services.GetService<MainPageViewModel>())
 		{
 
 		}
+
 		public MainPage(MainPageViewModel viewModel)
 		{
 			BindingContext = _viewModel = viewModel;
@@ -24,6 +26,9 @@ namespace Maui.Controls.Sample.Pages
 			label.Margin = new Thickness(15, 10, 20, 15);
 
 			verticalStack.Add(label);
+
+			var underlineLabel = new Label { Text = "underline", TextDecorations = TextDecorations.Underline };
+			verticalStack.Add(underlineLabel);
 
 			var button = new Button() { Text = _viewModel.Text, WidthRequest = 200 };
 			var button2 = new Button()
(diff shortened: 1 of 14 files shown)


hartez (approved):

src/Core/tests/DeviceTests/Handlers/Label/LabelHandlerTests.Android.cs
```diff
@@ -23,5 +25,33 @@ Task ValidateNativeBackgroundColor(ILabel label, Color color)
 				GetNativeLabel(CreateHandler(label)).AssertContainsColor(color);
 			});
 		}
-	}
+
+        PaintFlags GetNativeTextDecorations(LabelHandler labelHandler) =>
+            GetNativeLabel(labelHandler).PaintFlags;
+
+        [Fact(DisplayName = "[LabelHandler] TextDecorations Initializes Correctly")]
+        public async Task TextDecorationsInitializesCorrectly()
```
rachelkang, line 33: Why isn't the test being added to LabelHandlerTests?

rachelkang (commented):
need to rebase main-handler here too

src/Core/src/Platform/iOS/LabelExtensions.cs
```diff
@@ -23,5 +24,31 @@ public static void UpdateTextColor(this UILabel nativeLabel, ILabel label)
 				nativeLabel.TextColor = textColor.ToNative(textColor);
 			}
 		}
+
+		public static void UpdateTextDecorations(this UILabel nativeLabel, ILabel label)
+		{
+			if (nativeLabel.AttributedText != null && !(nativeLabel.AttributedText?.Length > 0))
```
mattleibow, line 30: How does `AttributedText` work? I see we always set `Text`, so will this be null, or does iOS create some wrapper for us?

src/Core/src/Platform/iOS/LabelExtensions.cs
```diff
@@ -23,5 +24,31 @@ public static void UpdateTextColor(this UILabel nativeLabel, ILabel label)
...
+				return;
+
+			var textDecorations = label?.TextDecorations;
+
+			var newAttributedText = nativeLabel.AttributedText != null ? new NSMutableAttributedString(nativeLabel.AttributedText) : new NSMutableAttributedString(label?.Text ?? string.Empty);
```
mattleibow, line 35: How does this fit with the check above? At the top, if there is not attrText, then bail. Bu here we check again.

PR #13865, opened 2021-02-25
[Handlers] Map CornerRadius property in ButtonHandlers

jsuarezruiz:
### Description of Change ###

Map CornerRadius property in ButtonHandlers.

### Platforms Affected ### 

- iOS
- Android

### PR Checklist ###

- [x] Include handlers tests


hartez:
I think we should hold off on this one until we have a better handle on what we're doing with shapes. If we can do a generic clip shape at the MAUI level, then we won't need to add a specific CornerRadius property; we can keep that at the Controls level and implement it using the clip shape.

GalaxiaGuy:
If this does ahead (I don't know what the plan is with shapes), should it be based on this:

https://github.com/dotnet/maui/issues/10

hartez:
Going to close this for now; we're close to having the underlying parts in MAUI that we can look at doing clipping shapes again for universal rounded/beveled/whatever edges on things. 

If we decide not to go that direction, we can re-open and port this PR.

PR #13878, opened 2021-02-26
Wire up Controls.Layouts and Content Page

PureWeen:
### Description of Change ###

- wired up Forms.Init so Controls can run successfully
- Wired together various communication points for old style layouts and visual elements
- Moved IFrameWorkElement to VisualElement
- Moved Margin to IView

### Testing Procedure ###
- inside MyApp you'll see some rows you can uncomment to register old renderers
- inside MainPage you'll see a different set of controls you can add that are purely Compatibility Layouts and not new layouts

### PR Checklist ###


build.cake
```diff
@@ -903,11 +903,59 @@ Task("Android100")
                     .WithProperty("AndroidTargetFrameworks", "MonoAndroid10.0"));
     });
 
+
```
PureWeen, line 906: This provides a set of targets you can run that clean the obj/bin folders, compile the slns, and then open them in VS ready to rock

src/Compatibility/Core/src/Android/Renderers/PageRenderer.cs
```diff
@@ -221,6 +221,15 @@ void IOrderedTraversalController.UpdateTraversalOrder()
...
 		{
+			var deviceIndependentLeft = Context.FromPixels(l);
+			var deviceIndependentTop = Context.FromPixels(t);
+			var deviceIndependentRight = Context.FromPixels(r);
+			var deviceIndependentBottom = Context.FromPixels(b);
+
+			var destination = Rectangle.FromLTRB(deviceIndependentLeft, deviceIndependentTop,
+				deviceIndependentRight, deviceIndependentBottom);
+
+			(Element as IFrameworkElement)?.Arrange(destination);
```
PureWeen, line 232: I figure this will get removed once we have more of the window/page parts of MAUI built out

src/Controls/samples/Controls.Sample/Controls.Sample.csproj
```diff
@@ -3,12 +3,6 @@
     <TargetFrameworks>netstandard2.1;Xamarin.iOS10;MonoAndroid10.0</TargetFrameworks>
   </PropertyGroup>  
   <Import Project="..\..\..\..\.nuspec\Microsoft.Maui.Controls.MultiTargeting.targets" />
-  <!--<ItemGroup Condition="$(TargetFramework.StartsWith('Xamarin.iOS')) == true ">
```
PureWeen, line 6: these projects don't compile with msbuild for some reason. I don't think sdk.extras can handle the conditionals

src/Controls/src/Core/HandlerImpl/VisualElement.Impl.cs
```diff
@@ -0,0 +1,103 @@
...
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Text;
+using Microsoft.Maui.Layouts;
+
+namespace Microsoft.Maui.Controls
+{
+	public partial class VisualElement : IFrameworkElement
+	{
+		private IViewHandler _handler;
```
PureWeen, line 11: Move IFrameworkElement up to VisualElement so that Page also implements IFrameworkElement

src/Controls/src/Core/Layout.cs
```diff
@@ -307,7 +307,10 @@ internal static void LayoutChildIntoBoundingRegion(View child, Rectangle region,
 			region.Y += margin.Top;
 			region.Height -= margin.VerticalThickness;
 
-			child.Layout(region);
+			if (child is IFrameworkElement fe)
```
PureWeen, line 310: Arrange calls Layout and does extra stuff to make it work if it has purely handler based controls. We might not need this if statement

src/Core/src/Hosting/MauiServiceProvider.cs
```diff
@@ -33,6 +28,12 @@ public MauiServiceProvider(IMauiServiceCollection collection)
 				baseType = baseType.BaseType;
 			}
 
+			foreach (var interfac in serviceType.GetInterfaces())
```
PureWeen, line 31: I swapped interfaces with base types. It seems like if there's something registered against a concrete base type that should win..  Though I could see cases where you want the interface to win so perhaps it needs to walk each base type and check the concrete type/interfaces there to pick the winner?

src/Core/src/Platform/Android/MauiAppCompatActivity.cs
```diff
@@ -29,25 +29,24 @@ protected override void OnCreate(Bundle? savedInstanceState)
...
 
 			//Hack for now we set this on the App Static but this should be on IFrameworkElement
 			App.Current.SetHandlerContext(window.MauiContext);
 
-			var content = window.Page.View;
+			var content = (window.Page as IView) ?? 
+				window.Page.View;
 
 			CoordinatorLayout parent = new CoordinatorLayout(this);
-			NestedScrollView main = new NestedScrollView(this);
```
PureWeen, line 42: There's not really a reason to have a scrollview at the root. If user wants a scrollview they can add a scroll view. Thanks to this PR!!

src/Compatibility/Core/src/Android/Compatibility.Android.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="MSBuild.Sdk.Extras">
   <PropertyGroup>
     <Description>Android Compatibility Backend for Microsoft MAUI</Description>
-    <AssemblyName>Microsoft.Maui.Controls.Compatibility</AssemblyName>
+    <AssemblyName>Microsoft.Maui.Controls.Compatibility.Android</AssemblyName>
```
PureWeen, line 4: MSBUILD can't handle this project and the iOS one having the same assembly name. It gets confused. Since we aren't packaging these versions of the dlls this is hopefully fine and doesn't ultimately lead to weird confusing scenarios

src/Compatibility/Core/src/iOS/Compatibility.iOS.csproj
```diff
@@ -3,7 +3,7 @@
   <Import Condition="'$(EnvironmentBuildPropsImported)' != 'True'" Project="..\..\..\..\..\eng\Environment.Build.props" />
   <PropertyGroup>
     <Description>iOS Compatibility Backend for Microsoft.Maui</Description>
-    <AssemblyName>Microsoft.Maui.Controls.Compatibility</AssemblyName>
+    <AssemblyName>Microsoft.Maui.Controls.Compatibility.iOS</AssemblyName>
```
PureWeen, line 6: MSBUILD can't handle this project and the Android one having the same assembly name. It gets confused. Since we aren't packaging these versions of the dlls this is hopefully fine and doesn't ultimately lead to weird confusing scenarios

src/Compatibility/Core/src/iOS/Forms.cs
```diff
@@ -167,14 +168,18 @@ public static void SetFlags(params string[] flags)
...
 				return;
+
+			ActivationState = activationState;
 			IsInitialized = true;
 
 			Microsoft.Maui.Controls.Internals.Registrar.RegisterRendererToHandlerShim(RendererToHandlerShim.CreateShim);
-			Color.SetAccent(ColorExtensions.AccentColor.ToColor());
+
+			//TODO: MAUI Accent Color?
+			Color.SetAccent(Color.FromRgba(50, 79, 133, 255));
```
PureWeen, line 182: AccentColor was moved to the maui.core dll which has this value marked as internal

src/Core/src/MauiApp.cs
```diff
@@ -5,7 +5,7 @@ namespace Microsoft.Maui
 {
 	public abstract class MauiApp : App
 	{
-		public abstract IWindow GetWindowFor(IActivationState state);
+		public abstract IWindow CreateWindowFor(IActivationState state);
```
mattleibow, line 8: If you think about it, this is not a very .NET name. Why can't we just use `CreateWindow`?

src/Core/src/Core/IFrameworkElement.cs
```diff
@@ -22,7 +22,5 @@ public interface IFrameworkElement
 
 		double Width { get; }
 		double Height { get; }
-
-		Thickness Margin { get; }
```
mattleibow, line 26: We may still want the margin here as if we do something like a shape, we still want that to have a margin. What was your thought behind the move?

src/Controls/src/Core/HandlerImpl/VisualElement.Impl.cs
mattleibow: I like this because if we ever do shapes or some non-view control, we can still lay it out.

src/Compatibility/Core/src/Android/Forms.cs
```diff
@@ -266,17 +271,19 @@ internal static void SendViewInitialized(this VisualElement self, global::Androi
 		}
 
 		static void SetupInit(
-			Context activity,
+			IMauiContext context,
 			Assembly resourceAssembly,
 			InitializationOptions? maybeOptions = null
 		)
 		{
+			var activity = (Activity)context.Context;
```
mattleibow, line 279: Why do we cast as an Activity? Can it just stay as Context since we never use the type.

src/Compatibility/Core/src/Android/Forms.cs
```diff
@@ -185,9 +186,12 @@ public static Color GetColorButtonNormal(Context context)
 			return _ColorButtonNormal;
 		}
 
+		public static void Init(IActivationState activationState) =>
+			Init(activationState.Context, activationState.SavedInstance);
+
 		// Provide backwards compat for Forms.Init and AndroidActivity
 		// Why is bundle a param if never used?
-		public static void Init(Context activity, Bundle bundle)
+		public static void Init(IMauiContext context, Bundle bundle)
```
mattleibow, line 194: How would someone migrating their app use this? I would expect there might be some extra code to add, but maybe we can wrap these things for them automatically? Otherwise, people will not know what do do.

src/Core/src/Platform/Android/MauiAppCompatActivity.cs
```diff
@@ -29,25 +29,24 @@ protected override void OnCreate(Bundle? savedInstanceState)
 			if (mauiApp.Services == null)
 				throw new InvalidOperationException("App was not initialized");
 
-			var window = mauiApp.GetWindowFor(null!);
+			var mauiContext = new HandlersContext(mauiApp.Services, this);
+			var window = mauiApp.CreateWindowFor(new ActivationState(savedInstanceState, mauiContext));
 
-			window.MauiContext = new HandlersContext(mauiApp.Services, this);
+			window.MauiContext = mauiContext;
```
mattleibow, line 35: As noted in my comment about the Ini, if this is just a simple wrap, then maybe we can do it for the users coming in from the old way?

More importantly, the `HandlersContext` needs to be renamed `MauiContext` to match iOS and the filename and the base interface.

src/Compatibility/Core/src/iOS/Forms.cs
mattleibow: Maybe it should be made public because people will expect it to be there. I also see it in ColorExtensions in compatibility. Maybe we should be using that?

src/Compatibility/Core/src/iOS/Forms.cs
```diff
@@ -167,14 +168,18 @@ public static void SetFlags(params string[] flags)
 				Profile.Enable();
 		}
 
-		public static void Init()
+		public static void Init(IActivationState activationState)
```
mattleibow, line 171: Like with Android, we can wrap all the things for the user if nothing was specified.

src/Core/src/Layouts/LayoutExtensions.cs
```diff
@@ -43,5 +43,13 @@ public static Rectangle ComputeFrame(this IFrameworkElement frameworkElement, Re
 			return new Rectangle(bounds.X + margin.Left, bounds.Y + margin.Top,
 				frameWidth, frameHeight);
 		}
+
+		static Thickness GetMargin(this IFrameworkElement frameworkElement)
```
mattleibow, line 47: This is only needed because the framework element no longer has margin, but I got a comment on maybe we do want it still.

src/Controls/samples/Controls.Sample.SingleProject/MyApp.cs
```diff
@@ -13,7 +13,14 @@ public class MyApp : MauiApp
...
 				});
 
-		public override IWindow GetWindowFor(IActivationState state) =>
-			Services.GetService<IWindow>();
+		public override IWindow CreateWindowFor(IActivationState state)
+		{
+#if (ANDROID || IOS)
+
+			// This will probably go into a compatibility app or window
+			Microsoft.Maui.Controls.Compatibility.Forms.Init(state);
```
mattleibow, line 21: Maybe we should do this internally for now? None of the controls can really work without it once we add normal properties. For example, In my Label PR, as soon as I touch the font properties, then it requires that Init be called.

Maybe there is need for overloads and whatnot, so we shouldn't. But just a thought.

But also, maybe we should still call this in the activity and delegate instead of the shared code since this is a compat line?

src/Core/src/Core/IFrameworkElement.cs
PureWeen: Mainly that @hartez told me to move it :-)

src/Compatibility/Core/src/Android/Forms.cs
PureWeen: Honestly I don't think we'll even have an Init class once we ship.... The reason we're calling init right now is because we haven't broken up Forms/Platform/Device classes

Once those are deconstructed into Maui concepts we won't need Init. What we will need is a way to opt into Assembly scanning for old school renderers if people still want to enable that. That's something we will have on the builder

builder.ScanForCompatibilityRenderers()

or something like that

src/Compatibility/Core/src/iOS/Forms.cs
PureWeen: Same comment here as above. I think Init will go away once all is said and done. This is just a way to make it all work for now

src/Controls/samples/Controls.Sample.SingleProject/MyApp.cs
PureWeen: In line with what I mentioned above... I think Init will ultimately go away and all we are going to have left is for users to opt into old school assembly scanning

src/Core/src/Layouts/LayoutExtensions.cs
PureWeen: i'll just revert the margin changes for now to limit the number of things addressed in this PR

PureWeen: Actually, I'm going to just leave it like this for now. We currently don't have margin on VisualElement so we have to add extra code on Visualelement to now account for a margin that doesn't exist. Putting Margin on IView keeps it inline with our current structures and I think we can modify once we organize it all a bit better

src/Core/src/Core/IFrameworkElement.cs
PureWeen: https://github.com/xamarin/Xamarin.Forms/pull/13878#discussion_r584893478

src/Core/src/Platform/Android/MauiAppCompatActivity.cs
PureWeen: > As noted in my comment about the Ini, if this is just a simple wrap, then maybe we can do it for the users coming in from the old way?

Once we have more defined windows/app/builder classes build then I think that will address it for users coming from old way.

I imagine the instructions will go
- Make your currrent "FormsAppCompatActibvity" inherit from MauiAppCompatActivity
- Delete all our code
- If you were Initing other third party libraries then you can go to the builder and do it like this... Or you can still do it here

src/Core/src/MauiApp.cs
PureWeen: Changed

mattleibow (approved):

src/Controls/src/Core/HandlerImpl/ContentPage.Impl.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Maui.Controls
+{
+	// TODO: We don't currently have any concept of a page in Maui
+	// so this just treats it as a layout for now
+	public partial class ContentPage : Microsoft.Maui.ILayout
```
rmarinho, line 9: Can we do a IPage instead since we have that? or better in a new PR

IPage : ILayout

?

PureWeen: Initially I had done that...   In the end I don't think `IPage : ILayout` so I'm just going for a really light touch approach here.

Once someone implements a PageHandler then I think all this will shake out in a more accurate way

PR #13881, opened 2021-02-26
Make .NET 6 Packable

Redth:
We are going to start by enabling `-t:Pack` to work on .NET 6.0 projects for initial MAUI previews.

- Added common nuget package properties to the `Directory.Build.props`
- Added more `-net6.csproj` versions of projects
- Build.Tasks project now is responsible for packaging the various .targets/.props, build tasks and their dependencies
- Cleaned up strong naming on the build tasks project
- Removed old net desktop profiles from projects (these were needed for vs2017)
- Script:  `pwsh ./.nuspec/package.ps1 -packageVersion "0.1.0-p2"` builds everything to `./artifacts/`


src/Compatibility/Core/src/Android.FormsViewGroup/Compatibility.Android.FormsViewGroup-net6.csproj
```diff
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net6.0-android</TargetFrameworks>
+    <TargetFrameworks>net6.0-ios;net6.0-android</TargetFrameworks>
```
mattleibow, line 3: Why is iOS here if this is just for Android?

src/Controls/src/Core.Design/toolbox/Microsoft.Maui.toolbox.xml
```diff
@@ -0,0 +1,205 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<Catalog Version="3.3">
```
mattleibow, line 3: This file looks new, where did it come from?

Redth: I didn't delete the old one.. I initially thought it might be useful to keep the old one for control gallery or something, building on old TFM's because this is what the designer uses to map some things out, but we're changing control gallery namespaces and such too anyway so it's probably not particularly useful to keep the old one.  Fixed in another commit.

src/Compatibility/Core/src/Android.FormsViewGroup/Compatibility.Android.FormsViewGroup-net6.csproj
Redth: Ahh yes, good catch, not sure why this was left with iOS after some testing.

build.cake
```diff
@@ -664,10 +664,17 @@ Task("provision")
     .IsDependentOn("provision-monosdk"); // always provision monosdk last otherwise CI might fail
 
 Task("NuGetPack")
-    .Description("Build and Create Nugets")
-    .IsDependentOn("Restore")
-    .IsDependentOn("BuildForNuget")
-    .IsDependentOn("_NuGetPack");
+    .Description("Build and Create Nugets").Does(()=> {
```
mattleibow, line 667: This will stop all the builds from running. Maybe we can have a new target that the "packaging" agent does? Then we can have another artifact that we release. I'm worried that something will be skipped. At least keep `BuildForNuget`

src/Controls/src/Build.Tasks/Controls.Build.Tasks.csproj
```diff
@@ -10,52 +10,79 @@
...
+		<None Include="$(TargetPath)" Visible="False" Pack="True" PackagePath="buildTransitive;build" />
+
+
+		<None Include="..\..\..\..\.nuspec\Microsoft.Maui.Controls.props" Visible="False" Pack="True" PackagePath="buildTransitive;build" />
+		<None Include="..\..\..\..\.nuspec\Microsoft.Maui.Controls.targets" Visible="False" Pack="True" PackagePath="buildTransitive;build" />
+		<None Include="..\..\..\..\.nuspec\Microsoft.Maui.Controls.SingleProject.props" Visible="False" Pack="True" PackagePath="buildTransitive;build" />
+		<None Include="..\..\..\..\.nuspec\Microsoft.Maui.Controls.SingleProject.targets" Visible="False" Pack="True" PackagePath="buildTransitive;build" />
+
+		<None Include="..\..\..\..\.nuspec\Microsoft.Maui.Controls.DefaultItems.props" Visible="False" Pack="True" PackagePath="buildTransitive\Microsoft.Maui.Controls.Build.Tasks.props;build\Microsoft.Maui.Controls.Build.Tasks.props" />
+		<None Include="..\..\..\..\.nuspec\Microsoft.Maui.Controls.DefaultItems.targets" Visible="False" Pack="True" PackagePath="buildTransitive\Microsoft.Maui.Controls.Build.Tasks.targets;build\Microsoft.Maui.Controls.Build.Tasks.targets" />
```
mattleibow, line 50: Maybe we should rename the file instead . Or, is this meant to be here at all? Shouldn't this be in MS.Maui.Core?

src/Controls/src/Build.Tasks/Controls.Build.Tasks.csproj
```diff
@@ -80,5 +107,5 @@
 			<_CopyItems Include="$(TargetDir)*.dll" />
 		</ItemGroup>
 		<Copy SourceFiles="@(_CopyItems)" DestinationFolder="..\..\..\..\.nuspec\$(TargetFramework)\" ContinueOnError="true" Retries="0" />
```
mattleibow, line 109: We need the copy or it won't build in the IDE from a clean checkout as it wont exist. But, what I will say is that copying to .nuspec is not the best, maybe we should copy to a different location and also include the targets and props there.

PureWeen: @StephaneDelcroix  ?

src/Controls/src/Core/VisualElement.cs
```diff
@@ -920,7 +920,7 @@ internal void InvalidateStateTriggers(bool attach)
 
 		internal void MockBounds(Rectangle bounds)
 		{
-#if NETSTANDARD2_0
+#if NETSTANDARD2_0 || NET6_0
```
PureWeen, line 923: Can we just delete this if/def? 

This was only here because of us building against NS1_0

PureWeen (approved):

src/Controls/src/Build.Tasks/Controls.Build.Tasks.csproj
mattleibow: I fixed this. The signing stuff still needs to be done again.

eng/pipelines/handlers.yml
```diff
@@ -121,7 +121,51 @@ stages:
...
+            errorActionPreference: stop
+
+          - pwsh: |
+              & dotnet tool install --global boots
+              & boots $(Android.Msi)
+              & boots $(iOS.Msi)
+            displayName: install .NET workloads
+            errorActionPreference: stop
+
+          - pwsh: ./.nuspec/package.ps1 -configuration Release -packageVersion 0.0.1-preview
```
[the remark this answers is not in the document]
mattleibow: Also the version numbers are not the same, note resizetizer and build.tasks
image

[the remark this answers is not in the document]
rmarinho: Why don't we use what we had? Version.Targets that uses GitInfo? I would just move to use SemVer2 ..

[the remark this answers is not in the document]
mattleibow: We should. I'm not sure how gitinfo work tbh. Do I just pack and things work?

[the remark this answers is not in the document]
mattleibow: it was the net6 projects that seemed to have a different task order meaning the gitinfo ran after the nuspec

PR #13883, opened 2021-02-26
Fix UWP threading exception

### LINKED ISSUE #8840: [Bug] Using fonts from a non-UI thread results in exception with HRESULT 0x8001010E (RPC_R_WRONG_THREAD) on UWP
### Description

Using fonts from a non-UI thread, even when the font is not being used in a page's UI hierarchy, results in an exception with HRESULT 0x8001010E (RPC_R_WRONG_THREAD) on UWP.

This prevents pages being prepared away from the UI thread before they are actually required, meaning that an obvious performance improvement technique cannot be used on UWP if fonts are used at all (the same problem happens if XAML is used as well).

### Steps to Reproduce

1. Add a Button to a page, with the Button defined as follows:

                        new Button
                        {
                            Text = "Use a font on a non-UI thread",
                            Command = new Command(() =>
                            {
                                Task.Run(() =>
                                {
                                    Xamarin.Forms.Label label = new Xamarin.Forms.Label
                                    {
                                        Text = "Hello",
                                        FontAttributes = FontAttributes.Italic
                                    };
                                });
                            })
                        }

2. Run the code and press the Button
3. The exception occurs, even though the View with the FontAttribute being set is not hooked into the UI hierarchy.

### Expected Behavior

No exception. Font-related properties should be set as per the code.

### Actual Behavior

Exception with HRESULT 0x8001010E (RPC_R_WRONG_THREAD)

### Basic Information

- Version with issue: 4.3.0.991211 (but this has been a problem for a long time)
- IDE: VS2019
- Platform Target Frameworks: 
  - UWP:  18362

charlesroddie:
### Description of Change ###

Removes an instance of an inverted dependency where creating a Xamarin.Forms object (including a Label with a font size) causes an exception, leading to crashes on user systems when done off the UI thread.

### Issues Resolved ### 

- fixes #8840

### Platforms Affected ### 

- UWP

- UWP apps will not get unexpected crashes when creating Xamarin.Forms objects off the UI thread.
- There is a rare instance where sizing of fonts could be different (developer has changed ControlContentThemeFontSize manually and then not used font sizes).

### Testing Procedure ###
Force the creation of the following object off the UI thread:
```fsharp
Label(FontSize = 10.)
```
This should not give an exception.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.UAP/FontExtensions.cs
+++ b/Xamarin.Forms.Platform.UAP/FontExtensions.cs
@@ -56,11 +56,7 @@ namespace Xamarin.Forms.Platform.UWP
 			switch (size)
 			{
 				case NamedSize.Default:
-					if(DefaultFontSize == double.NegativeInfinity)
-					{
-						DefaultFontSize = (double)WApplication.Current.Resources["ControlContentThemeFontSize"];
-					}
-					return DefaultFontSize;
+					return 14.667;
 				case NamedSize.Micro:
 					return 15.667;
 				case NamedSize.Small:
--- a/Xamarin.Forms.Platform.UAP/FontExtensions.cs
+++ b/Xamarin.Forms.Platform.UAP/FontExtensions.cs
@@ -15,7 +15,6 @@ namespace Xamarin.Forms.Platform.UWP
 	{
 		[ThreadStatic]
 		static Dictionary<string, FontFamily> FontFamilies = new Dictionary<string, FontFamily>();
-		static double DefaultFontSize = double.NegativeInfinity;
 
 		public static void ApplyFont(this Control self, Font font)
 		{


charlesroddie:
Review please?

jfversluis:
Now that we're so close to the sunsetting of Xamarin.Forms unfortunately we won't be able to take this in anymore, we're really sorry about that. Nevertheless, thank you so much for your time and effort that you have put into this PR.

Please have a look at the evolution of Xamarin.Forms, [.NET MAUI](https://github.com/dotnet/maui). A lot of development has been going on there. Hopefully this issue was already fixed in that codebase. If not, feel free to port this over to there.

Again, thank you so much for being a contributor and Xamarin.Forms user!

PR #13884, opened 2021-02-27
Font AppHostBuilder Extension

Redth:
### Description of Change ###

Single Project will process `SharedFont` MSBuild items into native platform font assets.  On android this means a font in the assets folder, on iOS it adds fonts as BundleResource's and generates the Info.plist font registrations.  In both cases it's possible to load the font by filename alone, using a bit of platform specific code.

This adds the platform specific code for loading the fonts by filename, and loads them into the FontRegistrar for use.  This Is a bit of a hack in that it still uses the EmbeddedFontLoader service and we should eventually clean up that code so this service is more suitably named, and the FontRegistrar has fewer code paths that we don't intend to use.  For now this gets us Single Project font supports out to preview, and we can evaluate fonts a bit more based on what's in Microsoft.Maui.Graphics too.

This adds also some default inclusion patterns for font files in the single project targets, and a font to the single project sample which should be included properly as a native font.

Eventually we may also decide to use source code generators to create the font registrar invocation automatically based on the existence of the msbuild SharedFont items.

### API Changes ###

Added:

```csharp
public static class AppHostBuilderExtensions
{
	public static IAppHostBuilder RegisterFont(this IAppHostBuilder builder, string filename, string alias = null)
	{
		// ...
	}
}
```

### Platforms Affected ### 

- Core (AppHostBuilder)
- iOS
- Android

### PR Checklist ###


src/Controls/src/Core/FontRegistrar.cs
```diff
@@ -19,6 +19,52 @@ public static void Register(ExportFontAttribute fontAttribute, Assembly assembly
...
+			if (!File.Exists(fontBundlePath))
+				return false;
+
+			using (var fontStream = File.OpenRead(fontBundlePath))
+			{
+				return TryLoadFont(cacheKey, fontStream);
+			}
+#elif __ANDROID__ || ANDROID
+			var assets = Android.App.Application.Context.Assets;
+			using (var fontStream = assets.Open(fontName))
```
mattleibow, line 60: We probably need to do the paths better. On iOS we got a few locations, Android is just at the root. I personally will always use a subfolder. 

Maybe we need to allow for some property when resizing and here to provide a sub-path.

src/Controls/src/Core/FontRegistrar.cs
```diff
@@ -43,7 +89,11 @@ public static (bool hasFont, string fontPath) HasFont(string font)
 			}
 			catch (Exception ex)
 			{
+#if !__ANDROID__ && !__IOS__
```
mattleibow, line 92: Is this missing the IOS and ANDROID?

rmarinho:
Hello @Redth  please move this PR to https://github.com/dotnet/maui

PR #13888, opened 2021-02-28
[Handlers] Fonts and Label FontFamily, FontAttributes & FontSize

mattleibow:
### Description of Change ###

Changes in this PR were to add support for fonts and font styles for Label, but required a few more things to be pulled down. This also includes _some_ of the work by @Redth in #13884

However, the final product is quite nice as registering a font that was added to the app (not the embedded ones, but the resizetizer ones) is just this:

```csharp
builder.ConfigureFonts((hostingContext, fonts) => {
    fonts.AddFont("dokdo_regular.ttf", "Dokdo");
});
```

In the end, this PR has a few things:

* The small part that actually sets the `Font`, `FontSize` and `FontAttributes` in the handlers  
  There is a current limitation in the system that might be resolved by #13878. There is no correct way yet to access the service provider, so this PR just pulls it from the `MauiApp.Current.Context` in the Android and iOS `LabelHandler`. After #13878 is merged, we should be able to get the context or provider directly from the view we are mapping.
* Dodgy, but functional support for singletons in the `MauiServiceProvider`  
  This is currently just keeping a dictionary with all the singleton instances that get created. There is no disposal process and there is no support for the scopes.
* In order to get the singleton support, I had to rework the `MauiServiceCollection` to not actually set up the instantiation, but moved it to `MauiServiceProvider` where the lifetime of the descriptor could be read. In addition, I had to remove the dictionary that the collection was using so I could get the entire descriptor to read other props. 
* In order to support fonts, I added `IAppHostBuilder ConfigureFonts(Action<HostBuilderContext, IFontCollection> configureDelegate);` to the `IAppHostBuilder` to manage the font collection until the app is built.  
  This new collection may be fine, but we may also want to maybe make this a bit more generic so that we don't have to keep adding things to the `IAppHostBuilder` every time we need to get a new resource. But, then again, this may not actually be too bad since not many things need registration and a service could be used.
* Moved the `FontRegistrar` down to core and created a `IFontRegistrar` that allows for registration of embedded and app package fonts. 
* In order to separate the font creating management and font use management, there is a `IFontManager` that will be used to go from Maui fonts into platform fonts. 
* Moved the concept of `IEmbeddedFontLoader` into Core so that fonts can be persisted in temporary storage. This is a poorly named object as this really is something that takes a stream and saves it on the device, and then registers the font with the OS. However, I kept the name for backwards compat with Forms. However, we really might want to rename to `IFontInstaller` or something. There is also the related `EmbeddedFont` type that literally is a filename and a stream. The installer saves the stream to disk using the filename...

Bugs fixed:

* A bug in resizetizer that would crash when no fonts/images were provided.
* Make sure to run `Forms.Init()` in the Controls sample so that we don't throw when `Device.Xxx` is not initialized. This is currently used in the existing controls to get default font sizes and things.
* Renamed the views in the Android XML layouts to use the new namespaces.
* Fixed the `[XmlnsDefinition]` attributes after the rename and split. Things like `Color` that were moved to Core could no longer be found.

Extras in here too:

* A build tasks solution file that will allow all the tasks to be pre-built before trying to use them.
* Fixed for the yaml to actually build and run the tests after the big rename.
* Added a compatibility sample to ensure the XAML still works and the existing `[EmbeddedFont]` attributes.

### Issues Resolved ### 

- fixes #

### API Changes ###

 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### PR Checklist ###


programmation:
Can I enter a plea to remove `internal` from this code please. There are many situations when we will need to use Font logic in our own controls, and having this `internal` means we will need to duplicate all of this code in our code in order to conform with Xamarin Forms' font handling. Thanks.

src/Compatibility/Core/src/iOS/Extensions/FontExtensions.cs
```diff
@@ -1,120 +1,17 @@
...
-using System.Linq;
-using Microsoft.Maui.Controls.Compatibility.Internals;
-using UIKit;
-using System;
 using Microsoft.Maui.Controls.Internals;
+using UIKit;
 
 namespace Microsoft.Maui.Controls.Compatibility.Platform.iOS
 {
 	public static partial class FontExtensions
```
mattleibow, line 6: These classes are basically reduce to being extension methods that call into the font manager.

src/Compatibility/Core/src/iOS/Forms.cs
```diff
@@ -33,6 +31,11 @@ public static class Forms
 	{
 		public static bool IsInitialized { get; private set; }
 
+		static IFontManager s_fontManager;
+
+		internal static IFontManager FontManager =>
```
mattleibow, line 36: For the time being, instead of storing the fonts in a static dictionary in `FontRegistrar` I split the functionality into the `FontRegistrar` that is the part that manages a list of fonts and `FontManager` that manages the font lifetime. Both are instances and can be obtained via the service provider in Core. However for forms, everything expects a static object. So to fix that, I added a static property to Forms, which is just for compatibility.

This might be wrong as the provider will return a different instance to this property as it is new-ed up here. Not sure what the correct way to do this is.

src/Controls/samples/Controls.Sample.Droid/MainActivity.cs
```diff
@@ -10,7 +10,10 @@ public class MainActivity : MauiAppCompatActivity
 	{
 		protected override void OnCreate(Bundle savedInstanceState)
 		{
+			Microsoft.Maui.Controls.Compatibility.Forms.Init(this, savedInstanceState);
```
mattleibow, line 13: Even though this is still compat, it is actually needed because fonts require the use of some members of `Device` which are not yet available in core. We can probably remove this in a new PR and do it right-er.

src/Controls/src/Core/Registrar.cs
```diff
@@ -271,6 +271,8 @@ static Registrar()
 			Registered = new Registrar<IRegisterable>();
 		}
 
+		public static IFontRegistrar FontRegistrar { get; } = new FontRegistrar();
```
mattleibow, line 274: This again is not using the service provider (like the `FontManager` on `Forms`) because we need this ready _before_ the registrar runs. This is used by the `RegisterAll` method, which is the initial entry point to register the things. Not sure what the correct way to use the service provider in a static context.

src/Controls/src/Core/Registrar.cs
```diff
@@ -412,6 +414,13 @@ public static void RegisterAll(Type[] attrTypes, InitializationFlags flags)
 				Profile.FrameEnd(frameName);
 			}
 
+			if (FontRegistrar is FontRegistrar fontRegistrar)
+			{
+				var type = Registered.GetHandlerType(typeof(EmbeddedFont));
+				if (type != null)
+					fontRegistrar.SetFontLoader((IEmbeddedFontLoader)Activator.CreateInstance(type));
+			}
```
mattleibow, line 422: This is interesting. Once I have registered all the fonts and all the "handlers", I then need to let the registrar know what the handler is to use when installing the fonts.

src/Core/src/Fonts/EmbeddedFont.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+using System.IO;
+
+namespace Microsoft.Maui
+{
+	public class EmbeddedFont
```
mattleibow, line 6: This is a bad name as this is passed to an "OS font installer" concept. This is really a filename to use when saving this stream to disk.

src/Core/src/Fonts/FontManager.Android.cs
```diff
@@ -0,0 +1,192 @@
...
+	public class FontManager : IFontManager
+	{
+		readonly ConcurrentDictionary<(string fontFamily, FontAttributes fontAttributes), Typeface?> _typefaces =
+			new ConcurrentDictionary<(string fontFamily, FontAttributes fontAttributes), Typeface?>();
+
+		readonly IFontRegistrar _fontRegistrar;
+
+		Typeface? _defaultTypeface;
+
+		public FontManager(IFontRegistrar fontRegistrar)
```
mattleibow, line 18: The font manager requires the font registrar because it looks up the font in the registrar, and then opens and caches it here.

src/Core/src/Fonts/FontRegistrar.cs
```diff
@@ -0,0 +1,146 @@
...
+
+		readonly Dictionary<string, (string Filename, string? Alias)> _nativeFonts =
+			new Dictionary<string, (string Filename, string? Alias)>();
+
+		readonly Dictionary<string, (bool Success, string? Path)> _fontLookupCache =
+			new Dictionary<string, (bool Success, string? Path)>();
+
+		IEmbeddedFontLoader? _fontLoader;
+
+		public FontRegistrar(IEmbeddedFontLoader? fontLoader = null)
```
mattleibow, line 23: The font registrar requires the font loader (installer) because before any fonts can actually be used, it needs to be installed. So, we can't return a font that is not yet installed.

src/Core/src/Handlers/Label/LabelHandler.Android.cs
```diff
@@ -30,5 +32,28 @@ public static void MapTextColor(LabelHandler handler, ILabel label)
...
+
+		public static void MapFontAttributes(LabelHandler handler, ILabel label)
+		{
+			MapFont(handler, label);
+		}
+
+		static void MapFont(LabelHandler handler, ILabel label)
+		{
+			var context = MauiApp.Current?.Context ?? throw new InvalidOperationException($"The MauiApp.Current.Context can't be null.");
+			var fontManager = context.Services.GetRequiredService<IFontManager>();
```
mattleibow, line 54: This is sus, but for now it is the only way to get the `IMauiContext`. I am doing it in the handler to reduce the junk in the extension methods and other places.

src/Core/src/Hosting/FontCollectionExtensions.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Microsoft.Maui.Hosting
+{
+	public static class FontCollectionExtensions
+	{
+		public static IFontCollection AddFont(this IFontCollection fontCollection, string filename, string? alias = null)
```
mattleibow, line 5: Just one for now, but we could add support for adding embedded fonts and maybe some other things.

src/Core/src/Hosting/IMauiServiceCollection.cs
```diff
@@ -1,11 +1,10 @@
...
-using System.Collections.Generic;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.Maui.Hosting
 {
-	public interface IMauiServiceCollection
-		: IServiceCollection, IDictionary<Type, Func<IServiceProvider, object?>?>
+	public interface IMauiServiceCollection : IServiceCollection
 	{
+		bool TryGetService(Type serviceType, out ServiceDescriptor? descriptor);
```
mattleibow, line 8: The `ServiceDescriptor` is needed to actually learn if the service is a single ton or transient. No support yet for scoped.

src/Core/src/Hosting/MauiServiceProvider.cs
```diff
@@ -1,16 +1,21 @@
...
 namespace Microsoft.Maui.Hosting
 {
-
 	class MauiServiceProvider : IMauiServiceProvider
 	{
-		IDictionary<Type, Func<IServiceProvider, object?>?> _collection;
+		IMauiServiceCollection _collection;
+
+		// TODO: do this properly and support scopes
+		IDictionary<ServiceDescriptor, object?> _singletons;
```
mattleibow, line 13: No lifetime management and might leak. I see the ASP.NET folks have a whole system in place, however since this is an embedded mobile app - things might not be too bad for now. But, looking at the ASP.NET system, there are more complex things to manage scopes and all of that. Once that gets done, this can go away for a real implementation.

src/Core/src/Platform/Android/LabelExtensions.cs
```diff
@@ -20,7 +21,18 @@ public static void UpdateTextColor(this TextView textView, ILabel label,  Maui.C
 			else
 			{
 				textView.SetTextColor(textColor.ToNative());
-			}				
+			}
+		}
+
+		public static void UpdateFont(this TextView textView, ILabel label, IFontManager fontManager)
```
mattleibow, line 27: I decided to pass in the required font manager here to keep things clean down here and let the handlers be a more interop-y thing.

mattleibow:
Need to add tests...

Redth (approved):

src/Core/src/Hosting/MauiServiceCollection.cs
```diff
@@ -1,110 +1,76 @@
...
-			if (item.ImplementationType != null)
-				_factories[item.ServiceType] = new Func<IServiceProvider, object?>(provider => Activator.CreateInstance(item.ImplementationType));
-			else if (item.ImplementationInstance != null)
-				_factories[item.ServiceType] = new Func<IServiceProvider, object?>(provider => item.ImplementationInstance);
-			else if (item.ImplementationFactory != null)
-				_factories[item.ServiceType] = item.ImplementationFactory;
-			else
-				throw new InvalidOperationException($"You need to provide an {nameof(item.ImplementationType)} or a {nameof(item.ImplementationInstance)} ");
+			if (_descriptors.Contains(item))
+				return;
```
rmarinho, line 26: I m not sure if we should throw or log something here to say to the user we don't support adding several implementations of the same type.

mattleibow: Should we throw or maybe override existing? So "last one wins"

rmarinho: we need to override, lets say Maui register ButtonHandler for IButton, and the user wants to register a CustomButtonHandler for IButton. the last one should win.

mattleibow: It actually is already the last one wins. if that exact instance of the descriptor is in already, it skips. If not, it overrides the service in the dictionary.

In actual fact, the descriptors list could have double registrations, but the dictionary always is the last one.

rmarinho (approved):

PR #13908, opened 2021-03-02
Break up legacy and Maui arrange code

PureWeen:
### Description of Change ###

The call to arrange on Layout to make sure that new style handlers get arranged wasn't accounting for the fact that that arrange code is going to add an additional margin. This code just splits up the old Layout call and the new Arrange call so they don't call each other and can just act predictably according to its scenario.

### Testing Procedure ###
- make sure the unit tests all pass

### PR Checklist ###

### DIFF
--- a/src/Controls/src/Core/HandlerImpl/VisualElement.Impl.cs
+++ b/src/Controls/src/Core/HandlerImpl/VisualElement.Impl.cs
@@ -56,7 +56,20 @@ namespace Microsoft.Maui.Controls
 
 			var newRect = this.ComputeFrame(bounds);
 
-			Layout(newRect);
+			Bounds = newRect;
+			Handler?.SetFrame(Bounds);
+		}
+
+		// TODO: MAUI
+		// This is here to support layout calls from legacy code
+		// This should go away once we get everything piping through
+		// Maui based layout code
+		public void Layout(Rectangle bounds)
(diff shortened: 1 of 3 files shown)


src/Controls/src/Core/HandlerImpl/VisualElement.Impl.cs
```diff
@@ -56,7 +56,20 @@ protected virtual void ArrangeOverride(Rectangle bounds)
...
-			Layout(newRect);
+			Bounds = newRect;
+			Handler?.SetFrame(Bounds);
+		}
+
+		// TODO: MAUI
+		// This is here to support layout calls from legacy code
+		// This should go away once we get everything piping through
+		// Maui based layout code
+		public void Layout(Rectangle bounds)
```
PureWeen, line 67: I moved this code here so that it's closer to the Arrange call that will eventually replace it

src/Controls/src/Core/StackLayout.cs
```diff
@@ -69,17 +69,6 @@ protected override void LayoutChildren(double x, double y, double width, double
 			}
 		}
 
-		// IFrameworkElement Measure
```
PureWeen, line 72: This code is no longer needed. This exact method was already moved up to the base class

PR #13919, opened 2021-03-03
Additional changes to propagate measures and arranges between layouts

PureWeen:
### Description of Change ###

- No longer sets IsArrangeValid from Layout. If Layout is called by legacy code that shouldn't stop the Arrange call
- Override GetSizeRequest on Layout2 to forward calls to measure because even though GetSizeRequest has been obsolete forever it's still called from different parts of the code
- Added tests for different arrangement scenarios of mixing Layout2 layouts and Controls layouts 

### Testing Procedure ###
- unit tests pass
- test different variations of layouts/controls in the sample appp

### PR Checklist ###

### DIFF
--- a/src/Controls/src/Core/HandlerImpl/VisualElement.Impl.cs
+++ b/src/Controls/src/Core/HandlerImpl/VisualElement.Impl.cs
@@ -67,9 +67,6 @@ namespace Microsoft.Maui.Controls
 		public void Layout(Rectangle bounds)
 		{
 			Bounds = bounds;
-			// If Layout is called without arrange getting called this ensures
-			// all the necessary arranged parts are set so that handlers will layout			
-			
 			Handler?.SetFrame(Bounds);
 		}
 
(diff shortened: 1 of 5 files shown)


src/Controls/src/Core/HandlerImpl/VisualElement.Impl.cs
```diff
@@ -69,7 +69,7 @@ public void Layout(Rectangle bounds)
 			Bounds = bounds;
 			// If Layout is called without arrange getting called this ensures
 			// all the necessary arranged parts are set so that handlers will layout
```
mattleibow, line 71: Is this comment still valid?

src/Controls/src/Core/Layout.cs
```diff
@@ -64,7 +64,7 @@ public void Remove(IView child)
 		}
 	}
 
-	public abstract class Layout : View, ILayout, ILayoutController, IPaddingElement
+	public abstract class Layout : View, ILayout, ILayoutController, IPaddingElement, IFrameworkElement
```
mattleibow, line 67: Should ILayout be an IFrameworkElement?

PureWeen: It is but I had to add the interface here so that I could add the explicit implementation of measure lower down in the code

src/Controls/src/Core/HandlerImpl/VisualElement.Impl.cs
PureWeen: Removed

mattleibow (approved):
CI is the final gate, but looks good!

PR #14019, opened 2021-03-16
[C] add source info to binding failure logs

### LINKED ISSUE #14006: [Enhancement] Include source info for binding failure debug output
## Summary

Debug output for binding failures currently looks like "Binding: Message", but it doesn't say where the binding failure came from in a source XAML file. If the binding failure has source info, it should be included in the debug output.

## API Changes

n/a

## Intended Use Case

Visual Studio's Hot Reload feature uses the BindingFailed event to know about binding failures and their source location. But... when the app first starts, before Hot Reload is initialized, binding failures are only available to VS in the debug output window. Those binding failures don't have source info, so double clicking on them in the binding failures pane in VS does nothing.

The selected row happens on boot and has no source info:

image

The debug output currently looks like this:

```
[0:] Binding: 'FooNavigateCommand' property not found on 'DataBindingDemos.MainPage', target property: 'Xamarin.Forms.TextCell.Command'
[0:] Binding: 'FooRed' property not found on 'DataBindingDemos.RgbColorViewModel', target property: 'Xamarin.Forms.Slider.Value'
[0:] Binding: 'FooGreen' property not found on 'DataBindingDemos.RgbColorViewModel', target property: 'Xamarin.Forms.Slider.Value'
```

To include source info and not break existing VS debug output parsers, source info should be added on a new line:

```
[0:] Binding: 'FooNavigateCommand' property not found on 'DataBindingDemos.MainPage', target property: 'Xamarin.Forms.TextCell.Command'
Binding source: line=1, column=10, file="path in quotes.xaml"
[0:] Binding: 'FooRed' property not found on 'DataBindingDemos.RgbColorViewModel', target property: 'Xamarin.Forms.Slider.Value'
Binding source: line=2, column=15, file="path in quotes.xaml"
[0:] Binding: 'FooGreen' property not found on 'DataBindingDemos.RgbColorViewModel', target property: 'Xamarin.Forms.Slider.Value'
Binding source: line=3, column=20, file="path in quotes.xaml"
```

The line, column, and file values come directly from the internal SourceInfo object. Both lines need to be written with a single debug output call so that there is no chance for other debug output to appear between them. The line prefix of "Binding source:" doesn't matter much, just need something that can be recognized in a regular expression.

StephaneDelcroix:
- fixes #14006

### DIFF
--- a/Xamarin.Forms.Core/Xaml/Diagnostics/BindingDiagnostics.cs
+++ b/Xamarin.Forms.Core/Xaml/Diagnostics/BindingDiagnostics.cs
@@ -12,14 +12,28 @@ namespace Xamarin.Forms.Xaml.Diagnostics
 
 		internal static void SendBindingFailure(BindingBase binding, string errorCode, string message, params object[] messageArgs)
 		{
-			Log.Warning(errorCode, message, messageArgs);
-			BindingFailed?.Invoke(null, new BindingBaseErrorEventArgs(VisualDiagnostics.GetXamlSourceInfo(binding), binding, errorCode, message, messageArgs));
+			var sourceinfo = VisualDiagnostics.GetXamlSourceInfo(binding);
+			if (sourceinfo != null)
+			{
+				var sourceinfoline = $"\nBinding source: line={sourceinfo.LineNumber}, column={sourceinfo.LinePosition}, file=\"{sourceinfo.SourceUri}\"";
+				Log.Warning(errorCode, message + sourceinfoline, messageArgs);
+			}
+			else
+				Log.Warning(errorCode, message, messageArgs);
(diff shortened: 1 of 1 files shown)


jsuarezruiz (approved):

StephaneDelcroix:
do not merge yet. it's still unclear if it's required or not. we're waiting for @spadapet feedback.

spadapet:
I commented in the issue #14006 about why I think this fix can be discarded. With everything updated, the issue doesn't repro when I debug the Android project. Also, when I built this PR's change and debugged it, there was no source info for the binding anyway, so the change didn't actually write out new debug output.

jfversluis:
Seems like this isn't needed anymore. Let us know if anything changes.

PR #14029, opened 2021-03-18
[macOS] Fix back button for nested nav pages

### LINKED ISSUE #8843: [Bug] macOS: no navigation bar (or any UI widgets to go back/forward) when pushing NavigationPages
### Description

When using navigation pages, other platforms (e.g. Android, iOS, GTK) have visual widgets to go back and forward, but macOS lacks them.

### Steps to Reproduce

1. Create a XamarinForms app with NavigationPages; e.g. I created a very simple one here: https://github.com/knocte/geewallet/tree/bugreport/noUIforNavigationInMac
2. Run the linked app above (note: checkout branch `bugreport/noUIforNavigationInMac`)
3. Click on the label to go one page forward.

### Actual Behavior
User cannot go back to previous page, as there's no UI widget to do it:
actualMacOs

In Android there is a "back arrow":
actualAndroid

In iOS too:
actualIOS

Even in GTK!:
actualGTK

### Expected Behavior
There should be a visual UI widget to go back to previous page.
An example of such widget you can find on the macOS preferences:
expectedMacOS

### Basic Information

- Version with issue: any! (currently running 4.3.x)
- Last known good version: None, AFAIK
- IDE: VS4Mac 8.3.10.2
- Affected Devices: running macOS catalina

### Screenshots

See above

### Reproduction Link

https://github.com/knocte/geewallet/tree/bugreport/noUIforNavigationInMac

aarani:
### Description of Change ###

In the current implementation of NativeToolbarTracker,
having a nested navigation page causes the toolbar to get
overwritten by the child navigation page which in turns
causes the back button to be unusable because it only moves
in the scope of the child navigation stack so to mitigate
the issue this commit adds the possiblity for the back button
to move out of the child navigation stack back into the parent
navigation stack.

### Issues Resolved ### 

- fixes #8843

### API Changes ###

 
 None

### Platforms Affected ### 

- macOS

Toolbar back button now can move out of a nested navigation page into the parent page.

Not applicable

### Testing Procedure ###

Build and run the test case repo provided by the author of the issue before and after this change and compare the toolbar behaviour after moving to the new page

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue8843.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue8843.cs
@@ -1,0 +1,78 @@
+﻿using System;
+using System.Threading.Tasks;
+using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+
+#if UITEST
+using Xamarin.Forms.Core.UITests;
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if UITEST
+	[Category(UITestCategories.Navigation)]
+#endif
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 8843, "[Bug] [macOS] Fix back button for nested nav pages",
+		PlatformAffected.macOS)]
+	public partial class Issue8843 : TestContentPage
+	{
+		public Issue8843()
+		{
+#if APP
+			InitializeComponent();
+#endif
+		}
+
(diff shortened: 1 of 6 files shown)


aarani:
I accidentally force-pushed this which caused it to close this is still ready for review

jsuarezruiz (changes requested):
Could you include a sample https://github.com/xamarin/Xamarin.Forms/tree/5.0.0/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared to validate the issue?

jfversluis:
Tried to revive this and see if we could make it work, unfortunately while testing I see some unpredictable behavior if I navigate back and forth more than once. Not going to merge this to avoid the risk on any regressions. Thanks for your time and effort on this!

PR #5237, opened 2019-02-13
[Android] TalkBack now reads name and helptext on buttons.

ryl:
### Description of Change ###

Modified `AutomationPropertiesProvider` such that `AutomationProperties.Name` and `AutomationProperties.HelpText` will be read by TalkBack.

### Issues Resolved ### 

- fixes #5150
- fixes #1529

### API Changes ###
 
 None

### Platforms Affected ### 

- Android

### Behavioral/Visual Changes ###

TalkBack will read `AutomationProperties.Name` and `AutomationProperties.HelpText` on a button if provided.

### Before/After Screenshots ### 

Not applicable

### Testing Procedure ###

In the Control Gallery, visit G5150. Labels containing the ContentDescription for each button should appear with appropriate text.

### PR Checklist ###

- [x] Has automated tests 
- [x] Rebased on top of the target branch at time of PR
- [x] Changes adhere to coding standard

PR #14089, opened 2021-03-31
Fix up Android Accessibility behind a flag

PureWeen:
### Description of Change ###
- added Xamarin.AndroidX.Navigation.UI so I can use the resource id for reading the back button as "Navigate up"
- added AccessibilityDelegateAutomationId so that if you set an AutomationId and the HelpText it'll read back the HelpText instead of the AutomationID
- When setting AutomationId on controls that aren't implicitly accessible we need to turn "ImportantForAccessibility" off so that TalkBack doesn't start reading it
- Disable extracting the defaults for hint and CD
- Remove the "Touch" settings on the PlatformRenderer
- Remove Clickable on the PageRenderer
- Fix Shell so that it doesn't read out the ViewPager if the page has no top tabs

### API Changes ###
Added:
 - Accessibility_Experimental Flag so that you can toggle these changes on and off. 

### Platforms Affected ### 
- Android

### Testing Procedure ###
- Turn this flag on and then click around in Control Gallery on the various TalkBack scenarios for Android. You'll see a flag you can uncomment inside FormsAppCompatActivity. The AutomationId page is a pretty good one to test. Before the flag it reads the AutomationId but then after the flag it continues to read the text

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.Android/AccessibilityDelegateAutomationId.cs
+++ b/Xamarin.Forms.Platform.Android/AccessibilityDelegateAutomationId.cs
@@ -3,12 +3,14 @@ using System.Collections.Generic;
 using System.Text;
 using Android.Views.Accessibility;
 using AndroidX.AppCompat.Widget;
+using AndroidX.Core.View;
+using AndroidX.Core.View.Accessibiity;
 using Xamarin.Forms.Platform.Android.FastRenderers;
-using AAccessibilityDelegate = Android.Views.View.AccessibilityDelegate;
 
 namespace Xamarin.Forms.Platform.Android
 {
-	class AccessibilityDelegateAutomationId : AAccessibilityDelegate
+	class AccessibilityDelegateAutomationId : AccessibilityDelegateCompat
+
 	{
 		BindableObject _element;
 
@@ -18,7 +20,7 @@ namespace Xamarin.Forms.Platform.Android
 		}
 
 
-		public override void OnInitializeAccessibilityNodeInfo(global::Android.Views.View host, AccessibilityNodeInfo info)
+		public override void OnInitializeAccessibilityNodeInfo(global::Android.Views.View host, AccessibilityNodeInfoCompat info)
 		{
 			base.OnInitializeAccessibilityNodeInfo(host, info);
 
--- a/Xamarin.Forms.Platform.Android/AccessibilityDelegateAutomationId.cs
+++ b/Xamarin.Forms.Platform.Android/AccessibilityDelegateAutomationId.cs
@@ -1,0 +1,49 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Android.Views.Accessibility;
+using AndroidX.AppCompat.Widget;
+using Xamarin.Forms.Platform.Android.FastRenderers;
+using AAccessibilityDelegate = Android.Views.View.AccessibilityDelegate;
+
+namespace Xamarin.Forms.Platform.Android
+{
+	class AccessibilityDelegateAutomationId : AAccessibilityDelegate
+	{
+		BindableObject _element;
+
+		public AccessibilityDelegateAutomationId(BindableObject element) : base()
+		{
+			_element = element;
+		}
+
+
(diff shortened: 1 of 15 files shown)


Xamarin.Forms.Platform.Android/FastRenderers/AutomationPropertiesProvider.cs
```diff
@@ -30,52 +31,96 @@ internal static void GetDrawerAccessibilityResources(global::Android.Content.Con
...
+
+			// User has specifically said what they want
+			if (AutomationProperties.GetIsInAccessibleTree(element) == true)
+				return false;
+
+			// User has explicitly set name and help text so we honor that
+			if (!String.IsNullOrWhiteSpace(ConcatenateNameAndHelpText(element)))
+				return false;
+
+			if (element is Layout ||
```
PureWeen, line 47: This set of controls isn't going to change

This is a bug that's very specific to XF 5 and won't surface in MAUI as we won't be using ContentDescription

Xamarin.Forms.Platform.Android/AccessibilityDelegateAutomationId.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Android.Views.Accessibility;
+using AndroidX.AppCompat.Widget;
+using Xamarin.Forms.Platform.Android.FastRenderers;
+using AAccessibilityDelegate = Android.Views.View.AccessibilityDelegate;
+
+namespace Xamarin.Forms.Platform.Android
+{
+	class AccessibilityDelegateAutomationId : AAccessibilityDelegate
```
PureWeen, line 11: This is very similar to what @ryl created here

https://github.com/xamarin/Xamarin.Forms/pull/5237/files#diff-d62639210acd753c36ed2128b141cb10e5bc1b4f5f1d1d95593823572f0c08f3R1

Xamarin.Forms.Platform.Android/FastRenderers/AutomationPropertiesProvider.cs
```diff
@@ -30,52 +32,102 @@ internal static void GetDrawerAccessibilityResources(global::Android.Content.Con
 			resourceIdClose = context.Resources.GetIdentifier($"{automationIdParent}{s_defaultDrawerIdCloseSuffix}", "string", context.ApplicationInfo.PackageName);
 		}
 
+		static bool ShoudISetImportantForAccessibilityToNoIfAutomationIdIsSet(AView control, Element element)
```
rachelkang, line 35: Can we come up with a better, shorter name for this bool?

Xamarin.Forms.Platform.Android/Renderers/ShellToolbarTracker.cs
```diff
@@ -451,7 +451,10 @@ protected virtual void UpdateToolbarIconAccessibilityText(Toolbar toolbar, Shell
 			else if (image == null ||
 				toolbar.SetNavigationContentDescription(image) == null)
 			{
-				toolbar.SetNavigationContentDescription(R.String.Ok);
+				if(CanNavigateBack)
+					toolbar.SetNavigationContentDescription(Resource.String.nav_app_bar_navigate_up_description);
+				else
+					toolbar.SetNavigationContentDescription(R.String.Ok);
```
rachelkang, line 457: should we still set it to ok? maybe we should just not set it to anything? then maybe it'll read like the native default that just reads as "unlabeled button"

rachelkang (approved):

Xamarin.Forms.Platform.Android/Renderers/ShellToolbarTracker.cs
DDSoftDennie: Is there an option for a "null" or "default"?  Because if I read the code it's not really OK....

PureWeen: > Because if I read the code it's not really OK....

What do you mean "read the code?"

Issue #13910, opened 2021-03-02
[Bug] [iOS] Clearing ObservableCollection causes NullReferenceException


Janek91:
### Description
Clear of ObservableCollection with some items binded to CollectionView causes NullReferenceException at: https://github.com/xamarin/Xamarin.Forms/blob/a10c2c6218e592bc6896f0fcbd51ae34e1f4d9dc/Xamarin.Forms.Platform.iOS/CollectionView/ObservableItemsSource.cs#L148
```
System.NullReferenceException: Object reference not set to an instance of an object
  at Xamarin.Forms.Platform.iOS.ObservableItemsSource.Reload () [0x00025] in D:\a\1\s\Xamarin.Forms.Platform.iOS\CollectionView\ObservableItemsSource.cs:148 
  at Xamarin.Forms.Platform.iOS.ObservableItemsSource.CollectionChanged (System.Collections.Specialized.NotifyCollectionChangedEventArgs args) [0x0005a] in D:\a\1\s\Xamarin.Forms.Platform.iOS\CollectionView\ObservableItemsSource.cs:132 
  at Xamarin.Forms.Platform.iOS.ObservableItemsSource.CollectionChanged (System.Object sender, System.Collections.Specialized.NotifyCollectionChangedEventArgs args) [0x0002d] in D:\a\1\s\Xamarin.Forms.Platform.iOS\CollectionView\ObservableItemsSource.cs:108 
  at System.Collections.ObjectModel.ObservableCollection`1[T].OnCollectionChanged (System.Collections.Specialized.NotifyCollectionChangedEventArgs e) [0x00018] in /Library/Frameworks/Xamarin.iOS.
framework/Versions/Current/src/Xamarin.iOS/external/corefx/src/System.ObjectModel/src/System/Collections/ObjectModel/ObservableCollection.cs:263 
  at MvvmCross.ViewModels.MvxObservableCollection`1[T].<>n__0 (System.Collections.Specialized.NotifyCollectionChangedEventArgs e) <0x110541000 + 0x0005a> in <03f32c17c8e741219cee9c3d186bfa64>:0 
  at MvvmCross.ViewModels.MvxObservableCollection`1+<>c__DisplayClass7_0[T].<OnCollectionChanged>b__0 () [0x00000] in /_/MvvmCross/ViewModels/MvxObservableCollection.cs:72 
[...]

PR #14076, opened 2021-03-29
Fixed the memory leak in the collectionview and gallery

jtorvald:
### Description of Change ###
Event handlers on the TemplatedCell are not being removed which results in the control to be left in memory which keeps the page where it is loaded also to stay behind in memory.

I added an unbind method on the TemplatedCell and unbind the event handlers in the ItemsViewController.
Also the TemplatedItemsList had a collectionchanged event which was not removed so I removed that as well.

If there is a better way to achieve this, please let me know and I will fix it in a better way.

Xamarin.Forms.Core/TemplatedItemsList.cs
Xamarin.Forms.Platform.iOS/CollectionView/ItemsViewController.cs
Xamarin.Forms.Platform.iOS/CollectionView/TemplatedCell.cs

All the other files are eventhandler fixes in the Gallery code.

### Issues Resolved ### 

- fixes #9787

### API Changes ###
Added: Unbind() on TemplatedCell.cs to remove events

### Platforms Affected ### 
- iOS
- Android

### Behavioral/Visual Changes ###

None

### Before/After Screenshots ### 

Not applicable

### Testing Procedure ###

Go to the templated collectionview pages in the gallery and go back. Repeat couple of times and see page count increasing in the Xamarin Profiler.

### PR Checklist ###

- [ ] Targets the correct branch
- [ ] Tests are passing (or failures are unrelated)

PR #14111, opened 2021-04-05
fix memory leak (and some crashes) in ItemsViewController (iOS)

tmijieux:
### Description of Change ###

When a CollectionView is removed from a layout or removed from view hierarchy (My use case is the following: when the user rotate the screen the layout changes. So, in my code, I recreate a new layout with a new CollectionView), the BindingContext is set to null, hence the ItemsSourceProperty is set to null and the ItemsSource property in the  Xamarin.Forms.Platform.iOS.ItemsViewController will be updated (through ItemsViewRenderer.OnElementPropertyChanged).

The problem is, if the IItemsViewSource is an Xamarin.Forms.Platform.iOS.ObservableItemsSource , the dispose method on the old ItemsSource value is not called and that dispose method is supposed to unsubscribe from the event delegate.
In the end, that object is still subscribed on the ObservableCollection in my ViewModel that has not reached the end of its lifetime(I continue using the same viewmodel), hence the ObservableItemSource is leaked because still referenced by the CollectionChanged event delegate, but the ItemsViewController which is supposed to own the reference to that ObservableItemsSource has no reference anymore.

The change in this PR is:
- Disposing pre-existing ObservableItemsSource(if any) when setting ItemsSource property to a new value (in ItemsViewController)

### Issues Resolved ### 
I did not yet do a complete check of what issues could be affected by this change.
One of the crash i encountered myself was exactly the same as in #13910 (same stacktrace) (when on INotifyCollectionChanged collection call the CollectionChanged event with the Reset event).

The reason this crash was happening is (i think) because the ObservableItemsSource, or the referenced views (if any) are removed from view hierarchy and they are not in a valid state anymore.

I will work on creating an issue with a minimal reproduction for my problem as soon as this PR is submitted.
EDIT: -> #14113 

### API Changes ###
None

### Platforms Affected ### 
- iOS

None

Not applicable

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS.UnitTests/ItemsViewControllerLeakTest.cs
+++ b/Xamarin.Forms.Platform.iOS.UnitTests/ItemsViewControllerLeakTest.cs
@@ -1,0 +1,140 @@
+﻿using System.Collections.Generic;
+using System.Collections.Specialized;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+using NUnit.Framework;
+
+namespace Xamarin.Forms.Platform.iOS.UnitTests
+{
+
+	class MyTestAdapter : List<int>, INotifyCollectionChanged
+	{
+		public event NotifyCollectionChangedEventHandler CollectionChanged;
+
+		public int NumberOfListener { get => CollectionChanged?.GetInvocationList().Length ?? 0; }
+		public bool ExceptionHappened { get; private set; } = false;
+
+		public new void Add(int item)
+		{
+			base.Add(item);
+			OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Add, item));
+		}
+
+		public new void Clear()
+		{
+			base.Clear();
+			OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
+		}
+
+		private void OnCollectionChanged(NotifyCollectionChangedEventArgs args)
+		{
+			if (CollectionChanged != null)
+			{
+				try
+				{
+					CollectionChanged.Invoke(this, args);
+				}
+				catch (System.Exception)
+				{
+					ExceptionHappened = true;
+				}
+			}
+		}
+	}
+
+	class MyTestViewModel : System.ComponentModel.INotifyPropertyChanged
+	{
+		private MyTestAdapter _myTestAdapter;
+		public MyTestAdapter MyTestAdapter
+		{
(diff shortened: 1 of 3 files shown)


jtorvald:
Same issue with multiple other leaky items was also fixed in #14076

rmarinho:
Is it possible to add a test case so we can make sure it fixes the issue ?

tmijieux:
@rmarinho do you mean a unit test ? I'm currently trying to add one. I figured it would land in the `Xamarin.Forms.Platform.iOS.UnitTest` namespace,...  I'm not yet sure how to run the tests that lies in this namespaces(currently updating stuff on my mac server to see if it is possible to run on device, but i'm not sure this is the way to go...)

If just a simple repro project is needed a already linked one with the associated issue:

https://github.com/xamarin/Xamarin.Forms/issues/14113
https://github.com/tmijieux/XFBugItemsViewController/

rmarinho:
I was saying a a sample page on the Issues gallery, and you can add a UITest

tmijieux:
i added two test,
-  the first one `DoesNotLeak`does not seem to work yet, (it pass both before and after the fix)   
- the second one `DoesNotCrash` seem to work as expected (fail before and pass after) .


I was about to look into my first test, but about at the same time i figured i would rebase on the lastest commit of xamarin/Xamarin.Forms:5.0.0 to see if things changed in between, but now i cannot deploy the ControlGallery to my iphone anymore (there is a build error that i can get rid of for now that have to do with a path length issue seemingly 😡 ) i'm not even sure that my deploy issue is related to the rebase or not, but i cant recall on what commit i was based before...

tmijieux:
Updated the tests, both tests now fail without  the fix and pass with the fix.

> I was saying a a sample page on the Issues gallery, and you can add a UITest

I'm not sure what you expect exactly, can you give a little more guidance please(is there a guideline I should read) ?

I put my tests in the unit tests because i just mimicked what was done in other unit tests to reproduce my bug, and the code in UITests seemed harder to grasp for me at first. But my feeling about the tests and the bugs is that they are not exactly trivial, so i understand if you want me to take them out of the "unit" tests.

jsuarezruiz:
The added test are passing.

Captura de pantalla 2021-07-07 a las 11 36 55

jsuarezruiz (approved):

rmarinho (approved):

rmarinho (changes requested):
I think a best fix is calling 

`ItemsSource?.Dispose();`

On the ` UpdateItemsSource()` method

rmarinho (approved):

PR #14137, opened 2021-04-09
WebView Navigation Now Uses Deferral Token 

### LINKED ISSUE #12720: [Bug] WebView Navigating Race Condition

Axemasta:
### Description of Change ###

`WebView`'s `Navigating` event does not allow for async code to be executed to determine whether to cancel the navigation. The implementation creates a race condition and makes the cancellation functionality confusing, the following code will not cancel the `WebView` navigation event:

```csharp
public MainPage()
{
    WebView.Navigating += async (s, e) =>
    {
        await Task.Delay(2000);

        e.Cancel = true;
    };
}
```

This makes it impossible to delay navigation in a `WebView` whilst doing a check such as seeing if the user can navigate to a url, without writing a hacky workaround ontop of this behaviour.

I make a first attempt at this PR in November and it was suggested by @PureWeen to refactor the code to use a DeferralToken, used by `Shell` to achieve the functionality (old pr: #12756). I've made a new PR since I made a completely new branch.

### Issues Resolved ### 

- fixes #12720 

### API Changes ###

Added:
 - Added `DeferralToken`, `DeferralTokenEventArgs`

Changed:
 - Renamed `ShellDeferralToken` => `DeferralToken`
 - Removed deferral token functionality from `ShellNavigatingArgs` into `DeferrableEventArgs`
 - `WebViewNavigatingArgs` now inherit from `DeferralTokenEventArgs`
 - `WKWebViewRenderer` now uses the deferral token to handle navigation decisions
 
New API Usage:
```csharp
WebView.Navigating += async (s, e) =>
{
    var token = e.GetDeferral();

    await Task.Delay(2000);

    var shouldCancel = CanBrowseUrl(e.Url);

    if (shouldCancel)
        e.Cancel();

    token.Complete();
};
```

### Platforms Affected ### 

- Core / All
- iOS Currently

This needs to be discussed but it would break and app using `WebViewNavigatingArgs.Cancel` to cancel navigation. I did slap an `Obsolete` attribute onto the property (and rename to `OldCancel` whilst the PR is work in progress, but it causes the control gallery to fail building. Maybe it can be added back when the PR is good to go & i go back through the control gallery and change all references to deferral.

Not applicable

### Testing Procedure ###

- Open the file `Xamarin.Forms.Controls.Issues.Shared.Issue12720.xaml.cs`
- You will find 3 event handlers in the constructor, for each event (only ever comment 1 in)
- Comment in the constructor, comment out the others
- Run the Control Gallery app
- Go to Test Cases
- Find issue G12720
- Use the web browser:

**OnNavigatingWithoutDeferral**
No impact to browser

**OnNavigatingWithDeferral**
User can't browse to any sites containing `bbc.co.uk` such as:
 - `https://www.bbc.co.uk/news/`

**OnNavigatingWithDeferralAsync**
All navigation occurs normally unless the user attempts to browse to any sites containing `bbc.co.uk` such as:
 - `https://www.bbc.co.uk/news/`

Requests to this site will be evaluated with a task that increases in duration, and always get cancelled

### Todo Before Merge ###

I've put this PR up to discuss the refactors I've made with the team before committing more time, this PR should be marked as do not merge until all work has been completed.

- [ ] Ensure Shell functioning correctly (affected by deferral refactor)
- [ ] Update Android WebViewRenderer
- [ ] Update GTK WebViewRenderer
- [ ] Update macOS WebViewRenderer
- [ ] Update Tizen WebViewRenderer
- [ ] Update WPF WebViewRenderer
- [ ] Move code from deferral folder to appropriate place (I'm using that folder to keep organised in my work, I'm not very used to working in a codebase with so little folders 😖 )
- [ ] Remove mfractor from git ignore (otherwise git tried to commit about 50 files)
- [ ] Cleanup `WebNavigationArgs` (base class or interface):
To get my code working I ended up changing the base class to `DeferrableEventArgs`, I added an interface for the `WebNavigationArgs`, we need to discuss what you would like to see about this. I'd personally prefer the interface.

### PR Checklist ###


Xamarin.Forms.Core/Shell/ShellNavigationManager.cs
```diff
@@ -333,14 +333,10 @@ public static void ApplyQueryAttributes(Element element, IDictionary<string, str
 
 		public void HandleNavigating(ShellNavigatingEventArgs args)
 		{
-			if (!args.DeferredEventArgs)
-			{
-				Navigating?.Invoke(this, args);
-			}
-			else
-			{
+			if (args.DeferralRequested)
```
Axemasta, line 336: i've made this change to make the code more readable

Xamarin.Forms.Core/Xamarin.Forms.Core.csproj
```diff
@@ -22,6 +22,7 @@
 	</ItemGroup>	
 	<ItemGroup>
 	  <Folder Include="Xaml\Diagnostics\" />
+	  <Folder Include="Deferral\" />
```
Axemasta, line 25: Need to remove before a merge

Axemasta:
@PureWeen I've finally got around to putting this PR up with your suggestions. What are your thoughts on how this is implemented?

Since these are technically breaking API changes would this be rejected as part of 5.0.0?

Axemasta:
Will this PR be addressed? How much effort would it be to fork the control & ship it as a seperate control (ie `WebView2`)?

Axemasta:
I went ahead and completely forked this control. I've implemented the deferral token stuff so the navigating event is able to cancel navigation from a background thread! Introducing [SuperWebView](https://github.com/Axemasta/SuperWebView).

I'll keep this PR open because ideally everything I've done in my custom control should come standard with `WebView` 😁

Axemasta:
@jfversluis thanks for the update, I realised quite early into Forms 5 that this PR would never be able to make its way into Xamarin due to the sunsetting of XF & the no api breaks policy in version 5, that is precisely why I forked the control and rolled my "own" implementation of it in my [SuperWebView](https://github.com/Axemasta/SuperWebView) package since I had a project that desperately needed the functionality (it's been in production since for over a year and works flawlessly).

Its definitely something I want to bring into Maui, I made a PR with the changes a bit too early in Maui (Before the WebView handler was implemented). I've been waiting for the right time because for now the Maui team seem very busy getting the .Net 7 GA version out. A little disappointed we couldn't make the change before a GA release and this is the perfect time for Api breaks!

There will be a PR coming soon, my issue and PR on the Maui repo have no interaction from anyone so I've been a little hesitant to put in the work if it goes unnoticed :(

Issue #14211, opened 2021-04-23
[Bug] Xamarin.Forms.Maps - Pin.SendMarkerClick() not implemented


joeweininger:
### Description

Pin.SendMarkerClick() is not implemented. Are there any plans to implement this in future releases?

using System.ComponentModel;
/// <summary>To be added.</summary>
/// <returns>To be added.</returns>
/// <remarks>To be added.</remarks>
[EditorBrowsable (EditorBrowsableState.Never)]
public bool SendMarkerClick ();

### Steps to Reproduce

1. Create a Pin
2. Call the SendMarketClick() method of that pin
3. Nothing happens

### Expected Behavior
Firing the SendMarketClick method should open the pin info window.

### Actual Behavior
nothing happens.

### Basic Information

- Version with issue: Xamarin Forms 5.0.0.2012
- Last known good version: N/A
- Platform Target Frameworks: 
  - iOS:  v12-v14
  - Android: v6-v11
  - UWP:  N/A
- Android Support Library / AndroidX Version: 
- NuGet Packages: Xamarin.Forms.Maps
- Affected Devices: All

### Environment

<details><summary>Show/Hide Visual Studio info</summary>

```
```

</details>

rachelkang:
Hi, @joeweininger - thanks for submitting this issue. `Pin.SendMarkerClick()` is implemented, but it's behavior is different from what you described. It just returns a bool as specified [here](https://docs.microsoft.com/en-us/dotnet/api/xamarin.forms.maps.pin.sendmarkerclick?view=xamarin-forms).

Anyhow, I see you opened a PR! Thank you for your contribution - we'll be sure to take a look :)

jfversluis:
Thanks for this suggestion! As Xamarin.Forms is now in maintenance mode, this will not happen anymore for Xamarin.Forms. We're only adding bugfixes and stability fixes.

If this is still important to you, make sure to check the [.NET MAUI repo](https://github.com/dotnet/maui) and see if it's already on the roadmap. If not, feel free to open a discussion to discuss a change first or open an issue with a detailed feature request. Thanks!

PR #14223, opened 2021-04-29
Add the ability to open the pin info window from code. Added option t…

joeweininger:
…o disable the bottom right Android toolbar.
### Description of Change ###

It is currently not possible to open the pin info window on the Xamarin.Forms.Maps through the code, only when the user clicks it. I added code so if you pass in the pin to the map you can open the pin info window. 

Also for Android only I added the ability to disable the bottom right tooltip window. This window only appears if the user clicks the pin and not if you call the pin open window from code behind.

### Issues Resolved ### 
Not currently created issue for this just the bug I submitted below. 

https://github.com/xamarin/Xamarin.Forms/issues/14211

### API Changes ###
none

Added:
 
To Map.cs in Xamarin.Forms.Maps

public void ShowPinInfoWindow(Pin pin)
public bool HasMapToolbarEnabled

To iOS and Android MapRender in the respective maps project

void OnMapShowPinInfoWindow(Pin pin)

To Android only MapRender

map.UiSettings.MapToolbarEnabled = Map.HasMapToolbarEnabled;

Changed:
 none
 
 Removed:
none
 
 -->
 
 None

### Platforms Affected ### 

- iOS and Android

None, unless you call into the method to view the pin info window or disable the Android toolbar at the bottom right.

Screen Shot 2021-04-29 at 11 48 51 AM

Screen Shot 2021-04-29 at 11 47 08 AM

### Testing Procedure ###
You can test this change by adding Pins to a Xamarin.Forms.Map and then call this function with your pin 

Map.ShowPinInfoWindow(yourPin);

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Maps.Android/MapRenderer.cs
+++ b/Xamarin.Forms.Maps.Android/MapRenderer.cs
@@ -26,6 +26,8 @@ namespace Xamarin.Forms.Maps.Android
 	{
 		const string MoveMessageName = "MapMoveToRegion";
 
+		const string MapShowPinInfoWindowMessageName = "MapShowPinInfoWindow";
+
 		static Bundle s_bundle;
 
 		bool _disposed;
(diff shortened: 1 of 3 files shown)


jfversluis:
hey @joeweininger thank you so much for this wonderful PR and functionality and it's your first PR with us as well. That makes me even more sorry that we did not respond to this at all. I could go on and explain all the reasons why, it won't matter much. The only thing I can say is we're striving to do better with .NET MAUI, I'm making it my personal goal.

As for this one. Since Xamarin.Forms is now in maintenance mode, I can't really add new APIs anymore. That means I won't be able to take this one in. However, you should be able to transition this easily into .NET MAUI I think it this is something that still interests you. Personally, I think this is a lof of people would want so if you're still up for it I would encourage you to open the feature request on the .NET MAUI repo and let's implement it there. When you decide to do so, please ping me so I can make sure we do better this time.

Thanks for understanding!

PR #14166, opened 2021-04-16
[Gtk] Implement RadioButton

JunielKatarn:
### Description of Change ###

Adds missing implementation of RadioButton for Gtk.

### Issues Resolved ### 

None

### API Changes ###

Added:
 - class Xamarin.Forms.Platform.GTK.Controls.RadioButton
 - Xamarin.Forms.Platform.GTK.Renderers.RadioButtonRenderer

### Platforms Affected ### 

- Gtk

### Behavioral/Visual Changes ###

RadioButtons under Gtk get rendered, shown, and behave consistently with other platforms.

### Before/After Screenshots ### 

Not applicable

### Testing Procedure ###

1. Start ControlGallery.Gtk.
2. Open the RadioButtonGroup page.
3. Verify RadioButtons show up and grouping behavior works.

### PR Checklist ###

- [ ] Targets the correct branch
- [ ] Tests are passing (or failures are unrelated)

PR #14235, opened 2021-05-02
[GTK] Some Shapes API support

knocte:
### Description of Change ###

Supported Shapes:
  - Ellipse
  - Rectangle
  - Path

Supported Path Segments:
  - Arc
  - Line

Supported Brushes:
  - SolidColorBrush

(Strokes implementation will come in a 2nd PR.)

Co-authored-by: Afshin Arani &lt;afshinarani79@gmail.com&gt;

### API Changes ###
NOTE: even if this could be considered an enhancement, any code that used Shapes with XamarinForms5.0 (or older) running on Gtk would have crashed or contained bugs (obviously), so this is essentially a bugfix. So: None, no API changes.

### Platforms Affected ### 
- GTK

### Testing Procedure ###
Manual with our own app.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.GTK/Controls/Shapes/EllipseView.cs
+++ b/Xamarin.Forms.Platform.GTK/Controls/Shapes/EllipseView.cs
@@ -1,0 +1,21 @@
+﻿using System;
+using Cairo;
+using Xamarin.Forms.Platform.GTK.Extensions;
+
+namespace Xamarin.Forms.Platform.GTK.Controls
+{
+	public class EllipseView : ShapeView
+	{
+		protected override void Draw(Gdk.Rectangle area, Context cr)
+		{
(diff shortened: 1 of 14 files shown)


jsuarezruiz:
@knocte Looks nice, thanks for the contribution. Do you think could be possible to include Polyline and Polygon to have a PR including support to all the basic shapes at the same time?

knocte:
> Do you think could be possible to include Polyline and Polygon to have a PR including support to all the basic shapes at the same time?

@jsuarezruiz we have just pushed a 2nd commit that implements this.

jfversluis:
Same as #14166 

Actually not adding new stuff to Forms, but since this is community backed and not adding to our maintenance load, let's goooo 🚀

Issue #14510, opened 2021-08-15
[Bug] Xamarin.Forms.Platform.iOS.EntryRendererBase`1[TControl].GetClearButtonTintImage crash


AndreasReitberger:
### Description
Using iOS 13 on an iPhone 8 simulator leads to following crash while using entry box with ClearButtonVisibility to set "WhileEditing".

```xaml
<Entry 
      Grid.Column="1"
      Placeholder="{x:Static localization:Strings.SettingsPrintServerConnectionAddress}"
      Text="{Binding ServerAddress, Mode=TwoWay}" 
      Style="{StaticResource DefaultBorderlessEntry}"
      ClearButtonVisibility="WhileEditing"
      Margin="16,2,16,4"
      />
```
Newer iOS versions are working fine.

### Steps to Reproduce

1. Setup an iPhone 8 simulator running iOS 13.1
2. Insert the xaml code above
3. See the exception

### Expected Behavior
No exception.

### Actual Behavior
Exception gets thrown.
`	exc	{System.NullReferenceException: Object reference not set to an instance of an object   at Xamarin.Forms.Platform.iOS.EntryRendererBase`1[TControl].GetClearButtonTintImage (UIKit.UIImage image, UIKit.UIColor color) [0x00000] in D:\a\1\s\Xamarin.Forms.Platform.iO…}	
`
### Basic Information

- Version with issue: SR4
- Platform Target Frameworks: 
  - iOS:  iOS 13.1 on iPhone 8 (Simulator and real device)
- Affected Devices: iPhone 8 (Simulator / real device)

### Stacktrace

Xamarin.Forms.Platform.iOS.EntryRendererBase`1[TControl].GetClearButtonTintImage (UIKit.UIImage image, UIKit.UIColor color) [0x00000] in D:\\a\\1\\s\\Xamarin.Forms.Platform.iOS\\Renderers\\EntryRenderer.cs:591 \n at Xamarin.Forms.Platform.iOS.EntryRendererBase`…"

AndreasReitberger:
Seems to be related to: https://github.com/xamarin/Xamarin.Forms/issues/14479

jfversluis:
Fixed with #14479

Issue #14479, opened 2021-07-31
[Bug]  Xamarin.forms iOS Latest Stable Version Crash: System.NullReferenceException: Object reference not set to an instance of an object   at Xamarin.Forms.Platform.iOS.EntryRendererBase`1[TControl].GetClearButtonTintImage


pasha-o:
### Description
**Visual studio 2019 16.10.4, updated Xamain.forms to 5.0.0.2083. it builds on iOS as Android, however, it crashes on iOS phone SE version 12.5.4**

2021-07-29 20:34:26.249 [].iOS[24513:823966] System.NullReferenceException: Object reference not set to an instance of an object
  at Xamarin.Forms.Platform.iOS.EntryRendererBase`1[TControl].GetClearButtonTintImage (UIKit.UIImage image, UIKit.UIColor color) [0x00000] in <4c65493b0ca44a809e1174d1a41683bd>:0 
  at Xamarin.Forms.Platform.iOS.EntryRendererBase`1[TControl].UpdateClearButtonColor () [0x0009d] in <4c65493b0ca44a809e1174d1a41683bd>:0 
  at Xamarin.Forms.Platform.iOS.EntryRendererBase`1[TControl].UpdateClearButtonVisibility () [0x00020] in <4c65493b0ca44a809e1174d1a41683bd>:0 
  at Xamarin.Forms.Platform.iOS.EntryRendererBase`1[TControl].OnElementChanged (Xamarin.Forms.Platform.iOS.ElementChangedEventArgs`1[TElement] e) [0x001a5] in <4c65493b0ca44a809e1174d1a41683bd>:0 
  at SpotLee.iOS.Renderers.CustomEntryRenderer.OnElementChanged (Xamarin.Forms.Platform.iOS.ElementChangedEventArgs`1[TElement] e) [0x00001] in C:\Dev\Repos\SensoDrive.Mobile\SpotLee\SpotLee.iOS\Renderers\CustomEntryRenderer.cs:14
 
  at Xamarin.Forms.Platform.iOS.VisualElementRenderer`1[TElement].SetElement (TElement element) [0x0017a] in <4c65493b0ca44a809e1174d1a41683bd>:0 
[...]

PR #14526, opened 2021-08-23
Fix Entry crashing on iOS < 14 

rotorgames:
### Description of Change ###

Fixing crashes on iOS < 14 if ClearButtonVisibility of Entry is WhileEditing

### Issues Resolved ### 

Fixes:
- #14479 
- #14510

### API Changes ###

 
 None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###

- Use iOS simulator with iOS 13 or lower
- Set ClearButtonVisibility of Entry to WhileEditing
- The app must not crash when a page is opening

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/EntryRenderer.cs
@@ -50,6 +50,8 @@ namespace Xamarin.Forms.Platform.iOS
 
 		UIImage _defaultClearImage;
 
+		UIButton ClearButton => Control?.ValueForKey(new NSString("clearButton")) as UIButton;
+
 		public EntryRendererBase()
 		{
 		}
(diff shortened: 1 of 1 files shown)


rotorgames:
@jfversluis Re-open this, please. Explanations: https://github.com/xamarin/Xamarin.Forms/pull/14566#issuecomment-924098252

jfversluis:
Reopening this to make sure we have another look

VitalyKnyazev:
I am getting new exception on iOS 12.4 after updating to SR6:
Name: NSRangeException Reason: Cannot remove an observer ... for the key path "sublayers" from <CALayer 0x600007b11e80> because it is not registered as an observer.

It happens on EntryRenderer Dispose call at this line:
ClearButton?.Layer?.RemoveObserver(this, new NSString("sublayers"));

xleon:
I'm getting this crash too. 
It doesn't matter what `ClearButtonVisibility` is set to. It crashes anyway when disposing (normally when navigating away from the page where the Entry is). 
As a workaround I'm overriding  `Dispose()` method with a try/catch in my custom `EntryRenderer`:

```csharp
{
    try
    {
        base.Dispose(disposing);
    }
    catch (Exception)
    {
        // TODO remove this hack on newer version of Xamarin.Forms: https://github.com/xamarin/Xamarin.Forms/pull/14526
    }
}
```

PR #14648, opened 2021-09-23
[Core] Update Parent of NavPage childrens doing PopModalAsync

### LINKED ISSUE #12703: [Bug] OnParentSet null is not called when popping modal stack
It is common to have a pattern such as
```c#
        protected override void OnParentSet()
        {
            base.OnParentSet();

            if (this.Parent == null)
            {
                // clean up here
            }
        }
```
to clean up stuff in your ContentPages. This works in a modal stack and a normal navigation stack, ie the system nulls parent and OnParentSet is called when a page is navigated away form.

However, if an entire modal stack is popped via `Navigation.PopModalAsync()`, OnParentSet null is not invoked on the pages. Expected behaviour is that OnParentSet null should be invoked in a reverse order in which they were pushed to the stack. My workaround for the time being is the following which accomplishes this:

```c#
            // a popped modal stack doesn't set Parent to null unless done manually
            this.ModalPopped += (object sender, ModalPoppedEventArgs e) =>
            {
                var oldStack = e.Modal.Navigation.NavigationStack.Reverse().ToList();
                foreach (var oldPage in oldStack)
                    oldPage.Parent = null;
            };
```

Thanks.

jsuarezruiz:
### Description of Change ###

Update Parent of NavPage childrens doing PopModalAsync.

### Issues Resolved ### 

- fixes #12703

### API Changes ###
 
 None

### Platforms Affected ### 

- Core/XAML (all platforms)

None

Not applicable

### Testing Procedure ###
Can use Issue 1230. If OnParentSet method is invoked and Parent is null, the test has passed. 

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue12300.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue12300.cs
@@ -7,6 +7,7 @@ using Xamarin.Forms.PlatformConfiguration;
 using Xamarin.Forms.PlatformConfiguration.iOSSpecific;
 using NavigationPage = Xamarin.Forms.NavigationPage;
 using Page = Xamarin.Forms.Page;
+using System.Diagnostics;
 
 #if UITEST
 using Xamarin.UITest;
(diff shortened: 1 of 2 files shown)


jfversluis:
I think at this point it's safe to say this won't be happening for Xamarin.Forms anymore. Let's make it happen for .NET MAUI!

Issue #14657, opened 2021-09-24
[Bug] Is XF Shell navigation leaking memory?


danielkraut:
### Description

Every time app adds a page to navigation stack via `Shell.Current.GoToAsync("MyPage")` a new instance is created, which is to be expected. But after navigating back (either with `Shell.Current.GoToAsync("..")` or with back button) that page is never released from memory or at least is seems that it is the case. I have tried to check it with a finalizer but it is never called, which may be an implementation detail but having also some private object created then its finalizer is also never called and that seems like a bug. I also tried to use Xamarin Profiler which also shows that alive objects are only growing.

When using the older navigation style - `PushAsync, PopAsync` - finalizers are called properly. I will attach a minimal example app to demonstrate the issue which is just the project autogenerated by Visual Studio stripped down to just 2 pages.

If this behavior is expected I would like to get a clarification how XF.Shell manages memory. Right now I am dealing with my own application which after some time only increases its memory usage and never releases it and I am wondering if XF.Shell may be responsible for that. This same issue have already been discussed here https://github.com/xamarin/Xamarin.Forms/issues/11476 but I don't think it was resolved in any way as the responders focused on Messagin centre but I am not using that.

### Steps to Reproduce

1. Create Xamarin Forms Shell application
2. Create 2 pages - one home page and the other registered with `Routing.RegisterRoute`
3. Add button to navigate from main page to other page and back
4. Add private field to other page which contain object of some custom class
5. Add finalizer to other page and custom class, put breakpoint or writing to console to the finalizer
6. Navigate back nad forth between main page and other page
7. Observer that finalizers are never called

### Expected Behavior

Allocated data is released when a page is removed from navigation stack.

### Actual Behavior

Finalizer of the removed page or finalizer of any object the page has reference to is never called.

### Basic Information

- Version with issue: 5.0.0-sr3, 5.0.0-sr4
- Last known good version: unknown
- Platform Target Frameworks: 
  - Android: 10, 11 (haven't tried other OS/platforms)
- NuGet Packages: none
- Affected Devices: all

### Environment

Visual Studio: Visual Studio 2019 16.9.3, 16.11.2

### Screenshots

![obrazek](https://user-images.githubusercontent.com/33424430/134741667-619caa40-1f37-4a58-9e4f-7be490ffc6b7.png)

### Reproduction Link

Minimal example:
[...]

PR #14717, opened 2021-10-11
[Android] Fix content pages not Garbage collected

danielkraut:
### Description of Change ###

Changes I am certail of that fixes (at least sometimes, read next paragraph) content page instances not garbage collected:
* `ShellToolbarTracker` - calling `RemoveDrawerListener(...)` in dispose method
    * after adding this the simplest content page with no content were garbage collected
* `ShellContentFragment` - explicitly setting page TitleView to `null`
    * this change allows pages to be collected when a page has set custom content via `Shell.TitleView` (you need to add at least some layout with content, e.g. `StackLayout` with `Label`)
    * This is my first time browsing throught Xamarin.Forms codebase and I have no idea if this is the best place to set title to null, I would appreciate other devs feedback but it certainly fixed my issue

Those changes allow _simple_ pages to be collected by GC. By _simple_ I mean pages with default controls (buttons, labels, etc.) and with custom TitleView, but in my application I have used third-party controls (Syncfusion) and those pages are still not collected. I am certain it is a Shell issue because using those controls in non-Shell application seems fine.

Other changes don't necessarily affect my issue but are something I noticed when debugging Xamarin Forms. Calling `_shellPageContainer.Dispose();` is must-have in my opinion but calling RemoveView/RemoveAllViews is something I need more feedback from other devs if it is necessary or not.

### Issues Resolved ### 

#14657 - It doesn't fix the issue completely, more investigation is still required.

### API Changes ###

None

### Platforms Affected ### 

- Android

Applications should use less memory because _simple_ page instances will be garbage collected.

Not applicable

### Testing Procedure ###

I was able to test only Android platform (Android 10 and 11).

In generall, just implement a finalizer on any page that is created and destroyed on the fly, i.e. a page when navigated to, a new instance is created unlike a pages which are added to flyout menu, those are usually have single instance for the lifetime of application. That finalizer is never called.

Also I have created automatic test to demonstrated the issue. Callign 

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue14657.cs
+++ b/Xamarin.Forms.Controls.Issues/Xamarin.Forms.Controls.Issues.Shared/Issue14657.cs
@@ -1,0 +1,136 @@
+﻿using Xamarin.Forms.CustomAttributes;
+using Xamarin.Forms.Internals;
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+#if UITEST
+using Xamarin.Forms.Core.UITests;
+using Xamarin.UITest;
+using NUnit.Framework;
+#endif
+
+namespace Xamarin.Forms.Controls.Issues
+{
+#if UITEST
+	[Category(UITestCategories.Github10000)]
+	[Category(UITestCategories.Shell)]
+	[Category(UITestCategories.TitleView)]
+#endif
+	[Preserve(AllMembers = true)]
+	[Issue(IssueTracker.Github, 14657, "Shell pages are not released from memory", PlatformAffected.Android)]
+	public class Issue14657 : TestShell 
+	{
+		static int pageCount = 0;
+		protected override void Init()
+		{
+			Routing.RegisterRoute(nameof(Issue14657_ChildPage), typeof(Issue14657_ChildPage));
+			
+			var rootPage = CreateRootPage();
+
+			AddContentPage(rootPage);
+		}
+
+		ContentPage CreateRootPage()
+		{
+			var rootPage = CreateContentPage("Home page");
+			rootPage.Content = new StackLayout()
+			{
+				Children =
+				{
+					new Button()
+					{
+						Command = new Command(CollectMemory),
+						Text = "Force GC",
+						AutomationId = "GC_14657"
+					},
+					new Button()
+					{
+						Command = new Command(async () => await GoToChild()),
+						Text = "Go to child page",
+						AutomationId = "GoToChild_14657"
+					}
+				}
+			};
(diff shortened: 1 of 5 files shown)


jfversluis:
Hey @danielkraut! I just triggered the pipeline :) I did read your comments and had a first quick look through your code. It seems to make sense to me so far. Thank you for this!

jfversluis:
Hey @danielkraut thanks for your patience on this one! I've been looking at it for a bit and seems all OK to me. If you have been digging into the more complex scenarios, feel free to do a follow up PR. Thanks for all your time and efforts on this one!

jfversluis (approved):

danielkraut:
> Hey @danielkraut thanks for your patience on this one! I've been looking at it for a bit and seems all OK to me. If you have been digging into the more complex scenarios, feel free to do a follow up PR. Thanks for all your time and efforts on this one!

Hi @jfversluis, yes I have been doing much more and it has been time consuming. I could not wait for this to be merged and actually did not most of the Shell advantages so in the mean time I have rewritten our application without the Shell. I have found that also not all pages are being GC'd. I though I would provide another PR but have found this - once you set `NavigationPage`/`Shell`.`TitleView` to any "complex" view (`StackLayout`, `Grid`, etc.) your pages will not be GC'd in **Debug build**. In Release build they eventually will be, but you have to make sure to do a clean and then build (I also manually removed the `obj` folder and uninstalled the application from target device just to be sure, this method worked) every time, otherwise this might not work. That process is quite slow as you might imagine. Our CI creates new build when we deploy our app so this is not a real problem in production but it is quite painful to debug. Also this problem starts from XF 5.0.0.2021 so my guess is it is related to XAML hot reload. I don't know how would I troubleshoot hot reload issues so I eventually gave up.

Summary - `ContentPage`s are not garbage collected if you:

* Use Xamarin Forms 5.0.0.2012 or higher, and
* use Debug build, and
* set `TitleView` to any `Layout` or other complex `View`

In Release build they should be but make sure you create clean build to be sure.

@jfversluis Since we don't use XF Shell any more I will not test changes in this PR in production but I think that you can close issue #14657, if GC releated issues are okay in debug build.

jfversluis:
Thank you so much for your detailed explanation and investigation Daniel, very much appreciated! I hope you will stick with us for .NET MAUI and continue to be a contributor this way :)

sahurgithub:
I'm not sure if anyone else is having this issue but the change on [line 173](https://github.com/xamarin/Xamarin.Forms/blob/5.0.0/Xamarin.Forms.Platform.Android/Renderers/ShellContentFragment.cs#L173) seems to be removing the TitleView from the Shell. This gets triggered when switching tabs as well and when the user goes back to a tab which has a TitleView, its now gone because the code set it to null.

PR #14971, opened 2021-12-09
added missing iOS constructor in ListViewRenderer

narciszait:
### Description of Change ###

Added a missing constructor in ListViewRenderer

### Issues Resolved ### 
Could not find it. But I saw it in the app crash logs:

ObjCRuntime.RuntimeException

Failed to marshal the Objective-C object 0x28382b600 (type: Xamarin_Forms_Platform_iOS_ListViewRenderer_ListViewDataSource). Could not find an existing managed instance for this object, nor was it possible to create a new managed instance (because the type 'Xamarin.Forms.Platform.iOS.ListViewRenderer+ListViewDataSource' does not have a constructor that takes one IntPtr argument).

### API Changes ###
Added:
- public ListViewRenderer(IntPtr handle)

### Platforms Affected ### 
- iOS

By having that missing constructor, we avoid that exception

### Testing Procedure ###
I saw it in our crash logs.  
Don't know necessarily how to reproduce it.

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/ListViewRenderer.cs
@@ -53,6 +53,12 @@ namespace Xamarin.Forms.Platform.iOS
 
 		}
 
+		[Internals.Preserve(Conditional = true)]
+		public ListViewRenderer(IntPtr handle)
+		{
+
+		}
+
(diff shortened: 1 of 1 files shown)


jsuarezruiz (commented):
Could you share a more detailed logs to try to know when this happens?. Thanks for the PR and the feedback.

jfversluis (approved):
Not sure it will fix the issue but I guess it doesn't hurt as well :)

PR #14982, opened 2021-12-14
[iOS] Fix crash on iOS LabelRenderer UpdateTextColor

fedemkr:
### Description of Change ###

When the label is `TextType=Html` and the html is big, when you send the app to background change the theme, go back then the app and close the current page it has high changes to crash because the `Element` is `null`. So this fixes it by checking that before updating the color.

### Issues Resolved ### 

- I couldn't find an issue to link to this, should I create it?

### API Changes ###

 
 None

### Platforms Affected ### 

- iOS

None

Not applicable

### Testing Procedure ###

- Create a view that can be open and close from another part
- On that view set a Label with TextType HTML and bind a large HTML (In my particular case, I had one that was like 150 chars styled with different colors depending on the char type, i.e. number, special, alphabet, etc)
- Open the app and open that view
- Put the app on background
- Change the light/dark theme
- Open the app again and without leaving the HTML to load properly (sometimes it takes time to apply the correct format) close the view (back button or in-app cancel/close view)
- The app crashes

### PR Checklist ###

### DIFF
--- a/Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
+++ b/Xamarin.Forms.Platform.iOS/Renderers/LabelRenderer.cs
@@ -561,6 +561,9 @@ namespace Xamarin.Forms.Platform.MacOS
 
 		void UpdateTextColor()
 		{
+			if (IsElementOrControlEmpty)
+				return;
+				
 			if (IsTextFormatted)
 			{
 				UpdateFormattedText();


jfversluis:
Hey @fedemkr thanks so much for your contribution! Don't worry about the issue, I will check it out from the info you gave here :)

jfversluis (approved):

jfversluis:
@fedemkr there is a NuGet now available for this, would you be able to grab the NuGet [as described here](https://github.com/xamarin/Xamarin.Forms/wiki/Preview-Packages) and me know if this actually fixes the issue you meant to fix?

jfversluis:
Going to merge this for now as I don't see any way this could cause too much trouble. If you still get to test it, please let me know the results!

fedemkr:
Just tested the Nuget PR preview and I can confirm that the fix works. Thanks @jfversluis 😄

